PR #1836, opened 2022-04-18
Revamp format heading

harish-sethuraman:
Revamp format heading to use the same method with diff params instead of having separate methods that do the same thing


zurfyx (approved):
Neat, thank  you! ✨

PR #1837, opened 2022-04-18
Adds disabled state css for action items

harish-sethuraman:
Although the button was disabled it was kinda not very different from other buttons. So this might actually be better
Before: 
Screenshot 2022-04-18 at 1 38 36 PM

After: 
Screenshot 2022-04-18 at 1 38 53 PM


zurfyx (approved):
LGTM!

PR #1838, opened 2022-04-18
Add prettier, eslint and flow checks to workflows

harish-sethuraman:
I once sent a PR to prettify all files and still this is not fixed and came back with some other's commit 
So I have added prettier flow and eslint checks in the workflow to better check if all files are prettified or not so we can catch here if it is force pushed. 
- https://github.com/facebook/lexical/pull/1754

Changes:
- Removed `pretty-quick` (IDK why we might actually need this. If we want this let me know)
- Added `npm-run-all` (we shall later use it run multiple tests-e2e in parallel or serial)
- Introduced new job to run prettifier, eslint and flow check

Commits:
1. Change config files 
2. prettify all files


.github/workflows/test.yml
```diff
@@ -16,6 +16,31 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  lint:
```
zurfyx, line 19: Isn't this job very similar to `integrity`? It's just adding prettier? Should they be combined?

zurfyx (commented):
Thank you

.github/workflows/test.yml
harish-sethuraman: > Isn't this job very similar to integrity? It's just adding prettier? Should they be combined?

Yes it is very similar. Sorry just noticed. I've used the `integrity` check instead. Thanks :)

zurfyx (approved):

harish-sethuraman:
Any idea on why the workflows didn't run?

zurfyx:
> Any idea on why the workflows didn't run?

Seems like they just queued

Issue #1811, opened 2022-04-17
Excalidraw Discard button should show a close modal


zurfyx:
_Feedback from [flibben](https://www.reddit.com/user/flibben/) @ Reddit_

Similar to #1794

pabloasc:
HI, can I work in this one?

thegreatercurve:
Closing as I added this in a recent PR.

PR #1847, opened 2022-04-18
Open a confirm modal on discard excalidraw

pabloasc:
Addresses #1811 

image

Modal does not appear in case you didin't make any change in excalidraw, it appears only in case there is something created with this tool.


zurfyx (approved):
LGTM, thank you!

PR #1848, opened 2022-04-18
setEditorState ignoreSelection flag

zurfyx:
I can see that this is potential use for setEditorState, selection is something that we don't want to replicate in a few cases, so I'm proposing to build this in ~~as an optional flag on `setEditorState`~~:

1. Restore from server
2. Side-by-side render view (i.e. https://github.com/facebook/lexical/discussions/1825)

Update: I'm proposing to build this as an optional flag for Rich/PlainTextPlugin. To do this we leverage batched updates. Why? Because `setEditorState` should ideally return the same EditorState that you have passed it.

In any case, I believe this is a temporary approach, because, as soon as we move `setEditorState` to be part of an update we can drop the selection easily with a $ function. Currently, it's complicated because an EditorState is immutable

### DIFF
--- a/packages/lexical-react/src/__tests__/unit/PlainRichTextPlugin.test.js
+++ b/packages/lexical-react/src/__tests__/unit/PlainRichTextPlugin.test.js
@@ -16,7 +16,12 @@ import LexicalContentEditable from '@lexical/react/LexicalContentEditable';
 import {HeadingNode, QuoteNode} from '@lexical/rich-text';
 import {TableCellNode, TableNode, TableRowNode} from '@lexical/table';
 import {$rootTextContentCurry} from '@lexical/text';
-import {$createParagraphNode, $createTextNode, $getRoot} from 'lexical';
+import {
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/shared/PlainRichTextUtils.js
```diff
@@ -9,20 +9,25 @@
 
 import type {EditorState, LexicalEditor} from 'lexical';
 
-import {$createParagraphNode, $getRoot, $getSelection} from 'lexical';
+import {
```
fantactuka, line 12: Not really related to the PR, but do we still use this utils file?

fantactuka (approved):

trueadm:
I think this might be somewhat over-engineered. Rather than add all this code into core, couldn't someone just do:

`const editorStateWithoutSelection = editorState.clone(null)`

zurfyx:
Agree with @trueadm, just keeping the unit test for EditorState then

PR #1851, opened 2022-04-18
feat: Make the SVG result resizable

### LINKED ISSUE #1813: Excalidraw - Make the SVG result resizable
_Feedback from [flibben](https://www.reddit.com/user/flibben/) @ Reddit_

We can improve Excalidraw's UX by making it resizable so that the user can easily pick their final size. Given Excalidraw provides a SVG for us this makes this issue a lot easier/doable

Whoever takes this issue may want to get some inspiration from the ImagePlugin/Node implementation

100paperkite:
trying to resolve #1813

I'm not sure if the image caption should be included in SVG result.. Should I include it?
Also, Is it okay with reusing the ImageResizer component?


packages/lexical-playground/src/nodes/ImageNode.jsx
```diff
@@ -98,7 +98,7 @@ function LazyImage({
   );
 }
 
-function ImageResizer({
+export function ImageResizer({
```
zurfyx, line 101: Can we move this to a separate file since it's now reused by two separate modules? Maybe into `packages/lexical-playground/src/ui`

zurfyx (approved):
Beautiful! Thanks

packages/lexical-playground/src/nodes/ImageNode.jsx
100paperkite: Of course. Thanks!

BjoernRave:
how easy do you think would it be to also make it "resizable" i.e. being able to drag only one side

zurfyx:
Thanks! I think @BjoernRave's comment on making the resize handles more intuitive makes sense but we can iterate on this in future PRs. Want to create an issue for this?

trueadm:
Did the CI run on this PR?

PR #1855, opened 2022-04-18
remove unwanted cross-env in package.json

harish-sethuraman:
There are way too many `cross-env` in the `package.json` that I think is unwanted. 

https://github.com/kentcdodds/cross-env#usage

```
You can set multiple environment variables at a time:

{
  "scripts": {
    "build": "cross-env FIRST_ENV=one SECOND_ENV=two node ./my-program"
  }
}
```


zurfyx (approved):
Nice!

PR #1856, opened 2022-04-18
Fix collab selections possibly doubling up when spanning text nodes

zwily:
I noticed while playing with collaboration that if a user highlighted multiple adjacent text blocks, then the remote user would see blocks in the middle double-highlighted. (See screenshot). This turned out to be because the DOM function `getClientRects()` can return multiple identical rects for some elements. The fix is just to ignore duplicate rects.

Screen Shot 2022-04-18 at 10 07 08 AM

### DIFF
--- a/packages/lexical-yjs/src/SyncCursors.js
+++ b/packages/lexical-yjs/src/SyncCursors.js
@@ -296,7 +296,6 @@ function updateCursor(
         prevRect.top === selectionRect.top &&
         prevRect.left === selectionRect.left &&
         prevRect.width === selectionRect.width &&
-        prevRect.height === selectionRect.height &&
         prevRect.height === selectionRect.height
       ) {
         selectionRects.splice(i--, 1);
--- a/packages/lexical-yjs/src/SyncCursors.js
+++ b/packages/lexical-yjs/src/SyncCursors.js
@@ -283,6 +283,8 @@ function updateCursor(
   let selectionRectsLength = selectionRects.length;
   const selectionsLength = selections.length;
 
+  let prevRect;
+
   for (let i = 0; i < selectionRectsLength; i++) {
     const selectionRect = selectionRects[i];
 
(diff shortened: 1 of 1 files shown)


packages/lexical-yjs/src/SyncCursors.js
```diff
@@ -285,6 +285,25 @@ function updateCursor(
...
+    // merge all the rects together into a set of rects that don't overlap, so we don't
+    // generate backgrounds that are too dark.
+    if (i > 0) {
+      const prevRect = selectionRects[i - 1];
+      if (
+        prevRect.top === selectionRect.top &&
+        prevRect.left === selectionRect.left &&
+        prevRect.width === selectionRect.width &&
+        prevRect.height === selectionRect.height
+      ) {
```
zurfyx, line 300: We can merge with the condition below?

packages/lexical-yjs/src/SyncCursors.js
```diff
@@ -285,6 +285,25 @@ function updateCursor(
 
   for (let i = 0; i < selectionRectsLength; i++) {
     const selectionRect = selectionRects[i];
+
+    // Exclude a rect that is the exact same as the last rect. getClientRects() can return
+    // the same rect twice for some elements. A more sophisticated thing to do here is to
+    // merge all the rects together into a set of rects that don't overlap, so we don't
+    // generate backgrounds that are too dark.
+    if (i > 0) {
+      const prevRect = selectionRects[i - 1];
```
zurfyx, line 294: Instead of accessing the previous elements, we can instead just keep track of the variable outside the loop and would not also simplify the conditionals but also be more efficient

zurfyx (approved):

zurfyx:
Makes sense to me, thanks!

zwily:
> Makes sense to me, thanks! Also do you mind adding a summary with a more detailed description on the issue and why this is a problem? Or link an issue?
> 
> EDIT: Ah, just found the convo on Discord, mind copy-pasting that here for historical reference?

No problem - it's just one message and a couple screenshots:

> Something interesting I noted while playing with the collab cursor highlighting... If you highlight spans 2 text nodes it works fine, but once it spans 3 nodes, the middle node gets double highlighting:

Screen Shot 2022-04-18 at 10 06 54 AM
Screen Shot 2022-04-18 at 10 07 08 AM

zurfyx:
Thanks for iterating on this!

packages/lexical-yjs/src/SyncCursors.js
```diff
@@ -283,14 +283,34 @@ function updateCursor(
...
   for (let i = 0; i < selectionRectsLength; i++) {
     const selectionRect = selectionRects[i];
-    if (selectionRect.width + rootPadding === rootRect.width) {
-      // Exclude selections that span the entire element
+
+    // Exclude a rect that is the exact same as the last rect. getClientRects() can return
+    // the same rect twice for some elements. A more sophisticated thing to do here is to
+    // merge all the rects together into a set of rects that don't overlap, so we don't
+    // generate backgrounds that are too dark.
+    const isDuplicateRect =
```
zwily, line 295: @zurfyx Does this look better?

Issue #1536, opened 2022-03-23
Flatten the testing matrix


acywatson:
We're running into concurrency limits on macOS in the CI and it's also just wasting compute. We should flatten the testing matrix and run the various tests suites serially instead of in parallel on separate machines. This will save the cost of repeatedly doing all the npm setup.

harish-sethuraman:
I tried to do so but seems like it also takes quite a lot of time. Most of the time spent is on running the e2e test and not on doing npm setup since we already have cached deps. You can take a look at #1858 in case that is what you mean in this issue

acywatson:
Thanks for the time you put into this - I think, rather than overall time, what we were going for here was fewer separate machines used to keep us away from the concurrency limits. If it's going to take a lot longe to execute them, we might want to just keep doing it the way we're doing it. As you said, everything is cached, so the setup overhead is relatively low.

Let's just abandon this for now. Thanks again for the effort!

PR #1858, opened 2022-04-18
Flattened the test matrix

harish-sethuraman:
May be? This should reduce CI time since we don't install deps again and again for other browsers and jobs should be less as well.

Before: Tests take 16m 20s to complete
- https://github.com/facebook/lexical/issues/1536


acywatson:
As discussed in the issue, we're going to table this for now, since it seems like the changes cause some serious overall execution speed issues.

PR #1862, opened 2022-04-19
Fix alignment issue with root level DecoratorNodes

### LINKED ISSUE #1696: Bug: We can't horizontally align a tweet
Lexical version: 0.2.0

## Steps To Reproduce

1. Go to Lexical Playground
2. Click on Insert and then click on Tweet
3. Insert a link to a tweet(if you want to go fast: https://twitter.com/trueadm/status/1514298427840024581) and then click on Confirm
4. Wait for loading the tweet properly on editor's area and then click on the right ride of the tweet element and try to align horizontally by clicking on Align > Center Align

You will see an error overlay with the following message:

Screenshot_2022-04-13_21-04-02

`Uncaught Error: Expected node root to have closest block element node.`

## The current behavior

We can't horinzontally align a tweet

## The expected behavior

We can horizontally align a tweet or at least have no errors being thrown

thegreatercurve:
Should fix #1696.

### DIFF
--- a/packages/lexical-playground/src/nodes/YouTubeNode.jsx
+++ b/packages/lexical-playground/src/nodes/YouTubeNode.jsx
@@ -62,7 +62,7 @@ export class YouTubeNode extends DecoratorNode<React$Node> {
   }
 
   createDOM(): HTMLElement {
-    return document.createElement('div');
+    return document.createElement('span');
   }
 
   updateDOM(): false {
--- a/packages/lexical-playground/src/nodes/YouTubeNode.jsx
+++ b/packages/lexical-playground/src/nodes/YouTubeNode.jsx
@@ -62,7 +62,9 @@ export class YouTubeNode extends DecoratorNode<React$Node> {
   }
 
   createDOM(): HTMLElement {
-    return document.createElement('span');
+    const elem = document.createElement('div');
+    elem.style.display = 'inline-block';
+    return elem;
   }
 
   updateDOM(): false {
--- a/packages/lexical-playground/src/plugins/TwitterPlugin.js
+++ b/packages/lexical-playground/src/plugins/TwitterPlugin.js
@@ -10,15 +10,19 @@
 import type {LexicalCommand} from 'lexical';
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {$findMatchingParent, mergeRegister} from '@lexical/utils';
 import {
+  $createParagraphNode,
   $getSelection,
   $isRangeSelection,
-  $isRootNode,
   COMMAND_PRIORITY_EDITOR,
+  COMMAND_PRIORITY_HIGH,
   createCommand,
+  INSERT_TEXT_COMMAND,
 } from 'lexical';
 import {useEffect} from 'react';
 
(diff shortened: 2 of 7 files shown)


packages/lexical-playground/src/plugins/TwitterPlugin.js
```diff
@@ -31,26 +35,42 @@ export default function TwitterPlugin(): React$Node {
...
           }
-        }
-        return true;
-      },
-      COMMAND_PRIORITY_EDITOR,
+          const focusNode = selection.focus.getNode();
+          const parentEmbedNode = $findMatchingParent(focusNode, (node) =>
+            $isEmbedNode(node),
+          );
+          return parentEmbedNode !== null;
```
fantactuka, line 70: Is it the way to prevent content within embed node? Copy pasting can do that too, so I wonder if having transform listener on embed node would do better

trueadm:
Rather than creating an EmbedNode, couldn't the DecoratorNode just listen to the align command and react accordingly?

packages/lexical-playground/src/plugins/ToolbarPlugin.jsx
```diff
@@ -521,7 +523,9 @@ function InsertYouTubeDialog({
   activeEditor: LexicalEditor,
   onClose: () => void,
 }): React$Node {
-  const [text, setText] = useState('');
+  const [text, setText] = useState(
+    'https://www.youtube.com/watch?v=zKUpf1Vx0vs',
```
Proful, line 527: thanks for cool recommendation :)

packages/lexical-playground/src/nodes/YouTubeNode.jsx
```diff
@@ -62,7 +62,8 @@ export class YouTubeNode extends DecoratorNode<React$Node> {
   }
 
   createDOM(): HTMLElement {
-    return document.createElement('div');
+    const elem = document.createElement('span');
+    return elem;
```
harish-sethuraman, line 66: May be? 
```suggestion
    return document.createElement('span');;
```

zurfyx:
I thought we were considering the possibility of the DecoratorNode handling the alignment by itself by listening to commands rather than an ElementNode which forces to handle the highlighted issues by @fantactuka. Where did we leave that conversation?

thegreatercurve:
> Rather than creating an EmbedNode, couldn't the DecoratorNode just listen to the align command and react accordingly?
> I thought we were considering the possibility of the DecoratorNode handling the alignment by itself by listening to commands rather than an ElementNode which forces to handle the highlighted issues by @fantactuka. Where did we leave that conversation?

@trueadm @zurfyx Probably should have put this in the description. I've got a branch in which I first tried this approach (https://github.com/facebook/lexical/compare/decorator-parent-level?expand=1), 

I'm adding `__format` to the state of each embed decorator node, so that it works for undo/redo and copy/paste. 

The problem though is whenever we trigger `createDOM`, whenever the alignment changes, the contents within the newly create element will reload, creating weird loading flashes: 


https://user-images.githubusercontent.com/7748470/163979099-daa73f4a-e2c1-47c6-aea1-c65b1d81bcee.mov

thegreatercurve:
Closing as I realised there's probably a better way of doing this #1867

PR #1864, opened 2022-04-19
Fix issue with inserting paragraphs between text nodes.

### LINKED ISSUE #1861: Bug: Inserting line break between two text nodes does not break a line correctly
Inserting line break between two text nodes currently insert a new line/paragraph without breaking the line from the text caret. It works as expected if your caret is within a text node though. Please see video below for repro steps.

Lexical version: 0.2.2

## Steps To Reproduce

https://user-images.githubusercontent.com/118264/163897483-61fbd625-2063-4a48-8eff-f6e37c3ec8a4.mp4

Link to code example:

acywatson:
Oops! Added a test so we won't hit this one again.

Fixes #1861

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,10 +2,12 @@
 
 packages/**/dist/*.js
 packages/**/build/*.js
+packages/**/npm/**/*
 packages/**/config/*.js
 packages/playwright
 packages/playwright-core
 packages/**/vite.config.js
+**/*.md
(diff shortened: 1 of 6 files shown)


zurfyx (approved):

trueadm (approved):

trueadm:
Does this work for when the cursor as at offset 0 on the node after the bold node too?

PR #1867, opened 2022-04-19
Fix bug with alignment for root level decorator nodes v2

thegreatercurve:
Should fix https://github.com/facebook/lexical/issues/1696.

https://user-images.githubusercontent.com/7748470/163989526-98345536-3dbb-43b9-9bd7-63ecb5923b38.mov

### DIFF
--- a/packages/lexical-playground/src/nodes/TweetNode.jsx
+++ b/packages/lexical-playground/src/nodes/TweetNode.jsx
@@ -7,7 +7,7 @@
  * @flow strict
  */
 
-import type {LexicalNode, NodeKey} from 'lexical';
+import type {ElementFormatType, LexicalNode, NodeKey} from 'lexical';
 
 import {DecoratorNode} from 'lexical';
 import * as React from 'react';
@@ -19,6 +19,7 @@ const getHasScriptCached = () =>
   document.querySelector(`script[src="${WIDGET_SCRIPT_URL}"]`);
 
 type TweetComponentProps = $ReadOnly<{
+  format: ?ElementFormatType,
   loadingComponent?: React$Node,
   onError?: (error?: Error) => void,
   onLoad?: () => void,
@@ -26,6 +27,7 @@ type TweetComponentProps = $ReadOnly<{
 }>;
 
 function TweetComponent({
+  format,
   loadingComponent,
   onError,
   onLoad,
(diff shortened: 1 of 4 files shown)


packages/lexical-playground/src/nodes/TweetNode.jsx
```diff
@@ -105,7 +111,18 @@ export class TweetNode extends DecoratorNode<React$Node> {
   }
 
   decorate(): React$Node {
-    return <TweetComponent loadingComponent="Loading..." tweetID={this.__id} />;
+    return (
+      <TweetComponent
```
zurfyx, line 115: Should we create a BlockComponent wrapper on top that handles the container bits for you automatically? This layer can also listen to `FORMAT_TEXT` commands which would solve centering when clicking the node and can be reused within Twitter and Youtube and possibly others soon (so I'd even consider making it part of @lexical/react)

thegreatercurve:
Found another related issue whilst fixing this one: #1868

packages/lexical-playground/src/nodes/TweetNode.jsx
```diff
@@ -72,15 +74,19 @@ function TweetComponent({
...
+        style={{display: 'inline-block', width: '550px'}}
+        ref={containerRef}
+      />
+    </div>
   );
 }
 
 export class TweetNode extends DecoratorNode<React$Node> {
   __id: string;
+  __format: ?ElementFormatType;
```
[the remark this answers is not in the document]
trueadm: Why put it on DecoratorNode? Inline decorators wouldn't use it? It's best to keep it here IMO.

packages/lexical-playground/src/nodes/TweetNode.jsx
trueadm: That's a good shout!

thegreatercurve: I think this and the above both seem like sensible suggestions. What do you think @trueadm? This probably has implications for the work you mentioned last week about adding fake cursors for `DecorateNodes`, as they're not currently focusable by themselves?

trueadm: You should be adding selection to embeds, in a similar way to how ImageNode and ExcalidrawNode has the ability to work with NodeSelection.

trueadm (approved):

packages/lexical-playground/src/nodes/TweetNode.jsx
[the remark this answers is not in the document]
thegreatercurve: If we add any more instances might be worth adding an intermittent extension of `AlignableDecoratorNode`/`BlockDecoratorNode`, but probably overkill in the mean time.

Issue #1849, opened 2022-04-18
Bug: Text replacement not working after newline in mobile Safari


jclem:
When typing a text replacement abbreviation directly after a newline, the text of the previous line is included what Safari sees as the current text, and therefore the replacement does not happen.

Lexical version: Whatever is on playground.lexical.dev as of Apr. 18, 2022 9:27AM ET.

## Steps To Reproduce

1. Type some text, like "Hello" on a line
2. Hit the return key to enter a new line
3. Type a text replacement abbreviation

Link to code example: n/a

## The current behavior

The replacement is not triggered, because if the previous line is "hello", and the abbreviation is "omw", Safari thinks the current text is "helloomw".

https://user-images.githubusercontent.com/412282/163815350-cac5bdc9-7068-490a-a3eb-ab96bc21b866.mp4

## The expected behavior

The expected behavior is that the first line contains the text that was typed, and on the second line, text replacement occurs. In this case, the "omw" abbreviation should be replaced with "On my way!"

https://user-images.githubusercontent.com/412282/163815598-fb0daf3a-7a02-47f2-8ba2-18384195b5fe.mp4

trueadm:
I know what is causing this. I’ll look into it tomorrow when I’m back at work.

trueadm:
I have a fix up: https://github.com/facebook/lexical/pull/1869. This should also fix the other issue too.

Issue #1846, opened 2022-04-18
Feature: Capitalize first letter after return in mobile Safari


jclem:
In most text inputs on iOS, including in mobile Safari, after the user hits the return key to insert a new line, the next letter is auto-capitalized. This does not appear to be the case with Lexical.

I believe ProseMirror is the only editor toolkit that is able to preserve this behavior (albeit in a way that's occasionally buggy), so it's possible this is not fixable without some new APIs being introduced in Safari.

I've filed this as a feature request, because almost no other editor toolkits preserve this behavior; it didn't feel right to file it as a bug.

**Expected behavior:**

https://user-images.githubusercontent.com/412282/163811330-02f4b7ee-304a-4d28-9315-f429319a1116.mp4

**Observed behavior:**

https://user-images.githubusercontent.com/412282/163811128-3be0f33b-b8e8-4a6c-b820-4791b2f6275d.mp4

trueadm:
Do you know how ProseMirror is making this work? We could likely do something similar.

jclem:
Amazing @trueadm—thanks! Faster to fix than for me to answer your question 😄

PR #1869, opened 2022-04-19
Avoid preventDefault on pressing enter

trueadm:
Fixes https://github.com/facebook/lexical/issues/1849 and fixes https://github.com/facebook/lexical/issues/1846.

If we have support for `beforeinput`, we can use that event for signal rather than blocking the return/enter key press via `preventDefault()`. This allows iOS to function better – although it's a bit strange how this is the only platform that suffers from this issue :P

### DIFF
--- a/packages/lexical-rich-text/src/index.js
+++ b/packages/lexical-rich-text/src/index.js
@@ -68,6 +68,7 @@ import {
   PASTE_COMMAND,
   REMOVE_TEXT_COMMAND,
 } from 'lexical';
+import {CAN_USE_BEFORE_INPUT} from 'shared/environment';
 
 export type InitialEditorStateType = null | string | EditorState | (() => void);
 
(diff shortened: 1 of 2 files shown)


zurfyx (approved):

packages/lexical-rich-text/src/index.js
```diff
@@ -605,6 +606,13 @@ export function registerRichText(
           return false;
         }
         if (event !== null) {
+          // If we have beforeinput, then we can avoid blocking
+          // the default behavior. This ensures that the browser/OS
+          // can intercept that we're actually inserting a paragraph,
+          // and autocomplete, autocapitialize etc work as intended.
+          if (CAN_USE_BEFORE_INPUT) {
```
fantactuka: Can we check for ios?

PR #1871, opened 2022-04-19
Excalidraw fixes

rulikrulit:
- Set readOnly mode when Excalidraw is open to avoid unnecessary updates
- Move Excalidraw Modal to a portal to avoid interacting with Lexical and other wrappers event listeneners
- Fix Excalidraw image size
- Remove Excalidraw Node if the content of the saved drawing is empty

### DIFF
--- a/packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
+++ b/packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
@@ -37,9 +37,8 @@ function ExcalidrawComponent({
   nodeKey: NodeKey,
 }): React.Node {
   const [editor] = useLexicalComposerContext();
-  const shouldModalOpenInitially = data === '[]' && !editor.isReadOnly();
   const [isModalOpen, setModalOpen] = useState<boolean>(
-    shouldModalOpenInitially,
+    data === '[]' && !editor.isReadOnly(),
   );
   const buttonRef = useRef<HTMLElement | null>(null);
   const [isSelected, setSelected, clearSelection] =
--- a/packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
+++ b/packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
@@ -7,6 +7,7 @@
  * @flow strict
  */
 
+import type {ExcalidrawElementFragment} from './ExcalidrawModal';
 import type {EditorConfig, LexicalEditor, LexicalNode, NodeKey} from 'lexical';
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
@@ -35,8 +36,11 @@ function ExcalidrawComponent({
   data: string,
   nodeKey: NodeKey,
 }): React.Node {
-  const [isModalOpen, setModalOpen] = useState<boolean>(data === '[]');
   const [editor] = useLexicalComposerContext();
+  const shouldModalOpenInitially = data === '[]' && !editor.isReadOnly();
+  const [isModalOpen, setModalOpen] = useState<boolean>(
+    shouldModalOpenInitially,
+  );
   const buttonRef = useRef<HTMLElement | null>(null);
   const [isSelected, setSelected, clearSelection] =
     useLexicalNodeSelection(nodeKey);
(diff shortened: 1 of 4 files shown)


packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
```diff
@@ -35,8 +36,11 @@ function ExcalidrawComponent({
   data: string,
   nodeKey: NodeKey,
 }): React.Node {
-  const [isModalOpen, setModalOpen] = useState<boolean>(data === '[]');
   const [editor] = useLexicalComposerContext();
+  const shouldModalOpenInitially = data === '[]' && !editor.isReadOnly();
```
zurfyx, line 40: Let's move logic inside so that it isn't recomputed every time

packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
```diff
@@ -59,6 +63,13 @@ function ExcalidrawComponent({
     [editor, isSelected, nodeKey, setSelected],
   );
 
+  // Set editor to readOnly if excalidraw is open to prevent unwanted changes
+  useEffect(() => {
+    if (isModalOpen) {
+      editor.setReadOnly(true);
+    }
+  }, [isModalOpen, editor]);
+
```
zurfyx, line 72: Do we need a useEffect for this? We know when we show/hide the Exalidraw modal, no?

zurfyx (commented):
> Set readOnly mode when Excalidraw is open to avoid unnecessary updates

When does the editor trigger updates when the modal is opened?

packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
rulikrulit: If you are doing copy/paste or you are in a read mode the modal may be hidden and it is controlled externally. useEffect will only be triggered when the value changes regardless the source which is exactly what I was looking for. Please, let me know if you see any better way to do it.

rulikrulit:
> > Set readOnly mode when Excalidraw is open to avoid unnecessary updates
> 
> When does the editor trigger updates when the modal is opened?

If currently you open an excalidraw and press backspace it is going to remove the excalidraw node in Editor because editor is listening to keyboard events.

packages/lexical-playground/src/nodes/ExcalidrawNode/index.jsx
zurfyx: It's taken into account only use as part of the default value of `useState` but it's computed in every single rerender. That's why for expensive functions you want to use `useMemo`. In this case, it's not an expensive function but can easily made more optimal by just moving this logic inside the `useState` param

zurfyx (commented):
Had another pass through the PR. Do you mind explaining why we need the modal open logic? I presume we want the modal to open when you click the button and the modal to close when save or discard?

```
const [open, setOpen] = useState(false);
// useEffect -> registerCommand
if (command) {
  setOpen = true;
}
// Exalidraw component
portal(
  <Excalidraw onSave={setOpen -> false} onHide={setOpen -> false}
);
```

I'm also unsure what the benefits of setting the editor to readonly is. Will the editor read any event once the portal is up?

rulikrulit:
> Do you mind explaining why we need the modal open logic?

There are some other cases. For example, the modal should open when you create a new node and the modal should never open in readOnly mode

rulikrulit:
> Will the editor read any event once the portal is up?

It will also read events if you click outside the portal. And assuming the Editor is covered by the modal window you will be doing changes without seeing them.

zurfyx:
> For example, the modal should open when you create a new node and the modal should never open in readOnly mode

That should be up to the plugin not the node itself. In this case, the toolbar should handle the readOnly mode and "disable" buttons that are forbidden.

> > Will the editor read any event once the portal is up?
> 
> It will also read events if you click outside the portal. And assuming the Editor is covered by the modal window you will be doing changes without seeing them.

Ah good point, sounds like we should perfect the modal component to have an invisible or darker background that once you click on it it dismisses the modal. (Well, this can be done in many ways but the point is that the modal should be dismissed when clicking outside rather than letting you interact with the outside world while the modal is open, it doesn't make sense UX wise. Do you agre?

zurfyx (approved):

PR #1876, opened 2022-04-19
fix: missing types in LexicalCode

wobsoriano:
Adds types of `getCodeLanguages` and `getDefaultCodeLanguage` in `@lexical/code`


zurfyx (approved):
Thanks!

PR #1877, opened 2022-04-19
Disable prettier

zurfyx:
CI is breaking for some reason, `npm run prettier` doesn't pass


acywatson (changes requested):
I don't know if this is the right solution, because it was fixing a real problem. Can we just fix the underlying issue?

PR #1883, opened 2022-04-20
Fix list outdent & indent bug

### LINKED ISSUE #1421: Bug: When selecting multiple list items and outdent & indent new items are created
Lexical version: latests

## Steps To Reproduce

https://user-images.githubusercontent.com/1071113/156743667-fe9bef3a-2e84-4cef-858c-93bccd32d415.mov

1. Create a list with elements with different indentation
2. Select all the items of the list
3. Press shift + tab to outdent
4. Press shift to indent
5. A list item is created (bug)
6. If you repeat steps 2 & 3 more list items will be created

Link to code example: Can be replicated using playground

## The current behavior

Pressing shift + tab and shift is creating new list items.

## The expected behavior

Pressing shift + tab and shift should only indent and outdent.

davidnguyen7:
Fixes #1421


packages/lexical-list/src/formatList.js
```diff
@@ -185,7 +187,14 @@ export function $handleIndent(listItemNodes: Array<ListItemNode>): void {
         if ($isListNode(nextInnerList)) {
           const children = nextInnerList.getChildren();
           innerList.append(...children);
-          nextInnerList.remove();
+          nextSibling.remove();
```
acywatson, line 190: Thanks for working on this!

I had to re-familiarize myself with the logic here. It looks like the root problem is that we're trying to process nodes from the listItemNodes param that were part of the original indented selection, but have been "handled" already (in this case, removed). Specifically, if we have a tree like this:

root
  └ (6) list 
    ├ (5) listitem 
    | └ (7) text "a"
    ├ (8) listitem 
    | └ (9) text "b"
    └ (86) listitem 
      └ (87) list 
        └ (10) listitem 
          └ (13) text "c"

And we select the whole and indent, then the algo does something like this:

1) for listItemNode[0] with key 5: it has no nested siblings to merge with, so we indent it by creating a new nested list and append it to it. Now we have:

root
  └ (6) list 
        └ (88) listitem
              └ (89) list
                    ├ (5) listitem 
                        └ (7) text "a"
         ├ (8) listitem 
              └ (9) text "b"
          └ (86) listitem 
              └ (87) list 
                  └ (10) listitem 
                        └ (13) text "c"

2) for listItemNode[1] key(8), it has nested lists on both sides, so we merge them all together by appending (8) to (89) and (10) to (89). Then, we remove (87), but not (86)! So, we have a dangling ListItem (86).

3) for listItemNode[2] key (86), we removed it's child list, so it no longer passes the isNestedListNode early return check at the top. It ends up getting appended to the previous sibling nested list.

4) for listItemNode[3] key (10), we end up just indenting it much like we did the first list item node.

Interestingly, I think just your one change of ```nextInnerList.remove()``` -> ```nextSibling.remove()``` fixes the immediate issue. This works because removing a node doesn't actually immediately change that nodes children - it just removes it from its parents children, effectively dereferencing it and waiting for the GC to clean it up. Since the children aren't removed, it still passes the isNestedListNode early return check on line 175.

Even though this is the simplest fix, I'm not sure if it's the right one for two reasons:

1) I'm not sure it's a good idea to rely on the implementation detail that children are still reference in removed parents. This could probably change at some point in the future. This scenario has already made me rethink it.

2) I think we are generally doing more work than we need to in this function, and the approach of removing "handled" nodes from the list of nodes to process could save some iterations.

The one thing I would take issue with is the way we're looping through all the listItemNodes to check their keys and then mutate the listItemNodes. Instead, I might suggest adding the removed node keys to a Set or something else with constant lookup time. Then, we can check that Set for the node key in the early return conditional and ignore it  if it has been removed.

```
const removed = new Set();
...
if (isNestedListNode(listItemNode) || removed.has(listItemNode.getKey()) { return };
...
```

The other thing I would ask it that we try to write a simple regression or e2e test for this.

acywatson (approved):

PR #1890, opened 2022-04-20
fix: incorrect import in lexical-rich-text types

wobsoriano:
replaces `HeadingNode` import to `LexicalEditor`

Reference https://github.com/facebook/lexical/blob/main/packages/lexical-rich-text/flow/LexicalRichText.js.flow#L16


acywatson (approved):
Looks right, thanks!!

PR #1895, opened 2022-04-20
Add embed block to playground

thegreatercurve:
- Add embed block to playground to handle selection for Twitter and YouTube nodes.
- Abstract FORMAT_ELEMENT_COMMAND logic for alignment into this block as well.

https://user-images.githubusercontent.com/7748470/164219042-2a7521c6-7029-4c16-8706-09ebc2bbb48e.mov


packages/lexical-playground/src/ui/EmbedBlock.jsx
```diff
@@ -0,0 +1,134 @@
...
+    useLexicalNodeSelection(nodeKey);
+  const ref = useRef();
+  const onDelete = useCallback(
+    (payload) => {
+      if (isSelected && $isNodeSelection($getSelection())) {
+        const event: KeyboardEvent = payload;
+        event.preventDefault();
+        editor.update(() => {
+          const node = $getNodeByKey(nodeKey);
+          if ($isYouTubeNode(node) || $isTweetNode(node)) {
```
zurfyx, line 57: Can we make this more generic? Maybe reusing the property in the previous comment? Same in the other places in this file

packages/lexical-playground/src/ui/EmbedBlock.jsx
```diff
@@ -0,0 +1,134 @@
...
+import {$isTweetNode} from '../nodes/TweetNode.jsx';
+import {$isYouTubeNode} from '../nodes/YouTubeNode.jsx';
+
+type Props = $ReadOnly<{
+  children: React$Node,
+  format: ?ElementFormatType,
+  nodeKey: NodeKey,
+}>;
+
+export default function EmbedBlock({
```
zurfyx, line 41: What exactly is an embed? Is it the same concept as isInline for ElementNode? In this case should we just introduce a property to DecoratorNode

zurfyx (commented):
This is looking really good!

packages/lexical-playground/src/ui/EmbedBlock.jsx
thegreatercurve: If I make this more generic by adding `isInline` to the `DecoratorNode`, is this component abstract enough to live in `@lexical/react`? I guess it could be a common use case.

thegreatercurve: I've made this a little bit more generic by extending DecoratorNode and then adding the format property to its state. I can then do the check for that node in here instead. 

We have `isTopLevel` on the DecoratorNode already, so I've reusing that property as well to check it's not a child.

packages/lexical-playground/src/ui/EmbedBlock.jsx
thegreatercurve: An embed is basically a top level block DecoratorNode, which can be align, but which not allow text to flow around it. I've tried to abstract this by extending from DecoratorNode. Let me know what you think.

packages/lexical-react/src/LexicalBlockDecoratorNode.js
```diff
@@ -0,0 +1,39 @@
...
+import {DecoratorNode} from 'lexical';
+
+export class LexicalBlockDecoratorNode extends DecoratorNode<React$Node> {
+  __format: ?ElementFormatType;
+
+  createDOM(): HTMLElement {
+    return document.createElement('div');
+  }
+
+  isTopLevel(): true {
```
zurfyx, line 21: I believe this is the most correct for now but this also limits the use cases to the very top. For example, it would not be correct to place it inside another ElementNode. 

EDIT: Actually, do we need this? See next comment

packages/lexical-react/src/LexicalBlockWithAlignableContents.jsx
```diff
@@ -0,0 +1,134 @@
...
+  useEffect(() => {
+    return mergeRegister(
+      editor.registerCommand(
+        FORMAT_ELEMENT_COMMAND,
+        (payload) => {
+          if (isSelected) {
+            const selection = $getSelection();
+            if ($isNodeSelection(selection)) {
+              const node = $getNodeByKey(nodeKey);
+              if ($isDecoratorNode(node) && node.isTopLevel()) {
```
zurfyx, line 77: $isBlockDecoratorNode?

packages/lexical-react/src/LexicalBlockWithAlignableContents.jsx
```diff
@@ -0,0 +1,134 @@
...
+import * as React from 'react';
+import {useCallback, useEffect, useRef} from 'react';
+
+type Props = $ReadOnly<{
+  children: React$Node,
+  format: ?ElementFormatType,
+  nodeKey: NodeKey,
+}>;
+
+export function LexicalBlockWithAlignableContents({
```
zurfyx, line 40: LexicalBlock? AlignableContents sounds oddly specific

zurfyx (approved):
LGTM, that's awesome! Looking forward to use this already in WWW. Adding a couple nits

I wonder if the rest of team have additional thoughts on this but I'll stamp already

tylerjbainbridge:
> I see where you're coming from but you're probably influenced by the fact you've seen both versions. A block in this case literally means a CSS block. What else do you have in mind?

Okay, true, maybe just drop the Lexical and swap the order of the types? -> `DecoratorBlockNode`

That way it's still very clear it's a `Decorator` type node, but one with special properties.

packages/lexical-playground/src/nodes/YouTubeNode.jsx
```diff
@@ -9,31 +9,19 @@
...
-      }
-
-      previousYouTubeIDRef.current = videoID;
-    }
-  }, [onLoad, videoID]);
-
+function YouTubeComponent({format, nodeKey, videoID}: YouTubeComponentProps) {
   return (
-    <div style={{textAlign: format}}>
+    <LexicalBlockWithAlignableContents format={format} nodeKey={nodeKey}>
```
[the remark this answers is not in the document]
tylerjbainbridge: One less thing to remember to do when building `DecoratorNodes`. Especially since I bet this will be a pretty common use case.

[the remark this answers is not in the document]
thegreatercurve: > If the Decorator is a LexicalBlockDecoratorNode -> wrap it's decorate() return value in a LexicalBlockWithAlignableContents with the required props.

it's a good shout. I did think about it, but I think it would be difficult just because you can only ever have one `decorate` method on the nodes that extend from `DecoratorNode`. 

It will just look up the chain of inheritance until it finds the one instance it needs, and then uses whatever is within that node.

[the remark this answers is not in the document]
thegreatercurve: Also, it might be useful to keep `DecoratorBlockNode` separate from all the command listener, in case someone wants to overwrite the command priorities (for example), but still use `DecoratorBlockNode`

packages/lexical-react/src/LexicalBlockWithAlignableContents.jsx
thegreatercurve: @zurfyx Maybe just keep it as `BlockWithAlignableContents` for the moment? We can revisit if we add in more functionality later. I guess this div could also become resizable and could be used to wrap the image, excalidraw and poll nodes.

thegreatercurve:
> > I see where you're coming from but you're probably influenced by the fact you've seen both versions. A block in this case literally means a CSS block. What else do you have in mind?
> 
> Okay, true, maybe just drop the Lexical and swap the order of the types? -> `DecoratorBlockNode`
> 
> That way it's still very clear it's a `Decorator` type node, but one with special properties.

@tylerjbainbridge  Good shout. Just updated the naming.

PR #1896, opened 2022-04-20
add default error handler in createEditor config

alajillo:
add default error handler when user don't put onError in createEditor config.

### DIFF
--- a/packages/lexical/src/LexicalEditor.js
+++ b/packages/lexical/src/LexicalEditor.js
@@ -253,7 +253,7 @@ export function createEditor<EditorContext>(editorConfig?: {
     ParagraphNode,
     ...(config.nodes || []),
   ];
-  const onError = config.onError;
+  const onError = config.onError || console.warn;
   const isReadOnly = config.readOnly || false;
 
   const registeredNodes = new Map();


alajillo:
> 
thank you for comment :)
yes i'm reffering to who people don't use type system.  in another way it maybe be better solution that edit docs to tell use errorHandler

packages/lexical/src/LexicalEditor.js
```diff
@@ -253,7 +253,7 @@ export function createEditor<EditorContext>(editorConfig?: {
     ParagraphNode,
     ...(config.nodes || []),
   ];
-  const onError = config.onError;
+  const onError = config.onError || console.warn;
```
trueadm, line 256: `onError` is always meant to be passed in, that's why the type isn't optional. I wonder if we should error if one is not passed in instead of defaulting to console.warn, which can lead to confusion?

alajillo: > `onError` is always meant to be passed in, that's why the type isn't optional. I wonder if we should error if one is not passed in instead of defaulting to console.warn, which can lead to confusion?

well i think it doesn't lead to confusion. using typescript is best way but we should tell them who use javascript that onError is not optional.

PR #1900, opened 2022-04-20
Do not reconcile selection during readOnly

trueadm:
If the editor is in a read only mode, do not attempt to reconcile selection.


acywatson (approved):
yep

PR #1901, opened 2022-04-21
Use TS type predicates for $isXNode checks

nhardy:
Uses TypeScript type predicates for `$isXNode` and `$isXSelection` checks in the `.d.ts` files. Looks like this is already happening in flow.


acywatson:
You may need to run ```npm run prettier:fix``` locally to get that integrity check to pass.

zurfyx (approved):

zurfyx:
Thanks!

PR #1902, opened 2022-04-21
Base markdown export & import/export toggle for the playground

fantactuka:
- Some basic markdown exporting
- Updated md button in the playground to be a bit more functional: toggling editor content between its markdown representation within code block and back to rich text.

https://user-images.githubusercontent.com/132642/164358063-5adc98bd-19eb-40d1-baf9-7e9ae05329f2.mov

### DIFF
--- a/packages/lexical-markdown/src/convertToMarkdown.js
+++ b/packages/lexical-markdown/src/convertToMarkdown.js
@@ -1,0 +1,157 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import type {ElementNode, LexicalNode, TextFormatType, TextNode} from 'lexical';
+
+import {$isLinkNode} from '@lexical/link';
+import {$getRoot, $isElementNode, $isLineBreakNode, $isTextNode} from 'lexical';
+
+import {
+  getAllMarkdownCriteriaForParagraphs,
+  getAllMarkdownCriteriaForTextNodes,
+} from './utils';
+
+export function $convertToMarkdownString(): string {
+  const output = [];
+  const children = $getRoot().getChildren();
+
+  for (const child of children) {
+    const result = exportTopLevelElementOrDecorator(child);
+    if (result != null) {
+      output.push(result);
+    }
+  }
+
+  return output.join('\n');
+}
+
+function exportTopLevelElementOrDecorator(node: LexicalNode): string | null {
+  const blockTransformers = getAllMarkdownCriteriaForParagraphs();
+  for (const transformer of blockTransformers) {
+    if (transformer.export != null) {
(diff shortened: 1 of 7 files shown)


packages/lexical-markdown/src/convertToMarkdown.js
```diff
@@ -0,0 +1,160 @@
...
+function encodeURLTitle(url: string): string {
+  return url.replace(/\\/g, '\\\\').replace(/\[/g, '\\[').replace(/\]/g, '\\]');
+}
+
+function $hasSingleListChild(node: ElementNode): boolean {
+  return node.getChildrenSize() === 1 && $isListNode(node.getFirstChild());
+}
+
+export function $convertToMarkdownString(): string {
+  const nodes = $dfs();
```
acywatson, line 43: Would it not be better to just visit the nodes as we go through the tree the first time instead of DFS'ing to get them into a list, then traversing the list?

tylerjbainbridge: Would be cool to have a version of `$dfs` with a callback you can use to do something at each node.

```js
$dfsEach((node) => {
  // Do something
});
```

packages/lexical-markdown/src/convertToMarkdown.js
```diff
@@ -0,0 +1,160 @@
...
+  let text = '';
+  let shouldAppendFirstLineBreak = false;
+
+  const closingBracketStack: Array<{depth: number, endText: string}> = [];
+  const listStack: Array<{count: number, depth: number, tag: 'ul' | 'ol'}> = [];
+  const quoteStack: Array<{depth: number}> = [];
+  const codeStack: Array<{depth: number}> = [];
+  for (let i = 0; i < nodes.length; i++) {
+    const {depth, node} = nodes[i];
+    if ($isHeadingNode(node)) {
```
tylerjbainbridge, line 53: This can be changed in the future, but since it's already broken down by node type I wonder if this middle portion would be cleaner if implemented as an `exportMarkdown` and `importMarkdown` methods on `LexicalNodes`. 

But maybe since we'll only ever need to support what Markdown supports (basic core types) we can just keep it isolated here. Just an idea :)

tylerjbainbridge (approved):
Looks good to me! Will Table support come in a later version?

packages/lexical-markdown/src/convertToMarkdown.js
fantactuka: I think supported node types should be passed from outside as much as possible (e.g. for cases when users would want to have custom HeadingNode but still make it convertable in/from markdown), will be addresses in a follow ups as this one is barebones setup to unblock the rest

packages/lexical-markdown/src/convertToMarkdown.js
fantactuka: Put up different version that does exactly that

tylerjbainbridge: Beautiful!

PR #1906, opened 2022-04-21
Add predicate to  $isLeafNode in .flow file

NicolasRoelandt:
Using $isLeafNode with Flow typing does not refine the type. This is because it's not declared as a predicate function (`%checks`). In order to have a predicate function, Flow needs the implementation. I suggest to add the function implementation in the flow file. This is not ideal as it duplicates things : the implementation in the flow file could get out of sync from the actual implementation, but in my opinion this is still an improvement.

I only do it for Flow here, as I am not very familiar with TypeScript

### DIFF
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -828,7 +828,9 @@ declare export function $getNearestNodeFromDOMNode(
 ): LexicalNode | null;
 declare export function $getNodeByKey<N: LexicalNode>(key: NodeKey): N | null;
 declare export function $getRoot(): RootNode;
-declare export function $isLeafNode(node: ?LexicalNode): boolean;
+export function $isLeafNode(node: ?LexicalNode): boolean %checks {
+  return $isTextNode(node) || $isLineBreakNode(node) || $isDecoratorNode(node);
+}
 declare export function $setCompositionKey(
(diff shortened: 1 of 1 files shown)


tylerjbainbridge (approved):
Thanks @NicolasRoelandt!

acywatson:
This makes sense,  but do we need to do it for every type guard we expose in other flow files too?

NicolasRoelandt:
@acywatson true, this was more of a PoC as I wasn't sure if we were ok with duplicating the implementation.
 If we want to do this, I can have a look at the other predicates

packages/lexical/flow/Lexical.js.flow
```diff
@@ -828,7 +828,9 @@ declare export function $getNearestNodeFromDOMNode(
 ): LexicalNode | null;
 declare export function $getNodeByKey<N: LexicalNode>(key: NodeKey): N | null;
 declare export function $getRoot(): RootNode;
-declare export function $isLeafNode(node: ?LexicalNode): boolean;
+export function $isLeafNode(node: ?LexicalNode): boolean %checks {
```
trueadm, line 831: You can't put in the body of definitions in the declaration files. We instead need to use the predicate syntax. See https://github.com/facebook/lexical/pull/1914 :)

PR #1907, opened 2022-04-21
change test command to e2e command

harish-sethuraman:
There is a `npm run test` in the readme `contributing to lexical` section. So I've changed the command to run e2e test.


acywatson (approved):
Makes sense - I think they can look at the package.json scripts to see what other commands are available.

PR #1908, opened 2022-04-21
Addition of Table of Contents (#1865)

### LINKED ISSUE #1865: Table of contents plugin
Could potentially (but not necessarily) look like this
tableofcontents

firstChairCoder:
Added table of contents to README.md

This commit resolves #1865.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,29 @@
 
 **⚠️ Lexical is currently in early development and APIs and packages are likely to change quite often.**
 
+
+## Table of Contents
+- [Getting started with React](#getting-started-with-react)
+- [Lexical is a framework](#lexical-is-a-framework)
+  * [Editor instances](#editor-instances)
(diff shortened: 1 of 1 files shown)


zurfyx:
Thanks and sorry for the confusion! The issue is referring to a brand new plugin for the playground editor. No strong feelings about having a table of contents on the README but where it is now it adds clutter to the a key area of the README which is the introduction.

I wonder what other folks in the team think about this

firstChairCoder:
Thanks for getting back so soon. I did see the signs going through recent previous commits. Sorry about that, I will revise.

zurfyx:
What do you think of the same approach but below the playground and sandbox links? And only the 4 heading ones?

firstChairCoder:
That can be done. Thanks. I will also try to handle the original issue as intended.

zurfyx (approved):

Issue #1911, opened 2022-04-21
Copy-pasting header loses format


zurfyx:
May apply to other nodes

https://user-images.githubusercontent.com/193447/164510007-7c6e255e-8d58-4d1f-936b-d4f58a2e759e.mov

trueadm:
Fixed with https://github.com/facebook/lexical/pull/1913

PR #1913, opened 2022-04-21
Simplify extraction of text

trueadm:
Fixes https://github.com/facebook/lexical/issues/1911. Turns out the fast-path logic wasn't too smart after all.


zurfyx:
That was fast, thanks!

zurfyx (approved):

acywatson (approved):

PR #1915, opened 2022-04-21
Change timeout config

acywatson:
I don't think tests that hit the global timeout are being properly retried. Let's try this config.

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -26,12 +26,10 @@ const config = {
   retries: IS_CI ? 4 : 1,
   timeout: 150000,
   use: {
-    
-    
     navigationTimeout: 30000,
     // this causes issues in the CI on on current version.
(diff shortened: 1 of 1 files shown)


fantactuka:
Would it help to add job/step timeout too?
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

By default job can run up to 6 hours

acywatson:
> Would it help to add job/step timeout too? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes
> 
> By default job can run up to 6 hours

Probably, yea - I'd like to follow up in a separate PR just to simplify and isolate this change.

fantactuka (approved):

Issue #1904, opened 2022-04-21
Bug: Lexical selection is saying wrongly that some nodes belong to the selection


Fetz:
Lexical version: 0.2.3

## Steps To Reproduce

1. create 3 words separated by multiple spaces
2. make part of those words (I think this is optional) with inline styles (bold, italic)
3. make the left and right words link
4. make the center word a link
5. if steps are follow correctly the anchorNode and focusNodes will be the textNodes outside and will not belong to the selection
6. But if you check the lexical selection it will say that that the anchor nodes and focus nodes are selected.

https://user-images.githubusercontent.com/1071113/164434689-0bd7b2c5-4ee7-4527-894d-87e3c12d7f4a.mov

## The current behavior
Lexical Selection and DOM Selection have the same anchorNode and focusNode but Lexical says wrongly that these nodes are selected but they are not.

## The expected behavior
Either:
- Lexical selection will return different anchorNode and focusNode then the browser that belongs to always to selection
- or it should use the same DOM Selection anchorNode and focusNode but should detect correctly if these nodes are in the selection or not.

trueadm:
Lexical normalizes DOM selection under-the-hood to reflect consistent values at an abstract level – so you shouldn't expect them to match up. Forcing them match up is what causes so many issues, because different browsers have very different ideas as to what should be selected. Stick with using Lexical's selection model if you can. :)

Fetz:
@trueadm but in this scenario that I highlighted in the video anchorNode and focusNode both in lexical and the browser are the same (I extracted the text content to double check I got the same content) but that anchorNode or focusNode either shouldn't be returning selected or Lexical should instead set the anchorNode/focusNode the nodes inside the selection.

trueadm:
Ah, sorry. I just realized I was looking at the wrong thing above. The logic is working correctly, it's just you're reading the selection of an editor state in the past. If you want to get the latest selection, you need to do `$getSelection` inside `editor.update()`. Doing it in a read will only get the selection of the editor state when it was created.

trueadm:
This is what I mean by normalization. The DOM selection is at the very end of the whitespace text node (you can see this in your debug view) and also at the start of the other whitespace text node (also in your debug view). If you're at the boundary of two text nodes, you're technically selecting the interception of both of them. Given in this case it's an element – a link node, you're selecting the link node.

Fetz:
Sadly is not what is happening in that specific example that I created.
A - Left whitespace
B - Link that is selected
[...]

PR #1917, opened 2022-04-21
Fix boundary selection issues

trueadm:
Fixes https://github.com/facebook/lexical/issues/1904. This was an interesting one. Logically, we were missing some selection resolution code around the handling of inline elements – but by fixing that, it also made it clear that some of the logic for handling links and character offset resolution was also a bit loose. This PR fixes that and adds a regression test to capture that main issue.


fantactuka (approved):

zurfyx:
Impressive PR as usual. Seems like some E2E tests are genuine failures?

acywatson:
Somehow the CI is running for a very long time as playwright waits for an element to be visible to click on, retrying literally over 1000 times. We need to add a timeout somewhere - maybe in the workflow, as @fantactuka suggested, or somewhere in the playwright config (or both)

Issue #1885, opened 2022-04-20
Bug: Typescript Error for theme


Proful:
Lexical version: 0.2.3

## Steps To Reproduce

1. packages/lexical-playground/src/themes/PlaygroundEditorTheme.js

## The current behavior
Above theme file in typescript complaining with below error

```
Property 'list' is incompatible with index signature.
  Type '{ listitem: string; nested: { listitem: string; }; olDepth: string[]; ul: string; }' is not assignable to type 'string | Record<string, string>'.
    Types of property 'nested' are incompatible.
      Type '{ listitem: string; }' is not assignable to type 'string'.ts(2530)
```

zurfyx:
I believe it's just a problem with the Lexical.d.ts types? Want to submit a quick PR?

PR #1919, opened 2022-04-22
fix: editor theme class name type for list

wobsoriano:
Fixes https://github.com/facebook/lexical/issues/1885

```typescript
type EditorThemeClasses = {
  list?: {...};
  // Handle other generic values
  [key: string]:
    | EditorThemeClassName
    | Record<
        string,
        | EditorThemeClassName
        | Array<EditorThemeClassName>
        | Record<string, EditorThemeClassName>
      >;
}
```

This only handles the current structure for `list` key. If we want it recursive I can update it to

```typescript
type EditorThemeRecursiveClassName = {
  [key: string]: EditorThemeClassName | Array<EditorThemeClassName> | EditorThemeRecursiveClassName
}

type EditorThemeClasses = {
  list?: {...};
  [key: string]: EditorThemeClassName | EditorThemeRecursiveClassName;
}
```

### DIFF
--- a/packages/lexical/Lexical.d.ts
+++ b/packages/lexical/Lexical.d.ts
@@ -213,7 +213,14 @@ export type EditorThemeClasses = {
     h5?: EditorThemeClassName;
   };
   // Handle other generic values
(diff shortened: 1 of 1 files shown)


acywatson:
You might need to ```npm run prettier:fix``` to get that integrity check to pass

PR #1920, opened 2022-04-22
cleanup LexicalNode

davidnguyen7:
Moved `isComposing` to `TextNode` and removed `getTextContentSize` from `LexicalNode`

### DIFF
--- a/packages/lexical-code/src/index.js
+++ b/packages/lexical-code/src/index.js
@@ -571,7 +571,7 @@ function updateAndRetainSelection(
       anchorOffset +
       anchorNode.getPreviousSiblings().reduce((offset, _node) => {
         return (
-          offset + ($isLineBreakNode(_node) ? 0 : _node.getTextContentSize())
+          offset + ($isLineBreakNode(_node) ? 0 : _node.getTextContent().length)
         );
       }, 0);
   }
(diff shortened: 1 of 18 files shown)


zurfyx (commented):
Thanks! The composition change looks good to me but I'm not sure why we're getting rid of `getTextContentSize`

davidnguyen7:
> Thanks! The composition change looks good to me but I'm not sure why we're getting rid of `getTextContentSize`

It's to ensure anything text-related is handled by `TextNode`, and also to prepare to remove `getTextContext` in `LexicalNode`.

acywatson:
> Thanks! The composition change looks good to me but I'm not sure why we're getting rid of `getTextContentSize`

Why would we not get rid of that? Save some bytes, probably. It's trivial to get the length of the text content, which is all this is doing.

trueadm:
@davidnguyen7 

> It's in preparation for the removal of getTextContext in LexicalNode, to ensure anything text-related is handled by TextNode.

FWIW, `getTextContent()` is not exclusive to text nodes. DecoratorNode's can also specific content, as can LineBreakNodes and also ElementNodes (although maybe element nodes can just extract from their children).

acywatson:
@davidnguyen7 

Thanks again for doing this - we actually belatedly realized that this is a breaking change and that getTextContentSize actually serves as a useful abstraction that we may want to keep around in the interest of maintaining a consistent cross-platform interface (think native iOS/Android).

We're going to revert this one for now and consider it a bit more as a team.

Issue #1726, opened 2022-04-14
Feature: Add descriptive element titles for better accessibility


WeaverOfTheWeb:
Icons aren't always the best for describing a feature/function so it would be helpful to have titles added to each interactive element describing it's functionality.

![Screenshot 2022-04-14 145749](https://user-images.githubusercontent.com/1872377/163406637-dea9db71-7cac-407e-97c8-18777513ae59.png)

acywatson:
I agree with this, in principle, but there are some accessibility gotchas with using just the title attribute. I think we might need to go a bit deeper, as described here (linked from MDN docs where I was refreshing on title attr):

https://inclusive-components.design/tooltips-toggletips/

In any case, even just the title attribute would probably be an improvement over what's there now.

furstenberg:
Yes, as already mentioned there are issues with using the title attribute. 

MDN page for title attribute has some [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns) listed. 

The title attribute is also mentioned in the [A11Y Project checklist](https://www.a11yproject.com/checklist/):

> The title attribute has numerous issues, and should not be used if the information provided is important for all people to access. An acceptable use for the title attribute would be labeling an iframe element to indicate what content it contains.

thegreatercurve:
Fixed by #1932

PR #1924, opened 2022-04-22
Add title to toolbar plugins

always-maap:
add a title with a shortcut to toolbar plugins

Screen Shot 1401-02-02 at 14 13 11

### DIFF
--- a/packages/lexical-playground/src/utils/toolbar-shortcut.js
+++ b/packages/lexical-playground/src/utils/toolbar-shortcut.js
@@ -1,9 +1,0 @@
-const isApple = /(Mac|iPhone|iPad)/i.test(navigator.platform);
-
-export const toolbarShortcut = {
-  bold: isApple ? 'Bold (⌘B)' : 'Bold (Ctrl+B)',
-  italic: isApple ? 'Italic (⌘I)' : 'Italic (Ctrl+I)',
-  redo: isApple ? 'Redo (⌘Y)' : 'Undo (Ctrl+Y)',
-  underline: isApple ? 'Underline (⌘U)' : 'Underline (Ctrl+U)',
-  undo: isApple ? 'Undo (⌘Z)' : 'Undo (Ctrl+Z)',
-};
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/utils/toolbar-shortcut.js
```diff
@@ -0,0 +1,9 @@
+const isApple = /(Mac|iPhone|iPad)/i.test(navigator.platform);
```
thegreatercurve, line 1: We have a utility called `IS_APPLE` in `packages/shared`  which does the same thing.

packages/lexical-playground/src/utils/toolbar-shortcut.js
```diff
@@ -0,0 +1,9 @@
...
+
+export const toolbarShortcut = {
+  bold: isApple ? 'Bold (⌘B)' : 'Bold (Ctrl+B)',
+  italic: isApple ? 'Italic (⌘I)' : 'Italic (Ctrl+I)',
+  redo: isApple ? 'Redo (⌘Y)' : 'Undo (Ctrl+Y)',
+  underline: isApple ? 'Underline (⌘U)' : 'Underline (Ctrl+U)',
+  undo: isApple ? 'Undo (⌘Z)' : 'Undo (Ctrl+Z)',
+};
```
thegreatercurve, line 9: Can we put these inline as props in `packages/lexical-playground/src/plugins/ToolbarPlugin.jsx`? 

There's probably not much point in abstracting them into a utils file if we're not reusing them elsewhere.

thegreatercurve (commented):
Thanks for spotting and making these changes @always-maap! LGTM but just left a couple of comments.

always-maap:
> Thanks for spotting and making these changes @always-maap! LGTM but just left a couple of comments.

fixed your comments

zurfyx (approved):
Thanks!

Just noticed @thegreatercurve's comments, want to have another pass yourself and merge?

thegreatercurve (approved):
Thanks @always-maap!

acywatson:
Overall, this looks like an improvement, but I was hoping we could make it accessible. See the thread here:

https://github.com/facebook/lexical/issues/1726

@zurfyx @thegreatercurve

PR #1930, opened 2022-04-22
Adds github link to lexical playground

harish-sethuraman:
May be could have added it inside the react app 😛  But I thought adding it to the HTML is faster and renders way before the playground even mounts


fantactuka:
LGTM, although I kinda liked clean index.html :)

fantactuka (approved):

trueadm:
Maybe we can move this to being in the App.js rather than adding to index.html?

harish-sethuraman:
bump

PR #1931, opened 2022-04-22
fix popup Position when text is selected

michael-fasst:
Before the popup was hiding the text =>

https://user-images.githubusercontent.com/93589687/164795677-5a764bf1-3442-4f44-8c28-a6c908e48823.mov

Afters changes => 

https://user-images.githubusercontent.com/93589687/164795745-6be76537-d0a3-489d-8052-c6d3055e794b.mov


trueadm (approved):
Much better!

PR #1932, opened 2022-04-22
Add aria-labels to toolbar and footer bar.

thegreatercurve:
Add descriptive `aria-labels` to all buttons in the toolbar and in the actions footer bar, which just show an icon.  

This should be used as the default button label by screen readers, overwriting the shortened `title` attribute introduced in #1924, 

The `title` attribute also has some accessibility shortcomings, highlighted in #1726.

### DIFF
--- a/packages/lexical-playground/src/plugins/ActionsPlugin.jsx
+++ b/packages/lexical-playground/src/plugins/ActionsPlugin.jsx
@@ -190,8 +190,8 @@ export default function ActionsPlugin({
             connected ? 'Disconnect' : 'Connect'
           } Collaborative Editing`}
           aria-label={`${
-            connected ? 'Disconnect' : 'Connect'
-          } from a collaborative editing server`}>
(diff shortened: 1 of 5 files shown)


trueadm (approved):
Looks good to me.

packages/lexical-playground/src/plugins/ActionsPlugin.jsx
```diff
@@ -126,16 +126,18 @@ export default function ActionsPlugin({
             'action-button action-button-mic ' +
             (isSpeechToText ? 'active' : '')
           }
-          title="Mic"
-          aria-label="Mic">
+          title="Speect To Text"
```
trueadm, line 129: Huge a11y win.

bouiboui: Shouldn't this be "Speech"?
```suggestion
          title="Speech To Text"
```

Issue #1938, opened 2022-04-23
Slow Performance in Safari


jclem:
In Safari, Lexical performance becomes unacceptably slow in as little as ~400 characters in a plaintext document.

This [this CodeSandbox](https://codesandbox.io/s/admiring-flower-16f01b?file=/src/Editor.js) as an example. In just a few hundred characters, the latency on my M1 MacBook Air makes the editor difficult to use.

This sandbox is using:

- Lexical 0.2.4
- React 18

I've been testing with this short document:

```
1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890

1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890

1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890

1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
```

merelinguist:
I can reproduce this on 2019 MacBook Air in Safari 15.4. The latency is especially obvious to me when adding linebreaks. There's a noticeable delay between the character being added and the cursor jumping. This doesn't seem to be a problem in Chrome.

trueadm:
Make sure you disable rendering the debug tree view component, it can add unnecessary lag.

fantactuka:
Hi @merelinguist do you mind checking if `0.2.5` fixed newlines insertion latency

jclem:
Update on the original issue: The latency, as I was beginning to suspect, appears to be in paints. In a small browser window, input latency is acceptable. In a large window when lots of text is visible, latency increases sharply.

trueadm:
Do you observe the same issues with other content editable based text editors? Could it be due to some CSS or feature on the Lexical playground?

jclem:
> Do you observe the same issues with other content editable based text editors? Could it be due to some CSS or feature on the Lexical playground?

Tested in ProseMirror on their homepage demo. Whitespace entry (including new lines) does not cause a full editor paint. Only the line being edited tends to be repainted.

trueadm:
@jclem It's likely to do with something on the playground then. We have some plugins that do a `getBoundingClientRect` on selection changing. That could be part of the reason, or it could be some CSS property. Does this occur for you on this example? Have you tried hiding the debug panel (it's in the settings tab, bottom left of screen).

You could also try another example with Lexical, such as: https://codesandbox.io/s/lexical-plain-text-example-g932e

jclem:
[...]

PR #1943, opened 2022-04-24
Fix bugs around pressing the enter key in Safari

trueadm:
This should fix some of the issues with Safari in https://github.com/facebook/lexical/issues/1938. Specifically, calling `preventDefault` on the enter key causes a big pause in Safari. Given we already have a codepath for iOS, we can re-use it for Safari on desktop too. I also noticed that pressing enter that Safari was not scrolling correctly to the new content – so this also fixes that.

### DIFF
--- a/packages/lexical/src/LexicalReconciler.js
+++ b/packages/lexical/src/LexicalReconciler.js
@@ -828,12 +828,17 @@ function reconcileSelection(
   const nextFocusOffset = focus.offset;
   let nextAnchorNode = anchorDOM;
   let nextFocusNode = focusDOM;
+  let skipNativeSelectionDiff = false;
 
   if (anchor.type === 'text') {
     nextAnchorNode = getDOMTextNode(anchorDOM);
+  } else {
+    skipNativeSelectionDiff = true;
   }
   if (focus.type === 'text') {
     nextFocusNode = getDOMTextNode(focusDOM);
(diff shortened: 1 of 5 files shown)


packages/lexical/src/LexicalReconciler.js
```diff
@@ -842,8 +847,11 @@ function reconcileSelection(
   }
 
   // Diff against the native DOM selection to ensure we don't do
-  // an unnecessary selection update.
+  // an unnecessary selection update. We also skip this check if
+  // we're moving selection to within an element, as this can
+  // sometimes be problematic around scrolling.
   if (
+    !skipNativeSelectionDiff &&
```
zurfyx, line 854: What about a selection that starts on a linebreak and ends on a text? Is this the intended behavior too? I'm also not sure about the implications on elements

zurfyx (approved):

packages/lexical/src/LexicalReconciler.js
trueadm: There's no harm in reconciling selection. We avoid it for performance reasons, but then we don't get scrolling :/

trueadm: Couldn't get it working with collab anyway, so commenting out for now. Will address in a follow up.

PR #1945, opened 2022-04-24
added superscript and subscript buttons in playground

animeshk874:
Adds superscript and subscript buttons in the playground. This is an extension of this PR -> https://github.com/facebook/lexical/pull/1903

Screenshot 2022-04-24 at 6 14 28 PM


packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -71,6 +71,14 @@
 .PlaygroundEditorTheme__textUnderlineStrikethrough {
   text-decoration: underline line-through;
 }
+.PlaygroundEditorTheme__textSubscript {
+  font-size: 12px;
+  vertical-align: sub !important;
```
acywatson, line 76: why was this necessary? I'm mostly just curious.

acywatson (approved):
I don't have a problem with this - some others might have opinions on how the toolbar real estate should be used, so will let this bake for a little while and get some other eyes on it.

trueadm:
Can these move into another dropdown or something? They're not common enough to use up so much space on mobile.

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
animeshk874: Just in case both `sub` and `sup` buttons are clicked, this will make sure that sub gets the priority. That's how it's handled [here](https://github.com/facebook/lexical/pull/1903/files)

animeshk874:
> Can these move into another dropdown or something? They're not common enough to use up so much space on mobile.

Makes sense. Will do that!

animeshk874:
> I don't have a problem with this - some others might have opinions on how the toolbar real estate should be used, so will let this bake for a little while and get some other eyes on it.

Sure thing. :)

animeshk874:
> > Can these move into another dropdown or something? They're not common enough to use up so much space on mobile.
> 
> Makes sense. Will do that!

@trueadm Should I merge `Strikethrough`, `Sub`, `Sup` and `Link` into a new dropdown and call it `More`? Because creating a new dropdown just for `Sub` and `Sup` will still take the same amount of space as it is taking now

trueadm:
I’d maybe keep link separate but you could collapse the others. :)

trueadm:
Screenshot 2022-04-26 at 00 13 48

@animeshk874 I wonder if there is a better icon to use? The three dots don't come over as formatting things? It's a minor nit, but I think we should make the toolbar as awesome as possible! :)

animeshk874:
> Screenshot 2022-04-26 at 00 13 48
> 
> @animeshk874 I wonder if there is a better icon to use? The three dots don't come over as formatting things? It's a minor nit, but I think we should make the toolbar as awesome as possible! :)

Hi @trueadm! I see the problem here. I will update the icon soon. I found this icon that I think makes more sense that the 3 dots.
Screenshot 2022-04-26 at 11 37 35 AM

The problem with this is that it's used for `Case transforms (uppercase, lower case etc.)`,  but here we don't have that option. So maybe adding this icon would mean introducing text transforms in lexical?
p.s - just in case we have plans of introducing text transform to Lexical, I would be happy to work on it

trueadm:
I'd be fine with that, I think. We can always change it in the future.

acywatson:
@animeshk874 I'm not sure why these tests are failing - maybe try rebasing onto main?

animeshk874:
> @animeshk874 I'm not sure why these tests are failing - maybe try rebasing onto main?

Have merged `main` into this branch. If the tests still fail, I will look into it.

animeshk874:
@acywatson Have fixed the failing test. It was happening because the `strikethrough` button was moved inside a dropdown.

trueadm:
I think we can drop the "More" Text from the button, right? I mean "More" Doesn't really mean anything contextually, you could argue all the other dropdowns are also "More".

animeshk874:
> I think we can drop the "More" Text from the button, right? I mean "More" Doesn't really mean anything contextually, you could argue all the other dropdowns are also "More".

I had kept it there for consistency as all the other dropdown buttons have a text. But if you think that would be confusing, then I can remove it.

animeshk874:
@trueadm have removed the `More` text. This is how it looks right now:
Screenshot 2022-04-29 at 10 27 48 AM

acywatson (approved):

PR #1949, opened 2022-04-24
Separate clean script

acywatson:
I went down a small rabbit hole on this. Wanted to add removing npm folders to the clean, then ended up decided I could use the global packages list to do that, then decided that slightly coupled the build to the npm release (which doesn't matter at all probably, in practice), then decided to make a separate script so we could clean without building, if desired.

So, here we are.

Mostly a waste of time, but probably not a bad idea to break the conditional logic out of this build script eventually, all things considered.


package.json
```diff
@@ -16,12 +16,12 @@
     "start": "NODE_ENV=development concurrently \"npm:collab\" \"npm run dev --prefix packages/lexical-playground\"",
     "start:website": "npm run start --prefix packages/lexical-website-new -- --port 3001",
     "dev": "npm run dev --prefix packages/lexical-playground",
-    "build": "node scripts/build.js --clean",
-    "build-prod": "node scripts/build.js --prod --clean",
-    "build-release": "node scripts/build.js --prod --release --clean",
-    "build-www": "node scripts/build.js --clean --www && node scripts/build.js --www --prod && npm run prepare-www",
+    "build": "npm run clean && node scripts/build.js",
+    "build-prod": "npm run clean && node scripts/build.js --prod",
+    "build-release": "npm run clean && node scripts/build.js --prod --release",
```
harish-sethuraman, line 21: It would actually be nice we we could re use the existing commands?
```suggestion
    "build": "npm run clean && node scripts/build.js",
    "build-prod": "npm run build --prod",
    "build-release": "npm run build-prod --release",
```

acywatson: I thought about this, but I think it's a little more complicated. I don't think the --prod and --release arguments are passed through to the build script here with the construction you're suggesting.

acywatson: it's actually not as complicated as I thought - just have to do the standard -- separator to separate the npm args from the ones to the actual script. I wasn't sure if the arguments would be passed to the second script after the && operator.

trueadm (approved):

package.json
harish-sethuraman: > I thought about this, but I think it's a little more complicated.

Okies


> I don't think the --prod and --release arguments are passed through to the build script here with the construction you're suggesting.

Not sure why they aren't working here until we actually make them something like `-- --prod`

You can resolve this conversation

package.json
```diff
@@ -17,12 +17,12 @@
     "start:website": "npm run start --prefix packages/lexical-website-new -- --port 3001",
     "dev": "npm run dev --prefix packages/lexical-playground",
     "build": "npm run clean && node scripts/build.js",
-    "build-prod": "npm run clean && node scripts/build.js --prod",
-    "build-release": "npm run clean && node scripts/build.js --prod --release",
-    "build-www": "npm run clean && node scripts/build.js --www && node scripts/build.js --www --prod && npm run prepare-www",
+    "build-prod": "npm run build -- --prod",
+    "build-release": "npm run clean && npm run build-prod -- --release",
+    "build-www": "npm run clean && npm run build -- --www && npm run build -- --www --prod && npm run prepare-www",
```
harish-sethuraman, line 22: we might not need the `npm run clean` here?

package.json
acywatson: They might be different ways of achieving similar ends. You can read more about passing args to npm scripts here:

https://docs.npmjs.com/cli/v8/commands/npm-run-script#description

It may be "--" prefixed args vs non that makes the difference in how they are parsed.

package.json
```diff
@@ -16,13 +16,13 @@
...
     "start:website": "npm run start --prefix packages/lexical-website-new -- --port 3001",
     "dev": "npm run dev --prefix packages/lexical-playground",
-    "build": "node scripts/build.js --clean",
-    "build-prod": "node scripts/build.js --prod --clean",
-    "build-release": "node scripts/build.js --prod --release --clean",
-    "build-www": "node scripts/build.js --clean --www && node scripts/build.js --www --prod && npm run prepare-www",
+    "build": "npm run clean && node scripts/build.js",
+    "build-prod": "npm run build -- --prod",
+    "build-release": "npm run clean && npm run build-prod -- --release",
+    "build-www": "npm run clean && npm run build -- --www && npm run build -- --www --prod && npm run prepare-www",
```
harish-sethuraman, line 22:
```suggestion
    "build-release": "npm run build-prod -- --release",
    "build-www": "npm run build -- --www && npm run build -- --www --prod && npm run prepare-www",
```

harish-sethuraman: Did them for you 😅  Sorry for the mess though

PR #1950, opened 2022-04-24
Improve typing perf accuracy

trueadm:
This should change the measurement heuristics in cases where UI changes might occur in selectionchange, before the setTimeout fires. This causes the toolbar UI to be included in typing metrics, which isn’t correct.


acywatson (commented):
Can I ask you to add a small amount of context to the PR description for clarity/posterity?

zurfyx (approved):

PR #1952, opened 2022-04-24
Revert "cleanup LexicalNode (#1920)"

acywatson:
As discussed here: https://github.com/facebook/lexical/pull/1920

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.js
+++ b/packages/lexical/src/nodes/LexicalTextNode.js
@@ -323,10 +323,6 @@ export class TextNode extends LexicalNode {
     return self.__style;
   }
 
-  isComposing(): boolean {
-    return this.__key === $getCompositionKey();
-  }
-
   isToken(): boolean {
     const self = this.getLatest();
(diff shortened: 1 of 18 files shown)


thegreatercurve (approved):

trueadm (approved):

packages/lexical/src/nodes/LexicalTextNode.js
```diff
@@ -323,10 +323,6 @@ export class TextNode extends LexicalNode {
     return self.__style;
   }
 
-  isComposing(): boolean {
-    return this.__key === $getCompositionKey();
-  }
-
```
zurfyx, line 329: Do we want to keep the `isComposing` changes? Or maybe on a separate PR?

zurfyx (approved):

packages/lexical/src/nodes/LexicalTextNode.js
trueadm: Separate PR is fine imo.

Issue #1954, opened 2022-04-24
Bug: Can't center align an empty paragraph


acywatson:
Lexical version:

## Steps To Reproduce

1. type some text into an empty editor
2. center align the text
3.  delete the text
4. try to left align
5. Observe that paragraph is still center-aligned (cursor is in middle).

Link to code example:  N/A

## The current behavior

Cursor stays in the center

## The expected behavior

Cursor moves to the left.

PR #1955, opened 2022-04-24
Fix block align bug

trueadm:
Fixes https://github.com/facebook/lexical/issues/1954.


thegreatercurve (approved):

trueadm:
Come to think about it, this isn’t the right fix. getNodes should return the element if it’s one of the points, and the element has no children

PR #1959, opened 2022-04-25
Fix the types definition for LexicalOnChangePlugin

Proful:


thegreatercurve (approved):
Thanks!

zurfyx (commented):
Thanks! Can we fix LexicalOnChangePlugin.js.flow too please?

zurfyx (approved):

Issue #1953, opened 2022-04-24
Enhancement: We shouldn't be initializing skipped tests.


acywatson:
Currently, we run the initialize util, which navigates to the local playground, in a beforeEach. The logic for skipping tests is inside the test itself. We need to to reorganize this somehow such that if we know a test will be skipped, we don't pay the cost of actually initializing.

There are several ways to think about this, the most straightforward probably being just moving the initialize inside the tests themselves and checking the skip condition before. For suites that are all skipped, we can probably move the skip to the beforeEach.

Example: https://github.com/facebook/lexical/blob/3b70199ea4a83e809ac3f57d8c62067b5b15a34c/packages/lexical-playground/__tests__/e2e/File.spec.mjs#L28

trueadm:
I don't believe we do this anymore! :)

PR #1960, opened 2022-04-25
Optimize E2E & regression tests

davidnguyen7:
Related to #1953

Moves skip condition checking from inside to before the tests are run to prevent unnecessary initialization.


packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -21,8 +21,8 @@ export const IS_WINDOWS = process.platform === 'win32';
 export const IS_LINUX = !IS_MAC && !IS_WINDOWS;
 export const IS_COLLAB =
   process.env.E2E_EDITOR_MODE === 'rich-text-with-collab';
-const IS_RICH_TEXT = process.env.E2E_EDITOR_MODE !== 'plain-text';
-const IS_PLAIN_TEXT = process.env.E2E_EDITOR_MODE === 'plain-text';
+export const IS_RICH_TEXT = process.env.E2E_EDITOR_MODE !== 'plain-text';
```
zurfyx, line 24: Should we delete `appSettings.isRichText = ` in L36?

zurfyx (commented):
Thanks! The constant part LGTM but I'm not sure about moving skip to `describe`, it makes less obvious that it the test is being skipped. While I think it's very opinionated, I'd rather have the verbose version

packages/lexical-playground/__tests__/utils/index.mjs
davidnguyen7: It'd still need to be there, seeing as how removing it would cause the app to default to the rich-text editor, breaking plain-text testing.

davidnguyen7:
Would moving skip to outside describe as shown below be more understandable at a glance?
```js
test.skip(IS_PLAIN_TEXT);

test.describe('foo', () => ...)
```

zurfyx:
> Would moving skip to outside describe as shown below be more understandable at a glance?
> 
> ```js
> test.skip(IS_PLAIN_TEXT);
> 
> test.describe('foo', () => ...)
> ```

The thing is that new users who are adding new tests probably won't read any further than their own test so this kind potentially cause mass unintended skips

packages/lexical-playground/__tests__/utils/index.mjs
zurfyx: You're right, I misread that

acywatson:
> > Would moving skip to outside describe as shown below be more understandable at a glance?
> > ```js
> > test.skip(IS_PLAIN_TEXT);
> > 
> > test.describe('foo', () => ...)
> > ```
> 
> The thing is that new users who are adding new tests probably won't read any further than their own test so this kind potentially cause mass unintended skips
> 
> Since this is very opinionated, we can also leave this PR as-is for a day or two to see if other members in the team have any opinions on this

I agree with having the move verbose version. I think we should probably run test.skip in each test case, then run initialize in after it (if not skipped). This is a bit more verbose, but make it crystal clear what is happening for each test.

davidnguyen7:
Makes sense to me. I'll close this PR and submit a new one once I come up with a better approach.

PR #1963, opened 2022-04-25
Fix formatting on embeds

trueadm:
Embed nodes do not correctly update. It turned out that the format logic was missing from the constructor and the clone methods, which was why it was never being kept on the node.

### DIFF
--- a/packages/lexical-react/src/LexicalDecoratorBlockNode.js
+++ b/packages/lexical-react/src/LexicalDecoratorBlockNode.js
@@ -7,12 +7,17 @@
  * @flow strict
  */
 
-import type {ElementFormatType, LexicalNode} from 'lexical';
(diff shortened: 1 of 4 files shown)


packages/lexical-react/src/LexicalDecoratorBlockNode.js
```diff
@@ -27,11 +32,6 @@ export class DecoratorBlockNode extends DecoratorNode<React$Node> {
     self.__format = format;
   }
 }
-
-export function $createDecoratorBlockNode(): DecoratorBlockNode {
```
trueadm, line 31: I couldn't see why anyone would create a decorator block directly here, so I removed it.

zurfyx (approved):

Issue #1966, opened 2022-04-25
Bug: 


Fetz:
Lexical version: 0.2.4

## Steps To Reproduce

1. Go to https://playground.lexical.dev/
2. Copy a lorem ipsum paragraph
> Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam venenatis risus ac cursus efficitur. Cras efficitur magna odio, lacinia posuere mauris placerat in. Etiam eu congue nisl. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Nulla vulputate justo id eros convallis, vel pellentesque orci hendrerit. Pellentesque accumsan molestie eros, vitae tempor nisl semper sit amet. Sed vulputate leo dolor, et bibendum quam feugiat eget. Praesent vestibulum libero sed enim ornare, in consequat dui posuere. Maecenas ornare vestibulum felis, non elementum urna imperdiet sit amet.
3. Paste and make it a list item
4. Paste and make a normal paragraph
5. Do a selection in both the paragraph & in the list item
    - Both scenarios the browser selection indicator and lexical debugger selection will not match
7. Do a selection in a word in the middle of a list item and paste and a new line, the pasted content will not match the selection.

https://user-images.githubusercontent.com/1071113/165082457-9022c6cf-21e5-4099-84cd-0764b6b0b8cc.mov

## The current behavior
- Debugger Lexical selection is not matching what is selected by the selection indicator.
- Selecting a text in a middle of list item, copy and pasting is not matching the selection. 

## The expected behavior
- Debugger Lexical selection should match what is selected by the selection indicator.
- Selecting a text in a middle of list item, copy and pasting is not matching the selection.

Fetz:
@trueadm will your PR: https://github.com/facebook/lexical/pull/1962 address this?

trueadm:
> Debugger Lexical selection should match what is selected by the selection indicator.

Can this be made into a separate issue please?

Fetz:
> Debugger Lexical selection should match what is selected by the selection indicator.

@trueadm created the following issue #1970 to track this.

PR #1968, opened 2022-04-25
Fix equation node handling on Android

trueadm:
Fixes https://github.com/facebook/lexical/issues/1966.


StormVanDerPol:
I checked the preview, the case where you insert only an "equation" node and nothing else seems to be fixed, but it looks like inserting some text, _then_ inserting an equation node, and finally hitting backspace still breaks.

I apologise, the reproduction steps in the issue I created seem to not catch all cases where this bug happens

zurfyx (approved):

PR #1971, opened 2022-04-25
CodeHighlight setFormat do not mark dirty

zurfyx:
If we're not changing node values, let's just not mark the node dirty which is more expensive? #1966


trueadm:
What's this even used for?

fantactuka:
Should not be used anyways and less code 👍

fantactuka (approved):

Proful:
Any thoughts on making it work instead of blocking (b/i/u etc).

It was working on Draft js and quite handy. Please think to provide option.

Issue #1981, opened 2022-04-25
Bug: When we select few lines, indent and outdent feature only applied to the last line.


yuuminakamura:
When we select few lines, indent and outdent feature only applied to the last line.

Lexical version: latest

![Screen Shot 2022-04-25 at 1 05 07 PM](https://user-images.githubusercontent.com/46723838/165166886-78285f4e-f114-44b7-a34f-81a81e4dee79.png)

## Steps To Reproduce

1. Select the list of words
2. Use indent / outdent feature

```
                <span
                    onClick={() => {
                        editor.dispatchCommand(INDENT_CONTENT_COMMAND);
                    }}
                    className={'toolbar-item spaced '}
                    aria-label="Format Indent"
                >
                    <FormatIndentIncreaseIcon />
                </span>
                <span
                    onClick={() => {
                        editor.dispatchCommand(OUTDENT_CONTENT_COMMAND);
                    }}
                    className={'toolbar-item spaced '}
                    aria-label="Format Outdent"
                >
                    <FormatIndentDecreaseIcon />
                </span>

```

PR #1982, opened 2022-04-25
Improve multi element indentation

trueadm:
Fixes https://github.com/facebook/lexical/issues/1981. Introduces, `ElementNode.canIndent()` method to help assist in getting the right behavior for specific elements.

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.js
+++ b/packages/lexical-list/src/LexicalListItemNode.js
@@ -227,6 +227,10 @@ export class ListItemNode extends ElementNode {
     return this;
   }
 
+  canIndent(): false {
+    return false;
+  }
+
   insertBefore(nodeToInsert: LexicalNode): LexicalNode {
     const siblings = this.getNextSiblings();
     if ($isListItemNode(nodeToInsert)) {
--- a/packages/lexical-list/src/LexicalListItemNode.js
+++ b/packages/lexical-list/src/LexicalListItemNode.js
@@ -228,6 +228,7 @@ export class ListItemNode extends ElementNode {
   }
 
   canIndent(): false {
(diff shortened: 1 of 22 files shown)


packages/lexical-rich-text/src/index.js
```diff
@@ -328,6 +329,34 @@ function onCutForRichText(event: ClipboardEvent, editor: LexicalEditor): void {
   });
 }
 
+function handleIndentAndOutdent(
+  editor: LexicalEditor,
```
fantactuka, line 333: Probably don't need editor param

packages/lexical-rich-text/src/index.js
```diff
@@ -478,50 +507,41 @@ export function registerRichText(
...
+            if ($isTextNode(node)) {
+              const textContent = node.getTextContent();
+              const character = textContent[textContent.length - 1];
+              if (character === '\t') {
+                editor.dispatchCommand(DELETE_CHARACTER_COMMAND, true);
+              }
+            }
+          },
+          (block) => {
+            if (block.getIndent() !== 0) {
```
fantactuka, line 540: It's better to extract into variable, as for lists it actually goes up in a tree to check nesting level

packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -227,6 +227,10 @@ export class ListItemNode extends ElementNode {
     return this;
   }
 
+  canIndent(): false {
+    return false;
```
fantactuka, line 231: Maybe worth adding comment that indent/outdent for lists handled separately from default logic (hence canIndent: false) as otherwise it looks confusing if haven't seen formatList first

fantactuka (approved):

PR #1983, opened 2022-04-25
Add forEachChild to ElementNode

trueadm:
This is a nice utility function that improves the computation without needing a second O(n) traversal of the element's children.

### DIFF
--- a/packages/lexical/src/nodes/LexicalElementNode.js
+++ b/packages/lexical/src/nodes/LexicalElementNode.js
@@ -53,11 +53,9 @@ export class ElementNode extends LexicalNode {
     return self.__indent;
   }
   forEachChild(cb: (node: LexicalNode, index: number) => void): void {
-    for (
-      let i = 0, children;
-      i < (children = this.getLatest().__children).length;
-      i++
-    ) {
+    const self = this.getLatest();
+    const children = self.__children;
+    for (let i = 0; i < children.length; i++) {
       const childNode = $getNodeByKey<LexicalNode>(children[i]);
       if (childNode !== null) {
         cb(childNode, i);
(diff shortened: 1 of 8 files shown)


fantactuka:
From offline discussion: it can be called during update and `__children` list will be mutated while looping through it:
```js
editor.update(() => {
  element.forEachChild(child => child.insertAfter($createTextNode()))
})
```

Maybe keep it as internal util and only call for cases when iteration does not mutate children list?

trueadm:
@fantactuka I reverted the logic because it caused issues. In fact, the change was actually not correct. According to the spec of `Array.forEach`:

> The range of elements processed by forEach is set before the first call to callbackfn. Elements which are appended to the array after the call to forEach begins will not be visited by callbackfn.

So, as soon as you call `someArray.forEach`, the elements in that array do not grow.

packages/lexical/src/nodes/LexicalElementNode.js
```diff
@@ -52,6 +52,16 @@ export class ElementNode extends LexicalNode {
     const self = this.getLatest();
     return self.__indent;
   }
+  forEachChild(cb: (node: LexicalNode, index: number) => void): void {
+    const self = this.getLatest();
+    const children = self.__children;
+    for (let i = 0; i < children.length; i++) {
```
zurfyx, line 58: This doesn't guarantee that all the children will be visited if the children change while the traversal is ongoing, will it?

zurfyx (approved):
No strong opinions on this PR, I guess it's similar to `getChildrenSize`

trueadm:
Decided against in the end due to the concerns around the live-ness of the children. If you mutate children part way through, it would differ from array.forEach, meaning we'd have to create a clone – undoing the perf wins.

PR #1984, opened 2022-04-26
Fix collab bugs

trueadm:
This PR fixes collaboration issues around lists. Specifically, how the value of list items is handled. For some reason, we never put it on the node as a property, which mean it could never be syncronized across clients. :/

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.js
+++ b/packages/lexical-list/src/LexicalListItemNode.js
@@ -31,26 +31,35 @@ import {
 import invariant from 'shared/invariant';
 
 import {$createListNode, $isListNode} from './';
-import {$handleIndent, $handleOutdent} from './formatList';
+import {
+  $handleIndent,
+  $handleOutdent,
+  updateChildrenListItemValue,
+} from './formatList';
 
 export class ListItemNode extends ElementNode {
+  __value: number;
+
   static getType(): string {
     return 'listitem';
   }
 
   static clone(node: ListItemNode): ListItemNode {
-    return new ListItemNode(node.__key);
+    return new ListItemNode(node.__value, node.__key);
   }
 
-  constructor(key?: NodeKey): void {
+  constructor(value?: number, key?: NodeKey): void {
     super(key);
+    this.__value = value === undefined ? 1 : value;
   }
-
-  // View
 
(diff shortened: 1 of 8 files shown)


packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -1249,155 +1238,159 @@ test.describe('Tables', () => {
...
-              <p class="PlaygroundEditorTheme__paragraph"><br /></p>
-            </td>
-          </tr>
-        </table>
-        <p class="PlaygroundEditorTheme__paragraph"><br /></p>
-      `,
-      {ignoreClasses: true},
-    );
+    // Selection of cells is not synced in collab, but we still want to
+    // ensure this doesn't break collab too.
```
fantactuka, line 1242: Can add `ignoreInlineStyles` next to `ignoreClasses` to strip "style" attributes for html assert

trueadm: I tried `ignoreInlineStyles` but it didn't seem to have any effect on these tests. Who knows :/

fantactuka (approved):

packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -233,10 +268,12 @@ export class ListItemNode extends ElementNode {
   }
 
   insertBefore(nodeToInsert: LexicalNode): LexicalNode {
-    const siblings = this.getNextSiblings();
     if ($isListItemNode(nodeToInsert)) {
-      // mark subsequent list items dirty so we update their value attribute.
-      siblings.forEach((sibling) => sibling.markDirty());
+      const parent = this.getParentOrThrow();
+      if ($isListNode(parent)) {
+        // mark subsequent list items dirty so we update their value attribute.
```
acywatson, line 274: Comment is no longer valid.

Also, looks like we used to update values just for subsequent nodes (next siblings) since anything prior shouldn’t change value. Now we update all children every time. Is that what we want?

trueadm: Addressed feedback. It now only does next siblings now.

PR #1985, opened 2022-04-26
feat: drop down keyboard navigation

### LINKED ISSUE #1923: Bug: Toolbar Keyboard Accessibility
Lexical version: The one on [the playground site](https://playground.lexical.dev/)

## Steps To Reproduce

1. Focus the toolbar’s “Formatting Options” dropdown
2. Activate the “dropdown” using space bar
3. Try to use the arrow keys to select an option from that list
4. Tab away to the next control
5. Keep on tabbing until you go past the last control of the toolbar
6. Hit shift+tab to try and go back to the toolbar

## The current behavior

At step 3: It's a custom select that doesn't respond to the up/down arrow keys to choose an option, so it won’t work
At step 5: The editor is focused

## The expected behavior

At step 3: For the arrow keys to work _(they don't)_
At step 4: For the custom select to disappear _(it remains visible)_
At step 5: (?) To not go into the editor, but make a round trip to the first control of the toolbar
At step 6: To get the focus back into the toolbar (or, alternatively: provide a shortcut to toggle focus between the toolbar and the editor field?)

ivan-ngchakming:
Added keyboard navigation feature to drop down UI component

- Set focus to first drop down item on open
- Allow navigation using `Tab`
- Allow navigation using arrow keys (`ArrowUp`, `ArrowDown`)
- Allow closing drop down using `Escape` and return focus to the drop down button

closes #1923

### Screens

tosDemo


packages/lexical-playground/src/ui/DropDown.jsx
```diff
@@ -7,11 +7,118 @@
...
+      setHighlightedItem((prev) => {
+        const index = items.indexOf(prev) - 1;
+        return items[index === -1 ? items.length - 1 : index];
+      });
+    } else if (key === 'ArrowDown' || key === 'Tab') {
+      setHighlightedItem((prev) => items[items.indexOf(prev) + 1]);
+    }
+  };
+
+  const contextValue = {
```
trueadm, line 102: Should be memoized to prevent React doing extra context updates.

zurfyx: @ivan-ngchakming - `useMemo` will preserve the same object reference which means that Context.Provider will see as the same exact value and won't trigger further re-renders when that is the case.

Without `useMemo`, every time this components rerenders you're essentially generating a brand new object/value.

always-maap:
I think when the drop-down is open, pressing `Tab` should close and focus on the next element but currently, it cycles between items

acywatson:
> I think when the drop-down is open, pressing `Tab` should close and focus on the next element but currently, it cycles between items

Is that correct? I'm not sure. @a11yHolli can your clarify the correct behavior for us here?

ivan-ngchakming:
> I think when the drop-down is open, pressing `Tab` should close and focus on the next element but currently, it cycles between items 

I believe this would be consistent with the native select. 


I am currently working on my finals, will work on this in a few days as soon as I am free.

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -6,10 +6,118 @@
...
+    [registerItem],
+  );
+
+  useEffect(() => {
+    if (items && !highlightedItem) {
+      setHighlightedItem(items[0]);
+    }
+
+    // $FlowFixMe: I am not sure why this is wrong
+    highlightedItem?.current?.focus();
```
acywatson, line 109: We typically avoid optional chaining - I don't think we have the transpilation in place necessary to support various targets we support. 

Does it work if you do something likee:

```
if (highlightedItem !== null && highlightedItem.current instanceof HTMLElement) {
  highlightedItem.current.focus();
}
```

acywatson (commented):
This looks good to me now - thanks!

packages/lexical-playground/src/ui/DropDown.tsx
trueadm: What if you did:
```
const highlightedItemElem: null | HTMLElement = highlightedItem.current
if (highlightedItemElem !== null) {
  highlightedItemElem.focus();
} 
```

always-maap (commented):
what exactly is the active class doing?
without it the button is not full width

Ordibehesht-30-1401 00-24-22

packages/lexical-playground/src/ui/DropDown.tsx
ivan-ngchakming: I tried out a few ways
```ts
    // works
    if (highlightedItem && highlightedItem.current instanceof HTMLElement) {
      highlightedItem.current.focus();
    }

    // also works, since highlightedItem.current is either `HTMLElement` or `null`
    if (highlightedItem && highlightedItem.current) {
      highlightedItem.current.focus();
    }

    // type error: highlightedItem is possibly 'undefined' in `highlightedItem.current instanceof HTMLElement`
    if (highlightedItem !== null && highlightedItem.current instanceof HTMLElement) {
      highlightedItem.current.focus();
    }

    // type error: highlightedItem is possibly 'undefined'
    const highlightedItemElem: null | HTMLElement = highlightedItem.current
    if (highlightedItemElem !== null) {
      highlightedItemElem.focus();
    }
```

If its all the same, I will go with the second option as its more concise, but still satisfying all the type checks?
```ts
    if (highlightedItem && highlightedItem.current) {
      highlightedItem.current.focus();
    }
```

ivan-ngchakming:
> what exactly is the active class doing? without it the button is not full width

It was behaving like this before this PR if I remember correctly. I am guessing the style that should be applied on it is missing, maybe there should be some kind of icon or indicator to show the style applied on the selected text? But the non-active buttons not being full width issue would still needs to be fixed.

image

acywatson (approved):

PR #1987, opened 2022-04-26
Flow ignore npm folder

zurfyx:
Ignore the autogenerated NPM (release) folder


harish-sethuraman:
Should we do the same for lock files as well? I don't think we have lock files in packages? lock files might get pushed sometimes as well?

fantactuka (approved):

trueadm (approved):

PR #1989, opened 2022-04-26
Persist collapsed formats across selection updates

trueadm:
This better revises formatting that was introduced in https://github.com/facebook/lexical/pull/1962. Specifically, we try to persist selection formats for collapsed selection if we're on the same offset and key within the given threshold.


fantactuka (approved):
It's amazing, fixing long-standing blocker for markdown text formatting as well!

PR #1993, opened 2022-04-26
Fix __cachedText to reflect new lines for ElementNodes

### LINKED ISSUE #1588: ElementNodes text is ignored during reconciliation
**Problem**

In rich text user types a new Paragraph. `registerTextContentListener` should be called and have `\n\n` between paragraphs, currently handled by `elementNode.getTextContent()`. Instead, neither `registerTextContentListener` nor the cached text content has `\n\n`. This works fine when nodes are dirty and cached is ignored.

Turns out the text in ElementNode is ignored during reconciliation. Instead, we're always looking at the children to build the final content.

Even if it wasn't ignored, this would still be a problem with the current approach as the combined text can potentially change with an overwritten `getTextContent`.

**Solution 1**

Fix the reconciler to keep track of text subtrees independently and dump the text to the editorTextContent only at the end of traversing the nodes. Also, consider ElementNode `getTextContent`.

**Solution 2**

Enforce `getTextContent` to be predictable and always constant on ElementNode. ElementNodes `getTextContent` should always return a combination of all the children text (in other words, should never be overwritten). For custom use case, users can implement their own version of text content.

Additional, to cater cases like ParagraphNode, when we need to append `\n\n` at the back we can add an optional `getPostamble` method for ElementNodes that gets treated separately by the reconciler.

zurfyx:
ElementNodes delimiters were never taken into account in the computation of __cachedText. This meant that `getTextContent()` was working properly on a dirty EditorState but not after the update. See #1588 for further details

Closes #1588 ; Addresses the immediate bug pointed out by https://github.com/facebook/lexical/discussions/1980

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "engineStrict": true,
   "scripts": {
-    "start": "NODE_ENV=development concurrently \"npm:collab\" \"npm run dev --prefix packages/lexical-playground\"",
+    "start": "cross-env NODE_ENV=development concurrently \"npm:collab\" \"npm run dev --prefix packages/lexical-playground\"",
     "start:website": "npm run start --prefix packages/lexical-website-new -- --port 3001",
     "dev": "npm run dev --prefix packages/lexical-playground",
     "build": "npm run clean && node scripts/build.js",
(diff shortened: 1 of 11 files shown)


trueadm (approved):
You could also add an e2e test and assert the root text content, as it's exposed on the editor element.

zurfyx:
Created https://github.com/facebook/lexical/issues/2004 to track this (I'll work on this) but merging this now to resolve the immediate issue

PR #1994, opened 2022-04-26
Make the ImageResizer handle more intuitive

### LINKED ISSUE #1974: Feature: Make the ImageResizer handles more intuitive
I would like to be able to drag and resize, for example an SVG, on only one axis, as known from other Vector graphic software or photoshop.

Origin of issue: https://github.com/facebook/lexical/pull/1851#issuecomment-1108310710

100paperkite:
resolve #1974 

- I manually set minHeight, minWidth to 100px.
- It maintains the proportion of the image If dragged on a diagonal button.
- Direction of cursor pointer changes depending on starting drag button 

### Image

(Default maxWidth of image is 500px)

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -838,16 +838,16 @@ select.font-family {
   background-color: rgba(60, 132, 244, 0.5);
 }
 
-.editor-shell .editor-image .image-resizer-sw {
+.editor-shell .editor-image .image-resizer-n {
   display: block;
   width: 7px;
   height: 7px;
   position: absolute;
-  bottom: -2px;
-  left: -6px;
+  top: -6px;
+  left: 48%;
   background-color: rgb(60, 132, 244);
   border: 1px solid #fff;
-  cursor: sw-resize;
+  cursor: n-resize;
 }
 
 .editor-shell .editor-image .image-resizer-ne {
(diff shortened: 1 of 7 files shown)


packages/lexical-playground/src/index.css
```diff
@@ -886,6 +886,54 @@ select.font-family {
...
+  height: 7px;
+  position: absolute;
+  bottom: 48%;
+  left: -6px;
+  background-color: rgb(60, 132, 244);
+  border: 1px solid #fff;
+  cursor: w-resize;
+}
+
+.editor-shell .editor-image .image-resizer-nw {
```
fantactuka, line 925: Could probably use some common class like image-resizer to contain most of these properties. Not critical for a playground but would be nice

packages/lexical-playground/src/ui/ImageResizer.jsx
```diff
@@ -51,26 +64,64 @@ export default function ImageResizer({
...
+    const ew = direction === Direction.east || direction === Direction.west;
+    const ns = direction === Direction.north || direction === Direction.south;
+    const nwse =
+      (direction & Direction.north && direction & Direction.west) ||
+      (direction & Direction.south && direction & Direction.east);
+
+    const cursorDir = ew ? 'ew' : ns ? 'ns' : nwse ? 'nwse' : 'nesw';
+
+    if (editorRootElement !== null) {
+      editorRootElement.style.setProperty(
```
fantactuka, line 90: Maybe even put it on body in case pointer dragged out of editor?

fantactuka (commented):
Not a fan of skewed images 😄 but feature-wise it's a nice one 👍

packages/lexical-playground/src/ui/ImageResizer.jsx
zurfyx: Do we need the two of them? `editorRootElement` and `document.body`?

zurfyx (approved):
Great work! Thank you

Issue #1957, opened 2022-04-24
Bug: Improve Lexical <-> Lexical Copy & Paste For Text Selections


tylerjbainbridge:
As of now `Lexical <-> Lexical` copy + paste clones the entire structure from the top level all the way to the way down to the selected nodes, which is inconsistent with how we build the tree for external (HTML) copy & paste.

Take the following example, if you copy text within a Header you get the following two cases:
`Lexical<->Lexical`: includes the parent header element
`External (HTML)`: only the selected text

https://user-images.githubusercontent.com/13852400/165001643-3d324170-aa07-4d7e-9b0e-b190b8ab4fa2.mov

This is _fine_ for paragraphs, headers, etc, but for deeply nested element trees such as Tables and Lists this is very problematic. Selecting any text within a Table or List will include the the entire tree from the top level all the way down to the text node. 

E.g. A RangeSelection within a TableCell or ListItem will include the following nodes `table -> table row -> table cell -> paragraph -> text`, which essentially is a tiny slice of the table and not expected behavior (see the video below).

https://user-images.githubusercontent.com/13852400/165001783-47a5167f-5542-4fb2-9d27-6411d329ec40.mov

This is what is causing #1916. 

[You can see the source code for `Lexical<->Lexical` copy + paste here](https://github.com/facebook/lexical/blob/3349ffbaf4f6902f8ccb71494b1535129f16d1d4/packages/lexical-selection/src/index.js#L163).

tylerjbainbridge:
I propose that we improve [`$cloneContents`](https://github.com/facebook/lexical/blob/3349ffbaf4f6902f8ccb71494b1535129f16d1d4/packages/lexical-selection/src/index.js#L163) by only including nodes that are "selected" in the editor, rather than building a clone from the bottom all the way up. This will make it's behavior consistent with [`$convertSelectedLexicalContentToHtml`](https://github.com/facebook/lexical/blob/f9ee530c9012b9de3ff7469a11ffb548861cdb25/packages/lexical-clipboard/src/clipboard.js#L117) and more predictable overall.

If we DO want to include the parent `ElementNode` with a `RangeSelection` isolated in the same parent, I propose that we introduce a new method on `ElementNode` that signals it's coupled with its `TextNode` children. This might be useful for Paragraphs, Headers, etc.

Open to suggestions for the new behaviors, but I think we should start by only copying selected content internally (i.e. the text itself in this case) until we have a more consistent & predictable solution in mind.

[...]

PR #1996, opened 2022-04-26
Improve Lexical -> HTML and Lexical -> Lexical Copy and Paste Data Model Conversion

tylerjbainbridge:
Solves the problems described in #1957 and the `Lexical -> HTML` & `Lexical -> Lexical (Clone)` conversion functions now have identical outputs, which will make debugging issues in the future much easier.

This PR also adds an `extractWithChild` method to `ElementNodes` that can be used to ensure there's a tightly coupled relationship between parent and child. This is useful situations where the `ElementNode` provides important context for its children e.g. `Heading -> TextNode` or `List -> ListItem`.

NOTE: Most of the line count is from the new tests in `clipboard.test.js`. 

`Lexical -> Lexical (Clone)` and `Lexical -> HTML` are tested individually and then they're both ran through the same processing code and compared to each other ensure the end output is the same.

https://user-images.githubusercontent.com/13852400/165403382-c8b326f9-bbce-4789-8d88-f9392f252fe7.mov

### DIFF
--- a/packages/lexical-clipboard/src/__tests__/unit/clipboard.test.js
+++ b/packages/lexical-clipboard/src/__tests__/unit/clipboard.test.js
@@ -1,0 +1,418 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {$createLinkNode, LinkNode} from '@lexical/link';
+import {$createListItemNode, $createListNode} from '@lexical/list';
+// import {$createHorizontalRuleNode} from '@lexical/react';
+import {$createTableNodeWithDimensions} from '@lexical/table';
+import {$dfs} from '@lexical/utils';
+import {
+  $createGridSelection,
+  $createParagraphNode,
+  $createTextNode,
+  $getRoot,
+  $getSelection,
+  $setSelection,
+} from 'lexical';
+import {initializeUnitTest} from 'lexical/src/__tests__/utils';
+
+import {
+  $cloneSelectedLexicalContent,
+  $convertSelectedLexicalContentToHtml,
+  $generateNodes,
+  $generateNodesFromDOM,
+} from '../../clipboard';
+
+// No idea why we suddenly need to do this, but it fixes the tests
+// with latest experimental React version.
+global.IS_REACT_ACT_ENVIRONMENT = true;
+
+export function setAnchorPoint(point) {
+  let selection = $getSelection();
+  if (selection === null) {
+    const dummyTextNode = $createTextNode();
+    dummyTextNode.select();
+    selection = $getSelection();
+  }
+  const anchor = selection.anchor;
+  anchor.type = point.type;
+  anchor.offset = point.offset;
+  anchor.key = point.key;
+}
+
+export function setFocusPoint(point) {
+  let selection = $getSelection();
+  if (selection === null) {
+    const dummyTextNode = $createTextNode();
+    dummyTextNode.select();
+    selection = $getSelection();
+  }
+  const focus = selection.focus;
+  focus.type = point.type;
+  focus.offset = point.offset;
+  focus.key = point.key;
+}
+
+function expectMatchingOutput(editor, cloneState, htmlString) {
+  const parser = new DOMParser();
+  const dom = parser.parseFromString(htmlString, 'text/html');
+
+  const htmlToLexicalTreeRoot = $createParagraphNode();
+  htmlToLexicalTreeRoot.append(...$generateNodesFromDOM(dom, editor));
+
+  const lexicalToLexicalTreeRoot = $createParagraphNode();
+  lexicalToLexicalTreeRoot.append(...$generateNodes(cloneState));
+
+  const htmlToLexicalTreeNodes = $dfs(htmlToLexicalTreeRoot);
+  const lexicalToLexicalNodes = $dfs(lexicalToLexicalTreeRoot);
+
+  expect(htmlToLexicalTreeNodes.length).toBe(lexicalToLexicalNodes.length);
+
+  for (let i = 0; i < htmlToLexicalTreeNodes.length; i++) {
+    const htmlToLexicalNode = htmlToLexicalTreeNodes[i];
+    const lexicalToLexicalDFSNode = htmlToLexicalTreeNodes[i];
+
+    expect(htmlToLexicalNode.depth).toBe(lexicalToLexicalDFSNode.depth);
+
+    Object.keys(htmlToLexicalNode.node).forEach((property) => {
+      if (property !== '__key') {
+        expect(htmlToLexicalNode.node[property]).toBe(
+          lexicalToLexicalDFSNode.node[property],
+        );
+      }
+    });
+  }
+}
+
+const fillEditorWithComplexData = () => {
+  const root = $getRoot();
+
+  const paragraph = $createParagraphNode();
+  const nodesToInsert = [];
+
+  root.append(paragraph);
+
+  setAnchorPoint({
+    key: paragraph.getKey(),
+    offset: 0,
+    type: 'element',
+  });
+  setFocusPoint({
+    key: paragraph.getKey(),
+    offset: 0,
+    type: 'element',
+  });
+
+  const link = $createLinkNode('https://');
+  link.append($createTextNode('ello worl'));
+
+  nodesToInsert.push(
+    $createTextNode('h').toggleFormat('bold'),
+    link,
+    $createTextNode('d').toggleFormat('italic'),
+  );
+
+  const table = $createTableNodeWithDimensions(3, 3);
+  nodesToInsert.push(table);
+
+  const tableCell1TextNode = table.getFirstDescendant();
+  tableCell1TextNode.setTextContent('table cell text!');
+
+  const list = $createListNode('ul');
+
+  for (let i = 0; i < 4; i++) {
+    const listItemNode = $createListItemNode();
+    listItemNode.append(
+      $createTextNode(`${i + 1}: Lorem ipsum dolor sit amet`),
+    );
+    list.append(listItemNode);
+  }
+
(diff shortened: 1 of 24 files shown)


tylerjbainbridge:
PR is ready for review.

packages/lexical-clipboard/src/clipboard.js
```diff
@@ -56,115 +57,156 @@ export function getHtmlContent(editor: LexicalEditor): string | null {
...
+    const nodeKeys = $appendSelectedLexicalNodesToClone(
+      editor,
+      selection,
+      topLevelNode,
+      nodeMap,
+      range,
+      true,
+    );
+    if (nodeKeys.length) {
+      range.push(...nodeKeys);
```
trueadm, line 199: Same as above.

packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -289,6 +292,26 @@ export class ListItemNode extends ElementNode {
...
+    selection: RangeSelection | NodeSelection | GridSelection,
+  ): boolean {
+    if (!$isRangeSelection(selection)) {
+      return false;
+    }
+    const anchorNode = selection.anchor.getNode();
+    const focusNode = selection.focus.getNode();
+    const isBackward = selection.isBackward();
+    const selectionLength = isBackward
+      ? selection.anchor.offset - selection.focus.offset
```
trueadm, line 307: Extract focus and anchor out so we don't keep doing long property accessors. Also, does this work if the anchor and focus point is the same node?

packages/lexical/src/nodes/LexicalParagraphNode.js
```diff
@@ -16,9 +15,9 @@ import type {
   DOMConversionMap,
   DOMConversionOutput,
   DOMExportOutput,
-  LexicalNode,
 } from '../LexicalNode';
 
+import {LexicalNode} from '../LexicalNode';
```
trueadm, line 20: What is this change for?

packages/lexical-clipboard/src/clipboard.js
```diff
@@ -56,115 +57,156 @@ export function getHtmlContent(editor: LexicalEditor): string | null {
...
+  }
+  return {nodeMap, range};
+}
+
 export function $getLexicalContent(editor: LexicalEditor): string | null {
   const selection = $getSelection();
   if (selection !== null) {
     const namespace = editor._config.namespace;
-    const state = $cloneContents(selection);
+    const state = $cloneSelectedLexicalContent(editor, selection);
```
trueadm, line 209: Can it be `$cloneSelectedContent`. I don't think we need Lexical in the name.

packages/lexical-clipboard/src/clipboard.js
```diff
@@ -56,115 +57,156 @@ export function getHtmlContent(editor: LexicalEditor): string | null {
...
+      editor,
+      selection,
+      topLevelNode,
+      container,
+    );
   }
   return container.innerHTML;
 }
 
+export function $appendSelectedLexicalNodesToClone(
```
trueadm, line 129: `$appendSelectedNodesToClone` is shorter.

packages/lexical-clipboard/src/clipboard.js
```diff
@@ -56,115 +57,156 @@ export function getHtmlContent(editor: LexicalEditor): string | null {
...
     const childNode = children[i];
-
-    if (childNode.isSelected()) {
-      const newElement = $convertSelectedLexicalNodeToHTMLElement(
-        editor,
-        selection,
-        childNode,
-      );
-      if (newElement) element.append(newElement);
+    const shouldIncludeChild = $appendSelectedLexicalNodesToHTML(
```
trueadm, line 82: This can be `$appendSelectedNodesToHTML`

packages/lexical/src/nodes/LexicalParagraphNode.js
tylerjbainbridge: Remnants of a removed changed, my bad.

packages/lexical-list/src/LexicalListItemNode.js
tylerjbainbridge: Updated the solution so this is no longer needed.

packages/lexical-clipboard/src/clipboard.js
```diff
@@ -56,115 +57,156 @@ export function getHtmlContent(editor: LexicalEditor): string | null {
...
-        const splitNodes = node.splitText(offset);
-        nodeToConvert = splitNodes[0];
-      }
-    }
+  currentNode: LexicalNode,
+  parentElement: HTMLElement | DocumentFragment,
+): boolean {
+  let shouldInclude = currentNode.isSelected();
+  const shouldExclude =
+    $isElementNode(currentNode) && currentNode.excludeFromCopy();
```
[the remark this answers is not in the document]
tylerjbainbridge: I fixed this and added a comment, but I think this should actually be changed at the `$generateNodes` level. The range + nodeMap approach creates some issues when the core structure of the tree is modified while copying. E.g. when parents no longer reference the right children, a node moves up or down the tree, etc.

An earlier version of this PR built a new tree from scratch in memory without appending to the document, but this caused some strange behavior because the clones use the same ID and the editor instance was updating the actual nodes in state. Also, the way "cloned" nodes work is a bit dangerous because if you use `.getWritable()` you're actually modifying the original node too since it looks it up by key.

I just wanted to flag that this data model is susceptible to some weird bugs right now.

packages/lexical-clipboard/src/clipboard.js
```diff
@@ -56,115 +57,198 @@ export function getHtmlContent(editor: LexicalEditor): string | null {
...
-    if ($isTextNode(node) && (isAnchor || isFocus)) {
-      const [anchorOffset, focusOffset] = selection.getCharacterOffsets();
-      const isBackward = selection.isBackward();
+  currentNode: LexicalNode,
+  parentElement: HTMLElement | DocumentFragment,
+): boolean {
+  let shouldInclude = currentNode.isSelected();
+  const shouldExclude =
+    $isElementNode(currentNode) && currentNode.excludeFromCopy('html');
+  let clone = $cloneWithProperties<LexicalNode>(currentNode);
```
acywatson, line 72: Do we clone basically just so we can splitText without mutating the current tree?

packages/lexical-clipboard/src/clipboard.js
```diff
@@ -56,115 +57,198 @@ export function getHtmlContent(editor: LexicalEditor): string | null {
...
+  selection: RangeSelection | NodeSelection | GridSelection,
+): string {
+  const container = document.createElement('div');
+  const root = $getRoot();
+  const topLevelChildren = root.getChildren();
+  for (let i = 0; i < topLevelChildren.length; i++) {
+    const topLevelNode = topLevelChildren[i];
+    $appendSelectedNodesToHTML(editor, selection, topLevelNode, container);
+  }
+  return container.innerHTML;
```
acywatson, line 121: To make this work on the server, I wonder if it'd be better to have exportDOM return some abstract interface that we can then render directly to a string here. For instance, it could be something like:

```
interface ExportDOMResult = {
  tag: string; // p, div, ul, li, etc...,
  style: {[key: string]: string},
  ...
}
```

packages/lexical-clipboard/src/clipboard.js
tylerjbainbridge: Yeah, that's right, and for when the `__children` array changes. I noted this above, but I think building a new tree in memory with deep cloned nodes is the safest (and most intuitive) approach moving forward.

packages/lexical-clipboard/src/clipboard.js
tylerjbainbridge: I think this is a step in the right direction, but some nodes, add extra elements and move their children around during `exportDOM` step e.g. Tables add a `colgroup` and move the child rows into a `tbody`. 

Maybe `exportDOM` could return a similar structure to what you proposed above, but one that allows for nested structures? Not sure. Probably out of the scope of this PR.

I do worry that making `exportDOM` radically different than `createDOM` is pretty confusing though.

packages/lexical-clipboard/src/__tests__/unit/clipboard.test.js
```diff
@@ -0,0 +1,403 @@
...
+  $convertSelectedContentToHtml,
+  $generateNodes,
+  $generateNodesFromDOM,
+} from '../../clipboard';
+
+// No idea why we suddenly need to do this, but it fixes the tests
+// with latest experimental React version.
+global.IS_REACT_ACT_ENVIRONMENT = true;
+
+export function setAnchorPoint(point) {
```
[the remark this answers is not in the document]
tylerjbainbridge: Whoops!!

packages/lexical-clipboard/src/clipboard.js
acywatson: I was worried the signature of exportDOM might be hard to change later, but the return value is abstracted into an object we can control the shape of, so I think we should be OK.

acywatson (approved):

packages/lexical-clipboard/src/clipboard.js
tylerjbainbridge: As long as exportDOM returns _something_ that can be translated into an HTML string we should be able to toy with and refine these APIs as we discover limitations :)

trueadm:
Any chance for some copy and paste e2e tests? Can’t we check the clipboard contents in e2e tests for everyone we do a copy in each test that does a copy? Like an assertion on the clipboard content?

tylerjbainbridge:
> Any chance for some copy and paste e2e tests? Can’t we check the clipboard contents in e2e tests for everyone we do a copy in each test that does a copy? Like an assertion on the clipboard content?

We can do that, but these functions are consumed directly by the ones that add the strings to the clipboard. So in a way we'd be testing that the browser clipboard API works rather than the actual changes introduced in this PR. 

Definitely open to doing this in a follow up PR though.

PR #1998, opened 2022-04-27
Markdown import/export/shortcuts

fantactuka:
### YAMD - Yet another markdown
Another markdown attempt, this PR includes core files to support shortcuts, import and export. Follow up PR (I wish we had stacks here) will show the way it's used and extended to support custom tags and nodes

### Shortcuts
Markdown logic relies on transformers that define how text and nodes are being processed. There are three types of transformers:
- _Block transformers_: `#` heading, `>` quote, `1.` list, etc. Top level elements, often have child nodes.
- _Text format transformers_: `*` italic `**` bold, `~~` strikethrough, etc. Can wrap other nodes (links), can be nested into each other (e.g. `*Hello __world__*`.
- _Text match transformers_: `[text](url)` link, `alt` images or any other text matching RegEx pattern. Can not have any nested nodes.

### Shortcuts
Since this logic is executed while typing it tries to have quick checks before running any matching. For text format and text match transformers it has internal maps:
```js
{
  "*": [bold transformer, italic transformer, bold italic transformer],
  "~": [strikethrough transformer],
...
  ")": [image transformer, link transformer]
}
```
When char under caret matches any of the map keys it tries to apply those transformers (although logic for text format and text matchers is slightly different given format might span across multiple text nodes). For blocks it's little less effective, as they are triggered by spaces (and spaces unlike `*` or `~` are more common, hence checked more often).

### Import
Executes following for each line of text (with exception for multi-line code blocks):
- Convert into top level element
- Process text format tags (`*`, `**`, etc) starting for top most and then recursively processing any nested ones
- Process text matches last (e.g. links `[](url)`) since they always applied over single text node and should be leaf nodes

### Export
Recursively runs over tree and generating output string. Nothing fancy, except maybe for text formats it deduplicates opening/closing tags if sibling nodes have same format

### API
Transformers are passed from outside, hence users can define their own subset of markdown rules and/or extend existing ones (e.g. add `^` to support superscript text format, or add tables with `|` block transformer). As of now, each transformer contains all settings needed for shortcuts, import and export. This approach follows current implementation with a good reason: you define one transformer that will be passed whenever markdown is used (whether it shortcuts, import or export), so there're less chances to forget writing separate one in three places. One cons: block and text match transformers look a bit bloated, especially if don't someone don't need export/import and only care about shortcuts, or vise versa:
```js
TextMatchTransformer:
  // Export:
  export: (...) => string | null,   // conversion logic
  nodeType: string, // node types that should use this transformer (not added yet, but will improve exporter lookup time)
  
  // Shortcuts:
  regExp: RegExp,
  trigger: string, // trigger char (last char of closing tag, e.g. ")" for images/links)

  // Import
  importRegExp: RegExp,

  // Import & shortcuts:
  replace: (node: TextNode, match: RegExp$matchResult) => void,

BlockTransformer =
  // Export
  export: (...) => string | null, // : conversion logic
  nodeType: string, // node types that should use this transformer (not added yet, but will improve exporter lookup time)

  // Import & shortcuts:
  regExp: RegExp,
  replace: (...) => void, // conversion logic
```

Text format transformers look way nicer:
```
{ format: ['strikethrough'], tag: '~~' }
```
which is enough for all conversions


packages/lexical-markdown/src/v2/MarkdownTransformers.js
```diff
@@ -0,0 +1,248 @@
...
+  }),
+  type: 'block-match',
+};
+
+export const QUOTE: BlockTransformer = {
+  export: (node, exportChildren) => {
+    return $isQuoteNode(node) ? '> ' + exportChildren(node) : null;
+  },
+  regExp: /^>\s/,
+  replace: replaceWithBlock(() => $createQuoteNode()),
```
a-xin, line 153: There might be a small issue for block quotes here:

Exporting Rich-text to Markdown should add an empty line after the quote block in Markdown if followed by a paragraph.

And going from Markdown to Rich-text should probably remove that empty line after a block quote to allow switching back and forth between Markdown and Rich-text without adding a new empty lines on every conversion.

https://spec.commonmark.org/0.30/#block-quotes
https://spec.commonmark.org/0.30/#example-247

packages/lexical-markdown/src/v2/MarkdownTransformers.js
```diff
@@ -0,0 +1,248 @@
...
+  type: 'block-match',
+};
+
+export const CODE: BlockTransformer = {
```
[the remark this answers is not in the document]
fantactuka: Started with ElementTransformer, but then thought it might be misleading due to inline elements which are not supposed to be handled here. Alternative could be TopLevelElementTransformer, but maybe it's too long?

packages/lexical-markdown/src/v2/MarkdownExport.js
```diff
@@ -0,0 +1,185 @@
...
+        output += tag;
+      }
+    }
+  }
+  return output;
+}
+
+// Get next or previous text sibling a text node, including cases
+// when it's a child of inline element (e.g. link)
+function getTextSibling(node: TextNode, backward: boolean): TextNode | null {
```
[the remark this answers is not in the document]
fantactuka: This one used to get nearby text nodes. Usually it's trivial when sibling is text node right away, but consider following (three links in a row):

```js
link { key: 1 }
  text { key: 2 }
link { key: 3 }  
  text { key: 4 }
link { key: 5 }  
  text { key: 6 }
```

when exporting text node key#4 need to see if neighbour text nodes have same format (to deduplicate open/close tags), it would check that there're no immediate siblings for node#4, that its parent is inline element. Then get parent's siblings, see that those are also inline elements, and would take their appropriate leaf text nodes

packages/lexical-markdown/src/v2/MarkdownExport.js
```diff
@@ -0,0 +1,185 @@
...
+  TextFormatTransformer,
+  TextMatchTransformer,
+} from './MarkdownTransformers';
+import type {ElementNode, LexicalNode, TextFormatType, TextNode} from 'lexical';
+
+import {$getRoot, $isElementNode, $isLineBreakNode, $isTextNode} from 'lexical';
+
+export function createMarkdownExport(
+  blockTransformers: Array<BlockTransformer>,
+  textFormatTransformers_: Array<TextFormatTransformer>,
```
acywatson, line 21: nit: I actually feel like prefixing with underscore at the beginning might be better, if we need a prefix at all.

packages/lexical-markdown/src/v2/MarkdownTransformers.js
fantactuka: Thanks for pointing out, it appears blockquotes are more complex than other block modifiers. Also supporting `\n` vs `\n\n` as a newline indicator is another important topic: would be nice to have it as an option. I'll work on it in follow ups

packages/lexical-markdown/src/v2/MarkdownExport.js
```diff
@@ -0,0 +1,185 @@
+/**
```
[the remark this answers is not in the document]
fantactuka: It is a different implementation, so had to put it somewhere and didn't come up with anything better than v2 🙂

acywatson:
Another question - this doesn't support inserting markdown at the selection point, right? importMarkdown will just replace the entire editor state with the nodes generated from provided markdown?

fantactuka:
> Another question - this doesn't support inserting markdown at the selection point, right? importMarkdown will just replace the entire editor state with the nodes generated from provided markdown?

Ah, you're right. Should have selection as an optional argument for both import & export
I'll probably merge this one and open a follow up as I have all playground code and e2e tests in next PR and jumping between branches is no joy

fantactuka:
> Do we have some tests for all this?

#2000 has this markdown @ playground and related e2e's this one is only for core logic as I wanted it to be reviewed in isolation which does not sounds like a great idea anymore :)

acywatson (approved):
This is a beast - nice work

packages/lexical-markdown/src/v2/MarkdownExport.js
Proful: I too felt the same :)

Issue #1927, opened 2022-04-22
Shouldn't be able to click outside the Modal (i.e. Excalidraw)


zurfyx:
Maybe we should close the modal when this happens and offer a method to handle modal discard

Additionally, you should not be able to tab outside the modal

siva-kannan3:
@zurfyx I would like to work on this issue. Can I be assigned?

siva-kannan3:
@zurfyx I would love to work on this issue. Can I be assigned?

siva-kannan3:
would it be nice to add click outside support for other modals too like table, poll?

PAkerstrand:
This issue was closed by #1999, correct?

PR #1999, opened 2022-04-27
support for onclickoutside of excalidraw modal

siva-kannan3:
Addresses this issue [1927](https://github.com/facebook/lexical/issues/1927).


zurfyx (commented):
Thank you for taking a stab on the modal! While your implementation is valid it may need some tweaks:

1. Events outside the modal shouldn't trigger instantly
2. It helps user retain the focus if they can the background visually darkened

Hence, I propose a version that's closer to - https://www.w3schools.com/howto/howto_css_modals.asp

A background div with some opacity and a content modal (like the current). When you click outside the content the modal disappears and the first click handles closing the modal.

siva-kannan3:
@zurfyx I have added the background overlay for excalidraw modal as mentioned in the second point.
Can you explain the first point a bit detail ?

zurfyx:
Thanks! What you have makes sense now. Happy to merge as-is but we'll probably need some follow up PRs on a discard modal once you click outside as it's easy to accidentally lose content.

Actually, I even wonder whether we need save/discard action buttons, perhaps it should (always) be automatically saved. @rulikrulit thoughts on this?

zurfyx (approved):
Mind rebasing/merging with main and see if the E2E test failures still persist?

PR #2000, opened 2022-04-27
Tests, playground usage, table/image/hr/tweet transformers

fantactuka:
Playground use with custom markdown transformers: 
- Image
- Katex equation
- Table (extremely basic, lots to improve)
- Tweet (more for a demo purpose and no real usage, although markdown allows html-like tags to be used so it can be a legal way to allow embeds within markdown)

https://user-images.githubusercontent.com/132642/165447693-3a01b590-603b-4928-af92-dd3e405669a5.mov

### DIFF
--- a/packages/lexical-playground/src/plugins/MarkdownTransformers.js
+++ b/packages/lexical-playground/src/plugins/MarkdownTransformers.js
@@ -1,0 +1,228 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import type {TextFormatTransformer} from '../../../lexical-markdown/src/v2/MarkdownTransformers';
+import type {BlockTransformer, TextMatchTransformer} from '@lexical/markdown';
+import type {ElementNode, LexicalNode} from 'lexical';
+
+import {v2} from '@lexical/markdown';
+import {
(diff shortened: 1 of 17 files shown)


packages/lexical-playground/src/plugins/MarkdownTransformers.js
```diff
@@ -0,0 +1,228 @@
...
+
+export const IMAGE: TextMatchTransformer = {
+  export: (node, exportChildren, exportFormat) => {
+    if (!$isImageNode(node)) {
+      return null;
+    }
+    return `![${node.getAltText()}](${node.getSrc()})`;
+  },
+  importRegExp: /!(?:\[([^[]*)\])(?:\(([^(]+)\))/,
+  regExp: /!(?:\[([^[]*)\])(?:\(([^(]+)\))$/,
```
Proful, line 66: Scary to see these regex. May be it just me. Wish there is one line of comment there to read the code quickly.

zurfyx: Don't disagree with that but from the Transformer we can guess it's `foo`?

zurfyx (approved):
That's awesome work, especially after having seen the offline demo! Side note: we got some feedback around the bottom action bar button not being very UX friendly. Perhaps we need to revisit the UI at some point

packages/lexical-playground/src/plugins/MarkdownTransformers.js
Proful: Sounds good, thanks.

PR #2003, opened 2022-04-27
fix: 404 on 'edit this page' link

samuelmartineau:
Hi there again,

I am not familiar with docusaurus so feel free to reject this PR if the fix is wrong.
Edit page call-to-action links to a 404 due to a double "docs" slug in the URL.

Capture d’écran 2022-04-27 à 18 03 39


zurfyx (approved):
Makes sense to me! Thanks

PR #2005, opened 2022-04-27
Revert "Separate clean script"

fantactuka:
Reverts facebook/lexical#1949

* Blocks www sync, need to figure root cause


thegreatercurve (approved):

fantactuka:
Fixed in #2007

PR #2008, opened 2022-04-27
Fix selection bug

trueadm:
This revises the logic and makes it special case only for iOS/Safari where we have an anchor type of element. We always need to set the selection in this case.


fantactuka (approved):
🎉

PR #2012, opened 2022-04-28
Adjust first node when applying format to selection

fantactuka:
Setting anchor to an end of a text node right before link node, selecting link and calling selection.toggleFormat caused crash since Selection#formatText didn't adjust for inline elements. Note that updated version relies on current usage of inline elements (links) that are non empty.


trueadm (approved):

trueadm:
An e2e test for this would be good, as this behavior might differ between browsers.

fantactuka:
> An e2e test for this would be good, as this behavior might differ between browsers.

Not sure if it's possible to replicate this issue via UI test. It was discovered with markdown transformation, where selection is set programmatically over nodes range. When doing it over UI it adjusts offset properly, here's the scenario:

If put caret before link and start expanding selection over the link, then offset moves from the end of previous text sibling to the beginning of link's text:

https://user-images.githubusercontent.com/132642/165878678-975b6c33-4061-45e4-a7a2-311a64bb5483.mov

PR #2018, opened 2022-04-28
Rename $rootTextContentCurry -> $rootTextContent

zurfyx:
The idea behind the curry suffix was to be able to provide an alternative that worked immediately on the read clause. For $textContent, which is widely used, the curry suffix is unnecessary (it doesn't take any parameter) and seemed confusing (maybe more now because someone killed the alternative function). So just cleaning this up and making it a single function without the suffix


trueadm:
Bare with me, I've been using this extensively in a PR.

trueadm:
@zurfyx All ready to rebase and merge this!

thegreatercurve:
@zurfyx Can this be merged?

PR #2042, opened 2022-04-29
Improve insertText handling of inline elements

trueadm:
This improves on the heuristics of `insertText`, so that it better handles inline elements. Working with @acywatson on this, so it might need some further work.

### DIFF
--- a/packages/lexical/src/LexicalSelection.js
+++ b/packages/lexical/src/LexicalSelection.js
@@ -808,8 +808,6 @@ export class RangeSelection implements BaseSelection {
       let lastElement = $isElementNode(lastNode)
         ? lastNode
         : lastNode.getParentOrThrow();
-      const shouldInsertAfterFirstElement =
-        firstElement.isInline() && firstNode.getNextSibling() === null;
       let lastElementWasInline = false;
 
       // If the last element is inline, we should instead look at getting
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalSelection.js
```diff
@@ -848,55 +863,60 @@ export class RangeSelection implements BaseSelection {
...
+
+        if (lastNodeChild.isAttached()) {
           if (
-            lastNodeChild.is(firstNode) ||
-            ($isElementNode(lastNodeChild) &&
-              lastNodeChild.isParentOf(firstNode))
+            !selectedNodesSet.has(lastNodeChild) ||
+            lastNodeChild.is(lastNode) ||
+            (lastElementWasInline &&
+              lastNodeChild.is(lastNode.getParentOrThrow()))
```
acywatson, line 891: The use of lastNode seems a little confusing here. For instance,``` lastNodeChild.is(lastNode.getParentOrThrow())``` kind of suggests that a child of "lastNode" and be the same as the parent of "lastNode".  I think it goes back to line 862 where we start to call the children of lastElement "lastNodeChildren".  Maybe they should be lastElementChildren?

packages/lexical/src/LexicalSelection.js
```diff
@@ -800,12 +800,27 @@ export class RangeSelection implements BaseSelection {
...
+      // We have to get the parent elements before the next section,
+      // as in that section we might mutate the lastNode.
       const firstElement = $isElementNode(firstNode)
         ? firstNode
         : firstNode.getParentOrThrow();
-      const lastElement = $isElementNode(lastNode)
+      let lastElement = $isElementNode(lastNode)
         ? lastNode
         : lastNode.getParentOrThrow();
+      const shouldInsertAfterFirstElement =
```
acywatson, line 811: I think we only use this constant once. If we want to keep if for readability, that's probably OK. But given how it's used below it seems to me we might as well inline the expression.

acywatson (approved):
Seems to solve the immediate problem(s), couple of thoughts around the naming.

packages/lexical/src/LexicalSelection.js
trueadm: Addressed with a comment. :)

packages/lexical/src/LexicalSelection.js
trueadm: Agreed, moved it inline.

PR #2046, opened 2022-04-30
Headless mode `@lexical/headless`

fantactuka:
- Adds `@lexical/headless` with `createHeadlessEditor`
- Adds `_headless` to editor
- Skips reconciler and dom selection sync if is headless
- Unit tests are running in node environment to check `editor.update`, `editor.setEditorState` and `editor.register***`

E.g. can run markdown import/export from node:

Screen Shot 2022-04-30 at 7 45 54 AM


tylerjbainbridge (approved):
This looks really cool Maksim! I imagine support for JSDom would be really useful for headless HTML import/export in Node.

fantactuka:
> This looks really cool Maksim! I imagine support for JSDom would be really useful for headless HTML import/export in Node.

Thought about dependency-free package for server, but somehow completely forgot about html export which _does_ require jsdom even thought editor itself does not need to be "rendered". I wonder if headless needs to embed any specific dom implementation (like jsdom) or let users to pick any of their choice, WDYT?

PR #2048, opened 2022-05-01
added HighlightKeywordsPlugin

pfftdammitchris:
Added **HighlightKeywordsPlugin**

Link to [codesandbox code](https://codesandbox.io/s/highlightkeywordsplugin-svj1n3?file=/src/App.tsx)

Configuration/usage:

```tsx
<HighlightKeywordsPlugin
  keywords={{
    '=.builtIn.': {
      color: 'teal',
      match: /=\.builtIn\./,
    },
    funcName: {
      color: 'magenta',
      match: /funcName/,
    },
  }}
/>
```

### DIFF
--- a/packages/lexical-playground/src/plugins/HighlightKeywordsPlugin.js
+++ b/packages/lexical-playground/src/plugins/HighlightKeywordsPlugin.js
@@ -1,0 +1,121 @@
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {$createTextNode, EditorConfig, LexicalEditor, TextNode} from 'lexical';
+import React from 'react';
+
+export class HighlightKeywordNode extends TextNode {
+  __color: string;
+  __keyword: string;
+
+  static getType() {
+    return 'highlight-keyword';
+  }
+
+  static clone(node: HighlightKeywordNode): HighlightKeywordNode {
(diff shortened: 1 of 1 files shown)


fantactuka:
Thanks for your contribution!

There's an [existing Keywords plugin](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/KeywordsPlugin), which you might want to check. Best part of it – it shows the usage of `useLexicalTextEntity` which is build to solve exactly the kind of problem you're addressing in this PR: take RegExp and replace its matched plain text with some custom node.

packages/lexical-playground/src/plugins/HighlightKeywordsPlugin.js
```diff
@@ -0,0 +1,121 @@
...
+                .substring(0, wordIndex);
+
+              const textContentAfter = node
+                .getTextContent()
+                .substring(wordIndex + len);
+
+              if (textContentBefore) {
+                const prevSibling = $createTextNode(textContentBefore);
+                highlightedNode.insertBefore(prevSibling);
+              }
```
fantactuka, line 90: Lexical has `TextNode#splitText` method for this kind of cases, it will also handle selection for you when splitting one single node into multiple. It's commonly used over the code base so take a look!

PR #2050, opened 2022-05-02
Checklist support

fantactuka:
Check list support
- Using `ListNode.__listType` instead of `ListNode.__tag` to differentiate lists
- `ListItemNode.__checked` to track state
- Markdown shortcut, import & export
- Toolbar updates to use list type instead of tag

⚠️ Breaking change ⚠️
`$createListNode(tag, start?)` -> `$createListNode(listType, start?)`
- Added backward compatibility mapping in case tag name is passed instead of list type

https://user-images.githubusercontent.com/132642/166244757-306a6d53-f719-45bd-8f83-234bfc82fd64.mov


packages/lexical-react/src/shared/useList.js
```diff
@@ -82,6 +93,57 @@ export default function useList(editor: LexicalEditor): void {
...
+      editor.update(() => {
+        const node = $getNearestNodeFromDOMNode(target);
+        if ($isListItemNode(node)) {
+          node.toggleChecked();
+        }
+      });
+      event.preventDefault();
+    }
+  }
+}
```
fantactuka, line 149: Few notes on this:
- Ideally core only provides new list type and `__checked` to store item state, and plugins can define how that __checked status is being updated
- This implementation checks that list item was clicked within first 20px and toggles check state; uses `:before` styling to render checked/unchecked icons. Which means there's no need for extra elements in a tree.
  - Accessibility?
  - RTL/LTR?
  - Get rid of hardcoded 20px? (e.g. use 1em to rely on a font size)

Could use some ideas here

tylerjbainbridge:
When this was first implemented for the Unified Editor we had an issue where the checkbox was being treated as part of the contenteditable and you could move the cursor on either side and even delete it. That's not an issue here right? It can be solved with some CSS tricks.

packages/lexical-list/flow/LexicalList.js.flow
```diff
@@ -17,9 +17,12 @@ import type {
 import {ElementNode} from 'lexical';
 
 type ListNodeTagType = 'ul' | 'ol';
-declare export function $createListItemNode(): ListItemNode;
+export type ListType = 'number' | 'bullet' | 'check';
```
tylerjbainbridge, line 20: What about when old editor state is loaded in and has the type set to "ol" or "ul"? Is there a translation step to account for the breaking change?

fantactuka:
> When this was first implemented for the Unified Editor we had an issue where the checkbox was being treated as part of the contenteditable and you could move the cursor on either side and even delete it. That's not an issue here right? It can be solved with some CSS tricks.

Yea, this solution does not render anything additional on top of bullet-list and adds checkbox with css. I wonder if `pointerdown` handler should be moved into standalone plugin, then core will only contain required data to handle check lists (`__checked`, `__listType`), while users will be able to write own plugins to handle checking/unchecking line items

tylerjbainbridge (approved):

trueadm:
I have some concerns around the semantics and a11y, specifically that we're not using an actual `<input>` checkbox anywhere. Have you tested this with NVDA/VoiceOver to make sure it works properly and the items get announced right?

packages/lexical-react/src/shared/useList.js
```diff
@@ -82,6 +93,57 @@ export default function useList(editor: LexicalEditor): void {
...
+  while (node) {
+    if (node.__lexicalEditor) {
+      return (node.__lexicalEditor: LexicalEditor);
+    }
+    node = node.parentNode;
+  }
+  return null;
+}
+
+function handlePointerDown(event) {
```
trueadm, line 128: I don't feel very confident about using pointer events to track selecting the input. This won't work with speech to text or screen readers. Instead we should probably be using a `click` event, which should get tracked. Plus, you should be able to use keyboard tab to move to an item and press "Space" to select it, which happens with native `<input>` elements, as they will fire a click/change event.

trueadm (changes requested):
I think there are too many holes here around a11y and event handling. Let's see if we can tackle them?

fantactuka:
Adding `tabIndex`, `role=checkbox` and `click` handler seem to make it working better for voiceover (although still need pointerdown handler to prevent caret moving when clicking on check mark). Not quite sure whether `tabindex` will do any damage, but works fine on mac browsers with voice over:

https://user-images.githubusercontent.com/132642/166310566-04eae157-696e-4d40-8073-14d7c0c13ec0.mov

packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -187,7 +192,9 @@ export class ListItemNode extends ElementNode {
   }
 
   insertNewAfter(): ListItemNode | ParagraphNode {
-    const newElement = $createListItemNode();
+    const newElement = $createListItemNode(
+      this.__checked == null ? undefined : false,
```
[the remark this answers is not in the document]
fantactuka: > Can `$createListItemNode` have a default parameter value?
Value only required if it's a checklist, otherwise it's undefined (for bullet/number list)

trueadm:
> Adding `tabIndex`, `role=checkbox` and `click` handler to `<li>` seem to make it working better for voiceover (although still need pointerdown handler to prevent caret moving when clicking on check mark). Not quite sure whether `tabindex` will do any damage, but works fine on mac browsers with voice over:
> 
> Note: video does not include css change to remove blue outline styles for focused `<li>`
> 
>  Screen.Recording.2022-05-02.at.2.56.58.PM.mov

Can you check on Windows with NVDA too please? Maybe reach out to @a11yHolli for feedback if you give her a URL?

a11yHolli:
> > Adding `tabIndex`, `role=checkbox` and `click` handler to `<li>` seem to make it working better for voiceover (although still need pointerdown handler to prevent caret moving when clicking on check mark). Not quite sure whether `tabindex` will do any damage, but works fine on mac browsers with voice over:
> > Note: video does not include css change to remove blue outline styles for focused `<li>`
> > Screen.Recording.2022-05-02.at.2.56.58.PM.mov
> 
> Can you check on Windows with NVDA too please? Maybe reach out to @a11yHolli for feedback if you give her a URL?

This has problems because you can only select/deselect the checkbox with a mouse. We need to find a way for users to move focus to the checkbox and select/deselect it.

@fantactuka I was thinking about this some more after we talked and almost think we should treat this like an interactive character, allowing the user to move the cursor backwards in front of the checkbox with arrow keys in order to select / deselect it using space and/or enter.

Some of the challenges I see with that approach:

- It's not a very elegant solution for unix users, because unix typically announces the character behind the cursor not in front of it so we may have to reverse the behavior for unix meaning when the cursor is behind the checkbox the user can't use space or enter to adjust text instead they can only use it to toggle selection
- You couldn't use space to adjust the position of the checkbox (you can't do that now so maybe this isn't as big of a deal)
- When the checkbox is initially spawned we would still want the cursor to be behind the checkbox so users can just start writing. I'm not sure how obvious it would be that you need to go back one character to interact with the checkbox.

We could address some of those concerns by using the unix approach for both windows and unix. 
When the cursor is directly behind the checkbox we allow space to select/deselect the checkbox. 
That solves some of the problems and creates a problem that you couldn't use space/enter directly behind a checkbox. Maybe we could mitigate this by only using space to select the checkbox. I don't think it would be a huge loss not to be able to put a space directly after a checkbox, and I don't think it would necessarily be expected that you would toggle it with enter. 

https://user-images.githubusercontent.com/78975337/166341128-b79a80c2-8c52-455a-b9ab-ce23213d09f1.mp4

fantactuka:
- Checkbox can be selected pressing arrow key when caret at the very beginning of a list item; if checkbox is already selected and arrow button pressed again - it allows caret to move as usual to a previous line
- Can move between check list items via up/down arrows


https://user-images.githubusercontent.com/132642/166505251-0a139681-7bce-49bc-a49a-8fd235d35fe4.mov

fantactuka:
Follow up refactoring planned:

- move `register*` calls from `@lexical/react/useList` to `@lexical/list/registerBulletAndNumberLists` (will make it easier for non-react frameworks to hook lists)
- move `register*` calls from `LexicalCheckListPlugin` to `@lexical/list/registerCheckList` (keep separate from two other list types as our implementation is a bit opinionated so we want to keep an option to use a different approach for check/uncheck handling)

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -192,11 +192,56 @@
...
-  margin: 8px 32px 8px 32px;
+  margin: 8px 32px;
+}
+.PlaygroundEditorTheme__listItemChecked,
+.PlaygroundEditorTheme__listItemUnchecked {
+  position: relative;
+  margin-left: 8px;
+  margin-right: 8px;
+  padding-left: 24px;
+  padding-right: 24px;
```
harish-sethuraman, line 203:
```suggestion
  margin: 0px 8px;
  padding: 0px 24px;
```
may be?

trueadm (approved):

trueadm:
Awesome stuff! :)

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
fantactuka: I really don’t want to change top/bottom margin or padding and only change sides to make space for checkbox

a11yHolli:
This works beautifully in NVDA but has some issues in JAWS. I spent some time trying to diagnose the issue and it seems to be related to the virtual PC cursor.

JAWS has two types of cursors a virtual PC cursor and a PC cursor. The Virtual PC Cursor is the read mode and the PC cursor is the edit mode. Usually JAWS switches automatically between these two modes and the user doesn't even need to be aware of what mode they are in. Usually in a contenteditable JAWS switches automatically to PC cursor, however it seems like either the dropdown autocomplete or moving into the checkbox is forcing JAWS into Virtual PC cursor and then the space command isn't getting passed to the checkbox. I can fix it by forcing JAWS back to the PC cursor with a shortcut (insert + space > z).

NVDA also has both these modes but doesn't seem to be having a problem switching. 

Here is a video of both experiences.
https://user-images.githubusercontent.com/78975337/166813964-87e77b8a-4f21-442f-869b-68d89824f872.mp4

PR #2054, opened 2022-05-03
fix: dispatchCommand payload typescript type

gutenye:
payload is optional, so that `editor.dispatchCommand(command)` won't show a type error in typescript.

Screen Shot 2022-05-03 at 18 00 21

### DIFF
--- a/packages/lexical/Lexical.d.ts
+++ b/packages/lexical/Lexical.d.ts
@@ -146,7 +146,7 @@ export declare class LexicalEditor {
     klass: Class<T>,
     listener: Transform<T>,
   ): () => void;
-  dispatchCommand<P>(type: string, payload: P): boolean;
+  dispatchCommand<P>(type: string, payload?: P): boolean;
   hasNodes(nodes: Array<Class<LexicalNode>>): boolean;
   getDecorators<X>(): Record<NodeKey, X>;
   getRootElement(): null | HTMLElement;


zurfyx:
Thanks! Actually the payload should not be optional, I fixed this here for Flow - https://github.com/facebook/lexical/pull/2030

Imagine you have a FormatCommand that is supposed to send 'bold' or 'italic'. If it's optional, it means it can potentially send undefined, that's not intended! Instead, you want the generic type to handle nullish values for you

gutenye:
Okay, then the playground also needs to be fixed

https://github.com/facebook/lexical/blob/31e3b1889a4ef570a04597c04d75efa4e13aea88/packages/lexical-playground/src/plugins/ToolbarPlugin.jsx#L631

to

```js
editor.dispatchCommand(INSERT_UNORDERED_LIST_COMMAND, null); 
```

zurfyx:
> Okay, then the playground also needs to be fixed
> 
> https://github.com/facebook/lexical/blob/31e3b1889a4ef570a04597c04d75efa4e13aea88/packages/lexical-playground/src/plugins/ToolbarPlugin.jsx#L631
> 
> to
> 
> ```js
> editor.dispatchCommand(INSERT_UNORDERED_LIST_COMMAND, null); 
> ```

The payload for this command is type `void` isn't it?

gutenye:
Yes, and the typescript expected 2 arguments, so we need to provide an extra `null` for second argument.

PR #2060, opened 2022-05-03
Adjust octocat colors to match rest of background

fantactuka:
Very important change:

Before 

Screen Shot 2022-05-03 at 3 52 30 PM

After 

Screen Shot 2022-05-03 at 3 52 24 PM


zurfyx (approved):
> Very important change:

Agree with the importance of the change, thank you for taking the lead on this. The repo is now safe 🙇‍♂️

PR #2063, opened 2022-05-04
Fix decorator content selection bug

trueadm:
When selection is within the content of a DecoratorNode, selection should be `null`.


zurfyx (approved):

thegreatercurve:
@trueadm This breaks selection when you drag your mouse over a decorator node to select it and the surrounding elements. This is what I would expect to happen: 

Uploading Screen Recording 2022-05-04 at 14.09.21.mov…

trueadm:
@thegreatercurve Mind checking on https://github.com/facebook/lexical/pull/2064?

PR #2064, opened 2022-05-04
Better fix for decorator content selection bug

trueadm:
A follow up to https://github.com/facebook/lexical/pull/2063, that applies a better heuristic for collapsed selection in a decorator.


thegreatercurve (approved):
Thanks for the fix!

PR #2066, opened 2022-05-04
Fix copy/paste bug for decorator nodes

### LINKED ISSUE #1868: Bug: Copy/paste broken for root level decorator nodes
Copy and paste appears to be broken for root level DecoratorNodes, such as YouTubeNode, TweetNode and HorizontalRuleNode. 

When you copy/paste to a new editor, any root level DecoratorNodes gets appended out of order to the end of the new document, inside a paragraph tag. 

Sometimes, I also see the below error message as well

|Before|After|
|---|---|
|Screenshot 2022-04-19 at 12 09 32|Screenshot 2022-04-19 at 12 09 48|

Screenshot 2022-04-19 at 12 07 20

## Steps To Reproduce

1. Add several root level decorator nodes into an editor.
2. Copy and paste them to another editor.

thegreatercurve:
I thought this was a bigger bug than it turned out to be. 

Fixes #1868

### DIFF
--- a/packages/lexical-playground/src/nodes/TweetNode.jsx
+++ b/packages/lexical-playground/src/nodes/TweetNode.jsx
@@ -117,6 +117,10 @@ export class TweetNode extends DecoratorBlockNode<React$Node> {
       />
     );
   }
+
+  isTopLevel(): true {
+    return true;
+  }
 }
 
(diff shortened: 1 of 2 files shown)


tylerjbainbridge (approved):
Good catch!!

trueadm (approved):

PR #2071, opened 2022-05-04
Fix add/remove classes order

fantactuka:
Added/removed styles might overlap (e.g. property-specific classes in stylex) so make sure we delete first and then add needed classes

```
existing classes = [a, d]
classes to add = [a, b, c]
classes to remove = [a, d]

existing classes should become [a, b, c]
```

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.js
+++ b/packages/lexical-list/src/LexicalListItemNode.js
@@ -377,12 +377,11 @@ function $setListItemThemeClassNames(
     }
   }
 
-  if (classesToAdd.length > 0) {
-    addClassNamesToElement(dom, ...classesToAdd);
-  }
-
   if (classesToRemove.length > 0) {
     removeClassNamesFromElement(dom, ...classesToRemove);
+  }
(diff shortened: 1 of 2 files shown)


packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -377,13 +377,12 @@ function $setListItemThemeClassNames(
...
   }
 
-  if (classesToAdd.length > 0) {
-    addClassNamesToElement(dom, ...classesToAdd);
-  }
-
   if (classesToRemove.length > 0) {
     removeClassNamesFromElement(dom, ...classesToRemove);
   }
+  if (classesToAdd.length > 0) {
```
acywatson, line 383: Don't we have some similar logic in other nodes?

acywatson (approved):

packages/lexical-list/src/LexicalListItemNode.js
fantactuka: It used in code block, but its order is already matching (remove first, add after)

PR #2073, opened 2022-05-04
Remove size compare scripts

thegreatercurve:
These haven't been used in months. In fact, we're now using `danger.js` to detect build size differences, which was implemented recently by @acywatson. This can therefore probably be removed.


trueadm:
Hold on. I added those commands… give me some time to think why lol

trueadm:
Okay, from what I remember, I took this from the React repo that also uses Danger. It uses this script to prepare for Danger. So we probably want to keep this?

trueadm:
Yeah go for it. I commented it out then as Dangerbot doesn't work on private repos.

trueadm (approved):

trueadm (approved):

PR #2074, opened 2022-05-04
Split out commenting logic into different modules

trueadm:
This breaks out some of the commenting logic so it can be re-used by other modules or packages. Specifically:

- add `@lexical/mark` that contains the `MarkNode` and all associated logic to it.


thegreatercurve:
I think you need to add it to `scripts/build-flow.sh` and `scripts/npm/packages.js` as well.

fantactuka (approved):

Issue #2079, opened 2022-05-05
Bug: caret color changes when paragraph is empty 


NicolasRoelandt:
I've notice an issue that could be a bit problematic in dark mode. In the screenshot I changed the base text color to red because it was easier to show, but imagine the same with dark background with white text.
Imagine you set your base text color to red. The caret will change color once we type text:
<img width="217" alt="Screenshot 2022-05-05 at 10 17 08" src="https://user-images.githubusercontent.com/8594193/166894998-d2e717cb-a970-418d-a4a4-48c69d576047.png"><img width="201" alt="Screenshot 2022-05-05 at 10 16 23" src="https://user-images.githubusercontent.com/8594193/166894902-bf6ec363-2d8f-4277-bf57-0623717c9ceb.png">

The reason is that the caret color will take the text color, but when your paragraph is empty, there is no text color. 
You can work around it it by applying a caret-color style to your ContentEditable. In my opinion this is not great as you need to specify this color in two different places (one which is not even inside your theme).

I had a small look at this and my first attempt was to apply the text.base style to paragraph node : https://github.com/facebook/lexical/compare/main...caret-color but then I figured out we would have to do this for headings, code, etc, which wouldn't scale well. So I think this might need a deeper discussion.

Lexical version: 0.2.5

## Steps To Reproduce

1. Apply the changes in https://github.com/facebook/lexical/compare/main...caret-issue to make you text color red
2. Notice how the caret color is different once you start typing text

## The current behavior
Caret color does not depend of text color when text is empty

## The expected behavior
They should be in sync

trueadm:
Addressed in https://github.com/facebook/lexical/pull/2080 :)

NicolasRoelandt:
Following the discussion in https://github.com/facebook/lexical/pull/2080, it seems that it will be a won't fix as it's by design. How can I mark it as won't fix?

PR #2080, opened 2022-05-05
Fix character styles position + caret color

trueadm:
Fixes https://github.com/facebook/lexical/issues/2079.

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -910,6 +910,7 @@ select.font-family {
 
 .emoji {
   color: transparent;
+  caret-color: rgb(5, 5, 5);
   background-size: 16px 16px;
   background-position: center;
   background-repeat: no-repeat;
--- a/packages/lexical-playground/src/plugins/CharacterStylesPopupPlugin.jsx
+++ b/packages/lexical-playground/src/plugins/CharacterStylesPopupPlugin.jsx
@@ -34,7 +34,7 @@ function setPopupPosition(
 ): void {
   let top = rect.top - 8 + window.pageYOffset;
   let left =
-    rect.left + 230 + window.pageXOffset - editor.offsetWidth + rect.width;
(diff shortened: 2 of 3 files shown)


zurfyx (approved):

NicolasRoelandt:
I am not sure I understand how this fixes https://github.com/facebook/lexical/issues/2079. I've just tried to apply the changes to https://github.com/facebook/lexical/compare/main...caret-issue and the issue is still there.

trueadm:
@NicolasRoelandt It fixes the issue by not setting the caret-color, meaning the content editable cusor color will correctly that of the content editable. In your issue, it looks like you're apply a style to the text node. Given there is no text node when the paragraph is empty, how can it inherit the color? Try setting the color on the content editable itself.

trueadm:
@NicolasRoelandt I'm struggling to understand the exact problem here. This is working as intended, and is consistent with how other web text editors work (I checked Google Docs, Quip and ProseMirror).

If you're on an empty paragraph, the inherited cursor color should always be that of the parapgrah, as the selection point inherits that of the paragraph, given there is no text. When you have text content, selection moves to the text, rather than the paragraph and thus takes the cursor color from the text color of the text node.

So you shouldn't need to be trying to keep the text.base color in sync with the caret color, as you should never be setting the text color in `text.base`. In fact, I'm not even sure why we have `text.base`, we should get rid of it.

NicolasRoelandt:
@trueadm 
> I checked Google Docs, Quip and ProseMirror
I see. In Quip, if I change the text color the caret changes only once I start typing. But I feel it's a bit of a different scenario where the color is given by the user and not by the surface itself.

> as you should never be theming the text color in text.base. In fact, I'm not even sure why we have text.base, we should get rid of it.

This probably makes sense, as the color should be set in the content editable styles. I am just trying to think if there are styles you might want to apply **only** to text nodes and not to anything else

PR #2086, opened 2022-05-05
Add TabFocusPlugin

trueadm:
This plugin will attempt to restore selection to the previous location in an editor when you tab focus back into the editor.


zurfyx (approved):
So there's no way to know that the key caused focus..

PR #2088, opened 2022-05-05
Add a cache for selection.getNodes()

trueadm:
Today, we call `selection.getNodes()` a lot. One such case is when you do `node.isSelected()`. If you do this in a hot path, it can cause a lot of slow-down. We can however apply a caching heuristic to `selection.getNodes()`. Upon the first pass we can cache the return value on the selection object. We just need to ensure we invalidate this cache correctly so we don't wrongly use the cache when something might have changed. My heuristic is:

- When calling `set` on a point, or add/clear/delete, thus "mutation" the selection.
- When calling `$setSelection`
- When mutating a node, via `getWritable()`

I might have missed some cases but let's see how this works out. In perf testing this dramatically improves performance in some cases.

### DIFF
--- a/packages/lexical-rich-text/src/index.js
+++ b/packages/lexical-rich-text/src/index.js
@@ -550,7 +550,6 @@ export function registerRichText(
           (block) => {
             const indent = block.getIndent();
             if (indent !== 0) {
-              debugger;
               block.setIndent(indent - 1);
             }
           },
--- a/packages/lexical-rich-text/src/index.js
+++ b/packages/lexical-rich-text/src/index.js
@@ -550,6 +550,7 @@ export function registerRichText(
           (block) => {
             const indent = block.getIndent();
             if (indent !== 0) {
+              debugger;
               block.setIndent(indent - 1);
             }
           },
--- a/packages/lexical/src/LexicalSelection.js
+++ b/packages/lexical/src/LexicalSelection.js
@@ -529,16 +529,20 @@ export class RangeSelection implements BaseSelection {
     if ($isElementNode(lastNode)) {
       lastNode = lastNode.getDescendantByIndex(focus.offset);
     }
+    let nodes;
+
     if (firstNode.is(lastNode)) {
       if (
         $isElementNode(firstNode) &&
         (firstNode.getChildrenSize() > 0 || firstNode.excludeFromCopy())
       ) {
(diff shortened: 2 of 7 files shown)


packages/lexical/src/LexicalSelection.js
```diff
@@ -289,6 +299,9 @@ export class NodeSelection implements BaseSelection {
         nodes.push(node);
       }
     }
+    if (!isCurrentlyReadOnlyMode()) {
```
fantactuka, line 302: Is it because we don't reconcile selection for read mode? Wonder if we have editor in a readmode and then insert html or run markdown conversion for it - will it miss all the benefits of this cache?

trueadm: It's more for if you try and do `editorState.read(() => $getSelection().getNodes() )`. The selection is frozen and immutable here, so we can't mutate it to add the cache.

fantactuka: Ah right, it's not about editor's readmode, but .read vs .update. Thanks!

packages/lexical/src/LexicalSelection.js
```diff
@@ -510,7 +538,11 @@ export class RangeSelection implements BaseSelection {
       }
       return [firstNode];
     }
-    return firstNode.getNodesBetween(lastNode);
+    const nodes = firstNode.getNodesBetween(lastNode);
+    if (!isCurrentlyReadOnlyMode()) {
+      this._nodesCache = nodes;
+    }
+    return nodes;
```
fantactuka, line 545: Would it make sense to cache all of these cases (couple early returns above)? Would avoid `getDescendantByIndex` calls

fantactuka (approved):

packages/lexical-rich-text/src/index.js
```diff
@@ -550,6 +550,7 @@ export function registerRichText(
           (block) => {
             const indent = block.getIndent();
             if (indent !== 0) {
+              debugger;
```
acywatson, line 553: Huh?

fantactuka: Easter egg

trueadm: It was a test to see if you actually review my PRs :P

PR #2097, opened 2022-05-06
Show excalidraw nodes only in case elements have been created

pabloasc:
Only include the modal on open excalidraw. The rest of the elements include them only in case something has been already created with the tool.


zurfyx (approved):
LGTM, thanks!

PR #2109, opened 2022-05-08
Fix Firefox composition bug with emojis

trueadm:
We have reports of emojis being inserted multiple times on Windows + Firefox using the Windows emoji picker. We actually fixed this issue in the past: https://github.com/facebook/lexical/pull/738. However, we can't use setTimeout, as it means that we invalidate composition too late.

In order to make the composition sequence work the same in FF as it does in Chrome/Webkit – we can defer ending composition till the subsequent `input` event comes in. Thus making the sequence behave like `compositionstart` -> `input` -> `compositionend`, even though `input` occurs after `compositionend` in Firefox.

### DIFF
--- a/packages/lexical/src/LexicalEvents.js
+++ b/packages/lexical/src/LexicalEvents.js
@@ -136,6 +136,7 @@ let lastKeyDownTimeStamp = 0;
 let rootElementsRegistered = 0;
 let isSelectionChangeFromReconcile = false;
 let isInsertLineBreak = false;
+let isFirefoxEndingComposition = false;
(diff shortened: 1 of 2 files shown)


acywatson:
It looks like your previous commit to rebuild the package-lock ran the CI correctly. Did you consider just rebasing on top of that instead of disabling the cache? Ideally, I'd like to see if we can figure out what the actual problem with the cache is.

trueadm:
@acywatson This is based on the commit. I have no idea how or why the CI doesn't work for this PR, I've closed it and created 3 new branches and none work. So I just deleted the cache from the script and now they run…

acywatson (approved):

Issue #1242, opened 2022-02-08
Migrate the codebase from flow to typescript


justinhaaheim:
@justinhaaheim is tracking plans and progress in separate documents. If you have questions or want to get involved reach out to him.

ivan-ngchakming:
Hi, I'd like to help with this. Is there any updates on this and how can I get involve?

maasencioh:
### Packages
- [x] lexical-clipboard
- [x] lexical-code
- [x] lexical-dragon
- [x] lexical-file
- [x] lexical-hashtag
- [x] lexical-headless
- [x] lexical-history
- [ ] lexical-link
- [ ] lexical-list
- [x] lexical-mark
- [ ] lexical-markdown
- [ ] lexical-offset
- [ ] lexical-overflow
- [x] lexical-plain-text
- [x] lexical-playground
- [ ] lexical-react
- [x] lexical-rich-text
- [ ] lexical-selection
- [ ] lexical-table
- [ ] lexical-text
- [ ] lexical-utils

### Special Packages
- [ ] lexical
- [ ] shared
- [ ] lexical-yjs

FezVrasta:
Can you share anything on the reason behind the migration?

PR #2110, opened 2022-05-08
Migrate initial few packages to TypeScript

thegreatercurve:

- Migrated 3 fairly small packages to test initial TS migration:
    -  `lexical-code`
    -  `lexical-history`
    -  `lexical-dragon`
- Refactored build steps to include TS files.
- Add TS transforms for Jest unit tests.  
- Fixed several small typing bugs which were picked up by TS.

### Migration steps

I ran the below script to update the directories automatically, and then went through each one fixing the result TypeScript compiler errors, and Lint errors which prevents Vite from building the playground.

```bash
#!/usr/bin/env bash
DIRECTORIES=(
    "lexical-code" 
    "lexical-history" 
    "lexical-dragon" 
)

    # Preserve commit history
    for i in $(find ./packages/$directory -iname "*.js" -not -path "*__tests__*" -not -path "*dist*");
    done
    for i in $(find ./packages/$directory -iname "*.jsx" -not -path "*__tests__*" -not -path "*dist*");
    done

    for i in $(find ./packages/$directory -iname "*.ts" -not -path "*__tests__*" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is FindAndReplace/g" $i && sed -i "" "s/: LexicalNode>/ = LexicalNode>/g" $i;
    done
    for i in $(find ./packages/$directory -iname "*.tsx" -not -path "*__tests__*" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is FindAndReplace/g" $i && sed -i "" "s/: LexicalNode>/ = LexicalNode>/g" $i;
    done
    
done

npm run prettier:fix;
```

### Bundle sizes

All looks good with the bundle sizes. We're still making the most out of Google Closure Compiler: 
|package|main|TS branch|
|---|---|---|
|`LexicalHistory.prod.js`|3.2K|3.2K|
|`LexicalCode.prod.js`|8.3K|8.3K|
|`LexicalDragon.prod.js`|1.1k|1.1K|


thegreatercurve:
I'll test syncing this back to WWW tomorrow, but I think it should be okay as we've not actually touched any of the underlying type files, and it gets exported back to WWW as vanilla JS.

scripts/build.js
```diff
@@ -188,22 +191,24 @@ async function build(name, inputFile, outputFile, isProd) {
...
+        ? typescript({tsconfig: path.resolve('./tsconfig.build.json')})
+        : babel({
+            babelHelpers: 'bundled',
+            babelrc: false,
+            configFile: false,
+            exclude: '/**/node_modules/**',
+            plugins: [
+              '@babel/plugin-transform-flow-strip-types',
+              [
+                require('./error-codes/transform-error-messages'),
```
trueadm, line 206: Do we keep our error codes in TS?

thegreatercurve: Good shout. It's hard to say as none of these packages use invariant, but I've replaced the typescript rollup plugin and added a babel preset instead, which should do the compilation and still transform the error messages. The bundle sizes are the same as well.

thegreatercurve:
Just to confirm, this PR doesn't introduce any new lint or compilation errors from any of these packages. They've all been fixed.

packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts
```diff
@@ -12,18 +12,17 @@ import {
...
   $getSelection,
+  $isRangeSelection,
+  EditorConfig,
 } from 'lexical';
-import {initializeUnitTest} from 'lexical/src/__tests__/utils';
 
-// No idea why we suddenly need to do this, but it fixes the tests
-// with latest experimental React version.
-global.IS_REACT_ACT_ENVIRONMENT = true;
+import {initializeUnitTest} from '../../../../lexical/src/__tests__/utils';
```
acywatson, line 19: Can we set up aliases so we can get some nicer-looking paths here?

packages/lexical-history/LexicalHistory.ts
```diff
@@ -7,6 +7,4 @@
  *
  */
 
-'use strict';
```
acywatson, line 10: Why remove this? I didn't think it was actually related to typescript.

echarles:
Don't have much time to review or help on that, but just want to confirm moving to typescript will help me to navigate and contribute to the sources.

Quick question: it sounds like this PR leaves the existence of `d.ts` files (eg the packages/lexical-code/LexicalCode.d.ts). Usually the definitions are create automatically by the typescript compiler. Is it here to stay or just temporary?

thegreatercurve:
> Quick question: it sounds like this PR leaves the existence of `d.ts` files (eg the packages/lexical-code/LexicalCode.d.ts). Usually the definitions are create automatically by the typescript compiler. Is it here to stay or just temporary?

It should just be temporary. Ideally, I want to migrate everything and then set `emitDeclarationOnly` to true in the`tsconfig.build.json`, so it automatically creates the types, and then use [`flowgen`](https://github.com/joarwilk/flowgen) to automatically create the Flow types, but this should be in another PR.

Also, I don't want to let the type definitions between TS and Flow diverge at all, as our internal codebase uses Flow, so I want to make sure adding `emitDeclarationOnly` too soon doesn't introduce any unforeseen issues on our end.

echarles:
> It should just be temporary. Ideally, I want to migrate everything and then set emitDeclarationOnly to true in thetsconfig.build.json, so it automatically creates the types, and then use [flowgen](https://github.com/joarwilk/flowgen) to automatically create the Flow types, but this should be in another PR.
> Also, I don't want to let the type definitions between TS and Flow diverge at all, as our internal codebase uses Flow, so I want to make sure adding emitDeclarationOnly too soon doesn't introduce any unforeseen issues on our end.

Sure, better having a safe migration path to ensure the other flow packages not yet migrated are not impacted. I guess once all the packages will be typescript enabled, `emitDeclarationOnly` could be enabled and the existing `d.ts` residing in the `src` folders could be removed.

thegreatercurve:
> I guess once all the packages will be typescript enabled, `emitDeclarationOnly` could be enabled and the existing `d.ts` residing in the `src` folders could be removed.

That's my plan! At the moment the `.d.ts` files are maintained manually, hence the bugs, so the sooner I can automate that the better the type definitions should be.

packages/lexical-history/LexicalHistory.ts
thegreatercurve: It was creating a general `eslint` warning: 
```
'use strict' is unnecessary inside of modules
```

packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts
thegreatercurve: Yeah, I think I left it in as there was something weird going on with the Jest cache. It seems to be working now though.

thegreatercurve:
> I'll test syncing this back to WWW tomorrow, but I think it should be okay as we've not actually touched any of the underlying type files, and it gets exported back to WWW as vanilla JS.

Tested this on our internal codebase with a test sync, and all E2Es are passing.

packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts
```diff
@@ -12,18 +12,16 @@ import {
...
   $getRoot,
   $getSelection,
+  $isRangeSelection,
+  EditorConfig,
 } from 'lexical';
 import {initializeUnitTest} from 'lexical/src/__tests__/utils';
 
-// No idea why we suddenly need to do this, but it fixes the tests
-// with latest experimental React version.
-global.IS_REACT_ACT_ENVIRONMENT = true;
```
acywatson, line 20: So the tests run fine without this now? Nice.

packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts
```diff
@@ -44,7 +42,7 @@ describe('LexicalCodeNode tests', () => {
         expect(codeNode.createDOM(editorConfig).outerHTML).toBe(
           '<code class="my-code-class" spellcheck="false"></code>',
         );
-        expect(codeNode.createDOM({theme: {}}).outerHTML).toBe(
+        expect(codeNode.createDOM({theme: {}} as EditorConfig).outerHTML).toBe(
```
acywatson, line 45: Sorry if you answered this already, but is there no way to avoid casting this?

packages/lexical-code/src/index.ts
```diff
@@ -549,9 +558,13 @@ function getHighlightNodes(tokens): Array<LexicalNode> {
       const {content} = token;
       if (typeof content === 'string') {
         nodes.push($createCodeHighlightNode(content, token.type));
-      } else if (content.length === 1 && typeof content[0] === 'string') {
+      } else if (
+        Array.isArray(content) &&
```
acywatson, line 562: Seems like the calls to Array.isArray here are redundant, but I could see why you might want to leave it this way to avoid nesting conditionals.

packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts
thegreatercurve: Yeah, we can define it in the `globals` of the `jest.config.js` instead.

packages/lexical-code/src/index.ts
thegreatercurve: This is because of the types from Prism. Content can sometimes no be an array, so we should just check for that: 
```
const content: Prism.Token | (string | Prism.Token)[]
```

packages/lexical-code/src/index.ts
```diff
@@ -140,8 +143,8 @@ export class CodeHighlightNode extends TextNode {
 
 function getHighlightThemeClass(
   theme: EditorThemeClasses,
-  highlightType: ?string,
-): ?string {
+  highlightType: string | null | undefined,
+): string | null | undefined {
```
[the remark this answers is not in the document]
thegreatercurve: Yeah, definitely, but I'd say we maybe do this in bulk in a later PR. I want to keep the types as consistent with what they represent in Flow, so we don't get any issues when we migrate other packages.

packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts
thegreatercurve: We can and I shall!

PR #2113, opened 2022-05-08
Fix CI Caching

acywatson:
We only cache the node_modules in the project root. Before https://github.com/facebook/lexical/pull/2073 was merged, we depended on dangerjs, which had a dependency on https-proxy-agent, so by coincidence, playwright was able to resolve that module because it was there in the root node_modules as a dependency of dangerjs.

Since we use a fork of playwright locally in the packages directory, a different node_modules gets created in packages/playwright-core as part of the workspace install. We don't cache this, and we don't run an npm install on a cache hit, so https-proxy-agent isn't there when playwright needs it to download browsers.

This fixes that by caching the playwright-core local node_modules. 

It also adds a GitHub Secret to the cache key which can be used to manually bust the cache from the settings without making a code change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,13 +95,13 @@ jobs:
         with:
           path: |
             node_modules
+            packages/playwright-core/node_modules
             ~/Library/Caches/ms-playwright
           key: ${{ runner.os }}-v4-${{ hashFiles('package-lock.json') }}
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
       - name: Download browsers
-        if: steps.cache.outputs.cache-hit != 'true'
         run: npx playwright install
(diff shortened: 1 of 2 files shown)


trueadm (approved):

trueadm:
It looks like you just changed v4 to `{{ secrets.CACHE_VERSION }}`. Was there another fix I'm missing?

.github/workflows/test.yml
```diff
@@ -139,8 +140,9 @@ jobs:
         with:
           path: |
             node_modules
+            packages/playwright-core/node_modules
```
acywatson, line 143: @trueadm this is the actual fix

acywatson:
> It looks like you just changed v4 to `{{ secrets.CACHE_VERSION }}`. Was there another fix I'm missing?

yea I put a comment on the fix in one of the workflows

PR #2116, opened 2022-05-09
Fix all of the playground type errors

thegreatercurve:
- Fix all of the playground type errors.
- Remove old `$FlowFixMes`.


packages/lexical-playground/src/Editor.tsx
```diff
@@ -57,6 +57,13 @@ import YouTubePlugin from './plugins/YouTubePlugin';
 import ContentEditable from './ui/ContentEditable';
 import Placeholder from './ui/Placeholder';
 
+interface customWindow extends Window {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  parent: any;
+}
```
zurfyx, line 63: Why do we need a custom Window for this?

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawModal.tsx
```diff
@@ -40,11 +37,11 @@ type Props = {
   /**
    * Handle modal closing
    */
-  onHide: () => mixed;
+  onHide: () => unknown;
```
zurfyx, line 40: I guess we could just force `void` but that's fine too?

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawModal.tsx
```diff
@@ -40,11 +37,11 @@ type Props = {
...
   /**
    * Callback when the save button is clicked
    */
-  onSave: (elements: ReadonlyArray<ExcalidrawElementFragment>) => mixed;
+  onSave: (elements: ReadonlyArray<ExcalidrawElementFragment>) => unknown;
```
zurfyx, line 44: ^

zurfyx (approved):

packages/lexical-playground/src/Editor.tsx
thegreatercurve: `right` is undefined on the `frames`

packages/lexical-react/LexicalMarkdownShortcutPlugin.d.ts
```diff
@@ -6,6 +6,8 @@
  *
  */
 
+import type {Transformer} from '@lexical/markdown';
```
thegreatercurve, line 9: +1 for TS @fantactuka. This created a TS error as we were importing the wrong kind of Transformer type.

zurfyx (approved):

PR #2120, opened 2022-05-09
[WIP] Deserialize/serialize

trueadm:


packages/lexical-serialize/src/__tests__/unit/LexicalSerialize.test.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {initializeUnitTest} from '../../../../lexical/src/__tests__/utils';
```
acywatson, line 9: should be able to use absolute path here

packages/lexical/src/LexicalSerialize.js
```diff
@@ -0,0 +1,62 @@
...
+  T: BaseSerializer<SerializedNode>,
+>(editor: LexicalEditor): SerializedNode {
+  const serializer = editor._serializer;
+  if (serializer === null) {
+    invariant(
+      false,
+      'Serializer not defined. You can pass a new serializer via EditorConfig -> createEditor({ serializer })',
+    );
+  }
+  return editor.getEditorState().read(() => {
```
trueadm, line 28: It should allow to pass in an editorState as an argument too.

packages/lexical/src/LexicalSerialize.js
```diff
@@ -45,18 +48,22 @@ export function $deserializeRoot<
...
     );
   }
+  flushRootMutations(editor);
+
+  const pendingEditorState = createEmptyEditorState();
+  editor._pendingEditorState = pendingEditorState;
+  pendingEditorState._flushSync = true;
+  editor._dirtyType = FULL_RECONCILE;
+  editor._compositionKey = null;
   editor.update(() => {
```
trueadm, line 58: Should we being any update in `$serializeRoot`. Isn't the intention to return a fresh `EditorState`?

trueadm:
I also think it's important that each $deserialize and $serialize function on the nodes do not handle children, instead the Serializer class should do this.

zurfyx:
> I also think it's important that each `$deserialize` and `$serialize` function on the nodes do not handle children, instead the Serializer class should do this.

Shouldn't it be the responsibility of each node to understand whether to process children?

trueadm:
I don't think so. Its children should be formed by the tree logic. This is important if certain children should be skipped or re-ordered etc

packages/lexical/src/LexicalSerialize.js
```diff
@@ -70,28 +74,19 @@ export function serializeEditorState<SerializedNode: BaseSerializedNode>(
...
-  // $FlowFixMe Refine type
-  const serializedRootNode: SerializedRootNode<SerializedNode> = json;
-  const previousPendingEditorState = editor._pendingEditorState;
-  const pendingEditorState = createEmptyEditorState();
-  editor._pendingEditorState = pendingEditorState;
-  editor.update(() => {
-    // No need .deserialize already does that
-    // $getRoot().markDirty(); // Create pendingEditorState
+  const serializedRootNode = json;
+  return createEditorState(editor, (pendingEditorState: EditorState) => {
```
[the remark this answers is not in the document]
trueadm: Make it should be called internalCreateEditorState, so people don't use it elsewhere lol

Issue #2119, opened 2022-05-09
Bug: Binding to webcomponent shadow root fails


yuzuquats:
I'm working on converting lexical into an es6 module + webcomponent. I've been able to successfully bundle an es6 module (using webpack). Using it as a top-level component in the HTML document root works, but including it as a webcomponent doesn't.

Lexical version: 0.25

## Steps To Reproduce

1. import a bundled `lexical.bundle.min.js` file
2. bind Lexical to an editor in the root DOM (ie. document.getElementById("editor")), and observe that it works
3. bind Lexical to an editor in the shadow DOM (this.shadowRoot.querySelector("#editor"), and observe that it doesn't work.
  - The shadow dom's editor does receive attributes for `data-lexical-editor=true` among others
  - The shadow dom does not receive key inputs

Link to code example:

https://user-images.githubusercontent.com/9470098/167487085-048ee412-6564-47cc-91f4-285bd9342a58.mov

https://github.com/yuzuquats/lexical-web-component

Any help would be appreciated! I'm fairly new to npm so if there's anything obvious I'm missing I'd love to learn!

trueadm:
That's strange. Lexical attaches the event listeners directly to the content editable (apart from selection, which is document level). Do you maybe have a codesandbox I could try?

yuzuquats:
let me see if i can put together a code sandbox - the github should have a couple of quick steps to repro in the meantime. 

In any case, I've done some digging and it seems like the root cause might be because we're using window.event. Specifically, I can confirm that key events (onBeforeInput) is called but the [selection is null - thus early returning](https://github.com/facebook/lexical/blob/49543fbb3160af953a6f63b6ce7a4e8a65ad7b4b/packages/lexical/src/LexicalEvents.js#L369). In fact, it seems like selection is not actually being instantiated within the [`beginUpdate` method](https://github.com/facebook/lexical/blob/e27b5691a7540f528efb3232f0432b65e5cae92f/packages/lexical/src/LexicalUpdates.js#L602). And this seems to be because the way we're grabbing the current active event is through [window.event](https://github.com/facebook/lexical/blob/39ff2253c1b5e8a131ec7c24763c5fe407b94505/packages/lexical/src/LexicalSelection.js#L2188) which is [null/inaccurate for shadow trees](https://developer.mozilla.org/en-US/docs/Web/API/Window/event).

I'm not sure if this is the only issue with webcomponent but it's the immediate reason why `insertTextInput` events are being swallowed.

[...]

PR #2126, opened 2022-05-10
Track last event for when in Shadow DOM

trueadm:
Fixes https://github.com/facebook/lexical/issues/2119. In the Shadow DOM, `window.event` returns `undefined`. In this case, we need to use the last known event that the Lexical event system knows of in order to get some heuristic working. I'm unsure of if this will work in all cases, so leaving as as draft for now.


yuzhouu:
`window.event` is marked as **Deprecated**, why not just replace it?

trueadm:
Holding off on this, as seems there's far more work to get Shadow DOM support working with Selection.

PR #2130, opened 2022-05-10
Fix typo in playground: speect → speech

LukasMurdock:
Action button title/aria-label says 'speect' instead of 'speech'.


trueadm (approved):

trueadm:
Thank you!

PR #2131, opened 2022-05-10
Set up excalidraw package

rulikrulit:
https://user-images.githubusercontent.com/6481831/168819578-1f1ad824-42c6-401b-afac-61527a53006f.mov


packages/lexical-excalidraw/src/index.tsx
```diff
@@ -25,7 +25,7 @@ import {
 import * as React from 'react';
 import {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 
-import ImageResizer from '../../ui/ImageResizer';
+import ImageResizer from '../../lexical-playground/src/ui/ImageResizer';
```
fantactuka, line 28: Similar to Acy's comment, you don't want this package to depend on playground UI components

I wonder if it needs any built-in rendering at all as in most cases each app has its own modal component. Maybe it can accept `showDialog({ children, onClose, onDelete, ... })` property that when called will be responsible to render dialog, add onClose/onDelete handlers and render children inside of it. Then it also will be up to a user to implement "Are you sure you want to delete ... " things and outside click handler

zurfyx:
1. Let's remove playground dependencies - duplicate if necessary (for now - and we can abstract later)
2. Remove index.css. There are many open source solutions like styled but considering we're planning to abstract this later maybe just hardcode the CSS styles into JS?
3. Rebase on top of https://github.com/facebook/lexical/pull/2148

zurfyx (approved):

trueadm:
Looks good, let's get the merge conflicts resolved!

n2o:
I'd love to see this feature in the next release. It would be great if you @rulikrulit could fix the last conflicts 🥳

trueadm:
@rulikrulit Please can you fix the merge conflicts.

n2o:
Three checks to go :D Would love to use this in our tools 👍

thegreatercurve:
I'm going to close this as I don't think we should be importing any components or CSS from Playground, and the bulk of this package relies on Playground imports. 

It creates an unsustainable dependency, if someone tries to use the components in production in their app. Our team needs to be free  to make changes in the playground without fear of it affecting others. 

Users are still able to copy and paste any of the features of the playground for use in their own app.

PR #2133, opened 2022-05-11
Remove unused ref in Playground ToolbarPlugin

hanford:
This PR cleans up the Toolbar from the Playground, removing a ref that's never set.

Feel free to close this, if `mouseDownRef` is needed, but on master it's never actually used which caused me some confusion as I was replicating the Toolbar


trueadm (approved):
Nice one! This was probably missed in a refactor.

PR #2134, opened 2022-05-11
editorUpdate -> skipRootElementFocus

zurfyx:
All updates (except readonly and `collaboration`) are forced selection reconciliation. This is not ideal when the user programmatically wants to update the content without forcing focus.

This PR adds an optional update flag, `skipRootElementFocus`, that will skip the root element focus when passed as true.

It is worth noting that the use of the flags is unideal for a11y reasons. For example, when you add an image you want to refocus the content to understand the changes but there may be edge cases where it's still suitable (and also was a demanded feature).

### DIFF
--- a/packages/lexical/src/LexicalReconciler.js
+++ b/packages/lexical/src/LexicalReconciler.js
@@ -812,15 +812,13 @@ function reconcileSelection(
   const focusOffset = domSelection.focusOffset;
   const activeElement = document.activeElement;
   const rootElement = editor._rootElement;
-  // TODO: make this not hard-coded, and add another config option
-  // that makes this configurable.
-  if (
-    editor._updateTags.has('collaboration') &&
-    activeElement !== rootElement
-  ) {
+  const rootElementHasFocus =
+    rootElement !== null &&
+    activeElement !== null &&
+    rootElement.contains(activeElement);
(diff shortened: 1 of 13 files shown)


trueadm:
Won’t this cause selection to be out of sync from the editor state?

trueadm:
@zurfyx I think I'd like to know more about this in maybe a VC at some point.

> This PR also fixes (WIP) update bundling when tags are present, in particular a current issue with collab and bootstrap. It is dangerous to merge updates with different tags because of batching, the next update potentially does not want to share the same behavior.

I think this should be another PR, no?

packages/lexical/src/LexicalReconciler.js
```diff
@@ -881,10 +879,6 @@ function reconcileSelection(
     );
   }
 
-  // Diff against the native DOM selection to ensure we don't do
```
trueadm, line 884: Do you want to remove this comment?

ed0wolf:
@zurfyx what is the status of this PR, is it dead or can we expect to get it merged?

This is causing us some issues and there don't seem to be any work arounds that don't involve some pretty involved focus management from other elements.

trueadm:
So many changes have happened since, it might be easier to start this PR from scratch. Closing for now.

PR #2140, opened 2022-05-11
Update Lexical.d.ts#dispatchCommand type

hanford:
This PR corrects the `dispatchCommand` type, making payload optional

This allows consumers to not explicitly pass in undefined which seems like a reasonable DX improvement.

i.e. https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx#L624

Feel free to close if I'm missing something!

### DIFF
--- a/packages/lexical/Lexical.d.ts
+++ b/packages/lexical/Lexical.d.ts
@@ -148,7 +148,7 @@ export declare class LexicalEditor {
     klass: Class<T>,
     listener: Transform<T>,
   ): () => void;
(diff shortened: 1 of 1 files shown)


zurfyx:
This is intentional https://github.com/facebook/lexical/pull/2030. While I do agree your premise, the reason why we introduced this is because `?T` may not always be correct. Let's say your T is `boolean`, you could potentially omit it and produce a bad payload. Let me know if you find a better way to type this

PR #2143, opened 2022-05-11
Fix @lexical/rich-text package.json

hanford:
The latest version 0.2.8 doesn't work for me, I assume it's because LexicalRichText.ts doesn't exist in the bundle


thegreatercurve:
Larger fix here #2144

PR #2144, opened 2022-05-11
Fix extensions

acywatson:
These entrypoints should point to the js files in the bundle.

### DIFF
--- a/packages/lexical-code/package.json
+++ b/packages/lexical-code/package.json
@@ -9,7 +9,7 @@
   ],
   "license": "MIT",
   "version": "0.2.8",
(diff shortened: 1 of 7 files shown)


thegreatercurve (approved):

hanford:
@acywatson will a new release happen after this is merged or should I patch the packages for now? Also thanks for the speedy fix

thegreatercurve:
> @acywatson will a new release happen after this is merged or should I patch the packages for now? Also thanks for the speedy fix

I'm going to do another release now, so it should be fixed in less than an hour.

PR #2146, opened 2022-05-12
Replace element node with list item instead of appending. Fix #2142

### LINKED ISSUE #2142: Bug: Stuck in List after changing blockTypes
https://www.loom.com/share/1900dbcd34c54485aff35984bff4efcb

Lexical version: 0.2.7 (latest at time of report)

## Steps To Reproduce

On the playground:

1. Empty out editor state
2. Create a bulleted list (`- foo`)
3. Turn list into `Quote`
4. Turn Quote back into bulleted list

## The current behavior
That should be enough to reproduce the bug, I'm now "stuck" in a list am not able to remove the list without a page refresh

## The expected behavior
I should _leave_ the bulleted list upon clicking delete on an empty list

fantactuka:
Fix for #2142

### Before:

https://user-images.githubusercontent.com/132642/167981375-624f4bb0-2cde-49f9-9bfc-7231008183fd.mov

### After:

https://user-images.githubusercontent.com/132642/167981418-2619bc3d-305d-4ab0-bf62-50995b739a69.mov

### DIFF
--- a/packages/lexical-list/src/formatList.js
+++ b/packages/lexical-list/src/formatList.js
@@ -146,12 +146,13 @@ function createListOrMerge(node: ElementNode, listType: ListType): ListNode {
   const previousSibling = node.getPreviousSibling();
   const nextSibling = node.getNextSibling();
   const listItem = $createListItemNode();
+  listItem.append(...node.getChildren());
   if (
     $isListNode(previousSibling) &&
     listType === previousSibling.getListType()
   ) {
-    listItem.append(node);
     previousSibling.append(listItem);
+    node.remove();
     // if the same type of list is on both sides, merge them.
     if ($isListNode(nextSibling) && listType === nextSibling.getListType()) {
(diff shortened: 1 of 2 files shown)


packages/lexical-list/src/formatList.js
```diff
@@ -146,12 +146,13 @@ function createListOrMerge(node: ElementNode, listType: ListType): ListNode {
   const previousSibling = node.getPreviousSibling();
   const nextSibling = node.getNextSibling();
   const listItem = $createListItemNode();
+  listItem.append(...node.getChildren());
```
trueadm, line 149: Avoid spreading into append, use for loop instead

fantactuka: Updated w/ .splice() to prevent spread. Once TS migration & linked lists changes are done will do a PR to support append(Array<Nodes>) as it touches plenty of files

acywatson (approved):
This looks good, thanks.

PR #2151, opened 2022-05-12
Migrate LexicalClipboard to TypeScript (5/n)

thegreatercurve:

- Migrated 1 package solo as it created some type errors with core:
    -  `lexical-clipboard`

### Migration steps

I ran the below script to update the directories automatically, and then went through each one fixing the result TypeScript compiler errors, and Lint errors which prevents Vite from building the playground.

```bash
#!/usr/bin/env bash
DIRECTORIES=(
    "lexical-clipboard"
)

    # Preserve commit history
    for i in $(find ./packages/$directory -iname "*.js" -not -path "*dist*");
    done
    for i in $(find ./packages/$directory -iname "*.jsx" -not -path "*dist*");
    done

    for i in $(find ./packages/$directory -iname "*.ts" -not -path "*__tests__*" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is FindAndReplace/g" $i;
    done
    for i in $(find ./packages/$directory -iname "*.tsx" -not -path "*__tests__*" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is FindAndReplace/g" $i;
    done
    
done

npm run prettier:fix;
```

### Bundle sizes

All looks good with the bundle sizes after running `npm run build-www`. We're still making the most out of Google Closure Compiler: 
|package|main|TS branch|
|---|---|---|
|`LexicalClipboard.prod.js`|6.2K|6.2K|


packages/lexical/src/nodes/LexicalElementNode.js
```diff
@@ -131,12 +132,9 @@ export class ElementNode extends LexicalNode {
     }
     return node;
   }
-  getDescendantByIndex(index: number): LexicalNode {
+  getDescendantByIndex<T: LexicalNode>(index: number): null | T {
     const children = this.getChildren();
     const childrenLength = children.length;
-    if (childrenLength === 0) {
-      return this;
```
thegreatercurve, line 138: Took this out so the method behavior is consistent with other descendant find methods - i.e. doesn't return itself if no child is found. Happy to add back in if there's a reason it needs to be here though.

packages/lexical-clipboard/src/__tests__/unit/clipboard.test.ts
```diff
@@ -47,11 +56,17 @@ function setAnchorPoint(point) {
...
 function setFocusPoint(point) {
   let selection = $getSelection();
+
   if (selection === null) {
     const dummyTextNode = $createTextNode();
     dummyTextNode.select();
     selection = $getSelection();
   }
+
+  if ($isNodeSelection(selection)) {
```
zurfyx, line 66: Why are we checking for node selection exclusively here? What about grid selection?

packages/lexical/src/nodes/LexicalElementNode.js
```diff
@@ -56,7 +56,7 @@ export class ElementNode extends LexicalNode {
     const self = this.getLatest();
     return self.__indent;
   }
-  getChildren(): Array<LexicalNode> {
+  getChildren<T: Array<LexicalNode>>(): T {
```
zurfyx, line 59: Any T other than LexicalNode may be incorrect, children can be of multiple types. I'd say let's stick to the previous version. Same in the ones below

thegreatercurve: The colon constricts the generic type that can be accepted, it's not just a default. So, if I try to get an array of strings I see the below type error: 

image

packages/lexical-clipboard/src/__tests__/unit/clipboard.test.ts
thegreatercurve: Because only a NodeSelection doesn't have a focus property on the class: 
image

trueadm (approved):
Nice work!

PR #2152, opened 2022-05-12
Add Toolbar Color Picker 

### LINKED ISSUE #892: Color picker
Color picker on toolbar

- [ ] Text color
- [ ] Highlight color

WinmezzZ:
#### Fixes
#892 

#### Features
- add text-color picker
- add bg-color picker

#### Screens
https://user-images.githubusercontent.com/40154766/168049431-438d131d-e920-491b-8793-6c038907f82b.mov


fantactuka:
Looks great!
Icon is a bit too similar to formatting dropdown (A vs Aa), but can't see any alternatives in bootstrap icons, maybe other folks will have suggestions

fantactuka (approved):

WinmezzZ:
> Looks great! Icon is a bit too similar to formatting dropdown (A vs Aa), but can't see any alternatives in bootstrap icons, maybe other folks will have suggestions

I think so.

 Maybe you can provide a better icon for it.

WinmezzZ:
What's wrong with checks?

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -57,12 +62,14 @@ export default function DropDown({
...
     }
-  }, [dropDownRef, buttonRef, showDropDown]);
+  }, [dropDownRef, buttonRef, showDropDown, stopCloseOnClickSelf]);
 
   return (
     <>
       <button
-        aria-label={buttonAriaLabel || buttonLabel}
+        aria-label={
+          typeof buttonLabel === 'string' ? buttonLabel : buttonAriaLabel
```
acywatson, line 71: This is what's breaking all the tests - some of the tests rely in finding elements by ARIA label, but the components also provide buttonLabel. With this change, you're overriding all the buttonAriaLabels with the value of buttonLabel (unless buttonLabel is a react component, which it isn't in every other case except yours).

I'm not really sure why you'd want the ARIA label to be JSX anyway. I would change this and just pass in button label and aria label separately.

acywatson (changes requested):

packages/lexical-playground/src/ui/DropDown.tsx
WinmezzZ: Got it, thanks.

I will fix this

WinmezzZ:
@acywatson 
Had been changed but also breaked the test?

WinmezzZ:
Need review changes

acywatson (approved):
Looks right to me now, thanks!

Would love to see an e2e test or two for this, but since it's relying on existing core library functionality and not exposing a new plugin for users to consume, I think we are OK without it for now.

Thanks for taking the time to put this together!

PR #2156, opened 2022-05-12
anotate onError is not optional type for who use vanilla-js in docs

alajillo:
Instead of setting default onError on config  in createEditor.  i edit some docs for who use vanilla-js that onError is not optional. so include onError on config.

### DIFF
--- a/packages/lexical-website-new/docs/getting-started/quick-start.md
+++ b/packages/lexical-website-new/docs/getting-started/quick-start.md
@@ -21,6 +21,7 @@ const config = {
   theme: {
     ...
   },
+  onError : console.warn // onError is not optional type
 };
 
 const editor = createEditor(config);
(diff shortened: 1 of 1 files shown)


packages/lexical-website-new/docs/getting-started/quick-start.md
```diff
@@ -21,6 +21,7 @@ const config = {
   theme: {
     ...
   },
+  onError : console.warn // onError is not optional type
```
zurfyx, line 24:
```suggestion
  onError: console.error
```

zurfyx (approved):
Should we make this an error instead? While Lexical can recover from most errors (will restore previous EditorState) this is really bad

packages/lexical-website-new/docs/getting-started/quick-start.md
alajillo: thank you! okay i'm going to fix it :)

PR #2157, opened 2022-05-12
Add unstable serialization logic for node JSON parsing

trueadm:
UPDATE (co-authored by @acywatson):

This is now ready to review. The typings were a bit fiddly and still may not be perfect, but I think we're in a place now where we can review, hopefully merge, and iterate.

To restate the main goal here - we're trying to provide a flexible serialization/deserialization API and in the process, to decouple our internal state representation from the serialized storage format we provide/recommend out-of-the-box. Success means we have given the core team AND maintainers of Node subclasses in userland the flexibility to implement backwards compatible (additive) changes to node serialization "schemas".

We have two slightly different approaches for TypeScript and Flow, but the main challenge is leveraging inheritance to create a good developer experience around serialization, while also creating as much type safety as we can to prevent errors that might cause data loss or loss backwards incompatibility between two versions of a serialized editor state.

## Flow

Flow looks like this: [try flow](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDKeATgJYCGMZAXngCYCiMeAtngHYYByc9eAHmLkqtBr34A+MAF4wAb1RgwAYwAWZGPRKcAXGACCJEhSxDSlanXoS8kgNxKw9MjpUYycDvoDkMDCQ+YAA+YD4kGDBBoRwArjAwjspQcCSsFBi+LFAY0WEqnBikeeFkAOZquSFhAFaxAM4eUFglPklgZBz8XPpxrABGpI4AvugqMBT19WDMbIW2Ck4A+kvqmtp6hsamAo3kHGUOy0subh5eWQElEVHVcQntKylpGVl4OSUFXMXVpRVVoR8dUaZGarUeS063UyYD6gxIjiceAAHjhUhgAFKEADy3HMIis4j4dgAFABKfTCSxiJgsdhcWz46nWWzSRTKZQ6DCxEgcRYcjlrLQ6bxgADaAF0ADROAWnPDuTyi+4wGUC5KpdIwnw+NXqqGFfQABj1Auw+F8eDphV1so5ADdSPUlfoAIx24btUajVAotERMDmghU0TWAAKFBMZRMODUjJDhJsxOkcicCZpEajMbjxIAaq6maGiVJqunw5GKNGKLHbLmAEyFxOs9C+1HowO4YMWIv0TOV7O1gtl4t2WT8sAAOinw9pcwZxMbNNZeqDvhwFarsdtykdJGdFzArr1U4nqFGfvbQaI3cTfc3Of49cXLOTY-Zk+nN4zG4HeaHX5fKQ9RUOA4BgMMSDgHB9D2Toyj1Xd91FOtjynM8xgmKYwDvX9+DAFEii6aZZnpHhiXHFYQLAiCoJggI4MRAULwiLFcWfEdJHJSkAIYEj5gXGdWXHTk8G5XlhI5E96lifASAnZjMRxbhyVNDlVzCdcs2rNRt3VKjwMg6DAw0eoJ0o0CDKg1SdydF0wDrD0vXQZRGgyMgVA6Vh-UUtjBOTEk7XqHik34biCW-LSawE4LlycClsJ-bSFnfZQQI4RpYXIuQODwJAcKS4lyXaZQOGJMylmeLUxyC8LALwCdKoyYrMv4cqDS4aqYrK9qMGa0rWpWeVFS8TrapHCchvODhmtBMASRq5lxsQpVZBkORXTJCSSrK8zqMMscfH4KAKHiXJmuGfCYCCjooDmhae1sCdlpGta5DrTaUo5fr6t2yycFGxaQvq-SaJwc67S5Hk+W+pyfQUjt8GvMb6AAYQs-Kov4digZTccTxnDGH0EPygKcHQqBgLA0b22iwFgg4V07Xx9IAWk0-ttN0k90NQcZJmmanwMSzGCAIzh6GmQnkuOcmEip9HDLo-Yyma4BgDAAB1AgVAoDgfAwfDdzASHxN1rAMA0A4wFYVICHqfAVFBdzjN1pHAb4+c8LUUg8FUtW6c6AowDNi24LABAyAt4OwAAImejgY7AVJY6DROSQQDR1GDnRYTwBgGEDOBjfeToCAtsvOw6PkyC89FWO4Ta7X9iOEmtigAGsCEjzy0SmMh+hYQv8ORQj6DAKAeXLuS7QU+vsdZLi3Z7D2yKxknR0+k2+U+5QpJk0h5LbFilJUu1lHUnxWfZ+9dIFWXKcF0HfFvh1bIPd0BU9JwfRcjA3I8mu3k57r04k3dWWtVCu3XFhCmYBe7OgHnbfeJA+ZYSvJ0MA3sc66zHjoKApcwD9EmAXEaV4cFgHjqeAU90mzEjCoDR+wsibz1fKEAmTD4xdVJsoeKjDIpEwkmlDK30xw5SQHwjmIsip2m+uVRqBs5A0KXGVeRfUdqQi6IUAGD1uqaK4GogaJxXAKimto2hrVJpKhmrdeaXD6pXlevkCybMmE+A+mfOmdiJz33ljTHAHpLrXR3mAf2eDS6kXDqkdu0xJjDwdkUegHilF1W8XgCmvi-rNWULNWxyNHrx1Wutdx6oOTJPGiDQyWSvrqIqVBA6R0Tr+DaB4i6Vpro5LKUDJ6b8+SOPeltaphjan-UUV44ZVTRifwhqJKGLU8Cw1QEAA).

We use type spreading and inexact types to create flexible type definitions that can be refined in the subclass definition of importJSON (the argument it receives is a union of all serialized subclass types, which can be refined in the function logic). For exportJSON, we unfortunately can't specify a return type more specific than the base class type. Otherwise, the next subclass will not be able to align with the exportJSON signature in the superclass.

This approach is not perfect, but if you play around in the example above, you can see that it provides decent soundness via the importJSON refinement, which is important for trying to prevent mistakes int the adapter logic for various versions. exportJSON is looser, and we'll recommend unit tests to verify the schema for that one (I added some for our core nodes in this PR).

## TypeScript

TypeScript looks like this: [TypeScript Playground](https://www.typescriptlang.org/play?ssl=59&ssc=1&pln=60&pc=1#code/C4TwDgpgBAymBOECGATAPAFQIwBooYCYA+KAXigG8oBtAaSgEsA7KAUQA8BjAGwFcUIaANYQQAewBm+XFBHiphIgF0AXNLpKoAXygAyfAQDcAKGOhIsCPAZJuDAF4QUAGQjsGnWwDkxAspWMoKHMINQBnYGsmAHMcQKgANyswhjEmNSZeAFsAIys4rVMQy2tbBydWbggsiCZgHz9yOERUNAp4zgALBm4URHSoAEF4eCQQNBgrGztHFzcPb18IIhMglAZETmBUgYBybkjdqAAfKF34YG4j08zublWoCTF4LKRgNX2ICWBrs85a4BWX7nBjRTo-E5nABWvAiDAkIF2D2YAjqGWyeXgJi0eEmpRmTlc7k83Aay1Mxh4SDCYTYVRqdTJASCbjAz2AACkYAB5LwACgAlGo8dNyihKtUAUyqPEgohgLx4Cx2kFVVAuj0+rU1NQlHhZar1pttml0Xc4mqgk8Xm8PlVvrt9ZbGExUe8oAAGC2WkIfCD0gGOg1BJLwFKmqBYA06eKFQrGVns4LgaAisqzACKvDEgKZTQQyHQKuTkA+AEds4CkfFQ+GBlgHhWc6EoBEotFsbipumKgHGUsiKYqTSoFnm0y3IDXbSJQz6ktmVAAPpLpuA8KRZjRfy7Tr+7hiass9hsi5c3mC4XdgkoMe5hcytXyxXK4NQAB0n7CvEg8HfibPHl+QFb01QAejAqAfFbTcthLLdHmeEtoCQ2sdjwXgmGtBUmDef0QCgABaYI92QqBJCgAAiEJKMeBh-RQR5bG4WkciQTghGCMQYPbKA3wgqA90QKA8k8WEUKkYBSM4MQ7gYOsXRI6BmEBMMIC2HZkPfKAAHVoE8FgsgYbZojwkj5KgNkaQYHJuAInIQH4yCUDEBCkCgeBMO2Gp1T3DioCQUzmAiJSPIgBUlScMiEhsKBonCjAU0UqToHYhVbFba8HDeTSD2iDx3ycqAAAUxGs2zoGAbiXNCsIkB861EmSdDLObOppjswrnV9M41wgINnTQiNcDfPq1Ck+T3xXProz4oJ4yCBgslPTkgL5MIstmMkr3xMU7wgMkhVHSsDofA0ZKYEKmAXcgmAgAB3Y7xyWQUHiCa6BCm1cTv8Dbdq2pZ3z6t6woilgPogB441MADgDItMbwAYTEWT9rJAA1LB-GaQs2niHrdhk2TCL6gamrDHY1BGoIie4AAxNLnnRXIrGMHESlFTMTrJQcExPJNigRsVkdR7mlnRghsYLVpiyGgYCG9dhIiQEX6cZ+BmcxAou3+pxVbR8WsF5wXNr1lHuANxoOZ7FB9bFgRMchIXZjt56HaMIduGpWlXfvPxJ1qFBaUt6BixXWmGa2JmoEyFn4H8D0Hlh89gJ2zmzdFt3Q7m1VnyVRc1U-d9v1-f9+cAi8QLfAnaZJk6ybVOW1AVt8I-V8bujCL626j+BZtjUxFuW9kU-W02UG262kfNkOjpDgv1TSK6bpjh6oF906BFeg0Ia+vrfvHskgZOkHd-D83I6q+PyD+9OJ8Bnur5B+EoD5XYldGVXL+eI5mEy3X74CAFAvIIt8bZHw-irC+6sQaFCfOFF8MclhQzZhSAQVJhIXRCsAAAzGnG2G8ySGCAA) 

The approach here is to implement a Spread-like utility type with override functionality for collisions, similar to how it works in flow. Using TypeScript, we can be more specific about the return value of exportJSON, which actually helped me catch a few bugs in my own code in this PR. On the downside, we can't leverage disjoint unions to do refinement with TypeScript (or at least I haven't figured out how) so the version property is actually far less useful. Instead, we need to use "in" type guards to refine the types.

The next steps here are:

- [ ] Extensively test the actual end-to-end json export/parseEditorState process (via the unstable_ editor APIs)
- [ ] Write more unit tests to verify the schemas of the other nodes we expose in other packages.
- [ ] Documentation on how to use these APIs safely and correctly.

---
Extends from https://github.com/facebook/lexical/pull/2147 and https://github.com/facebook/lexical/pull/2120.

This PR introduces two APIs on all nodes. `exportJSON()` and `importJSON`. Additionally, there are two new unstable prefixed methods that aim to replace their originals in the future, they are `unstable_parseEditorState` on the `editor` object and `unstable_toJSON` on the `EditorState` object.

Another thing that PR does, it ensure that we using the public API methods on nodes rather than the private values, when importing and exporting JSON. This should help improve consistency and remove the tendency to store data that is more of an implementation detail.


packages/lexical-playground/src/appSettings.ts
```diff
@@ -27,7 +27,7 @@ export const isDevPlayground: boolean =
 
 export const DEFAULT_SETTINGS: Settings = {
   disableBeforeInput: false,
-  emptyEditor: isDevPlayground,
+  emptyEditor: false,
```
acywatson, line 30: Can we change this back?

trueadm: Ah, I keep doing that lol.

packages/lexical/src/LexicalNode.js
```diff
@@ -610,6 +610,22 @@ export class LexicalNode {
     return null;
   }
 
+  // $FlowFixMe: TODO
+  exportJSON(): Object {
+    invariant(false, 'exportJSON: base method not extended');
+  }
+
+  static importJSON(
+    // $FlowFixMe: TODO
```
acywatson, line 619: Can this actually be avoided?

packages/lexical/src/LexicalEditor.js
```diff
@@ -584,6 +590,16 @@ export class LexicalEditor {
...
   }
+  unstable_parseEditorState<SerializedNode>(
+    maybeStringifiedEditorState: string | SerialzedEditorState<SerializedNode>,
+    updateFn?: () => void,
+  ): EditorState {
+    const serialzedEditorState =
+      typeof maybeStringifiedEditorState === 'string'
+        ? JSON.parse(maybeStringifiedEditorState)
+        : maybeStringifiedEditorState;
+    return unstable_parseEditorState(serialzedEditorState, this, updateFn);
```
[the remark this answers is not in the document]
trueadm: I envisaged that once we give some time for people to add the import/export JSON methods to their nodes, then we can swap them around.

packages/lexical/src/LexicalNode.js
```diff
@@ -610,6 +610,22 @@ export class LexicalNode {
...
+    serialzedNode: Object,
```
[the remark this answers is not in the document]
trueadm: For this, it was intentional. Flow didn't like trying to apply an interface here, TS did though. Given no one extends LexicalNode directly, this should be fine, as we don't expose LexicalNode.

packages/lexical/src/LexicalUpdates.js
```diff
@@ -279,6 +285,69 @@ export function parseEditorState(
   return editorState;
 }
 
+type InternalSerializedNode = {
+  children?: Array<InternalSerializedNode>,
+  type: string,
+};
+
+function parseSerializedNode<SerializedNode: InternalSerializedNode>(
```
acywatson, line 293: Seems weird to put this in Updates. Maybe we should just put it on the editor?

packages/lexical/src/LexicalUpdates.js
```diff
@@ -279,6 +285,69 @@ export function parseEditorState(
...
+        registeredNodes,
+      );
+      node.append(childNode);
+    }
+  }
+  return node;
+}
+
+export function unstable_parseEditorState<SerializedNode>(
+  serialzedEditorState: SerialzedEditorState<SerializedNode>,
```
acywatson, line 319: typo here

packages/lexical/src/LexicalEditorState.js
```diff
@@ -15,25 +15,34 @@ import type {
...
   _selection: null | ParsedSelection,
 };
-
+// TODO: deprecated
 export type JSONEditorState = {
   _nodeMap: Array<[NodeKey, LexicalNode]>,
   _selection: null | ParsedSelection,
 };
 
+export interface SerialzedEditorState<SerializedNode> {
```
acywatson, line 42: typo here

packages/lexical/src/LexicalUpdates.js
trueadm: It needs access to the closure lexical variables, that's why it's in here.

packages/lexical/src/LexicalNode.js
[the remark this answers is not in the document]
acywatson: I meant it should be serializedNode, not serialzedNode

Re: the typing, I came to the same conclusion. We probably don't need a TODO if we don't plan to fix it. I agree that we can probably do better than this with TS, if that's what you're waiting for.

packages/lexical/src/LexicalNode.js
```diff
@@ -610,6 +610,22 @@ export class LexicalNode {
...
+  ): LexicalNode {
+    invariant(
+      false,
+      'LexicalNode: Node %s does not implement .importJSON().',
```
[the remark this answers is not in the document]
acywatson: Hmmm, not sure I'm following. In the example above if you call ReallyCoolNode.importJSON, it will delegate to CoolNode.importJSON, which will happily return an object missing the __reallyCoolFactor property.

packages/lexical/src/LexicalUpdates.js
```diff
@@ -279,6 +285,69 @@ export function parseEditorState(
...
+  const registeredNode = registeredNodes.get(type);
+  if (registeredNode === undefined) {
+    invariant(false, 'parseEditorState: type "%s" + not found', type);
+  }
+  const registeredNodeClass = registeredNode.klass;
+  const node = registeredNodeClass.importJSON(serialzedJSONNode);
+  const children = serialzedJSONNode.children;
+  if ($isElementNode(node) && Array.isArray(children)) {
+    for (let i = 0; i < children.length; i++) {
+      const serialzedJSONChildNode = children[i];
```
acywatson, line 314: typo lol

packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -96,6 +104,26 @@ export class ListItemNode extends ElementNode {
...
+  ): ListItemNode {
+    const node = new ListItemNode(serializedNode.value, serializedNode.checked);
+    node.setFormat(serializedNode.format);
+    node.setIndent(serializedNode.indent);
+    node.setDirection(serializedNode.direction);
+    return node;
+  }
+
+  exportJSON<SerializedNode>(): SerializedListItemNode<SerializedNode> {
+    // $FlowFixMe: Flow limitation
```
acywatson, line 118: I just noticed all these - this is the interface limitation in Flow that I ran into. So I guess the plan is to ignore all these for now and then just convert to TS

thegreatercurve (approved):
🔥

packages/lexical-code/src/index.ts
```diff
@@ -266,6 +303,24 @@ export class CodeNode extends ElementNode {
     };
   }
 
+  static importJSON<SerializedNode>(
+    serializedNode: SerializedCodeNode<SerializedNode>,
+  ): CodeNode {
+    const node = $createCodeNode(serializedNode.language);
+    node.setFormat(serializedNode.format);
+    node.setIndent(serializedNode.indent);
+    node.setDirection(serializedNode.direction);
```
tylerjbainbridge, line 312: Does this API solve the problem with new fields/deprecated fields? Is this where you'd add a default in some cases?

tylerjbainbridge (approved):

packages/lexical-link/src/index.js
```diff
@@ -67,6 +74,24 @@ export class LinkNode extends ElementNode {
...
+    node.setFormat(serializedNode.format);
+    node.setIndent(serializedNode.indent);
+    node.setDirection(serializedNode.direction);
+    return node;
+  }
+
+  exportJSON<SerializedNode>(): SerializedLinkNode<SerializedNode> {
+    return {
+      ...super.exportJSON(),
+      type: 'link',
```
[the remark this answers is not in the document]
tylerjbainbridge: Or maybe you'd want to be able to "override" a node by calling the plugin after? Not sure, feels potentially confusing and hard to debug.

trueadm (commented):
This looks awesome @acywatson. Don't forget to run the extract codes script though!

Update: updated codes.json

global.d.ts
```diff
@@ -0,0 +1,15 @@
+declare global {
```
thegreatercurve, line 1: There's already a `global.d.ts` in the `typedefs` folder.

global.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  interface Window {
+    twttr: {
+      widgets: {
+        createTweet(id: string, el: Element);
+      };
+    };
+  }
+}
+
+export type Spread<T1, T2> = {[K in Exclude<keyof T1, keyof T2>]: T1[K]} & T2;
```
zurfyx, line 15: How are other folks going to use Spread for their custom nodes?

global.d.ts
trueadm: Mind patching up this PR accordingly?

packages/lexical-code/src/index.ts
```diff
@@ -59,6 +62,24 @@ import {
 
 const DEFAULT_CODE_LANGUAGE = 'javascript';
 
+type SerializedCodeNode = Spread<
+  {
+    language: string | null | undefined;
+    type: 'code';
+    version: 1;
```
zurfyx, line 69: Do we still need version numbers? AFAIK it was proposed by the Flow team but later you pointed out you still had problems with inheritance (and refinement?)

packages/lexical-link/src/index.js
```diff
@@ -119,6 +145,12 @@ export function $isLinkNode(node: ?LexicalNode): boolean %checks {
   return node instanceof LinkNode;
 }
 
+export type SerializedAutoLinkNode = {
+  ...SerializedLinkNode,
+  type: 'autolink',
+  ...
+};
+
```
zurfyx, line 153: Is this the Flow approach? `...Extended` + `invariant`? Also, no version here?

packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -96,6 +105,23 @@ export class ListItemNode extends ElementNode {
...
+    node.setDirection(serializedNode.direction);
+    return node;
+  }
+
+  exportJSON(): SerializedElementNode {
+    return {
+      ...super.exportJSON(),
+      checked: this.getChecked(),
+      type: 'listitem',
+      value: this.getValue(),
```
zurfyx, line 121: [Opinionated] Mixing static + class is confusing:

When I write `importJSON` I have to think of indent but when I wrote `exportJSON` I do not. That makes it even harder than just having `static` methods, because now I have to think what I have as a part of the class and what comes as part of the prototype. And I have to understand everything anyway because the `importJSON` requires me to understand them all

packages/lexical/flow/Lexical.js.flow
```diff
@@ -843,3 +863,58 @@ export type EventHandler = (
  * LexicalVersion
  */
 declare export var VERSION: string;
+
+// Serialization
+
+export type SerializedLexicalNode = {
+  type: string,
+  version: number,
+  ...
```
zurfyx, line 872: Do we need the spread? I was hoping `{...Extended, something else}` would do the trick but pretty sure you've played with this extensively

zurfyx (commented):
LGTM, some thoughts:

1. You may want to add copy-paste to the todo list above. Ideally we want to add this new format to the list which can provide more fidelity than HTML when possible (namespaces match)
2. I've noticed that some nodes have version, others not. Added a comment to one of these occurrences
3. Can we have a test/example that showcases multiple versions? Ideally somewhere in the code base so that TS can attest this works as intended

global.d.ts
thegreatercurve: Yeah, this is probably my biggest concern as well. It's probably not great DevX to require users to have to use utility types. I wonder if we can make this just an implementation detail on some of the base classes.

acywatson: I agree, but this is sort of a limitation of TypeScript. Their team made the explicit decision to not natively support spread (or some other) syntax for this use case and they recommend using mapped types instead:

https://github.com/Microsoft/TypeScript/pull/13288#issuecomment-412230721

I think the concern here can be easily addressed in either of two ways:

1) We can export and alias or abstraction over this utility type like ```type ExtendedSerializedNode<T1, T2>```. I think it's pretty normal for users to import and use a library-specific type from that library (think React.Element or something).

2) Users can simply copy the utility type. Not ideal, but also not that big of a deal, IMO.

I don't think it's possible to make it an implementation detail in the base classes. The problem is that if you don't allow type field overrides for subclass serialized types, you can't further extend that class because the method signature can't be matched (at least in the experimentation I have done so far).

packages/lexical-link/src/index.js
acywatson: The missing version is a mistake, but yes the Flow approach can be an invariant, an if condition or whatever other method of refinement you want to use. The point is that you have to differentiate between the two potential argument types in order to access the properties of the subclass type.

packages/lexical-list/src/LexicalListItemNode.js
acywatson: I think I understand what you're saying, but I don't understand why it's a problem or how we might improve on this situation without introducing other trade-offs.

packages/lexical-playground/src/nodes/EmojiNode.tsx
```diff
@@ -47,6 +61,31 @@ export class EmojiNode extends TextNode {
...
+    node.setMode(serializedNode.mode);
+    node.setStyle(serializedNode.style);
+    return node;
+  }
+
+  exportJSON(): SerializedEmojiNode {
+    return {
+      ...super.exportJSON(),
+      className: this.getClassName(),
+      type: 'emoji',
```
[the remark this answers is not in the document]
acywatson: Yea, I actually mentioned this to @trueadm last night in chat. I was originally thinking why not just call the static getType method on the class during serialization, so you never have a mismatch. I came up with some reason not to do that, but I can't recall what it was. I concluded that we should use an invariant instead when the exported type doesn't match the class type, but I'm open to doing it either way.

packages/lexical/flow/Lexical.js.flow
acywatson: Yea, I tried to make this work multiple times/ways. IIRC, the problem comes in with inheritance. You get method signature collisions when you try to further extend an exact type.

acywatson:
> 1. You may want to add copy-paste to the todo list above. Ideally we want to add this new format to the list which can provide more fidelity than HTML when possible (namespaces match)

Yea, we originally discussed actually removing that serialization format altogether for copy and paste and just relying on HTML. However, I think the use case of copy and pasting into the same editor (e.g., you want to change the order of sections in a document you're composing) is probably common enough to warrant keeping this format.

> 2. I've noticed that some nodes have version, others not. Added a comment to one of these occurrences

Fixed.

> 3. Can we have a test/example that showcases multiple versions? Ideally somewhere in the code base so that TS can attest this works as intended

Sure - for now I would suggest you look at the TS and Flow examples linked in the PR description.

global.d.ts
thegreatercurve: Since the above comment, there's been a few more useful util functions which have been added to TypeScript than just `Exclude`, so I think we could do something like the below instead, using `Omit`, if the basic aim is to exclude the properties of the superclass if they exist on the sub class: 
```
Omit<T1, keyof T2> & T2
```
Feel free to merge. We can experiment later.

packages/lexical-list/src/LexicalListItemNode.js
```diff
@@ -262,8 +288,13 @@ export class ListItemNode extends ElementNode {
   }
 
   getIndent(): number {
+    // If we don't have a parent, we are likely serializing
+    const parent = this.getParent();
+    if (parent === null) {
+      return this.getLatest().__indent;
+    }
```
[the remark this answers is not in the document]
trueadm: That's correct. In serialziation, there might not be a parent (if you are serializing a detached child).

packages/lexical-playground/src/nodes/EmojiNode.tsx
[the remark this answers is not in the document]
trueadm: That's right. If we move to automating getType, then we'd also need to check if the class has its own exportJSON, and doesn't inherit it from the super class. Either way works.

PR #2158, opened 2022-05-12
Ensure selection is not prematurely nulled out on blur

trueadm:
When the user moves selection outside the editor, we should aim to keep the selection object the same within the editor.

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Focus.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Focus.spec.mjs
@@ -6,7 +6,14 @@
  *
  */
 
-import {expect, focusEditor, initialize, test} from '../utils/index.mjs';
+import {
+  click,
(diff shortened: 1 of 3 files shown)


thegreatercurve (approved):

acywatson:
Could we write a test for this?

GiacomoRandazzo:
Hi, could you share the rationale behind this decision?

I have the following situation and I'm trying to figure out the best way to go about it.

The style of a `DecoratorNode` changes when it is selected with a `NodeSelection`.
When the editor is blurred, the internal state's selection will stay a `NodeSelection`, and the style won't go back to normal.

The solutions I'm considering are:
- on editor blur, check for `NodeSelection` and update it to a `null` selection
- add `tabindex="-1"` to the `DecoratorNode`'s element, focus the element if the node gets selected. Then style the node wrt its focused state

PR #2160, opened 2022-05-12
Update LexicalMarkdownShortcutPlugin.d.ts

hanford:
In https://github.com/facebook/lexical/pull/2124 the `transformers` prop was made optional, but the type was never made optional. This fixes that!


fantactuka:
Thanks! Do you mind updating .flow and .jsx files too?

hanford:
@fantactuka updated!

fantactuka (approved):

PR #2162, opened 2022-05-12
Adjust selection to be after decorator node when moving selection to the end of decorator/linebreak

### LINKED ISSUE #2161: Bug: error replacing a TextNode for a DecoratorNode
There's an error when the node that is needed to be replaced is a DecoratorNode instead a TextNode. I'm using a code similar to this one (see the sandbox below):
```ts
useEffect(() => {
    return editor.registerNodeTransform(TextNode, (textNode: TextNode) => {
      let match = getHashtagMatch(textNode.getTextContent());

      let currentNode = textNode;
      while (match !== null) {
        let nodeToReplace: TextNode;
        currentNode.markDirty();

        if (match.start === 0) {
          [nodeToReplace, currentNode] = currentNode.splitText(match.end);
        } else {
          [, nodeToReplace, currentNode] = currentNode.splitText(
            match.start,
            match.end,
          );
        }
        const replacementNode = $createSampleLinkNode(
          nodeToReplace.getTextContent(),
        );
        nodeToReplace.replace(replacementNode);
        if (!currentNode) return;
        match = getHashtagMatch(currentNode.getTextContent());
      }
    });
  }, [editor]);
```

The current workaround is to add manually a selection
```ts
if (!currentNode) {
  replacementNode.selectNext();
  return;
} else {
  currentNode.select(0, 0);
}
```

Lexical version: 0.2.9

## Steps To Reproduce

In the sandbox below write a `#hastag`, it should be replaced in a transformer to a `DummyNode`

[Link to code example](https://codesandbox.io/s/lexical-replacement-selection-birosu?file=/src/dummy/DummyPlugin.js)

## The current behavior
I get two errors:
```
Uncaught Error: updateEditor: selection has been lost because the previously selected nodes have been removed and selection wasn't moved to another node. Ensure selection changes after removing/replacing a selected node.
```
and
``` 
Uncaught Error: Point.getNode: node not found
```

## The expected behavior
In this case replace any hastag with a dummy span

fantactuka:
Replacing selected node with decorator/linebreak does not adjust selection properly. It supposed to move selection to the end of newly inserted node, which didn't happen for non-text/element nodes

Fix #2161

### DIFF
--- a/packages/lexical/src/LexicalSelection.js
+++ b/packages/lexical/src/LexicalSelection.js
@@ -155,7 +155,7 @@ function $createPoint(
 }
 
 function selectPointOnNode(point: PointType, node: LexicalNode): void {
-  const key = node.__key;
+  let key = node.__key;
   let offset = point.offset;
   let type = 'element';
   if ($isTextNode(node)) {
@@ -163,6 +163,18 @@ function selectPointOnNode(point: PointType, node: LexicalNode): void {
     const textContentLength = node.getTextContentSize();
     if (offset > textContentLength) {
       offset = textContentLength;
+    }
+  } else if (!$isElementNode(node)) {
+    const nextSibling = node.getNextSibling();
+    if ($isTextNode(nextSibling)) {
+      key = nextSibling.__key;
+      offset = 0;
+    } else {
+      const parentNode = node.getParent();
+      if (parentNode) {
+        key = parentNode.__key;
+        offset = node.getIndexWithinParent() + 1;
+      }
     }
   }
   point.set(key, offset, type);
@@ -179,7 +191,7 @@ export function $moveSelectionPointToEnd(
     } else {
       selectPointOnNode(point, node);
     }
-  } else if ($isTextNode(node)) {
+  } else {
     selectPointOnNode(point, node);
   }
 }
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalSelection.js
```diff
@@ -164,6 +164,12 @@ function selectPointOnNode(point: PointType, node: LexicalNode): void {
     if (offset > textContentLength) {
       offset = textContentLength;
     }
+  } else if (!$isElementNode(node)) {
+    const parentNode = node.getParent();
+    if (parentNode) {
+      key = parentNode.__key;
+      offset = node.getIndexWithinParent() + 1;
```
fantactuka, line 171: @trueadm Not quite sure about this one:
When have `<decorator><text>` and calling `selectPointOnNode` to put selection after decorator node, then selecting next text node seems to be a better approach. But if node after decorator is not text (another decorator) or does not exist (decorator is a last child) then need to use decorators parent for selection. Q: is it better to address these cases separately, or always set selection on element node (decorator's parent) and let reconciler to adjust it

trueadm: It's correct to move it to the parent element node if the next sibling is not a text node.

trueadm (approved):

PR #2165, opened 2022-05-13
Apply colspan and rowspan from external tables

### LINKED ISSUE #2101: Bug: colSpan is ignored in Table cell node
Colspan and rowspan are not respected when you copy paste a table from external website . 

Lexical version: 0.2.5

## Steps To Reproduce

1. Go to a website like https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_th_colspan
2. Copy paste table to lexical playground 

 
Link to code example:

https://github.com/facebook/lexical/blob/main/packages/lexical-table/src/LexicalTableCellNode.js

1. The createDom method doesnt  attach colspan .
2.  convertTableCellNodeElement doesnt pass colspan
3.  GridCellNode doesnt have rowspan member variable as well . 

## The current behavior
Table's which have row and colspan gets murky when copy pasted

## The expected behavior
WYSWIG

mmarkelov:
close #2101


acywatson:
Thanks for the PR! Looks like a few issues with the tests - I think the assertions just need to be updated to accommodate the changes.

mmarkelov:
@acywatson yeah! My bad, forgot tests

packages/lexical-table/src/LexicalTableCellNode.js
```diff
@@ -37,6 +37,8 @@ export type TableCellHeaderState = $Values<typeof TableCellHeaderStates>;
 export class TableCellNode extends GridCellNode {
   __headerState: TableCellHeaderState;
   __width: ?number;
+  __colSpan: number;
```
acywatson, line 40: The only concern I have here is that this may technically be a breaking change since it will change the schema of the serialized editor state. We are working on a way to decouple the serialization format from the node properties, after which this won't really be a problem. In the meantime, it looks like we're only adding properties and we have defaults here that should make it backwards compatible.

acywatson (approved):
I'm good with this, but will let @tylerjbainbridge take a quick look as the owner of tables.

acywatson (changes requested):
Actually, looks like we still have test failures - the CI was showing a green check earlier and fooled me 😅

packages/lexical-table/src/LexicalTableCellNode.js
mmarkelov: @acywatson idk, i suppose that i can implement the logic when these default properties won't be added during the serialization

tylerjbainbridge (changes requested):
I'm pretty sure this is going to break selection and a lot of table operations (insert row/column, delete row/column, apply header state, etc) because they iterate through the table assuming each row has the same amount of columns using basic for-loops. There needs to be some new logic that accounts for colspan & rowspan added here. 

This feature is planned but will require changing a lot of core code within [`TableSelection`](https://github.com/facebook/lexical/blob/main/packages/lexical-table/src/LexicalTableSelection.js), [`GridSelection`](https://github.com/facebook/lexical/blob/84948d70fc499f8e638ed7c6e51eda3135c9fff5/packages/lexical/src/LexicalSelection.js#L334), and [`TableUtils`](https://github.com/facebook/lexical/blob/main/packages/lexical-table/src/LexicalTableUtils.js).

Happy to approve the PR once support is added in these places & the tests pass, but that's going to be a lot of work.

mmarkelov:
@tylerjbainbridge yeah, now. I see. Thanks for your comment. Will close it for a now, cause I tried to implement all cases, but it requires a lot of changes

PR #2166, opened 2022-05-13
Skip markdown links parsing

### LINKED ISSUE #2159: Bug: Image URLs that contain markdown formatting don't get parsed correctly when converting to/from markdown
Lexical version: 0.2.9

## Steps To Reproduce

This is reproducible on the Lexical playground:

1. Insert an image that contains some type of markdown formatting in the URL. For example, the following URL contains underscores (markdown italics): https://storage.googleapis.com/dev_papyrus_images/ae676267a8839fec36f0e29551c64369
2. The image is inserted fine:

image

3. Now, toggle to markdown, and the image URL also looks correct: 
image

4. Toggle back to HTML and the image URL gets malformed, since the underscores get parsed into italics:

image

## The current behavior
Conversions to/from markdown don't preserve the image, if the URL contains markdown formatting.

## The expected behavior
The URL should be parsed correctly, and thus the image should be preserved when converting to/from markdown.

mmarkelov:
close #2159

Actually don't know the best approach. There are multiple places where `URL_MATCHER` is used. So maybe it should be moved somewhere to shared directory?


packages/lexical-markdown/src/v2/MarkdownImport.js
```diff
@@ -219,6 +219,9 @@ function importTextMatchTransformers(
   }
 }
 
+const URL_MATCHER =
```
acywatson, line 222: As you pointed out, seems like this is used in more than one place, so we should probably centralize.

I'm thinking the lexical-utils package might be a good place. Maybe instead of exporting the regex, though, we just make a util function like ```isUrl(text: string): boolean```?

acywatson (commented):
Thanks for the PR!

One comment on the code organization, @fantactuka to review the implementation as the markdown SME.

acywatson:
Also, it would be great to have a regression test for this.

packages/lexical-markdown/src/v2/MarkdownImport.js
mmarkelov: I'm not sure if I did it in right way, cause i got some ts warnings, so I'm not quite sure about correct import

fantactuka:
Hi @mmarkelov,
thank you for the contribution! 

There was a similar issue in shortcuts plugin (handling markdown transformations as you type) - #2136 but the fix was not applied to import. I think we should also rely on `intraword` flag here instead of checking for url pattern (at least for the current implementation it's the way to go).

As of isURL helper - it sounds good to move it to shared, but I wonder if it's better to move RegExp only instead of wrapping it into helper: you might want to use it via `.test` or `.match` to get matching index. WDYT?

acywatson:
> As of isURL helper - it sounds good to move it to shared, but I wonder if it's better to move RegExp only instead of wrapping it into helper: you might want to use it via `.test` or `.match` to get matching index. WDYT?

I recommended a helper function instead of the raw regex because generally I think that regex can be a bit more confusing to deal as a utility than just some functions that do whatever you want them to do (isURL, getHostname, parseURL, etc...).

I actually don't mind either way here, though.

packages/lexical-markdown/src/v2/MarkdownImport.js
```diff
@@ -232,12 +232,16 @@ function findOutermostMatch(
...
     // before using match to find transformer
-    const fullMatchRegExp =
-      textTransformersIndex.fullMatchRegExpByTag[match.replace(/^\s/, '')];
+    const symbol = match.replace(/^\s/, '');
+    const fullMatchRegExp = textTransformersIndex.fullMatchRegExpByTag[symbol];
     if (fullMatchRegExp == null) {
       continue;
     }
 
+    if (textTransformersIndex.transformersByTag[symbol].intraword === false) {
```
fantactuka, line 241: It should not blindly skip non-intraword transformers, it should check whether current match is somewhere within a word or surrounded with space/punctuation/newline. Let me take over it, along with couple of improvements for playground markdown

fantactuka, line 241: It should not blindly skip non-intraword transformers, it should check whether current match is somewhere within a word or surrounded with space/punctuation/newline. Let me take over it, along with couple of improvements for playground markdown

mmarkelov:
@fantactuka thanks a lot!
Close it, will be fixed with https://github.com/facebook/lexical/pull/2191

PR #2173, opened 2022-05-14
delete unuse variable dfsAncestor

sanfengliao:
The variable dfsAncestor in LexicalNode.js looks useless. Although it is used for conditional judgment in the code, the code executed when the condition is true only reassigns dfsAncestor, which has no effect on the overall logic of the function getNodesBetween.


zurfyx (approved):
Hmm, makes sense, makes we wonder what the intent behind this was..

acywatson (approved):
Nice find here, also not sure why this was there in the first place, but logically this seems like the right thing to do.

PR #2174, opened 2022-05-14
[Lexical Playground]: Fixes Excalidraw saved scenes rendering

Fausto95:
Hi there, Excalidraw core team member here 👋🏽

I was just playing around with the playground and realized that the saved scenes weren't being rendered because of the `appState` props null checking, we don't really need the `appState` when rendering the scenes.

Capture d’écran 2022-05-14 à 22 55 30


thegreatercurve (approved):
Thanks for the fix!

PR #2178, opened 2022-05-15
Update licence header in files to use Meta 

thegreatercurve:


acywatson:
Looks like this is the right thing to do - thank you!

acywatson (approved):

PR #2182, opened 2022-05-15
Another small copy fix in some of the file headers

thegreatercurve:


acywatson (approved):
Nice - I guess you're looking at the build script as the source of truth? It looks like it's correct, but I guess there are several alternative grammatical constructions we're allowed to use, since the automated checkup appears to be passing for all of these files.

thegreatercurve:
> Nice - I guess you're looking at the build script as the source of truth? It looks like it's correct, but I guess there are several alternative grammatical constructions we're allowed to use, since the automated checkup appears to be passing for all of these files.

I think the logic in the build script is just to add the header to the WWW file headers on build. We don't really have a check to make sure it exists in all files. Could we worth adding in the future though?

PR #2135, opened 2022-05-11
Migrate more packages to TypeScript (2/n)

thegreatercurve:
Continues work on #1242. This PR includes: 

- Migrated 3 fairly small packages to test initial TS migration:
    -  `lexical-rich-text`
    -  `lexical-plain-text`
    -  `lexical-file`
    -  `lexical-hashtag`
    -  `lexical-headless`
- Fixed several small typing bugs which were picked up by TS.

### Migration steps

I ran the below script to update the directories automatically, and then went through each one fixing the result TypeScript compiler errors, and Lint errors which prevents Vite from building the playground.

```bash
#!/usr/bin/env bash
DIRECTORIES=(
    "lexical-code" 
    "lexical-history" 
    "lexical-dragon" 
)

for directory in ${DIRECTORIES[@]}; do
    # Preserve commit history
    for i in $(find ./packages/$directory -iname "*.js" -not -path "*__tests__*" -not -path "*dist*");
        do git mv "$i" "$(echo $i | rev | cut -d '.' -f 2- | rev).ts";
    done
    for i in $(find ./packages/$directory -iname "*.jsx" -not -path "*__tests__*" -not -path "*dist*");
        do git mv "$i" "$(echo $i | rev | cut -d '.' -f 2- | rev).tsx";
    done

    for i in $(find ./packages/$directory -iname "*.ts" -not -path "*__tests__*" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is FindAndReplace/g" $i && sed -i "" "s/: LexicalNode>/ = LexicalNode>/g" $i;
    done
    for i in $(find ./packages/$directory -iname "*.tsx" -not -path "*__tests__*" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is FindAndReplace/g" $i && sed -i "" "s/: LexicalNode>/ = LexicalNode>/g" $i;
    done
    
    npx @khanacademy/flow-to-ts --write "./packages/$directory/**/{!*.d.ts,*.ts}";
    npx @khanacademy/flow-to-ts --write "./packages/$directory/**/*.tsx";
done

npm run prettier:fix;
```

### Bundle sizes

All looks good with the bundle sizes after running `npm run build-www`. We're still making the most out of Google Closure Compiler: 
|package|main|TS branch|
|---|---|---|
|`LexicalHistory.prod.js`|3.2K|3.2K|
|`LexicalCode.prod.js`|8.3K|8.3K|
[...]

PR #2183, opened 2022-05-16
Migrate `lexical-mark` to TypeScript

maasencioh:
Help on #1242

### DIFF
--- a/packages/lexical-mark/src/MarkNode.ts
+++ b/packages/lexical-mark/src/MarkNode.ts
@@ -170,6 +170,6 @@ export function $createMarkNode(ids: Array<string>): MarkNode {
   return new MarkNode(ids);
 }
 
-export function $isMarkNode(node?: LexicalNode | null): node is MarkNode {
+export function $isMarkNode(node: LexicalNode | null): node is MarkNode {
   return node instanceof MarkNode;
 }
--- a/packages/lexical-mark/src/MarkNode.ts
+++ b/packages/lexical-mark/src/MarkNode.ts
@@ -33,7 +33,7 @@ export class MarkNode extends ElementNode {
     return new MarkNode(Array.from(node.__ids), node.__key);
   }
 
-  constructor(ids: Array<string>, key?: NodeKey): void {
+  constructor(ids: Array<string>, key?: NodeKey) {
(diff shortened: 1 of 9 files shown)


packages/lexical-mark/src/MarkNode.ts
```diff
@@ -168,6 +170,6 @@ export function $createMarkNode(ids: Array<string>): MarkNode {
   return new MarkNode(ids);
 }
 
-export function $isMarkNode(node: ?LexicalNode): boolean %checks {
+export function $isMarkNode(node?: LexicalNode | null): node is MarkNode {
```
thegreatercurve, line 173: We want to make sure we pass something to this method: 
```suggestion
export function $isMarkNode(node: LexicalNode | null): node is MarkNode {
```

thegreatercurve (commented):
LGTM but I would leave out the `shared` packages migration, at least for the moment. 

I had issues trying to get those to work interoperably between Flow and TS in #2135, hence the failing build here as well. It was easier just to create a temporary and separate `shared-ts` package which we use in the TypeScript files. 

My idea is to remove `shared` and rename `shared-ts` back to `shared` once everything has been migrated.

thegreatercurve:
Thanks for helping to work on this @maasencioh! 🙏

maasencioh:
> My idea is to remove shared and rename shared-ts back to shared once everything has been migrated.
Thanks I just saw your PR, so for the moment I'll revert that package

thegreatercurve:
I think you need to update this line in the build script as well to `.ts`: https://github.com/facebook/lexical/blob/main/scripts/build.js#L541

maasencioh:
@thegreatercurve thank you, sorry, a lot of new learning about how to do it, but after, I'm pretty sure to be able to do it with fewer fix commits 😅

thegreatercurve (approved):

thegreatercurve:
> @thegreatercurve thank you, sorry, a lot of new learning about how to do it, but I'm pretty sure to be able to do it with fewer fix commits next time 😅

@maasencioh It's really no problem at all! All of the commits on a PR get squashed on merge any way, so feel free to add as many as you like.

Issue #2123, opened 2022-05-10
can't enter CJK characters properly


arayutw:
When you enter CJK characters from your iOS15 smartphone, the first character will be duplicated.

![IMG_0159](https://user-images.githubusercontent.com/5499895/167518925-b6bd1465-0692-4aa9-b3cb-3d184c65ed86.PNG)
If you enter `おはよう`, `おはおはよう` will be entered.

Can be reproduced here.
https://playground.lexical.dev/

trueadm:
Thanks for reporting this. I'll dig into it when I next get the chance :)

Issue #2168, opened 2022-05-14
Bug: If `placeholder` exists, the first character entered will be selected.


kyoya0819:
Lexical version: 0.2.9

## Steps To Reproduce

I do not have a computer with a non-Japanese keyboard, so I am writing about how to reproduce this on a Japanese keyboard.

1. Set `placeholder` to something other than `undefined`.
2. Selects the input area.
3. Change the character input mode to alphanumeric input mode.
4. Changes the character input mode to Kana (Japanese) input mode.
5. Type some characters.

~~This bug can also be seen in [Lexical's Playground](https://lexical-playground.vercel.app/).~~
This bug can also be seen in [the sample embedded on the home page](https://lexical.dev/).

## The current behavior

https://user-images.githubusercontent.com/45027336/168419957-56384162-746a-4261-acd5-91edb65855c6.mov

## The expected behavior

If I set `placeholder` to `undefined`, I will not get this error.
However, I'll get a type definition error.

https://user-images.githubusercontent.com/45027336/168419862-6fd80e0c-6088-4be1-8fd5-9212857b845b.mov

I am not very good at English, so if I am wrong, I would appreciate it if you could let me know.
Thanks for reading! Thanks for the great library!

trueadm:
What browser and operating system are you using?

kyoya0819:
I forgot to post the information. My apologies.

As far as I could confirm, only the combination of macOS and Chrome could confirm this phenomenon.

Here are the respective versions.
macOS: 12.3.1
Chrome: 101.0.4951.64

Also, let me make a correction regarding the first Issue.
The same behavior was observed on the Lexical site, not on the "Playground" but on the sample embedded on the home page.

trueadm:
Hmm, I wonder if this is related to plain text mode. On the playground website, if you click in the bottom left, on the toggle and switch to Plain text mode. Does this issue occur then too?

trueadm:
I'm finding it really hard to reproduce this issue. Every-time I enter a character on either playground, or on the website example (which is a plain text example), it works fine. I'm also on macOS + Chrome.

kyoya0819:
> Hmm, I wonder if this is related to plain text mode. On the playground website, if you click in the bottom left, on the toggle and toggle off "Rich text" mode. Does this issue occur then too?

Both with Rich mode off and on, Playground was unable to reproduce the behavior of the video in this issue.

[...]

PR #2185, opened 2022-05-16
Fix Safari IME issues

trueadm:
This should fix https://github.com/facebook/lexical/issues/2123 and possibly https://github.com/facebook/lexical/issues/2168. It also fixes some other composition issues that might affect Chrome + FF too.

### DIFF
--- a/packages/lexical/src/LexicalEvents.js
+++ b/packages/lexical/src/LexicalEvents.js
@@ -579,7 +579,7 @@ function onCompositionStart(
         !selection.isCollapsed() ||
         selection.anchor.getNode().getFormat() !== selection.format
       ) {
-        // We insert an zero width character, ready for the composition
+        // We insert a zero width character, ready for the composition
         // to get inserted into the new node we create. If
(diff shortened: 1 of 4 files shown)


packages/lexical/src/LexicalEvents.js
```diff
@@ -561,19 +570,20 @@ function onCompositionStart(
...
         event.timeStamp < lastKeyDownTimeStamp + ANDROID_COMPOSITION_LATENCY ||
         // FF has issues around composing multibyte characters, so we also
         // need to invoke the empty space heuristic below.
-        (IS_FIREFOX && anchor.type === 'element') ||
+        anchor.type === 'element' ||
         !selection.isCollapsed() ||
         selection.anchor.getNode().getFormat() !== selection.format
       ) {
-        // We insert an empty space, ready for the composition
+        // We insert an zero width character, ready for the composition
```
thegreatercurve, line 582:
```suggestion
        // We insert a zero width character, ready for the composition
```

thegreatercurve (approved):

Issue #2049, opened 2022-05-02
Bug: Formatting of Inline code elements is lost when content is copied from external web page


reads21:
Formatting of Inline code elements is lost when content is copied from external web page. In the dom we  see  inline "<code> code</code>" element  missing when pasted in lexical .

Lexical version: 0.25

## Steps To Reproduce

1. navigate to Mozilla mdn website https://developer.mozilla.org/en-US/docs/Web/API/Node/normalize and copy the first paragraph like the one's pasted below. 

 "The normalize() method of the [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node) interface puts the specified node and all of its sub-tree into a normalized form. In a normalized sub-tree, no text nodes in the sub-tree are empty and there are no adjacent text nodes."
 
2.  Paste it into lexical playground to see the "normalize()" text unformatted . 

 
Link to code example:
 
https://playground.lexical.dev/

## The current behavior
Inline code element's formmating is missing when content is from external web page.

## The expected behavior
Inline code elements formatting should be preserved when pasted into lexical.

mmarkelov:
I guess that the `format` keep only one value, so only one style will be applied to pasted text. I can try to take it, and make `format` an array or set maybe

fantactuka:
Fixed by #2607

PR #2187, opened 2022-05-16
Add code to html conversions

mmarkelov:
close https://github.com/facebook/lexical/issues/2049

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
@@ -296,6 +296,51 @@ test.describe('CopyAndPaste', () => {
     });
   });
 
+  test(`Copy and paste code`, async ({isPlainText, page, browserName}) => {
+    test.skip(isPlainText);
+
+    await focusEditor(page);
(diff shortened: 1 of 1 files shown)


tylerjbainbridge:
In `importDOM` for `code` you should check if it's an inline code element and not a code block. If it's a code block, you can return null, otherwise convert to text node.


```js
...
code: (node: Node) => {
  if (node is code block) {
    return null;
  }

  return {
    conversion: convertTextFormatElement,
    priority: 0,
  }
},
...
```

acywatson:
> Won't this match every `<code>` tag with the TextNode and not the CodeNode?

It will, but this still might be the right thing to do. CodeNode is not a core node, so this would allow us to handle pasted HTML code tags even if CodeNode is not registered.

I think the full solution involves also adding a higher priority conversion on CodeNode/CodeHighlightNode, but I think it could be done in two PRs.

Either way, we need some test coverage for this, which can be added here:

https://github.com/facebook/lexical/blob/main/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs

trueadm:
@tylerjbainbridge This adds tests, is this something we can get in for increased coverage?

Issue #1988, opened 2022-04-26
Copy-pasting new lines, lists and headings from Mac notes


zurfyx:
Currently creating additional lines, bullet points and headings are not marked as such

https://user-images.githubusercontent.com/193447/165287489-6a37f70c-9ceb-4d05-acbe-1f975d1a086d.mov

trueadm:
This has been greatly improved since.

PR #2188, opened 2022-05-16
Remove spaces from clipboard html

mmarkelov:
Some investigation on https://github.com/facebook/lexical/issues/1988

This is because copied html contains line breakers, which lexical convert to TextNode. I guess that `htmlString` should be inline, without any breakers


fantactuka:
Full clean up of \n's might be a bit too much (e.g. might want to keep it within `<pre>`). Also not sure about ` {2}` part - if trying to remove duplicated spaces you'd still want to keep at least one: `hello _ _ _ world` -> `hello _ world`.

@acywatson @tylerjbainbridge WDYT about this one, should tabs/empty lines be cleaned up upfront on on node level

tylerjbainbridge:
> Full clean up of \n's might be a bit too much (e.g. might want to keep it within `\<pre\>`). Also not sure about ` {2}` part - if trying to remove duplicated spaces you'd still want to keep at least one: `hello _ _ _ world` -> `hello _ world`.
> 
> @acywatson @tylerjbainbridge WDYT about this one, should tabs/empty lines be cleaned up upfront on on node level

Definitely agree re: duplicated spaces, I personally don't like the idea of additional spaces being lost upon saving because they're most likely there for a reason.

And regarding `/n`- we're just using the browser's native `innerHtml` serialization to build the string and like Maksim said there may be some cases where the `/n`s are important like within `<pre>` tags or text.

I think this should be handled in the `$generateNodesFromDOM(...)` function and just avoid creating unnecessary text nodes for line breaks.

mmarkelov:
@tylerjbainbridge @fantactuka I tried to put this logic into `$generateNodesFromDOM(...)`, but this is a challenge for me. I understood that removing the whole spaces was not a good idea. But what if we remove new lines only between tags, i guess it should work. What do you think guys?
https://regex101.com/r/MXqZMS/1

fantactuka:
> @tylerjbainbridge @fantactuka I tried to put this logic into `$generateNodesFromDOM(...)`, but this is a challenge for me. I understood that removing the whole spaces was not a good idea. But what if we remove new lines only between tags, i guess it should work. What do you think guys? https://regex101.com/r/MXqZMS/1

It's a better approach, but it yet gonna fall short for `<pre>` tag :( Looked a bit more into in and Tyler's suggestion around `$generateNodesFromDOM` makes sense, or even more specific `$createNodesFromDOM` where DOM node transformed into lexical node. E.g. you can check if current element tag is PRE and keep text nodes with `\n` content, or skip them otherwise. Do you want to give another try or I can take over it?

mmarkelov:
@fantactuka thanks for your report. IDK, but shouldn't it be the part of `LexicalTextNode`?
Maybe here:
```js
function convertTextDOMNode(domNode: Node): DOMConversionOutput {
  return {node: domNode.textContent.match(/\n+/g) ? null : $createTextNode(domNode.textContent)};
}
```

trueadm:
Can we confirm we still need this? I noticed lately this has already improved?

fantactuka:
#2431 fixing this issue newlines, while preserving it within `<pre>`

PR #2189, opened 2022-05-16
Fix editor-state.md

trueadm:
We should be recommending people to use `JSON.stringify(editorState)` directly.


thegreatercurve:
Why shouldn't we recommend using `.toJSON`? I thought it has some specialized formatting for the different selection types: https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalEditorState.js#L93-L134

thegreatercurve (approved):

trueadm:
@thegreatercurve `toJSON` is an implementation detail that `JSON.stringify` looks for on objects (as part of the JS spec).

PR #2191, opened 2022-05-16
Add intraword format support for Markdown import

### LINKED ISSUE #2159: Bug: Image URLs that contain markdown formatting don't get parsed correctly when converting to/from markdown

fantactuka:
- Similar to #2136 which added intraword into markdown shortcuts (`_` should not be converted into italic if not surrounded with space/punctuation/newline), handling it during markdown import
- Couple fixes for playground transformers (table & tweet)

Fix #2159

### DIFF
--- a/packages/lexical-playground/src/plugins/MarkdownTransformers.ts
+++ b/packages/lexical-playground/src/plugins/MarkdownTransformers.ts
@@ -31,6 +31,7 @@ import {
   $isTableRowNode,
   TableCellHeaderStates,
   TableCellNode,
+  TableNode,
 } from '@lexical/table';
 import {
   $createParagraphNode,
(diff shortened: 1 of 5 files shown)


packages/lexical-playground/src/plugins/MarkdownTransformers.ts
```diff
@@ -203,12 +202,27 @@ export const TABLE: ElementTransformer = {
       }
     }
 
-    parentNode.replace(table);
+    const previousSibling = parentNode.getPreviousSibling();
```
acywatson, line 205: This seems right, but wondering if we need to test it?

trueadm (approved):

packages/lexical-playground/src/plugins/MarkdownTransformers.ts
fantactuka: Added couple e2e for tables, which exposed issue with focus being triggered during reconciliation (#2209), I'll merge this one to clean up PR queue and add e2e as part of that PR

Issue #2195, opened 2022-05-17
Bug: Custom Decorator nodes not rendered after editor state is restored from local storage


vineetdigit:
When the editor state containing custom decorator nodes is stored in the local storage, and later retrieved, parsed and passed down to the RichTextPlugin, then the custom decorator nodes are **not** rendered. However, paragraph/text nodes are rendered just fine. The expected behavior is that all nodes are rendered. This problem is at least reproducible in the codesandbox below.

Lexical version: 0.2.9

## Steps To Reproduce

I have created a custom decorator node named Excalidraw which currently renders just a div. 

The codesandbox has instructions on the app screen to reproduce the issue.

https://codesandbox.io/s/frosty-aryabhata-e4pkke-e4pkke

## The current behavior
Only paragraph/text nodes are rendered.

## The expected behavior
paragraph/text, as well as custom decorator node, must render

trueadm:
I'm hoping that https://github.com/facebook/lexical/pull/2200 fixes this issue due to lifecycle sequencing.

PR #2200, opened 2022-05-17
Improve useDecorators sequencing

trueadm:
Hopefully fixes https://github.com/facebook/lexical/issues/2195.

### DIFF
--- a/packages/lexical-react/src/shared/useDecorators.js
+++ b/packages/lexical-react/src/shared/useDecorators.js
@@ -34,7 +34,7 @@ export function useDecorators(editor: LexicalEditor): Array<React.Node> {
     // ensuring that we set the value.
     setDecorators(editor.getDecorators());
   }, [editor]);
-  
+
   // Return decorators defined as React Portals
   return useMemo(() => {
     const decoratedPortals = [];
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/shared/useDecorators.js
```diff
@@ -27,6 +27,14 @@ export function useDecorators(editor: LexicalEditor): Array<React.Node> {
       });
     });
   }, [editor]);
+
+  useEffect(() => {
+    // If the content editable mounts before the subscription is added, then
+    // nothing will be rendered on initial pass. We can get around that by
+    // ensuring that we set the value.
+    setDecorators(editor.getDecorators());
+  }, [editor]);
```
fantactuka, line 36: Should decorator listener registering move into useEffect too? Would be kind of natural flow: on after render set current decorators and subscribe to future changes, or layout effect still gives any advantages?

trueadm: No need, as we listen to the changes above. It would just be additional overhead to do that again.

fantactuka (approved):

tylerjbainbridge (approved):

Issue #2199, opened 2022-05-17
Bug: incorrect caret position when using mac's built-in text expander


shaoshing:
When using mac's built-in text expander feature to replace text, Lexical inserts the wrong caret position after text replacement.

Lexical version: 

## Steps To Reproduce

1. In keyboard preference (System Preferences > Keyboard > Text), add some text replacements. e.g. "omw" > "On my way!" 
2. Go to playground (https://playground.lexical.dev/)
3. Type "omw,"
4. Observe the text get changed to "On my way!," with the text caret position changed to in front of the comma.

Lexical playground
https://user-images.githubusercontent.com/118264/168899380-2c31307c-fe61-4da4-9584-e6ecd9e88bff.mp4

For comparison, here's the behavior in a regular input
https://user-images.githubusercontent.com/118264/168899643-48013141-b94f-4e3e-8159-6c2d2e2c7d9e.mp4

## The current behavior

The caret position should be after the comma.

trueadm:
What's the behavior in other contenteditables?

shaoshing:
Both Draft and Quill have the expected behavior by inserting the caret at the end

https://user-images.githubusercontent.com/118264/168902759-36034aec-2e44-4539-87f0-44661593d657.mp4

https://user-images.githubusercontent.com/118264/168902766-e745125f-280a-4e6c-bcf3-0c4631b75af1.mp4

trueadm:
I'll look tomorrow :)

PR #2203, opened 2022-05-18
Fix text replacement event handling

trueadm:
Fixes https://github.com/facebook/lexical/issues/2199. When handling text replacements, we need to handle the logic at a later stage than beforeinput. That's because we don't know what the selection should be after the replacement.

### DIFF
--- a/packages/lexical/src/LexicalEvents.js
+++ b/packages/lexical/src/LexicalEvents.js
@@ -77,6 +77,7 @@ import {
   $updateSelectedTextFromDOM,
   $updateTextNodeFromDOMContent,
   dispatchCommand,
+  doesContainGrapheme,
   getDOMTextNode,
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalEvents.js
```diff
@@ -530,10 +531,17 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
   updateEditor(editor, () => {
     const selection = $getSelection();
     const data = event.data;
+    const possibleTextReplacement =
```
fantactuka, line 534: Is it applicable for non-mac/ios?

trueadm: I believe so!

thegreatercurve (approved):

PR #2205, opened 2022-05-18
Migrate EditorState doc to typescript

always-maap:
Since the lexical team is migrating to typescript I thought it would be appropriate to convert examples to typescript. Is it something you are interested in? I can work on it. either JS/TS or all in ts


thegreatercurve:
Thanks for the contribution @always-maap! Tbh, I think we should keep the docs as vanilla JS only for the moment to lower the barrier to entry for non-TS users. Once we've done (alot) more work on our docs, we could expand our examples to also include TS examples.

PR #2206, opened 2022-05-18
Upgrade build.js to use rollup plugin

trueadm:
This switches from our custom GCC plugin to using the Rollup GCC plugin from the AMP project.


tylerjbainbridge (approved):

thegreatercurve (approved):

scripts/build.js
```diff
@@ -70,10 +69,6 @@ const wwwMappings = {
   'react-dom': 'ReactDOMComet',
 };
```
acywatson, line 71: Why don’t we need this anymore?

scripts/build.js
```diff
@@ -270,7 +265,7 @@ ${source}`;
     exports: 'auto',
     externalLiveBindings: false,
     file: outputFile,
-    format: 'cjs',
+    format: 'cjs', // change between es and cjs modules
```
acywatson, line 268: Is the idea to manually change this when we want produce esm?

trueadm: The idea is to probably build both in the future. @thegreatercurve was interested in doing that at some point, as it would require us to add some more fields to all the package.jsons etc.

PR #2209, opened 2022-05-18
Focus into editor after reconciliation is done

fantactuka:


trueadm (approved):

trueadm:
Very nice!

trueadm:
Couldn't `triggerMutationListeners` be moved afterall?

fantactuka:
> Couldn't `triggerMutationListeners` be moved afterall?

Removed that for now, as not yet finished with focus command adjustments:

Moving `rootElement.focus()` to run somewhere after reconciliation does not solve the issue, since `domSelection.setBaseAndExtent` that is called within `reconcileSelection` also triggering focus event. Moving `setBaseAndExtent` outside of reconciler (to be called after it similar to .focus()) does not sound right

Thought to add a flag to defer focus command handlers to run after reconciler, but then it will break usage of .preventDefault (in case it's used)

trueadm:
> Thought to add a flag to defer focus command handlers to run after reconciler, but then it will break usage of .preventDefault (in case it's used)

I was thinking about doing this too, with a capture listener on the `window`. You can't preventDefault on focus/blur anyway, it doesn't do anything. You can just re-dispatch the event on the content editable after reconciliation.

fantactuka:
> However, maybe we're missing a trick here? Should we just reconcileSelection after we reconcile the root, and after we do the other work?

Not sure I follow here, as selection is already reconciled right after root (only mutation observer is restarted in between). Did you mean moving focus/setBaseAndExtent to run after reconciler or was it something else?

trueadm:
> Not sure I follow here, as selection is already reconciled right after root (only mutation observer is restarted in between). Did you mean moving focus/setBaseAndExtent to run after reconciler or was it something else?

After we clear the maps/sets/freeze everything. Right now we do `updateEditorState`, but maybe that should just be a direct call to `reconcileRoot` and below, we do `reconcileSelection` and remove `updateEditorState` entirely.

fantactuka:
> After we clear the maps/sets/freeze everything. Right now we do `updateEditorState`, but maybe that should just be a direct call to `reconcileRoot` and below, we do `reconcileSelection` and remove `updateEditorState` entirely.

Yea, it might make code easier to adjust. I wonder how much selection.setBaseAndExtent is a part of reconciler, because  it could be something like this:
```js
function commitPendingUpdates() {
  ...
  try {
    reconcileRoot();
    reconcileSelection();
  } catch(e) {
    recoverIfFailed()
  } finally {
    syncSelectionWithDomAndFocusIfNeeded() // <- deferred .focus(), .setBaseAndExtent() and scrollIntoView() calls
    ...
  }
  ...
}



```

acywatson:
Sorry to jump in here randomly, but what problem is this solving?

fantactuka:
> Sorry to jump in here randomly, but what problem is this solving?

Preventing .focus from being called while reconciling selection. It dispatches synchronous FOCUS_COMMAND which means code will be executed in the middle of reconciliation. While we want reconciliation to be isolated from any outside changes

trueadm:
@fantactuka You don't want to put it in finally, as that will always trigger even in the exception case?

fantactuka:
> @fantactuka You don't want to put it in finally, as that will always trigger even in the exception case?

Agh, right, recovering from error will run its own commitPendingUpdates and will do own selection sync

PR #2210, opened 2022-05-18
Support Code Theme Switch Manually

WinmezzZ:
In the playground, the existing code theme was flat, even missing some token styles

For most of people, they had accustomed different code theme, I think we can provide a way to let them switch between their favorite code theme

Like this:

https://user-images.githubusercontent.com/40154766/169103510-601b44b6-7f7e-48ea-98a3-5a01ce750221.mov

All of themes from [prism-themes](https://github.com/PrismJS/prism-themes), these just part of it, so people can add more of theme into theirs project.

**Note**: 
- Removed the original code token style
- Fix missing token: Add All of [prism token](https://prismjs.com/tokens.html) className
- For performance optimization, all theme style files are dynamically imported

I think this tutorial is very helpful and meaningful


packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -97,13 +97,10 @@
...
 }
 .PlaygroundEditorTheme__code {
-  background-color: rgb(240, 242, 245);
-  font-family: Menlo, Consolas, Monaco, monospace;
   display: block;
-  padding: 8px 8px 8px 52px;
-  line-height: 1.53;
-  font-size: 13px;
-  margin: 0;
+  padding: 8px 8px 8px 52px !important;
```
acywatson, line 101: What was the problem here? Could we not solve it with specificity?

tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "target": "ES2019",
     "lib": ["esnext", "dom", "dom.iterable"],
     "jsx": "react-jsx",
+    "module": "ESNext",
```
acywatson, line 6: What's this for? Enabling dynamic imports?

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
WinmezzZ: In others prism theme, they have diffirent line-height or font-size, they have a higher CSS priority than these

So I add `!important` for these

tsconfig.json
WinmezzZ: Right! otherwise, a TS error will occur

> Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node12', or 'nodenext'.

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
WinmezzZ: I think this should be irrelevant since this is just playground example code and people can also use all the styles of Prism Theme themselves

packages/lexical-playground/src/themes/PlaygroundEditorTheme.ts
```diff
@@ -14,36 +14,38 @@ const theme: EditorThemeClasses = {
   characterLimit: 'PlaygroundEditorTheme__characterLimit',
   code: 'PlaygroundEditorTheme__code',
   codeHighlight: {
-    atrule: 'PlaygroundEditorTheme__tokenAttr',
```
fantactuka, line 17: e2e tests are failing since class names in tokens has changes, so it's either keep names as it is now (`tokenAttr` instead of `token_attr-name` for all themes) or need to update e2e tests to accommodate new class names

WinmezzZ: Changed e2e tests className but also failed, I think i need some help for these because i am not good at writting unit test.

WinmezzZ:
If no one can help me solve the tests problem, I will close this pull request

acywatson:
> If no one can help me solve the tests problem, I will close this pull request

Hey @WinmezzZ - we can definitely help! What's the issue? It looks like you just need to change the classNames in the tests:

```
   -   <span class="PlaygroundEditorTheme__token_entity" data-lexical-text="true">
   +   <span class="PlaygroundEditorTheme__token_operator" data-lexical-text="true">
```

This [result in the tests](https://github.com/facebook/lexical/runs/6529964438?check_suite_focus=true#step:8:306), for example, means that the test was expecting:

```
class="PlaygroundEditorTheme__token_entity"
```
but the actual HTML has this:

```
 class="PlaygroundEditorTheme__token_operator"
```

These two need to match for the test to pass. You need to do one of two things:

1) If the code is functioning as intended, you should update the HTML in the test assertion to the correct className

2) Otherwise, if the code is not functioning as intended and there should be a different className here, then you need to fix the error in the code.

It sounds like we're in case 1.

WinmezzZ:
I don't have much experience with unit testing, I think I need you guys to help me with PR

I would be very grateful to that


https://github.com/WinmezzZ/lexical

WinmezzZ:
Turn off the PR because it's not active

PR #2215, opened 2022-05-18
Migrate 5 more packages to TypeScript (8/n)

thegreatercurve:

- Migrated 1 package::
    -  `lexical-list`
    -  `lexical-markdown`
    -  `lexical-offset`
    -  `lexical-overflow`
    -  `lexical-link`

### Migration steps

I ran the below script to update the directories automatically, and then fixed each compiler linter error manually:

```bash
#!/usr/bin/env bash
DIRECTORIES=(
  "lexical-list"
  "lexical-markdown"
  "lexical-offset"
  "lexical-overflow"
  "lexical-link"
)

    # Preserve commit history
    for i in $(find ./packages/$directory -iname "*.js" -not -path "*dist*");
    done
    for i in $(find ./packages/$directory -iname "*.jsx" -not -path "*dist*");
    done

    for i in $(find ./packages/$directory -iname "*.ts" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is ManuallyFindAndReplace/g" $i && sed -i "" "s/: LexicalNode>/ = LexicalNode>/g" $i;
    done
    for i in $(find ./packages/$directory -iname "*.tsx" -not -path "*dist*" -not -path "flow");
        do sed -i "" "s/boolean %checks/node is ManuallyFindAndReplace/g" $i && sed -i "" "s/: LexicalNode>/ = LexicalNode>/g" $i;
    done
    
done

npm run prettier:fix;
```

### Bundle sizes

Bundle sizes are consistent with `main`. Below are some sample packages: 
|package|main|TS branch|
|---|---|---|
|`LexicalList.prod.js`|11K|11K|
|`LexicalOveflow.prod.js`|879B|879B|
|`LexicalOffset.prod.js`|3.6k|3.6K|


packages/lexical-link/src/index.ts
```diff
@@ -19,16 +19,18 @@ import type {
 } from 'lexical';
 
 import {addClassNamesToElement} from '@lexical/utils';
+import {Spread} from 'globals';
```
acywatson, line 22: I thought we migrated this to some other path in the serialization PR?

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -63,21 +64,23 @@ export class ListItemNode extends ElementNode {
...
 
-  constructor(value?: number, checked?: boolean, key?: NodeKey): void {
+  constructor(value?: number, checked?: boolean, key?: NodeKey) {
     super(key);
     this.__value = value === undefined ? 1 : value;
     this.__checked = checked;
   }
 
-  createDOM(config: EditorConfig): HTMLElement {
+  createDOM(config: EditorConfig, _editor?: LexicalEditor): HTMLElement {
```
acywatson, line 73: ?

packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
```diff
@@ -75,32 +76,46 @@ describe('LexicalListNode tests', () => {
 
     test('ListNode.createDOM()', async () => {
       const {editor} = testEnv;
+
       await editor.update(() => {
-        const listNode = $createListNode('ul', 1);
+        const listNode = $createListNode('bullet', 1);
```
acywatson, line 81: Why change this tag name?

packages/lexical/flow/Lexical.js.flow
```diff
@@ -346,17 +346,17 @@ declare export class LexicalNode {
   isSelected(): boolean;
   getKey(): NodeKey;
   getIndexWithinParent(): number;
-  getParent(): ElementNode | null;
-  getParentOrThrow(): ElementNode;
-  getTopLevelElement(): null | ElementNode;
-  getTopLevelElementOrThrow(): ElementNode;
-  getParents(): Array<ElementNode>;
+  getParent<T: ElementNode>(): T | null;
```
acywatson, line 349: Won't this require us to pass in type args at every callsite?

acywatson: Or it works because you provide a default type arg on the base class?

packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
thegreatercurve: This is the [actual type](https://github.com/facebook/lexical/blob/main/packages/lexical-list/LexicalList.d.ts#L21), but we exclude the tests from the Flow config, so it was never picked up.

packages/lexical-list/src/LexicalListItemNode.ts
thegreatercurve: Hmm not sure why this is here. Removed.

packages/lexical/flow/Lexical.js.flow
thegreatercurve: It works as a default value for `ElementNode`, hence why you don't always have to pass the value, but it also ensures that any type that is explicitly defined (i.e. `getParent<RootNode>()` conforms the `ElementNode`, so it has to be an ElementNode or extend ElementNode.

packages/lexical-link/src/index.ts
thegreatercurve: Yeah, I changed it back in the serialization PR as it caused some unit tests to break. Globals works by default as well.

acywatson (approved):
I'm good with this - thanks!

PR #2220, opened 2022-05-19
Support Strikethrough and italic paste from Google Docs

### LINKED ISSUE #2216: Bug: Paste strikethrough and italic from Google Docs
Some formatting (strikethrough and italic) are lost when pasting from Google docs. This is likely because Google docs doesn't use semantic HTML and instead uses inline styles. In the past, I've needed to detect when a paste is coming from Google docs and then look at inline styles. I've done this by seeing if the paste payload includes `docs-internal-guid` and then looking at nodes `font-style` (for italic) and `text-decoration` for strikethrough.

Lexical version: 0.2.9 (Latest at time of issue creation)

## Steps To Reproduce

1. Go here: https://docs.google.com/document/d/1qLjUKNWU-VK_TQsUY18qqyA_gtIpJOsNAMFh4et21Fw/edit?usp=sharing
2. Copy contents
3. Paste into sandbox

Link to code example:
https://codesandbox.io/s/lexical-rich-text-example-forked-mxx7z8?file=/src/Editor.js

## The current behavior
Formatting is lost

## The expected behavior
Formatting should be retained

hanford:
closes #2216

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.js
+++ b/packages/lexical/src/nodes/LexicalTextNode.js
@@ -771,7 +771,7 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
   const hasBoldFontWeight = span.style.fontWeight === '700';
   // Google Docs uses span tags + text-decoration for strikethrough text
   const hasLinethroughTextDecoration = span.style.textDecoration === 'line-through';
-  // Google Docs uses span tags + text-decoration for italic text
+  // Google Docs uses span tags + font-style for italic text
   const hasItalicFontStyle = span.style.fontStyle === 'italic';
   
   return {
--- a/packages/lexical/src/nodes/LexicalTextNode.js
+++ b/packages/lexical/src/nodes/LexicalTextNode.js
@@ -768,23 +768,23 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
   // $FlowFixMe[incompatible-type] domNode is a <span> since we matched it by nodeName
   const span: HTMLSpanElement = domNode;
   // Google Docs uses span tags + font-weight for bold text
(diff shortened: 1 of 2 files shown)


packages/lexical/src/nodes/LexicalTextNode.js
```diff
@@ -768,11 +768,23 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
   // $FlowFixMe[incompatible-type] domNode is a <span> since we matched it by nodeName
   const span: HTMLSpanElement = domNode;
   // Google Docs uses span tags + font-weight for bold text
-  const hasBoldFontWeight = span.style.fontWeight === '700';
+  const hasBoldFontWeight = span.style.fontWeight === "700";
+  // Google Docs uses span tags + text-decoration for strikethrough text
+  const hasLinethroughTextDecoration =
+    span.style.textDecoration === "line-through";
```
hanford, line 774: Prettier wants this to be on two lines ¯\_(ツ)_/¯

thegreatercurve (approved):
Thanks @hanford! LGTM. Thanks for the detailed issue report as well.

thegreatercurve:
Run `npm run prettier:fix` to fix the CI error.

acywatson:
Awesome! Thanks for fixing this.

Can we add a quick test or two here?:

https://github.com/facebook/lexical/blob/71824d1be3d8595b791285f1c6a1ab55d8a9250d/packages/lexical-utils/src/__tests__/unit/LexicalEventHelpers.test.js#L318

As a separate issue, these unit tests are probably in the wrong place, as whole. We need to put them somewhere more discoverable for verifying HTML copy+paste.

hanford (commented):
Added some tests!

acywatson (approved):

PR #2221, opened 2022-05-19
Fixing Placeholder Bug found on web

bmr5:
Updating useLexicalCanShowPlaceholder to run the placeholder checking logic before the useLayoutEffect to solve for a very specific race condition.


zurfyx (approved):
LGTM! Thanks ps: you may have to run Prettier

PR #2222, opened 2022-05-19
Fix placeholder bug

bmr5:
Fixing Race conditions in the useLayoutEffect


zurfyx (approved):
LGTM!

packages/lexical-react/src/shared/useCanShowPlaceholder.js
```diff
@@ -21,6 +21,11 @@ export function useCanShowPlaceholder(editor: LexicalEditor): boolean {
   );
 
   useLayoutEffect(() => {
+    const isComposing = editor.isComposing();
+    const currentCanShowPlaceholder = editorState.getEditorState().read(
```
bmr5, line 25: missed this - will update

PR #2227, opened 2022-05-20
Deprecated module types bug fix for WWW build

thegreatercurve:
We should be rewriting the imports for the deprecated file types in WWW. I had to manually adjust them in the last sync.


acywatson (approved):
should be right - thanks!

PR #2229, opened 2022-05-20
Fix window undefined issue when SSR

thegreatercurve:
#2203 is causing an internal Meta error that `window.getSelection()` is undefined when server-side rendered. 

I think we should be fine to bypass this check completely as we don't need it until the editor is loaded and interactable

### DIFF
--- a/packages/lexical/src/LexicalEvents.js
+++ b/packages/lexical/src/LexicalEvents.js
@@ -543,6 +543,7 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
     const selection = $getSelection();
     const data = event.data;
     const possibleTextReplacement =
+      CAN_USE_DOM &&
       event.inputType === 'insertText' &&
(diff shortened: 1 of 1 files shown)


trueadm (approved):

packages/lexical/src/LexicalEvents.js
```diff
@@ -561,7 +562,7 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
         isFirefoxEndingComposition = false;
       }
       dispatchCommand(editor, INSERT_TEXT_COMMAND, data);
-      if (possibleTextReplacement) {
+      if (possibleTextReplacement && CAN_USE_DOM) {
```
trueadm, line 565: It might be better to put the CAN_USE_DOM as part of the `possibleTextReplacement` assignment. As `possibleTextReplacement` is used twice.

PR #2236, opened 2022-05-22
Revert "Fix window undefined issue when SSR"

trueadm:
Reverts facebook/lexical#2229

Come to think of it, I'm not sure this is the right fix. @thegreatercurve this code path is on a DOM event listener, which means there must be a window present for it to fire to begin with?


thegreatercurve:
@trueadm It's weird. We had a load of SSR errors last week, but they've since flattened out since the original PR landed. I guess maybe it's similar to a stale caching issue.

thegreatercurve (approved):

PR #2238, opened 2022-05-23
Update CharacterStylesPopupPlugin.tsx

hanford:
Remove unused ref


packages/lexical-playground/src/plugins/CharacterStylesPopupPlugin.tsx
```diff
@@ -74,7 +74,6 @@ function FloatingCharacterStylesEditor({
   isUnderline: boolean;
 }): JSX.Element {
   const popupCharStylesEditorRef = useRef<HTMLDivElement | null>(null);
-  const mouseDownRef = useRef(false);
```
hanford, line 77: This ref is never set

trueadm (approved):
Thank you!

PR #2239, opened 2022-05-23
move toggleLink to lexical/link

haikyuu:
`toggleLink` isn't a React specific function. As such I moved it to lexical/link so that it can be used by other frameworks


packages/lexical-link/LexicalLink.d.ts
```diff
@@ -54,3 +54,4 @@ export function $isAutoLinkNode(
 ): node is AutoLinkNode;
 
 export var TOGGLE_LINK_COMMAND: LexicalCommand<string | null>;
+export function toggleLink(url: null | string): void;
```
acywatson, line 57: I think we need to add this to Flow types as well. Otherwise, looks good!

haikyuu: @acywatson Done 👍

PR #2241, opened 2022-05-23
Remove unstable JSON serialization functions + unify copy+paste to be HTML

trueadm:
> Note: This PR is a major breaking change, so we should version the next version 0.3 (I believe we planned to do this anway).

This PR does the following:

- removes the `unstable_` prefixed JSON import/export serialization functions, instead makes their functionality use the original `toJSON` and `parseEditorState` methods. The new editor state does NOT include selection.
- Removes adding `x-lexical-editor-state` from clipboard, instead we only put on HTML and parse HTML accordingly.
- Fixes a bunch of issues with HTML, including rogue empty line breaks + missing import/export HTML definitions.
- Removes `namespace` from config, as no longer needed.
- Updates Flow/TS types accordingly.

This PR should remove quite a lot of logic around parsing logic from the core.


packages/lexical/src/LexicalUpdates.js
```diff
@@ -247,44 +241,6 @@ function $applyAllTransforms(
   editor._dirtyElements = dirtyElements;
 }
 
-// TODO: once unstable_parseEditorState is stable, swap that for this.
-export function parseEditorState(
-  parsedEditorState: ParsedEditorState,
-  editor: LexicalEditor,
-): EditorState {
```
zurfyx, line 254: ⚠️ Currently used in WWW (PoC @acywatson)

zurfyx (approved):
So ultimately we're saying that both HTML and JSON have the same fidelity. The only difference with JSON is that is can handle versions in a more structured manner..

acywatson:
> So ultimately we're saying that both HTML and JSON have the same fidelity. The only difference with JSON is that is can handle versions in a more structured manner..

I’m not sure that it’s true that HTML has the same fidelity as JSON. Have we found a way to represent nested editors in HTML?  The problem with less than perfect fidelity is that copying and pasting within the same editor (like moving sections of a document around) can result in data loss. I think the tolerance for this is higher between editors, but I would generally expect copying within the same editor to reproduce the pasted content almost perfectly.

trueadm:
> I'm hesitant to accept this PR because without x-lexical-editor-state there's no way to copy + paste decorator nodes around in the same document unless I'm missing something.

I added logic to copy some of the decorators to this PR already. Check out `ImageNode` it also handles the nested caption too. :)

tylerjbainbridge:
> > I'm hesitant to accept this PR because without x-lexical-editor-state there's no way to copy + paste decorator nodes around in the same document unless I'm missing something.
> 
> I added logic to copy some of the decorators to this PR already. Check out `ImageNode` it also handles the nested caption too. :)

Decorator Nodes that have semantic HTML equivalents (ImageNode, HorizontalRuleNode, etc) are easy enough to import/export to HTML but `YoutubeEmbedNode`, `PollNode`, `ExcalidrawNode`, etc, are more complicated to import and export losslessly until we have a robust Decorator<->HTML solution.

tylerjbainbridge:
> I did ExcalidrawNode and EquationNode. Need to do the others, but I used an attribute as signal.

Ahh, I see that now. I guess I was imagining something added to the core serialization logic to support Decorator Nodes by default so people don't have to add identifying data-attributes and serialized node properties themselves, but this is probably fine for now.

packages/lexical-link/src/index.ts
```diff
@@ -131,7 +131,7 @@ export class LinkNode extends ElementNode {
 function convertAnchorElement(domNode: Node): DOMConversionOutput {
   let node = null;
   if (domNode instanceof HTMLAnchorElement) {
-    node = $createLinkNode(domNode.href);
+    node = $createLinkNode(domNode.getAttribute('href'));
```
acywatson, line 134: wondering what the difference is between this and a direct property access?

trueadm: Misses the // from http://

packages/lexical-playground/src/nodes/ImageNode.tsx
```diff
@@ -327,6 +329,25 @@ export type SerializedImageNode = Spread<
   SerializedLexicalNode
 >;
 
+function convertImageElement(domNode: Node): null | DOMConversionOutput {
+  if (domNode instanceof HTMLImageElement) {
+    const {alt: altText, src} = domNode;
+    const showCaption = domNode.getAttribute('data-show-caption') === 'true';
+    const captionJSON = domNode.getAttribute('data-caption-json');
```
acywatson, line 336: Should we namespace these with "lexical" to further lower the probability of collisions?

packages/lexical-link/src/index.ts
acywatson: That seems like really odd behavior. It just cuts the hacks out of the protocol? Docs seem to say otherwise too:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/href

I don't have a problem with this, more curious than anything.

packages/lexical/src/LexicalEditor.js
```diff
@@ -574,31 +564,22 @@ export class LexicalEditor {
     }
     this._pendingEditorState = editorState;
     this._dirtyType = FULL_RECONCILE;
+    this._dirtyElements.set('root', false);
```
acywatson, line 567: What is this change for?

trueadm: Fixed a long standing issue that this PR brought about. It's actually one that might make you smile!

When we build up the new EditorState for parsing the JSON, we use the typical node APIs, which means we assign to the dirtyElements, dirtyLeaves Maps/Sets. We were never clearing these down after, so we were polluting them mistakenly! We now properly reset them to what they were before – which means if you were to reconcile an editor state, nothing can be dirty as it's historical, but that also means that logic that skips doing anything when there are no dirty elements now incorrectly triggers – undo/redo/collab etc.

Marking the root as dirty gets around this and ensures they all work correctly (they worked before because the serializer incorrectly added them as per above).

packages/lexical-playground/src/nodes/ImageNode.tsx
trueadm: Yeah I can do.

thegreatercurve (approved):

PR #2242, opened 2022-05-23
Rename insert text command

trueadm:
The `INSERT_TEXT_COMMAND` is misleading. It appears to be a signal for a user to detect some user text insertion, and allowing them to override it or change what is coming in. Actually, this couldn't be further from the truth – it's actually an internal signal used to let plugins know of a controlled text insertion, rather than a non-controlled text insertion (we let the browser do it and do not override). This PR renames it to `CONTROLLED_TEXT_INSERTION` to better communicate intent.

I also added a piece in the FAQ to explain how to go about listening for text insertions the correct way – using node transforms.

### DIFF
--- a/packages/lexical-headless/src/__tests__/unit/LexicalHeadlessEditor.test.js
+++ b/packages/lexical-headless/src/__tests__/unit/LexicalHeadlessEditor.test.js
@@ -19,7 +19,7 @@ import {
   $createTextNode,
   $getRoot,
   COMMAND_PRIORITY_NORMAL,
-  INSERT_TEXT_COMMAND,
+  CONTROLLED_TEXT_INSERTION_COMMAND,
   ParagraphNode,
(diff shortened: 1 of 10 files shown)


zurfyx (approved):

echarles:
This issue is closed but would still love to have feedback on the following. I kind-of understand the argument around the a non-controlled text insertion (documented on https://lexical.dev/docs/faq#how-do-i-listen-for-user-text-insertions), but there I see there is a `DELETE_CHARACTER_COMMAND` that can be used and is invoked on each character deletion.
 https://github.com/facebook/lexical/blob/b804f2a3eeb407a58d2b78b91d51cab62ed09bd0/packages/lexical/src/LexicalCommands.ts#L17

What is the difference between character insertion or deletion. If we can listen to deletion, why can't we listen on insertion?

PR #2246, opened 2022-05-23
Move HTML<->Lexical functions to new package, @lexical/html.

### LINKED ISSUE #1834: HTML import/export utility
There seems to be some good demand in utilities to import and export from HTML. This comes from developers who either store it as HTML on the backend already (coming from other libraries) or want it to render.

Leaving aside that HTML is not as effective as EditorState I think we should explore the possibilities to enable such API. In a way we already have this in the @lexical/clipboard, it's just a matter of abstracting it in a way and ergonomic way that other developers can use as well as the copy-paste functionality.

tylerjbainbridge:
This PR strips out the clipboard HTML logic into generic utility functions for converting `Lexical` -> `HTML` and `HTML` -> `Lexical`. 

Closes #1834.
Website documentation added in #2249. 

Disclaimer: These functions are currently being tested in the `@lexical/clipboard` package, but at some point we can move some of them here.

How to use 👇 

### Exporting
```js
// When converting to HTML you can pass in a selection object to narrow it
// down to a certain part of the editor's contents.
const htmlString = $generateHtmlFromNodes(editor, selection | null);
```

### Importing
First we need to parse the HTML string into a DOM instance.
```js
// In the browser you can use the native DOMParser API to parse the HTML string.
const dom = parser.parseFromString(htmlString, textHtmlMimeType);

// In a headless environment you can use a package such as JSDom to parse the HTML string.
const dom = new JSDOM(htmlString);
```
And once you have the DOM instance.
```js

// Once you have the lexical nodes you can initialize an editor instance with the parsed nodes.
const editor = createEditor({ ...config, nodes });

// Or insert them at a selection.
selection.insertNodes(nodes);
```


packages/lexical-headless/src/LexicalHtml.ts
```diff
@@ -0,0 +1,236 @@
+/**
```
acywatson, line 1: Should we have @lexical/html instead?

packages/lexical-selection/src/index.ts
```diff
@@ -894,3 +894,48 @@ export function createRectsFromDOMRange(
   }
   return selectionRects;
 }
+
+export function $splitTextNode(
```
acywatson, line 898: What's going on here? I thought this already existed.

tylerjbainbridge: It needed to get moved here in order to be used by `/clipboard` and `/headless`, otherwise `/headless` would have to import all of `/clipboard` which feels wrong.

packages/lexical-headless/src/LexicalHtml.ts
tylerjbainbridge: @trueadm can probably add some color to this decision, but we decided to just add this directly to the `/headless` package so HTML processing can easily be done on the server.

tylerjbainbridge:
Website documentation added in #2249.

packages/lexical-selection/src/index.ts
acywatson: Oh I got this confused with TextNode.splitText - I thought it was in the core already.

I guess the follow up question would be if /selection is the best place for it, as opposed to /text maybe?

We had a situation before where a ton of things that were very loosely related to selection were all in "LexicalSelectionHelpers", which was a bit confusing.

tylerjbainbridge: Seems to me like it could fit nicely in either package, but happy to move it to the text if you want to avoid `/selection` getting too messy!

packages/lexical-headless/src/LexicalHtml.ts
acywatson: That kind of makes sense, but why not just pull in lexical/html on the server if you want to do that? There's no immediate problem here, it just seems like we're grouping things that are not strictly related. Why should clipboard depend on headless?

packages/lexical-selection/src/index.ts
acywatson: We can see what others think - I don't mind if it's here.

packages/lexical-headless/src/LexicalHtml.ts
tylerjbainbridge: I agree, I think it'd fit nicely in a `@lexical/html` package that can be used on the server. Curious what others think though.

trueadm: Let's put it in the html package. Makes more sense :)

packages/lexical-selection/src/index.ts
trueadm: It's fine to be here IMO, as it requires selection too.

acywatson (approved):
Looks right, nice work!

trueadm (approved):

trueadm:
I think you're missing the new package from build.js and a few other places?

scripts/build.js
```diff
@@ -51,6 +51,7 @@ const wwwMappings = {
   '@lexical/hashtag': 'LexicalHashtag',
   '@lexical/headless': 'LexicalHeadless',
   '@lexical/history': 'LexicalHistory',
+  '@lexical/html': 'LexicalHtml',
```
tylerjbainbridge, line 54: @trueadm it's here

acywatson:
> 

Everything looked right to me in there...

scripts/build.js
tylerjbainbridge: Everything compiles and builds locally, so I think it's in all the right places.

trueadm:
I was looking at a stale changes page, by bad!

packages/lexical-html/README.md
```diff
@@ -0,0 +1,35 @@
...
+// In the browser you can use the native DOMParser API to parse the HTML string.
+const parser = new DOMParser();
+const dom = parser.parseFromString(htmlString, textHtmlMimeType);
+
+// In a headless environment you can use a package such as JSDom to parse the HTML string.
+const dom = new JSDOM(htmlString);
+```
+And once you have the DOM instance.
+```js
+const nodes = $generateNodesFromDOM(editor, dom);
```
stof, line 27: should this really be a function named with a `$` prefix if one of its intended usage is to call it before initializing the editor ? The doc says that $-prefixed functions can only be called in the lexical scope of an editor closure, and this is not the case here.

PR #2251, opened 2022-05-23
Trim whitespace before applying text formatting during markdown export

### LINKED ISSUE #2250: Bug: Many `text-format` nodes with whitespace break markdown export
Lexical version: 0.29.0

When typing and combining different `text-format`'s on text nodes, i.e. "**bold** _italic_" it's easy to break the markdown export as it doesn't properly handle whitespace.

## Steps To Reproduce
1. type `cmd + b` to enable bold
2. type `foo ` (note the space) <-- the trailing space causes the bug
3. type `cmd + b` to disable bold
4. type `cmd + i` to enable italic
5. type `bar`
6. export markdown

You'll see the markdown export is `**foo ***bar*` which yields broken markdown.

## The current behavior

Whitespace can jumbles up the markdown leaf conversion

## The expected behavior
`@lexical/markdown` should be whitespace aware and properly apply `text-format` styles _around_ whitespace

hanford:
When typing and combining different text-format's on text nodes, i.e. "**bold** _italic_" it's easy to break the markdown export as it currently applies the formatting characters around included whitespace.

i.e. `"  foo  "` becomes `"**  foo  **"` when we actually want it to be `"   **foo**   "`

This PR fixes that

---

Before:
Screen Recording 2022-05-23 at 2 48 10 PM

After:
Screen Recording 2022-05-23 at 2 50 05 PM

closes #2250

### DIFF
--- a/packages/lexical-markdown/src/v2/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/v2/MarkdownExport.ts
@@ -123,7 +123,13 @@ function exportTextFormat(
   textContent: string,
   textTransformers: Array<TextFormatTransformer>,
 ): string {
-  let output = textContent;
+  // This function handles the case of a string looking like this: "   foo   "
+  // Where it would be invalid markdown to generate: "**   foo   **"
+  // We instead want to trim the whitespace out, apply formatting, and then
+  // bring the whitespace back. So our returned string looks like this: "   **foo**   "
+  const frozenString = textContent.trim();
+  let output = frozenString;
+
   const applied = new Set();
 
   for (const transformer of textTransformers) {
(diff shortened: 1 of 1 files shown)


packages/lexical-markdown/src/v2/MarkdownExport.ts
```diff
@@ -123,7 +123,13 @@ function exportTextFormat(
   textContent: string,
   textTransformers: Array<TextFormatTransformer>,
 ): string {
-  let output = textContent;
+  // This function handles the case of a string looking like this: "   foo   "
```
hanford, line 126: feel free to remove if this comment doesn't add value

hanford:
Unless I'm mistaken, I think these test failures are flakey and not a result of my PR

acywatson (approved):
Looks good! Can we cover this with a test here?:

https://github.com/facebook/lexical/blob/main/packages/lexical-playground/__tests__/e2e/Markdown-v2.spec.mjs

@fantactuka as code owner.

PR #2254, opened 2022-05-24
Add MaxLengthPlugin

trueadm:
This PR adds a MaxLengthPlugin, and an associated `@lexical/selection` helper function `trimTextContentFromAnchor` to make all the logic work well. Internally, we expose `RootNode` to be treated always as a dirty element if any of its children are dirty too.

### DIFF
--- a/packages/lexical-selection/src/index.ts
+++ b/packages/lexical-selection/src/index.ts
@@ -19,7 +19,9 @@ import type {
 } from 'lexical';
 
 import {
+  $createTextNode,
   $getDecoratorNode,
+  $getNodeByKey,
   $getPreviousSelection,
   $isDecoratorNode,
   $isElementNode,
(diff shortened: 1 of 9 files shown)


tylerjbainbridge (approved):
👍

packages/lexical-selection/src/index.ts
```diff
@@ -894,3 +896,121 @@ export function createRectsFromDOMRange(
...
+  let remaining: number = delCount;
+
+  if ($isElementNode(currentNode)) {
+    const descendantNode = currentNode.getDescendantByIndex(anchor.offset);
+    if (descendantNode !== null) {
+      currentNode = descendantNode;
+    }
+  }
+
+  while (remaining > 0 && currentNode !== null) {
```
tylerjbainbridge, line 920: Is there any reason (I'm sure there is...) that this loop can't be as simple as going to the previous text node sibling (this could be a good core node method) and deleting the node/slicing it's content?

tylerjbainbridge: I'm just struggling to understand all the cases this needs to cover from a high level

Issue #2230, opened 2022-05-20
Feature: maxLength


jack-guy:
Setting a maxlength for an editor was something that was [quite complicated](https://stackoverflow.com/a/46071494/1008741) using Draft.js. It's also a common need - particularly for developers relying on an editor primarily for plaintext functionality (as a replacement for `<textarea />`).

I see that there's some WIP work being done on [LexicalCharacterLimitPlugin](https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalCharacterLimitPlugin.tsx) but testing this feature out in the playground seems to show a different behavior than I expected - rather than limiting input it simply highlights input that's outside the character limit - 
![image](https://user-images.githubusercontent.com/957238/169620937-1a17b89e-b367-4ad4-8bd9-a840a89e2b32.png)

I also tried to poke around the source code for how I might implement this myself. My current, very naive implementation:
```
const LexicalMaxLengthPlugin = ({ maxLength }) => {
    const [editor] = useLexicalComposerContext();

    useEffect(() => {
        editor.registerTextContentListener(textContent => {
            if (textContent.length > maxLength) {
                editor.update(
                    () => {
                        $getSelection().deleteCharacter(true);
                    },
                    {
                        tag: 'history-merge',
                    },
                );
            }
        });
    }, [editor]);

    return null;
};
```
this has many problems (I wasn't really expecting it to work)
- very slow to delete excessive characters
- using a bad update pattern
- even with the `history-merge` tag, the history stack is broken. A keypress to enter a character beyond the maxlength is still added to the stack.

I'm curious if functionality to mimic the behavior of `<textarea maxlength=""` can be accomplished using Lexical as it exists today, or if there will need to be some new APIs implemented. In particular without some pretty tedious and potentially slow node manipulation it's not clear to me how one could truncate excess characters in a single update.

My team is trying to be an early adopter of Lexical - we're in the process of migrating our application to React 18 and draft.js is partially blocking that. Thanks!

trueadm:
> I'm curious if functionality to mimic the behavior of <textarea maxlength="" can be accomplished using Lexical as it exists today, or if there will need to be some new APIs implemented. In particular without some pretty tedious and potentially slow node manipulation it's not clear to me how one could truncate excess characters in a single update.

It should definitely be possible to do this. I might take a stab at hacking something together here!

trueadm:
Check out https://github.com/facebook/lexical/pull/2254!

[...]

PR #2258, opened 2022-05-24
Fix some bugs around MaxLength plugin

trueadm:
This fixes some bugs reported in #2230. Specifically:

- History plugin should ignore a dirty leaf if it hasn't actually changed.
- We should adjust the anchor offset in the case it comes before the remaining offset.

### DIFF
--- a/packages/lexical-history/src/index.ts
+++ b/packages/lexical-history/src/index.ts
@@ -192,6 +192,26 @@ function getChangeType(
   return OTHER;
 }
 
+function isTextNodeUnchanged(
+  key: NodeKey,
+  prevEditorState: EditorState,
+  nextEditorState: EditorState,
+): boolean {
+  const prevNode = prevEditorState._nodeMap.get(key);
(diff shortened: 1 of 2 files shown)


packages/lexical-history/src/index.ts
```diff
@@ -192,6 +192,26 @@ function getChangeType(
...
+  if ($isTextNode(prevNode) && $isTextNode(nextNode)) {
+    return (
+      prevNode.__type === nextNode.__type &&
+      prevNode.__text === nextNode.__text &&
+      prevNode.__mode === nextNode.__mode &&
+      prevNode.__detail === nextNode.__detail &&
+      prevNode.__style === nextNode.__style &&
+      prevNode.__format === nextNode.__format
+    );
+  }
```
fantactuka, line 211: Was thinking if we should have TextNode.is() similar to selection's Point, but does not seem to be that useful outside of this case

PR #2263, opened 2022-05-24
Fix Listeners

acywatson:
Spreads are loosely transpiled internally and only work with Arrays, not Sets.

### DIFF
--- a/packages/lexical/src/LexicalUpdates.ts
+++ b/packages/lexical/src/LexicalUpdates.ts
@@ -601,8 +601,7 @@ export function triggerListeners(
   editor._updating = isCurrentlyEnqueuingUpdates;
 
   try {
-    const listeners = [...editor._listeners[type]];
-
+    const listeners = [...editor._listeners[type].values()];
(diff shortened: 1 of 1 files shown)


trueadm:
Woah, why are we doing array spreads at all? This should be a simple Array.from(). It's not only faster but doesn't transpile to shit.

acywatson:
> Woah, why are we doing array spreads at all? This should be a simple Array.from(). It's not only faster but doesn't transpile to shit.

Yea, I know - this was changed in the TS migration. When I try to use Array.from now, there are bunch of type errors. I will see if I can figure it out, but would love to get this sync done :)

trueadm (approved):

acywatson:
> > Woah, why are we doing array spreads at all? This should be a simple Array.from(). It's not only faster but doesn't transpile to shit.
> 
> Yea, I know - this was changed in the TS migration. When I try to use Array.from now, there are bunch of type errors. I will see if I can figure it out, but would love to get this sync done :)

Fixed it properly - thanks for the push :)

PR #2269, opened 2022-05-25
Provide legacy editor conversion

trueadm:
This PR introduces a legacy EditorState JSON -> `EditorState` conversion. Making it easier to upgrade and convert from the old data that people might have stored, into an EditorState, where they can do `JSON.stringify` and get the new format.

```js
import { unsupported_convertLegacyJSONEditorState } from '@lexical/utils';

// Put in legacyJsonEditorState, which is using the old format!
const editorState = unsupported_convertLegacyJSONEditorState(editor, legacyJsonEditorState);
console.log(JSON.stringify(newEditorState)); // New format!

```

### DIFF
--- a/packages/lexical-react/src/LexicalAutoFocusPlugin.ts
+++ b/packages/lexical-react/src/LexicalAutoFocusPlugin.ts
@@ -7,7 +7,6 @@
  */
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
-import { unsupported_convertLegacyJSONEditorState } from '@lexical/utils';
 import {useEffect} from 'react';
 
 export function AutoFocusPlugin(): null {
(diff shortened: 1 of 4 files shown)


fantactuka (approved):

trueadm:
> There are no changes in importing "legacy" editor state via setEditorState, right?

You can set the created editor state like any other. This is a temporary stop gap, as we know that we'll be changing the core of EditorState soon, so when we do that, we'll need to remove this helper entirely.

thegreatercurve (approved):

acywatson (approved):
This should do nicely! Thanks!

fameoflight:
Maybe a dumb question but is lexical schema version stored in Json?  So we can show errors when editor is updated but state passed is old version 

Also wondering do I need to apply this if I was using editor state from 0.29.

trueadm:
@fameoflight In 0.3.0 we introduced formal JSON formats, so people could store their EditorState in a backwards compatible way (with typed serialized nodes, including versioning fields) in a place that was easily loadable/resumable. This format is not compatible with 0.2.9, but this PR provides a temporary way of converting the old JSON format to and EditorState, so you can generate the new JSON format from it and replace any places you may have stored the old format with the new one.

fameoflight:
Thank you. You are awesome dude :)

PR #2273, opened 2022-05-25
Tune MaxLengthPlugin behavior for restoration

trueadm:
Based off further feedback from https://github.com/facebook/lexical/issues/2230#issuecomment-1137838765. We can actually apply a heuristic where if the last editor state content was already at the max length, we just try and restore that editor state entirely.

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -393,12 +393,11 @@ export function $restoreEditorState(
   editor: LexicalEditor,
   editorState: EditorState,
 ): void {
+  const FULL_RECONCILE = 2;
   const nodeMap = new Map(editorState._nodeMap);
   const activeEditorState = editor._pendingEditorState;
   activeEditorState._nodeMap = nodeMap;
-  for (const [, node] of nodeMap) {
-    node.markDirty();
-  }
+  editor._dirtyType = FULL_RECONCILE;
   const selection = editorState._selection;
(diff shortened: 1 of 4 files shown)


packages/lexical-utils/src/index.ts
```diff
@@ -382,3 +388,17 @@ export function unstable_convertLegacyJSONEditorState(
...
+
+export function $restoreEditorState(
+  editor: LexicalEditor,
+  editorState: EditorState,
+): void {
+  const nodeMap = new Map(editorState._nodeMap);
+  const activeEditorState = editor._pendingEditorState;
+  activeEditorState._nodeMap = nodeMap;
+  for (const [, node] of nodeMap) {
+    node.markDirty();
```
fantactuka, line 400: Maybe `editor._dirtyType = FULL_RECONCILE`?

fantactuka (approved):

packages/lexical-utils/src/index.ts
trueadm: I thought that, but then won't other node transforms not run?

fantactuka: Didn't know that's the difference! But if it's a restored state, then it should already have all transformers resolved (similar as in history stack we don't want any transformers when undo/redo)

trueadm: The big difference is that is done via `setEditorState`. I guess it's fine to make that change though, we just need to keep on eye on this.

PR #2277, opened 2022-05-26
Unexpected language change for code block is fixed #2271

### LINKED ISSUE #2271: Bug: Unexpected language change for Code Block
Lexical version: latest

## Steps To Reproduce

1. Select new Code Block
2. Change the language to Python for example
3. Insert copied text

## The current behavior
The language changed to default JavaScript

## The expected behavior
Should keep Python

https://user-images.githubusercontent.com/22371328/170352258-e9ed6ab3-5c17-4a01-a401-9cb6dcb23a4d.mov

ozanyurtsever:
As the issue points, whenever you paste a new block of code when the code block is empt even though you have selected a different language for selector, it returns back to the javascript mode, which is the default language. I have created a new variable to store the selected language and then use it in the node creation process.

### DIFF
--- a/packages/lexical-code/src/index.ts
+++ b/packages/lexical-code/src/index.ts
@@ -64,6 +64,7 @@ import {
 import {Spread} from 'libdefs/globals';
 
 const DEFAULT_CODE_LANGUAGE = 'javascript';
+let selectedLanguage: string;
 
 type SerializedCodeNode = Spread<
   {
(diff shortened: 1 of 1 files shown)


fantactuka:
Will it work if add code block, then another one with different language, and then paste something into the first one?

thegreatercurve:
Closing this as we can't have module specific variables specific to nodes, as it will be overwritten if there is >1 CodeNode.

PR #2295, opened 2022-05-30
Feat: Add clear formatting button to playground

### LINKED ISSUE #2190: Feature: Add a "Clear formatting" button to the Playground
Add a "Clear formatting" to the Playground which will restore all text content to an unstyled paragraph.

im-adithya:
Closes #2190

https://user-images.githubusercontent.com/64399555/170961978-128faa38-23e9-4cdb-8e03-63e5f64d8f11.mp4

**Note:** Is it okay to leave the hashtag as is? Because lexical changes the node from text to hashtag whenever a string starts with "#". But due to changing the background color to default, the light blue background of the hashtag is being removed.

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
@@ -852,7 +852,7 @@ export default function ToolbarPlugin(): JSX.Element {
           }
         }
       }
-      // Hande buttons
+      // Handle buttons
       setFontSize(
         $getSelectionStyleValueForProperty(selection, 'font-size', '15px'),
       );
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
@@ -930,12 +930,8 @@ export default function ToolbarPlugin(): JSX.Element {
         selection.getNodes().forEach((node) => {
           if ($isTextNode(node)) {
             node.setFormat(0);
+            node.setStyle('');
           }
-        });
-        $patchStyleText(selection, {
-          'background-color': '#fff',
-          color: '#000',
-          'font-family': 'Arial',
         });
       }
       activeEditor.dispatchCommand(FORMAT_ELEMENT_COMMAND, 'left');
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
@@ -936,11 +936,8 @@ export default function ToolbarPlugin(): JSX.Element {
           'background-color': '#fff',
           color: '#000',
           'font-family': 'Arial',
-          'font-size': '15px',
         });
-        $wrapLeafNodesInElements(selection, () => $createParagraphNode());
       }
-      activeEditor.dispatchCommand(TOGGLE_LINK_COMMAND, null);
       activeEditor.dispatchCommand(FORMAT_ELEMENT_COMMAND, 'left');
     });
   }, [activeEditor]);
@@ -1002,9 +999,9 @@ export default function ToolbarPlugin(): JSX.Element {
     <div className="toolbar">
       <button
         onClick={clearFormatting}
-        title="Clear Formatting"
+        title="Clear text formatting"
         className="toolbar-item spaced"
-        aria-label="Clear">
+        aria-label="Clear all text formatting">
         <i className="format clear" />
       </button>
       <button
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
@@ -998,13 +998,6 @@ export default function ToolbarPlugin(): JSX.Element {
   return (
     <div className="toolbar">
       <button
-        onClick={clearFormatting}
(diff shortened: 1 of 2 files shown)


trueadm:
Could use an e2e test to confirm it works right?

im-adithya:
@trueadm Can you please let me know about handling the hashtag node as discussed above? That would help me in writing e2e for this functionality!

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
```diff
@@ -920,6 +922,29 @@ export default function ToolbarPlugin(): JSX.Element {
...
+          'background-color': '#fff',
+          color: '#000',
+          'font-family': 'Arial',
+          'font-size': '15px',
+        });
+        $wrapLeafNodesInElements(selection, () => $createParagraphNode());
+      }
+      activeEditor.dispatchCommand(TOGGLE_LINK_COMMAND, null);
+      activeEditor.dispatchCommand(FORMAT_ELEMENT_COMMAND, 'left');
+    });
```
thegreatercurve, line 945: Apologies, I was basing the original behaviour for this on the "Paste without format" behaviour in Google Docs and Word, but the "Clear All Formatting" behaviour is slightly different in that it doesn't necessarily remove links, hashtags, or list elements. 

It just clears the format and any text color, but NOT the size. More from [the Word docs here](https://support.microsoft.com/en-us/office/clear-all-text-formatting-c094c4da-7f09-4cea-9a8d-c166949c9c80): 
> You can easily clear all formatting (such as bold, underline, italics, color, superscript, subscript, and more) from your text and return your text to its default formatting styles.

I think therefore we can just do the below:

```suggestion
    activeEditor.update(() => {
      const selection = $getSelection();
      if ($isRangeSelection(selection)) {
        $selectAll(selection);
        selection.getNodes().forEach((node) => {
          if ($isTextNode(node)) {
            node.setFormat(0);
          }
        });
        $patchStyleText(selection, {
          'background-color': '#fff',
          color: '#000',
          'font-family': 'Arial',
        });
      }
      activeEditor.dispatchCommand(FORMAT_ELEMENT_COMMAND, 'left');
    });
```

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
```diff
@@ -975,6 +1000,13 @@ export default function ToolbarPlugin(): JSX.Element {
 
   return (
     <div className="toolbar">
+      <button
+        onClick={clearFormatting}
+        title="Clear Formatting"
+        className="toolbar-item spaced"
+        aria-label="Clear">
```
thegreatercurve, line 1007:
```suggestion
        aria-label="Clear all text formatting">
```

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
```diff
@@ -975,6 +1000,13 @@ export default function ToolbarPlugin(): JSX.Element {
 
   return (
     <div className="toolbar">
+      <button
+        onClick={clearFormatting}
+        title="Clear Formatting"
```
thegreatercurve, line 1005: Might be a little clearer for accessibility: 
```suggestion
        title="Clear text formatting"
```

thegreatercurve (changes requested):
LGTM! Just clarified some of the behaviour in the comments above which should also answer the hashtag question

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
```diff
@@ -975,6 +1000,13 @@ export default function ToolbarPlugin(): JSX.Element {
 
   return (
     <div className="toolbar">
+      <button
+        onClick={clearFormatting}
```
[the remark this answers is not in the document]
im-adithya: Thanks for the valuable inputs! Made all the necessary changes!

im-adithya:
@thegreatercurve shouldn't we leave the nodes which already have some pre-specified text-color (quote node and link node) or background-color (hashtags, mentions) while using the `$patchStyleText` helper function? Attaching these screenshots for reference:

Before Clearing:
Screenshot from 2022-05-30 21-20-36

After Clearing:
Screenshot from 2022-05-30 21-20-33

thegreatercurve:
> @thegreatercurve shouldn't we leave the nodes which already have some pre-specified text-color (quote node and link node) or background-color (hashtags, mentions) while using the `$patchStyleText` helper function? Attaching these screenshots for reference:
> 
> Before Clearing: Screenshot from 2022-05-30 21-20-36
> 
> After Clearing: Screenshot from 2022-05-30 21-20-33

@im-adithya Very good point. You're right, in other editors, they essentially reset the non-TextNodes to its original styles. 

In that case, I think we can still solve the issue with the existing APIs. There is a `setStyle` method on TextNode, which we can use to reset things such as font, and highlight or text colors. 

So, if you remove `$patchStyleText`, you can then do something like the below in your loop:

```js
 if ($isTextNode(node) || $isHashTahNode(node) || $isMentionNode(node)) {
    node.setFormat(0);
    node.setStyle('');
 }
```

im-adithya:
Done! Checking for just `$textNode` worked fine as the other two were extending from it!

thegreatercurve (approved):
LGTM. Thanks for working on this @im-adithya.

Issue #2255, opened 2022-05-24
Add Tests for MaxLength functionality


acywatson:
#2254 introduced a MaxLength plugin to the playground:

https://github.com/facebook/lexical/blob/90aad493addf837f05af86cd2a28871083e6e73e/packages/lexical-playground/src/plugins/MaxLengthPlugin.tsx#L14

as well as a helper function that contains most of the logic for enforcing the max length:

https://github.com/facebook/lexical/blob/90aad493addf837f05af86cd2a28871083e6e73e/packages/lexical-selection/src/index.ts#L904

It would be great to have an e2e test for the MaxLength plugin, as well as a suite of unit tests for the new helper function in lexical/selection.

im-adithya:
@acywatson please review :)

thegreatercurve:
Fixed in #2296

PR #2296, opened 2022-05-30
Add e2e test for MaxLengthPlugin

im-adithya:
Adds e2e for #2255 

https://user-images.githubusercontent.com/64399555/170990797-ea6962b4-b571-476f-9de9-4027b1cdaee1.mp4

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {
+  assertHTML,
+  focusEditor,
+  html,
+  initialize,
+  test,
+} from '../utils/index.mjs';
+
+test.describe('MaxLength', () => {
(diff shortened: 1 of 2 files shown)


zurfyx (approved):
Thank you, additional stuff you can test:

1. Selection
2. Pasting text on an empty/non-empty editor
3. IME on boundaries
4. Emojis (that can have variable size)

im-adithya:
Thanks for the inputs @zurfyx! This helped me find a bug(?) in MaxLengthPlugin.
I'm not quite sure because I don't exactly know how emojis work in lexical. As far as the issue is concerned, it is as follows:

https://user-images.githubusercontent.com/64399555/171148508-4f4dc82d-baeb-4b42-a75a-1bbc2bb20b73.mp4
1) Type any random 30 characters
2) Copy this emoji 💏 or any other emoji of length 2 (occupying a length of 2, otherwise this emoji is actually of length 11 as shown by `'💏'.length` which also can be an error if lexical is supposed to treat this as an 11-length emoji, but `a....a💏` (28 `a`s) works fine in MaxLength)
3) Remove a character and paste the emoji at the end (and ONLY at the end, otherwise the behavior is proper)

The entire editor gets cleared and the Block Format dropdown disappears! (I've found out that it is because the `blockType` of editor gets changed to root after being cleared)

I would like to work on this after getting to know a bit more about emoji guidelines as that would help in a better understanding of the functionality!

PR #2297, opened 2022-05-30
Fixed dropdown position in playground example

x-games:
In the playground, if page scroll is present then position of dropdown is not correct.


fantactuka (approved):

fantactuka:
Thanks for the fix!

PR #2300, opened 2022-05-31
Add includeHeaders argument to INSERT_TABLE_COMMAND

dbud:
`$createTableNodeWithDimensions` has an option to toggle headers on table cells. 
`INSERT_TABLE_COMMAND` enabled by LexicalTablePlugin should also accept this option (to disable marking first row/column as headers).


tylerjbainbridge (approved):
Great! Looks good to me. Thanks for making this change 👍

PR #2304, opened 2022-05-31
`@lexical/markdown` Fix strikethrough TextFormatTransformer order

### LINKED ISSUE #2303: Bug: Combining bold italic strikethrough can generate invalid markdown
Combining Bold, Italic and strikethrough formats can generate invalid markdown

Lexical version: 0.2.9

## Steps To Reproduce

1. Enable Bold, type (`Bold `)
2. Enable Italic, type (`Italic `)
3. Enable strike through, type (`bold italic strikethrough`)
4. View generated markdown

Note that the generated markdown is:
```
**bold *italic ~~strikethrough***~~
```

Which escapes strikethrough incorrectly, it _should_ be

```
**bold *italic ~~strikethrough~~***
```

## The current behavior
Invalid markdown is generated

## The expected behavior
Proper markdown should be generated

Current _bad_ state
Screen Shot 2022-05-31 at 12 42 06 PM

Screen Shot 2022-05-31 at 12 36 29 PM

Fixed state with open PR:
Screen Shot 2022-05-31 at 12 42 33 PM
Screen Shot 2022-05-31 at 12 36 55 PM

hanford:
This seems to fix it on my end!

closes #2303


fantactuka:
Unfortunately it's not about order of transformers, but more about order of opening tags, e.g. 

```
**bold ~~strikethrough~~**
~~strikethrough **bold**~~
```

Looks like [this part](https://github.com/facebook/lexical/blob/main/packages/lexical-markdown/src/v2/MarkdownExport.ts#L145-L154) will need to carry a stack of open tags in order to close them in a proper order

packages/lexical-markdown/src/v2/MarkdownExport.ts
```diff
@@ -139,21 +140,31 @@ function exportTextFormat(
...
 
       if (!hasFormat(previousNode, format)) {
         output = tag + output;
       }
+    }
+  }
 
-      // Prevent adding closing tag if next sibling will do it
-      const nextNode = getTextSibling(node, false);
+  appliedOrder.forEach(transformer => {
```
hanford, line 156: I wasn't able to do this in one iteration, given the small number of TextTransforms that doesn't seem to be an issue?

hanford: Spent most of my day looking at this and made some progress, but not with a solution I'm happy with:

```js
function getTransformOrder(node, textTransformers, reverse) {
  let appliedOrder = [];
  const applied = new Set();
  for (const transformer of textTransformers) {
    const format = transformer.format[0];

    if (hasFormat(node, format) && !applied.has(format)) {
      // Multiple tags might be used for the same format (*, _)
      applied.add(format); // Prevent adding opening tag is already opened by the previous sibling
      // Add most recent textTransform to beginning of appliedOrder where
      // we'll later iterate last-in-first-out when appending closing tags
      appliedOrder = [transformer, ...appliedOrder];
    }
  }

  return appliedOrder
}

function exportTextFormat(node, textContent, textTransformers) {
  let frozenString = textContent.trim()
  let output = frozenString
  
  // get siblings
  const previousSiblingNode = getTextSibling(node, true);
  const nextSiblingNode = getTextSibling(node, false);


  // get previous node formats
  let prevNode = getTransformOrder(previousSiblingNode, textTransformers)
  // get next node formats
  let nextNode = getTransformOrder(nextSiblingNode, textTransformers)

  // get current node formats
  let appliedOrder = getTransformOrder(node, textTransformers)

  // filter out common formats between current and previous nodes
  const prevFormats = appliedOrder.filter(x => !prevNode.includes(x))
  // filter out common formats between current and next nodes
  const nextFormats = appliedOrder.filter(x => !nextNode.includes(x))

 // reorder current node transforms contextually with which formats are surrounding us
  const reordered = [
    ...new Set([
      ...prevFormats,
      ...appliedOrder,
      ...nextFormats,
    ])
  ]

  reordered.forEach((transformer) => {
    const format = transformer.format[0];
    const tag = transformer.tag;
    if (!hasFormat(previousSiblingNode, format)) {
      output = tag + output
    } 

    // Prevent adding closing tag if next sibling will do it
    if (!hasFormat(nextSiblingNode, format)) {
      output += tag;
    }
  });

  return textContent.replace(frozenString, output);
} 
```

The best solution I could come up with, is just being more naive and removing the next/previous sibling code, it's a nice optimization to have, but I'm able to break it in a number of different ways in the playground

i.e. why not just wrap whatever lexical gives us as a leaf node with the appropriate tags rather than having to be smart about it what's around and traversing nearby nodes?
 
Would you be opposed to me removing the `getTextSibling` function and simplifying the tag additions?

fantactuka: Thanks for putting so much effort into this one! One problem with removing check for prev/next nodes will cause tags duplication:
```
<b>bold text <s>strikethrough</s></b>
->
**bold text ****~~strikethrough~~**
instead of
**bold text ~~strikethrough~~**
```

I'm thinking whether import/export would need to have double pass through:
Where in first pass we process nodes tree into objects like
```
{
  elementTag: '# ',
  children: [
    { text: 'hello', tags: ['*', '~~'] },
    { text: 'world', tags: ['*'] },
  ]
}
```
and then second one will generate final string using stack for open/close tags. Maybe it's still possible to do it in a single run tho. I'll give it a try over the weekend, but appreciate any ideas or suggestions!

JGonzalvoZipDev:
This is great!

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :trophy: This code is awesome&nbsp; | &nbsp;**Tags:** Readable, Efficient, Reusable

acywatson:
@fantactuka @hanford can we get this merged or closed?

digitalgopnik:
@hanford What do we need to finish merging this fix? :slightly_smiling_face:

yunhanw5:
Hi @hanford @fantactuka, Is there an alternate solution that made this closed? If not, will this be merged soon to fix the wrong format order? We are facing similar issues when applying underline and bold; the wrong markdown is **aa++bb**++ where it should be **aa++bb++**. Thanks

PR #2288, opened 2022-05-27
Confine character popup to rootElementRect

im-adithya:
Fixes #2287 

https://user-images.githubusercontent.com/64399555/170713005-71aa49ee-71f1-4c06-a837-2122cf96a623.mp4

**Note:** Would require suggestions on how to handle the popup if the screen size is less than that of popup (or if we can ignore that case altogether)

PR #2306, opened 2022-06-01
Confine link editor to rootElementRect

im-adithya:
Similar to #2288

https://user-images.githubusercontent.com/64399555/171335923-887131f9-8988-4359-bfb8-9da0efaf8c4e.mp4

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
@@ -149,7 +149,11 @@ function getSelectedNode(selection: RangeSelection): TextNode | ElementNode {
   }
 }
 
-function positionEditorElement(editor, rect) {
+function positionEditorElement(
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
```diff
@@ -192,6 +201,7 @@ function FloatingLinkEditor({editor}: {editor: LexicalEditor}): JSX.Element {
     }
 
     const rootElement = editor.getRootElement();
+    const rootElementRect = rootElement.getBoundingClientRect();
```
zurfyx, line 204: `updateLinkEditor` can trigger quite often and this doesn't necessarily have to run unless the the selection is collapsed or need to be hidden, should we move this inside the `if` and `else if` (you can copy-paste)

zurfyx (approved):
Thank you!

PR #2309, opened 2022-06-01
fix mispositioning of treeview caret

im-adithya:
As the tree view uses " " to display strings, the caret (^) is mispositioned by one character each time text is selected.

Before:
Screenshot from 2022-06-01 17-49-02

After:
Screenshot from 2022-06-01 17-48-55


zurfyx (approved):
Makes sense, thanks!

PR #2321, opened 2022-06-02
Disable checklist items in readOnly mode

rulikrulit:
https://user-images.githubusercontent.com/6481831/171696493-6e2e66f9-0284-4729-8083-5269a38f9818.mov


fantactuka (approved):

fantactuka:
Wonder if this behaviour should be optional, but then it means will need to have onChange for read mode which may look weird

PR #2322, opened 2022-06-02
Add split node method

im-adithya:
This brings up the closed PR #1410 again with mainly the `splice` method fix as `splice` usually returns the removed items (we have been returning the modified node instead) and have changed the usage in `append` method as well.

Regarding the selection @trueadm was [talking about](https://github.com/facebook/lexical/pull/1410#issuecomment-1057833155), the `splitText` method doesn't use a built-in method like `splice` so we had to hard code the selection restoration but here we have `splice` which handles the selection updates for us!

Here's an example (ignore the usage of Normal button for the update, have used it just for the sake of testing. It essentially finds the parent of `listItemNode` and applies `splitText` method on its parent (i.e. `ListNode`) instead of wrapping leaf nodes)

https://user-images.githubusercontent.com/64399555/171723528-5f2b1be9-c916-44d6-b596-c72000e4974f.mp4


packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -428,7 +431,43 @@ export class ElementNode extends LexicalNode {
...
+    const thisIndex = children.indexOf(self.__key);
+    if (thisIndex === -1) {
+      invariant(false, 'Node is not a child of its parent');
+    }
+
+    const beforeNode = self;
+    const beforeNodeWritable = beforeNode.getWritable();
+
+    // create a node of the same type
+    const afterNode = new self.constructor();
```
im-adithya, line 456: @trueadm is there any helper function we can use here?
Which can return a new instance given the type of `elementNode`?

trueadm: Nodes all have a `clone` method that would be the right approach instead of doing this.

im-adithya: But using clone copies the key which makes the child nodes attach to the same parent node again... Is there any way to get the last key so that we can increment it and then update the clone with the new key?
Or can we stick with this because constructor always gives the latest key (and we have to add `// @ts-expect-error` in either cases to ignore the error)

im-adithya: I didn't get you correctly. Here's what I'm replacing the above line with:
```const afterNode = self.constructor.clone(self);```

And this expects a node as an argument. An example from `ListNode`:
```
static clone(node: ListNode): ListNode {
    const listType = node.__listType || TAG_TO_LIST_TYPE[node.__tag];

    return new ListNode(listType, node.__start, node.__key);
  }
```

Comment (trueadm) [reply packages/lexical/src/nodes/LexicalElementNode.ts:456]: What you need to do is store the key before you call `clone.()` and make it undefined. Then when you call clone, you'll get a new key, and you can re-apply the old key if needed.

Comment (im-adithya) [reply packages/lexical/src/nodes/LexicalElementNode.ts]: Done @trueadm, please review :D

File: packages/lexical/src/nodes/LexicalElementNode.ts [hunk 2]
```diff
@@ -428,7 +432,44 @@ export class ElementNode extends LexicalNode {
...
 
-    return writableSelf;
+    return removedNodes;
+  }
+  splitNode(index: number): [ElementNode, LexicalNode, ElementNode] {
+    errorOnReadOnly();
+    const self = this.getLatest();
+    const splitChild = self.getChildAtIndex(index);
+    if (splitChild == null) {
+      invariant(false, 'cannot split node, the child is out of bound');
```
acywatson, line 442: Make sure we run ```npm run extract-codes``` to updates the codes.json with these invariant messages.

packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -428,7 +432,44 @@ export class ElementNode extends LexicalNode {
...
+
+    // remove elements after a split point and save them to the variable
+    const nodesToMove = beforeNodeWritable.splice(
+      index,
+      childrenLength - index,
+      [],
+    );
+    afterNode.append(...nodesToMove);
+
+    return [beforeNodeWritable, splitChild, afterNode];
```
acywatson, line 472: I think we should just return beforeNode here, not the mutable one.

packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -428,7 +432,44 @@ export class ElementNode extends LexicalNode {
...
+    const beforeNode = self;
+    const beforeNodeWritable = beforeNode.getWritable();
+
+    // create a node of the same type
+    // @ts-expect-error
+    const afterNode = self.constructor.clone(self);
+    $setNodeKey(afterNode, undefined);
+
+    // add the node right after the node we are splitting
+    writableParent.splice(thisIndex + 1, 0, [afterNode]);
```
acywatson, line 462: can't we just use self.insertAfter?

acywatson (commented):
I think it would be great to polish this up and get it merged - I think we have a use case for it in @lexical/list and probably elsewhere. However, this would definitely need tests before we can get it in.

packages/lexical/src/nodes/LexicalElementNode.ts
im-adithya: Done!

im-adithya:
> this would definitely need tests before we can get it in.

I'm not sure what to test this on. Is it just Listnodes...


and ParagraphNodes with emojis?

Screenshot from 2022-07-07 20-59-07

packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
```diff
@@ -613,4 +613,50 @@ describe('LexicalElementNode tests', () => {
       });
     });
   });
+
+  describe('splitNode', () => {
```
im-adithya, line 617: @acywatson done!

acywatson:
Had some internal discussion around this. We don't want it on ElementNode because it will probably be used pretty rarely and will add bloat to the core library. If you don't mind moving this to lexical-utils, that would remove the size concern.

im-adithya:
Sure, no issues :)

im-adithya:
All unit tests are passing locally, not sure what's happening here

acywatson:
> All unit tests are passing locally, not sure what's happening here

```
/home/runner/work/lexical/lexical/packages/shared/src/invariant.ts:17
    throw new Error('Internal Lexical error: invariant() is meant to be replaced at compile ' +
          ^

Error: Internal Lexical error: invariant() is meant to be replaced at compile time. There is no runtime version.
    at invariant (/home/runner/work/lexical/lexical/packages/shared/src/invariant.ts:21:9)
    at splitNode (/home/runner/work/lexical/lexical/packages/lexical/src/LexicalUtils.ts:1101:14)
    at /home/runner/work/lexical/lexical/packages/lexical/src/__tests__/unit/LexicalUtils.test.ts:215:34
    at beginUpdate (/home/runner/work/lexical/lexical/packages/lexical/src/LexicalUpdates.ts:798:5)
    at updateEditor (/home/runner/work/lexical/lexical/packages/lexical/src/LexicalUpdates.ts:905:5)
    at LexicalEditor.update (/home/runner/work/lexical/lexical/packages/lexical/src/LexicalEditor.ts:396:43)
    at /home/runner/work/lexical/lexical/packages/lexical/src/__tests__/unit/LexicalUtils.test.ts:214:22
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/home/runner/work/lexical/lexical/packages/lexical/src/__tests__/unit/LexicalUtils.test.ts:213:19)
    ```
    
Not 100% sure, but I think your test is hitting your out-of-bounds invariant, which is giving this message because we're not running tests against the built bundles (in which the invariant would be transformed).  We should improve how the tests handle this, but I suspect one of your tests cases is actually failing.

**Comment by acywatson** [comment]
Can we clean this one up and fix that unit test?

**Comment by im-adithya** [comment]
> but I suspect one of your tests cases is actually failing.

Not just one, the entire set of tests were wrong and weren't even being checked (hence passing) :facepalm:, fixed it!

File: packages/lexical/src/nodes/LexicalElementNode.ts [hunk 6]
```diff
@@ -319,13 +320,14 @@ export class ElementNode extends LexicalNode {
     start: number,
     deleteCount: number,
     nodesToInsert: Array<LexicalNode>,
-  ): this {
+  ): Array<LexicalNode> {
```
acywatson, line 323: mm, I just realized this is a breaking change, but it does match the JS Array.splice signature, so I'm good.

fantactuka:
QQ: to my understanding this would work 1-2 levels deep, is it worth supporting splitting nodes up until root/shadow root?

DaniGuardiola:
@fantactuka I have a use case for splitting up to the root!

thegreatercurve:
Closing this as it was addressed instead in #3342.

PR #2324, opened 2022-06-02
Add default exportDOM and importDOM methods

tylerjbainbridge:
This PR adds robust fallback logic for `exportDOM` and `importDOM` on the base `LexicalNode` class.

https://user-images.githubusercontent.com/13852400/171728151-9b5dfa4a-c239-4376-9b89-935242d697cf.mov

### DIFF
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -136,7 +136,20 @@ function getConversionFunction(
 ): DOMConversionFn | null {
   const {nodeName} = domNode;
 
-  const cachedConversions = editor._htmlConversions.get(nodeName.toLowerCase());
+  const baseCacheConversions =
+    domNode instanceof HTMLElement &&
+    domNode.hasAttribute('data-lexical-node-type')
+      ? editor._htmlConversions.get('*')
+      : [];
+
+  const nodeNameCacheConversions = editor._htmlConversions.get(
+    nodeName.toLowerCase(),
+  );
+
+  const cachedConversions = [
+    ...nodeNameCacheConversions,
+    ...baseCacheConversions,
+  ];
 
   let currentConversion: DOMConversion | null = null;
 
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -143,7 +143,7 @@ function getConversionFunction(
       : [];
 
   const nodeNameCacheConversions =
-    editor._htmlConversions.get(nodeName.toLowerCase()) ?? [];
+    editor._htmlConversions.get(nodeName.toLowerCase()) || [];
 
   const cachedConversions = [
     ...nodeNameCacheConversions,
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -142,9 +142,8 @@ function getConversionFunction(
       ? editor._htmlConversions.get('*')
       : [];
 
-  const nodeNameCacheConversions = editor._htmlConversions.get(
-    nodeName.toLowerCase(),
-  );
+  const nodeNameCacheConversions =
+    editor._htmlConversions.get(nodeName.toLowerCase()) ?? [];
 
   const cachedConversions = [
     ...nodeNameCacheConversions,
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -145,10 +145,9 @@ function getConversionFunction(
   const nodeNameCacheConversions =
     editor._htmlConversions.get(nodeName.toLowerCase()) || [];
 
-  const cachedConversions = [
-    ...nodeNameCacheConversions,
-    ...baseCacheConversions,
-  ];
+  const cachedConversions = Array.from(nodeNameCacheConversions).concat(
+    baseCacheConversions,
+  );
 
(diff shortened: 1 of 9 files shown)


fantactuka (approved):

packages/lexical-html/src/index.ts
```diff
@@ -136,7 +136,19 @@ function getConversionFunction(
...
+  const baseCacheConversions =
+    domNode instanceof HTMLElement &&
+    domNode.hasAttribute('data-lexical-node-type')
+      ? editor._htmlConversions.get('*')
+      : [];
+
+  const nodeNameCacheConversions =
+    editor._htmlConversions.get(nodeName.toLowerCase()) || [];
+
+  const cachedConversions = [
```
trueadm, line 148: Use Array.from rather than spread arrays.

packages/lexical/src/LexicalEditor.ts
```diff
@@ -312,13 +312,15 @@ export function createEditor(
...
         });
-        if (proto instanceof DecoratorNode || proto instanceof ElementNode) {
+        if (
           // eslint-disable-next-line no-prototype-builtins
-          if (!klass.hasOwnProperty('importDOM')) {
-            console.warn(
-              `${name} should implement "importDOM" method to ensure HTML serialization (important for copy & paste) works as expected`,
-            );
-          }
+          !klass.hasOwnProperty('importDOM') &&
```
trueadm, line 317: What if I have importDOM but not exportDOM?

packages/lexical/src/LexicalNode.ts
```diff
@@ -609,20 +600,48 @@ export class LexicalNode {
...
-
     const element = this.createDOM(editor._config, editor);
-
+    const serializedNode = this.exportJSON();
+    element.setAttribute('data-lexical-node-type', this.getType());
+    element.setAttribute(
+      'data-lexical-node-json',
+      JSON.stringify(serializedNode),
+    );
+    element.setAttribute('data-lexical-editor-key', editor.getKey());
```
trueadm, line 610: It's core, you can use editor.__key.

packages/lexical/src/LexicalNode.ts
```diff
@@ -609,20 +600,48 @@ export class LexicalNode {
...
-      const element = editor.getElementByKey(this.getKey());
-      return {
-        element: element ? (element.cloneNode() as HTMLElement) : null,
-      };
-    }
-
     const element = this.createDOM(editor._config, editor);
-
+    const serializedNode = this.exportJSON();
+    element.setAttribute('data-lexical-node-type', this.getType());
```
trueadm, line 605: It's core you can use this.__type

packages/lexical/src/LexicalNode.ts
```diff
@@ -609,20 +600,48 @@ export class LexicalNode {
...
+    const proto: any = this.prototype.constructor;
+    return {
+      // Catch-all key because we don't know the nodeName of the element returned by exportDOM.
+      '*': (domNode: Node) => {
+        if (!(domNode instanceof HTMLElement)) return null;
+        const editorKey = domNode.getAttribute('data-lexical-editor-key');
+        const nodeType = domNode.getAttribute('data-lexical-node-type');
+        if (editorKey == null || nodeType == null) return null;
+        const editor = getActiveEditor();
+        if (editorKey === editor.getKey() && nodeType === proto.getType()) {
```
trueadm, line 625: Again, you can use the properties directly.

packages/lexical/src/LexicalNode.ts
```diff
@@ -609,20 +600,48 @@ export class LexicalNode {
...
 
   static importDOM(): DOMConversionMap | null {
-    return null;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const proto: any = this.prototype.constructor;
+    return {
+      // Catch-all key because we don't know the nodeName of the element returned by exportDOM.
+      '*': (domNode: Node) => {
+        if (!(domNode instanceof HTMLElement)) return null;
+        const editorKey = domNode.getAttribute('data-lexical-editor-key');
```
trueadm, line 621: We might want to ensure editor's use createUID, I think it might be just an incriminator, which means it might break between refreshing as the id might be the same.

packages/lexical/src/LexicalEditor.ts
tylerjbainbridge: That case is actually fine since the fallback `exportDOM` will use the output of `createDOM` and most likely if the author of the Node is writing their own `importDOM` they'll account for that `HTML` `tagName`.

This is done all over Lexical Core and in the Playground. If you add the inverse warning virtually every node fails the check and clogs the console despite being valid.

packages/lexical/src/LexicalNode.ts
tylerjbainbridge: Yup, just checked, it's incrementing.

That _might_ be what we want (re: refreshing), lol. But you're right, two editors in different windows will likely have the same key despite not originating from the same editor.

trueadm: Let's make the editor key use `createUID` then, it's already available in core somewhere I believe?

trueadm (approved):

PR #2332, opened 2022-06-03
fix placeholder overflow

im-adithya:
Just a small fix
### Before
Screenshot from 2022-06-03 15-25-32
### After
Screenshot from 2022-06-03 15-25-11


trueadm (approved):
Thanks

PR #2343, opened 2022-06-05
Autocomplete v2

zurfyx:
Autocomplete is one of the Lexical's oldest and unmaintained playground plugins and the quality behind the implementation has fallen behind the rest of the plugins.

https://user-images.githubusercontent.com/193447/172059339-a9ea5590-fb09-4f37-9a6e-93b8afd4b3b7.mov

V2 is a rewrite that supports the following:

**Collab**

Autocomplete will show up in Collab and multiple users can interact with their own autocompletes. At the same time, autocompletion is private, not just for the sake of readability but it's also important for a privacy point of view (i.e. autocomplete suggestions based on previous user patterns).

**Inert**

We want to get rid of inert since early this year. It introduces various selection edge cases that were introduced exclusively for autocomplete. From an a11y stand point we can do the same with a DecoratorNode so there's no need to support this additional functionality.

**Modular**

Not in this PR but some of the methods already hint towards. The idea is to eventually move this module to @lexical/react and support a custom Server and search heuristics.

**Mobile**

Swipe support for mobile.


trueadm (approved):
Beautiful

PR #2346, opened 2022-06-06
doc: fix wrong components names

samuelmartineau:
Introduced by 0.3.0 global renaming.

I haven't checked if it's generalized to all the documentation 🤔


trueadm (approved):
Thanks

PR #2359, opened 2022-06-07
Autogenerate TypeScript declaration files

thegreatercurve:
- Remove `.d.ts` files and set absolute path aliases to `.ts` files.
- Fix any resulting type errors.
- Add logic to build pipeline to emit type declarations, and then move them into relevant `dist` folders. 
- Refactored the build scripts to run in sequence. (It takes a little while longer, but avoids some OOM errors I was getting with an early implementation).

Tested the integration by running `npm run prepare-release` and then copying files over to a test TypeScript integration of Lexical which uses both `lexical` and `@lexical/react`,


trueadm (approved):

thegreatercurve:
Sorry to ask but can I get a re-review? I had to fix some bugs in the playground that weren't covered in the original tsconfig

trueadm (approved):

zurfyx (approved):

echarles:
That is a great move. I see the flow folders are still present. If an API changes, the d.ts will be automatically generated, does the flow files still need to be manually updated (disclaimer: I have never used flow). Nothing is written about that in https://github.com/facebook/lexical/blob/main/CONTRIBUTING.md

PR #2363, opened 2022-06-08
paste: import direction, format and indent information

im-adithya:
Currently, we remember the formatting information of ElementNodes on copying, but we don't apply them on paste.

https://user-images.githubusercontent.com/64399555/172604781-2b453b80-ee9c-47a3-98e0-0629b66bfbb1.mp4


tylerjbainbridge:
Hi @im-adithya, @trueadm and I came up with a plan that'll fix this directly in core. It should be resolved by tomorrow!

tylerjbainbridge:
Resolved by #2370!

PR #2370, opened 2022-06-08
Add stringified LexicalNodes to clipboard for lossless Lexical->Lexical copy and paste.

tylerjbainbridge:
Turns out that adding default `importDOM` and `exportDOM` methods that are completely lossless is very difficult and risky, because someone can add their own custom method(s) and not catch all of the data on the DOM node. 

To prevent this issue when copying and pasting between Lexical documents we're reusing the `exportJSON` and `importJSON` methods (which most of your Nodes should already have) to stringify the selection and add it to the clipboard under `'application/x-lexical-editor'`. Now there's one less thing to worry about as a developer and you can be certain copy and paste is _completely_ lossless when in the same Lexical namespace.

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -10,7 +10,6 @@ import type {
   GridSelection,
   LexicalEditor,
   LexicalNode,
-  NodeKey,
   NodeSelection,
   RangeSelection,
 } from 'lexical';
@@ -35,6 +34,7 @@ import {
   $isLineBreakNode,
   $isRangeSelection,
   $isTextNode,
+  $parseSerializedNode,
   $setSelection,
   GridNode,
   SELECTION_CHANGE_COMMAND,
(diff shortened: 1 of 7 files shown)


trueadm (approved):
LGTM. Don't forget to run `extract-codes` to build the new invariant codes file.

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -382,3 +317,129 @@ function $mergeGridNodesStrategy(
...
+  let clone = $cloneWithProperties<LexicalNode>(currentNode);
+  clone =
+    $isTextNode(clone) && selection != null
+      ? $sliceSelectedTextNodeContent(selection, clone)
+      : clone;
+  const children = $isElementNode(clone) ? clone.getChildren() : [];
+
+  const serializedNode = exportNodeToJSON(clone);
+
+  // TODO: TextNode calls getTextContent() (NOT node.__text) within it's exportJSON method
```
trueadm, line 374: I'm a little confused by this, as many nodes call methods rather than using __properties directly, and this is actually preferred. Why doesn't using `getTextContent` work here? Isn't `$cloneWithProperties` creating a new key for the node?

tylerjbainbridge: Nope, it's not, it just creates another instance of the node with the same key.

And if you manually generate a new key in the cloneWithProperties method without adding to the editor state node map (which you wouldn't want to do here), it throws an error.

tylerjbainbridge: It's a good thing that they use the methods to set/get the properties, yes, but it's a bad thing (imo) that when you call those methods on a **clone**, you might be modifying or reading the original node because of the getLatest and getWritable calls.

tylerjbainbridge: This still feels really risky because the term `clone` entails.. well, that it's a clone, and **not** the original. And right now you could accidentally mess up your editor state by calling methods on a "clone", which _feels_ safe, and have no idea.

tylerjbainbridge: It's fine for this PR, but moving forward we should rethink this naming and maybe use the term "duplicate" or something like that.

PR #2371, opened 2022-06-09
add code threshold for highlighting

lateefazeez:
## Starter Task 1
 - Add option to highlight plugin to stop running highlighting after certain code length threshold
 - Threshold currently set to 5
Screen Shot 2022-06-08 at 6 06 23 PM
Screen Shot 2022-06-08 at 6 06 59 PM


packages/lexical-code/src/index.ts
```diff
@@ -716,7 +716,8 @@ function codeNodeTransform(node: CodeNode, editor: LexicalEditor) {
         const highlightNodes = getHighlightNodes(tokens);
         const diffRange = getDiffRange(node.getChildren(), highlightNodes);
         const {from, to, nodesForReplacement} = diffRange;
-        if (from !== to || nodesForReplacement.length) {
+        console.log("Length: ", code.length)
```
trueadm, line 719: Console.log?

packages/lexical-playground/src/plugins/CodeHighlightPlugin.ts
```diff
@@ -14,7 +14,7 @@ export default function CodeHighlightPlugin(): JSX.Element | null {
   const [editor] = useLexicalComposerContext();
 
   useEffect(() => {
-    return registerCodeHighlighting(editor);
+    return registerCodeHighlighting(editor, 50);
```
fantactuka, line 17: You can use 50 as a test value, but for a playground it should be 5000-10000

packages/lexical-code/src/index.ts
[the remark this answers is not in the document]
lateefazeez: ooops! will make sure to always take them out

packages/lexical-playground/src/plugins/CodeHighlightPlugin.ts
lateefazeez: Noted

thegreatercurve:
@lateefazeez Maybe just rebase this branch onto main again and then push it to remote? For some reason the E2E test suite doesn't to seem to have run completely.

PR #2375, opened 2022-06-09
Fix bad target issue for backspace/delete

trueadm:
When we handle backspace/delete keys we should first validate the target is not within another editable field.

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -6,7 +6,7 @@
  *
  */
 
-import type {
+ import type {
   DOMConversionMap,
   DOMConversionOutput,
   EditorConfig,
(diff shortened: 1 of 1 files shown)


packages/lexical-rich-text/src/index.ts
```diff
@@ -669,6 +675,10 @@ export function registerRichText(
     editor.registerCommand<KeyboardEvent>(
       KEY_BACKSPACE_COMMAND,
       (event) => {
+        // @ts-ignore event.target will be a HTMLElement here
+        if (isTargetWithinDecorator(event.target)) {
```
thegreatercurve, line 679: The docs I've seen around this issue appear to suggest just casting as below: 

```suggestion
        if (isTargetWithinDecorator(event.target as HTMLElement)) {
```

thegreatercurve (approved):

PR #2378, opened 2022-06-09
Prevent content duplication for initialEditorState function

fantactuka:
Content might be duplicated as it's assumed `initialEditorState` is only affects content once. The fix would be to check whether root is empty within `initialEditorState` but I think core should do that check instead


packages/lexical-rich-text/src/index.ts
```diff
@@ -355,7 +354,12 @@ function initializeEditor(
         break;
       }
       case 'function': {
-        editor.update(initialEditorState, updateOptions);
+        editor.update(() => {
```
zurfyx, line 357: If I'm reading this right this is inside an `editor.update` already

zurfyx (approved):
LGTM, I guess this can be a problem with double invoke effects?

packages/lexical-plain-text/src/index.ts
```diff
@@ -146,7 +144,12 @@ function initializeEditor(
       }
 
       case 'function': {
-        editor.update(initialEditorState, updateOptions);
+        editor.update(() => {
+          const root = $getRoot();
+          if (root.isEmpty()) {
+            initialEditorState();
```
trueadm, line 150: Can we also pass `editor` to this function so it's a bit more useful?

packages/lexical-rich-text/src/index.ts
fantactuka: Lol this made me sweating! I see what you mean here, but opened VS code and couldn't see it anymore :)
It's all Github with semi-collapsed lines, if you expand lines in between then `editor.update` only affects if branch, but not else part

packages/lexical-rich-text/src/index.ts
```diff
@@ -330,8 +330,7 @@ function initializeEditor(
   } else if (initialEditorState === undefined) {
     editor.update(() => {
```
trueadm, line 331: It still is in an update. This update.

fantactuka: It ends before `else`, so `switch (typeof initialEditorState) {` part has its own .update call. (Need to expand collapsed lines)

trueadm: Ah yeah, so confusing!

trueadm:
Needs a prettier run, otherwise, awesome stuff!

trueadm:
I think you might have missed updating the Flow definition files?

trueadm:
It just occurred to me that this PR might not cause the root element to have the latest content. If the root is not empty, we should probably be doing `setEditorState(editorState)` to force a full reconciliation, in-case the DOM node for the content editable is empty (because React cleared it).

PR #2381, opened 2022-06-09
fix overflow issue in input field

im-adithya:
Reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1067136
This would cause issues when there's less space like this:
# Before
Screenshot from 2022-06-09 23-43-33
# After
Screenshot from 2022-06-09 23-47-57

### DIFF
--- a/packages/lexical-playground/src/ui/Input.css
+++ b/packages/lexical-playground/src/ui/Input.css
@@ -28,4 +28,5 @@
   padding-right: 10px;
   font-size: 16px;
   border-radius: 5px;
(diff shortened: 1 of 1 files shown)


trueadm (approved):
Thank you

im-adithya:
> Thanks for the fix! Do you mind adding before/after as I'm not quite sure what was broken

Sorry for that, updated!
(I was working on input for the color picker while I encountered this, so I only had that example, sorry for the confusion :P)

PR #2382, opened 2022-06-09
Fix/code highlighting split

lateefazeez:
## Split Code Highlighting Plugin
 - Split code highlighting plugin into highlighting and editor shortcuts (indent multiple lines, keep tab level on a new line

https://user-images.githubusercontent.com/53082437/172921529-6383f004-c075-4f9e-96ac-a38cb50459a3.mov

)


packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -0,0 +1,674 @@
...
+    }
+    if ($isLineBreakNode(node)) {
+      break;
+    }
+  }
+
+  return currentNode;
+}
+
+export function getLastCodeHighlightNodeOfLine(
```
fantactuka, line 97: This one should probably go into EditorShortcuts as it's not used here

packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -0,0 +1,674 @@
...
+  return cell.classList.contains('js-file-line');
+}
+
+function isGitHubCodeTable(table: HTMLTableElement): table is HTMLTableElement {
+  return table.classList.contains('js-file-line-container');
+}
+
+const LANGUAGE_DATA_ATTRIBUTE = 'data-highlight-language';
+
+export class CodeNode extends ElementNode {
```
fantactuka, line 175: Do you mind extracting CodeNode & all related helpers (including $isCodeNode, $createCodeNode) into own file, and same for CodeHighlightingNode. It should make highlighting plugin file easier to navigate through

packages/lexical-code/src/EditorShortcuts.ts
```diff
@@ -0,0 +1,490 @@
...
+        offset = i;
+        break;
+      }
+    }
+  }
+
+  return offset;
+}
+
+export function getStartOfCodeInLine(anchor: LexicalNode): {
```
fantactuka, line 119: This one does not seem to be used anywhere else, so no need to export it. Same applies to few other exports: try to move helpers into plugins files where they are used and only export node classes + helpers that are used as API ($isCodeNode, registerCodeHighlighting, etc). The rest can remain as local helpers without exporting

fantactuka (changes requested):

lateefazeez:
## Changes
 - I rearranged the codes, created separate files for CodeHighlightNode, CodeNode, Code Highlighter and EditorShortcuts as requested. Please review.

packages/lexical-code/LexicalCode.d.ts
```diff
@@ -78,7 +78,11 @@ declare function $isCodeHighlightNode(
   node: LexicalNode | null | undefined,
 ): node is CodeHighlightNode;
 
-declare function registerCodeHighlighting(editor: LexicalEditor): () => void;
+declare function registerCodeHighlighting(
+  editor: LexicalEditor,
+  threshold: number,
+): () => void;
+declare function registerCodeIndent(editor: LexicalEditor): () => void;
```
fantactuka, line 85: Let's remove `registerCodeIndent` from d.ts file for now, it is part of `registerCodeHighlighting` for now. I think we should eventually split them and call separately, but it'll be a breaking change so it's not urgent to do now

packages/lexical-code/LexicalCode.d.ts
```diff
@@ -78,7 +78,11 @@ declare function $isCodeHighlightNode(
   node: LexicalNode | null | undefined,
 ): node is CodeHighlightNode;
 
-declare function registerCodeHighlighting(editor: LexicalEditor): () => void;
+declare function registerCodeHighlighting(
+  editor: LexicalEditor,
+  threshold: number,
```
fantactuka, line 83: It should be optional argument, with `Infinity` as default value

packages/lexical-playground/src/plugins/CodeHighlightPlugin.ts
```diff
@@ -14,7 +14,7 @@ export default function CodeHighlightPlugin(): JSX.Element | null {
   const [editor] = useLexicalComposerContext();
 
   useEffect(() => {
-    return registerCodeHighlighting(editor);
+    return registerCodeHighlighting(editor, 50);
```
fantactuka, line 17: Let's have it 10000 for a playground

fantactuka (approved):
Couple minor changes needed, but looks good otherwise 👍

packages/lexical-playground/src/plugins/CodeHighlightPlugin.ts
lateefazeez: Okay, will do same as above

lateefazeez:
I updated this code to resolve all conflicts, but somehow not showing up here. So I created a new PR for this. Link below

https://github.com/MLH-Fellowship/lexical/pull/5/

PR #2385, opened 2022-06-09
fixed the names of plugins used in the docs

karam-qaoud:
The current plugins names used in the docs (under ```/react/plugins```) give the following error:

**_Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports._**

That's because the names of the plugins that are used in the ```import```s do not match the names that were given to the plugins when they were created and ```export```ed. For example:

**Incorrect**
```jsx
import {LexicalListPlugin} from '@lexical/react/LexicalListPlugin';
```
**correct**
```jsx
 import {ListPlugin} from '@lexical/react/LexicalListPlugin';
```

This is corrected by renaming the plugins names in the code to match the names that were given to the plugins upon exporting them.

### DIFF
--- a/packages/lexical-website-new/docs/react/plugins.md
+++ b/packages/lexical-website-new/docs/react/plugins.md
@@ -8,12 +8,12 @@ React-based plugins are using Lexical editor instance from `<LexicalComposer>` c
 
 ```jsx
 <LexicalComposer initialConfig={initialConfig}>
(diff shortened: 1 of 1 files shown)


trueadm (approved):
Thank you

PR #2387, opened 2022-06-10
Customizable DecoratorBlockNode via theme

zurfyx:
Decorator now has an additional config parameter like createDOM. Do we strictly need DOM like in createDOM? The answer is we don't, we can't potentially build a good abstraction on top that extract the styles. However, it's reality, it's annoying, usually these nodes are tied with the component itself, so for the user having to create an additional abstract to pass their own styles in a different way it's not convenient.

This way we can pass `theme.embedBlock` and the component will automatically be rendered correctly without any further modifications.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "clean": "node scripts/clean.js",
     "extract-codes": "node scripts/build.js --codes",
     "flow": "node ./scripts/check-flow-types.js",
-    "tsc-watch": "tsc -w",
     "collab": "cross-env HOST=localhost PORT=1234 npx y-websocket-server",
     "test-unit": "jest --selectProjects unit --testPathPattern",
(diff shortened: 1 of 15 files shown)


fantactuka:
Should it go as a breaking change?

zurfyx:
> Should it go as a breaking change?

Potentially, it's unfortunate we just released 0.3

fantactuka:
> Potentially, it's unfortunate we just released 0.3
Maybe change order of arguments or will it make API inconsistent with other methods?

trueadm:
It's a breaking change if it's changing the 1st argument, but not a breaking change if adding as 2nd argument. So let's swap the arguments around, problem solved!

thegreatercurve (approved):

Issue #2127, opened 2022-05-10
Feature: Developer Tools Web Extension


thegreatercurve:
- [x] Scaffold a new [cross-browser web extension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Build_a_cross_browser_extension) which works with Firefox, Chrome, Edge, and Safari.
- [x] Port the existing inline ASCI Lexical state tree representation to the extension.
- [ ] Create a more interactive representation of the Lexical state tree, akin to the Component view within [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en).
  - [x] Respective tree nodes should be collapsible or expandable.
  - [x] Respective tree nodes should highlight the relevant DOMNode on hover of the Lexical Node.
  - [x] Add slide out panel with all properties on the node.
  - [x] Add a view for the current selection.
  - [ ] Add the ability to time-travel between different editor states.
  - [ ] Add a log of all Lexical commands and their respective payloads.
  - [ ] View should be performant even for larger Lexical state trees.
  - [ ] View should have a suite of unit and E2E tests which test for basic Lexical states.
  - [ ] Add details of editor config.
- [ ] Build a CI pipeline within our Github Actions which handles publishing new extension versions to their respective web extension stores.
- [ ] Add the extension to each of our test browser environments, so we can better debug failing E2E tests.

thegreatercurve:
@noi5e Would you mind adding yourself as an assignee to this Issue? For some reason, I can't add you.

noi5e:
@thegreatercurve Huh, I don't see a button to add myself:
<img width="329" alt="Screen Shot 2022-06-13 at 9 37 34 AM" src="https://user-images.githubusercontent.com/4361605/173402319-8f803a1b-0392-4d11-9da1-21fd8701a8ea.png">

Is it true that only one person can be assigned to an issue at a time? It could be a question of access level too.

PR #2388, opened 2022-06-10
Create DevTools Browser Extension Scaffold

noi5e:
- Creates a scaffold for React Components-style panel within DevTools for Chrome, Firefox, and Edge.

Screen Shot 2022-06-10 at 2 57 45 PM

Currently this is completely unintegrated with Lexical's codebase and build flow.

To run locally:
1. cd `lexical/packages/lexical-devtools`
2. `npm run build`
3. add to Firefox/Chrome/Edge by loading the `build` folder or `manifest.json` as a temporary extension.
  - [example Firefox instructions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_first_WebExtension#trying_it_out)

---


fantactuka:
Could you record a video how to run it from chrome then in future we can use it as a reference for anyone who needs help with it?

packages/lexical-devtools/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "strict": true,
+    "forceConsistentCasingInFileNames": true,
+    "noFallthroughCasesInSwitch": true,
+    "module": "esnext",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "react-jsx"
+  },
```
thegreatercurve, line 22: This can probably just inherit from the base `tsconfig` to avoid having different TS configs per package.

```suggestion
  "compilerOptions": {
    "strict": true,
  },
  "extends": "../tsconfig.json" 
```

packages/lexical-devtools/public/devtools.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
+
+// eslint-disable-next-line no-undef
+chrome.devtools.panels.create('Lexical', 'logo192.png', 'index.html', () => {});
```
thegreatercurve, line 4: To remove the lint warning: 
```suggestion
chrome.devtools.panels.create('Lexical', 'logo192.png', 'index.html', () => {
    return;
});
```

thegreatercurve (approved):
Looks good! Thanks for working on this.

packages/lexical-devtools/public/devtools.js
noi5e: > To remove the lint warning:

@thegreatercurve Ah yeah, ESLint was actually signalling `chrome` as undefined here. It seems like `chrome` is defined in Firefox, Chrome, and Edge. Maybe `let chrome = chrome` would be a way to get around the linter without disabling it? Open to suggestions here.

packages/lexical-devtools/tsconfig.json
noi5e: > This can probably just inherit from the base tsconfig to avoid having different TS configs per package.

Sounds good, will commit

packages/lexical-devtools/public/devtools.js
fantactuka: Maybe similar to this: https://github.com/facebook/lexical/blob/c8b7add27d1435972a586a1c89130ba776748914/.eslintrc.js#L14

noi5e:
Hmm, the CI tests have been failing:
Screen Shot 2022-06-12 at 12 17 30 PM

Due to installing dependencies:
Screen Shot 2022-06-12 at 12 07 52 PM

I made this PR by creating a web extension scaffold **in a separate repo** [here](https://github.com/noi5e/lexical-dev-tools). Then I copied-and-pasted what I had into a new folder under `packages`. 

Maybe something about the config/dependencies from **that repo** is not playing nice with existing Lexical config.

I'm noticing that other Lexical `packages` have their own individual `package.json`. But it seems like they don't have a `package-lock.json` like this PR has.

This is kind of a shot in the dark, but I wonder if deleting the `package-lock.json` here will help this pass. Will play around with it.

noi5e:
> Could you record a video how to run it from chrome then in future we can use it as a reference for anyone who needs help with it? I'd probably even put it into readme/contribution section

@fantactuka In the next PR, I hope to fully integrate the package with Lexical's build process, with Vite etc. The goal in the next few days or so is to have the browser extension build and run automatically without any tinkering required. So hopefully having to run `npm run build` manually is just a temporary state of affairs 🤞.

I just included those instructions so maintainers could see what the scaffold looked like currently... But I made a video anyway!

https://user-images.githubusercontent.com/4361605/173254833-48af086c-abf5-40c0-b849-e7b09b283956.mov

> 1. cd packages/lexical-devtools
> 2. npm run build
> 3. add to Firefox/Chrome/Edge by loading the build folder or manifest.json as a temporary extension.

noi5e:
Screen Shot 2022-06-12 at 4 20 51 PM

Since I'm a first-time contributor, my commits don't trigger GitHub Actions... @thegreatercurve @fantactuka Can you run the workflow for me when you get a chance?

I deleted `package-lock.json` to see if that would help this pass the CI, let's see. 🙏 

I also added build instructions in `README.md`.

And changed `tsconfig.json` to inherit from base. I believe I can just delete `tsconfig.json`, but for some reason when I do that, the extension won't `build` properly. In the next PR, I'm going to try to build the package with Vite instead of Create React App, so I'll try again there.

noi5e:
Checks are passing now 🥳 :
Screen Shot 2022-06-12 at 5 33 07 PM

thegreatercurve:
@noi5e I'll merge this and you can create experiment with Vite in another separate PR.

PR #2394, opened 2022-06-12
Prevent dispatching redundant can undo/redo commands

fantactuka:
- Trigger CAN_UNDO/CAN_REDO only when its value changes (vs on each change causing toolbar rerendering)
- Trigger CAN_UNDO/CAN_REDO commands when clearing history

Misc:
- Add ts/tsx to precommit hooks

### DIFF
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -1,4 +1,5 @@
 module.exports = {
-  '*.(js|mjs|jsx|css|html|d.ts|js.flow)': 'prettier --write',
+  '*.(js|mjs|jsx|css|html|d.ts|js.flow|ts|tsx)': 'prettier --write',
   '*.(js|mjs|jsx)': ['flow focus-check', 'eslint --fix'],
+  '*.(ts|tsx)': ['eslint --fix'],
 };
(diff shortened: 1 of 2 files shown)


.lintstagedrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
-  '*.(js|mjs|jsx|css|html|d.ts|js.flow)': 'prettier --write',
+  '*.(js|mjs|jsx|css|html|d.ts|js.flow|ts|tsx)': 'prettier --write',
   '*.(js|mjs|jsx)': ['flow focus-check', 'eslint --fix'],
```
fantactuka, line 3: Do we even need these anymore? // @thegreatercurve

thegreatercurve, line 3:
```suggestion
  '*.(js|mjs|jsx|ts|tsx)': ['eslint --fix'],
```

thegreatercurve: You can remove the Flow check completely. We still use Eslint for the TS files so you can add that for the TS files:
```suggestion
  '*.(js|mjs|jsx|ts|tsx)': ['eslint --fix'],
```

thegreatercurve (approved):

PR #2395, opened 2022-06-12
Simplify clickable links checks

fantactuka:
For this plugin we want to let user selecting links with mouse but let clicking on links. It used to check pointer down & pointer up coordinates to detect selection, while it can be simplified to just checking if selection is collapsed.

### DIFF
--- a/packages/lexical-playground/src/plugins/ClickableLinkPlugin.ts
+++ b/packages/lexical-playground/src/plugins/ClickableLinkPlugin.ts
@@ -30,7 +30,6 @@ export default function ClickableLinkPlugin({
   const [editor] = useLexicalComposerContext();
   useEffect(() => {
     function onClick(e: Event) {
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/ClickableLinkPlugin.ts
```diff
@@ -23,26 +28,8 @@ export default function ClickableLinkPlugin({
...
-      hasMoved.current =
-        event.offsetX !== prevOffsetX || event.offsetY !== prevOffsetY;
-    }
-
     function onClick(e: Event) {
-      // Based on pointerdown/up we can check if cursor moved during click event,
-      // and ignore clicks with moves (to allow link text selection)
-      const hasMovedDuringClick = hasMoved.current;
-      hasMoved.current = false;
       // $FlowExpectedError[incompatible-cast] onClick handler will get MouseEvent, safe to cast
```
thegreatercurve, line 33:
```suggestion
```

thegreatercurve (approved):

PR #2402, opened 2022-06-13
fix: highlight in dropdown

### LINKED ISSUE #2401: Bug: blocktype highlight seems to be improper
How do we wish to highlight the selected blockType?

We currently seem to use a span of class active when selected, but this causes the UI to look improper on hovering over other block types as shown below

```
{blockType === 'paragraph' && <span className="active" />}
```

(Observe the background-color size)
<p>

</p>

Lexical version: Latest

## The expected behavior

Please let me know the expected behavior, did we miss adding a tick icon in the active span element or is this the desired behavior itself?

## Note
I would like to work on this!

im-adithya:
Closes #2401 

# Before
<p>

</p>

# After
Screenshot from 2022-06-14 00-24-50

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
@@ -623,6 +623,11 @@ function InsertEquationDialog({
   return <KatexEquationAlterer onConfirm={onEquationConfirm} />;
 }
 
+function dropDownActiveClass(active: boolean) {
+  if (active) return 'active dropdown-item-active';
+  else return '';
+}
+
 function BlockFormatDropDown({
   editor,
   blockType,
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
```diff
@@ -718,50 +718,80 @@ function BlockFormatDropDown({
       buttonIconClassName={'icon block-type ' + blockType}
       buttonLabel={blockTypeToBlockName[blockType]}
       buttonAriaLabel="Formatting options for text style">
-      <DropDownItem className="item" onClick={formatParagraph}>
-        <span className="icon paragraph" />
+      <DropDownItem
+        className={
+          'item ' +
+          (blockType === 'paragraph' ? 'active dropdown-item-active' : '')
```
fantactuka, line 724: Not a big deal, but maybe small helper to take current block type + expected one and output className?

fantactuka (approved):
LGTM! 🚢

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
im-adithya: Sure, I was thinking of it as well!

im-adithya: Done :D

PR #2407, opened 2022-06-13
Prevent orphan ListItems on clipboard from crashing the editor

### LINKED ISSUE #2405: Bug: (Copy + Paste) Safely handle orphan list items on the clipboard
Sometimes other rich text apps only include the list items on the clipboard without the parent list and this breaks `@lexical/list` since it expects all `ListItemNodes` to be within a `ListNode` parent. 

We need to catch this and safely create a parent `ListNode` that the `ListItemNode`s generated by `importDOM` are wrapped in.

Maybe something similar to `extractWithChild` where a node can define a strict parent relationship? Not sure.

cc: @acywatson

tylerjbainbridge:
Fixes #2405

> Sometimes other rich text apps only include the list items on the clipboard without the parent list and this breaks @lexical/list since it expects all ListItemNodes to be within a ListNode parent.

https://user-images.githubusercontent.com/13852400/173455405-c00e5b0a-8116-4a1d-b889-5f7c4c2236d1.mov

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -15,6 +15,7 @@ import type {
 } from 'lexical';
 
 import {$generateHtmlFromNodes, $generateNodesFromDOM} from '@lexical/html';
+import {$createListNode, $isListItemNode} from '@lexical/list';
 import {
   $cloneWithProperties,
   $sliceSelectedTextNodeContent,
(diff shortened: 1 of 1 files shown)


tylerjbainbridge:
cc: @hanford if you want to pull down locally and confirm the fix works as intended.

trueadm (approved):

hanford:
Seems to fix it on my end @tylerjbainbridge

PR #2415, opened 2022-06-15
Bundle Lexical DevTools with Vite

noi5e:
This does everything that PR #2388 did, it just creates a browser extension scaffold with Vite instead of Create React App/webpack.

Lots of file/folder structural changes, but not any functional difference for the user. Even the installation instructions in `README.md` are the same.

I'm still figuring out how to do the Vite config. Right now the assets are split into two folders:
1. `src` - React components
2. `public` - I made this folder as a workaround, in order to get Vite to include `manifest.json` (and everything it references) within the `build`. Following [Vite docs](https://vitejs.dev/guide/assets.html#the-public-directory).

It's pretty crucial that the browser extension `build` includes a `manifest.json`. It seems like Vite doesn't do this out of the box. 

An outside repo makes a [custom Vite plugin](https://github.com/JohnBra/vite-web-extension/blob/main/utils/plugins/make-manifest.ts) that will generate a `manifest.json`, and it looks relatively easy to replicate. I haven't done this yet, will save it for another PR.

---


noi5e:
I think this should be a simple fix, but I don't have the access to resolve the conflict: 
Screen Shot 2022-06-14 at 7 12 03 PM

thegreatercurve:
@noi5e You shouldn't need any kind of special access to resolve the merge conflict. Update `main` on your fork, and then rebase your branch on top of that before merging into this repo.

thegreatercurve (changes requested):
Set up looks fine, but how com we've got rid of TypeScript? That should work out of the box for Vite. See the `react-ts` example here: https://vitejs.dev/guide/#trying-vite-online

packages/lexical-devtools/index.html
```diff
@@ -0,0 +1,17 @@
...
+      rel="icon"
+      type="image/svg+xml"
+      href="/src/images/lexical-white.png" />
+    <!-- <link rel="manifest" href="/src/manifest.json" /> -->
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Lexical Developer Tools</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/main.jsx"></script>
```
thegreatercurve, line 15: You can import `.tsx` files directly and it will update them to `.js`.

```suggestion
    <script type="module" src="/src/main.tsx"></script>
```

packages/lexical-devtools/src/main.jsx
```diff
@@ -0,0 +1,12 @@
+import './index.css';
```
thegreatercurve, line 1: This file should be a `.tsx` file so we get the TypeScript checks.

packages/lexical-devtools/src/main.jsx
```diff
@@ -0,0 +1,12 @@
...
+
+import React from 'react';
+import ReactDOM from 'react-dom/client';
```
thegreatercurve, line 4: TypeScript forces modules which don't explicitly have a `default` export to be imported in this way below: 
```suggestion
import * as React from 'react';
import * as ReactDOM from 'react-dom/client';
```

More info: https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports

packages/lexical-devtools/src/main.jsx
thegreatercurve, line 1: You'll get a lint error that this file needs a header with the Meta licence. If you're in VSCode and right click on the error, and then click `fix`, it should just automatically add the header text it in for you.

thegreatercurve:
> @thegreatercurve Thanks for the assist, and helping this to pass the CI! Looks like the last few reviews you mentioned were covered by your commit as well.
> 
> This might be ready to merge, although I do have a slight question mark about deleting `tsconfig.json` and if this is now adequately covered by TypeScript from the root config (and not, let's say, from a global install in my environment?). I think it is? But not sure. Dropping a note in chat about this.

@noi5e your folder is included in the `includes` of the base `tsconfig`, so if running `tsc` in the root folder doesn't cause any errors, you should be all good!

PR #2421, opened 2022-06-15
[0.5] Remove INERT mode

zurfyx:
The reason why we created INERT nodes was solely to support the Autocomplete plugin, the hint had to have special selection rules that weren't supported by any of the existing modes.

With the a11y work done on DecoratorNodes late in 2021, INERT nodes are no longer a must have, we can potentially do the same with DecoratorNodes which reinforces a single way to do things, other than simplifying and reducing the size of the core.

This PR gets rids of INERT nodes completely.

At the same time, I'm leveraging this opportunity to simplify `getTextContent` and drop the `includeDirectionless` param as well. The directionless param was a gimmick to complement the customization of `getTextContent` but it no longer plays nicely with the cache rules we introduced. We could potentially optimize it but I think that, if there's no real life cases where it can come useful, it's better if we just kill it for now.


thegreatercurve (approved):

trueadm (approved):

thegreatercurve:
Closes #2730

trueadm:
Don’t we still want to remove inert nodes?

PR #2424, opened 2022-06-15
Add `eslint-plugin-header` to lint for correct MIT licence header

### LINKED ISSUE #1499: Feature:  ESLint rule for licensing
Write a custom ESLint rule to check for license at the top of files and, ideally, to add them via --fix if not present.

A Node script that does this is one alternative.

thegreatercurve:
Add `eslint-plugin-header` to ensure correct Meta licence header is being applied to all files.

Closes #1499

### DIFF
--- a/packages/lexical-dragon/LexicalDragon.ts
+++ b/packages/lexical-dragon/LexicalDragon.ts
@@ -1,0 +1,9 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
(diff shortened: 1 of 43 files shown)


packages/lexical-dragon/LexicalDragon.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+module.exports = require('./dist/LexicalDragon.js');
```
thegreatercurve, line 9: Not sure why this was never exported, but this should fix any issue.

zurfyx (approved):

PR #2427, opened 2022-06-15
fix: add styles on copy

im-adithya:
Currently we face an issue while pasting from Lexical <-> Lexical, the styles and formatting get copied as they should be (Thanks to #2370) But the problem is that the toolbar doesn't recognize the styles that are being pasted. You can see the issue in the video below:

# Before

https://user-images.githubusercontent.com/64399555/173834698-d79eef4f-771d-4585-abdc-bf9904f0a7cc.mp4

This is apparently because all the styles are stored in a Map which only gets set on applying a style, this PR adds the functionality to add styles to the Map even on paste.

# After

https://user-images.githubusercontent.com/64399555/173834724-424b655b-50e7-4e5a-b35d-b5520c62b55d.mp4

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -17,6 +17,7 @@ import type {
 import {$generateHtmlFromNodes, $generateNodesFromDOM} from '@lexical/html';
 import {$createListNode, $isListItemNode} from '@lexical/list';
 import {
+  $addNodeStyle,
   $cloneWithProperties,
   $sliceSelectedTextNodeContent,
 } from '@lexical/selection';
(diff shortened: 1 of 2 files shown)


im-adithya:
Welcome, @zurfyx!
I'm working on it next! #2370 has introduced the feature of copying the LexicalString (so as to make it easier while converting to nodes) to the clipboard rather than the HTMLString. So now that there's no crossing of the two, we can focus entirely on the copy-paste functionality for purely HTMLStrings, like the ones we get on copying from Google Docs separately.

tylerjbainbridge:
> Thank you, the code looks correct! Just one question, have you had a chance to play around with Microsoft Word or Google Docs? As far as I know, they carry some good CSS that we may have to drop at some point, mostly wondering about the correctness and compatibility across different editors.
> 
> cc @tylerjbainbridge since he's been looking into this area a lot recently

It should be noted that this doesn't effect bold, italic, etc, because they use the built in `format` flag on the `TextNode`, this only applies to the more vague `styles` property on `TextNodes`.

So I'm not super familiar with `TextNode` styles, but this does seems like a fine solution for the current implementation. However long term I'm definitely wary of the styles being stored as a string on the `TextNode` instead of storing a style JS object that gets translated to inline styles during `createDOM`. 

All this string parsing seems a bit fickle and risky to me.

Maybe @trueadm can shed some light on this design decision?

tylerjbainbridge (approved):

PR #2429, opened 2022-06-15
Fix exposed private methods

zurfyx:
Screen Shot 2022-06-15 at 3 20 54 pm

⚠️ While I tried not to make any breaking changes, there are still minor (but still interesting to bundle in because they're sort of fixes):
- LexicalNode
- LexicalEditor
- EditorState
- RangeSelection
- NodeSelection

All of the above will now only be exported as types, instead of interfaces or classes. These were always meant to be exported this way but originally were exported as classes because of Flow limitations.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -238,7 +238,7 @@ function $setPointValues(
   point.type = type;
 }
 
-interface BaseSelection {
+export interface BaseSelection {
   clone(): BaseSelection;
   dirty: boolean;
   extract(): Array<LexicalNode>;
--- a/packages/lexical/src/index.ts
+++ b/packages/lexical/src/index.ts
@@ -6,21 +6,50 @@
  *
  */
 
-export * from './LexicalCommands';
-export * from './LexicalEditor';
-export * from './LexicalEditorState';
-export * from './LexicalEvents';
-export * from './LexicalNode';
-export * from './LexicalSelection';
-export * from './LexicalUpdates';
-export * from './LexicalUtils';
-export * from './LexicalVersion';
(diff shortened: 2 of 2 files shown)


trueadm (approved):

thegreatercurve:
@zurfyx What methods do we want to exclude? If it's only the 7 below prefixed internal, can we still export everything from the other files? 

```
packages/lexical/dist/LexicalSelection.d.ts:
  135: export declare function internalMakeRangeSelection(anchorKey: NodeKey, anchorOffset: number, focusKey: NodeKey, focusOffset: number, anchorType: 'text' | 'element', focusType: 'text' | 'element'): RangeSelection;
  139: export declare function internalCreateSelection(editor: LexicalEditor): null | RangeSelection | NodeSelection | GridSelection;
  140: export declare function internalCreateRangeSelection(lastSelection: null | RangeSelection | NodeSelection | GridSelection, domSelection: Selection | null, editor: LexicalEditor): null | RangeSelection;

packages/lexical/dist/LexicalUpdates.d.ts:
  25: export declare function internalGetActiveEditor(): null | LexicalEditor;

packages/lexical/dist/LexicalUtils.d.ts:
  33: export declare function internalMarkNodeAsDirty(node: LexicalNode): void;
  34: export declare function internalMarkSiblingsAsDirty(node: LexicalNode): void;
  44: export declare function internalGetRoot(editorState: EditorState): RootNode;
```

trueadm:
> I think there's probably a greater potential for us to forget to re-export new public methods at the root. We do a release, and then get bug reports that certain functionality is missing.

We should be using said APIs in our unit tests or via plugins on the playground, so this shouldn't be an issue.

PR #2431, opened 2022-06-15
Remove redundant newlines on pasting html

fantactuka:
Before

https://user-images.githubusercontent.com/132642/173874728-2782a7d1-f1c0-4531-a966-1ea3e78011b9.mov

After

https://user-images.githubusercontent.com/132642/173874726-13f067fb-1f46-494c-b0a4-44342e8b710b.mov

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
@@ -1774,4 +1774,39 @@ test.describe('CopyAndPaste', () => {
       `,
     );
   });
+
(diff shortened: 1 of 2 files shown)


zurfyx (approved):
Nvm, test case is clear 👌

fantactuka:
> So you're saying that an empty text node with just `\n` is not correct?
We use `<br>` for newlines within non-pre nodes, so \n with spaces can be skipped. I'm not yet sure about spaces, whether we should trim it or replace repeating spaces at all. I can imagine copying "a  ...multiple spaces... b" should preserve all spaces within, so this diff only addresses newlines (original bug from #1988)

PR #2435, opened 2022-06-16
Fix `lexical-offset` type errors

thegreatercurve:
Not sure why this wasn't picked up in the CI, but we're getting type errors in `lexical-offset` which block the internal sync. In this PR, I'm just simplifying the types.

### DIFF
--- a/packages/lexical-offset/src/index.ts
+++ b/packages/lexical-offset/src/index.ts
@@ -387,23 +387,20 @@ function $searchForNodeWithOffset(
   return null;
 }
 
-function $createInternalOffsetNode<
-  TChild extends OffsetNode,
-  TParent extends OffsetElementNode = OffsetElementNode,
->(
-  child: null | TChild,
+function $createInternalOffsetNode(
+  child: null | OffsetNode,
   type: 'element' | 'text' | 'inline',
   start: number,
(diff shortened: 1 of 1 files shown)


packages/lexical-offset/src/index.ts
```diff
@@ -484,15 +481,18 @@ function $createOffsetNode(
...
+    end,
+    key,
+    parent,
+  );
 
-  offsetMap.set(key, offsetNode);
+  offsetMap.set(key, offsetNode as OffsetNode);
 
-  return offsetNode;
+  return offsetNode as OffsetNode;
```
zurfyx, line 495: Why do we need to cast this, shouldn't this function return an `OffsetNode` instead of an object?

zurfyx (approved):
Stamp to unblock the sync

packages/lexical-offset/src/index.ts
thegreatercurve: It should, but `OffsetNode` is a complex union of different object types. Returning `OffsetNode` creates a tonne of new type conflicts which we need to solve.

Long term, Flow let us get away with this, but if we enable strict mode in TS, we have to reckon with all these hacky types.

PR #2441, opened 2022-06-16
feat: Replace select with dropdown for font size and font family

### LINKED ISSUE #2333: Question: Is there any reason behind using select for font-family and size?
Would like to submit a PR modifying the select dropdowns to improve the UI if there isn't any!

im-adithya:
Closes #2333

<p>

</p>


im-adithya:
This also seems to _fix_ the _bug_ which we can observe when we highlight two different font-families (except fo Arial) like below:

Screenshot from 2022-06-16 20-50-29

[Courier Sans on left and Times New Roman on right but still toolbar shows Arial because of select's value setting to default on receiving "" string (So, it's completely fine on the back end of it.)]

But how do we wish to show that the two fonts are different? Google Docs shows a blank box as it shows its fonts in a defined box, is this fine to show it this way? Or should we leave space for the user to understand?

Screenshot from 2022-06-16 20-55-01

zurfyx:
Thanks, that looks awesome! Appreciate your multiple contributions to the project!

> Google Docs shows a blank box 

So that MS Word; I think you're right in that we probably want to show a blank box just like the other editors. Unfortunately, that may be somewhat expensive since we'll have to check the format of every single node. I wonder if we should cache or throttle or do some other optimization to remediate this..

im-adithya:
BUMP! @zurfyx

thegreatercurve (approved):
Thanks @im-adithya!

acywatson:
Can we fix this tests here?

im-adithya:
Done, must be because of the buggy PR that was reverted a few days ago

thegreatercurve:
> Done, must be because of the buggy PR that was reverted a few days ago

@im-adithya Buggy PR? This looks good to merge to me.

im-adithya:
> @im-adithya Buggy PR? This looks good to merge to me.

Not this, I was referring to this PR (#2675) which on rebase a week ago broke the tests here (Reverted in #2708)

It is ready now, though!

thegreatercurve:
Cool. Just merged! Thanks again @im-adithya

Issue #2398, opened 2022-06-13
Bug: Image node resizing bug in Safari


AaronGazzola:
When resizing an image node in the lexical playground while using Safari, the cursor changes to text instead of ew-resize. 
Also, when dragging the resize beyond the width of the editor, all the text following the image is selected.

Lexical version: 0.3.3

## Steps To Reproduce

1. Add an image node to the lexical playground
2. Resize the image, and drag the cursor beyond the width of the editor

Link to code example: 
Reproducible in playground

## The current behavior
In safari, the text cursor is shown while resizing an image, and all the text in the editor after the image is selected if the cursor is dragged beyond the editor width

## The expected behavior
The ew-resize cursor should be displayed while resizing, and text should not be selected while resizing.

PR #2444, opened 2022-06-16
fix(lexical-playground): image node resizing bug in safari

LuciNyan:
### Description
This is fixed by dynamically adding `-webkit-user-select: none` to body when resizing

https://user-images.githubusercontent.com/22126563/174114795-47c42850-3f63-4083-be46-a3018ba8f6b3.mov

#2398

### DIFF
--- a/packages/lexical-playground/src/ui/ImageResizer.tsx
+++ b/packages/lexical-playground/src/ui/ImageResizer.tsx
@@ -40,6 +40,10 @@ export default function ImageResizer({
   showCaption: boolean;
 }): JSX.Element {
   const buttonRef = useRef(null);
+  const userSelect = useRef({
+    priority: '',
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/ui/ImageResizer.tsx
```diff
@@ -98,6 +102,17 @@ export default function ImageResizer({
         `${cursorDir}-resize`,
         'important',
       );
+      userSelect.current.value = document.body.style.getPropertyValue(
```
zurfyx, line 105: You can potentially hoist userSelect but no big deal, the rest is likely to be more expensive

zurfyx (approved):

zurfyx:
Thank you Luci!

Issue #2450, opened 2022-06-16
Bug: Pasting text into paragraph with link and line breaks 


fantactuka:
## Steps To Reproduce

1. Type "Hello [link] world"
2. Select "He"
3. CMD+C, CMD+V
4. Content will be "Hello  world[link]" while it's expected to keep "Hello [link] world"

https://user-images.githubusercontent.com/132642/174150412-3f7ed291-b446-4720-9109-408d263dbc2b.mov

PR #2451, opened 2022-06-16
Adjust selection.insertNodes to handle inline elements (links)

fantactuka:
Not a real fix, code is a mess and need better understanding on why we re-insert siblings into current target. 

Details: #2450


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1404,14 +1404,21 @@ export class RangeSelection implements BaseSelection {
           lastChild.selectNext();
         }
       }
+      const $isBlockElementNode = (node) =>
```
trueadm, line 1407: This is a hot function, so it's best to move this function out so it's not created each time.

PR #2453, opened 2022-06-17
Add default value for undefined case in markdown transformers

imnoahcook:
Easy review

Context:
I was playing with the rich text playground and noticed that if you only used the `ELEMENT_TRANFORMERS` from `@lexical/markdown`, you will get a runtime error as it tries to iterate over the textFormat and textMatch fields, which are undefined in this case.

They're undefined because none of the element transformers have these fields in them.

I consider this a bug, because there shouldn't be any reason someone not to be able to use
> for quote
1. for list
etc..
without using the rest of the markdown exports

Solution: 
Add a default empty array value to `text-format` and `text-match`

### DIFF
--- a/packages/lexical-markdown/src/v2/utils.ts
+++ b/packages/lexical-markdown/src/v2/utils.ts
@@ -41,8 +41,8 @@ export function transformersByType(transformers: Array<Transformer>): Readonly<{
 
   return {
     element: byType.element as Array<ElementTransformer>,
-    textFormat: byType['text-format'] as Array<TextFormatTransformer>,
-    textMatch: byType['text-match'] as Array<TextMatchTransformer>,
+    textFormat: (byType['text-format'] ?? []) as Array<TextFormatTransformer>,
+    textMatch: (byType['text-match'] ?? []) as Array<TextMatchTransformer>,
   };
 }
 


imnoahcook:
Screen Shot 2022-06-16 at 11 27 47 PM
Screenshot for context on what happens when you try to add `ELEMENT_TRANSFORMERS` to the `MarkdownShortcutPlugin` on the `rich-text-playground` example

acywatson:
Just needs the CLA signed, when you have a chance.

acywatson:
Ah, I forgot, we don't use nullish coalescing in Lexical for some reason that I can't quite recall at the moment. Something. about not wanting to set up the Babel plugin to transpile it, I think. You can use a logical OR, instead.

packages/lexical-markdown/src/v2/utils.ts
```diff
@@ -41,8 +41,8 @@ export function transformersByType(transformers: Array<Transformer>): Readonly<{
 
   return {
     element: byType.element as Array<ElementTransformer>,
```
fantactuka, line 43: Somehow missed that before, but maybe it's good idea to add default for by element too?

fantactuka (approved):
Minor update, looks good otherwise!

acywatson:
It looks like the playground vercel build is failing because of a missing script that was added recently. Rebasing might fix that.

imnoahcook:
Resolved changed by @fantactuka and rebased to pass tests.

Issue #2446, opened 2022-06-16
[Discussion] Rename CharacterStylesPopupPlugin to FloatingToolbar?


zurfyx:
It always takes me 30 ages to find this plugin, I wonder if we should rename it

harish-sethuraman:
may be `FloatingFormatter`?

thegreatercurve:
We have other floating toolbars some maybe something more specific like `TextFormatFloatingToolbar`?

harish-sethuraman:
TextFormatFloatingToolbar -> `FloatingTextFormatToolbar` may be? It would be easy to grasp if floating was at the start?

thegreatercurve:
Closed by #2456

PR #2456, opened 2022-06-17
renamed character styles popup plugin

harish-sethuraman:
Addressing  
- https://github.com/facebook/lexical/issues/2446

I felt like `FloatingFormatter` seemed like a better name. Feel free to suggest other names if any. 😅


zurfyx:
Well, `FloatingToolbar` is the only one that returns proper results when Googling -> https://www.google.com/search?q=floating+toolbar

Also, appreciate you're taking this! If you want you make as well rename some similar wording inside this file like:
- useCharacterStylesPopup
- FloatingCharacterStylesEditor

harish-sethuraman:
`character-style-popup` should I rename styles too?

zurfyx:
> `character-style-popup` should I rename classnames and styles too?

That would be awesome. Also, @thegreatercurve just suggested `TextFormatFloatingToolbar`, I quite like that one, WDYT?

harish-sethuraman:
TextFormatFloatingToolbar -> `FloatingTextFormatToolbar` may be? It would be easy to grasp if floating was at the start?

zurfyx:
> TextFormatFloatingToolbar -> `FloatingTextFormatToolbar` may be? It would be easy to grasp if floating was at the start?

Prob won't make a different when you search via CMD + P but no strong opinions on this!

zurfyx (approved):
Thank you!

PR #2458, opened 2022-06-17
Migrate to TS strict mode 1/n

thegreatercurve:
Writing these as smaller PRs as some of these types fixes involve changing business logic.


zurfyx:
That's awesome, thank you for following up on this!

zurfyx (approved):

Issue #1631, opened 2022-04-07
Run unit tests against built bundles


acywatson:
Our last couple of releases have included issues related to bad imports. It's hard to catch these in dev since they don't fail the build in the local environment. One strategy we can use is running our unit tests against the built bundles, which should catch these problems before we release.

acywatson:
Fixed in #2460

PR #2460, opened 2022-06-17
CI check against prod bundle

zurfyx:
# Prod fires prevention

It happened once and just happened twice. We build and thoroughly test our dev bundle only to find that our prod bundle is completely flawed and can potentially **crash the entire site**.

The best way to mitigate this is by extending our test suite to test with our production bundle as well as dev. My hunch is that we don't need to test the entire matrix like we do for dev but we should at least test one browser + collab, this should already give us some good confidence in that the prod bundle is behaving just as well as the dev one.


thegreatercurve:
Closes #1631

zurfyx:
As per @trueadm's offline suggestion, I'm reducing the workload while still retaining the runs that give us the most value

thegreatercurve (approved):

PR #2462, opened 2022-06-17
Fix button--outline hover color dark mode

zakiego:
"Visit Playground" text color becomes invisible when using dark mode

https://user-images.githubusercontent.com/78015359/174300865-2e290ec4-88bb-4568-8879-72d99c6b5056.mp4

### DIFF
--- a/packages/lexical-website-new/src/css/custom.css
+++ b/packages/lexical-website-new/src/css/custom.css
@@ -40,3 +40,7 @@ html[data-theme='dark'] .docusaurus-highlight-code-line {
 .footer__links {
   margin-bottom: 3rem;
 }
+
+.button--outline:hover {
+  color: white;
+}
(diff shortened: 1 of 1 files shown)


packages/lexical-website-new/src/css/custom.css
```diff
@@ -40,3 +40,7 @@ html[data-theme='dark'] .docusaurus-highlight-code-line {
 .footer__links {
   margin-bottom: 3rem;
 }
+
+.button--outline:hover {
```
thegreatercurve, line 44: Can we do something like the below if this is only conditional on it being in dark mode? 

```suggestion
html[data-theme='dark'] .button--outline:hover {
```

thegreatercurve (changes requested):

thegreatercurve (approved):
Thanks!

PR #2465, opened 2022-06-17
fix(lexical): caret at wrong place when paste

### LINKED ISSUE #2459: Bug: Copy/Pasting formatted text from/to the editor places the caret at wrong place
Lexical version: 0.3.5

## Steps To Reproduce

https://codesandbox.io/s/lexical-rich-text-example-5tncvy

1. Enter into the first paragraph: AA
2. Enter into the second paragraph: BCB
3. Select C
4. Make it italic
5. Select BCB
6. CTRL+C -> copy
7. Place the Caret between A|A
8. CTRL+V -> paste

## The current behavior
Caret placed to: AB|CBA

## The expected behavior
Caret should ABCB|A

https://www.youtube.com/watch?v=bvkq1GlC0xI

LuciNyan:
### Description
Fix bug: Copy/Pasting formatted text from/to the editor places the caret at wrong place.

The reason for this bug appears to be that at some point the last element inserted is not equal to the last element

https://user-images.githubusercontent.com/22126563/174320921-b8ca7305-2fe9-41cc-abd6-b3270249e478.mov

closes #2459

### DIFF
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
@@ -2398,6 +2398,77 @@ describe('LexicalSelectionHelpers tests', () => {
           '<p dir="ltr"><span data-lexical-text="true">Existing text...foo</span></p>',
         );
       });
+
+      test('a paragraph with a child text and a child italic text and a child text', async () => {
+        const editor = createTestEditor();
+
+        const element = document.createElement('div');
+
+        editor.setRootElement(element);
+
+        await editor.update(() => {
+          const root = $getRoot();
+
+          const paragraph = $createParagraphNode();
+          const text = $createTextNode('AE');
+
+          paragraph.append(text);
+          root.append(paragraph);
+
+          setAnchorPoint({
+            key: text.getKey(),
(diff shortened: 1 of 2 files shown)


tylerjbainbridge (commented):
I think this makes sense to me, but this is a really important part of the codebase and want to make sure it's 100% right. 

I see that one of the `for` loops is going backwards, but this new logic only ever considers the `lastNode` as the last `child` which deviates from the previous behavior.

I guess we need to figure out why one loop goes backwards and if it's important for certain situations.

LuciNyan:
> I think this makes sense to me, but this is a really important part of the codebase and want to make sure it's 100% right.

I agree with you, I think maybe we need to ask some people who are familiar with this piece of code to take a look.

tylerjbainbridge:
cc: @trueadm if you have a chance to take a look at this PR. I know you're the one who is most familiar.

I'll take a look at the meantime and try to familiarize myself with the different cases.

fantactuka:
Would be great to have couple e2e's to prevent regression in future

LuciNyan:
> Would be great to have couple e2e's to prevent regression in future

Thanks for the reminder!

fantactuka (approved):
Looks good! 👍

PR #2466, opened 2022-06-17
add e2e tests for MaxLength plugin

im-adithya:
Follow up to #2296, as suggested by https://github.com/facebook/lexical/pull/2296#pullrequestreview-989515743
(as the issue is fixed by #2428)


zurfyx (approved):
Thank you! I believe that another interesting case would be pasting 💏 at the edge (like the lorem ipsum at L84) but that's a solid test already

PR #2472, opened 2022-06-19
Fix component/plugin import names in get started section

alesmenzel:
Fixes incorrect plugin names used in JSX to match imported names and comments out the theme ellipsis so the code is copyable into a test js runtime like Codesandbox.


fantactuka (approved):
Looks great, thank you!

trueadm (approved):
Thanks

PR #2467, opened 2022-06-17
Revert "add e2e tests for MaxLength plugin (#2466)"

zurfyx:
This reverts commit e5b574dcce6dd03470ff780fcc72b82619db8333.

PR #2478, opened 2022-06-20
chore: add e2e tests for maxlength plugin

im-adithya:
Follow up to #2467 which failed e2e

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
@@ -71,6 +71,7 @@ test.describe('MaxLength', () => {
   });
 
   test(`can restrict emojis on boundaries`, async ({page}) => {
+    await focusEditor(page);
     await pasteFromClipboard(page, {
       'text/plain': 'lorem ipsum dolor sit amet, consectetur adipiscing elit',
     });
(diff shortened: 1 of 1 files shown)


thegreatercurve (approved):
Thanks!

thegreatercurve:
Actually looks like there may be a bug with the plugin or the E2Es. Some of these E2Es are failing legitimately.

thegreatercurve (changes requested):
Actually looks like there may be a bug with the plugin or the E2Es. Some of these E2Es are failing legitimately.

fantactuka (approved):

fantactuka:
Somehow messed up with tabs and pressed approve in a wrong one 🤷‍♂️

im-adithya:
Fixed! I wasn't using focusEditor in collab mode which caused the error.
The tests are still failing though but because of another issue...
I don't understand why this is happening:
```
-   <span data-lexical-text="true">lorem ipsum dolor sit amet, c</span>
+   <span data-lexical-text="true">lorem ipsum dolor sit amet, c&nbsp;</span>
```
This same test was passing before I added the `focusEditor` line ([check here](https://github.com/facebook/lexical/runs/6972866135)) but is failing [above](https://github.com/facebook/lexical/runs/7005622566).

But I can't seem to reproduce this in my localhost (in firefox)... And this error doesn't seem to occur in mac.

Why is this happening?

thegreatercurve (approved):

acywatson:
Looks like we still have failing tests here

im-adithya:
@acywatson @thegreatercurve ready to merge!

PR #2482, opened 2022-06-20
Migrate to TS strict mode 3/n 

thegreatercurve:
Writing these as smaller PRs as some of these types fixes involve changing business logic. This PR reduces the number of TS errors from 129 to 0 and enables strict mode for the build config (i.e. excluding Playground).

To test TS strict mode, run `tsc -p tsconfig.build.json`. 

**This PR needs to be reviewed carefully as it touches core business logic.**

#2469

### DIFF
--- a/packages/lexical/src/LexicalMutations.ts
+++ b/packages/lexical/src/LexicalMutations.ts
@@ -44,7 +44,7 @@ export function getIsProcesssingMutations(): boolean {
   return isProcessingMutations;
 }
 
-function updateTimeStamp(event) {
+function updateTimeStamp(event: Event) {
   lastTextEntryTimeStamp = event.timeStamp;
 }
 
@@ -62,7 +62,8 @@ function isManagedLineBreak(
   return (
     // @ts-expect-error: internal field
     target.__lexicalLineBreak === dom ||
-    dom['__lexicalKey_' + editor._key] !== undefined
+    // @ts-ignore We intentionally add this to the Node.
+    dom[`__lexicalKey_${editor._key}`] !== undefined
   );
 }
 
@@ -90,7 +91,9 @@ function handleTextMutation(
   }
 
   const text = target.nodeValue;
-  $updateTextNodeFromDOMContent(node, text, anchorOffset, focusOffset, false);
+  if (text) {
+    $updateTextNodeFromDOMContent(node, text, anchorOffset, focusOffset, false);
+  }
 }
 
 function shouldUpdateTextNodeFromMutation(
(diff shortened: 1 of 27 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -819,9 +813,9 @@ export class TextNode extends LexicalNode {
   }
 }
 
-function convertSpanElement(domNode: HTMLSpanElement): DOMConversionOutput {
+function convertSpanElement(domNode: Node): DOMConversionOutput {
```
zurfyx, line 816: Shouldn't the conversion be done beforehand then? `convertSpanElement(whatever as HTMLSpanElement)`

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -836,7 +844,9 @@ function beginUpdate(
     }
   } catch (error) {
     // Report errors
-    editor._onError(error);
+    if (error instanceof Error) {
```
zurfyx, line 847: The error in the catch part isn't automatically an Error?

zurfyx (approved):

zurfyx:
Awesome stuff, looks like we're already TS strict with this one!

packages/lexical/src/nodes/LexicalTextNode.ts
thegreatercurve: Updated the DOMConversion types to handle this earlier.

thegreatercurve:
> Awesome stuff, looks like we're already TS strict with this one!

I'm going to fix the playground errors as well. There's a lot but they're easy fixes compared to these.

packages/lexical/src/LexicalMutations.ts
```diff
@@ -90,7 +91,9 @@ function handleTextMutation(
   }
 
   const text = target.nodeValue;
-  $updateTextNodeFromDOMContent(node, text, anchorOffset, focusOffset, false);
+  if (text) {
```
trueadm, line 94: `text !== undefined`, as `text` can be an empty string

packages/lexical/src/LexicalUtils.ts
```diff
@@ -516,13 +519,15 @@ export function $updateSelectedTextFromDOM(
...
       }
 
-      $updateTextNodeFromDOMContent(
-        node,
-        textContent,
-        anchorOffset,
-        focusOffset,
-        isCompositionEnd,
-      );
+      if (textContent) {
```
trueadm, line 522: textContent can be empty string here

PR #2490, opened 2022-06-22
Fix infinite recursion in insertText in RangeSelection

### LINKED ISSUE #2484: Bug: Inserting text in partially formatted link
Lexical version: latest - v0.3.0

## Steps To Reproduce

1. Run the playground locally
2. Create _(or reuse)_ a link node with some text, I used "Lexical website"
3. Format part of the `Link`-node contents, such as "Lexical **web**site"
4. Try to add some additional text in the middle of the link at the end of the formatted content
5. This now creates a StackOverflow-exception at `RangeSelection.insertText`

**Exception in playground:**
image

**Exception with `npm start` locally:** 
image

**Link to code example:**
reproduction

## The current behavior
Stack Overflow exception when inserting text after the formatted `TextNode` inside a `Link`-node

## The expected behavior
Good question. Either expanding the formatted `TextNode`, or the non-formatted one. No Stack Overflow would be a good start. But the most consistent behaviour would probably to insert content into the formatted text node, right?

PAkerstrand:
This PR fixes #2484 by adding additional guards in `RangeSelection#insertText` to ensure the proper branch is taken in order to avoid infinite recursion for situations where text is inserted into nodes where the `parentNode` has `canInsertTextBefore` and/or `canInsertTextAfter` return `false`.

**Test failures before the fix:**

https://user-images.githubusercontent.com/463105/175008972-848431e1-8701-45b0-9aaa-0381fdf7b2ff.mp4

https://user-images.githubusercontent.com/463105/175008977-6624611c-308f-40cf-af95-899b029eae4b.mp4

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
@@ -741,6 +741,189 @@ test.describe('Links', () => {
     });
   });
 
+  test(`It can insert text inside a link after a formatted text node`, async ({
+    page,
+  }) => {
+    await focusEditor(page);
+    const linkText = 'This is the bold link';
+    await page.keyboard.type(linkText);
+
+    // Select all characters
+    await selectCharacters(page, 'left', linkText.length);
+
+    // Make it a link
+    await click(page, '.link');
+
+    await assertHTML(
+      page,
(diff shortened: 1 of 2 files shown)


fantactuka (approved):
👍

PR #2493, opened 2022-06-22
Add mouse over and mouse leave command

aquarius-wing:
Add two command support, and a example to use.
The example maybe there is some code to be optimized.
If there is some thing I need to change the code, please let me know.

### DIFF
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -53,6 +53,7 @@ import TwitterPlugin from './plugins/TwitterPlugin';
 import YouTubePlugin from './plugins/YouTubePlugin';
 import ContentEditable from './ui/ContentEditable';
 import Placeholder from './ui/Placeholder';
+import ShowKeyPlugin from "./plugins/ShowKeyPlugin";
 
 const skipCollaborationInit =
   // @ts-ignore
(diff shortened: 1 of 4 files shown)


thegreatercurve:
Thanks for contributing @aquarius-wing! 

I don't think we should include this in the playground, just because it's a fairly rare use case. Plus, the playground is has to work on mobile, which doesn't support mouse events. 

To achieve what you want, a simpler solution involves omiting the whole command system altogether and just using `$getNearestNodeFromDOMNode` in an event listener on the root element. Something like the below:

```js
rootElement.addEventListener(evemt => {
    editor.read(() => {
        $getNearestNodeFromDOMNode(event)
     })
})
```

aquarius-wing:
Thanks for reply, so if I delete the code in playground, my pull request can be accepted?

aquarius-wing:
Emmm, actually, I use these commands to develop a block tool, or row tool you can call it.
Just like notions.so website, so I must need to know which element node the user mouse over, and then show the block tool to help people change the style of element node. Finnally when user's mouse went out, I will hide the block tool.

aquarius-wing:
> > Emmm, actually, I use these commands to develop a block tool, or row tool you can call it.
> > Just like notions.so website, so I must need to know which element node the user mouse over, and then show the block tool to help people change the style of element node. Finnally when user's mouse went out, I will hide the block tool.
> 
> Got it - it is possible to just do that the way John suggested? You can listen for the events you're interested in an event listener on the root element, then find the nearest LexicalNode using the utility function he provided?

Thank you! I know how to develop my block tool

PR #2497, opened 2022-06-22
chore(lexical-playground): remove redundant code

LuciNyan:
### Description
I'm sorry. I made a mistake in this [PR](https://github.com/facebook/lexical/pull/2027). This is a redundant piece of code. I don't think we need it at all, because there is `rangeSelection.applyDOMRange(range);` to set the selection later.

We can remove it, and it won't make any difference.


thegreatercurve (approved):
Thanks!

PR #2502, opened 2022-06-23
fix(lexical): "selection.format" is not set correctly

### LINKED ISSUE #2498: Bug: When removing an italic/underline/strikethrough style from multiple paragraphs, the style change is visually applied but not reflected in the selection state
When removing an italic/underline/strikethrough style from multiple paragraphs, the style change is applied (the text visibly loses its italic/underline/strikethrough format as appropriate) but not reflected in the selection state (thus the toolbar does not untoggle the italic/underline/strikethrough icon).

This only seems to happen when the selection spans multiple paragraphs, and affects only italic, underline, and strikethrough styles. It does not seem to affect bold/subscript/superscript/link/code styles.

Lexical version: 0.3.5

## Steps To Reproduce

1. Create two paragraphs of text. Remove all styling on them.
2. Select all and toggle to bold.
3. Toggle to one of: italic/underline/strikethrough.
4. Untoggle the italic/underline/strikethrough style.
5. Untoggle the bold style.

Link to code example:

_(N/A: reproduced via manual interaction)_

## The current behavior

_Note: After each action in these steps, before taking the screenshot, I unfocus the text field as otherwise the popup styling menu gets in the way of seeing the results. Fortunately, the Selection state is unaffected by unfocusing the text field._

1) Create two paragraphs of text. Remove all styling on them:

image

✅ Correctly represents starting state.

2) Select all and toggle to bold:

image

✅ All text becomes bold, and the bold icon in the toolbar becomes active.

3) Additionally toggle to italic:

image

✅ All text becomes both bold and italic, and both the bold icon and the italic icon in the toolbar become active.

4) Untoggle the italic style:

image

❌ Although the text state is visually correct, the toolbar italic icon remains highlighted (incorrectly) along with the bold icon (correctly). Note how the selection state incorrectly shows `bold, italic` rather than just `bold`.

ℹ️ Of note, if we select all again at this step, the state corrects itself.

5) Untoggle the bold style:

image

✅ Both the text and selection state are correct, showing that the issue does not extend to the bold style.

## The expected behavior

1-3) As above.
4) The selection state should become just `bold` and the toolbar should correspondingly show only the bold icon as active.
5) As above.

LuciNyan:
### Description
It seems that the problem is mainly related to the `selectionchange` event not being fired correctly. The update of the button style on the toolbar depends on the selectionchange event. However, when focusNode, focusOffset, anchorNode, and anchorOffset are exactly the same and selection has a format applied, the event is not fired correctly.

https://user-images.githubusercontent.com/22126563/175352875-77839c7a-06c6-450d-8dd3-d6eac56c9fb7.mov

closes #2498

### DIFF
--- a/packages/lexical-markdown/src/v2/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/v2/MarkdownShortcuts.ts
@@ -242,6 +242,7 @@ function runTextFormatTransformers(
       openNodeText.slice(0, openTagStartIndex) +
         openNodeText.slice(openTagStartIndex + tagLength),
     );
+    const selection = $getSelection();
     const nextSelection = $createRangeSelection();
     $setSelection(nextSelection);
     // Adjust offset based on deleted chars
@@ -269,6 +270,10 @@ function runTextFormatTransformers(
       if (nextSelection.hasFormat(format)) {
         nextSelection.toggleFormat(format);
       }
+    }
+
+    if ($isRangeSelection(selection)) {
+      nextSelection.format = selection.format;
     }
 
     return true;
--- a/packages/lexical-playground/__tests__/e2e/TextFormatting.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/TextFormatting.spec.mjs
@@ -20,6 +20,8 @@ import {
   assertHTML,
   assertSelection,
   click,
+  evaluate,
+  expect,
   focusEditor,
   html,
   initialize,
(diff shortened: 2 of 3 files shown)


LuciNyan:
It seems to be some CI issues.

image

fantactuka:
Wonder if we can get away without selection change event by setting combined `anchor | focus` format in case of multi-node selection by adding

```js
this.format = firstNextFormat | lastNextFormat;
```
after first/last formats are calculated here:
https://github.com/facebook/lexical/blob/b70d217b28aa5997d6e1fdbde445aed5c83ab1e8/packages/lexical/src/LexicalSelection.ts#L1123-L1146


similar to what we do on selection change for multi nodes:
https://github.com/facebook/lexical/blob/b70d217b28aa5997d6e1fdbde445aed5c83ab1e8/packages/lexical/src/LexicalEvents.ts#L260-L267

LuciNyan:
> Wonder if we can get away from extra selection change event by setting combined `anchor | focus` format in case of multi-node selection by adding

Yes! this solution is really great.  Thank you! @fantactuka I've learned a lot from you. <3

fantactuka (approved):

LuciNyan:
Hello @fantactuka! 
I'm a little confused, when we apply new logic, it seems to change the behavior of some parts of markdown. Like below, when I type "\~\~\_\*\*hello world**_~~!" we get different results. I haven't done any markdown development before, so I'd like to know which result is the expected one.

New:
image

Old:
image

fantactuka:
Old version is correct as for some reason after markdown is being processed selection's format is not set correctly (it should be 0 (no formatting) as all markdown style tags are closed when typing "!". I'll take a look into it

PR #2504, opened 2022-06-24
Fix $generateHtmlFromNodes behavior if selection is null

yicrotkd:
Nothing is generated if selection is null. This PR fixes $generateHtmlFromNodes to output whole editor contents if selection is null.

https://lexical.dev/docs/concepts/serialization

> When generating HTML from an editor you can pass in a selection object to narrow it down to a certain section or pass in null to convert the whole editor.

I think it might be more natural to generate it even when the selection is undefined.

### DIFF
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -67,7 +67,7 @@ export function $generateHtmlFromNodes(
 
   for (let i = 0; i < topLevelChildren.length; i++) {
     const topLevelNode = topLevelChildren[i];
-    if (selection) {
+    if (selection !== undefined) {
       $appendNodesToHTML(editor, selection, topLevelNode, container);
     }
   }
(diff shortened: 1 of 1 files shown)


packages/lexical-html/src/index.ts
```diff
@@ -67,7 +67,7 @@ export function $generateHtmlFromNodes(
 
   for (let i = 0; i < topLevelChildren.length; i++) {
     const topLevelNode = topLevelChildren[i];
-    if (selection) {
+    if (selection !== undefined) {
```
zurfyx, line 70: Remove? It can never be undefined anyway
```suggestion
```

zurfyx (approved):
Thanks! Exporting all nodes when selection is null looks reasonable to me

packages/lexical-html/src/index.ts
yicrotkd: Thanks for the review!

`selection` is an optional parameter and can be undefined.

https://github.com/facebook/lexical/pull/2504/files#diff-c57dd373e8904083f1fe1b2a900f1a8b7e7e6a8e5f737891e4642d87083e9858L56

The code that didn't pass selection before will break, but is it OK to make it a required parameter?

tylerjbainbridge (approved):

yicrotkd:
I rebased my branch to the latest main branch. (Hoping CI passes)

PR #2506, opened 2022-06-24
fix(lexical-playground): Floating toolbar displayed on composition

### LINKED ISSUE #2344: Bug: Double-byte character conversion brings up the toolbar on macOS
Lexical version: 0.3.1（Latest）

## Steps To Reproduce

1. Occurs on macOS (not on Windows10 or 11）
2. Use Japanese IME (Mac default or Google IME)
3. In lexical playground, type Double-byte characters and convert it
4. Toolbar appears

## The current behavior

https://user-images.githubusercontent.com/77480273/172114418-57282b9c-3e63-4fde-b176-aeba4ce7c11f.mov

## The expected behavior

Works the same as windows

https://user-images.githubusercontent.com/77480273/172114440-17276237-a47e-40cc-a8c1-dd41cb1087c4.mov

LuciNyan:
### Description
Float toolbar should not be displayed during composition.

https://user-images.githubusercontent.com/22126563/175507437-674a9c3a-2eba-4498-8b0c-2dbd876e717f.mov

closes #2344

### DIFF
--- a/packages/lexical-playground/src/plugins/TextFormatFloatingToolbarPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/TextFormatFloatingToolbarPlugin.tsx
@@ -270,9 +270,8 @@ function useTextFormatFloatingToolbar(
 
   const updatePopup = useCallback(() => {
     editor.getEditorState().read(() => {
-      const isComposition = editor._compositionKey !== null;
       // Should not to pop up the floating toolbar when using IME input
-      if (isComposition) {
+      if (editor.isComposing()) {
         return;
       }
       const selection = $getSelection();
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/plugins/TextFormatFloatingToolbarPlugin.tsx
```diff
@@ -270,6 +270,11 @@ function useTextFormatFloatingToolbar(
 
   const updatePopup = useCallback(() => {
     editor.getEditorState().read(() => {
+      const isComposition = editor._compositionKey !== null;
```
trueadm, line 273: This is a private property. Opt to use editor.isComposing() instead :)

thegreatercurve (approved):

PR #2508, opened 2022-06-24
fix: maxlength bug in code node

im-adithya:
Follow up to #2428

The MaxLength emoji bug still persists in codeNode because while trimming content, it retrieves the text content of the previous node which is done only when this check passes: [packages/lexical-selection/src/index.ts#L984](https://github.com/facebook/lexical/blob/main/packages/lexical-selection/src/index.ts#L984)

In codeNode, the code highlight node replaces the text node but fails to match __type==="text" above, causing the issue.

### DIFF
--- a/packages/lexical-selection/src/index.ts
+++ b/packages/lexical-selection/src/index.ts
@@ -18,6 +18,7 @@ import type {
   TextNode,
 } from 'lexical';
 
+import {$isCodeHighlightNode} from '@lexical/code';
 import {
   $createTextNode,
   $getDecoratorNode,
@@ -981,7 +982,10 @@ export function trimTextContentFromAnchor(
         .getEditorState()
         .read(() => {
           const prevNode = $getNodeByKey(key);
-          if ($isTextNode(prevNode) && prevNode.isSimpleText()) {
+          if (
+            ($isTextNode(prevNode) && prevNode.isSimpleText()) ||
+            $isCodeHighlightNode(prevNode)
+          ) {
             return prevNode.getTextContent();
           }
           return null;


packages/lexical-selection/src/index.ts
```diff
@@ -981,7 +982,10 @@ export function trimTextContentFromAnchor(
         .getEditorState()
         .read(() => {
           const prevNode = $getNodeByKey(key);
-          if ($isTextNode(prevNode) && prevNode.isSimpleText()) {
+          if (
+            ($isTextNode(prevNode) && prevNode.isSimpleText()) ||
+            $isCodeHighlightNode(prevNode)
```
fantactuka, line 987: I don't think we should add code package as a dependency to selection, it will also fail for any other custom subclass of text node

acywatson: Yea, we need to find a more general heuristic for this. I guess CodeHighlight probably shouldn't return true from isSimpleText - not sure how/where else that is used. I also don't necessarily want to introduce a different heuristic.

im-adithya: > I don't think we should add code package as a dependency to selection, it will also fail for any other custom subclass of text node

Oops! Didn't see the big picture, you're right hashtag node (and all others extending textnode) fail(s) too!

im-adithya: > guess CodeHighlight probably shouldn't return true from isSimpleText

```
isSimpleText(): boolean {
    return this.__type === 'text' && this.__mode === 0;
}
```
It's a simple check to ensure the type is 'text' and mode is set to 0, but in codeHighlightNode, the type is 'code' so it fails anyway. I never came across mode yet, what is it used for?

If that's something enough to check, we can just call `getMode()` and check if it's `normal`...

im-adithya: BUMP! @acywatson @fantactuka

thegreatercurve: The `mode` is whether or not the whole word, individual character, or whole phrase can be deleted on a single backspace. We use this mainly with the mention nodes.

thegreatercurve:
Closing this as I don't think there's a way around the `lexical/code` cross package dependency, and that's something we really want to avoid introducing.

PR #2510, opened 2022-06-24
Fetch Lexical State Tree in DevTools App

noi5e:
Screen Shot 2022-06-24 at 11 37 52 AM

# Features
- Injects a script via [devtools.inspectedWindow.eval()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/inspectedWindow/eval) onto the page to `registerUpdateListener` and listen to `editorState` changes
- `editorState` changes get passed up to the browser extension [background script](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background)
- The background script messages the React app in the DevTools panel, which updates the UI via hooks
- Tested in Firefox, Chrome, and Edge
- **Other Feature:** popup window for browser icon:
Screen Shot 2022-06-24 at 12 12 37 PM

## Note on the App Flow
Currently the `editorState` updates get passed into the browser extension this way:
  1. User loads the extension, `content.js` sets up a [messaging connection](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onConnect) with `background.js`
  3. User opens the browser's DevTools console, `devtools.js` injects a `registerUpdateListener` onto the page
  4. When the `editorState` changes, the page does a `window.postMessage`
  5. `content.js` reads the `postMessage` and passes the `editorState` up to `background.js`
  6. If the Lexical DevTools panel has been opened, `background.js` sends a message to `App.tsx`, where the UI is updated.

The messaging is complicated right now for a few reasons. The main one is that I'm trying to not put too many things into `devtools.inspectedWindow.eval()` for security reasons. In the future, I think the messaging could be made a lot simpler and handled completely by `devtools.js`, without needing a content or background script. For example: 
  1. `devtools.js` sets up a messaging connection between the inspected window, the Lexical DevTools React app, and itself.
  2. `devtools.js` injects a `registerUpdateListener` onto the page AND a listener for `window.postMessage`
  3. `devtools.js` messages the Lexical DevTools React app itself

---


packages/lexical-devtools/public/background.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ */
+'use strict';
+
+// Create messaging connection to send editorState updates to Lexical DevTools App.
+let ports = {}; // Each tab will have a separate messaging port for the devTools app & the inspectedWindow's content script, eg. { tabId: { reactPort, contentScriptPort } }
+
+// The Lexical DevTools React UI sends a message to initialize the port.
+// eslint-disable-next-line no-undef
+chrome.runtime.onConnect.addListener(function (port) {
```
noi5e, line 15: I'm still trying to find a way to not disable the linter here. `chrome` is defined within the browser extension of course. 

I experimented a bit with making a custom `.eslintrc` like someone suggested, but I didn't have much luck. Pointers welcome!

packages/lexical-devtools/src/App.tsx
```diff
@@ -9,15 +9,41 @@
...
 
 import * as React from 'react';
+import {useEffect, useRef, useState} from 'react';
 
 import logo from './images/lexical-white.png';
 
 function App(): JSX.Element {
+  const [count, setCount] = useState<number>(0);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [editorState, setEditorState] = useState<object>({});
```
noi5e, line 19: Temporary linter disable here, because the app isn't technically doing anything with `editorState` yet. I will remove it in the next PR.

packages/lexical-devtools/src/App.tsx
```diff
@@ -9,15 +9,41 @@
...
+
+  useEffect(() => {
+    // create and initialize the messaging port to receive editorState updates
+    port.current = chrome.runtime.connect();
+    port.current.postMessage({
+      name: 'init',
+      tabId: chrome.devtools.inspectedWindow.tabId,
+      type: 'FROM_APP',
+    });
+  }, []);
```
noi5e, line 30: TBH, I could use some guidance in best practices for making these hooks, there's a lot of knowledge gaps I have here.

noi5e:
Screen Shot 2022-06-24 at 12 24 04 PM

Okay, I guess now's a good time to start learning TypeScript. 📖

packages/lexical-devtools/src/App.tsx
thegreatercurve: This hook seems to be correct, apart from updating the dependency array to include the ref. React should provide you with a linting error telling you what should be included in the dependency array.

This is a useful guide on how to use effect hooks: https://blog.logrocket.com/guide-to-react-useeffect-hook

thegreatercurve (changes requested):
This looks good to me. My only concern is that we're using JS in most of these files when we should instead be using TS.

You will need to add an `input` and `output` to your Vite config [like in this example](https://github.com/JohnBra/vite-web-extension/blob/main/vite.config.ts#L25-L37) to transform the TS into JS.

packages/lexical-devtools/public/background.js
```diff
@@ -0,0 +1,33 @@
...
+  port.onMessage.addListener(function (message) {
```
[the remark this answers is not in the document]
noi5e: Done!

packages/lexical-devtools/public/devtools.js
```diff
@@ -8,5 +8,30 @@
...
+    editor.registerUpdateListener(({editorState}) => {
+      window.postMessage(
+        {
+          editorState,
+          type: 'FROM_PAGE',
+        },
+        '*',
+      );
+    });
+  `,
```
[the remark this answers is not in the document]
noi5e: > This logic can be outside the eval so it lints correctly, and then you can call it via a function call in a string literal i.e. .eval(${registerAllUpdateListeners()}``

Gave this a shot, didn't quite run as expected, although I might have the syntax wrong:
Screen Shot 2022-06-27 at 10 45 27 PM

Seems to be a scoping thing maybe?

Screen Shot 2022-06-27 at 10 45 37 PM

packages/lexical-devtools/src/App.tsx
```diff
@@ -9,15 +9,41 @@
...
 
 function App(): JSX.Element {
+  const [count, setCount] = useState<number>(0);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [editorState, setEditorState] = useState<object>({});
+  const port = useRef();
+
+  useEffect(() => {
+    // create and initialize the messaging port to receive editorState updates
+    port.current = chrome.runtime.connect();
```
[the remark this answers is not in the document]
noi5e: Done!

packages/lexical-devtools/src/App.tsx
noi5e: > This hook seems to be correct, apart from updating the dependency array to include the ref. React should provide you with a linting error telling you what should be included in the dependency array.

Added the port object here in the latest commits. 

This is the current state of the linter errors:
Screen Shot 2022-06-27 at 11 21 12 PM

I think I need to define a `Port` type and describe `onMessage` and `postMessage` as its methods. Is that right? Looks like this might be the main barrier to getting this merged. Will have to study up on this.

packages/lexical-devtools/public/background.js
noi5e: Addressed here, yet to be implemented: https://github.com/facebook/lexical/pull/2510#discussion_r907140259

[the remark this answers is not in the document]
noi5e: > This looks good to me. My only concern is that we're using JS in most of these files when we should instead be using TS.

> You will need to add an input and output to your Vite config [like in this example](https://github.com/JohnBra/vite-web-extension/blob/main/vite.config.ts#L25-L37) to transform the TS into JS.

Nice, this weekend I was working on exactly that in a separate branch. I just opened up a separate PR for this change at #2540.

> This should be something like the below. You can then add the chrome windows types using npm i --save-dev @types/chrome

packages/lexical-devtools/src/App.tsx
noi5e: Since @thegreatercurve's will be away for a bit, curious if @fantactuka or @acywatson have a suggestion here. 

I'm starting from scratch with TypeScript and a little guidance could help since this is the main thing blocking this PR right now:

> I think I need to define a Port type and describe onMessage and postMessage as its methods. Is that right? Looks like this might be the main barrier to getting this merged. Will have to study up on this.

thegreatercurve: @noi5e Use `CMD+Click` to click on `connect()` to see what return type is being coming back from `port.current`. It looks like an empty object, hence why you're getting the errors.

Ideally, the return type from port.current should look like the below: 
```ts
type CurrentPort = {
  postMessage: (
    message: {
       name: string; 
       tabId: string;
       type: string 
    }
  ) => void;
  onMessage: {
    addListener: (message: {editorState: React.SetStateAction<object>}) => void
  }
}
```

For the moment though, these are quite advanced types to try to define yourself, especially if you're not experience with TS, so "cast" the types to `any` to remove the type checks, disable the lint warning, and one of us can then do a fix PR on top to add the types. 

This would look like:
```ts
(port.current as any).postWindow(
(port.current as any).onMessage.
```

packages/lexical-devtools/public/background.js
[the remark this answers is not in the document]
acywatson: I think we can keep separate PRs, if that's easier for you to manage.

packages/lexical-devtools/public/content.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+'use strict';
+
+// eslint-disable-next-line no-undef
+let port = chrome.runtime.connect();
+
+port.postMessage({
+  name: 'init',
+  type: 'FROM_CONTENT',
```
acywatson, line 15: can these be constants somewhere in a different shared module, maybe?

noi5e:
Okay, I think I've taken care of all the TypeScript errors in DevTools (or disabled the linter so they can be taken care of in #2540).

This still isn't passing CI for some reason. The failures look like they don't have anything to do with my code: 
Screen Shot 2022-06-30 at 1 02 17 PM
Screen Shot 2022-06-30 at 1 02 42 PM

I did install `@types/chrome` in the last commit. Maybe there's a chance that did something weird to the dependency tree?

packages/lexical-devtools/public/content.js
noi5e: > can these be constants somewhere in a different shared module, maybe?

Addressing this and moving the discussion [here](https://github.com/facebook/lexical/pull/2540#issuecomment-1174292738)

noi5e:
Screen Shot 2022-07-04 at 1 37 09 PM

Just one failing test now, seems like it doesn't have anything to do with my changes.

Once this passes CI, this will be ready to merge.

Just to keep tabs on the main suggestions from code review:
1. `chrome.runtime.onConnect.addListener` - This needs `@types/chrome` to be installed, which is done in #2540.
2. `port.current.onMessage.addListener` - Still don't know enough TypeScript to handle this, @thegreatercurve suggesting changing this type to `any`, and then having staff write the type properly.
3. ES6-style imports and modules in `content.js` and `background.js` - handled in #2540

noi5e:
I think this is ready to merge! @fantactuka @thegreatercurve @acywatson

PR #2511, opened 2022-06-24
Hide placeholder when readonly

zurfyx:


packages/lexical-react/src/shared/useCanShowPlaceholder.ts
```diff
@@ -9,29 +9,40 @@
...
-    });
+    }
+    update();
+    return mergeRegister(
+      editor.registerUpdateListener(() => {
+        update();
+      }),
+      editor.registerReadOnlyListener(() => {
+        update();
+      }),
```
fantactuka, line 44: No need for closures?

fantactuka (approved):

acywatson:
Looks like the new test is failing

packages/lexical-react/src/shared/useCanShowPlaceholder.ts
zurfyx: Leveraging it's an always a listener to rely on `editor.getEditorState()`. But I'll rename `update` because it's misleading

PR #2513, opened 2022-06-25
Fix Windows CI Runs

acywatson:
This fixes the Windows CI runs, which are suddenly failing apparently because of the colon in the command names in the npm scripts. 

Simple repo to repro the issue: https://github.com/acywatson/glowing-succotash/runs/7063676429?check_suite_focus=true

I spent a long time debugging this, and I still can't understand why this started happening. I opened some topics on the GitHub Actions support forum which are pending approval. Hopefully we will get some answers there. Until then, this does 2 things:

1) changes the node version to be compatible with the ranges supported by some dependencies (previously there were a bunch of warning about this in the test output).

2) removes all the colons and replaces them with hyphens in the npm scripts. Ultimately, we should refactor all these scripts to use a single node script or something, but for now this will unblock us.


fantactuka (approved):

tylerjbainbridge:
Great catch!

PR #2515, opened 2022-06-25
Fixed getStyleObjectFromRawCSS function to work with unformatted css

karam-qaoud:
…trings


zurfyx (approved):
Thanks Karam! Can we write a unit test for this please? Since this method is not exposed it may be better to test `$getSelectionStyleValueForProperty` instead that eventually calls this

PR #2518, opened 2022-06-26
fix: resize on mobile

LuciNyan:
### Description
Fix resize on mobile


zurfyx (commented):
Thanks! Can we use a new wrapper over these controls instead? CSS classes can potentially be minified and this code tends to be copy-pasted

I wonder if we need the mobile condition too, it would be cheaper without it and I don't think the added performance cost for desktop is noticeable

LuciNyan:
@zurfyx Thank you for your suggestion! It's really great! <3

packages/lexical-playground/src/ui/ImageResizer.tsx
```diff
@@ -151,6 +154,7 @@ export default function ImageResizer({
       setStartCursor(direction);
       onResizeStart();
 
+      controlWrapper.classList.add('image-control-wrapper--resizing');
```
zurfyx, line 157: Let's just hardcode the style here, otherwise we will face the same problem because the classname is not variable

zurfyx (approved):
Thanks, minor comment, otherwise looking good!

packages/lexical-playground/src/ui/ImageResizer.tsx
LuciNyan: I hope I'm not bothering you. I'd really like to understand "CSS classes can potentially be minified and this code tends to be copy-pasted". If that's convenient, I wonder if you could make it more easier to understand. Thank you very, very much!

PR #2534, opened 2022-06-27
Lexical Typeaheads

### LINKED ISSUE #2114: Feature: slash command
It would be great to have a slash command ready to be used (in core or as an example), see the notion like experience.

Untitled4

tylerjbainbridge:
This PR adds the base `LexicalTypeaheadMenuPlugin ` and a `ComponentPickerPlugin ` example + new `MentionsPlugin` for the Playground that utilize it.

As you can see from the video below, this is a powerful design that allows for dynamic menu options to be populated from the query (e.g. `10x5` to create a table of that size). The custom searching logic and UI are decoupled from the base Plugin which allows for any search, cache, and ranking solutions you may need.

https://user-images.githubusercontent.com/13852400/176021961-24266f5e-b3b3-4795-ab7e-2b6f363a749f.mov

Closes #2114

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -242,8 +242,9 @@ function ShortcutTypeahead<TOption extends TypeaheadOption>({
   }, [resolution.match.matchingString]);
 
   useEffect(() => {
+    const rootElement = editor.getRootElement();
+
     function positionMenu() {
-      const rootElement = editor.getRootElement();
       const containerDiv = anchorElementRef.current;
       containerDiv.setAttribute('aria-label', 'Typeahead menu');
       containerDiv.setAttribute('id', 'typeahead-menu');
@@ -256,16 +257,19 @@ function ShortcutTypeahead<TOption extends TypeaheadOption>({
         containerDiv.style.display = 'block';
         containerDiv.style.position = 'absolute';
         if (!containerDiv.isConnected) {
-          rootElement.setAttribute('aria-controls', 'typeahead-menu');
           document.body.append(containerDiv);
         }
         anchorElementRef.current = containerDiv;
+        rootElement.setAttribute('aria-controls', 'typeahead-menu');
       }
     }
     positionMenu();
     window.addEventListener('resize', positionMenu);
     return () => {
       window.removeEventListener('resize', positionMenu);
+      if (rootElement !== null) {
+        rootElement.removeAttribute('aria-controls');
+      }
     };
   }, [editor, resolution, options]);
 


thegreatercurve:
CC @im-adithya

echarles:
This is awesome. Loving the screencast and can't wait to see this top requested feature merged. I guess a next iteration could add a plugin that would decide which options to show based on the position in the document.

Screenshot 2022-06-30 at 17 55 58

tylerjbainbridge:
> This is awesome. Loving the screencast and can't wait to see this top requested feature merged. I guess a next iteration could add a plugin that would decide which options to show based on the position in the document.
> 
> Screenshot 2022-06-30 at 17 55 58

That's a great idea! I should have this PR merged tomorrow by the way. Sorry for the delay, I was on vacation and away from a computer. Excited to get this in though.

tsconfig.json
```diff
@@ -105,6 +105,9 @@
       "@lexical/react/LexicalRichTextPlugin": [
         "./packages/lexical-react/src/LexicalRichTextPlugin.tsx"
       ],
+      "@lexical/react/LexicalShortcutMenuPlugin": [
+        "packages/lexical-react/src/LexicalSlashShortcutMenuPlugin.tsx"
```
acywatson, line 109: Is there really any reason for it to be a "slash" menu, specifically? It feels like this piece could be configurable and maybe shouldn't be part of the name. For instance, I can trigger it with @ or $ or whatever.

packages/lexical-react/src/LexicalSlashShortcutMenuPlugin.tsx
```diff
@@ -0,0 +1,607 @@
...
+  }
+  return null;
+}
+
+function getPossibleMatch(text): QueryMatch | null {
+  const match = checkForSlashMatch(text, 0);
+  return match;
+}
+
+function getTextUpToAnchor(selection: RangeSelection): string | null {
```
acywatson, line 183: It looks like a lot of this is copied over from mentions - is there any way we can refactor this so we don't have it all in two places? Is here a generic typeahead component or utility that we could make?

We've also had a few people in the community trying to make typeahead like functionality, so some shared, generic logic to make this easier could be really valuable, in addition to being easier to maintain.

packages/lexical-react/src/LexicalSlashShortcutMenuPlugin.tsx
```diff
@@ -0,0 +1,607 @@
...
+ * Walk backwards along user input and forward through entity title to try
+ * and replace more of the user's text with entity.
+ */
+function getFullMatchOffset(
+  documentText: string,
+  entryText: string,
+  offset: number,
+): number {
+  let triggerOffset = offset;
+  for (let ii = triggerOffset; ii <= entryText.length; ii++) {
```
acywatson, line 236: I'm sure I'm missing something, but why is this a double "i"?

packages/lexical-react/src/LexicalSlashShortcutMenuPlugin.tsx
```diff
@@ -0,0 +1,607 @@
...
+    this.keyboardShortcut = options.keyboardShortcut;
+    this.onSelect = options.onSelect;
+    this.ref = {current: null};
+  }
+
+  setRefElement = (element: HTMLElement | null) => {
+    this.ref.current = element;
+  };
+
+  isMatch = (query: string) => {
```
acywatson, line 123: Is the idea here that the consumer of this plugin would be able to call isMatch on each option in the OptionResolverFn? This seems a little restrictive/opinionated. Can we simplify and just let the user filter in the OptionResolverFn however they want? I guess they can do that now, but it's not clear why we need isMatch or even an instance of TypeaheadOption. Why not just make TypeaheadOption an interface and let people filter against whatever key/values?

acywatson (commented):
Overall, looks solid - I think we can factor out some of the common logic, make it configurable, and shift more of the specific logic into the playground plugin.

packages/lexical-react/src/LexicalSlashShortcutMenuPlugin.tsx
tylerjbainbridge: I'll take a crack at reusing more logic today.

tylerjbainbridge:
@acywatson I rewrote a lot of this to make it much more generic. Now the consumer only need to implement a handful of simple methods to build their very own `Typeahead`. I refactored our `MentionsPlugins` to use `LexicalTypeaheadMenuPlugin` as a proof of concept in addition to building a `ComponentPickerMenuPlugin` for the Playground.

image

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
```diff
@@ -0,0 +1,387 @@
...
+                const codeNode = $createCodeNode();
+                selection.insertNodes([codeNode]);
+                selection.insertRawText(textContent);
+              }
+            }
+          }),
+      }),
+      new ComponentPickerOption('Divider', {
+        icon: <i className="icon horizontal-rule" />,
+        keywords: ['horizantal rule', 'divider', 'hr'],
```
BobPritchett, line 253: horizontal

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
```diff
@@ -0,0 +1,388 @@
...
+        (n) =>
+          new ComponentPickerOption(`Heading ${n}`, {
+            icon: <i className={`icon h${n}`} />,
+            keywords: ['heading', 'header', `h${n}`],
+            onSelect: () =>
+              editor.update(() => {
+                const selection = $getSelection();
+                if ($isRangeSelection(selection)) {
+                  $wrapLeafNodesInElements(selection, () =>
+                    // @ts-ignore
```
acywatson, line 189: What's wrong here?

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
```diff
@@ -0,0 +1,388 @@
...
+    [editor],
+  );
+
+  return (
+    <>
+      {modal}
+      <LexicalTypeaheadMenuPlugin<ComponentPickerOption>
+        onQueryChange={setQueryString}
+        onSelectOption={onSelectOption}
+        matchResolverFn={checkForTriggerMatch}
```
acywatson, line 356: match could also seem to be related to the option filtering. Should we standardize on "trigger" for all thing related to what causes the typeahead to activate?

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
```diff
@@ -368,7 +368,7 @@ function InsertImageUriDialogBody({
   );
 }
 
-function InsertImageUploadedDialogBody({
+export function InsertImageUploadedDialogBody({
```
acywatson, line 371: I think this breaks the encapsulation of plugins and sort of couples Toolbar with whatever imports these things - should separate them out into a different shared place if they're going to be reused?

acywatson:
This looks good to me - once we get the mentions tests passing, it's good to go.

acywatson (approved):

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
tylerjbainbridge: There are a few main components to this typeahead
1. `matchResolverFn`: when do I open
2. `onQueryChange`: what is the current search query
3. `menuRenderFn`: what do I look like
4. `options`: what options
5. `onSelectOption`: what do I do when an option is selected

So to answer your question, the `trigger`, aka `matchResolverFn`, defines when the menu will open and the `queryString` can be used to filter the options whether it's in memory or asynchronously on the server.

They could be related moved into one function, but there's some processing that needs to be done on the match/textnode to derive the query string and that can be standardized (and used for `onQueryChange`).

I'm open to consolidating if you think it's confusing though :)

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
tylerjbainbridge: That's a good point. I'd argue that since this is for the playground it's OK, but I can split them into a shared file with all the modals if you think it's more clear.

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
```diff
@@ -0,0 +1,388 @@
...
+  option: ComponentPickerOption;
+}) {
+  let className = 'item';
+  if (isSelected) {
+    className += ' selected';
+  }
+  return (
+    <li
+      key={option.key}
+      tabIndex={-1}
```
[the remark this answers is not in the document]
tylerjbainbridge: Yup, that's handled in the core component picker via Lexical Events.

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -0,0 +1,566 @@
...
+    matchingString: string,
+  ) => void;
+  options: Array<TOption>;
+  menuRenderFn: MenuRenderFn<TOption>;
+  matchResolverFn: MatchResolverFn;
+};
+
+type MatchResolverFn = (text: string) => QueryMatch | null;
+
+export function LexicalTypeaheadMenuPlugin<TOption extends TypeaheadOption>({
```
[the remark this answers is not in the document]
tylerjbainbridge: Definitely, I think `LexicalTypeaheadMenuPlugin` could actually be used to solve that use case directly. Happy to hack on something basic for the Playground as a proof of concept.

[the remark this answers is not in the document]
tylerjbainbridge: Basically `LexicalTypeaheadMenuPlugin` constructs an invisible div positioned over the matching content and lets you attach a portal to it. So you could theoretically position suggested text to the right of the cursor, but it might require some changes in the core plugin.

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
acywatson: I was thinking more about just changing the name from matchResolverFn to like "triggerFn" or something. I don't think it's a big deal though, tbh.

packages/lexical-playground/src/plugins/ToolbarPlugin.tsx
acywatson: I'm good either way - I think people copy and paste out of the playground sometimes, and we should be careful about the example we set, but in this case I would expect people to rely on the @lexical/react component and build there own functionality up from there, so should be fine to leave as-is.

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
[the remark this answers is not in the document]
acywatson: > Basically `LexicalTypeaheadMenuPlugin` constructs an invisible div positioned over the matching content and lets you attach a portal to it. So you could theoretically position suggested text to the right of the cursor, but it might require some changes in the core plugin.

Yea or you could make the content of the TypeaheadNode be two differently styled text nodes, one representing what the user types and the other representing the autocomplete suggestion.

We can iterate on this.

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
tylerjbainbridge: It's because Typescript can't confirm that ``h${n}`` is of the type `HeadingTagType`, but we know that the #s will be 1-3 from the loop above.

packages/lexical-playground/src/plugins/ComponentPickerPlugin.tsx
tylerjbainbridge: I think you're right. I'll change the naming :)

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -0,0 +1,572 @@
...
+      });
+    };
+
+    const removeUpdateListener = editor.registerUpdateListener(updateListener);
+
+    return () => {
+      activeRange = null;
+      removeUpdateListener();
+    };
+  }, [editor, triggerFn, onQueryChange, resolution]);
```
adri1wald, line 556: Hi, could you explain why this `useEffect` has `resolution` as a dep?

tylerjbainbridge: Good catch, it looks like this isn't necessary here and was an oversight on my end. Shouldn't cause any major issues in the meantime, but I'll get a PR up that removes it.

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -0,0 +1,572 @@
...
+    return () => {
+      const rootElem = editor.getRootElement();
+      if (rootElem !== null) {
+        rootElem.removeAttribute('aria-activedescendant');
+      }
+    };
+  }, [editor]);
+
+  useLayoutEffect(() => {
+    if (options === null) {
```
adri1wald, line 319: I'm also confused about the `options === null` check here and in the `registerCommand` callbacks. As far as I can tell, `options` is always an array (if the user follows the typescript contract). I suspect you meant to check whether `options.length === 0`?

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -0,0 +1,572 @@
...
+        },
+        COMMAND_PRIORITY_LOW,
+      ),
+      editor.registerCommand<KeyboardEvent>(
+        KEY_ESCAPE_COMMAND,
+        (payload) => {
+          const event = payload;
+          if (options === null || selectedIndex === null) {
+            return false;
+          }
```
adri1wald, line 372: This would block closing of the typeahead menu when there are no options (assuming the `options === null` checks are replaced with length checks).

Issue #2536, opened 2022-06-27
Feature: Ensure the type passed to `createCommand` is retained, inferred, and enforced


PAkerstrand:
Issue created from a Q-and-A post over at Discord: https://discord.com/channels/953974421008293909/955972012541628456/990966752110325771

I really dig the command system in Lexical, but there are two things that irks me a little bit:

1. Strongly typing the command seems like a great idea, but the underlying types makes it a bit hard to debug the code. Basically, `createCommand()`-call just creates an empty object `{}` that's used as a unique key for a `Map`. This is great, but given that approach, whenever a command is dispatched, the `type` is simply `{}`. Would it make sense to have some internal field or name to trace which command is being dispatched?

2. Each `LexicalCommand` is generic in terms of the payload, so you can have a `LexicalCommand<boolean>`, or `LexicalCommand<KeyboardEvent>`. However, the `LexicalCommand<T>` is really just an alias for `Readonly<Record<string, unknown>>`, so the required type of the payload isn't saved. This also makes us lose type safety down the line, where you can do

```ts
const SOME_EVENT = createCommand<boolean>();

// ...

editor.dispatchCommand(SOME_EVENT, 'Not a boolean...');
```
and TypeScript will happily accept it. It'd be good if the payload was constrained by the type of the command here.

Additionally, when you `registerCommand` you need to repeat the type of the payload. Given that the type information was stored inside of the command itself, this could then be inferred, along the lines of:

```ts
editor.registerCommand(SOME_EVENT, (payload) => { /* do stuff with inferred boolean */ }
```

Finally, if the `registerCommand` could infer the payload type from the command itself, then we could ensure that the command handler has the expected signature. With the current approach there is nothing preventing a very similar error as in the `dispatchCommand` example, where the `registerCommand`-handler expects a payload differing from the one provided to `createCommand`.

_____

As a starting point to the second bullet, I've created a PoC draft PR. There are still some type checking errors introduced by this change, related to `LexicalEvents`:

```
packages/lexical/src/LexicalEvents.ts:514:68 - error TS2345: Argument of type 'InputEvent' is not assignable
  to parameter of type 'string'.

514         dispatchCommand(editor, CONTROLLED_TEXT_INSERTION_COMMAND, event);
                                                                       ~~~~~

packages/lexical/src/LexicalEvents.ts:521:68 - error TS2345: Argument of type 'InputEvent' is not assignable
  to parameter of type 'string'.

[...]

PR #2537, opened 2022-06-27
Capture the expected payload type in commands

PAkerstrand:
Implements inference and enforcement of the payload type for `LexicalCommand`, as discussed in point 2. in #2536.

### DIFF
--- a/packages/lexical-code/src/index.ts
+++ b/packages/lexical-code/src/index.ts
@@ -1160,24 +1160,22 @@ export function registerCodeHighlighting(editor: LexicalEditor): () => void {
     ),
     editor.registerCommand(
       KEY_ARROW_UP_COMMAND,
-      (payload: KeyboardEvent): boolean =>
-        handleShiftLines(KEY_ARROW_UP_COMMAND, payload),
+      (payload): boolean => handleShiftLines(KEY_ARROW_UP_COMMAND, payload),
       COMMAND_PRIORITY_LOW,
     ),
     editor.registerCommand(
       KEY_ARROW_DOWN_COMMAND,
-      (payload: KeyboardEvent): boolean =>
-        handleShiftLines(KEY_ARROW_DOWN_COMMAND, payload),
+      (payload): boolean => handleShiftLines(KEY_ARROW_DOWN_COMMAND, payload),
       COMMAND_PRIORITY_LOW,
     ),
     editor.registerCommand(
       MOVE_TO_END,
-      (payload: KeyboardEvent): boolean => handleMoveTo(MOVE_TO_END, payload),
+      (payload): boolean => handleMoveTo(MOVE_TO_END, payload),
       COMMAND_PRIORITY_LOW,
     ),
     editor.registerCommand(
       MOVE_TO_START,
-      (payload: KeyboardEvent): boolean => handleMoveTo(MOVE_TO_START, payload),
+      (payload): boolean => handleMoveTo(MOVE_TO_START, payload),
       COMMAND_PRIORITY_LOW,
     ),
   );
--- a/packages/lexical-plain-text/src/index.ts
+++ b/packages/lexical-plain-text/src/index.ts
@@ -70,7 +70,7 @@ function onCopyForPlainText(
   event.preventDefault();
   editor.update(() => {
     const clipboardData =
-      event instanceof ClipboardEvent ? event.clipboardData : null;
+      event instanceof KeyboardEvent ? null : event.clipboardData;
     const selection = $getSelection();
 
     if (selection !== null) {
@@ -95,7 +95,7 @@ function onPasteForPlainText(
   editor.update(() => {
     const selection = $getSelection();
     const clipboardData =
-      event instanceof ClipboardEvent ? event.clipboardData : null;
+      event instanceof InputEvent ? null : event.clipboardData;
 
     if (clipboardData != null && $isRangeSelection(selection)) {
       $insertDataTransferForPlainText(clipboardData, selection);
--- a/packages/lexical-plain-text/src/index.ts
+++ b/packages/lexical-plain-text/src/index.ts
@@ -6,7 +6,7 @@
  *
  */
 
-import type {EditorState, LexicalEditor} from 'lexical';
+import type {CommandPayloadType, EditorState, LexicalEditor} from 'lexical';
 
 import {
   $getHtmlContent,
@@ -64,12 +64,13 @@ const updateOptions: {
 } = options;
 
 function onCopyForPlainText(
-  event: ClipboardEvent,
+  event: CommandPayloadType<typeof COPY_COMMAND>,
   editor: LexicalEditor,
 ): void {
   event.preventDefault();
   editor.update(() => {
-    const clipboardData = event.clipboardData;
+    const clipboardData =
+      event instanceof ClipboardEvent ? event.clipboardData : null;
     const selection = $getSelection();
 
     if (selection !== null) {
@@ -87,13 +88,14 @@ function onCopyForPlainText(
 }
 
 function onPasteForPlainText(
-  event: ClipboardEvent,
+  event: CommandPayloadType<typeof PASTE_COMMAND>,
   editor: LexicalEditor,
 ): void {
   event.preventDefault();
   editor.update(() => {
     const selection = $getSelection();
-    const clipboardData = event.clipboardData;
+    const clipboardData =
+      event instanceof ClipboardEvent ? event.clipboardData : null;
 
     if (clipboardData != null && $isRangeSelection(selection)) {
       $insertDataTransferForPlainText(clipboardData, selection);
@@ -101,7 +103,10 @@ function onPasteForPlainText(
   });
 }
 
-function onCutForPlainText(event: ClipboardEvent, editor: LexicalEditor): void {
+function onCutForPlainText(
+  event: CommandPayloadType<typeof CUT_COMMAND>,
+  editor: LexicalEditor,
+): void {
   onCopyForPlainText(event, editor);
   editor.update(() => {
     const selection = $getSelection();
@@ -378,7 +383,7 @@ export function registerPlainText(
       },
       COMMAND_PRIORITY_EDITOR,
     ),
-    editor.registerCommand<ClipboardEvent>(
+    editor.registerCommand(
       COPY_COMMAND,
       (event) => {
         const selection = $getSelection();
@@ -392,7 +397,7 @@ export function registerPlainText(
       },
       COMMAND_PRIORITY_EDITOR,
     ),
-    editor.registerCommand<ClipboardEvent>(
+    editor.registerCommand(
       CUT_COMMAND,
       (event) => {
         const selection = $getSelection();
@@ -406,7 +411,7 @@ export function registerPlainText(
       },
       COMMAND_PRIORITY_EDITOR,
     ),
-    editor.registerCommand<ClipboardEvent>(
+    editor.registerCommand(
       PASTE_COMMAND,
       (event) => {
         const selection = $getSelection();
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -7,6 +7,7 @@
  */
 
 import type {
+  CommandPayloadType,
   DOMConversionMap,
   DOMConversionOutput,
   EditorConfig,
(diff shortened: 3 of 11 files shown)


packages/lexical-code/src/index.ts
```diff
@@ -1150,24 +1150,22 @@ export function registerCodeHighlighting(editor: LexicalEditor): () => void {
     ),
     editor.registerCommand(
       KEY_ARROW_UP_COMMAND,
-      (payload: KeyboardEvent): boolean =>
-        handleShiftLines(KEY_ARROW_UP_COMMAND, payload),
+      (payload): boolean => handleShiftLines(KEY_ARROW_UP_COMMAND, payload),
```
PAkerstrand, line 1153: With the change to `LexicalCommand`, the payload is not inferred from the `command`

packages/lexical/src/LexicalCommands.ts
```diff
@@ -19,8 +19,9 @@ export const DELETE_CHARACTER_COMMAND: LexicalCommand<boolean> =
 export const INSERT_LINE_BREAK_COMMAND: LexicalCommand<boolean> =
   createCommand();
 export const INSERT_PARAGRAPH_COMMAND: LexicalCommand<void> = createCommand();
-export const CONTROLLED_TEXT_INSERTION_COMMAND: LexicalCommand<string> =
-  createCommand();
+export const CONTROLLED_TEXT_INSERTION_COMMAND: LexicalCommand<
+  InputEvent | string
+> = createCommand();
```
PAkerstrand, line 24: This was actually incorrectly typed in the command itself, as we dispatch the payload as either a `string` or, in `LexicalEvents.ts`, an `InputEvent`

packages/lexical-rich-text/src/index.ts
```diff
@@ -388,11 +390,15 @@ function onPasteForRichText(
   });
 }
 
-function onCopyForRichText(event: ClipboardEvent, editor: LexicalEditor): void {
+function onCopyForRichText(
+  event: CommandPayloadType<typeof COPY_COMMAND>,
```
thegreatercurve, line 394: Could we just use a simple union type here instead? It might be a little more readable and `CommandPayloadType` feels slightly too general if we're not using it a lot elsewhere. 

```suggestion
  event: ClipboardEvent | KeyboardEvent,
```

thegreatercurve (approved):

thegreatercurve:
Closes #2536

thegreatercurve:
Thanks for looking into this @PAkerstrand!

packages/lexical-plain-text/src/index.ts
```diff
@@ -64,12 +64,13 @@ const updateOptions: {
 } = options;
 
 function onCopyForPlainText(
-  event: ClipboardEvent,
+  event: CommandPayloadType<typeof COPY_COMMAND>,
```
zurfyx, line 67: This should be a clipboard event?

zurfyx (commented):
Thanks for taking a stab at this, we certainly missed this when we migrated to TS and it's pretty important to get this right! Turns out TS resolves the type whereas Flow takes the first type available so we can no longer leverage the empty object and unused generic trick

Wrote down a couple comments, the idea is correct but I think the clipboard types needs fixing..
I wrote a simple version outlining the gist of the problem, it comes down to a current issue on the existing codebase I think (lmk if I'm mistaken, I checked pretty quick) https://github.com/facebook/lexical/commit/5f743c9104fe4f8f5651ed16b7c4d3a901ed6137

packages/lexical-plain-text/src/index.ts
PAkerstrand: Correct, my bad!

packages/lexical-rich-text/src/index.ts
PAkerstrand: @thegreatercurve I actually started with just a union type. The question then becomes, do we want to provide a utility type for extracting the payload in a nice way from the `lexical` package, to make it a bit easier for other users if they want to extract a function for delegating to from the command handlers? Inside of the command handlers the payload is inferred, but that's only the case if/when it is inlined. If you want to do something like this:

```ts
type MY_COMMAND = createCommand<string | SomeOtherType>();

// ...
editor.registerCommand(MY_COMMAND, myCommandHandler);

// ...
function myCommandHandler(payload: string | SomeOtherType) {}
```

The idea here was to provide functionality for this, such that the payload type and the argument type are can easily be kept in sync if that union type would grow for some reason

```ts
function myCommandHandler(payload: CommandPayloadType<typeof MY_COMMAND>) {}
```

If we do want to provide it, it would make a lot of sense to adorn it with a proper TSDoc comment to show its purpose and some examples of how to use it.

I guess it's also a question of:
* How often do the generic type of a command really change?
* Where would we want type errors in usage to pop up?

packages/lexical/src/LexicalEditor.ts
```diff
@@ -167,7 +167,13 @@ export const COMMAND_PRIORITY_HIGH = 3;
 export const COMMAND_PRIORITY_CRITICAL = 4;
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export type LexicalCommand<T = never> = Readonly<Record<string, unknown>>;
+export type LexicalCommand<T = never> = Readonly<{
```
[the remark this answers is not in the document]
PAkerstrand: No, I don't think we do. I guess one could argue that it'd be convenient to have some way to describe "payloadless" commands, where you could just do `editor.dispatchCommand(THE_COMMAND)`, but this does not help with that

packages/lexical-plain-text/src/index.ts
PAkerstrand: One of the events it can receive is a `ClipboardEvent`. However, the `COPY_COMMAND` is also dispatched with a `KeyboardEvent` from `LexicalEvents.ts`

acywatson: Should it? I originally had the same thought, but I think this is actually a command listener, not a DOM event listener, in which case I think the abstraction is appropriate, right?

zurfyx: Ah yes, makes sense! I'm happy with the proposed changes, will let @acywatson do the final review! Thanks

packages/lexical/src/LexicalEditor.ts
```diff
@@ -166,14 +166,12 @@ export const COMMAND_PRIORITY_NORMAL = 2;
...
-   * Capture the type of the payload this command expects. This ensures that `dispatchCommand` and `registerCommand` can
-   * correctly enforce and infer the payload later on
-   */
-  _payloadType?: T;
-}>;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-interface
+export interface LexicalCommand<TPayload> {}
+export type CommandPayloadType<C> = C extends LexicalCommand<infer R>
+  ? R
+  : never;
```
PAkerstrand, line 173: I moved this type helper here since it is quite tightly tied to the `LexicalCommand`-type

packages/lexical/src/LexicalEditor.ts
```diff
@@ -166,14 +166,12 @@ export const COMMAND_PRIORITY_NORMAL = 2;
...
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export type LexicalCommand<T = never> = Readonly<{
-  /**
-   * Capture the type of the payload this command expects. This ensures that `dispatchCommand` and `registerCommand` can
-   * correctly enforce and infer the payload later on
-   */
-  _payloadType?: T;
-}>;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-interface
+export interface LexicalCommand<TPayload> {}
```
PAkerstrand, line 170: I changed this from a type alias into a "proper interface". That allows the type itself to capture and remember its value, instead of the hack with an unused field

packages/lexical/src/LexicalEditor.ts
```diff
@@ -636,7 +634,10 @@ export class LexicalEditor {
     return true;
   }
 
-  dispatchCommand<P>(type: LexicalCommand<P>, payload: P): boolean {
+  dispatchCommand<
+    TCommand extends LexicalCommand<unknown>,
+    TPayload extends CommandPayloadType<TCommand>,
+  >(type: TCommand, payload: TPayload): boolean {
```
PAkerstrand, line 640: @thegreatercurve Now we have some internal usage of the `CommandPayloadType`, namely to ensure that the type of the payload matches what's extracted from the type of the first argument

packages/lexical/src/LexicalEvents.ts
```diff
@@ -482,7 +482,7 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
     if (inputType === 'insertText') {
       if (data === '\n') {
         event.preventDefault();
-        dispatchCommand(editor, INSERT_LINE_BREAK_COMMAND, undefined);
+        dispatchCommand(editor, INSERT_LINE_BREAK_COMMAND, false);
```
PAkerstrand, line 485: New type errors that popped up from the stricter enforcement.

A `LexicalCommand<boolean>` no longer accepts `undefined` or `null` as payload

packages/lexical/src/LexicalEvents.ts
```diff
@@ -967,31 +967,67 @@ export function addRootElementEvents(
             if (!editor.isReadOnly()) {
               switch (eventName) {
                 case 'cut':
-                  return dispatchCommand(editor, CUT_COMMAND, event);
+                  return dispatchCommand(
+                    editor,
+                    CUT_COMMAND,
+                    event as ClipboardEvent,
+                  );
```
PAkerstrand, line 974: Are we fine with these type casts, or do we need to refactor the code such that TS can properly infer the type itself?

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1098,11 +1099,10 @@ export function isFirefoxClipboardEvents(): boolean {
...
 
-export function dispatchCommand<P>(
-  editor: LexicalEditor,
-  type: LexicalCommand<P>,
-  payload: P,
-): boolean {
+export function dispatchCommand<
+  TCommand extends LexicalCommand<unknown>,
+  TPayload extends CommandPayloadType<TCommand>,
+>(editor: LexicalEditor, type: TCommand, payload: TPayload): boolean {
```
PAkerstrand, line 1105: Same signature as in [`LexicalEditor.ts`](https://github.com/facebook/lexical/pull/2537/commits/754065483f6ede2f40eedd481956498866f569e7#diff-b38ac058e19ebb0f68aff9acf5c52df0683a2229bf882d2262c71508b8fb6f72R637-R640)

packages/lexical/src/LexicalEditor.ts
```diff
@@ -167,7 +167,13 @@ export const COMMAND_PRIORITY_HIGH = 3;
 export const COMMAND_PRIORITY_CRITICAL = 4;
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export type LexicalCommand<T = never> = Readonly<Record<string, unknown>>;
+export type LexicalCommand<T = never> = Readonly<{
+  /**
+   * Capture the type of the payload this command expects. This ensures that `dispatchCommand` and `registerCommand` can
+   * correctly enforce and infer the payload later on
+   */
+  _payloadType?: T;
```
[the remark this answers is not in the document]
PAkerstrand: I pushed a change in approach in order to tackle this issue. See https://github.com/facebook/lexical/pull/2537/commits/754065483f6ede2f40eedd481956498866f569e7

packages/lexical/src/LexicalEditor.ts
```diff
@@ -167,7 +167,30 @@ export const COMMAND_PRIORITY_HIGH = 3;
 export const COMMAND_PRIORITY_CRITICAL = 4;
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export type LexicalCommand<T = never> = Readonly<Record<string, unknown>>;
+export type LexicalCommand<TPayload> = Record<string, never>;
```
PAkerstrand, line 170: Turns out it was the `Readonly` type helper that breaks type inference later on. `Record<string, never>` signifies an empty object, which is what the commands currently are, but still allows the generic type to be extracted later on.

PAkerstrand:
> I keep getting these types of failures in the test suite, and it's usually on Windows and composition. Not sure whether it's coincidence or if I've accidentally introduced some errors? Not sure how I could introduce connection refused errors though.
> 
> image

____

This seems to be the core issue of the failed runs, we get an `EMFILE`-error emitted. Might be some file descriptor leakage somewhere?

image

acywatson:
@thegreatercurve - the author let me know he is on vacation for the next 3 weeks - can you take this over and see if we need to make any changes or if we can merge it as-is?

PAkerstrand:
> @thegreatercurve - the author let me know he is on vacation for the next 3 weeks - can you take this over and see if we need to make any changes or if we can merge it as-is?
> 
> The CI failure is the Windows thing I fixed and should pass on rebase (or we can ignore that signal and merge as-is if you're good)

Correct ☝🏼 

I’m completely OK with you pushing any changes required to get this merged. I am somewhat around to be able to respond, but won’t be coding for the next few weeks

PR #2547, opened 2022-06-28
Add ability to set format and details on TextNode by `string`

### LINKED ISSUE #2491: Feature: Text Node format and detail setters should accept a string
`setFormat` and `setDetail` both accept numbers for format and detail type, but `toggleFormat` accepts a `TextFormatType` corresponding to the format type. 

We can still store the formats as numbers for the sake of saving space during serialization, but these should all accept a `TextFormatType` for developer experience.

thegreatercurve:
Small improvement to the API of `TextNode`, to make the logic consistent with `toggleFormat`

Closes #2491

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -25,6 +25,7 @@ import invariant from 'shared/invariant';
 
 import {
   COMPOSITION_SUFFIX,
+  DETAIL_TYPE_TO_DETAIL,
   IS_BOLD,
   IS_CODE,
   IS_DIRECTIONLESS,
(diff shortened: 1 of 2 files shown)


zurfyx (approved):

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -500,17 +503,21 @@ export class TextNode extends LexicalNode {
     return;
   }
 
-  setFormat(format: number): this {
+  // TODO 0.4 This should just be a `string`.
```
fantactuka, line 506: Does it mean we won't be able to set multiple formats, eg bold and italic

PR #2554, opened 2022-06-29
fix(lexical): deselecting a image/poll node by clicking

### LINKED ISSUE #2457: Bug: Deselecting a decorator node by clicking an empty paragraph node sets selection to null
When a decorator node is selected, clicking an empty paragraph node clears the selection, rather than selecting the paragraph.

Lexical version: 0.3.5

## Steps To Reproduce

1. In the lexical playground, insert and select an image node
2. Click on an empty paragraph node

Link to code example: reproducible in playground

## The current behavior
When a decorator node is selected, clicking on an empty paragraph clears the selection

## The expected behavior
Clicking on an empty paragraph node while the decorator node is selected should select the paragraph node with a collapsed range selection

LuciNyan:
### Description
After we have selected an image/poll. When we click on an empty paragraph node or the end of a paragraph that ends with an image/poll, the `domAnchor.nodeType` will be ELEMENT_NODE

https://user-images.githubusercontent.com/22126563/176342847-82d30cc5-7094-4f9d-abd7-cbf428a15b02.mov

closes #2457

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -66,6 +66,7 @@ import {
 import {KEY_MODIFIER_COMMAND} from './LexicalCommands';
 import {
   COMPOSITION_START_CHAR,
+  DOM_ELEMENT_TYPE,
   DOM_TEXT_TYPE,
   DOUBLE_LINE_BREAK,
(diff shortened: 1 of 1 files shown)


zurfyx (approved):
Sounds reasonable to me, thanks!

PR #2555, opened 2022-06-29
fix(lexical): Text with underline format is stripped out on paste

### LINKED ISSUE #2552: Bug: Text with 'underline' format is stripped out on paste
Pasting rich text from other editors which support styling (e.g. Google Docs, Apple Notes and etc) removes the 'underline' style.

Lexical version: `0.3.5`

## Steps To Reproduce

1. Copy a rich text with 'underline' style from any other rich text editor
2. Paste it into the lexical playground editor

https://user-images.githubusercontent.com/1575198/176225987-f3a0188b-2323-4a77-92e5-47c320f8d93b.mov

**Note:** this is not just the playground issue. I am seeing the same issue while testing in my env. 

## The current behavior
Rich text is pasted without 'underline' style.

## The expected behavior
'underline' style should be preserved as other styles (e.g. bold or italic) do.

LuciNyan:
### Description
function `convertSpanElement` does not handle `text-decoration: underline`

closes #2552 

https://user-images.githubusercontent.com/22126563/176366558-08a7e3d8-f317-4bb3-89b1-7e598f971ebd.mov

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -830,11 +830,13 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
   const span = domNode as HTMLSpanElement;
   // Google Docs uses span tags + font-weight for bold text
   const hasBoldFontWeight = span.style.fontWeight === '700';
-  // Google Docs uses span tags + text-decoration for strikethrough text
+  // Google Docs uses span tags + text-decoration: line-through for strikethrough text
   const hasLinethroughTextDecoration =
     span.style.textDecoration === 'line-through';
   // Google Docs uses span tags + font-style for italic text
   const hasItalicFontStyle = span.style.fontStyle === 'italic';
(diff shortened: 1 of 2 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -847,6 +849,9 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
       if ($isTextNode(lexicalNode) && hasItalicFontStyle) {
         lexicalNode.toggleFormat('italic');
       }
+      if ($isTextNode(lexicalNode) && hasUnderlineTextDecoration) {
+        lexicalNode.toggleFormat('underline');
+      }
```
zurfyx, line 854: Nit; mind adding an early return at the top, would be faster than reevaluating this 4 times
```
if (!$isTextNode(lexicalNode)) {
  return;
}
```

zurfyx (approved):
Thanks Luci!

zurfyx:
Feel free to follow on another PR, going to close 0.3.6 now

PR #2571, opened 2022-06-30
fix: import color and vertical align info from html string

im-adithya:
https://user-images.githubusercontent.com/64399555/176701853-0ef15dee-18c7-4b92-8498-11381ee0aeae.mp4

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -837,20 +837,49 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
   const hasItalicFontStyle = span.style.fontStyle === 'italic';
   // Google Docs uses span tags + text-decoration: underline for underline text
   const hasUnderlineTextDecoration = span.style.textDecoration === 'underline';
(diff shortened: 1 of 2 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -837,21 +837,49 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
...
   // Google Docs uses span tags + text-decoration: underline for underline text
   const hasUnderlineTextDecoration = span.style.textDecoration === 'underline';
+  // Google Docs uses span tags + vertical-align to specify subscript and superscript
+  const verticalAlign = span.style.verticalAlign;
+  // Google Docs uses span tags + color, background-color for coloring
+  const backgroundColor = span.style.backgroundColor;
+  const textColor = span.style.color;
+
+  //TODO: font-size and coloring of subscript & superscript
+  let cssString = '';
```
fantactuka, line 847: Should it be defined within forChild as otherwise same color/background duplicate styles could be appeneded

acywatson (approved):

PR #2573, opened 2022-06-30
Ensure to call existing listeners only (not newly added ones)

fantactuka:
Like what we do with update and readOnly listeners #1602, add tests and fixes for command and mutation listeners to ensure we do not execute listeners that were added during current update cycle. This can cause an infinite listeners loop if we register to listener within synchronous useEffect call


zurfyx (approved):

acywatson (approved):
Looks good, thanks

PR #2574, opened 2022-06-30
Website error codes - lexical.dev/error/<code>

zurfyx:
Share error codes on the Website, previously just a `codes.json` in the repo
Screen Shot 2022-06-30 at 6 29 30 pm
Screen Shot 2022-06-30 at 6 31 45 pm


acywatson:
Nice! This will be super helpful. Was there no way to get the error message from the URL? Like as a query string param or even parsing the path? That way we could have one page instead of one for each message?

zurfyx:
> Nice! This will be super helpful. Was there no way to get the error message from the URL? Like as a query string param or even parsing the path? That way we could have one page instead of one for each message?

Good point, a query param would work but not sure how to do it with paths.. I can explore this further and see what I can find, this would certainly simplify the code if that's possible

acywatson:
> but not sure how to do it with paths

window.location.pathname or something should work, then parse/validate the string and get the code out.

> > Nice! This will be super helpful. Was there no way to get the error message from the URL? Like as a query string param or even parsing the path? That way we could have one page instead of one for each message?
> 
> Good point, a query param would work but not sure how to do it with paths (which is cleaner).. I can explore this further and see what I can find, this would certainly simplify the code if that's possible.
> 
> However, this approach is also lighter. We don't need to fetch for all error codes, just the one we're looking for (tho probably won't make a noticeable difference).

fantactuka:
> However, this approach is also lighter. We don't need to fetch for all error codes, just the one we're looking for (tho probably won't make a noticeable difference).

IDK about this, `codes.json` is 100 lines of text, which is incomparable with amount of "good stuff" we load to render it, single error details page would be easier to maintain. Although `lexical.dev/errors/<code>` looks nicer than `lexical.dev/errors?code=<code>`, but at cost of 100 files 🤷‍♂️

acywatson (approved):
Looks good, thanks!

PR #2579, opened 2022-07-01
fix: subscript and superscript can co-exist

im-adithya:
Screenshot from 2022-07-01 15-30-18


acywatson:
Thanks! can you try rebasing onto main to trigger a new CI run and see if we get all checks? :D

BrianHung:
ProseMirror has a flag for this if it helps: https://prosemirror.net/docs/ref/#model.MarkSpec.excludes

im-adithya:
@acywatson, this is ready for review!

thegreatercurve:
Closing this for the moment as the PR is becoming a little stale and because, generally, we need to rethink how we handling storing these formats, probably moving away from byte code to just an object of string literals or a set. It should be a little less performant but much more user friendly.

PR #2580, opened 2022-07-01
chore: use keyboard shortcuts

im-adithya:
Uses keyboard shortcuts introduced in previous PRs

### DIFF
--- a/packages/lexical-playground/__tests__/keyboardShortcuts/index.mjs
+++ b/packages/lexical-playground/__tests__/keyboardShortcuts/index.mjs
@@ -171,6 +171,15 @@ export async function moveRight(page, numCharacters = 1, delayMs) {
   }
 }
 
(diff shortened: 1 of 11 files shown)


packages/lexical-playground/__tests__/keyboardShortcuts/index.mjs
```diff
@@ -171,6 +171,15 @@ export async function moveRight(page, numCharacters = 1, delayMs) {
   }
 }
 
+export async function hitBackspace(page, numCharacters = 1, delayMs) {
```
acywatson, line 174: I know this is a bit annoying, but I think it should be "pressBackspace" for consistency with the playwright API and the other util functions (like "pressShiftEnter")

acywatson (approved):
Looks good, thanks!

Issue #2576, opened 2022-06-30
Bug:  export function for TextMatchTransformer / text-match doesn't get triggered for TextNode's


lukasjaronis:
`export` function for `TextMatchTransformer` / `text-match` doesn't get triggered for `TextNode`'s

`text-match` works with `ElementNodes` perfectly fine.

Lexical version: 0.3.6

## Steps To Reproduce

1. Click below link
2. Make sure `console` tab is open
3. Type plain text into the editor (should be `TextNodes` from my understanding).
4. Click the bottom left grey button (markdown transformation)
5. Analyze console logs

Link to code example:

(https://codesandbox.io/s/dreamy-shaw-9zg9ek?file=/src/plugins/MarkdownTransformers.ts)

## The current behavior

`export` function does not trigger at all for `TextNode`'s. 

## The expected behavior

`export` function should get called and let me check what node it is, `$isMyNode(node)` so i can `export` the `TextNode` in the markdown that I wish. `replace` function works perfectly fine from the looks of it.

PR #2582, opened 2022-07-01
Reverse MarkdownExport loop order to take TextMatchTransformers into account first

lukasjaronis:
Issue was that `TextMatchTransformers` were being skipped completely for export. Reversing the order in the loop has fixed this problem. I did a quick test on my local machine and `export` method is being triggered successfully for `TextNode`'s.


acywatson:
Code seems fine to me - let's make sure @fantactuka is good and we can merge.

acywatson:
fixes #2576

fantactuka (approved):

PR #2592, opened 2022-07-04
Added TableOfContentsPlugin

karam-qaoud:


packages/lexical-react/src/LexicalTableOfContents.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+function $insertHeadingIntoTableOfContents(
+  prevHeading: HeadingNode | null,
+  newHeading: HeadingNode | null,
+  currentTableOfContents: Array<[NodeKey, string, string]>,
+): Array<[NodeKey, string, string]> {
+  if (!newHeading) {
+    return currentTableOfContents;
+  }
+  const newNode: [NodeKey, string, string] = [
```
zurfyx, line 25: Let's not call this node, node means a completely different thing
```suggestion
  const newEntry: [NodeKey, string, string] = [
```

packages/lexical-react/src/LexicalTableOfContents.tsx
```diff
@@ -0,0 +1,170 @@
...
+import {useEffect, useState} from 'react';
+
+import {useLexicalComposerContext} from './LexicalComposerContext';
+
+function $insertHeadingIntoTableOfContents(
+  prevHeading: HeadingNode | null,
+  newHeading: HeadingNode | null,
+  currentTableOfContents: Array<[NodeKey, string, string]>,
+): Array<[NodeKey, string, string]> {
+  if (!newHeading) {
```
zurfyx, line 22:
```suggestion
  if (newHeading === null) {
```

packages/lexical-react/src/LexicalTableOfContents.tsx
```diff
@@ -0,0 +1,170 @@
...
+    return currentTableOfContents;
+  }
+  const newNode: [NodeKey, string, string] = [
+    newHeading.getKey(),
+    newHeading.getTextContent(),
+    newHeading.getTag(),
+  ];
+  let newTableOfContents: Array<[NodeKey, string, string]> = [];
+  if (!prevHeading) {
```
zurfyx, line 31:
```suggestion
  if (prevHeading === null) {
```

zurfyx (commented):
Awesome work!

1. Can we rename this module to LexicalTableOfContents__EXPERIMENTAL while it's not complete? You also need to create the `LexicalTableOfContents__EXPERIMENTAL.flow.js`

I see that this PR only contains the module itself, if you prefer adding the playground integration later that's fine but let's not postpone for long, the best way to test an externally exposed plugin works is by showcasing it in the playground. You can add it as a feature (on the left bottom gear)

Added a few comments but they're all nits

packages/lexical-playground/src/Editor.tsx
```diff
@@ -82,81 +84,92 @@ export default function Editor(): JSX.Element {
...
+            {isAutocomplete && <AutocompletePlugin />}
+            <ActionsPlugin isRichText={isRichText} />
+          </div>
+          {showTreeView && <TreeViewPlugin />}
+        </div>
+        <LexicalTableOfContents__EXPERIMENTAL>
+          {(tableOfContents) => {
+            return <TableOfContentsList tableOfContents={tableOfContents} />;
+          }}
+        </LexicalTableOfContents__EXPERIMENTAL>
```
zurfyx, line 171: Let's wrap it with a feature, like Autocomplete or MaxLength above for now

packages/lexical-playground/src/Editor.tsx
```diff
@@ -82,81 +84,92 @@ export default function Editor(): JSX.Element {
...
-        />
-        {isRichText ? (
-          <>
-            {isCollab ? (
-              <CollaborationPlugin
-                id="main"
-                providerFactory={createWebsocketProvider}
-                shouldBootstrap={!skipCollaborationInit}
-              />
+      <div className="editor-and-table-of-contents">
```
zurfyx, line 87: Am I misreading this or the fragment and div could be merged together?

packages/lexical-react/flow/LexicalTableOfContents__EXPERIMENTAL.js.flow
```diff
@@ -0,0 +1,14 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import type {NodeKey} from '../../lexical/flow/Lexical';
+
+declare export function LexicalTableOfContentsPlugin(
+  children: (tableOfContents: Array<[NodeKey, string, string]>) => React$Node,
```
zurfyx, line 13:
```suggestion
  children: (tableOfContents: Array<[NodeKey, string, HeadingTagType]>) => React$Node,
```

packages/lexical-playground/src/nodes/TableOfContentsList.tsx
```diff
@@ -0,0 +1,52 @@
+/**
```
zurfyx, line 1: This folder is reserved for Lexical nodes, there are no Lexical nodes here, this is merely a plugin. Let's move it the plugins folder. Also I'd pack it all together (with the Editor.tsx changes included)

```
import css..
function TableOfContentsPlugin() {
  return <LexicalTableOfContents>some stuff</LexicalTableOfContents>
}
```

Then you can just call this plugin directly on Editor.tsx and it will be neater

karam-qaoud:
Thank you for the feedback! Will fix these and commit in a few.

packages/lexical-playground/src/Editor.tsx
karam-qaoud: Could you elaborate more on what you mean by this, please?

zurfyx: Like this!
Screen Shot 2022-07-05 at 11 24 16 am

zurfyx:
Also, can we get rid of package.json? We don't commit this unless we intentionally add/remove packages

echarles:
I have tried the vercel deployed playground on https://lexical-playground-qd3p1sce7-fbopensource.vercel.app and it shows an empty page with an error in the console https://lexical.dev/docs/error?code=8 `LexicalComposerContext.useLexicalComposerContext: cannot find a LexicalComposerContext``

So I have built from source the branch and the result is really looking good, with ability to navigate to the heading from the TOC. Here is a quick preview (would be great to have one more comprehensive in the first comment of this PR).

Untitled5

zurfyx (commented):
Thanks! Some more feedback, some of it a must-have for prior merging, other can be iterated on in future PRs:

Extensibility support:
1. Heading can potentially change. It won't be the case for the default heading but someone can potentially override the `updateDOM` method that would cause a re-render and consequently change the observer reference
Performance:
1. The useEffect as it is triggered every time the state changes, consequently unsubscribing and resubscribing to all headingNodes again. We can track updated nodes so that we know when exactly to unsubscribe and resubscribe.
2. Do we even need to track all nodes or can we just track a single node and dynamically move the observer as the nodes become visible/hidden?

UX:
1. The editor overflows the page when pasting a very long text, try with a lipsum paragraph -> https://www.lipsum.com/
2. The table of content dots are colored immediately after the node becomes visible. It would be much better UX to do a sort of GDocs approach where the node is marked visible only when it near the top or beyond (the opposite)
3. The table of contents doesn't follow the user while we scroll down the page

Dev:
1. Let's set the table of contents features to default `false` while it's not fully polished

karam-qaoud:
Thanks @zurfyx for the detailed feedback! Will fix these bugs in a few and close this pr for the present time. Then, we can start reiterating in future pr's for performance and style improvement!

karam-qaoud:
Hi @zurfyx Just deleted the `IntersectionObserver()` and used the normal `Element.getBoundingClientRect()` instead. While we are not observing all the nodes anymore through `IntersectionObserver()`,  I still think  that using `Element.getBoundingClientRect()` repeatedly in a scroll listener is expensive as well. I tried to mitigate that cost using debouncing but ultimately should trade it off for something faster and more performant.

Issue #2559, opened 2022-06-29
Feature: Improve serialization flexibility of built-in nodes.


StormVanDerPol:
Currently `exportDOM` handles the way nodes serialize to html, which makes sense in the context of the editor DOM, but it's possible a user would want the output html obtained from `$generateHtmlFromNodes` to have a different shape. So I think  It would help for a user to somehow be able to change how the built-in nodes serialize to html.

Example case: `TextNode` currently serializes to either a `strong`, `span` or `em` tag with the CSS classes you find in the theme for formatting. It's possible a user would rather serialize `TextNode` to an actual text node, wrapped in `u`, `em` and `strong` tags for formatting, hypothetically because all their existing html is already structured like this.

thegreatercurve:
This should have been solved with the introduction of replaceable base nodes in #3367

Issue #1262, opened 2022-02-10
Decoupling plugin and nodes


fantactuka:
### Problem
Plugins and nodes are tightly coupled and it makes it harder to extend default nodes and maintaining existing plugins functionality.

#### Example
Let's take markdown plugin (copy/paste logic is also a good example), it exports bunch of node creation helpers (`$createHeadingNode`, `$createCodeBlockNode`, `$createListNode`, etc).

Then if I want to extend HeadingNode behaviour, for example to add `id` attribute to each heading so it can be used as an anchor (smth like `<h1 id="getting-started-with-react">Getting started with React</h1>`, see our [README](https://github.com/facebook/lexical#getting-started-with-react) doing it). To do so I'd extend default node and add new behaviour on top:

```js
class AnchorHeadingNode extends HeadingNode {
  static getType() {
    return 'heading';
  }

  createDOM<EditorContext>(config: EditorConfig<EditorContext>): HTMLElement {
    const element = super.createDOM(config);
    element.setAttribute('id', getAnchorID(this));
    return element;
  }

  updateDOM(prevNode: HeadingNode, dom: HTMLElement): boolean {
    dom.setAttribute('id', getAnchorID(this));
    return false;
  }
}
```

But now the problem is that all plugins that might insert HeadingNode (markdown, toolbar, copy/paste), they all keep using default HeadingNode, because they all use `$createHeadingNode` imported from default heading node file.

#### Potential solution:
Replace `$createHeadingNode()` implementation from
```js
function $createHeadingNode(tag): HeadingNode {
  return new HeadingNode(tag);
}
```
to
```js
function $createHeadingNode(tag): HeadingNode {
  const NodeConstructor = $getNodeFromRegistry(HeadingNode);
  return new NodeConstructor(tag);
}

...
// and smth like:
function $getNodeFromRegistry<T>(nodeKlass: T): T {
  return $getActiveEditor()._registeredNodes.get(nodeKlass.getType());
}
```

This will allow passing `AnchorHeadingNode` into LexicalComposer, and since it has the same type ('heading'), it'll be used whenever $createHeadingNode is used, so all callsites that create heading will use our extended heading node.

wenerme:
> I don't think we want plugins to work with extended nodes automatically. That's magic and it will cause more bugs than it will solve. 

This is how tiptap Extension works, extend exists Mark & Node, add extract attributes.

> <FooPlugin heading={HeadingNode} />

[...]

PR #2594, opened 2022-07-04
[RFC] Allow users to configure HTML exports on core nodes

acywatson:
Addresses: https://github.com/facebook/lexical/issues/2559

Still thinking through the API here - the idea so far is to allow configurability directly at the @lexical/html level, but also allow overriding defaults for the entire editor via the Editor configuration.

This also needs tests.

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -15,7 +15,7 @@ import type {
   SerializedTextNode,
 } from 'lexical';
 
-import {$generateHtmlFromNodes, $generateNodesFromDOM} from '@lexical/html';
+import {$generateHtmlFromNodes, $generateNodesFromDOM, DOMExportMap} from '@lexical/html';
 import {$createListNode, $isListItemNode} from '@lexical/list';
 import {
   $addNodeStyle,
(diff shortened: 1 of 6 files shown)


fantactuka:
Would it require having configurable importing too?

As a side note, should we go back to idea of command-based core nodes overrides where you can redefine which paragraph, heading, list nodes are used. Then to address custom html export/import it'd be possible to inherit core node, override its dom import/export methods and use instead of built-in node:

```
class BetterParagraphNode extends ParagraphNode {
  exportDOM() {
    return ...
  }

  static importDOM() {
    ...
  }
}

editor.registerCommand(CREATE_PARAGRAPH_NODE, nodeFactory(BetterParagraphNode))
```

More details: https://github.com/facebook/lexical/issues/1262#issuecomment-1139069241

acywatson:
> Would it require having configurable importing too?
> 
> As a side note, should we go back to idea of command-based core nodes overrides where you can redefine which paragraph, heading, list nodes are used. Then to address custom html export/import it'd be possible to inherit core node, override its dom import/export methods and use instead of built-in node:
> 
> ```js
> class BetterParagraphNode extends ParagraphNode {
>   exportDOM() {
>     return ...
>   }
> 
>   static importDOM() {
>     ...
>   }
> }
> 
> editor.registerCommand(CREATE_PARAGRAPH_NODE, nodeFactory(BetterParagraphNode))
> ```
> 
> More details: [#1262 (comment)](https://github.com/facebook/lexical/issues/1262#issuecomment-1139069241)

That's a really good idea. If we solve that, then this problem sort of solves itself without having to plumb through a bunch of configuration. Let me see if I can get that to work...

ahhshm:
This would be great. I want to serialize the `dir` attribute of nodes to html and this could help me. Any progress have been made?

thegreatercurve:
Closing this as this is no longer relevant with @trueadm's recent PR to add in configurable node replacements.

PR #2598, opened 2022-07-05
Added dragend to list of rootElementEvents

stuartrobinson3007:
Looks like 90% of the work to add dragend was done here: https://github.com/facebook/lexical/pull/1986

Only missing to add 'dragend' to the list of events.


acywatson:
Looks right, thanks!

acywatson (approved):

PR #2607, opened 2022-07-06
Fix pasting inline code blocks

### LINKED ISSUE #2605: Bug: Copy/Paste Pasting on Linebreak node moves text multiple lines
Lexical version: [v0.3.6]

## Steps To Reproduce

1. Create a code block with an empty first line
2. Copy something
3. Paste it on to the first line of the code block

You can run this on the playground. I believe it happens for 

https://user-images.githubusercontent.com/93895461/177620932-d82e1f67-3bc2-4df8-af99-afe8a9e3db83.mov

Link to code example:

## The current behavior
Pastes at the end of the code block

## The expected behavior
Paste where the cursor is

fantactuka:
Fixes #2605

### DIFF
--- a/packages/lexical-code/src/index.ts
+++ b/packages/lexical-code/src/index.ts
@@ -265,10 +265,20 @@ export class CodeNode extends ElementNode {
 
   static importDOM(): DOMConversionMap | null {
     return {
-      code: (node: Node) => ({
-        conversion: convertPreElement,
-        priority: 0,
-      }),
+      // Typically <pre> is used for code blocks, and <code> for inline code styles
+      // but if it's a multi line <code> we'll create a block. Pass through to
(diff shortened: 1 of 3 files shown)


acywatson:
Does it also close this: https://github.com/facebook/lexical/pull/2187

and fix this: #2049 ?

acywatson (approved):

acywatson:
> This may be a result of a different commit and may be intentional, but pasting things from dark theme gives a dark background to the text now
> 
>  Screen.Recording.2022-07-07.at.2.47.22.PM.mov

Yea, I'm pretty sure this came from a different commit that implemented support for "background-color" inline styles on pasted HTML. Not sure if this is the behavior we want. If so, we should probably at least allow "paste without formatting" or "paste without styles"

PR #2616, opened 2022-07-08
fix: Verify if there are text nodes before continue

Negan1911:
There are cases (like, when the only item in selection is a decorator block) that you will not have text nodes in selection.

This used to fail with `Cannot read properties of undefined (reading 'getFormatFlags')
` previously, now it currently returns abruptly without failing.

Previously:

https://user-images.githubusercontent.com/3399236/178030889-eed3780a-6550-4b55-80f0-7af98dfe09fd.mov

Now:

https://user-images.githubusercontent.com/3399236/178030969-ca3994ec-8def-4a94-889b-fe49ee967745.mov

(basically it will not fail)

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1052,6 +1052,9 @@ export class RangeSelection implements BaseSelection {
     const selectedTextNodesLength = selectedTextNodes.length;
 
     if (selectedTextNodesLength === 0) {
+      this.toggleFormat(formatType);
+      // When changing format, we should stop composition
+      $setCompositionKey(null);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1050,6 +1050,9 @@ export class RangeSelection implements BaseSelection {
       }
     }
     const selectedTextNodesLength = selectedTextNodes.length;
+
+    if (!selectedTextNodesLength) return;
```
zurfyx, line 1054:
```suggestion
    if (selectedTextNodesLength === 0) {
      return;
    }
```

zurfyx (approved):
While the user should ideally do some validation beforehand this makes sense to me (especially when triggered via a command), thank you!

acywatson:
Looks good - I just approved the CI run, so let's make sure that passes before merging.

acywatson (changes requested):
Looks like we have some failing e2e and unit tests.  You can ignore the console.warn - I'll fix that in a separate PR.

Negan1911:
I found that I broke some valid test cases by returning without toggling the format in, this should do it now :)

acywatson (approved):
Sorry, didn't mean to block this for so long - looks like all tests are passing now.

PR #2623, opened 2022-07-09
fix(lexical): Japanese IME issue

### LINKED ISSUE #2613: Bug: Setting a pending format at end of text before beginning a (Japanese) composition in Safari errors with "Lexical node does not exist in active editor state"
Lexical version:

## Steps To Reproduce

1. Open the lexical playground in Safari (desktop)
2. Select all the text and delete it
3. Switch to Japanese input
4. Press <kbd>a</kbd> followed by <kbd>Enter</kbd> to commit あ
5. Press <kbd>⌘</kbd>+<kbd>B</kbd> to set bold on the collapsed selection
6. Press <kbd>a</kbd> to begin a new composition

## The current behavior

An error is produced and the candidate window shows up with an "a" and no candidate text in the editor.

In the current version of the playground which uses the production version of Lexical, the error is:

```
Error: Minified Lexical error #69...
```

image

Using a similar setup with a dev build of Lexical, the error is:

```
Error: Lexical node does not exist in active editor state. Avoid using the same node references between nested closures from editor.read/editor.update.
```

image

## The expected behavior

It should not error. 

This appears to come about from this block:

https://github.com/facebook/lexical/blob/82bb692f221b813d9702b3c8a83aebff0788f92a/packages/lexical/src/LexicalEvents.ts#L391-L419

It looks like the reference to `node` used inside the `setTimeout` is now stale.

LuciNyan:
## Description
This bug also exists on chrome. The operation described in issue can be determined by all the conditions in the following code. This seems to be not as expected, and I wonder if this particular piece of logic should be restricted to the Android platform.

### Steps To Reproduce
Open the lexical playground in Safari (desktop)
Select all the text and delete it
Switch to Japanese input
Press a followed by Enter to commit あ
Press ⌘+B to set bold on the collapsed selection
Press a to begin a new composition

```js
if (composedText) {
      updateEditor(editor, () => {

          const anchor = selection.anchor;
          const node = anchor.getNode();
          const prevNode = node.getPreviousSibling();

          if (
            anchor.offset === 0 &&
            $isTextNode(node) &&
            $isTextNode(prevNode) &&
            node.getTextContent() === COMPOSITION_START_CHAR &&
            prevNode.getFormat() !== selection.format
          ) {
            const prevTextContent = prevNode.getTextContent();

            if (composedText.indexOf(prevTextContent) === 0) {
              const insertedText = composedText.slice(prevTextContent.length);
              dispatchCommand(
                editor,
                CONTROLLED_TEXT_INSERTION_COMMAND,
                insertedText,
              );
              setTimeout(() => {
                updateEditor(editor, () => {
                  node.select();
                });
              }, ANDROID_COMPOSITION_LATENCY);
            }
          }
        }
      });
    }
```

closes #2613


packages/shared/src/environment.ts
```diff
@@ -21,6 +21,9 @@ export const IS_APPLE: boolean =
 export const IS_FIREFOX: boolean =
   CAN_USE_DOM && /^(?!.*Seamonkey)(?=.*Firefox).*/i.test(navigator.userAgent);
 
+export const IS_ANDROID =
+  CAN_USE_DOM && /android/.test(navigator.userAgent.toLowerCase());
```
zurfyx, line 25: Nit, you could've used case insensitive search just like FF above

```suggestion
  CAN_USE_DOM && /android/i.test(navigator.userAgent);
```

zurfyx (approved):
Thanks Luci! Did you have a chance to test this condition on Android to verify if it requires further work?

LuciNyan:
> Did you have a chance to test this condition on Android to verify if it requires further work?

I don't have the suitable device to verify it. I think we need the help of an Android user.

birtles:
Android seems pretty broken to me, both with and without this change.

On Firefox for Android, the input is repeated:

https://user-images.githubusercontent.com/1232595/178377320-55a73c67-2765-492c-a5e3-deab86ffefa8.mp4

On Chrome for Android, it's not possible to commit the second string:

https://user-images.githubusercontent.com/1232595/178377482-039a0233-e4d3-4b6d-a6af-09ed41e12bd6.mp4

I see the same results, with and without this patch so I don't think this patch regresses anything.

packages/lexical/src/LexicalEvents.ts
```diff
@@ -384,7 +385,7 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
     // TODO: evaluate if this is Android only. It doesn't always seem
     // to have any real impact, so could probably be refactored or removed
     // for an alternative approach.
-    if (composedText) {
+    if (IS_ANDROID && composedText) {
```
trueadm, line 388: I wonder if we even need this logic for Android. We can probably remove it if it doesn’t help Android anymore.

LuciNyan: I removed this part of the code, but I'm a little concerned about whether it will be needed in certain boundary cases.

birtles: With the updated preview (i.e. with the code removed), following the STR from this PR now works correctly on both Firefox for Android (Beta) and Chrome for Android for me.

LuciNyan: > With the updated preview (i.e. with the code removed), following the STR from this PR now works correctly on both Firefox for Android (Beta) and Chrome for Android for me.

Thank you for your feedback!

acywatson: Awesome - thanks for the help and feedback here @birtles @LuciNyan!

I'm still waiting on an Android device to be delivered to me, but I will go ahead and merge this, going off your assertion that it is working as expected on Android.

PR #2625, opened 2022-07-11
Port ASCII State Tree to Browser Extension

noi5e:
## Features
- Fetches JSON `editorState._nodeMap` on initial app loading, and on subsequent updates.
- Renders tree `<Node>`s recursively.

https://user-images.githubusercontent.com/4361605/178168320-1aca6b43-cb8b-4325-9483-f4bce942a4cb.mov

---

### DIFF
--- a/packages/lexical-devtools/src/background/index.ts
+++ b/packages/lexical-devtools/src/background/index.ts
@@ -16,11 +16,12 @@ const tabsToPorts: Record<
 // The Lexical DevTools React UI sends a message to initialize the port.
 chrome.runtime.onConnect.addListener((port: chrome.runtime.Port) => {
   port.onMessage.addListener((message) => {
-    if (!port.sender || !port.sender.tab) {
+    if ((!port.sender || !port.sender.tab) && !message.tabId) {
+      // in the DevTools React App, port.sender is undefined, so we make an exception for messages that have tabId defined.
       return;
     }
 
-    const tabId = port.sender.tab.id;
+    const tabId = port?.sender?.tab?.id || message.tabId;
 
     if (!tabId) {
       return;
--- a/packages/lexical-devtools/src/background/index.ts
+++ b/packages/lexical-devtools/src/background/index.ts
@@ -16,14 +16,15 @@ const tabsToPorts: Record<
 // The Lexical DevTools React UI sends a message to initialize the port.
 chrome.runtime.onConnect.addListener((port: chrome.runtime.Port) => {
   port.onMessage.addListener((message) => {
-    if ((!port.sender || !port.sender.tab) && !message.tabId) {
-      // in the DevTools React App, port.sender is undefined, so we make an exception for messages that have tabId defined.
-      return;
-    }
+    let tabId;
 
-    const tabId = port?.sender?.tab?.id || message.tabId;
-
-    if (!tabId) {
+    if (port.sender && port.sender.tab) {
+      tabId = port.sender.tab.id;
+    } else if (message.tabId) {
+      // in the DevTools React App, port.sender is undefined within FROM_APP messages
+      // instead tabId is sent within message payload
+      tabId = message.tabId;
+    } else {
       return;
     }
 
--- a/packages/lexical-devtools/src/content/index.ts
+++ b/packages/lexical-devtools/src/content/index.ts
@@ -21,7 +21,11 @@ window.addEventListener('message', function (event) {
     return;
   }
 
-  if (event.data.type && event.data.type === 'FROM_PAGE') {
+  if (
+    event.data.type &&
+    event.data.type === 'FROM_PAGE' &&
+    event.data.name === 'editor-update'
+  ) {
     const nodeMap = IS_FIREFOX
       ? event.data.editorState._nodeMap
       : Object.fromEntries(event.data.editorState._nodeMap); // workaround, for some reason Object.fromEntries fails without console.error in Firefox
--- a/packages/lexical-devtools/src/content/index.ts
+++ b/packages/lexical-devtools/src/content/index.ts
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
+const IS_FIREFOX = navigator.userAgent.indexOf('Firefox') >= 0;
+
 const port = chrome.runtime.connect();
 
 port.postMessage({
@@ -20,8 +22,11 @@ window.addEventListener('message', function (event) {
   }
 
   if (event.data.type && event.data.type === 'FROM_PAGE') {
+    const nodeMap = IS_FIREFOX
+      ? event.data.editorState._nodeMap
+      : Object.fromEntries(event.data.editorState._nodeMap); // workaround, for some reason Object.fromEntries fails without console.error in Firefox
     port.postMessage({
-      editorState: event.data.editorState._nodeMap, // placeholder, sending _nodeMap for now because Chrome & Edge auto-serialize postMessages. sending the whole editorState throws a JSON serialization error due to its 'circular' JSON structure
+      editorState: {nodeMap},
       name: 'editor-update',
       type: 'FROM_CONTENT',
     });
--- a/packages/lexical-devtools/src/content/index.ts
+++ b/packages/lexical-devtools/src/content/index.ts
@@ -5,8 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-const IS_FIREFOX = navigator.userAgent.indexOf('Firefox') >= 0;
-
 const port = chrome.runtime.connect();
 
 port.postMessage({
@@ -26,11 +24,8 @@ window.addEventListener('message', function (event) {
     event.data.type === 'FROM_PAGE' &&
     event.data.name === 'editor-update'
   ) {
-    const nodeMap = IS_FIREFOX
-      ? event.data.editorState._nodeMap
-      : Object.fromEntries(event.data.editorState._nodeMap); // workaround, for some reason Object.fromEntries fails without console.error in Firefox
     port.postMessage({
-      editorState: {nodeMap},
+      editorState: event.data.editorState,
       name: 'editor-update',
       type: 'FROM_CONTENT',
     });
--- a/packages/lexical-devtools/src/panel/components/App/index.css
+++ b/packages/lexical-devtools/src/panel/components/App/index.css
@@ -9,12 +9,12 @@
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  font-size: 24px;
+  font-size: 2em;
   color: white;
 }
 
 .loading-view {
   background: #222;
   color: #fff;
-  font-size: 36px;
+  font-size: 3em;
 }
--- a/packages/lexical-devtools/src/panel/components/App/index.css
+++ b/packages/lexical-devtools/src/panel/components/App/index.css
@@ -2,26 +2,19 @@
   text-align: center;
 }
 
-.App-logo {
-  height: 40vmin;
-  pointer-events: none;
-}
-
 .App-header {
   background-color: #282c34;
-  min-height: 100vh;
+  font-family: monospace;
   display: flex;
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  font-size: calc(10px + 2vmin);
+  font-size: 24px;
   color: white;
 }
 
-.App-link {
-  color: #61dafb;
-}
-
-button {
-  font-size: calc(10px + 2vmin);
+.loading-view {
+  background: #222;
+  color: #fff;
+  font-size: 36px;
 }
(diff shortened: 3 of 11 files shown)


noi5e:
I expect there to be tons of TypeScript errors when the CI finishes... I'm going to spend the next few days brushing up on TypeScript to get past them, but I thought I'd make the PR early and get feedback on it simultaneously.

packages/lexical-devtools/src/panel/components/App/index.tsx
```diff
@@ -33,23 +34,32 @@ function App(): JSX.Element {
...
-    );
+    (port.current as any).onMessage.addListener((message) => {
+      setCount(count + 1);
+      setIsLoading(false);
+      // workaround for Firefox, message.editorState.nodeMap is either a vanilla object or a Map structure
+      const newNodeMap =
+        message.editorState.nodeMap instanceof Map
+          ? Object.fromEntries(message.editorState.nodeMap) // we need object dot notation to work TreeView, so we convert Map to Object
+          : message.editorState.nodeMap;
+      setNodeMap(newNodeMap);
```
noi5e, line 45: I found out that `editorState` has a `toJSON()` method, so this will probably be simplified a lot in the next PR.

packages/lexical-devtools/src/content/index.ts
```diff
@@ -19,9 +21,16 @@ window.addEventListener('message', function (event) {
...
+    event.data.type &&
+    event.data.type === 'FROM_PAGE' &&
+    event.data.name === 'editor-update'
+  ) {
+    const nodeMap = IS_FIREFOX
+      ? event.data.editorState._nodeMap
+      : Object.fromEntries(event.data.editorState._nodeMap); // workaround, for some reason Object.fromEntries fails without console.error in Firefox
     port.postMessage({
-      editorState: event.data.editorState._nodeMap, // placeholder, sending _nodeMap for now because Chrome & Edge auto-serialize postMessages. sending the whole editorState throws a JSON serialization error due to its 'circular' JSON structure
+      editorState: {nodeMap},
```
noi5e, line 33: See [comment](https://github.com/facebook/lexical/pull/2625/files#r917471663)

packages/lexical-devtools/src/panel/components/Node/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  type: string;
+}): JSX.Element {
+  return (
+    <li key={lexicalKey}>
+      ({lexicalKey}) {type} {text ? '"' + text + '"' : ''}
+      {children.length > 0 ? (
+        <ul>
+          {children.map((child) => (
+            <Node {...child} />
+          ))}
```
noi5e, line 28: There's some really cool [layout stuff](https://github.com/facebook/lexical/blob/30ad13bcc637dfc61cb15aaa45e3d13bcf7d9602/packages/lexical-react/src/LexicalTreeView.tsx) going on with `<pre>` in the existing `LexicalTreeView` which I hope can be implemented later.

packages/lexical-devtools/src/content/index.ts
```diff
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
+const IS_FIREFOX = navigator.userAgent.indexOf('Firefox') >= 0;
```
thegreatercurve, line 8: There's already a utility method for `IS_FIREFOX` you can import instead

packages/lexical-devtools/src/content/index.ts
thegreatercurve: Why do we need the Firefox logic if this is only initially going to be working on Chrome?

packages/lexical-devtools/src/panel/components/App/index.tsx
```diff
@@ -33,23 +34,32 @@ function App(): JSX.Element {
...
-        <p>
-          <code>editorState</code> updates: {count}
-        </p>
       </header>
+      {isLoading ? (
+        <>
+          <div className="loading-view">
+            <p>Loading...</p>
+          </div>
+        </>
```
thegreatercurve, line 59:
```suggestion
          <div className="loading-view">
            <p>Loading...</p>
          </div>
```

packages/lexical-devtools/src/panel/components/TreeView/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const depthFirstSearch = (node) => {
+    const {
+      __cachedText: cachedText,
+      __key: lexicalKey,
+      __text: text,
+      __type: type,
+    } = node;
+    const branch = {cachedText, lexicalKey, text, type};
+
+    const children = [];
```
thegreatercurve, line 32:
```suggestion
    const children: Array<NodeKey >= [];
```

packages/lexical-devtools/src/content/index.ts
noi5e: > Why do we need the Firefox logic if this is only initially going to be working on Chrome?

Oops, just to clarify:
- The app currently works in Chrome, Firefox, and Edge (manually tested everything). I use the `chrome` namespace because `chrome` works in all three browsers (even Firefox). `browser`, the other alternative, doesn't work in Chrome and Edge ([docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Chrome_incompatibilities#chrome._and_browser._namespace))
- I think I'll be able to get rid of the `IS_FIREFOX` logic completely. I didn't realize that `getEditorState()` has a `toJSON()` method, which will make this logic totally unneeded. (It's kind of a long story, but has to do with serialization differences between the different browsers) It should be a simple fix, but I'll focus on getting over the TypeScript hurdle in this PR first.

packages/lexical-devtools/src/background/index.ts
```diff
@@ -16,11 +16,12 @@ const tabsToPorts: Record<
...
 chrome.runtime.onConnect.addListener((port: chrome.runtime.Port) => {
   port.onMessage.addListener((message) => {
-    if (!port.sender || !port.sender.tab) {
+    if ((!port.sender || !port.sender.tab) && !message.tabId) {
+      // in the DevTools React App, port.sender is undefined, so we make an exception for messages that have tabId defined.
       return;
     }
 
-    const tabId = port.sender.tab.id;
+    const tabId = port?.sender?.tab?.id || message.tabId;
```
noi5e, line 24: I had to modify this after rebasing on your changes because `port.sender` is `undefined` in messages from the DevTools app. Just to summarize the locations which send messages to `background.js`:

- `content.js` 
  - the browser extension injects this content script, once per browser tab.
  - **`port.sender` is defined in messages from `content.js` (`FROM_CONTENT`)**
- `devtools.js`
  - this script runs an `eval()` statement to inject a `registerUpdateListener` into the tab (which we can't do from a content script).
  - the `eval()` statement also sets up messaging between the page and the background script.
  - `devtools.js` also sets up the actual DevTools panel, and loads the `index.html` where the React app is located
  - **`port.sender` is defined in messages from the page (`FROM_PAGE`)**
- `App.tsx`
  - this is the React app within the DevTools panel.
  - for some reason, **`port.sender` is UNDEFINED in messages from the React app (`FROM_APP`)**
  - my workaround for this is to lookup the `inspectedWindow`'s `tabId` and [post that with the message](https://github.com/MLH-Fellowship/lexical/blob/f3aafca34c33183e3cd5d70c072dca716cf2f906/packages/lexical-devtools/src/panel/components/App/index.tsx#L26-L31)

**TLDR:** The logic here has to accommodate messages `FROM_APP`, where `port.sender` is `undefined`, but `message.tabId` **IS**.

Right now I have `const tabId = port?.sender?.tab?.id || message.tabId;` as a workaround. I'm sure that's not great, but `port.sender.tab.id || message.tabId` throws a TS `possibly undefined` error. Any suggestions?

noi5e: Side note: I know the whole messaging setup is pretty complicated. 

I think it's possible to just have the page communicate with the DevTools panel, or at least remove `content.js` as the middleman. But I think both scenarios would require putting more messaging (and possible security risk?) in the `eval()` statement. Saving that as a stretch goal for now.

thegreatercurve: > TLDR: The logic here has to accommodate messages FROM_APP, where port.sender is undefined, but message.tabId IS.
> Right now I have const tabId = port?.sender?.tab?.id || message.tabId; as a workaround. I'm sure that's not great, but port.sender.tab.id || message.tabId throws a TS possibly undefined error. Any suggestions?

@noi5e can you just put it in a conditional and reassign it if only one of the cases is met else return out of the function? Something like the below:
```js
let tabId;

if (port.sender && port.sender.tab) {
    tabId = port.sender.tab.id;
} else if (message.tabId) {
    tabId = message.tabId;
} else {
    return;
}

```

thegreatercurve (changes requested):
LGTM, but just requesting changes for the TS errors.

packages/lexical-devtools/src/background/index.ts
noi5e: > @noi5e can you just put it in a conditional and reassign it if only one of the cases is met, else return out of the function? Something like the below:

Thanks, that did it! 🎊

noi5e:
@thegreatercurve Looks like we're not in bad shape here! 

Can you look at the types I made and let me know how to improve them?

packages/lexical-devtools/src/panel/components/App/index.css
```diff
@@ -2,26 +2,19 @@
...
 .App-header {
   background-color: #282c34;
-  min-height: 100vh;
+  font-family: monospace;
   display: flex;
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  font-size: calc(10px + 2vmin);
+  font-size: 24px;
```
thegreatercurve, line 12: You might as well use `ems`

```suggestion
  font-size: 2em;
```

packages/lexical-devtools/src/panel/components/App/index.css
```diff
@@ -2,26 +2,19 @@
...
-.App-link {
-  color: #61dafb;
-}
-
-button {
-  font-size: calc(10px + 2vmin);
+.loading-view {
+  background: #222;
+  color: #fff;
+  font-size: 36px;
```
thegreatercurve, line 19:
```suggestion
  font-size: 3em;
```

packages/lexical-devtools/src/panel/components/App/index.tsx
```diff
@@ -5,24 +5,27 @@
...
-import logo from '../../../images/lexical-white.png';
+import TreeView from '../TreeView';
 
 function App(): JSX.Element {
   const [count, setCount] = useState<number>(0);
-  const [, setEditorState] = useState<EditorState | null>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [nodeMap, setNodeMap] = useState<DevToolsTree | Record<string, never>>(
+    {},
+  );
```
thegreatercurve, line 21: Would this just work? 

```suggestion
  const [nodeMap, setNodeMap] = useState<DevToolsTree>({});
```

packages/lexical-devtools/types.ts
```diff
@@ -0,0 +1,19 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+export interface DevToolsTree {
+  [key: string]: DevToolsNode;
+}
+
+export interface DevToolsNode {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [x: string]: any;
```
thegreatercurve, line 14: This should be fine if we're planning on indiscriminately printing all keys and values on the node object.

thegreatercurve (approved):
LGTM @noi5e. A couple of suggested changes, but otherwise good to merge.

packages/lexical-devtools/src/panel/components/App/index.tsx
```diff
@@ -32,10 +33,13 @@ function App(): JSX.Element {
 
   useEffect(() => {
     if (port.current !== null) {
+      // message listener for editorState updates from inspectedWindow
       port.current.onMessage.addListener(
```
acywatson, line 37: Do you need to remove this listener ever?

acywatson (approved):
Nice work

packages/lexical-devtools/types.ts
noi5e: > This should be fine if we're planning on indiscriminately printing all keys and values on the node object.

Yeah, I hope so, mostly I was just copying from the existing [LexicalNode](https://github.com/facebook/lexical/blob/645227ebd0f61942d48897b03b0701a4da884508/packages/lexical/src/LexicalNode.ts#L160-L166). Can revisit later once I get a sense of what properties DevTools will be dealing with

noi5e:
> Do you need to remove this listener ever?

Oops yeah, forgot to do this. Just updated the `useEffect` hooks to return a `port.disconnect`.

After the CI passes, I think this is finally good to merge!

PR #2643, opened 2022-07-14
fix(lexical): calculate range selection formatting

### LINKED ISSUE #2626: Bug: Selecting text ending in bold and then typing produces one character without bold and the remainder in bold
Lexical version: Playground version (0.3.7?)

Testing with Firefox Nightly (104) and Safari desktop.

## Steps To Reproduce

1. Open [lexical playground](https://playground.lexical.dev/)
2. Select all the text and delete it
3. Type "Hello there"
4. Select "there" and make it bold, e.g. by pressing <kbd>Ctrl</kbd> + <kbd>B</kbd>
5. Select the whole string, "Hello there", from left to right
6. Type something, like "Hi again"

## The current behavior

"H**i again**" is produced with "H" not in bold, and "i again" in bold.

## The expected behavior

I guess none of it should be bold. That would match Word's behavior.

If that's difficult, at least making all of the typed text would be more consistent.

## Screen shots

With the text selected:

image

After typing "H" (notice we have the text node for "H" with _no_ format but the selection has `{ format: bold }`).

image

After typing "i again" (notice the selection format has been applied to the remaining text

image

I suppose we need to detect that text is being replaced and reset the selection's format to the format of the anchor node?

LuciNyan:
## Description
Fix range selection formatting.

It looks like this bug is related to the format of our rangeSelection calculation. Here our calculation is not quite the same as the google doc. As an example:

- In google doc. When we have three letters 1(italic) 2(bold) 3(underline), and we select them, the selection will show that the format is nothing at this time. Only when all textNodes have a format at the same time will the selection show that it has a format.
image

- And in lexical, our logic is that as long as the anchorNode or focusNode has a certain style, we will show that selection.format has that style.
image

- When we can calculate the selection range correctly, the problem mentioned in the issue does not exist anymore.

----
closes #2626 

https://user-images.githubusercontent.com/22126563/178920657-20ba0a72-c574-4635-ab35-472c8fca2d1b.mov

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -69,6 +69,7 @@ import {
   DOM_ELEMENT_TYPE,
   DOM_TEXT_TYPE,
   DOUBLE_LINE_BREAK,
+  IS_ALL_FORMATTING,
 } from './LexicalConstants';
 import {internalCreateRangeSelection, RangeSelection} from './LexicalSelection';
 import {updateEditor} from './LexicalUpdates';
@@ -255,17 +256,14 @@ function onSelectionChange(
           }
         }
       } else {
-        const focus = selection.focus;
-        const focusNode = focus.getNode();
-        let combinedFormat = 0;
+        let combinedFormat = IS_ALL_FORMATTING;
 
-        if (anchor.type === 'text') {
-          combinedFormat |= anchorNode.getFormat();
-        }
-
-        if (focus.type === 'text' && !anchorNode.is(focusNode)) {
-          combinedFormat |= focusNode.getFormat();
-        }
+        const nodes = selection.getNodes();
+        nodes.forEach((node) => {
+          if ($isTextNode(node)) {
+            combinedFormat &= node.getFormat();
+          }
+        });
 
         selection.format = combinedFormat;
       }
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -259,11 +259,16 @@ function onSelectionChange(
         let combinedFormat = IS_ALL_FORMATTING;
 
         const nodes = selection.getNodes();
-        nodes.forEach((node) => {
+        const nodesLength = nodes.length;
+        for (let i = 0; i < nodesLength; i++) {
+          const node = nodes[i];
           if ($isTextNode(node)) {
             combinedFormat &= node.getFormat();
+            if (combinedFormat === 0) {
+              break;
+            }
           }
-        });
+        }
 
         selection.format = combinedFormat;
       }
--- a/packages/lexical-playground/__tests__/e2e/Toolbar.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Toolbar.spec.mjs
@@ -6,11 +6,18 @@
  *
  */
 
-import {selectAll} from '../keyboardShortcuts/index.mjs';
+import {
+  selectAll,
+  selectCharacters,
+  toggleBold,
+  toggleItalic,
+  toggleUnderline,
+} from '../keyboardShortcuts/index.mjs';
 import {
   assertHTML,
   click,
   evaluate,
+  expect,
   focus,
   focusEditor,
   html,
@@ -274,5 +281,28 @@ test.describe('Toolbar', () => {
         </p>
       `,
     );
+  });
+
+  test('When we select three textNodes with different formatting at the same time, the selection formatting should show no formatting at all', async ({
+    page,
+    isPlainText,
+    isCollab,
+  }) => {
+    test.skip(isPlainText || isCollab);
+    await focusEditor(page);
+
+    await toggleBold(page);
+    await page.keyboard.type('A ');
+    await toggleBold(page);
+    await toggleItalic(page);
+    await page.keyboard.type('B ');
+    await toggleItalic(page);
+    await toggleUnderline(page);
+    await page.keyboard.type('C');
+    await selectCharacters(page, 'left', 5);
+
+    const actives = await page.$$('div.toolbar button.toolbar-item.active');
+    await page.pause();
+    expect(actives.length).toEqual(0);
   });
 });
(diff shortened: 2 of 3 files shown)


zurfyx (commented):
Thank you for digging into this Luci! While I don't disagree with the new proposed heuristic, the root of the problem is hiding in the beforeInput behavior. Notice how it still behaves oddly when doing the reverse thing `<bold>Hello</bold> world`

Also, is this GDocs heuristic? Haven't played with it extensively but my hunch is that they're using the first node instead (that is focus when selection is reversed, anchor otherwise)

LuciNyan:
> Thank you for digging into this Luci! While I don't disagree with the new proposed heuristic, the root of the problem is hiding in the beforeInput behavior. Notice how it still behaves oddly when doing the reverse thing `<bold>Hello</bold> world`
> 
> Also, is this GDocs heuristic? Haven't played with it extensively but my hunch is that they're using the first node instead (that is focus when selection is reversed, anchor otherwise)

Based on your concern, I have tried the same on onenote and microsoft word, both of which have this logic. It doesn't seem to be unique to GDocs. 
You can see that when the second row of 2 3 has the same formatting and the selection contains only 2 3, the selection shows their common formatting.

https://user-images.githubusercontent.com/22126563/178949099-08497b79-152a-4f62-8968-13e2016ab99a.mov

https://user-images.githubusercontent.com/22126563/178949174-6fe64068-f2e5-400e-94e1-b095e8573b99.mov

LuciNyan:
> Notice how it still behaves oddly when doing the reverse thing `<bold>Hello</bold> world`

This actually seems like a logical result, and I've tried it on google doc and onenote as well as microsoft. The new text will inherit the style of the top textNode in the deleted text.

https://user-images.githubusercontent.com/22126563/178950840-03188825-1a6b-45b4-b191-84dab3a83e78.mov

https://user-images.githubusercontent.com/22126563/178951473-067e9b8e-5d0e-4de0-a1d2-5ba1de28d665.mov

https://user-images.githubusercontent.com/22126563/178951618-301e554d-0822-4097-9139-960097e57ad9.mov

packages/lexical/src/LexicalEvents.ts
```diff
@@ -255,17 +256,14 @@ function onSelectionChange(
...
-
-        if (anchor.type === 'text') {
-          combinedFormat |= anchorNode.getFormat();
-        }
+        let combinedFormat = IS_ALL_FORMATTING;
 
-        if (focus.type === 'text' && !anchorNode.is(focusNode)) {
-          combinedFormat |= focusNode.getFormat();
-        }
+        const nodes = selection.getNodes();
```
acywatson, line 261: This looks like the right solution, behaviorally, but I'm a little bit concerned about putting selection.getNodes and iteration through the result in this path. This could get a bit hot if someone is drag selecting a bunch of nodes.

I know we do some caching in getNodes and I actually haven't been able to detect a practical slowdown in my environment, but I want to test this a bit more. Will circle back shortly.

@zurfyx @fantactuka do you seen any concerns here?

LuciNyan: > This looks like the right solution, behaviorally, but I'm a little bit concerned about putting selection.getNodes and iteration through the result in this path. This could get a bit hot if someone is drag selecting a bunch of nodes.
> 
> I know we do some caching in getNodes and I actually haven't been able to detect a practical slowdown in my environment, but I want to test this a bit more. Will circle back shortly.

Thank you for the reminder! I think we can break the iteration early when the combinedFormat is equal to 0. This will avoid a lot of unnecessary calculations.

zurfyx: @acywatson, @LuciNyan - Had a look over the cost. A potential optimization could be to store a cache of the result of an arbritary array of `selection.getNodes` so that we can reuse the value later if it hasn't changed.

However, this is not immediate because `selection._cachedNodes` changes at every selection changes. We would have to rebuild the logic to have a more stable array that avoids changing reference at every selection.

That said, while I think the above optimizations can be interesting performance-wise, I don't foresee a noticeable regression as part of the introduction of this logic for the majority of the users.
1. We are currently already running `selection.getNodes` on every selection change (on the Playground). The toolbar, floating toolbar and TreeView trigger it multiple times
2. This logic only runs on non-collapsed selection

I think it's worth listing down as an issue to optimize later as it can potentially yield some wins even beyond this PR (and non-collapsed selection changes) but we can likely do without them now.

packages/lexical/src/LexicalEvents.ts
```diff
@@ -255,16 +256,17 @@ function onSelectionChange(
...
-        if (anchor.type === 'text') {
-          combinedFormat |= anchorNode.getFormat();
-        }
-
-        if (focus.type === 'text' && !anchorNode.is(focusNode)) {
-          combinedFormat |= focusNode.getFormat();
+        let combinedFormat = IS_ALL_FORMATTING;
+
+        const nodes = selection.getNodes();
+        for (let i = 0; i < nodes.length; i++) {
```
zurfyx, line 262: Nit
```suggestion
        const nodesLength = nodes.length;
        for (let i = 0; i < nodeLength; i++) {
```

zurfyx (approved):
LGTM, thank you Luci, will wait until @acywatson's stamp too since he wanted to have a second pass to it

packages/lexical/src/LexicalEvents.ts
fantactuka: Wonder if it's worth adding giving it's a hot path and can be costly when selecting whole document. IIRC quip does anchor/focus check only

LuciNyan: > Wonder if it's worth adding giving it's a hot path and can be costly when selecting whole document. IIRC quip does anchor/focus check only

Or can we do the combinedFormat intersection operation at the same time as selection.getNodes?

Let's say we add a new method to selection like `getCommonFormat`, then it will do the same thing as getNodes, it will iterate over each node, but do the intersection operation at the same time. Once the result of the combinedFormat is 0, we can quit iterating. So even if the user selects the whole document, it looks like we won't be traversing all the nodes. By doing this, I think we'll be able to exit the traversal pretty quickly in most cases.

I'm wondering if you think this would be better?

acywatson:
I'm good with this - I don't see any issues at the moment. I do agree that this is an area we could probably optimize in the future.

Nice work and thanks again for the contribution!

PR #2645, opened 2022-07-14
Remove `IntentionallyMarkedAsDirtyElement` from public API

### LINKED ISSUE #1375: Avoid exposing IntentionallyMarkedAsDirtyElement
This is an internal implementation detail and really shouldn't be exposed.

thegreatercurve:


acywatson:
Can we mark this for 0.4? Technically a breaking change.

zurfyx (commented):
We're still exposing this value to public via editor update, do we want to delete it? I think the idea behind this issue is not to expose this value somehow but it won't be straightforward not that it's already exposed and used in the HistoryPlugin

thegreatercurve:
> We're still exposing this value to public via editor update, do we want to delete it? I think the idea behind this issue is not to expose this value somehow but it won't be straightforward not that it's already exposed and used in the HistoryPlugin

If this is more than just a small type fix, I'll close it until we have to look at the next major release.

PR #2650, opened 2022-07-15
Add text fallbacks to YouTubeNode and TweetNode on conversion to CodeBlock

### LINKED ISSUE #1779: Bug: CodeBlock formatting deletes custom nodes like YoutubeNode, TweetNode etc in Lexical Playground.
When we select a paragraph node and the paragraph contains a TweetNode or Poll or any kind of custom node and if we select all and try to format it with code block the custom nodes are deleted and its not rendered. 

I am attaching a video for reference. 

https://user-images.githubusercontent.com/21175404/163665718-b311a42a-8ca8-4a8f-b1f6-890a1c623eb6.mp4

I think the bug got introduced while fixing this #1644 issue. Only text data is retrieved from the selection and inserted back into the code block. 

Some of the fixes that I thought of : 
  Maybe the custom nodes can be parsed back to text and can be inserted into the code block and can be reparsed back to custom nodes again or maybe custom nodes can be ignored and only text can be formatted ( Not entirely sure how it would look for mentions and emojis ).

thegreatercurve:
https://user-images.githubusercontent.com/7748470/179281529-88300cee-385e-4876-81ae-16f82e6e7e28.mov

Closes #1779

### DIFF
--- a/packages/lexical-playground/src/nodes/TweetNode.tsx
+++ b/packages/lexical-playground/src/nodes/TweetNode.tsx
@@ -187,6 +187,11 @@ export class TweetNode extends DecoratorBlockNode {
     return this.__id;
   }
 
+  getTextContent(): string {
+    // This is called when we attempt to convert this DecoratorNode into a TextNode, such as when we convert a range selection into a CodeBlock.
+    return `https://twitter.com/i/web/status/${this.__id}`;
+  }
+
   decorate(editor: LexicalEditor, config: EditorConfig): JSX.Element {
     const embedBlockTheme = config.theme.embedBlock || {};
     const className = {
--- a/packages/lexical-playground/src/nodes/TweetNode.tsx
+++ b/packages/lexical-playground/src/nodes/TweetNode.tsx
@@ -187,11 +187,6 @@ export class TweetNode extends DecoratorBlockNode {
     return this.__id;
   }
 
-  getTextContent(): string {
-    // This is called when we attempt to convert this DecoratorNode into a TextNode, such as when we convert a range selection into a CodeBlock.
-    return `https://twitter.com/i/web/status/${this.__id}`;
-  }
-
   decorate(editor: LexicalEditor, config: EditorConfig): JSX.Element {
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/nodes/TweetNode.tsx
```diff
@@ -187,6 +187,11 @@ export class TweetNode extends DecoratorBlockNode {
     return this.__id;
   }
 
+  getTextContent(): string {
+    // This is called when we attempt to convert this DecoratorNode into a TextNode, such as when we convert a range selection into a CodeBlock.
+    return `https://twitter.com/i/web/status/${this.__id}`;
```
acywatson, line 192: This will make getTextContentSize return the length of the URL. Is that what we want? I'm looking at all the places we use getTextContent and getTextContentSize and wondering if this is as innocuous as it seems.

acywatson: Maybe the thing to do is override getTextContentSize as well.

zurfyx: @acywatson - Shouldn't `getTextContentSize` be equivalent to `getTextContent().length`? Or am I missing something? That said, if we consider ImageNode as a reference DecoratorNode I'd argue that an alt is a better fit for this `getTextContent` but personally I don't have any objections towards the proposed solution either

thegreatercurve: Thinking about it, semantically, we probably shouldn't be using the text content as a fallback value any way, especially if it potentially has some business logic usage as discussed above. 

I've implemented an easier fix where we convert the affected decorator nodes to text nodes before we attempt to insert insert everything from the selection into the new code block. 

It generally keeps the same affect.

zurfyx (approved):
No objections as a first pass to fix the bug but I wonder if we should generalize this case so that people don't have to handle this manually.

I.e does it make sense to always convert decorator nodes to text via getTextContext leveraging selection.insertRawText?

PR #2651, opened 2022-07-17
Specify the return type of getNearestNodeOfType.

y-hiraoka:
Specify the return type of `$getNearestNodeOfType`.

an example:

```ts
const parentList = $getNearestNodeOfType(anchorNode, ListNode);
const listType = parentList?.getListType();
```

Currently `listType` is of type `any`, but after this PR it will be of type `ListType | undefined`.

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -119,7 +119,7 @@ export function $getNearestNodeOfType<T extends ElementNode>(
     parent = parent.getParent();
   }
 
-  return parent;
+  return null;
 }
 
 export function $getNearestBlockElementAncestorOrThrow(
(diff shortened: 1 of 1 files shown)


zurfyx:
@y-hiraoka Looks like TS is failing. Perhaps we need a more specific selector? I.e. `T extends`

Also, can you check the Flow types and make sure they're up to date?

zurfyx:
Thanks for checking the flow types!

> So the type of `parent` is not narrowed down and the function tries to return `ElementNode | LexicalNode` when it should return null.

What about just simply returning `return null`?

y-hiraoka:
@zurfyx 

>What about just simply returning return null?

That's a great idea!

I have modified the file.

zurfyx:
> @zurfyx 
> 
> 
> 
> >What about just simply returning return null?
> 
> 
> 
> That's a great idea!
> 
> 
> 
> I have modified the file.

Thank you, LGTM!

PR #2654, opened 2022-07-18
Autolink default protocol

### LINKED ISSUE #2653: Autolink does not set http protocol
Type www.example.com, click on it, you are redirected to `http://localhost:3000/www.example.com`

zurfyx:
Closes #2653

Should probably follow up with user-entered URLs, even though these are in a better position because you can change them

### DIFF
--- a/packages/lexical-playground/src/plugins/AutoLinkPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/AutoLinkPlugin.tsx
@@ -26,7 +26,7 @@ const MATCHERS = [
       index: match.index,
       length: fullMatch.length,
       text: fullMatch,
-      url: fullMatch.startsWith('http') ? fullMatch : `http://${fullMatch}`,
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/plugins/AutoLinkPlugin.tsx
```diff
@@ -18,14 +18,16 @@ const EMAIL_MATCHER =
...
-    );
+    if (match === null) {
+      return null;
+    }
+    const fullMatch = match[0];
+    return {
+      index: match.index,
+      length: fullMatch.length,
+      text: fullMatch,
+      url: fullMatch.startsWith('http') ? fullMatch : `http://${fullMatch}`,
```
hanford, line 29: I wonder if `https://` is a better assumption. Don't feel strongly but very rarely see `http://` these days

acywatson: Yea, that probably makes sense.

thegreatercurve (approved):

PR #2655, opened 2022-07-19
fix(doc): RichTextPlugin placeholder

unvalley:
Thank you for amazing work.

The `RichTextPlugin` currently accepts a placeholder as `JSX.Element | string`. 
But, docs are still documented as `null`. This PR fixes it.

https://github.com/facebook/lexical/blob/48bfe0809b4ea74ef09dbece193e8d1cc1f638d1/packages/lexical-react/src/LexicalRichTextPlugin.tsx#L30


acywatson:
Looks good! Sign the CLA and we'll get it merged!

acywatson (approved):

unvalley:
Thanks! I've signed CLA and expect the check will run again.

PR #2662, opened 2022-07-20
chore: remove files that should not be submitted

LuciNyan:
## Description
remove .swp file. It looks like this file was not supposed to be submitted


thegreatercurve:
Thanks @LuciNyan!

PR #2672, opened 2022-07-21
fix(lexical-list): remove list breaks if selection in empty

### LINKED ISSUE #2640: Bug: playground remove list breaks if selection in empty nested list item
Clicking current block type in toolbar supposed to convert selected block to paragraph. For lists, if you try to do this while the selection is in an empty nested list item, we get an Error code #19 (selection has been lost). I believe it's because of a bug in the REMOVE_LIST_COMMAND.

Lexical version: playground

## Steps To Reproduce

1. Go to playground
2. Create a nested list item in the preloaded content's bullet list
3. Click on bullet list in toolbar

### Works

https://user-images.githubusercontent.com/34961840/178792075-d8ac1005-9d13-4ec7-ac22-32d1214bcaa5.mov

### Fails

https://user-images.githubusercontent.com/34961840/178790727-9bc91c69-4704-434e-9004-9b9f18f71e89.mov

LuciNyan:
## Description
When the anchor and focus of our selection fall on the textNode, we don't have to change the selection because the textNode will be appended to the newly generated paragraph.

When selection is in empty nested list item, selection is actually on the listItemNode. When the corresponding listItemNode is deleted and replaced by the newly generated paragraph, we should manually set the selection's focus and anchor to the newly generated paragraph.

closes #2640

https://user-images.githubusercontent.com/22126563/180176157-f72e29ce-4cfc-4a70-8dc4-43719f7be7e1.mov

### DIFF
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -238,6 +238,14 @@ export function removeList(editor: LexicalEditor): void {
             insertionPoint.insertAfter(paragraph);
             insertionPoint = paragraph;
 
+            if (listItemNode.__key === selection.anchor.key) {
+              selection.anchor.set(paragraph.getKey(), 0, 'element');
+            }
+
+            if (listItemNode.__key === selection.focus.key) {
+              selection.focus.set(paragraph.getKey(), 0, 'element');
+            }
+
             listItemNode.remove();
           }
         });
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -238,6 +238,12 @@ export function removeList(editor: LexicalEditor): void {
             insertionPoint.insertAfter(paragraph);
             insertionPoint = paragraph;
 
+            // When the anchor and focus fall on the textNode
+            // we don't have to change the selection because the textNode will be appended to
+            // the newly generated paragraph.
+            // When selection is in empty nested list item, selection is actually on the listItemNode.
+            // When the corresponding listItemNode is deleted and replaced by the newly generated paragraph
+            // we should manually set the selection's focus and anchor to the newly generated paragraph.
             if (listItemNode.__key === selection.anchor.key) {
               selection.anchor.set(paragraph.getKey(), 0, 'element');
             }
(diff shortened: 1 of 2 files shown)


packages/lexical-list/src/formatList.ts
```diff
@@ -238,6 +238,20 @@ export function removeList(editor: LexicalEditor): void {
...
+            // the newly generated paragraph.
+            // When selection is in empty nested list item, selection is actually on the listItemNode.
+            // When the corresponding listItemNode is deleted and replaced by the newly generated paragraph
+            // we should manually set the selection's focus and anchor to the newly generated paragraph.
+            if (listItemNode.__key === selection.anchor.key) {
+              selection.anchor.set(paragraph.getKey(), 0, 'element');
+            }
+
+            if (listItemNode.__key === selection.focus.key) {
+              selection.focus.set(paragraph.getKey(), 0, 'element');
```
zurfyx, line 252: This selection is not normalized, see $normalizeSelection

zurfyx (commented):
Thanks for taking a stab on this! Seems like the problem is deeper than a selection issue.

First of, note that `.remove` recovers the selection automatically when the selection is correctly placed. Suspect bad behavior when this does not happen.

Second, the behavior you described in the test is probably not what we're after, see how GDocs does it. It moves the list item node as a paragraph item.

Third, I would change a few things on the implementation:
1. Remove items should work on the current item list node rather than trying to reorganize all the nodes
Screen Shot 2022-07-21 at 5 39 45 pm Why are we looking at all the top nodes?
3. I'm suspicious about the correctness of splice, probably worth revisiting selection logic there to make sure it's being recovered gracefully
4. forEach is not recommended

Let me know if you want me to help with this

cc @acywatson

zurfyx (approved):
Your fix is correct! Added more details on the linked PR but we can merge this.

(As for the fist point about the behavior I think we should revisit this, not sure this behavior is very desirable but your fix addresses the bug so it's correct)

LuciNyan:
> Added more details on the linked PR but we can merge this.

Thank you @zurfyx ! That's awesome! I learned a lot from your PR.  You've solved a big potential problem. <3

PR #2673, opened 2022-07-21
Separate `@lexical/code` into more atomic modules

thegreatercurve:


packages/lexical-code/src/index.ts
```diff
@@ -6,1206 +6,6 @@
...
-    editor.registerCommand(
-      MOVE_TO_START,
-      (payload): boolean => handleMoveTo(MOVE_TO_START, payload),
-      COMMAND_PRIORITY_LOW,
-    ),
-  );
-}
+export * from './CodeHighlighter';
+export * from './CodeHighlightNode';
+export * from './CodeNode';
```
zurfyx, line 11: Should we use named export rather than `*`? It did lead to accidentally exported modules in the past

zurfyx (approved):

packages/lexical-code/src/index.ts
acywatson: Agree

acywatson (approved):
Fine with this, but curious what prompted it.

thegreatercurve:
> Fine with this, but curious what prompted it.

A change that Lateef had made. This rebases some of his changes and updates them with recent changes to lexical/code

PR #2674, opened 2022-07-21
chore(lexical-playground): make plugin directory clear

LuciNyan:
## Description

As the number of plugins grew, xxx.css and xxx.tsx files were mixed together under lexical-playground/src/plugins. This pr just groups the css and tsx files into their respective folders according to the plugin.


acywatson:
> LGTM, I wonder what other folks think about this @thegreatercurve @acywatson

I actually prefer it this way - nested directories are pretty cheap and this encourages good code organization for contributors.

One thing that does bother me is relative import paths, but we can fix that later :)

PR #2675, opened 2022-07-21
Add E2E test for TableOfContentsPlugin

karam-qaoud:


packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -385,6 +389,16 @@ export async function getHTML(page, selector = 'div[contenteditable="true"]') {
   return element.innerHTML();
 }
 
+export async function getElement(
```
thegreatercurve, line 392: @karam-qaoud Can you not just use `getEditorElement` to get your element? The logic seems basically the same

thegreatercurve (commented):
Might need to delete the `packages/lexical-playground/src/plugins/.TableOfContentsPlugin.tsx.swp` file that keeps being committed, or add it to the .gitignore.

packages/lexical-playground/__tests__/utils/index.mjs
karam-qaoud: getEditorElement gets you the editor itself, but this method gets you any element within the editor/page.

thegreatercurve: Can we refactor `getEditorElement` to use `getElement` then? Keeps things more DRY.

thegreatercurve (approved):

zurfyx (approved):

thegreatercurve:
@karam-qaoud Does this fix the page freezing issue

karam-qaoud:
@thegreatercurve Yes, `packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx` has the fix.

PR #2677, opened 2022-07-21
Resolve selection for orphan selected children

zurfyx:
PR https://github.com/facebook/lexical/pull/2672 revealed a bigger issue in core. When the children have selection and the parent is delete the selection is not moved to the nearest parent node.

While this doesn't fix the specific list issue (@LuciNyan's fix is correct) this should fix the underlying issue that can arise in similar situations, especially programatically.

https://user-images.githubusercontent.com/193447/180285979-9dca37ab-d5fa-4b5d-8060-51b52f94aa67.mov


packages/lexical/src/LexicalUtils.ts
```diff
@@ -1176,3 +1176,37 @@ export function $addUpdateTag(tag: string): void {
...
+  if ($hasParent(anchorNode, parentNode)) {
+    anchor.set(parentNode.__key, 0, 'element');
+  }
+  if ($hasParent(focusNode, parentNode)) {
+    focus.set(parentNode.__key, 0, 'element');
+  }
+  return selection;
+}
+
+function $hasParent(child: LexicalNode, targetNode: LexicalNode): boolean {
```
acywatson, line 1203: Is this more like hasAncestor?

acywatson: Not part of the public API, so no big deal

packages/lexical/src/__tests__/utils/index.tsx
```diff
@@ -401,3 +408,12 @@ export function createTestEditor(
   });
   return editor;
 }
+
+export function $assertRangeSelection(selection): RangeSelection {
+  expect($isRangeSelection(selection)).toBeTruthy();
+  if ($isRangeSelection(selection)) {
+    return selection;
+  } else {
+    throw new Error('??');
```
acywatson, line 417: wat

acywatson (approved):
Couple nits, but merge-at-will

packages/lexical/src/__tests__/utils/index.tsx
zurfyx: I'll add an invariant like a normal person :D

PR #2679, opened 2022-07-22
Collapse and Expand DevTools Tree Nodes

noi5e:
DevTools nodes now accordion open and closed.

https://user-images.githubusercontent.com/4361605/180367895-56839bad-2669-4000-90dc-86a2bf59f295.mov

---

### DIFF
--- a/packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
+++ b/packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
@@ -7,31 +7,40 @@
  */
 import {DevToolsNode} from 'packages/lexical-devtools/types';
 import * as React from 'react';
+import {Fragment} from 'react';
 
 function TreeNode({
-  lexicalKey,
   __text,
   __type,
   children,
-}: {
-  children: Array<DevToolsNode>;
-  lexicalKey: string;
-  __text?: string;
-  __type: string;
-}): JSX.Element {
+  nestingLevel,
+  lexicalKey,
+}: DevToolsNode): JSX.Element {
+  const indentation = '  '.repeat(nestingLevel);
+  const marker = children.length > 0 ? '▼ ' : '  ';
+  const nodeString = ` (${lexicalKey}) ${__type} ${
+    // ▶ right unicode
+    __text ? '"' + __text + '"' : ''
+  }`;
+  const childNodes =
+    children.length > 0 ? (
+      <Fragment>
+        {children.map((child) => (
+          <TreeNode {...child} />
+        ))}
+      </Fragment>
+    ) : (
+      ''
+    );
+
   return (
-    <li key={lexicalKey}>
-      ({lexicalKey}) {__type} {__text ? '"' + __text + '"' : ''}
-      {children.length > 0 ? (
-        <ul>
-          {children.map((child) => (
-            <TreeNode {...child} />
-          ))}
-        </ul>
-      ) : (
-        ''
-      )}
-    </li>
+    <Fragment key={lexicalKey}>
+      {indentation}
+      {marker}
+      {nodeString}
+      {'\n'}
+      {childNodes}
+    </Fragment>
   );
 }
 
--- a/packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
+++ b/packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
@@ -11,18 +11,18 @@ import {DevToolsNode} from 'packages/lexical-devtools/types';
 import * as React from 'react';
 import {Fragment, useState} from 'react';
 
-import Marker from '../Marker';
+import Chevron from '../Chevron';
 
 function TreeNode({
   __text,
   __type,
   children,
-  nestingLevel,
+  depth,
   lexicalKey,
 }: DevToolsNode): JSX.Element {
   const [isExpanded, setIsExpanded] = useState(true);
 
-  const handleMarkerClick = () => {
+  const handleChevronClick = () => {
     setIsExpanded(!isExpanded);
   };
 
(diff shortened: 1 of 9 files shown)


packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
```diff
@@ -5,33 +5,55 @@
...
-            <TreeNode {...child} />
-          ))}
-        </ul>
+        <Chevron
+          handleClick={handleChevronClick}
+          icon={isExpanded ? '▼' : '▶'}
+        />
       ) : (
-        ''
+        <button className="indentation">▶</button>
```
noi5e, line 51: CSS hack here for the sake of spacing. This particular `button.indentation` element appears invisible.

The existing non-DevTools [TreeView](https://github.com/facebook/lexical/blob/30ad13bcc637dfc61cb15aaa45e3d13bcf7d9602/packages/lexical-react/src/LexicalTreeView.tsx) is formatted as a `<pre>` element.

Devtools's TreeView is formatted as nested `<div>`s, so I'm still working on getting the styling to be perfect. Work in progress.

packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
```diff
@@ -5,33 +5,55 @@
...
+    <div className="tree-node" key={lexicalKey}>
       {children.length > 0 ? (
-        <ul>
-          {children.map((child) => (
-            <TreeNode {...child} />
-          ))}
-        </ul>
+        <Chevron
+          handleClick={handleChevronClick}
+          icon={isExpanded ? '▼' : '▶'}
```
zurfyx, line 48: Prefer the Unicode equivalent instead for the arrows

zurfyx (approved):

zurfyx:
LGTM!

packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
```diff
@@ -5,33 +5,55 @@
...
+  const nodeString = ` (${lexicalKey}) ${__type} ${
+    __text ? '"' + __text + '"' : ''
+  }`;
+  const childNodes =
+    children.length > 0 ? (
+      <Fragment>
+        {children.map((child) => (
+          <TreeNode {...child} />
+        ))}
+      </Fragment>
```
thegreatercurve, line 38:
```suggestion
      <>
        {children.map((child) => (
          <TreeNode {...child} />
        ))}
      </>
```

packages/lexical-devtools/src/panel/components/TreeView/index.css
```diff
@@ -4,14 +4,10 @@
   color: #fff;
   margin: 0;
   padding: 1em;
-  font-size: 1em;
-  overflow: auto;
+  font-size: 1.2em;
+  overflow-x: scroll;
```
thegreatercurve, line 8: Why should this always show a scroll bar? What if the editor has nothing in it?

packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
thegreatercurve: You can probably just put the ternary in the `Chevron` component as well, and use an `isExpanded` prop here instead.

thegreatercurve:
Nic work @noi5e!

thegreatercurve (approved):

packages/lexical-devtools/src/panel/components/TreeView/index.tsx
```diff
@@ -19,29 +19,35 @@ function TreeView({
...
+    node: DevToolsNode,
+    depth: number,
+  ): DevToolsNode => {
     const children: Array<DevToolsNode> = [];
 
     if (Object.prototype.hasOwnProperty.call(node, '__children')) {
       node.__children.forEach((childKey: string) => {
         const child = nodeMap[childKey];
-        children.push(depthFirstSearch(child));
+        children.push(depthFirstSearch(child, depth + 1)); // recursive call
```
acywatson, line 32: Not sure the comment adds anything here

packages/lexical-devtools/src/panel/components/TreeView/index.tsx
```diff
@@ -19,29 +19,35 @@ function TreeView({
   viewClassName: string;
   nodeMap: DevToolsTree;
 }): JSX.Element {
-  const depthFirstSearch = (node: DevToolsNode): DevToolsNode => {
+  // takes flat JSON structure, nests child comments inside parents
+  const depthFirstSearch = (
+    node: DevToolsNode,
+    depth: number,
```
acywatson, line 25: Should this default to 0? Is there a case where you'd want to start at something other than zero other than in the implementation? Probably not a big deal since this isn't exported or anything.

noi5e:
Pushed some new commits to take care of everyone's comments including this one:

> Should this default to 0? Is there a case where you'd want to start at something other than zero other than in the implementation? Probably not a big deal since this isn't exported or anything.

I rewrote `depthFirstSearch` with default parameters, and also passing in `nodeMap` as an additional parameter. Seems like a better piece of code, thanks for the suggestion:

```
  const depthFirstSearch = (
    map: DevToolsTree = nodeMap,
    nodeKey = 'root',
    depth = 0,
  ): DevToolsNode => {
    const node = map[nodeKey];
    const children: Array<DevToolsNode> = [];

    if (Object.prototype.hasOwnProperty.call(node, '__children')) {
      node.__children.forEach((childKey: string) => {
        children.push(depthFirstSearch(map, childKey, depth + 1));
      });
    }

    return {
      ...node,
      __type: node.__type,
      children,
      depth,
      lexicalKey: node.__key,
    };
  };
```

noi5e:
This seems merge-ready, just failing some unrelated e2e tests in lexical-playground. This is the based on the most recent `main` as far as I know...

PR #2683, opened 2022-07-22
fix: path to icons

im-adithya:
The icons related to comments are missing https://playground.lexical.dev/; this PR aims to fix that
Screenshot from 2022-07-22 17-48-57
.


zurfyx (approved):
Thanks!

PR #2686, opened 2022-07-22
Lexical Features

zurfyx:
# Lexical Features

With the introduction of rich text late 2021, we have gone a long way to have full support for the common and not so common rich text features to mimic the most complex and complete text editors like GDocs and MS Word.

However, while a full composer editor would benefit from most/all of these, smaller composers prefers a reduced set of functionality. For example, no tables or strikethrough text.

The features that are provided via self-contained plugins are easy to toggle but the ones that come bundled into core nodes or into other multi-purpose plugins are more complicated to toggle. For example, strikethrough or other TextNode formatting. These are only possible to control via transforms.

Transforms are not very convenient. Not only they are verbose to type but also they can potentially break with the introduction of new Lexical versions. For example, if the next version of Lexical were to introduce a new type of format. 

> Note that the formatting can come from UI (toolbar), keyboard shortcuts or copy-paste.

## Features proposal

This new features flag would allow developers to control the node behavior and would be up to the owner to maintain and support toggling new features.

```
nodes: [
  nodeRegistration(TextNode, {
    features: ['strikethrough'],
  }),
  HeadingNode,
]
```

The optional features array and node registration allows the developer to safely control and rollout new features.

Then the developer can control this feature on the node via the 'lexical' exposed `hasFeature`

```
  setFormat(format: TextFormatType | number): this {
    errorOnReadOnly();
    self.__format =
      typeof format === 'string' ? TEXT_TYPE_TO_FORMAT[format] : format;
    if (
      (self.__format & TEXT_TYPE_TO_FORMAT.strikethrough) !== 0 &&
      !hasFeature(this.constructor, 'strikethrough')
    ) {
      console.warn(`LexicalTextNode feature disabled: strikethrough`);
      self.__format ^= IS_STRIKETHROUGH;
    }
    return self;
  }
```

### DIFF
--- a/packages/lexical-playground/src/nodes/PlaygroundNodes.ts
+++ b/packages/lexical-playground/src/nodes/PlaygroundNodes.ts
@@ -6,8 +6,6 @@
  *
  */
 
-import type {Klass, LexicalNode} from 'lexical';
-
 import {CodeHighlightNode, CodeNode} from '@lexical/code';
 import {HashtagNode} from '@lexical/hashtag';
 import {AutoLinkNode, LinkNode} from '@lexical/link';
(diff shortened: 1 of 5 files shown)


acywatson:
Could we do this with a plugin that intercepts and no-ops the text formatting command?

acywatson:
Closing as stale - feel free to reopen if we need to continue the discussion here.

Issue #2671, opened 2022-07-21
Feature: Add target attribute to LinkNode


andrewchmr:
It would be great to support `target` for `LinkNode` attribute. It is currently impossible to specify if a link should be open in a new tab.

Example of link with `target="_blank"`:
```
<a href="https://lexical.dev" target="_blank" rel="noreferrer noopener"><span>Lexical</span></a>
```

Example of wanted implementation:

```
export class LinkNode extends ElementNode {
  __url: string;
  __target?: string;
  
  constructor(url: string, target?: string, key?: NodeKey) {
    super(key);
    this.__url = url;
    this.__target = target;
  }
  
  createDOM(config: EditorConfig): HTMLAnchorElement {
    const element = document.createElement('a');
    element.href = this.__url;

    if (this.__target === '_blank') {
      element.target = this.__target;
      element.rel = 'noreferrer noopener';
    }

    addClassNamesToElement(element, config.theme.link);

    return element;
  }
  
  ...
```

andrewchmr:
> @andrewchmr This seems like a sensible suggestion to me, as well as possibly adding add `rel` as a key on the node itself. Want to open a PR and add the above code? We can review it pretty quickly.

Sure, I will open a PR 🙂

PR #2687, opened 2022-07-22
feat: Link node with target and rel

andrewchmr:
Issue: https://github.com/facebook/lexical/issues/2671

Support `target` and `rel` for LinkNode

Ideas for the future:
- Add an option in the [playground](https://playground.lexical.dev/) if link should be opened in the new tab

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -35,7 +35,7 @@ export type SerializedLinkNode = Spread<
     type: 'link';
     url: string;
     target?: string;
-    rel?: string;
+    relationship?: string;
     version: 1;
   },
   SerializedElementNode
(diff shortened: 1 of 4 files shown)


packages/lexical-link/src/index.ts
```diff
@@ -111,10 +137,28 @@ export class LinkNode extends ElementNode {
...
+  setTarget(target: string | undefined): void {
+    const writable = this.getWritable();
+    writable.__target = target;
+  }
+
+  getRel(): string | undefined {
+    return this.getLatest().__rel;
+  }
+
+  setRel(rel: string | undefined): void {
```
thegreatercurve, line 153: Can we name this to be something more explicit, like `setRelationship`?

thegreatercurve:
LGTM @andrewchmr! It just seems to be breaking a few tests to do with the Autolink node.

andrewchmr:
> LGTM @andrewchmr! It just seems to be breaking a few tests to do with the Autolink node.

I fixed tests with `Autolink`, but there are 2 failing tests that don't seem to be related to my changes 🤔

image

thegreatercurve:
@andrewchmr This is weird, I can't actually seeing any tests failing in the CI, especially the unit tests. Can you try rebasing onto main and then force pushing an update to trigger another CI run? It might be some weird caching issue with Github Actions

thegreatercurve:
The failing unit tests are due to a the coverage reports returning errors. Given that we don't use the coverage reports, we can probably just delete them: https://github.com/facebook/lexical/pull/2699

thegreatercurve (approved):
LGTM @andrewchmr! Thanks for making these backwards compatible and adding in the tests.

thegreatercurve:
Closes #2671

fameoflight:
Any ideas on how can I enable target="_blank" as default for any link

giokallis:
Is there anything new with @fameoflight's proposal ? I also don't understand how to use the target in LinkNode!

tylerklose:
> Is there anything new with @fameoflight's proposal ? I also don't understand how to use the target in LinkNode!

I tried achieving this by using the following, but have had no luck:
```javascript
  const URL_MATCHER =
    /((https?:\/\/(www\.)?)|(www\.))[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)/

  const MATCHERS = [
    (text: string) => {
      const match = URL_MATCHER.exec(text)
      return (
        match && {
          index: match.index,
          length: match[0].length,
          text: match[0],
          url: match[0],
          attributes: { rel: "noopener", target: "_blank" }
        }
      )
    },
  ]
  
  ...
  
  <AutoFocusPlugin />
  ```

Issue #1282, opened 2022-02-12
Feature: Add PrettierPlugin for code block formatting


trueadm:
We probably need the API to detect nodes being removed/added before we can do this, but it would great to have a plugin that adds an auto format button to code blocks so we can format the content. This was an internally requested feature too.

LuciNyan:
We now have a code block that supports many languages, I wonder if we need to introduce a prettier parser for each language when we implement this feature. If we support more languages later, do we need to add the corresponding parser as well?

PR #2688, opened 2022-07-22
feat(lexical-playground): prettier code

LuciNyan:
## Description
There are some languages that do not have official support for the prettier plugin, so the `prettier` button will not be displayed.

If there is a syntax error, the user should be given some feedback of the error. I wondering what is the best form of feedback we should give?

closes: #1282 

https://user-images.githubusercontent.com/22126563/180471432-1db5e9d1-7bb8-4c41-a50b-be3dd0a882b0.mov

when syntax error

https://user-images.githubusercontent.com/22126563/180592974-6ad0fb48-5325-4482-924d-bb4be7f8214a.mov


zurfyx:
> I wondering what is the best form of feedback we should give?

VSCode does nothing, you could potentially show some non-intrusive text, open to ideas!

LuciNyan:
> VSCode does nothing, you could potentially show some non-intrusive text, open to ideas!

Thank you for your quick reply! <3 

Wondering if other folks have any good ideas. 🤔

acywatson:
> > VSCode does nothing, you could potentially show some non-intrusive text, open to ideas!
> 
> Thank you for your quick reply! <3
> 
> Wondering if other folks have any good ideas. 🤔

I mean, I think ideal scenario might be something like the squiggly underlines that show a popover on mouseenter. Not sure how feasible that is.

A good first step might be an additional icon that shows up in the codeblock on error and, when hovered, shows a popover with line number and error message.

LuciNyan:
> A good first step might be an additional icon that shows up in the codeblock on error and, when hovered, shows a popover with line number and error message.

Thanks, this idea looks great!

> I mean, I think ideal scenario might be something like the squiggly underlines that show a popover on mouseenter. Not sure how feasible that is.

I really like the effect! There are currently some issues with the line number display in the code block though, I'll get to this after I fix that one.

LuciNyan:
I added the error message. It's ready for review now.

packages/lexical-code/src/CodeHighlightNode.ts
```diff
@@ -72,6 +72,10 @@ export const CODE_LANGUAGE_MAP: Record<string, string> = {
   text: 'plain',
 };
 
+export function normalizeCodeLang(lang: string) {
+  return CODE_LANGUAGE_MAP[lang] || lang;
+}
+
 export function getLanguageFriendlyName(lang: string) {
   const _lang = CODE_LANGUAGE_MAP[lang] || lang;
```
thegreatercurve, line 80:
```suggestion
  const _lang = normalizeCodeLang(lang);
```

packages/lexical-playground/src/plugins/CodeActionMenuPlugin/components/PrettierButton/index.css
```diff
@@ -0,0 +1,15 @@
...
+  position: relative;
+}
+
+.code-action-menu-container .prettier-wrapper .code-error-tips {
+  padding: 5px;
+  border-radius: 4px;
+  color: #fff;
+  background: #222;
+  margin-top: 4px;
+
```
thegreatercurve, line 11:
```suggestion
```

thegreatercurve (approved):
LGTM! Thanks again @LuciNyan. Great work!

packages/lexical-code/src/CodeHighlightNode.ts
LuciNyan: Have fixed it 🚀

PR #2689, opened 2022-07-22
fix(lexical-playground): LexicalTypeaheadMenuPlugin import

elv1n:
Use the correct import path that available in NPM package


thegreatercurve (approved):

thegreatercurve:
Thanks @elv1n!

PR #2694, opened 2022-07-23
fix(playground): fix rendering Exclidraw

ImSingee:
This PR is nearly the same as #2174.

The Exclidraw feature isn't available in the current version, and this PR fixes it.

The reason is that in https://github.com/facebook/lexical/pull/2496/files#diff-f45027bac56b3f85c3bec26a2889fb714d521d060db7b1d9554ff6d1fc3421efR88 , the change in #2174 is reverted and the bug appears again.


acywatson:
Oh, wow - nice catch. Really, really appreciate you digging in to find and fix this!

thegreatercurve (approved):

PR #2697, opened 2022-07-25
Changed type of includeHeaders parameter

alinamusuroi:
Changed type for includeHeaders parameter from string to boolean to match the type of the parameter from the $createTableNodeWithDimensions function.


thegreatercurve (approved):
Thanks @alinamusuroi!

PR #2700, opened 2022-07-25
fix: collapse block on deleting at beginning

im-adithya:
Removes the block on removing character instead of merging in above elementNode:

https://user-images.githubusercontent.com/64399555/180785206-a5d1e8c5-3453-447b-be3e-63f575971b2c.mp4

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1736,6 +1736,32 @@ export class RangeSelection implements BaseSelection {
         if ($isElementNode(nextSibling) && !nextSibling.canExtractContents()) {
           return;
         }
+
+        if (
+          anchorNode.getNextSibling() === null &&
+          $isElementNode(nextSibling) &&
+          nextSibling.collapseAtStart(this)
+        ) {
+          return;
+        }
+      }
+
+      if (isBackward && anchor.offset === 0) {
+        let element;
+        if (
+          anchor.type === 'text' &&
+          anchorNode.getPreviousSibling() === null
+        ) {
+          element = anchorNode.getParentOrThrow();
+        }
+
+        if (anchor.type === 'element') {
+          element = anchorNode;
+        }
+
+        if (element && element.collapseAtStart(this)) {
+          return;
+        }
       }
       this.modify('extend', isBackward, 'character');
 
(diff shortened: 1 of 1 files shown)


acywatson:
Sorry for the delay here - I have been thinking about whether this is the behavior we want. Generally, it looks like things don't work this way in other editors, such as Quip or Google Docs. It also seems to cause problems with Images (maybe other things? I only tested images).

acywatson:
Does anyone else have thoughts on this? @fantactuka @zurfyx @thegreatercurve @karam-qaoud @tylerjbainbridge

zurfyx:
Sorry for the delay and thank you for your proposal. I am sort on @acywatson's side on GDocs/Quip respect, while we don't necessarily have to stick to the same patterns, standarization when possible it is a positive thing (as long as they have the proper a11y/UX requirements).

Besides that, undo-redo doesn't seem to work properly, notice that the text is being dropped and it also won't work when the parent Element is not immediately the top Element (for example, try wrapping it in a mark node by commenting on it).

im-adithya:
> Generally, it looks like things don't work this way in other editors, such as Quip or Google Docs

I was thinking along the lines of text input component (similar to what we see in slack and discord)

> It also seems to cause problems with Images (maybe other things? I only tested images).

Oops! Will fix that, but would we want this?

acywatson:
I don't think this is behavior we want by default, but I wonder whether it's possible to add this behavior somehow through the public API rather than through modifying the core. I guess you would have to intercept the commands that call deleteCharacter and change the behavior there. Theoretically could b done in a plugin, I think?

im-adithya:
Just an idea:
Maybe we can set a flag like `isChatBox` for people who intend to use Lexical in their chat apps' message-sending-box for message formatting (like Meta is currently using in WhatsApp if I'm not wrong) and add this feature there? Not just this, but we can integrate some more features necessary for chat-textboxes rather than text editors, for example, emojis like these `:emojis:`

trueadm:
Couldn't you just add a custom `collapseAtStart` to your node that represents your chat element and in there prevent the default behavior?

acywatson:
> Couldn't you just add a custom `collapseAtStart` to your node that represents your chat element and in there prevent the default behavior?

Yes, let's close this one out.

PR #2704, opened 2022-07-26
Add LexicalAutoEmbedPlugin and (Playground)AutoEmbedPlugin.

tylerjbainbridge:
This PR lets you easily add `EmbedConfig`s for all of your app's embedded content which will act as instructions whether the entry point is a `ComponentPicker`, a `Toolbar`, or from the `AutoEmbedMenu`, it's all one config.

See what it's like to build a new embed using an `EmbedConfig` in #2705 which adds the `FigmaEmbed` seen in action below.

https://user-images.githubusercontent.com/13852400/180896879-df6cef7a-cea3-42ee-84fb-0f146feb086b.mov

```jsx
// Extend the base EmbedConfig to add things like icons, descriptions, etc.
interface PlaygroundEmbedConfig extends EmbedConfig {
  // Human readable name of the embeded content e.g. Tweet or Google Map.
  contentName: string;

  // Icon for display.
  icon?: JSX.Element;

  // An example of a matching url https://twitter.com/jack/status/20
  exampleUrl: string;

  // For extra searching.
  keywords: Array<string>;

  // Embed a Figma Project.
  description?: string;
}
```

```jsx
// The FigmaEmbed will be added in a follow up PR, but here is the config.
export const FigmaEmbedConfig: PlaygroundEmbedConfig = {
  ...
  insertNode: (editor: LexicalEditor, result: EmbedMatchResult) => {
    editor.dispatchCommand(INSERT_FIGMA_COMMAND, result.id);
  },
  parseUrl: (text: string) => {
    const match =
      /https:\/\/([\w.-]+\.)?figma.com\/(file|proto)\/([0-9a-zA-Z]{22,128})(?:\/.*)?$/.exec(
        text,
      );

    if (match != null) {
      return {
        id: match[3],
        url: match[0],
      };
    }

    return null;
  },
};
```


packages/lexical-history/src/index.ts
```diff
@@ -238,7 +238,7 @@ function createMergeActionGetter(
 
     // If applying changes from history stack there's no need
     // to run history logic again, as history entries already calculated
-    if (tags.has('historic')) {
+    if (tags.has('history-undo') || tags.has('history-redo')) {
```
fantactuka, line 241: This is a breaking change as third party plugins can rely on historic tag

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -153,7 +155,7 @@ function getFullMatchOffset(
 }
 
 /**
- * Split Lexica TextNode and return a new TextNode only containing matched text.
+ * Split Lexica; TextNode and return a new TextNode only containing matched text.
```
acywatson, line 158: This doesn't seem quite right

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -42,7 +44,7 @@ export type QueryMatch = {
 
 export type Resolution = {
   match: QueryMatch;
-  range: Range;
```
acywatson, line 45: Why are we doing this?

packages/lexical-rich-text/src/index.ts
```diff
@@ -385,6 +385,7 @@ function onPasteForRichText(
       clipboardData != null &&
       ($isRangeSelection(selection) || $isGridSelection(selection))
     ) {
+      editor._updateTags.add('paste');
```
acywatson, line 388: can we use update options?

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
tylerjbainbridge: Lol

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -116,11 +116,8 @@ export function $insertDataTransferForRichText(
     try {
       const parser = new DOMParser();
       const dom = parser.parseFromString(htmlString, 'text/html');
-      return $insertGeneratedNodes(
-        editor,
-        $generateNodesFromDOM(editor, dom),
-        selection,
-      );
+      const nodes = $generateNodesFromDOM(editor, dom);
```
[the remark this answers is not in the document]
tylerjbainbridge: This was left in by accident, I was originally going to analyze the count of the top level nodes being inserted via paste to determine if the popover should open but `dirtyLeaves` ended up working just fine.

packages/lexical-history/src/index.ts
tylerjbainbridge: Good catch- I'll add it back. I didn't end up using it anyways.

packages/lexical-clipboard/src/clipboard.ts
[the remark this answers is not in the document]
tylerjbainbridge: I'm going to keep the change because it's consistent with the rest of the function.

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
tylerjbainbridge: To make the Resolution object more flexible. Before it was just used with a Browser Range to calculate the shape of the selection (for Typeaheads), but now you can attach the Popover to any Lexical node as long as you supply a getRect function.

packages/lexical-plain-text/src/index.ts
```diff
@@ -98,6 +98,7 @@ function onPasteForPlainText(
       event instanceof InputEvent ? null : event.clipboardData;
 
     if (clipboardData != null && $isRangeSelection(selection)) {
+      editor._updateTags.add('paste');
```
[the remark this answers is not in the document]
tylerjbainbridge: Not easily accessible from these entrypoints. There's no `updateEditor` or `setEditorState`, the methods used for paste insert from selection and lose the context of the paste.

packages/lexical-rich-text/src/index.ts
tylerjbainbridge: From above

> Not easily accessible from these entrypoints. There's no updateEditor or setEditorState, the methods used for paste insert from selection and lose the context of the paste.

packages/lexical-plain-text/src/index.ts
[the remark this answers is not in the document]
acywatson: Sorry, I don't think I fully understand - you're in an editor.update callback here, so instead of manually setting the _updateTags props, can you not pass the tag in the EditorUpdateOptions, which is the second argument to editor.update method?

acywatson:
One point of clarification on the tags - other than that, I'm good once the tests pass.

packages/lexical-plain-text/src/index.ts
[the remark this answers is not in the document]
tylerjbainbridge: Oh! I see now. I was confused about how EditorUpdateOptions worked. Yes, in that case I can definitely add it.l

acywatson (approved):

PR #2706, opened 2022-07-26
OnChangePlugin ignoreInitialChange -> ignoreHistoryMergeTagChange

zurfyx:
The `ignoreInitialChange` property is a trap. While, it's logically well implemented it doesn't account for plugins that may load later in time and cause nodes to be dirty even when this results to 0 DOM changes.

This race condition is accentuated when we use `editorState` inside LexicalComposer instead as it loads even before other `useLayoutEffects`

Hence, I'm proposing killing this property and replacing it for `ignoreHistoryMergeTag` that practically speaking does the same job we intended this one to do.

### DIFF
--- a/packages/lexical-react/flow/LexicalOnChangePlugin.js.flow
+++ b/packages/lexical-react/flow/LexicalOnChangePlugin.js.flow
@@ -10,6 +10,7 @@
 import type {EditorState, LexicalEditor} from 'lexical';
 
 declare export function OnChangePlugin({
+  ignoreHistoryMergeTagChange?: boolean,
   ignoreInitialChange?: boolean,
   ignoreSelectionChange?: boolean,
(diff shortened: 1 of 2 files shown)


thegreatercurve:
Why do we even need these props? I thought the `initialEditorState` only gets memoized once on mount, and before all of the plugins, because the plugins depend on the editor instance?

thegreatercurve (approved):

zurfyx:
> Why do we even need these props? I thought the `initialEditorState` only gets memoized once on mount, and before all of the plugins, because the plugins depend on the editor instance?

Yes, when `editorState` is passed on the LexicalComposer initialConfig it is run before any of the plugins. But that only fixes `ignoreInitialChange`, there's still a  need for `ignoreHistoryMergeTagChange` because of transforms.

PR #2728, opened 2022-07-29
Highlight DOM Nodes onHover of Lexical DevTools Node

noi5e:
https://user-images.githubusercontent.com/4361605/181681718-e1288646-d7bd-4c6e-9baf-77fa4c9d3da0.mov

---

### DIFF
--- a/packages/lexical-devtools/src/devtools/index.ts
+++ b/packages/lexical-devtools/src/devtools/index.ts
@@ -16,66 +16,118 @@ chrome.devtools.panels.create(
   },
 );
 
-// const port = chrome.runtime.connect();
-
-// port.postMessage({
-//   name: 'init',
-//   tabId: chrome.devtools.inspectedWindow.tabId,
-//   type: 'FROM_DEVTOOLS',
-// });
-
 chrome.runtime.onConnect.addListener((port) => {
   port.onMessage.addListener((message) => {
     if (message.name === 'highlight') {
-      // console.log(`highlight: ${message.lexicalKey}`);
+      chrome.devtools.inspectedWindow.eval(`
+        highlight('${message.lexicalKey}');
+      `);
     }
 
     if (message.name === 'dehighlight') {
-      // console.log(`dehighlight: ${message.lexicalKey}`);
+      chrome.devtools.inspectedWindow.eval(`
+        dehighlight();
+      `);
     }
   });
 });
 
 function handleShown() {
-  // Use devtools.inspectedWindow.eval() to get editorState updates. For more info on security concerns:
+  // Security concerns related to chrome.devtools.inspectedWindow.eval():
   //   https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts
   //   https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/inspectedWindow/eval
-  chrome.devtools.inspectedWindow
-    // Attach a registerUpdateListener within the window to subscribe to changes in editorState.
-    // After the initial registration, all editorState updates are done via browser.runtime.onConnect & window.postMessage
-    .eval(
-      `
-      const editor = document.querySelectorAll('div[data-lexical-editor]')[0]
-        .__lexicalEditor;
-      
-      const initialEditorState = editor.getEditorState();
-      
-      // custom JSON serialization
-      // existing editorState.toJSON() does not contain Lexical ._key property
-      const serializeEditorState = (editorState) => {
-        const nodeMap = Object.fromEntries(editorState._nodeMap);
-        return {nodeMap};
-      };
-      
+
+  // query document for Lexical editor instance.
+  // TODO: add support multiple Lexical editors within the same page
+  // lexicalKey is attached to each DOM node like so: DOMNode[__lexicalKey_{editorKey}]
+  chrome.devtools.inspectedWindow.eval(`
+    const editorDOMNode = document.querySelectorAll(
+      'div[data-lexical-editor]',
+    )[0];
+    const editor = editorDOMNode.__lexicalEditor;
+    const editorKey = editorDOMNode.__lexicalEditor._key;
+    const lexicalKey = '__lexicalKey_' + editorKey;
+  `);
+
+  // depth first search to find DOM node with lexicalKey
+  chrome.devtools.inspectedWindow.eval(`
+    const findDOMNode = (node, targetKey) => {
+      if (targetKey === 'root') {
+        return editorDOMNode;
+      }
+
+      if (node[lexicalKey] && node[lexicalKey] === targetKey) {
+        return node;
+      }
+    
+      for (let i = 0; i < node.childNodes.length; i++) {
+        const child = findDOMNode(node.childNodes[i], targetKey);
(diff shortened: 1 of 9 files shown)


packages/lexical-devtools/src/devtools/index.ts
```diff
@@ -16,46 +16,118 @@ chrome.devtools.panels.create(
...
+    };
+
+    const dehighlight = () => {
+      highlightOverlay.style.display = 'none';
+    };
+  `);
+
+  // append highlight overlay <div> to document
+  chrome.devtools.inspectedWindow.eval(`
+    const highlightOverlay = document.createElement('div');
```
noi5e, line 91: I realized a bit late that a lot of the stuff in `devtools.js` can be moved into `content.js`. Particularly the parts that don't depend on the `editor` object, but are just declaring functions and variables in the `window` scope.

This would be ideal since they would be linted as TS in `content.js`, and they wouldn't be run in an `eval`. 

I can either move it in this PR with another commit, or early in the next PR.

noi5e: > I realized a bit late that a lot of the stuff in devtools.js can be moved into content.js. 

Actually forget about it, I spent the last couple of days working on this, but I can't get around the fact that `content.js` cannot `document.querySelector` the `editorDOMNode`. This is because content scripts have limited access to the DOM.

I don't think I'll be able to get around having the highlight/dehighlighting code within `devtools`.

@thegreatercurve This is otherwise ready to merge!

noi5e: > I don't think I'll be able to get around having the highlight/dehighlighting code within devtools.

I've been experimenting with some of the approaches in this [StackOverflow post](https://stackoverflow.com/questions/9515704/use-a-content-script-to-access-the-page-context-variables-and-functions). 

Some of it seems promising. In particular, I think the way to go in the future will be to have the content script inject something via `document.createElement('script')`. React DevTools takes this approach while still making limited use of `inspectedWindow.eval()`.

I tried to get this approach running yesterday, but for some reason it doesn't quite work yet. I think it has something to do with timing, like the script is being injected before the DOM finishes loading. I'll keep experimenting for now, but I think whatever changes I make should be the subject of a new PR.

thegreatercurve (approved):

packages/lexical-devtools/src/devtools/index.ts
thegreatercurve: Try defer loading of the script by setting `async` to be `true` on the script, or add an event listener on the window for after the document has fully loaded.

packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
```diff
@@ -40,15 +53,25 @@ function TreeNode({
       ''
     );
 
+  const hoverClassName = isHovered ? ' hover' : '';
+  const treeNodeClassName = 'tree-node' + hoverClassName;
```
acywatson, line 57: nit: template literals are a bit easier to deal with

packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
```diff
@@ -40,15 +53,25 @@ function TreeNode({
...
-      )}
-      {nodeString}
-      {<br />}
+    <div className="tree-node-wrapper" key={lexicalKey}>
+      <div
+        className={treeNodeClassName}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
+        style={{paddingLeft: leftIndent}}>
+        <span style={{width: '1.2em'}}>&nbsp;</span>
```
acywatson, line 67: should 1.2 be a constant somewhere?

acywatson (approved):

packages/lexical-devtools/src/devtools/index.ts
```diff
@@ -16,46 +16,118 @@ chrome.devtools.panels.create(
...
+      }
+    
+      for (let i = 0; i < node.childNodes.length; i++) {
+        const child = findDOMNode(node.childNodes[i], targetKey);
+        if (child) return child;
+      }
+    
+      return null;
+    };
+  `);
```
thegreatercurve, line 70: @noi5e Maybe do this in another PR, but if you want to be able to use the eval code outside of the eval statement so you get actual TS errors, just out it in it's own function scope:

```suggestion
  const getEvalCode = () => {
    const findDOMNode = (node, targetKey) => {
      if (targetKey === 'root') {
        return editorDOMNode;
      }

      if (node[lexicalKey] && node[lexicalKey] === targetKey) {
        return node;
      }
    
      for (let i = 0; i < node.childNodes.length; i++) {
        const child = findDOMNode(node.childNodes[i], targetKey);
        if (child) return child;
      }
    
      return null;
    };
  }

  chrome.devtools.inspectedWindow.eval(`${getEvalCode()}`);
```

packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
```diff
@@ -17,15 +17,28 @@ function TreeNode({
   __text,
   __type,
   children,
+  deHighlightDOMNode,
   depth,
+  highlightDOMNode,
   lexicalKey,
 }: DevToolsNode): JSX.Element {
   const [isExpanded, setIsExpanded] = useState(true);
+  const [isHovered, setIsHovered] = useState(false);
```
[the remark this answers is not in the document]
noi5e: > Can you just use a hover pseudo CSS class to control the color of the nodes in the tree view?
> You can probably get rid of this and just control the styling with CSS pseudo classes.

Oops. Totally forgot the `:hover` selector existed. Fixed with the latest commit.

packages/lexical-devtools/src/devtools/index.ts
noi5e: > Try defer loading of the script by setting async to be true on the script, or add an event listener on the window for after the document has fully loaded.

Nice, I tried some of these. I went with another workaround where the DevTools app sends a message, requesting to load the `editorState` the moment the DevTools panel is opened in the browser console. Seems to work fine.

The good news is that I got all of the code out of the `eval()` statement and into an injected script, and it works great. I'll bundle the changes into another PR right after this one is merged.

packages/lexical-devtools/src/panel/components/TreeNode/index.tsx
noi5e: Okay, it's using CSS variables now...

This is a little weird, not sure if this is best practice, but I'm using `getComputedStyle` here to read the CSS variable from the DOM in `TreeView`:
```
const monospaceWidth = getComputedStyle(
    document.documentElement,
  ).getPropertyValue('--monospace-character-width');
```

In order to pass it to `TreeNode` here:
```
const leftIndent = depth * parseFloat(monospaceWidth) + 'em';
```

noi5e:
This is merge ready if the fix [here](https://github.com/facebook/lexical/pull/2728#discussion_r936229910) is okay, tests are mostly passing except for some unrelated stuff.

Once this is merged, I'll open a PR for the changes mentioned [here](https://github.com/facebook/lexical/pull/2728#discussion_r936216159).

thegreatercurve (approved):

Issue #2731, opened 2022-07-29
Bug: typo in Documentation. It should be ReactNode instead of React$Node


iampava:
## Steps To Reproduce

1. Navigate to this page in the Docs: https://lexical.dev/docs/concepts/nodes
2. Look at the code snippets. They mention `React$Node`

## The expected behavior

The Docs should mention `ReactNode`

iampava:
Merged!

PR #2732, opened 2022-07-29
Bug: typo in Documentation. It should be ReactNode instead of React (…

iampava:
…https://github.com/facebook/lexical/issues/2731)

### DIFF
--- a/examples/decorators.md
+++ b/examples/decorators.md
@@ -4,7 +4,7 @@ Decorator node is a way to embed non-text components into the editor. It can be 
 
 Here's an example of how you can create a decorator node for embedding a video:
 
-```js
+```ts
 export class VideoNode extends DecoratorNode {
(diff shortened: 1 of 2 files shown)


iampava:
This fixes: https://github.com/facebook/lexical/issues/2731

examples/decorators.md
```diff
@@ -4,7 +4,7 @@ Decorator node is a way to embed non-text components into the editor. It can be
 
 Here's an example of how you can create a decorator node for embedding a video:
 
-```js
+```ts
```
iampava, line 7: Changed these to `ts` as the language is not JavaScript, but this leads to a slightly different result when applying syntax highlighting in browser:

Previous
image


Now
image

Your call if you want me to revert this :)

thegreatercurve (approved):

PR #2735, opened 2022-07-29
fix: `timeoutId` type

shanpriyan:
- Fixed `timeoutId` type (`setTimeout` returns a number  in browser  and `Timeout` object in node)
- Removed `window.` prefix, since it might break during **SSR**

### DIFF
--- a/packages/lexical-playground/src/hooks/useReport.ts
+++ b/packages/lexical-playground/src/hooks/useReport.ts
@@ -31,11 +31,13 @@ const getElement = (): HTMLElement => {
   return element;
 };
 
-export default function useReport(): (arg0: string) => number {
-  const timer = useRef<number | null>(null);
+export default function useReport(): (
+  arg0: string,
+) => ReturnType<typeof setTimeout> {
+  const timer = useRef<ReturnType<typeof setTimeout> | null>(null);
   const cleanup = useCallback(() => {
     if (timer !== null) {
-      clearTimeout(timer.current as number);
+      clearTimeout(timer.current as ReturnType<typeof setTimeout>);
     }
 
     if (document.body) {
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/hooks/useReport.ts
```diff
@@ -52,9 +52,9 @@ export default function useReport(): (arg0: string) => NodeJS.Timeout {
       // eslint-disable-next-line no-console
       console.log(content);
       const element = getElement();
-      clearTimeout(timer.current as NodeJS.Timeout);
+      clearTimeout(timer.current as number);
       element.innerHTML = content;
-      timer.current = setTimeout(cleanup, 1000);
+      timer.current = window.setTimeout(cleanup, 1000);
```
zurfyx, line 57: Let's not use the `window.` prefix (here and other occurrences, especially NPM packages), it will fail on some frameworks like Remix when loaded on the server-side, especially now that we support headless mode

zurfyx (commented):
Thank you!

packages/lexical-playground/src/hooks/useReport.ts
shanpriyan: After removing `window.` prefix before setTimeout, typescript is expecting `NodeJS.Timeout` instead of type `number`. 
image
The solutions would be to use,
1. type `ReturnType<typeof setTimeout>` instead of `number` for timerId
ref: https://stackoverflow.com/a/51040768/17948948

(OR)

2. Keep the type `number` for timerId and add the below in *tsconfig.json*
ref: https://stackoverflow.com/a/62323045/17948948
```js
{
  "compilerOptions": {
     ...
    "types": []
  }
}
```
which one shall I choose?

shanpriyan: Done!

thegreatercurve (approved):

PR #2736, opened 2022-07-30
fix: composing text input interruption when editor is at composing state

geminiwen:
When I type some Chinese for appending text to HashtagNode (Editor is in composing state), lexical should prevent default action and stop inserting text for TextNode.

https://user-images.githubusercontent.com/1413056/181918441-07c6bcf2-785f-469e-b11e-07e2a11c2699.mov


trueadm:
This doesn't seem like the right fix. We definitely want composed text to not go to the default place, and instead modify the selection during compositionstart instead?

acywatson:
Closing due to inactivity - feel free to re-open if someone wants to pick this back up.

PR #2738, opened 2022-07-31
feat: add iframe plugin

im-adithya:
Adds IFrame Plugin similar to YoutubePlugin //cc @thegreatercurve
Screenshot from 2022-07-31 15-51-30


acywatson:
Not opposed, but how does this play with AutoEmbedPlugin?

@tylerjbainbridge

echarles:
> Not opposed, but how does this play with AutoEmbedPlugin?

I guess these are similar but still different cases.

I have tried the vercel playground with https://www.google.com but google refused to display.

Screenshot 2022-08-07 at 08 03 57

Also, I am not able to delete the iframe.

acywatson:
> I guess these are similar but still different cases.

I don't think so - rendering in an iframe is one strategy for embedding, so I think this is redundant.

echarles:
I would say that `AutoEmbedPlugin` looks for a regexp matching a URL, e.g. for figma https://www.figma.com/file/LKQ4FJ4bTnCSjedbRpk931/Sample-File. If the user pastes such URL, he is proposed for a `FigmaNode` rendred as a `FigmaComponent` which happens to contain an iframe.

This PR allows to directly embed an iframe with any URL to showcase any website content in the Lexical editor.

acywatson:
> I would say that `AutoEmbedPlugin` looks for a regexp matching a URL, e.g. for figma https://www.figma.com/file/LKQ4FJ4bTnCSjedbRpk931/Sample-File. If the user pastes such URL, he is proposed for a `FigmaNode` rendred as a `FigmaComponent` which happens to contain an iframe.
> 
> This PR allows to directly embed an iframe with any URL to showcase any website content in the Lexical editor.

I guess that makes sense, but couldn't you just make a regex that matches any URL and then render an iframe with AutoEmbed? I feel like there's some duplication of intent here. I will let others weigh in - I might be wrong. If no one else has thoughts, we can merge this.

im-adithya:
> I have tried the vercel playground with https://www.google.com/ but google refused to display.

@echarles try using this link: https://www.google.com/webhp?igu=1, it's because of the SAMEORIGIN header, [more here](https://stackoverflow.com/questions/8700636/how-to-show-google-com-in-an-iframe)

> Also, I am not able to delete the iframe.

Checked it just now, works fine for me. Please let me know if you happen to observe it still.

echarles:
Thx @im-adithya 

Tested with https://www.google.com/webhp?igu=1 and the iframe displays google.

I can also delete the iframe in the last playground.

im-adithya:
> how does this play with AutoEmbedPlugin?

AutoEmbedPlugin only works when we specify EmbedConfig for each type ([reference](https://github.com/facebook/lexical/blob/0fa3f72cbeb21b9879e6935280e6c869b0364b72/packages/lexical-playground/src/plugins/AutoEmbedPlugin/index.tsx#L46)), and here I am not specifying any for two reasons:
1) It would override the other configs saying "Embed an IFrame" instead of being more specific (like "Embed Tweet") when a link with a specified Config is pasted.
2) We cannot predict whether the site has an IFrame option or not beforehand, and that might lead to a screen like the above.


So, for now, this would only work when we select it from the toolbar and will not overlap with AutoEmbedPlugin in any way.

echarles:
Looking more at the source code, I now better understand @acywatson question about the relationship with the AutoEmbedPlugin. I was at some point also asking myself why not adding a IFrameEmbedConfig to the existing 3 ones in https://github.com/facebook/lexical/blob/abcdd57f0dca304a12cbcd4ffc5bc54369bb0419/packages/lexical-playground/src/plugins/AutoEmbedPlugin/index.tsx#L148-L152

> It would override the other configs saying "Embed an IFrame" instead of being more specific (like "Embed Tweet") when a link with a specified Config is pasted.

Have not tried, but it may be that the first match regexp rule will define which Embed is chosen. Would a catch-all rule for a IFrameEmbedConfig that at the end of the embed list be a solution?

> We cannot predict whether the site has an IFrame option or not beforehand, and that might lead to a screen like the above.

Ok, so we don't want all pasted linked be matched by the IFrameEmbedConfig and rendered in a preview IFrame, potentially broken. But when you paste a link, you are proposed with a hover menu option, so that would not be an issue (well, pasting a link is quite common, so the user will have to choose `Dismiss` when he pastes a link).

Screenshot 2022-08-11 at 08 57 54

acywatson:
I'm going to close this one for now - we can potentially revisit later

PR #2741, opened 2022-08-01
fix: getTopLevelElement for decoratorNode

im-adithya:
If a node is inserted while a `decoratorNode` is selected, it gives an error as it fails to receive a `topLevelElement`

Fix: The `topLevelElement` should be itself during a node selection so that the inserted node can be added using `insertAfter`

## Before
https://user-images.githubusercontent.com/64399555/182132955-f6fe368e-a68c-4bad-ad98-d862ae5495e6.mp4
## After
https://user-images.githubusercontent.com/64399555/182132988-7ec067ac-7342-4a07-90a0-11b0055ca318.mp4

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -13,7 +13,13 @@ import type {Klass} from 'lexical';
 
 import invariant from 'shared/invariant';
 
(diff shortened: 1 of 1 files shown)


thegreatercurve (approved):
Seems like a legitimate fix to me as decorator nodes are valid at the root child level. Thanks @im-adithya.

PR #2743, opened 2022-08-01
Table of contents style improvements

karam-qaoud:


packages/lexical-playground/src/Editor.tsx
```diff
@@ -159,9 +159,7 @@ export default function Editor(): JSX.Element {
         )}
         {isAutocomplete && <AutocompletePlugin />}
         <ActionsPlugin isRichText={isRichText} />
-        <div className="toc">
```
tylerjbainbridge, line 162: Is there any reason you need the wrapper div here?

karam-qaoud: No, I already got rid of that! did you mean this div `<div>{showTableOfContents && <TableOfContentsPlugin />}</div>`?

packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
```diff
@@ -1,74 +1,78 @@
...
+.heading3 {
+  margin-left: 20px;
+}
+
+.selectedHeading {
+  color: #3578e5;
+  position: relative;
+}
+
+.xselectedHeading::before {
```
zurfyx, line 14: Let's find a better name for this

packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
```diff
@@ -1,74 +1,78 @@
...
 
-.circle {
+.heading3 {
+  margin-left: 20px;
+}
+
+.selectedHeading {
+  color: #3578e5;
+  position: relative;
+}
```
zurfyx, line 12: Have you tested the specificity of these selectors? I wonder if we need to be more specific to prevent conflicts i.e. `.toc .heading2`

zurfyx (approved):

PR #2760, opened 2022-08-04
Fix typo in community page

Josh-Cena:
@yangshun knows that I-dont-like-typos 😄


thegreatercurve (approved):
LGTM! You can tell where we originally copied this README from 😅

Josh-Cena:
Yep, as Meta projects always do :P

PR #2761, opened 2022-08-04
feat: Replace select with dropdown for code

LuciNyan:
feat: replace select with dropdown for code

image


thegreatercurve (approved):
LGTM! We keep saying this @LuciNyan, but thanks again! Your contributions are invaluable.

PR #2764, opened 2022-08-04
Handle `insertTranspose` for `beforeinput` event

### LINKED ISSUE #2751: Bug: Ctrl-T on Mac should transpose characters
In a native Mac text field, pressing Ctrl-T will transpose the adjacent characters and move one character rightwards. i.e. given

```
a b c
 ^ cursor
```

pressing Ctrl-T will give

```
b a c
   ^ cursor
```

This shortcut doesn't work in Lexical. Reproed on lexical.dev and in Workplace UFI.

fantactuka:
Pressing `^ + T` works in Safari since it triggers onbeforeinput with `inputType=insertText`, while chrome uses `inputType=insertTranspose`:

fix #2751

https://user-images.githubusercontent.com/132642/182932055-f9a849a7-28f8-40c8-a320-312a2d4afb44.mov

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -436,7 +436,7 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
     const anchorNode = anchor.getNode();
     const focusNode = focus.getNode();
 
-    if (inputType === 'insertText') {
+    if (inputType === 'insertText' || inputType === 'insertTranspose') {
       if (data === '\n') {
         event.preventDefault();
         dispatchCommand(editor, INSERT_LINE_BREAK_COMMAND, false);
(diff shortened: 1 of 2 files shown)


thegreatercurve (approved):

thegreatercurve:
Tested with other Emacs key bindings (i.e. CTRL+A to go to the start of a line) to verify fix works.

packages/lexical/src/LexicalEvents.ts
```diff
@@ -436,7 +436,7 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
     const anchorNode = anchor.getNode();
     const focusNode = focus.getNode();
 
-    if (inputType === 'insertText') {
+    if (inputType === 'insertText' || inputType === 'insertTranspose') {
```
thegreatercurve, line 439: Maybe just add a comment to the above explaining why we need this fix.

PR #2765, opened 2022-08-04
feat: styles dropper in playground

im-adithya:
[NOT READY YET - waiting for approval]

I don't know if this can be a useful feature or not, but quite frequently, I copy-paste similarly-styled text and then modify it instead of typing the text without styling and then applying the same styles again (usually when there's a paragraph of text in between them). This becomes more annoying if more styles are applied as we have to apply them again one by one.

This PR aims to _fix_ that issue(?) by allowing users to _pick and drop_ styles from a text node to the selected ones.

### Demo

https://user-images.githubusercontent.com/64399555/182938924-cfb14b14-db79-4a27-8110-eca3b773dbf8.mp4


packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -999,11 +1017,30 @@ export default function ToolbarPlugin(): JSX.Element {
...
+              });
+            }
+          });
+        }
         return false;
       },
       COMMAND_PRIORITY_CRITICAL,
     );
-  }, [editor, updateToolbar]);
+  }, [activeEditor, dropperSelection, editor, isDropper, updateToolbar]);
```
echarles, line 1043: You add a bunch of deps here, will that cause which will trigger the hook. Is there a risk for too many calls here. Sounds to me not, as there is the ` if (isDropper) ` but wanted to double check with you.

echarles:
Works great for me on the vercel playground. just left a minor question.

acywatson:
Yea, I think a feature like this is cool and useful for the playground.

Not sure what the state of this is at the moment in terms of review-readiness, but there seem to be a few bugs in the Vercel preview. Selection behaves pretty strangely, in general, and the behavior of this feature seems to not work correctly with text inside lists, for instance.

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -999,11 +1017,30 @@ export default function ToolbarPlugin(): JSX.Element {
       (_payload, newEditor) => {
         updateToolbar();
         setActiveEditor(newEditor);
+        if (isDropper) {
+          newEditor.update(() => {
+            const newSelection = $getSelection();
+            const node = newSelection?.getNodes()[0];
```
acywatson, line 1023: What is the intent here? We probably want to use anchor or focus, depending on selection "backwardness"?

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -999,11 +1017,30 @@ export default function ToolbarPlugin(): JSX.Element {
...
+            $setSelection(dropperSelection);
```
acywatson, line 1024: This might be more of an opinion, but I'm not sure this is the right way to do this, from a UX standpoint. To me, it makes more sense to select something, hit the button to copy the styles, then select something else and press the button again to paste the styles. The behavior of automatically applying the styles I copied to the next thing I select is a bit weird to me, for whatever reason.

Do similar features work this way in other text editors?

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
acywatson: Might make the most sense to just move this into a separate command listener. Another option would be to create a separate function like updateToolbar.

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
echarles: > Do similar features work this way in other text editors?

I have checked google docs and have not found that feature. I remember GIMP has that, but had hard time to make sense how to use it.

In this case, I find the current proposed user experience quite good once you know how to use it.

echarles:
> but there seem to be a few bugs in the Vercel preview. Selection behaves pretty strangely, in general, and the behavior of this feature seems to not work correctly with text inside lists, for instance.

Did a few more test and when I pick italic and apply to a bold text, it becomes bold+italic, which is inline with the implementation, but may not be what the user would expect (he would expect italic only).

> this feature seems to not work correctly with text inside lists, for instance.

Tried to pick and apply format to text in a list (bullets), and it worked for me.

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
im-adithya: As this gets triggered on `SELECTION_CHANGE`, I assumed that the selection the user makes would be a point selection (anchor === focus) and get the first node of it (which would be nothing but itself as it's the only node) and apply its styles.

This is the reason behind the strange selection issues that you were mentioning above... You can reproduce it if you consciously try selecting a bunch of text instead of clicking at one place while the dropper is selected (by knowingly sliding the mouse while you select :P).

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
im-adithya: What if we change the cursor to make it more obvious that _you're picking something now_?
Like this:
image

Most color droppers work this way, so this may be easy to understand.

Although I'm not sure about whether I implemented this correctly as I used the SettingsContext by adding `showDropper` to it. (and hence appears in the link above `http://localhost:3000/?showDropper=true` when being used, which makes me think it's not a good choice)

im-adithya:
> Did a few more test and when I pick italic and apply to a bold text, it becomes bold+italic, which is inline with the implementation, but may not be what the user would expect (he would expect italic only).

Oops, forgot about that; thanks for reminding :D

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
im-adithya: Sure!

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
echarles: > What if we change the cursor to make it more obvious that you're picking something now?

That would be super useful

thegreatercurve:
@im-adithya Can this be moved in the toolbar to after the font color and background highlight drop downs? That probably makes more contextual sense

thegreatercurve:
Closing this PR due to it becoming stale. Also, since we initially talked about this PR, I think we've decided to somewhat simplify the contents of the playground to remove lots of additional plugins that could overwhelm first time visitors.

Issue #2755, opened 2022-08-04
Bug: LexicalTypeaheadMenuPlugin never cleans up its anchor element


Morphyish:
Lexical version: 0.3.8

## Steps To Reproduce

1. Go to the [playground](https://playground.lexical.dev/)
2. Open the mention menu (by typing `@s` somewhere in the editor)
3. Delete the `s`, then type it again
4. Repeat the 3rd step multiple time

## The current behavior

A new `<div aria-label="Typeahead menu" id="typeahead-menu" />` element is created everytime, and never removed from the DOM.

This can get out of hand pretty fast if you have a lot of plugins relying on it (on the Playground it also happens everytime you use the `/` command).

![image](https://user-images.githubusercontent.com/39383180/182835880-6a771518-7ff7-40ca-84ae-8e45e155ecbc.png)

## The expected behavior

Have a single anchor element for every instance of the editor.

This can probably be achieved by storing the `anchorElementRef` in the plugin instead of the `ShortcutTypeahead` component, and passing it down as a prop.

PR #2768, opened 2022-08-05
fix: Single anchor element per LexicalTypeaheadMenuPlugin instance

Morphyish:
Fixes https://github.com/facebook/lexical/issues/2755

Creating the anchorElement in the Plugin itself and passing it down to the `ShortcutTypeahead` avoid creating a lot of unnecessary nodes that are never removed.

Every plugin that relies on `LexicalTypeaheadMenuPlugin` will get its anchor element that will then be properly updated with the new position, as it should.

Please note that I had to remove the `startTransition` when setting the `resolution` to `null`. 
The transition was causing the `/` menu to flicker, but not the mentions. I'm not sure if it's a specific implementation issue in either of those playground plugins, but as hiding the menu should not cause any unnecessary loading state on the UI, it felt safe to remove.

Demo

https://user-images.githubusercontent.com/39383180/183049115-05158813-c31e-4978-9520-660fbd0c10bc.mp4


packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -224,13 +224,15 @@ function startTransition(callback: () => void) {
...
   editor,
+  anchorElementRef,
   resolution,
   options,
   menuRenderFn,
   onSelectOption,
 }: {
   close: () => void;
   editor: LexicalEditor;
+  anchorElementRef: MutableRefObject<HTMLElement>;
```
thegreatercurve, line 235: I think we want to use `forwardRef` here instead of passing the ref object as a prop: https://reactjs.org/docs/forwarding-refs.html

thegreatercurve (commented):
Thanks for the contribution @Morphyish! One small fix, but I'll let @tylerjbainbridge review this fully as this was his component.

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
Morphyish: Would that work? 
The HTML element is created in the parent, the prop is used to pass down the ref so I can update the attributes on the element inside the `useRef` hook.

But maybe I should move the entire `useEffect` hook in the parent, and only pass down the `anchorElementRef.current` value, maybe that would make more sense?

tylerjbainbridge (approved):
This makes sense to me, good idea!

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
thegreatercurve: I think you could probably use it with `useImperativeHandle` and keep all the original ref logic in the child. This is still all good though.

PR #2770, opened 2022-08-05
Selection#formatText to retain selection and handle all text nodes

fantactuka:
A few bugs in formatText affected selection retention and nodes formatting in case selection starts with non-text node:

1. Selection changes after formatting

https://user-images.githubusercontent.com/132642/183076029-8d848c8b-4427-4c9e-bb1e-aa9c78ae1ca5.mov

https://user-images.githubusercontent.com/132642/183076594-afb91e3d-6e51-44f6-994c-b35420246be8.mov

2. Selection formats only first text node

https://user-images.githubusercontent.com/132642/183076292-32adb1b1-6104-408c-8a3f-ce0f409c90e1.mov

3. Backward selection (focus is before anchor) converted to a regular one (anchor goes first). Select word backward, make it bold, press Shift + left arrow. Expected selection to keep expanding backward

https://user-images.githubusercontent.com/132642/183086040-c8f09519-25ef-4f45-933e-39d15822f002.mov

Overall logic:
- Get selected text nodes (skip first one if selection starts on its end)
- Get start/end text offsets including cases with non-text anchor/focus
- Split first/last nodes if selection starts or ends inside text node and format it
- Format all text nodes in between
- Move selection only in case when it was on text node that was split, otherwise keep it as is (including cases when it's on element node)

Adjusted e2e that expected backward selection to be converted to a regular one

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/TextFormatting.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/TextFormatting.spec.mjs
@@ -183,9 +183,9 @@ test.describe('TextFormatting', () => {
       `,
     );
     await assertSelection(page, {
-      anchorOffset: 0,
+      anchorOffset: 5,
       anchorPath: [0, 1, 0],
-      focusOffset: 5,
+      focusOffset: 0,
       focusPath: [0, 1, 0],
     });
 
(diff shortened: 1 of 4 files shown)


zurfyx (approved):
Woah, you rewrote the whole formatText function, impressive 🙇‍♂️

Awjin:
I ran into this bug yesterday, and like magic this fix was released today. Love this lib and the solid work you all are doing, it is very much appreciated.

PR #2777, opened 2022-08-06
chore: Move useCollaborationContext to dedicated file

Morphyish:
This PR moves `useCollaborationContext` out of the `LexicalCollaborationPlugin` and into its own file.

`LexicalCollaborationPlugin` imports functions from `@lexical/yjs`, which has a `yjs` peer dependency, and also from `y-websocket`.
At the same time, `useCollaborationContext` is used in `LexicalNestedComposer`.

The combination of those 2 was causing the application to crash if you tried to use `LexicalNestedComposer` without adding `yjs` and `y-websocket` as a dependency on your project.

`yjs` should not be mandatory to use nested composers.

By splitting the 2, it should hopefully solves this issue, as the new context file only imports types from `yjs` and has no runtime dependencies to either `yjs` nor `y-websocket`.


echarles:
> By splitting the 2, it should hopefully solves this issue, as the new context file only imports types from yjs

Thx for this architecture move. I wonder why the new context file still needs to import types from yjs?

Morphyish:
> > By splitting the 2, it should hopefully solves this issue, as the new context file only imports types from yjs
> 
> Thx for this architecture move. I wonder why the new context file still needs to import types from yjs?

The `yjsDocMap` in `CollaborationContextType` still needs to know what the yjs Doc type is, for typescript validation, and I couldn't figure out a way to make it work through generic types.

But, as this is a type import, it shouldn't have an impact at runtime, and as the `CollaborationContext` is thighly coupled to yjs anyway, it felt ok to leave it here instead of having to import the `Doc` type everytime you want to use the context.

acywatson:
This looks right to me - that collab test failure is probably a flake. Executed a re-run to see if we can get a check there.

We can see if we can get the Doc type out of there later - for now, this is an improvement.

acywatson (approved):

PR #2778, opened 2022-08-06
Inject DevTools Script in Browser Extension

noi5e:
This moves all of the code that was previously running as an inline string in `devtools.inspectedWindow.eval()` like so:
Screen Shot 2022-08-05 at 6 34 11 PM

And moves it to a new `inject.ts` file:
Screen Shot 2022-08-05 at 6 34 51 PM

---

### DIFF
--- a/packages/lexical-devtools/src/content/index.ts
+++ b/packages/lexical-devtools/src/content/index.ts
@@ -5,6 +5,22 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
+import {IS_FIREFOX} from 'shared/environment';
+
+// const CAN_USE_DOM =
+//   typeof window !== 'undefined' &&
+//   typeof window.document !== 'undefined' &&
+//   typeof window.document.createElement !== 'undefined';
+// const IS_FIREFOX =
+//   CAN_USE_DOM && /^(?!.*Seamonkey)(?=.*Firefox).*/i.test(navigator.userAgent);
+
+// for security reasons, content scripts cannot read Lexical's changes to the DOM
+// in order to access the editorState, we inject this script directly into the page
+const script = document.createElement('script');
+script.src = chrome.runtime.getURL('inject.js');
+document.documentElement.appendChild(script);
+if (script.parentNode) script.parentNode.removeChild(script);
+
 const port = chrome.runtime.connect();
 
 port.postMessage({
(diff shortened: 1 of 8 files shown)


packages/lexical-devtools/src/content/index.ts
```diff
@@ -5,6 +5,21 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
+import {IS_FIREFOX} from 'shared/environment';
+
+declare global {
+  interface DocumentEventMap {
+    editorStateUpdate: CustomEvent;
+  }
+}
```
noi5e, line 14: For the `highlight` `CustomEvent`. I'm unsure if a `declare global` is what I'm supposed to be doing here. @thegreatercurve

packages/lexical-devtools/src/content/index.ts
```diff
@@ -31,3 +46,34 @@ window.addEventListener('message', function (event) {
...
+    type: 'FROM_CONTENT',
+  });
+});
+
+port.onMessage.addListener((message) => {
+  if (message.name === 'highlight') {
+    const data = {lexicalKey: message.lexicalKey};
+    const detail = IS_FIREFOX
+      ? // eslint-disable-next-line no-undef
+        cloneInto(data, document.defaultView) // see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts#cloneinto
```
noi5e, line 63: This is the only TS error I couldn't find a way to handle. This is a [Firefox-only browser extension API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts#cloneinto).

There's a definition written in [this package](https://github.com/emoji-gen/clone-into), should I just copy it?

noi5e: Okay, I just went ahead and wrote a `cloneInto` type. After all, it's just deep cloning the same object.

But now I get this error:
Screen Shot 2022-08-06 at 2 43 35 PM

Where to go from here? Declare the definition on the `window`?

noi5e: Here's the current type definition from `types.ts`:
```
export type cloneInto = (
  arg: {data: {lexicalKey: string}},
  arg2: WindowProxy,
) => {data: {lexicalKey: string}};
```

noi5e: Something like this within `content.ts`?
Screen Shot 2022-08-06 at 4 40 39 PM

packages/lexical-devtools/src/inject/index.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+import {EditorState} from 'lexical';
+import {LexicalHTMLElement, LexicalKey} from 'packages/lexical-devtools/types';
+
+declare global {
+  interface DocumentEventMap {
```
thegreatercurve, line 12: Why declare this in 2 places?

packages/lexical-devtools/src/content/index.ts
thegreatercurve: You can do, it will just get referenced in this module scope either way.

noi5e:
@thegreatercurve My current fix for [cloneInto](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts#cloneinto) being undefined is a little weird, you might want to take a look. 

I ended up dropping a `@ts-ignore` to declare `cloneInto` on `globalThis`. This is following the lead of this [third-party wrapper](https://github.com/emoji-gen/clone-into). Not sure what my alternative is at this point, there doesn't seem to be a DefinitelyTyped definition of this Firefox-only API.

Otherwise this is merge ready!

PR #2780, opened 2022-08-07
chore(lexical-playground): typos, improved build

Dattaya:
Thanks for open sourcing this awesome editor. 
Some things that I noticed while trying to build the playground, they're unrelated but each in a separate commit.


acywatson (approved):
Looks right, thanks!

Issue #2779, opened 2022-08-07
Bug: DecoratorNodes content should be collaborative


echarles:
I am not sure if this is a bug or a feature request, so put it as bug for now as a user.

Lexical version: 0.3.8

## Steps To Reproduce

1. Run the playground in collaborative mode
2. Insert a decorator node, eg an equation or a sticky node
3. Change the content of that node

## The current behavior

The change of the content of the decorator nodes are not collaborative. This is probably logical from a developer perspective based on the current code (nothing is forseen to propagate the changes), but as a user, this is confusing.

## The expected behavior

The user would expect the content change to be collaborative.

![Untitled9](https://user-images.githubusercontent.com/226720/183277782-d58ddf95-7d11-4e66-b3c6-70dc152c0c62.gif)

![Untitled5](https://user-images.githubusercontent.com/226720/183277784-ba7adc83-7bca-407b-a4e7-33a2ebd61006.gif)

trueadm:
This used to work, so it's a recent regression. It's likely to do with changes to nested editors.

PR #2781, opened 2022-08-07
Fix nested editors in collab

trueadm:
Fixes https://github.com/facebook/lexical/issues/2779. We now use a boolean flag to determine if collab is ready and active.

### DIFF
--- a/packages/lexical-react/src/__tests__/unit/utils.tsx
+++ b/packages/lexical-react/src/__tests__/unit/utils.tsx
@@ -20,10 +20,12 @@ import {ContentEditable} from '../../LexicalContentEditable';
 import {RichTextPlugin} from '../../LexicalRichTextPlugin';
 
 function Editor({doc, provider, setEditor}) {
-  const {yjsDocMap} = useCollaborationContext();
+  const context = useCollaborationContext();
 
   const [editor] = useLexicalComposerContext();
 
+  const {yjsDocMap} = context;
+  context.isCollabActive = true;
   yjsDocMap.set('main', doc);
 
   setEditor(editor);
--- a/packages/lexical-playground/src/nodes/EquationNode.tsx
+++ b/packages/lexical-playground/src/nodes/EquationNode.tsx
@@ -62,6 +62,12 @@ function EquationComponent({
   );
 
   useEffect(() => {
+    if (!showEquationEditor && equationValue !== equation) {
+      setEquationValue(equation);
+    }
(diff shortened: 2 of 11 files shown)


echarles:
Thx @trueadm, just tested on https://lexical-playground-git-fix-collab-nested-editors-fbopensource.vercel.app/split/index.html?isCollab=true 

The sticky notes content are synced, but the equation do not sync for me.

echarles:
The image caption is also synced.

These changes looks to correctly address the nested editor case, but the equation editor is another case that does not use the nested editor.

A general question for me is "What are the requirements to make a Decorator node syncable?"

trueadm:
> A general question for me is "What are the requirements to make a Decorator node syncable?"

Decorator nodes should sync automatically. If you use the Lexical primitive values, it should "just work". I'll take a look at the equation editor, but it's likely a regression in that plugin somewhere along the lines too.

acywatson:
Do we have any tests around this? Might be nice to add that so this doesn't regress again.

acywatson (approved):

echarles:
> Decorator nodes should sync automatically. If you use the Lexical primitive values, it should "just work". 

Thx. I will dig more into the code to better understand the logic and hopefully come with a PR with some documentation on that.

> Update: was a simple bug in the equation node logic, it wasn't updating the React state with the value on the node

Equation are now synced in the playground. Thx.

fantactuka (approved):
Looks great, completely missed that we've used `main` as collab indicator for plugins

packages/lexical-react/src/__tests__/unit/utils.tsx
```diff
@@ -22,10 +22,12 @@ import {ContentEditable} from '../../LexicalContentEditable';
 import {RichTextPlugin} from '../../LexicalRichTextPlugin';
 
 function Editor({doc, provider, setEditor}) {
-  const {yjsDocMap} = useCollaborationContext();
+  const context = useCollaborationContext();
 
   const [editor] = useLexicalComposerContext();
 
+  const {yjsDocMap} = context;
+  context.isCollabActive = true;
```
fantactuka, line 30: Should this be within useEffect to unset isCollabActive on onmount?

trueadm: Good point. Mind taking over this PR? I won’t be able to work on it again till I’m back in a few weeks.

Issue #2610, opened 2022-07-07
Full RTL direction support


Nirlah:
The following is a short overview of bi-directional editing surfaces and my suggestion for improving Lexical’s implementation. _(Continued from the [Discord discussion](https://discord.com/channels/953974421008293909/955972012541628456/993985520747155476))_

### Overrview
Modern editing surfaces’ default behavior today is to support bi-directional content by setting the direction based on the first letter’s language. This can be seen in the current Lexical implementation, _Apple Notes_, _Gmail_, and _Google Docs_:

https://user-images.githubusercontent.com/1909713/177728347-054678d1-cf3e-47d5-817e-c5041ff144ef.mp4

One implementation detail crucial for longer-form and complex content (like a letter or an article) is maintaining the previous paragraph setting if the user manually sets it. _Microsoft Office Word_ gets this correctly:

https://user-images.githubusercontent.com/1909713/177722676-d603b861-daad-42b4-8378-f053bc455fa9.mp4

_The recordings demonstrate Hebrew, but the behavior is similar in Arabic._

### Implementation suggestions

The following are my suggestions for scalable bi-directional support in Lexical:
1. Maintain the current dynamic evaluation of direction as the default behavior.
2. Add support for manually toggling/setting the node’s direction.
3. If a node has a `direction` manually set, it should be respected. When parsing a serialized editor state, the current implementation dynamically evaluates the content, disregarding the provided configuration.
4. One (or both) of the following will allow composing more complex content (letter/article) without constant friction:
  4.1. If the root node’s direction is _manually_ set, it should be the default direction for child nodes.
  4.2. When creating the following paragraph, if the previous paragraph has a manual direction set, it should be copied into the next paragraph.

Nirlah:
> > If a node has a direction manually set, it should be respected. (When parsing a serialized editor state, the current implementation dynamically evaluates the content, disregarding the provided configuration.)
> 
> QQ: Why it would not respect manual direction in case of deserializing, e.g. if you save editor state and then init new editor with serialized state wouldn't you want to keep that manually set direction?

Yes, the parenthesis describes the current Lexical implementation that ignores the node’s `direction` value when deserializing.

milkrong:
what is status of this features?

sopost-kirt:
[...]

PR #2785, opened 2022-08-08
fix: export align and direction info

### LINKED ISSUE #2775: $generateHtmlFromNodes does not serialize node with applied "text-align" styles.
Lexical version: @lexical/html - 0.3.8

## Steps To Reproduce
See the linked reproduction using the default "RichText" editor.

1. Create a text node
2. set node to be align "center"
3. click the serialize button
4. View serialized html in console and notice that the nodes have no "styles" applied to them. 

Link to code example:
https://codesandbox.io/s/nervous-babbage-y9brct?file=/src/plugins/SerializePlugin.js

## The current behavior
No styles are being serialized

## The expected behavior
text-align styles should be serialized.

im-adithya:
Fixes #2775

### DIFF
--- a/packages/lexical/src/nodes/LexicalParagraphNode.ts
+++ b/packages/lexical/src/nodes/LexicalParagraphNode.ts
@@ -150,14 +150,8 @@ function convertParagraphElement(domNode: Node): DOMConversionOutput {
   const p = domNode as HTMLElement;
   const node = $createParagraphNode();
   const textAlign = p.style.textAlign;
-  const direction = p.getAttribute('dir');
   if (textAlign) {
     node.setFormat(p.style.textAlign as ElementFormatType);
-  }
-  // TODO: This currently doesn't work as it gets over-ridden by dynamic
-  // evaluation of text content during reconciliation.
-  if (direction === 'rtl' || direction === 'ltr') {
-    node.setDirection(direction);
   }
   return {node};
 }
--- a/packages/lexical/src/nodes/LexicalParagraphNode.ts
+++ b/packages/lexical/src/nodes/LexicalParagraphNode.ts
@@ -13,7 +13,10 @@ import type {
   DOMExportOutput,
   LexicalNode,
 } from '../LexicalNode';
-import type {SerializedElementNode} from './LexicalElementNode';
+import type {
+  ElementFormatType,
+  SerializedElementNode,
+} from './LexicalElementNode';
 import type {Spread} from 'lexical';
 
 import {getCachedClassNameArray} from '../LexicalUtils';
@@ -66,6 +69,25 @@ export class ParagraphNode extends ElementNode {
 
     if (element && this.isEmpty()) {
       element.append(document.createElement('br'));
+    }
+
+    const format = this.getFormatType();
+    if (element && format) {
+      element.style.cssText += 'text-align: ' + format;
+    }
+
+    const direction = this.getDirection();
+    if (element && direction) {
+      element.setAttribute('dir', direction);
+
+      const ltrClassName = editor._config.theme.ltr;
+      const rtlClassName = editor._config.theme.rtl;
+
+      if (direction === 'ltr' && ltrClassName) {
+        element.classList.add(ltrClassName);
+      } else if (direction === 'rtl' && rtlClassName) {
+        element.classList.add(rtlClassName);
+      }
     }
 
     return {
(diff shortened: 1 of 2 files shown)


thegreatercurve:
Related to #2610

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -63,11 +64,22 @@ export class ParagraphNode extends ElementNode {
...
 
+    if (element && format !== 0) {
+      element.style.cssText = 'text-align: ' + ELEMENT_FORMAT_TO_TYPE[format];
+    }
+
+    if (element && dir) {
+      element.setAttribute('dir', dir);
+      element.classList.add('PlaygroundEditorTheme__' + dir);
+    }
+
```
thegreatercurve, line 82: The playground class name is user definable and there are some convenience methods on the super class we can use instead:

```suggestion
    if (element) {
      element.style.cssText += 'text-align: ' + this.getFormatType();
    }

    const direction = this.getDirection();

    if (element && direction) {
      element.setAttribute('dir', direction);

      const ltrClassName = editor._config.theme.ltr;
      const rtlClassName = editor._config.theme.rtl;

      if (direction === 'ltr' && ltrClassName) {
        element.classList.add(ltrClassName);
      } else if (direction === 'rtl' && rtlClassName) {
        element.classList.add(rtlClassName);
      }
    }
```

im-adithya: Got it!
Is this intentional? As we would be having the format set to 0 when nothing is applied, which results in `text-align: ;` being added
```
'text-align: ' + this.getFormatType();
```

thegreatercurve:
LGTM @im-adithya. Just a couple of small changes. 

It's not essential to this PR, but if you have some spare time, it might be worth adding a unit test for this in `@lexical/html`, which will test `$generateNodesFromDOM` to make sure the correct direction and alignment are being added. 

You can use the below logic to test the editor update:

```
const html = '<p class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr" dir="ltr"><span>asdfasdf</span></p><p class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr" dir="ltr"><span>asdfasdf</span></p><p class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__rtl" dir="rtl"><span>شسيبشسيب</span></p><p class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__rtl" dir="rtl" style="text-align: left;"><span>شسيبشسيب</span></p>'

editor.update(() => {
    const dom = parser.parseFromString(html, 'text/html');
    const root = $getRoot();
    root.clear().append(...nodes);
});
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
thegreatercurve: Just updated that to be `if (element && this.getFormatType() !== '') {`

im-adithya:
Sure!

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -124,8 +146,20 @@ export class ParagraphNode extends ElementNode {
...
+function convertParagraphElement(domNode: Node): DOMConversionOutput {
+  const p = domNode as HTMLElement;
+  const node = $createParagraphNode();
+  const textAlign = p.style.textAlign;
+  const direction = p.getAttribute('dir');
+  if (textAlign) {
+    node.setFormat(p.style.textAlign as ElementFormatType);
+  }
+  // TODO: This currently doesn't work as it gets over-ridden by dynamic
+  // evaluation of text content during reconciliation.
```
im-adithya, line 158: Related to https://github.com/facebook/lexical/issues/2610

im-adithya:
> Fixes #2610

I don't think this solves the issue yet as the reconciliation still overrides the RTL setting, though this can be considered as one step towards complete implementation of Bidirectional Support.

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -124,8 +146,20 @@ export class ParagraphNode extends ElementNode {
...
+  if (direction === 'rtl' || direction === 'ltr') {
+    node.setDirection(direction);
+  }
+  return {node};
```
thegreatercurve, line 162: @im-adithya Maybe remove this for the moment if this doesn't solve the original bug.

im-adithya: Sure!

im-adithya: Should I change the behavior in unit tests (breaking currently) or add this back?
Because this does make the helper functions work as expected (which is what we are testing in unit tests), but the bug occurs in the playground due to reconciliation...

tylerjbainbridge:
@im-adithya I'm closing this for now, but feel free to reopen if you're able to drive it to completion.

PR #2790, opened 2022-08-08
chore(auto-link-plugin): fix invariant message for node registration check

echarles:
Fix the message when checking for the node registration in the autolink plugin.


acywatson:
Thanks for this! I think we need to run ```npm run extract-codes``` to update the codes.json file.

echarles:
> Thanks for this! I think we need to run npm run extract-codes to update the codes.json file.

Done.

In the meantime my CLA is confirmed.

acywatson (approved):

PR #2798, opened 2022-08-09
Add TableOfContentsPlugin to wwwSyncs

karam-qaoud:


zurfyx (changes requested):
Actually, hold on, my bad, we may not need this since it lives inside `@lexical/react`

PR #2802, opened 2022-08-09
Fixed flow return type for TableOfContents plugin

karam-qaoud:


fantactuka (approved):

packages/lexical-react/flow/LexicalTableOfContents__EXPERIMENTAL.js.flow
```diff
@@ -10,8 +10,8 @@
 import type {HeadingTagType} from '@lexical/rich-text';
 import type {NodeKey} from 'lexical';
 
-declare export function LexicalTableOfContentsPlugin(
+declare export default function LexicalTableOfContentsPlugin({
```
thegreatercurve, line 13: This shouldn't be a default export but a named export like the other React packages.

thegreatercurve (approved):

PR #2811, opened 2022-08-10
prevent button from submitting forms

gdelmas:
in cases where the debugger is a decendent of a form, the button will by default submit. preventing this.

### DIFF
--- a/packages/lexical-react/src/LexicalTreeView.tsx
+++ b/packages/lexical-react/src/LexicalTreeView.tsx
@@ -154,7 +154,8 @@ export function TreeView({
               setTimeTravelEnabled(true);
             }
           }}
-          className={timeTravelButtonClassName}>
+          className={timeTravelButtonClassName}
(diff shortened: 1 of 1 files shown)


acywatson (approved):
Thanks!

packages/lexical-react/src/LexicalTreeView.tsx
```diff
@@ -145,7 +145,8 @@ export function TreeView({
     <div className={viewClassName}>
       {!timeTravelEnabled && totalEditorStates > 2 && (
         <button
-          onClick={() => {
+          onClick={(event) => {
+            event.preventDefault();
```
a-xin, line 149: LGTM, but thinking that explicitly setting the default behaviour of the button by adding a type is more descriptive: `type="button"`

gdelmas: that is much better. changed

PR #2814, opened 2022-08-11
fixed getStyleObjectFromRawCSS to handle css values with a colon

dosatross:
Currently if a css property has a colon in its value the rest of the value following the colon gets cut off.
Ran into this when using a font family with a prefix e.g. `font-family: prefix:Arial`

### DIFF
--- a/packages/lexical-selection/src/index.ts
+++ b/packages/lexical-selection/src/index.ts
@@ -310,8 +310,8 @@ function getStyleObjectFromRawCSS(css: string): Record<string, string> {
 
   for (const style of styles) {
     if (style !== '') {
-      const [key, ...value] = style.split(':');
-      styleObject[key.trim()] = value.join(':').trim();
+      const [key, value] = style.split(/:([^]+)/); // split on first colon
+      styleObject[key.trim()] = value.trim();
     }
   }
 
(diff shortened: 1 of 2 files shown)


acywatson:
Nice, thanks for writing a test!

packages/lexical-selection/src/index.ts
```diff
@@ -310,8 +310,8 @@ function getStyleObjectFromRawCSS(css: string): Record<string, string> {
 
   for (const style of styles) {
     if (style !== '') {
-      const patch = style.split(':');
-      styleObject[patch[0].trim()] = patch[1].trim();
+      const [key, ...value] = style.split(':');
+      styleObject[key.trim()] = value.join(':').trim();
```
acywatson, line 314: I guess another way to go at this would be Regex capture groups, which might be more performant than all this string manipulation. Thoughts?

dosatross: im not great with regex but gave it a go - lmk what you think

karam-qaoud: Great catch, the regex version looks much cleaner!

acywatson (approved):
Looks right to me, pending CI pass

packages/lexical-selection/src/index.ts
```diff
@@ -310,8 +310,8 @@ function getStyleObjectFromRawCSS(css: string): Record<string, string> {
 
   for (const style of styles) {
     if (style !== '') {
-      const patch = style.split(':');
-      styleObject[patch[0].trim()] = patch[1].trim();
+      const [key, value] = style.split(/:([^]+)/); // split on first colon
```
acywatson, line 313: I had actually forgotten you could use the "^" character this way in JS RegEx!

Issue #2806, opened 2022-08-10
Bug: xss vulnerability within links?


maxhelsel:
Lexical version: lexicon playground 

## Steps To Reproduce

1. type a word
2. highlight word
3. click add link icon
4. type anything in there, something like 'javacript:alert('xss')'

seems you can enter anything inside the link prompt

![Screen Shot 2022-08-10 at 10 47 43 AM](https://user-images.githubusercontent.com/46327674/183936846-04884c29-10e1-4588-8198-6304c0cb0e86.png)

Link to code example: [https://playground.lexical.dev/](https://playground.lexical.dev/)

acywatson:
Yea, we should probably have some basic, but configurable, validation here. Also might be nice to set the rel attribute to head off reverse tabnabbing.

Seems like a good first issue :)

heesungjang:
I would love to work on this issue! Could you please assign this to me if not assigned to anybody yet? thanks!

PR #2825, opened 2022-08-12
fix: added basic validation for unsafe url and unsafe data url

heesungjang:
XSS-video (1)

Hi, this is my very first open source pull request. Any comment would be a great learning source for me.

### Problem

xss vulnerability 

- when adding an url for a link, where was no validation logic for potential javascript injection for unsafe urls (javascript: or data:).

### Solution

- I have implemented function called sanitizeUrl that validates event.target.value before it's passed to setLinkUrl().

- When user enters unsafe url string starts with javascript: or data: , it returns default https://

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28813,7 +28813,7 @@
       "dev": true,
       "requires": {
         "@types/node": "*",
-        "playwright-core": "1.23.0-next-alpha-trueadm-fork"
+        "playwright-core": "1.23.1"
       }
     },
     "@polka/url": {
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
@@ -258,6 +258,22 @@ function FloatingLinkEditor({editor}: {editor: LexicalEditor}): JSX.Element {
     return true;
   }, [editor]);
 
+  const sanitizeUrl = (url: string): string => {
+    /** A pattern that matches safe  URLs. */
+    const SAFE_URL_PATTERN =
+      /^(?:(?:https?|mailto|ftp|tel|file|sms):|[^&:/?#]*(?:[/?#]|$))/gi;
+
+    /** A pattern that matches safe data URLs. */
+    const DATA_URL_PATTERN =
+      /^data:(?:image\/(?:bmp|gif|jpeg|jpg|png|tiff|webp)|video\/(?:mpeg|mp4|ogg|webm)|audio\/(?:mp3|oga|ogg|opus));base64,[a-z0-9+/]+=*$/i;
+
+    url = String(url).trim();
+
+    if (url.match(SAFE_URL_PATTERN) || url.match(DATA_URL_PATTERN)) return url;
+
+    return `https://`;
+  };
+
   useEffect(() => {
     const onResize = () => {
       editor.getEditorState().read(() => {
@@ -310,7 +326,7 @@ function FloatingLinkEditor({editor}: {editor: LexicalEditor}): JSX.Element {
           className="link-input"
           value={linkUrl}
           onChange={(event) => {
-            setLinkUrl(event.target.value);
+            setLinkUrl(sanitizeUrl(event.target.value));
           }}
           onKeyDown={(event) => {
             if (event.key === 'Enter') {


zurfyx:
That's a really good find, thank you Heesung! Given that JavaScript can be injected into the doument in multiple ways (copy-paste, import, collab) I wonder if you should validate either:
1. The current state
2. Warn before leaving to another domain

**1. The current state**

Create a plugin that removes a bad href or extend the @lexical/link

```
registerNodeTransform(LinkNode, (node) => {
 ...
});
registerNodeTransform(AutoLinkNode, (node) => {
 ...
});
```

Override createDOM HTML in @lexical/link

```
element.href = sanitize(this.__url);
```

**2. Warn before leaving**

Add a new React plugin that intercepts when clicking links and adds a modal/intertitial when going outside the domain.

---

For the sake of simplicity and given that it's rare to find a case where JavaScript would be directly supported as part of the input, I'd probably just go for 'Override createDOM HTML in @lexical/link'. WDYT?

zurfyx:
Nit; let's drop the changes on your package-json lock in this PR. Create a separate PR so that we can track it independently please!

acywatson:
Fixes #2806

acywatson:
> That's a really good find, thank you Heesung! Given that JavaScript can be injected into the doument in multiple ways (copy-paste, import, collab) I wonder if you should validate either:
> 
> 1. The current state
> 2. Warn before leaving to another domain
> 
> **1. The current state**
> 
> Create a plugin that removes a bad href or extend the @lexical/link
> 
> ```
> registerNodeTransform(LinkNode, (node) => {
>  ...
> });
> registerNodeTransform(AutoLinkNode, (node) => {
>  ...
> });
> ```
> 
> Override createDOM HTML in @lexical/link
> 
> ```
> element.href = sanitize(this.__url);
> ```
> 
> **2. Warn before leaving**
> 
> Add a new React plugin that intercepts when clicking links and adds a modal/intertitial when going outside the domain.
> 
> For the sake of simplicity and given that it's rare to find a case where JavaScript would be directly supported as part of the input, I'd probably just go for 'Override createDOM HTML in @lexical/link'. WDYT?

Good ideas, but I think we need to be careful about injecting to much opinion into the packages that hold the lower-level shared functionality. Whatever we do in there should probably have an escape hatch configuration option (to override the sanitization pattern, for instance).

We should also think about the UX here. If we just return "https://" as the href of the HTMLAnchorElement returned from createDOM, then we'll just put a broken link in the document that will look like a working link. What we probably need to do is either unwrap the contents of the LinkNode via a transform or something or just leave that up to the package users.

"warn before leave" or any sort of "shimming" behavior strikes me as a pretty surface-specific thing that probably shouldn't live in the base package unless it's totally configurable.

Maybe there is some other approach we're not thinking about yet?

zurfyx:
@acywatson The plugins can potentially live on the playground but I've got no objection into sanitizing links by default as this is like the 99% desired behavior and easy to miss this security hole. That said, I do agree the `https://` prefix is unideal and would also be nice to provide a customize the sanitization logic.

acywatson:
> @acywatson The plugins can potentially live on the playground but I've got no objection into sanitizing links by default as this is like the 99% desired behavior and easy to miss this security hole. That said, I do agree the `https://` prefix is unideal and would also be nice to provide a customize the sanitization logic.

@zurfyx agree, I'm just not sure where to put this logic. If you put it in @lexical/link, how do you notify the caller about a validation error when they try to input an invalid URL? I guess we can plumb through an onValidation listener or something?  If you put it in LinkPlugin command listener, your sort of have the same problem. How do you handle validation in the failure case and report that back to the editor so they can handle it however they want? I guess we could fire another LINK_URL_VALIDATION_FAILURE_COMMAND, but a callback in the props seems more straightforward.

If we silently sanitize (i.e., just change the url) in our implementation, that might create unexpected behavior, right?

@heesungjang do you have any ideas here?

heesungjang:
Hi,

First of all, thank you both @acywatson and @zurfyx for your opinion.

To be honest, I'm trying to digest the conversation between @acywatson and @zurfyx. I would need more time to look closer at the lexical source code.

However, currently, I'm looking into other open source editors like [TOAST UI](https://github.com/nhn/toast-ui.doc) and others to see how they've tackled the current issue.

heesungjang:
I agree with the @zurfyx for sanitizing the links by default as that is the desired behavior for link plugin. And therefore, I believe it is reasonable to implement a sanitize url function in @lexical/link that would like something like below

```jsx
whiteListProtocols = ['https', 'ftp', 'ftps', 'mailto', 'news', 'irc', ...]
sanitizeUrl(__url, whiteListProtocols): {
        //...sanitizing

	if(not allowed protocol){
	// return the url without unsafe protocol
  // e.g) if javascript:alert("") --> alert("")
	}
}  

element.href = this.sanitizeUrl(this.__url)
```

However, about notifying the caller for an invalid input error, I'm not sure to which extent it should notify them. For example, some users might put a link without a protocol at all like "youtube.com". Is this considered an invalid URL too?

acywatson:
> I agree with the @zurfyx for sanitizing the links by default as that is the desired behavior for link plugin. And therefore, I believe it is reasonable to implement a sanitize url function in @lexical/link that would like something like below
> 
> ```js
> whiteListProtocols = ['https', 'ftp', 'ftps', 'mailto', 'news', 'irc', ...]
> sanitizeUrl(__url, whiteListProtocols): {
>         //...sanitizing
> 
> 	if(not allowed protocol){
> 	// return the url without unsafe protocol
>   // e.g) if javascript:alert("") --> alert("")
> 	}
> }  
> 
> element.href = this.sanitizeUrl(this.__url)
> ```
> 
> However, about notifying the caller for an invalid input error, I'm not sure to which extent it should notify them. For example, some users might put a link without a protocol at all like "youtube.com". Is this considered an invalid URL too?

The concern I have, overall, is that we're taking an opinionated stance on how any consumer of @lexical/link should deal with "unsafe" or "invalid" URLs. We're saying that the right thing to do in that case is always to silently chop off the protocol and create the link anyway with the remaining URL string. I think that people will ultimately want to customize that behavior, to some extent. For instance, Google Docs shows an error in the UI saying "please use a valid URL".

However, thinking about this as I'm writing, what you're doing can serve as a sane default behavior, and people can choose to sanitize the URL before they even call toggleLink and do whatever they want in the UI at that point.

The only other nit I have is that the preferred terminology is now "allowlist", rather than "whitelist". Other than that, I'm ok with your approach.

heesungjang:
I will create a separate PR so that we can track it independently!

thegreatercurve (approved):

package-lock.json
```diff
@@ -28813,7 +28813,7 @@
       "dev": true,
       "requires": {
         "@types/node": "*",
-        "playwright-core": "1.23.0-next-alpha-trueadm-fork"
+        "playwright-core": "1.23.1"
```
thegreatercurve, line 28816: Why has this been upgraded here? What was failing about our local fork?

heesungjang: I have deleted this fork repository from github and local when this PR was closed previously. Do you mind if i open seperate PR without package-lock.json?

acywatson: > I have deleted this fork repository from github and local when this PR was closed previously. Do you mind if i open seperate PR without package-lock.json?

You can do it as a separate PR if you want - no problem

heesungjang: @acywatson opened new PR #2900 !!

PR #2831, opened 2022-08-13
fix: set cursor grab when image can be dragged

LuciNyan:
## Description
set cursor grab when image can be dragged. 

#### Additionally
I've tried a lot but still can't resolve the problem of not being able to set the cursor when dragging

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -937,7 +937,6 @@ i.prettier-error {
 .editor-shell .editor-image img.focused {
   outline: 2px solid rgb(60, 132, 244);
   user-select: none;
-  cursor: move;
   cursor: grab;
 }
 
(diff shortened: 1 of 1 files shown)


fantactuka:
Have you tried to set cursor style to document body while dragging?

LuciNyan:
> Have you tried to set cursor style to document body while dragging?

Yes, I tried, but it didn't work.

fantactuka:
Even with "!important"? 😯

LuciNyan:
> Even with "!important"? 😯

Yes... This is strange. Will it work on windows? I'm curious.

fantactuka (approved):

fantactuka:
> > Even with "!important"? 😯
> 
> Yes... This is strange. Will it work on windows? I'm curious.

Looks like browser forces cursor if has drag data: https://bugs.chromium.org/p/chromium/issues/detail?id=1196385

PR #2833, opened 2022-08-13
fix: add fallback for code formatting

im-adithya:
Adds fallback for FigmaNode similar to Youtube and Twitter

https://user-images.githubusercontent.com/64399555/184502809-94206350-ce8d-4b82-b57a-5e2d0fe78b42.mp4


fantactuka (commented):
Should it use getTextConteny instead? Can override it for all these decorators and simplify this code. It also used for plain text clipboard so copying decorator won't end up with getting nothing for a plan text

zurfyx (approved):

zurfyx:
Thanks but let's work on @fantactuka before merging (sorry, read his comment too late)

im-adithya:
Done!

Issue #1517, opened 2022-03-23
API Documentation: lexical core


acywatson:
Add API documentation for Lexical Core. We can probably just copy the TypeScript declarations for now and add a little bit of context to each one. Later, these will be auto-generated.

Issue #1527, opened 2022-03-23
API Docs: @lexical/react


acywatson:
Add API documentation for @lexical/react. We can probably just copy the TypeScript declarations for now and add a little bit of context to each one. Later, these will be auto-generated.

etrepum:
These are autogenerated, but the exports do still need more doc comments

Issue #1525, opened 2022-03-23
API Docs: @lexical/table


acywatson:
Add API documentation for @lexical/table. We can probably just copy the TypeScript declarations for now and add a little bit of context to each one. Later, these will be auto-generated.

Issue #1524, opened 2022-03-23
API Docs: @lexical/list


acywatson:
Add API documentation for @lexical/list. We can probably just copy the TypeScript declarations for now and add a little bit of context to each one. Later, these will be auto-generated.

PR #2837, opened 2022-08-15
Add autogenerated TypeDoc docs

acywatson:
This should help a bit with documentation woes. We'll have to iteratively go through and add comments to functions for this to be really helpful.

One thing to note is that the titles of the "modules" in the sidebar are wrong and can't be overridden without a comment in the file header (has to be at the top of the file). We have a lint rule that was silently overwriting all my changes in the pre-commit 🙄  so I'll need to fix that to support the module name override eventualy.


zurfyx (approved):
As per our offline discussion, I think it's a good idea even if that doesn't solve all our documentation needs 👍

acywatson:
Closes #1517 #1519 #1520 #1521 #1522 #1523 #1524 #1525 #1527

PR #2841, opened 2022-08-15
Support selection in readonly mode

acywatson:
I saw this:

https://github.com/facebook/lexical/pull/1339

which looks like the only code path where this function is used. I can't seen any regression around Decorator selection, but maybe there is something I am missing - not sure what the original issue was here.


fantactuka (approved):

acywatson:
Closing this because it turns out that is has a lot of unexpected effects.

PR #2842, opened 2022-08-16
Only normalize text nodes into list item nodes if they are not empty

### LINKED ISSUE #2807: Bug: $generateNodesFromDOM creates empty list items from whitespace between actual list items
Lexical version: 0.3.8

## Steps To Reproduce

1. Use the following code to initialize Lexical:

```js
  editorState: (editor) => {
    const dom = parser.parseFromString(
      "<ul><li>foo</li> <li>bar</li> <li>banana</li></ul>",
      "text/html"
    );
    const rootNode = $getRoot();
    rootNode.append(...nodes);
  }
```

2. Observe that the empty spaces between the `<li>` elements get converted to empty list item nodes:

image

3. If you delete the extra spaces, the list will have the correct number of list items.

Link to code example: https://codesandbox.io/s/lexical-list-parsing-bug-ktssb4?file=%2Fsrc%2FEditor.js%3A1849-2160

## The current behavior

Empty list items are created from whitespace text nodes between list item nodes.

## The expected behavior

The empty whitespace between list items should be ignored.

knpwrs:
Fixes #2807. The issue with wrapping empty text nodes in list item nodes as a default normalization is that this assumes the incoming HTML is minified.

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -6,14 +6,13 @@
  *
  */
 
-import {$isTextNode, Spread} from 'lexical';
-
 import {
   addClassNamesToElement,
   removeClassNamesFromElement,
 } from '@lexical/utils';
 import {
   $createTextNode,
+  $isTextNode,
   DOMConversionMap,
   DOMConversionOutput,
   EditorConfig,
@@ -23,6 +22,7 @@ import {
   LexicalNode,
   NodeKey,
   SerializedElementNode,
+  Spread,
 } from 'lexical';
 
 import {$createListItemNode, $isListItemNode, ListItemNode} from '.';
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -247,7 +247,7 @@ function normalizeChildren(nodes: Array<LexicalNode>): Array<ListItemNode> {
           normalizedListItems.push(wrapInListItem(child));
         }
       });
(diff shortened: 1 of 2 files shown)


knpwrs:
Which tests? I don't see any that exercise `importDOM`.

acywatson:
> Which tests? I don't see any that exercise `importDOM`.

You can write a test here:

https://github.com/facebook/lexical/blob/b7887a02780e40640cf0e7582212f9c29730c14d/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs#L1516

Just copy and paste it over and change the title, pasted HTML, and assertion.

Looks like there's also a lint error.

knpwrs:
This should be ready for another workflow run

acywatson:
> This should be ready for another workflow run

Not sure why all the checks didn't run - looks like only 3 did. Can you push a dummy commit and see if that gives me the option to re-trigger the workflow?

knpwrs:
Just pushed an empty commit in 1f373ff

knpwrs:
It seems like at least some of the failures aren't related to this change.

acywatson:
> It seems like at least some of the failures aren't related to this change.

In the test output, I see some flaky tests, but the only failure is the test you wrote for this. I think you need to remove the html prefix before the backticks in your assertions - there is some magic we do in the assertion to remove whitespace and compare the HTML strings. Instead, you can use single or double quotes and put the HTML on a single line.

thegreatercurve:
Closing this due to PR becoming stale and failing E2Es. Feel free to reopen though later.

PR #2846, opened 2022-08-17
Adjust Typeahead Styles

tylerjbainbridge:
Right now the Typeahead Menu moves as you type and positions itself with the cursor instead of the trigger, this PR locks the left value at the trigger.

Before:
image

After:
image


thegreatercurve (approved):
LGTM

fantactuka (approved):
👍

PR #2848, opened 2022-08-18
Small toolbar style fix

thegreatercurve:
The toolbar sometimes shows it's scrollbar at max width if the first dropdown contains a long enough string #2848

Fixes the below:

image

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -48,7 +48,7 @@ header h1 {
 .editor-shell {
   margin: 20px auto;
   border-radius: 2px;
-  max-width: 1000px;
+  max-width: 1100px;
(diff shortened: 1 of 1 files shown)


fantactuka:
We might need to consider nested dropdowns at this point 😁
Not sure if it's still the case, but some collab failed in the past as it wasn't able to click on right-most dropdown (as it's using split view to show two editors). In that case will need to increase playwright browser's start width to be maxWidth x2

fantactuka (approved):

thegreatercurve:
> We might need to consider nested dropdowns at this point 😁

@fantactuka Yeah, I'm starting to think we should have several smaller playgrounds with different examples than our one kitchen sink example. This is somtehing other editors do better.

PR #2855, opened 2022-08-19
⏸ [0.7] Remove indentation keyboard shortcuts in RTE set up

### LINKED ISSUE #2854: Feature: allow rich-text editor Tab-to-Indent to be disabled
Hi there!

We're looking into lexical as part of a feature allowing students to compose rich text responses to an online textbook. However, we're also trying to pick a rich-text editor that will be keyboard accessible.

The problem we're having is that because Tab and Shift-Tab are captured in the editor (to provide indent/unindent support), once the RTE has focus it's not possible to escape it with Tab/Shift-Tab, making other elements on the page inaccessible by keyboard or switch devices.

Could we get an option to disable this indent/unindent behavior for Lexical rich-text editor instances? The expectation is that pressing Tab while inside of the RTE will go to the next focusable item, and pressing Shift-Tab will go to the previous focusable item.

thegreatercurve:
Stealing the tab keyboard shortcut prevents users from being able to tab in and out of the editor using TAB or SHIFT+TAB.

We should remove this altogether, so that users don't proactively have to add in plugins to make their editors more accessible.

Fixes #2854

### DIFF
--- a/packages/lexical-react/src/LexicalTabIndentationPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTabIndentationPlugin.tsx
@@ -1,0 +1,50 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {
+  $getSelection,
+  $isRangeSelection,
+  COMMAND_PRIORITY_EDITOR,
+  INDENT_CONTENT_COMMAND,
+  KEY_TAB_COMMAND,
+  OUTDENT_CONTENT_COMMAND,
+} from 'lexical';
+import {useEffect} from 'react';
+
+/**
+ * This plugin adds the ability to indent content using the tab key. Generally, we don't
+ * recommend using this plugin as it could negatively affect acessibility for keyboard
+ * users, causing focus to become trapped within the editor.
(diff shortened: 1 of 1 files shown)


acywatson:
Just to externalize our internal discussion -

This is a breaking change - it will change the behavior of every editor using the rich-text-plugin. If we decide to move forward, we can handle this with a migration path in the documentation, but we need to bump the version to 0.4 for this change.

acywatson:
@thegreatercurve can we include this in 0.4 except just pull the logic you deleted out into a separate plugin that we can offer as a migration path to restore the functionality to rich-text-plugin users?

EgonBolton:
So, do you finally plan to completely remove the command?
I insist, just as I said here https://github.com/facebook/lexical/issues/2854
I think it would be nice if it could be optionally enabled.

Awjin:
> @thegreatercurve can we include this in 0.4 except just pull the logic you deleted out into a separate plugin that we can offer as a migration path to restore the functionality to rich-text-plugin users?

+1 preserving the current behavior as a plugin. Correct in/out-denting is one of Lexical's standout features. No other RTE library does this correctly for nested lists.

acywatson:
> So, do you finally plan to completely remove the command?
I insist, just as I said here https://github.com/facebook/lexical/issues/2854
I think it would be nice if it could be optionally enabled.


Yea, I think we're on the same page. We're just going to make the core stuff optimized for accessibility by default, but have a plugin that people can use to add this behavior and configure their own escape mechanisms for accessibility.

acywatson (approved):

PR #2859, opened 2022-08-20
Add watch mode for auto-gen doc comments in dev

acywatson:
This makes it so that adding a JSDOC comment to a function in one of the packages automatically does an incremental rebuild of the website in dev mode (npm run start:website)


echarles:
Thx @acywatson. This is very useful.

Issue #2115, opened 2022-05-09
Feature: draggable block


echarles:
It would be great to have draggable blocks ready to be used (in core or as an example), see the notion like experience.

![Untitled5](https://user-images.githubusercontent.com/226720/167362676-73ad647d-12e4-4aa5-8d21-97ff9f99d82c.gif)

echarles:
Work is being done in https://github.com/facebook/lexical/pull/2027 that could help to get the block drag/drop feature alive.

oskar-szwajkowski:
Could you share your opinions about how could it be achieved in lexical?

Up to this point I was using Draft.js, and there I implemented drag & drop by wrapping each block inside draggablewrapper, then saving blockkey on drag event, and when dropping, changing order of blocks array to reflect drag & drop behavior

Should implementing custom drag & drop in lexical works similarly? However I cannot figure out the way to create general 'wrapper' around every non-inline element
(Other use case of that wrapper would be count up amount of open comments in this paragraph/block)

DaniGuardiola:
Bumping this as we also need this feature really badly. What @oskar-szwajkowski shared is quite similar to our thinking. Two main challenges stand out:

- Overriding the nodes to wrap them with a drag and drop handle. Ideally, in a universal way (wrap all block nodes without duplicating too much code) and with the possibility of using react elements. Thinking that https://github.com/facebook/lexical/issues/1262 will be a step forward here.
- Showing drop targets in the right places to visually indicate where the block would land if dropped.

Would really love to learn how to implement something like this.

hustlefueled:
+1 to @DaniGuardiola and @oskar-szwajkowski .

Being able to readjust content blocks will be super helpful. Currently, you need to cut and paste elements to realign and this creates formatting issues. 

Would love to see this come to Lexical.

acywatson:
Acknowledging this thread - thanks for the feedback and thoughts here! Will try to follow up with some ideas on how we might be able to achieve something like this.

LuciNyan:
Hi  @oskar-szwajkowski.

I'm sorry for missing your mention. For this feature, I tried a few different implementations. In my opinion the drag and drop implementation should not be intrusive to any other code. The user is free to choose whether or not to add this plugin, so I think the best implementation would be to add a draggable menu. When dragging happens it's essentially just dragging an icon, and we can customize the content displayed by way of setDragImage.

songispm:
anything new?

echarles:
[...]

PR #2860, opened 2022-08-20
feat(lexical-playground): draggable block

LuciNyan:
## Description
Add draggable block plugin

https://user-images.githubusercontent.com/22126563/190201193-3784dc18-7214-4b48-afba-ade0b5aa288d.mov

Close: #2115


hustlefueled:
I tried the deployment on the playground. Don't think this works yet. Or am I too early here? 🤔

LuciNyan:
> I tried the deployment on the playground. Don't think this works yet. Or am I too early here? 🤔

Yes, It's WIP, there is currently no drop handling.

LuciNyan:
Hi @hustlefueled. Now it's working properly, but I haven't finished processing all the cases.

echarles:
Tested on the deployed playground and it works beautifully for paragraphs, images... move. Maybe the handle could be vertically positioned at the top instead of at the middle (not sure what other editors do).

hustlefueled:
Works for me too now. Looks good.

LuciNyan:
> This PR is amazing! @LuciNyan do you plan any changes to it, otherwise can help with rebasing if needed and it should go to a playground!

Hi @fantactuka. This PR is not yet able to be used, it has some flaws and doesn't cover all cases. I'm sorry I may not be able to fix them for a while because my baby is sick, but I'll be back as soon as I can.

LuciNyan:
Hi @fantactuka @acywatson! This PR is ready to be reviewed! XD

packages/lexical-playground/src/index.css
```diff
@@ -51,7 +51,7 @@ header h1 {
   max-width: 1100px;
   color: #000;
   position: relative;
-  line-height: 20px;
+  line-height: 1.7;
```
acywatson, line 54: Is this somehow related to the draggable block functionality?

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.css
```diff
@@ -0,0 +1,34 @@
+.draggable-block-menu {
+  padding: 2px 1px;
+  cursor: grab;
+  position: absolute;
+  left: -10000px;
+  top: -10000px;
+
```
acywatson, line 7: nit: unnecessary line break

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+  const anchorElementRect = anchorElem.getBoundingClientRect();
+  const topLevelNodeKeys = getTopLevelNodeKeys(editor);
+
+  let blockElem: HTMLElement | null = null;
+
+  editor.getEditorState().read(() => {
+    let index = getCurrentIndex(topLevelNodeKeys.length);
+    let direction = TBD;
+
+    while (index >= 0 && index < topLevelNodeKeys.length) {
```
fantactuka, line 72: It's a lot of calculations for on move handler :) I think it should be good to go since it adds great feature to the playground, but might be interesting to improve it further, e.g. what are your thoughts on:

```js
let lastTarget: HTMLElement | null = null;
let lastTopNode: LexicalNode | null = null;

const onMove = (event: MouseEvent) => {
  const target = event.target as HTMLElement;

  if (lastTarget === target) {
    return;
  }

  lastTarget = target;

  editor.update(() => {
    const node = $getNearestNodeFromDOMNode(target);

    if (node === null) {
      return;
    }

    const topLevelNode = node.getTopLevelElement();

    if (topLevelNode === null || topLevelNode === lastTopNode) {
      return;
    }

    lastTopNode = topLevelNode;
    positionMenuForNode(topLevelNode);
  });
};

const onMouseLeave = () => {
  positionMenuForNode(null);
}

rootElement.addEventListener('mouseleave', onMouseLeave);
rootElement.addEventListener('mousemove', onMove);


```

fantactuka (approved):

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -324,7 +314,7 @@
   list-style-position: inside;
 }
 .PlaygroundEditorTheme__listItem {
-  margin: 8px 32px;
+  margin: 0 32px;
```
acywatson, line 317: Not sure how these styling changes are related?

acywatson (commented):
Looks great! Couple of questions, but I would like to ultimately move this to @lexical/react.  We can merge it into the playground for now and then do that part later.

Thank you!

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
DaniGuardiola: This isn't a great constrain to enforce for this plugin to work, because margin collapse is very useful for text content. I'm guessing block margins conflict with showing the drop targets in the right places. Hopefully there's a solution to that that doesn't involve disallowing margins in block nodes.

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+const SPACE = 4;
+const TARGET_LINE_HALF_HEIGHT = 2;
+const DRAGGABLE_BLOCK_MENU_CLASSNAME = 'draggable-block-menu';
+const DRAG_DATA_FORMAT = 'application/x-lexical-drag-block';
+const TEXT_BOX_HORIZONTAL_PADDING = 28;
+
+const Downward = 1;
+const Upward = -1;
+const TBD = 0;
```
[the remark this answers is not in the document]
LuciNyan: Thx! Looks 'Indeterminate' better. XD

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
LuciNyan: Before:
image

After:
image

This is partly because it is difficult to calculate the correct position of the "drop target", and I do think that rendering the line height by font-size * line-height(1.7) is the right way to go. Maybe I'm missing something and I'm wondering if there's a particular reason why we need to choose a fixed line-height(20px) + margin-bottom to render the line height the way we want?

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
LuciNyan: This approach is really great and reduces a lot of calculations. That's what I did when I first implemented it, and it works well in some scenarios. But it can't handle all scenarios. As an example:

image

When I mouse over the margin area and to the left of the margin area, it does not work properly. Of course, we can solve this problem by wrapping these elements in a div and replacing the margin with padding. I tried to do this, but I found two problems with it.

1. we need to make changes to the LexicalReconciler implementation.
2. we need to ensure that the outer layer of the visual content of all the TopLevelNodes is padding. Just like this:

image

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
LuciNyan: > Hopefully there's a solution to that that doesn't involve disallowing margins in block nodes.

If 'margin collapse' is necessary, I suppose I could add some calculations, such as calculating the margin of the next sibling element, then comparing the sizes to get the correct gap, and dividing by two to get the final position. But by looking at the 'Before' image above, you can see that even this calculated position is not the visual "middle" of the gap between the two TopLevelNodes. This doesn't look perfect. Of course we can add some offset to this to make it look good, but I still want to know if there is a better way.

DaniGuardiola: I wonder if it'd be worth it to provide options to customize the position depending on the block, e.g. setting a "drop indicator position" per block. Not sure how the final position could be computed even then though, since the two blocks could be setting conflicting positions.

DaniGuardiola (commented):
I've been playing with this plugin for a bit and it's pretty awesome! Some thoughts:

- Locally, I had to add anchorEl.scrollTop to the computed top of both the drag handler and the drop target to account for scrollable containers.
- The drop action seems a little flaky. Sometimes it won't work, depending on where your mouse is when releasing the mouse.

LuciNyan:
> * The drop action seems a little flaky. Sometimes it won't work, depending on where your mouse is when releasing the mouse.

I have noticed that this happens when I drag to the very end of the content and then drag down. Do you mean this scenario? If you are talking about other scenarios, Could you provide the steps to reproduce them?

tylerjbainbridge (approved):

fantactuka:
Thanks @LuciNyan for amazing feature! Any improvements can go separately, and I believe many people would be happy to play around this feature as it is

silvergravel:
Any ETA on when / if this implementation will be shifted to `@lexical/react`?

amilich:
> Any ETA on when / if this implementation will be shifted to `@lexical/react`?

is this usable in the react library yet?

ivailop7:
> > Any ETA on when / if this implementation will be shifted to `@lexical/react`?
> 
> 
> 
> is this usable in the react library yet?

This has been available for a year+ in the playground. The plugin code is available inside the playground folder in the repo.

nickpapciak:
I can't get this plugin working in react. I tried copying the way it was done from the playground folder but it did not work. Could I get some help on how to use the DraggableBlockPlugin?

iamdipansh:
> I can't get this plugin working in react. I tried copying the way it was done from the playground folder but it did not work. Could I get some help on how to use the DraggableBlockPlugin?

same, i'm facing the same problem

iamdipansh:
> I can't get this plugin working in react. I tried copying the way it was done from the playground folder but it did not work. Could I get some help on how to use the DraggableBlockPlugin?

I'm having the same issue.

PR #2863, opened 2022-08-20
refactor: add root event handler more readable

yuanyan:


packages/lexical/src/LexicalEvents.ts
```diff
@@ -118,35 +118,31 @@ import {
...
-  ['cut', PASS_THROUGH_COMMAND],
-  ['copy', PASS_THROUGH_COMMAND],
-  ['dragstart', PASS_THROUGH_COMMAND],
-  ['dragover', PASS_THROUGH_COMMAND],
-  ['dragend', PASS_THROUGH_COMMAND],
-  ['paste', PASS_THROUGH_COMMAND],
-  ['focus', PASS_THROUGH_COMMAND],
-  ['blur', PASS_THROUGH_COMMAND],
-  ['drop', PASS_THROUGH_COMMAND],
+  ['keydown', onKeyDown as RootElementEventHandler],
```
acywatson, line 126: It seems weird that we have to cast this?

acywatson (commented):
This seems fine to me, aside from the casting. Not sure why we had that PASS_THROUGH_COMMAND. Let's see if others find it an improvement, as well.

trueadm:
I'm not really sure this makes the handling logic more readable as there seems more indirection and more casting going on.

acywatson:
> I'm not really sure this makes the handling logic more readable as there seems more indirection and more casting going on.

PASS_THROUGH_COMMAND seems more indirect to me. Why should there be frozen empty objects in the list of events instead of just event handlers?

That said, I'm not bothered either way tbh

acywatson:
Thanks for the contribution here, but I don't really see the benefits outweighing the potential costs here. I agree that the current PASS_THROUGH situation makes very little sense from a readability standpoint, but the risk of a performance regression here outweighs the benefits of rearranging this code, I think.

Issue #2856, opened 2022-08-19
Bug: RangeSelection does not update when NodeSelection active


gdelmas:
As can be seen in the video, usually a RangeSelection updates in real-time while selecting. Though when there is a prior NodeSelection the selection only gets updated after mouse up. This leads to two problems:
- the node appears selected at the same time while other text is selected already
- when the cursor is released outside of the editor the selection is not applied at all (but is shown)

I'd propose to fix by real-time selecting in all instances.

https://user-images.githubusercontent.com/2949816/185684805-6ef0401d-8886-4f37-bcc5-e4dd7f9160d8.mov

PR #2872, opened 2022-08-22
Fix delayed RangeSelection on click when NodeSelection/GridSelection

zurfyx:
https://user-images.githubusercontent.com/193447/186028874-b2a0d401-4f4b-43ea-9156-a57d2a8cbbf8.mov

Closes https://github.com/facebook/lexical/issues/2856


trueadm:
Why was this original code added? Would be good to see what the original intent was

zurfyx:
> Why was this original code added? Would be good to see what the original intent was

My hunch is that it was trying to do the exact same except that it never considered that one edge case. Hence, why I copy-pasted the comment into the other file

PR #2878, opened 2022-08-23
Exported DEFAULT_TRANSFORMERS array in react LexicalMarkdownShortcutPlugin

kevinansfield:
The `LexicalMarkdownShortcutPlugin` package extends the defaults from `@lexical/markdown` with a HR transform but didn't expose that default list making it harder for anyone wanting to use the defaults plus their own extensions.

### DIFF
--- a/packages/lexical-react/flow/LexicalMarkdownShortcutPlugin.js.flow
+++ b/packages/lexical-react/flow/LexicalMarkdownShortcutPlugin.js.flow
@@ -9,6 +9,8 @@
 
 import type {Transformer} from '@lexical/markdown';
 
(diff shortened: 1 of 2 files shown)


kevinansfield:
> Could you add it to .flow/.d.ts for proper exports typing

@fantactuka I've just added it to `flow/LexicalMarkdownShortcutPlugin.js.flow`, is that what you meant? Sorry, I'm completely new to flow so wasn't aware it was needed 🙂

zurfyx (approved):
Thank you!

acywatson:
Thanks!

PR #2881, opened 2022-08-24
add the corresponding import to react doc

signalwerk:
In [«Getting Started with React»](https://lexical.dev/docs/getting-started/react) there are the imports shown. But since they are not quite obvious (for a beginner) it's probably good to add them here, at least in the first example so people can then figure out the other imports easily themselves (same logic for all other components)


trueadm (approved):
Thanks!

PR #2882, opened 2022-08-24
Replaced `addTransform` with `registerNodeTransform` in transforms doc

kevinansfield:
`addTransform` was renamed a couple of times (first to `addNodeTransform` then `registerNodeTransform`) but the doc page on transforms still referenced the original naming.

### DIFF
--- a/packages/lexical-website-new/docs/concepts/transforms.md
+++ b/packages/lexical-website-new/docs/concepts/transforms.md
@@ -92,9 +92,9 @@ Transforms are very specific to a type of node. This applies to both the declara
 
 ```js
 // Won't trigger
-editor.addTransform(ParagraphNode, ..)
+editor.registerNodeTransform(ParagraphNode, ..)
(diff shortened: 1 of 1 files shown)


kevinansfield:
Something for a separate issue - it's not entirely clear what "rules 2 and 1" refers to in the  "You move a node via `replace`, rules 2 and 1 are applied."  text.

trueadm (approved):
Thanks

packages/lexical-website-new/docs/concepts/transforms.md
```diff
@@ -107,7 +107,7 @@ You remove a node, its parent is marked dirty, also the node's immediate sibling
 You move a node via `replace`, rules 2 and 1 are applied.
```
trueadm, line 107: I don't even know what this is too, must have been a typo. We can probably remove this line.

acywatson:
Thank you!

PR #2884, opened 2022-08-24
Improve heuristics around node selection and keyboard navigation

trueadm:
Keyboard navigation today around node selection is a bit wonky and broken. Specifically, you can't actually select a node using the keyboard arrow keys properly, nor can you escape node selection and move back to range selection using the keyboard arrow keys.

This PR fixes that problem and also tackles a few accessibility issues along the way with image nodes and horizontal rule nodes too.

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -63,8 +63,10 @@ import {
   INDENT_CONTENT_COMMAND,
   INSERT_LINE_BREAK_COMMAND,
   INSERT_PARAGRAPH_COMMAND,
+  KEY_ARROW_DOWN_COMMAND,
   KEY_ARROW_LEFT_COMMAND,
   KEY_ARROW_RIGHT_COMMAND,
+  KEY_ARROW_UP_COMMAND,
   KEY_BACKSPACE_COMMAND,
   KEY_DELETE_COMMAND,
   KEY_ENTER_COMMAND,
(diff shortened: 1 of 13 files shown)


thegreatercurve (approved):

packages/lexical/src/LexicalEditor.ts
```diff
@@ -798,14 +798,15 @@ export class LexicalEditor {
...
           onUpdate: () => {
-            rootElement.removeAttribute('autocapitalize');
-
             if (callbackFn) {
               callbackFn();
             }
           },
         },
       );
+      setTimeout(() => {
```
acywatson, line 807: What was the problem here?

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1652,6 +1653,14 @@ export class RangeSelection implements BaseSelection {
     // Handle the selection movement around decorators.
     const possibleNode = $getDecoratorNode(focus, isBackward);
     if ($isDecoratorNode(possibleNode) && !possibleNode.isIsolated()) {
+      // Enable it possible to move selection from range selection to
```
acywatson, line 1656: nit: "make it possible"?

packages/lexical-rich-text/src/index.ts
```diff
@@ -705,15 +708,59 @@ export function registerRichText(
       },
       COMMAND_PRIORITY_EDITOR,
     ),
+    editor.registerCommand<KeyboardEvent>(
+      KEY_ARROW_UP_COMMAND,
+      (event) => {
+        const selection = $getSelection();
+        if ($isNodeSelection(selection)) {
```
acywatson, line 715: Might we consider pulling the logic that handles this case into a separate function that can be re-used across the arrow key command listeners?

trueadm: I considered it but it but it made things less readable as the logic is different slightly in each one.

PR #2885, opened 2022-08-24
LexicalTypeaheadMenuPlugin - Increase priority for keyboard commands

ttillberg:
This would enable us to use keyboard shortcuts on `LexicalTypeaheadMenuPlugin` dropdown menu  without necessarily triggering conflicting key commands (ESC, ARROW, ENTER) declared by custom plugins. Currently only the opposite is possible since the priority is already at the lowest. 

Some more detailed background: https://stackoverflow.com/questions/73471798/lexical-key-command-conflict-and-priority


tylerjbainbridge (approved):
Makes sense to me. Thank you!

PR #2894, opened 2022-08-25
Add `SerializedParagraphNode` type export

matthew-plusprogramming:


zurfyx (approved):
Thank you!

acywatson:
You need to export it fro ParagraphNode as well, I think

matthew-plusprogramming:
Quick question @zurfyx. The documentation says that the `SerializedLineBreak` type is not exported, is there a reason? For my use case I think I may also need to handle this type. If it's okay to export, I'll open another PR for that as well

thegreatercurve:
@matthew-plusprogramming You need to export `SerializedParagraphNode` from `LexicalParagraphNode` as well. You can add export for `SerializedLineBreakNode` here as well.

matthew-plusprogramming:
> @matthew-plusprogramming You need to export `SerializedParagraphNode` from `LexicalParagraphNode` as well. You can add export for `SerializedLineBreakNode` here as well.

Thanks for the response. Sorry I've been out for the last few days. The `SerializedParagraphNode` and `SerializedLineBreakNode` types are already being exported in their respective files. If I'm missing anything still, please let me know but I'll push up the `SerializedLineBreakNode` export as well then re-request review!

matthew-plusprogramming:
Reopened with new source branch name and commit https://github.com/facebook/lexical/pull/2914

PR #2896, opened 2022-08-26
Try removing playwright

thegreatercurve:


trueadm:
Have the changes to Playwright been made and landed upstream now?

thegreatercurve:
> Have the changes to Playwright been made and landed upstream now?

It was removed as part of #2825 and the tests were passing, but it looks like we were still importing the fork elsewhere.

PR #2899, opened 2022-08-26
Read only validation server

trueadm:
This PR does two things, on our path to realizing how to validate comments during read-only mode:
- Many of the plugins on the playground have now been made lazy and code-split to improve performance also enable the editor env to be run in a browser-less environment.
- Added a simple editor state verification server that only permits comments being added/removed from an editor state.

To run the server, do `npm run validation` before using the playground.


packages/lexical-playground/src/server/validation.ts
```diff
@@ -0,0 +1,133 @@
...
+    return true;
+  }
+  const prevEditorState = editor.getEditorState();
+  const nextEditorState = editor.parseEditorState(stringifiedJSON);
+  editor.setEditorState(nextEditorState);
+  editor.update(() => {
+    const root = $getRoot();
+    sanitizeNode(root);
+  });
+  await Promise.resolve().then();
```
thegreatercurve, line 75: Why do we need this?

thegreatercurve (approved):

packages/lexical-playground/src/server/validation.ts
trueadm: Updates are async.

PR #2901, opened 2022-08-26
Attempt transform of NodeSelection to RangeSelection on mouseDown

zurfyx:
Different approach (old https://github.com/facebook/lexical/pull/2872) by levering mouse down.

Techical considerations:
- The mousedown event alone is insufficient for to track the new selection. The DOM anchor/focus nodes will still be marked as null.
- Mousedown + mouseup combination is unreliable. FF has a different order: Chrome/Safari: mousedown -> selectionchange -> mouseup; FF (sometimes): mousedown -> mouseup -> selectionchange
- Since mousedown + selectionChange seems reliable across browsers I'm relying on this order in this PR.

https://user-images.githubusercontent.com/193447/186977135-0c511363-d2ce-4d23-a7a7-0fc6654a6ea3.mov

Closes https://github.com/facebook/lexical/issues/2856

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ npm
 .env.test.local
 .env.production.local
 .ts-temp
+.docusaurus
 e2e-screenshots
 test-results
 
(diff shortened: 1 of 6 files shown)


trueadm (approved):
I love it, so easy to follow too. Can you add the e2e for horizontal break to confirm it working again? Thanks!

PR #2910, opened 2022-08-30
Provide markdown plugin node dependencies

trueadm:
The markdown plugin should ensure that certain Lexical nodes are registered with the editor before attempting to apply a markdown transform. To do so, we can specify a list of dependencies for each transform, resulting in invariants if the node hasn't been registered.

This PR also introduces `$getEditor` to the public Lexical API, so the currently active editor is retrievable without having to pass around the editor binding.

### DIFF
--- a/packages/lexical-markdown/flow/LexicalMarkdown.js.flow
+++ b/packages/lexical-markdown/flow/LexicalMarkdown.js.flow
@@ -21,7 +21,7 @@ export type Transformer =
   | TextMatchTransformer;
 
 export type ElementTransformer = {
-  dependencies: Array<Class<LexicalNode>>;
+  dependencies: Array<Class<LexicalNode>>,
   export: (
     node: LexicalNode,
     traverseChildren: (node: ElementNode) => string,
@@ -44,7 +44,7 @@ export type TextFormatTransformer = $ReadOnly<{
 }>;
 
 export type TextMatchTransformer = $ReadOnly<{
-  dependencies: Array<Class<LexicalNode>>;
+  dependencies: Array<Class<LexicalNode>>,
   export: (
     node: LexicalNode,
     exportChildren: (node: ElementNode) => string,
--- a/packages/lexical-markdown/flow/LexicalMarkdown.js.flow
+++ b/packages/lexical-markdown/flow/LexicalMarkdown.js.flow
@@ -21,6 +21,7 @@ export type Transformer =
   | TextMatchTransformer;
 
 export type ElementTransformer = {
+  dependencies?: Array<Class<LexicalNode>>;
   export: (
     node: LexicalNode,
     traverseChildren: (node: ElementNode) => string,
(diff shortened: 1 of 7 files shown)


thegreatercurve (approved):

fantactuka:
Should it instead run dependencies check on mount and throw as other plugins do? You can pass explicit list of markdown transformers if want to skip some for unregistered nodes

fantactuka (approved):

patdx:
I've been experimenting with lexical and after upgrading to 0.4.0 I got this new error message and found my way here.

I checked the main document page for the [markdown plugin](https://lexical.dev/docs/packages/lexical-markdown) and my setup seemed to match the guide so it was not immediately clear how to resolve the issue. However, I found that if I import and set the nodes on the `initialConfig`, like `nodes: [HeadingNode, QuoteNode, CodeNode, ListNode, ListItemNode, LinkNode],`, then it started working again. Does this fix seem correct?

```ts
import { CodeNode } from '@lexical/code';
import { LinkNode } from '@lexical/link';
import { ListItemNode, ListNode } from '@lexical/list';
import { TRANSFORMERS } from '@lexical/markdown';
import { LexicalComposer } from '@lexical/react/LexicalComposer';
import { ContentEditable } from '@lexical/react/LexicalContentEditable';
import { LinkPlugin } from '@lexical/react/LexicalLinkPlugin';
import { ListPlugin } from '@lexical/react/LexicalListPlugin';
import { MarkdownShortcutPlugin } from '@lexical/react/LexicalMarkdownShortcutPlugin';
import { RichTextPlugin } from '@lexical/react/LexicalRichTextPlugin';
import { HeadingNode, QuoteNode } from '@lexical/rich-text';
import ToolbarPlugin from './plugins/toolbar-plugin';

<LexicalComposer
  initialConfig={{
    namespace: 'MyEditor',
    theme: {},
    onError,
    nodes: [HeadingNode, QuoteNode, CodeNode, ListNode, ListItemNode, LinkNode],
  }}
>
  <ToolbarPlugin />
  <div
    data-comment="lexical-editor-shell"
    className="relative flex flex-1 flex-col border"
  >
    <RichTextPlugin
      contentEditable={
        <ContentEditable className="prose block max-w-none flex-1 cursor-text resize-y overflow-auto p-2 text-sm outline-none" />
      }
      placeholder={
        <div className="pointer-events-none absolute inset-2 touch-none">
          Enter some text...
        </div>
      }
    />
  </div>
  <LinkPlugin />
  <ListPlugin />
  <MarkdownShortcutPlugin transformers={TRANSFORMERS} />
</LexicalComposer>;
```

tmb:
We're also experiencing this error. We solved it by pulling the required nodes out of `TRANSFORMERS` before instantiating the editor: that said, it would be nice if the Markdown plugin added the correct nodes to the editor before instantiating itself: 

```typescript
  const transformerNodes = TRANSFORMERS.map((t) => {
    if (t.type === "element") {
      return t.dependencies;
    } else if (t.type === "text-match") {
      return t.dependencies;
    }

    return undefined;
  })
    .filter((t): t is Klass<LexicalNode>[] => !!t)
    .flat();
```

PR #2911, opened 2022-08-30
feat: allow escaped characters within TextFormatTransformer

digitalgopnik:
Will extend that escaped characters won't be respected during TextFormatTransformers.

Before:
before
The text was formatted to italic, even if the asterix was escaped.

After:
after

If that pull-request is done, i could add another one with a new transformer for replacing the backslash from escaped characters, when moving from markdown-mode to wysiwyg-mode.

### DIFF
--- a/packages/lexical-react/flow/LexicalComposer.js.flow
+++ b/packages/lexical-react/flow/LexicalComposer.js.flow
@@ -14,7 +14,7 @@ import type {InitialEditorStateType as InitialEditorStateRichTextType} from '@le
 type Props = {
   initialConfig: $ReadOnly<{
     editor__DEPRECATED?: LexicalEditor | null,
-    readOnly?: boolean,
+    editable?: boolean,
     namespace: string,
     nodes?: $ReadOnlyArray<Class<LexicalNode>>,
     theme?: EditorThemeClasses,


fantactuka:
Thanks for looking into it! One thing to check is whether it'll be failing in Safari with negative/positive lookbehind in RegExp https://caniuse.com/?search=Negative%20Lookbehind

digitalgopnik:
@fantactuka Thanks for your response! Definitely a thing to check. But currently i don't really know, if it's the right way to change the "main"-regexp for textFormats. Maybe it's better to let it match, if there is one in the textContent and to re-check it afterwards, if it contains escaped-chars? Or better follow the current approach and to extend the regexp? What do you think?

I'm not quite sure, if its possible to write only one regexp, e.g. to match something that does contain asterisk and does not contain escaped asterisk :laughing: :exploding_head:

digitalgopnik:
Got it now (nearly done). Unfortunately safari needs an extra sausage (as we say in germany). Not sure if anyone would use safari on desktop **and** will write escaped markdown there. Neither on iOS. If the User-Agent will recognize Safari or iOS, the old regexp will be used (without lookbehind and lookahead). With a bit of luck we will get [this](https://bugs.webkit.org/show_bug.cgi?id=174931) resolved this decade / century? 😃 I mean at least lexical can support firefox- and chrome-users, if they write escaped markdown. What do you think?

acywatson (approved):
I think this is fine, but will let @fantactuka merge it.

PR #2935, opened 2022-09-02
Prevent nested editor event duplication

trueadm:
When we have nested Lexical editors, we run into issues where events are duplicated on the parent editor due to how event propagation works. Rather than block the native event propagation – which can cause bugs around UIs that use other event systems (React's for example) we locally stop propagation for Lexical aware things.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -748,9 +748,12 @@ function onCompositionEnd(
 }
 
 function onKeyDown(event: KeyboardEvent, editor: LexicalEditor): void {
+  if (hasStoppedLexicalPropagation(event)) {
+    return;
+  }
   lastKeyDownTimeStamp = event.timeStamp;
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -921,6 +923,20 @@ function onDocumentSelectionChange(event: Event): void {
...
+  // We attach a special property to ensure the same event doesn't re-fire
+  // for parent editors.
+  // @ts-ignore
+  event._lexicalHandled = true;
+}
+
+function hasStoppedLexicalPropagation(event: Event): boolean {
+  // @ts-ignore
+  const stopped = event._lexicalHandled === true;
+  stopLexicalPropagation(event);
```
fantactuka, line 936: I guess we won’t use it often, but someone gonna miss the fact it mutated event. Could it be called explicitly instead?

fantactuka (approved):

Issue #2758, opened 2022-08-04
Bug(playground): Escape key doesn't close the link editor


ahhshm:
Pressing the escape key doesn't close the link editor in the playground. The logic actually exists ([link](https://github.com/facebook/lexical/blob/5a1c20da44eba7968523c40f076ba6ba43fed1bd/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx#L324)) but it doesn't work, at least in Chromium 104.0.5112.79. It's probably a bug or even the intentional behavior of chromium because I don't have the same problem in Firefox.

Shannon26:
Hello @ahhshm, I tried this on Microsoft Edge which runs on Chromium. But this seems to work for me. Pressing Esc keep close the link editor.

nikmish1:
<img width="1164" alt="Screenshot 2022-09-12 at 2 49 15 PM" src="https://user-images.githubusercontent.com/30014840/189618177-4215ecc0-1b80-4772-ab32-e80076ed8da1.png">
The link editor doesn’t close on the press of the ESC button I tried in chrome and firefox.

ahhshm:
> May I pick this issue?

Sure. Thank you.

PR #2941, opened 2022-09-04
fix: playground bug(floating link editor not closing with 'Escape' key

heesungjang:
fixed playground bug - closing the floating link editor with the "Escape" key did not work (https://github.com/facebook/lexical/issues/2758).

- previously, it did not work because the logic for the 'escape' keyDown only existed and applied to link-input (input tag). [#link](https://github.com/facebook/lexical/blob/5a1c20da44eba7968523c40f076ba6ba43fed1bd/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx#L324)

- however, when isEditMode is false, it actually conditionally renders a different tag which is a div element.

It was not ideal to add tabindex={0} to this div element and onKeyDown. This was because the div element then had to get focused which will prevent the user from range selecting the link text or moving the cursor in between link texts.


heesungjang:
> Thanks for working on it! Downside of the fix is that after pressing Escape editor loses focus (it's intentional to allow getting out of editor with keyboard only, [see here](https://github.com/facebook/lexical/blob/main/packages/lexical-rich-text/src/index.ts#L896-L908)). With that said, I think fix should also utilize KEY_ESCAPE_COMMAND listener, that if link modal is opened will close it (and return false in command handler to prevent default behaviour from the code pointer above)

thanks @fantactuka ! I will look into that approach.

heesungjang:
Hi, @fantactuka. I have utilized KEY_ESCAPE_COMMAND to close the floating link editor. 

It prevents the default editor behavior(escape editor with escape key) if floating link editor is opened.

heesungjang:
HI @fantactuka this is now ready for the review.

thegreatercurve (approved):
Sorry for the late review! This look goos to me though!

thegreatercurve:
Closes #2758

PR #2942, opened 2022-09-04
Add type to commands for logging 

thegreatercurve:
In the future, when we add logging for commands to the developer extension, we need a way of getting access to the command types as plain strings. At the moment we can't do that as the commands are originally instantiated as empty objects.

This diffs adds in an optional type object for the command object, which we can use in the logger.

image

### DIFF
--- a/examples/decorators.md
+++ b/examples/decorators.md
@@ -62,7 +62,7 @@ As any other custom Lexical node, decorator nodes need to be registered _before_
 // Create a custom command with a typed payload.
 type CommandPayload = string;
 export const INSERT_VIDEO_COMMAND: LexicalCommand<CommandPayload> =
-  createCommand();
+  createCommand('INSERT_VIDEO_COMMAND');
 
 function VideoPlugin(): ReactNode {
   const [editor] = useLexicalComposerContext();
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -341,7 +341,7 @@ export const TOGGLE_LINK_COMMAND: LexicalCommand<
       rel?: string;
     }
   | null
-> = createCommand();
+> = createCommand('TOGGLE_LINK_COMMAND');
 
 export function toggleLink(
   url: null | string,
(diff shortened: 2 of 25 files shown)


trueadm:
What’s the effect on bundle sizes?

acywatson:
> What’s the effect on bundle sizes?

If it's significant can we transpile them to IDs like we do for errors?

thegreatercurve:
> What’s the effect on bundle sizes?

@trueadm Nothing to prod at the moment as we're using it in the `__DEV__` env, but it's a small increase if we include it in prod for error logging purposes. Current numbers: 

Before: 
```
271558  4 Sep 20:29 Lexical.dev.js
27964  4 Sep 20:30 Lexical.js.flow
77259  4 Sep 20:29 Lexical.prod.js
```
After:
```
272419  4 Sep 20:27 Lexical.dev.js
27990  4 Sep 20:28 Lexical.js.flow
77259  4 Sep 20:27 Lexical.prod.js
```

> If it's significant can we transpile them to IDs like we do for errors?

@acywatson I don't think we can as this needs to be usable for developers who create their own custom commands as well.

acywatson:
> > What’s the effect on bundle sizes?
> 
> @trueadm Nothing to prod at the moment as we're using it in the `__DEV__` env, but it's a small increase if we include it in prod for error logging purposes. Current numbers:
> 
> Before:
> 
> ```
> 271558  4 Sep 20:29 Lexical.dev.js
> 27964  4 Sep 20:30 Lexical.js.flow
> 77259  4 Sep 20:29 Lexical.prod.js
> ```
> 
> After:
> 
> ```
> 272419  4 Sep 20:27 Lexical.dev.js
> 27990  4 Sep 20:28 Lexical.js.flow
> 77259  4 Sep 20:27 Lexical.prod.js
> ```
> 
>  > If it's significant can we transpile them to IDs like we do for errors?
> 
> @acywatson I don't think we can as this needs to be usable for developers who create their own custom commands as well.

Yea, but I’d expect the vast majority of commands to be in core, not added by users. We can use the id if it exists or just fallback to the string name maybe.

Juice might not be worth the squeeze - just a thought.

PR #2943, opened 2022-09-04
Fix Safari selection bug

### LINKED ISSUE #2797: Bug: Space between paragraph nodes selected in Safari
In the Lexical playground in Safari 15.3, the space between paragraph nodes can be selected. The Tree View Plugin indicates that both paragraphs on either side of the highlighted space is selected but text formatting can't be applied.

Lexical version: 0.3.8

## Steps To Reproduce

1. Navigate to the lexical playground in Safari
2. Place the cursor at the beginning of a paragraph node, immediately before the first character.
3. Click and drag the selection to the end of the paragraph node above. 
4. The space between the paragraph nodes should be selected
image

Link to code example: playground.lexical.dev

## The current behavior
Space between paragraph nodes can be selected in Safari

## The expected behavior
Content outside of nodes should not be selected.

thegreatercurve:
When a `position` property is declared on a DOM element, Safari will show the range selection highlighting between 2 elements when the anchor node is at the end of an element and the focus node at the start of the next sibling.

Removing the styling should fix this issue without causing any regressions.

Before
image

After
image

Closes #2797

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -16,7 +16,6 @@
 .PlaygroundEditorTheme__paragraph {
   margin: 0;
   margin-bottom: 8px;
-  position: relative;
 }
 .PlaygroundEditorTheme__paragraph:last-child {
   margin-bottom: 0;


thegreatercurve:
It may cause some UX confusion, so logged a WebKit bug report as well: https://bugs.webkit.org/show_bug.cgi?id=244780

trueadm (approved):
Nice find!

PR #2945, opened 2022-09-05
Fix editable property being forced to true

### LINKED ISSUE #2944: Bug: Editable: false can't be set via initialConfig (LexicalComposer)
The change from readOnly to editable does not work properly with initial config, because the line that sets the editable mode was just logically swapped with the new editable setup:

LexicalComposer.src
`editor.setEditable(isEditable || true);` (line 92)
Said line will, when isEditable is undefined or false, always resolve to true and thus the composer is never going to end up with a read only / not editable mode.

Lexical version: 0.4

## Steps To Reproduce

1. Set an initial config of a LexicalComposer component to `{ editable: false }`

## The current behavior
Read only / editable mode can't be set for LexicalComposer via initialConfig

## The expected behavior
`editor.setEditable( isEditable !== undefined ? isEditable : true);`

TimFL:
Fixes #2944

### DIFF
--- a/packages/lexical-react/src/LexicalComposer.tsx
+++ b/packages/lexical-react/src/LexicalComposer.tsx
@@ -89,7 +89,7 @@ export function LexicalComposer({initialConfig, children}: Props): JSX.Element {
   useLayoutEffect(() => {
     const isEditable = initialConfig.editable;
     const [editor] = composerContext;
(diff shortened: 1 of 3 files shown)


trueadm (approved):

thegreatercurve:
Thanks for spotting and putting together the fix @TimFL.

PR #2948, opened 2022-09-05
[0.5] Treat undefined selection the same as null in $generateNodesFromHtml

acywatson:
This is a breaking change, so we'll need to go to 0.5.

Previously, we treated undefined differently than null here for backwards compat. undefined would end up just returning an empty string, which is super confusing. Now, leaving off the selection argument completely will return the entire contents of the editor converted to HTML.

### DIFF
--- a/packages/lexical/src/nodes/LexicalElementNode.ts
+++ b/packages/lexical/src/nodes/LexicalElementNode.ts
@@ -493,12 +493,7 @@ export class ElementNode extends LexicalNode {
   }
   extractWithChild(
     child: LexicalNode,
-    selection:
-      | RangeSelection
-      | NodeSelection
-      | GridSelection
-      | null
-      | undefined,
+    selection: RangeSelection | NodeSelection | GridSelection | null,
     destination: 'clone' | 'html',
   ): boolean {
     return false;
--- a/packages/lexical/src/nodes/LexicalElementNode.ts
+++ b/packages/lexical/src/nodes/LexicalElementNode.ts
@@ -493,7 +493,12 @@ export class ElementNode extends LexicalNode {
   }
   extractWithChild(
     child: LexicalNode,
-    selection: RangeSelection | NodeSelection | GridSelection | null,
+    selection:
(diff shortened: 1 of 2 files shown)


packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -485,7 +485,12 @@ export class ElementNode extends LexicalNode {
   }
   extractWithChild(
     child: LexicalNode,
-    selection: RangeSelection | NodeSelection | GridSelection | null,
+    selection:
+      | RangeSelection
+      | NodeSelection
+      | GridSelection
+      | null
+      | undefined,
```
trueadm, line 493: This change is confusing as it's user-land facing. We never promote the idea of selection ever being undefined, maybe we can revise elsewhere to prevent this leaking out?

acywatson: Yea good call

zurfyx (commented):
No objections, but what about `selection?: RangeSelection | NodeSelection | GridSelection | null = null`?

zurfyx (approved):

acywatson:
> No objections, but what about `selection?: RangeSelection | NodeSelection | GridSelection | null = null`? It's odd to see a function handling undefined in Lexical

We could - in Typescript an optional param is the same as the union of that params type with "undefined", so what you're suggesting there is practically the same. Also optional params can't be followed by required params, so we'd have to reorder the function params. Maybe that's fine since we're making a breaking change anyway?

acywatson:
> > No objections, but what about `selection?: RangeSelection | NodeSelection | GridSelection | null = null`? It's odd to see a function handling undefined in Lexical
> 
> We could - in Typescript an optional param is the same as the union of that params type with "undefined", so what you're suggesting there is practically the same. Also optional params can't be followed by required params, so we'd have to reorder the function params. Maybe that's fine since we're making a breaking change anyway?

nvm, I forgot this is an internal function, so re-ordering doesn't matter too much.

PR #2953, opened 2022-09-07
Apply history differently

abelsj60:
**Issue**

- It does not currently seem to be possible to restore a `historyStack` when trying to load a saved `editorState.` 
- The reason is that the [undo](https://github.com/facebook/lexical/blob/main/packages/lexical-history/src/index.ts#L365) and [redo](https://github.com/facebook/lexical/blob/main/packages/lexical-history/src/index.ts#L336) functions in the history package try to apply each `editorState` in the `historyStack` to the instance that produced it. This "historical" instance was saved to the `historyStack`, along with the `editorState` we care about, via the [`registerHistory`](https://github.com/facebook/lexical/blob/main/packages/lexical-history/src/index.ts#L380) function. 
- [More info from Discord](https://discord.com/channels/953974421008293909/953974421486436393/1016796173027250176).

**Reason**

- I ran into this problem while attempting to pass Lexical editors around my UI (outside LexicalContext). 
- I achieve this by: 

  a.  Saving the `editorState` and `historyStack` of the current editor, 
  b. Unmounting the current Lexical editor, 
  c. Mounting a new Lexical editor, and 
  d. Passing the saved `editorState` and `historyStack` into it as initial config. 

- While the `editorState` worked as expected, I was not able to undo/redo through the preexisting `historyStack` entries. 

**Analysis**

- The problem here is that the new Lexical editor does not match the editor instances that are found in each of the `historyStack`'s preexisting entries. 
  - You can see the mismatch by comparing the new instance's `.__key` to the `.__key` property on the instances found in the preexisting `historyStack.`
  - This seems to mean that Lexical is running the `undo` and `redo` functions on a now non-existent editor.

**Conclusion**

- The preexisting `historyStack` will work with the new editor instance by modifying the `undo` and `redo` functions to apply the preexisting `editorState` to the current editor instance. It's passed into each function as an argument. 
  - In fact, this "active" editor instance is currently used to dispatch the "canUndo" and "canRedo" commands.

- This pull request makes this change. The test suite seems to pass. The e2e tests seem to pass to the same degree as they currently do on the main branch.
  - I also: 
    a. Added an explanatory comment to each function, and 
    b. Changed a `.length` test for more consistency between the two functions
  - I signed the contributor agreement yesterday.

- I'm currently addressing this via patch-package, but would love to see the change migrate upstream. 

- Please let me know if I've misunderstood something or how I can further help. 

Thanks so much for the time. -j

### DIFF
--- a/packages/lexical-history/src/index.ts
+++ b/packages/lexical-history/src/index.ts
@@ -313,6 +313,9 @@ function createMergeActionGetter(
 }
 
 function redo(editor: LexicalEditor, historyState: HistoryState): void {
+  // historyState is a record of the history that should be restored, while the
+  // incoming editor instance is the object to which it should be applied
+
   const redoStack = historyState.redoStack;
   const undoStack = historyState.undoStack;
 
@@ -333,7 +336,7 @@ function redo(editor: LexicalEditor, historyState: HistoryState): void {
     historyState.current = historyStateEntry || null;
 
     if (historyStateEntry) {
-      historyStateEntry.editor.setEditorState(historyStateEntry.editorState, {
+      editor.setEditorState(historyStateEntry.editorState, {
         tag: 'historic',
       });
     }
@@ -341,11 +344,13 @@ function redo(editor: LexicalEditor, historyState: HistoryState): void {
 }
 
 function undo(editor: LexicalEditor, historyState: HistoryState): void {
+  // historyState is a record of the history that should be restored, while the
+  // incoming editor instance is the object to which it should be applied
+
   const redoStack = historyState.redoStack;
   const undoStack = historyState.undoStack;
-  const undoStackLength = undoStack.length;
 
-  if (undoStackLength !== 0) {
+  if (undoStack.length !== 0) {
     const current = historyState.current;
     const historyStateEntry = undoStack.pop();
 
@@ -361,7 +366,7 @@ function undo(editor: LexicalEditor, historyState: HistoryState): void {
     historyState.current = historyStateEntry || null;
 
     if (historyStateEntry) {
-      historyStateEntry.editor.setEditorState(
+      editor.setEditorState(
         historyStateEntry.editorState.clone(historyStateEntry.undoSelection),
         {
           tag: 'historic',


acywatson:
@fantactuka did you have thoughts you wanted to add here?

fantactuka:
Main concern from discord discussion:
Editor in history stack was added for nested editors (e.g., image's caption or sticky notes in the playground). History stack does not store whole editor state when nested one is updated, instead it only persists that nested editor data, and during undo/redo needs correct editor instance to apply that state

abelsj60:
I hear you. This sounds like an absolute bar to merging this pull request as is — correct?

My concern is this:

There seem to be at least two use cases for the historyStack: 

(a) Restoring a set of editor instances that haven't unmounted, and 
(b) Restoring a set of editor instances that have mounted.

Right now, (a) has been solved, but (b) hasn't. 

I tend to think restoring a historyStack is as reasonable as handling nested editors, but you may see this as an edge case?

Nested editors do seem complicated — I absolutely haven't understood them until @fantactuka's example above. 

To work within the confines of the history package, I guess you'd have to apply the old editorStates to the newly generated instances? That'd mean crawling down the editor tree and matching the old state up with the new instance that should represent it. Maybe you could use some sort of editor-instance dictionary on the historyStack when Lexical mounts? 

I'm not sure. 

Do you, @acywatson, or anyone have any thoughts. Orphaning history doesn't seem right to me. I'd try to apply an instance stored on context to a newly mounted editor, but I think the initialConfig says passing a full editor instance is deprecated?

I'm going to have to do _something_ on my end — stick with the patch or try something else. 

If you've thoughts in terms of where I might take this pull request, I can give it a whirl. 

Thanks again for the continued interest.

-j

acywatson:
> I hear you. This sounds like an absolute bar to merging this pull request as is — correct?
> 
> My concern is this:
> 
> There seem to be at least two use cases for the historyStack:
> 
> (a) Restoring a set of editor instances that are mounted, and (b) Restoring a set of editor instances that are unmounted.
> 
> Right now, (a) has been solved, but (b) hasn't. (I think?)
> 
> I tend to think restoring an old historyStack is important, but you may see this as an edge case?
> 
> Nested editors do seem complicated — I absolutely haven't understood them until @fantactuka's example above.
> 
> To work with the history package, I guess I'd have to apply the old editorStates to the newly generated instances? That'd mean crawling down the editor tree and matching the old state to the new instance that represents it? Maybe I could use some sort of editor-instance dictionary on the historyStack when Lexical mounts?
> 
> I'm not sure.
> 
> Do you, @acywatson, or anyone have any thoughts?
> 
> I'm going to have to do _something_ on my end — stick with the patch or try something else. (I'd try to apply an instance stored on state to a newly mounted editor, but I think the initialConfig says passing a full editor instance is deprecated?)
> 
> If you've thoughts in terms of where I might take this pull request, I can give it a whirl.
> 
> Thanks again for the continued interest.
> 
> -j

I wonder if Editors could have serializable stable identifiers that could then be used to associate history state entries to the correct editor on undo/redo.

I guess the fundamental use case here is just initializing editor with a saved state that also has a history. If I'm understanding correctly, this would fail in any case where you try to pass an initialHistory into a different editor than it was generated in. I guess it's simpler without nested editors, because you could just go go through the historyStack and change all the editor references to point to the new one.

abelsj60:
> 

Hey @acywatson. Sorry for the delay, I got busy with a separate set of Remix-related issues. 

Some background explanation: My Lexical editor instances can be dragged and dropped around the page. Each drop places the Lexical editor into a sibling div. Thus it unmounts and remounts. This is a good interface and a fair structure. But, preserving the Lexical history is the fly in the ointment. 

Now, your comments: 

I think your suggestion sounds accurate. You have a more sophisticated understanding of this than I do, but some kind of stable and serializable ID on an editor instance sounds like a good way to sync old history with the new editor instances. 

But, doesn't the editor instance have a `__key` property?  

Why do you think this key isn't stable and serializable over time? I'm probably missing something?

I can see my current patch is a problem. This PR will have to be closed, unless I can contribute something better. 

In theory, I'd like to be able to do this: 

- Build an object — a map of history — by looping through the parsed json of the "saved" instances. The "old" key from each instance will become a key on this object map. 
- Each of these keys will then receive the new "new" editor's key as a value, as soon as it can be identified, at least.
- (The problem, of course, is that I'm not entirely sure how to link the two, i.e., match the new key to the old key. I haven't tried to loop through old editor state on mount, though, so maybe something will be obvious when I do...?)

Anyway, this might let me do what you suggest, update the history to work with the new instances? 

Am I right about that, or does this sequence sound insufficient to you, generally speaking?

The main alternative in my mind is React.Context. 

There's an existing [issue on Github with Genei that discusses](https://github.com/facebook/lexical/discussions/2481) how to use React.Context to manage many editor instances. If the editor instances stay in state, then maybe the history stack will just keep working because the editor instances won't be lost when the component that "shows" the Lexical editor unmounts? 

Any thoughts are appreciated... I'll mess with this, so will at least close the loop when I figure something out should others pass this way in the future.   -j

PR #2954, opened 2022-09-07
Fix link breaking when formatting on

pgmccullough:
As can be seen at https://codesandbox.io/s/lexical-rich-text-example-5tncvy, if any sort of formatting is active (bold, italic, etc...) while typing a link, the formatting will be disabled on the text of the detected url, and on subsequent keystrokes, the url will break.

Screen Shot 2022-09-07 at 4 11 04 PM

By applying the parent node's format to the autolink, this problem is resolved.


fantactuka (approved):
Thank you!

PR #2963, opened 2022-09-08
docs: selection cases

echarles:
Add to the documentation examples for 2 cases related to selection.

### DIFF
--- a/packages/lexical-website/docs/concepts/selection.md
+++ b/packages/lexical-website/docs/concepts/selection.md
@@ -60,4 +60,31 @@ has moved to another editor on the page. This can also happen when trying to sel
 
 ## Working with selection
 
-> TODO
+This section lists a few use cases around selection usage.
+
+### Get the current node
+
+If nothing is selected, how can I get the node where the cursor is at?
+
+```js
+const selection = $getSelection();
+if (selection.isCollapsed()) { // Nothing is selected.
+  // Get the anchor of the selection (works also with focus).
+  const anchorNode = selection.anchor.getNode();
+  console.log('This is the node where the cursor is at', anchorNode);
+}
+```
+
+### Set a node to be selected
+
+How can I explicitly set the selection to a node, so the user sees the node selected in its editor?
+
+```js
+// Let's assume you have a nodeKey.
+const node = $getNodeByKey(nodeKey);
+const nodeSelection = $createNodeSelection();
+nodeSelection.add(node.__key);
+$setSelection(nodeSelection);
+```
+
+> TODO More case to cover.


packages/lexical-website/docs/concepts/selection.md
```diff
@@ -60,4 +60,31 @@ has moved to another editor on the page. This can also happen when trying to sel
...
+```
+
+### Set a node to be selected
+
+How can I explicitly set the selection to a node, so the user sees the node selected in its editor?
+
+```js
+// Let's assume you have a nodeKey.
+const node = $getNodeByKey(nodeKey);
+const nodeSelection = $createNodeSelection();
```
acywatson, line 85: NodeSelection should only apply to DecoratorNodes - might be best to call that out here. Otherwise, you'd want to use RangeSelection probably.

echarles: I have been using the $createNodeSelection to select non decorator nodes. The NodeSelection has also a method `insertText` which makes me think that is not only for Decorator nodes. I can document the way you propose if you confirm.

acywatson: Yea, the intent behind NodeSelection is selecting one or more decorator nodes:

https://github.com/facebook/lexical/pull/1332

It will probably work in a lot of cases if you add non-decorators to it, but we are doing a lot of checks in various places to see if something is a RangeSelection, which is what we typically expect when a non-decorator is selected.

Also I don't think insertText is actually implemented in that class.

TextNode and ElementNode both also expose a "select" method that sets the current RangeSelection to that particular node (with optional offset args).

So, I would say the canonical way to do this for non-decorators is using that API, but for decorators you want to use NodeSelection.

thegreatercurve:
Closing this due to the PR becoming stale.

packages/lexical-website/docs/concepts/selection.md
echarles: @acywatson sorry I have lost track of this. I am trying your suggestion with the following snippet. I can see in the TreeView the selection being updated to the first node (with anchor 0 and focus 3), but the editor does not show the text being selected, so the use does not really see that the text is selected.

```jsx
        editor.update(() => {
          const root = $getRoot();
          const node = root.getChildren()[0];
          const selected = (node as ElementNode).select(0, 3);
          $setSelection(selected);
        });
```

This is what I would expect as end-result, the blue selection being show, but it does not.

Screenshot 2022-11-08 at 09 08 59

echarles:
@thegreatercurve I have replied on https://github.com/facebook/lexical/pull/2963#discussion_r1016264219. Could you reopen this? Thx.

packages/lexical-website/docs/concepts/selection.md
thegreatercurve: You have to call `const selected = node.getFirstChild().select(0, 3);` to see what you expect, otherwise it tries to apply the selection to the parent element of the text node.

thegreatercurve:
Closing this as we need to make the use cases for different selection types a lot clearer in the docs.

packages/lexical-website/docs/concepts/selection.md
echarles: That worked. Thx for the help.

echarles:
> Closing this as we need to make the use cases for different selection types a lot clearer in the docs.

That was indeed the case of this PR, but I am out of time and don't have enough knowledge to write it down easily. I hope someone else will pick that.

PR #2968, opened 2022-09-08
Reorder Keys

acywatson:
I know this seems very silly (and it is, actually), but it's a workaround for an issue with our internal javascript parser.


acywatson:
Closed in favor of doing it in the build: https://github.com/facebook/lexical/pull/2970

PR #2969, opened 2022-09-08
Clean up redundant newlines during pasting

fantactuka:
When parsing html we should ignore newlines except if it's within `<pre>` or code-like tags where it needs to be preserved. Although we can check for explicit`<pre>` tag, we can't easily check for code-like cases (e.g. github gists use tables for it) so this PR introduces "preformatted" flag for DOMTransformerFn, which will indicate that nested content should preserve newlines, otherwise it'll be removed.

Before:

https://user-images.githubusercontent.com/132642/189204784-3fa241d7-f07f-4543-b997-9bf6d0b312a6.mov

After:

https://user-images.githubusercontent.com/132642/189204778-6db2ba17-7727-466b-ba77-6253ffeef2aa.mov

### DIFF
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -290,12 +290,13 @@ export function $isCodeNode(
 }
 
 function convertPreElement(domNode: Node): DOMConversionOutput {
-  return {node: $createCodeNode()};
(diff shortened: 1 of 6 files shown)


acywatson (approved):
Works for me - thanks!

PR #2981, opened 2022-09-09
Fixed exportDOM for paragraph node

zifahm:
Now paragraph node exports DOM related text-align, dir and text-indent.


tylerjbainbridge (approved):
Makes sense to me.

acywatson:
Thanks! A test would be nice for this, too.

zifahm:
Sure adding test soon

PR #2988, opened 2022-09-10
Set Playground Sticky initialNodes

jagregory:
**Work in progress**

Updating any Playground plugins that use nested editors to use the new `initialNodes` prop to restrict which nodes can be inserted into the nested editors.


dalakatt:
[https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/#tutorial](url)

thegreatercurve:
I'm not sure if this is necessary. I'm pretty sure we do want to allow custom nodes within the Sticky Note example.

PR #2989, opened 2022-09-10
docs: loadContent clarification

ly3xqhl8g9:
The documentation on saving/loading editor state is unclear what kind of data is expected from the `initialEditorState` which leads to confusion as in #1937. Mocked-up the `loadContent` function to return an `'empty'` initial editor state to make it clear.


zurfyx (approved):
Thank you!

PR #2999, opened 2022-09-12
Capture pendingDecorators after garbage collecting detached decorators

### LINKED ISSUE #2993: Bug: `DecoratorNode`'s component does not call `useEffect` cleanup until after remount
`DecoratorNode.decorate` can return JSX element to render React components. Surprisingly, when you delete a `DecoratorNode`, `useEffect` cleanup functions are **not** called. Even more surprisingly, when you undo the delete, the `useEffect` cleanup function is called.

Lexical version: 0.4.1

## Steps To Reproduce

1. Clone https://github.com/adri1wald/lexical-decorator-node-useeffect-cleanup-repro
2. Follow steps outlined in README

## The current behavior

Cleanup doesn't happen when DecoratorNode is delete. It happens if and when the DecoratorNode is remounted.

## The expected behavior

Cleanup should happen at the expected point in the component lifecycle, i.e. when it unmounts / when it looks like it unmounts.

adri1wald:
Fixes #2993 

Have made changes as suggested by @trueadm. Cursory testing within our app suggests this change fixes the useEffect cleanup bug I noted in #2993, although I cannot speak to any other side effects it might have.

### DIFF
--- a/packages/lexical/src/LexicalUpdates.ts
+++ b/packages/lexical/src/LexicalUpdates.ts
@@ -508,7 +508,6 @@ export function commitPendingUpdates(editor: LexicalEditor): void {
   const dirtyElements = editor._dirtyElements;
   const normalizedNodes = editor._normalizedNodes;
   const tags = editor._updateTags;
-  const pendingDecorators = editor._pendingDecorators;
   const deferred = editor._deferred;
 
   if (needsUpdate) {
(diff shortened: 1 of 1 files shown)


trueadm (approved):

trueadm:
Awesome. Thanks for taking the time to do this!

adri1wald:
Signed bout 30 min ago! Think it's just taking its time to go through.

fantactuka (approved):
Thanks!

PR #1802, opened 2022-04-17
Add preserveEmptyParent param to LexicalNode.remove

acywatson:
Will add a test, just wanted to write the explanation while it's fresh:

The root cause of this issue is that, in $wrapLeafNodesInElements, we call [$removeParentEmptyElements](https://github.com/facebook/lexical/blame/0e159ecb6bf4002777c91305312bd4182ff94391/packages/lexical-selection/src/index.js#L507) to try to remove all the nodes that are empty after their former leaves are appended to their new parents (the "elements" in "$wrapLeafNodesInElements"). This function tries to crawl up the tree, removing newly-empty nodes as it goes (uses getLatest to access the pending state).

The problem with this is that the implementation of LexicalNode.remove() actually [removes the parent under certain conditions](https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalNode.js#L98-L104). So, as we're crawling up a nested structure, we remove a node, and it's parent also gets removed at the same time, which effectively short-circuits $removeParentEmptyElements, since it correctly detects that the latest version of the current node it's on doesn't have a parent anymore.

I tried a few ways to fix this. The next best one was probably trying to crawl up the tree without removing any nodes until we got to the highest empty one, then just removing that one and thereby removing all it's children. The issue became actually detecting emptiness in that case, since we're not removing as we go anymore, it gets harder to tell if a node is truly empty without recursively checking all of the siblings.

At the end, I decided that having removeNode also remove the parent in some cases, while convenient, is also not necessarily obvious or semantically correct behavior. It is probably usually the right thing to do, but shouldn't we have some low-level API for *just* removing a node without the side-effect of removing it's parent as well?

As a compromise, I introduced an optional flag to condition this behavior on.

Fixes #1128

PR #3000, opened 2022-09-12
Fix (Known) Table Bugs

### LINKED ISSUE #1597: Can't edit formula when inside a table cell
I can type at the end but it's impossible for me to place the cursor in the middle using my mouse

https://user-images.githubusercontent.com/193447/161524494-769a3662-66e2-4250-a505-56b9d54dfc89.mov

### LINKED ISSUE #2010: Bug: Cannot move selection below table
Lexical version: lexical playground

## Steps To Reproduce

1. insert a table inside existing text
2. start selection above table and move it down using arrows

## The current behavior
Selection gets stuck at the table bottom, instead of moving to the next line

## The expected behavior
Selection moves to the next line below the table

image

### LINKED ISSUE #2752: Bug: Cannot place cursor outside table
When inserting a table if you delete everything after and before the table, the cursor remains inside the table and you can’t add anything else after or before. To exit the table, you need to move the cursor inside a cell and press the down arrow to move below the table — see video attached.

https://user-images.githubusercontent.com/44519061/182714077-2a9dd636-80bf-4f12-a86c-2ee11bfbd829.mov

## Steps To Reproduce

1. Insert a table
2. Delete everything after and before the table
3. Click outside the table

### LINKED ISSUE #2793: Bug: Table cell breaks when content is switched from one styling to another
Table cells break when changing style of text from one type to another (for example, when inserting into a cell a "Heading 1" and then switching to normal text - see video). This behaviour seems to not appear when changing one type of list to another.

https://user-images.githubusercontent.com/44519061/183514925-52aa2f88-f7b9-4be7-9a64-e9180b399c25.mp4

## Steps To Reproduce

1. Insert a "Heading 1" into a cell
2. Press on "Normal" styling

tylerjbainbridge:
Long term the solution is to improve on the React-only approach MVP here (#2929) but this gets the existing solution in much better shape.

This PR fixes the following known table bugs (seen list at bottom).

Most notably, now you can change the block type within a `TableCell` without it breaking.

https://user-images.githubusercontent.com/13852400/189770404-f6fe9a07-289b-4c4e-9530-e067341cb9e8.mov

- [X] If there is no content before the table, when pressing Cmd+A and then the delete key, the table is not deleted completely. One single cell remains and the options menu popover button is misplaced.
- [X] When pressing Cmd+Shift+Delete inside a cell, the table breaks and the cell is deleted.
- [X] After deleting the table, no content can be inserted into Composer without creating a new selection.
- [X] Table cells currently break when content is switched from one type of styling to another (closes #2793).
- [X] Content from cell cannot be easily selected/ can’t move cursor inside text from cell.
- [X] Cannot place cursor outside table (closes #2752)
- [X] Closes #1597
- [X] Closes #2010 
- [X] Closes #2793

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -925,11 +925,18 @@ export function applyTableHandlers(
             DEPRECATED_$isGridSelection(prevSelection)) &&
           tableSelection.gridSelection !== selection
         ) {
-          tableSelection.updateTableGridSelection(
-            DEPRECATED_$isGridSelection(selection) && tableNode.isSelected()
-              ? selection
-              : null,
-          );
+          if (
+            DEPRECATED_$isGridSelection(selection) &&
+            selection.gridKey === tableSelection.tableNodeKey
+          ) {
+            tableSelection.updateTableGridSelection(selection);
+          } else if (
+            !DEPRECATED_$isGridSelection(selection) &&
+            DEPRECATED_$isGridSelection(prevSelection) &&
+            prevSelection.gridKey === tableSelection.tableNodeKey
+          ) {
+            tableSelection.updateTableGridSelection(null);
+          }
           return false;
         }
 
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -10,6 +10,7 @@ import type {TableNode} from './LexicalTableNode';
 import type {Cell, Cells, Grid} from './LexicalTableSelection';
 import type {
   GridSelection,
+  LexicalCommand,
   LexicalEditor,
   LexicalNode,
   NodeSelection,
@@ -20,7 +21,9 @@ import type {
 import {TableCellNode} from '@lexical/table';
 import {$findMatchingParent} from '@lexical/utils';
 import {
+  $createParagraphNode,
   $createRangeSelection,
+  $createTextNode,
   $getNearestNodeFromDOMNode,
   $getPreviousSelection,
   $getSelection,
@@ -31,6 +34,8 @@ import {
   COMMAND_PRIORITY_CRITICAL,
   CONTROLLED_TEXT_INSERTION_COMMAND,
   DELETE_CHARACTER_COMMAND,
+  DELETE_LINE_COMMAND,
+  DELETE_WORD_COMMAND,
   DEPRECATED_$isGridSelection,
   FOCUS_COMMAND,
   FORMAT_TEXT_COMMAND,
@@ -39,6 +44,7 @@ import {
   KEY_ARROW_RIGHT_COMMAND,
   KEY_ARROW_UP_COMMAND,
   KEY_BACKSPACE_COMMAND,
+  KEY_DELETE_COMMAND,
   KEY_TAB_COMMAND,
   SELECTION_CHANGE_COMMAND,
 } from 'lexical';
@@ -85,6 +91,8 @@ export function applyTableHandlers(
       if (event.button !== 0) {
         return;
       }
+
+      isMouseDown = true;
 
       const cell = getCellFromTarget(event.target as Node);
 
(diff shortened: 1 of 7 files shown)


tylerjbainbridge:
Will look into the failing tests tomorrow.

zurfyx (approved):

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -106,6 +113,7 @@ export function applyTableHandlers(
 
   // This is adjusting the focus of the selection.
   tableElement.addEventListener('mousemove', (event: MouseEvent) => {
+    // console.log({isRangeSelectionHijacked});
```
fantactuka, line 116: 👆

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -597,83 +605,155 @@ export function applyTableHandlers(
...
+      const clearCell = () => {
+        const newParagraphNode = $createParagraphNode();
+        const textNode = $createTextNode();
+        newParagraphNode.append(textNode);
+        tableCellNode.append(newParagraphNode);
+        tableCellNode.getChildren().forEach((child) => {
+          if (child !== newParagraphNode) {
+            child.remove();
+          }
+        });
```
fantactuka, line 669: Does it need to append first, and the clear the rest? Otherwise it can be a bit easier: 
```
tableCellNode.clear().append($createParagraphNode())
```

(Empty text node might not be needed to?)

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -597,83 +605,155 @@ export function applyTableHandlers(
...
+      };
 
+      if (
+        command === DELETE_LINE_COMMAND &&
+        parentElementNode.getPreviousSiblings().length === 0
```
fantactuka, line 674: Do we need to get all siblings?
```
parentElementNode.getPreviousSibling() !== null
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -597,83 +605,155 @@ export function applyTableHandlers(
...
+
+      if (
+        command === DELETE_CHARACTER_COMMAND ||
+        command === DELETE_WORD_COMMAND
+      ) {
+        if (
+          selection.isCollapsed() &&
+          selection.anchor.offset === 0 &&
+          parentElementNode === nearestElementNode &&
+          nearestElementNode.getPreviousSiblings().length === 0
```
fantactuka, line 688: Similar here

packages/lexical/src/LexicalEvents.ts
```diff
@@ -761,6 +761,8 @@ function onKeyDown(event: KeyboardEvent, editor: LexicalEditor): void {
 
   const {keyCode, shiftKey, ctrlKey, metaKey, altKey} = event;
 
+  // debugger;
```
fantactuka, line 764: 🙈

packages/lexical/flow/Lexical.js.flow
```diff
@@ -722,6 +722,7 @@ declare export class ElementNode extends LexicalNode {
   canInsertTextBefore(): boolean;
   canInsertTextAfter(): boolean;
   isInline(): boolean;
+  isRequiredParent(): boolean;
```
fantactuka, line 725: To be honest, I'm not sure I fully understand this method semantics, per code it flags tables, but what is its usage in API (e.g., when other element node need to override it). On the other hand, if it's just a temp thing to live with while we figuring out future of tables, maybe it should be prefixed with `unstanble_` or similar?

acywatson: I was going to comment something very similar - not sure the intent is totally conveyed by the method name. I gather I would use this when I want to check if some operation should stop at an earlier ancestor than the root?

tylerjbainbridge: I'm actually getting rid of this part of the PR- Gerard was working on a fix for this same issue and his solution (#3001) is a bit more robust and uses a more semantically clear method name.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
tylerjbainbridge: If an ElementNode is cleared it'll also remove the empty parents as well, which destroys the table.

acywatson: > If an ElementNode is cleared it'll also remove the empty parents as well, which destroys the table. So we have to do it in this kind of roundabout way...

Oh I think I ran into this before and made a fix for it.

https://github.com/facebook/lexical/pull/1802

It was a little controversial at the time, so might not be the right place to talk about adding it to clear().

I still do and probably always will think it's wrong for LexicalNode.remove() to remove empty parents as well.

PR #3009, opened 2022-09-13
[0.5] Correct definition of isTopLevel; introduce DecoratorNode->isInline, ElementNode->isShadowRoot

zurfyx:
This PR fixes the definition of the `isTopLevel`

**What is `isTopLevel`?**

Initially isTopLevel meant the node immediately after the RootNode, either an ElementNode or a DecoratorNode. In 2022 we expanded this to become a value that we can define on DecoratorNodes, that will help towards clipboard pasting -> top level nodes don't get a paragraph included.

The second definition is misplaced, what we meant instead is whether they'd be block level nodes, which matches the current behavior and the most correct UX behavior (i.e. if you add a horizontal node instead a table cell it should show inside the table cell instead of somewhere else). At the same time, the inline property should guarantee that you can never place nodes inside paragraph.

Separately, on #3001 we reused the isTopLevel definition for yet another slightly different behavior. Instead, what we meant is `shadowRoot` and this can work hand in hand is `isTopLevel` and `isInline`.

**This PR**

This PR removes the bad definitions of `isTopLevel` and introduces an `isInline` property for DecoratorNode and `isShadowRoot` for ElementNode.

It also modifies the behavior of `getTopElementNode` to work well with `isShadowRoot`.

At the same time, there's subtle differences with node insertion like aforementioned. See video below.

Before

https://user-images.githubusercontent.com/193447/190021199-1fcf6aaa-fda1-405e-87e2-87e111edc440.mov

After

https://user-images.githubusercontent.com/193447/190021227-761df6b2-34df-4555-a91a-6c4fe07d100f.mov

---

Relevant PRs: #2741 and #3001

Thanks @fantactuka for helping brainstorm the behavior and names!


packages/lexical-playground/src/plugins/FigmaPlugin/index.tsx
```diff
@@ -27,7 +27,7 @@ export default function FigmaPlugin(): JSX.Element | null {
       INSERT_FIGMA_COMMAND,
       (payload) => {
         const figmaNode = $createFigmaNode(payload);
-        $insertBlockNode(figmaNode);
+        $insertNodeToTopLevel(figmaNode);
```
fantactuka, line 30: insertNodeToNearestRoot?

fantactuka (approved):

packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -480,7 +480,7 @@ export class ElementNode extends LexicalNode {
   isInline(): boolean {
     return false;
   }
-  isTopLevel(): boolean {
+  isShadowRoot(): boolean {
```
acywatson, line 483: Can we add typedoc comments to these methods, at least? I understand the concept of a shadowRoot and a topLevel and an inline node, but I really don't think it will be obvious to the community how this important concept applies to their integrations.

Ultimately, we might even need a separate subheading under "Concepts" to clarify these things.

PR #3011, opened 2022-09-14
feat(lexical-react): add initialEditorState for LexicalCollaborationPlugin

strdr4605:
This PR is a follow-up of [this discord thread](https://discord.com/channels/953974421008293909/1013899629894565979/1013900284247953498).

I tested this feature, by connecting the WebsocketProvider to my custom Websocket server which removes the YDoc from persistence when there were no connections (last user disconnected)

Here is the `initialEditorState` that I passed to `CollaborationPlugin`

```js
              <CollaborationPlugin
                id="main"
                providerFactory={createWebsocketProvider}
                shouldBootstrap={!skipCollaborationInit}
                initialEditorState={`{"root":{"children":[{"children":[{"detail":0,"format":0,"mode":"normal","style":"","text":"1","type":"text","version":1}],"direction":null,"format":"","indent":0,"type":"paragraph","version":1},{"children":[{"detail":0,"format":0,"mode":"normal","style":"","text":"22","type":"text","version":1}],"direction":null,"format":"","indent":0,"type":"paragraph","version":1},{"children":[{"detail":0,"format":0,"mode":"normal","style":"","text":"333","type":"text","version":1}],"direction":null,"format":"","indent":0,"type":"paragraph","version":1}],"direction":null,"format":"","indent":0,"type":"root","version":1}}`}
              />

```

lexical-init-state-collab

### DIFF
--- a/packages/lexical-react/src/LexicalCollaborationPlugin.ts
+++ b/packages/lexical-react/src/LexicalCollaborationPlugin.ts
@@ -13,6 +13,7 @@ import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {useEffect, useMemo} from 'react';
 import {WebsocketProvider} from 'y-websocket';
 
+import {InitialEditorStateType} from './LexicalComposer';
 import {
   CursorsContainerRef,
   useYjsCollaboration,
@@ -26,6 +27,7 @@ export function CollaborationPlugin({
   shouldBootstrap,
   username,
   cursorsContainerRef,
+  initialEditorState,
 }: {
   id: string;
   providerFactory: (
@@ -36,6 +38,7 @@ export function CollaborationPlugin({
   shouldBootstrap: boolean;
   username?: string;
   cursorsContainerRef?: CursorsContainerRef;
+  initialEditorState?: InitialEditorStateType;
 }): JSX.Element {
   const collabContext = useCollaborationContext(username);
 
@@ -69,6 +72,7 @@ export function CollaborationPlugin({
     color,
     shouldBootstrap,
     cursorsContainerRef,
+    initialEditorState,
   );
 
   collabContext.clientID = binding.clientID;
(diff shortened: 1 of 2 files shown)


fantactuka:
Thanks for working on it!

fantactuka (approved):

fantactuka:
@zurfyx Just realized that 0.5 has plans to move initialEditorState to composer component. Idea of this PR is to provide default editor state when collab server does not have any data, which means LexicalComposer will still have to have initialEditorState=null, and the CollabPlugin can optionally have initialEditorState (in case users want something more than empty paragraph). Any concerns?

acywatson:
@strdr4605 thanks for this - can you fix the linter error? If so we can merge this for 0.5

zurfyx:
Thank you for your contribution @strdr4605! We discussed this as a team, while we think this is correct we want to double down on the initiative to move the initialEditorState to LexicalComposer https://github.com/facebook/lexical/pull/3032.

@fantactuka is working on the Collab equivalent, a LexicalCollaborationComposer, which besides the initialEditorState it will also remove the need to have a separate Collab plugin and have all the collab basics initializing altogether.

strdr4605:
> Thank you for your contribution @strdr4605! We discussed this as a team, while we think this is correct we want to double down on the initiative to move the initialEditorState to LexicalComposer https://github.com/facebook/lexical/pull/3032.
> 
> 
> 
> @fantactuka is working on the Collab equivalent, a LexicalCollaborationComposer, which besides the initialEditorState it will also remove the need to have a separate Collab plugin and have all the collab basics initializing altogether.

Ok. Waiting for the implementation!

strdr4605:
Also, while preparing for our release. Testing the collaborative editing with ~4-6 users, we had some should not happen errors from lexical. And we are not sure where the problem is, lexical? Yjs or our websocket implementation. 

I would suggest to test collaborative with ~4 users (different browsers), maybe running 2 instances of playground with yjs ws servers connected one persistence DB.

fantactuka:
We recently updated errors so it should be easier to spot what has happened: https://github.com/facebook/lexical/pull/2971

trueadm:
Don’t forget to update any documentation if needed.

pmioulet:
Is there anyway to avoid the initialEditorState to end up in the undo stack that comes with the history plugin? I would like to avoid the full undo of a document you first load as it does not make much sense. Thanks

PR #3016, opened 2022-09-14
Fix image copying

trueadm:
Fixes copying and pasting of images using node selection

### DIFF
--- a/packages/lexical-playground/src/nodes/ImageNode.tsx
+++ b/packages/lexical-playground/src/nodes/ImageNode.tsx
@@ -114,7 +114,8 @@ export class ImageNode extends DecoratorNode<JSX.Element> {
 
   exportDOM(): DOMExportOutput {
     const element = document.createElement('img');
-    element.setAttribute('src', this.__src);
+    // Add host to URL or it will get stripped from clipboard
+    element.setAttribute('src', window.location.href + this.__src);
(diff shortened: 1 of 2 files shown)


zurfyx (commented):
Will this work for base64 images?

fantactuka:
To @zurfyx point:
exportDOM receives editor as an argument, would it be too weird to get node's `img` DOM element and get its src property (which will include full normalized url):

Screen Shot 2022-09-14 at 4 03 56 PM

fantactuka:
To @zurfyx point: thought it would be possible to get normalized url from node's dom element, but it's hard to get to it since it's a decorator. But can just create it in memory:
```
const image = new Image();
image.src = this.__src;
return image.src; // this will be normalized
```

acywatson (approved):

zurfyx:
> To @zurfyx point: thought it would be possible to get normalized url from node's dom element, but it's hard to get to it since it's a decorator. But can just create it in memory:
> 
> ```
> const image = new Image();
> image.src = this.__src;
> return image.src; // this will be normalized
> ```

But if you do window.location.href on a base64 it will break the src no?

Fonger:
If what we need is host, why location.href here?

acywatson:
This won't work, per the comments. We need to fix it some other way.

Issue #3010, opened 2022-09-14
Bug: Lexical nodes not persisting styles present in original HTML


manish2612:
Lexical version: 0.4.1

## Steps To Reproduce

1. Generate HTML using Lexical APIs
2. Use generated HTML from step 1 and provide to Lexical editor. use generateNode API provided by Lexical.
3. Resulting content in editor will not have style which it supposed to have

Following HTML was generated by Lexical and saved in DB. We provide it as is to Lexical and generate nodes (https://lexical.dev/docs/concepts/serialization) from it

```
<p class="RMXContentTheme__paragraph"
 data-lexical-node-type="paragraph"
 data-lexical-node-json='{"children":[],"direction":"ltr","format":"","indent":0,"type":"paragraph","version":1}'
 data-lexical-editor-key="rpoel">
 <span
   data-lexical-node-type="text"
   data-lexical-node-json='{"detail":0,"format":0,"mode":"normal","style":"color: #e00000;background-color: #f8e71c;","text":"manish","type":"text","version":1}'
   data-lexical-editor-key="rpoel"
   style="color: rgb(224, 0, 0); background-color: rgb(248, 231, 28)">
       Hello World
   </span>
</p>
```

## The current behavior

After deserializing HTML content **its NOT PERSISTING style properties** which were present in HTML
```
<p
  class="RMXContentTheme__paragraph"
  data-lexical-node-type="paragraph"
  data-lexical-node-json='{"children":[],"direction":"ltr","format":"","indent":0,"type":"paragraph","version":1}'
  data-lexical-editor-key="qktmf">
  <span
    data-lexical-node-type="text"
    data-lexical-node-json='{"detail":0,"format":0,"mode":"normal","style":"","text":"manish","type":"text","version":1}'
    data-lexical-editor-key="qktmf">
    Hello World
  </span>
</p>
```

## The expected behavior

After deserializing HTML content **it SHOULD PERSIST style properties** which were present in HTML
```
<p class="RMXContentTheme__paragraph"
 data-lexical-node-type="paragraph"
 data-lexical-node-json='{"children":[],"direction":"ltr","format":"","indent":0,"type":"paragraph","version":1}'
 data-lexical-editor-key="rpoel">
 <span
   data-lexical-node-type="text"
   data-lexical-node-json='{"detail":0,"format":0,"mode":"normal","style":"color: #e00000;background-color: #f8e71c;","text":"manish","type":"text","version":1}'
   data-lexical-editor-key="rpoel"
[...]

PR #3017, opened 2022-09-14
Preserve clipboard color/background-color for TextNode

zurfyx:
https://user-images.githubusercontent.com/193447/190220228-b55485a4-e1cc-4626-8449-cd69e8bbccb2.mov
(Recorded forcing HTML paste)

Not the most complete fix but I'd rather us allow list these properties one by one than having a wildcard and randomly importing all the props from other sources

Closes https://github.com/facebook/lexical/issues/3010

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -836,6 +836,10 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
   const hasUnderlineTextDecoration = span.style.textDecoration === 'underline';
   // Google Docs uses span tags + vertical-align to specify subscript and superscript
   const verticalAlign = span.style.verticalAlign;
+  const color = span.style.color;
+  const backgroundColor = span.style.backgroundColor;
+
+  const rawStyles: string[] = [];
 
   return {
     forChild: (lexicalNode) => {
@@ -859,6 +863,16 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
       }
       if (verticalAlign === 'super') {
         lexicalNode.toggleFormat('superscript');
+      }
+
+      if (color !== undefined) {
+        rawStyles.push(`color: ${color}`);
+      }
+      if (backgroundColor !== undefined) {
+        rawStyles.push(`background-color: ${backgroundColor}`);
+      }
+      if (rawStyles.length > 0) {
+        lexicalNode.setStyle(rawStyles.join(';'));
       }
 
       return lexicalNode;


acywatson (commented):
The problem with this is that editors that don’t support background color (for example) will now have the background color added to text nodes with no way to change it. The user could paste without formatting, but this is all-or-nothing.  The right approach is ultimately to allow overriding the importDOM on core nodes either via an Editor API or Proxies

zurfyx:
> The problem with this is that editors that don’t support background color (for example) will now have the background color added to text nodes with no way to change it. The user could paste without formatting, but this is all-or-nothing. The right approach is ultimately to allow overriding the importDOM on core nodes either via an Editor API or Proxies

Good point, sounds to me like the preferred approach would be to define a set of styles that we support and prevent users to freely pass `setStyle`. Instead, an array of allowlisted values would control what gets set and not, and importDOM wouldn't have to worry about this.

To pass this array I believe we could leverage Proxies.

I'll park this PR for this until Proxies are ready

trueadm:
> The problem with this is that editors that don’t support background color (for example) will now have the background color added to text nodes with no way to change it. The user could paste without formatting, but this is all-or-nothing. The right approach is ultimately to allow overriding the importDOM on core nodes either via an Editor API or Proxies

Isn't this how Google Docs and Word already do things though? Shouldn't we be following suite. I'm fine with the implementation.

acywatson:
> > The problem with this is that editors that don’t support background color (for example) will now have the background color added to text nodes with no way to change it. The user could paste without formatting, but this is all-or-nothing. The right approach is ultimately to allow overriding the importDOM on core nodes either via an Editor API or Proxies
> 
> Isn't this how Google Docs and Word already do things though? Shouldn't we be following suite. I'm fine with the implementation.

I mean, obviously I disagree. We did this exact thing in an earlier PR and had to revert it due to internal feedback. You can tell people to just paste as plain text, but many don’t know how to do that and honestly shouldn’t have to imo. So few surfaces support colored text that it this doesn’t seem like a sane default.

trueadm:
> I mean, obviously I disagree. We did this exact thing in an earlier PR and had to revert it due to internal feedback. You can tell people to just paste as plain text, but many don’t know how to do that and honestly shouldn’t have to imo. So few surfaces support colored text that it this doesn’t seem like a sane default. Google Docs and Word aren’t the right things to benchmark against here because every surface we support isn’t a full-fledged word processor. In fact, that’s a small minority, I guess.

Makes sense. Let's abandon this idea then.

acywatson:
Yea, thinking more about this (and reading internal feedback about GDocs) this is not the right thing to default to, so I'm closing. The solution here would be to add support for these properties in userland. If that's hard now, we should make it easier, not build this into the core.

Issue #2500, opened 2022-06-23
Feature: Export `createEmptyEditorState` and other LexicalEditorState functions


nc:
By default EditorState manipulation functions like `createEmptyEditorState` are not exported. It'd be great to have these also exported by default so it's easier to create and manage EditorState.

trueadm:
We don’t want people manually constructing EditorStates as it won’t work properly outside an update cycle and it won’t work with the reconciler properly. So it’s unlikely we’ll ever expose this function as it’s an implementation detail.

ProofOfKeags:
@trueadm how are we supposed to supply the `initialEditorState` argument to `LexicalComposer` if there is no way to construct an editor state...?

PR #3018, opened 2022-09-14
Export createEmptyEditorState

caburj:
Useful for resetting the editor to a blank slate.

### DIFF
--- a/packages/lexical/src/index.ts
+++ b/packages/lexical/src/index.ts
@@ -22,7 +22,7 @@ export type {
   SerializedEditor,
   Spread,
 } from './LexicalEditor';
-export type {EditorState, SerializedEditorState} from './LexicalEditorState';
+export type {EditorState, SerializedEditorState, createEmptyEditorState} from './LexicalEditorState';
 export type {
   DOMChildConversion,
   DOMConversion,


zurfyx (approved):
LGTM!

trueadm:
We've avoiding doing this in the past: https://github.com/facebook/lexical/issues/2500

caburj:
> We've avoiding doing this in the past: #2500

I see. Can you propose a way to clear the content of the Editor instance? Can I do something like getting the root node and clearing its content in the ‘update’ context?

acywatson:
> > We've avoiding doing this in the past: #2500
> 
> I see. Can you propose a way to clear the content of the Editor instance? Can I do something like getting the root node and clearing its content in the ‘update’ context?

$getRoot().clear() should work I think

acywatson:
Add docs around this:

https://github.com/facebook/lexical/pull/3019

caburj:
> $getRoot().clear() should work I think

I see. I didn’t look further. I just found that ‘clear’ method is in ElementNode which is a super of RootNode. Thanks @acywatson.

acywatson:
> > $getRoot().clear() should work I think
> 
> I see. I didn’t look further. I just found that ‘clear’ method is in ElementNode which is a super of RootNode. Thanks @acywatson.

Thanks for taking the time to put up a PR! That's always welcome.

PR #3020, opened 2022-09-14
[0.5] $wrapLeafNodesInElements -> $wrapNodes

zurfyx:
The "LeafNodes" part is a lie. Plus, given it's part of the @lexical/selection package we can simplify the name

### DIFF
--- a/packages/lexical-selection/README.md
+++ b/packages/lexical-selection/README.md
@@ -95,12 +95,12 @@ Expands the current Selection to cover all of the content in the editor.
 export function $selectAll(selection: RangeSelection): void;
 ```
 
-#### `$wrapLeafNodesInElements`
+#### `$wrapNodes`
 
 Attempts to wrap all leaf nodes in the Selection in ElementNodes returned from createElement. If wrappingElement is provided, all of the wrapped leaves are appended to the wrappingElement. It attempts to append the resulting sub-tree to the nearest safe insertion target.
 
 ```ts
-export function $wrapLeafNodesInElements(
+export function $wrapNodes(
   selection: RangeSelection,
   createElement: () => ElementNode,
   wrappingElement?: ElementNode,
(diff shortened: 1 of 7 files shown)


packages/lexical-selection/README.md
```diff
@@ -95,12 +95,12 @@ Expands the current Selection to cover all of the content in the editor.
 export function $selectAll(selection: RangeSelection): void;
 ```
 
-#### `$wrapLeafNodesInElements`
+#### `$wrapNodes`
 
 Attempts to wrap all leaf nodes in the Selection in ElementNodes returned from createElement. If wrappingElement is provided, all of the wrapped leaves are appended to the wrappingElement. It attempts to append the resulting sub-tree to the nearest safe insertion target.
```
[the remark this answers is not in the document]
acywatson: https://typedoc.org/guides/doccomments/

packages/lexical-selection/src/index.ts
```diff
@@ -586,8 +586,7 @@ function $removeParentEmptyElements(startingNode: ElementNode): void {
   }
 }
 
-// TODO 0.6 Rename to $wrapDescendantNodesInElements
-export function $wrapLeafNodesInElements(
+export function $wrapNodes(
```
acywatson, line 589: Not even sure this should take a selection. Does it wrap every node in the Selection into the wrapper? The semantics seem off.

acywatson (changes requested):
Couple comments

packages/lexical-selection/src/index.ts
zurfyx: It's quite a magical function, it takes a selection because based on the selection it can wrap on one way or another (unlike `$wrapNodeInElement`). I personally find it more explicit this way (especially after a chat with @fantactuka).

Passing a list of nodes in would require some change of behavior, in which case you probably want to use the use function instead.

Let's connect offline if you have strong feeling about this one

acywatson (approved):

PR #3027, opened 2022-09-16
Fix nested mark creation when wrapping forward selection

### LINKED ISSUE #3026: Bug: Inconsistent behavior wrapping selections with mark nodes when intersecting with other mark nodes
Lexical version: v0.4.1

### First Case: Backward selection intersecting with existing mark node

01-backward

This correctly ✅ results in three mark nodes. The first references the first comment, the second mark node references both the first and second comments, and the third mark node references only the second comment:

image

Here is an exported playground state (gzipped to allow attaching to Github): [Playground 2022-09-15T21 45 47.320Z.lexical.gz](https://github.com/facebook/lexical/files/9579256/Playground.2022-09-15T21.45.47.320Z.lexical.gz)

### Second Case: Forward selection intersecting with existing mark node

02-forward

This incorrectly ❌ results in two mark nodes, the first of which references the second comment, and the second of which references both the first and second comments (which is incorrect, because the second comment should only refer to part of the text wrapped in the second mark node):

image

Here is an exported playground state (gzipped to allow attaching to Github): [Playground 2022-09-15T21 48 33.896Z.lexical.gz](https://github.com/facebook/lexical/files/9579257/Playground.2022-09-15T21.48.33.896Z.lexical.gz)

Link to code example: The latest playground: https://github.com/facebook/lexical/tree/v0.4.1

## The current behavior

Backward selections intersecting with mark nodes get wrapped and split properly, but forward selections intersecting with mark nodes do not get wrapped and split properly.

## The expected behavior

Both forward and backward selections which intersect with mark nodes should get wrapped and split properly.

knpwrs:
As noted in #3026, creating a nested mark while wrapping a forward selection is currently broken because the logic sets all inline elements as the target node to be wrapped, whereas if the inline element is a mark node we should continue in to the existing mark node to split its children, rather than wrap the entire mark node. This only affects forward selections because the forward selection contains the mark node, whereas a backward selection which ends inside another mark node does not include that mark node.

I also added some comments and renamed one variable for additional clarity in following the logic. The comments may be able to be improved, they reflect my understanding of this code just coming to it from the outside.

Fixes #3026

### DIFF
--- a/packages/lexical-mark/src/index.ts
+++ b/packages/lexical-mark/src/index.ts
@@ -77,13 +77,16 @@ export function $wrapSelectionInMarkNode(
           ? splitNodes[1]
           : splitNodes[0];
     } else if ($isMarkNode(node)) {
-      // Case 2: the node is a mark node and we can ignore it, moving on to its
-      // children. Note that when we make a mark inside another mark, it may
-      // utlimately be unnested by a call to
+      // Case 2: the node is a mark node and we can ignore it as a target,
+      // moving on to its children. Note that when we make a mark inside
+      // another mark, it may utlimately be unnested by a call to
       // `registerNestedElementResolver<MarkNode>` somewhere else in the
       // codebase.
+
       continue;
     } else if ($isElementNode(node) && node.isInline()) {
+      // Case 3: inline element nodes can be added in their entirety to the new
+      // mark
       targetNode = node;
     }
 
--- a/packages/lexical-mark/src/index.ts
+++ b/packages/lexical-mark/src/index.ts
@@ -41,15 +41,19 @@ export function $wrapSelectionInMarkNode(
   const startOffset = isBackward ? focusOffset : anchorOffset;
   const endOffset = isBackward ? anchorOffset : focusOffset;
   let currentNodeParent;
-  let currentMarkNode;
+  let lastCreatedMarkNode;
 
   // We only want wrap adjacent text nodes, line break nodes
   // and inline element nodes. For decorator nodes and block
-  // element nodes, we stop out their boundary and start again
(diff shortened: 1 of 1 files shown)


acywatson:
> As noted in #3026, creating a nested mark while wrapping a forward selection is currently broken because the logic sets all inline elements as the target node to be wrapped, whereas if the inline element is a mark node we should continue in to the existing mark node to split its children, rather than wrap the entire mark node. This only affects forward selections because the forward selection contains the mark node, whereas a backward selection which ends inside another mark node does not include that mark node.
> 
> I also added some comments and renamed one variable for additional clarity in following the logic. The comments may be able to be improved, they reflect my understanding of this code just coming to it from the outside.
> 
> Fixes #3026

I think this makes sense. If I understand the logic correctly, in the case you demoed in the issue:

1) We select part of a paragraph and wrap it in a mark node, splitting the text into two text nodes (one inside the mark and one out).

2) We select across the boundary of the first mark node and wrap that selection in a mark node.

2a) $wrapSelectionInMarkNode walks through the nodes in the selection, wrapping them in a new mark node. When it encounters the existing mark node (created in one) it skips it and moves to its first child. 

2b) Because that child has a different parent than the last child of the mark node we just created, we end up setting lastCreatedMarkNode to undefined and therefore ultimately creating a new mark node that contains all the rest of the nodes in the selection.

This makes sense to me and appears to work in the scenarios I've tested manually. I'd love to have to an e2e for this, but we can follow up with it, as I would prefer to have this fix in sooner rather than later.

Thanks for digging into this!

acywatson (approved):

acywatson (approved):

PR #3028, opened 2022-09-16
Allow custom check to show node ids in tree view

knpwrs:
This is just a quick PR to allow custom logic for showing node ids in the tree view, while still being backwards compatible with the existing implementation.


fantactuka:
This PR adds pretty specific prop to show, and ideally you'd want it to be more generic. That said, I'd suggest to take a look into Lexical dev tools which would show you all node props. It's not published in Chrome extensions store, but you can install it locally, [see here](https://github.com/facebook/lexical/tree/main/packages/lexical-devtools#local-development)


https://user-images.githubusercontent.com/132642/190685947-d466fac5-740e-4835-9852-a229c811d836.mov

PR #3038, opened 2022-09-20
⏸ TextNode setFormat/setDetail to accept a number only

zurfyx:
Initially I thought it was best to do a full replacement in favor for constants https://github.com/facebook/lexical/pull/3037 just like we have for commands now. Constants are easy to follow and type.

I quitted for two reasons:
1. User churn
2. No bundle size win as I expected. Unfortunately there's no good way to minimize ES Module bundles and the constant names have a decent size

However, what we have now is still not professional. A setter and a getter should return the same value, ideally what we store on the node.

In this PR I'm making setFormat and setFormat to work exclusively with numbers and additional provide the toggle and has functions for both.

Closes https://github.com/facebook/lexical/pull/3037

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -496,16 +496,13 @@ export class TextNode extends LexicalNode {
     return;
   }
 
-  // TODO 0.5 This should just be a `string`.
-  setFormat(format: TextFormatType | number): this {
+  setFormat(format: number): this {
     const self = this.getWritable();
-    self.__format =
-      typeof format === 'string' ? TEXT_TYPE_TO_FORMAT[format] : format;
+    self.__format = format;
     return self;
   }
 
-  // TODO 0.5 This should just be a `string`.
-  setDetail(detail: TextDetailType | number): this {
+  setDetail(detail: number): this {
     const self = this.getWritable();
     self.__detail =
       typeof detail === 'string' ? DETAIL_TYPE_TO_DETAIL[detail] : detail;
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -28,7 +28,6 @@ import {
   DETAIL_TYPE_TO_DETAIL,
   IS_BOLD,
   IS_CODE,
-  IS_DIRECTIONLESS,
   IS_ITALIC,
   IS_SEGMENTED,
   IS_STRIKETHROUGH,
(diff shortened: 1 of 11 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -523,16 +512,9 @@ export class TextNode extends LexicalNode {
...
-    const self = this.getWritable();
-    self.__detail ^= IS_DIRECTIONLESS;
-    return self;
-  }
-
-  toggleUnmergeable(): this {
-    const self = this.getWritable();
-    self.__detail ^= IS_UNMERGEABLE;
-    return self;
+  toggleDetail(detail: TextDetailType): this {
```
fantactuka, line 515: Will it benefit to have this exposed vs toggleUnmergeable/toggleDirectionless? It will bring similar confusion that you've tried to avoid when setDetail takes number, and toggleDetail uses strings

zurfyx: @fantactuka What about format (bold, italics, etc), should they be toggleBold/toggleItalics etc?

fantactuka: Ah, I see now 👍

fantactuka (approved):

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -496,19 +489,15 @@ export class TextNode extends LexicalNode {
...
     const self = this.getWritable();
-    self.__format =
-      typeof format === 'string' ? TEXT_TYPE_TO_FORMAT[format] : format;
+    self.__format = format;
     return self;
   }
 
-  // TODO 0.5 This should just be a `string`.
-  setDetail(detail: TextDetailType | number): this {
+  setDetail(detail: number): this {
```
thegreatercurve, line 498: This API has already been reported as confusing by users. Why should `toggleDetail` accept a string but `setDetail` accept a number?

thegreatercurve (changes requested):

thegreatercurve (approved):

trueadm:
I'm not super sold on this. The `__detail` property was never meant to be exposed really, it was used to store internal information only. Having a set/get for it is more confusing IMO.

I think set/get Format is a different thing entirely here too. That is meant to be exposed.

trueadm (changes requested):
Let's rethink the __detail changes.

trueadm (approved):

trueadm:
Is this going in 0.7?

thegreatercurve:
Closing this due to PR becoming stale.

PR #3039, opened 2022-09-20
⏸ ElementNode to manipulate format as a string

zurfyx:
Follow up of https://github.com/facebook/lexical/pull/3038. setFormat on an ElementNode !== LexicalTextNode. They are completely different, it's not even a toggle, merely a string. In fact, I'd argue that it should be called `align`, but I quitted because of user churn when doing this:

```
export type SerializedElementNodeV1 = Spread<
  {
    children: Array<SerializedLexicalNode>;
    direction: 'ltr' | 'rtl' | null;
    format: ElementAlignType;
    indent: number;
  },
  SerializedLexicalNode
>;

export type SerializedElementNodeV2 = Spread<
  {
    children: Array<SerializedLexicalNode>;
    direction: 'ltr' | 'rtl' | null;
    align: ElementAlignType;
    indent: number;
  },
  SerializedLexicalNode
>;
```

In any case, there's still room to do better. Since the current value is stored a string it's already easy to manipulate and read, so I'm just dropping all the copy-pasted logic from toggles and making it a simple getter-setter. 

If we need additional "formats" in the future, it should logically fit in another property.


packages/lexical/src/LexicalReconciler.ts
```diff
@@ -118,19 +111,29 @@ function setElementIndent(dom: HTMLElement, indent: number): void {
...
-    setTextAlign(domStyle, '');
-  } else if (format === IS_ALIGN_LEFT) {
-    setTextAlign(domStyle, 'left');
-  } else if (format === IS_ALIGN_CENTER) {
-    setTextAlign(domStyle, 'center');
-  } else if (format === IS_ALIGN_RIGHT) {
-    setTextAlign(domStyle, 'right');
-  } else if (format === IS_ALIGN_JUSTIFY) {
-    setTextAlign(domStyle, 'justify');
+  switch (format) {
```
fantactuka, line 117: Do we need this switch here as it's 1:1 format to style mapping now?

fantactuka (approved):

trueadm:
Is this going in 0.7?

thegreatercurve:
Closing this due to PR becoming stale.

PR #3051, opened 2022-09-22
Improve table selection handling when there are no siblings

tylerjbainbridge:
Before

https://user-images.githubusercontent.com/13852400/191811851-aa1c7b6d-188d-411a-b1a5-0aa3a5d37248.mov

After

https://user-images.githubusercontent.com/13852400/191811862-da68fa77-9192-4c61-b3ff-62edf3ea2710.mov

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -929,16 +929,12 @@ export function applyTableHandlers(
             ) {
               const gridSelection = DEPRECATED_$createGridSelection();
               const tableKey = tableNode.getKey();
-              const tableFirstDescendant = tableNode.getFirstDescendant();
-              const tableLastDescendant = tableNode.getLastDescendant();
 
-              const firstCell =
-                tableFirstDescendant &&
-                $findMatchingParent(tableFirstDescendant, $isTableCellNode);
+              const firstCell = tableNode
+                .getFirstChildOrThrow()
+                .getFirstChild();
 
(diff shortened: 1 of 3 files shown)


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -916,6 +920,39 @@ export function applyTableHandlers(
             $addHighlightStyleToTable(tableSelection);
 
             return true;
+          } else if (selectionIsInsideTable) {
```
tylerjbainbridge, line 923: This basically just checks if it's a range selection located inside of a table and if all of the cells are selected. If so, it converts the range selection to a grid selection.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -916,6 +920,39 @@ export function applyTableHandlers(
...
+              const tableFirstDescendant = tableNode.getFirstDescendant();
+              const tableLastDescendant = tableNode.getLastDescendant();
+
+              const firstCell =
+                tableFirstDescendant &&
+                $findMatchingParent(tableFirstDescendant, $isTableCellNode);
+
+              const lastCell =
+                tableLastDescendant &&
+                $findMatchingParent(tableLastDescendant, $isTableCellNode);
```
fantactuka, line 941: Would it be easier to do tableNode.getFirstChild.getFirstChild and tableNode.getLastChild.getLastChild to get first/last cell?

fantactuka (approved):

PR #3056, opened 2022-09-23
Dispatch CAN_{REDO,UNDO}_COMMAND after clearing history

wilbertom:
Without this, registered events to `CAN_REDO_COMMAND` and `CAN_UNDO_COMMAND` can have invalid true values after clearing the history.

You can replicate the bug by:

1. Opening the playground. Notice the redo/undo buttons are disabled by default.
2. Create an undo/redo history for the editor.
3. Now both the undo and redo buttons are enabled.
4. Hit the import button button and select a `.lexical` file.
5. After the import, both buttons are enabled even though clicking them doesn't change anything.

The UI was never notified that the editor can't undo/redo anymore because its history was cleared:

These were never called on `ToolbarPlugin`:

```ts
      activeEditor.registerCommand<boolean>(
        CAN_UNDO_COMMAND,
        (payload) => {
          setCanUndo(payload);
          return false;
        },
        COMMAND_PRIORITY_CRITICAL,
      ),
      activeEditor.registerCommand<boolean>(
        CAN_REDO_COMMAND,
        (payload) => {
          setCanRedo(payload);
          return false;
        },
        COMMAND_PRIORITY_CRITICAL,
      ),
```

This is my first day knowing about Lexical and first time writing typescript so sorry if there's any bugs.

Question, what is the difference between `CLEAR_EDITOR_COMMAND` and `CLEAR_HISTORY_COMMAND`. Should that also dispatch the `CAN_*_COMMAND`s?

### DIFF
--- a/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
+++ b/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
@@ -14,6 +14,10 @@ import {$createQuoteNode} from '@lexical/rich-text/src';
 import {$wrapNodes} from '@lexical/selection/src';
 import {
   $createRangeSelection,
+  CAN_REDO_COMMAND,
+  CAN_UNDO_COMMAND,
+  CLEAR_HISTORY_COMMAND,
+  COMMAND_PRIORITY_CRITICAL,
   LexicalEditor,
   SerializedElementNode,
   SerializedTextNode,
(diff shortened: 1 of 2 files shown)


fantactuka (approved):
👍

zurfyx:
LGMT, thank you!

> Question, what is the difference between CLEAR_EDITOR_COMMAND and CLEAR_HISTORY_COMMAND. Should that also dispatch the CAN_*_COMMANDs?

Clear editor is used to order the editor to clear its content, the bin button at the bottom. Given that it currently clears the history I believe it **does make sense** to replicate what you did for this one.

That said, my hunch is that the current logic in the history plugin is not being triggered due to the command priorities (as I can undo just fine after clearing). Could you check both this missing logic and the command priority on the follow-up PR if you got some additional bandwidth? Thank you!

PR #3057, opened 2022-09-23
feat: add export/import DOM for horizontal rule node

samuelmartineau:
Hi the team,

Feel free to reject this pull request if you had a reason to remove exportDOM/importDOM for HorizontalRuleNode here https://github.com/facebook/lexical/pull/2324.

Currently, we lose the information during import/export since the horizontal rule is replaced by a div due to the DecoratorNode type (the root decorated with the React Component)


zurfyx (approved):
Sounds like an accidental removal as you mentioned, thanks!

Issue #3058, opened 2022-09-23
Bug: Pasting from Google Docs adds extra paragraphs / linebreaks


Piliuta:
Copying and pasting text from Google docs with empty paragraphs are converted into linebreaks wrapped into a paragraph.

Lexical version: 0.4.1

## Steps To Reproduce

1. Create a Google Doc with multiple paragraphs
2. Copy the text
3. Paste into a lexical editor

https://user-images.githubusercontent.com/1575198/191975463-957761eb-e091-4f5d-aecf-766c17bbe0cb.mov

<img width="1120" alt="image" src="https://user-images.githubusercontent.com/1575198/191975550-61c756c7-8589-4466-b7cc-71a8ed5f097c.png">

## The current behavior
What's considered "paragraph" in Google Doc is converted into a linebreak(s) wrapped with paragraph in lexical

## The expected behavior
Paragraphs should be created as paragraph nodes when pasted from Google Docs. This is how any other editors handle it.

PR #3063, opened 2022-09-24
fix(lexical-clipboard): pasting from google docs

LuciNyan:
## Description
Fix pasting from google docs adds extra paragraphs / linebreaks

Closes: #3058

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
@@ -2168,7 +2168,6 @@ test.describe('CopyAndPaste', () => {
         </p>
         <p class="PlaygroundEditorTheme__paragraph">
           <br />
-          <br />
         </p>
       `,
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
```diff
@@ -2168,7 +2168,6 @@ test.describe('CopyAndPaste', () => {
         </p>
         <p class="PlaygroundEditorTheme__paragraph">
           <br />
-          <br />
         </p>
```
LuciNyan, line 2171: I was wondering why the expectation here is two br. Should it be changed here to one br ?

acywatson: Probably a situation where we copied the output HTML into the test assertion and just missed this. Single BR should be right inside an empty paragraph.

acywatson (approved):

PR #3068, opened 2022-09-25
chore(lexical-playground): add resizability back to regular tables

Dattaya:
Idk if it was an accident or on purpose, but regular tables are no longer resizable, because TableCellResizer was removed. It doesn't look like this plugin clashes with the new experimental table at a first glance.


zurfyx (approved):
It probably was, thank you!

PR #3069, opened 2022-09-25
pref(lexical-playground): reduce rendering consumption by creating compositing layers

LuciNyan:
## Description
Reduce rendering consumption by creating compositing layers

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -1141,6 +1141,9 @@ i.prettier-error {
 
 .table-cell-action-button-container {
   position: absolute;
+  top: 0;
+  left: 0;
+  will-change: transform;
 }
 
 .table-cell-action-button {
--- a/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.css
+++ b/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.css
@@ -2,9 +2,11 @@
   border-radius: 4px;
   padding: 2px 1px;
   cursor: grab;
+  opacity: 0;
   position: absolute;
-  left: -10000px;
-  top: -10000px;
+  left: 0;
+  top: 0;
+  will-change: transform;
 }
 
 .draggable-block-menu .icon {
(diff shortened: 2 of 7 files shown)


acywatson:
Not sure about this one - was there a problem here you were trying to address?

LuciNyan:
> Not sure about this one - was there a problem here you were trying to address?

I don't think the previous implementation is a "problem", but it looks better with transform. It just moves the layer instead of reflowing it.

zurfyx:
I believe that the point @acywatson is trying to make is whether the proposed approach has noticeable a performance win. It is correct that `translate` is the preferred approach performance-wise as opposed to `top` and `left` that will always run on CPU. However, your code becomes more complicated to understand after that. I will delegate the approval to @acywatson since he's been the first the approach this PR.

trueadm:
This should be much smoother on mobile as they generally have decent GPU power and suffer from over constrained CPU cores. Have you done any mobile comparisons?

acywatson:
> I believe that the point @acywatson is trying to make is whether the proposed approach has noticeable a performance win. It is correct that `translate` is the preferred approach performance-wise as opposed to `top` and `left` that will always run on CPU. However, your code becomes more complicated to understand after that. I will delegate the approval to @acywatson since he's been the first the approach this PR.

I'm fine with the translation change, but was more wondering if will-change is the right thing to use here. Generally you dpn't need to start messing with things like this unless there's an explicit performance issue you're trying to solve.

This might be a noticeable improvement on mobile - would be interesting to see the comparison.

acywatson (approved):

LuciNyan:
> > I believe that the point @acywatson is trying to make is whether the proposed approach has noticeable a performance win. It is correct that `translate` is the preferred approach performance-wise as opposed to `top` and `left` that will always run on CPU. However, your code becomes more complicated to understand after that. I will delegate the approval to @acywatson since he's been the first the approach this PR.
> 
> I'm fine with the translation change, but was more wondering if will-change is the right thing to use here. Generally you dpn't need to start messing with things like this unless there's an explicit performance issue you're trying to solve.
> 
> This might be a noticeable improvement on mobile - would be interesting to see the comparison.

Hi @acywatson! I really appreciate the questions you asked. Your question made me think seriously about whether so is really worth it. I had only read in many places that transform was better than left top, but I hadn't explored the issue myself. But after trying it myself, I honestly don't see a significant difference. Can you give me some advice on testing, I really want to know if we should use transform instead of left top now, or if both are ok!

PR #3080, opened 2022-09-28
Double-trigger hack for high-fidelity Node/Grid selection

zurfyx:
## Problem

Copying a Node will only store `text/html` into the clipboard. This differs from copying a RangeSelection when we copy `text/plain`, `text/html` and `application/x-lexical-editor` to fulfill a wider range of use cases. For example, when pasting content we would rather have the JSON serialized content as its the exact copy (most fidelity) Lexical can offer.

The current Node and Grid selection copy relies on the Clipboard API that is still relatively new, to the point that even some everly green browsers do not support setting more than one item at once.

At the same time, the fact that only the newest browsers can rely on Clipboard means that older browsers that we still want to support will be unable to copy Node or Grid.

## Technical problem

The main difference between a RangeSelection and a NodeSelection comes down the DOMSelection. A RangeSelection copy will trigger a ClipboardEvent (via the RootNode copy event listener) whereas the "fake" selection we performs on DecoratorNodes has to be captured via KeyboardEvent. **A KeyboardEvent does not give us access to ClipboardEvent** that ultimately can set arbitrary data onto the clipboard.

## Double-trigger hack

Through experimentation, it turns out we can do better. We can programmatically trigger a copy event after programmatically remapping the selection to a valid range selection.

This hack gives us a second copy event we can now remap just like a normal RangeSelection.

Screen Shot 2022-09-27 at 4 26 01 pm
Screen Shot 2022-09-27 at 4 26 21 pm

## Limitations

While it has wider browser version support it still doesn't work with Firefox. It turns out that neither Clipboard nor this hack work. For Firefox, we'll have to rely on Clipboard once it moves away from the experimental `dom.events.asyncClipboard.clipboardItem`

Screen Shot 2022-09-27 at 4 30 35 pm

## Future work

I already moved this method onto @lexical/clipboard because I think that it has the potential to:
1. Handle custom selection
2. Programmatically trigger copy events (i.e. from toolbar)
3. Plain text to leverage this method to enable use cases such as copy-paste for MentionNodes

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -185,8 +185,7 @@ function $basicInsertStrategy(
 
     /**
      * There's no good way to add this to importDOM or importJSON directly,
-     * so this is here in order to safely correct faultimport { copyToClipboard__EXPERIMENTAL } from './clipboard';
-y clipboard data
+     * so this is here in order to safely correct faulty clipboard data
      * that we can't control and avoid crashing the app.
      * https://github.com/facebook/lexical/issues/2405
      */
@@ -574,6 +573,7 @@ export function copyToClipboard__EXPERIMENTAL(
       clipboardEventTimeout = null;
     }, EVENT_LATENCY);
     document.execCommand('copy');
+    element.remove();
   }
 }
 
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -6,15 +6,6 @@
  *
  */
 
-import type {
-  GridSelection,
-  LexicalEditor,
-  LexicalNode,
-  NodeSelection,
-  RangeSelection,
-  SerializedTextNode,
-} from 'lexical';
-
 import {$generateHtmlFromNodes, $generateNodesFromDOM} from '@lexical/html';
 import {$createListNode, $isListItemNode} from '@lexical/list';
 import {
@@ -34,18 +25,26 @@ import {
   $isTextNode,
   $parseSerializedNode,
   $setSelection,
+  COMMAND_PRIORITY_CRITICAL,
+  COPY_COMMAND,
   DEPRECATED_$createGridSelection,
   DEPRECATED_$isGridCellNode,
   DEPRECATED_$isGridNode,
   DEPRECATED_$isGridRowNode,
   DEPRECATED_$isGridSelection,
   DEPRECATED_GridNode,
+  GridSelection,
+  LexicalEditor,
+  LexicalNode,
+  NodeSelection,
+  RangeSelection,
   SELECTION_CHANGE_COMMAND,
+  SerializedTextNode,
 } from 'lexical';
 import {IS_FIREFOX} from 'shared/environment';
 import invariant from 'shared/invariant';
 
-export function $getHtmlContent(editor: LexicalEditor): string | null {
+export function $getHtmlContent(editor: LexicalEditor): string {
   const selection = $getSelection();
 
   if (selection == null) {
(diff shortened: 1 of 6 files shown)


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -183,7 +185,8 @@ function $basicInsertStrategy(
 
     /**
      * There's no good way to add this to importDOM or importJSON directly,
-     * so this is here in order to safely correct faulty clipboard data
+     * so this is here in order to safely correct faultimport { copyToClipboard__EXPERIMENTAL } from './clipboard';
```
acywatson, line 188: ?

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -491,3 +495,109 @@ export function $generateNodesFromSerializedNodes(
...
+    // the listener will be quickly freed so that the user can reuse it again
+    clipboardEventTimeout = window.setTimeout(() => {
+      removeListener();
+      clipboardEventTimeout = null;
+    }, EVENT_LATENCY);
+    document.execCommand('copy');
+  }
+}
+
+// TODO shouldn't pass editor (pass namespace directly)
```
acywatson, line 580: Downside is forcing users to deal with the concept of a namespace, I guess. Do they have to do that elsewhere?

packages/lexical-playground/src/nodes/TweetNode.tsx
```diff
@@ -173,8 +173,10 @@ export class TweetNode extends DecoratorBlockNode {
   }
 
   exportDOM(): DOMExportOutput {
+    console.info('asdf');
     const element = document.createElement('div');
     element.setAttribute('data-lexical-tweet-id', this.__id);
+    element.innerText = this.getTextContent();
```
acywatson, line 179: ?

packages/lexical-clipboard/src/clipboard.ts
acywatson: Could also just access the active editor and get the namespace from that in the update closure

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -491,3 +495,109 @@ export function $generateNodesFromSerializedNodes(
...
+    return;
+  }
+
+  const rootElement = editor.getRootElement();
+  const domSelection = document.getSelection();
+  if (rootElement !== null && domSelection !== null) {
+    const element = document.createElement('span');
+    element.style.cssText = 'position: fixed; top: -1000px;';
+    element.append(document.createTextNode('#'));
+    rootElement.append(element);
```
trueadm, line 549: You don’t seem to be removing it later?

packages/lexical-clipboard/src/clipboard.ts
fantactuka: I remember for one of the PRs Dominic added $getActiveEditor, but then removed as it was not used in the end. Maybe it's a valid usecase to bring it back

packages/lexical-playground/src/nodes/TweetNode.tsx
zurfyx: Followed up in https://github.com/facebook/lexical/pull/3088

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -56,13 +56,15 @@ export function $getHtmlContent(editor: LexicalEditor): string | null {
...
-    return null;
+    return '';
   }
 
   return $generateHtmlFromNodes(editor, selection);
 }
 
-export function $getLexicalContent(editor: LexicalEditor): string | null {
+// TODO Return a blank string instead
+// TODO Rename to $getJSON
```
thegreatercurve, line 66: Should this be done in a future feature version? 
```suggestion
// TODO 0.6. Rename to $getJSON
```

PR #3082, opened 2022-09-28
Add collapsible container plugin

fantactuka:
Kind of an experiment with nested elements (vs going w/decorator and nested editors). Main part is in transformers that enforce nesting structure (and if it's violated unwrap content and convert collapsible content into regular paragraphs)
```
CollapsibleContainer
  > CollapsibleTitle
  > CollapsibleContent[isShadowRoot]
    > Paragraph
    > Heading
    > ...
```

https://user-images.githubusercontent.com/132642/192806800-292aa729-c5e6-41d1-9ba6-02259f1d9d6e.mov

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
@@ -88,6 +88,7 @@ import TextInput from '../../ui/TextInput';
 import {getSelectedNode} from '../../utils/getSelectedNode';
 import {sanitizeUrl} from '../../utils/sanitizeUrl';
 import {EmbedConfigs} from '../AutoEmbedPlugin';
+import {INSERT_COLLAPSIBLE_COMMAND} from '../CollapsiblePlugin';
 import {INSERT_EQUATION_COMMAND} from '../EquationsPlugin';
 import {INSERT_EXCALIDRAW_COMMAND} from '../ExcalidrawPlugin';
 import {INSERT_IMAGE_COMMAND} from '../ImagesPlugin';
@@ -1120,6 +1121,14 @@ export default function ToolbarPlugin(): JSX.Element {
               className="item">
               <i className="icon sticky" />
               <span className="text">Sticky Note</span>
+            </DropDownItem>
+            <DropDownItem
+              onClick={() => {
+                editor.dispatchCommand(INSERT_COLLAPSIBLE_COMMAND, undefined);
+              }}
+              className="item">
+              <i className="icon plus" />
+              <span className="text">Collapsible container</span>
             </DropDownItem>
             {EmbedConfigs.map((embedConfig) => (
               <DropDownItem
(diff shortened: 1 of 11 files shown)


packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -1121,6 +1122,14 @@ export default function ToolbarPlugin(): JSX.Element {
               <i className="icon sticky" />
               <span className="text">Sticky Note</span>
             </DropDownItem>
+            <DropDownItem
+              onClick={() => {
+                editor.dispatchCommand(INSERT_COLLAPSIBLE_COMMAND, undefined);
+              }}
+              className="item">
+              <i className="icon plus" />
```
thegreatercurve, line 1130: Maybe `arrow-right` instead?

thegreatercurve (commented):
LGTM! Out of interest, how badly does this mess up browser search for the document? You could possibly look at adding something like `hidden=until-found` but it's still in early development: https://developer.chrome.com/articles/hidden-until-found

thegreatercurve (approved):

Nirlah:
I like the nested nodes approach, it could be nice to have draggable blocks support inside the container and from/to the outside.

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
acywatson: +1

zurfyx (approved):
Awesome work! Not sure how to best validate the CollapsiblePlugin logic besides assuming that you came across them via trial and error and that we potentially still missed some that we will have to address over time.

acywatson:
Never got a chance to finish this, but a while back I was messing around with doing this leaning more on browser native behaviors and semantic html:

https://codesandbox.io/s/lexical-plain-text-example-forked-jrrb38?file=/src/nodes/ToggleNode.js

Yours is certainly way better, just thought I would share

fantactuka:
> Never got a chance to finish this, but a while back I was messing around with doing this leaning more on browser native behaviors and semantic html:
> [codesandbox.io/s/lexical-plain-text-example-forked-jrrb38?file=/src/nodes/ToggleNode.js](https://codesandbox.io/s/lexical-plain-text-example-forked-jrrb38?file=/src/nodes/ToggleNode.js)

I completely forgot about details/summary tags, thank you! It's actually a way better html structure as it both solves collab problem (collapsed/open state won't be synchronized) and auto-open if content matches CMD+F query

acywatson:
> > Never got a chance to finish this, but a while back I was messing around with doing this leaning more on browser native behaviors and semantic html:
> > [codesandbox.io/s/lexical-plain-text-example-forked-jrrb38?file=/src/nodes/ToggleNode.js](https://codesandbox.io/s/lexical-plain-text-example-forked-jrrb38?file=/src/nodes/ToggleNode.js)
> 
> I completely forgot about details/summary tags, thank you! It's actually a way better html structure as it both solves collab problem (collapsed/open state won't be synchronized) and auto-open if content matches CMD+F query

PR #3086, opened 2022-09-29
Merge markdown E2E tests into one file

thegreatercurve:
We only have 1 markdown version now so the E2Es can be merged.


acywatson:
I wonder if one big file is slower and than two smaller ones. It's better to consolidate these for now though, since we don't need references to versioned markdown

acywatson (approved):

PR #3088, opened 2022-09-29
Tweet fallback to text on copy

zurfyx:
Very opinionated PR but I think having a fallback to text for non-Lexical editors is a much better experience

https://user-images.githubusercontent.com/193447/193140469-eb026434-73a2-4155-978d-6ee0af7aacae.mov

### DIFF
--- a/packages/lexical-playground/src/nodes/TweetNode.tsx
+++ b/packages/lexical-playground/src/nodes/TweetNode.tsx
@@ -175,6 +175,8 @@ export class TweetNode extends DecoratorBlockNode {
   exportDOM(): DOMExportOutput {
     const element = document.createElement('div');
     element.setAttribute('data-lexical-tweet-id', this.__id);
+    const text = document.createTextNode(this.getTextContent());
+    element.append(text);
     return {element};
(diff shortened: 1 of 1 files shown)


thegreatercurve:
@zurfyx Maybe want to do the same for YouTube nodes as well?

thegreatercurve (approved):

packages/lexical-playground/src/nodes/TweetNode.tsx
```diff
@@ -175,6 +175,8 @@ export class TweetNode extends DecoratorBlockNode {
   exportDOM(): DOMExportOutput {
     const element = document.createElement('div');
     element.setAttribute('data-lexical-tweet-id', this.__id);
+    const text = document.createTextNode(this.getTextContent());
+    element.append(text);
```
acywatson, line 179: I would even say a link for this type of thing - differs by decorator, I suppose.

thegreatercurve:
@zurfyx Merged this ahead of a sync in error. I think the suggestions can be applied in a separate PR though.

Issue #3060, opened 2022-09-23
Refactor `lexical/selection` into separate files


thegreatercurve:
All of the logic in `lexical/selection/index.js` is contained in one file. We should instead refactor this file into multiple files instead.

abdullahranginwala:
Hello! First time contributor here. Just to clarify is this -> packages/lexical-selection/src/index.ts the file you are referring to? Secondly, are there any guidelines on how you want to code to be split up in multiple files?

abdullahranginwala:
@thegreatercurve Thank you for reaching out. I looked at the code and realised that each function takes the selection type as a parameter. How can I split the code into multiple files based on the selection type?

thegreatercurve:
@abdullahranginwala Can you just import the relevent classes into your utils file?

PR #3094, opened 2022-10-01
Refactored code into a utils file

abdullahranginwala:
Refactored the common the logic in lexical/selection/index.js as discussed in #3060


thegreatercurve (changes requested):
@abdullahranginwala There's not much point to having a common utils file and then just one index file. Ideally, we also have other files for each selection class (e.g. `RangeSelection`, `NodeSelection` and `GridSelection`), a `utils` file for common functions between each class, and then an `index` file which just re-exports everything.

thegreatercurve:
Closing this as someone else submitted a fix for the same Issue: #3145

PR #3104, opened 2022-10-03
Typeaheads: Remove hard-coded "bottom" alignment (Breaking change)

tylerjbainbridge:
Lexical is a rich text editing framework, not a popover framework, so it's a slippery slope to start including dynamic positioning with our plugins. It should be up to the developer to wire our `anchor` to a popover solution of their choice or use a React Portal and menu with some basic margin.

When this plugin was originally created it was designed with the common "bottom" alignment hard coded by forcing the `anchor` to be positioned below the line (see red box below).

image

This works for common use cases but quickly gets in the way as soon as you want to use a popover with auto alignment that may shift positioning to the top of the anchor when it'd otherwise be off screen.

Now the `anchor` is positioned "over" the selection so you can handle the menu positioning however you'd like.

image

### DIFF
--- a/packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
+++ b/packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
@@ -221,17 +221,19 @@ export function LexicalAutoEmbedPlugin<TEmbedConfig extends EmbedConfig>({
       ) =>
         anchorElement && nodeKey != null
           ? ReactDOM.createPortal(
-              <MenuComponent
-                options={options}
-                selectedItemIndex={selectedIndex}
-                onOptionClick={(option: AutoEmbedOption, index: number) => {
-                  setHighlightedIndex(index);
-                  selectOptionAndCleanUp(option);
-                }}
-                onOptionMouseEnter={(index: number) => {
-                  setHighlightedIndex(index);
-                }}
-              />,
+              <div className="typeahead-popover">
+                <MenuComponent
+                  options={options}
+                  selectedItemIndex={selectedIndex}
+                  onOptionClick={(option: AutoEmbedOption, index: number) => {
+                    setHighlightedIndex(index);
+                    selectOptionAndCleanUp(option);
+                  }}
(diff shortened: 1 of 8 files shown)


packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
```diff
@@ -221,17 +221,19 @@ export function LexicalAutoEmbedPlugin<TEmbedConfig extends EmbedConfig>({
...
-                selectedItemIndex={selectedIndex}
-                onOptionClick={(option: AutoEmbedOption, index: number) => {
-                  setHighlightedIndex(index);
-                  selectOptionAndCleanUp(option);
-                }}
-                onOptionMouseEnter={(index: number) => {
-                  setHighlightedIndex(index);
-                }}
-              />,
+              <div className="typeahead-popover">
```
zurfyx, line 224: Will this work fine on WWW builds after minification? Perhaps we should pass style classes as props?

zurfyx (approved):

packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
fantactuka: Is it legit to use theme here?

tylerjbainbridge: This was an accident, I meant to add this to the playground implementation of the menu. Good catch!

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -738,6 +745,7 @@ export function LexicalTypeaheadMenuPlugin<TOption extends TypeaheadOption>({
 }
 
 type NodeMenuPluginProps<TOption extends TypeaheadOption> = {
+  anchorConfig?: AnchorConfig;
```
tylerjbainbridge, line 748: Also I need to remove this before merging.

PR #3106, opened 2022-10-04
Typeaheads: Add scroll command and increase priority

tylerjbainbridge:
When using virtualization or other special rendering methods you may want to listen for the `SCROLL_TYPEAHEAD_OPTION_INTO_VIEW_COMMAND` command.

This PR also increases the priority of typeahead command listeners since when the menu is open any keystrokes may be relevant to the typeahead's internal behavior. For example, there was an issue reported where Table key navigation took priority over Typeaheads that were open inside of them. This PR and #3107 solve this.

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -13,12 +13,14 @@ import {
   $getSelection,
   $isRangeSelection,
   $isTextNode,
-  COMMAND_PRIORITY_NORMAL,
+  COMMAND_PRIORITY_CRITICAL,
+  createCommand,
   KEY_ARROW_DOWN_COMMAND,
   KEY_ARROW_UP_COMMAND,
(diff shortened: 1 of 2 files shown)


packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -319,6 +322,11 @@ export function useDynamicPositioning(
   }, [targetElement, editor, onVisibilityChange, onReposition, resolution]);
 }
 
+export const SCROLL_TYPEAHEAD_OPTION_INTO_VIEW_COMMAND: LexicalCommand<{
```
fantactuka, line 325: Might need flow type as well

fantactuka (approved):

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
tylerjbainbridge: Good call

PR #3108, opened 2022-10-04
Support inline annotations in code blocks

acywatson:
WIP - still have some issues with selection jumping around weirdly.

Not super happy with the fact that @lexical/code needs to depend on @lexical/mark for this.

### DIFF
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -7,7 +7,6 @@
  */
 
 // eslint-disable-next-line simple-import-sort/imports
-// eslint-disable-next-line simple-import-sort/imports
 import type {
   LexicalCommand,
   LexicalEditor,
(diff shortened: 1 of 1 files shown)


packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -199,6 +203,30 @@ function textNodeTransform(node: TextNode, editor: LexicalEditor): void {
...
 }
 
+function getCodeContent(node: CodeNode): string {
+  let textContent = '';
+  const children = node.getChildren();
+  const childrenLength = children.length;
+  for (let i = 0; i < childrenLength; i++) {
+    const child = children[i];
+    if ($isMarkNode(child)) {
+      textContent += `/*${MARK_START_PREFIX}_${child.getIDs().join('_')}_*/`;
```
fantactuka, line 213: This might not work for languages with different comment syntax and break tokenization

fantactuka: Alternative could be counting mark start and end offset to reapply it after tokenization (given it does not change content length)

acywatson: That's a good point.

The offset thing could work

PR #3109, opened 2022-10-04
fix(lexical-playground): draggable blockplugin uses unexposed apis

### LINKED ISSUE #3100: Bug: `DraggableBlockPlugin` uses unexposed apis
Getting `Failed to resolve import "lexical/LexicalUtils"` when trying to use `DraggableBlockPlugin` from the playground. `getElementByKeyOrThrow` is exported by `lexical/LexicalUtils` but the npm package only contains `Lexical.{dev,prod}.js` and `getElementByKeyOrThrow` is not exported anywhere else.

Lexical version: 0.5

image

LuciNyan:
Closes #3100

### DIFF
--- a/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
@@ -18,7 +18,6 @@ import {
   DROP_COMMAND,
   LexicalEditor,
 } from 'lexical';
-import {getElementByKeyOrThrow} from 'lexical/src/LexicalUtils';
 import * as React from 'react';
 import {DragEvent as ReactDragEvent, useEffect, useRef, useState} from 'react';
 import {createPortal} from 'react-dom';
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
```diff
@@ -18,7 +18,6 @@ import {
   DROP_COMMAND,
   LexicalEditor,
 } from 'lexical';
-import {getElementByKeyOrThrow} from 'lexical/src/LexicalUtils';
```
acywatson, line 21: We could also just export it, I guess?

LuciNyan: I noticed that in other plugins, editor.getElementByKey is used when you need to get the element by key. whether we export getElementByKeyOrThrow or not, it looks better to use editor.getElementByKey consistently here. What do you think?

acywatson: I don't have a strong preference - it's mostly just a convenience method, but now you've gone from throwing an error in the case of a key lookup miss to just silently failing. If that's the behavior you want, then this is fine.

acywatson (approved):

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
LuciNyan: Thank you for the detailed explanation here. <3

Issue #1711, opened 2022-04-14
Bug: npm install error caused by install chromium


Junezm:
Lexical version: 0.2.0

## Steps To Reproduce

1. git clone repo
2. npm install

Link to code example:

## The current behavior
 Command: node install.js
Arguments: 
Directory: /Users/frontend/editor/lexical/node_modules/puppeteer
Output:
The chromium binary is not available for arm64.
If you are on Ubuntu, you can install with: 

 sudo apt install chromium

 sudo apt install chromium-browser

/Users/frontend/editor/lexical/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:115
                    throw new Error();
## The expected behavior
the package link-preview-generator unused, so I remove it, then I can install successful

trueadm:
Try running your terminal with compatibility mode if you're using an M1 machine. That's what I have to do. Then everything works great :)

shijie-todd:
> 

How to run terminal with compatibility mode ?

kevinansfield:
If anyone else runs into this, note that you can install chromium manually via homebrew and set some env vars to stop point puppeteer at the right place. The instructions here work well https://linguinecode.com/post/how-to-fix-m1-mac-puppeteer-chromium-arm64-bug

Timebutt:
I ran into the same problem this morning and dropped a [PR](https://github.com/facebook/lexical/pull/3114) to resolve the problem.

PR #3114, opened 2022-10-05
Fix npm install on M1 macs

Timebutt:
This PR removes the unused `link-preview-generator` package which has `puppeteer` `8.0.0` listed as one of its dependencies, which in turn ensures you cannot run `npm i` in the root of this project on an M1 Mac. One of the suggestion in [this issue](https://github.com/facebook/lexical/issues/1711) mentions you can run your terminal in compatibility mode which is not something you want if you are running a shiny M1 chip ;)


acywatson (approved):
Thank you!

PR #3116, opened 2022-10-05
Fix when the selection is anchored to the end of a text node followed by a non-text node

### LINKED ISSUE #3115: Bug: $patchStyleText throws if the selection anchor is at the end of a text node before an inline element
Lexical version: 0.5.0

## Steps To Reproduce

1. Set up a selection with the anchor set to the end of a text node that is followed by a non-text node (e.g. inline element)
2. Pass said selection to `$patchStyleText`

Link to code example: (PR forthcoming)

## The current behavior

Normally `$patchStyleText` detects a selection anchored to the end of a text node and pushes the anchor to the next node, however it won't do that if the next node is not a text node: https://github.com/facebook/lexical/blob/f8ac4240e69add5a392ba4b605114d2bfdbfe829/packages/lexical-selection/src/index.ts#L378-L387

As a result, when we later try to split the text node, we won't get back any second item and hence `firstNode` will become undefined: https://github.com/facebook/lexical/blob/f8ac4240e69add5a392ba4b605114d2bfdbfe829/packages/lexical-selection/src/index.ts#L415-L421

That will case an error when we later try to read from it in `$patchNodeStyle`: https://github.com/facebook/lexical/blob/f8ac4240e69add5a392ba4b605114d2bfdbfe829/packages/lexical-selection/src/index.ts#L342-L343

## The expected behavior

It should not error.

One approach would be simply to make the original fixup apply even if the next sibling is not a text node. Alternatively we could simply _not_ split the node if we detect that our offset it as the end of it.

birtles:
Fixes #3115.

As mentioned in that issue, it might be better to fix this by checking before splitting instead, I'm not sure.

### DIFF
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
@@ -8,7 +8,7 @@
 
 import {$createLinkNode} from '@lexical/link';
 import {$createHeadingNode} from '@lexical/rich-text';
-import {$cloneContents} from '@lexical/selection';
+import {$cloneContents, $patchStyleText} from '@lexical/selection';
 import {
   $createParagraphNode,
   $createTextNode,
@@ -17,6 +17,7 @@ import {
   $getSelection,
   $isNodeSelection,
   $isRangeSelection,
+  RangeSelection,
   TextNode,
 } from 'lexical';
 import {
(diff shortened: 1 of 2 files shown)


birtles:
Actually, I think I want to try adding a test for when there is no next sibling and see of I can produce a similar error.

If so, that would suggest we need to do the check before splitText.

I'll have a look tomorrow.

acywatson:
@birtles Thanks for the fix and sorry for the delay - if you can just resolve the conflicts I'm happy to merge this! Thank you!

acywatson (approved):

birtles:
> @birtles Thanks for the fix and sorry for the delay - if you can just resolve the conflicts I'm happy to merge this! Thank you!

Thanks @acywatson! I've rebased and resolved the conflicts so hopefully this is ready to merge. (I believe the failed e2e-mac test is unrelated.)

PR #3118, opened 2022-10-05
Fix GridSelection comparison

zurfyx:
https://user-images.githubusercontent.com/193447/194163012-f8f06978-afee-48c7-8a20-efabd706c7c5.mov

This PR also prevents TableCellNode properties from being copied over, which was being done accidentally by the previous behavior. Is this what the best behavior? It is with the currently supported feature for TableCells but won't be as soon as we add more. Unfortunately, the good fix for this second issue is to delegate all this Grid logic to it's own package. https://github.com/facebook/lexical/issues/3121


a-xin (approved):
Looking great, and just double-checked this does fix GridSelection for my test case

PR #3119, opened 2022-10-05
Add flushSync option to update()

fantactuka:
This would allow synchronous updates (e.g. for tests and some cases of headless mode):
```js
editor.update(() => {
  ...
}, { flushSync: true });

// Update and listeners are already executed

```

### DIFF
--- a/packages/lexical-playground/src/nodes/TableComponent.tsx
+++ b/packages/lexical-playground/src/nodes/TableComponent.tsx
@@ -6,7 +6,7 @@
  *
  */
 
-import type {EditorState, RangeSelection, TextFormatType} from 'lexical';
+import type {RangeSelection, TextFormatType} from 'lexical';
 
 import {
   $generateJSONFromSelectedNodes,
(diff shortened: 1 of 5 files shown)


trueadm:
I don't think we want to call it `flushSync` as this would encourage people to always use it. `discrete` would be more accurate and follow the terminology browsers and React uses for the same behaviors.

fantactuka:
> discrete

Thought react uses flushSync, but discrete sounds good too

fantactuka:
@trueadm Another option I had in mind is to add `$flushSync()` util to be called within updateFn. It might make it even more explicit than a flag

trueadm:
> Thought react uses flushSync, but discrete sounds good too

That's true, I guess I meant as the underlying terminology. I'm just worried everyone will do this rather than use async updates.

trueadm:
> Another option I had in mind is to add $flushSync() util to be called within updateFn. It might make it even more explicit than a flag

Gets complicated with nested updates though and what should happen when others happen etc.

acywatson:
> That's true, I guess I meant as the underlying terminology. I'm just worried everyone will do this rather than use async updates.

Would this really be a problem? You'd have to go looking for this API and then maybe we could ameliorate the problem with some documentation around valid use cases for it.

trueadm:
Lexical used to have a flush sync flag, we actually killed it. Everyone used that in their tests and then their production surfaces were in async causing a different behavior.

fantactuka:
> Lexical used to have a flush sync flag, we actually killed it. Everyone used that in their tests and then their production surfaces were in async causing a different behavior.

With that I don't like neither of solutions anymore :) But using internal _flushSync flag is even more confusing, so I'll move on with `discrete`

acywatson (approved):

PR #3124, opened 2022-10-06
Add className prop to typeahead plugin

acywatson:
This allows us to add styles to the underlying typeahead menu container.

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -583,10 +583,10 @@ export function useBasicTypeaheadTriggerMatch(
 function useMenuAnchorRef(
   resolution: Resolution | null,
   setResolution: (r: Resolution | null) => void,
+  className?: string,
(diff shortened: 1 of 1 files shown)


thegreatercurve (approved):

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -583,10 +583,10 @@ export function useBasicTypeaheadTriggerMatch(
 function useMenuAnchorRef(
   resolution: Resolution | null,
   setResolution: (r: Resolution | null) => void,
+  className?: string,
```
tylerjbainbridge, line 586: Can you change it to `anchorClassName`? Just want to make to make sure that people don't think this is how you style the typeahead menu itself.

tylerjbainbridge (approved):

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
acywatson: Yea, good call

PR #3126, opened 2022-10-06
Don't apply element based offsets to text nodes in $patchStyleText

### LINKED ISSUE #3125: Bug: $patchStyleText only patches part of the children when the selection ends at the end of an element
Lexical version: 0.5.0

## Steps To Reproduce

1. Set up a selection whose right-hand side is the end of an _element_ (i.e. offset set to the number of children in the element)
2. Pass said selection to `$patchStyleText`

Link to code example: (PR forthcoming)

## The current behavior

Inside `$patchStyleText` we'll fetch the nodes in the selection using `RangeSelection.getNodes()`. This will return a series of nodes including the end element and its text children.

_However_ the last node returned will not be the element the selection ends on but rather its text children. As a result `lastNode` will point to a text node but `endOffset` refers to the child offset within the end element, _not_ an offset within `lastNode`.

Nevertheless, we'll go ahead and split the node using `endOffset`: https://github.com/facebook/lexical/blob/737f8811254ff499755ac5ab5535a736fe9b142a/packages/lexical-selection/src/index.ts#L429-L431

As a result, only a portion of the selected element's children will be formatted.

## The expected behavior

We shouldn't apply child-based offsets to the text of the text node children themselves.

Note that `insertText` has a similar bug but I haven't had a chance to file it yet.

birtles:
Fixes #3125

### DIFF
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
@@ -2932,4 +2932,143 @@ describe('$patchStyleText', () => {
         '<p dir="ltr"><span style="text-emphasis: filled;" data-lexical-text="true">b</span></p>',
     );
   });
+
+  test('can patch a selection that ends on an element', async () => {
+    const editor = createTestEditor();
+    const element = document.createElement('div');
+    editor.setRootElement(element);
+
+    await editor.update(() => {
+      const root = $getRoot();
+
+      const paragraph = createParagraphWithNodes(editor, [
+        {
+          key: 'a',
+          mergeable: false,
+          text: 'a',
+        },
+      ]);
+
+      root.append(paragraph);
+
+      const link = $createLinkNode('https://');
+      link.append($createTextNode('link'));
+
+      const a = $getNodeByKey('a');
+      a.insertAfter(link);
+
+      setAnchorPoint({
+        key: 'a',
+        offset: 0,
+        type: 'text',
+      });
+      // Select to end of the link _element_
+      setFocusPoint({
+        key: link.getKey(),
+        offset: 1,
+        type: 'element',
+      });
+
+      const selection = $getSelection() as RangeSelection;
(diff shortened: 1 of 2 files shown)


packages/lexical-selection/src/index.ts
```diff
@@ -367,15 +367,14 @@ export function $patchStyleText(
   const firstNodeTextLength = firstNodeText.length;
   const focusOffset = focus.offset;
   let anchorOffset = anchor.offset;
-  let startOffset;
-  let endOffset;
   const isBefore = anchor.isBefore(focus);
-  startOffset = isBefore ? anchorOffset : focusOffset;
-  endOffset = isBefore ? focusOffset : anchorOffset;
+  const [start, end] = isBefore ? [anchor, focus] : [focus, anchor];
```
trueadm, line 371: Do we need this change? It adds more logical runtime overhead compared to before.

packages/lexical-selection/src/index.ts
```diff
@@ -425,6 +434,14 @@ export function $patchStyleText(
       const lastNodeText = lastNode.getTextContent();
       const lastNodeTextLength = lastNodeText.length;
 
+      // The last node might not actually be the end node
+      //
+      // If not, assume the last node is fully-selected unless the end offset is
+      // zero.
+      if (lastNode.getKey() !== end.key && endOffset) {
```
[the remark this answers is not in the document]
birtles: Fixed.

packages/lexical-selection/src/index.ts
birtles: No, it's not needed. Removed.

birtles:
I believe the test failures in e2e-mac are unrelated pre-existing flakiness.

acywatson:
> I believe the test failures in e2e-mac are unrelated pre-existing flakiness.

Yea I think you're right - re-running tests now. If you can resolve conflicts here too, that'd be awesome. Sorry for the churn.

birtles:
I'll rebase this once #3116 is merged since it will conflict with that.

birtles:
Rebased and ready for review

acywatson (approved):

PR #3137, opened 2022-10-07
Fix insertNodes when replacing content adjacent to an inline element

### LINKED ISSUE #3136: Bug: Calling `insertNodes` when a text node is selected, followed by an inline element, causes the nodes to be inserted inside or after the element
Lexical version: 0.5.0

## Steps To Reproduce

1. Create a paragraph with a text node, followed by an inline element, e.g. a link element
2. Fully-select the text node
3. Call `insertNodes` providing some content to replace the text node with

e.g.

```ts
await editor.update(() => {
  const root = $getRoot();

  root.append(paragraph);

  const text = $createTextNode('Existing text...');
  paragraph.append(text);

  setAnchorPoint({
    key: text.getKey(),
    offset: 0,
    type: 'text',
  });
  setFocusPoint({
    key: text.getKey(),
    offset: 'Existing text...'.length,
    type: 'text',
  });

  if (!$isRangeSelection(selection)) {
    return;
  }
  selection.insertNodes([$createTextNode('foo')]);
});
```

## The current behavior

Given the above example, the following (simplified) HTML will be produced:

```html
<p><a href=\"https://"><span>link</span></a><span>foo</span></p>
```

Notice that the inserted text has been inserted _after_ the sibling inline element, despite the fact that we selected text _before_ the inline element.

## The expected behavior

The inserted nodes should replace the selected content. That is, they should be inserted before the sibling inline element, i.e.

```html
<p><span>foo</span><a href=\"https://"><span>link</span></a></p>
```

## Analysis

This appears to occur because at the start of `RangeSelection.insertNodes` we collapse the selection by removing its content:

https://github.com/facebook/lexical/blob/06e347070ade5ca07ac7d27ef45a544d4f8efb2a/packages/lexical/src/LexicalSelection.ts#L1199-L1201

As a result of this, the selection will be moved to the remaining child, specifically it will be put on the link _element_ at offset 0.

Following that, we try to resolve the target node to insert content _after_ with the following steps:

https://github.com/facebook/lexical/blob/06e347070ade5ca07ac7d27ef45a544d4f8efb2a/packages/lexical/src/LexicalSelection.ts#L1208-L1218

This makes sense for a paragraph that has since been emptied, but in our case `placementNode` will be `null` and we will end up setting the `target` to the inline element (i.e. the link in this example).

From here, the behavior depends on the type of content we are trying to insert:

- For a text node, for example, we'll end up inserting it _after_ the inline element: https://github.com/facebook/lexical/blob/06e347070ade5ca07ac7d27ef45a544d4f8efb2a/packages/lexical/src/LexicalSelection.ts#L1390-L1391
- For a block element, we'll end up trying to merge its contents and insert its content inside the inline element

birtles:
Fixes #3136.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1197,7 +1197,37 @@ export class RangeSelection implements BaseSelection {
   insertNodes(nodes: Array<LexicalNode>, selectStart?: boolean): boolean {
     // If there is a range selected remove the text in it
     if (!this.isCollapsed()) {
+      const selectionEnd = this.isBackward() ? this.anchor : this.focus;
+
+      const nextSibling = selectionEnd.getNode().getNextSibling();
+      const nextSiblingKey = nextSibling ? nextSibling.getKey() : null;
+
+      const prevSibling = selectionEnd.getNode().getPreviousSibling();
+      const prevSiblingKey = prevSibling ? prevSibling.getKey() : null;
+
       this.removeText();
+
+      // If the selection has been moved to an adjacent inline element, create
+      // a temporary text node that we can insert the nodes after.
+      if (this.isCollapsed() && this.focus.type === 'element') {
+        let textNode;
+
+        if (this.focus.key === nextSiblingKey && this.focus.offset === 0) {
+          textNode = $createTextNode();
+          this.focus.getNode().insertBefore(textNode);
+        } else if (
+          this.focus.key === prevSiblingKey &&
+          this.focus.offset === this.focus.getNode().getChildrenSize()
+        ) {
+          textNode = $createTextNode();
+          this.focus.getNode().insertAfter(textNode);
+        }
+
+        if (textNode) {
+          this.focus.set(textNode.__key, 0, 'text');
+          this.anchor.set(textNode.__key, 0, 'text');
+        }
+      }
     }
     const anchor = this.anchor;
     const anchorOffset = anchor.offset;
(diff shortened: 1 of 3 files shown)


birtles:
Vercel gives me a 404 when I try to view the playground failure so I'm not sure what's going on there

trueadm:
Please could you add an e2e test for this given the core nature of insert nodes? Thank you

birtles:
> Please could you add an e2e test for this given the core nature of insert nodes? Thank you

Sure, it's the start of a long weekend here but I'll get onto it next week.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1197,7 +1197,27 @@ export class RangeSelection implements BaseSelection {
   insertNodes(nodes: Array<LexicalNode>, selectStart?: boolean): boolean {
     // If there is a range selected remove the text in it
     if (!this.isCollapsed()) {
+      const selectionEnd = this.isBackward() ? this.anchor : this.focus;
+      const nextSiblingKey = selectionEnd.getNode().getNextSibling()?.getKey();
```
acywatson, line 1201: You can't use safe navigation in this repo - this is why the integrity check and vercel build are failing.

birtles:
> Please could you add an e2e test for this given the core nature of insert nodes? Thank you

E2E test added.

birtles:
Actually, I think this same situation might also arise when removing content from the end of a block.

birtles:
(Test failures appear to be pre-existing flakiness that I can reproduce without this patch applied.)

birtles:
> Actually, I think this same situation might also arise when removing content from the end of a block.

Updated the PR to handle the end of block case too.

fantactuka (approved):

PR #3140, opened 2022-10-07
feat: add icons to extension and change build

im-adithya:
Adds icons to the Lexical Devtools Extension and also modifies the build to accommodate different browsers.

#2127


packages/lexical-devtools/chrome/build.js
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
+/* eslint-disable header/header */
```
thegreatercurve, line 2: Can we add the licence header in here?

thegreatercurve (approved):
Great work @im-adithya! This looks great!

packages/lexical-devtools/chrome/build.js
im-adithya: Done!

packages/lexical-devtools/chrome/manifest.json
```diff
@@ -0,0 +1,38 @@
...
+    "default_icon": {
+      "16": "icons/16-disabled.png",
+      "32": "icons/32-disabled.png",
+      "48": "icons/48-disabled.png",
+      "128": "icons/128-disabled.png"
+    },
+    "default_popup": "src/popup/index.html"
+  },
+  "devtools_page": "src/devtools/index.html",
+  "content_security_policy": "script-src 'self' 'unsafe-eval'; object-src 'self'",
```
acywatson, line 24: Is all of this actually required? Doesn't seem to be in the previous manifest.

packages/lexical-devtools/package.json
```diff
@@ -5,7 +5,13 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "build:chrome": "cross-env NODE_ENV=production node ./chrome/build",
+    "build:chrome:local": "cross-env NODE_ENV=development node ./chrome/build",
```
acywatson, line 10: maybe use ":dev" instead of ":local" for consistency with the other scripts in the repo.

im-adithya: Done!

packages/lexical-devtools/chrome/manifest.json
im-adithya: Felt useful ([docs](https://developer.chrome.com/docs/apps/contentSecurityPolicy/)). Also saw it in React Devtools so thought of adding it here as well!

PR #3145, opened 2022-10-08
ref(selection): separate @lexical/selection/index into multiple files

tranqhuy46:
Hello, first-time contributor here. I read the code & your comments, and this is my solution, have a review at your convenience, and feel free to request changes (if any). #3060


thegreatercurve:
Thanks for the contribution @tranqhuy46. Great work!

This seems good to merge from me. The file structure is a lot clearer. I've double checked we're not exporting anything new that we don't want to expose, but it all seems good.

Closes #3060

thegreatercurve (approved):

PR #3147, opened 2022-10-09
FF Node/Grid copy-paste support

zurfyx:
Follow up of https://github.com/facebook/lexical/pull/3080. Turns out `setStart/setEnd` does the trick.

https://user-images.githubusercontent.com/193447/194762136-21a82e99-cd45-48b5-9fa0-ea8f407da893.mov

Screen Shot 2022-10-09 at 10 21 32 am

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -41,7 +41,6 @@ import {
   SELECTION_CHANGE_COMMAND,
   SerializedTextNode,
 } from 'lexical';
-import {IS_FIREFOX} from 'shared/environment';
 import invariant from 'shared/invariant';
 
 export function $getHtmlContent(editor: LexicalEditor): string {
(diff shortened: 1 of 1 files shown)


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -548,8 +525,9 @@ export async function copyToClipboard__EXPERIMENTAL(
   element.style.cssText = 'position: fixed; top: -1000px;';
   element.append(document.createTextNode('#'));
   rootElement.append(element);
-  const range = document.createRange();
-  range.selectNodeContents(element);
+  const range = new Range();
```
acywatson, line 528: Do we have an e2e test for this? Would be nice to ensure consistent behavior across browsers.

zurfyx: Don't think there's a good way to E2E this unfortunately, it's limited by Playwright. For reference, this is the Clipboard hack we have on the test right now

https://github.com/facebookexternal/Outline//blob/e4d6b21761c2acb5012721e177a467096aef315c/packages/lexical-playground/__tests__/utils/index.mjs#L300

```
 const copyEvent = new ClipboardEvent('copy');
    Object.defineProperty(copyEvent, 'clipboardData', {
```

acywatson: I was thinking we could try using keyboard shortcuts or something, as a user would? Haven't looked at these tests in a while, so I'll leave it to you.

acywatson (approved):

packages/lexical-clipboard/src/clipboard.ts
zurfyx: The clipboard won't work, that's why the E2E hack

Fonger:
Where can I get this Clipboard Viewer apps for macOS?

PR #3177, opened 2022-10-13
Fix nested editor cut

fantactuka:
Null check got lost when moving to experimental copy helper. It avoids cut/copy event from being handled in non-active editor (e.g., when running cut in nested editor)

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -374,12 +374,15 @@ async function onCutForRichText(
   event: CommandPayloadType<typeof CUT_COMMAND>,
   editor: LexicalEditor,
 ): Promise<void> {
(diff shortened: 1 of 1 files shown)


acywatson (approved):

acywatson:
> Null check got lost when moving to experimental copy helper. It avoids cut/copy event from being handled in non-active editor (e.g., when running cut in nested editor)

Do we not have test coverage here?

zurfyx:
Thank you @fantactuka!

> > Null check got lost when moving to experimental copy helper. It avoids cut/copy event from being handled in non-active editor (e.g., when running cut in nested editor)
> 
> Do we not have test coverage here?

No, we don't, it's very hard to test copy-paste as we cannot mimic the browser in Playwright

Issue #3089, opened 2022-09-30
Feature: Decorator error boundary


fantactuka:
Complex RTE surfaces might have multiple decorator nodes and their internal rendering logic might fail unrelated to editor itself, and it will crash whole editor and user will lose all content. To isolate that change we can wrap each individual decorator into `<ErrorBoundary fallback={...}>` or provide some sort of centralized way of generating error boundaries through Rich/PlainTextPlugins:

```jsx
<RichTextPlugin
  contentEditable={<ContentEditable />}
  placeholder={placeholder}
  onDecorate={(decorator, editor) => (
    <ErrorBoundary fallback={() => <Ooops />} onError={editor._onError}>
      {decorator}
    </ErrorBoundary>
  )}
/>
```

PR #3178, opened 2022-10-13
Built-in Error/SuspenseBoundaries for React DecoratorNodes

zurfyx:
It happens. Your video component crashes; or your component comes with some asynchronous GraphQL behavior that makes the component suspend. Both of them currently need developer awareness as otherwise it can break the whole editor (yes, not having a Suspense layer on a Decorator makes React swallow all the DOM).

We can make this better on our `@lexical/react` framework. We can ship this by default. And that's exactly what this PR fixes.

https://user-images.githubusercontent.com/193447/195731936-a30b2397-9bf8-4f1c-97d7-9893b083ed98.mov

## API

**TypeScript**

```
<RichTextPlugin
  contentEditable={
    <div className="editor-scroller">
      <div className="editor" ref={onRef}>
        <ContentEditable />
      </div>
    </div>
  }
  placeholder={placeholder}
  ErrorBoundary={ErrorBoundary}
/>
export default function ErrorBoundary({children, onError}: Props): JSX.Element {
  return (
    <ReactErrorBoundary
      fallback={
        <span className="ErrorBoundary__container">
          React crashed. Please,{' '}
          <a href="https://github.com/facebook/lexical/issues/new/choose">
            file a task
          </a>
          .
        </span>
      }
      onError={onError}>
      {children}
    </ReactErrorBoundary>
  );
}
```

**Flow and WWW**

```
ErrorBoundary={({children, onError}) => (
  <ErrorBoundary
    fallback={() => <span>Crashed</span>}
    onError={onError}>
    {children}
  </ErrorBoundary>
)}
```

Or just simply (because it doesn't require fallback) -- don't really endorse this

```
ErrorBoundary={ErrorBoundary}
```

## Why custom ErrorBoundary component support but not for Suspense?

ErrorBoundary is a must have. Users shouldn't have to understand about error propagation on nested editor and the private `editor._onError` property.

At the same time, Suspense is a fairly common user-visible state and it's a good practice to customize the glimmer to adjust the UI best. This can be done easily by adding another Suspense layer on the app. However, this one will act as a fallback mechanism in case the user didn't.

## ⚠️ 0.6 breaking change

It is unwise to have ErrorBoundaries with a fallback to `null`. For newcomers, we may want to provide a built-in way to postpone this effort and simplify it. We can iterate on this later. `react-error-boundary` thinks the same way:

> Uncaught Error: react-error-boundary requires either a fallback, fallbackRender, or FallbackComponent prop

Second, we don't want to ship an opinionated implementation of React Error Boundary, just like React never did. For WWW, we have a internal-specific ErrorBoundary that we certainly want to reuse.

---

Kudos to @fantactuka for the ErrorBoundary idea


fantactuka:
Fixes #3089

packages/lexical-playground/package.json
```diff
@@ -31,6 +31,7 @@
     "prettier": "^2.3.2",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-error-boundary": "^3.1.4",
```
fantactuka, line 34: QQ: is this one required, given default value is some sort of cope of react-error-boundary?

fantactuka (approved):

packages/lexical-playground/package.json
zurfyx: Yeah, but it is just for the playground; other people can just their own replacement

acywatson (approved):

PR #3182, opened 2022-10-14
ToC hook + fixes

zurfyx:
1. Fix generation on nodes on updates
2. Added support for non-top-level headings
3. Remove empty headings from ToC
4. Add optional title on ToC

For 2 unfortunately this makes the algorithm less optimal as nodes can be updated in many multiples ways and not necessarily mark HeadingNode as dirty. Added a debounce on long documents to compensate.


packages/lexical-react/src/useLexicalTableOfContents.ts
```diff
@@ -0,0 +1,43 @@
...
+        const currentNodeKeys = [];
+        for (const {node} of $dfs()) {
+          if ($isHeadingNode(node) && node.getTextContentSize() > 0) {
+            currentNodeKeys.push(node.getKey());
+          }
+        }
+        setNodeKeys(currentNodeKeys);
+      });
+    }, DEBOUNCE_MS);
+    return editor.registerUpdateListener(() => {
```
fantactuka, line 34: Why do we need to run dfs? Can this be mutation listener for created/deleted/updated?

zurfyx: Mutation listener won't catch when a child node is moved (i.e. HeadingNode is wrapped by another component like a character limit or table) and you move the parent

fantactuka: Then maybe `$nodesOfType()` as it won't generate new array for all nodes

acywatson:
@zurfyx what's up with this?

Issue #2886, opened 2022-08-24
Inline examples in the documentation 


thegreatercurve:
> 🚨 PLEASE DO NOT WORK ON THIS ISSUE: This work is earmarked for a Meta intern.

Compared to other rich text editors, our documentation lacks inline/iframe examples of various Lexical concepts and plugins.

As we have opened up Lexical to the OSS community, more and more functionality is being added to the [official playground](https://playground.lexical.dev/).

This creates a kitchen sink example which is confusing to navigate for first time and existing users alike. It also blocks many developers who try to navigate the playground's codebase to simply copy and paste example code. 

To solve this issue, my proposal is to break down the playground into several smaller examples, each one highlighting a different piece of functionality, or different plugin.

These example can then be collated into `iframes` in an "Examples" section within our docs.

Some examples of other inline example pages from other RTEs: 
- [TipTap examples page](https://tiptap.dev/examples/default)
- [ProseMirror examples page](https://prosemirror.net/examples/)
- [SlateJS examples page](https://www.slatejs.org/examples/richtext)
- [QuillJS examples page](https://quilljs.com/playground/)
- [TinyMCE examples page](https://www.tiny.cloud/docs/tinymce/6/basic-example/)

This work will also involve changing some of our E2Es which rely exclusively on the playground.

echarles:
> These example can then be collated into iframes in an "Examples" section within our docs.

Probably the quickest to get something. Another way is to implement lexical as a Docusaurus plugin.

thegreatercurve:
Looks like Docusaurus has recently added some nice functionality for [live code blocks](https://docusaurus.io/docs/markdown-features/code-blocks#interactive-code-editor) in a native plugin.

Depending on the complexity of the example I'm not sure if we'd need to also support having multiple tabs for different files, similar to what you get on TipTap's examples page.

echarles:
There are other examples that allow to show live code and also generate a button to open that code in codesanbox

- https://syncedstore.org/docs/react
- https://plate.udecode.io/docs/plugins/alignment

I have looked at a docusaurus plugin that would allow to embed a codesandbox in a docusaurus webpage but have not found any. Maybe it exists somewhere, or it should not be too difficult to implement. This would allow to directly embed the existing codesanbox plain and rich text.

acywatson:
I remember seeing that React as using this on their new website:

https://sandpack.codesandbox.io/

echarles:
[...]

PR #3185, opened 2022-10-15
Add inline plugin example for images 

akmarzhan1:
- [x] added a new subpage for the images example in the playground (under “/images-demo”)
- [x] added a new page and the example on the website (under “Demos”)

Issue [#2886](https://github.com/facebook/lexical/issues/2886)

Might be possible issues with cross-origin access (now I am embedding a page that doesn't exist yet in the playground, but the embed is blocked when trying for the main playground).


akmarzhan1:
I am not sure why the website deployment is failing, but I am guessing that it is because docusaurus v1 doesn't support react-router v6. I had some issues with react-router versions (v5 and v6) and docusaurus initially but they got fixed after I updated docusaurus on my computer (npm installs v1 and not v2). The package.json already had docusaurus v2 so I don't know what to fix.

thegreatercurve:
@akmarzhan1 The reason you're getting the 404 locally is because your `iframe` attribute of `src` is pointing at the production, `http://playground.lexical.dev/image-demo/`, when the new page in this PR hasn't been deployed yet.

My recommendation would be to move all of the changes you've made in `packages/lexical-playground`, and then to add your Image plugin app into `packages/lexical-website/src/pages/examples/image` instead. 

Remove all of the usages of React Router, as Docusaurus has its own router, and then export your `ImagePlayground` as the main component from this page.

Then, in `packages/lexical-website/docs/demos/advanced-examples/images` point your iframe `src` to a relative URL of wherever you host your images demo (e.g. `src="/demos/image"`) and it should appear even in local development.

Something like the below.

image

Does that sort of make sense?

akmarzhan1:
Thanks for the feedback! The 404 error was on purpose because I couldn't find a way to embed the playground without errors. 

Just to clarify, does this mean I should revert the playground to its original state (remove all the changes) and instead add the image plugin app fully to the website?

thegreatercurve:
> Thanks for the feedback @thegreatercurve! The 404 error was on purpose because I couldn't find a way to embed the playground without errors. 
> 
> Just to clarify, does this mean I should revert the playground to its original state (remove all the changes) and instead add the image plugin app fully to the website?

Yeah, that's right. That should be a lot simpler than having to wrestle with the playground as well. Everything will just be handled by the same router

akmarzhan1:
> Yeah, that's right. That should be a lot simpler than having to wrestle with the playground as well. Everything will just be handled by the same router

Makes sense. I will try to do that and will let you know if I have any questions.

akmarzhan1:
@thegreatercurve I created a new PR draft (I figured it would be easier to create a new branch since I needed to revert all the changes to the playground), here is the link: https://github.com/facebook/lexical/pull/3216

PR #3190, opened 2022-10-16
docs: Fixed broken link

SimonProper:
Fixed faulty file extension in the url for the useList example in the Commands docs, ending in a 404 page.
Changed the file extension from .ts to .js.

### DIFF
--- a/packages/lexical-website/docs/concepts/commands.md
+++ b/packages/lexical-website/docs/concepts/commands.md
@@ -55,7 +55,7 @@ const formatBulletList = () => {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


zurfyx (approved):
LGTM, thanks, or alternatively we upgrade the link to point to a newer (.ts) commit

SimonProper:
Happy to help! 

That seems like a fair point, there are more links in that doc that points to older branches. Is there a reason for that?
I could update all the links to point to the current files in the main branch.

zurfyx:
> Happy to help!
> 
> That seems like a fair point, there are more links in that doc that points to older branches. Is there a reason for that? I could update all the links to point to the current files in the main branch.

Not really, they're just outdated. I'll go ahead and merge this for now

PR #3204, opened 2022-10-18
Fix frozen selection

acywatson:
Internal test fails because we freeze Selection in dev and  [this PR] (https://github.com/facebook/lexical/pull/3177/files) causes us to try to get a reference to the frozen one via read callback and then try to modify it in a subsequent update.


thegreatercurve (approved):

fantactuka (approved):
Thanks for fixing it!

PR #3207, opened 2022-10-18
fix: fixed dual text appearance problem and removed outline scrollbar

karam-qaoud:
* The outline had a problem where it rendered headings text twice. This was due to the plugin being called two times - merge conflict side effect or something? 
* Hid the outline scrollbar, and decreased its width so it fits next to the editor.
* Moved the scroll callback numbers to constants for better readability and modularity.

### DIFF
--- a/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx
@@ -54,7 +54,6 @@ function TableOfContentsList({
   }
   function isHeadingAboveViewport(element: HTMLElement): boolean {
     const elementYPosition = element?.getClientRects()[0].y;
-    // console.log(editor, elementYPosition);
     return elementYPosition < MARGIN_ABOVE_EDITOR;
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx
```diff
@@ -44,15 +47,19 @@ function TableOfContentsList({
...
-    return elementYPosition >= 0.26 && elementYPosition <= 9;
+    return (
+      elementYPosition >= MARGIN_ABOVE_EDITOR &&
+      elementYPosition <= MARGIN_ABOVE_EDITOR + HEADING_WIDTH
+    );
   }
   function isHeadingAboveViewport(element: HTMLElement): boolean {
     const elementYPosition = element?.getClientRects()[0].y;
-    return elementYPosition <= 0;
+    // console.log(editor, elementYPosition);
```
zurfyx, line 57: ^

zurfyx (approved):
Thank you!

PR #3209, opened 2022-10-18
feat: Extend support for link target and rel to lexical-react’s AutoLinkPlugin

acusti:
this is a follow-up to #2687 (which resolved issue #2671) that extends support for setting link target and rel to the `AutoLinkPlugin` component via its `matchers` prop. it adds optional `rel` and `target` fields to the `LinkMatcherResult` type and forwards those along to `$createAutoLinkNode`, or uses `linkNode.setRel` / `linkNode.setTarget` in `handleLinkEdit` if they have changed.

i also updated the lexical-react docs plugins page’s `AutoLinkPlugin` section to adopt the default protocol behavior introduced in #2654 and added commented-out example usage of the new LinkMatcher `rel` and `target` fields ([preview](https://lexical-git-fork-acusti-feat-link-node-with-ec78d1-fbopensource.vercel.app/docs/react/plugins#lexicalautolinkplugin)). i wasn’t sure how to properly document it, so if there is a better way (e.g. just a single comment at the beginning or ending of the LinkMatcher object mentioning the optional fields or something else entirely), please let me know.

i didn’t find an open issue for this, though there is a question about it in the original PR thread: https://github.com/facebook/lexical/pull/2687#issuecomment-1207463130. i’d be happy to open one if that would be helpful.

### DIFF
--- a/packages/lexical-link/flow/LexicalLink.js.flow
+++ b/packages/lexical-link/flow/LexicalLink.js.flow
@@ -65,7 +65,13 @@ declare export class AutoLinkNode extends LinkNode {
   static clone(node: AutoLinkNode): AutoLinkNode;
   insertNewAfter(selection: RangeSelection): null | ElementNode;
 }
-declare export function $createAutoLinkNode(url: string): AutoLinkNode;
+declare export function $createAutoLinkNode(
+  url: string,
+  attributes?: {
+    rel?: null | string,
+    target?: null | string,
+  },
+): AutoLinkNode;
 declare export function $isAutoLinkNode(
   node: ?LexicalNode,
 ): boolean %checks(node instanceof AutoLinkNode);
@@ -81,6 +87,5 @@ declare export var TOGGLE_LINK_COMMAND: LexicalCommand<
 >;
 declare export function toggleLink(
   url: null | string,
-  target?: null | string,
-  rel?: null | string,
+  attributes: {target?: null | string, rel?: null | string},
 ): void;
--- a/packages/lexical-link/flow/LexicalLink.js.flow
+++ b/packages/lexical-link/flow/LexicalLink.js.flow
@@ -16,6 +16,10 @@ import type {
 } from 'lexical';
 import {addClassNamesToElement} from '@lexical/utils';
 import {$isElementNode, ElementNode} from 'lexical';
+export type LinkAttributes = {
+  rel?: null | string,
+  target?: null | string,
+};
 declare export class LinkNode extends ElementNode {
   __url: string;
   __target: null | string;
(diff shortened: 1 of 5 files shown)


acywatson (commented):
I was originally not sure about making this part of matcher result, but it makes sense because you might want different values here for different matcher types.

Maybe it makes more sense to add one property to the LinkMatcherResult that's a bag of attributes that can be added to the link:

LinkMatcherResult = {
  ...
  attributes: {
    rel?: ?string
    target?: ?string
  }
}

acusti:
@acywatson that would match the API of `$createAutoLinkNode`; i’m good with that change. should i implement it?

> I was originally not sure about making this part of matcher result, but it makes sense because you might want different values here for different matcher types.

in my case, i am using a lexical editor for user generated content, so for this instance of the editor, i’m planning to set [`target: 'ugc'`](https://developers.google.com/search/blog/2019/09/evolving-nofollow-new-ways-to-identify) on all autolinked links (but not mailto links).

fantactuka:
$createAutoLinkNode accepts `attributes` as second argument, maybe exposing a type like `LinkAttributes` and then returning `attributes` object from matcher with that type (as Acy suggested above) would simplify extending it in future?

acusti:
@fantactuka i added an exported `LinkAttributes` type in [`packages/lexical-link/src/index.ts`](https://github.com/facebook/lexical/pull/3209/files#diff-ad5ea781745f188d648a0a8b2f30e487a79e3c9b0fa9c01a6ff74385f06d17d6R33), and i’m importing it into [`packages/lexical-react/src/LexicalAutoLinkPlugin.ts`](https://github.com/facebook/lexical/pull/3209/files#diff-d62fd402ee470b9082046b4a435f031dd8539650efc92ac2ea645f8ded2a4c57R9) and using it for a new optional `attributes` key in the `LinkMatcherResult` type. does that look like what you were hoping for (cc: @acywatson)?

acywatson (approved):
this looks right to me.

acusti:
i’m happy to rebase this against latest `main` or do anything else that would be useful for making this PR mergeable, just let me know if there’s anything like that i can do. also, i mentioned this when i opened the PR, but just to be thorough: i wanted to add something to the docs to reference the new optional attributes, so i put a commented-out line in the AutoLinkPlugin usage example: https://github.com/facebook/lexical/pull/3209/files#diff-d59fe907ca78808fad44ab494f4014c15040d7e97abd9270edab884936115bbfR136

i didn’t come across any other references to optional attributes like that in lexical-react’s plugins doc, so i just went for the single-line comment, but if there is a different way you all would prefer, please let me know. or if i should just remove that line from the docs and allow it to be something people discover via typescript types or browsing through source code, that’s also fine with me.

PR #3213, opened 2022-10-18
Refine backspace handling in Tables

tylerjbainbridge:
This is a patch. The real solution is to refine [the way selection is modified after hitting backspace](https://github.com/facebook/lexical/blob/573a716ef97218927007943689c420a8749107ce/packages/lexical/src/LexicalSelection.ts#L1818). The focus should not be moved to a node with a different `shadowRoot`, but right now this is how native domSelection works.

Before:

https://user-images.githubusercontent.com/13852400/196534120-69f71430-b0b5-4686-bfac-2e4ed95b04d2.mov

After:

https://user-images.githubusercontent.com/13852400/196534211-dad4ada4-9db1-435d-b9ea-cb3ba433ba33.mov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21912,36 +21912,36 @@
       }
     },
     "packages/lexical": {
-      "version": "0.5.1-next.0",
+      "version": "0.5.0",
       "license": "MIT"
     },
     "packages/lexical-clipboard": {
       "name": "@lexical/clipboard",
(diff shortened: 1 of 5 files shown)


zurfyx (approved):
Makes sense, let's ship the workaround for now and investigate the long-term fix more carefully

Issue #3211, opened 2022-10-18
Feature: keep editor__DEPRECATED property until there's editing history serialization available


kfancy:
Scenario: editing multiple independent pieces of content in the same browser window session.

So far, I can:

1. get a copy of the history in an editor instance
2. get a copy of the editor in an editor instance
3. store both of those separately for re-use
4. "close" content editing by un-mounting the LexicalComposer
5. later in same browser window session, use the history and the editor instance to "rehydrate" an instance of LexicalComposer to continue where I left off editing, complete with history intact

I understand that the history is referenced by the editor instance, and that reference gets severed at serialization.

The **REAL** feature request is to have a way to store and rehydrate editor history for undo/redo for use in different content editing sessions (say, one now and one tomorrow).

Until that is possible, using the `editor__DEPRECATED` property I can at least get halfway there by locally storing the editor and history at least for use in the same session.

I'm unsure of the timeline for actual "DEPRECATION" of that property, but I'd like to request that it is NOT deprecated until history is serializable.

I've tried to a few of the suggestions offered on a variety of GitHub issue discussions and Discord threads, so far this is the only way to get it to work.

etrepum:
There's no need for this feature when you can create LexicalComposerContext directly and specify externalHistoryState to the HistoryPlugin

PR #3214, opened 2022-10-18
Add MultiEditorStore to Lexical/react

abelsj60:
**Issue** 

Editor instances are not easily portable in Lexical/react today. 

This pull request proposes to solve the problem by introducing a MultiEditorStore to the Lexical/react package. The store makes top-level Lexical editor instances instantly portable, crucially, without changing the existing Lexical/react model. 

_Key elements_

```
- New MultiEditorStore.tsx
- New MultiEditorStoreCtx.ts

- Updates to LexicalComposer
- Updates to LexicalComposerContext
- Updates to LexicalNestedComposer
- Updates to LexicalHistoryPlugin

- New documentation
- New unit test file

- Sundry config updates and some ancillary file debugging (e.g., ImageComponent)
```

**Reason**

Many user interfaces require portable top-level Lexical instances.

Portable means that developers can easily move a Lexical editor around a React tree without losing its history or state. Unfortunately, this does not currently seem to be true. I'll illustrate the problem with a brief, practical example. 

Imagine two sets of tabs. One is on the right, the other the left. Say the user wants to move a tab containing a Lexical editor from the right to the left group. This is done by dragging the "tab button" from right to left. The user will expect the editor's state and history — undo/redo — to stay the same before and after the drag-n-drop. 

Rots a ruck. [Component libraries](https://www.radix-ui.com/docs/primitives/components/tabs) often break tabs into sibling groups of divs: 

```
<LeftTabGroup>
  <TabRoot>
    ...
  </TabRoot>
</LeftTabGroup>
<RightTabGroup>
  <TabRoot>
    <TabButton>
      <ButtonList />
    </TabButtons>
    <TabContents>
      <TabContent />
    </TabContents>
  </TabRoot>
</RightTabGroup>
```

In this scenario, React will — among other things — unmount the TabContent component containing Lexical, mount a new one in the new tab group, and instantiate a new LexicalComposer inside it. Philosophically, Lexical expects developers to keep the user's content true by passing Lexical's previous editorState to the new composer as it mounts. 

There are at least two problems with this, though. 

_1. History_

History gets lost because the Lexical HistoryState contains a series of references to editor instances. If the composer that created the top-level instance dies, references to it (and its nested children) will no longer work. 

Saving the editor instance outside the composer via a custom plugin is no refuge because you can't pass an active instance to a composer anymore. This means the new composer can't host the old instance, you must create a new one by passing editorState instead. While the composer could host foreign instances in the past, this functionality is marked deprecated, and probably for good reason. You probably want to encourage people to save a lightweight state — not full instances. 

_2. onChange_

If you decide to use editorState and forget history, you may interfere with Lexical's simple elegance. The reason is that many developers will probably do what I did — use the OnChangePlugin to save editorState to React whenever it fires. While you might pass the instance out of the composer and serialize it on demand, I'd bet a lot of devs — needing to clear tickets quickly — will just rig it through onChange. I did...!

Unfortunately, my concern is that this perverts Lexical's core thesis — "React is not good for contenteditables." 

After all, doesn't triggering setState on every change feel like taking the drawers out of a heavy bureau before moving it — only to stack them back on top before lifting it up? You do all this work to get out of React, then you reactivate it!

I think a MultiEditorStore can address these issues. 

**Analysis**

The MultiEditorStore makes the LexicalComposer — and its siblings — a little smarter about editor creation. 

When installed, the multi-editor store allows Lexical/react to make its own decisions about when to create a new editor instance and when to load an existing one. More important, it does this without changing the core Lexical/react model. 

In other words, the LexicalComposer and the LexicalNestedComposer remain responsible for creating new editors and the HistoryPlugin remains responsible for creating new historyStates. The store's no more than a searchable repository.

You can review this pull request for the store's operational details. The gist is that developers add a "multiEditorStoreKey" to the initialConfig of their LexicalComposer(s). On mount, each composer uses the key to check the store for an active editor instance. If found, it'll be used. If not, the composer will create a new instance as usual and key it into the store. 

The HistoryPlugin works the same way.

(Of course, these lookups only occur if a store's active. The logic's ignored otherwise.)

Ultimately, the goal of the Lexical/react package is to help React developers use Lexical without duplicating core functionality. The multi-editor store does just that. Plugins just can't fully match its seamless capabilities. 

As a result, I think it a good drop-in component for the core Lexical/react library. 

**Conclusion**

The MultiEditorStore offers a timely enhancement for Lexical's portability in a React app. 

I've already patched this into my own project's Lexical code and am happy to contribute it upstream. Please let me know if I can help in some way. I'm happy to do what I can. Lexical's a great project and I'd be glad to further its overall mission.

-j


fantactuka:
Thanks for all the research, James!

I remember one of discussions on Discord, and the issue with history was that it's not transferable (both within the page and across pages, e.g., by storing on server/localstorage). To my understanding this PR addresses within the page case by moving editor instance to multi editor store that would be somewhere higher in a component tree. With that I wonder how much that would be different from lifting <LexicalComposer> + <HistoryPlugin> in components tree to wrap all multi-tabs (or multi-step-wizard) where you need to re-render editor with maintained history stack.

Also regarding `2. onChange`: storing in a ref (vs state) would avoid re-rendering

abelsj60:
> Thanks for all the research, James!
> 
> I remember one of discussions on Discord, and the issue with history was that it's not transferable (both within the page and across pages, e.g., by storing on server/localstorage). To my understanding this PR addresses within the page case by moving editor instance to multi editor store that would be somewhere higher in a component tree. With that I wonder how much that would be different from lifting + in components tree to wrap all multi-tabs (or multi-step-wizard) where you need to re-render editor with maintained history stack.
> 
> Also regarding `2. onChange`: storing in a ref (vs state) would avoid re-rendering

I'm sorry I can't just talk this out with you face to face, so I'll write it all down, best I can. I realize the length is not ideal, but I don't how else to walk you through where I'm coming from. 

1) Yes, this addresses history and editor instances within a page, not between page loads. 

Maybe I could figure out the page load thing, but I can only do one thing at a time, and I had to take a break from my crazy-ambitious-over-my-head-too-much for-one-person-to-build-you're-a-madman app to get this working because I felt I needed it above all. Also, history still doesn't totally work right, and, to my eye, I think it may need an overhaul. 

Moreover, saving historyState between sessions hasn't really been a Lexical goal. I got this drift on my last pull request when Acy postulated a way of doing it. It seemed pretty clear that preserving history is not on the agenda right now, so the on-page portability I focused on seems to fit into the library's current scope. 

Finally, I think one can fudge this. Restoring history during a session, but not between loads, is the sort of very annoying functionality that I have seen before as a user, writer, and consumer of the Internet. Losing it without a reload? No.

2) Efficiency. 

It's funny I brought this up. It's always my last thought. I didn't really bring it up here because I run on Lighthouse. 

I more brought it up as a matter of principal. If you've got an instance from the Lexical library, your instinct is that you should be able to grab it from anywhere and read its state. That's why you made the contenteditable an instance outside React! 

But, you're right, there's code floating around that uses a plugin to lift the composer's editor instance up into your tree for use anywhere. With that, you could grab state without onChange from just about anywhere in React. 

Nevertheless, here's my thought: 

I felt I needed the store because history wasn't portable, even with that plugin. Then, once my store worked, I wondered why I should ever need to code that plugin or anything like it myself, particularly as I don't think it covers historyStates. (We'll get back to this.)

I guess what I'm saying is, you've got such nice bindings for React, and given that this issue of portability keeps coming up, why shouldn't your React library include a drop-in component to cover it?

I mean, to my eye, portability is no less important than your markdown, table of contents, and autoscroll plugins. It's a core, often sought-after type of functionality that can be handled better when integrated with the rest of your library. It polishes some rough edges, and it can make at least some devs more productive, faster. Someone's metrics will like that.

That said, I did my best with this attempt. It wasn't a half-hearted effort, but I'm certainly not Lexical core... 

3) Can't I just make it work by lifting some state/context? 

I don't think so, but I'd like to walk through it. I think you're saying — I could use a plugin to lift the editor instance and another plugin to lift the historyState, and that should make for a happy path because the editor instance will live on? 

(Or are you saying I treat each tab as a nested lexical composer with a full composer above it? I thought of that. I just couldn't do it. It may technically work, but it's very muddy and all I could see was heartache down that path...)

I hope I have that right. I don't think it works. 

Now, I'm either right or very stupid. I often am very stupid. I've learned to live with that. 

But, let's give this a whirl. Maybe we can debug the situation together. Here goes...

We've got the tab structure above, courtesy of Radix — two tab groups, each consisting of a button list and list of content panels. The button and the content panels are siblings, so you can't just drag a button and take the content panel with it. 

Now, let's say we've taught React to create each set of tabs via these string[]s from state (useState).

rightTabGroup: [ '1', '2', '3']
leftTabGroup: ['a', 'b', 'c']

Let's also say there's a Lexical editor in the content panel for `tabId: 3` with a key of `ezbug`.
Finally, we'll say `ezbug` has filled up a historyState's undoStack, meaning it has a bunch of instances keyed `ezbug`.

(I logged all this a few weeks ago and eventually realized what was going on by tracking these keys. Did I get something wrong? I was in tears at the time, so maybe I couldn't see right. Please, do correct me if I'm wrong...)

OK. You drag `tabId: 3` from right to left, cleverly using the HTML 5 drag-n-drop API to update React state:

rightTabGroup: [ '1', '2']
leftTabGroup: ['a', 'b', 'c', '3']

This means that:

In the right group, React unmounts the tab button and the content panel for `tabId: 3`. 
In the left group, React mounts  a brand new tab button and a brand new content panel for `tabId: 3`.

_Case 1: No plugin saves editor instances, a shared history plugin saves shared history stacks_

When we moved `tabId: 3` left, we replicated its editor's content by passing an editorState to the LexicalComposer that's now mounting on the left. We also handed the original shared historyStack to the new composer's HistoryPlugin. 

- Undo and redo both fail because those editors are for an instance with the key of `ezbug`. 
- The new composer, however, has created a new instance with the key of `sosad`.

```
To review, as you pointed out back in the day, this is how undo/redo work: 

      historyStateEntry.editor.setEditorState(
        historyStateEntry.editorState.clone(historyStateEntry.undoSelection),
        {
          tag: 'historic',
        },
      );

They restore off instances, not editorStates. Thus, it's imperative that the instance key matches whatever instance the user's interacting with when hitting ctrl+z. If it doesn't, nothing happens. Don't I know it — that's how this started.
```

_Case 2: A plugin saves editor instances, a shared history plugin saves shared history stacks_

We now have the same situation, but we've added a plugin for editor instances and we're storing them in context, beside the history stacks, way up high in our tree. 

We perform the same drag operation. We again use editorState to restore the Lexical editor's content on the left when the drop is done. (Do you have a question about that? We'll come back to it in a second.)

- Undo and redo both fail because those editors are for an instance with the key of `ezbug`. 
- The new composer, however, has created a new instance with the key of `sosad`.

Still doesn't work. Now for this matter of editorState. Well, these are Lexical's rules of engagement.

You guys deprecated the ability to restore an editor instance in a LexicalComposer. I've taken this seriously. Let's assume, then that the only way to instantiate a composer is with an editorState. If that's the case, there's no way to connect the old histories — the one's keyed `ezbug` — with the new instance on the left — the one that's keyed `sosad`. 

I don't think it matters how high I lift the context — these things will always be apples and oranges. 

Now, your team's concern seems to be that you don't want people saving and restoring instances, but there shouldn't be any philosophical problem with passing them around once mounted in an app. That's Lexical current, local work.  

Thus, the store feels right to me. It follows the model.

Finally, you may say, we'll fix the history thing. Number one, when? Number two, I honestly hate to say this again, but history isn't fully working. It breaks with some frequency, particularly with nested composers and decorator/react nodes. My sense is that history won't be fixed so fast, and as I said earlier, my observation is that it isn't even on the list right now.

Even if you do fix it, I'm back to the beginning. 

A store seems super useful! Maybe you can do all this without it — but it's more work. Shouldn't Lexical consumers be able to port the editor and the history around React in one motion and without effort?

Call the store experimental, like Tanstack does with some React-Query functionality. React-Query is my absolute favorite library of all time, so, I figure that's probably a pretty good practice. TL nailed it. He just did! Kill the store if you have to.

I think it resolves the history problem while an app is active and it makes life easy without interfering in any way. 

But, you guys have to think it worthwhile. I felt I needed it, so I built it, and I tried to do it close to right.

You pays your money, you takes your chances... Thanks for reading... -j

abelsj60:
> Thanks for all the research, James!
> 
> I remember one of discussions on Discord, and the issue with history was that it's not transferable (both within the page and across pages, e.g., by storing on server/localstorage). To my understanding this PR addresses within the page case by moving editor instance to multi editor store that would be somewhere higher in a component tree. With that I wonder how much that would be different from lifting `<LexicalComposer> + <HistoryPlugin>` in components tree to wrap all multi-tabs (or multi-step-wizard) where you need to re-render editor with maintained history stack (see https://codesandbox.io/s/lexical-plain-text-example-forked-rqdl3i?file=/src/Editor.js where `<Tabs>` force-rerenders PlainTextPlugin when switching between tabs due to different `key` prop, you can type, switch between tabs and undo/redo works just fine)
> 
> Also regarding `2. onChange`: storing in a ref (vs state) would avoid re-rendering
> 
> UPD: `<HistoryPlugin>` is not even needed to be lifted in components tree, only LexicalComposer + SharedHistoryContext which would hold root editor state + shared history stack

Also thanks so much for that sandbox! I can do a fuller discussion tomorrow or Thursday, if you want. 

But, generally, I think you're showing a different setup. I suppose my real point is that devs have a wide range of requirements, and I'm proposing a store that can help Lexical cover an even broader range of them.

I hope it's helpful. I certainly appreciate your willingness to think it through with me...!

I couldn't mean that more. -j

fantactuka:
Ok, so correct me if I'm wrong here in your case you can't lift `<LexicalComposer>` because each tab has multiple composers? In that case I wonder whether it's better to keep optional `editor` prop for composer (the one that deprecated now) and let application code to keep editor reference (e.g., like in your case, put it into context or whatever approach is appropriate there).

abelsj60:
> Ok, so correct me if I'm wrong here in your case you can't lift `<LexicalComposer>` because each tab has multiple composers? In that case I wonder whether it's better to keep optional `editor` prop for composer (the one that deprecated now) and let application code to keep editor reference (e.g., like in your case, put it into context or whatever approach is appropriate there).

Hey @fantactuka. 

1. You are exactly correct. I think we're on a similar page now. I can't lift the LexicalComposer out of the tab. 
   - Specifics don't matter, but a tab could have multiple composers. Still, even if a tab only had one, the end result would be the same — not everything you want to do with Lexical will happen inside the LexicalComposer.  
       - I'd wager other devs will confront more situations like this that don't involve tabs. 
2. re: Undeprecating the `{ initialConfig: editorInstance }` property. 
   - If you took the time to deprecate it, I'd imagine you confronted a number of situations that caused you to think it would be problematic over time. There's a sort of inherent beauty in limiting instantiation to editorState. And, as @trueadm said a couple weeks ago, the editorState is very lightweight. 
       -  I keep finding that the code that's best is the code that's simplistic and...lightweight. 
   - Still, here's an issue from kent-place (in Discord) asking for you to undeprecate that property! It's very well said. 
       - https://github.com/facebook/lexical/issues/3211
3. This PR is probably a win if it caused you all to stop a second to reconsider this topic.
      - Feel free to close the Pull Request or ask me to, if you'd rather go in that direction. I'm not sure I would at this point. I've come to think Lexical/react would be better with a drop-in context provider. But, I can understand that an "official" store won't be flexible enough for all people in all situations. That could prove a hassle for you.
      - And, of course, my code may not be the code you want. I can accept that, although, I happily stand by it...

Let me know if there's something else I can do to help.

-j

fantactuka:
One thing to mention here: I don't think there's a need to store nested editors in your context, having root only should be sufficient to move editor instance around tabs

abelsj60:
> One thing to mention here: I don't think there's a need to store nested editors in your context, having root only should be sufficient to move editor instance around tabs

Ahhh! Now we're in the thick of it!!

I'm not storing nested editors! I'm only storing a list of their keys! The reason is "configuration." 

(In fact, this was a bit of a refactor, as I didn't understand this initially. Once I did, I moved to the string[] of keys...)

If you don't store these keys, then you will "re-set up" the nested editor in the NestedLexicalComposer every time it runs on remount. This means always re-configuring everything from the nested editor's collaboration status to any properties attached to it during the composer's initial useMemo function. 

(I named the condition/pre-condition `isRemountable` in both composers to help indicate this. No good?)

As you say, in a remount, the nested editor is fully configured b/c the root editor/root editor in the historyStack has a reference to it, so it still exists.

This means — as I see it — you only have to option: 
- (a) Always allow set up to fully run on remount or 
- (b) Vary how set up is done. It runs on initial mount, then is fully — or partially — skipped on remount.

Sadly, you'll have no control of this if you don't track these lists! (But, total control with them.)**

I choose option (b).

Right now, option (a) may be harmless, but...
- It will probably always be confusing! 
    - Why? It can cause the developer to misunderstand how things work. That's not great.
    - (If you want, I can point to an example of this kind of potential misdirection in the existing LexicalNestedComposer.)
- Over time, a full set up on _remount_ could wipe out some important configuration.  
    - I'm not sure if this is the current case — I tried to address yjs collaboration effectively.

We've got an identical situation with the HistoryPlugin, which is why I again followed the model and tracked all history keys. In a remount, you'll need to know what's active in order to differentiate the configuration of an initial mount and a remount.

I was thinking of refactoring this into a plugin, but then you lose the ability to intelligently control this configuration. 

I _think_ a store like this probably requires integration into all composers for production-grade reliability. No?

--
** I'd love to see an "introspection" class on Lexical editors, kinda akin to a...database (do I have that right)?
```
- You're using editors almost like a database, so it makes some sense, no? If a root editor could be 
introspected, you wouldn't need to track the history keys or nested editor keys to make the store 
work effectively. 

- I saw someone ask for a method to get active nodes on an instance in Discord recently. Same idea!
```

Issue #3157, opened 2022-10-11
Bug: Impossible to set cursor between decorator nodes using mouse


Piliuta:
Setting the cursor between two non-inline decorator nodes is not possible to do.

Lexical version: 0.5.0

## Steps To Reproduce

1. Add two non-inline decorator nodes (e.g. tweet node)
2. Ensure there are no other nodes between them (e.g. removing empty space is not possible due to this issue https://github.com/facebook/lexical/issues/3155, use drag actions to stack the nodes)
3. Attempt to set cursor position anywhere between the nodes using the mouse

https://user-images.githubusercontent.com/1575198/195076123-d7c0ff98-d698-42f8-9376-689f2bb4b1ec.mov

## The current behavior
- Click on the decorator node selects, typing in the "selected" does nothing
- Click before or after the decorator node sets the cursor position after the last decorator node

## The expected behavior
_**Proposal**_ clicks on the left or right side of a non-inline decorator node will set the caret before or after the node accordingly.

trueadm:
The way we tackle this is to ensure you can select the nodes, then upon pressing left/right you can move the cursor to in between. Not all decorators have node selection enabled on them, but horizontal rule nodes do and you can see this behavior there.

Piliuta:
> The way we tackle this is to ensure you can select the nodes, then upon pressing left/right you can move the cursor to in between. Not all decorators have node selection enabled on them, but horizontal rule nodes do and you can see this behavior there.

Such behavior is not practical for a few reasons:
- It is not intuitive and not very common. You have to be really a pro user to know it is possible.
- _As you already mentioned_, not all decorators have node selection and thus, setting the caret between such nodes is not possible.
- Node selection can be implemented differently if their content is supposed to act differently to click events and selection is triggered say with some sort of a handler button, in which case that flow to set the caret becomes even more confusing.
- After all, most of the powerful editors allow clicking near _nodes_ to set the caret before or after, so missing such a feature does feel great.

Is this something that can be considered added to the lexical as a core behavior? Do you have any plans for it?

trueadm:
I think this should still be handled as part of the decorator as they’re meant to own and control selection - not the core. So if you wanted localized click handling you could add a click handled to the editor element that tracked +/- 10px of edge and moved range selection upon the click. This could be packages up as a React hook or something so nodes can use them as a one liner

zurfyx:
Closing as emulated cursors should address this.

Issue #2495, opened 2022-06-22
Bug: Inserting into the first line might cause scroll


nextend:
Lexical version: 0.3.5

## Steps To Reproduce

1. Place the caret into the first line
2. Use selection.insertNodes([...], false);

## The current behavior
In special cases scrollIntoViewIfNeeded() gets triggered. 

## The expected behavior
It should not scroll

https://www.youtube.com/watch?v=JXcazlsW-YM

This is the console.log in `scrollIntoViewIfNeeded()` which you can see in my video:
```
if (Math.floor(rect.top) < Math.floor(rootRect.top)) {
    element.scrollIntoView();
    console.log(element, rootElement, rect.top, rootRect.top);
}
```

Maybe here should be a check if rootElement is scrollable and only trigger this statement if it is.

ttillberg:
I'm experiencing something similar because my elements are styled in such way the very bottom intersects the `rootRect` by 4px. I was able to fix it by moving around the styles but agree this behaviour should ultimately only be triggered if the editor is scrollable. and/or if it's intended.  This has been a "mystery bug" in our source for some time (it only occurs when hitting DEL key in a scrollable window) and it took a while to track it down to [that very line of code](https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalUtils.ts#L1166) and I occurs to me other people will struggle with the same issue at some point.

Thoughts:

- Should `rootElement` be the first scrollable parent above instead of the `contentEditable` ? 
- Doesn't `LexicalAutoScrollPlugin.js` make this code redundant ? 
- I can't see how this code has any use at all since `root` is the `contentEditable` and is not meant to be scrollable (as per `Playground` example)

rubenmoya:
The scroll in the playground is funky too.

Once there is scroll in the editor it starts to scroll down to the bottom on every new line. Then when you delete a line it moves up slowly.

Any workarounds to solve this?

![lexical](https://user-images.githubusercontent.com/905225/196193935-cf3441a0-525d-40c8-932d-6db2c53ca059.gif)

thegreatercurve:
This should have been fixed with #3378. I can confirm that the last GIF is no longer replicable. 

<img width="1038" alt="image" src="https://user-images.githubusercontent.com/7748470/203809080-7ff51360-4cc2-48bf-9b32-97b4275d6ffe.png">

rubenmoya:
I can confirm it is fixed, thank you so much @thegreatercurve @trueadm! 👏

[...]

PR #3220, opened 2022-10-19
Allow skipping auto-scroll action on selection change

Piliuta:
## Problem

Lexical does not provide a way to control whether it should auto-scroll or not when the selection changes.

**Context:**

I am working on the implementation of a "click to set cursor near decorator nodes" feature https://github.com/facebook/lexical/issues/3157. I am using lexical selection to detect where the custom caret is supposed to be. Since lexical (which matches with the native APIs) does not allow distinguish if the caret is set before or after an element, I chose to treat such selection as "after the previous node" (as opposed to "before the current/offset node") because input changes from this state create a new block between the nodes, so it feels natural to visually indicate the caret at the end of "the previous node".

```
node1[caret]
node2
```

_any input changes format the content as_

```
node1
[whatever is typed or pasted from being in the previous state]
node2
```

Such positioning of the fake caret leads to misleading auto-scroll, e.g. if the click action is called near the `node1` it auto-scrolls to `node2` which is confusing because the fake caret is set near the `node1`.

**Other issues:**

Possibly can be used to fix other issues https://github.com/facebook/lexical/issues/2495

## Solution
Add a new `skip-scroll-into-view` tag to be able to skip auto-scroll behavior while selection updates.

```js
editor.update(() => {
  // selection updates that are not supposed to auto-scroll
}, { tag: 'skip-scroll-into-view' });
```

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -2742,6 +2742,7 @@ export function updateDOMSelection(
     );
 
     if (
+      !tags.has('skip-scroll-into-view') &&
       nextSelection.isCollapsed() &&
       rootElement !== null &&
       rootElement === activeElement


acywatson:
No objections here, but we should probably document this API and organize these tags somehow. If we keep relying on this API as way for users to configure core behavior, we need to be more explicit.

No need to change this PR, IMO - just a note for the team.

Issue #3152, opened 2022-10-10
Revisit generics


zurfyx:
Possible improvement to Flow typing of .getChildren()?

[...]

PR #3226, opened 2022-10-19
Handle bad list item children gracefully

zurfyx:
Generics are a lie https://github.com/facebook/lexical/issues/3152#issuecomment-1273576531

Turns out there is a way to make this fail.. we need to investigate the root of the problem. Fixing the existing crash scenario

---

See https://github.com/facebook/lexical/issues/3227

### DIFF
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -283,10 +283,6 @@ export function updateChildrenListItemValue(
         if (prevValue !== nextValue) {
           child.setValue(nextValue);
         }
-      } else {
-        if (__DEV__) {
-          invariant(false, 'Expected ListNode child to be a ListItemNode');
-        }
       }
     }
   }
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -270,16 +270,26 @@ export function removeList(editor: LexicalEditor): void {
 
 export function updateChildrenListItemValue(
   list: ListNode,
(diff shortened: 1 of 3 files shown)


packages/lexical-list/src/formatList.ts
```diff
@@ -270,16 +270,26 @@ export function removeList(editor: LexicalEditor): void {
...
+        const prevValue = child.getValue();
+        const nextValue = $getListItemValue(child);
+
+        if (prevValue !== nextValue) {
+          child.setValue(nextValue);
+        }
+      } else {
+        if (__DEV__) {
+          invariant(false, 'Expected ListNode child to be a ListItemNode');
+        }
```
zurfyx, line 289: @fantactuka I wonder what you think about this. While I'm all in into enforcing document consistency, at this point it's already too late. The harm has already been done, the best we can do is to handle it gracefully.

On the contrary, when caught right on the update operation that's causing this the editor can revert the update before any commit has triggered

zurfyx:
OK, I think this is just a "fix" to the actual problem but we should prioritize this https://github.com/facebook/lexical/pull/3226 which didn't stop this from happening, as well of the actual problem which is still unkown.

packages/lexical-list/src/formatList.ts
acywatson: We need to change how list items are nested and we can get rid of all of this anyway

zurfyx: Deleted invariant

fantactuka (approved):

PR #3230, opened 2022-10-20
Show there is a bug in ElementNode.hasFormat

wilbertom:
Opening a PR to show how the tests fail after adding this commit.

### DIFF
--- a/packages/lexical/src/nodes/LexicalElementNode.ts
+++ b/packages/lexical/src/nodes/LexicalElementNode.ts
@@ -241,9 +241,10 @@ export class ElementNode extends LexicalNode {
   hasFormat(type: ElementFormatType): boolean {
     if (type !== '') {
       const formatFlag = ELEMENT_TYPE_TO_FORMAT[type];
-      return (this.getFormat() & formatFlag) !== 0;
+      return this.getFormat() === formatFlag;
     }
-    return false;
+
+    return this.getFormat() === 0;
   }
 
   // Mutators
--- a/packages/lexical/src/nodes/LexicalElementNode.ts
+++ b/packages/lexical/src/nodes/LexicalElementNode.ts
@@ -240,8 +240,8 @@ export class ElementNode extends LexicalNode {
   }
   hasFormat(type: ElementFormatType): boolean {
     if (type !== '') {
-      const formatFlag = ELEMENT_TYPE_TO_FORMAT[type];
-      return (this.getFormat() & formatFlag) !== 0;
+      const formatFlag = 1 << ELEMENT_TYPE_TO_FORMAT[type];
+      return ((1 << this.getFormat()) & formatFlag) !== 0;
     }
     return false;
   }
--- a/packages/lexical/src/nodes/LexicalElementNode.ts
+++ b/packages/lexical/src/nodes/LexicalElementNode.ts
@@ -241,10 +241,9 @@ export class ElementNode extends LexicalNode {
   hasFormat(type: ElementFormatType): boolean {
     if (type !== '') {
       const formatFlag = ELEMENT_TYPE_TO_FORMAT[type];
-      return this.getFormat() === formatFlag;
+      return (this.getFormat() & formatFlag) !== 0;
     }
-
-    return this.getFormat() === 0;
+    return false;
   }
 
   // Mutators
--- a/packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
+++ b/packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
@@ -118,7 +118,7 @@ describe('LexicalElementNode tests', () => {
         expect(node.hasFormat('center')).toEqual(false);
         expect(node.hasFormat('right')).toEqual(false);
         expect(node.hasFormat('justify')).toEqual(false);
-        expect(node.hasFormat('')).toEqual(true);
+        expect(node.hasFormat('')).toEqual(false);
       });
     });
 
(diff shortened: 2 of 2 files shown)


wilbertom:
Ran into this bug while implementing an `active` indicator on toolbar buttons similar to how Playground does.

Notice how multiple buttons end up active because `hasFormat` is returning true.

Screenshot 2022-10-20 at 01-23-11 Screenshot
Screenshot 2022-10-20 at 01-22-57 Screenshot
Screenshot 2022-10-20 at 01-22-29 Screenshot


I could also be misusing the API :man_shrugging: . Here's my code:

```ts
class SomeClass {
  active(editor: LexicalEditor): boolean {

    if (!$isRangeSelection(selection) && !$isNodeSelection(selection)) {
      return false;
    }

    return this.allNodesHaveFormat(selection.getNodes(), this.format);
  }

  allNodesHaveFormat(nodes: LexicalNode[], format: ElementFormatType): boolean {
    return all(nodes, node => this.nodeHasFormat(node, format));
  }

  nodeHasFormat(node: LexicalNode, format: ElementFormatType): boolean {
    const element = $getNearestBlockElementAncestorOrThrow(node);
    return element.hasFormat(format);
  }
}
```

`SomeClass.active` does get called in a context where the `$` functions should work.

```ts
  React.useEffect(() => {
    return mergeRegister(
      editor.registerUpdateListener(({editorState}) => {
        editorState.read(() => {
          setActive(someClassInstance.active(editor))
        });
      }),
    )
  }, [editor]);

```

trueadm:
This doesn't seem right. We use bitwise as you can have multiple formats active at once.

zurfyx:
Oh, the lovely `format` function.. it always causes confusion even within the team. This method should be called `align` instead (it was inherited from TextNode but works in different ways). Unfortunately, we can't fully rename it because that would imply some annoying JSON serialization changes we don't want to push forward.

trueadm:
> @trueadm what is ElementFormatType `justify`? Is the idea that a node could be aligned `left`, `center` or `right`, and justified at the same time?
> 
> Or is it that `ElementNode.__format` is encoding more than just the `ElementFormatType`?
> 
> Sorry for the false alarm.

The idea was that by being bitwise, we might want to add additional format types in the future. Right now it's kind of exclusive to alignment, but that was more of an implementation detail that has existed since forever.

wilbertom:
Thank you both for the context. I can think of two possible ways of fixing it then.

## Change bitwise values

If I understand correctly. The issue is that `1`, `2`, `3`, and `4` don't work as bitwise values.

For example, checking if a node with align left(`1`) has align right(`3`) returns true because their values `&`ed return non 0.

```python
1 == 0b01
3 == 0b11
```
```python

  0b01
& 0b11
-------
  0b01
```

Should we change it to:

```ts
-export const IS_ALIGN_LEFT = 1;
-export const IS_ALIGN_CENTER = 2;
-export const IS_ALIGN_RIGHT = 3;
-export const IS_ALIGN_JUSTIFY = 4;
+export const IS_ALIGN_LEFT = 1 << 1;
+export const IS_ALIGN_CENTER = 1 << 2;
+export const IS_ALIGN_RIGHT = 1 << 3;
+export const IS_ALIGN_JUSTIFY = 1 << 4;
```
Would this break serialized/deserialized `EditorState`?

## Add ElementNode.hasAlignment

Or can we add a `ElementNode.hasAlignment(t: ElementAlignmentType): boolean` that only has to worry about alignment and implement it how I implemented `hasFormat` on 7b3a5d5f4de12c96def68db5287cfbdbbc31678d ?

```ts
type ElementAlignmentType = 'left' | 'center' | 'right' | 'justify' | '';

type ElementFormatType = ElementAlignmentType; // don't break the existing type
```

Later, if some other formatting options make their way here, it can be changed to:

```ts
type ElementFormatType = ElementAlignmentType | SomethingElse
```

wilbertom:
Or:

```ts
diff --git a/packages/lexical/src/nodes/LexicalElementNode.ts b/packages/lexical/src/nodes/LexicalElementNode.ts
index 402dbcbe..377dc79f 100644
--- a/packages/lexical/src/nodes/LexicalElementNode.ts
+++ b/packages/lexical/src/nodes/LexicalElementNode.ts
@@ -240,8 +240,8 @@ export class ElementNode extends LexicalNode {
   }
   hasFormat(type: ElementFormatType): boolean {
     if (type !== '') {
-      const formatFlag = ELEMENT_TYPE_TO_FORMAT[type];
-      return (this.getFormat() & formatFlag) !== 0;
+      const formatFlag = 1 << ELEMENT_TYPE_TO_FORMAT[type];
+      return ((1 << this.getFormat()) & formatFlag) !== 0;
     }
     return false;
   }
```

I'm going to guess this is the correct solution.

trueadm:
```
-export const IS_ALIGN_LEFT = 1;
-export const IS_ALIGN_CENTER = 2;
-export const IS_ALIGN_RIGHT = 3;
-export const IS_ALIGN_JUSTIFY = 4;
+export const IS_ALIGN_LEFT = 1 << 1;
+export const IS_ALIGN_CENTER = 1 << 2;
+export const IS_ALIGN_RIGHT = 1 << 3;
+export const IS_ALIGN_JUSTIFY = 1 << 4;
```

Ah this seems like a mistake on my side of things. Given this, your original PR makes much more sense, as this isn't using bitwise (as it was intended to). As you allude to, changing this is a bigger breaking change. I think we should stick with your original. Sorry for the churn.

thegreatercurve:
Closing this due to becoming stale.

PR #3231, opened 2022-10-20
Fix exportJSON return types for ParagraphNode and LineBreakNode

mattreid1:
The `exportJSON()` types for `ParagraphNode` and `LineBreakNode` did not match the `importJSON()` types for those nodes. This PR makes the `exportJSON()` return type consistent with the expected type for `importJSON()` - the same way as it is for all other nodes.

The Flow file was also updated with the types, but this caused an error as the type `type` in `SerializedLexicalNode` is a string, but the `SerializedParagraphNode` and `SerializedLineBreakNode` types use the literal strings `'paragraph'` and `'linebreak'` respectively and so `$FlowExpectedError` was added with a justification.

### DIFF
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -626,7 +626,8 @@ declare export class LineBreakNode extends LexicalNode {
   static importJSON(
     serializedLineBreakNode: SerializedLineBreakNode,
   ): LineBreakNode;
-  exportJSON(): SerializedLexicalNode;
+  // $FlowExpectedError[incompatible-extend] 'linebreak' is a literal string
+  exportJSON(): SerializedLineBreakNode;
 }
 declare export function $createLineBreakNode(): LineBreakNode;
 declare export function $isLineBreakNode(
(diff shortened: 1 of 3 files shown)


fantactuka:
Would you be interested to update types for rich-text nodes too (code, quote, heading, list)? Seems like there are a few classes with `exportJSON(): SerializedElementNode`

acywatson:
> Would you be interested to update types for rich-text nodes too (code, quote, heading, list)? Seems like there are a few classes with `exportJSON(): SerializedElementNode`

This was originally done to prevent forcing every node subclasses to add a FlowFixMe or ts-ignore on these methods.

mattreid1:
> > Would you be interested to update types for rich-text nodes too (code, quote, heading, list)? Seems like there are a few classes with `exportJSON(): SerializedElementNode`
> 
> This was originally done to prevent forcing every node subclasses to add a FlowFixMe or ts-ignore on these methods.

I wouldn't mind adding a `FlowExpectedError` or `FlowFixMe` to the subclasses if you like? TypeScript didn't seem to complain when I tested it.

I think doing so would make handling serialized nodes easier. Adding in the `FlowExpectedError` is a one-off job, whereas developers will be continually using these methods and serialized nodes themselves. Although I do admit I'm not terribly familiar with Flow itself and if doing so may cause other side effects

acywatson:
> I think doing so would make handling serialized nodes easier. Adding in the `FlowExpectedError` is a one-off job,

You might be right about this - I don't remember testing with ExpectError here. Let me check it

acywatson:
> I think doing so would make handling serialized nodes easier. Adding in the `FlowExpectedError` is a one-off job,

You might be right about this - I don't remember testing with ExpectError here. Let me check it

trueadm (approved):

acywatson:
@trueadm once again, this was done for a reason. I think TS errors will happen to anyone subclassing ParagraphNode and overriding this method. Would have appreciated a check in before merging.

trueadm:
@acywatson my bad, I jumped the gun on this one through triage. Feel free to revert.

Issue #3176, opened 2022-10-13
Make TreeView configurable


zurfyx:
Should be able to hide/show its features and make it more customizable. 
- node tree
-  selection
-  editor
-  time travel

juanjo75es:
I added a push request with an improvement to make TreeView more configurable as requested in this issue.
https://github.com/facebook/lexical/pull/3232
Please, let me know if you find that something should be improved. I have a large experience as Software Engineer but I'd just like to get some experience in open-source and React JS projects to prove I can contribute to this kind of project.

fantactuka:
Should it all be in dev tools instead or do we plan to expand tree view in parallel?

PR #3232, opened 2022-10-20
Added TreeView button to toggle visibility

juanjo75es:
As suggested in https://github.com/facebook/lexical/issues/3176
lexical-capture

My goal is to prove I am able to contribute to React JS projects, so feel free to comment me ways to improve my code as if I had a responsibility in this project.

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -1255,6 +1255,7 @@ button.action-button:disabled {
   position: absolute;
   background: none;
   color: #fff;
+  z-index: 4;
 }
 
 .debug-timetravel-button:hover {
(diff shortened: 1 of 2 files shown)


juanjo75es:
I have cloned the main branch and compiled it without modifications and some tests have also to be canceled as they keep retrying forever, exactly as in my branch.  No idea what is going on...
I am also trying to write a new test and it fails for the same reason.

juanjo75es:
I was running the server on the wrong port and that's why my tests failed. 
Now I see that there are 7 tests failing in chromium. Four are shared with the main branch and three of them are new ones. Of these three, at least one is failing because it is depending on the previous implementation of TreeView, and one is failing because the output has a different HTML format but reproducing it manually seems to work as expected and it doesn't fail in Firefox. Indeed, in all three cases reproducing the tests manually seem to perform as expected.

For example, this line requires a pre element in the output but the new implementation does not use pre.
`.querySelector(`${_parentSelector} > .tree-view-output pre`)`

acywatson:
Not sure what the status is here - closing to clean up but reopen if you want to take another crack at it.

PR #3233, opened 2022-10-20
chore: add new `eslint-plugin` with rule `no-optional-chaining` locally

### LINKED ISSUE #3074: Add an eslint rule to prevent optional chaining
Add an eslint rule to prevent optional chaining. We don't support it and it will fail the build. Can probably use this, but I don't think we want to depend on this package, necessarily (hasn't been updated in a while and seems to require some additional config that we might not want).

https://github.com/mysticatea/eslint-plugin-es/blob/master/lib/rules/no-optional-chaining.js

shanpriyan:
Resolves #3074

1. Created a new `eslint-plugin` inside the project directory and added a new rule (`no-optional-chaining`) in it.
2. Added the plugin in `.eslintrc.js` to detect the usage of optional chaining
3. Added the locally created `eslint-plugin-lexical` to _package.json_

Rule reference: https://github.com/mysticatea/eslint-plugin-es/blob/master/lib/rules/no-optional-chaining.js

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -84,7 +84,11 @@ module.exports = {
       },
     },
     {
-      files: ['packages/**/src/__tests__/**', 'packages/lexical-playground/**'],
+      files: [
+        'packages/**/src/__tests__/**',
+        'packages/lexical-playground/**',
+        'packages/lexical-devtools/**',
+      ],
       rules: {
         'lexical/no-optional-chaining': OFF,
       },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,7 +15,12 @@ const ERROR = 2;
 
 module.exports = {
   // Prettier must be last so it can override other configs (https://github.com/prettier/eslint-config-prettier#installation)
-  extends: ['fbjs', 'plugin:react-hooks/recommended', 'prettier'],
+  extends: [
+    'fbjs',
+    'plugin:react-hooks/recommended',
+    'plugin:lexical/all',
+    'prettier',
+  ],
 
   globals: {
     JSX: true,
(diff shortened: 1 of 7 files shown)


shanpriyan:
Run `yarn lint` or `npm run lint` to lint all files

image
image

acywatson:
Pretty sweet! Let's ignore test files and the playground, I guess

shanpriyan:
> Looks like we also need to ignore lexical-devtools

Done 🚀

abelsj60:
@shanpriyan Hi. I'm having some kind of problem with this eslint rule today. 

I've just synced my Lexical fork so I can update my local copy. However, I can't complete the commit with all the changes because ESLint is causing Husky to die on this plugin. 

Specifically: 

1. Husky reports:
```
ESLint couldn't find the plugin "eslint-plugin-lexical".
```
2. Then ESLint says:
```
ESLint: 7.32.0

ESLint couldn't find the plugin "eslint-plugin-lexical".

(The package "eslint-plugin-lexical" was not found when loaded as a Node module from the directory "/Users/james/Desktop/code/lexical".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install eslint-plugin-lexical@latest --save-dev
...
```
3. Finally, when taking ESLint's advice to do a `npm install ...`, I get the following error:
```
npm ERR! Cannot convert undefined or null to object
```

Do you have any advice on how to fix this? Is it possible that you added a file system reference to your "npm registry list", a la https://stackoverflow.com/a/41200925 ? If so, what would I need to add?

I can't even seem to remove the rule, as each time I do, Husky seems to "magically" reinsert it. 

I'm definitely missing something here. Any thoughts that could help?

abelsj60:
> @shanpriyan Hi. I'm having some kind of problem with this eslint rule today.
> 
...

Ha! I asked for help in the Eslint Discord and the recommendation seems to have worked right quick:

`npm link ./eslint-plugin`

Looks like this solves the problem. I'm cautiously optimistic!

PR #3234, opened 2022-10-20
Various fixes in $wrapNodes

EgonBolton:
- rename the $wrapNodes function to $setBlocksType. I know that $wrapLeafNodesInElements was recently renamed to $wrapNodes. But nothing is actually being wrapped. Only certain nodes are replaced or converted to others.
- wrapNodes/setBlockTypes greatly simplified: went from ~300 lines of code to ~80
- when applied on an elementNode, the replace method now also adds its descendants to the replace node.
- fixed bug in wrapNodes/setBlocktype: when the selection started or ended with empty elements. (I added a test too).


packages/lexical/src/LexicalNode.ts
```diff
@@ -639,6 +641,7 @@ export class LexicalNode {
 
   replace<N extends LexicalNode>(replaceWith: N): N {
     errorOnReadOnly();
+    const selection = $getSelection()?.clone();
```
acywatson, line 644: Take these safe nav operators out to get the integrity check passing. There's a PR for a lint rule for this.

EgonBolton: done. And what about e2e tests? I get an error installing playwright. @acywatson

packages/lexical-selection/src/range-selection.ts
```diff
@@ -7,304 +7,81 @@
...
-  LexicalNode,
-  NodeKey,
-  Point,
-  RangeSelection,
-  TextNode,
-} from 'lexical';
 
+import {$isCodeNode} from '@lexical/code';
+import {$isListItemNode} from '@lexical/list';
+import {$isHeadingNode, $isQuoteNode} from '@lexical/rich-text';
```
fantactuka, line 13: We should not have dependency on rich-text package within core (lexical, lexical-selection)

packages/lexical/src/LexicalNode.ts
```diff
@@ -654,8 +658,13 @@ export class LexicalNode {
     writableReplaceWith.__parent = newParent.__key;
     removeNode(this, false);
     internalMarkSiblingsAsDirty(writableReplaceWith);
-    const selection = $getSelection();
+    if ($isElementNode(this)) {
+      this.getChildren().forEach((child: LexicalNode) => {
```
trueadm, line 662: Not sure what this is for?

EgonBolton: Replace would copy the children to the new node. Before, the function was called and then doing that additional work externally.

packages/lexical-selection/src/range-selection.ts
EgonBolton: Makes sense. I think the most reasonable thing would be to move $setBlocksType to Rich-text instead of putting it in lexical-selection, don't you think so?

fantactuka: Fair point! But that said, these dependencies should not be added either way, e.g., rich text should not depend on code and/or lists. This hardcoded check would also mean that no other element node will be able to use this method

packages/lexical-selection/src/range-selection.ts
```diff
@@ -7,304 +7,81 @@
...
-  }
+function isBlock(node: LexicalNode) {
+  return (
+    $isElementNode(node) &&
+    ($isParagraphNode(node) ||
+      $isHeadingNode(node) ||
+      $isListItemNode(node) ||
+      $isQuoteNode(node) ||
+      $isCodeNode(node))
+  );
```
Piliuta, line 84: Is the purpose of this function to check if the node can be inline or not? Can't we simply use `isInline` method?

```suggestion
  return !node.isInline();
```

_If this works, we do not even need to define this as a separate function._

packages/lexical-selection/README.md
```diff
@@ -95,15 +95,14 @@ Expands the current Selection to cover all of the content in the editor.
...
 
-Attempts to wrap all nodes in the Selection in ElementNodes returned from createElement. If wrappingElement is provided, all of the wrapped leaves are appended to the wrappingElement. It attempts to append the resulting sub-tree to the nearest safe insertion target.
+Converts all nodes in the selection that are of one block type to another specified by parameter
 
 ```ts
-export function $wrapNodes(
+export function $setBlocksType(
   selection: RangeSelection,
-  wrappingElement?: ElementNode,
```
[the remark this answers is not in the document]
EgonBolton: > This may be just me though.

Exact. That parameter is not used throughout the repository.

And for me, putting it there would violate the principle of single responsibility.
The purpose of this function is not to wrap (hence why $wrap was a bad name). What the function basically does is replace nodes.

packages/lexical-selection/src/range-selection.ts
Piliuta: _Without getting deep into it._ The reason I proposed to use `isInline` is that `isBlock` function made me think that the logic uses it to attempt to skip steps when something is inline. It might be better to rename it to something different if that's not the case.

packages/lexical-selection/README.md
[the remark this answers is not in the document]
Piliuta: Yeah, makes sense. Just raised my hand, but it should be easy to live without it. (Moreover, I think my use case actually should be fixed on the Quote node level)

packages/lexical-selection/src/range-selection.ts
EgonBolton: Actually, there are many types of nodes. Think of table, row, cell, list, listitem... etc.
One option would be to add a $isBlockNode function to each node, as was done with ShadowRoot. But I am reluctant to add new concepts and complexity to the definition of a node.
I think something pretty close to what I put in the comment above should work.

packages/lexical-selection/src/range-selection.ts
EgonBolton: solved!

EgonBolton:
> The e2e failure is probably due to your changes to LexicalNode.replace. Haven't checked, but that method is probably called somewhere in the code path exercised by that test.

Hi, @acywatson sorry for the delay. I have already fixed all the problems except this one. Are you sure the problem is with the replace method? I look at the e2e test log and it seems that the problem is with the IME tests. Is it possible that I messed up something related to the playwright fork?

acywatson:
> > The e2e failure is probably due to your changes to LexicalNode.replace. Haven't checked, but that method is probably called somewhere in the code path exercised by that test.
> 
> Hi, @acywatson sorry for the delay. I have already fixed all the problems except this one. Are you sure the problem is with the replace method? I look at the e2e test log and it seems that the problem is with the IME tests. Is it possible that I messed up something related to the playwright fork?

Not sure how you could have done that. It's definitely the IME tests that are failing, but I think it's more likely because of your code changes that because of playwright. I mentioned replace because that's a very widely used method and is probably used somewhere in the code path exercised by that failing test.

EgonBolton:
> Not sure how you could have done that. It's definitely the IME tests that are failing, but I think it's more likely because of your code changes that because of playwright. I mentioned replace because that's a very widely used method and is probably used somewhere in the code path exercised by that failing test.

@acywatson I've tried restoring the two replace functions I modified (on LexicalNode and LexicalListItemNode) and those tests keep failing.

The tests do not fail any internal process. It just breaks when page.keyboard.imeSetComposition() is called, because "it's not a function".

Any other test or even any other method of page.keyboard in Playwright seems to work fine. :/

EgonBolton:
> are these failures happening locally? In the Github CI logs, I see mismatched HTML output as the cause for the failure.

Effectively. On my pc I get: "page.keybard.imeSetComposition is not a function". In CI there is an HTML mismatch.

However, I am seeing that it marks many things as discrepancies that, as far as I can see, are not:
image

EgonBolton:
Ok, it cost but I found the error.

The problem in local was that I had to update Playwright because the fork version does not allow the VS Code extension to work (please I would appreciate it if you could update the fork).

The problem on the server was with the `Composition › IME › Can type Hiragana via IME part way through a mention` test. For some reason it was inserting a zero width space in chromium, so I had to slightly modify the test.

Now the tests pass, except collab-linux, I don't know why. Any ideas?

EgonBolton:
Hello, sorry for the inconvenience. Do you know how I can debug for Linux? Or is this PR ready to pass? How can I move this forward?

EgonBolton:
I have already resolved the conflicts. Can you please approve this? @acywatson

packages/lexical/src/LexicalNode.ts
```diff
@@ -623,8 +625,14 @@ export class LexicalNode {
...
 
-  replace<N extends LexicalNode>(replaceWith: N, restoreSelection = true): N {
+  replace<N extends LexicalNode>(
+    replaceWith: N,
+    restoreSelection = true,
+    includeChildren?: boolean,
+  ): N {
     errorOnReadOnly();
+    let selection = $getSelection();
+    if (selection !== null) selection = selection.clone();
```
acywatson, line 635: Why do you have to clone selection here?

packages/lexical-selection/src/range-selection.ts
```diff
@@ -6,313 +6,64 @@
...
- * It attempts to append the resulting sub-tree to the nearest safe insertion target.
+ * Converts all nodes in the selection that are of one block type to another specified by parameter
  *
  * @param selection
  * @param createElement
- * @param wrappingElement
  * @returns
  */
-export function $wrapNodes(
+export function $setBlocksType(
```
acywatson, line 33: I like how much you were able to simplify this, but I'm a bit worried about untested edge cases here. Internally, we don't have semantic versioning, so we can't simply roll back to the previous version if this function blows something up. My recommendation would be to change $setBlocksType to $setBlocksType_experimental and leave $wrapNodes in place. We'll be able to merge this in without a breaking change then and we can test it on the playground. Then in 0.8, we can deprecate $wrapNodes and drop the experimental from $setBlocksType.

Also, I'm not sure about introducing the concept of "Blocks". It's a bit confusing alongside Elements. Instead, can we just call it $setElementTypes?

acywatson:
> I have already resolved the conflicts. Can you please approve this? @acywatson

Test are passing - that piece looks good. A couple of logistical concerns I commented on. Thanks for sticking with this one. It's a bit of a beast to change critical functions like $wrapNodes.

packages/lexical-selection/src/range-selection.ts
EgonBolton: $setBlocksType_experimental seems good to me.

I agree that the concept of "block" can be ambiguous, but it is certainly a subset of "element". Blocks are those nodes that you would expect could be converted to another type in any text editor. You cannot pass LinkNode to $setBlocksType. On the other hand, this is a concept that I didn't introduce myself... [it was already in the toolbar plugin](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx#L93).

In fact, I plan to open a discussion or issue as soon as this PR is approved about whether they should not be a single node, expanding on what I said [here ](https://github.com/facebook/lexical/issues/1262#issuecomment-1186170960
) and [here](https://github.com/facebook/lexical/issues/1262#issuecomment-1292436961
)
That was the origin of this PR actually.

packages/lexical/src/LexicalNode.ts
EgonBolton: Because replace can break the selection and is necessary to restore it. In fact, you've pointed out to me that my implementation can also remove the extra restore selection parameter that was just recently added in another PR.

EgonBolton:
what about now? @acywatson

acywatson (approved):

Issue #3041, opened 2022-09-20
Bug: `selection.insertNodes` inserts 'top level' nodes inline


Piliuta:
This is a core issue in the `selection.insertNodes` method. Using this method to insert top-level nodes (`isTopLevel: true`) inserts nodes inline. So if the selection is in a middle of a paragraph, it inserts the node inside the paragraph.

This was discovered in the copy and paste functionality, but really can affect anything that uses this method.

Lexical version: 0.4.1

## Steps To Reproduce

1. Create a selectable top level decorator node. I use `ImageNode` from the lexical library, but made it top level
2. Add an image in the editor
3. Select and copy the image node
4. Paste it in the middle of a text - the image inserted inline

_**Note: There is also a fix for copying image src - not sure if this is a relative path issue or what though.**_ 

```diff
diff --git a/packages/lexical-playground/src/nodes/ImageNode.tsx b/packages/lexical-playground/src/nodes/ImageNode.tsx
index 3adf5b24..a87adc7b 100644
--- a/packages/lexical-playground/src/nodes/ImageNode.tsx
+++ b/packages/lexical-playground/src/nodes/ImageNode.tsx
@@ -42,8 +42,8 @@ export interface ImagePayload {
 
 function convertImageElement(domNode: Node): null | DOMConversionOutput {
   if (domNode instanceof HTMLImageElement) {
-    const {alt: altText, src} = domNode;
-    const node = $createImageNode({altText, src});
+    const {alt: altText, dataset} = domNode;
+    const node = $createImageNode({altText, src: dataset.src});
     return {node};
   }
   return null;
@@ -115,6 +115,7 @@ export class ImageNode extends DecoratorNode<JSX.Element> {
   exportDOM(): DOMExportOutput {
     const element = document.createElement('img');
     element.setAttribute('src', this.__src);
+    element.setAttribute('data-src', this.__src);
     element.setAttribute('alt', this.__altText);
     return {element};
   }
@@ -202,6 +203,14 @@ export class ImageNode extends DecoratorNode<JSX.Element> {
     return this.__altText;
   }
 
+  isTopLevel(): boolean {
+    return true;
+  }
+
+  isInline(): boolean {
+    return false;
+  }
+
   decorate(): JSX.Element {
     return (
       <Suspense fallback={null}>
```

https://user-images.githubusercontent.com/1575198/191313699-8099f2ed-0de9-4e41-80ae-5662550a193a.mov

## The current behavior
Top-level nodes inserted inline.

[...]

PR #3238, opened 2022-10-21
Fix pasting top level nodes inline

PSmolic:
Resoves: https://github.com/facebook/lexical/issues/3041
## Problem
Pasting top level nodes (isInline set to `false`) within a paragraph places node in paragraph instead of breaking paragraph and placing node between.

### Steps to reproduce 
https://user-images.githubusercontent.com/4542049/197205190-a048afb8-059c-4934-9cc8-5827b491bf60.mov

## Solution
https://user-images.githubusercontent.com/4542049/197206901-6afc13cc-7a12-416b-9fc3-c0b9ae3140a3.mov

## Note
This is still partially broken for `headings` and `quotes` because when they get broken the second part becomes paragraph instead of whatever node it originally was but it is still an improvement on current behavior.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -14,7 +14,6 @@ import type {TextFormatType} from './nodes/LexicalTextNode';
 
 import {$splitNode} from '@lexical/utils';
 import {IS_CHROME} from 'shared/environment';
-import getDOMSelection from 'shared/getDOMSelection';
 import invariant from 'shared/invariant';
 
 import {
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1438,7 +1438,8 @@ export class RangeSelection implements BaseSelection {
         if (
           $isRangeSelection(this) &&
           $isDecoratorNode(node) &&
-          !node.isInline()
+          !node.isInline() &&
+          $isTextNode(target)
         ) {
           this.insertParagraph();
           target = this.focus
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1433,7 +1433,21 @@ export class RangeSelection implements BaseSelection {
         ($isDecoratorNode(target) && !target.isInline())
       ) {
         lastNode = node;
-        target = target.insertAfter(node, false);
+        // when pasting top level node in the middle of paragraph
+        // we need to split paragraph instead of placing it inline
+        if (
+          $isRangeSelection(this) &&
+          $isDecoratorNode(node) &&
+          !node.isInline()
+        ) {
+          this.insertParagraph();
+          target = this.focus
+            .getNode()
+            .getTopLevelElementOrThrow()
+            .insertBefore(node);
+        } else {
+          target = target.insertAfter(node, false);
+        }
       } else {
         const nextTarget: ElementNode = target.getParentOrThrow();
         // if we're inserting an Element after a LineBreak, we want to move the target to the parent
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1440,11 +1440,11 @@ export class RangeSelection implements BaseSelection {
         if (
           $isRangeSelection(this) &&
           $isDecoratorNode(node) &&
-          !$isDecoratorNode(target) &&
+          ($isElementNode(target) || $isTextNode(target)) &&
           !node.isInline()
         ) {
-          let splitNode;
-          let splitOffset;
+          let splitNode: ElementNode;
+          let splitOffset: number;
 
           if ($isTextNode(target)) {
             splitNode = target.getParentOrThrow();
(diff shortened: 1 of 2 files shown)


Piliuta:
@acywatson could you take a look into this PR? This fixes the issue reported here https://github.com/facebook/lexical/issues/3041

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1418,7 +1418,21 @@ export class RangeSelection implements BaseSelection {
         ($isDecoratorNode(target) && !target.isInline())
       ) {
         lastNode = node;
-        target = target.insertAfter(node);
+        // when pasting top level node in the middle of paragraph
```
acywatson, line 1421: You sort of addressed my feedback in the summary - this won't work for any type of ElementNode except paragraph. Well, it will, but it won't work perfectly.  I wonder if what we need here is splitNode:

https://github.com/facebook/lexical/pull/2322

Piliuta: _Not sure how far that PR is from shipping._ Would it make sense to have this PR merged and when/if #2322 is done, we make a follow-up change to improve this logic?

acywatson:
I do think this is an improvement in this on specific scenario, but I'm concerned about regressions in untested cases. We can't keep merging special cases into insertNodes until we have more robust testing in place. @fantactuka is working on this a bit, IIRC

Piliuta:
> I do think this is an improvement in this on specific scenario, but I'm concerned about regressions in untested cases. We can't keep merging special cases into insertNodes until we have more robust testing in place. @fantactuka is working on this a bit, IIRC

Since the work is still in progress but the problem exists now (I do think this a bad issue because it distorts the state), would you propose any workarounds that can be done outside the lexical?

acywatson:
> > I do think this is an improvement in this on specific scenario, but I'm concerned about regressions in untested cases. We can't keep merging special cases into insertNodes until we have more robust testing in place. @fantactuka is working on this a bit, IIRC
> 
> Since the work is still in progress but the problem exists now (I do think this a bad issue because it distorts the state), would you propose any workarounds that can be done outside the lexical?

@Piliuta Looks like #3342 can help here @fantactuka

Piliuta:
> > > I do think this is an improvement in this on specific scenario, but I'm concerned about regressions in untested cases. We can't keep merging special cases into insertNodes until we have more robust testing in place. @fantactuka is working on this a bit, IIRC
> > 
> > 
> > Since the work is still in progress but the problem exists now (I do think this a bad issue because it distorts the state), would you propose any workarounds that can be done outside the lexical?
> 
> @Piliuta Looks like #3342 can help here @fantactuka

That looks great. Thank you.
Do you think it is okay to simply replace `insertParagraph` with the new split method? Are you okay with the conditions we have here?

Also, I see that `$insertNodeToNearestRoot` https://github.com/facebook/lexical/pull/3344 does half of the `selection.insertNodes` job, is it meant to replace some functionality in the selection (at least for inserting non-inline nodes case?)

fantactuka:
> Also, I see that $insertNodeToNearestRoot https://github.com/facebook/lexical/pull/3344 does half of the selection.insertNodes job, is it meant to replace some functionality in the selection (at least for inserting non-inline nodes case?)

They are a bit different (e.g., $insertNodeToNearestRoot does not current non-collapsed selection content). I'd expect that $splitNode can be useful there, but in general, `insertNodes` needs big rework as it became a pile of patches for specific cases and still have a bunch of bugs around decorators and inline elements

Piliuta:
@acywatson @fantactuka does look good now as a patch for the current issue until `insertNodes` is refactored?

acywatson:
> @acywatson @fantactuka does look good now as a patch for the current issue until `insertNodes` is refactored?

I'm good with this - just fix the type error and that should clean up the remaining test failures

packages/lexical/src/LexicalSelection.ts
trueadm: We have $splitNode now AFAIK.

PSmolic: @trueadm I updated it to use `splitText()` (`$splitNode()` does not support splitting text) and `$splitNode()` the other day.
https://github.com/facebook/lexical/pull/3238/files#diff-78b86cfd1ec931911a488d41ac3b779722121b8657aa59062b78856fa63cae12R1433-R1445

acywatson (approved):
Looks good. I'd like to rework this function soon, but this fixes an important issue for now.

trueadm:
@acywatson Are we okay to merge this PR?

thegreatercurve:
Merging this as the changes LGTM.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -12,6 +12,7 @@ import type {LexicalNode, NodeKey} from './LexicalNode';
 import type {ElementNode} from './nodes/LexicalElementNode';
 import type {TextFormatType} from './nodes/LexicalTextNode';
 
+import {$splitNode} from '@lexical/utils';
```
trueadm, line 15: Ouch. You ca't pull in external modules here.

Issue #3240, opened 2022-10-22
Bug: Sticky Note Plugin crashes Playground on import


timaxlucas:
Lexical version: 0.5.1-next.1

## Steps To Reproduce

1. Open [Playground](https://playground.lexical.dev/)
2. Create a Sticky Note
3. Export
4. Import

```
TypeError: Cannot set properties of undefined (setting 'theme')
    at LexicalNestedComposer.1647df3c.js:1:575
    at Object.useMemo (main.fa4a2c4e.js:21:69182)
    at c.useMemo (main.fa4a2c4e.js:1:8266)
    at i (LexicalNestedComposer.1647df3c.js:1:443)
    at ma (main.fa4a2c4e.js:21:62859)
    at fu (main.fa4a2c4e.js:21:119869)
    at cc (main.fa4a2c4e.js:21:108839)
    at lc (main.fa4a2c4e.js:21:108767)
    at ac (main.fa4a2c4e.js:21:108630)
    at Xu (main.fa4a2c4e.js:21:105437)
```

timaxlucas:
Update: There are several issues with the StickyNote plugin. Thanks #3241 - this almost fixes the persistency issue.

1. StickyNode is sometimes recreated onMouseUp (when its dropped)
2. Weird behaviour of note dragging if `document.body` is scrollable.

I'll try to my best to do another PR.

timaxlucas:
PR #3241 does fix the crash, but nested editorState isn't restored on import. PR #3246 fixes both.

PR #3241, opened 2022-10-22
fix editor crash when importing editorState with nested stickyNote

karam-qaoud:
This is to fix issue #3240 

Looks like the StickyNote class expects to receive a `caption` of type `LexicalEditor`, but actually gets a `Caption` type - an `EditorState` wrapped inside an object. This causes the editor to crash as it tries to access properties which `Caption` type doesn't have. Though we can use the `editorState` inside the newly defined `Caption` type to create a new Lexical editor.

### DIFF
--- a/packages/lexical-playground/src/nodes/StickyComponent.tsx
+++ b/packages/lexical-playground/src/nodes/StickyComponent.tsx
@@ -55,9 +55,9 @@ export default function StickyComponent({
   y,
   nodeKey,
   color,
-  editorCaption,
+  captionEditor,
 }: {
-  editorCaption: LexicalEditor;
+  captionEditor: LexicalEditor;
   color: 'pink' | 'yellow';
   nodeKey: NodeKey;
   x: number;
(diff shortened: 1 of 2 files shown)


timaxlucas:
Thanks for your help! I just tested your fix. Unfortunely the nested editor does not persist data. It has something to do with `importJSON` and the newly created `Caption` type.

karam-qaoud:
@timaxlucas I just noticed that the value of  `caption` in `exportJSON` was always undefined, therefore the `editorState` of StikyNote was not persisted when exporting. Is this the error you had? I just committed a fix to that. If you experienced a different problem, please feel free to write down the steps to reproduce and paste the error message if you have it.

timaxlucas:
> @timaxlucas I just noticed that the value of `caption` in `exportJSON` was always undefined, therefore the `editorState` of StikyNote was not persisted when exporting. Is this the error you had? I just committed a fix to that. If you experienced a different problem, please feel free to write down the steps to reproduce and paste the error message if you have it.

I just tested it and it still didn't persist the data. See #3246 for a working fix.

PR #3243, opened 2022-10-23
Expose Tokenizer interface to accept custom tokenizer other than Prism at @lexical/code

saint1991:
Prism has powerful functionality to handle various languages including custom language, 
but some other tools like [ANTLR](https://www.antlr.org) have their own grammar definition and parser generator.

This PR exposes the Tokenizer interface at `registerCodeHighlighting` in @lexical/code plugin to handle such a case.

### DIFF
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -58,6 +58,26 @@ import {
 
 import {$isCodeNode, CodeNode} from './CodeNode';
 
+type TokenContent = string | Token | (string | Token)[];
+
+export interface Token {
+  type: string;
+  content: TokenContent;
+}
+
+export interface Tokenizer {
+  tokenize(code: string, language?: string): (string | Token)[];
+}
+
+export const PrismTokenizer: Tokenizer = {
+  tokenize(code: string, language?: string): (string | Token)[] {
+    return Prism.tokenize(
+      code,
+      Prism.languages[language || ''] || Prism.languages[DEFAULT_CODE_LANGUAGE],
+    );
+  },
+};
+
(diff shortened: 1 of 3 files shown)


packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -654,13 +679,20 @@ function handleMoveTo(
   return true;
 }
 
-export function registerCodeHighlighting(editor: LexicalEditor): () => void {
+export function registerCodeHighlighting(
+  editor: LexicalEditor,
+  tokenizer?: Tokenizer,
```
saint1991, line 684: To keep compatibility, add tokenizer as an optional argument.
If it is not specified, PrismTokenizer will be used like before (at line 693).

packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -58,6 +58,26 @@ import {
...
+export interface Token {
+  readonly type: string;
+  readonly content: TokenContent;
+}
+
+export interface Tokenizer {
+  tokenize(code: string, language?: string): (string | Token)[];
+}
+
+export const PrismTokenizer: Tokenizer = {
```
saint1991, line 72: Tokenizer implementation using Prism that is the same as the original implementation and this will be used as default.

fantactuka:
Much needed PR!
0.6 release could introduce a breaking change where tokenizer is required and prism one is not included into the bundle

Could you please add comment for the tokenizer argument:
```
TODO: 0.6 make required
```

danestves:
This will be perfect, I want to use other code highlight than Prism!

acywatson (approved):

fantactuka (approved):

PR #3245, opened 2022-10-23
Handle left part of match

Neewd:
The`importTextMatchTransformers` use the splitText method to handle the match and continue the loop. 
However it never use the left part of the match as the split text method returns an array of text node. 
Thus if you have a future textMatchTransformers match in the left part (before the replaceNode) it won't be caught.

I just catch the left part of the match and do a recursive call to the same method to handle this.

Let me know if there any other way of doing it.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -289,14 +289,22 @@ function importTextMatchTransformers(
 
       const startIndex = match.index || 0;
       const endIndex = startIndex + match[0].length;
-      let replaceNode;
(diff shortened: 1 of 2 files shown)


fantactuka:
Thank you! Could you add a test for this? Fix looks reasonable

Neewd:
> Thank you! Could you add a test for this? Fix looks reasonable

Didn't see any related tests directory for this but for sure I can create one ;)

Neewd:
> Thank you! Could you add a test for this? Fix looks reasonable

Done ;)

acywatson (approved):

PR #3246, opened 2022-10-23
fix playground Sticky Note export/import causing crash

timaxlucas:
Fixes issue #3240 where export/import of StickyNote crashes Playground. Now with working state export/restore.

### DIFF
--- a/packages/lexical-playground/src/nodes/StickyNode.tsx
+++ b/packages/lexical-playground/src/nodes/StickyNode.tsx
@@ -44,7 +44,7 @@ export class StickyNode extends DecoratorNode<JSX.Element> {
   __x: number;
   __y: number;
   __color: StickyNoteColor;
-  __captionEditor: LexicalEditor;
+  __caption: LexicalEditor;
 
   static getType(): string {
     return 'sticky';
@@ -55,7 +55,7 @@ export class StickyNode extends DecoratorNode<JSX.Element> {
       node.__x,
       node.__y,
       node.__color,
-      node.__captionEditor,
+      node.__caption,
       node.__key,
     );
   }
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/nodes/StickyNode.tsx
```diff
@@ -32,7 +33,7 @@ export type SerializedStickyNode = Spread<
     xOffset: number;
     yOffset: number;
     color: StickyNoteColor;
-    caption: LexicalEditor;
+    caption: EditorState;
```
fantactuka, line 36: Thanks for the fix!

Here it should be `SerializedEditor`, and the `exportJSON` would use `this.__captionEditor.toJSON()` (could use ImageNode.tsx as a reference). On another note, caption (vs captionEditor) seemed ok for the props and for some consistency with other nested editors usages.

timaxlucas: Thanks for the feedback, @fantactuka. I fixed it in my latest commit.

packages/lexical-playground/src/nodes/StickyNode.tsx
```diff
@@ -54,36 +55,45 @@ export class StickyNode extends DecoratorNode<JSX.Element> {
...
+      captionEditor,
     );
   }
 
   constructor(
     x: number,
     y: number,
     color: 'pink' | 'yellow',
-    caption?: LexicalEditor,
+    captionEditor?: LexicalEditor,
```
fantactuka, line 84: Could you roll it back to `caption` too? Although it's not exposed as "official" plugin, some would copy it to their code base. And if by any chance it uses collaboration, then this will be a breaking change, since in collab mode node properties are mapped to Yjs data types. Since there's no strong need to rename it, let's keep it as is

timaxlucas: sure! Should the class attribute be `__caption` too?

fantactuka: Yes, argument name is less important

timaxlucas: done

PR #3247, opened 2022-10-24
Allow clearing styles with $patchStyleText

### LINKED ISSUE #3235: Bug: No way to clear styles with $patchStyleText
Lexical version: 0.5.0

## Steps To Reproduce

1. Apply style using `$patchStyleText`, e.g. `$patchStyleText(selection, { 'text-emphasis': 'filled' })`.
2. Attempt to clear said style

## The current behavior

There is no way (that I can find) to clear the style.

`$patchStyleText`'s `patch` argument expects a value of type `Record<string, string>`.

If you set the value to the empty string, then you can remove the visual effect, but because any affected text nodes' `__style` is still set, they won't be merged.

## The expected behavior

The `patch` argument should be something like `Record<string, string | undefined>` (or alternatively `null` if that's more idiomatic). Then you could clear the style using `$patchStyleText(selection, { 'text-emphasis': undefined })`.

I'm happy to make up a PR for this if you agree with the request. I believe it mostly involves getting `getCSSFromStyleObject` to skip any properties where the value is `undefined`. I could use guidance on whether `undefined` or `null` is more idiomatic or if we should support both.

birtles:
Fixes #3235

### DIFF
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -301,9 +301,22 @@ export function $addNodeStyle(node: TextNode): void {
   CSS_TO_STYLES.set(CSSText, styles);
 }
 
-function $patchNodeStyle(node: TextNode, patch: Record<string, string>): void {
+function $patchNodeStyle(
+  node: TextNode,
+  patch: Record<string, string | null>,
+): void {
   const prevStyles = getStyleObjectFromCSS(node.getStyle());
-  const newStyles = prevStyles ? {...prevStyles, ...patch} : patch;
+  const newStyles = Object.entries(patch).reduce<Record<string, string>>(
+    (styles, [key, value]) => {
+      if (value === null) {
+        delete styles[key];
+      } else {
+        styles[key] = value;
+      }
+      return styles;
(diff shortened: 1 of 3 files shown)


packages/lexical-selection/src/lexical-node.ts
```diff
@@ -301,17 +301,30 @@ export function $addNodeStyle(node: TextNode): void {
...
+  );
   const newCSSText = getCSSFromStyleObject(newStyles);
   node.setStyle(newCSSText);
   CSS_TO_STYLES.set(newCSSText, newStyles);
 }
 
 export function $patchStyleText(
   selection: RangeSelection | GridSelection,
-  patch: Record<string, string>,
+  patch: Record<string, string | undefined | null>,
```
acywatson, line 327: I know you asked this and I didn't respond in time, but I think we should remove undefined from the union.  We generally try to avoid using undefined explicitly in Lexical's public API (implicit undefineds in option params, for example, notwithstanding.

One other small request - can you update the Flow type for this function?

birtles: Sure thing! All done.

PR #3248, opened 2022-10-24
Improve Collab cursor, add custom cursorColor

strdr4605:
- fix(lexical-yjs): remove height from collab cursor name

We remove the selection height because it does not work with a common CSS style

```css
* {
  box-sizing: border-box;
}
```

Screenshot 2022-10-24 at 16 47 26

when we remove the height it works

Screenshot 2022-10-24 at 16 50 22

---

- feat(collab): add cursorColor prop for LexicalCollaborationPlugin

I had to create a sibling for the caret span in order to give devs the possibility to provide any color for the caret, name, and selection.

---

Also, I think that we should add the possibility to specify a className (for the name at least) to give more freedom.

CC @fantactuka

### DIFF
--- a/packages/lexical-yjs/src/SyncCursors.ts
+++ b/packages/lexical-yjs/src/SyncCursors.ts
@@ -148,10 +148,10 @@ function createCursorSelection(
 ): CursorSelection {
   const color = cursor.color;
   const caret = document.createElement('span');
-  caret.style.cssText = `position:absolute;top:0;bottom:0;right:-1px;width:1px;background-color:rgb(${color});z-index:10;`;
+  caret.style.cssText = `position:absolute;top:0;bottom:0;right:-1px;width:1px;background-color:${color};z-index:10;`;
   const name = document.createElement('span');
   name.textContent = cursor.name;
-  name.style.cssText = `position:absolute;left:-2px;top:-16px;background-color:rgb(${color});color:#fff;line-height:12px;font-size:12px;padding:2px;font-family:Arial;font-weight:bold;white-space:nowrap;`;
+  name.style.cssText = `position:absolute;left:-2px;top:-16px;background-color:${color};color:#fff;line-height:12px;font-size:12px;padding:2px;font-family:Arial;font-weight:bold;white-space:nowrap;`;
   caret.appendChild(name);
   return {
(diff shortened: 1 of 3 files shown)


packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -148,10 +148,10 @@ function createCursorSelection(
 ): CursorSelection {
   const color = cursor.color;
   const caret = document.createElement('span');
-  caret.style.cssText = `position:absolute;top:0;bottom:0;right:-1px;width:1px;background-color:rgb(${color});z-index:10;`;
+  caret.style.cssText = `position:absolute;top:0;bottom:0;right:-1px;width:1px;background-color:${color};opacity:1;z-index:10;`;
   const name = document.createElement('span');
   name.textContent = cursor.name;
-  name.style.cssText = `position:absolute;left:-2px;top:-16px;background-color:rgb(${color});color:#fff;line-height:12px;height:12px;font-size:12px;padding:2px;font-family:Arial;font-weight:bold;white-space:nowrap;`;
+  name.style.cssText = `position:absolute;left:-2px;top:-16px;background-color:${color};opacity:1;color:#fff;line-height:12px;font-size:12px;padding:2px;font-family:Arial;font-weight:bold;white-space:nowrap;`;
```
fantactuka, line 154: Does this need opacity?

fantactuka:
LGTM, as for customization, it might become too much of properties to pass around through multiple plugins. Maybe it would be possible to pass renderCursor function that would deal with dom creation/styling/insertion rather than passing separate flags/props to it (including cursor container)

strdr4605:
> LGTM, as for customization, it might become too much of properties to pass around through multiple plugins. Maybe it would be possible to pass renderCursor function that would deal with dom creation/styling/insertion rather than passing separate flags/props to it (including cursor container)

Will see if we need more customization.

As a workaround, we could style cursors through cursorContainerRef

packages/lexical-yjs/src/SyncCursors.ts
strdr4605: No😊

acywatson (approved):

PR #3254, opened 2022-10-25
Select end if root has no children in focus logic

thegreatercurve:
It shouldn't matter if the editor is empty when called `editor.focus`. The cursor should still appear somewhere in the editor when the method is called.

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -816,12 +816,10 @@ export class LexicalEditor {
           if (selection !== null) {
             // Marking the selection dirty will force the selection back to it
             selection.dirty = true;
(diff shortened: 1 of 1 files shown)


zurfyx (approved):
I'm would dare to say this was intentional but I can't recall why..

thegreatercurve:
> I'm would dare to say this was intentional but I can't recall why..

I looked through the blame history but didn't see anything. I guess we'll see!

acywatson:
> > I'm would dare to say this was intentional but I can't recall why..
> 
> I looked through the blame history but didn't see anything. I guess we'll see!

I think this is why:

https://github.com/facebook/lexical/actions/runs/3322384417/jobs/5492978426

collab tests on webkit failing starting with this diff.

PR #3256, opened 2022-10-25
Refactor `isHighlighting` flag into object literal

### LINKED ISSUE #3257: Code highlighting doesn't work with more than one code block
Code highlighting seems to not work when there is more than one code block on the page.

image

thegreatercurve:
The `isHighlighting` flag gets set to `true` when there is more than 1 `CodeNode` rendered on the page on load, causing subsequent `CodeNodes` to not get highlighted.

Fixes #3257

### Before
image

### After
image

### DIFF
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -250,26 +250,31 @@ function updateCodeGutter(node: CodeNode, editor: LexicalEditor): void {
 // Using `skipTransforms` to prevent extra transforms since reformatting the code
 // will not affect code block content itself.
 //
-// Using extra flag (`isHighlighting`) since both CodeNode and CodeHighlightNode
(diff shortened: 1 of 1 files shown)


fantactuka (approved):

packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -291,16 +298,18 @@ function codeNodeTransform(
...
           return true;
         }
+
         return false;
       });
     },
     {
       onUpdate: () => {
-        isHighlighting = false;
+        delete nodesCurrentlyHighlighting[nodeKey];
```
trueadm, line 312: Can we avoid using an object and delete and instead use a Set?

thegreatercurve: Yeah, it's not really much different from an obj literal though, if I use `performance.now` at the beginning and end and slow down the CPU. Still, might be easier to read #3258

Issue #3155, opened 2022-10-11
Bug: Press "Delete" on a paragraph after decorator nodes removes the node


Piliuta:
When the cursor focus is set on a paragraph next to a decorator node (not inline) pressing the "Delete" ("Backspace") key removes the node.

Lexical version: 0.5.0

## Steps To Reproduce

1. Add any non-inline decorator node to the editor (e.g. Horizontal Rule)
2. Add a paragraph next to this node
3. Press "Delete" key

https://user-images.githubusercontent.com/1575198/195064627-51352f1a-5f50-4aee-ad19-d6b3a571b243.mov

## The current behavior
Pressing "Delete" key before a non-inline decorator node removes the node.

## The expected behavior
Removing an empty paragraph should not remove the decorator node on pressing "Delete" key.
If the paragraph (or the start of whatever non-inline block) is not empty it should remove the node above, i.e. if the merge cannot be resolved into a valid state.

thegreatercurve:
@Piliuta What browser and OS are you using for this bug? is this I'm struggling to replicate this on Chrome.

Piliuta:
> @Piliuta What browser and OS are you using for this bug? is this I'm struggling to replicate this on Chrome.

_I am on Mac OS in Chrome._

I am sorry, I meant to say "that button on the right top corner that I never look at" the "Delete" (or "Backspace" on windows) button 😳. I updated the description.

thegreatercurve:
Just to confirm, if the paragraph node after a decorator node is empty, on hitting "Delete" whilst selection is within the empty paragraph node, you expect the empty paragraph to disappear and the decorator node to become selected? Similar to how it is on ProseMirror.

https://user-images.githubusercontent.com/7748470/195388657-4c7d2374-079b-4a4a-a852-afeb680a0e5e.mov

I can't find the related issue but I think I've seen this mentioned elsewhere. I think the discussion boiled down to it isn't really a bug, it's just a choice we made that if a user hits delete/backspace, the decorator node or character before the carat should be deleted.

Maybe someone else knows different?

Piliuta:
> Just to confirm, if the paragraph node after a decorator node is empty, on hitting "Delete" whilst selection is within the empty paragraph node, you expect the empty paragraph to disappear and the decorator node to become selected? Similar to how it is on ProseMirror.

Yes, exactly. And this is how many other advanced editors behave (or at least they do not remove the content ahead). 

[...]

Issue #3156, opened 2022-10-11
Bug: Caret has wrong visual position when focus is set between decorator nodes


Piliuta:
Setting the caret position between non-inline decorator nodes indicates the caret is in the wrong place even though lexical state recognizes its position correctly, e.g. typing is allowed between the nodes.

Lexical version: 0.5.0

## Steps To Reproduce

1. Add two non-inline decorator nodes (e.g. horizontal rule)
2. Add some text after or/and before the nodes
3. Move the caret between the decorator nodes, e.g
  a.  Set the cursor on the start of the paragraph next to the last decorator node
  b. Press Arrow Left twice - the first time it selects the node, second time it _sets_ the caret position before the decorator node

https://user-images.githubusercontent.com/1575198/195071814-73b4ed40-4f73-4e94-aeaa-4328d59b9181.mov

## The current behavior
Caret position between non-inline decorator nodes is indicated after the last decorator node.

## The expected behavior
The caret position is indicated between the decorator nodes.

trueadm:
This is both a known issue and also not really a Lexical bug at all. It’s a browser bug and one we’ve been asking browser vendors to fix for a while. :/

trueadm:
@Piliuta You could emulate a cursor using a div that looks like a cursor and absolutely position it ontop to where you think it should be in the cases where this might happen. I believe there was another issue where I think we spoke about this, but can't seem to find that issue. You could make this behavior work in a plugin that emulates the cursor with the fake one if it thinks it might be lost (i.e. RangeSelection with element selection between decorators).

Piliuta:
> @Piliuta You could emulate a cursor using a div that looks like a cursor and absolutely position it ontop to where you think it should be in the cases where this might happen. I believe there was another issue where I think we spoke about this, but can't seem to find that issue. You could make this behavior work in a plugin that emulates the cursor with the fake one if it thinks it might be lost (i.e. RangeSelection with element selection between decorators).

That's exactly what I had in mind as a spare option. I'll try it. Thanks

> > Thanks for reply. Is there any link about this browser bug ? Such as a chromium bug issue link
> 
> https://bugs.chromium.org/p/chromium/issues/detail?id=384357

This report is closed and does not seem to be exactly the same problem, no?

trueadm:
I probably linked the wrong one. Sorry I’m on my phone and not my MacBook whilst on parental leave. When I get back I’ll find it for you :)

DaniGuardiola:
[...]

PR #3261, opened 2022-10-26
Fix deleting empty paragraphs

PSmolic:
Fixes: https://github.com/facebook/lexical/issues/3155
## Problem
When the cursor focus is set on a paragraph next to a decorator node (not inline) pressing the "Delete" ("Backspace") key removes the node.

## Note
Fixing deletion behavior makes bug described here a bit more noticeable: https://github.com/facebook/lexical/issues/3156

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/HorizontalRule.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/HorizontalRule.spec.mjs
@@ -350,8 +350,9 @@ test.describe('HorizontalRule', () => {
     page,
     browserName,
     isPlainText,
+    isCollab,
   }) => {
-    test.skip(isPlainText);
+    test.skip(isPlainText || isCollab);
 
     await focusEditor(page);
 
@@ -363,12 +364,7 @@ test.describe('HorizontalRule', () => {
       page,
       html`
         <p class="PlaygroundEditorTheme__paragraph"><br /></p>
-        <div
-          contenteditable="false"
-          style="display: contents;"
-          data-lexical-decorator="true">
-          <hr />
-        </div>
+        <hr class="" contenteditable="false" data-lexical-decorator="true" />
         <p class="PlaygroundEditorTheme__paragraph"><br /></p>
       `,
     );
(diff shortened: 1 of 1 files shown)


acywatson:
Your test is failing on mac+webkit (and possibly other suites) even after 2 re-runs.

Assuming you fix that, the problem would be that after deleting the paragraph after the DecoratorNode, it's now impossible to type after the DecoratorNode. That's because we put ParagraphNode after DecoratorNode to serve the functional purpose of allowing the user to select and type after the DecoratorNode.

Piliuta:
> Assuming you fix that, the problem would be that after deleting the paragraph after the DecoratorNode, it's now impossible to type after the DecoratorNode. That's because we put ParagraphNode after DecoratorNode to serve the functional purpose of allowing the user to select and type after the DecoratorNode.

That's a valid selection though. Lexical sets such as the selection between decorator nodes and it is possible to type there (the only issue is that the browser does not reflect such selection correctly which is mentioned in this PR https://github.com/facebook/lexical/issues/3156).

In this PR it would set the cursor after the node where typing is possible and it creates a new paragraph.

acywatson:
> > Assuming you fix that, the problem would be that after deleting the paragraph after the DecoratorNode, it's now impossible to type after the DecoratorNode. That's because we put ParagraphNode after DecoratorNode to serve the functional purpose of allowing the user to select and type after the DecoratorNode.
> 
> That's a valid selection though. Lexical sets such as the selection between decorator nodes and it is possible to type there (the only issue is that the browser does not reflect such selection correctly which is mentioned in this PR #3156).
> 
> In this PR it would set the cursor after the node where typing is possible and it creates a new paragraph.

Ok, I see thanks. Let me sync with Dominic about the bug he mentioned in that other PR

trueadm:
I believe this has since been fixed, but maybe we can keep your e2e test? Please could you rebase and see if it passes?

trueadm (approved):

trueadm:
Looks like your test doesn't work in collab mode. You could try excluding it from running in collab CI tests.

PSmolic:
So many tests :) I pushed an update to skip that test in collab mode.
Do you think there is a sense in adding collab and plain text mode test commands to readme? I can add a PR if you think that would be beneficial.

trueadm:
> Do you think there is a sense in adding collab and plain text mode test commands to readme? I can add a PR if you think that would be beneficial.

That would be great. Please!

PR #3262, opened 2022-10-26
Fix nodesOfType return type

acywatson:
This is a little more complicated than I thought, but I think the issue is that the TS compiler isn't convinced by the instanceof and __type checks that nodesOfType will actually return T and not a different subtype of LexicalNode. My solution is to cast to T, because these checks should verify that all nodesOfType are of the same subtype of LexicalNode at runtime.

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -1026,21 +1026,19 @@ export function setMutatedNode(
   }
 }
 
-export function $nodesOfType<T extends LexicalNode>(
-  klass: Klass<T>,
-): Array<LexicalNode> {
+export function $nodesOfType<T extends LexicalNode>(klass: Klass<T>): Array<T> {
   const editorState = getActiveEditorState();
   const readOnly = editorState._readOnly;
   const klassType = klass.getType();
   const nodes = editorState._nodeMap;
-  const nodesOfType = [];
+  const nodesOfType: Array<T> = [];
   for (const [, node] of nodes) {
     if (
       node instanceof klass &&
       node.__type === klassType &&
       (readOnly || node.isAttached())
     ) {
-      nodesOfType.push(node);
+      nodesOfType.push(node as T);
     }
   }
   return nodesOfType;


zurfyx (approved):

Stopa:
How about defining a type guard inside the function - 

```ts
function nodeIsKlass(node: LexicalNode): node is T {
  return node instanceof klass && node.__type === klassType && (readOnly || node.isAttached());
}
```

and then directly filtering out the relevant nodes -
```ts
return Array.from(nodes.values()).filter(nodeIsKlass);
```

Feels a bit cleaner to me as you don't need to `as` anything

thegreatercurve (approved):

acywatson:
> How about defining a type guard inside the function -
> 
> ```ts
> function nodeIsKlass(node: LexicalNode): node is T {
>   return node instanceof klass && node.__type === klassType && (readOnly || node.isAttached());
> }
> ```
> 
> and then directly filtering out the relevant nodes -
> 
> ```ts
> return Array.from(nodes.values()).filter(nodeIsKlass);
> ```
> 
> Feels a bit cleaner to me as you don't need to `as` anything
> 
> Or the same type guard can be used in the `for of` loop that's currently used to iterate as well

Not a bad idea! Is there an advantage in terms of type safety? To me it seems like a type guard here is just a slightly more verbose way of casting. 

Wrt to the Array.from + filter, I'm not sure about the performance of that in hot paths against the for/of iteration.

Stopa:
> Not a bad idea! Is there an advantage in terms of type safety? To me it seems like a type guard here is just a slightly more verbose way of casting. I'd grant the reusability would be a bit better (if we defined it outside the function).

I'm not 100% sure, but I don't think there's much advantage in terms of type safety other than type guard way is more explicit from the language itself and so might be less error-prone for moving-copying-pasting code than holding checks and casting separate.

acywatson:
I appreciate the feedback, but I don't really see the advantage of using a type guard here at this point. Casting is a language feature just as typeguards are a language feature. I'm going to go with the current approach for now.

PR #3268, opened 2022-10-26
Do not reconcile terminating linebreak for inline elements

fantactuka:
If link empty link (inline element) is the list child it would've had `<br>` added in dom


zurfyx (approved):
LGTM, need package-lock?

acywatson (approved):

PR #3273, opened 2022-10-26
Don't cache empty strings

acywatson:
https://github.com/facebook/lexical/pull/3247 revealed a gnarly bug in the style logic. We were sharing references across keys. this should only happen in the case of an empty string though, so this fixes it.


zurfyx (approved):
I see the problem but I'm not sure why `''` is the only collision..

acywatson:
> I see the problem but I'm not sure I understand why `''` is the only collision..

Yea you're actually right. I thought I could get away without copying the styles to a new object every time (this is what we used to do before) but we could have the same problem with any pre-existing key. Good call.

PR #3277, opened 2022-10-27
Add root theme class

fantactuka:
We already had `root` theme class but it was never applied to root node, it's quite useful for font styles (especially for nested editors which otherwise might inherit styles from its decorator)


acywatson (approved):
Nice

tylerjbainbridge (approved):

trueadm:
We removed it as it clashed with React from what I remember. Given you render the content editable and the class with React, Lexical might break this.

Issue #3278, opened 2022-10-27
Can't delete table paragraphs


zurfyx:
https://user-images.githubusercontent.com/193447/198363341-72c0ae69-9f85-4954-b5db-e680b71f75a9.mov

PR #3279, opened 2022-10-27
selection.modify to respect ShadowRoot

zurfyx:
Tables had some logic to override the deletion logic that was incomplete. This prevented new lines from being deleted.

However, when attempting to fix this (and switching the DELETE_CHARACTER command to be unhandled) it uncovered a bigger problem, the fact that selection can expand beyond a ShadowRoot and cause cells to be deleted. This shouldn't be the responsibility of tables, so I'm handling the ShadowRoot case appropriately in Selection but rather generic.

Closes https://github.com/facebook/lexical/issues/3278

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -9,7 +9,7 @@
 /* eslint-disable no-constant-condition */
 import type {EditorConfig, LexicalEditor} from './LexicalEditor';
 import type {RangeSelection} from './LexicalSelection';
-import type {Klass} from 'lexical';
+import type {Klass, RootNode} from 'lexical';
 
 import invariant from 'shared/invariant';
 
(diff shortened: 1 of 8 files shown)


packages/lexical/src/LexicalNode.ts
```diff
@@ -590,6 +590,20 @@ export class LexicalNode {
     return this.getTextContent().length;
   }
 
+  /**
+   * Returns the nearest ShadowRoot or RootNode
+   */
+  getRoot(): RootNode | ElementNode {
```
acywatson, line 596: getRootOrShadowRoot?

Sounds clunky but not sure about making this the same name (without $) as the method that always returns the true root.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1768,7 +1766,7 @@ export class RangeSelection implements BaseSelection {
     // from getTargetRanges(), and is also better than trying to do it ourselves
     // using Intl.Segmenter or other workarounds that struggle with word segments
     // and line segments (especially with word wrapping and non-Roman languages).
-    $moveNativeSelection(
+    moveNativeSelection(
```
acywatson, line 1769: y? Looks like we use $ methods inside

zurfyx: Hm, we don't

acywatson (approved):

packages/lexical/src/LexicalNode.ts
zurfyx: > TBH, wouldn't add LexicalNode#getRoot() and keep it as a util method, as our node classes already bloated -- @fantactuka 

Makes sense, moved away from LexicalNode

tylerjbainbridge (approved):

PR #3280, opened 2022-10-28
Fix pasting text near decorator nodes

Piliuta:
## Problem
Pasting text near decorator nodes loses the parent element and inserts children nodes inline as direct children of the root node.

**Steps to reproduce:**
https://user-images.githubusercontent.com/1575198/198649730-9be20a68-687d-47cc-b759-c6f26a2589fb.mov

(**Note:** There is a visual glitch where the cursor is set near the last HR whereas in fact it is set between the nodes. This is reported separately here https://github.com/facebook/lexical/issues/3156)

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste.spec.mjs
@@ -2804,4 +2804,74 @@ test.describe('CopyAndPaste', () => {
       `,
     );
   });
+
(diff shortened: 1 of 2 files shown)


Piliuta:
This is ready for review. `e2e-collab-prod` failed, but I think this is a flaky test.

acywatson:
I think this is fine. We're really trying to avoid making insertNodes more complicated until we can refactor it, but this seems like a reasonable bug fix rather than a new special case.

acywatson (approved):

PR #3282, opened 2022-10-28
Add mentions plugin example

akmarzhan1:
- [x] created a CodeSandbox example for the mentions plugin [here](https://codesandbox.io/s/lexical-mention-plugin-example-ojn42n)
- [x] added a new page and the example on the website (under “/docs/demos/plugins/mentions”)

image


tylerjbainbridge (approved):
This looks great! Thank you :)

acywatson:
Is it possible to make the iframe longer by default? You could see more of the code in IDE and scroll less maybe.

acywatson:
> @acywatson Yep, I increased the height of the iframe, is this better or should I increase it more?

I think it's good! We can always tinker with it later.

SalahAdDin:
Where is it? I can't find the url.

PR #3292, opened 2022-10-31
Optional decorator nodes selection with keyboard navigation

Piliuta:
## Problem

While navigating within the editor using keyboard keys (Arrow Left/Right) it automatically captures decorator nodes as a selection step. Whereas it is a cool feature, in some products such behavior can be not the desired one.
There is no easy way to override this behavior as it mostly requires reimplementing lexical's [selection modify](https://github.com/facebook/lexical/blob/a2d745fa10f593600938c0c0670f2c8b373935a5/packages/lexical/src/LexicalSelection.ts#L1704) handler to be able to handle text selection correctly.

## Solution

Add a new flag to allow optionally disable automatic selection as a step of keyboard selection navigation.

**Note:** This is different from `isIsolated` which prevents selection altogether.

### DIFF
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -734,7 +734,7 @@ declare export class DecoratorNode<X> extends LexicalNode {
   decorate(editor: LexicalEditor, config: EditorConfig): X;
   isIsolated(): boolean;
   isInline(): boolean;
-  isSelectable(): boolean;
+  isKeyboardSelectable(): boolean;
 }
 declare export function $isDecoratorNode(
   node: ?LexicalNode,
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -734,6 +734,7 @@ declare export class DecoratorNode<X> extends LexicalNode {
   decorate(editor: LexicalEditor, config: EditorConfig): X;
   isIsolated(): boolean;
   isInline(): boolean;
+  isSelectable(): boolean;
 }
 declare export function $isDecoratorNode(
   node: ?LexicalNode,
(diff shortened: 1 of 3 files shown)


trueadm:
Maybe it makes sense to call it `isKeyboardSelectable` instead? Might be confusing with those that still work with click selecting.

Piliuta:
> isKeyboardSelectable

Totally makes sense. Pushed an update.

Piliuta:
> Maybe it makes sense to call it `isKeyboardSelectable` instead? Might be confusing with those that still work with click selecting.

Totally makes sense. Pushed an update.

acywatson (approved):

zurfyx:
I have mixed feelings about this PR. I believe it's fine from a technical perspective but a11y wise it's not ideal. Hence, why I'm unsure it's a good idea to promote this pattern.

Screen readers will not announce (all) the content when skipped over, so ideally we can focus and guide the screen readers when focused on a DecoratorNode. This is in a way connected to Mentions and the reason why (unlike other editors) they are not implemented as tokens. Tokens are hard to read and edit and also behave inconsistency depending on the software and platform.

Piliuta:
> I have mixed feelings about this PR. It's fine from a technical perspective but a11y wise it's not ideal. Hence, why I'm unsure it's a good idea to promote this pattern.
> 
> Screen readers will not announce (all) the content when skipped over, so ideally we can focus and guide the screen readers when focused on a DecoratorNode. This is in a way connected to Mentions and the reason why (unlike other editors) they are not implemented as tokens. Tokens are hard to read and edit and also behave inconsistency depending on the software and platform.

@zurfyx thanks for the comment. I am not sure that understand - what's your proposal here?

zurfyx:
Thanks for the PR! After a discussion with the team, we are OK with Lexical to provide the flexibility to do so

PR #3293, opened 2022-10-31
Preserve selection for pending state

fantactuka:
In non-headless mode there's a separate logic that either reuses previous selection or relies on dom. For headless we just copy previous selection


packages/lexical/src/LexicalUpdates.ts
```diff
@@ -809,6 +809,8 @@ function beginUpdate(
   try {
     if (editorStateWasCloned && !editor._headless) {
       pendingEditorState._selection = internalCreateSelection(editor);
+    } else if (editor._headless && currentEditorState._selection != null) {
+      pendingEditorState._selection = currentEditorState._selection.clone();
```
trueadm, line 813: If the editor state wasn't cloned then should you be mutating it?

PR #3294, opened 2022-10-31
Preserve selection for pending state

fantactuka:
Follow up for https://github.com/facebook/lexical/pull/3293#discussion_r1009527659

### DIFF
--- a/packages/lexical/src/LexicalUpdates.ts
+++ b/packages/lexical/src/LexicalUpdates.ts
@@ -807,10 +807,14 @@ function beginUpdate(
   activeEditor = editor;
 
   try {
-    if (editorStateWasCloned && !editor._headless) {
-      pendingEditorState._selection = internalCreateSelection(editor);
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalUpdates.ts
```diff
@@ -807,10 +807,14 @@ function beginUpdate(
   activeEditor = editor;
 
   try {
-    if (editorStateWasCloned && !editor._headless) {
-      pendingEditorState._selection = internalCreateSelection(editor);
-    } else if (editor._headless && currentEditorState._selection != null) {
-      pendingEditorState._selection = currentEditorState._selection.clone();
+    if (editorStateWasCloned) {
+      if (editor._headless) {
+        if (currentEditorState._selection != null) {
```
trueadm, line 812: Hoist? 
`const currentSelection = currentEditorState._selection;`

PR #3295, opened 2022-10-31
[0.6] Make RichTextPlugin ErrorBoundary mandatory

zurfyx:
Remove the TODO, make it mandatory now. For the sake of simplicity, I also introduced LexicalErrorBoundary that should work for newcomers.

Screen Shot 2022-10-31 at 11 58 46 AM


acywatson (approved):

packages/lexical-react/src/LexicalErrorBoundary.tsx
```diff
@@ -16,17 +14,21 @@ type Props = {
   onError: (error: Error) => void;
 };
 
-export default function ErrorBoundary({children, onError}: Props): JSX.Element {
+export default function LexicalErrorBoundary({
```
strdr4605, line 17: Why was it decided that Error will be exported as the default in contrary to other plugins from `@lexical/react/**`?
cc @zurfyx

Issue #2647, opened 2022-07-15
Bug: Quote: Conversion between paragraph and quotes with links creates extra elements


Piliuta:
Lexical version: 0.3.7

## Steps To Reproduce

1. Open the playground page https://playground.lexical.dev/
2. Create any paragraph with links in it
3. Select all text of the paragraph (Note: should a range selection, i.e. not collapsed)
4. Apply "Quote" format - (1 + x links) quotes created
5. Select text on the Quote and turn it into Normal text - 

https://user-images.githubusercontent.com/1575198/179219061-7fdfbc96-4b00-4e87-a6c4-cbfb3f0427d0.mov

## The current behavior
The paragraph is converted to Quote with extra Quotes above it. _Every link creates an extra Quote_
The Quote is converted into multiple paragraphs

## The expected behavior
The paragraph is converted to a single Quote. The quote is converted to one paragraph.

Notes: Looks like the problem is caused by the `$wrapLeafNodesInElements` method [somewhere here](https://github.com/facebook/lexical/blob/645227ebd0f61942d48897b03b0701a4da884508/packages/lexical-selection/src/index.ts#L667-L698) that generates extra elements.

acywatson:
Related to #2648

Piliuta:
I was able to find where the problem comes from:

When we have a text selection like `text [link](google.com)` and wrap it using the `$wrapLeafNodesInElements` function:
- it loops through all children elements (including nested ones), in the examples the list of nodes is
```
[
  { type: 'text', text: 'text ' },
  { type: 'link', url: 'google.com ' },
  { type: 'text', text: 'link ' },
]
```
- [1st loop] it gets children from the parent of the first node, appends them to the wrapped element and changes parent relations on them to the new node
https://github.com/facebook/lexical/blob/5a1c20da44eba7968523c40f076ba6ba43fed1bd/packages/lexical-selection/src/index.ts#L688-L689
- [2nd loop] does not do anything because the link node is not considered a leaf (I think this is a bug too, but this is not the source of the problem)
https://github.com/facebook/lexical/blob/5a1c20da44eba7968523c40f076ba6ba43fed1bd/packages/lexical-selection/src/index.ts#L677
or otherwise, it would be ignored because this is a first-level leaf (because already wrapped)
https://github.com/facebook/lexical/blob/5a1c20da44eba7968523c40f076ba6ba43fed1bd/packages/lexical-selection/src/index.ts#L678
[...]

PR #3301, opened 2022-11-01
Fix wrapping nodes with nested leaf nodes

Piliuta:
Fixes https://github.com/facebook/lexical/issues/2647

The wrapping nodes function should skip nested leaf nodes if their parents are already wrapped (appended).

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -194,7 +194,7 @@ export function $wrapNodesImpl(
     }
   }
 
-  const movedLeafNodes: Set<NodeKey> = new Set();
+  const movedNodes: Set<NodeKey> = new Set();
 
   // Move out all leaf nodes into our elements array.
   // If we find a top level empty element, also move make
@@ -210,7 +210,7 @@ export function $wrapNodesImpl(
     if (
       parent !== null &&
       $isLeafNode(node) &&
-      !movedLeafNodes.has(node.getKey())
+      !movedNodes.has(node.getKey())
     ) {
       const parentKey = parent.getKey();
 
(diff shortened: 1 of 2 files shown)


packages/lexical-selection/src/range-selection.ts
```diff
@@ -225,6 +225,9 @@ export function $wrapNodesImpl(
         parent.getChildren().forEach((child) => {
           targetElement.append(child);
           movedLeafNodes.add(child.getKey());
+          if ($isElementNode(child)) {
+            child.getChildrenKeys().forEach((key) => movedLeafNodes.add(key));
```
zurfyx, line 229: L206 doesn't look right.
1. It doesn't account for ShadowRoot
2. It doesn't account for nested structures, just the first layer.

This impacts your logic because then it may make sense to just check whether the parent has been moved (when processing the node). This way we can support multi-depth efficiently (rather than just the first layer).

That said, as it is now, I have to admit it solves the existing problem given the existing incomplete code. Would you at least mind adding a comment on top of the loop so that we have an easier time next time someone attempts to fix this? 

Can be just your PR comment `// Skip nested leaf nodes if the parent has already been moved`

L227 is also a lie; it may contain non-leaf nodes. The variable should be renamed

zurfyx (approved):
Thank you, added a comment, but overall LGTM!

packages/lexical-selection/src/range-selection.ts
Piliuta: > L206 doesn't look right.
> 
> 1. It doesn't account for ShadowRoot
> 2. It doesn't account for nested structures, just the first layer.
> 
> This impacts your logic because then it may make sense to just check whether the parent has been moved (when processing the node). This way we can support multi-depth efficiently (rather than just the first layer).

I wasn't able to find a different use case to test it, so ended up checking just that one level. I was proposing a different fix in the comment here https://github.com/facebook/lexical/issues/2647#issuecomment-1205121519 (second proposal) which checks whether the parent was moved, however it also captures one level only since I am not sure how to loop through all levels of a node clearly. If you can give me some tips on how to handle it, I can try to have a better fix.


> That said, as it is now, I have to admit it solves the existing problem given the existing incomplete code. Would you at least mind adding a comment on top of the loop so that we have an easier time next time someone attempts to fix this?
> 
> Can be just your PR comment `// Skip nested leaf nodes if the parent has already been moved`

Added

> L227 is also a lie; it may contain non-leaf nodes. The variable should be renamed

Updated

zurfyx:
Thank you, appreciate the unit test!

PR #3324, opened 2022-11-05
Add list plugin example

akmarzhan1:
- [x] created a CodeSandbox example for the list plugin [here](https://codesandbox.io/s/lexical-list-plugin-example-00d642)
- [x] added a new page and the example on the website (under “/docs/demos/plugins/lists”)

image


akmarzhan1:
I have only added the bullet list and numbered list because there were some issues with the checklist (although I imported all of the necessary components, checklists were showing up as bullet lists) and I wasn't sure whether I needed to add the examples from the list plugin only or also from the checklist plugin. 

Let me know if I need to fix anything.

fantactuka:
> I have only added the bullet list and numbered list because there were some issues with the checklist (although I imported all of the necessary components, checklists were showing up as bullet lists) and I wasn't sure whether I needed to add the examples from the list plugin only or also from the checklist plugin.

Checklist requires [some css](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css#L319) to render "checkbox"

acywatson (approved):
Not sure about the package-lock change - did you add a new package?

akmarzhan1:
> Not sure about the package-lock change - did you add a new package?

Also not sure where the changes came from since I didn't add new packages. Should I revert those differences?

akmarzhan1:
> some css

That makes sense! I totally missed these, will add and update

akmarzhan1:
@fantactuka added the check list css and now it works:
image

@acywatson reverted the package-lock changes.

Let me know if I need to fix anything else.

PR #3329, opened 2022-11-08
Filter out mixed files + content copy pasting

fantactuka:
When copying, MSWord adds content as an image to the clipboard (apart from various text formats data), and our file copy/pasting logic which is executed before text content parsing inserts that image instead of actual text content:

https://user-images.githubusercontent.com/132642/200468353-3922b892-8045-43eb-86bd-45b445798900.mov

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -407,7 +407,7 @@ async function onCutForRichText(
 // control this with the first boolean flag.
 export function eventFiles(
   event: DragEvent | PasteCommandType,
-): [boolean, Array<File>] {
+): [boolean, Array<File>, boolean] {
   let dataTransfer: null | DataTransfer = null;
   if (event instanceof DragEvent) {
     dataTransfer = event.dataTransfer;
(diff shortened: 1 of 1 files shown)


packages/lexical-rich-text/src/index.ts
```diff
@@ -416,9 +416,14 @@ export function eventFiles(
...
   if (dataTransfer === null) {
-    return [false, []];
+    return [false, [], false];
   }
-  return [dataTransfer.types.includes('Files'), Array.from(dataTransfer.files)];
+
+  const types = dataTransfer.types;
+  const hasFiles = types.includes('Files');
+  const hasContent =
+    types.includes('text/html') || types.includes('text/plain');
```
fantactuka, line 425: @zurfyx Not sure which check is better: text/html | plain/text or just check if there's anything else except 'Files'

thegreatercurve (approved):

packages/lexical-rich-text/src/index.ts
zurfyx: I'd say what you have now is best as many apps including Mac Finder tend to include a variety of formats

zurfyx (approved):
Thank you Maksim!

PR #3336, opened 2022-11-08
feat: make icons and popups dynamic & change file structure

im-adithya:
- [x] Changes code structure
- [x] Changes popup and icon when Lexical is found
- [x] Opens Lexical in Devtools only when there's Lexical


packages/lexical-devtools/src/background/index.ts
```diff
@@ -10,9 +10,15 @@
...
   number,
-  {contentScriptPort?: chrome.runtime.Port; reactPort?: chrome.runtime.Port}
+  {
+    contentScriptPort?: chrome.runtime.Port;
+    reactPort?: chrome.runtime.Port;
+    devtoolsPort?: chrome.runtime.Port;
+  }
 > = {};
 
+const IS_FIREFOX: boolean = navigator.userAgent.indexOf('Firefox') >= 0;
```
thegreatercurve, line 20: I think we have a common util for this in `packages/shared`

packages/lexical-devtools/src/background/index.ts
```diff
@@ -28,20 +34,26 @@ chrome.runtime.onConnect.addListener((port: chrome.runtime.Port) => {
       return;
     }
 
-    if (message.name === 'init' && message.type === 'FROM_APP') {
+    if (message.name === 'init' && port.name === 'react-app') {
```
thegreatercurve, line 37: I'm presuming we're taking some of this logic from React DevTools? If so, is this the same port name and could it cause a clash?

thegreatercurve (approved):
LGTM to me! Thanks @im-adithya! Can you maybe update the PR description to include some screenshots of the panel looks like without Lexical etc..

packages/lexical-devtools/src/background/index.ts
im-adithya: No, this suggests that the port is established from the devtools panel _app_ (built with react, hence react-app).

We have two ports for the devtools now, one ('devtools') is from `main.js` used to communicate with the background by adding listeners to update whenever the Editor changes, and the other('react-app') is from the panel we are creating in the chrome devtools used for highlighting and dehighlighting

packages/lexical-devtools/src/background/index.ts
im-adithya: That is linked to DOM:
https://github.com/facebook/lexical/blob/bcee9cc89beaea3db8e83baa37b57efea7e0bcc1/packages/shared/src/environment.ts#L27-L28
So I thought of using this, should I add this to `packages/shared`?

thegreatercurve: Ah okay! And you don't want to check if the document exists because you're in a Developer Tab window?

im-adithya: Yes!

PR #3342, opened 2022-11-09
Add $splitNode & $copyNode

fantactuka:
1. `$splitNode` would allow splitting current tree up until root or shadow root and enhance insertions. As of now `$insertNodeoNearestRoot` takes nearest top-level child and appends node after it, which is not very precise. Here's an example with nested lists:
```
- Top level
    - Text before <split here> text after
    - More nested items
- More top level items
```
with selection on `<split here>`, `$insertNodeoNearestRoot` would insert node after the whole list. `$splitNode` aims to allow splitting nested trees to insert where user would expect, e.g. calling `$splitNode(listItem, 1)` would split it into two separate lists:

```
- Top level
    - Text before

-
    - text after
    - More nested items
- More top level items
```

It should address limitation of #2322 and allow splitting any arbitrary nested structure vs only dealing with 1-2 levels

2. `$copyNode` is needed for `$splitNode` implementation, and is already used internally (with a hacky workaround), so adding it to the official API

Follow up PR will update `$insertNodeoNearestRoot` to use splitText & splitNode to insert block level nodes more precise

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -8,12 +8,14 @@
  */
 
 import {
+  $copyNode,
   $createParagraphNode,
   $getRoot,
   $getSelection,
   $isElementNode,
   $isNodeSelection,
   $isRangeSelection,
+  $isRootOrShadowRoot,
   $isTextNode,
   $setSelection,
   createEditor,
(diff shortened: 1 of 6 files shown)


fantactuka:
One thing I'm not 100% sure about is whether `$splitNode` should guaranteed to return two new trees, mainly for the cases when split happens on the edge of the node. E.g., the simplest scenario is calling splitNode on empty paragraph, as of now it'll return two empty paragraphs. Alternatively, if the left tree is empty and each level is element node with single child (except top-most, which is empty) we can delete it.

packages/lexical-utils/src/index.ts
```diff
@@ -441,3 +443,45 @@ export function $wrapNodeInElement(
   elementNode.append(node);
   return elementNode;
 }
+
+export function $splitNode(
```
acywatson, line 447: should it be splitElement?

packages/lexical-utils/src/index.ts
```diff
@@ -441,3 +443,45 @@ export function $wrapNodeInElement(
...
+  node: ElementNode,
+  offset: number,
+): [ElementNode | null, ElementNode] {
+  let startNode = node.getChildAtIndex(offset) as LexicalNode;
```
acywatson, line 451: How does it work if you want to split the tree at a specific offset in a TextNode? I guess you have to first splitText, then go to the parent and $splitNode and then put the pieces of the text into the correct new parent.

packages/lexical-utils/src/index.ts
```diff
@@ -441,3 +443,45 @@ export function $wrapNodeInElement(
...
+  if (startNode == null) {
```
acywatson, line 452: Why cast and then check null here? Can we just get rid of the cast?

packages/lexical-utils/src/index.ts
fantactuka: Yep, do that in https://github.com/facebook/lexical/pull/3344

packages/lexical-utils/src/index.ts
fantactuka: Fair point, initially id had support for text nodes (with splitText), but I thought it might be too much for a single method and it would be better to see if it'll be ever needed (in addition to $insertNodeToNearestRoot use case)

acywatson (approved):

zurfyx (approved):

PR #3344, opened 2022-11-09
Improve $insertNodeToNearestRoot

fantactuka:
Using `$splitNodes` allows insertion in the middle of the text with nodes tree being split up until nearest root:

https://user-images.githubusercontent.com/132642/200885233-6a0f0983-63fe-4be6-b12a-b1ef1a9c5946.mov

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -8,12 +8,14 @@
  */
 
 import {
+  $copyNode,
   $createParagraphNode,
   $getRoot,
   $getSelection,
   $isElementNode,
   $isNodeSelection,
   $isRangeSelection,
+  $isRootOrShadowRoot,
   $isTextNode,
   $setSelection,
   createEditor,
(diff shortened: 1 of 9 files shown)


acywatson:
I think there are places in lists where we can use $splitNode too

packages/lexical-utils/src/index.ts
```diff
@@ -416,21 +416,38 @@ export function $restoreEditorState(
...
+    let splitNode: ElementNode;
+    let splitOffset: number;
+
+    if ($isTextNode(focusNode)) {
+      splitNode = focusNode.getParentOrThrow();
+      splitOffset = focusNode.getIndexWithinParent();
+      if (focusOffset > 0) {
+        splitOffset += 1;
+        focusNode.splitText(focusOffset);
+      }
```
zurfyx, line 431: Nit; makes me think that the splitNode API should probably consider any kind of node and you can do your own best judgement there, like computing the TextNode offset or other nodes like LineBreakNode

zurfyx (approved):

packages/lexical-utils/src/index.ts
fantactuka: Yeah, I think we can roll it out with elements support only and see if it'll be useful elsewhere (apart from $insertNodeToNearestRoot). But based on your and Acy's comments I think it would make sense to have it, will do a bit later as this part was more time-pressing

trueadm:
This causes many issues when called on the RootNode.

PR #3350, opened 2022-11-10
Change the plugin to `OnChangePlugin` in the React Editor State Example

atgctg:
I just noticed that the `RichTextPlugin` also has a `Lexical` prefix, so this was probably done intentionally.

Feel free to close this PR.  (Just found importing `LexicalOnChangePlugin` confusing at first.)


zurfyx:
Thank you. Yes, all exported modules should always have the `Lexical` prefix, otherwise it's a mistake in our side. The only time when we don't do this is in our playground and that's because it just resembles a user-facing app that won't ever be exposed to NPM.

Issue #2648, opened 2022-07-15
Bug: Block selection from start to end affects next block when converted to other blocks


Piliuta:
Lexical version: 0.3.7

## Steps To Reproduce

1. Open the playground page https://playground.lexical.dev/
2. Create two paragraphs
3. Select the first one (from the very end, e.g. use double-click)
4. Apply any block format (e.g. heading or quote)

https://user-images.githubusercontent.com/1575198/181064038-183b3c2c-ddf3-4b4c-93b0-8af03805e567.mov

## The current behavior
The selected paragraph and the paragraph next to it are converted into the new format.

## The expected behavior
Only the selected paragraph is converted into the format. The paragraph next to it remains as is.

acywatson:
This is interesting - there's actually only one paragraph in the editor state - it looks like a new one is actually inserted before the existing one when you apply the QuoteNode formatting.

This is probably an issue with $wrapLeafNodesInElements, which has always been a bit buggy, especially around inline elements.

Piliuta:
@acywatson I just realized that I attached the wrong video to this issue (already updated). This issue is different from what I posted here https://github.com/facebook/lexical/issues/2647 and _I feel like they might actually be different bugs_. Could you take another look?

zeeyang:
Triple click block selection includes the line break node, which puts selection range end at the beginning of next paragraph. $wrapLeafNodesInElements just needs to be smarter about not including the next paragraph if selection range is at the beginning.

Piliuta:
@acywatson I am thinking to make a fix for this issue. I can see at least two different approaches here:
1. Fix the function that wraps nodes to be smarter as @zeeyang [proposed](https://github.com/facebook/lexical/issues/2648#issuecomment-1211028229). 
2. Change how lexical handles selection via triple-click overriding default behavior to select only the block from start to end. 

I see different editors handle it differently, so not sure what's the preferable solution here.

Any preferences or concerns here?

Piliuta:
> After triple click I would expect the carriage return to be selected. i.e. delete after triple click should completely remove the paragraph including carriage return. I tested on several editors and this seems to be the universal expectation. I think option 2 goes against this expected behavior.

Tested in JIRA, Google Docs, Coda, Discord products, and Slate framework - neither of these remove the carriage return 🤷🏻‍♂️

zeeyang:
[...]

PR #3362, opened 2022-11-14
Do not capture trailing node while wrapping nodes when it is on the edge of the selection

Piliuta:
Resolves https://github.com/facebook/lexical/issues/2648

TODOs:
- [ ] Fix unit tests
- [x] Add backward direction selection case

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -68,6 +68,37 @@ export function $wrapNodes(
   createElement: () => ElementNode,
   wrappingElement: null | ElementNode = null,
 ): void {
+  const selectionEnd = selection.isBackward()
+    ? selection.anchor
+    : selection.focus;
+  if (
+    $isRangeSelection(selection) &&
+    !selection.isCollapsed() &&
+    selectionEnd.offset === 0
+  ) {
+    let trailingNode = selectionEnd.getNode();
+    while (trailingNode !== null) {
+      const parentOffset = trailingNode.getIndexWithinParent();
+      trailingNode = trailingNode.getParentOrThrow();
+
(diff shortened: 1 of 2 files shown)


packages/lexical-selection/src/range-selection.ts
```diff
@@ -53,6 +53,38 @@ function isPointAttached(point: Point): boolean {
...
+    let trailingNode = selectionEnd.getNode();
+    while (trailingNode !== null) {
+      const parentOffset = trailingNode.getIndexWithinParent();
+      trailingNode = trailingNode.getParentOrThrow();
+
+      if ($isRootOrShadowRoot(trailingNode)) {
+        if (selection.isBackward()) {
+          const collapsedEndSelection = selection.clone();
+          collapsedEndSelection.modify('move', false, 'character');
+          collapsedEndSelection.modify('move', true, 'character');
```
Piliuta, line 74: This is a hacky workaround to move the anchor backward. The first `modify` is needed to deselect and the second is to actually move.
I was trying to collapse (as `collapsedEndSelection.focus = collapsedEndSelection.anchor`) and move, but that did not work somehow, and was moving the focus instead of the anchor.

Piliuta:
This fix works in both selection directions, but I am having a hard time fixing tests. Unit tests are failing because it is missing some mocks for the dom selection function and it is hard to understand what it needs. e2e are probably browser discrepancies?

@thegreatercurve I noticed you self-assigned this issue https://github.com/facebook/lexical/issues/2648
Assuming you are going to work on it, this PR could be helpful. Likely does need to much changes.

thegreatercurve:
I've not personally had time yet to look at the underlying issue, but will consult this PR when I do so. I'm going to close it for the moment though due to becoming stale, and the fact it's still WIP.

PR #3368, opened 2022-11-16
Move default language setting to Tokenizer

saint1991:
### ~1. Add a preserve feature to CodeNode~

~In the original implementation, CodeNode is removed when the backward deletion occurs at the beginning of the Node.~
~In some use cases, implementing a code editor, such behavior is not needed.~
~The "Preserve" setting is for making it controllable.~

### 2. Move the default language setting to Tokenizer 

The former PR #3243 enabled to use arbitrary Tokenizer other than Prism and 'javascript' is not always appropriate to the default so make it a Tokenizer setting.

This PR contains 2 changes but I'm happy to divide them if required.

### DIFF
--- a/packages/lexical-code/flow/LexicalCode.js.flow
+++ b/packages/lexical-code/flow/LexicalCode.js.flow
@@ -22,7 +22,7 @@ import {ElementNode, TextNode} from 'lexical';
 declare export class CodeNode extends ElementNode {
   static getType(): string;
   static clone(node: CodeNode): CodeNode;
-  constructor(key?: NodeKey): void;
+  constructor(language: ?string, key?: NodeKey): void;
   createDOM(config: EditorConfig): HTMLElement;
   updateDOM(prevNode: CodeNode, dom: HTMLElement): boolean;
   insertNewAfter(
@@ -33,7 +33,7 @@ declare export class CodeNode extends ElementNode {
   setLanguage(language: string): void;
   getLanguage(): string | void;
 }
-declare export function $createCodeNode(language?: string): CodeNode;
+declare export function $createCodeNode(language: ?string): CodeNode;
 declare export function $isCodeNode(
   node: ?LexicalNode,
 ): boolean %checks(node instanceof CodeNode);
@@ -71,6 +71,7 @@ export interface Token {
   content: TokenContent;
 }
 export interface Tokenizer {
+  defaultLanguage: string;
   tokenize(code: string, language?: string): (string | Token)[];
 }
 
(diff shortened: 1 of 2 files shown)


packages/lexical-website/docs/demos/plugins/embeds/_category_.yml
```diff
@@ -1,2 +1,2 @@
 label: 'Embed Plugins'
-position: 1
\ No newline at end of file
+position: 1
```
saint1991, line 2: just apply pretteir

saint1991:
I have no idea how to fix the failed e2e tests...
It seems to be occurring in the main branch randomly.

thegreatercurve (changes requested):
@saint1991 The default language changes seems fine, but I think we'd want to leave out the preserve flag. Users can just implement this themselves by extending the CodeNode and overwriting the collapseAtStart method.

saint1991:
@thegreatercurve 

Thanks for your kind review!
Thanks to #3367, the code something like below works as  I expected.

Thank you so much for the great work!
I've removed commit 1. which is no longer needed as you mentioned.

```


export class CodeEditorNode extends CodeNode {
  static getType(): string {
    return 'code-editor';
  }

  collapseAtStart(): true {
    return true;
  }
}

export const $createCodeEditorNode = (language?: string | null | undefined): CodeEditorNode => {
  return new CodeEditorNode(language);
}

const editor = createEditor({
  nodes: [
    CodeEditorNode,
    CodeNode: {
      replace: CodeNode, 
      with: (node: CodeNode) => $createCodeEditorNode(node.getLanguage())
    },
    CodeHighlightNode,
  ]
})
```

packages/lexical-code/flow/LexicalCode.js.flow
```diff
@@ -33,7 +33,7 @@ declare export class CodeNode extends ElementNode {
   setLanguage(language: string): void;
   getLanguage(): string | void;
 }
-declare export function $createCodeNode(language?: string): CodeNode;
+declare export function $createCodeNode(language?: string | null): CodeNode;
```
thegreatercurve, line 36: This should just be enough:
```suggestion
declare export function $createCodeNode(language: ?string): CodeNode;
```

thegreatercurve (approved):

fantactuka:
I'm curious if it's better to have default language as a param to `registerCodeHighlighting` and `CodeHighlightingPlugin`. Then you can have single tokenizer but custom default language based on specific surface needs

saint1991:
I think it's also OK, but I put the setting here because the supported languages depend on Tokenizer's implementation. The setting 'javascript' is one of the proper choices for PrismTokenizer because Prism supports it.

In terms of the use of the library, the difference is that users should specify the default language even when they use PrismTokenizer by default.

PR #3369, opened 2022-11-16
Add collapsible container plugin example

akmarzhan1:
- [x] created a CodeSandbox example for the collapsible container plugin [here](https://codesandbox.io/s/lexical-collapsible-container-plugin-example-90nt6t)
- [x] added a new page and the example on the website (under “/docs/demos/plugins/collapsible-container”)

Collapsible Container Plugin

### DIFF
--- a/packages/lexical-website/docs/demos/plugins/collapsible-container.md
+++ b/packages/lexical-website/docs/demos/plugins/collapsible-container.md
@@ -1,0 +1,16 @@
+---
+id: "collapsible-container"
+title: "Collapsible Container Plugin"
+sidebar_label: "Collapsible Container Plugin"
+---
+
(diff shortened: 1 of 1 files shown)


thegreatercurve:
@akmarzhan1 I'm not sure if we necessarily need the header with the text "Collapsible Container Plugin Example. Note: this build shows how to use collapsible containers", and the grey background around the text editor. Is that visible within the iframe embed in the docs? 

LGTM otherwise!

thegreatercurve (approved):

akmarzhan1:
> @akmarzhan1 I'm not sure if we necessarily need the header with the text "Collapsible Container Plugin Example. Note: this build shows how to use collapsible containers", and the grey background around the text editor. Is that visible within the iframe embed in the docs?
> 
> LGTM otherwise!

Yep, it is visible, would you prefer if I removed these headers from all of the examples?

PR #3371, opened 2022-11-17
LinkPlugin to support link paste

zurfyx:
Looks like a desirable feature, at least based on the usage inside Meta. Let's make this built-in into the LinksPlugin

https://user-images.githubusercontent.com/193447/202585095-a04f669b-51af-450a-8ab1-56dccc0d4a6e.mov


acywatson (approved):
Could also be a separate plugin, but I think it can make sense to do it this way too.

PR #3376, opened 2022-11-18
Enable CodeHighlightNode to hold token metadata

saint1991:
Tokenize process sometimes produces some other metadata than the token type.
For example, which 2 tokens are pair of braces.

To enable keeping such information at CodeHighlightNode adds a field to hold serializable metadata.

### TODO

- [ ] test


thegreatercurve:
Thanks for the contribution @saint1991!

Unfortunately, I don't think this is really needed any more with the introduction of allowing users to replace core nodes themselves #3367. This should give you the ability to create your own node from CodeNode and then add in any additional serialisable data.

Issue #3349, opened 2022-11-10
Simplify placeholder creation for custom layouts


zurfyx:
From internal feedback the placeholder creation is complicated. It is particularly hard when custom nodes (beyond a simple TextNode) are involved. For example, our Workplace Post Composer structure that contains a Header + Body as a placeholder. We should research a better and generic way to simplify this problem.

> Text placeholder needs lots of extra moves to setup correctly. It gonna need to be simplified?? Maybe building it atop of paragraph node and absolute positioning?

thegreatercurve:
#3379

PR #3379, opened 2022-11-18
⏸ [0.7] Customize Placeholder visibility

zurfyx:
**Edit:**
Revisited the whole PR based on @trueadm and @thegreatercurve's comments. Now it's a 0.7 breaking change and the placeholder makes it possible to pass a callback functio with an editable param to allow customization.

---

Provi @ Discord
> Is there a way to show the placeholder when isEditable is false? I want to disable an input and show in placeholder that it is disabled
> looks like it's intentionally disabled

That's a valid concern. The reason why I removed it in #3030 was to promote the use Lexical for long-content form read-only views where the placeholder wouldn't make much sense. However, it still makes sense on shorter form content, and in fact, the `<input disabled>` and `<textarea disabled>`  shows the placeholder when disabled, which makes me reconsider whether that was a good solution in the first place.

Unfortunately, I can't just change the current behavior without causing a breaking change which would be undesirable but we can still provide the option to customize the non-editable placeholder via a new {Plain/Rich}TextPlugin property (`placeholderNonEditable`) while preserving the original behavior.

That said, this is likely to change in the not so distant future. We are reconsidering how we package RichText and PlainText to provide further flexibility while making it easier for newcomers to bootstrap a rich text editor quickly.


thegreatercurve:
> Unfortunately, I can't just change the current behavior without causing a breaking change which would be undesirable but we can still provide the option to customize the non-editable placeholder via a new {Plain/Rich}TextPlugin property (placeholderNonEditable) while preserving the original behavior.

By adding another new property, we're just creating 2 breaking changes we need to fix. I'd say just make `placeholder` always visible in a disabled state, regardless of the length of text, like it is on native `textareas` and `inputs`, and then do a 0.7 release.

zurfyx:
Revisited the whole PR based on @trueadm and @thegreatercurve's comments. Now it's a 0.7 breaking change and the placeholder makes it possible to pass a callback functio with an `editable` param to allow customization.

packages/lexical-playground/src/Editor.tsx
```diff
@@ -171,7 +171,7 @@ export default function Editor(): JSX.Element {
                 contentEditable={
                   <ContentEditable className="TableNode__contentEditable" />
                 }
-                placeholder={''}
```
trueadm, line 174: This should still work, right?

zurfyx: Why not just `placeholder={null}` instead?

thegreatercurve:
Closes #3349

packages/lexical-playground/src/Editor.tsx
trueadm: Strings should be allowed too.

trueadm (approved):

PR #3392, opened 2022-11-21
Add LexicalNodeEventPlugin to @lexical/react

trueadm:
This plugin makes it easy and simple to attach event listeners to specific types of Lexical node without needing to manually do the wiring.

### DIFF
--- a/packages/lexical-playground/src/ui/Modal.tsx
+++ b/packages/lexical-playground/src/ui/Modal.tsx
@@ -48,10 +48,11 @@ function PortalImpl({
         onClose();
       }
     };
-    if (modalRef.current !== null) {
-      modalOverlayElement = modalRef.current?.parentElement;
+    const modelElement = modalRef.current;
+    if (modelElement !== null) {
+      modalOverlayElement = modelElement.parentElement;
       if (modalOverlayElement !== null) {
-        modalOverlayElement?.addEventListener('click', clickOutsideHandler);
+        modalOverlayElement.addEventListener('click', clickOutsideHandler);
       }
     }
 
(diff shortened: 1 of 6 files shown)


packages/lexical-playground/src/ui/Modal.tsx
```diff
@@ -48,10 +48,11 @@ function PortalImpl({
         onClose();
       }
     };
-    if (modalRef.current !== null) {
-      modalOverlayElement = modalRef.current?.parentElement;
+    const modelElement = modalRef.current;
+    if (modelElement !== null) {
+      modalOverlayElement = modelElement.parentElement;
```
trueadm, line 53: I just removed experimental syntax as part of this PR because it annoyed me :P

packages/lexical-react/src/LexicalNodeEventPlugin.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  useLayoutEffect(() => {
+    return editor.registerMutationListener(nodeType, (mutations) => {
+      editor.getEditorState().read(() => {
+        for (const [key, mutation] of mutations) {
+          const element: null | HTMLElement = editor.getElementByKey(key);
+
+          if (mutation === 'created' && element !== null) {
+            element.addEventListener(eventType, (event: Event) => {
+              listenerRef.current(event, key);
```
acywatson, line 36: Since the listener is a ref, i guess this also solves the issue of stale React state values in the click handler, right?

trueadm: Yep

fantactuka: Should probably pass editor instance as well

acywatson (approved):

PR #3396, opened 2022-11-22
refactor: React composer initialConfig to exported type

SimonProper:
Extracted the initialConfig  for LexicalComposer components from the component props into a separate type that is also exported.  

This allows you to import and use the type to get autocompletion and typescript assistant when creating a new editor.  
As some of the properties are required in the type, Typescript will warn you directly instead of errors in the client. 
Example of use: 
```typescript
import { InitialConfigType } from "@lexical/react/LexicalComposer";

const initialConfig: InitialConfigType = {
    namespace: "MyEditor",
    onError,
    theme: editorTheme,
    nodes: [HeadingNode, ImageNode],
    editorState: InitialEditorState,
}
```
Personally found this especially helpful when trying to set the editorState property. That one is a bit tricky and Typescript helps at least a bit from passing the wrong types.


SimonProper:
@zurfyx tried to copy and paste around the types in the flow file, haven't use flow before so please have a look that its correct. 

Also updated the PR with latest changes on main.

zurfyx:
Thanks!

PR #3399, opened 2022-11-23
Add more invariants around insertion of text nodes into root

trueadm:
It shouldn't be possible to insert text nodes into the root node. We can streamline this in the future with schemas but until then we should make it fail hard if you attempt to as a developer.


zurfyx:
Is this PR empty? I can see you only updates codes.json

zurfyx (approved):

Issue #3365, opened 2022-11-15
Bug: `Expected valid LexicalSelection` when trying to copy from nested editor


zignis:
Lexical version: Latest (playground)

## Steps To Reproduce

1. Visit [playground](https://playground.lexical.dev)
2. Add a sample image and fill in some random caption
3. Copying the text throws this exception: `Expected valid LexicalSelection` in the browser console.

Link to code example: -

## The current behavior
An exception is thrown when trying to copy some text off the nested editor.

## The expected behavior
Should be able to copy content from nested editors. Seems to be a bug with `@lexical/selection` module.

PR #3401, opened 2022-11-23
Apply event propagation between all event types

trueadm:
Fixes https://github.com/facebook/lexical/issues/3365.


fantactuka:
Similar issue was fixed for cutting in nested editor via checking for null selection (https://github.com/facebook/lexical/blob/102b01d808f53b560ddbd3f7dcc53ba34d24e4e1/packages/lexical-rich-text/src/index.ts#L389-L391), I guess it's obsolete now with this change

fantactuka (approved):

PR #3411, opened 2022-11-23
Fix $splitNode Flow type

zurfyx:


packages/lexical-utils/flow/LexicalUtils.js.flow
```diff
@@ -79,4 +79,7 @@ declare export function $wrapNodeInElement(
   createElementNode: () => ElementNode,
 ): ElementNode;
 
-declare export function $splitNode<T: LexicalNode>(node: T): T;
+declare export function $splitNode(
+  node: ElementNode,
+  offset: number,
```
trueadm, line 84: This doesn't seem right, it should take …offsets like the node method, right?

trueadm: Oh this is the element version. Ignore me.

thegreatercurve (approved):

Issue #3417, opened 2022-11-24
Bug: Selection issues around block level decorator nodes


Piliuta:
Typing or pasting text does not work if the editor has only one (block) decorator node in the content.

Lexical version: 0.6.3

## Steps To Reproduce

1. Remove all content
2. Add a block node decorator (not a horizontal rule, that has been fixed)
3. Remove all spaces before and after the node
4. Set focus on the node and change cursor position to left or right (press Arrow Left/Right)
5. Attempt to type or paste any text - nothing happens.
 
**Notes:**
- Pressing Enter does work
- This is not an issue if content has any other (even empty) nodes

https://user-images.githubusercontent.com/1575198/203802959-25acb8a6-05a7-490c-9883-9d782c1413cb.mov

Piliuta:
@trueadm this is still a problem with other block decorator nodes. The horizontal rule was an example, but this is a problem with all such nodes. I do not think this should be fixed on the node level. It needs a general fix.

trueadm:
@Piliuta This doesn't seem like a Lexical bug though. The problem is how selection is handled by browser rendering engines in respect to content editable. `<hr>` as a block level element can have selection around it, but oddly, `<div>` can't it seems. I'm not sure there is a general fix here. Maybe you can help me do some more research around what types of element work and do not (using a standard content editable element).

Piliuta:
> @Piliuta This doesn't seem like a Lexical bug though. The problem is how selection is handled by browser rendering engines in respect to content editable. `<hr>` as a block level element can have selection around it, but oddly, `<div>` can't it seems. I'm not sure there is a general fix here. Maybe you can help me do some more research around what types of element work and do not (using a standard content editable element).

Not promising anything, but I can take a look. We should keep this issue open as it does not resolve the problem completely.

trueadm:
I wonder if someone has filed a browser bug around this before. I feel this should be something browser vendors should be fixing.

trueadm:
For what it's worth, making the container `<div>` for your decorator `inline` or `inline-block` seems to fix the selection issues too.

Piliuta:
> For what it's worth, making the container `<div>` for your decorator `inline` or `inline-block` via CSS `display` seems to fix the selection issues too.

Yeah, it does work for me. Do you think this can cause other side effects or it can be set as a style by default for decorator nodes?

Piliuta:
[...]

PR #3418, opened 2022-11-24
Fix browser selection with horizontal rule

trueadm:
Fixes https://github.com/facebook/lexical/issues/3417.

This is an interesting problem. It seems having a `<div>` as a top level block in the contenteditable causes selection to break around it – even if it uses `display: contents`. Having the `<hr>` be the top level block works though. I hate content editable sometimes.


thegreatercurve (approved):
LGTM. Just the failing tests.

PR #3424, opened 2022-11-25
Add link list properties to nodes

trueadm:
This PR introduces the node properties required for the work to move Lexical's internal structure to a doubly linked list in the future. Note: we don't actually do anything with these properties in this PR.


packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -55,6 +55,10 @@ export class ElementNode extends LexicalNode {
   /** @internal */
   __children: Array<NodeKey>;
   /** @internal */
+  __first: null | NodeKey;
+  /** @internal */
+  __last: null | NodeKey;
```
acywatson, line 60: Consider adding “__size” here as well

trueadm: Nice one.

acywatson:
I was wondering if it'd make more sense to develop this feature on a branch, as opposed to merging it in piecemeal. Makes it harder for others to understand what has and hasn't been done. Will leave it up to you though.

acywatson (approved):

trueadm:
@acywatson I wanted to get this in just to simply ensure we have the properties on the nodes. Having them there means that people might not assume they're not there and do breaking change stuff with Lexical's internal node structures (not that they should).

Issue #1158, opened 2022-01-21
Regression: bad code break: Delete empty line should move next line up, but also changes styling 


kraisler:
Deleting empty line where next line is H1 wrongly removes H1 styling.
https://user-images.githubusercontent.com/1952520/150588773-bd0feaee-fe56-4831-bd77-7b65e2306710.mov

yuzhouu:
another similar bug

Uploading 150588773-bd0feaee-fe56-4831-bd77-7b65e2306710.mov…

PR #3425, opened 2022-11-25
Fix a couple of bugs with Headings creation

### LINKED ISSUE #1170: Heading turns into normal paragraphed text when new line at the front
Similar to #1158 but the other way around is also bugged

In the video:
1. Correct
2. Debatable (Quip works this way but not GDocs, Word nor Pages)
3. Nope, certainly bugged

https://user-images.githubusercontent.com/193447/150975793-ce696608-f3e2-4f58-80b3-48af2b1f5567.mov

thegreatercurve:
On backspace at the start of heading, they shouldn't collapse into a paragraph, and on enter in the middle of a heading, they should stay as a heading.

Closes #1170 #1158

Before 

https://user-images.githubusercontent.com/7748470/204026666-11d75e18-8890-438b-b284-f01853005c62.mov

After

https://user-images.githubusercontent.com/7748470/204026688-55507bcc-3e83-4b5d-91c4-252789c2f894.mov

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -299,8 +299,9 @@ export class HeadingNode extends ElementNode {
 
   // Mutation
   insertNewAfter(selection?: RangeSelection): ParagraphNode | HeadingNode {
+    const selectionOffset = selection ? selection.anchor.offset : 0;
     const newElement =
-      selection && selection.anchor.offset < this.getTextContentSize()
+      selectionOffset < this.getTextContentSize() && selectionOffset > 0
         ? $createHeadingNode(this.getTag())
         : $createParagraphNode();
     const direction = this.getDirection();
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -298,9 +298,9 @@ export class HeadingNode extends ElementNode {
   }
 
   // Mutation
-  insertNewAfter(selection: RangeSelection): ParagraphNode | HeadingNode {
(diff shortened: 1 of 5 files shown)


packages/lexical-rich-text/src/index.ts
```diff
@@ -297,20 +298,25 @@ export class HeadingNode extends ElementNode {
...
     const direction = this.getDirection();
     newElement.setDirection(direction);
     this.insertAfter(newElement);
     return newElement;
   }
 
-  collapseAtStart(): true {
-    const paragraph = $createParagraphNode();
+  collapseAtStart(selection: RangeSelection): true {
+    const newElement =
```
trueadm, line 313: Shouldn't this be checking if the prevSibling is an empty element instead of checking the offsets?

packages/lexical/flow/Lexical.js.flow
```diff
@@ -494,7 +494,6 @@ type TextPointType = {
   getNode: () => TextNode,
   set: (key: NodeKey, offset: number, type: 'text' | 'element') => void,
   getCharacterOffset: () => number,
-  isAtNodeEnd: () => boolean,
```
trueadm, line 497: Was this never used anywhere? Not even internally?

thegreatercurve: No, it was either never implemented or it has been removed for a while. There's another method called `$isNodeAtEnd` which takes a selection point instead.

packages/lexical-rich-text/src/index.ts
thegreatercurve: Yeah, that's probably a better way of doing this one.

trueadm (approved):

Issue #3428, opened 2022-11-25
Bug:  correcting a word using Grammarly will also insert word where the cursor is if in a different paragraph.


scvet-ct:
Lexical version: 0.5.0 

## Steps To Reproduce

1. enable grammarly extension (turn off autocorrect)
2. write a paragraph
3. start a new paragraph with an incorrectly spelled word.
4. click back into the previous paragraph
5. hover over the incorrectly spelled word (don't click on it) and then click on Grammarly's suggested change to select it
6. the word will change but also be placed in the previous paragraph where the cursor is.

Link to code example:
This happens in the current version of Lexical Playground.

## The current behavior
<img width="426" alt="image" src="https://user-images.githubusercontent.com/114913876/203997116-72a87c73-fc95-4b5d-ab28-28ed31c53af8.png">
<img width="418" alt="image" src="https://user-images.githubusercontent.com/114913876/203997187-0afd7703-f390-4473-8ace-22c84e6e92b3.png">

## The expected behavior
<img width="426" alt="image" src="https://user-images.githubusercontent.com/114913876/203997116-72a87c73-fc95-4b5d-ab28-28ed31c53af8.png">
<img width="349" alt="image" src="https://user-images.githubusercontent.com/114913876/203997262-08cfcb51-f9f0-47ec-8607-05d125067c0b.png">

trueadm:
I'm a muppet and somehow undid the fix before landing that PR. I fixed in https://github.com/facebook/lexical/pull/3438.

scvet-ct:
Amazing, thank you for fixing it so quickly. Will update our version of lexical when it's in main 👍

PR #3429, opened 2022-11-25
Fix text replacement via insertText

trueadm:
Without regressing https://github.com/facebook/lexical/issues/2199 and https://github.com/facebook/lexical/pull/2853, this PR aims to fix text replacement for both Grammarly but also MacOS/iOS text replacement too, which we seem to have regressed on.

Fixes https://github.com/facebook/lexical/issues/3428.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -198,7 +198,10 @@ function $shouldPreventDefaultAndInsertText(
     // If we are replacing a range with a single character or grapheme, and not composing.
     (((!isBeforeInput &&
       (!CAN_USE_BEFORE_INPUT ||
-        // Attempt to detect execCommand('insertText')
+        // We check to see if there has been
+        // a recent beforeinput event for "textInput". If there has been one in the last
+        // 50ms then we proceed as normal. However, if there is not, then this is likely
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -166,6 +170,58 @@ let collapsedSelectionFormat: [number, number, NodeKey, number] = [
   0,
 ];
 
+// This function is used to determine if Lexical should attempt to override
+// the default browser behavior for insertion of text and use its own internal
+// heuristics. This is an extremely important function, and makes much of Lexical
+// work as intended between different browsers and across word, line and character
+// boundary/formats. It also is important for text replacement, node schemas and
+// composition mechanics.import { getActiveEditor } from './LexicalUpdates';
```
acywatson, line 178: Looks like an import statement somehow got tacked to this comment

acywatson:
Given that it seems pretty hard to test this, wondering if some additional context in the PR description could be useful. Any sense of what caused the regression? How is the timestamp hack fixing it?

acywatson (approved):
Seems fine :)

packages/lexical/src/LexicalEvents.ts
trueadm: LOL

Issue #3426, opened 2022-11-25
Feature: Improving selection around decorator nodes


trueadm:
Today, there are a variety of issues with selection and Lexical's decorator nodes.

- Block level decorator nodes can block selection from happening at all. https://github.com/facebook/lexical/issues/3417
- Inline level decorator nodes cause selection to appear in the wrong place or to be lost entirely.
- The cursor often doesn't show in the right place.

Taking inspiration from ProseMirror, we can opt to use two tricks to get around these problems:

- Have a virtual emulated cursor which is just a non editable div with no content that shows a horizontal cursor. This would be applied by the reconciler between boundaries of block level decorator elements. This fixes the first issue and the third issue.
- Use images without a src to be handled by the reconciler. These allow selection and the cursor to be shown in the right place. The images would be inserted around decorators that are block level (have no text next to them).

The complexity of this task is implementing this elegantly into the existing reconciler and teaching the selection system to identify and ignore these special elements we insert into the DOM.

Hopefully this should fix the many issues open around this problem space: https://github.com/facebook/lexical/issues/3417 https://github.com/facebook/lexical/issues/1965 https://github.com/facebook/lexical/issues/3156 https://github.com/facebook/lexical/issues/3157 https://github.com/facebook/lexical/issues/3143 https://github.com/facebook/lexical/issues/2285

mehmetscoskun:
I have a case where the images are like block and they are rendered at the same level with paragraphs. You cannot add text next to it. My problem was that I could not put my cursor between images. I fixed that with these steps:
- I added `::before` (top) and `::after` (bottom) pseudo elements to the image.
- I gave them enough heights, so that they look like margins.
- I added a `NodeEventPlugin` with click for the image nodes. When you click those pseudo elements:
           - it moves selection to before/after the image node, if there is some text. 
           - it creates a paragraph node before/after the image node, if there is no text and moves selection there.

In this fix, if you move cursor afterwards, it does not remove the paragraph node you just created but that's fine by me.

PR #3434, opened 2022-11-27
⏸ [0.7] Add block emulated cursors

trueadm:
This PR adds emulated block cursors to Lexical. Taking inspiration from ProseMirror, we can opt to use a trick to get around these problems:

- Have a virtual emulated cursor which is just a non editable div with no content that shows a horizontal cursor. This would be applied by the reconciler between boundaries of block level decorator elements. 

This logic works by teaching the entire core the concept of a "block cursor element". This element gets inserted when selection can exist in a position where the browser can't render a native cursor accurately between block elements. We insert the block cursor to show a blinking horizontal line that can be themed too.

Fixes https://github.com/facebook/lexical/issues/3417. Partly addresses https://github.com/facebook/lexical/issues/3426.

### DIFF
--- a/packages/lexical/src/LexicalReconciler.ts
+++ b/packages/lexical/src/LexicalReconciler.ts
@@ -634,7 +634,14 @@ function getFirstChild(element: HTMLElement): Node | null {
 }
 
 function getNextSibling(element: HTMLElement): Node | null {
-  return element.nextSibling;
+  let nextSibling = element.nextSibling;
+  if (
+    nextSibling !== null &&
+    nextSibling === activeEditor._blockCursorElement
+  ) {
+    nextSibling = nextSibling.nextSibling;
+  }
+  return nextSibling;
 }
 
 function reconcileNodeChildren(
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -8,6 +8,7 @@
 
 import type {
   CommandPayloadType,
+  EditorConfig,
   EditorThemeClasses,
   IntentionallyMarkedAsDirtyElement,
   Klass,
@@ -1032,7 +1033,7 @@ function resolveElement(
   return block.getChildAtIndex(isBackward ? offset - 1 : offset);
 }
 
-export function $getDecoratorNode(
+export function $getAdjacentNode(
   focus: PointType,
   isBackward: boolean,
 ): null | LexicalNode {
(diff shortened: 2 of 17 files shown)


packages/lexical/src/LexicalReconciler.ts
```diff
@@ -628,7 +628,14 @@ function getFirstChild(element: HTMLElement): Node | null {
 }
 
 function getNextSibling(element: HTMLElement): Node | null {
-  return element.nextSibling;
+  let nextSibling = element.nextSibling;
+  if (
+    nextSibling !== null &&
+    nextSibling === activeEditor._blockCursorElement
```
zurfyx, line 634: Do we still need this if we are removing the cursor before reconciliation?

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1291,3 +1292,100 @@ export function errorOnInsertTextNodeOnRoot(
...
+
+export function removeDOMBlockCursorElement(
+  blockCursorElement: HTMLElement,
+  editor: LexicalEditor,
+  rootElement: HTMLElement,
+) {
+  rootElement.style.removeProperty('caret-color');
+  editor._blockCursorElement = null;
+  const parentElement = blockCursorElement.parentElement;
+  if (parentElement !== null) {
```
zurfyx, line 1330: Nit; isn't it guaranteed to exist?

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1291,3 +1292,100 @@ export function errorOnInsertTextNodeOnRoot(
...
+    !node.isInline()
+  );
+}
+
+export function removeDOMBlockCursorElement(
+  blockCursorElement: HTMLElement,
+  editor: LexicalEditor,
+  rootElement: HTMLElement,
+) {
+  rootElement.style.removeProperty('caret-color');
```
zurfyx, line 1327: Why remove the caret-color property? The other cursor is still around no?

zurfyx (approved):
Brilliant work!

Heads up, it seems like blur doesn't really work, when you click outside it still blinks
Screen Shot 2022-11-28 at 3 56 50 PM

packages/lexical/src/LexicalReconciler.ts
trueadm: We don't remove it before reconciliation of the root, only selection.

packages/lexical/src/LexicalUtils.ts
trueadm: Nope. Mutations are the devil in the detail.

trueadm:
@zurfyx

> Heads up, it seems like blur doesn't really work, when you click outside it still blinks

I was unable to repro, how did you do that?

packages/lexical/src/LexicalUtils.ts
trueadm: It will use the CSS property instead. This will only remove the inline style.

zurfyx:
> @zurfyx
> 
> > Heads up, it seems like blur doesn't really work, when you click outside it still blinks
> 
> I was unable to repro, how did you do that?

I'm pulled on the latest Fix collab so I believe I got the latest

https://user-images.githubusercontent.com/193447/204382683-ced3da9e-45b6-44b0-a44a-ecc6eb3d8860.mov

trueadm:
@zurfyx Honestly, I can't repro this at all :O

There's logic in the handling of blur to remove the cursor. Can you check if that fires locally?

zurfyx:
> @zurfyx Honestly, I can't repro this at all :O
> 
> There's logic in the handling of blur to remove the cursor. Can you check if that fires locally?

Ah, turns out to be Microsoft Editor: Spelling & Grammar Checker. For some reason, this forces the selection to be dirty once again and causes a second reconciliation that eventually updates the block but doesn't remove it. Does this help?

Piliuta:
I am seeing an issue where after inserting twitter embed I cannot do anything in the editor. This is not an issue on the main playground.

https://user-images.githubusercontent.com/1575198/204555920-f018bab3-0b2a-4459-96b2-f0036e357f5d.mov

trueadm:
@Piliuta I'm struggling to repro. Can you pull the latest changes please?

Piliuta:
> @Piliuta I'm struggling to repro. Can you pull the latest changes please?

Thanks. This works after the last commit on the latest deployment https://lexical-playground-ax3d7u2it-fbopensource.vercel.app/

PR #3435, opened 2022-11-27
Adds type exports for various SerializedNodes

tolkienfan2:
Would be easier to be able to directly import all lexical serialized node types from the lexical packages instead of needing to install all the sub-packages just for the types.


tolkienfan2:
Failing tests probably caused by incompatible node version. Will close this PR and open a new one once this is fixed.

trueadm:
Not sure we even want to export them from the Lexical package as these other packages are optional and not core.

PR #3436, opened 2022-11-27
Exports serialized node types

tolkienfan2:
Exports serialized node types from main lexical package.

### DIFF
--- a/packages/lexical/src/index.ts
+++ b/packages/lexical/src/index.ts
@@ -7,24 +7,6 @@
  *
  */
 
- export type {SerializedCodeNode} from './../../lexical-code/src';
- export type {
-   SerializedAutoLinkNode,
-   SerializedLinkNode,
- } from './../../lexical-link/src';
- export type {
-   SerializedListItemNode,
-   SerializedListNode,
- } from './../../lexical-list/src';
- export type {
-   SerializedHeadingNode,
-   SerializedQuoteNode,
- } from './../../lexical-rich-text/src';
- export type {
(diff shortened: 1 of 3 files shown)


tolkienfan2:
I really don't know why the test above is failing with the message Error: prettier exit code 1.
When I run npm prettier in my local branch, it runs without any errors.

packages/lexical/src/index.ts
```diff
@@ -7,6 +7,24 @@
...
+ } from './../../lexical-list/src';
+ export type {
+   SerializedHeadingNode,
+   SerializedQuoteNode,
+ } from './../../lexical-rich-text/src';
+ export type {
+   SerializedTableCellNode,
+   SerializedTableNode,
+   SerializedTableRowNode,
+ } from './../../lexical-table/src';
```
thegreatercurve, line 27: This should generally be fine, but we don't want to re-export these from the core as it will just inflate the size of core for everyone. These should all be exported from the `index` files in their respective packages.

thegreatercurve (changes requested):
Thanks @tolkienfan2! Just one small change.

packages/lexical/src/index.ts
tolkienfan2: Oh, ok. So I should just close this PR in this case?

thegreatercurve: No, just delete the above few lines and you should be fine.

thegreatercurve (approved):

PR #3437, opened 2022-11-27
Don't unnecessarily mark text nodes as dirty

### LINKED ISSUE #1605: Revise methods that mark nodes as dirty
`setTextContent`, `setMode` etc. do not need to mark the node dirty if the original content is the same as the one being applied.

```
setTextContent(nextText) {
  if (this.getLatest().__text === nextText) {
    // Ignore, do not mark dirty
    return;
  }
  this.getWritable().__text = nextText;
}
```

thegreatercurve:
Closes #1605.

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -538,14 +538,20 @@ export class TextNode extends LexicalNode {
   setMode(type: TextModeType): this {
     const mode = TEXT_MODE_TO_TYPE[type];
     const self = this.getWritable();
+    if (self.__mode === mode) {
+      return self;
+    }
     self.__mode = mode;
     return self;
   }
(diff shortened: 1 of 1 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -538,14 +538,20 @@ export class TextNode extends LexicalNode {
...
     self.__mode = mode;
     return self;
   }
 
   setTextContent(text: string): this {
-    const writableSelf = this.getWritable();
-    writableSelf.__text = text;
-    return writableSelf;
+    const self = this.getWritable();
+    if (self.__text === text) {
```
trueadm, line 550: This will still make the node dirty as `getWrtiable` is the thing making it dirty.

trueadm (approved):

PR #3444, opened 2022-11-28
check CAN_USE_DOM before accessing document

ProvidenceXz:
Fix for Server Side Rendering, `document is undefined` error when invoking `LexicalComposer`

### DIFF
--- a/packages/lexical-react/src/LexicalComposer.tsx
+++ b/packages/lexical-react/src/LexicalComposer.tsx
@@ -25,6 +25,7 @@ import {
 } from 'lexical';
 import {useMemo} from 'react';
 import * as React from 'react';
+import {CAN_USE_DOM} from 'shared/canUseDOM';
 import useLayoutEffect from 'shared/useLayoutEffect';
 
 const HISTORY_MERGE_OPTIONS = {tag: 'history-merge'};
@@ -127,7 +128,7 @@ function initializeEditor(
       if (root.isEmpty()) {
         const paragraph = $createParagraphNode();
         root.append(paragraph);
-        const activeElement = document.activeElement;
+        const activeElement = CAN_USE_DOM ? document.activeElement : null;
         if (
           $getSelection() !== null ||
           (activeElement !== null && activeElement === editor.getRootElement())


packages/lexical-react/src/LexicalComposer.tsx
```diff
@@ -127,7 +128,7 @@ function initializeEditor(
       if (root.isEmpty()) {
         const paragraph = $createParagraphNode();
         root.append(paragraph);
-        const activeElement = document.activeElement;
+        const activeElement = CAN_USE_DOM ? document.activeElement : null;
```
zurfyx, line 131: Can we move this to `environment.ts`? Just next to `documentMode`. This will make it easier to reuse later

zurfyx (approved):
Thank you! Mind updating `useYjsCollaboration` as well? It has the same initialization logic

packages/lexical-react/src/LexicalComposer.tsx
ProvidenceXz: Do you mean to export `const activeElement = CAN_USE_DOM ? document.activeElement : null;` from `environment.ts`?

trueadm:
CLA needs signing too :)

packages/lexical-react/src/LexicalComposer.tsx
trueadm: Actually, it's fine to keep it outside. There's something strange about adding more logic to that module as it breaks tree shaking when you do :/

trueadm:
> Thank you! Mind updating useYjsCollaboration as well? It has the same initialization logic

That's in a `useEffect` right? That won't happen on the server AFAIK.

ProvidenceXz:
> CLA needs signing too :)

I think it took some time to propagate - should be good now :)

zurfyx:
> > Thank you! Mind updating useYjsCollaboration as well? It has the same initialization logic
> 
> That's in a `useEffect` right? That won't happen on the server AFAIK.

Hold on, you've got a point, but actually that's the same for LexicalComposer given the useLayoutEffect implementation we have?

```
const useLayoutEffectImpl: typeof useLayoutEffect = CAN_USE_DOM
  ? useLayoutEffect
  : useEffect;
```

ProvidenceXz:
Just re-checked and the error is triggered in SSR build. Here's the trace:
```
[=== ] info  - Generating static pages (0/46)
Error occurred prerendering page "/editor". Read more: https://nextjs.org/docs/messages/prerender-error
ReferenceError: document is not defined
    at [path]/node_modules/@lexical/react/LexicalComposer.prod.js:95:32
    at ke ([path]/node_modules/lexical/Lexical.prod.js:122:1)
    at A ([path]/node_modules/lexical/Lexical.prod.js:123:230)
    at Pe.update ([path]/node_modules/lexical/Lexical.prod.js:178:245)
    at initializeEditor ([path]/node_modules/@lexical/react/LexicalComposer.prod.js:89:13)
    at [path]/node_modules/@lexical/react/LexicalComposer.prod.js:66:8
    at Object.Ja [as useMemo] ([path]/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:27:240)
    at Object.exports.useMemo ([path]/node_modules/react/cjs/react.production.min.js:23:113)
    at LexicalComposer ([path]/node_modules/@lexical/react/LexicalComposer.prod.
```
If this is the case, maybe we should fix for the `useYjsCollaboration` too?

trueadm:
> > > Thank you! Mind updating useYjsCollaboration as well? It has the same initialization logic
> > 
> > 
> > That's in a `useEffect` right? That won't happen on the server AFAIK.
> 
> You've got a point, but actually that's the same for LexicalComposer given the useLayoutEffect implementation we have?
> 
> ```
> const useLayoutEffectImpl: typeof useLayoutEffect = CAN_USE_DOM
>   ? useLayoutEffect
>   : useEffect;
> ```

It's in a `useMemo` not an effect it seems?

trueadm (approved):

PR #3447, opened 2022-11-29
patch(Playground): add `type="button"` to <button /> components

ScottAgirs:
If this is merged, this will add `type` attribute to `<button />` components in playground.

This will fix an issue where clicking on any given button unintentionally submits a form, if the Lexical editor is used within a `<form />`


thegreatercurve:
This seems sensible if a user copy and pastes the code from the playground.

thegreatercurve (approved):

Issue #3457, opened 2022-11-30
Feature: According to the socket push data to achieve incremental document update


wuyxp:
I'm using lexical as our document component. Documents carry large amounts of data, and we support collaborative editing of documents. So, we think we can capitalize on the ability of reconcile, maximizing the performance of reducing the data network transfer overhead and lexical redraw.

Let's think about it in two parts

1. In terms of data storage, neither html string storage nor json structure obtained by editor.getEditorState() can be used. 
<img width="385" alt="image" src="https://user-images.githubusercontent.com/12847220/204692567-9276f48d-1dc1-40b9-a2c7-2c49d3aeda6d.png">

The above two storage methods will store data in a complete data structure, which is not suitable for partial transmission when incremental push. So what we want to save is the following structure. store the list structure with the key in the _nodeMap to the server so that it can be stored incrementally and also contains the key information to reduce dom updates when they need to be updated.
```ts
const obj = [...editor?.getEditorState()._nodeMap.entries()].reduce((obj, [key, value]) => (obj[key] = value, obj), {})
setEditorContentJSON(JSON.stringify(obj))
```
<img width="646" alt="image" src="https://user-images.githubusercontent.com/12847220/204692651-c044b4f8-c571-4803-bcd8-6af8c1185ebf.png">

2. We can do this when partial data is pushed through the socket.
```ts 
     const editorState = editor?.getEditorState()
      const nodeMap = editorState?._nodeMap
// First declare the default map data to be updated
      const dirtyMap = new Map()

// 先将当前 state 的所有 map 进行获取并且设置
      nodeMap?.forEach((value) => {
        value && dirtyMap.set(value.__key, value)
      })

// Obtain and set all maps of the current state
      dirtyArr.forEach((value) => {
        value && dirtyMap.set(value.__key, value)
      })

// Finally, all data is mounted to the corresponding prototype method by type.
              Object.entries(editorState).forEach(([key, value]) => {
                switch (value.__type) {
                  case 'root':
                    dirtyMap.set(key,Object.assign($createRootNode(),value))
                    break;
                  case 'paragraph':
                    dirtyMap.set(key,Object.assign($createParagraphNode(),value))
                    break;
                  case 'text':
                    dirtyMap.set(key,Object.assign($createTextNode(),value))
                    break;
                }
              })

// Then the map is generated to generate the state of the letter, and the set is updated
              const newState = new EditorState(dirtyMap)
              editor?.setEditorState(newState)
[...]

PR #3458, opened 2022-11-30
feat: Change the export editorState

wuyxp:
reference: https://github.com/facebook/lexical/issues/3457

### DIFF
--- a/packages/lexical/src/index.ts
+++ b/packages/lexical/src/index.ts
@@ -23,7 +23,8 @@ export type {
   SerializedEditor,
   Spread,
 } from './LexicalEditor';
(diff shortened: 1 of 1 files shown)


thegreatercurve:
@wuyxp Thanks for the detailed issue. 

I can’t speak for your specific use case, but unfortunately, I don’t think we should expose the EditorState class. Overriding the private node map could lead to unexpected errors, and I think exposing the class could lead to general confusion for users trying to figure out how to imperatively set editor state.

If you’re trying to handle partial transmission of collaborative editor state through a WebSocket, we offer bindings with yJS, which we’re still working to perfect (https://github.com/facebook/lexical/tree/main/packages/lexical-yjs).

PR #3474, opened 2022-12-01
Check for validity of next sibling as text node

thegreatercurve:


packages/lexical-text/src/index.ts
```diff
@@ -255,7 +256,8 @@ export function registerLexicalTextEntity<T extends TextNode>(
       if (
         match.start === 0 &&
         $isTextNode(prevSibling) &&
-        prevSibling.isTextEntity()
+        prevSibling.isTextEntity() &&
+        $isTextNode(nextSibling)
```
trueadm, line 260: shouldn't this be a separate thing, as this is when match.start is at the start? I'd imagine it to be only when the nextSibling is also an entity too?

PR #3476, opened 2022-12-01
Upgrade Playwright webkit version

trueadm:


thegreatercurve (approved):

acywatson:
Probably should just update the playwright fork, eh?

trueadm:
@acywatson I tried. I think they removed a bunch of the code to make the ime stuff work, so I doubt we can. They closed the original PR too :(

zurfyx (approved):

acywatson:
> @acywatson I tried. I think they removed a bunch of the code to make the ime stuff work, so I doubt we can. They closed the original PR too :(

Damn, that's going to be a problem eventually, I guess.

PR #3478, opened 2022-12-01
Improve selection performance on large updates

trueadm:
Moving selection on large DOM sets is not cheap in Chrome. It's actually one of the most expensive things you can do on Webkit/Chromium. In order to improve performance, you can separate DOM mutations, and in doing so make it easier for the recalculation of DOM layers to apply to selection.

The easiest way to approach tackling this problem is using a simple `requestAnimationFrame` however, this then moves much of the intended logic for dealing with problems within Lexical to being async. In reality, we don't always need this to be async and in most cases sync is fine. For when we do have async, Lexical can gradually pick-up, but this is mostly an edge-case.

### DIFF
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
@@ -79,6 +79,25 @@ jest.mock('shared/environment', () => {
   return {...originalModule, IS_FIREFOX: true};
 });
 
+Range.prototype.getBoundingClientRect = function (): DOMRect {
+  const rect = {
+    bottom: 0,
+    height: 0,
+    left: 0,
+    right: 0,
+    top: 0,
+    width: 0,
+    x: 0,
+    y: 0,
+  };
+  return {
+    ...rect,
+    toJSON() {
+      return rect;
+    },
+  };
+};
+
 describe('LexicalSelection tests', () => {
   let container = null;
 
--- a/packages/lexical/src/LexicalGC.ts
+++ b/packages/lexical/src/LexicalGC.ts
@@ -40,37 +40,31 @@ export function $garbageCollectDetachedDecorators(
 
 function $garbageCollectDetachedDeepChildNodes(
   node: ElementNode,
-  parentKey: NodeKey,
   prevNodeMap: NodeMap,
   nodeMap: NodeMap,
   dirtyNodes: Map<NodeKey, IntentionallyMarkedAsDirtyElement>,
 ): void {
-  const children = node.__children;
-  const childrenLength = children.length;
+  let child = node.getFirstChild();
 
-  for (let i = 0; i < childrenLength; i++) {
-    const childKey = children[i];
-    const child = nodeMap.get(childKey);
-
-    if (child !== undefined && child.__parent === parentKey) {
-      if ($isElementNode(child)) {
-        $garbageCollectDetachedDeepChildNodes(
-          child,
-          childKey,
-          prevNodeMap,
-          nodeMap,
-          dirtyNodes,
-        );
-      }
-
-      // If we have created a node and it was dereferenced, then also
-      // remove it from out dirty nodes Set.
-      if (!prevNodeMap.has(childKey)) {
-        dirtyNodes.delete(childKey);
-      }
-
-      nodeMap.delete(childKey);
+  while (child !== null) {
+    const childKey = child.__key;
+    const nextSibling = child.getNextSibling();
+    if ($isElementNode(child)) {
+      $garbageCollectDetachedDeepChildNodes(
+        child,
+        prevNodeMap,
+        nodeMap,
+        dirtyNodes,
+      );
     }
+
+    // If we have created a node and it was dereferenced, then also
+    // remove it from out dirty nodes Set.
+    if (!prevNodeMap.has(childKey)) {
+      dirtyNodes.delete(childKey);
+    }
+    nodeMap.delete(childKey);
+    child = nextSibling;
   }
 }
 
@@ -104,7 +98,6 @@ export function $garbageCollectDetachedNodes(
         if ($isElementNode(node)) {
           $garbageCollectDetachedDeepChildNodes(
             node,
-            nodeKey,
             prevNodeMap,
             nodeMap,
             dirtyElements,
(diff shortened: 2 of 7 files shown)


packages/lexical/src/LexicalSelection.ts
```diff
@@ -2791,10 +2792,7 @@ export function updateDOMSelection(
     !(domSelection.type === 'Range' && isCollapsed)
   ) {
     // If the root element does not have focus, ensure it has focus
-    if (
-      rootElement !== null &&
-      (activeElement === null || !rootElement.contains(activeElement))
-    ) {
+    if (activeElement === null || !rootElement.contains(activeElement)) {
```
thegreatercurve, line 2795: I think you might need to update the same logic in the AutoFocusPlugin as well.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2804,44 +2802,61 @@ export function updateDOMSelection(
...
+    // a "selectionchange" event, although it will be asynchronous.
+    try {
+      // When updating more than 1000 nodes, it's actually better to defer
+      // updating the selection till the next frame. This is because Chrome's
+      // Blink engine has hard limit on how many DOM nodes it can redraw in
+      // a single cycle, so keeping it to the next frame improves performance.
+      // The downside is that is makes the computation within Lexical more
+      // complex, as now, we've sync update the DOM, but selection no longer
+      // matches.
+      if (dirtyLeavesCount > 1000) {
```
thegreatercurve, line 2816: Is this hard limit the same in other browsers as well?

thegreatercurve (approved):

packages/lexical/src/LexicalSelection.ts
trueadm: I don't think so. `rootElement` always exists in this function. Not sure why we had the check before, as it's done in the calling function already.

packages/lexical/src/LexicalGC.ts
```diff
@@ -40,37 +40,31 @@ export function $garbageCollectDetachedDecorators(
...
-      // If we have created a node and it was dereferenced, then also
-      // remove it from out dirty nodes Set.
-      if (!prevNodeMap.has(childKey)) {
-        dirtyNodes.delete(childKey);
-      }
+  let child = node.getFirstChild();
+
+  while (child !== null) {
+    const childKey = child.__key;
+    const nextSibling = child.getNextSibling();
```
fantactuka, line 51: Curios why this would work faster given now nextSibling uses indexof

packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
```diff
@@ -79,6 +79,25 @@ jest.mock('shared/environment', () => {
   return {...originalModule, IS_FIREFOX: true};
 });
 
+Range.prototype.getBoundingClientRect = function (): DOMRect {
```
fantactuka, line 82: Can go into utils as mockSelectionRangeBoundingRect or into jest setup hook to have it always executed before test run

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2804,44 +2803,61 @@ export function updateDOMSelection(
...
+    // a "selectionchange" event, although it will be asynchronous.
+    try {
+      // When updating more than 1000 nodes on Chrome, it's actually better to defer
+      // updating the selection till the next frame. This is because Chrome's
+      // Blink engine has hard limit on how many DOM nodes it can redraw in
+      // a single cycle, so keeping it to the next frame improves performance.
+      // The downside is that is makes the computation within Lexical more
+      // complex, as now, we've sync update the DOM, but selection no longer
+      // matches.
+      if (IS_CHROME && dirtyLeavesCount > 1000) {
```
fantactuka, line 2817: 🤯

packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
trueadm: We need to do a round on cleaning up unit tests anyway, so this should be maybe part of that.

packages/lexical/src/LexicalGC.ts
trueadm: Oops. Just spotted this comment.

trueadm: https://github.com/facebook/lexical/pull/3481

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2804,44 +2803,61 @@ export function updateDOMSelection(
...
-      }
-    }
-
-    markSelectionChangeFromDOMUpdate();
-  } catch (error) {
-    // If we encounter an error, continue. This can sometimes
-    // occur with FF and there's no good reason as to why it
-    // should happen.
-  }
+  if (!tags.has('skip-scroll-into-view'))
```
Piliuta, line 2806: @trueadm any reasons `domSelection.setBaseAndExtent` https://github.com/facebook/lexical/pull/3478/files#diff-78b86cfd1ec931911a488d41ac3b779722121b8657aa59062b78856fa63cae12L2810 was moved inside this block?

It causes a selection discrepancy issue for me where lexical and dom selections do not match after focusing lexical editor via a custom logic that sets cursor near decorator nodes.

I have an `editor.update` call like this:


```js
// a mouse down handler

editor.update(() => {
  // determine offset position

  // ...

  // focus the editor if it is a click from other elements
  if (!rootElement.contains(document.activeElement)) {
    rootElement.focus({ preventScroll: true });
  }

  // ...

  // set Lexical selection here
  $getRoot().select(offset, offset);
}, { tag: 'skip-scroll-into-view' });
```

I am using this tag because the auto-scroll behavior here is not accurate because lexical attempts to make the node at the `offset` position fully visible, however, it is not practical as the cursor is set between the `offset` and `offset - 1` (in my implementation it is even worth as I am showing the cursor next to the element at the `offset - 1` position) and visibility of the whole node is not needed and even confusing as the content jumps further from where the custom cursor is visible.

trueadm: Are you referring to the fact we make it async now with rAF?

trueadm: Good point. I'll address.

Piliuta: Thank you 🙏

PR #3487, opened 2022-12-04
Inform users that they need to install headless

phillipb:
Currently, it's not obvious that `@lexical/headless` needs to be installed. Other packages like rich-text don't need to be installed. This is a simple PR to make docs a little more clear.


fantactuka (approved):

fantactuka:
👍

Issue #3488, opened 2022-12-04
Bug: Weird Backspace key behavior in nodes containing inputs


goale:
When I try to delete text in a node that contains an input element by hitting Backspace it ends up deleting text after the node. Playground does seem to have this issue too with Poll node.

Lexical version: 0.6.5

## Steps To Reproduce

1. Go to [Lexical Playground](https://playground.lexical.dev/)
2. Insert a Poll node and type in some poll options
3. Put the cursor on the text after the inserted node
4. Put the cursor back in one of the input options
5. Try to delete characters with Backspace

![Peek 2022-12-04 16-52](https://user-images.githubusercontent.com/4661017/205514383-07997790-ffce-43dd-acdc-a99c44cf14ff.gif)

## The current behavior
Characters are being deleted in text before or after the node
## The expected behavior
Characters are being deleted in the input element

trueadm:
I’ll look into this. Seems like an obvious issue looking at the video

hamodhalsatry:
MOHAMMED

hamodhalsatry:
> MOHAMMED

hamodhalsatry:
Halo

PR #3491, opened 2022-12-05
Fix decorator input selection

trueadm:
Turns out we had a strange bug in `isSelectionCapturedInDecoratorInput`. Fixes https://github.com/facebook/lexical/issues/3488.

### DIFF
--- a/packages/lexical-playground/src/nodes/PollComponent.tsx
+++ b/packages/lexical-playground/src/nodes/PollComponent.tsx
@@ -210,7 +210,7 @@ export default function PollComponent({
           cb(node);
         }
       },
-      { onUpdate },
+      {onUpdate},
     );
   };
 
--- a/packages/lexical-playground/src/nodes/PollComponent.tsx
+++ b/packages/lexical-playground/src/nodes/PollComponent.tsx
@@ -51,7 +51,10 @@ function PollOptionComponent({
   option: Option;
   options: Options;
   totalVotes: number;
-  withPollNode: (cb: (pollNode: PollNode) => void) => void;
+  withPollNode: (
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/nodes/PollComponent.tsx
```diff
@@ -186,13 +199,25 @@ export default function PollComponent({
...
+          cb(node);
+        }
+      },
+      {
+        onUpdate() {
+          if (onUpdate) {
+            onUpdate();
+          }
+        },
+      },
```
fantactuka, line 219:
```suggestion
      { onUpdate },
```

fantactuka (approved):

packages/lexical-playground/src/nodes/PollComponent.tsx
```diff
@@ -93,9 +96,19 @@ function PollOptionComponent({
...
+            const value = target.value;
+            const selectionStart = target.selectionStart;
+            const selectionEnd = target.selectionEnd;
+            withPollNode(
+              (node) => {
+                node.setOptionText(option, value);
+              },
+              () => {
+                target.selectionStart = selectionStart;
+                target.selectionEnd = selectionEnd;
```
fantactuka, line 109: It's a bit weird pattern to follow as it might be hard to add selection restoration in more complex embeds. I thought lexical would avoid handling input completely when selection is within decorator input.

trueadm: I guess you could avoid doing this if you marked the update as discrete too. However, I've had to do this in React before, it's because the setState for React is async from the change event AFAIK.

PR #3498, opened 2022-12-06
Fix insertion when the existing anchor node is orphaned

### LINKED ISSUE #3460: Bug: Text sometimes not updated in Safari since 0.6.4 (regession from #3429)
Lexical version: 0.6.4

Given a playwright test such as the following:

```ts
test.describe('Suite', () => {
  test.only('Edits stuff', async ({
    page,
  }) => {
    await page.goto('/page/with/editor');
    await page.
      .locator('[data-rich-text-input-front] [data-lexical-editor]')
      .fill('Front');

    await pageOne
      .locator('[data-rich-text-input-front] [data-lexical-editor]')
      .fill('Front updated');

    await expect(
      pageOne.locator('[data-rich-text-input-front] [data-lexical-editor]')
    ).toHaveText('Front updated');
  });
});
```

I get the following test failure after updating from 0.6.3 to 0.6.4 in **webkit** only.

```log
    Error: expect(received).toHaveText(expected)

    Expected string: "Front updated"
    Received string: "Front"
```

Having debugged the difference between the two versions I see the following:

Firstly, we call `$shouldPreventDefaultAndInsertText` from `onBeforeInput.updateEditor` with inputType `'insertText'`.

In Chromium in 0.6.3 and 0.6.4 `$shouldPreventDefaultAndInsertText` will return `true` due to the following condition evaluating to `true`:

https://github.com/facebook/lexical/blob/a2f3b2fe157bcd20d69c1e85b9203d6c52d244ab/packages/lexical/src/LexicalEvents.ts#L216-L220

In Chromium, the `domAnchorNode` points to the wrapping `<div>` element and hence will not match the text node for the `backingAnchorElement` (which is the child `<span>` element). As a result we will call `preventDefault()` and dispatch a `CONTROLLED_TEXT_INSERTION_COMMAND`.

In Webkit, however, in both 0.6.3 and 0.6.4, the above condition will return `false` since `domAnchorNode` points to text node child of `backingAnchorElement`. Hence none of the conditions in `$shouldPreventDefaultAndInsertText` evaluate to `true` and we _don't_ dispatch a `CONTROLLED_TEXT_INSERTION_COMMAND`.

Since we don't call `preventDefault()` we get a call to `onInput`.

In `onInput` we call `$shouldPreventDefaultAndInsertText` again but _this_ time, when we have an `'input'` event, `domAnchorNode` points to the the `Text` node on the `<span>` (containing "Front updated") back the `backingAnchorElement` points to an orphaned `<span>` whose `Text` node contains just "Front". Since these are different `Text` nodes, the backing anchor element check evaluates to `true` and `$shouldPreventDefaultAndInsertText` returns `true`.

From this point on the behavior differs between 0.6.3 and 0.6.4.

In 0.6.3, we unconditionally dispatch `CONTROLLED_TEXT_INSERTION_COMMAND`:

https://github.com/facebook/lexical/blob/1367743625e8ea79c08d881d4cecbe9195da2fcc/packages/lexical/src/LexicalEvents.ts#L616

However, in 0.6.4, as of #3429, we only dispatch a `CONTROLLED_TEXT_INSERTION_COMMAND` if the text in the DOM appears to differ from the result of splicing the event `data` with the anchor node's text:

https://github.com/facebook/lexical/blob/a2f3b2fe157bcd20d69c1e85b9203d6c52d244ab/packages/lexical/src/LexicalEvents.ts#L692-L706

In the case of this particular test, the combination of `anchorNode.getTextContent()` and `data` matches the text content of `domSelection.anchorNode` so we _don't_ dispatch a `CONTROLLED_TEXT_INSERTION_COMMAND`.

(Specifically, `anchorNode.getTextContent()` is `"Front"`, but the selection encompasses the whole string from 0 to 5 so we end up just considering `event.data` which is `"Front updated"`.)

As a result, `anchorNode` is never updated to match the DOM. And presumably at some later point we clobber the DOM with the value in `anchorNode`.

If I force the condition to evaluate to true and hence dispatch a `CONTROLLED_TEXT_INSERTION_COMMAND` the test passes again.

In Firefox, `beforeinput` uses `inputType` of `'insertCompositionText'` which is ignored by `onBeforeInput` so it also proceeds to `onInput`. However, it will fail the branch `$shouldPreventDefaultAndInsertText` in `onInput` (unlike Safari, `backingAnchorElement` appears to point to the live `<span>` element in Firefox) and proceed to update the selected text from the DOM:

https://github.com/facebook/lexical/blob/a2f3b2fe157bcd20d69c1e85b9203d6c52d244ab/packages/lexical/src/LexicalEvents.ts#L726-L727

birtles:
Fixes #3460.

Consists of three changes:

1. In #3429 we tweaked `$shouldPreventDefaultAndInsertText` so that the check for replacing a range with a single character or grapheme is only applied for input events (not beforeinput events) unless we determine that it is a dangling input event but we likely meant to `&&` that condition with the remaining conditions (not `||` it).

   See analysis here: https://github.com/facebook/lexical/issues/3460#issuecomment-1334744872

   This patch effectively changes the `||` to and `&&` (and drops a few no-longer-needed braces in the process).

2. We have observed in some circumstances (thought to be when using a newer version of Playwright) that Webkit can end up in a state where, during the input event, the DOM node for our lexical selection anchor has been orphaned.

   The second change in this patch makes `$shouldPreventDefaultAndInsertText` detect this case and return true (i.e. cause us to dispatch a controlled text insertion in this case).

3. Finally, in order to ensure that we do, in fact, dispatch a controlled text insertion in the case described in (2), this patch also alters the condition inside `onInput` to ensure we perform the insertion when the DOM node for our lexical anchor is orphaned.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -87,6 +87,7 @@ import {
   getAnchorTextFromDOM,
   getDOMTextNode,
   getEditorsToPropagate,
+  getInsertedText,
   getNearestEditorFromDOMNode,
   getWindow,
   isBackspace,
@@ -219,6 +220,11 @@ function $shouldPreventDefaultAndInsertText(
       domAnchorNode !== getDOMTextNode(backingAnchorElement)) ||
     // If the DOM selection element is orphaned
     (backingAnchorElement !== null && !backingAnchorElement.isConnected) ||
+    // If the DOM has not been updated for an input event
+    (!isBeforeInput &&
+      domAnchorNode &&
+      getInsertedText(selection, text) !==
+        getAnchorTextFromDOM(domAnchorNode)) ||
     // Check if we're changing from bold to italics, or some other format.
     anchorNode.getFormat() !== selection.format ||
     // One last set of heuristics to check against.
@@ -692,7 +698,6 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
       const backingAnchorElement = getActiveEditor().getElementByKey(
         selection.anchor.key,
       );
-      const offset = anchor.offset;
       // If the content is the same as inserted, then don't dispatch an insertion.
       // Given onInput doesn't take the current selection (it uses the previous)
       // we can compare that against what the DOM currently says.
@@ -704,9 +709,7 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
         // If the DOM node for our anchor is orphaned, however, we need to do
         // the text insertion.
         (backingAnchorElement !== null && !backingAnchorElement.isConnected) ||
-        anchorNode.getTextContent().slice(0, offset) +
-          data +
-          anchorNode.getTextContent().slice(offset + selection.focus.offset) !==
+        getInsertedText(selection, data) !==
           getAnchorTextFromDOM(domSelection.anchorNode)
       ) {
         dispatchCommand(editor, CONTROLLED_TEXT_INSERTION_COMMAND, data);
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -223,6 +223,7 @@ function $shouldPreventDefaultAndInsertText(
     // If the DOM has not been updated for an input event
     (!isBeforeInput &&
       domAnchorNode &&
+      !anchorNode.isComposing() &&
       getInsertedText(selection, text) !==
         getAnchorTextFromDOM(domAnchorNode)) ||
     // Check if we're changing from bold to italics, or some other format.
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -196,15 +196,14 @@ function $shouldPreventDefaultAndInsertText(
     // If we're working with a non-text node.
     !$isTextNode(anchorNode) ||
     // If we are replacing a range with a single character or grapheme, and not composing.
-    (((!isBeforeInput &&
+    (!isBeforeInput &&
       (!CAN_USE_BEFORE_INPUT ||
         // We check to see if there has been
         // a recent beforeinput event for "textInput". If there has been one in the last
         // 50ms then we proceed as normal. However, if there is not, then this is likely
         // a dangling `input` event caused by execCommand('insertText').
-        lastBeforeInputInsertTextTimeStamp < timeStamp + 50)) ||
-      textLength < 2 ||
-      doesContainGrapheme(text)) &&
+        lastBeforeInputInsertTextTimeStamp < timeStamp + 50) &&
+      (textLength < 2 || doesContainGrapheme(text)) &&
       anchor.offset !== focus.offset &&
       !anchorNode.isComposing()) ||
     // Any non standard text node.
@@ -218,6 +217,8 @@ function $shouldPreventDefaultAndInsertText(
       backingAnchorElement !== null &&
       !anchorNode.isComposing() &&
       domAnchorNode !== getDOMTextNode(backingAnchorElement)) ||
+    // If the DOM selection element is orphaned
+    (backingAnchorElement !== null && !backingAnchorElement.isConnected) ||
     // Check if we're changing from bold to italics, or some other format.
     anchorNode.getFormat() !== selection.format ||
     // One last set of heuristics to check against.
@@ -688,6 +689,9 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
       if (domSelection === null) {
         return;
       }
+      const backingAnchorElement = getActiveEditor().getElementByKey(
+        selection.anchor.key,
+      );
       const offset = anchor.offset;
       // If the content is the same as inserted, then don't dispatch an insertion.
       // Given onInput doesn't take the current selection (it uses the previous)
@@ -697,6 +701,9 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
         selection.isCollapsed() ||
         !$isTextNode(anchorNode) ||
         domSelection.anchorNode === null ||
+        // If the DOM node for our anchor is orphaned, however, we need to do
+        // the text insertion.
+        (backingAnchorElement !== null && !backingAnchorElement.isConnected) ||
         anchorNode.getTextContent().slice(0, offset) +
           data +
           anchorNode.getTextContent().slice(offset + selection.focus.offset) !==
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -713,6 +713,8 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
           getAnchorTextFromDOM(domSelection.anchorNode)
       ) {
         dispatchCommand(editor, CONTROLLED_TEXT_INSERTION_COMMAND, data);
+      } else {
+        $updateSelectedTextFromDOM(false);
       }
 
       const textLength = data.length;
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -510,6 +510,29 @@ export function getAnchorTextFromDOM(anchorNode: Node): null | string {
   return null;
 }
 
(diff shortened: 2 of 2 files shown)


trueadm (approved):
This is so detailed and well executed. Thank you for digging into this. Did you confirm this works with browser extensions still too? I’ll check Grammarly later and the macOS keyboard text replacement tool and merge if I don’t find any issues

birtles:
Wow, that's ironic. The code that was meant to fix #3460 made the (previously passing) test for it fail.

birtles:
> This is so detailed and well executed. Thank you for digging into this. Did you confirm this works with browser extensions still too? I’ll check Grammarly later and the macOS keyboard text replacement tool and merge if I don’t find any issues

Thanks! Unfortunately it looks like I have more work to do just getting the tests to pass first. After that I'm happy to check Grammarly. Are there any specific STR to try?

trueadm:
> > This is so detailed and well executed. Thank you for digging into this. Did you confirm this works with browser extensions still too? I’ll check Grammarly later and the macOS keyboard text replacement tool and merge if I don’t find any issues
> 
> Thanks! Unfortunately it looks like I have more work to do just getting the tests to pass first. After that I'm happy to check Grammarly. Are there any specific STR to try?

Just some manual tests from things that I’ve found problematic - mainly browser extensions (Microsoft editor and grammarly), the onscreen keyboard text replacement thing on macOS, system emoji inserter on Mac and windows and system spellcheck

trueadm:
Seems to have an issue with graphemes now. I'm not sure making `$shouldPreventDefaultAndInsertText` more complex is the right angle to tackle this, as it affects so many different things and is extremely fragile because of that.

If you can create an e2e test that emulates the behavior of the latest Playwright then I can also dig into this with you – maybe by manually emulating the event sequence (beforeinput, DOM mutations, input) manually in the test. I can take another stab at trying to update Playwright but that's proving to be really difficult too.

birtles:
> Seems to have an issue with graphemes now. I'm not sure making `$shouldPreventDefaultAndInsertText` more complex is the right angle to tackle this, as it affects so many different things and is extremely fragile because of that.

Yes, I agree.

I think this PR & issue got a bit derailed when I discovered the odd-looking condition introduced in #3429.

That is, we had the following:

```js
    (((!isBeforeInput &&
      (!CAN_USE_BEFORE_INPUT ||
        // We check to see if there has been
        // a recent beforeinput event for "textInput". If there has been one in the last
        // 50ms then we proceed as normal. However, if there is not, then this is likely
        // a dangling `input` event caused by execCommand('insertText').
        lastBeforeInputInsertTextTimeStamp < timeStamp + 50)) ||
      textLength < 2 ||
      doesContainGrapheme(text)) &&
      anchor.offset !== focus.offset &&
      !anchorNode.isComposing()) ||
```

which I believe was supposed to be:

```js
    (!isBeforeInput &&
      (!CAN_USE_BEFORE_INPUT ||
        // We check to see if there has been
        // a recent beforeinput event for "textInput". If there has been one in the last
        // 50ms then we proceed as normal. However, if there is not, then this is likely
        // a dangling `input` event caused by execCommand('insertText').
        lastBeforeInputInsertTextTimeStamp < timeStamp + 50) &&
      (textLength < 2 || doesContainGrapheme(text)) &&
      anchor.offset !== focus.offset &&
      !anchorNode.isComposing()) ||
```

(Since otherwise `$shouldPreventDefaultAndInsertText` will return true for _any_ `input` event preceded by a `beforeinput` event with a non-collapsed selection and where we're not composing.)

But when I fix that, the test added in #3470 (the "Front updated" one) fails.

So, if my understanding is correct and the code introduced in #3429 is buggy, then I think the first step would be getting the newly added E2E test to pass with the above-mentioned code fixed--which might mean revisiting how to fix #3428 / #3429.

> If you can create an e2e test that emulates the behavior of the latest Playwright then I can also dig into this with you – maybe by manually emulating the event sequence (beforeinput, DOM mutations, input) manually in the test. I can take another stab at trying to update Playwright but that's proving to be really difficult too.

Sure! I can have a go at that but it's probably more important to revisit the fix for #3428 / #3429 first.

birtles:
> Another plot twist – updating to the latest Playwright still doesn't cause that test to fail. It must be something else?

That's interesting. I've simplified my app down to the following and it still reproduces for me:

```tsx
  <RichTextPlugin
    contentEditable={<ContentEditable />}
    ErrorBoundary={ErrorBoundary}
    placeholder={<div>Yer</div>}
  />
</LexicalComposer>
```

So my guess is either:

- There's something in the lexical playground setup that is triggering a different behavior
- There's another event handler somewhere in my app that is triggering the different behavior

Webkit tends to take different code paths based on the presence of even a single no-op event handler so it might be hard to track down what is tickling the different behavior here.

trueadm:
> > Another plot twist – updating to the latest Playwright still doesn't cause that test to fail. It must be something else?
> 
> That's interesting. I've simplified my app down to the following and it still reproduces for me:
> 
> ```tsx
> <LexicalComposer initialConfig={initialConfig}>
>   <RichTextPlugin
>     contentEditable={<ContentEditable />}
>     ErrorBoundary={ErrorBoundary}
>     placeholder={<div>Yer</div>}
>   />
> </LexicalComposer>
> ```
> 
> So my guess is either:
> 
> * There's something in the lexical playground setup that is triggering a different behavior
> * There's another event handler somewhere in my app that is triggering the different behavior
> 
> Webkit tends to take different code paths based on the presence of even a single no-op event handler so it might be hard to track down what is tickling the different behavior here.

Could it be about focus/selection? The playground auto-selects the content editable. I wonder if that's the big difference?

trueadm:
> I think this PR & issue got a bit derailed when I discovered the odd-looking condition introduced in https://github.com/facebook/lexical/pull/3429.

I agree, that does look wrong. Let me try and fix that as per your suggestion and make things work. I think this might be a separate problem from what you're seeing though, which might be related to the fact your environment might not be behaving the same as the playground – specifically around focus or selection.

birtles:
> Could it be about focus/selection? The playground auto-focuses the content editable via the `<AutoFocusPlugin />`. I wonder if that's the big difference?

I just tried throwing the `<AutoFocusPlugin />` in but it didn't seem to make a difference.

I also tried ripping things out of playground to on `main` to see if I can get that test to fail there but no luck.

trueadm:
@birtles Are you using React 18?

birtles:
> @birtles Are you using React 18?

Oh I'm really sorry! That's it! I'm using Preact. I didn't even think about that being a factor. Sorry!

trueadm:
> Oh I'm really sorry! That's it! I'm using Preact. I didn't even think about that being a factor. Sorry!

Should we close this PR and issue then?

birtles:
> Should we close this PR and issue then?

I think we can close this PR but probably leave the issue open for now.

I'd like to look at the issue after we have a proper fix in place for #3428 / #3429.

I think it's still worth fixing because:

- We don't know for sure if Preact is to blame
- Even if it is, Lexical is framework agnostic and its conceivable that another framework or even a future version of React or Playwright could generate the same situation
- This used to work before 0.6.3 so it should be possible to get it to work again

With a proper fix for #3428 / #3429 maybe this case will just work, or maybe it will be a one-liner to fix. In any case most of this PR won't be needed.

trueadm:
Whilst Lexical is framework agnostic, you’re also using React based plugins. Preact isn’t a 1:1 swap in replacement for React 18, as much as they may market it as such. This has been a common pattern even before React 18 too (I used to work on the React core team).

birtles:
> At the same time, it would be good to maybe setup a code sandbox with Preact so we can further dig into this :) unless this might be related to running in sync mode. We run React 18 in concurrent mode, which is how you're meant to use React 18, but maybe that could be fixing the issue too.

Yes, that would be good. Unfortunately I need to spend a bit of time first debugging why certain cases of deletion are broken for us after updating to 0.7.0 so it might be a while before I can get back to this. (I suspect either 6b7d82842adac1832209b58cac476faeb4c48752 or d30c88202e8d68e7ab79c9d2bf45366de0cea794 but they're big changesets so it's going to take a while.)

PR #3505, opened 2022-12-06
Improve Auto Embed

tylerjbainbridge:
This PR adds support for async parseURL calls, which may be necessary for making sure something is embeddable. It also lets parse URL add arbitrary data to its match result which may be useful for caching async results.

Most of the changed lines here relate to adding async support on the playground.

### DIFF
--- a/packages/lexical-playground/src/plugins/AutoEmbedPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/AutoEmbedPlugin/index.tsx
@@ -228,17 +228,19 @@ export function AutoEmbedDialog({
 
   const validateText = useMemo(
     () =>
-      debounce(() => {
-        const urlMatch = URL_MATCHER.exec(text);
-        if (embedConfig != null && text != null && urlMatch != null) {
-          Promise.resolve(embedConfig.parseUrl(text)).then((parseResult) => {
-            setEmbedResult(parseResult);
-          });
+      debounce((inputText) => {
+        const urlMatch = URL_MATCHER.exec(inputText);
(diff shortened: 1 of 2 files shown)


trueadm (approved):

packages/lexical-playground/src/plugins/AutoEmbedPlugin/index.tsx
```diff
@@ -16,7 +16,8 @@ import {
   URL_MATCHER,
 } from '@lexical/react/LexicalAutoEmbedPlugin';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
-import {useState} from 'react';
+import {debounce} from 'lodash-es';
```
trueadm, line 19: Let’s avoid pulling in lodash. Just create a denounce function instead

packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
```diff
@@ -158,17 +169,20 @@ export function LexicalAutoEmbedPlugin<TEmbedConfig extends EmbedConfig>({
...
       if ($isLinkNode(linkNode)) {
-        const result = activeEmbedConfig.parseUrl(linkNode.__url);
+        const result = await Promise.resolve(
+          activeEmbedConfig.parseUrl(linkNode.__url),
+        );
         if (result != null) {
           editor.update(() => {
             activeEmbedConfig.insertNode(editor, result);
+            if (linkNode.isAttached()) {
+              editor.update(() => {
```
fantactuka, line 181: Probably no need for this editor.update anymore

PR #3510, opened 2022-12-07
Add warn only once to all console.warn instances

### LINKED ISSUE #1012: Warning system
Implement a warning system just like invariant

Special requisites:
1. Should trigger only once

thegreatercurve:
Closes #1012


trueadm:
You might need to reference the new module in jest.config.js and ts.config.json.

PR #3528, opened 2022-12-10
Fix paste for LexicalLinkPlugin

trueadm:
If we're trying to paste content in over a range of selection that contains an element, then avoid applying the link logic.

### DIFF
--- a/packages/lexical-react/src/LexicalLinkPlugin.ts
+++ b/packages/lexical-react/src/LexicalLinkPlugin.ts
@@ -11,6 +11,7 @@ import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {mergeRegister} from '@lexical/utils';
 import {
   $getSelection,
+  $isElementNode,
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/LexicalLinkPlugin.ts
```diff
@@ -66,9 +67,13 @@ export function LinkPlugin({validateUrl}: Props): null {
               if (!validateUrl(clipboardText)) {
                 return false;
               }
-              editor.dispatchCommand(TOGGLE_LINK_COMMAND, clipboardText);
-              event.preventDefault();
-              return true;
+              // If we select nodes that are elements then avoid applying the link.
+              if (!selection.getNodes().find(node => $isElementNode(node))) {
```
fantactuka, line 71:
```suggestion
              if (selection.getNodes().some(node => $isElementNode(node))) {
```

fantactuka (approved):

PR #3532, opened 2022-12-12
Add collaboration-init tag when initializeEditor for yjs collaboration

strdr4605:
Follow up to:
- #3520 

This will allow having a hook that will tell when the editor has initiated the collaboration.

```js
/**
 * @param {boolean} shouldRegisterEvent - pass isCollabEnabled
 * @returns {boolean} true when lexical editor has content from the collaboration session
 */
export const useEditorIsInCollab = (shouldRegisterEvent: boolean): boolean => {
  const [isInCollab, setIsInCollab] = useState(false);

  useEffect(() => {
    if (!shouldRegisterEvent || isInCollab) {
      return undefined;
    }
    return editor.registerUpdateListener(({ tags }) => {
      if (tags.has('collaboration') || tags.has('collaboration-init')) {
        setIsInCollab(true);
      }
    });
  }, [editor, isInCollab, shouldRegisterEvent]);
}
```

### DIFF
--- a/packages/lexical-react/src/shared/useYjsCollaboration.tsx
+++ b/packages/lexical-react/src/shared/useYjsCollaboration.tsx
@@ -303,11 +303,15 @@ function initializeEditor(
             case 'string': {
               const parsedEditorState =
                 editor.parseEditorState(initialEditorState);
-              editor.setEditorState(parsedEditorState, {tag: 'history-merge'});
+              editor.setEditorState(parsedEditorState, {
+                tag: 'collaboration-init',
+              });
               break;
             }
             case 'object': {
-              editor.setEditorState(initialEditorState, {tag: 'history-merge'});
+              editor.setEditorState(initialEditorState, {
+                tag: 'collaboration-init',
+              });
               break;
             }
             case 'function': {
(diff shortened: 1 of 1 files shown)


fantactuka:
This one is a bit tricky to change since there're some plugins (e.g., onChange) that rely on history-merge tag. At this point 'history-merge' is used as "(likely) not user-triggered change" which is often used by plugins that are initializing editor or auto-adjusting its initial state. See summary of [#2706](https://github.com/facebook/lexical/pull/2706) with details from @zurfyx.

fantactuka:
Added some extra details in your original [Q&A post](https://github.com/facebook/lexical/discussions/3520)

strdr4605:
Same as "collaboration" tag, "(likely) not CURRENT user-triggered change".
When I tested, I always get 2 events. first is "history-merge" probable from another initialization.
and second is "history-merge" from CollabPlugin.

When there is a Yjs Doc then the second event is "collaboration".

---

If you think that this may break other parts of the editor. 
How can we achieve the goal of knowing then the editor is ready to make changes that will be propagated to the collab session?

fantactuka:
Mentioned it in Q&A post, but you might want to use "synced" event from your collab provider. Having "init" event for collab is not 100% reliable anyway as you can always have concurrent edits from other clients during client boot

strdr4605:
Ok, will try to use "synced" event from your collab provider. If it still does not solve the issue, we can come back to this topic.

PR #3540, opened 2022-12-12
Pass replaced node's key to new node

moy2010:
Fix for displaying placeholder text for CustomParagraphNode example

### DIFF
--- a/packages/lexical-website/docs/concepts/node-replacement.md
+++ b/packages/lexical-website/docs/concepts/node-replacement.md
@@ -15,7 +15,7 @@ const editorConfig = {
         {
             replace: ParagraphNode,
             with: (node: ParagraphNode) => {
-                return new CustomParagraphNode();
+                return new CustomParagraphNode(node.__key);
             }
         }
     ]


trueadm:
Please don’t do this. In fact we should add an invariant in the return call to check if the keys are the same. They should never be the same! This will break many of the internal heuristics around Lexical’s garbage collector if you force it.

moy2010:
> Please don’t do this. In fact we should add an invariant in the return call to check if the keys are the same. They should never be the same! This will break many of the internal heuristics around Lexical’s garbage collector if you force it.

Ah, gotcha! Do you know how can we make the placeholder text work with the custom paragraph example?

trueadm:
The issue is here:

https://github.com/facebook/lexical/blob/main/packages/lexical-text/src/index.ts#L117

This should be using an instanceof check rather than checking the type.

moy2010:
> The issue is here:
> 
> https://github.com/facebook/lexical/blob/main/packages/lexical-text/src/index.ts#L117
> 
> This should be using an instanceof check rather than checking the type property. Maybe you can change the PR accordingly?

Sorry, I didn't see the last part of your message and created a new PR instead. I did it on the browser since I haven't cloned the repo locally yet, so hopefully it typechecks.

acywatson:
Closing in favor of #3542

PR #3541, opened 2022-12-12
Change mousedown event type to fix selection Chromium bug on Android

### LINKED ISSUE #3538: Bug: Gboard causes cursor to jump back when navigating
Lexical version: 0.7.1

## Steps To Reproduce

1. Use an (Android?) phone with Gboard
2. Put the cursor somewhere in the editor input that is not the end of a line
3. Click somewhere else in the editor input

Link to code example: https://codesandbox.io/s/lexical-plain-text-example-g932e

## The current behavior
When attempting to move the cursor, it jumps back to the current location immediately. Doesn't happen if the cursor is at the end of a line, doesn't happen if the keyboard is minimized.

Only happens if Gboard is enabled. Tested it on 3 different Android phones, all had the same behavior. Not sure if it happens on iOS as well.

### Video example
https://user-images.githubusercontent.com/1776412/207128608-edc8dda5-63a7-4303-a5c2-56898109512a.mp4

## The expected behavior
The cursor moves and stays were you put it

thegreatercurve:
Closes #3538 

Validated that it fixes the issue on mobile, and doesn't cause regressions with #2901 on desktop.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -132,7 +132,7 @@ const PASS_THROUGH_COMMAND = Object.freeze({});
 const ANDROID_COMPOSITION_LATENCY = 30;
 const rootElementEvents: RootElementEvents = [
   ['keydown', onKeyDown],
-  ['mousedown', onMouseDown],
+  ['pointerdown', onPointerDown],
   ['compositionstart', onCompositionStart],
   ['compositionend', onCompositionEnd],
   ['input', onInput],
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -391,6 +392,7 @@ function onMouseDown(event: MouseEvent, editor: LexicalEditor) {
       }
     });
   }
+  event.stopPropagation();
```
fantactuka, line 395: Is it needed for non-touch devices? Could also limit it to android only

trueadm (approved):

zurfyx (approved):
Thank you

PR #3543, opened 2022-12-12
Don't throw if getPreviousSiblings can't find a parent

acywatson:
This method used to safely return an empty list instead of throwing if the parent isn't found. This changes it back to that approach, so the API matches similar methods and the linked list change remains an implementation detail.

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -337,7 +337,10 @@ export class LexicalNode {
 
   getPreviousSiblings<T extends LexicalNode>(): Array<T> {
     const siblings: Array<T> = [];
-    const parent = this.getParentOrThrow();
+    const parent = this.getParent();
+    if (parent === null) {
+      return siblings;
+    }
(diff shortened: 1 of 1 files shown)


fantactuka:
Curious if it's possible for node to have null parent but non-null siblings. If that's the case then getPrev/NextSiblings will work differently. That said, should we have check for parent in getNextSiblings too to avoid it?

acywatson:
> Curious if it's possible for node to have null parent but non-null siblings. If that's the case then getPrev/NextSiblings will work differently. That said, should we have check for parent in getNextSiblings too to avoid it?

I suppose anything is possible? We never seemed to have that issue before, when this method worked this way. 

Maybe the best solution ultimately is to make some of these methods invariant if they're called on RootNode. It's pretty nonsensical. But then you have to check for RootNode everywhere.

trueadm (approved):
Makes sense. I think I did this by accident

acywatson:
> Makes sense. I think I did this by accident

I figured that might be the case - thanks!

PR #3558, opened 2022-12-15
Fix dom inline style not work

nwa2018:
fix when transform html to lexical, the inline style not work. 

for example, when the html string is 

```html
<p class="PlaygroundEditorTheme__paragraph" dir="ltr"><span style="background-color: rgb(208, 2, 27);">a</span><span style="background-color: rgb(255, 255, 255); color: rgb(208, 2, 27);">a</span><span style="background-color: rgb(255, 255, 255); font-size: 10px;">a</span></p>
```

and run 
```js
      const dom = parser.parseFromString('<p class="PlaygroundEditorTheme__paragraph" dir="ltr"><span style="background-color: rgb(208, 2, 27);">a</span><span style="background-color: rgb(255, 255, 255); color: rgb(208, 2, 27);">a</span><span style="background-color: rgb(255, 255, 255); font-size: 10px;">a</span></p>', 'text/html');
      $getRoot().select();
      $insertNodes(nodes);
```

lexical will behave like this without the inline style
image

while the normal case should behave like this

image

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -862,6 +862,8 @@ function convertSpanElement(domNode: Node): DOMConversionOutput {
         lexicalNode.toggleFormat('superscript');
       }
 
+      lexicalNode.setStyle(span.style.cssText);
+
       return lexicalNode;
     },
     node: null,


acywatson:
We don't want to do this in the core. See the discussion here: 

https://github.com/facebook/lexical/pull/3017

This was also attempted previously in a different PR and reverted.

The solution here for now is to extend TextNode and override importDOM. If this is coming up continually, maybe we can add some configuration to enable it, but it cannot be the default.

PR #3564, opened 2022-12-15
Allow subclass node of equivalent type to override the lexical default node

nwa2018:
currently I use all the lexical package except `lexical-playground` to build my editor, and as the product requirement, I need to store html string and pass the value to lexical editor, and thus I will use the [html---lexical](https://lexical.dev/docs/concepts/serialization#html---lexical) api to init my editor value.

however, the `importDom` api in `TextNode` and `ParagraphNode` do not satisfy all scenarios, such as, inline style not work for `TextNode`, can refer to [#3558](https://github.com/facebook/lexical/pull/3558) and [#3017](https://github.com/facebook/lexical/pull/3017)

Then I found a way to overwrite the TextNode

``` js

class CustomTextNode extends TextNode {
        // custom logic
    }
}

const PlaygroundNodes: Array<
  | Klass<LexicalNode>
  | {
      replace: Klass<LexicalNode>;
      with: <T extends { new (...args: any): any }>(node: InstanceType<T>) => LexicalNode;
    }
> = [
  {
    replace: CustomTextNode,
    with: (node: TextNode) => {
      return new CustomTextNode(node.__text);
    },
  },
  HeadingNode,
  ListNode,
  ListItemNode,
  ...
]
```

in this way, I can rewrite `importDOM` function to fix the deserialization problem in my case. But it will cause error as the snapshot below

image

So I submit this pull request to fix it. If not, do you have any idea how to solve this deserialization problem gracefully? thank you for your time.

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -830,13 +830,15 @@ function errorOnTypeKlassMismatch(
     );
   }
   const editorKlass = registeredNode.klass;
-  if (editorKlass !== klass) {
+  if (editorKlass !== klass && !(editorKlass.prototype instanceof klass)) {
     invariant(
       false,
-      'Create node: Type %s in node %s does not match registered node %s with the same type',
+      'Create node: Type %s in node %s does not match registered node %s with the same type or %s is not a subclass of %s',
       type,
       klass.name,
       editorKlass.name,
+      editorKlass.name,
+      klass.name
     );
   }
 }


trueadm:
You also need to register your `CustomTextNode` before you do the replacement.

nwa2018:
> You also need to register your `CustomTextNode` before you do the replacement.

Yeah, I pass the `PlaygroundNodes` which contains my `CustomTextNode` to `LexicalComposer`, and it works. 

But it will causes the following error. So I submit this pull request to fix it. If the node with the same type to the default lexical Node and is the subclass of the original lexical Node, will not cause the below error.

image

trueadm:
Your `CustomTextNode` also needs a `static getType()` method that returns a different type than `text`.

trueadm:
I understand what you're trying to do, but you should never duplicate types. The standard `TextNode` in core is the only node that can have the type of `text`. Given you've replaced the base with your custom one, that should suffice. You shouldn't need to also have the same type anywhere. If you find there are differences then that's a bug we need to fix separately.

acywatson:
This doesn’t make sense to me. Why can’t you just give the subclass a different type, as dominic suggested? What breaks when you do that?

Issue #3562, opened 2022-12-15
Bug: [Playground] Image gets stuck in resizing state


SharatChandar:
When resizing an Image over a long distance in lexical playground, the draggable event of the image is getting triggered. This causes the image to be stuck in the resizing state.
Lexical version:  "0.7.1",

## Steps To Reproduce

1. Add an image in playground
2. Resize over a considerable amount of distance by dragging the resize pointer

## Video

https://user-images.githubusercontent.com/27125472/207875504-2b69c7c9-5798-43ac-a801-cb694acd7040.mp4

## The current behavior
Image gets stuck in resize state

## The expected behavior
Image should not be draggable while resizing

SharatChandar:
Fixed this by checking for resizing before setting image draggable to true in packages/lexical-playground/src/nodes/ImageComponent.tsx. Can i take up this issue myself and raise a PR ?

acywatson:
> Fixed this by checking for resizing before setting image draggable to true in packages/lexical-playground/src/nodes/ImageComponent.tsx. Can i take up this issue myself and raise a PR ?

Yes, PR welcome. Thank you!

acywatson:
> Fixed this by checking for resizing before setting image draggable to true in packages/lexical-playground/src/nodes/ImageComponent.tsx. Can i take up this issue myself and raise a PR ?

Yes, PR welcome. Thank you!

PR #3565, opened 2022-12-15
Image gets stuck in resizing state fix (#3562)

SharatChandar:
Resolves : #3562

Issue: When resizing an Image over a long distance in lexical playground, the draggable event of the image is getting triggered. This causes the image to be stuck in the resizing state.

https://user-images.githubusercontent.com/27125472/207942913-2fa4b848-9d2d-48d4-abe3-11d291958643.mp4

Fix: Added a condition to not set the draggable flag if the image is in resizing state. 

https://user-images.githubusercontent.com/27125472/207944605-6c161d87-f56f-44f0-90be-971e40e23187.mp4

### DIFF
--- a/packages/lexical-playground/src/nodes/ImageComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ImageComponent.tsx
@@ -319,7 +319,7 @@ export default function ImageComponent({
     settings: {showNestedEditorTreeView},
   } = useSettings();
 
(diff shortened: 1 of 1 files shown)


fantactuka (approved):
👍

HitaishLokeshEntr:
This same issue is occuring inside of a table. Is there a fix for that ?

PR #3569, opened 2022-12-16
Use the actual node count rather than the dirty node count

trueadm:
We should be using the actual node count here rather than the dirty node count.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -2742,7 +2742,7 @@ export function updateDOMSelection(
   domSelection: Selection,
   tags: Set<string>,
   rootElement: HTMLElement,
-  dirtyLeavesCount: number,
+  nodeCount: number,
 ): void {
   const anchorDOMNode = domSelection.anchorNode;
   const focusDOMNode = domSelection.focusNode;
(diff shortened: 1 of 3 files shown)


thegreatercurve (approved):

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2851,7 +2851,7 @@ export function updateDOMSelection(
       // The downside is that is makes the computation within Lexical more
       // complex, as now, we've sync update the DOM, but selection no longer
       // matches.
-      if (IS_CHROME && dirtyLeavesCount > 1000) {
+      if (IS_CHROME && nodeCount > 1000) {
```
adri1wald, line 2854: @trueadm We've encountered an issue in our editor that arose after this change where the selection shoots back up to the top of the editor after appending nodes to the end of the editor programmatically. Was wondering you have had any idea what might cause this? If nothing immediately obvious comes to mind, happy to produce a minimal repro

trueadm: A repro would be great. This was intended to improve the perf by deferring the setting of selection. Does the issue still persist if you use `setTimeout` or `queueMicroTask`?

Issue #3570, opened 2022-12-17
Bug: call $setSelection in editorState.read()


yf-yang:
I notice `$setSelection` only calls `getActiveEditorState`( without `errorOnReadOnly` and `getActiveEditor`), which means I am able to update selection in a `editorState.read` call. Is it a desired behavior?

To reproduce:
``` typescript
editor.getEditorState().read(() => {
      const selection = $getSelection();
      $setSelection(null);
});
```

Desired behavior: throw an error.

trueadm:
Nice spot. Feel free to put up a fix and we can get it merged :)

yf-yang:
A suggestion: 
In my understanding, `LexicalNode.getLatest` and `LexicalNode.getWritable` not only provide a consistent node from pending editorState, but also checks if write permission is allowed. Similarly, when I need to implement a `$` API function with write permission, I may need to
1. call a built-in `$` write function.
2. call a LexicalNode method that writes something.
3. modifies something managed by myself, but prohibits the function being called in a read context.
For case 3, `errorOnReadOnly` is not exported from the core package. I think providing some utility functions like `getWritable()` that wraps `errorOnReadOnly` is necessary.

trueadm:
Calling `getWritable()` will always do the readonly check, so we advise non core packages to do that. In fact, we've tried to remove calls to `errorOnReadOnly` where we know `getWritable()` is being called in core too – to keep it uniform (we might have missed some places though).

PR #3572, opened 2022-12-17
check read only flag during $setSelection

yf-yang:
Closes https://github.com/facebook/lexical/issues/3570


trueadm (approved):

fantactuka:
It might be used for commenting in read only 

Cc @acywatson

trueadm:
> It might be used for commenting in read only

That's not the same thing right? That's editable mode, which we stopped calling read only mode because of the confusion.

PR #3578, opened 2022-12-17
More extensive fixes for Lexical in iframes

trueadm:
A better solution than what I did in https://github.com/facebook/lexical/pull/3573.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -185,10 +185,11 @@ function $shouldPreventDefaultAndInsertText(
   const anchor = selection.anchor;
   const focus = selection.focus;
   const anchorNode = anchor.getNode();
-  const domSelection = getDOMSelection();
+  const editor = getActiveEditor();
+  const domSelection = getDOMSelection(editor._window);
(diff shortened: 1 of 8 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -958,7 +959,14 @@ function getRootElementRemoveHandles(
...
 
 function onDocumentSelectionChange(event: Event): void {
-  const domSelection = getDOMSelection();
+  const target = event.target as null | Element | Document;
+  const targetWindow =
+    target == null
+      ? null
+      : target.nodeType === 9
+      ? (target as Document).defaultView
+      : (target as Element).ownerDocument.defaultView;
```
thegreatercurve, line 968: I'd say put this in utils if it's used more than once.

thegreatercurve (approved):

packages/lexical/src/LexicalEvents.ts
trueadm: It looks more complicated than it is. It just checks if the event.target is the document and if not, get's the document to then find the window (default view). The reason why we can't just check if `target === document` is because of iframes and cross realms.

PR #3592, opened 2022-12-19
Add missing usages of `$applyNodeReplacement`

thegreatercurve:


trueadm (approved):

acywatson:
Generally you don't need to add $applyNodeReplacement to your subclasses unless you plan to expose them as a library, which I think is probably rare in userland. I would say let's not put this in every creator on the playground for people to just copy and paste everywhere. The problem this is solving is allowing customization of core node behavior without changing the core library itself. The concept is not necessarily useful or applicable in places like TweetNode, where we don't even expose the node.

thegreatercurve:
@acywatson Fair point. I've just added it in to the exported nodes now.

acywatson (approved):

PR #3595, opened 2022-12-19
Preserve empty parent during replacement

fantactuka:
When node is being replaced we should preserve its empty parent (in case it has `canBeEmpty: false`), as otherwise parent might be deleted and replacement fail. E.g., LinkNode has `canBeEmpty: false` and attempting to replace its text node child would fail. 

~~*Will add test, want to see CI results in case I've missed anything*~~

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -645,7 +645,7 @@ export class LexicalNode {
     const prevKey = self.__prev;
     const nextKey = self.__next;
     const parentKey = self.__parent;
-    removeNode(self, false);
+    removeNode(self, false, true);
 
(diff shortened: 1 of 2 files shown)


thegreatercurve (approved):
LGTM. Accepting to unblock, but maybe @trueadm would have the best insights here.

acywatson:
https://github.com/facebook/lexical/pull/1802

It would probably break some things, but I'm still personally of the opinion that this behavior should be the default. It's not intuitive that removing a specific node would also sometimes remove it's parent.

fantactuka:
Pure removal might require parent cleanup but as a part of replacement it does not make much sense

trueadm (approved):

PR #3600, opened 2022-12-20
Add size limit report

fantactuka:
Adds PR size report for lexical core, plain and rich text packages

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -1,0 +1,15 @@
+name: 'size'
+on:
+  pull_request:
+    branches:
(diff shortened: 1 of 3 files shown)


fantactuka:
Looking into failed "size" GH action, it seems like it tries to switch to main branch and run size limit as well which fails as we don't have it in main yet. So we probably will need to merge this PR to get proper size report

acywatson (approved):
Can we get rid of the dangerJS dependency? I think it was attempting to serve the same purpose.

fantactuka:
> dangerJS

Don't see it anywhere

PR #3601, opened 2022-12-20
Select node after updating text content to ensure cursor position at end

JeremyBernier:
Problem I was getting was that `prevSibiling.setTextContent(newTextContent)` was moving the cursor back to the end of `previousText` (ie. the middle of the `newTextContent` string).

So if `newTextContent = previousText + concatText = "users" + ".firstname"`, then the existing code causes the cursor to jump back to the middle after the 's' (eg. `users|.firstname` where '|' is the cursor) instead of the end of the word (`users.firstname|` where it should be)

```
[prevSibiling]node
[users.]firstNa|
[users.]firstNam|
[users.|firstName]
```

Selecting `prevSibiling` after updating the text content seems to fix this.

### DIFF
--- a/packages/lexical-text/src/index.ts
+++ b/packages/lexical-text/src/index.ts
@@ -193,8 +193,8 @@ export function registerLexicalTextEntity<T extends TextNode>(
           if (diff > 0) {
             const concatText = text.slice(0, diff);
             const newTextContent = previousText + concatText;
-            prevSibling.select();
             prevSibling.setTextContent(newTextContent);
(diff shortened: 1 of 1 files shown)


acywatson (approved):

acywatson (changes requested):
Let's look into the failing test - otherwise looks good

acywatson:
I tried this out locally and it does break adjacent hashtags. I can't repro the problem with the text entity implementations that we have on the playground, so closing this for now.

PR #3607, opened 2022-12-21
Fix bad rebase

trueadm:
This line was never meant to be here.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -2840,39 +2840,38 @@ export function updateDOMSelection(
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalSelection.ts
```diff
@@ -2840,39 +2840,38 @@ export function updateDOMSelection(
     }
   }
 
-  if (!tags.has('skip-scroll-into-view'))
```
trueadm, line 2843: This is the rogue line.

fantactuka (approved):

packages/lexical/src/LexicalSelection.ts
fantactuka: image

PR #3608, opened 2022-12-21
[0.8] feat: set the update tag from yjs based on the origin

husseinraoouf:
This PR makes the updates tags for updates coming from yjs change based on origin, so if the changes came from UndoManger the tag is set to `collaboration-history` and if the updates came from other places like websocket the tag is set to `collaboration`, 

this is to be able to know which updates came from the user's interaction and which updates came from other users through yjs

### DIFF
--- a/packages/lexical-react/src/shared/useYjsCollaboration.tsx
+++ b/packages/lexical-react/src/shared/useYjsCollaboration.tsx
@@ -8,7 +8,6 @@
 
 import type {Binding} from '@lexical/yjs';
 import type {LexicalEditor} from 'lexical';
-import {Doc, Transaction, UndoManager, YEvent} from 'yjs';
 
 import {mergeRegister} from '@lexical/utils';
 import {
@@ -36,6 +35,7 @@ import * as React from 'react';
 import {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {createPortal} from 'react-dom';
 import {WebsocketProvider} from 'y-websocket';
+import {Doc, Transaction, UndoManager, YEvent} from 'yjs';
 
 import {InitialEditorStateType} from '../LexicalComposer';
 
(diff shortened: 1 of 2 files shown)


fantactuka:
I'd call it `historic` to allow plugins to rely on the same tag for both collab and non-collab cases, and it should go as a part of release (0.8) as it's a breaking change.

husseinraoouf:
@fantactuka Makes sense I will update the pr

> it should go as a part of release (0.8) as it's a breaking change.

What is the expected release date for 0.8, I am just asking because I will be blocked until it is released

packages/lexical-react/src/shared/useYjsCollaboration.tsx
```diff
@@ -104,8 +104,9 @@ export function useYjsCollaboration(
       events: Array<YEvent<any>>,
       transaction: Transaction,
     ) => {
+      const isFromUndoManger = transaction.origin instanceof UndoManager
```
trueadm, line 107: Maybe we can hoist out `transaction.origin` to `origin` so we don't repeat it. Also we can move the boolean to be within the if statement too.

husseinraoouf: update the pr with the requests changes

husseinraoouf:
@fantactuka @trueadm update the pr with the requested changes, thank you for the review

acywatson:
> Can I ask What is the expected release date for 0.8

Theres's no specific release date for 0.8, but we can do it in the next few days probably.

ivailop7:
Hey @acywatson any outlook on merging this one in soon?

acywatson:
> Hey @acywatson any outlook on merging this one in soon?

It needs to pass the integrity check first - looks like a prettier error

packages/lexical-yjs/src/SyncEditorStates.ts
```diff
@@ -234,7 +235,7 @@ export function syncLexicalUpdateToYjs(
       // types a character and we get it, we don't want to then insert
       // the same character again. The exception to this heuristic is
       // when we need to handle normalization merge conflicts.
-      if (tags.has('collaboration')) {
+      if (tags.has('collaboration') || tags.has('historic')) {
```
acywatson, line 238: Do we need to do this part in order to accomplish the goal set out in the PR description? This seems like a behavioral change for the core, whereas I thought this PR was just trying to pass through the tags.

PR #3609, opened 2022-12-21
Fix getSelectedNode logic for forward selection

moy2010:
Fixes `getSelectedNode` logic for forward selection.

### DIFF
--- a/packages/lexical-playground/src/utils/getSelectedNode.ts
+++ b/packages/lexical-playground/src/utils/getSelectedNode.ts
@@ -22,6 +22,6 @@ export function getSelectedNode(
   if (isBackward) {
     return $isAtNodeEnd(focus) ? anchorNode : focusNode;
   } else {
-    return $isAtNodeEnd(anchor) ? focusNode : anchorNode;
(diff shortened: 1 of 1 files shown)


trueadm:
Can you explain why this change makes sense? As I'm struggling to understand it. Maybe a video or tests of some kind?

moy2010:
> Can you explain why this change makes sense? As I'm struggling to understand it. Maybe a video or tests of some kind?

Sure!

You can test this by copy/pasting this in the playground:

I am a [link](http://example.com/)!

If you start selecting the word `link` starting from the letter `l`, `isLink` will be `true`, but it should stop being `true` as soon as you select the exclamation mark, since that is no longer a `LinkNode`. 

If you do the same bu by selecting the word backwards you should see the correct behaviour.

moy2010:
As for the explanation for why this makes sense, if we take the example given above you can see that the node being selected at the end of the selection (focus) is a `TextNode`, thus the existing logic which returns the `focusNode` if `$isAtNodeEnd(anchor)` returns `true` is incorrect.

trueadm (approved):
Thank you for the explanation

PR #3610, opened 2022-12-21
feat(website): migrates website to typescript

ixahmedxi:
This is the first PR of many that will update the lexical website and its codebase.

This PR introduces the following changes:

- Migrates the codebase to TypeScript.
- Removes unused markdown page example.


packages/lexical-website/src/components/HomepageFeatures/index.tsx
```diff
@@ -6,6 +6,8 @@
  *
  */
 
+/* eslint-disable @typescript-eslint/no-var-requires */
+
 import Translate from '@docusaurus/Translate';
 import clsx from 'clsx';
 import React from 'react';
```
thegreatercurve, line 13:
```suggestion
import * as React from 'react';
```

packages/lexical-website/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "@tsconfig/docusaurus/tsconfig.json",
```
thegreatercurve, line 2: I don't know if this has any specific compiler options that are required by docusaurus, but you can always just extends our base config:

```suggestion
  "extends": "../../tsconfig.json",
```

thegreatercurve:
LGTM. Just the build errors that come up block accepting the PR. To replicate the errors, run `npm run build` or `npm run ci-check` locally.

packages/lexical-website/tsconfig.json
ixahmedxi: We do need these docusaurus compile options for typescript to know the types and the`@site` imports, I couldn't extend from two tsconfigs so hence why this is happening, I could however add the docusaurus typescript options to the base tsconfig but I don't know how you guys feel about that one. I could also copy paste the base config options into the website tsconfig.

thegreatercurve:
Closing this for the moment due to staleness.

Issue #3160, opened 2022-10-11
Bug: `SELECTION_CHANGE_COMMAND` event is not fired on node selection


Piliuta:
Listeners of the `SELECTION_CHANGE_COMMAND` command do not receive events if the selection changes to a node.

Lexical version: 0.5.0

## Steps To Reproduce

1. Use the modified codesandbox example that listens for the event
2. Add some text and horizontal rules (HR button in the toolbar)
3. Do any text selection - in the console, you will see `SELECTION_CHANGE_COMMAND, RangeSelection(...)` output
4. Select horizontal rule - click on it (which is hard 🙈 ) or select using keyboard (set cursor to the start of a paragraph next to it and press ArrowLeft). When the node is selected it is reflected in the history plugin, however, the console log does not output anything.

Link to code example: https://codesandbox.io/s/busy-wing-1eop87

https://user-images.githubusercontent.com/1575198/195141496-f174784b-802a-4f07-80d8-2d0f6aecd5e0.mov

## The current behavior
Decorator node selection is not captured and does not trigger `SELECTION_CHANGE_COMMAND` event.

## The expected behavior
Whether it is a range -> node selection change or node -> node it should trigger the `SELECTION_CHANGE_COMMAND` event.

Piliuta:
> I encountered the same problem, any solution?

I ended up using `registerUpdateListener` https://lexical.dev/docs/concepts/listeners#registerupdatelistener instead of listening for the command. This fixed some issues on my end, although I am not sure if there are any downsides for using it (I did not notice so far).

LvChengbin:
@Piliuta I tried that too, but it doesn't work while using nested editors. I need to listen to `SELECTION_CHANGE_COMMAND` event to get the current active editor and it isn't triggered.

LvChengbin:
@Piliuta  I tried triggered the `SELECTION_CHANGE_COMMAND` by myself and solved my problem by a hack. `Lexical` has some strange code to handle the `selection` and `editor._selection`, it's messy and need to be re-designed and optimized.

LvChengbin:
Does any plan on this issue? My hack code doesn't work after updating lexical to the latest version.

trueadm:
We should probably invoke a SELECTION_CHANGE_COMMAND in the update phase where we trigger update listeners if going from RangeSelection -> NodeSelection by comparing prevSelection to nextSelection. I wonder if we should also invoke if NodeSelection's selected nodes change too?

[...]

PR #3623, opened 2022-12-22
selectionChange fix

trueadm:
Fixes https://github.com/facebook/lexical/issues/3160.

### DIFF
--- a/packages/lexical/src/LexicalUpdates.ts
+++ b/packages/lexical/src/LexicalUpdates.ts
@@ -21,7 +21,7 @@ import type {LexicalNode, SerializedLexicalNode} from './LexicalNode';
 
 import invariant from 'shared/invariant';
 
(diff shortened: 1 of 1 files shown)


thegreatercurve (approved):

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -596,7 +596,19 @@ export function commitPendingUpdates(editor: LexicalEditor): void {
       dirtyLeaves,
     );
   }
-
+  let selectionChanged = false;
+  if (!$isRangeSelection(pendingSelection) && pendingSelection !== null) {
+    if (currentSelection === null) {
+      if (pendingSelection !== null) {
```
acywatson, line 602: Do you need this condition here given the logical AND on line 600?

acywatson (approved):

PR #3639, opened 2022-12-25
[0.8] Apply node transform not only to the original node but also to the overriding node

saint1991:
This PR extends the node overrides: when "withKlass" is specified, node transforms are registered with both nodes by registerNodeTransform. 

### Background of this PR

Due to the implementation of Node Transforms, it is applied only to the exact same type of registered node.

According to [here](https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalEditor.ts#L679), it seems to determine nodes to apply the transform by the result of static LexicalNode#getType().

[Node overrides](https://lexical.dev/docs/concepts/node-replacement) is very powerful but due to the above implementation, Node Transforms added by some plugins do not work with Node overrides.

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -678,34 +678,40 @@ export class LexicalEditor {
     };
   }
 
+  private registerNodeTransformToKlass<T extends LexicalNode>(
+    klass: Klass<T>,
+    listener: Transform<T>,
+  ): RegisteredNode {
+    const type = klass.getType();
+
+    const registeredNode = this._nodes.get(type);
+
+    if (registeredNode === undefined) {
+      invariant(
+        false,
+        'Node %s has not been registered. Ensure node has been passed to createEditor.',
+        klass.name,
+      );
+    }
+    const transforms = registeredNode.transforms;
+    transforms.add(listener as Transform<LexicalNode>);
+
+    return registeredNode;
+  }
+
   registerNodeTransform<T extends LexicalNode>(
     klass: Klass<T>,
     listener: Transform<T>,
   ): () => void {
-    const register = <U extends LexicalNode>(kls: Klass<U>): RegisteredNode => {
-      const type = kls.getType();
+    const registeredNode = this.registerNodeTransformToKlass(klass, listener);
+    const registeredNodes = [registeredNode];
 
-      const registeredNode = this._nodes.get(type);
-
-      if (registeredNode === undefined) {
-        invariant(
-          false,
-          'Node %s has not been registered. Ensure node has been passed to createEditor.',
-          kls.name,
-        );
-      }
-      const transforms = registeredNode.transforms;
-      transforms.add(listener as Transform<LexicalNode>);
-
-      return registeredNode;
-    };
-
-    const registered = register(klass);
-    const registeredNodes = [registered];
-
-    const replaceWithKlass = registered.replaceWithKlass;
+    const replaceWithKlass = registeredNode.replaceWithKlass;
     if (replaceWithKlass != null) {
-      const registeredReplaceWithNode = register(replaceWithKlass);
+      const registeredReplaceWithNode = this.registerNodeTransformToKlass(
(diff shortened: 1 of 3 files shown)


packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -2137,4 +2104,141 @@ describe('LexicalEditor tests', () => {
...
+        const root = $getRoot();
+        const paragraph = $createParagraphNode();
+        const text = $createTextNode('123');
+        root.append(paragraph);
+        paragraph.append(text);
+        expect(text instanceof TestTextNode).toBe(true);
+        expect(text.getTextContent()).toBe('123');
+      });
+
+      expect(onError).not.toHaveBeenCalled();
```
saint1991, line 2143: Fixed the above case a bit because the original code did not detect failure.

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -2137,4 +2104,141 @@ describe('LexicalEditor tests', () => {
...
+    });
+
+    it('node transform against the nodes specified by "replace" should not be applied to the nodes specified by "with" when "withKlass" is not specified', async () => {
+      const onError = jest.fn();
```
saint1991, line 2147: This is the case I faced when using the lexical-code plugin.

The lexical-code plugin registers node transforms with TextNode and CodeHighlightNode but they don't work with node replacement.

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -2137,4 +2104,141 @@ describe('LexicalEditor tests', () => {
...
+        root.append(paragraph);
+        paragraph.append(text);
+        expect(text instanceof TestTextNode).toBe(true);
+        expect(text.getTextContent()).toBe('123');
+      });
+
+      expect(onError).not.toHaveBeenCalled();
+    });
+
+    it('node transform to the nodes specified by "replace" should not be applied to the nodes specified by "with" when "withKlass" is not specified', async () => {
```
saint1991, line 2146: This is a case I faced when using the lexical-code plugin.

The lexical-code plugin registers node transform with CodeHighlightNode but it does not work with node overrides.

packages/lexical/src/LexicalEditor.ts
```diff
@@ -676,23 +682,38 @@ export class LexicalEditor {
     klass: Klass<T>,
     listener: Transform<T>,
   ): () => void {
-    const type = klass.getType();
+    const register = <U extends LexicalNode>(kls: Klass<U>): RegisteredNode => {
```
trueadm, line 685: Avoid having an inline function, and instead extract this out.

saint1991: @trueadm 
Fixed it.

packages/lexical/src/LexicalEditor.ts
```diff
@@ -346,6 +347,8 @@ export function createEditor(editorConfig?: {
         with: <T extends {new (...args: any): any}>(
           node: InstanceType<T>,
         ) => LexicalNode;
+
+        withKlass?: Klass<LexicalNode>;
```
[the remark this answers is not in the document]
trueadm: Go for it :)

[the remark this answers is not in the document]
saint1991: @trueadm 
Sorry I tried it but it was not a good solution than I expected because there is no way to determine type parameters for value type of RegisteredNodes type.

I made it at the last commit but I'm happy to revert it if you want.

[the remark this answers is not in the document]
trueadm: Thats fine. :)

trueadm (approved):
This looks good, but it's a breaking change, so will need to go in 0.8

saint1991:
Thanks!

PR #3664, opened 2023-01-02
Add flow vscode extension

doong-jo:
If the repository has a file with a flow extension, but does not have an extension, syntax of that file is not supported, making it difficult to read.

As a result, developers have to find and install related extensions themselves. To improve this, add it to the extension list.


trueadm:
The bigger issue is that having both TS and Flow causes issues too.

thegreatercurve:
I'm not sure if we necessarily need this as an explicit recommendation for VSCode. We mainly just retain the Flow types for our own internal purposes.

Issue #3602, opened 2022-12-20
Bug: Odd behaviour with $getRoot().getTextContent() when copying and pasting it is stripping new lines.


OrderAndCh4oS:
When copying and pasting content in the rich text editor, hitting return causes the last piece of pasted text to strip all new lines.

Lexical version: ^0.7.4

## Steps To Reproduce

1. `console.log($getRoot().getTextContent())` in an `onChange handler.
2. Type some text with a couple of line breaks.
3. Copy the text with the line breaks and paste it underneath.
4. Hit return, and you should see the output has stripped line breaks from the logged content. 

Link to code example: [Example Project](https://codesandbox.io/s/lexical-get-text-content-bug-lymscb?file=/src/editor.js)

## The current behavior

Lines are being concatenated when reading $getRoot().getTextContent()

## The expected behavior

Text should maintain line breaks when reading $getRoot().getTextContent()

https://user-images.githubusercontent.com/1910955/208721042-09b044c7-d0bd-420f-bbe2-6c574769bda3.mov

PR #3669, opened 2023-01-03
Fix bug in reconciler sub-tree text content cache

trueadm:
Fixes https://github.com/facebook/lexical/issues/3602.

I think a bit of logic was missed from https://github.com/facebook/lexical/pull/1993. It's hard to fully know though. Would be good to have the eyes of @zurfyx on this.

### DIFF
--- a/packages/lexical/src/LexicalReconciler.ts
+++ b/packages/lexical/src/LexicalReconciler.ts
@@ -176,6 +176,9 @@ function createNode(
       const endIndex = childrenSize - 1;
       const children = createChildrenArray(node, activeNextNodeMap);
       createChildrenWithDirection(children, endIndex, node, dom);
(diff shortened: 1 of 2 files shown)


zurfyx (approved):

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -176,6 +176,9 @@ function createNode(
       const endIndex = childrenSize - 1;
       const children = createChildrenArray(node, activeNextNodeMap);
       createChildrenWithDirection(children, endIndex, node, dom);
+      if ($textContentRequiresDoubleLinebreakAtEnd(node)) {
```
acywatson, line 179: Is it possible to write a test for the scenario described in the issue?

PR #3671, opened 2023-01-04
Remove ts-ignore for createContext

yf-yang:
The reason tsc complains is that createContext is a generic function. Not sure if other similar calls need to be fixed. They look good.


thegreatercurve (approved):

thegreatercurve:
@yf-yang Thanks for the fix! Would you mind signing the CLA and I can merge this in?

acywatson (approved):
Thank you!

PR #3672, opened 2023-01-04
fix: type-ahead menu scroll problem

kirakiray:
fix an issue where the menu would cause the editor to scroll.

#### before

https://user-images.githubusercontent.com/5945154/211185806-6659b340-cf12-4e52-bb32-716862ba0c44.mp4

#### after

Only scrollIntoView will be done when there are not enough positions.

https://user-images.githubusercontent.com/5945154/210514951-e3e60bdf-497a-4b8e-bd68-455842d38f73.mp4

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -84,7 +84,10 @@ const scrollIntoViewIfNeeded = (target: HTMLElement) => {
   if (container) {
     const parentNode = target.parentNode as HTMLElement | null;
 
-    if (parentNode) {
+    if (
+      parentNode &&
+      /auto|scroll/.test(getComputedStyle(parentNode).overflow)
+    ) {
       const parentRect = parentNode.getBoundingClientRect();
 
       if (parentRect.top + parentRect.height > window.innerHeight) {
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -80,13 +80,19 @@ export type MenuRenderFn<TOption extends TypeaheadOption> = (
 
 const scrollIntoViewIfNeeded = (target: HTMLElement) => {
   const container = document.getElementById('typeahead-menu');
+
   if (container) {
-    const containerRect = container.getBoundingClientRect();
-    const targetRect = target.getBoundingClientRect();
-    if (targetRect.bottom > containerRect.bottom) {
+    const parentNode = target.parentNode as HTMLElement | null;
```
acywatson, line 85: Will this still work correctly if the target's parent node isn't the scroll container? It seems like we're making an assumption about the parent node here that I'm not sure will always be valid.

kirakiray: Added new code
If the parent container is not scrolling, use scrollIntoview directly to make sure it works
The template code for TypeaheadMenu has been finalized, and it is expected that the parent element will be the scrolling container for a long time to come

trueadm (approved):
Looks good to me.

trueadm:
Please can you sign the CLA?

kirakiray:
> Please can you sign the CLA?

signed

PR #3673, opened 2023-01-04
Fix: convert html 's' tag to node for strickethrough format is missing

moraeso:
importDOM method of the LexicalTextNode.ts is missing the node conversion portion for the HTML 's' tag.


acywatson (approved):
Thank you!

trueadm (approved):

PR #3676, opened 2023-01-05
Fix bug with ArrowDown/ArrowUp in RichTextPlugin

trueadm:
We should have been checking the decorator is inline or not.


a-xin (approved):
LGTM

Issue #3678, opened 2023-01-05
Bug: text node broken after removing selection as entering a composition symbol like korean.


conatus-recedendi:
Lexical version: 0.7.5 (latest)

This bug only happens in Chrome, Safari and not in FF.

## Steps To Reproduce

1. Select more than two text nodes (above one should be fully wrapped)
2. Enter any symbol affecting composition event like Korean symbol.
3. selection will be removed but symbol doesn't appear and can't write any word at the focusing text node 

https://user-images.githubusercontent.com/74358372/210831408-eef66347-c4aa-4baa-846d-5240e8895517.mov

## The current behavior

When the selected text nodes are replaced, the new node has empty TextContent and can't be modified. (until newline)

## The expected behavior

When the selected text nodes are replaced, the new one will have contents I typed and work correctly.

conatus-recedendi:
I found a temporary solution about this bug, but I'm not sure it doesn't have any side effects.

conatus-recedendi:
I found a temporary solution #3679  about this bug, but I'm not sure it doesn't have any side effects.

PR #3679, opened 2023-01-05
fix: force to set TextNode's content when compositionKey is invalid

conatus-recedendi:
fix bug in #3678

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -635,10 +635,14 @@ export function $updateTextNodeFromDOMContent(
       }
       const parent = node.getParent();
       const prevSelection = $getPreviousSelection();
+      const compositionKey = $getCompositionKey();
+      const nodeKey = node.getKey();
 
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalUtils.ts
```diff
@@ -635,10 +635,14 @@ export function $updateTextNodeFromDOMContent(
...
       const parent = node.getParent();
       const prevSelection = $getPreviousSelection();
+      const compositionKey = $getCompositionKey();
+      const nodeKey = node.getKey();
 
       if (
         node.isToken() ||
-        ($getCompositionKey() !== null && !isComposing) ||
+        (compositionKey !== null &&
+          nodeKey === compositionKey &&
```
trueadm, line 644: I'm struggling with this one. If we're not composing, then `compositionKey` will be `null` and will never match `nodeKey`?

trueadm (approved):

trueadm:
Thanks for this!

PR #3684, opened 2023-01-06
Add width and height to serialized image nodes

ebads67:
This PR adds width and height to image nodes when they are serialized or get converted back to Lexical nodes.


trueadm (approved):

trueadm:
Thank you.

PR #3690, opened 2023-01-09
insert node after parent when selection at the end

### LINKED ISSUE #3681: Bug: divider is inserted before list items when selection is at end
Lexical version: 0.7.5

## Steps To Reproduce

1. Create a list item.
2. Put selection at the end.
3. Insert a divider.

Note that this works fine for other block nodes I've tested, such as paragraphs or headings.

## The current behavior

The divider will be inserted before.

https://user-images.githubusercontent.com/8862967/210907681-5a33cb60-16f2-434f-8bb6-0064ece741a2.mp4

## The expected behavior

The divider is inserted after.

aldinezi:
This PR fixes inserting the divider before list item when the selection is at the end of the last list item. 

Fixes #3681

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -350,9 +350,19 @@ export function $insertNodeToNearestRoot<T extends LexicalNode>(node: T): T {
         splitNode = focusNode;
         splitOffset = focusOffset;
       }
-      const [, rightTree] = $splitNode(splitNode, splitOffset);
-      rightTree.insertBefore(node);
-      rightTree.selectStart();
+      const selectionAtLastOffset = focusNode.__text.length === focusOffset;
+      if (
+        splitNode.__type === 'listitem' &&
+        splitNode.__next == null &&
+        selectionAtLastOffset
+      ) {
+        const parentNode = splitNode.getParentOrThrow();
+        parentNode.insertAfter(node);
+      } else {
+        const [, rightTree] = $splitNode(splitNode, splitOffset);
+        rightTree.insertBefore(node);
+        rightTree.selectStart();
+      }
     }
   } else {
     if ($isNodeSelection(selection) || DEPRECATED_$isGridSelection(selection)) {


packages/lexical-utils/src/index.ts
```diff
@@ -350,9 +350,19 @@ export function $insertNodeToNearestRoot<T extends LexicalNode>(node: T): T {
         splitNode = focusNode;
         splitOffset = focusOffset;
       }
-      const [, rightTree] = $splitNode(splitNode, splitOffset);
-      rightTree.insertBefore(node);
-      rightTree.selectStart();
+      const selectionAtLastOffset = focusNode.__text.length === focusOffset;
+      if (
+        splitNode.__type === 'listitem' &&
```
trueadm, line 355: We shouldn't be looking for specific types here. We should be using the methods on those classes for signal instead, as this allows this logic to expand for other types of custom list items that people might create, or just other elements that work in the same manner.

trueadm: That's still coupling it to list items, which when I read this logic, seems for generic and likely applicable to other types of element that follow the same style? Can the `canInsertAfter` method give signal here?

aldinezi: True, it is coupling to list logic here. 
I have used `canInsertAfter`.

packages/lexical-utils/src/index.ts
```diff
@@ -350,9 +350,20 @@ export function $insertNodeToNearestRoot<T extends LexicalNode>(node: T): T {
         splitNode = focusNode;
         splitOffset = focusOffset;
       }
-      const [, rightTree] = $splitNode(splitNode, splitOffset);
-      rightTree.insertBefore(node);
-      rightTree.selectStart();
+      const selectionAtLastOffset =
+        focusNode.getTextContent().length === focusOffset;
```
trueadm, line 354: Doesn't this also need to check that focusNode is a text node? Selection can be of type `element` instead of `text` which would mean that selection isn't inside a text node and this logic might fail as focusOffset would be the index of the child inside the element, rather than that of the text node content.

packages/lexical-utils/src/index.ts
trueadm: Perfect

trueadm (approved):

trueadm:
This seems to be breaking some unit tests and end to end tests?

aldinezi:
> This seems to be breaking some unit tests and end to end tests?

@trueadm Apologies. It should be fine now.

trueadm:
Sorry, It looks like e2e tests are still failing.

aldinezi:
@trueadm do I need any particular package for running e2e tests locally? It gets stuck on this screen when I run `npm run debug-test-e2e`.
Screenshot 2023-01-10 at 15 48 11

aldinezi:
> Sorry, It looks like e2e tests are still failing.

Hey @trueadm, now it doesn't split the paragraph into two when inserting youtube embed, for example. 
I believe it would be solved if this case was handled only for the list nodes and it's extends.
So, if we replace `canInsertAfter` with `isListItemNode` it should all work. 
You can check my latest commit. Thank you.

trueadm:
I get that it's tempting to go back to that, but it might actually make more sense to understand why that method doesn't work first. As forcing it to be coupled with list items is also not ideal here. Maybe we need a new method like `canInsertAfter` to apply the heuristic you're looking for?

aldinezi:
> I get that it's tempting to go back to that, but it might actually make more sense to understand why that method doesn't work first. As forcing it to be coupled with list items is also not ideal here. Maybe we need a new method like `canInsertAfter` to apply the heuristic you're looking for?

Any suggestions or directions?

packages/lexical-utils/src/index.ts
```diff
@@ -350,9 +353,24 @@ export function $insertNodeToNearestRoot<T extends LexicalNode>(node: T): T {
...
-      rightTree.selectStart();
+      const selectionAtTextEnd =
+        $isTextNode(focusNode) && focusNodeTextLength === focusOffset;
+      if (
+        $isListItemNode(splitNode) &&
+        splitNode.__next == null &&
+        selectionAtTextEnd
+      ) {
+        const parentNode = splitNode.getParentOrThrow();
+        if ($isRootNode(parentNode)) {
```
acywatson, line 364: This is an invariant, as a ListItemNode can't be a direct child of RootNode.

packages/lexical-utils/src/index.ts
```diff
@@ -350,9 +353,24 @@ export function $insertNodeToNearestRoot<T extends LexicalNode>(node: T): T {
         splitNode = focusNode;
         splitOffset = focusOffset;
       }
-      const [, rightTree] = $splitNode(splitNode, splitOffset);
-      rightTree.insertBefore(node);
-      rightTree.selectStart();
+      const selectionAtTextEnd =
+        $isTextNode(focusNode) && focusNodeTextLength === focusOffset;
+      if (
+        $isListItemNode(splitNode) &&
```
acywatson, line 359: Can we just traverse up until we find a root or shadow root?

thegreatercurve:
Closing this for the moment due to PR becoming stale, but happy to re-open if issues above can be addressed.

PR #3693, opened 2023-01-09
Add config to updateDOM in ParagraphNode 

abelsj60:
The standard ParagraphNode does not expose the config property available to its updateDOM method. 

This can make for a nasty TypeScript problem when creating a custom ParagraphNode for the replacement API. The reason is that TypeScript will insist that there's no such config property. But there is. This PR will convince TS otherwise!

I encountered this problem while building a LinedCodeNode. This PR is 1/2 that will let it work without patches. 

-j

### DIFF
--- a/packages/lexical/src/nodes/LexicalParagraphNode.ts
+++ b/packages/lexical/src/nodes/LexicalParagraphNode.ts
@@ -49,7 +49,11 @@ export class ParagraphNode extends ElementNode {
     }
     return dom;
   }
-  updateDOM(prevNode: ParagraphNode, dom: HTMLElement): boolean {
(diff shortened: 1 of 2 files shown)


trueadm (approved):

abelsj60:
@trueadm Sorry, do I need to do something else here? I see the failing and timed out tests, but I'm not sure how this pull request can cause them. 

I ran `npm run test-e2e-collab-ci-webkit`, for example, on my local main branch (a fork), and everything failed or timed out there, too. 

Maybe there's something wrong with some of the test suites? Please let me know, I'm at a bit of a loss.

PR #3722, opened 2023-01-14
Make highlight code plugin movement optional

moy2010:
Add option to get movement across highlighted code behave as with normal text


zurfyx (commented):
You can override commands, would this work for you?

```
editor.registerCommand(MOVE_TO_END, () => true, COMMAND_PRIORITY_MEDIUM);
```

Note that the priority is higher than the one is CodeHighlighter

moy2010:
@zurfyx provided a better solution!

PR #3724, opened 2023-01-14
Fix behaviour of copy&pasting code highlight nodes

### LINKED ISSUE #3743: Bug: Copy&Paste between code blocks doesn't work
Lexical version: 0.76

## Steps To Reproduce

1. Create a code block with some content (i.e. `const text = "Hello world!"`).
2. Select the content via keyboard or mouse, and copy it.
3. Create a new empty code block.
4. Paste the copied text.

## The current behavior

The empty code block is turned into a paragraph when pasting the text.

## The expected behavior

The code block is maintained, and the pasted text is part of its contents.

moy2010:
Fix behaviour of copy&pasting code highlight nodes by properly wrapping them in a `CodeNode`.

Fixes #3743 .

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -6,6 +6,7 @@
  *
  */
 
+import {$createCodeNode, $isCodeHighlightNode} from '@lexical/code';
 import {$generateHtmlFromNodes, $generateNodesFromDOM} from '@lexical/html';
 import {$createListNode, $isListItemNode} from '@lexical/list';
 import {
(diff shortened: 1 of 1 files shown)


zurfyx (commented):
Thank you! Given the list fix, this looks reasonable to me. My main concern with this hacky clipboard fix is that the code module is large, looking at:

```
import 'prismjs/components/prism-clike';
import 'prismjs/components/prism-javascript';
import 'prismjs/components/prism-markup';
import 'prismjs/components/prism-markdown';
import 'prismjs/components/prism-c';
import 'prismjs/components/prism-css';
import 'prismjs/components/prism-objectivec';
import 'prismjs/components/prism-sql';
import 'prismjs/components/prism-python';
import 'prismjs/components/prism-rust';
import 'prismjs/components/prism-swift';
import 'prismjs/components/prism-typescript';
```

Perhaps we should pass it via props (Strategy pattern) or something. @fantactuka wdyt?

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -201,7 +202,15 @@ function $basicInsertStrategy(
 
     const isLineBreakNode = $isLineBreakNode(node);
 
-    if (
+    if ($isCodeHighlightNode(node)) {
```
trueadm, line 205: We definitely want to avoid pulling in specifics about code into this module. Rather than checking against the class, maybe we could leverage a method on the node class itself that can give signal? What heuristic would a code highlight node have that we could add to the base TextNode?

trueadm: Maybe some kind of validation around the parent? `isValidParent()`? This might also fit into the schema work that @tylerjbainbridge is working on.

PR #3726, opened 2023-01-16
Update listener in toc

Stopa:
Previously, the Table of Contents plugin didn't update correctly if a heading was updated. For example, if you enable ToC on the Lexical Playground and drag one heading after another (with the `DraggableBlockPlugin`), the ToC didn't update.

This PR adds handling of such events and includes minor code cleanup:

- Create a separate TS type for ToC that can be used throughout the plugin without needing to duplicate
- Introduces a utility function to transform a heading node to an array to be included in the ToC
- Updates the ToC only once per listener trigger - previously it was updated once for each mutation


thegreatercurve (approved):
LGTM! Thanks for the contribution.

PR #3731, opened 2023-01-16
Tweak CodeNode to allow for overrides

abelsj60:
This pull request is the first of at least two that are also needed to implement my LinedCodeNode without patches. 

I've extended my work on this node to allow it to override the official CodeNode via the override API so it can work with — I hope — Lexical's existing core library packages, notably @lexical/markdown.

_In this pull request:_ 

- Reduced priority level for several importDOM converters so an overriding CodeNode can take over
  - Without this, the official CodeNode will win the race to control how importDOM elements are handled
- Exposed the editor config in the CodeNode's updateDOM method
- Added a more generic return type to the CodeNode's collapseAtStart method so CodeNode overrides can apply more nuanced logic

Coming soon-ish:

A pull request to make the new `setBlocksType_experimental` function more flexible.

Recent related pull requests:

### DIFF
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -129,7 +129,7 @@ export class CodeNode extends ElementNode {
     return {
       // Typically <pre> is used for code blocks, and <code> for inline code styles
       // but if it's a multi line <code> we'll create a block. Pass through to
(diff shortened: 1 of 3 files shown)


fantactuka:
Loos good!

fantactuka (approved):

PR #3738, opened 2023-01-17
Attach checkbox click listeners to editor root element instead of document

### LINKED ISSUE #3657: Bug: Impossible to click on an image or a checkbox in a Chakra modal
I'm using a lot of editor instance in [Chakra UI modals](https://chakra-ui.com/docs/components/modal) and I get a persistent bug:

It's impossible to click on an image to resize it or to toggle a checkbox.

You can test it here: https://storybook.rolebase.io/?path=/story/richeditor--in-modal

Code of the story "InModal" I used to test it: https://github.com/lonestone/rolebase.io/blob/main/src/components/molecules/editor/lib/RichEditor.stories.tsx#L168 

I'm searching why but I don't understand what's going on.

Lexical version: 0.7.5

thegreatercurve:
Closes #3657

### DIFF
--- a/packages/lexical-react/src/LexicalCheckListPlugin.tsx
+++ b/packages/lexical-react/src/LexicalCheckListPlugin.tsx
@@ -35,8 +35,6 @@ export function CheckListPlugin(): null {
   const [editor] = useLexicalComposerContext();
 
   useEffect(() => {
-    listenPointerDown(editor.getRootElement());
-
     return mergeRegister(
       editor.registerCommand(
         INSERT_CHECK_LIST_COMMAND,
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/LexicalCheckListPlugin.tsx
```diff
@@ -35,6 +35,8 @@ export function CheckListPlugin(): null {
   const [editor] = useLexicalComposerContext();
 
   useEffect(() => {
+    listenPointerDown(editor.getRootElement());
```
trueadm, line 38: Better to be a root listener, in case the editor root element changes.

thegreatercurve: Ah yeah, of course!

trueadm (approved):

PR #3742, opened 2023-01-17
Remove preventDefault from $copyToClipboardEvent

### LINKED ISSUE #3698: Bug: copy Equation content cause Uncaught Error: Expected valid LexicalSelection
image

Lexical version: 0.7.5

## Steps To Reproduce

1. create equation.
2. click the equation, select content and press cmd+c.
3. check console and you can see 'Expected valid LexicalSelection' Error.

## The current behavior
equation content can NOT be copied.

## The expected behavior
equation content can be copied.

### LINKED ISSUE #3717: Bug: Copy/Cut from input field within DecoratorNode component (for example PollComponent)
Lexical version: v0.7.6

## Steps To Reproduce

1. Add Poll to the sandbox and add some text to option fields
2. Try to copy/cut from the option field

https://user-images.githubusercontent.com/4542049/212292551-6dac68b4-05d9-4151-b3e1-f4fb5bac734f.mov

## The current behavior
On Copy/Cut from the Poll text field error gets thrown:
```
main.22f383cc.js:21 Uncaught Error: Expected valid LexicalSelection
    at ev (main.22f383cc.js:21:247217)
    at av (main.22f383cc.js:21:250901)
    at main.22f383cc.js:21:252025
    at Vf (main.22f383cc.js:21:179169)
    at Hf (main.22f383cc.js:21:179802)
    at Wf (main.22f383cc.js:21:182038)
    at Tm.update (main.22f383cc.js:21:209325)
    at qf (main.22f383cc.js:21:178724)
    at HTMLDivElement.o (main.22f383cc.js:21:208073)
```

## The expected behavior
Text gets copied/cut to the clipboard.

thegreatercurve:
The error logic and prevent default in $copyToClipboardEvent is causing an issue for input or textareas in within decorator nodes. The native selection is overridden, but when no Lexical selection is found, an error is thrown.

Closes #3717 
Closes #3698 

https://user-images.githubusercontent.com/7748470/212957866-84e70cce-3f58-4f2d-8e20-7f391a53cace.mov

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -34,6 +34,7 @@ import {
   DEPRECATED_$isGridSelection,
   DEPRECATED_GridNode,
   GridSelection,
+  isSelectionWithinEditor,
   LexicalEditor,
   LexicalNode,
   NodeSelection,
@@ -572,6 +573,20 @@ function $copyToClipboardEvent(
   editor: LexicalEditor,
   event: ClipboardEvent,
 ): boolean {
+  const domSelection = window.getSelection();
+  if (!domSelection) {
+    return false;
+  }
+  const anchorDOM = domSelection.anchorNode;
+  const focusDOM = domSelection.focusNode;
+  if (
+    anchorDOM == null ||
+    focusDOM == null ||
+    !isSelectionWithinEditor(editor, anchorDOM, focusDOM)
+  ) {
+    return false;
+  }
+  event.preventDefault();
   const clipboardData = event.clipboardData;
   if (clipboardData === null) {
     return false;
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -47,7 +47,7 @@ export function $getHtmlContent(editor: LexicalEditor): string {
   const selection = $getSelection();
 
   if (selection == null) {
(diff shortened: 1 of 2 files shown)


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -572,7 +572,6 @@ function $copyToClipboardEvent(
   editor: LexicalEditor,
   event: ClipboardEvent,
 ): boolean {
-  event.preventDefault();
```
trueadm, line 575: We don't want to remove it, as it's serving a purpose for where we want to control the copy/paste. We just want to not do it if the activeElement is not the root element, right? In fact in that case we should just early return the whole function.

thegreatercurve: Okay, I've updated the check to use `isSelectionWithinEditor` from core which has the activeElement.

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -47,7 +48,7 @@ export function $getHtmlContent(editor: LexicalEditor): string {
   const selection = $getSelection();
 
   if (selection == null) {
-    throw new Error('Expected valid LexicalSelection');
+    return '';
```
trueadm, line 51: I don’t think we want to throw here right? This can happen in valid cases sometimes.

trueadm (approved):

PR #3747, opened 2023-01-18
fix(lexical-playground): NewTablePlugin is not draggable

smzst:
The Lexical Playground environment has two components, Table and Table (Experimental), the former draggable but not the latter.
I thought it would be desirable to have a component that is draggable like the Table component, so I fixed this as a bug.

<details>
<summary>Before/After</summary>

Before:
before-newtableplugin-is-not-draggable

After:
after-newtableplugin-is-not-draggable

</details>


thegreatercurve (approved):
LGTM! Thanks

PR #3749, opened 2023-01-18
Fix $generateJSONFromSelectedNodes with a custom selection passed in

### LINKED ISSUE #3745: Bug: $generateJSONFromSelectedNodes does not work with multiple different selections
**Lexical version:** 0.7.6

Hi! Thank you all for the great work! It's been awesome using Lexical so far.

I ran into this behavior recently and I wanted to check if it's a bug or if I'm doing something wrong.

I have an editor with some content in it. I'm trying to "cut" the editor content at the focus point (notice the cursor in front of "underline text") and save the "Head" and "Tail" parts into separate variables. I want the a selection that is in the middle of anode to cut the node where the selection is, and save the prefix & postfix in their respective head & tail variables. I also want it to **not visually select the text.**

To do this, I found `$generateJSONFromSelectedNodes`, which seems to do exactly what I want. However, it doesn't work when I pass selection created by `$createRangeSelection` into the function. In those cases, `$generateJSONFromSelectedNodes` only cuts  the **current** node into it's correct head & tail sections. It ignores the rest of the editor and doesn't go to the top or to the bottom of the editor.

The behavior I want only seems to work when directly modifying the editor selection returned from `$getSelection`. However, modifying the selection directly also visually selects the text, which is what I don't want.

Let me know if anything is confusing or more info is needed, happy to provide.

Screenshot 2023-01-17 at 7 19 53 PM

## Steps To Reproduce

**Confusing behavior**
This code with `headSelection` and `tailSelection` only ends up generating the JSON for the *current node*, and not until the top / bottom of the editor, even though `headSelection` should be anchored at the top, and `tailSelection` should be focused at the end.

```
editor.update(() => {
      if (!$isRangeSelection(selection)) return;

      const root = $getRoot();
      const previousSelectionAnchor = selection.anchor;
      const previousSelectionFocus = selection.focus;

      const firstChild = root.getFirstChild();
      if (!firstChild) return;

      const headSelection = $createRangeSelection();
      headSelection.anchor.set(
        firstChild.getKey(),
        0,
        firstChild.getType() as any
      );
      headSelection.focus.set(
        previousSelectionFocus.key,
        previousSelectionFocus.offset,
        previousSelectionFocus.type
      );

      const headNodes = $generateJSONFromSelectedNodes(
        editor,
        headSelection
      ).nodes;
      console.log("HEAD NODES: ", headNodes);

      const lastDescendent = root.getLastDescendant();
      if (!lastDescendent) return;

      const tailSelection = $createRangeSelection();
      tailSelection.anchor.set(
        previousSelectionFocus.key,
        previousSelectionFocus.offset,
        previousSelectionFocus.type
      );
      tailSelection.focus.set(
        lastDescendent.getKey(),
        lastDescendent.getTextContentSize(),
        lastDescendent.getType() as any
      );

      const tailNodes = $generateJSONFromSelectedNodes(
        editor,
        tailSelection
      ).nodes;

      console.log("TAIL NODES: ", tailNodes);
    });
```

**Working**
Notice the change from using `headSelection` and `tailSelection` to just using `selection`, which is returned on line 2) from `$getSelection()`.

```
editor.update(() => {
      if (!$isRangeSelection(selection)) return;

      const root = $getRoot();
      const previousSelectionAnchor = selection.anchor;
      const previousSelectionFocus = selection.focus;

      const firstChild = root.getFirstChild();
      if (!firstChild) return;

      selection.anchor.set(firstChild.getKey(), 0, firstChild.getType() as any);
      selection.focus.set(
        previousSelectionFocus.key,
        previousSelectionFocus.offset,
        previousSelectionFocus.type
      );

      const headNodes = $generateJSONFromSelectedNodes(editor, selection).nodes;
      console.log("HEAD NODES: ", headNodes);

      const lastDescendent = root.getLastDescendant();
      if (!lastDescendent) return;

      selection.anchor.set(
        previousSelectionFocus.key,
        previousSelectionFocus.offset,
        previousSelectionFocus.type
      );
      selection.focus.set(
        lastDescendent.getKey(),
        lastDescendent.getTextContentSize(),
        lastDescendent.getType() as any
      );

      const tailNodes = $generateJSONFromSelectedNodes(editor, selection).nodes;

      console.log("TAIL NODES: ", tailNodes);
    });
```

willwang888:
Fixes #3745 

Previously, passing in a Selection into `$generateJSONFromSelectedNodes` would result in the wrong JSON because the the main selection from `$getSelection()` was being used. We should be using the target selection that was passed in instead of the main editor selection.

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -8,7 +8,11 @@
 
 /* eslint-disable no-constant-condition */
 import type {EditorConfig, LexicalEditor} from './LexicalEditor';
-import type {RangeSelection} from './LexicalSelection';
+import type {
+  GridSelection,
+  NodeSelection,
+  RangeSelection,
+} from './LexicalSelection';
 import type {Klass} from 'lexical';
 
 import invariant from 'shared/invariant';
@@ -222,8 +226,10 @@ export class LexicalNode {
     return false;
   }
 
-  isSelected(): boolean {
-    const selection = $getSelection();
+  isSelected(
+    _selection?: null | RangeSelection | NodeSelection | GridSelection,
+  ): boolean {
+    const selection = _selection || $getSelection();
     if (selection == null) {
       return false;
     }
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -398,7 +398,8 @@ function $appendNodesToJSON(
   currentNode: LexicalNode,
   targetArray: Array<BaseSerializedNode> = [],
 ): boolean {
-  let shouldInclude = selection != null ? currentNode.isSelected() : true;
+  let shouldInclude =
+    selection != null ? currentNode.isSelected(selection) : true;
   const shouldExclude =
     $isElementNode(currentNode) && currentNode.excludeFromCopy('html');
   let target = currentNode;
--- a/packages/lexical/src/__tests__/unit/LexicalNode.test.ts
+++ b/packages/lexical/src/__tests__/unit/LexicalNode.test.ts
@@ -14,6 +14,7 @@ import {
   TextNode,
 } from 'lexical';
 
+import {$createRangeSelection} from '../..';
 import {LexicalNode} from '../../LexicalNode';
 import {$createParagraphNode} from '../../nodes/LexicalParagraphNode';
 import {$createTextNode} from '../../nodes/LexicalTextNode';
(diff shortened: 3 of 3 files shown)


packages/lexical/src/LexicalNode.ts
```diff
@@ -222,8 +226,10 @@ export class LexicalNode {
     return false;
   }
 
-  isSelected(): boolean {
-    const selection = $getSelection();
+  isSelected(
+    _selection?: null | RangeSelection | NodeSelection | GridSelection,
```
zurfyx, line 230: Given that many Lexical operations rely on selection, I wonder if it's best to just set it beforehand (in client's code) instead of passing a parameter every time. I.e.

```
let previousSelection = $getSelection();
$setSelection(x);
// $operation
$setSelection(previousSelection)
```

willwang888: @zurfyx I added selection as a parameter because:
1.  Many other function calls in this function have selection passed in, e.g. `sliceSelectedTextNodeContent` and `node.extractWithChild`. Given the parent function takes in a selection as an argument, which is then passed into those other function calls as args, I wonder if it would be inconsistent to set the selection at the top when we're already passing it into other places. I assume it's done like this because setting the main selection on the editor probably has implications for other things elsewhere in the tree, but I'm not positive.
2. It opens more one off use cases to check if a node is selected within a custom selection that is created / modified by `$createRangeSelection` or others, specifically if we don't want the selection to trigger an visual updates to the DOM.

Let me know if you all think differently — happy to change if it makes more sense but wanted to give my thoughts first 🙂

willwang888: The functions in `clipboard.ts` also don't seem to have any unit tests — perhaps they're accounted for E2E? 🤔

acywatson: I think @willwang888 has the right idea here. I actually wrote this exact same fix the exact same way before I realized there was already a fix. I don't think setting the selection makes sense because then you need to sometimes restore is back to what it previously was as you may not always wants to actually change the selection of the editor when you all this function.

For me , the new parameter makes sense - the two drawbacks I see are:

1) We don't use "_" prefix anywhere else that I know of - I know you are doing this to avoid a collision, but I would suggest we handle that in the function implementation. I might be wrong about how prevalent "_" prefix is in our codebase - let me know if you disagree/feel strongly.

2) Adding a single optional property constrains how we can make additional backwards compatible changes to this API in the future. We won't be able to add non-optional properties without breaking consumers. I decided it was probably unlikely that we'd add more arguments, and also if we have to then a breaking change is not the end of the world.

acywatson: > The functions in `clipboard.ts` also don't seem to have any unit tests — perhaps they're accounted for E2E? 🤔

Yes, clipboard functions are covered in e2e.  CopyAndPaste.spec.ts I believe.

acywatson (approved):

packages/lexical/src/LexicalNode.ts
willwang888: > We don't use underscore prefix anywhere else that I know of - I know you are doing this to avoid a collision, but I would suggest we handle that in the function implementation. I might be wrong about how prevalent undescore prefix is in our codebase - let me know if you disagree/feel strongly.

@acywatson  Agree, I don't see args prefixed with undersscore — happy to change this. Just pushed a fix to rename the function internal `selection` to `targetSelection` and removed the underscore from the arg!

willwang888:
@acywatson Friendly bump — is there anything else I need to do to merge this PR?

tylerjbainbridge (approved):
Makes sense to me. Thanks for improving the API @willwang888

PR #3751, opened 2023-01-19
Move horizontal rule plugin to lexical/react

### LINKED ISSUE #3680: Feature: official HorizontalRulePlugin
HorizontalRulePlugin is currently part of the playground code, but we're using it and we had to copy it into our codebase to use it, which comes with disadvantages like missing updates and having to do manual maintenance. This also feels weird because the node itself is actually shipped as part of the lexical packages.

thegreatercurve:
Closes #3680


DaniGuardiola:
Beat me to it, thanks :pray:

zurfyx (approved):

PR #3753, opened 2023-01-19
Edit CSS to fix marker alignment issue in Safari

### LINKED ISSUE #3752: Bug: [Playground] On Safari, using any list, cursor appears to the left of the ::marker until typing commences
Lexical version: Latest

## Steps To Reproduce

1. Go to playground on safari
2. Create a bulleted list
3. See location of cursor

## The expected behavior
Cursor should be to the right of `::marker`

Screen Shot 2023-01-19 at 10 54 02 AM

thegreatercurve:
There appears to be some issues with `list-style-position: inside;` on Safari, but we can probably just remove it as we're setting a margin-left any way. It's the same as putting the marker on the outside.

Closes #3752

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -277,42 +277,30 @@
 .PlaygroundEditorTheme__ol1 {
   padding: 0;
   margin: 0;
-  margin-left: 16px;
-  list-style-position: inside;
 }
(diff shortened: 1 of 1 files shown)


Amethystix:
Hi! Thanks for taking on this bug I reported so quickly; could you also apply the same style fix to ordered lists? It's the same problem there, you should just have to apply this to ol1 - ol5's markers and remove the same properties as that of unordered lists!

zurfyx (approved):

Issue #3740, opened 2023-01-17
Bug: Copy Pase of List elements b/w Playground & any Custom Instance is producing extra lines


gaurav21r:
**Since this bug pertains to a base level feature (Lists & Clipboard) I'm flagging as URGENT**

Lexical version: 0.6.0 to latest

## Steps To Reproduce

1. Make a list of any arbitrary depth in the playground https://playground.lexical.dev/
<img width="783" alt="image" src="https://user-images.githubusercontent.com/1716462/212914315-04747033-3a3f-4615-8236-5255bbe51053.png">

2. Copy paste this list (along with the heading & any other element(s) ) in any arbitrary custom instance of Lexical. For example, the *Rich Text* sandbox in the website https://lexical.dev itself, will give this:

<img width="458" alt="image" src="https://user-images.githubusercontent.com/1716462/212914782-04ea89b2-c314-4114-aac9-26978faf056d.png">

Link to code example: Codesandbox link of Rich Text Editor from the website itself: https://codesandbox.io/s/lexical-rich-text-example-5tncvy?from-embed

## The current behavior
There is an extra `\n` being added to every list item that has a sub list.

## The expected behavior
The playground editor and all other editors should behave consistently.

## Probable cause.
I couldn't understand the code much but I think this line may be the offender.
https://github.com/facebook/lexical/blob/68c64e942967e86238760858af259117fa9a8c7c/packages/lexical-clipboard/src/clipboard.ts#L126

<img width="614" alt="image" src="https://user-images.githubusercontent.com/1716462/212917454-bf9df20b-2c70-43ce-be6c-2f5277d119df.png">

The comment seems to suggest something to the issue.

bschwartz10:
Experiencing the same behavior of the insertion of an empty `editor-listitem` when parsing html content back into the editor

**parsing logic**
```
const root = $getRoot()
const parser = new DOMParser()
const dom = parser.parseFromString(value, 'text/html')
const nodes = $generateNodesFromDOM(editor, dom)         
root.append(...nodes)
```

**rendered text**
<img width="717" alt="Screen Shot 2023-01-19 at 11 58 56 AM" src="https://user-images.githubusercontent.com/21000661/213513192-16732ca7-fd73-4a79-87d1-7498eb60754d.png">

**Lexical editor w/ extra empty `editor-listitem`**
[...]

PR #3757, opened 2023-01-20
Fix list normalization on HTML paste.

acywatson:


fantactuka (approved):

gaurav21r:
@fantactuka Let me Pull master and confirm so we can close #3740

PR #3758, opened 2023-01-20
Fix equation transformer in Playground

focux:
The equation markdown transformer is not working when your LaTeX is only one character long.  For instance, you can use this text:
```
What is the $x$-intercept of line $k$?
```

**Current Result**
image

**Expected**
image

You can take a look at the new RegEx [here](https://regex101.com/r/2GgxP1/1)


fantactuka (approved):

fantactuka:
Looks good! Could you sign CLA, then we can merge this PR

focux:
@fantactuka I already signed the CLA but seems like it takes a bit to update it (?)

Issue #3761, opened 2023-01-21
Bug: can't use some shortcuts in inputs inside DecoratorNodes 


samyosm:
When we have an input inside a DecoratorNode and we try to do `Ctrl+Del` for example, which is suppose to delete one full word, it doesn't do anything.

[Screencast from 2023-01-21 05:57:46 PM.webm](https://user-images.githubusercontent.com/99157490/213890396-16c87be2-55a0-4a08-85dd-610883bef773.webm)

Lexical version:

## Steps To Reproduce

1. Go to playgroud
2. Insert a poll
3. Write something
4. Press `Ctrl+Del`
5. Notice nothing happening

## The current behavior
Nothing happens when `Ctrl+Del` in an input.

## The expected behavior
A word is deleted in an input when `Ctrl+Del` is pressed.

elbotho:
I'm also encountering this with e.g. `CMD+A` on mac.
@alisouran is there a PR with your fix?

PR #3766, opened 2023-01-23
Fix(#3761): Disable Lexical Key Events for DecoratorNodes

alisouran:


packages/lexical/src/LexicalEvents.ts
```diff
@@ -851,6 +851,28 @@ function onKeyDown(event: KeyboardEvent, editor: LexicalEditor): void {
...
+  }
+  const anchorDOM = domSelection.anchorNode;
+  const focusDOM = domSelection.focusNode;
+  let isInEditor = true;
+
+  updateEditor(editor, () => {
+    if (
+      anchorDOM !== null &&
+      focusDOM !== null &&
+      !isSelectionWithinEditor(editor, anchorDOM, focusDOM)
```
trueadm, line 866: Why not use this above rather than having the upateEditor?

alisouran: Because I get the following error in isSelectionWithinEditor catch:

```
Error: Unable to find an active editor. This method can only be used synchronously during the callback of editor.update().
    at getActiveEditor (LexicalUpdates.ts:109:7)
    at getNodeFromDOMNode (LexicalUtils.ts:398:18)
    at $getNearestNodeFromDOMNode (LexicalUtils.ts:413:18)
    at isSelectionCapturedInDecoratorInput (LexicalUtils.ts:118:22)
    at isSelectionWithinEditor (LexicalUtils.ts:140:8)
    at onKeyDown (LexicalEvents.ts:865:6)
    at HTMLDivElement.<anonymous> (LexicalEvents.ts:1089:15)
```

and it returned `false`.

trueadm: You can do a read of the existing editor state instead of always causing an update here.

trueadm: Interestingly we can refactor some bits to avoid that error. `$getNearestNodeFromDOMNode` already accepts passing of a given editor state. What if also allow `isSelectionCapturedInDecoratorInput` and `isSelectionWithinEditor` to take an optional editorState too? Then maybe we can avoid this update entirely.

alisouran: I did, but I think we have the same error due to the `editor.getActiveEditor()` in the `getNodeFromDOMNode` function.

trueadm: @alisouran That shouldn't be an issue, as `getNodeFromDOMNode` already takes an optional second argument for `editorState` that will skip the `editor.getActiveEditor` lookup if used.

alisouran: @trueadm I refactored the `isSelectionWithinEditor` to:
```ts
export function isSelectionWithinEditor(
  editor: LexicalEditor,
  anchorDOM: null | Node,
  focusDOM: null | Node,
): boolean {
  try {
    return (
      rootElement !== null &&
      rootElement.contains(anchorDOM) &&
      rootElement.contains(focusDOM) &&
      // Ignore if selection is within nested editor
      anchorDOM !== null &&
      !isSelectionCapturedInDecoratorInput(
        anchorDOM as Node,
        editor._editorState, // optional editorState
      ) &&
      getNearestEditorFromDOMNode(anchorDOM) === editor
    );
  } catch (error) {
    return false;
  }
}
```
and the `isSelectionCapturedInDecoratorInput` to:
```ts
export function isSelectionCapturedInDecoratorInput(
  anchorDOM: Node,
  editorState?: EditorState, // optional editorState
): boolean {
  const activeElement = document.activeElement as HTMLElement;

  if (activeElement === null) {
    return false;
  }
  const nodeName = activeElement.nodeName;

  return (
    $isDecoratorNode($getNearestNodeFromDOMNode(anchorDOM, editorState)) &&
    (nodeName === 'INPUT' ||
      nodeName === 'TEXTAREA' ||
      (activeElement.contentEditable === 'true' &&
        // @ts-ignore iternal field
        activeElement.__lexicalEditor == null))
  );
}
```

but when I log the error in catch, I get:
```
Error: Unable to find an active editor. This method can only be used synchronously during the callback of editor.update().
    at getActiveEditor (LexicalUpdates.ts:109:7)
    at getNodeFromDOMNode (LexicalUtils.ts:404:18)
    at $getNearestNodeFromDOMNode (LexicalUtils.ts:419:18)
    at isSelectionCapturedInDecoratorInput (LexicalUtils.ts:121:22)
    at isSelectionWithinEditor (LexicalUtils.ts:143:8)
    at onKeyDown (LexicalEvents.ts:870:6)
    at HTMLDivElement.<anonymous> (LexicalEvents.ts:1092:15)
```

trueadm: Rightly so. I've put up https://github.com/facebook/lexical/pull/3766 to address the issue you had. Thank you for the help working this one out. I also noticed that we need to capture the correct DOM selection, so opted to use our helper there (important for iframes).

PR #3769, opened 2023-01-24
Fix line delete bug in history

abelsj60:
This pull request fixes a bug that prevents the deletion of empty lines from entering the history stack. 

I debugged it as I thought it was related to my LinedCodeNode, but the bug's in the core. 

To reproduce:

1. Go to the Playground
2. Select the end of a paragraph 
3. Hit enter
4. Move the cursor to the start of the new empty line
5. Hit backspace to delete it
6. Hit Command + Z (or its Windows corollary)
7. Behold — nothing happens

The same can be achieved with lists, for instance. The reason is a function named `isTextNodeUnchanged` in the history package. When there's just one `dirtyLeaf`, it checks to see if it's an unchanged TextNode. If it is, the package bails. 

Unfortunately, as it stands, this function sweeps line deletes into the bin, too. 

I believe this can be fixed by checking to see if there's a mismatch between the current and previous selection.


trueadm:
Looks like there failing e2e tests for this?

abelsj60:
> Looks like there failing e2e tests for this?

I'm not sure. I ran `npm run test-unit` before filing and everything passed, as long as I swapped @ts-expect-error for @ts-ignore, which otherwise wreaks havoc for me. Everything passed, one skipped, which is what I always see, I think.

I'll have to look at this. 

Thing is, this happens every time I PR with the e2e's, I believe. I think you had to rerun one of my last ones to get it to work. 

And its sister PR also failed, but John (?) merged it in, and so I basically now just look at the e2es as advisory.

I don't know, I'll take a look later (sigh). Maybe it's real this time, but mostly, I feel powerless.

abelsj60:
> Looks like there failing e2e tests for this?

@trueadm, @acywatson  I could use some help. 

I think the e2e results are a red herring. I dug into them and walked away with no real confidence in them. 

**A. Test confidence** 

The tests that broke on my second run are different than the first run.

Run one problems: 
- `e2e-collab-mac (16.15.1, webkit)`
- `e2e-linux (16.15.1, chromium, rich-text, legacy-events)`

Run two problems: 
- `e2e-collab-linux (16.15.1, chromium)`
- `e2e-collab-windows (16.15.1, chromium)`

If you run them five more times, I bet you'll get five different results. Maybe one time they'll all pass.

**B. Timestamps and the editor**

These tests seem to break at random, for random reasons, maybe having to do with hanging scripts and loops that run too often (I got both errors as overlays when running Playwright in debug mode). Lexical relies on timestamps in its update process, so if something hangs, I can easily see the editor failing at random because it takes the wrong update path.

That's not all. @trueadm reported this bug to Playwright awhile back: 

https://github.com/microsoft/playwright/issues/5929

I found it still at play when running `AutoLinks.spec.mjs:60` in debug mode in the terminal and from the VSCode extension. When FireFox ran, the cursor failed to move left, and so an exclamation mark was inserted badly. 

**C. Manual replication**

I replicated a handful of the "failing" tests" by hand, I think in the complaining browser, and correctly produced the expected results. 

(Although, the second run claims problems in Windows and Linux, neither of which can I access.)

**D. The sidelines**

Finally, I don't think most of the tripped tests call my code? It should mostly only get called when a new line is deleted. At present, if you hit enter after an element node, such as Paragraph, go to the new line (offset 0), then press backspace to delete the line, the history stack breaks. This means nothing will happen when you hit ctrl + z. Very disconcerting.

I'm not sure what to do with this PR or its sibling. I think my code is probably fine. 

The only thing I could see is charting out which tests deal with deleting lines from offset 0 and making a table of the results I achieve by hand. I'm not really convinced that's any more confidence inducing than this message, however.

**_What do you recommend._** I'd like to see the two PRs merge as they help Lexical, too...? 

-j

trueadm (approved):

PR #3770, opened 2023-01-24
Better tab handling

abelsj60:
This pull request fixes a problem with the use of the hidden tab character ('\t'). 

In some scenarios, new characters can't be inserted when it's present. 

Here's a brief description of the new functionality from a new comment in `$updateSelectedTextFromDOM`:

      // The newly typed character may be missing from the DOM's textContent
      // when some types of tokenized nodes include a tab. For instance,
      // a LinedCodeNode <code><div><codeToken></div></code> will
      // result in the following misbehavior:

      // a. /tconst --> type 'd' at offset 1 --> /tconst
      //  - Missing 'd'
      // b. /tconst --> type 'd' at offset 3 --> /tcodnst
      //    --> type 'd' at offset 3 --> /tcodnst
      //  - Missing second 'd'

      // In these cases, we can fix the problem by manually inserting the
      // newly typed character where we know it should have been.

Here's an edited version of the [details I left on Discord](https://discord.com/channels/953974421008293909/955972012541628456/1066405018120503367): 

> I'm trying to figure out a tab character problem in a nested LexicalNode.
> 
> Say I have a TextNode, a code highlight, that contains a "\t" special character. 
> 
> - It's on its own like this due to Prisma tokenization: LexicalNode<\t>LexicalNode<const>... 
> - The \t is a hidden character, it's length is 0 unless stringified.
> - When the user types a character, the input event fires on Root and Lexical intercepts the request to decide if it should  handle the insert. 
> - If it takes over, Lexical uses the value of `event.data` via `$getNearestNodeFromDOMNode` to insert the text. If it doesn't take over, Lexical uses the value of `editor._window.getSelection()` (via `$getDOMSelection`) to check if there's a mismatch with the current node (apparently, called a "backing node"). 
> If there's a mismatch, an update occurs. If there isn't, it doesn't.
> - Unlike `event.data`, the value of `getSelection` is the full string of the browser's  AnchorNode (which Lexical semantically mirrors internally).
> 
> This is all fine and dandy. 
> 
> The problem is that, if there is a tab character at the start of the `.getSelection` value, one of three things will happen: 
> 
> 1. The new character is not included in the DOM text if the browser's AnchorNode is at offset 0 (AKA, one index after \t. 
> 2. A new character will only be included one time if the browser's AnchorNode is at a point greater than offset 0, e.g., \tcondst, but never \tconddst. 
> 3. A new character can't be added before the \t. 
> 
> Of course, the value of `event.data` is correct in all three cases.
> 
> In paragraphs, Lexical generally seems to use CSS instead of tab characters, so this problem doesn't seem to manifest. Further, I see different behavior than what you might see with the official CodeNode as my node places code in lines (divs). As a result, Lexical's algorithms behave differently. 


abelsj60:
@acywatson This is the sibling pull request I mentioned in https://github.com/facebook/lexical/pull/3769. It also contains e2es I don't understand. 

The first run had two failures that succeed on the the second run.

The second run had two new claimed failures:

`e2e-collab-linux`
`e2e-collab-windows`

On the third run, only `e2e-collab-linux` purported to fail. 

I can continue poking at it, but I'm really not sure what to make of this. I noticed that in GitHub actions, most current pull requests have at least one failing (or cuationary) test. 

I feel as if there is a piece of this picture I don't understand, and I don't know how to proceed as a result. 

Basically, I'm confused. 

Can you help?

-j

packages/lexical/src/LexicalUtils.ts
```diff
@@ -573,6 +573,35 @@ export function $updateSelectedTextFromDOM(
...
+
+      // a. /tconst --> type 'd' at offset 1 --> /tconst
+      //  - Missing 'd'
+      // b. /tconst --> type 'd' at offset 3 --> /tcodnst
+      //    --> type another 'd' at offset 3 --> /tcodnst
+      //  - Missing second 'd'
+
+      // In these cases, we can fix the problem by manually inserting the
+      // newly typed character where we know it should have been.
+      const hasTabCharacter = textContent.includes('\t');
```
trueadm, line 589: Isn't tab always at the start?

packages/lexical/src/LexicalUtils.ts
```diff
@@ -573,6 +573,35 @@ export function $updateSelectedTextFromDOM(
     let textContent = getAnchorTextFromDOM(anchorNode);
     const node = $getNearestNodeFromDOMNode(anchorNode);
     if (textContent !== null && $isTextNode(node)) {
+      // The newly typed character may be missing from the DOM's textContent
```
trueadm, line 576: I wonder if there's some way we can contain this heuristic to only code highlight nodes? Maybe if we introduce a new method on TextNode called `canContainTabs(): boolean` which is `false` by default and then implement it on CodeHighlightNode to return true. Then we can only validate this use case here and avoid applying it too all text insertions?

abelsj60: I like what you're saying. These helper methods are a good idea. 

I would've been afraid to be so specific with core on my own, though. 

To be clear, this problem comes up when you contain highlight nodes in lines/divs, which is what I'm doing. I'm pretty committed to offering it up once done, which it nearly is. Maybe others will use it, too.

You don't really see this problem arise right now in the official code package because it doesn't contain code in lines. In essence, Lexical has side stepped it. 

I'm using my own HighlightNode (`LinedCodeTextNode`), which has some specialized behavior over the official one. I'd be happy to add `canContainTabs()` to mine and target this code to check for it. 

Offhand, that...seems like it'd work — and ward against unexpected bugs.

Would that be enough?

abelsj60: > I wonder if there's some way we can contain this heuristic to only code highlight nodes? Maybe if we introduce a new method on TextNode called `canContainTabs(): boolean` which is `false` by default and then implement it on CodeHighlightNode to return true. Then we can only validate this use case here and avoid applying it too all text insertions?

I like what you're saying. These helper methods are a good idea.

I would've been afraid to be so specific with core on my own, though.

To be clear, this problem comes up when you contain highlight nodes in lines/divs, which is what I'm doing. I'm pretty committed to offering it up once done, which it nearly is. Maybe others will use it, too.

You don't really see this problem arise right now in the official code package because it doesn't contain code in lines. In essence, Lexical has side stepped it.

I'm using my own HighlightNode (LinedCodeTextNode), which has some specialized behavior over the official one. I'd be happy to add canContainTabs() to mine and target this code to check for it.

Offhand, that...seems like it'd work — and ward against unexpected bugs.

Would that be enough?

trueadm: The entire code node right now sucks (what we have). It should have been lined to begin with, with a div around each line to prevent wrapping and keep line numbers consistent. So anything you've done in this area would be advantageous and I'd love to bring it back into core.

packages/lexical/src/LexicalUtils.ts
abelsj60: Oh! Well, I wasn't sure.  

I think I probably used .includes because I know it will always find the tab while .startsWith or text[0]...may not?

Your approach is smarter. I'm happy to try it. I just always worry about being too smart with code. Every time I get smart with code, I regret it. Well, I regret it in life, too. But, I don't usually have to worry about that.

packages/lexical/src/LexicalUtils.ts
abelsj60: > The entire code node right now sucks (what we have). It should have been lined to begin with, with a div around each line to prevent wrapping and keep line numbers consistent. So anything you've done in this area would be advantageous and I'd love to bring it back into core.

That's motivating, thanks. I'll shift `canContainTabs` into place and pull it into this PR.

packages/lexical/src/LexicalUtils.ts
abelsj60: I know what it is. 

If you have a tab at the start of a line, move the anchor point to offset 0, then type, you'll get text before the tab. The tab gets pushed right or left. I guess I figured using .includes was more resilient to all these unknown scenarios.

trueadm: Makes sense! We definitely want to gate this to only code highlight nodes then, otherwise we're adding overhead to all other text inputs, which is every keypress.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -573,6 +573,35 @@ export function $updateSelectedTextFromDOM(
...
+      //  - Missing 'd'
+      // b. /tconst --> type 'd' at offset 3 --> /tcodnst
+      //    --> type another 'd' at offset 3 --> /tcodnst
+      //  - Missing second 'd'
+
+      // In these cases, we can fix the problem by manually inserting the
+      // newly typed character where we know it should have been.
+      const hasTabCharacter = textContent.includes('\t');
+
+      if (data && hasTabCharacter) {
```
trueadm, line 591: better to define `data`, as what if the value is "0"?

abelsj60: I just suddenly realized what you meant. 

I'll add `data.length > 0` to the condition. I think that's what you mean? I'll push the change after the tests finish. Last time I pushed a commit in the middle of running tests, GitHub swallowed it.

trueadm (approved):

PR #3771, opened 2023-01-24
fix: styling information of paragraph are missing while importing dom

Rajatava:
fixed the issue of missing styling information of paragraph while importing dom

### DIFF
--- a/packages/lexical/src/nodes/LexicalParagraphNode.ts
+++ b/packages/lexical/src/nodes/LexicalParagraphNode.ts
@@ -151,7 +151,7 @@ function convertParagraphElement(element: HTMLElement): DOMConversionOutput {
   const node = $createParagraphNode();
   if (element.style) {
     node.setFormat(element.style.textAlign as ElementFormatType);
-    const indent = parseInt(element.style.textIndent) / 20;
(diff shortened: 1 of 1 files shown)


packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -144,8 +147,16 @@ export class ParagraphNode extends ElementNode {
   }
 }
 
-function convertParagraphElement(): DOMConversionOutput {
-  return {node: $createParagraphNode()};
+function convertParagraphElement(element: HTMLElement): DOMConversionOutput {
+  const node = $createParagraphNode();
+  if (element.style) {
+    node.setFormat(element.style.textAlign as ElementFormatType);
+    const indent = parseInt(element.style.textIndent) / 20;
```
trueadm, line 154: Missing the radix base number from parseInt

trueadm (approved):
Thanks for dong this.

packages/lexical/src/nodes/LexicalParagraphNode.ts
Rajatava: thanks for the review, updated accordingly.

PR #3774, opened 2023-01-24
Letting other nodes with priority 1 to try to convert div elements if the element is not a codeNode

ebads67:
According to the docs the function mapped to a element name should return null if it's able to convert the html element to its node type. The codeNode only does this check inside the conversion function for `<div>` elements. So it does not let other nodes with priority 1 to try to convert the element to their type.

### DIFF
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -142,10 +142,17 @@ export class CodeNode extends ElementNode {
             }
           : null;
       },
-      div: (node: Node) => ({
-        conversion: convertDivElement,
-        priority: 1,
-      }),
+      div: (node: Node) => {
+        // domNode is a <div> since we matched it by nodeName
+        const div = node as HTMLDivElement;
+        if (!isCodeElement(div)){
+          return null;
+        }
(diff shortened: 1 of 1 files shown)


packages/lexical-code/src/CodeNode.ts
```diff
@@ -142,10 +142,17 @@ export class CodeNode extends ElementNode {
             }
           : null;
       },
-      div: (node: Node) => ({
-        conversion: convertDivElement,
-        priority: 1,
-      }),
+      div: (node: Node) => {
```
acywatson, line 145: This is a good catch!

My one concern here is that I'm not sure the that we always want to return null if isCodeElement is false. If you look at the implementation of convertDivElement, CodeNode still tries to handle divs that pass the isCodeChildElement check. So, with  this PR we'd stop handling those correctly, I think.

We could repeat the isCodeChildElement check here, as well, but that's sort of weird since we're coupling the condition to the implementation of convertDivElement.  My recommendation might be to extend the return type of convertDivElement to include "null" and just do a null check to see if we should return it as a conversion object or not.

Thoughts?

PR #3782, opened 2023-01-26
Remove remaining deprecated hooks

thegreatercurve:


zurfyx (approved):

acywatson (approved):
I assume we'll add this to 0.8?

thegreatercurve:
> I assume we'll add this to 0.8?

Tbh, I was just going to do it as a minor release. The hooks have been marked as deprecated since we made the library public, so not sure it necessarily constitutes a breaking change

thegreatercurve:
> > > I assume we'll add this to 0.8?
> > 
> > 
> > Tbh, I was just going to do it as a minor release. The hooks have been marked as deprecated since we made the library public, so not sure it necessarily constitutes a breaking change
> 
> I understand your perspective here, but it's definitely a breaking change. I doubt it will break many consumers, practically, but removing any public API can cause builds to suddenly fail if they're pulling in the latest patch version (which is common).

I think it's fine. I've done it with the other modules and it's not an unexpected breaking change if the module has forever been marked as deprecated.

acywatson:
@thegreatercurve Can you rebase this?

PR #3786, opened 2023-01-27
Fix copy paste wrong order in TableOfContents

Yidaotus:
Fixes a bug where the TableOfContents would display the wrong order of headings if the user cuts/copy and pastes headers.
It seems `getPreviousSiblings` does not return the correct order, so iterate over the siblings instead.

To reproduce:

- Enter a few headings and enable ToC in playground.
- Copy or cut one heading
- Paste the heading above or below another heading
- ToC will have the wrong order

Also cleans up the code a little bit:

- Don't redefine TableOfContentsEntry type in the TableOfContent Plugin. Just export the already existing type and use this instead.
- Wrong keys in heading render map. Keys should be attached to the direct childs.

### DIFF
--- a/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx
@@ -143,10 +143,9 @@ function TableOfContentsList({
         {tableOfContents.map(([key, text, tag], index) => {
           if (index === 0) {
             return (
-              <div className="normal-heading-wrapper">
+              <div className="normal-heading-wrapper" key={key}>
                 <div
                   className="first-heading"
-                  key={key}
                   onClick={() => scrollToNode(key, index)}
                   role="button"
                   tabIndex={0}>
@@ -162,9 +161,9 @@ function TableOfContentsList({
               <div
                 className={`normal-heading-wrapper ${
                   selectedKey === key ? 'selected-heading-wrapper' : ''
-                }`}>
+                }`}
+                key={key}>
                 <div
-                  key={key}
                   onClick={() => scrollToNode(key, index)}
                   role="button"
                   className={indent(tag)}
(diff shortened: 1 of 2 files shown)


packages/lexical-react/src/LexicalTableOfContents__EXPERIMENTAL.tsx
```diff
@@ -158,9 +162,10 @@ export default function LexicalTableOfContentsPlugin({
             } else if (mutation === 'updated') {
               const newHeading = $getNodeByKey<HeadingNode>(nodeKey);
               if (newHeading !== null) {
-                const prevHeading = newHeading
-                  .getPreviousSiblings()
-                  .find($isHeadingNode);
+                let prevHeading = newHeading.getPreviousSibling();
+                while (prevHeading && !$isHeadingNode(prevHeading)) {
```
zurfyx, line 166:
```suggestion
                while (prevHeading !== null && !$isHeadingNode(prevHeading)) {
```

packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx
```diff
@@ -11,7 +11,9 @@ import type {NodeKey} from 'lexical';
 import './index.css';
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
-import LexicalTableOfContents__EXPERIMENTAL from '@lexical/react/LexicalTableOfContents__EXPERIMENTAL';
+import LexicalTableOfContents__EXPERIMENTAL, {
+  TableOfContentsEntry,
```
zurfyx, line 15: Let's move this to a type

```
import type {TableOfContentsEntry} from '@lexical/react/LexicalTableOfContents__EXPERIMENTAL';`
```

zurfyx (approved):
Thank you! Just a couple nits

PR #3787, opened 2023-01-27
fix(lexical-playground): fix bugs related to ComponentPickerPlugin

smzst:
I have made the following two fixes:

1. https://github.com/facebook/lexical/commit/70eed93ed3231f24a12ae3dabf017a1035429772: When the ComponentPicker options are opened at a position overlapping the table, strings in the cell overlap and become difficult to read.
    * Intention: Another idea I had was to increase the z-index of the ComponentPicker options, but it doesn't feel right that only the table cells are floating, so I modified it as shown in PR.
1. https://github.com/facebook/lexical/commit/d67b6024ec0bb73d46b0b2d9a89d25baad291fe4: Missing `Table (Experimental)` in ComponentPicker options (maybe not a bug).

<details>
<summary>Before/After</summary>

### Before

before-bug-related-component-picker-plugin

### After

after-bug-related-component-picker-plugin

</details>

### DIFF
--- a/packages/lexical-playground/src/nodes/TableComponent.tsx
+++ b/packages/lexical-playground/src/nodes/TableComponent.tsx
@@ -622,7 +622,6 @@ function TableCell({
       ) : (
         <>
           <div
-            style={{position: 'relative', zIndex: 3}}
             dangerouslySetInnerHTML={{
               __html:
                 editorStateJSON === ''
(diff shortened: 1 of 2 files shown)


smzst:
E2E test CI is failing, but I don't think it's due to this fix, so I'm making it ready for review. 
I think it's the same problem as https://github.com/facebook/lexical/pull/3769#issuecomment-1407076527.

zurfyx (approved):
I can't repro the overlapping bug before and after, but overall the fix LGTM, thank you for adding the experimental tables onto the list

cc @tylerjbainbridge

tylerjbainbridge (approved):
Not sure why those position/zindex styles were on the Experimental Table Component in the first place, maybe @trueadm has insights. 

Looks good to me though!

PR #3789, opened 2023-01-28
Fix misplaced key in TableOfContents plugin

zignis:


zignis:
#3786 includes this fix, if merged.

zurfyx:
Thank you zignis, https://github.com/facebook/lexical/pull/3786 is ready to go and also bundles in some additional fixes, would be OK with merging his that would save him from merge conflicts?

Issue #3701, opened 2023-01-10
Bug: invalid state - list items wrapped in paragraph nodes instead of list


DaniGuardiola:
Lexical version: 0.7.5 (I think? I used the playground just now)

## Steps To Reproduce

1. Replicate what happens in the video below.

https://user-images.githubusercontent.com/8862967/211608329-c38f45cd-949f-46b5-bf33-203c45f39fb3.mp4

## The current behavior

List items are wrapped in paragraph blocks.

## The expected behavior

List items are *converted* into paragraph blocks individually.

## Insight / solution proposal

AFAIK, this bug is related to `$setBlocksType_experimental` and/or list normalization. Not sure what the correct way to handle this would be, but it makes sense that this happens because `$setBlocksType` does exactly that: turn the top level blocks contained in the selection into a different block type (the old `$wrapNodes`). After that happens, I'm guessing the list normalization should be responsible for turning each list item into a separate paragraph and flattening the parent paragraph. But I don't really know.

thegreatercurve:
@EgonBolton as the author of the above method, would you have any thoughts on how this should be implemented?

EgonBolton:
Oh I hadn't noticed this thread and I duplicated it last week! https://github.com/facebook/lexical/issues/3776

Yes, I already tried some things and I have an idea to solve it. If you want, assign it to me and I'll try to do it soon. @thegreatercurve

PR #3821, opened 2023-02-04
Fix converting nodes with nested lists

### LINKED ISSUE #3776: Bug: converting nodes doesn't work well with nested lists
Lexical version: 0.7.6

## Steps To Reproduce

Select and turn paragraphs the following:

Case 1:
```html
<ul>
  <li>1</li>
  <li>
    <ul>
      <li>1.1</li>
    </ul>
  </li>
</ul>
```

Case 2:
```html
<ul>
  <li>
    <ul>
      <li>1.1</li>
    </ul>
  </li>
</ul>
```

Link to code example: Playground

## The current behavior
Case 1:
```html
<p>1</p>
<p>
  <p>
    <li>1.1</li>
  </p>
</p>
```
Case 2:
```html
<ul>
  <li>
    <p>1.1</p>
  </li>
</ul>
```

## The expected behavior
case 1:
```html
<p>1</p>
<p style="padding-inline-start: 20px;">1.1</p>
```
case 2:
```html
<p style="padding-inline-start: 20px;">1.1</p>
```

Note: $wrapNodes (the previous implementation of $setBlocksType) failed with case 2 but not with case 1.

EgonBolton:
Resolves #3776 and #3701

Pending
- [ ] Add an explanatory description to the PR
- [ ] Add tests
- [ ] Fix bug when converting the following and trying to write immediately:
```
1
- 2
- 3
4
```


packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -157,42 +157,35 @@ export class ListItemNode extends ElementNode {
...
-        });
-      }
-      this.remove();
-
-      if (childrenLength === 1) {
-        list.remove();
+    if (!$isListNode(list)) return replaceWithNode;
+    if (list.__first === this.__key) {
+      list.insertBefore(replaceWithNode);
+    } else if (list.__last === this.__key) {
```
acywatson, line 166: Same with __last and __key

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -157,42 +157,35 @@ export class ListItemNode extends ElementNode {
...
+      list.insertAfter(replaceWithNode);
+    } else {
+      // Split the list
+      const newList = $createListNode(list.getListType());
+      let nextsibling = this.getNextSibling();
```
acywatson, line 171: nit: camel-case here

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -157,42 +157,35 @@ export class ListItemNode extends ElementNode {
...
+      list.insertAfter(replaceWithNode);
+      replaceWithNode.insertAfter(newList);
+    }
+    if (includeChildren) {
+      this.getChildren().forEach((child: LexicalNode) => {
+        replaceWithNode.append(child);
+      });
+    }
+    this.remove();
+    if (list.__size === 0) {
```
acywatson, line 186: impl detail - use getter methods

packages/lexical/src/LexicalUtils.ts
```diff
@@ -265,53 +265,23 @@ function internalMarkParentElementsAsDirty(
...
+  let prevSibling = node.getPreviousSibling();
+  if (prevSibling) prevSibling = prevSibling.getWritable();
+  const writableParent = oldParent.getWritable();
+
+  if (!prevSibling) writableParent.__first = node.__next;
+  else prevSibling.__next = node.__next;
+  if (!nextSibling) writableParent.__last = node.__prev;
+  else nextSibling.__prev = node.__prev;
+
+  writableParent.__size--;
```
acywatson: It's ok to do this here, since we're in the core

packages/lexical-selection/src/range-selection.ts
```diff
@@ -42,26 +42,25 @@ import {getStyleObjectFromCSS} from './utils';
  */
 export function $setBlocksType_experimental(
   selection: RangeSelection | GridSelection,
-
   createElement: () => ElementNode,
 ): void {
   if (selection.anchor.key === 'root') {
+    // Should we actually handle this case?
```
acywatson, line 48: Not sure if this is adding value - what was your train of thought re: whether or not we should handle this.

packages/lexical-selection/src/range-selection.ts
```diff
@@ -42,26 +42,25 @@ import {getStyleObjectFromCSS} from './utils';
...
-
   const nodes = selection.getNodes();
-  if (selection.anchor.type === 'text') {
-    let firstBlock = selection.anchor.getNode().getParent() as LexicalNode;
-    firstBlock = (
-      firstBlock.isInline() ? firstBlock.getParent() : firstBlock
-    ) as LexicalNode;
-    if (nodes.indexOf(firstBlock) === -1) nodes.push(firstBlock);
+  let maybeBlock = selection.anchor.getNode().getParentOrThrow();
+  if (nodes.indexOf(maybeBlock) === -1) nodes.push(maybeBlock);
```
acywatson, line 58: Not sure if we should be using indexOf like this, as it relies on referential equality of the nodes in this case. It might not cause issues, but if the node is cloned, the keys would be stable while the reference may not be.

packages/lexical-selection/src/range-selection.ts
EgonBolton: This `if` block was added because some unit tests initialize the selection with the anchor in `root` and therefore those tests would break. My doubt is if it is a real scenario that the selection is in root.
If the answer is yes, I will delete the comment.
If the answer is no, I should modify those tests and remove this block.

To my knowledge, the selection can be `null`, so I don't see the point of allowing the selection of root.

packages/lexical-selection/src/range-selection.ts
EgonBolton: I'm pretty sure that can't happen, `nodes` and `maybeBlock` are referentially retrieved from the selection, one immediately after the other.

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -157,42 +157,35 @@ export class ListItemNode extends ElementNode {
...
-        this.getChildren().forEach((child: LexicalNode) => {
-          replaceWithNode.append(child);
-        });
-      }
-      this.remove();
-
-      if (childrenLength === 1) {
-        list.remove();
+    if (!$isListNode(list)) return replaceWithNode;
+    if (list.__first === this.__key) {
```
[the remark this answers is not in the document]
EgonBolton: Do you suggest I put getFirstChild() and getLastChild()? Shouldn't ElementNode expose something like getFirst() and getLast() the same way it does with getKey()?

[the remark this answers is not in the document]
EgonBolton: Also, I see it being used in several places outside of lexical core:

image
image
image
image

packages/lexical-list/src/LexicalListItemNode.ts
EgonBolton: fixed the key (I replaced __key with getKey())
About __last waiting for response to the comment above.

packages/lexical-list/src/LexicalListItemNode.ts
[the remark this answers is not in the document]
acywatson: > Also, I see it being used in several places outside of lexical core:

To the extent this is being done, it probably shouldn't be. The definition of "core" is sometimes a bit loose, but it definitely doesn't include lexical-list.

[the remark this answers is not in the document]
acywatson: > Shouldn't ElementNode expose something like getFirst() and getLast() the same way it does with getKey()?

Sure, or just do it the other way you suggested or some other way.

> which by the way I'm curious as to why the build doesn't allow it

I don't think it's necessarily a principled decision to just never allow optional chaining, the build simply isn't configured to support it. I'll see if I can recall the original rationale (assuming there was one).

packages/lexical-selection/src/range-selection.ts
acywatson: Holding aside whether or not Selection should be allowed to move to root, in principle, it can currently move to root and that's intentional.

packages/lexical-selection/src/range-selection.ts
acywatson: Yea this is probably fine

packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -115,6 +115,14 @@ export class ElementNode extends LexicalNode {
...
     return self.__size;
   }
+  getFirstChildKey(): null | NodeKey {
+    const self = this.getLatest();
+    return self.__first;
+  }
+  getLastChildKey(): null | NodeKey {
+    const self = this.getLatest();
+    return self.__last;
+  }
```
EgonBolton, line 125: I'm not sure if getLatest() should go here.
getKey() is implemented like this:
```
 getKey(): NodeKey {
    // Key is stable between copies
    return this.__key;
  }
```
but getChildrenSize for example does use it:

```
getChildrenSize(): number {
     return self.__size;
   }
```

EgonBolton:
Unit tests pass on my pc. Please try re-running them.

acywatson:
> Unit tests pass on my pc. Please try re-running them.

Will look deeper at this tomorrow - seems like a legit failure, but not sure why yet. Unit tests are not typically flaky. In fact, I can't remember ever seeing a flaky failure.

EgonBolton:
Fixed. This is done!

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -158,42 +158,34 @@ export class ListItemNode extends ElementNode {
...
-        });
-      }
-      this.remove();
-
-      if (childrenLength === 1) {
-        list.remove();
+    if (!$isListNode(list)) return replaceWithNode;
+    if (list.getFirstChildKey() === this.getKey()) {
+      list.insertBefore(replaceWithNode);
+    } else if (list.getLastChildKey() === this.getKey()) {
```
fantactuka, line 166: This is the case, where I'd either go with either of following instead of introducing getLastChildKey()/getFirstChildKey() as API to keep it smaller:
- direct `.__first`, `.__last` usage (not always welcomed, but it's a core module that is very aware of implementation details)
- full `getFirstChild().getKey()` (getting first/last child is a constant time, so overhead is neglectable) 

WDYT?

EgonBolton: [I changed `__first` at @acywatson request](https://github.com/facebook/lexical/pull/3821#discussion_r1100299594). Personally `.__first` seems good to me, but I leave it for you to decide among yourselves.
The second option would require an extra line or two of code every time it happens because non null assertion is not allowed.

acywatson: Yea, I mean, it's not good hygiene for anything to depend on these internal properties unless we're in the core somewhere (core meaning lexical package). It makes more callsites we have to change if/when we change the implementation AND it sets a bad example for others looking at the code.

It's not the end of the world, though - we can change it back to __first for now and I will separately see if we can get optional chaining working at some point.

EgonBolton: Done!

acywatson (approved):

Issue #3815, opened 2023-02-03
Bug: In the middle of page, Layout shift while using textahead menu


blacktoast:
When the page is longer than the viewport, the layout shift occurs when you open the textahead menu in the middle of the document and press the downwards arrow or upwards arrow 

Is this the same error as #3716 ?

Lexical version:

## Steps To Reproduce

1. Go to https://playground.lexical.dev/
2. Copy and paste content so that content is more than one page long
3. scroll down to the middle part of the document page
4. write  `:`
5. try pressing the downwards arrow or upwards arrow 

Link to code example:

## The current behavior
when pressing the downwards arrow or upwards arrow, occur layout shift
![Kapture 2023-02-03 at 22 35 35](https://user-images.githubusercontent.com/10705018/216618929-add8e358-1eaa-4bb7-acfd-37af1556ef90.gif)

## The expected behavior
Is this the correct function?
I don't think it's the right function.

if It is a bug and I want to fix this bug, Should I have the textahead menu come to the center of the viewport?

PR #3826, opened 2023-02-05
Fix(#3716,#3815): Prevent layout shift while using typeahead menu

### LINKED ISSUE #3716: Bug: Layout shifts while using typeahead menu
While the page is longer than just the viewport height with the typeahead menu open, while navigating through keyboard, on reaching the last value and pressing bottom button again shifts the page. I don't think this is expected? 

Lexical version:

## Steps To Reproduce

2. Copy and paste content so that content is more than one page long
3. try `:a`
4. navigate using keyboard until last option and while at last option click the down button again. 

## The current behavior
While first element comes to focus through keyboard from the last element screen shifts.

https://user-images.githubusercontent.com/32865581/212252950-0cf61fc1-2728-4a6c-a9fc-cb99f16953b6.mov

## The expected behavior
This should'nt happen

blacktoast:
resolve #3716 and  #3815 

# Problem
When the amount of a document is more than one page and when users use the typeahead menu in the middle of the document, the layout shifts

# Solution
Kapture 2023-02-05 at 22 33 28

# Note
1. I changed the `if else statement` into an early return statement

2. I removed `parentNode`, because the previous `parentNode` was very similar to the target node. 
3. I made `typeaheadContainerNode`  for detecting the typeahead menu out of the viewport

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -80,24 +80,26 @@ export type MenuRenderFn<TOption extends TypeaheadOption> = (
 
 const scrollIntoViewIfNeeded = (target: HTMLElement) => {
   const container = document.getElementById('typeahead-menu');
+  if (!container) return;
 
-  if (container) {
-    const parentNode = target.parentNode as HTMLElement | null;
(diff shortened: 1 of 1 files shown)


thegreatercurve:
Great work! Thanks @blacktoast!

thegreatercurve (approved):

Issue #3546, opened 2022-12-13
Bug: [Playground] autolink escapes on second `.` when typing


zeitstein:
## Steps To Reproduce

Type in e.g. `https://www.foo.com` in Playground. Works as expected on paste.

## The current behavior

![image](https://user-images.githubusercontent.com/8146007/207295726-ef2b50a6-5d84-4385-805f-6aed8b1a2477.png)

prestonbrown-me:
some urls with subdomains do actually work. for example, typing `https://playground.lexical.dev` does work. I believe this has to do with the autolink plugin itself, rather than the URL matcher used.

NikolajLeischner:
I ran into the same issue and could track it to this change: https://github.com/facebook/lexical/commit/f29ab08dbf9c0b14b89ebef44fa243e712d817ad#diff-d62fd402ee470b9082046b4a435f031dd8539650efc92ac2ea645f8ded2a4c57R55

If you remove the "." from the `PUNCTUATION_OR_SPACE` regex, then the issue disappears.

thegreatercurve:
I'm not sure there's much we can do here. This is technically behaving correctly. 

The regex is accounting for valid URLs which omit their subdomains (i.e. `https://github.com`), but it doesn't make any distinction when such a URL uses "www" as it's actual domain. 

This is technically valid behaviour. Sites like `https://www.org` are a valid, and we want them to be recognised by the regex.

NikolajLeischner:
Closing this doesn't seem very helpful. As is, the autolink-plugin is useless. It doesn't recognize most URLs with sub-domains correctly, and no URLs if you add a protocol. And this is independent of the regex you provide the plugin with.

acywatson:
> Closing this doesn't seem very helpful. As is, the autolink-plugin is useless. It doesn't recognize most URLs with sub-domains correctly, and no URLs if you add a protocol.
Note that this behavior is independent of the matchers you provide the plugin with.

Agreed, this was closed prematurely, I think. We need to see what we can do to fix this.

Karibash:
> I ran into the same issue and could track it to this change: [f29ab08#diff-d62fd402ee470b9082046b4a435f031dd8539650efc92ac2ea645f8ded2a4c57R55](https://github.com/facebook/lexical/commit/f29ab08dbf9c0b14b89ebef44fa243e712d817ad#diff-d62fd402ee470b9082046b4a435f031dd8539650efc92ac2ea645f8ded2a4c57R55)
[...]

PR #3829, opened 2023-02-06
Fix a bug that caused the Node to be regenerated when editing the end of an AutoLinkNode

### LINKED ISSUE #3825: Bug: Each time a character is added to the tail, the AutoLinkNode is regenerated
Lexical version: 0.7.6

## Steps To Reproduce

1. Enter the URL.
2. Add text to the end of the URL.

## The current behavior

Each time a trailing character is added, the AutoLinkNode is regenerated.
If editing in the middle of a URL, the Node is not regenerated.

https://user-images.githubusercontent.com/40270352/216815279-af159889-1280-482d-871d-fd24d0b54e82.mov

## The expected behavior

Node is not regenerated even when editing the end of the URL.

Karibash:
close #3825 

https://user-images.githubusercontent.com/40270352/216878308-06bc05e5-422c-4b3e-aed9-8602ce005ee9.mov

### DIFF
--- a/packages/lexical-react/src/LexicalAutoLinkPlugin.ts
+++ b/packages/lexical-react/src/LexicalAutoLinkPlugin.ts
@@ -220,18 +220,24 @@ function handleLinkEdit(
 
 // Bad neighbours are edits in neighbor nodes that make AutoLinks incompatible.
 // Given the creation preconditions, these can only be simple text nodes.
-function handleBadNeighbors(textNode: TextNode, onChange: ChangeHandler): void {
+function handleBadNeighbors(
+  textNode: TextNode,
+  matchers: Array<LinkMatcher>,
+  onChange: ChangeHandler,
+): void {
   const previousSibling = textNode.getPreviousSibling();
   const nextSibling = textNode.getNextSibling();
   const text = textNode.getTextContent();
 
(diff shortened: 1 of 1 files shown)


trueadm:
Looks like the AutoLink e2e tests are failing?

acywatson:
Hmmm, on the vercel preview here: https://lexical-playground-git-fork-karibash-featur-ef48ce-fbopensource.vercel.app/

it seems like this is not working correctly. if I type in https://www.github.com, the ".com" is not part of the autolink.

Karibash:
@acywatson 
Hmm, that seems to be an existing bug.
I can reproduce it in the [current playground](https://playground.lexical.dev/).
It seems to be a bug caused by the number of characters in the domain.
The same phenomenon can be reproduced when there is a second level domain like co.jp.

https://user-images.githubusercontent.com/40270352/219257868-cb8c1d68-9595-4cfd-9f6f-5af8fe5215c6.mov

acywatson:
> 

Ah ok - weird, the video in the PR description seems to show it working.

Karibash:
I found an issue reporting the above problem, but it seems to be closed.
https://github.com/facebook/lexical/issues/3546

thegreatercurve:
Related issue aside this PR looks good to merge.

thegreatercurve (approved):

PR #3836, opened 2023-02-06
Make setBlocksType_experimental more flexible

abelsj60:
This pull request allows `setBlocksType_experimental` to process shadowRoots in two ways: 

- New options have been added to process all the children of a shadowRoot node when the selection touches any one of them. Processing can be done automatically via `collapseAtStart` or via a custom handler. 
 
    - Example: Take a LinedCodeNode or a List. This feature allows the entire node to be converted to basic paragraphs and text at once. As this is optional behavior, the developer can employ it as desired (it's almost always desired with a LinedCodeNode — otherwise things get weird...)

- Another new option lets developers append the newly created block nodes to a parent shadowRoot node. This is done by passing a custom function to `setBlocksType_experimental`. 

Note, similar functionality existed in $wrapNodes, but was lost when this function was added because its primary goal was a huge level of simplification. While successful at this, the ability to effectively handle shadowRoot (nested structures) was lost.

Here's an abridged usage example: 

_SET UP_
```
// EXISTING FUNCTION

const createElement = (blockNode: LexicalNode) => {
  const line = $createLinedCodeLineNode();
  line.append($createTextNode(blockNode.getTextContent()));
  return line;
};

// NEW OPTIONAL FUNCTION TO CREATE SHADOW ROOT NODES

const createParentBlock = (childBlocks: LexicalNode[]) => {
  const codeNode = $createLinedCodeNode();
  codeNode.append(...childBlocks as LinedCodeLineNode[]);

  const firstLine = codeNode.getFirstChild(); 
  ...
  return codeNode;
};

```
_USAGE_
```
// CURRENT FUNCTION WITH NEWLY ADDED OPTIONS OBJECT

$setBlocksType_experimental(selection, createElement, {
  createParentBlock,
  collapseShadowRootNodes: true,
});
```

All unit tests pass.


acywatson:
This is actually @EgonBolton's brain-child, so will let him weigh in on the changes here.

abelsj60:
> This is actually @EgonBolton's brain-child, so will let him weigh in on the changes here.

OK. I'm making some changes. Will push again, probably today. I hope. 

I may have to look at $wrapNodes, too, but I don't really want to. This one was kinda enough. 

If you can tell me, it would be helpful to know if you actually plan on deprecating $wrapNodes. As setBlocks... is now in the Playground, so I assumed you would. But, suddenly I'm not so sure. 

The bottom line is that $wrapNodes seems to be the only one of these two functions that can process shadowRoot, at least for creation, but...probably destruction, too, right now. My alternative is to stick with my own implementation and not use a Lexical library function for this. I don't love that idea either, but these are the only options I can see at the moment.

-j

EgonBolton:
I've spent a good deal of effort understanding what you're trying to do, but I'm not sure I've succeeded.

The only thing $wrapNodes could do that $setBlocksType doesn't, was allow an additional node to wrap around the replaced one. This can be done in a single line of code. Simply apply [$wrapNodeInElement](https://github.com/facebook/lexical/blob/1dc6efa69cc2a3f7cac312ff2c6a04dbbd2642e3/packages/lexical-utils/src/index.ts#L383) to the node you want to wrap. Another option would be to send the wrapper with its content in the createElement parameter of $setBlocksType, and do the replace on the child node.

> If you can tell me, it would be helpful to know if you actually plan on deprecating $wrapNodes? setBlocks is now in the Playground, so I assumed you would. But, suddenly I'm not so sure.

$setBlocksType is no longer just a matter of simplicity. $wrapNodes is not safe to use, as it has a lot of bugs (Some of which I haven't reported). I'm waiting for the latest PR on $setBlocksType which fixes some things to be merged, to deprecate $wrapNodes.

abelsj60:
> I've spent a good deal of effort understanding what you're trying to do, but I'm not sure I've succeeded.
> 
> The only thing $wrapNodes could do that $setBlocksType doesn't, was allow an additional node to wrap around the replaced one. This can be done in a single line of code. Simply apply [$wrapNodeInElement](https://github.com/facebook/lexical/blob/1dc6efa69cc2a3f7cac312ff2c6a04dbbd2642e3/packages/lexical-utils/src/index.ts#L383) to the node you want to wrap. Another option would be to send the wrapper with its content in the createElement parameter of $setBlocksType, and do the replace on the child node.
> 
> > If you can tell me, it would be helpful to know if you actually plan on deprecating $wrapNodes? setBlocks is now in the Playground, so I assumed you would. But, suddenly I'm not so sure.
> 
> $setBlocksType is no longer just a matter of simplicity. $wrapNodes is not safe to use, as it has a lot of bugs (Some of which I haven't reported). I'm waiting for the latest PR on $setBlocksType which fixes some things to be merged, to deprecate $wrapNodes.

Lol. "Trying." I like it. Story of my life.

Anyway, this code lets me transform shadowRoot nodes in two ways — collapse and create.  

Truthfully, I've created a nested structure — a `LinedCodeNode` — that isn't in the Playground right now. As a result, I probably have an unusual vantage point. That's the curse of an un-opinionated framework, I suppose. 

I'll try to briefly clarify before outlining my game plan: 

**New forms of shadowRoot** 

- _A. Creating shadowRoot nodes, more specifically my `LinedCodeNode`:_ 

  - My node is a contiguous box of many "blocks". 
  - I need to take all the things made by your `createElement` function and append them to the parent code block. 
  - I don't believe I could pass a `parentBlock` creator inside your `createElement` as that would not make a single parent for all the children. I believe it would make many separate sibling `LinedCodeNodes`.
    - In other words, I am using a single shadowRoot as a container for many children, AKA "blocks."  
    - I need to handle selection conversions in a way that works with this.

- _B. Removing shadowRoot nodes, more specifically my `LinedCodeNode`:_ 

  - The problem here is sneaky. 
  - Imagine this scenario. My `LinedCodeNode` has five lines and: 
    - I have a "collapsed" `selection` touching only one of those five lines, or 
    - I have an "uncollapsed" `selection` spanning two of those five lines. 
  - In this scenario, I didn't think it made sense to only transform one of those lines to something else when the user clicks a button. I felt it made more sense to collapse the whole thing — the whole `LinedCodeNode` — at once.
    - That's a problem for your function, however. 
    - The reason is that your function only processes the nodes that are included in the `selection`. This might not include the `LinedCodeNode` and all of its children (lines one, three, four, and five).
    - Thus, I need to expand the selected node range to cover all the lines in the `LinedCodeNode`. 
    - I have two options: 
        - Expand the selected node inside your function — which is what I did here. It now runs `collapseAtStart` on my node, updates the selected nodes with the paragraph nodes that `collapseAtStart` created and appended to root, and then continues on its merry way. It actually works very nicely, so far.
      - Run `collapseAtStart` and update the `selection` BEFORE calling your function. Truthfully, this last didn't occur to me. I knew I needed to update your code and I was probably so focused on that, that I didn't think of it. . 
    - I'll have to give the latter a shot. Maybe I could remove the `handleShadowRoot` parts, then. Oh happy day. 
  - (This pull request also covers several other weird edge cases, such as selecting backward from plain paragraphs into a `LinedCodeNode`, or between `LinedCodeNodes`. I'll have to re-evaluate those again, too.)

**In summary**: I believe _**A**_ remains a problem for me until, at least, a wrapping function is introduced that works with all the things that `createElement` produces at once. _**B**_ may be fixed without adding any code, but I'll have to try it out.

**My game plan**

I'll push some changes over the next day or so. 

@acywatson I'll ping you again when I do and you can sort out what you want to do. I'll need to do something. 

If I have to break off and go with my own function for converting a `selection`, I will. I'll just make it available with the rest of the `LinedCodeNode`'s code when I do.

abelsj60:
Oh, happy day. 

@EgonBolton, you did me a favor. I went back to the drawing board and managed to work around this function. 

I ended up with a better API for my node, as a result. I use this function a bit, and can easily swap it out.

@acywatson For what it's worth, I suspect the `$wrapNodes` idea that's explored here will need help in the future. 

To me, the current model views node transforms as an "implementation detail" of the LexicalEditor. 

I wonder if that'll work. Lexical's spirit is much: "anything goes." That's really its great strength.

As a result, I keep thinking that node transformation should be treated more like `importDOM` and `importJSON`. This might mean adding a normalization method to, at least, `ElementNodes`. Devs might use it to transform nested node sets into a format that can be easily processed into another node set without worrying about the previous one's quirks. 

Put another way, I'm thinking of a top-level node method — no matter how many levels "up" — that allows the whole container to be transformed into a form that Lexical — and the developers who use it — can manipulate as needed. 

This is how I was using `collapseAtStart` in this pull request. 

I'm either a huge fool — which I always think a fair bet — or I'm a little ahead of the curve right now. 

My `LinedCodeNode` lives in three dimensions — block-as-grandparent (code-node), blocks-as-parents (line divs), and a sequence of individual text nodes (text spans) in each line. As a result, I may not see shadowRoot in the same way now. 

In practice, I suspect one size won't always be able to fit all when transforming node sets from type-to-type. 

I'm not the only one going in this direction. 

I see people on Discord encapsulating paragraphs and text with some frequency. The only difference may be that they often have a single `TextNode` in their lowest-level block — I do not. To be fair, Lexical fought me every step of the way. 

For instance, `.replace` contains a failure point when swapping the original, single `TextNode` with a brand new sequence of text-node-code-tokens. It loses the `selection`, which must then be restored manually before the update concludes.

Still, I wouldn't be surprised to see more people want to encapsulate multiple text nodes within parents, AKA blocks. 

It's intuitive. Moreover, I also get the sense that this model may work nicely with the state of the art — generative AI. Using different types of text nodes to help track some words and sentences inside a text editor seems like it could help people control context when talking to ChatGPT and its ilk. It's almost as if everything's going...meta. 

-j

PS: I'll mark this PR closed a little later.

abelsj60:
To the future: 

The `LinedCodeNode`'s code — and docs on how to transform it to other node types on the fly (which uses `setBlocksType_experimental` under the hood) — can be found here, at [Lexical 401](https://github.com/abelsj60/lexical-401).

PR #3843, opened 2023-02-07
Add more configuration for heading in INSERT_TABLE_COMMAND

Sebastien-Ahkrin:
The idea behind this changement is to allow the user of the API to choice if the header will be only row, column or Both.

For example: 
```tsx
function onClick() {
  activeEditor.dispatchCommand(INSERT_TABLE_COMMAND, {
    columns,
    includeHeaders: true,
    rows,
  });
}
```

This will make the header as column and row like before.
image

but, if you add columns and row as an object for includeHeaders like that:
```tsx
function onClick() {
  activeEditor.dispatchCommand(INSERT_TABLE_COMMAND, {
    columns,
    includeHeaders: {columns: false, rows: true},
    rows,
  });
}
```
CleanShot 2023-02-07 at 17 10 42

Or like that if you make columns as true and rows as false 
image

Do this functionnality looks good for you ?

### DIFF
--- a/packages/lexical-playground/src/plugins/TablePlugin.tsx
+++ b/packages/lexical-playground/src/plugins/TablePlugin.tsx
@@ -102,7 +102,11 @@ export function InsertTableDialog({
   const [columns, setColumns] = useState('5');
 
   const onClick = () => {
-    activeEditor.dispatchCommand(INSERT_TABLE_COMMAND, {columns, rows});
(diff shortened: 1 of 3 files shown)


thegreatercurve (approved):
The semantics might be slightly odd as you'd always expect `includeHeaders` to be a boolean, but it's probably not worth introducing a breaking change over.

PR #3855, opened 2023-02-08
Revert "Fix $restoreEditorState"

thegreatercurve:
This is causing MaxLength E2Es to fail

Reverts facebook/lexical#3842


fantactuka (approved):

zurfyx:
> * Might be worth adding details on why it has to be reverted

+1. Don't have much context on this but your fix is changing the implementation from a deep clone to a shallow clone

PR #3865, opened 2023-02-09
[0.8] Block keyDown events in decorators

trueadm:
Expanding on the discussion in https://github.com/facebook/lexical/pull/3766.


thegreatercurve (approved):

trueadm:
After implementing this, I noticed that a ton of logic now fails around keyboard selection. So this is definitely not the path we want to go down. Closing.

PR #3878, opened 2023-02-10
Add KEY_DOWN_COMMAND

trueadm:
This enables people to listen to all keydown commands and block any further commands if they return true. This can be useful for controlling very specific use-cases rather than having to fight with a custom `keydown` event listener.


thegreatercurve (approved):
Thanks for adding in the docs

Issue #3880, opened 2023-02-10
typing "i" causes the composer entry box to freeze and no other characters are allowed to be typed.


zurfyx:
https://user-images.githubusercontent.com/193447/218162325-923bf6f7-f9cb-48ac-8f61-acc9df25dabb.mov

PR #3881, opened 2023-02-10
Fix autocapitalization

zurfyx:
This PR fixes autocapitalization on Mac #3880.

It turns out that MacOS with Safari can trigger a double `beforeInput` before `input`, which causes our the `insertText` `beforeInput` to get lost and never processed (we wait until onInput to detect the content that has been inputted via beforeInput - this is intentional to boost the performance of the most common operation (we let the browser do its own thing)).

We can fix this by processing the previous beforeInput when this rarely event happens.

Copying from the code comments

```
    // This represents the case when two beforeinput events are triggered at the same time (without a
    // full event loop ending at input). This happens with MacOS with the default keyboard settings,
    // a combination of autocorrection + autocapitalization.
    // Having Lexical run everything in controlled mode would fix the issue without additional code
    // but this would kill the massive performance win from the most common typing event.
    // Alternative, when this happens we can prematurely update our EditorState based on the DOM
    // content, a job that would usually be the input event's responsibility.
```

https://user-images.githubusercontent.com/193447/218163360-403d50cb-23b2-424d-a3a1-52a8f46341fb.mov

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Events.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Events.spec.mjs
@@ -1,0 +1,100 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {
+  assertHTML,
+  evaluate,
+  focusEditor,
+  html,
+  initialize,
+  test,
+} from '../utils/index.mjs';
+
+test.describe('Events', () => {
+  test.beforeEach(({isCollab, page}) =>
+    initialize({isAutocomplete: true, isCollab, page}),
+  );
+  test('Autocapitalization (MacOS specific)', async ({page, isPlainText}) => {
+    await focusEditor(page);
+    await page.keyboard.type('i');
+    await evaluate(page, () => {
+      const editable = document.querySelector('[contenteditable="true"]');
+      const span = editable.querySelector('span');
+      const textNode = span.firstChild;
+      function singleRangeFn(
+        startContainer,
+        startOffset,
+        endContainer,
+        endOffset,
+      ) {
+        return () => [
+          new StaticRange({
+            endContainer,
+            endOffset,
+            startContainer,
+            startOffset,
+          }),
+        ];
+      }
+      const character = 'S'; // S for space because the space itself gets trimmed in the assertHTML
+      const dataTransfer = new DataTransfer();
+      dataTransfer.setData('text/plain', 'I');
+      dataTransfer.setData('text/html', 'I');
+      const characterBeforeInputEvent = new InputEvent('beforeinput', {
+        bubbles: true,
+        cancelable: true,
+        data: character,
+        inputType: 'insertText',
+      });
+      characterBeforeInputEvent.getTargetRanges = singleRangeFn(
+        textNode,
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/__tests__/e2e/Events.spec.mjs
```diff
@@ -0,0 +1,100 @@
...
+        data: character,
+        inputType: 'insertText',
+      });
+      editable.dispatchEvent(characterBeforeInputEvent);
+      textNode.textContent += character;
+      // Selection should be set with the event; this won't quite work, but not sure how to fix it..
+      // document.getSelection().setBaseAndExtent(textNode, 2, textNode, 2);
+      editable.dispatchEvent(replacementBeforeInputEvent);
+      editable.dispatchEvent(characterInputEvent);
+      // Lexical will prevent default for the replacement input event via the clipboard logic; we
```
acywatson, line 84: Do we need this commented out code?

packages/lexical/src/LexicalEvents.ts
```diff
@@ -492,14 +493,27 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
 
     const data = event.data;
 
+    // This represents the case when two beforeinput events are triggered at the same time (without a
+    // full event loop ending at input). This happens with MacOS with the default keyboard settings,
+    // a combination of autocorrection + autocapitalization.
+    // Having Lexical run everything in controlled mode would fix the issue without additional code
+    // but this would kill the massive performance win from the most common typing event.
+    // Alternative, when this happens we can prematurely update our EditorState based on the DOM
```
acywatson, line 501: nit: Alternative -> Alternatively

packages/lexical/src/LexicalEvents.ts
```diff
@@ -528,6 +542,8 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
       ) {
         event.preventDefault();
         dispatchCommand(editor, CONTROLLED_TEXT_INSERTION_COMMAND, data);
+      } else {
```
acywatson, line 545: Should we have a condition here around the inputType, like above? Not sure what the value of it is in this specifics auto-cap scenario.

zurfyx: Not sure I understand your comment. The `else` case handles the specific non-controlled case. Autocaps is a combination of a keypress `beforeInput` event + a system `beforeInput` event.

acywatson: Sure man - I was just thinking generally about the broadness of an "else" case. Given that we have a series of specific conditions above it, I wanted to make sure that we wouldn't unintentionally be catching some cases here that we didn't intend to. I trust your judgment

acywatson: Sure man - I was just thinking generally about the broadness of an "else" case. Given that we have a series of specific conditions above it, I wanted to make sure that we wouldn't unintentionally be catching some cases here that we didn't intend to. I trust your judgment

acywatson (approved):

packages/lexical/src/LexicalEvents.ts
zurfyx: That's a good point, my understanding is that the `else` statement only catches the non-controlled mode here as  otherwise it would have been early returned, and there is no preexisting else condition because we do no logic when this happens

PR #3886, opened 2023-02-11
Improve markdown transformers

Godefroy:
I need to import a markdown document with tables containing markdown in each cell.

Problem: `$convertFromMarkdownString` is currently not called inside of table cells and doesn't take a target node as a parameter, it uses root node only.

The `TABLE` transformer needs to be adapted, as well as `createMarkdownImport` (which is used by `$convertFromMarkdownString`) and `createMarkdownExport`.

This pull request adds:
- Markdown import inside table cells
- Markdown export inside table cells
- Table row header import
- Emoji import

You can test with this markdown example:
```md
| Table | Col 1 | Col 2 |
|-------|-------|-------|
| Line 1 | *Content* | Multi\nlines |
| Line 2 | Emoji :smile: | `Code` |
```


packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts
```diff
@@ -50,6 +43,7 @@ import {
 } from '../../nodes/EquationNode';
 import {$createImageNode, $isImageNode, ImageNode} from '../../nodes/ImageNode';
 import {$createTweetNode, $isTweetNode, TweetNode} from '../../nodes/TweetNode';
+import emojiList from '../EmojiPickerPlugin/emoji-list';
```
acywatson, line 46: Plugins should be dependent directly on other plugins like this. You'd need to pull the common logic out to a shared dir somewhere.

packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts
```diff
@@ -239,37 +272,24 @@ function getTableColumnsSize(table: TableNode) {
...
-  // TODO:
-  // For now plain text, single node. Can be expanded to more complex content
-  // including formatted text
   const match = textContent.match(TABLE_ROW_REG_EXP);
-
-  if (!match || !match[1]) {
-    return null;
-  }
-
+  if (!match || !match[1]) return null;
```
acywatson, line 284: we typically don't do one line conditional like this (IIRC). Nothing against it personally, just a repo style consistency thing.

packages/lexical-markdown/src/index.ts
```diff
@@ -92,6 +92,8 @@ export {
   BOLD_UNDERSCORE,
   CHECK_LIST,
   CODE,
+  createMarkdownExport,
```
acywatson, line 95: @fantactuka is this what we want to do?

acywatson (changes requested):
CI integrity is failing for some reason - looks like prettier

packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts
Godefroy: I understand, but there are already 6 other plugins depending on other plugins like this, and plugins are tightly coupled with other folders (nodes, ui). I'm not sure how to make this change properly.

Do you have a preference between:

1. Moving `emoji-list.ts` in `utils` or another folder
2. Removing this commit to remove emojis transformer
3. Refactor playground transformers to set them from root component Editor to improve modularity.

packages/lexical-markdown/src/index.ts
Godefroy: The alternative is to adapt also `$convertToMarkdownString` and use it instead.

Should I do that?

packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts
acywatson: @Godefroy You can just do the first one

Godefroy:
Thanks @acywatson for the code review! I adapted my code.

Also, I took the liberty to remove `root.selectEnd();` from `createMarkdownImport` because I think it's not its job and it was causing the equivalent of an autoFocus when importing a markdown.

packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts
```diff
@@ -136,39 +150,75 @@ export const TWEET: ElementTransformer = {
...
+        output.push(`| ${rowOutput.map((_) => '---').join(' | ')} |`);
+      }
     }
 
     return output.join('\n');
   },
   regExp: TABLE_ROW_REG_EXP,
   replace: (parentNode, _1, match) => {
+    // Header row
+    if (TABLE_ROW_DIVIDER_REG_EXP.test(match[0])) {
```
acywatson, line 197: @zurfyx will this be a problem for the tables works you're currently doing? I doubt it, but wanted to check.

packages/lexical-playground/src/plugins/EmojiPickerPlugin/index.tsx
```diff
@@ -95,7 +95,7 @@ export default function EmojiPickerPlugin() {
 
   useEffect(() => {
     // @ts-ignore
-    import('./emoji-list.ts').then((file) => setEmojis(file.default));
+    import('../../utils/emoji-list.ts').then((file) => setEmojis(file.default));
```
acywatson, line 98: thanks for this

packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -41,16 +42,16 @@ type TextFormatTransformersIndex = Readonly<{
 
 export function createMarkdownImport(
   transformers: Array<Transformer>,
-): (markdownString: string) => void {
+): (markdownString: string, node?: ElementNode) => void {
```
acywatson, line 45: I'm ok with this - maybe a nit to call it targetNode or something that indicates that it's the parent of whatever the import creates.

Separately, why shouldn't markdown work more like HTML, where you just get back some nodes and then insert them wherever you want?

@fantactuka 

I'm ok with this change.

acywatson (approved):

PR #3894, opened 2023-02-13
Playground : Adding a wrapper div to support alignment (left, center, right) for Youtube Video. Adding exportDom function to the FigmaNode.

GuyoST:
PLayground : I added exportDom to FigmaNode, and modified little bit the existing exportDom in Youtube Node to support alignment (left, center, right), because we could align it in the editor, but it was not reflected in the HTML output.

### DIFF
--- a/packages/lexical-playground/src/nodes/FigmaNode.tsx
+++ b/packages/lexical-playground/src/nodes/FigmaNode.tsx
@@ -7,6 +7,7 @@
  */
 
 import type {
+  DOMExportOutput,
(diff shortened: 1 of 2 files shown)


thegreatercurve:
Can you add this to the `DecoratorBlockNode` instead which is the node which wraps these nodes? Something like this on that node should work: 
```
     element.style.textAlign = this.__format;
     return {element};
 }
```

thegreatercurve:
Closing this for the moment due to inactivity.

PR #3898, opened 2023-02-13
Add overridable method to cancel markdown

abelsj60:
@trueadm — I'm following your pattern for this PR, so thought I'd tag you.

This Pull Request allows the `TextNode` — and its overrides and extensions — to cancel Markdown shortcuts at will. 

I need it so my `LinedCodeNode` can block accidental Markdown usage within each line. 

Unfortunately, the Markdown plugin currently assumes that code will either have the 'code' format or a `CodeNode` parent. 

- In the first case, the current `code` package doesn't tag code with the format of 'code' (surprising, I know.)
    - I experimented with adding the 'code' format to my `LinedCodeTextNodes`, but the styling gets weird (monospacing).
    - I'm sure no one — including me — wants to _have_ to add the CSS necessary to combat that.
- In the second case, the `LinedCodeNode` puts `LinedCodeTextNodes` in divs, AKA, lines. _Their_ parent is a `CodeNode`, meaning the current test fails.

Rather than try to add something specific, I've added `canContainMarkdown` to the `TextNode`. It's `true` by default. My `LinedCodeTextNode` overrides it and sets it to `false`. As a result, Markdown won't appear in lines — a shocking sight!

The pattern is the same as my recent `canContainTabs` pull request. It's nice because it's generic and succinct.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/MarkdownShortcuts.ts
@@ -419,7 +419,11 @@ export function registerMarkdownShortcuts(
 
         const parentNode = anchorNode.getParent();
 
-        if (parentNode === null || $isCodeNode(parentNode)) {
+        if (
+          parentNode === null ||
+          !anchorNode.canContainMarkdown() ||
+          $isCodeNode(parentNode)
+        ) {
           return;
         }
 
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -707,6 +707,10 @@ export class TextNode extends LexicalNode {
     return false;
   }
 
+  canContainMarkdown(): boolean {
+    return true;
+  }
+
   splitText(...splitOffsets: Array<number>): Array<TextNode> {
     errorOnReadOnly();
     const self = this.getLatest();


acywatson:
This seems like a problem we could solve through the configuration of the Markdown plugin, either by using the existing transformer API or expanding it to accommodate this use case. I don't think we want to introduce the concept of Markdown to TextNodes.

abelsj60:
> This seems like a problem we could solve through the configuration of the Markdown plugin, either by using the existing transformer API or expanding it to accommodate this use case. I don't think we want to introduce the concept of Markdown to TextNodes.

Sure. I don't really care about what method's used for this.

What do you suggest I do _now_ to ensure that text nodes in a nested node like mine can't trigger Markdown shortcuts?  At the moment, if I type a hash in a line of code, for instance, I create an H1 in that line. This interferes with it being a line of code and also looks strange. 

I looked at the Markdown package and don't see any other external ways of shutting it off. 

What can I PR today to clean this up?

--ASIDE--

The plugin currently looks at anchorNodes. If the anchorNode is not a textNode, it is rejected, so it is working with text nodes, specifically. If the anchorNode is a textNode, processing continues. If the format is 'code,' it is rejected, although nothing is tagged this way right now. If not, processing continues. If the anchor/textNode is a child of code, it is rejected, if not, processing continues. This is where I get stuck.)

thegreatercurve:
@acywatson A general `canBeTransformed` property on the LexicalNode might be a good idea, much in the same way that we added `createParentElementNode` and `isParentRequired` in  #3800. This would unblock this PR for @abelsj60 and would allow us to remove the existing `$isCodeNode` import dependency, reducing the overall size of the markdown package.

acywatson:
> @acywatson A general `canBeTransformed` property on the LexicalNode might be a good idea, much in the same way that we added `createParentElementNode` and `isParentRequired` in #3800. This would unblock this PR for @abelsj60 and would allow us to remove the existing `$isCodeNode` import dependency, reducing the overall size of the markdown package.

@thegreatercurve @abelsj60 

canBeTransformed is slightly more palatable, but the issue is that we're polluting the core with a bunch of methods that don't provide much clarity around how they work in different situations. For instance, here how should I know that canBeTransformed just applies to Markdown conversions and not Node Transforms (or anything else that might be perceived as a "transform"). If it's just the former, then why are we leaking markdown-specific logic into the core? If it's the latter, why are we coupling markdown concerns with other transform-ish things?

We might be able to get around this with documentation but I think the real solution it making the Markdown plugin behavior more configurable and externalizing the default configuration that handles code nodes, for instance. This both breaks the dependency AND solves @abelsj60 problem by allowing him to customize the behavior however he wants without touching the core.

abelsj60:
> > @acywatson A general `canBeTransformed` property on the LexicalNode might be a good idea, much in the same way that we added `createParentElementNode` and `isParentRequired` in #3800. This would unblock this PR for @abelsj60 and would allow us to remove the existing `$isCodeNode` import dependency, reducing the overall size of the markdown package.
> 
> @thegreatercurve @abelsj60
> 
> canBeTransformed is slightly more palatable, but the issue is that we're polluting the core with a bunch of methods that don't provide much clarity around how they work in different situations. For instance, here how should I know that canBeTransformed just applies to Markdown conversions and not Node Transforms (or anything else that might be perceived as a "transform"). If it's just the former, then why are we leaking markdown-specific logic into the core? If it's the latter, why are we coupling markdown concerns with other transform-ish things?
> 
> We might be able to get around this with documentation but I think the real solution it making the Markdown plugin behavior more configurable and externalizing the default configuration that handles code nodes, for instance. This both breaks the dependency AND solves @abelsj60 problem by allowing him to customize the behavior however he wants without touching the core.

These are all reasonable thoughts, @acywatson. 

I can't say I didn't suspect this was your reasoning when I saw your first comment.

Right now, I understand a lot of Lexical better than I might in the near future. I may set it aside awhile, I'm not sure. 

Do you have some psuedocode of what you're looking for on this — I'd be happy to try something if you do. Modifying the plugin was my original thought, but @trueadm uses this method pattern that I like as well, and you guys are also careful about what you allow done to core plugins. It's both rational and responsible, but there seems a fair concern here, too. 

If you point me in a direction while I still understand your library, I can try to return something that resolves the problem in a satisfactory way.

acywatson:
We've discussed this and we aren't going to use this approach right now. We'll investigate a different approach to markdown transformer configurability.

PR #3913, opened 2023-02-15
Remove instanceof type checks against HTMLElement and subtypes

### LINKED ISSUE #3907: Bug: $generateNodesFromDOM fails in headless mode
Lexical version: 0.8.0

## Steps To Reproduce

1. In a Node.js environment, try to parse some HTML that contains links and lists using a headless editor and `$generateNodesFromDOM`, as per the documentation example.
2. :boom: `HTMLElement is not defined` or `HTMLAnchorElement is not defined`

The reason is that in a server environment, those browser-exclusive globals are not present. Polyfilling is hard and clunky, since it's usually just not necessary with libs like `JSDOM` or `happy-dom`, which create sandboxed environments rather than polluting the global space.

Offending lines that I found:

- Lists package: https://github.com/facebook/lexical/blob/3228323a2e32245386dacc0327d844de6390e5a6/packages/lexical-list/src/LexicalListNode.ts#L290
- Links package: https://github.com/facebook/lexical/blob/3228323a2e32245386dacc0327d844de6390e5a6/packages/lexical-link/src/index.ts#L237

Other that I found by grepping the repo but I haven't tested:

- Code package: https://github.com/facebook/lexical/blob/3228323a2e32245386dacc0327d844de6390e5a6/packages/lexical-code/src/CodeNode.ts#L71
- Checklist: https://github.com/facebook/lexical/blob/3228323a2e32245386dacc0327d844de6390e5a6/packages/lexical-react/src/LexicalCheckListPlugin.tsx#L162

## The current behavior

Crashes.

## The expected behavior

It works because that code doesn't depend on browser-only globals (or there is a documented, supported, and feasible solution like executing the conversion inside `JSDOM`'s sandbox or something like that).

acywatson:
This should get $generateNodesFromDOM working on the server.

Fixes #3907

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -389,3 +389,12 @@ export function $wrapNodeInElement(
   elementNode.append(node);
   return elementNode;
 }
+
+export function isHTMLAnchorElement(x: Node): x is HTMLAnchorElement {
+  return isHTMLElement(x) && x.tagName === 'A';
+}
+
+export function isHTMLElement(x: Node | EventTarget): x is HTMLElement {
+  // @ts-ignore-next-line - strict check on nodeType here should filter out non-Element EventTarget implementors
+  return x.nodeType === 1;
+}
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -33,7 +33,7 @@ import 'prismjs/components/prism-rust';
 import 'prismjs/components/prism-swift';
 import 'prismjs/components/prism-typescript';
 
-import {addClassNamesToElement} from '@lexical/utils';
+import {addClassNamesToElement, isHTMLElement} from '@lexical/utils';
 import {
   $applyNodeReplacement,
   $createLineBreakNode,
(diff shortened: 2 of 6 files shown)


DaniGuardiola:
Sweet <3

packages/lexical-utils/src/index.ts
```diff
@@ -389,3 +389,12 @@ export function $wrapNodeInElement(
   elementNode.append(node);
   return elementNode;
 }
+
+export function isHTMLAnchorElement(x: Node): x is HTMLAnchorElement {
+  return isHTMLElement(x) && x.tagName === 'A';
+}
```
fantactuka, line 395: nit: does not seem to be used much and can be inlined vs extracted as a special util

fantactuka (approved):

packages/lexical-utils/src/index.ts
acywatson: Potentially - my preference was to establish/reinforce a pattern of co-locating type-guards here. Also, as we've seen, the strategies here can change, which makes it nice to have it encapsulated in a separate place. The other issue is that you need the type predicate to refine the types, so it has to be a separate function whether it lives here or in LinkNode. I would personally say such a thing does not really relate directly to LexicalLink, semantically.

On the other hand, I think similar logic is inlined in ListNode/ListItemNode.

thegreatercurve (approved):

Issue #3763, opened 2023-01-23
Generalize setters/getters, clone, importJSON and exportJSON


GermanJablo:
# The Problems

In https://github.com/facebook/lexical/issues/1262#issuecomment-1186170960 I explained a series of problems when it comes to extending nodes.

At that time I thought that the best way to solve it was to group different nodes in the same class (as is done now with the different levels of Heading). In fact later [in the same thread, I offered a proof of concept](https://github.com/facebook/lexical/issues/1262#issuecomment-1292436961). I still think this could potentially reduce code repetition in very similar nodes like `headings` and `paragraphs`, but after thinking about it a lot I came to the conclusion that it is not a viable solution, at least for all the problems that I described.

After that the node replacement API came out. Although it solves some problems, I think the first 3 items of my initial post are still valid:

> 1. When converting one elementNode to another, the shared properties are lost (for example, convert an indented paragraph to a heading and the indentation will be lost). The reason for this is that since they are different classes, a new instance is required. Of course, by changing the type of a node one could also copy the desired properties to the new node. EDIT: It seems this has been fixed only for the indent case, and not for arbitrary user-defined properties.
> 2. Even if some properties are used only by some nodes, it would be ideal if the other nodes keep that information. The reason for this is that users often undo operations using the reverse operation instead of ctrl + z. For example, ListItem currently ignores the __indent property because it handles it with nesting. But if I have indented paragraphs, I would like the indents to be preserved when I convert them to a list and return them to paragraphs. Or if I have a checklist checked, convert it to something else, and return to the checklist, the "completed" property should be preserved. I recommend seeing [this notion blog article](https://www.notion.so/blog/data-model-behind-notion#type-and-properties) where they talk about this.
> 3. If the different types of elementNodes are subclassed and we have a shared property, then the same display logic must be defined in the updateDOM and createDOM of each node, which makes it very difficult to maintain the code. In Lexical right now there is a workaround with the __indent case that is handled in Reconcilier.ts. Besides being a bit confusing, the rest of us can't do the same for other shared properties we might want to add.

# Things I've Tried

So once I had the node replacements API available, I decided to ignore issues 1 and 2 for the moment and focus on issue 3. That is, I tried to reduce the excessive code repetition required to add a shared property to multiple nodes at once; and I did it using a combination of the node replacement API and the [mixin pattern](https://www.typescriptlang.org/docs/handbook/mixins.html
).

[...]

PR #3920, opened 2023-02-16
Generic clone method

GermanJablo:
This is the first of a series of changes I proposed in #3763 to simplify Lexical, remove code duplication, and improve DX.

In short, thanks to this PR no node needs to define a static clone() method.

The most important thing (and the first thing I recommend to check), is the (non-static) clone method that I put in LexicalNode.ts.

Almost whenever the clone() method was used, the __parent, __next, etc. properties were also copied. This was done with $cloneWithProperties or simply by repeating its logic as in getWritable(). The new clone() method I've implemented brings this in for free.

The only part of the repository where these properties are required to be "empty" is in $copyNode, so I set them to null there.

Based on my testing, in the most common case where those properties are required to be cloned, the new implementation is almost 60X faster.


packages/lexical/src/LexicalNode.ts
```diff
@@ -559,27 +557,8 @@ export class LexicalNode {
...
-      mutableNode.__indent = latestNode.__indent;
-      mutableNode.__format = latestNode.__format;
-      mutableNode.__dir = latestNode.__dir;
-    } else if ($isTextNode(latestNode) && $isTextNode(mutableNode)) {
-      mutableNode.__format = latestNode.__format;
-      mutableNode.__style = latestNode.__style;
-      mutableNode.__mode = latestNode.__mode;
-      mutableNode.__detail = latestNode.__detail;
-    }
+    const mutableNode = latestNode.clone();
```
EgonBolton, line 560: This is the part where I did a benchmark.
I measured the time it takes for Lexical to clone all the nodes [in this large document](https://www.slatejs.org/examples/huge-document). The result is that the new implementation is 59X times faster.

EgonBolton:
Why does the deployment fail?

packages/lexical/src/LexicalNode.ts
```diff
@@ -182,12 +182,11 @@ export class LexicalNode {
...
 
-  static clone(_data: unknown): LexicalNode {
-    invariant(
-      false,
-      'LexicalNode: Node %s does not implement .clone().',
-      this.name,
-    );
+  clone(): this {
+    let clone = Object.create(Object.getPrototypeOf(this));
+    const deepClone = JSON.parse(JSON.stringify(this));
```
EgonBolton, line 187: This is the most important part!
After trying a lot of things, I based on this [SO answer](https://stackoverflow.com/a/44782052/10476393), but added the deepClone so that if any node uses an object as a property, it doesn't cause problems (as jduhls warns in the subcomments).

packages/lexical/src/LexicalNode.ts
```diff
@@ -182,12 +182,11 @@ export class LexicalNode {
...
+    clone = Object.assign(clone, deepClone);
+    return clone;
```
EgonBolton, line 189: This is the most important part!
After trying a lot of things, I based on this [SO answer](https://stackoverflow.com/a/44782052/10476393), but added the deepClone so that if any node uses an object as a property, it doesn't cause problems (as jduhls warns in the subcomments).

acywatson:
> Why does the deployment fail?

No idea - it looks like playwright download failed during the npm install part of the build. A retry might succeed. I don't think it was anything you did.

packages/lexical/src/LexicalNode.ts
acywatson: Can you provide some more details on how you benchmarked this? Would like to repro it and want to make sure I'm experimenting under the same or similar conditions.

packages/lexical/src/LexicalNode.ts
trueadm: This looks like a very slow way to clone. `getPrototypeOf` is a de-opt in all JS VM engines and using JSON to stringify and parse will break in many cases. Have you seen `structuredClone` API?

packages/lexical/src/LexicalNode.ts
EgonBolton: I added these 2 console.time(). mutableNode1 is not used, it is only for benchmarking purposes.

```ts
  getWritable(): this {
    errorOnReadOnly();
    const editorState = getActiveEditorState();
    const nodeMap = editorState._nodeMap;
    const key = this.__key;
    // Ensure we get the latest node from pending state
    const latestNode = this.getLatest();
    const parent = latestNode.__parent;
    const cloneNotNeeded = editor._cloneNotNeeded;
    if (selection !== null) {
	@@ -559,27 +557,8 @@ export class LexicalNode {
      internalMarkNodeAsDirty(latestNode);
      return latestNode;
    }
    console.time("1");
    const mutableNode1 = latestNode.clone();
    console.timeEnd("1");
    console.time("2");
    // @ts-expect-error
    const mutableNode = constructor.clone(latestNode);
    mutableNode.__parent = parent;
    mutableNode.__next = latestNode.__next;
    mutableNode.__prev = latestNode.__prev;
    if ($isElementNode(latestNode) && $isElementNode(mutableNode)) {
      mutableNode.__first = latestNode.__first;
      mutableNode.__last = latestNode.__last;
      mutableNode.__size = latestNode.__size;
      mutableNode.__indent = latestNode.__indent;
      mutableNode.__format = latestNode.__format;
      mutableNode.__dir = latestNode.__dir;
    } else if ($isTextNode(latestNode) && $isTextNode(mutableNode)) {
      mutableNode.__format = latestNode.__format;
      mutableNode.__style = latestNode.__style;
      mutableNode.__mode = latestNode.__mode;
      mutableNode.__detail = latestNode.__detail;
    }
    console.timeEnd("2");
    cloneNotNeeded.add(key);
    mutableNode.__key = key;
    internalMarkNodeAsDirty(mutableNode);
    // Update reference in node map
    nodeMap.set(key, mutableNode);
    return mutableNode;
  }
```

Now I realize that I didn't put `mutableNode.__key = key;` in the console.time("2").

packages/lexical/src/LexicalNode.ts
EgonBolton: As far as `getPrototypeOf` is concerned, I first tried using new this.constructor(). The problem is that this fails on nodes with required constructor parameters like headingNode or listNode .

As far as `structuredClone` is concerned, it doesn't seem like it's any faster. I am doing something wrong?
image

acywatson: > using JSON to stringify and parse will break in many cases

I think the issue is more that it will break a lot of runtime objects that aren't JSON serializable (Map, Set, WeakMap, Date, etc, etc)

trueadm: That's why I suggested `structuredClone`. It works with those objects too.

acywatson: > That's why I suggested `structuredClone`. It works with those objects too.

Yea that's what I figured - I think @EgonBolton was focusing on the performance aspect, when it's really the difference in functionality that is key here.

EgonBolton: Ok.
I think Jest currently accepts structuredClone only in the alpha version? https://github.com/facebook/jest/pull/12631
I'm not sure, but it may also be necessary to upgrade to nodejs +17 in github actions since the structuredClone was incorporated there.

packages/lexical/src/LexicalNode.ts
```diff
@@ -182,12 +182,11 @@ export class LexicalNode {
     );
   }
 
-  static clone(_data: unknown): LexicalNode {
-    invariant(
-      false,
-      'LexicalNode: Node %s does not implement .clone().',
-      this.name,
-    );
+  clone(): this {
```
acywatson, line 185: Another thing I thought of - if we end up doing this, does this even need to be an instance method on LexicalNode? can it just be a standalone function?

packages/lexical/src/LexicalNode.ts
acywatson: I can take a crack at some upgrades there hopefully this evening. Thanks for the PR

packages/lexical/src/LexicalNode.ts
EgonBolton: In fact, I would like to know what criteria you use to decide when something should be a class method or a function. I often find myself wondering the same thing...

packages/lexical/src/LexicalNode.ts
acywatson: jest + node upgrades are merged

EgonBolton: I have merged main to PR, but Jest still does not accept `structuredClone`

packages/lexical/src/LexicalNode.ts
acywatson: Yea I don't actually care too much about this piece. Let's do the changes to clone implementation then we can mess with the API later if we want.

packages/lexical/src/LexicalNode.ts
acywatson: > I have merged main to PR, but Jest still does not accept `structuredClone`. Have you been able to make it work on your side @acywatson?

@EgonBolton I will check this evening

packages/lexical/src/LexicalNode.ts
```diff
@@ -188,18 +188,41 @@ export class LexicalNode {
...
+        }
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const cloned = new (this.constructor as new (...args: any[]) => this)(
+        ...Object.values(props),
+      );
+      return Object.assign(cloned, props);
+    }
   }
```
EgonBolton, line 226: @acywatson, as you can see here I offer two possible implementations of the clone method (the two "try" and "catch" blocks respectively).

The first strategy (the one with the "try" block) uses structuredClone. The problem is that [structuredClone can't clone DOM nodes](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#things_that_dont_work_with_structured_clone), lexicalEditor has DOM nodes, and therefore it was failing on nested editors.
The strategy I used a few commits back is to override the clone methods on imageNode and stickyNode, just like we're currently doing on the import/exportJSON PR.

However, then chatGPT came up with a generic way to clone nodes, even for nested editors or any non-serializable structure. It does not require structuredClone or its core-js polyfill. This is the strategy you can see in the catch block.
The problem with this strategy is that it can cause problems if the node's constructor has parameters that are not direct properties. For example, LinkNode groups three optional properties under the attributes object in the constructor.

So, we have different options:
1. Maintain both strategies using try-catch (the catch block accesses when you can't clone with structuredClone). The only way this won't work is if a node satisfies (a) it uses a non-clonable property like a nested editor and (b) its constructor uses parameters that aren't properties of the node. It would be a rather rare case for something like this to happen.
2. Leave only the try strategy (structured clone): the image and sticky nodes would have to define their own clone method, and each Lexical user would have to ensure that they do the same if using non-clonable properties.
3. Leave the catch strategy alone: LinkNode should be refactored and Lexical should document the aforementioned restriction on node constructors.

Perhaps to choose between those options it can help to know what strategy will be used for the import/exportJSON. Currently in that PR we are using a very similar strategy to (2), and @fantacktuka seems to [have an idea for how to automate the case of nested Lexical editors](https://github.com/facebook/lexical/pull/3931#discussion_r1118188574) as well.
I've tried to think of some strategy that would allow cloning and serializing/deserializing any node and in any condition, but for now this is as far as I've been able to go.

EgonBolton: Yeah, I don't like option 3 either.
I'm not sure which one I like better between 1 and 2, I think I'm happy with either one.
In the case of 2, the clone method of ImageNode and StickyNode require a slight refactoring. The method is not static and does not receive any parameters, but instead clones "this" ([see example](https://github.com/facebook/lexical/commit/87eea6371ff07b687968aa1a8c904d11cc790cb1)).

trueadm: I benchmarked the above code and it seems to be significantly slower than what we have today. It's almost 11x slower in fact. It's also 43x slower when we force the JIT and using d8, you can see the instruction set is now vastly more complicated.

One of the original reasons for wanting to promote a `clone` method originally was so that it forced the user to think about how they might clone their nodes – and as such, normalize the properties that their node has so that they're serializable and even clonable to begin with. There are some special edge-cases, like nested editors, but that's handled by exposing internal methods to help the user.

EgonBolton: I also benchmarked the two strategies and did not notice underperformance in either one. Can you explain how you did it?
Are you measuring the entire try-catch process, or both blocks separately?

On the other, both this PR and the serialization PR allow you to override clone, import/exportJSON.

trueadm: I created an update function that creates 10k heading nodes, each with a text node. I then run this function 10 times after the initial first time and then get captured time (excluding the first run, so it can warm up). I change between what's on main and what's in this PR.

acywatson: > One of the original reasons for wanting to promote a clone method originally was so that it forced the user to think about how they might clone their nodes – and as such, normalize the properties that their node has so that they're serializable and even clonable to begin with. There are some special edge-cases, like nested editors, but that's handled by exposing internal methods to help the user.

I haven't had a chance to verify the benchmarks - we should get to the bottom of that. Looking at this more, I can see how the object creation + clone + assign could be slower than calling the constructor directly, but it would be worth ensuring that we're testing what we'd intend to ship, which is just the "try" block.

On the other hand, while I understand why you would want people to think about cloning, I do see people constantly tripping over subclassing nodes because they have to do this themselves. They often forget to pass args to the constructor or pass the wrong args or just forget to even add a clone method. At that point, it's clear that most people don't understand what they're doing when they implement clone, so it just becomes a boilerplate footgun.

If we confirm the performance discrepancy, we can look at alternatives to making it easier for people to use clone.

trueadm: Completely. I didn't envisage people stumbling on this, but that might be down to lack of using TS. I think the best approach here is to allow for a custom `clone` but fallback to manually enumerating through each property on the class and passing that over to the clone object. In DEV time, we can check if a property has a prototype other than that of standard "object", pushing them to maybe adopt a custom `clone` method. This would improve performance on the current ideas in this PR and avoid needing a try/catch entirely.

EgonBolton: Ok, I have benchmarked 6 clone strategies that you can see in my [clone-benchmark branch](https://github.com/EgonBolton/lexical/blob/Clone-benchmark/packages/lexical/src/LexicalNode.ts#L192-L297).

An important point is that the various clones of this PR differ from the current static clone method in that they also include properties like `prev, next, size, etc.`. Most of the time when Lexical clones a node, it immediately proceeds to copy those properties as well.

That is why in the benchmark the `oldNode` strategy is the sum of the static method + the copy of these properties, as it appeared for example in getWritable. There is a related discussion on this topic [here](https://github.com/facebook/lexical/pull/3920/files/fe315852f81cfa5f92aa7826a9c09e77696b7bb8#r1108071691).

I don't know what I did differently when I commented that the same implementation was 59 times faster for me.
Either way, I have repeated the same experiment. I measured the time it takes for Lexical to clone all the nodes [in this large document](https://www.slatejs.org/examples/huge-document). Basically, I paste the document, and when I copy it again, Lexical clones all the nodes and they show up in the console at the time of each one. These are the results:

image

Although neither seems terribly slow, the `not-stringify-nor-structured` implementation seems like the perfect candidate. It works generically in ALL cases and is the second fastest of all after `no-valid-with-objects` (which is one of the most restrictive strategies). Both are faster than the current implementation (`oldnode`).

The `not-stringify-nor-structured` strategy is basically a mix between the `catch` block strategy we are discussing here (in the benchmark it would be `constructor-constraint`), and most others that use ` Object.create(Object.getPrototypeOf(this));`), instead of calling the constructor.

TO UPDATE: in the next 10 minutes I'm going to be pushing the `not-stringify-nor-structured` strategy to this PR.

Please let me know what you think.

trueadm: I’d be a bit skeptical if using copy/paste as the defining heuristic. It tests a different code path for cloning AFAIK. Can you instead test like I mentioned before? Have an editor.update that does the creation. Add console.time logs at the start of the update and end of the update function and measure that for a large document? Make sure you run the benchmark function a few times first to warm it up.

acywatson: I was thinking - even if this implementation is a little slower, we could theoretically leave the overrides in place on the core nodes and just make this the default impl for node subclasses. Then you get the dx and the perf. My hesitation would be that that leads users away from the pit of success from a perf standpoint and I don’t know if the dx win is quite enough to offset that.  If people are doing a lot of custom nodes and don’t optimize their clone methods, they’d have to dig to figure out why things were slow.

fantactuka: One more option to explore is if we can calculate list of clonable props during node registration in editor constructor. Or cache it during first clone call for particular node type

EgonBolton: Have you seen the benchmark? I don't see that there are different paths, the strategies are executed in the same method one after another.
I also just removed the first few iterations to get rid of the cold start thing you mentioned and it didn't change anything.

The current implementation is generic, it works with properties of any type, does not use cloning or requires exceptions, and according to my benchmark it is faster than the current implementation. 🙂

trueadm: @EgonBolton Copy and paste uses this codepath for cloning AFAIK: https://github.com/facebook/lexical/blob/main/packages/lexical-selection/src/lexical-node.ts#L66-L84

EgonBolton: Yes, and that was basically the code I was benchmarking.

> An important point is that the various clones of this PR differ from the current static clone method in that they also include properties like prev, next, size, etc.. Most of the time when Lexical clones a node, it immediately proceeds to copy those properties as well.
> 
> That is why in the benchmark the oldNode strategy is the sum of the static method + the copy of these properties, as it appeared for example in getWritable.

However, I have now tested it in this plugin and now find `not-stringify-nor-structured` ~20X slower than `oldClone`. The only one that comes close is `no-valid-with-objects` but still ~3 times slower.

```tsx
import {$createHeadingNode} from '@lexical/rich-text';
import {$createTextNode, $getRoot, LexicalNode, NodeMap} from 'lexical';

export function BenchmarkPlugin() {

  function handleBenchmark(cloneFunction: (node: LexicalNode) => void): void {
    editor.update(() => {
      const root = $getRoot();
      root.clear();
      for (let i = 0; i < 1000; i++) {
        root.append($createHeadingNode('h1').append($createTextNode('Hi')));
      }
      const nodes = editor._pendingEditorState?._nodeMap as NodeMap;
      cloneFunction(root); // warm up

      console.time('time');
      for (const node of nodes.values()) {
        cloneFunction(node);
      }
      console.timeEnd('time');
    });
  }

  return (
    <>
      <button onClick={() => handleBenchmark((node) => node.clone())}>
        Clone New
      </button>
      <button
        onClick={() => handleBenchmark((node) => node.constructor.clone(node))}>
        Clone old
      </button>
      <button
        onClick={() => handleBenchmark((node) => $cloneWithProperties(node))}>
        Clone Old w/props
      </button>
      <button
        onClick={() =>
          handleBenchmark((node) =>
            Object.assign(Object.create(Object.getPrototypeOf(node)), node),
          )
        }>
        No valid with objects
      </button>
    </>
  );
}
```

I don't really understand the difference of `$cloneWithProperties(node)` here with `oldClone` on my first benchmark, which measure the same code 🤔. I have reviewed my code again, and there was only a slight error in `oldClone` and another in `no-valid-with-objects` (I have already pushed to my branch), but they do not change the positions of the strategies:


image

My intuition tells me that these results in the table make a lot more sense. And in fact, I think it's a good idea to benchmark the strategies under normal conditions of use (copy and paste, writing a document, etc.). The only reason for the difference between the two benchmarks that I can think of is that in the plugin benchmark above the javascript engine is doing some magic when creating nodes immediately after others, in which case I don't think it should be a decision parameter.

trueadm: Thanks again for digging into this, I think it's important to understand what is happening before making any characterisations around implementations. What are your results for `$cloneWithProperties(node)` vs `oldClone`? What mistakes did you find too?

I have a hunch that something in the entire copy/paste flow is causing a de-opt, which means that the clone strategy optimizations are being alienated because of this. This is the beauty of JS VMs, some other code somewhere can actually undo the performance tweaks that were applied somewhere else. Likely, this is a hidden class problem that occurs only during copy/paste. The reason I suggested a simple benchmark was to isolate only what you're trying to benchmark here – the cloning itself. Otherwise, you're not actually capturing the impact fully of the change and instead you're capturing the change as a subsequence of an event flow – in this case copying and pasting.

What would be good to do, is find out what is causing the performance regression in copy/paste. This might be a bigger impact win on performance than anything we're discussing in this thread!

EgonBolton: `oldClone` **IS** `$cloneWithProperties`. It's basically the alias I gave to that console.time.

I just checked, and it's not like copy-paste is "turning off an optimization". I would rather say that the js engine does some optimization in `for (const node of nodes.values()) { node.constructor.clone(node) }`, which does not apply to `node.constructor.clone(node)` .

How do I know? replace in the snippet of the plugin that I gave you `i < 1000` by `i < 1`, and you will see how the different strategies automatically match the table that I placed from my first benchmark.

trueadm: I'd be happy to chat about this with you and explain what is happening at a VM level. What might be interesting for you to discover, is the bytecode being generated in each instance. If you put your function into a single script on your filesystem and run this command:

`
node --print-bytecode cloneMethod.js
`

You'll see the output from d8 :)

EgonBolton: It sounds interesting, and I would very much like to do it at some point. But it seems like it would take me more time than I'm willing to spend on this problem right now.

I have made one last effort and I have come up with a new strategy:
  ```ts
clone(): this {
     const clone = Object.create(Object.getPrototypeOf(this));
     for (const prop in this) {
       clone[prop] = this[prop];
     }
     return clone;
   }
```
In BenchmarkPlugin it is about twice as slow as `$cloneWithProperties` for `i < 1000`. On the other hand, in the benchmark where all strategies with alternation are tested, it is twice as fast as `$cloneWithProperties`:

image

So I decided to do a final test with the best candidate. I tried to do the copy test but in branches where no other strategy interfered. `main` for oldNode and this PR for the new strategy. The result? The new strategy is three times as fast:

image

Comment (trueadm) [reply packages/lexical/src/LexicalNode.ts]: That's a good fallback. We'd also have to warn people to not use private properties on their nodes though – but maybe that's okay given I've only seen a few instances of them being used.

Comment (acywatson) [reply packages/lexical/src/LexicalNode.ts]: Sure, we could add that, but I don't think that's what's holding this up. I just need to have chat with the core team and make sure we have general agreement on this. It's a large change.

**Comment by EgonBolton** [comment]
- I added the missing comment you mentioned to me.
- I removed key from the constructor as we discussed on Discord.
- Updated from the main branch and resolved conflicts.

The only problem now is that the unit test `Has correct text point after removal after merge` is failing and I can't figure out why.

I tried to replace the update editor with this, but it didn't work:
```ts
       editor.update(() => {
         const root = $getRoot();
        
         const a = $createTextNode('a');
         const bb = $createTextNode('bb');
         const empty = $createTextNode('');
         const cc = $createTextNode('cc');
         const d = $createTextNode('d');
         root.append(paragraph.append(a, bb, empty, cc, d));

         setAnchorPoint({
           key: bb.getKey(),
           offset: 1,
           type: 'text',
         });

         setFocusPoint({
           key: cc.getKey(),
           offset: 1,
           type: 'text',
         });
       });
```

potatowagon:

etrepum:
I do think we should go with this sort of approach eventually, but maybe a phased approach makes sense? Or at least something with a codemod?

I think the most important thing to do is to require all node classes to allow for a 0-argument constructor and then we can move all of the important and awkward static stuff to instances (clone, importJSON) post-construction. Passing the NodeKey on construction can be deprecated and then later removed. We could set and reset a module global on clone to seed `$setNodeKey` so that existingKey is never used.

etrepum:
This PR changes all of the constructor signatures and anyone subclassing them is going to have to change their super calls and constructor signatures accordingly to be compliant with the types.

Object.create/assign and JSON.parse/stringify are simply not sound solutions in the general case, it adds extra limitations on what properties are allowed and how the subclasses need to be implemented (private properties aren't enumerable or serialized by default for example). There are of course ways to work around that with the sort of boilerplate overrides that we do now, but I am very hesitant to propose a situation where the constructor code simply does not run for clones.

GermanJablo:
> Object.create/assign and JSON.parse/stringify are simply not sound solutions in the general case, it adds extra limitations on what properties are allowed and how the subclasses need to be implemented (private properties aren't enumerable or serialized by default for example). 

That's fine. The goal of this PR was that for non-JSON serializable nodes (which is the great minority), the user can continue using the manual method

PR #3925, opened 2023-02-16
fix: change tablecell return type for TableCellNode

Sebastien-Ahkrin:
Actually, if i use the node override like [that](https://lexical.dev/docs/concepts/node-replacement). 
i can't pass something else than "tablecell" in getType function. (Or i have to change with a as or to use a // @ts-ignore)

```tsx
export class CustomTableCellNode extends TableCellNode {

  public static getType(): 'tablecell' {
    return 'tablecell';
  }

  public static clone(node: TableCellNode): TableCellNode {
    return new CustomTableCellNode(
      node.__headerState,
      node.__colSpan,
      node.__width,
      node.__key,
    );
  }

  public createDOM(config: EditorConfig): HTMLElement {
    const dom = super.createDOM(config);
    return dom;
  }
}
```

```tsx
  public static getType(): 'tablecell' {
    console.log('getType called');
    // @ts-ignore
    return 'custom-tablecell';
  }
```

With this, now we should be able to use a string while extending TableCellNode.
Let me know if its good for you!

### DIFF
--- a/packages/lexical-table/src/LexicalTableCellNode.ts
+++ b/packages/lexical-table/src/LexicalTableCellNode.ts
@@ -40,7 +40,7 @@ export type TableCellHeaderState =
 export type SerializedTableCellNode = Spread<
   {
     headerState: TableCellHeaderState;
-    type: 'tablecell';
+    type: string;
     width?: number;
   },
   SerializedGridCellNode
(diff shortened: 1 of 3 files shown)


thegreatercurve (approved):
Thanks for the fix!

PR #3930, opened 2023-02-17
Account for children text content size in `ElementNode`'s `getTextContentSize`

### LINKED ISSUE #3772: Bug: `$getRoot().getTextContentSize()` does not account for overriden `getTextContentSize` in descendants
Lexical version: 0.7.6

## Steps To Reproduce

1. Create custom node that overrides both `getTextContent` and `getTextContentSize`, where `getTextContent().length !== getTextContentSize()`;
2. Call `$getRoot().getTextContent()` and `$getRoot().getTextContentSize()`, they will both account for the `getTextContent` override but not the `getTextContentSize` override.

Link to code example: https://codesandbox.io/s/lexical-gettextcontentsize-2if3vk (forked from https://codesandbox.io/s/lexical-plain-text-example-g932e).

In `EmojiNode.js` I override `getTextContent` to return `emoji:${this.__text}` and `getTextContentSize` to return `1`.
Add an emoji to the editor by typing `:)`, in the console it will log:

```
getTextContent emoji:🙂 
getTextContentSize 8
```

The output logged for `getTextContent` is expected, the one for `getTextContentSize` is not. I expect it to be `1` instead of `8`.

## Motivation

We have a custom equation node. To customize the text/plain clipboard behavior we return `"$" + this.equation + "$"` as text content. But in order to enforce a maximum length on the editor text content we override `getTextContentSize` to return the length of `this.equation`, without accounting for the delimiters "$".
Is there a better approach?

GiacomoRandazzo:
Fixes #3772 

In `ElementNode` I've duplicated the logic of `getTextContent` for `getTextContentSize()` to account for children where `getTextContentSize() !== getTextContent().length`.

I'm not sure that the changes are desirable.
In particular, `RootNode` overrides `getTextContent` to track the internal property `__cachedText`. 
`__cachedText` is changed manually when the editor is updated. I lack familiarity with the codebase but my guess is that it's a performance optimization.
In the current state of the PR, the same optimization is not applied to `getTextContentSize` in `RootNode`. 
What are the implications of this? I see that `getTextContentSize` is used internally in `LexicalSelection.ts`.

### DIFF
--- a/packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
+++ b/packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
@@ -255,12 +255,12 @@ describe('LexicalElementNode tests', () => {
     test('basic', async () => {
       await update(() => {
         expect($getRoot().getFirstChild().getTextContentSize()).toBe(
-          'FooBarBaz'.length,
+          $getRoot().getFirstChild().getTextContent().length,
         );
       });
     });
 
-    test('child getTextContentSize() does not match getTextContent().length', async () => {
+    test('child node getTextContentSize() can be overridden and is then reflected when calling the same method on parent node', async () => {
       await update(() => {
         const block = $createTestElementNode();
         const text = $createTextNode('Foo');
--- a/packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
+++ b/packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
@@ -251,6 +251,27 @@ describe('LexicalElementNode tests', () => {
     });
   });
 
+  describe('getTextContentSize()', () => {
+    test('basic', async () => {
+      await update(() => {
+        expect($getRoot().getFirstChild().getTextContentSize()).toBe(
+          'FooBarBaz'.length,
+        );
+      });
+    });
+
+    test('child getTextContentSize() does not match getTextContent().length', async () => {
+      await update(() => {
+        const block = $createTestElementNode();
+        const text = $createTextNode('Foo');
+        text.getTextContentSize = () => 1;
+        block.append(text);
+
+        expect(block.getTextContentSize()).toBe(1);
+      });
+    });
+  });
+
   describe('splice', () => {
     let block;
 
--- a/packages/lexical-playground/src/plugins/MaxLengthPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/MaxLengthPlugin/index.tsx
@@ -24,20 +24,19 @@ export function MaxLengthPlugin({maxLength}: {maxLength: number}): null {
         return;
       }
       const prevEditorState = editor.getEditorState();
-      const prevTextContent = prevEditorState.read(() =>
-        rootNode.getTextContent(),
+      const prevTextContentSize = prevEditorState.read(() =>
+        rootNode.getTextContentSize(),
       );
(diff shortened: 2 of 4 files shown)


packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
```diff
@@ -251,6 +251,27 @@ describe('LexicalElementNode tests', () => {
...
+  describe('getTextContentSize()', () => {
+    test('basic', async () => {
+      await update(() => {
+        expect($getRoot().getFirstChild().getTextContentSize()).toBe(
+          'FooBarBaz'.length,
+        );
+      });
+    });
+
+    test('child getTextContentSize() does not match getTextContent().length', async () => {
```
thegreatercurve, line 263:
```suggestion
    test('child node getTextContentSize() can be overridden and is then reflected when calling the same method on parent node', async () => {
```

packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
```diff
@@ -251,6 +251,27 @@ describe('LexicalElementNode tests', () => {
     });
   });
 
+  describe('getTextContentSize()', () => {
+    test('basic', async () => {
+      await update(() => {
+        expect($getRoot().getFirstChild().getTextContentSize()).toBe(
+          'FooBarBaz'.length,
```
thegreatercurve, line 258:
```suggestion
          $getRoot().getFirstChild().getTextContent().length,
```

thegreatercurve:
@GiacomoRandazzo Generally, I think this seems sensible and this is probably a use-case we should be accommodating.  

You're right that the `__cachedText` property on the root is a performance optimisation which is used during reconciliation. We need to keep the property on the root node, but it's probably okay to remove it from the logic of `getTextContent` on the root to fix the bug you've mentioned.

It's added as an performance optimization, but it seems we only really use it in the max length plugin.

GiacomoRandazzo:
I've updated the max length plugin and the `trimTextContentFromAnchor` function to account for `getTextContentSize()`.

---

I have an alternative proposal that might be more future proof.

As described in #3772, the context in which I need `getTextContent().length !== getTextContentSize()` is in copying a custom equation node with `$` delimiters around the equation and, at the same time, have the max length plugin account for the equation length without the delimiters.
Currently `getTextContent` is used both for copying as `text/plain` and to compute the node length for the max length plugin.

The current PR solves the problem by allowing `getTextContentSize` to differ from the length of the node text content. This could lead to inconsistencies in the internal logic of Lexical (e.g. what if a non-empty node has `getTextContentSize = () => 0`?). From a dev point of view is hard to grasp the implications of overriding `getTextContentSize()`.

An alternative solution is to have a way for custom nodes to override the behavior for copying as `text/plain`, similarly to how `exportDOM` is used to copy as `text/html`.
Lexical internals could assume `getTextContent().length === getTextContentSize()`, which is more future proof. 
At this point `getTextContentSize()` could be removed altogether to avoid confusion.

thegreatercurve:
Thanks for the proposal @GiacomoRandazzo! I like the idea but I think, given the rarity of the issue, introducing a whole new `export/importPlainText` API feels somewhat like overkill. If you overwrite `getTextContentSize` on core nodes whilst doing a node replacement, developers know they're doing so at their own risk of bugs. It's definitely less of an issue though in cases such as yours where it's a one-off extended node.

thegreatercurve (approved):

PR #3931, opened 2023-02-17
WIP: Generic ImportJSON/ExportJSON

GermanJablo:
This is the second in a series of changes I proposed in #3763 to simplify Lexical, remove code duplication, and improve DX.

In short, thanks to this PR, no node needs to define an `importJSON` or `exportJSON` method.

The most important thing (and the first thing I recommend checking) are the changes in `LexicalUpdate.ts` (for `importJSON`) and in `lexicalNode.ts` (for `exportJSON`).

### Pending:

- [ ] The `mode`, `direction` and `format` properties mismatch between the way they are exported and imported making the algorithm more inefficient. I have a PR in mind that would improve this. Either way, the PR as it is is now functional.
- [ ] There are currently 5 unit tests failing because exportJSON seems to work fine, it just changes the order of the keys. My question is, is it okay if I modify those tests to compare objects instead of the stringified editorState?


packages/lexical/src/LexicalUpdates.ts
```diff
@@ -309,17 +309,33 @@ function $parseSerializedNodeImpl<
...
+    type !== 'table'
+  ) {
+    // @ts-expect-error
+    node.setDirection(serializedNode.direction);
+  }
+  const prefix = '__';
+  const withUnderscore = Object.fromEntries(
+    Object.entries(serializedNode2).map(([k, v]) => [`${prefix}${k}`, v]),
+  );
+  node = Object.assign(node, withUnderscore);
```
EgonBolton, line 338: This is the most important part.

As a side note, one of the ideas I have for simplifying setters and getters is to use a generic proxy that accesses getLatest() and getWritable(), so probably the underscore `__` thing may not be necessary anymore. here or anywhere.

acywatson:
I’m not sure this makes sense. The purpose of this API is to allow custom nodes to control deserialization behavior, which allows for runtime backwards compatibility for Lexical states in storage. With this change, you’ve taken that flexibility and control away from developers and instead created a place in the core that will have to be continuously expanded and special-cased as new core node types are added.

packages/lexical/src/LexicalNode.ts
```diff
@@ -624,16 +624,32 @@ export class LexicalNode {
...
+      serializedNode.direction = this.getDirection();
+      serializedNode.format = this.getFormatType();
+      serializedNode.indent = this.getIndent();
+    }
+    if ($isTextNode(this)) {
+      serializedNode.mode = this.getMode();
+    }
+    serializedNode.type = this.getType();
+    serializedNode.version = 1;
+    return serializedNode;
```
EgonBolton, line 652: This is the other important part, which replaces `exportJSON()`

EgonBolton:
I've updated the PR to add `exportJSON`. As you can see, the new `exportJSON` method of LexicalNode is generic, but a node could override it by inheritance if someone wanted to.

I can do the same with `importJSON`.

What do you think?

acywatson:
> I've updated the PR to add `exportJSON`. As you can see, the new `exportJSON` method of LexicalNode is generic, but a node could override it by inheritance if someone wanted to.
> 
> I can do the same with `importJSON`.
> 
> What do you think @acywatson?

My recommendation would be for you to review this PR: https://github.com/facebook/lexical/pull/2157

The reason serialization works like this is not because we never thought about just serializing all the properties on the node - that was basically how it was done before. The idea was to create a type-safe abstraction between the serialized types and the internal node structures to insulate against changes to those internal structure breaking old stored Lexical states.

I appreciate the effort, but I don't think we have a lot of appetite to revisit this right now.

EgonBolton:
I have found in the last commit a way to autogenerate the types of the serialized nodes automatically. As you can see in the video, it works with inheritance.

https://www.loom.com/share/4601d7d7411c4eafbfd47d2281bade52

Maybe that solves the problem?

> I appreciate the effort, but I don't think we have a lot of appetite to revisit this right now.

I believe that what I propose is viable and has enormous advantages. But if you don't find value in these changes, please feel free to close the PR. I'd rather use a fork than be a nuisance :)

packages/lexical/src/LexicalNode.ts
```diff
@@ -155,8 +155,6 @@ export type DOMExportOutput = {
 export type NodeKey = string;
 
 export class LexicalNode {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  [x: string]: any;
```
EgonBolton, line 159: This conflicts with the solution I found for serialized node types.
I don't understand why it was done in the first instance. It doesn't seem like a good practice since it gives a false illusion of type-safety, when in fact it breaks it at the beginning of the class hierarchy.

acywatson:
> I believe that what I propose is viable and has enormous advantages. But if you don't find value in these changes, please feel free to close the PR. I'd rather use a fork than be a nuisance :)

@EgonBolton You're not being a nuisance, but I'm also perfectly happy for you to use a fork - it is open source, after all :)

Let me spend some time with your work and see if I can get comfortable with the functionality in the scenarios I care about. Specifically, I need to understand what happens when someone decides to change a node to add a new property in. How does deserialization work with old states that don't have that property? What about when a new state (containing the added property) is deserialized by an old editor? How does the type system behave when the user makes changes to the public serialization interface of a node so that they're warned about the two scenarios above and can react accordingly?

I'm not saying that your proposal doesn't have advantages (they might even be enormous ones!), but it is potentially very consequential for all of our internal systems, which means I need to spend my time verifying it, which impacts my other priorities. That's the only issue here - I'm happy to consider the idea on it's merits and we're better just for having seen the idea, whether we merge it now or not.

I have some free time this weekend, so I will try to check out the branch and verify some of these scenarios and see where I get.

EgonBolton:
I have marked the PR as WIP, because TS is taking me longer than I thought.
`[x: string]: any;` in LexicalNode was hiding a lot of errors from ts. Just to give an example, there were things like `this._rel` in `LinkNode` (one underscore instead of two) that were not being detected.

packages/lexical/src/LexicalNode.ts
```diff
@@ -862,3 +880,46 @@ function errorOnTypeKlassMismatch(
     );
   }
 }
+
+// In the following lines we follow the exportJSON process.
+
+// 1. Eliminate the methods leaving only the properties
```
acywatson, line 886: Did you come up with a similar approach for Flow?

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -643,8 +643,8 @@ function reconcileNode(
     nextNode.__cachedText !== editorTextContent
   ) {
     // Cache the latest text content.
-    nextNode = nextNode.getWritable();
-    nextNode.__cachedText = editorTextContent;
+    const root = nextNode.getWritable();
```
acywatson, line 646: Doesn't seem related?

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -309,17 +307,31 @@ function $parseSerializedNodeImpl<
     invariant(false, 'parseEditorState: type "%s" + not found', type);
   }
 
-  const nodeClass = registeredNode.klass;
-
-  if (serializedNode.type !== nodeClass.getType()) {
```
acywatson, line 314: Doesn't seem like this error message was accurate - maybe we still want this check?

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1538,7 +1538,7 @@ export function $splitNode(
       const [leftTree, rightTree, newParent] = recurse(parent);
       const nextSiblings = currentNode.getNextSiblings();
 
-      newParent.append(nodeToMove, ...nextSiblings);
+      (newParent as ElementNode).append(nodeToMove, ...nextSiblings);
```
acywatson, line 1541: why do we have to cast this now?

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -2226,9 +2226,8 @@ describe('LexicalEditor tests', () => {
           TestTextNode,
           {
             replace: TextNode,
-            // @ts-ignore
             with: (node: TextNode) => new TestTextNode(node.getTextContent()),
-
+            // @ts-ignore
```
acywatson, line 2230: ?

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
```diff
@@ -87,16 +79,6 @@ describe('LexicalNode tests', () => {
         });
       });
 
-      test('LexicalNode.clone()', async () => {
```
acywatson, line 90: ?

packages/lexical/src/LexicalEditor.ts
```diff
@@ -42,7 +42,11 @@ export type Spread<T1, T2> = Omit<T2, keyof T1> & T1;
 export type Klass<T extends LexicalNode> = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   new (...args: any[]): T;
-} & Omit<LexicalNode, 'constructor'>;
+} & {
+  getType: () => string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  importDOM?: () => any;
```
acywatson, line 48: not sure why we'd be changing this/suppressing the linter

packages/lexical-utils/src/index.ts
```diff
@@ -207,6 +208,8 @@ export type DOMNodeToLexicalConversionMap = Record<
   DOMNodeToLexicalConversion
 >;
 
+// TODO: It would be nice if this function returned
+// the type of findFn since sometimes it is a type predicate.
```
acywatson, line 212: File an issue instead of or at least in addition to a TODO

packages/lexical-selection/src/range-selection.ts
```diff
@@ -56,14 +56,13 @@ export function $setBlocksType_experimental(
 
   const nodes = selection.getNodes();
   if (selection.anchor.type === 'text') {
-    let firstBlock = selection.anchor.getNode().getParent() as LexicalNode;
-    firstBlock = (
-      firstBlock.isInline() ? firstBlock.getParent() : firstBlock
-    ) as LexicalNode;
-    if (nodes.indexOf(firstBlock) === -1) nodes.push(firstBlock);
+    let firstBlock = selection.anchor.getNode().getParent();
+    firstBlock =
```
acywatson, line 60: How is this related?

packages/lexical-markdown/src/__tests__/unit/MarkdownImport.test.ts
```diff
@@ -39,7 +40,8 @@ const IMAGE: TextMatchTransformer = {
 
 const EQUATION: TextMatchTransformer = {
   dependencies: [EquationNode],
-  export: (node, exportChildren, exportFormat) => {
+  // @ts-ignore
```
acywatson, line 43: Seems like this might be OK, but can you add a comment with an explanation? Not sure I understand how it's related to import/exportJSON

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -181,7 +163,7 @@ export class ListItemNode extends ElementNode {
         list.insertAfter(replaceWithNode);
         replaceWithNode.insertAfter(newList);
       }
-      if (includeChildren) {
+      if (includeChildren && $isElementNode(replaceWithNode)) {
```
acywatson, line 166: ?

acywatson:
> You can safely remove the importJSON and exportJSON method from each of your custom nodes.

This might not be the case - it depends on what they were doing in these methods, right? If they were, for instance, renaming a serialized property, or deliberately not serializing a property, then simply deleting will change the behavior.

acywatson:
> > You can safely remove the importJSON and exportJSON method from each of your custom nodes.
> 
> This might not be the case - it depends on what they were doing in these methods, right? If they were, for instance, renaming a serialized property, or deliberately not serializing a property, or mapping a serialized property to a node property with a different name, then simply deleting will change the behavior.

acywatson:
Sent a few questions over Discord. Overall, so far I think this might be a good and workable approach - I really like how much it simplifies the node creation process and I don't think we lose much in terms of type safety. There are some drawbacks/issues I see:

1) We are introducing some special-case logic in the core that determines which node properties to delete on export/import and which node properties to handle differently based on whether something is a TextNode or a Table or whatever. It won't be immediately clear to anyone looking at that why or when you would make a change to that logic.

2) I'm not sure I understand exactly how the new serializableNode type supposed to be used. Is this the type you return from exportJSON in a subclass? If so, I can't seem to get that to work as-is with HeadingNode:

```
 exportJSON(): serializableNode<this> {
    return {
      tag: "h1" // <-- type '{ tag: string; }' is not assignable to type 'serializableNode<this>'.
    }
  }
  ```

Even if this did work, I would expect it to force the node to return every non-specifically-omitted property on the class, right?

3) Given that we would be switching to using constructors directly rather than node creator functions, how would this work with $applyNodeReplacement? If I'm not mistaken, deserialized nodes would not have those replacements applied.

Those are the biggest things I see for now - would like to get the rest of the team to take a look, but I'm leaning supportive on this. Will try to keep the ball rolling as you update this - as mentioned earlier, lots of other things going on at the moment, but I am willing to try to move this forward - it just may take some time.

@zurfyx @fantactuka @trueadm

File: packages/lexical/src/LexicalReconciler.ts (hunk 5 above)
Comment (EgonBolton) [reply packages/lexical/src/LexicalReconciler.ts:646]: As we discussed on Discord, this and all subsequent comments on ts have to do with the `[x: string]: any;` line being removed, "exposing" a bunch of bugs that were hidden. This change and therefore all the ones it triggers are necessary so that `exportJSON` can be typesafe.

File: packages/lexical-list/src/LexicalListItemNode.ts (hunk 14 above)
Comment (EgonBolton) [reply packages/lexical-list/src/LexicalListItemNode.ts:166]: Again, this is the fault of ´[x: string]: any;´
append is not a method of LexicalNode.
If questions like this arise, I recommend deleting the addition in your IDE and seeing all the errors that are thrown by ts.

File: packages/lexical-markdown/src/__tests__/unit/MarkdownImport.test.ts (hunk 13 above)
Comment (EgonBolton) [reply packages/lexical-markdown/src/__tests__/unit/MarkdownImport.test.ts:43]: This is the fault of ´[x: string]: any;´
getEquationNode is not a method of LexicalNode.

File: packages/lexical-selection/src/range-selection.ts (hunk 12 above)
Comment (EgonBolton) [reply packages/lexical-selection/src/range-selection.ts:60]: There are methods that are not from LexicalNode, but from some subclass. Errors (previously invisible due to ´[x: string]: any;´) can be seen if you remove this change.

File: packages/lexical-utils/src/index.ts (hunk 11 above)
Comment (EgonBolton) [reply packages/lexical-utils/src/index.ts:212]: I thought it was fine for quick little suggestions, because I see them all over the repository without an open issue.
I guess opening an issue on every detail would take a lot of time, so if you prefer I just delete the to-do comments.

File: packages/lexical/src/LexicalEditor.ts (hunk 10 above)
Comment (EgonBolton) [reply packages/lexical/src/LexicalEditor.ts:48]: By removing `[x: string]: any;`, the line that was here (`Omit<LexicalNode, 'constructor'>;`) generated many errors (which also had a similar intent to allow almost anything.

The reason I had to specify getType and importDOM is because Klass doesn't include the static methods, which is weird because as far as I understand it's a type over classes, not over instances.

File: packages/lexical/src/LexicalNode.ts (hunk 4 above)
Comment (EgonBolton) [reply packages/lexical/src/LexicalNode.ts:886]: Not yet, I'll add this to the to-do list.

File: packages/lexical/src/LexicalUpdates.ts (hunk 6 above)
Comment (EgonBolton) [reply packages/lexical/src/LexicalUpdates.ts:314]: hmm you're right. It wouldn't be necessary in general, but should be left in case someone wants to override the `exportJSON` method.

**Comment by EgonBolton** [comment]
> so far I think this might be a good and workable approach - I really like how much it simplifies the node creation process and I don't think we lose much in terms of type safety. 

Thank you! In reality, this PR would make Lexical 1000x more type-safe than before. `[x: string]: any;` was wreaking havoc!

> 1. We are introducing some special-case logic in the core that determines which node properties to delete on export/import and which node properties to handle differently based on whether something is a TextNode or a Table or whatever. It won't be immediately clear to anyone looking at that why or when you would make a change to that logic.

Different case handling between TextNode and ElementNode is a very common pattern throughout Lexical. Now if you mean the line `if ( $isElementNode(node) && type !== 'tablecell' && type !== 'tablerow' && type !== 'table') { node.setDirection(serializedNode.direction); }` in `$parseSerializedNodeImpl` The explanation is as follows:

`direction`, `format` and `indent`, are defined and exported in all element nodes, except `direction` in those 3 table nodes. Well you can solve that by adding that `if` that I put or modify the unit test that checks the `exportJSON` of the tables.
However, I think these 3 properties should be part of the `setAttributes` that I proposed at https://github.com/facebook/lexical/issues/3763. Table nodes are examples of how we think of these 3 attributes more in terms of composition than inheritance.

____
I will answer the other questions later.

File: packages/lexical/src/LexicalUpdates.ts [hunk 15]
```diff
@@ -318,8 +316,31 @@ function $parseSerializedNodeImpl<
...
+    type !== 'tablerow' &&
+    type !== 'table'
+  ) {
+    node.setDirection(serializedNode.direction);
+  }
+  const prefix = '__';
+  const withUnderscore = Object.fromEntries(
+    Object.entries(serializedNode2).map(([k, v]) => [`${prefix}${k}`, v]),
+  );
+  node = Object.assign(node, withUnderscore) as unknown as LexicalNode;
```
fantactuka, line 343: It probably does not handle nested editors (e.g., see ImageNode.__caption)

EgonBolton: Yes, it's on the to-do list to check if structuredClone instead of JSON.parse works. I think it is solvable.

packages/lexical-website/docs/concepts/nodes.md
```diff
@@ -55,10 +55,9 @@ can output components from React, vanilla js or other frameworks.
 
 ## Node Properties
 
-Lexical nodes can have properties. It's important that these properties are JSON serializable too, so you should never
-be assigning a property to a node that is a function, Symbol, Map, Set, or any other object that has a different prototype
-than the built-ins. `null`, `undefined`, `number`, `string`, `boolean`, `{}` and `[]` are all types of property that can be
-assigned to node.
+Lexical nodes can have properties. You don't have to worry about serializable JSON properties like `null`, `undefined`, `number`, `string`, `boolean`, `{}` and `[]`.
+
+However, it is important that if you define a property that is not JSON serializable (function, symbol, map, array, cyclic object, etc.), you must define `exportJSON` and `static importJSON` methods on the node. An example of this is the `ImageNode` in the Playground code, which because the `__caption` property is a `LexicalEditor` requires custom serialization.
```
EgonBolton, line 60: This modification is a good summary of what this PR does. Basically, you can still define serialization, but it's only needed for non-serializable properties.

The only occasion or example where that is necessary is in `ImageNode`, which has an `importJSON` due to the `caption` property.

packages/lexical/src/LexicalUpdates.ts
EgonBolton: Resolved. Please read the comment I just updated above.

packages/lexical/src/__tests__/unit/LexicalSerialization.test.ts
```diff
@@ -111,16 +111,20 @@ describe('LexicalSerialization tests', () => {
...
+      expect(JSON.parse(stringifiedEditorState)).toStrictEqual(
+        JSON.parse(
```
EgonBolton, line 115: It's a pity that the Github diff doesn't highlight only the changed parts of the string like it does in VS Code.

What I had to do here and in a couple of other unit tests, is to add `"colSpan": 1` to the tableCell nodes, which for some reason were not being exported. This is probably because tableCell's `exportJSON` was exporting colSpan as `super.__colspan`.

I already mentioned this to @acywatson

packages/lexical/src/__tests__/unit/LexicalEditorState.test.ts
```diff
@@ -95,8 +95,10 @@ describe('LexicalEditorState tests', () => {
         $getRoot().append(paragraph);
       });
 
-      expect(JSON.stringify(editor.getEditorState().toJSON())).toEqual(
-        `{"root":{"children":[{"children":[{"detail":0,"format":0,"mode":"normal","style":"","text":"Hello world","type":"text","version":1}],"direction":"ltr","format":"","indent":0,"type":"paragraph","version":1}],"direction":"ltr","format":"","indent":0,"type":"root","version":1}}`,
+      expect(editor.getEditorState().toJSON()).toEqual(
+        JSON.parse(
+          `{"root":{"children":[{"children":[{"detail":0,"format":0,"mode":"normal","style":"","text":"Hello world","type":"text","version":1}],"direction":"ltr","format":"","indent":0,"type":"paragraph","version":1}],"direction":"ltr","format":"","indent":0,"type":"root","version":1}}`,
+        ),
```
EgonBolton, line 101: I had to modify this and about 3 other tests so that it compares the parsed JSON, not stringifed, because sometimes it is identical but only the order of the keys changes.

packages/lexical-website/docs/concepts/serialization.md
```diff
@@ -184,120 +184,37 @@ const editorState = editor.getEditorState();
...
-    // Notice the version is now 2
-    version: 2,
-  },
-  SerializedLexicalNode
->;
-
-export type SerializedTextNode = SerializedTextNodeV1 | SerializedTextNodeV2;
-```
+}
+```
```
EgonBolton, line 220: I tried to improve this section of the documentation. The version concept wasn't really explaining how breaking changes can be dealt with. I tried to explain how `exportJSON` and `importJSON` can solve this with an example. (Though I think it's better to sanitize the DB than to patch the serialization.)

Anyway, could you give me your feedback? What do you think of this section?

packages/lexical-link/src/index.ts
```diff
@@ -377,9 +325,7 @@ export function toggleLink(
       const firstNode = nodes[0];
       // if the first node is a LinkNode or if its
       // parent is a LinkNode, we update the URL, target and rel.
-      const linkNode = $isLinkNode(firstNode)
-        ? firstNode
-        : $getLinkAncestor(firstNode);
+      const linkNode = $findMatchingParent(firstNode, $isLinkNode) as LinkNode;
```
EgonBolton, line 328: This was one of the many ts problems I had to solve because of `[x: string]: any;`. In doing so, I noticed that there was already a function that did the same thing as `$getLinkAncestor`, called `$findMatchingParent`

packages/lexical/src/LexicalEditor.ts
fantactuka: So glad `[x: string]: any;` is going away 🙏

packages/lexical-website/docs/concepts/serialization.md
fantactuka: The idea was to have version control for nodes where you can safely add/change/delete properties from nodes. And version number would've helped to associate it with specific type of serialized node:

```
type SerializedParagraphNodeV1 = {
  dir: 'left-to-right' | 'right-to-left',
  version: 1
}

type SerializedParagraphNodeV2 = {
  dir: 'ltr' | 'rtl',
  version: 2,
}
```
then logic within import json would rely on version number rather than duck typing based on property values

packages/lexical/src/LexicalNode.ts
```diff
@@ -624,16 +617,32 @@ export class LexicalNode {
...
-    invariant(false, 'exportJSON: base method not extended');
-  }
-
-  static importJSON(_serializedNode: SerializedLexicalNode): LexicalNode {
-    invariant(
-      false,
-      'LexicalNode: Node %s does not implement .importJSON().',
-      this.name,
+  exportJSON() {
+    let serializedNode = JSON.parse(JSON.stringify(this.getLatest()));
```
fantactuka, line 621: Tried to run it on a doc with 5k nodes, got ~35ms vs ~6ms on main branch, it's not extremely slow, but I can imagine someone to have editor.toJSON within onChange handler. Tried to replace it with [looping through prefixed keys](https://gist.github.com/fantactuka/26dcc03c2b2a2cb43f894282c101d674) and it works better (~9ms) but it leaves more responsibility for custom nodes owners, who need to ensure their props are serializable. But we can try dev mode invariants that will warn about non-serializable values, but won't affect production users

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -314,12 +302,36 @@ function $parseSerializedNodeImpl<
...
+    delete serializedNode2.mode;
+    delete serializedNode2.direction;
+    if ($isTextNode(node)) {
+      node.setMode(serializedNode.mode);
+    }
+    if ($isElementNode(node)) {
+      delete serializedNode2.format;
+      node.setFormat(serializedNode.format);
+      if (type !== 'tablecell' && type !== 'tablerow' && type !== 'table') {
+        node.setDirection(serializedNode.direction);
```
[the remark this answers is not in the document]
EgonBolton: You mean to export the direction in those nodes right? I agree.
If that's okay with you, I'll have to modify the unit tests to include that property just like I had to do with colSpan

packages/lexical-website/docs/concepts/serialization.md
EgonBolton: Ok. I clarified again about the version property.

packages/lexical/src/LexicalNode.ts
EgonBolton: Excellent improvement, thank you very much!

> Tried to replace it with [looping through prefixed keys](https://gist.github.com/fantactuka/26dcc03c2b2a2cb43f894282c101d674) and it works better (~9ms) but it leaves more responsibility for custom nodes owners, who need to ensure their props are serializable or provide own import/export overrides.

If I'm not misunderstanding, that's true for both my implementation and the variation you've made. The purpose of this PR in principle is that it is only necessary to define serialization on non-serializable nodes.

Actually, I believe that automatic serialization could always be achieved by using something like [devalue](https://github.com/Rich-Harris/devalue
). However, I'm not sure what the implications of this are on performance or memory. Taking `caption` for example, I don't know if it would be good to serialize all editor properties like `pendingEditorState` and such. Maybe we can open a separate thread to discuss this separately?

acywatson: > Actually, I believe that automatic serialization could always be achieved by using something like [devalue](https://github.com/Rich-Harris/devalue). However, I'm not sure what the implications of this are on performance or memory.

1) Automatic serialization is not necessarily what we want - more like flexible serialization with sane defaults. 

2) Lexical core has no dependencies on any other library - we don't want to change that. Control over performance is one of the main reasons that was done.

> Taking caption for example, I don't know if it would be good to serialize all editor properties like pendingEditorState and such. 

I don't think it would be good. In the best case it's unnecessary bloat in editor states, in the worst case it encourages depending on implementation details that might change.

packages/lexical/src/__tests__/unit/LexicalEditorState.test.ts
acywatson: I don't see any issue with this - I had actually forgotten that toEqual recursively compares equality.

packages/lexical-link/src/index.ts
acywatson: I know this isn't within the overall intent of your PR, but it seems like the right thing to do here is either use a type argument to assert the type of the return value.

EgonBolton: Yes, I tried for a while and since I couldn't get it I had left it as a comment to-do. I deleted the comment when you told me. I didn't have time to open another issue.

packages/lexical/src/LexicalNode.ts
EgonBolton: Yeah, that's what I thought @acywatson 

@fantactuka, did you get it to work with your snippet?

even setting `direction`, `indent` and `version` there are some tests that break when I implement it.

fantactuka: > @fantactuka, did you get it to work with your snippet?

No, I didn't get to the point to run testes over it, can take a look closer tomorrow evening

We could limit auto-serialization to primitives (numbers, strings, bools, nulls) and nested editors (by just calling .toJSON on it). Everything else should require custom import/export. It should be possible to lint (class with no import/exportJSON and non-primitive props starting with "__"). I might be wrong, but I can only recall MarkNode and its sub-classes that use arrays/maps as serializable props

potatowagon:

PR #3943, opened 2023-02-19
Trigger `CAN_REDO_COMMAND` as false when `redoStack` goes down to 0

desclapez:
This fixes an issue in the lexical history plugin that does not notify that the redo action is unavailable when the user performs an update.

There is no bug reported that I can find, so here is the steps to reproduce the issue:

```
Steps to reproduce:
- Go to https://playground.lexical.dev/
- Make a change
- Undo
- Make another change

Expected: the redo button is disabled

Actual: the redo button is enabled
```

I've added a unit test to test a full sequence (probably more of that is needed for this PR but there were not enough tests for this plugin I believe).

Let me know if you need to open an issue and I will do it.


acywatson:
This looks right to me. @fantactuka?

fantactuka (approved):
👍

PR #3945, opened 2023-02-20
Clone objects from the CSS cache before mutating them

### LINKED ISSUE #3944: Bug: When setting styles on a collapsed selection, the styles object is not copied.
Lexical version: 59990233d496f50a365e68008df4bba2ade6aaf4 (0.8.0 branch)

## Steps To Reproduce

```ts
$getSelectionStyleValueForProperty($getSelection(), 'color', '');

$patchStyleText($getSelection(), {color: 'red'});

$getSelection().setStyle('');

$getSelectionStyleValueForProperty($getSelection(), 'color', '');
```

## The current behavior

The final call to `$getSelectionStyleValueForProperty` returns `'red'`.

## The expected behavior

It should return an empty string.

## Analysis

The problem is that the initial call to `$getSelectionStyleValueForProperty` will populate the CSS style cache (`CSS_TO_STYLES`) with an empty string pointing to an empty object then the subsequent call to `$patchStyleText`, when it operates on a collapsed selection, will directly modify that object without first cloning it.

birtles:
Fixes #3944.

This patch also fixes a bug in $getSelectionStyleValueForProperty that
was papered over by this bug--namely that when it is called on
a collapsed selection we should report the style on the selection
regardless of whether it agrees with the current node style.

### DIFF
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -247,11 +247,13 @@ export function $addNodeStyle(node: TextNode): void {
   CSS_TO_STYLES.set(CSSText, styles);
 }
 
-function $patchNodeStyle(
-  node: TextNode,
+function $patchStyle(
+  target: TextNode | RangeSelection,
   patch: Record<string, string | null>,
 ): void {
-  const prevStyles = getStyleObjectFromCSS(node.getStyle());
+  const prevStyles = getStyleObjectFromCSS(
+    'getStyle' in target ? target.getStyle() : target.style,
+  );
   const newStyles = Object.entries(patch).reduce<Record<string, string>>(
     (styles, [key, value]) => {
       if (value === null) {
@@ -264,7 +266,7 @@ function $patchNodeStyle(
     {...prevStyles} || {},
   );
   const newCSSText = getCSSFromStyleObject(newStyles);
-  node.setStyle(newCSSText);
+  target.setStyle(newCSSText);
   CSS_TO_STYLES.set(newCSSText, newStyles);
 }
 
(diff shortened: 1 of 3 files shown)


birtles:
After posting this, I was thinking I might try to drop the change to `$isTextNode` and try adding a `Readonly<>` annotation to the result of `getStyleObjectFromCSS` to catch bugs like this earlier.

birtles:
This is ready for review now.

I reverted the change to `$isTextNode` since, if we change the `$isXXXNode` helpers to take an `unknown` input, we should probably change all of them at the same time.

Also, I _haven't_ updated the return type of `getStyleObjectFromCSS` to `Readonly<>` yet because it turns out that function is exported so doing so is a potentially breaking change which should probably be made separately. I'll file a separate issue for that.

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -247,11 +247,13 @@ export function $addNodeStyle(node: TextNode): void {
   CSS_TO_STYLES.set(CSSText, styles);
 }
 
-function $patchNodeStyle(
-  node: TextNode,
+function $patchStyle(
```
thegreatercurve, line 250: I know it's probably not DRY, but to avoid shipping this as a breaking change, can we use the original method of `$patchNodeStyle`, but also add `$patchSelectionStyle`?

thegreatercurve (commented):
LGTM, but can we make one change to avoid having to ship this as a breaking change.

packages/lexical-selection/src/lexical-node.ts
birtles: Sure, can do. But just to check, why would this be a breaking change? Is $patchNodeStyle exported somewhere? (Sorry, on a phone now where it's hard to check)

thegreatercurve: Sorry @birtles, my mistake, I'm confusing this with another similarly named method we do export.

thegreatercurve (approved):

Issue #3947, opened 2023-02-21
Question: Why the method of getElementByKeyOrThrow was not exposed in lexical?


Corle-He:
Lexical version: latest ([v0.7.9](https://github.com/facebook/lexical/releases/tag/v0.7.9))

In the process of using lexical, I've noticed that there's a good method to getting an element by `NodeKey`

- [getElementByKeyOrThrow](https://github.com/facebook/lexical/blob/f8a0ba22a169b6e2b9e6b973a02b6a08318a4d53/packages/lexical/src/LexicalUtils.ts#L1176-L1188)

But , I am very curious why this method was not exposed in [lexical](https://github.com/facebook/lexical/blob/f8a0ba22a169b6e2b9e6b973a02b6a08318a4d53/packages/lexical/src/index.ts#L130-L148
)

So, just leave tow question:
1. Is there any concern?
2. Is it possible for lexical lexical to expose this method?

And it seems that there are other ways to get the element through Node

Hey-hy:
This problem can be solved in [PR](https://github.com/facebook/lexical/pull/3948)

thegreatercurve:
You can just use the below method to get a DOM element which is exposed directly on the editor instance:  
```
const domElement = editor.getElementByKey(nodeKey);
```

PR #3948, opened 2023-02-21
Export getElementByKeyOrThrow to lexical

Corle-He:
To fix: 

In the process of using lexical, I've noticed that there's a good method to getting an element by NodeKey

- [getElementByKeyOrThrow](https://github.com/facebook/lexical/blob/f8a0ba22a169b6e2b9e6b973a02b6a08318a4d53/packages/lexical/src/LexicalUtils.ts#L1176-L1188)

But , I am very curious why this method was not exposed in [lexical](https://github.com/facebook/lexical/blob/f8a0ba22a169b6e2b9e6b973a02b6a08318a4d53/packages/lexical/src/index.ts#L130-L148)

And it seems that there are other ways to get the element through Node

So i did this change

Closes: #3947

### DIFF
--- a/packages/lexical/src/index.ts
+++ b/packages/lexical/src/index.ts
@@ -145,6 +145,7 @@ export {
   $setSelection,
   $splitNode,
   isSelectionWithinEditor,
+  getElementByKeyOrThrow
 } from './LexicalUtils';
 export {$isDecoratorNode, DecoratorNode} from './nodes/LexicalDecoratorNode';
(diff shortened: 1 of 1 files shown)


fantactuka:
We already have editor.getElementByKey so I'd say if we want this to be exposed it should be moved into editor class instead

thegreatercurve:
Closing as added the same answer as @fantactuka in the Issue.

PR #3952, opened 2023-02-22
Check if parent node is root in order to remove it in trimTextContentFromAnchor

### LINKED ISSUE #3650: Bug: MaxLengthPlugin throws error when pasting large content
When MaxLength is enabled in Playground, clear the editor and paste large text into it, you'll find an error appears in console and the paste was rejected.

The error is: https://lexical.dev/docs/error?code=52

Lexical version: 0.7.5

## Steps To Reproduce

1. Clear the Playground Editor
2. Enable MaxLength Plugin
3. Open Console in DevTools
4. Copy Large Content Into Clipboard. I copied this entire page content (https://en.wikipedia.org/wiki/Project_Gutenberg)
5. Paste into Playground Editor
6. Check Console

## The current behavior

`main.7743b947.js:21 Uncaught Error: Minified Lexical error #52; visit https://lexical.dev/docs/error?code=52 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.`

image

## The expected behavior
The MaxLengthPlugin should not throw error.

theonastos:
Fixes #3650

I have added an extra condition in trimTextContentFromAnchor to only remove parent node if it is not the `RootNode`

### DIFF
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -23,6 +23,7 @@ import {
   $isElementNode,
   $isRangeSelection,
   $isTextNode,
+  $isRootNode,
   DEPRECATED_$isGridSelection,
 } from 'lexical';
 
@@ -183,7 +184,7 @@ export function trimTextContentFromAnchor(
     if (!$isTextNode(currentNode) || remaining >= textNodeSize) {
       const parent = currentNode.getParent();
       currentNode.remove();
-      if (parent != null && parent.getChildrenSize() === 0) {
+      if (parent != null && parent.getChildrenSize() === 0 && !$isRootNode(parent)) {
         parent.remove();
       }
       remaining -= textNodeSize + additionalElementWhitespace;


zurfyx:
Thank you! While I think the provided code is correct (as it can catch some corner case) having an empty RootNode often means the editor has not been initialized. Certainly, this shouldn't be a state that we land in after pasting content on an initialized editor. Do you mind digging deeper into why the MaxLength plugin ends up deleting all the children in the RootNode including the initial empty paragraph?

thegreatercurve:
Still waiting for above comment, but merging this for the moment because this fixes an internal error.

thegreatercurve (approved):

theonastos:
@zurfyx 
> Thank you! While I think the provided code is correct (as it can catch some corner case) having an empty RootNode often means the editor has not been initialized. Certainly, this shouldn't be a state that we land in after pasting content on an initialized editor. Do you mind digging deeper into why the MaxLength plugin ends up deleting all the children in the RootNode including the initial empty paragraph?

Yeap I get it, this has been bothering me, I'll have a look.

PR #3966, opened 2023-02-23
Fix getNodes of RangeSelection

kosei28:
This fixes a problem in RangeSelection's getNodes method, which incorrectly retrieves the nodes when the anchor is after the focus.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -601,25 +601,27 @@ export class RangeSelection implements BaseSelection {
     const anchor = this.anchor;
     const focus = this.focus;
     const isBefore = anchor.isBefore(focus);
-    let firstNode = isBefore ? anchor.getNode() : focus.getNode();
-    let lastNode = isBefore ? focus.getNode() : anchor.getNode();
-    const firstOffset = isBefore ? anchor.offset : focus.offset;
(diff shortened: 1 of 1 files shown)


acywatson:
Thanks for the PR - can you make an issue or a failing test to demonstrate the problem? It's easier to review the change with some context around the issue you're trying to address.

kosei28:
@acywatson 
I made the issue. #3973

packages/lexical/src/LexicalSelection.ts
```diff
@@ -600,25 +600,31 @@ export class RangeSelection implements BaseSelection {
...
     const anchor = this.anchor;
     const focus = this.focus;
+    const isBefore = anchor.isBefore(focus);
     let firstNode = anchor.getNode();
     let lastNode = focus.getNode();
+    let firstOffset = anchor.offset;
+    let lastOffset = focus.offset;
+
+    if (!isBefore) {
+      [firstNode, lastNode] = [lastNode, firstNode];
```
acywatson, line 610: nit: the typical pattern we use here is ternary assignment conditioned on isBefore. I think it's a bit clearer and consistent with what we do elsewhere.

thegreatercurve:
LGTM. Will merge if @acywatson is happy as well?

PR #3970, opened 2023-02-24
Make one previous node be selected when deleting a block

Karibash:
When a block is deleted, one previous node is selected.
This will improve the writing experience.

https://user-images.githubusercontent.com/40270352/221128816-4a476298-d177-46c2-a122-dc917d90d34b.mov


thegreatercurve (approved):
Seems fine to me! Thanks

Issue #3940, opened 2023-02-19
Feature: Export the matcher of AutoLinkPlugin


Karibash:
Currently, in order to implement an automatic linking function for URLs and mail addresses, matchers must define their own matchers using AutoLinkNode.
It would be useful to be able to use the URL and mail address matchers implemented in Playground as is.

```ts
import { AutoLinkPlugin, urlLinkMatcher, emailLinkMatcher } from '@lexical/react/LexicalAutoLinkPlugin';

const matchers = [
  urlLinkMatcher,
  emailLinkMatcher,
];

export default function LexicalAutoLinkPlugin(): JSX.Element {
  return <AutoLinkPlugin matchers={matchers} />;
}
```

acywatson:
> AutoLinkNode.

Maybe instead of exporting it we can make it the default in the plugin. The issue is then that we have to officially support, but maybe that’s ok.

PR #3972, opened 2023-02-24
Add a function to create a LinkMatcher based on a RegExp

Karibash:
close: #3940

### DIFF
--- a/packages/lexical-playground/src/plugins/AutoLinkPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/AutoLinkPlugin/index.tsx
@@ -8,12 +8,24 @@
 
 import {
   AutoLinkPlugin,
-  emailLinkMatcher,
-  urlLinkMatcher,
+  createLinkMatcherWithRegExp,
 } from '@lexical/react/LexicalAutoLinkPlugin';
 import * as React from 'react';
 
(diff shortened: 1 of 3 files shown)


acywatson (approved):

fantactuka:
Should we export factory function instead that will accept reg exp? There's a lot of url/email patterns and apps are often very sensitive about it

acywatson:
> Should we export factory function instead that will accept reg exp? There's a lot of url/email patterns and apps are often very sensitive about it

Yea, this might be a good idea. I was hesitant about this change for the same reason. We are getting into the business of maintaining reg ex patterns that are notoriously fickle. At the same time, it's nice for users to have something that usually works by default, and they can always eject and do their own pattern.

Factory function seems like a good idea, regardless

Karibash:
Is it like exporting a factory function like the one below?
```ts
function createLinkMatcherWithRegExp(
  regExp: RegExp,
  urlTransformer: (text: string) => string = (text) => text,
) {
  return (text: string) => {
    const match = regExp.exec(text);
    if (match === null) return null;
    return {
      index: match.index,
      length: match[0].length,
      text: match[0],
      url: urlTransformer(text),
    };
  };
}
```

Karibash:
@fantactuka @acywatson 
I have fixed it, please review.

acywatson (approved):
Yea, this looks right, thanks!

Issue #3085, opened 2022-09-29
Feature: Make collaboration plugin support IndexeddbPersistence


zephyraft:
I'm trying to add IndexeddbPersistence to the providerFactory of the collaboration plugin, so that the offline content can be saved in the local browser.
When I try something like the following code, it works fine without switching the namespace of LexicalComposer.
```
<LexicalCollaborationPlugin
    id={id}
    username={username}
    providerFactory={(id, map) =>
        let doc = map.get(id);
        if (doc === undefined) {
            doc = new Y.Doc();
            map.set(id, doc);
        } else {
            doc.load();
        }

        const _doc = doc;

        return new (class extends WebsocketProvider {
            private _indexeddbProvider: IndexeddbPersistence;

            constructor() {
                super(websocketAddr, id, _doc, {
                    connect: false,
                });
                this._indexeddbProvider = this.bindIndexeddb(id, _doc);
            }

            private bindIndexeddb(id: string, doc: Y.Doc) {
                const indexeddbPersistence = new IndexeddbPersistence(id, doc);
                indexeddbPersistence.whenSynced.then((persistence) => {
                    console.debug(
                    'loaded data from indexed db',
                    this.roomname
                    );
                });
                return indexeddbPersistence;
            }

            connect() {
                super.connect();
                this._indexeddbProvider = this.bindIndexeddb(this.roomname, this.doc);
            }

            disconnect() {
                super.disconnect();
                this._indexeddbProvider.destroy();
            }
        })();
    }
    shouldBootstrap={!skipCollaborationInit}
/>
```

However, **when the websocket connection fails, the document cannot be switched normally when the namespace is modified**.

https://github.com/facebook/lexical/blob/f0de0f57dd5f7f5ec04caf1449259cbc6230d060/packages/lexical-react/src/shared/useYjsCollaboration.tsx#L56

the doc state in `useYjsCollaboration` will only setDoc when the WebsocketProvider triggers the reload event, so when the websocket connection fails, the document will not switch properly.

Finally I tried adding the following code to `useYjsCollaboration` to make it work.
```
useEffect(() => {
    const newDoc = docMap.get(id);
    if (newDoc) {
        setDoc(newDoc);
    }
}, [id, docMap]);
```

Hope the collaboration plugin will add support for indexeddb or other providers.

fantactuka:
Could you explain what do you mean with "with/without switching the namespace"?

[...]

PR #3979, opened 2023-02-25
Cleanup excessive use of WebsocketProvider for types

fantactuka:
We always use `WebsocketProvider` as an example, but are not tied to this specific implementation and it also creates confusion for users as they assume collab provider works only with WS (#3085)


thegreatercurve (approved):

thegreatercurve:
@fantactuka You might need to delete and re-install the packages

Issue #3906, opened 2023-02-14
Bug: Collapsible Section toggle event doesn't update the node's state


ivailop7:
Lexical version: 0.8

## Steps To Reproduce

1. Add a Collapsible Section (expanded by default)
2. Check the value of "this.__open"
3. Click on the arrow to collapse the section
4. Check the value of "this.__open"

## The current behavior
The value of this.__open gets updated only when doing a shortcut (Ctrl + Enter) that triggers a command

## The expected behavior
this.__open to update respectively on toggling

## Fix
I've mentioned this in the discord chat. I'm happy to work on this. I need a pointer from someone from the core team to tell me where to put the `element.addEventListener("toggle", <method>, true)`. I tried it in a mutationListener to trigger a "node.setOpen", but I get an error about invalid activeEditor state, despite operating in editor.read() when I trigger the setOpen. I've also done it via window - `window.addEventListener("toggle", <method>, true)` in a useEffect in the plugin, but have to do a selection to get the node, which makes me feel there must be a simpler way.

What's the right place to add this logic in, maybe a RootListener? The fix is small, just unsure on what's the endorsed approach is. Direction would be appreciated!

PR #3983, opened 2023-02-26
Listen to Collapsible Section toggle event

ivailop7:
Fixes: #3906 

Before:
before

After:
after

Need assistance, to understand what's wrong with the definition of the useEffect that triggers this constant re-render until the first click. Secondly, is listening on window the right approach here or should I be using a rootListener? Thanks.

### DIFF
--- a/packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleContainerNode.ts
+++ b/packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleContainerNode.ts
@@ -91,7 +91,7 @@ export class CollapsibleContainerNode extends ElementNode {
 
   exportDOM(): DOMExportOutput {
     const element = document.createElement('details');
-    element.open = this.__open;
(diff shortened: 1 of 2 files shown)


fantactuka:
Checking it on every click is a bit too overwhelming, instead we could utilize toggle event that will ensure DOM and node property are in sync:

```
    const dom = document.createElement('details');
    dom.classList.add('Collapsible__container');
    dom.open = this.__open;
    dom.addEventListener('toggle', () => {
      const open = editor.getEditorState().read(() => this.getOpen());
      if (open !== dom.open) {
        editor.update(() => this.toggleOpen());
      }
    });
    return dom;
  }
```

That made me notice we have issue in `getOpen` method and it should be:

```
  getOpen(): boolean {
    return this.getLatest().__open;
  }
```

ivailop7:
Thank you @fantactuka , much better approach than mine. Incorporated the comments.

fantactuka (approved):

Issue #3890, opened 2023-02-12
Bug: Selecting color on the color picker creates multiple entries on the undo stack


TheSwilliam:
Lexical version: 0.8.0

## Steps To Reproduce

1. Select text
2. Expand the font color color picker and drag the handle around to adjust color

## The current behavior
Currently the color gets added to the undo stack on the mouse move event.

https://user-images.githubusercontent.com/69976884/218330664-261e5d2b-98ec-4faa-aa1b-0a3a2e3e5d9f.mov

## The expected behavior
The color should only get added to the undo stack when the mouse up event fires on the color picker. 

Figma has this working currently 

https://user-images.githubusercontent.com/69976884/218330901-f7ba4591-536c-4509-a156-0aa9a740f94c.mov

MohitBansal321:
Hey, I didn't see any mouse move event in it @TheSwilliam

KhalfounMehdi:
I will work on this one

PR #3984, opened 2023-02-27
Bugfix: ColorPicker color changing adds multiple entries in undo stack

KhalfounMehdi:
The bug is described in the following issue : #3890


thegreatercurve:
@KhalfounMehdi This generally looks fine! Would you mind just fixing the few lint errors?

KhalfounMehdi:
@thegreatercurve  actually i need to fix an bug that I noticed in my fix : 
- after I call the editor.update , the merge action returned is `HISTORY_MERGE` and it should be `HISTORY_PUSH`
I'll finish with the bug , I will fix the lining errors nad push

packages/lexical-playground/src/ui/ColorPicker.tsx
```diff
@@ -99,6 +101,11 @@ export default function ColorPicker({
     setSelfColor(newColor);
     setInputColor(newColor.hex);
   };
+  const HandleDropDownClose = React.useCallback(() => {
```
thegreatercurve, line 104:
```suggestion
  const HandleDropDownClose = useCallback(() => {
```

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -199,7 +201,22 @@ export default function DropDown({
       };
     }
   }, [dropDownRef, buttonRef, showDropDown, stopCloseOnClickSelf]);
+  function usePrevious(value: any) {
```
fantactuka, line 204: Unless we want this to be common util, you can just use ref without hook

zurfyx (changes requested):
Can we please revise the behavior of this PR and consider replicating the mouse up behavior from Figma (suggested in https://github.com/facebook/lexical/issues/3890)? The ColorPicker doesn't seem to be working correctly and also breaks the same time you open the picker

https://user-images.githubusercontent.com/193447/233785218-040de80e-ea65-4aa3-880b-c04ab21f7509.mov

thegreatercurve:
Closing due to PR becoming stale

PR #3985, opened 2023-02-27
[0.9] Hotfix - Remove $Wrapnodes

EgonBolton:
Now that https://github.com/facebook/lexical/pull/3821 is done, `$wrapNodes` should be removed.

The new function, `$setBlocksType`, is no longer just an improvement in simplicity. `$wrapNodes` contains many bugs when using 

1. in nested lists, where an item has multiple indentations from its parent item (see case 2 example at https://github.com/facebook/lexical/issues/3776) and 
2. in nested lists, when the list is partially selected, and there is a part outside of the selected list as well.

This PR renames `setBlocksType_experimental` to `setBlocksType`.

I know this was planned for 8.0. I don't know if you want to leave it for 9.0 now.


acywatson:
Should we deprecate before we remove? We can do the name changes in 0.9 and the removal in 0.10?

EgonBolton:
Yes of course. Just be aware that `$wrapNodes` can easily throw errors.

PR #3989, opened 2023-02-27
extend type for SerializedElementNode

strdr4605:
There can be use cases when you need to create an object of the type `SerializedEditorState`

```js
ort const serializedEditorState: SerializedEditorState = {
  root: {
    children: [
      {
        children: [{ text: 'test', type: 'text', version: 1 }],
        format: '',
        type: 'paragraph',
        version: 1,
      },
    ],
    direction: 'ltr',
    format: '',
    indent: 0,
    type: 'root',
    version: 1,
  },
};
```

```
Type '{ children: { text: string; type: string; version: number; }[]; format: string; type: string; version: number; }' is not assignable to type 'SerializedLexicalNode'.
  Object literal may only specify known properties, and 'children' does not exist in type 'SerializedLexicalNode'
```

---

An alternative solution is to always create `serializedEditorState` with an instance of the lexical editor.

But still looks like that type for `SerializedElementNode.children` should extend `SerializedLexicalNode`.

### DIFF
--- a/packages/lexical/src/nodes/LexicalElementNode.ts
+++ b/packages/lexical/src/nodes/LexicalElementNode.ts
@@ -39,7 +39,7 @@ import {
 
 export type SerializedElementNode = Spread<
   {
-    children: Array<SerializedLexicalNode>;
+    children: Array<SerializedLexicalNode & Record<string, unknown>>;
     direction: 'ltr' | 'rtl' | null;
     format: ElementFormatType;
     indent: number;


thegreatercurve:
What about using an optional generic instead? We at least retain some type safety then.
```ts
export type SerializedElementNode<
  T extends SerializedLexicalNode = SerializedLexicalNode,
> = Spread<
  {
    children: Array<T>;
```

You then pass the generic up through the various levels, until you specify the below in your serialized editor state:
```ts
const serializedEditorState: SerializedEditorState<
  SerializedElementNode<SerializedTextNode>
> = {
  root: {
```

thegreatercurve:
Actually, I've got a working branch, let me just open another PR: #3990

strdr4605:
I initially played with the same solution from #3990, but thought that 
```js
const serializedEditorState: SerializedEditorState<
  SerializedElementNode<SerializedTextNode | SerializedMentionNode | ...>
> 
```
will be too much boilerplate to write.😅
But I think we will be ok with #3990 😊

thegreatercurve:
> I initially played with the same solution from #3990, but thought that
> 
> ```js
> const serializedEditorState: SerializedEditorState<
>   SerializedElementNode<SerializedTextNode | SerializedMentionNode | ...>
> > 
> ```
> 
> will be too much boilerplate to write.😅 But I think we will be ok with #3990 😊

Let's see how we get on 😅. It's easier to start strict and then loosen the rules later, compared to the other way around, even if it requires a bit more boilerplate.

PR #3991, opened 2023-02-27
Doc: initial execution of listener function

gdelmas:
Updating the documentation to state that the root listener function is executed directly upon registration (see [implementation](https://github.com/facebook/lexical/blob/924b729deaab492736badd72089ef9831f00568e/packages/lexical/src/LexicalEditor.ts#L610)), as this  is unexpected behaviour for event listeners.


fantactuka:
I think it'd be better to add extra code into example to make it clear

zurfyx (approved):

acywatson:
> I think it'd be better to add extra code into example to make it clear (with rootElement.addListener, prevRootElement.removeListener)

I'll make a separate PR for this.

Issue #3791, opened 2023-01-28
Bug: Selection changes incorrectly when focus is before anchor across two text nodes with different font colors when setting a font color


TheSwilliam:
Lexical version: 0.7.5

## Steps To Reproduce

1. Create two text nodes. 
2. Set the color of the second node to a color
3. Then select right to left and select the entire second node and half of the first node.
4. then select a new color on the color picker

## The current behavior

https://user-images.githubusercontent.com/69976884/215290831-23e783b6-3881-49f3-90cc-597744c758de.mov

## The expected behavior
The selection should remain consistent and not change when the color changes.

Deekshithrathod:
@thegreatercurve Hey I'm new to open source, and would love to work on this issue. I hope you can assign it to me.

Ryczko:
Since the bug is still not resolved I will try to solve it

Deekshithrathod:
sure 👍

PR #4004, opened 2023-02-28
Fix selection on color change

Ryczko:
This condition will be more appropriate for the case where we have a single node selected. 
It fixes bug described here: https://github.com/facebook/lexical/issues/3791


thegreatercurve (approved):
Lovely work! Thanks @Ryczko

Issue #3999, opened 2023-02-28
Bug: list-style-position:inside problems


GermanJablo:
## Background on `list-style-position:inside`

- Added in ([#2334](https://github.com/facebook/lexical/pull/2334)) so that bullets follow text alignment.
- Removed in ([#3753](https://github.com/facebook/lexical/pull/3753)) because of a bug in Safari with the cursor ([Issue #3752](https://github.com/facebook/lexical/pull/3752))
- Added again in ([#3814](https://github.com/facebook/lexical/pull/3814)) so that bullets follow text alignment ([Issue #3807](https://github.com/facebook/lexical/issues/3807)). Safari bug reported to webkit.org

## Solution
The top answer of [this StackOverflow thread](https://stackoverflow.com/questions/28977320/how-do-i-get-the-bullet-points-of-a-ul-to-center-with-the-text) offers 2 solutions for bullets to follow text alignment. The first is to use `list-style-position: inside`, which as already seen is buggy in Safari.

The second solution is to align `ul` instead of `li`. This solves the alignment problem and the cursor bug in Safari.

### Advantages:
- Fixes the bug in Safari.
- Lists do not require theming or CSS to work well.
- As you can see running the snippets of the SO thread, the second solution ensures that the bullets are aligned. Not so with `list-style-position: inside`
- By using the default CSS properties, other possible bugs are avoided. For example, this is what happens to me when I try to use my `PlaceholderPlugin` with lists:
![image](https://user-images.githubusercontent.com/43938777/221925483-26efb718-3eb8-44d9-ba3e-eeb682de511b.png)

### Trade-off?
- There couldn't be `li` with different alignments within the same list (which seems reasonable to me).

____

Lexical version: 0.8.1

## The current behavior
- the bullets follow the alignment of the `li` BUT are not aligned with each other.
- `list-style-position: inside` is required and causes unwanted effects, like the one shown with the placeholder.
- The cursor appears to the left of the bullet in Safari.

## The expected behavior
- the bullets must follow the alignment of the text AND be aligned with each other.
- `list-style-position: inside` should not be required as it causes unwanted effects, like the one shown with the placeholder.
- The cursor appears to the right of the bullet in Safari.

ivailop7:
Closing for now

PR #4025, opened 2023-03-03
Change default indent to 40px

EgonBolton:
40px would be a better default because that's what browsers use for lists.
Indentations in paragraphs, headings, etc, would match that of the lists, and converting the nodes would not generate a strange displacement.

https://www.loom.com/share/6443d73269d04769bd88c00c8f1b973b

You may not see a perfect match with the list indents in the playground, because it's using some styles like `list-style-position: inside`.
Issue covered here: https://github.com/facebook/lexical/issues/3999


thegreatercurve:
@EgonBolton Thanks for the contribution! I'm on board with this change, even if we might initially encounter some issues with nested lists on smaller view ports. I think aiming for browser consistency though is a decent aim though.

Mind updating the failing E2Es and I can merge?

EgonBolton:
Cool! This is done :)

thegreatercurve (approved):

thegreatercurve:
@EgonBolton Sorry, I know this is annoying,  but do you mind fixing the merge conflicts and then I can merge?

Issue #3227, opened 2022-10-20
ListNode dangerously executes update logic on createDOM


zurfyx:
`createDOM` was designed to render on the DOM. It can happen synchronously and asynchronously but ultimately it translates node data into DOM.

List leverages createDOM to do a last-minute update logic, in particular in `updateChildrenListItemValue`

```
  createDOM(config: EditorConfig): HTMLElement {
    const element = document.createElement('li');
    const parent = this.getParent();

    if ($isListNode(parent)) {
      updateChildrenListItemValue(parent);
      updateListItemChecked(element, this, null, parent);
    }
    element.value = this.__value;
    $setListItemThemeClassNames(element, config.theme, this);

    return element;
  }
```

```
export function updateChildrenListItemValue(
  list: ListNode,
  children?: Array<ListItemNode>,
): void {
  (children || list.getChildren()).forEach((child: ListItemNode) => {
    const prevValue = child.getValue();
    const nextValue = $getListItemValue(child);

    if (prevValue !== nextValue) {
      child.setValue(nextValue);
    }
  });
}
```

This is dangerous because the error-recovery mechanism only applies to the update and transforms, not the reconciliation; and this can translate to a crashed UI and potentially bad collab updates.

trueadm:
I think this has been the cause of some complex collab bugs internally after looking at a recent error report. I think we should prioritise this.

acywatson:
> This is dangerous because the error-recovery mechanism only applies to the update and transforms, not the reconciliation

As discussed offline, I don't think is actually true. We [catch and attempt to recover from errors during reconciliation](https://github.com/facebook/lexical/blob/230903f0f536b8d1a84f2ca43dbb390e9a7f9063/packages/lexical/src/LexicalUpdates.ts#L470), AFAIK. That's not to say this logic is correct - just want to avoid red herrings here.

PR #4030, opened 2023-03-04
Fix list deserialization

acywatson:
This one was a bit gnarly and probably requires a deeper fix at some point. We've had reports of people encountering issues when deserializing lists, and it all goes back to this issue: https://github.com/facebook/lexical/issues/3227

What we're doing here is basically trying to update Lexical nodes during the reconciliation phase, which typically works fine, but encounters some issues in a few cases. One of those is when reconciliation runs asynchronously, which seems to be what's happening in the case of state being initialized from a string via LexicalComposer. Basically, we "freeze" the node map after the update cycle to ensure that state can't be further updated outside of that cycle. In this case, we were freezing the map, then createDOM was running, calling these list item update functions, which tried to update the nodes. There are several interrelated issues here, including the underlying structure/function of lists, which we will ultimately probably change (as discussed elsewhere).  

The fix I'm proposing here is to introduce a method to nodes that allows them to declare a transform on themselves, rather than registering it through the register API. This allows us to keep this logic encapsulated at the node level, rather than moving it to the plugin, and it should otherwise function the same.


packages/lexical/src/LexicalUpdates.ts
```diff
@@ -132,6 +132,7 @@ export function $applyTransforms(
 
   if (transformsArr === undefined) {
     transformsArr = Array.from(registeredNode.transforms);
+    transformsArr.push(node.transform.bind(node));
```
acywatson, line 135: I should probably do this in LexicalEditor onInit instead.

packages/lexical/src/LexicalNode.ts
```diff
@@ -636,6 +636,10 @@ export class LexicalNode {
     );
   }
 
+  transform(): void {
+    return;
```
acywatson, line 640: This should probably return a function and just be called once during init and then passed the node like other transforms.

moy2010:
One deserialization error that I've had before is caused by the fact that markdown's shortcut's regex for the numeric list matches on any number, so it's possible to create numeric lists that do not start from `1.`.

acywatson:
> One deserialization error that I've had before is caused by the fact that markdown's shortcut's regex for the numeric list matches on any number, so it's possible to create numeric lists that do not start from `1.`.

Oh weird. I guess starting a list at a  number other than 1 should be possible. Might need to look into that separately.

zurfyx (approved):

PR #4031, opened 2023-03-05
Prevent creation of malformed ordered lists

moy2010:
The existing regex for ordered lists matches on any numeric number, which then triggers parsing errors when attempting to parse the generated editor state JSON.


moy2010:
There are e2e tests failing, so I assume that the behaviour of the original regex was intended. I will test tomorrow if @acywatson PR fixes this as well.

PR #4035, opened 2023-03-06
After removing a ListItemNode, merge its siblings.

Awjin:
Properly handles siblings that contain nested lists.

# Before:

https://user-images.githubusercontent.com/3211874/223054925-c7a752bd-275e-4864-b29b-0be509729d33.mov

# After:

https://user-images.githubusercontent.com/3211874/223054952-5bb5f620-489a-46bf-8959-2e996c780fd8.mov

### DIFF
--- a/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
+++ b/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
@@ -15,6 +15,7 @@ import {
   ListItemNode,
   ListNode,
 } from '../..';
+import {expectHtmlToBeEqual, html} from '../utils';
 
 const editorConfig = Object.freeze({
   namespace: '',
@@ -27,10 +28,6 @@ const editorConfig = Object.freeze({
     },
   },
 });
-
-function stripLineBreaks(str: string): string {
-  return str.replace(/\n\s+/g, '');
-}
 
 describe('LexicalListItemNode tests', () => {
   initializeUnitTest((testEnv) => {
@@ -266,9 +263,13 @@ describe('LexicalListItemNode tests', () => {
           root.append(parent);
         });
 
-        expect(testEnv.outerHTML).toBe(
-          stripLineBreaks(`
-            <div contenteditable="true" style="user-select: text; white-space: pre-wrap; word-break: break-word;" data-lexical-editor="true">
+        expectHtmlToBeEqual(
+          testEnv.outerHTML,
+          html`
+            <div
+              contenteditable="true"
+              style="user-select: text; white-space: pre-wrap; word-break: break-word;"
+              data-lexical-editor="true">
               <ul>
                 <li value="1" dir="ltr">
                   <span data-lexical-text="true">A</span>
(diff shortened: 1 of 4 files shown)


packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
```diff
@@ -237,6 +241,616 @@ describe('LexicalListItemNode tests', () => {
...
+
+          const B_listItem = new ListItemNode();
+          B_listItem.append(new TextNode('B'));
+
+          parent.append(A_listItem, x, B_listItem);
+          root.append(parent);
+        });
+
+        expect(testEnv.outerHTML).toBe(
+          stripLineBreaks(`
```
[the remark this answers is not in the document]
Awjin: Done. Should I clean up the other specs too?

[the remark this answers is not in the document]
acywatson: > Done. Should I clean up the other specs too?

If you're so inclined, sure. I've been doing them ad-hoc when I encounter them in the normal course of coding lol.

packages/lexical-list/src/__tests__/utils.ts
```diff
@@ -0,0 +1,33 @@
+/**
```
Awjin, line 1: FYI these utils are pretty similar to those in `lexical-playground/__tests__/utils`. At some point it might be worth extracting both into `lexical/src/__tests__/utils`.

acywatson: Good point. We once had designs around exposing these utils to the world, as well. Could be interesting to explore at some point.

packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
```diff
@@ -196,8 +287,24 @@ describe('LexicalListItemNode tests', () => {
           listItemNode3.replace(paragraphNode);
         });
 
-        expect(testEnv.outerHTML).toBe(
-          '<div contenteditable="true" style="user-select: text; white-space: pre-wrap; word-break: break-word;" data-lexical-editor="true"><ul><li value="1" dir="ltr"><span data-lexical-text="true">one</span></li><li value="2" dir="ltr"><span data-lexical-text="true">two</span></li></ul><p><br></p></div>',
+        expectHtmlToBeEqual(
```
acywatson, line 290: ah, we already have a util called assertHTML - I'm so sorry for the churn. I should have been clearer.

We can merge this to get the fix and then do that cleanup later.

acywatson (approved):

packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
Awjin: Sounds good, thanks. I assume we'd want to extract `assertHTML` from `lexical-playground`'s test utils, into  `lexical`'s? If so we have to migrate the playground tests to `ts` first

PR #4036, opened 2023-03-06
[0.9] Remove experimental flag from content plugins

thegreatercurve:


zurfyx:
I would hold on this one, I had a PR around I never finished but I think this module need further work. Up to you really, stamping and you decide

thegreatercurve:
> I would hold on this one, I had a PR around I never finished but I think this module need further work. Up to you really, stamping and you decide

What further work is needed? Is it new features or bug fixes? If it's the latter, I'd keep the experimental flag on it.

PR #4037, opened 2023-03-06
[0.9] Remove experimental flag for copyToClipboard

thegreatercurve:


thegreatercurve:
@zurfyx Can you sense check that this flag is okay to be removed? The function has been in use for around 6 month without any reported bugs.

zurfyx (approved):

PR #4040, opened 2023-03-06
Fix EquationNode issues on Android 

### LINKED ISSUE #1965: Bug: Unable to backspace delete decorator nodes containing text on android chrome
When attempting to backspace delete a decorator node that contains DOM text, the keyboard is instead dismissed, and the decorator node stays in the editor.
Happens on various Android keyboard apps like GBoard and Microsoft SwiftKey.

This video should demonstrate it:
https://files.catbox.moe/3v1gdi.mp4
(Sorry the filesize limit on github wouldn't allow me to upload it directly)

Lexical version: 0.3.8
Android version: 12
Android Chrome version: 100
Microsoft SwiftKey Keyboard version: 8.10.12.4
Gboard version: 11

## Steps To Reproduce

1. go to https://playground.lexical.dev/ using e.g. latest android chrome, while using GBoard or Microsoft SwiftKey
2. insert a "equation" node (content does not matter)
3. Move selection to the end of the editor.
4. Hit backspace

EDIT:
Maybe it's best to also check variants of the reproduction steps above like:
2. insert some random text
3. insert an "equation" node
4. Move the selection to the end of the editor
5. Hit backspace

Since I noticed that [83dac84](https://github.com/facebook/lexical/pull/1968/commits/83dac84899305aa3cd0b43b2caadf0235fe6b5dd) from #1968 only solves the issue when using the first set of steps. My bad.

Link to code example:

## The current behavior

Hitting backspace does not delete the decorator node, instead, the keyboard is dismissed.

## The expected behavior

Hitting backspace should delete the decorator node, and the keyboard should not be dismissed.

EDIT 2: this was previously fixed but still happens, also right now firefox android behaves differently so I edited the title.

As for the behaviour on firefox: android tries to compose in the inner text when holding backspace, deleting some text prior to the decorator node. Placing the caret right in front of the decorator node and hitting backspace works as intended.

### LINKED ISSUE #3426: Feature: Improving selection around decorator nodes

thegreatercurve:
This PR fixes a Android bug. 

Hitting backspace at the end of an `EquationNode` causes selection to be lost. After much experimentation, I found that the fix involves adding `user-select: none;` to the decorator node, like with the inline `ImageNodes`, and to change the wrapping spaces around the equation node into `img` tags instead. 

I made some simplifications to the equation node logic as well.

Closes #3426
Closes #1965

### DIFF
--- a/packages/lexical-playground/src/ui/KatexRenderer.tsx
+++ b/packages/lexical-playground/src/ui/KatexRenderer.tsx
@@ -13,11 +13,11 @@ import {useEffect, useRef} from 'react';
 export default function KatexRenderer({
   equation,
   inline,
-  onClick,
+  onDoubleClick,
 }: Readonly<{
   equation: string;
   inline: boolean;
-  onClick: () => void;
+  onDoubleClick: () => void;
 }>): JSX.Element {
   const katexElementRef = useRef(null);
 
@@ -37,18 +37,18 @@ export default function KatexRenderer({
   }, [equation, inline]);
 
   return (
-    // We use spacers either side to ensure Android doesn't try and compose from the
+    // We use an empty image tag either side to ensure Android doesn't try and compose from the
     // inner text from Katex. There didn't seem to be any other way of making this work,
     // without having a physical space.
     <>
-      <span className="spacer"> </span>
+      <img src="#" alt="" />
       <span
         role="button"
         tabIndex={-1}
-        onClick={onClick}
+        onDoubleClick={onDoubleClick}
         ref={katexElementRef}
       />
-      <span className="spacer"> </span>
+      <img src="#" alt="" />
     </>
   );
 }
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -1623,6 +1623,15 @@ button.action-button:disabled {
   letter-spacing: -2px;
 }
 
+.editor-equation {
+  cursor: default;
+  user-select: none;
+}
+
+.editor-equation.focused {
+  outline: 2px solid rgb(60, 132, 244);
+}
+
 button.item i {
   opacity: 0.6;
 }
--- a/packages/lexical-playground/src/nodes/EquationComponent.tsx
+++ b/packages/lexical-playground/src/nodes/EquationComponent.tsx
@@ -8,10 +8,13 @@
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {mergeRegister} from '@lexical/utils';
-import { $getNodeByKey,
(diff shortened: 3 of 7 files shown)


acywatson:
Interesting! Seems like the chromium bug mentioned in that issue is still open, though?

thegreatercurve:
> Interesting! Seems like the chromium bug mentioned in that issue is still open, though?

@acywatson I actually realised that the fix for this original issue regarding the double delete when backspacing was a Chromium bug for which I've already submitted a report (https://bugs.chromium.org/p/chromium/issues/detail?id=1403455)

acywatson (approved):
So anyone trying to copy this functionality or implement something similar would need to do the same, I suppose. Maybe we should document it somewhere.

thegreatercurve:
> So anyone trying to copy this functionality or implement something similar would need to do the same, I suppose. Maybe we should document it somewhere.

I've added a comment in the `createDOM` method

packages/lexical-playground/src/ui/KatexRenderer.tsx
```diff
@@ -37,18 +37,18 @@ export default function KatexRenderer({
...
     // inner text from Katex. There didn't seem to be any other way of making this work,
     // without having a physical space.
     <>
-      <span className="spacer"> </span>
+      <img src="#" alt="" />
       <span
         role="button"
         tabIndex={-1}
-        onClick={onClick}
+        onDoubleClick={onDoubleClick}
```
zurfyx, line 48: How well does this work on Android/iOS?

PR #4046, opened 2023-03-07
Table - cell merging

zurfyx:
This PR adds support for cell merging on the stable tables implementation. We store this with a simple `__colSpan` and `__rowSpan` property on the GridCellNode.

There are many follow ups from this PR, including the fact that only basic cases work for now:
- Doesn't copy merged content onto the merged cell
- Rewrite `getShape()` - indexes fail when merged cells are present
- Fix paste (it's already crashing in the current version for some reason)
- Kill GridSelection - performance and budget issues (adds cost the Lexical Core plain text users that don't use tables)

Cell merging

https://user-images.githubusercontent.com/193447/223561223-16d84087-a07b-4fd3-afa6-f9ba2029e7df.mov

_(yes, the menu visibility needs a fix but we're considering dropping it for a better a11y alternative)_

Selection is guaranteed to be rectangular

https://user-images.githubusercontent.com/193447/223561217-3f1dc72d-1404-433f-bca3-5ff348ed98e3.mov

More complicated selection

https://user-images.githubusercontent.com/193447/223561222-a33aae54-14b5-4518-ac0c-8b73d2de834a.mov

### DIFF
--- a/packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
@@ -29,17 +29,74 @@ import {
   $getRoot,
   $getSelection,
   $isRangeSelection,
+  DEPRECATED_$isGridCellNode,
   DEPRECATED_$isGridSelection,
+  GridSelection,
 } from 'lexical';
 import * as React from 'react';
 import {ReactPortal, useCallback, useEffect, useRef, useState} from 'react';
 import {createPortal} from 'react-dom';
+import invariant from 'shared/invariant';
+
+function computeSelectionCount(selection: GridSelection): {
+  columns: number;
+  rows: number;
+} {
+  const selectionShape = selection.getShape();
+  return {
+    columns: selectionShape.toX - selectionShape.fromX + 1,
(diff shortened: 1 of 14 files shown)


packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -29,24 +29,82 @@ import {
...
+    currentRows.every((v) => v === currentRows[0])
+  );
+}
 
 type TableCellActionMenuProps = Readonly<{
   contextRef: {current: null | HTMLElement};
   onClose: () => void;
   setIsMenuOpen: (isOpen: boolean) => void;
   tableCellNode: TableCellNode;
+  cellMerge__EXPERIMENTAL: boolean;
```
thegreatercurve, line 99: We probably don't need to mark this as experimental if this is just in the playground. We'll just end up forgetting about it anyway
```suggestion
  cellMerge boolean;
```

thegreatercurve (approved):
This looks good to me but @tylerjbainbridge should probably have a look. @zurfyx Maybe add a couple of test cases for showing what happens when you insert/delete a column or row which contains or insersects one with a merged cell.

Issue #2845, opened 2022-08-17
Improve Documentation


acywatson:
Generic Issue for Documentation Improvement.

- [ ] Setup autogenerated API Docs
- [ ] Lexical lifecycle overview
- [ ] Example: how to generate HTML from Lexical EditorState
- [ ] Example: how to use Commands/Transformers to disable core functionality
- [ ] Example: how to set initial editor state
- [ ] Example: how to set up collaborative editing

Others?

fantactuka:
I'd add an example with initializing, updating, and persisting state (this should be most common use case)

DaniGuardiola:
Ideally I'd like to see JSdoc comments in all exported objects, functions, etc. Using `@example` and adding links to documentation is great too for IDE experience.

signalwerk:
I would like to see for [commands](https://lexical.dev/docs/concepts/commands) a description of the different priority levels. As a newcomer, it might avoid some confusion. (see #2868 as an example)

signalwerk:
I found [this file](https://github.com/facebook/lexical/blob/main/examples/decorators.md) that explains the idea and usage of decorator nodes way better than the [documentation](https://lexical.dev/docs/concepts/nodes). Is there a reason for not having this file integrated in the documentation? Or could I try to merge them with a PR, so we see the information on the documentation?

acywatson:
> I found [this file](https://github.com/facebook/lexical/blob/main/examples/decorators.md) that explains the idea and usage of decorator nodes way better than the [documentation](https://lexical.dev/docs/concepts/nodes). Is there a reason for not having this file integrated in the documentation? Or could I try to merge them with a PR, so we see the information on the documentation?

Yea, this is a good idea! I think we just overlooked it.

My idea would be to have a separate subheading under Concepts called "Decorators" that contains this content. Let me know if you need help doing that a PR would be welcome!

youngkyo0504:
um... I know Working with API Docs. How about example docs? 

where do I write example?

HarrySIV:
I'm trying `npm run start:website` and getting  `'TYPEDOC_WATCH' is not recognized as an internal or external command,
operable program or batch file.` Any idea how to fix this? I've `npm i` to install all dependencies to no avail.

acywatson:
Oh are you running windows? Easiest fix is to go into that script in the package.json and just locally delete TYPEDOC_WATCH. I will try to fix it to run on windows later

acywatson:
[...]

PR #4047, opened 2023-03-08
Improve Documentation #2845 - lexical/utils

HarrySIV:
With Acy's help I've tried to follow his example and add inline api documentation, though I might have gone a bit overboard. Please let me know if there's anything I can do to improve the PR.

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -283,11 +283,11 @@ type Func = () => void;
 /**
  * Returns a function that will execute all functions passed when called. It is generally used
  * to register multiple lexical listeners and then tear them down with a single function call, such
- * as React's useEffect hook. 
- * @example 
+ * as React's useEffect hook.
+ * @example
  * ```ts
  * useEffect(() => {
- *   return mergeRegister( 
+ *   return mergeRegister(
  *     editor.registerCommand(...registerCommand1 logic),
  *     editor.registerCommand(...registerCommand2 logic),
  *     editor.registerCommand(...registerCommand3 logic)
@@ -295,10 +295,10 @@ type Func = () => void;
  * }, [editor])
  * ```
  * In this case, useEffect is returning the function returned by mergeRegister as a cleanup
- * function to be executed after either the useEffect runs again (due to one of its dependencies 
- * updating) or the compenent it resides in unmounts, like when the component re-renders. 
- * Note the functions don't neccesarily need to be in an array as all arguements 
- * are considered to be the func argument and spread from there. 
+ * function to be executed after either the useEffect runs again (due to one of its dependencies
+ * updating) or the compenent it resides in unmounts.
+ * Note the functions don't neccesarily need to be in an array as all arguements
+ * are considered to be the func argument and spread from there.
  * @param func - An array of functions meant to be executed by the returned function.
  * @returns the function which executes all the passed register command functions.
  */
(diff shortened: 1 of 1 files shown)


HarrySIV:
I'm really unsure as to why the e2e tests are failing. I wonder if using typedoc inline comments inside another one caused an issue, but I don't really know.

acywatson:
e2es are just being flaky

the vercel deployment is failing because of this:

```


Expand 62 Lines
--
17:33:57.095 | SyntaxError: /vercel/path0/packages/lexical-website/docs/api/modules/lexical_utils.md: Expected corresponding JSX closing tag for <div>. (538:70)
17:33:57.095 | 536 \| eg. element = `}<div class="element-inner active small">{`
17:33:57.096 | 537 \| removeClassNamesFromElement(element, `}{`['active small', true, null]`}{`);
```

packages/lexical-utils/src/index.ts
```diff
@@ -226,22 +285,45 @@ export function $findMatchingParent(
 
 type Func = () => void;
 
+/**
+ * Executes each function passed into the mergeRegister function argument.
```
acywatson, line 289: Almost. What this actually does is returns a function that will execute all of the functions passed to mergeRegister when called. The purpose is to make it easier to register a bunch of lexical listeners and then tear them down with a single function call. For example, in a React useEffect.

This one could potentially benefit from an @example annotation

packages/lexical-utils/src/index.ts
```diff
@@ -226,22 +285,45 @@ export function $findMatchingParent(
...
+ * @returns undefined
+ */
 export function mergeRegister(...func: Array<Func>): () => void {
   return () => {
     func.forEach((f) => f());
   };
 }
 
+/**
+ *
```
acywatson, line 300: This attempts to resolve nested element nodes of the same type into a single node of that type. used mostly with marks/commenting

packages/lexical-utils/src/index.ts
```diff
@@ -226,22 +285,45 @@ export function $findMatchingParent(
...
+  /**
+   * @param node - node to be tested against the targetNode instance
+   * @returns true if there is a match, false otherwise
+   */
   const $isTargetNode = (node: LexicalNode | null | undefined): node is N => {
     return node instanceof targetNode;
   };
 
+  /**
+   * Looks for an ancestor of a node of type targetNode. Ensures there are not any children of the node of the same targetNode instance
```
acywatson, line 322: remove

packages/lexical-utils/src/index.ts
```diff
@@ -270,6 +352,10 @@ export function registerNestedElementResolver<N extends ElementNode>(
     return null;
   };
 
+  /**
```
acywatson, line 355: remove

packages/lexical-utils/src/index.ts
```diff
@@ -307,6 +393,12 @@ export function registerNestedElementResolver<N extends ElementNode>(
   return editor.registerNodeTransform(targetNode, elementNodeTransform);
 }
 
+/**
+ * Clones the editor and marks it as dirty to be reconciled. If there was a selection,
```
acywatson, line 397: I wonder how this is different from setEditorState... We can leave as-is for now and polish it up later.

packages/lexical-utils/src/index.ts
HarrySIV: This took me a minute to wrap my head around, but I think I understand how it works especially with useEffect. I hope my explanation is clear and helpful.

acywatson:
Hmm - looks like a prettier error in there somewhere. Can you run prettier locally to see if you can figure out what the issue is?

HarrySIV:
Looked like some extra whitespace at the end of some sentences. Sorry...

acywatson (approved):

PR #4071, opened 2023-03-09
Revert "chore: prevent action from running on forks"

thegreatercurve:
Realized we should revert this before doing the 0.9.0 release. Reverts facebook/lexical#4068


LuciNyan:
Hi @thegreatercurve, I'm really sorry if my PR caused any trouble for you. 😞 I'm just not sure where exactly the issue lies, so if you have a moment, could you please let me know? Thanks!

LuciNyan:
Thank you so much for explaining that to me! <3

acywatson (approved):

PR #4076, opened 2023-03-09
Sticky toolbar to assist in editing large content

ChronicLynx:
**Summary**

- Sticky toolbar
- Removed a duplicate property (resize defined twice)

**With a sticky toolbar you can more easily edit content in this situation:**

https://user-images.githubusercontent.com/29527680/224075876-7eb98c6b-530b-4d19-ad58-a223b823165c.mov


zurfyx (approved):
Thank you, interestingly, we actually have something like this internally but never ported it to OSS!

PR #4077, opened 2023-03-09
Fix toolbars that shouldn't appear in certain cases

WarrenLee19:
For example,
when the number of lines in the editor is more than 2,
gently double-click the end of the line, a toolbar will appear, and this toolbar will not process any text content.
In this case, it is obviously unnecessary to display the toolbar.

https://user-images.githubusercontent.com/48719861/224102364-2faa88a9-0fb7-4fed-9666-cc6240717a49.mov

image
image

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
@@ -307,18 +307,10 @@ function useFloatingTextFormatToolbar(
         setIsText(false);
       }
 
-      if ($isRangeSelection(selection) && !selection.isCollapsed()) {
-        const anchorOffset = selection.anchor.offset;
-        const focusOffset = selection.focus.offset;
-        if (
-          selection.anchor.key !== selection.focus.key &&
-          selection.dirty &&
-          focusOffset === 0 &&
-          anchorOffset > 0
-        ) {
-          setIsText(false);
-          return;
-        }
+      const rawTextContent = selection.getTextContent().replace(/\n/g, '');
+      if (!selection.isCollapsed() && rawTextContent === '') {
+        setIsText(false);
+        return;
       }
     });
   }, [editor]);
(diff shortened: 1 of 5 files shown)


zurfyx (approved):
Thanks, LGTM! We still do formatting on collapsed selection that will carry over the following written text but since we don't show the toolbar either on collapsed selection your solution makes sense to me

zurfyx:
Can you revise the 8 failing checks please?

WarrenLee19:
I'm going to modify these eight

WarrenLee19:
@zurfyx Hi, I have modified the judgment conditions. You can reprocess my pr. thanks

zurfyx:
Can you double check TextFormatting.spec.mjs? Seems like this file is failing consistently

WarrenLee19:
ok, let me try

WarrenLee19:
image

WarrenLee19:
@zurfyx hi, fixed it in TextFormatting.spec.mjs, please reprocess

packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
```diff
@@ -306,6 +306,20 @@ function useFloatingTextFormatToolbar(
...
+        if (
+          selection.anchor.key !== selection.focus.key &&
+          selection.dirty &&
+          focusOffset === 0 &&
+          anchorOffset > 0
+        ) {
+          setIsText(false);
+          return;
+        }
+      }
```
zurfyx, line 322: The problem with the current solution is that when your select spans multiple characers, it gets removed as soon as you take the first action. It should remain around.

Additionally, we probably want to handle this case in a more generic manner. We know that this happens when the selection is not collapsed and when there's no content.

Unfortunately, the `selection.getTextContent` will includes new lines spacers in the output, that's intentional because this function is usually used to export plain text content. We can either rebuild this or assume that this is one only edge case for now and run a simple regex to drop it. I'd probably go for the easy path for now and we can revisit this later since it's quite involved (if it makes any sense to invest in this at all).

```
        const rawTextContent = selection.getTextContent().replace(/\n/g, '');
        if (!selection.isCollapsed() && rawTextContent === '') {
          setIsText(false);
          return;
        }
```

WarrenLee19: @zurfyx  oh, i got it,  unfortunately my solution is not suitable, learn a lot, thanks😄

WarrenLee19:
@zurfyx hi, i has been modified based on your method and reset the TextFormatting.spec.mjs. thanks

zurfyx:
Thank you Warren!

zurfyx (approved):

PR #4082, opened 2023-03-09
Updated InsertTableDialog UX

ChronicLynx:
• Resolved bug allowing empty submission
• Ability to limit the amount of rows and columns
• Simplified labeling
• Disabled submit button when criteria is outside acceptable range 
• Added placeholders

**Before:**

https://user-images.githubusercontent.com/29527680/224183063-a4cfc040-8b97-43bc-9892-48647c81d57d.mov

**After:**

https://user-images.githubusercontent.com/29527680/224183202-e54a2f18-ba5e-4478-b6e3-28b7a90cab3a.mov

### DIFF
--- a/packages/lexical-playground/src/plugins/TablePlugin.tsx
+++ b/packages/lexical-playground/src/plugins/TablePlugin.tsx
@@ -103,14 +103,9 @@ export function InsertTableDialog({
   const [isDisabled, setIsDisabled] = useState(true);
 
   useEffect(() => {
-    if (
-      rows &&
-      +rows > 0 &&
-      +rows <= 500 &&
-      columns &&
-      +columns > 0 &&
-      +columns <= 50
-    ) {
+    const row = Number(rows);
+    const column = Number(columns);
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/TablePlugin.tsx
```diff
@@ -98,8 +98,24 @@ export function InsertTableDialog({
...
+  const [isDisabled, setIsDisabled] = useState(true);
+
+  useEffect(() => {
+    if (
+      rows &&
+      +rows > 0 &&
+      +rows <= 500 &&
+      columns &&
+      +columns > 0 &&
+      +columns <= 50
```
zurfyx, line 112: Nit; oftentimes the longer form `Number(..)` is preferred, it's just more readable. Also, performance wise you're casting 4 times when you can abstract this as a variable on top. It won't make an impact perf wise but I thought you might want to reflect on this

zurfyx (approved):
Thank you! Certainly a lot better

PR #4092, opened 2023-03-10
Add focus tag to editor.focus

strdr4605:
Issue:

`editor.focus()` triggers onChange event.

Example: https://codesandbox.io/s/lexical-focus-onchange-vot70x?file=/src/Editor.js

following 
- #4091

we could handle focus update in onChange

```js
function onChange(editorState: EditorState, editor: LexicalEditor, tags: Set<string>) {
  if (tags.has('focus')) {
    return;
  }

  // do the DB update.
}

...


<OnChangePlugin onChange={onChange} />
```

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -1087,6 +1087,7 @@ export class LexicalEditor {
               callbackFn();
             }
           },
+          tag: 'focus',
         },
       );
       // In the case where onUpdate doesn't fire (due to the focus update not


zurfyx:
Thanks, no objections towards this PR, but what exactly are you trying to do? The OnChange handler works very well a listener but if you're doing anything else than listening then a command handler or transform works best

strdr4605:
We need to save the editor state when there is a change in the editor.  We do this with:

```js
function onChange(editorState: EditorState, editor: LexicalEditor, tags: Set<string>) {
  // do the DB update
}
```

but the `editor.focus()` also triggers onChange, and there is no way to know in the `OnChangePlugin` if the change was triggered by `editor.focus()`. by adding a tag to `.focus()` we can handle it with tags passed to onChange.

Hope you understand what we are doing 😅

zurfyx:
Gotcha, makes sense! What about `ignoreSelectionChange`, I presume you also want to ignore RangeSelection and the like?

strdr4605:
> Gotcha, makes sense! What about `ignoreSelectionChange`, I presume you also want to ignore RangeSelection and the like?

Sure, have `<OnChangePlugin ignoreSelectionChange onChange={onChange} />` 😊. But this PR addresses focus updates.

I am unsure that the proposed changes are the best solution, but I think we need to fix this.

Maybe allow passing custom tag using `EditorFocusOptions`?🤔

zurfyx (approved):

zurfyx:
> > Gotcha, makes sense! What about `ignoreSelectionChange`, I presume you also want to ignore RangeSelection and the like?
> 
> Sure, have `<OnChangePlugin ignoreSelectionChange onChange={onChange} />` 😊. But this PR addresses focus updates.
> 
> I am unsure that the proposed changes are the best solution, but I think we need to fix this.
> 
> Maybe allow passing custom tag using `EditorFocusOptions`?🤔

Happy to merge this in, regardless of this change I just wanted to make sure that you were approaching this problem (on your app) in the best way possible.

Internally at Meta, when we don't always use real-time collaboration for autosaving purposes, we often use `ignoreSelectionChange` to drop any updates that we don't want to take into consideration when saving data. This includes selection and focus, neither of which are important. I didn't fully grasp if your use case was exactly the same but I wanted to make sure you understand all the possibilites and that you can determine the most optimal path forward!

PR #4097, opened 2023-03-10
Use broader return types in `HorizontalRuleNode`

montlebalm:
Use broader return types inside `HorizontalRuleNode` to make extending the node easier. Currently, subclasses cannot override these methods with anything other than what's specified by `HorizontalRuleNode`.


zurfyx (commented):
Thanks! Q: Why would a horizontal rule not be inline? Same for the `updateDOM` behavior. It feels to me like if you want to create a HorizontalRule that diverges too much from what we have on the playground you may want to create your own Node.

zurfyx:
> but could an extending class want to return true in order to update a DOM attribute

You're right! Let's just revert `isInline` then

zurfyx (approved):

zurfyx:
Thank you!

PR #4107, opened 2023-03-12
Improve Documentation #2845 - lexical/list

HarrySIV:
This PR adds documentation to the lexical/list module @ files: formatList.ts, LexicalListItemNode.ts, LexicalListNode.ts, and utils.ts.


HarrySIV:
Thank you! I will say, I was a bit worried about over explaining the logic, there are a lot of nested if checks and I didn't know if it was more helpful to be specific or general in some cases.

PR #4108, opened 2023-03-13
Make typing stricter and increase readability

SikandAlex:
As I examined this file in order to implement multiple `EntityMatch` objects for my own use case I discovered I wasn't getting very helpful type inference and I also didn't know immediately what  `(prevMatch === null || getMode(prevSibling) !== 0)` wad doing. 

I feel like this will make the file more readable for developers. Let me know if I should be directly importing the constant from somewhere instead of doing `!== 'normal'`. Also I'm not super familiar with the repo so not sure if this will break / fail some tests.

### DIFF
--- a/packages/lexical-text/src/index.ts
+++ b/packages/lexical-text/src/index.ts
@@ -16,7 +16,10 @@ import {
   $isParagraphNode,
   $isTextNode,
   TextNode,
+  TextModeType
 } from 'lexical';
+
+import {TEXT_TYPE_TO_MODE} from 'lexical/LexicalConstants'
 
 export type TextNodeWithOffset = {
   node: TextNode;
@@ -163,8 +166,8 @@ export function registerLexicalTextEntity<T extends TextNode>(
     node.replace(textNode);
   };
 
-  const getMode = (node: TextNode): number => {
-    return node.getLatest().__mode;
+  const getMode = (node: TextNode): TextModeType => {
+    return TEXT_TYPE_TO_MODE[node.getLatest().__mode];
   };
 
   const textNodeTransform = (node: TextNode) => {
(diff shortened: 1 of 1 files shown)


SikandAlex:
Alternately, we could just import `IS_NORMAL` from the constants file and just compare the number to that?

packages/lexical-text/src/index.ts
```diff
@@ -183,7 +186,7 @@ export function registerLexicalTextEntity<T extends TextNode>(
       const prevMatch = getMatch(combinedText);
 
       if (isTargetNode(prevSibling)) {
-        if (prevMatch === null || getMode(prevSibling) !== 0) {
+        if (prevMatch === null || getMode(prevSibling) !== 'normal') {
```
SikandAlex, line 189: The other option would be to import `IS_NORMAL` from the same constants file I added and compare against that, leaving the return type of `getMode` as a number.

packages/lexical-text/src/index.ts
```diff
@@ -163,8 +166,8 @@ export function registerLexicalTextEntity<T extends TextNode>(
     node.replace(textNode);
   };
 
-  const getMode = (node: TextNode): number => {
-    return node.getLatest().__mode;
+  const getMode = (node: TextNode): TextModeType => {
```
SikandAlex, line 169: This could be a one liner but I left it in the previous style. 

Personally I would do `const getMode = (node: TextNode): number => node.getLatest().__mode` for the least verbosity but not sure what Meta standard practice is.

packages/lexical-text/src/index.ts
zurfyx: Good point, comparing to a string is usually frown upon as they're no minifiable

zurfyx (commented):
Thank you, this PR is technically correct. For reference, I think the idea is that detail would be able to host more than one type, leveraging the same number to store multiple properties (in a cheaper way). It definitely doesn't seem like it's going to be the case anymore.

However, modifying this now is tricky. Modifying the property means that Collab will not work and we'll need a separate JSON serialization strategy (not ideal). Modifying the function signature is arguably fine, but we'd have to postpone it until 0.10 as it's a breakign change.

cc @acywatson @fantactuka @tylerjbainbridge

SikandAlex:
Sorry, I'm really busy with an upcoming deadline for work. I will return to this and shape the MR up in the next few days.

PR #4115, opened 2023-03-13
Fix ShadowRoot + X insertNode

### LINKED ISSUE #4114: Pasting a table on an empty paragraph crashes
https://user-images.githubusercontent.com/193447/224830607-adde6f98-eb36-4ec8-8f06-a8cec689f8f6.mov

zurfyx:
Turns out the append to ShadowRoot/RootNode condition was not correct. The reason why it's a rare crash is because usually we either don't insert multiple nodes at once or they fall under the replace/merge category (condition 1). This is not the case when you append a `ShadowRoot` followed by another (ElementNode).

Tables are ShadowRoot, and when pasting a table followed by a paragraph it falls under that condition.

Fixes #4114


tylerjbainbridge (approved):
Good catch!

PR #4116, opened 2023-03-13
Paste merged celled table

zurfyx:
Just adding copy-paste support

https://user-images.githubusercontent.com/193447/224838565-abc8f1f1-ac4f-4b56-b2cb-ee56813fa6db.mov

### DIFF
--- a/packages/lexical-table/src/LexicalTableCellNode.ts
+++ b/packages/lexical-table/src/LexicalTableCellNode.ts
@@ -58,12 +58,14 @@ export class TableCellNode extends DEPRECATED_GridCellNode {
   }
 
   static clone(node: TableCellNode): TableCellNode {
-    return new TableCellNode(
+    const tableNode = new TableCellNode(
       node.__headerState,
       node.__colSpan,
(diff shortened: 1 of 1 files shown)


thegreatercurve (approved):

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -243,6 +245,10 @@ export function convertTableCellNodeElement(
       ? TableCellHeaderStates.ROW
       : TableCellHeaderStates.NO_STATUS,
   );
+  if (domNode instanceof HTMLTableCellElement) {
```
thegreatercurve, line 248: I think you might need to add `CAN_USE_DOM` here to guard against any server side rendering bugs.

zurfyx: This is part of `importDOM`, it should be safe from SSR no?

acywatson: headless tho

PR #4117, opened 2023-03-13
Fix tables Collab E2E

zurfyx:
The viewport problem seems to be flaky; but adding the same patch to the remaining tests that use the contextual menu should do the trick for now.

Eventually, we have to fix or delete the contextual menu.


acywatson (approved):

fantactuka (approved):
🚀

PR #4124, opened 2023-03-14
[RFC] Initial stage of refactor for React Native support

amyworrall:
[Disclaimer: all references to Lexical React Native should not be treated as a commitment!]

Lexical React Native will use a different reconciler to Lexical JS, as it will not have access to the DOM. In order to make Lexical Core able to work in both situations, I'm starting a refactor. I've created a thing that I call a "Frontend Adapter" (name TBC, please make suggestions if you think of something better!). Each editor possesses one `FrontendAdapter`, which can be passed in upon creation. We'll have separate Frontend Adapters for DOM, headless, and React Native (eventually).

The Frontend Adapters will (eventually) encapsulate anything that is different between the different platforms. For example, they'll be responsible for kicking off the reconciler; for storing state (e.g. the Root Element for DOM) that only applies to one platform; etc. For now, this is the very first step. Eventually, things like `window` won't be a public method on `FrontendAdapter`, because every bit of code that requires talking to the window will be encapsulated within the specific FrontendAdapter subclass, and the public methods will be at a higher abstraction level.

Questions for discussion:

* Is the name "Frontend Adapter" good? Or do you have a better idea?
* Should I be merging these PRs as I go, or should I wait a little bit? If the latter, it'll be harder for me to track the rest of the code as it is updated, of course.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -86,14 +86,6 @@
         "yarn": ">=1"
       }
     },
-    "eslint-plugin": {
-      "name": "eslint-plugin-lexical",
-      "version": "1.0.0",
-      "dev": true,
-      "peerDependencies": {
-        "eslint": ">=4.19.1"
-      }
-    },
     "node_modules/@algolia/autocomplete-core": {
       "version": "1.7.1",
       "resolved": "https://registry.npmjs.org/@algolia/autocomplete-core/-/autocomplete-core-1.7.1.tgz",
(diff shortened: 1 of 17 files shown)


thegreatercurve:
Closing this for the moment due to the PR becoming stale. Also, internally, we need to focus on other project for the moment before this becomes viable to be merged.

PR #4142, opened 2023-03-17
Update "Spread" to work better with overrided nodes

montlebalm:
Update the `Spread` helper to make overriding nodes easier.

Consider this example where `CustomListNode` extends `ListNode`:
```js
type SerializedCustomListNode = Spread<{
  type: "custom-list";
  version: 1;
}, SerializedListNode>

class CustomListNode extends ListNode {
  exportJSON(): SerializedCustomListNode {
    return {
      ...super.exportJSON(),
      type: "custom-list",
      version: 1,
    };
  }
}
```

Previously, the line `type: 'custom-list'` would throw the type error:
```
Type '"list_module"' is not assignable to type '"list"'
```

That made it difficult to override nodes as you'd need to either disable the type warning or else return `type: "list"` as expected by `SerializedListNode`.

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -37,7 +37,7 @@ import {LineBreakNode} from './nodes/LexicalLineBreakNode';
 import {ParagraphNode} from './nodes/LexicalParagraphNode';
 import {RootNode} from './nodes/LexicalRootNode';
 
-export type Spread<T1, T2> = Omit<T2, keyof T1> & T1;
+export type Spread<T1, T2> = {[K in Exclude<keyof T1, keyof T2>]: T1[K]} & T2;
 
 export type Klass<T extends LexicalNode> = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any


thegreatercurve (commented):
This is what the Spread type used to be in #2434, but we changed it to use Omit to fix some unrelated type issues, shown in the PR. 

I think the easier fix is probably just to remove string literal types from the nodes and instead just use `string`.  

It's less typesafe, but we still have some decent runtime error handling around checking the node type.

acywatson:
>This is what the Spread type used to be in https://github.com/facebook/lexical/pull/2434, but we changed it to use Omit to fix some unrelated type issues, shown in the PR.

I think the easier fix is probably just to remove string literal types from the nodes and instead just use string.

It's less typesafe, but we still have some decent runtime error handling around checking the node type.

Also has the benefit of being a backwards compatible change - sorry @montlebalm - I forgot about that PR John linked when I recommended this fix.

thegreatercurve:
Closing in favour of #4143. I need to test it internally first to make sure it's not horrendously misguided, but this should fix your bug.

PR #4146, opened 2023-03-17
Color picker - fixes #4127

### LINKED ISSUE #4127: Bug: Color picker hex input impossible to use. Picker closes on input focus.
Attempting to input a hex color into the color picker immediately closes the color picker making it impossible to use.

Lexical version: Current

## Steps To Reproduce

Reproducible directly in [Lexical playground](https://playground.lexical.dev/)

1. Highlight text
2. Open color picker
3. Click hex input field

Tested and reproduced in Safari 16.2 /  Chrome 110.0.5481.77

HarrySIV:
This PR technically fixes #4127 the color picker closing immediately on color selection, including typing in the TextInput field, but there are a couple more issues here that could use fixing, I'm just not sure how to go about all of them.

1. It'd be nice if you could make the ::selection background clear so that you can see the color as it changes, but only while the color picker is open. I suppose I could do that with a useEffect hook and check if the picker isOpen or not, but I'm not exactly sure how to change the raw CSS for that. 
2. If you type in the hex code and accidentally type again, either hitting enter or something, it will replace the current selection with the key stroke. I'm not sure what the best way is to prevent that, but my first thought is don't allow the TextInput to autocomplete, instead require an enter keystroke or for the TextInput to lose focus in order to acknowledge the color change. 
I feel like that may be a big ask though.

### DIFF
--- a/packages/lexical-playground/src/ui/ColorPicker.tsx
+++ b/packages/lexical-playground/src/ui/ColorPicker.tsx
@@ -23,6 +23,7 @@ interface ColorPickerProps {
   color: string;
   children?: ReactNode;
   onChange?: (color: string) => void;
+  stopCloseOnClickSelf?: boolean;
   title?: string;
 }
 
@@ -52,6 +53,7 @@ export default function ColorPicker({
   children,
   onChange,
   disabled = false,
+  stopCloseOnClickSelf = true,
   ...rest
 }: Readonly<ColorPickerProps>): JSX.Element {
   const [selfColor, setSelfColor] = useState(transformColor('hex', color));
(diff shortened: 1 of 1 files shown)


zurfyx (approved):
Thanks! This makes sense to me but honestly I don't see the use case behind the `stopCloseOnClickSelf` property. Why would a dropdown ever close when you're clicking it? For example, I feel like the Insert and Align dropdowns we have do not have the correct behavior. That's a follow up item if you feel like fixing the remaining

HarrySIV:
I believe the default behavior for a lot of editors is to close the drop down on click, though I understand how someone who is editing might want to try out different looks before choosing one, and that having to constantly reopen the drop down menu can be annoying (I've been there). Or if they want to add like 5 horizontal rules.
Looking around at other color pickers, I see a lot of editors have an "ok" and "cancel" button that closes the drop down/modal and finalizes the change.
I can certainly go ahead and make the changes to insert and align items though.

PR #4147, opened 2023-03-17
DecoratorNode not disabling placeholder at root

### LINKED ISSUE #4145: Bug: placeholder should not be shown for single DecoratorNode
Lexical version: 0.9

image

It seems that `DecoratorNode` is not checked:

https://github.com/facebook/lexical/blob/5445f2ce45496a5305a1bd6c9df2bc8d7dfc737b/packages/lexical-text/src/index.ts#L114-L137

HarrySIV:
This PR fixes #4145 . TBH I'm not completely sure what the expected behavior of a DecoratorNode is supposed to be, or whether this PR might break something. I ran the unit tests and they passed, but I couldn't get the e2e tests to work even for the unchanged main.

### DIFF
--- a/packages/lexical-text/src/index.ts
+++ b/packages/lexical-text/src/index.ts
@@ -12,6 +12,7 @@ import type {Klass, LexicalEditor, LexicalNode, RootNode} from 'lexical';
 import {
   $createTextNode,
   $getRoot,
+  $isDecoratorNode,
   $isElementNode,
   $isParagraphNode,
   $isTextNode,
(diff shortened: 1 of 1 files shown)


packages/lexical-text/src/index.ts
```diff
@@ -114,7 +115,7 @@ export function $canShowPlaceholder(isComposing: boolean): boolean {
   for (let i = 0; i < childrenLength; i++) {
     const topBlock = children[i];
 
-    if ($isElementNode(topBlock)) {
+    if ($isElementNode(topBlock) || $isDecoratorNode(topBlock)) {
```
acywatson, line 118: This should be fine, but you can probably make this a separate condition and return early instead of proceeding to all those other checks - I don't think any of those will pass on a decorator.

cleverhare (commented):
LGTM :)

PR #4148, opened 2023-03-17
Close on click handling

HarrySIV:
An extension of PR #4146.


HarrySIV:
The color picker plugin must not have been included in the tests because, if I understand correctly, these tests look to be failing as the selection isn't being restored due to the drop down menu never closing...

zurfyx:
Thanks! I believe the menu should close when actions are taken but otherwise, if just clicking the blank space if should remain open. This should fix the broken tests as you suggested!

HarrySIV:
I'll work on it!

HarrySIV:
Okay, I'll be the first to admit, this logic looks really ugly, but it works _~almost~_ as intended. I'm still struggling to figure out how to handle the color picker besides turning it into a button to pop up a modal that let's you select the custom color. Also typing in the hex code doesn't automatically close the plugin, but I'd have to pass `handleClose` through the context and that proved to be difficult with how it's currently set up.

zurfyx:
Thank you, no objections toward merging this as long as the integrity check are fixed! However, I would be inclined on a less magical system. We don't need to know which elements on the page will close the page or not, we can delegate this responsibility to the user. The reason why I'm saying this is because any change to the type of elements that display the menu will break this logic.

I'm thinking of something like

```
<Dropdown>
  <DropdownItem>...</DropdownItem>
</Dropdown>
```

Where DropdownItem by default triggers dropdown close. To do this, I believe you can use some sort of Context to communite with the parent logic in a clean manner.

I've just skimmed through the code but I believe this is the idea behind React Boostrap https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Dropdown.tsx#L173

HarrySIV:
I completely agree and I pursued this idea first but I was running into trouble trying to work with the context that was already being passed. I know I can pass an object containing the useMemo function but I kept getting a weird type error that I couldn't solve, so I opted for this instead. I'd love to pass down a handleClickToClose type function where the user can pass it to their buttons to close the drop down. I'll take another crack at it, since it would be cleaner and more customizable for the developer.

HarrySIV:
Still having issues with Excalidraw and I'm not sure what to do about it.
This PR has sort of evolved so allow me to reiterate the intent. Essentially, the drop down menu should close when an action occurs that alters the editor contents. What I've done is forced the developer to normalize those actionable components to fit in as a DropDownItem component. I extended DropDownItem  to allow text input as well. Basically, anytime a button is clicked or the text input executes its desired function, the drop down menu is closed. Clicking outside of the drop down should also still close it. The only part of this puzzle that wasn't fixed was the custom color picker, which I think should still be a button that closes the drop down and opens a modal to select the custom color, and this setup will allow for that functionality.

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -17,59 +17,91 @@ import {
...
+        <TextInput
+          label={props.label}
+          onChange={(e: string) => props.onChange(e, closeDropDown)}
+          value={props.value}
+        />
+      ) : (
+        <button
+          className={props.className}
+          key={props.key}
+          onClick={async (e) => {
```
HarrySIV, line 82: This is hacky and I'm not exactly sure why it works but it does. I think the problem is the drop down is closing before the editor can update, but I'm not sure.

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -17,59 +17,91 @@ import {
...
+            await props.onClick(e);
```
HarrySIV, line 83: This is hacky and I'm not sure why it works but it does. I think the problem is the drop down is closing before the editor has time to update, but I'm not sure

packages/lexical-playground/src/ui/ColorPicker.tsx
```diff
@@ -75,11 +72,12 @@ export default function ColorPicker({
...
   );
 
-  const onSetHex = (hex: string) => {
+  const onSetHex = async (hex: string, closeDropDown: () => void) => {
     setInputColor(hex);
     if (/^#[0-9A-Fa-f]{6}$/i.test(hex)) {
       const newColor = transformColor('hex', hex);
-      setSelfColor(newColor);
+      await setSelfColor(newColor);
+      closeDropDown();
```
HarrySIV, line 80: Again, same hacky fix. It works but I'm not sure why.

HarrySIV:
Is this okay to merge?

Issue #4155, opened 2023-03-20
Bug: Popover / typeahead menus stop working when resizing window


AlessioGr:
https://user-images.githubusercontent.com/70709113/226220144-a0dca774-7aa4-490a-a794-f9fdecc2dcd2.mp4

This video showcases the issue

Lexical version: 0.8

## Steps To Reproduce

1. Resize the browser (Chrome) (sometimes it works, sometimes it doesn't. I've had the most luck when opening the browser console, as seen in the video). What also works reliably to reproduce the bug is putting the window in the split screen mode of "Arc Browser" and then resizing it.
2. Try to click on a link

Link to code example: https://playground.lexical.dev/

## The current behavior
The popover stops working

## The expected behavior
The popover should keep working

It seems that it might break some events, as the code which runs the position change for the popover menu is not run anymore in the first place.

acywatson:
Reproduced this on Chrome

etrepum:
The reason this happens is because the FloatingLinkPlugin is only installed when `!isSmallWidthViewport`. It might make sense to use more specific widths for each plugin, but this is working as specified.

```ts
  useEffect(() => {
    const updateViewPortWidth = () => {
      const isNextSmallWidthViewport =
        CAN_USE_DOM && window.matchMedia('(max-width: 1025px)').matches;

      if (isNextSmallWidthViewport !== isSmallWidthViewport) {
        setIsSmallWidthViewport(isNextSmallWidthViewport);
      }
    };
    updateViewPortWidth();
    window.addEventListener('resize', updateViewPortWidth);

    return () => {
      window.removeEventListener('resize', updateViewPortWidth);
    };
  }, [isSmallWidthViewport]);
```

PR #4158, opened 2023-03-20
UX: Floating link editor: better positioning

AlessioGr:
## Old:

https://user-images.githubusercontent.com/70709113/226237755-c9176cc2-8527-4a09-878d-5620e8d6c11f.mp4

## New:

https://user-images.githubusercontent.com/70709113/226237880-9a732cfa-171a-4e2f-ab5f-f4ddd4323b3a.mp4

The new floating link editor has the following benefits:
- Less distractions when writing, as it moves around in a more predictable way - always under the link. It doesn't move around when clicking on different positions in the link as seen in the old one. I found that to be pretty annoying and distracting
- The old one sometimes overlaps the link & positions itself incorrectly - as seen in the video. Very annoying! This does not happen with the new one anymore
- Even for longer selections, the floating link editor now positions itself correctly, directly under the link
- Simpler code & logic => better performance

This code also changes

`  let top = targetRect.top - floatingElemRect.height - verticalGap;`

to

`  let top = targetRect.top - verticalGap;`

in the setFloatingElemPosition.ts as that extra minus caused incorrect and non-predictable positioning. Instead of overriding  the file, I created a new one, to avoid breaking anything which uses that.

I think the reason this was glitchy is that floatingElemRect.height  does not have a height for the first click, but it does have a height for consecutive clicks, resulting in positioning inconsistency.

Having two almost identical setFloatingElemPosition files might be weird architecturally, so feel free to change this around if you have a better solution!

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
@@ -28,7 +28,7 @@ import * as React from 'react';
 import {createPortal} from 'react-dom';
 
 import {getSelectedNode} from '../../utils/getSelectedNode';
-import {setFloatingElemPosition} from '../../utils/setFloatingElemPosition';
+import {setFloatingElemPositionForLinkEditor} from '../../utils/setFloatingElemPositionForLinkEditor';
 import {sanitizeUrl} from '../../utils/url';
 
 function FloatingLinkEditor({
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -28,7 +28,7 @@ import * as React from 'react';
 import {createPortal} from 'react-dom';
 
 import {getSelectedNode} from '../../utils/getSelectedNode';
-import {setFloatingElemPosition} from '../../utils/setFloatingElemPosition';
+import {setFloatingElemPositionForLinkEditor} from '../../utils/setFloatingElemPositionForLinkEditor';
```
thegreatercurve, line 31: If we're not using it elsewhere, you can just outright delete `setFloatingElemPosition`.

thegreatercurve:
LGTM. Thanks for your contributions in help to fix the link popover @AlessioGr!

Does this fix #4155 as well?

thegreatercurve (approved):

AlessioGr:
> LGTM. Thanks for your contributions in help to fix the link popover @AlessioGr!
> 
> Does this fix #4155 as well?

I don't think it fixes that. The code I changed here doesn't even run when #4155 occurs

PR #4161, opened 2023-03-20
Add a title to the link node to support titles in markdown

themagickoala:
We have a use case for supporting titles in link markdown nodes (see [markdown spec](https://www.markdownguide.org/basic-syntax/#adding-titles "Even github does it!")). We're currently copying source code and modifying it inside our own project, but that's not really sustainable so here's a PR to add it to Lexical.

Note that this doesn't add support for adding the title via the WYSIWYG editor, but it will support adding it in Markdown and then converting back to rich text.


packages/lexical-react/src/LexicalTreeView.tsx
```diff
@@ -604,6 +608,15 @@ function printRelProperties(node: LinkNode) {
   return str;
 }
 
+function printTitleProperties(node: LinkNode) {
+  let str = node.getTitle();
+  // TODO Fix nullish on LinkNode
+  if (str != null) {
```
zurfyx, line 614: If I'm reading this right, it's guaranteed to be `null` or `string` no? So we can do `!== null` instead and drop the TODO?

zurfyx (approved):
LGTM, thank you! cc @fantactuka if you want to double check the MD stuff

packages/lexical-react/src/LexicalTreeView.tsx
themagickoala: Probably, but this was a copy and paste of target/rel versions, which are also string | null.

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -340,11 +343,11 @@ export const LINK: TextMatchTransformer = {
       return linkContent;
     }
   },
-  importRegExp: /(?:\[([^[]+)\])(?:\(([^()]+)\))/,
-  regExp: /(?:\[([^[]+)\])(?:\(([^()]+)\))$/,
+  importRegExp: /(?:\[([^[]+)\])(?:\((?:([^()\s]+)(?:\s"([^"]+)"\s*)?)\))/,
+  regExp: /(?:\[([^[]+)\])(?:\((?:([^()\s]+)(?:\s"([^"]+)"\s*)?)\))$/,
```
fantactuka, line 347: Any chance to check if " was escaped in title?

```
[Text](https://test.com "Title \" with escaped quote")
```

fantactuka (approved):

packages/lexical-markdown/src/MarkdownTransformers.ts
themagickoala: Good catch, I'll try and add a test case for that too, and fix the other unit test that's broken.

themagickoala:
@zurfyx can this be merged now?

themagickoala:
@fantactuka @zurfyx I've rebased this to fix the conflict created by my other PR (#4168) so it should be good to merge now I think.

packages/lexical-react/src/LexicalTreeView.tsx
acywatson: I'm fine with this - we can fix it now or we can merge this and fix them all in a separate PR -all good either way for me.

zurfyx:
Thank you! Sorry for the delay

PR #4162, opened 2023-03-20
Fix table stealing selectionchange

### LINKED ISSUE #3785: Bug: Cannot deselect selected playground content when selection includes a table and some other content
Lexical version: current playground version

## Steps To Reproduce

1. Start with the Lexical playground
2. Add a table
3. Click and drag a selection that includes both some part of the table, or even all of it, and some other text content on the page as well.

## The current behavior
A mouse click anywhere does not automatically deselect the selection.

## The expected behavior
A mouse click anywhere should deselect the selection, as it does in all other playground contexts.

### LINKED ISSUE #4138: Bug: Table prevents text deselection
As soon as a table is added to the rich-text-editor, the text (de)selection is bugged. 

Lexical version: 0.9.0

## Steps To Reproduce

1. Insert a table node into the editor. You can reproduce the issue in the playground (https://playground.lexical.dev/)
2. Select some text outside the table
3. Try to move the caret to a position inside the selected text. It won't work. You can only clear the selection by clicking elsewhere.

If you remove the table again, everything works as expected again.

zurfyx:
Turns out that the `mouseDown` logic inside the table plugin was handling more events than it should.

https://user-images.githubusercontent.com/193447/226449211-70ead2ad-3f37-4cf7-9565-4e007d6a7bac.mov

Fixes #3785 
Fixes #4138

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -37,6 +37,7 @@ import {
   DELETE_LINE_COMMAND,
   DELETE_WORD_COMMAND,
   DEPRECATED_$createGridSelection,
+  DEPRECATED_$isGridNode,
   DEPRECATED_$isGridSelection,
   FOCUS_COMMAND,
   FORMAT_TEXT_COMMAND,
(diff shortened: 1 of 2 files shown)


tylerjbainbridge (approved):
good catch!

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -141,21 +142,29 @@ export function applyTableHandlers(
```
fantactuka, line 142: Is this one needed at all?

fantactuka (approved):

thegreatercurve (approved):

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
zurfyx: Discussed offline; let's leave this for now as the comment above is important

```
  // Select entire table at this point, when grid selection is ready.
```

Basically means that it wasn't implemented yet 🙃

PR #4163, opened 2023-03-20
Fix TableResizer when multiple editors

### LINKED ISSUE #4157: Bug: Tables do not work if there are 2 or more lexical editors on one page
When there are 2 or more lexical editors on one page, the table throws a "TableCellResizer: Table cell node not found." error if you hover over it with your mouse.

Lexical version: 0.9.0

## Steps To Reproduce

1. Have 2 editors in one page (as seen in the playground of https://github.com/AlessioGr/lexical/tree/two-editors)
2. place a table and hover over it

Link to code example: https://github.com/AlessioGr/lexical/tree/two-editors

zurfyx:
The current module expect the cell to always be found. This won't be the case when there's multiple editors on the page.

Rewrote a chunk of the logic to
1. Leverage EditorState instead of HTML
2. throttle instead of setTimeout with different values depending on whether it's resizing - this should give the editor overall better performance

There's more stuff around to be fixed, the HTML `._cell` format won't work with merged cells but that's a story for another PR

https://user-images.githubusercontent.com/193447/226481598-93eb758a-30e5-40ec-86a9-591c5515983d.mov

Fixes #4157

### DIFF
--- a/packages/lexical-playground/src/App.tsx
+++ b/packages/lexical-playground/src/App.tsx
@@ -161,6 +161,7 @@ export default function PlaygroundApp(): JSX.Element {
   return (
     <SettingsContext>
       <App />
+      <App />
       <a
         href="https://github.com/facebook/lexical/tree/main/packages/lexical-playground"
         className="github-corner"
--- a/packages/lexical-playground/src/App.tsx
+++ b/packages/lexical-playground/src/App.tsx
@@ -161,7 +161,6 @@ export default function PlaygroundApp(): JSX.Element {
   return (
     <SettingsContext>
       <App />
-      <App />
       <a
         href="https://github.com/facebook/lexical/tree/main/packages/lexical-playground"
         className="github-corner"
--- a/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
@@ -93,7 +93,7 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
   }, []);
 
   useEffect(() => {
-    const THROTTLE_MS = draggingDirection !== null ? 20 : 200;
+    const throttleMs = draggingDirection !== null ? 20 : 200;
     const [onMouseMove, onMouseMoveDecommission] = throttle1(
       (event: MouseEvent) => {
         const target = event.target;
@@ -152,7 +152,7 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
           });
         }
       },
-      THROTTLE_MS,
+      throttleMs,
     );
 
     document.addEventListener('mousemove', onMouseMove);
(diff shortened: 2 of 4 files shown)


packages/lexical-playground/src/utils/throttle.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+export function throttle1<T1>(
```
thegreatercurve, line 9: Why does this have 1 at the end?

thegreatercurve (approved):

packages/lexical-playground/src/App.tsx
```diff
@@ -160,6 +160,7 @@ function App(): JSX.Element {
 export default function PlaygroundApp(): JSX.Element {
   return (
     <SettingsContext>
+      <App />
```
thegreatercurve, line 163: Remove this before merging

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1001,6 +1001,7 @@ export function getTableSelectionFromTableElement(
   return tableElement[LEXICAL_ELEMENT_KEY];
 }
 
+// TODO 0.10 Deprecate
```
thegreatercurve, line 1004: Maybe just create a PR for this now and add "[0.10.0]" to the title. Otherwise, we'll forget about this TODO.

packages/lexical-playground/src/utils/throttle.ts
zurfyx: 1 param, afaik there's no way to type-safe infinite args. We follow the same pattern internally with Reason

tylerjbainbridge (approved):

packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
```diff
@@ -105,40 +110,55 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
           return;
         }
 
-        if (targetRef.current !== target) {
-          targetRef.current = target as HTMLElement;
-          const cell = getCellFromTarget(target as HTMLElement);
-
-          if (cell && activeCell !== cell) {
```
bencarletonn, line 112: hi, just a heads up. it appears by removing this check, an editor update will be triggered each time the cursor is moved anywhere, as opposed to each time the cursor moves onto a new cell in the table. is this the intended behavior?

AlessioGr:
> @zurfyx Closing this due to the PR becoming but feel free to re-open if needed

Hey @thegreatercurve , what do you mean with "due to the PR becoming"?

thegreatercurve:
> > @zurfyx Closing this due to the PR becoming but feel free to re-open if needed
> 
> Hey @thegreatercurve , what do you mean with "due to the PR becoming"?

Forgot to mention the word "stale".

PR #4166, opened 2023-03-21
Improve Documentation - multiple modules

HarrySIV:
This PR adds documentation to multiple small modules including: file, hashtag, headless, history, link, and text. #2845


packages/lexical-history/src/index.ts
```diff
@@ -498,6 +507,10 @@ export function registerHistory(
   };
 }
 
+/**
+ * Creates an exmpty history state where the stacks are empty arrays and current is null.
```
acywatson, line 511: typo here - also maybe leave offf the part about stacks/current, as those are implementation details.

packages/lexical-text/src/index.ts
```diff
@@ -68,6 +75,12 @@ export function $findTextIntersectionFromCharacters(
   return null;
 }
 
+/**
+ * Determines if the root has any text content and can trim any whitespace if it does.
+ * @param isEditorComposing - Has the editor recieved input?
```
acywatson, line 80: isEditorComposing is about whether or not the editor is in "composition mode", meaning that an IME of some sort is active (like Gboard or the text suggestions that pop up when you write in Korean or Chinese, for example). Seems like kind of a useless parameter, tbh.

packages/lexical-text/src/index.ts
```diff
@@ -147,6 +181,25 @@ export function $canShowPlaceholderCurry(
...
+ * @example
+ * ```ts
+ *   useEffect(() => {
+    return mergeRegister(
+      ...registerLexicalTextEntity(editor, getMatch, targetNode, createNode),
+    );
+  }, [createNode, editor, getMatch, targetNode]);
+ * ```
+ * Where targetNode is the type of node containing the text you want to transform (like a text input),
+ * then getMatch uses a regex to find a #foo and then creates a HashtagNode containing #foo.
```
acywatson, line 196: not necessarily a HashtagNode, right? This can be used for mentions and many other things, as well.

HarrySIV: Yes, I was trying to give a good example but I suppose I was a bit too specific. I'll be more general.

packages/lexical-link/src/index.ts
```diff
@@ -246,13 +246,24 @@ function convertAnchorElement(domNode: Node): DOMConversionOutput {
   return {node};
 }
 
+/**
+ * Takes a URL and creates a LinkNode.
+ * @param url - The URL the LinkNode should direct to.
+ * @param attributes - Optional HTML a tag attributes { target, rel }
```
themagickoala, line 252: #4161 adds a third attribute to this interface: `title`. The conflicts between these two PRs will need resolving at some point.

HarrySIV:
Excalidraw is causing issues, I know a recent merge updated it.

HarrySIV:
Can this be merged?

acywatson (approved):

PR #4178, opened 2023-03-22
Allow passing in a custom Set of tags to ignore in $generateNodesFromDOM

amanharwara:
Currently there is no way of modifying the HTML tags that the `$generateNodesFromDOM` function ignores without maintaining a copy of both the `$generateNodesFromDOM` and `$createNodesFromDOM` functions in our own codebase. This PR adds a way to pass in a custom `Set` of HTML tags that get ignored when generating nodes from DOM.

### DIFF
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -32,6 +32,7 @@ import {$getRoot, $isElementNode, $isTextNode} from 'lexical';
 export function $generateNodesFromDOM(
   editor: LexicalEditor,
   dom: Document,
+  ignoreTags = IGNORE_TAGS,
 ): Array<LexicalNode> {
   let lexicalNodes: Array<LexicalNode> = [];
   const elements = dom.body ? dom.body.childNodes : [];
@@ -39,8 +40,8 @@ export function $generateNodesFromDOM(
   for (let i = 0; i < elements.length; i++) {
     const element = elements[i];
 
-    if (!IGNORE_TAGS.has(element.nodeName)) {
-      const lexicalNode = $createNodesFromDOM(element, editor);
+    if (!ignoreTags.has(element.nodeName)) {
+      const lexicalNode = $createNodesFromDOM(element, editor, ignoreTags);
 
       if (lexicalNode !== null) {
         lexicalNodes = lexicalNodes.concat(lexicalNode);
@@ -167,13 +168,14 @@ const IGNORE_TAGS = new Set(['STYLE']);
 function $createNodesFromDOM(
   node: Node,
   editor: LexicalEditor,
+  ignoreTags = IGNORE_TAGS,
   forChildMap: Map<string, DOMChildConversion> = new Map(),
   parentLexicalNode?: LexicalNode | null | undefined,
   preformatted = false,
 ): Array<LexicalNode> {
   let lexicalNodes: Array<LexicalNode> = [];
 
-  if (IGNORE_TAGS.has(node.nodeName)) {
+  if (ignoreTags.has(node.nodeName)) {
     return lexicalNodes;
   }
 
@@ -220,6 +222,7 @@ function $createNodesFromDOM(
       ...$createNodesFromDOM(
         children[i],
         editor,
+        ignoreTags,
         new Map(forChildMap),
         currentLexicalNode,
         preformatted ||
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -32,7 +32,7 @@ import {$getRoot, $isElementNode, $isTextNode} from 'lexical';
 export function $generateNodesFromDOM(
   editor: LexicalEditor,
   dom: Document,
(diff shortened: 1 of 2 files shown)


amanharwara:
> Could you explain your use case a bit more? I think it's possible to ignore tags now by adding those tags to importDOM() rules with high priority and returning `{ node: null }`.

We're working on a clipper extension that would allow users to clip a full page for example, and on some pages that include inline `<script>`s, those seem to get converted to `TextNode`s.

Admittedly haven't tried the `importDOM` method because I'm not sure what node that would go on? Since I can't modify the `importDOM` method of the `TextNode` itself, I'm guessing I'd have to create a custom node like `IgnoreScriptsNode` that extends `TextNode` and has an `importDOM` to ignore script tags?

Not entirely opposed to that idea, but we're also using `$generateNodesFromDOM` in another place where we don't care about ignoring those, so having that as a function argument feels better.

> I think it will still process the children though, which might not be what they want. Another option might be to remove the node from the DOM to avoid further child processing, if that's the case (untested).

In our particular usecase I don't think child processing would be much of an issue considering `<script>` tags can only really contain text.

acywatson:
> Admittedly haven't tried the `importDOM` method because I'm not sure what node that would go on? Since I can't modify the `importDOM` method of the `TextNode` itself, I'm guessing I'd have to create a custom node like `IgnoreScriptsNode` that extends `TextNode` and has an `importDOM` to ignore script tags?
> 
> Not entirely opposed to that idea, but we're also using `$generateNodesFromDOM` in another place where we don't care about ignoring those, so having that as a function argument feels better.

Yea you could do it with importDOM as you described. I wonder if we could have an optional importDOM prop in the editor config that allows you to register conversions without having to create or override a node. That is kind of unwieldy.

On the other hand, I don't have a huge problem with this API, but I might say that, if we go ahead with this, we should add an options bag instead and just make ignoreTags one property on that just to sort of put a lid on the proliferation of args here.

acywatson:
> It doesn't seem to actually ignore those though. I see the elements in the console log, but they still get converted to >TextNodes.

Yea I suspect there's actually a DOM TextNode nestled in that Script node, which is why I brought up the thing about the children. I'm fine with the change you're proposing, just make the arg like:

```
options: { ignoreTags: string[] }
```

amanharwara:
> Should we just always ignore script tags (similar to style)

That would be enough for our usecase, but as @acywatson mentioned, there might be someone who will want to convert script tags or maybe ignore some other tag in which case these changes allow for that.

I think it might make sense to do both - ignore both `script` and `style` tags by default, and allow users to not ignore those if they want to.

packages/lexical-html/src/index.ts
```diff
@@ -32,15 +32,16 @@ import {$getRoot, $isElementNode, $isTextNode} from 'lexical';
 export function $generateNodesFromDOM(
   editor: LexicalEditor,
   dom: Document,
+  {ignoreTags = IGNORE_TAGS}: {ignoreTags?: Set<string>} = {},
```
acywatson, line 35: I wonder if you should be able to just specify additional tags to ignore. The way this works, it seems like you'd have to manually copy over the IGNORE_TAGS baked into the core and then add your additional tags on top.

packages/lexical-html/src/index.ts
```diff
@@ -167,13 +168,21 @@ const IGNORE_TAGS = new Set(['STYLE']);
 function $createNodesFromDOM(
   node: Node,
   editor: LexicalEditor,
-  forChildMap: Map<string, DOMChildConversion> = new Map(),
-  parentLexicalNode?: LexicalNode | null | undefined,
-  preformatted = false,
+  {
+    ignoreTags = IGNORE_TAGS,
```
acywatson, line 172: nit: can't we just pass through the value from the caller, instead if setting a default value again here?

amanharwara: we can, at runtime that wouldn't be an issue since this function is only consumed by `$generateNodesFromDOM` which will always pass the value. setting the default here is mostly just to satisfy typescript so that it doesn't think that `ignoreTags` is `undefined` in the `if (ignoreTags.has(node.nodeName)) {` statement. we can also not set the default and change the if-statement ot `ignoreTags && ignoreTags.has(...)`
do you have a particular preference for one over the other?

packages/lexical-html/src/index.ts
amanharwara: that's true, but this does allow if for whatever reason the user wants to explicitly not ignore a `<style>` tag. although that would probably be a very niche usecase, so if you feel like it would be better to not allow overriding the default ignored tags and only allow extending them, then i can refactor it to be so.

amanharwara: also, should i add `SCRIPT` to the default ignored tags set?

zurfyx (approved):

acywatson:
We now ignore script tags by default: 

#4249 

Following up on the generic ignore solution in:

#4253

PR #4182, opened 2023-03-22
Table toolbar foundation

zurfyx:
Not adding the dropdown buttons just yet as I think we probably want to abstract the methods, but we're also actively rewriting the methods..

Screenshot 2023-03-22 at 6 46 15 PM


fantactuka (approved):
🚀

PR #4189, opened 2023-03-23
Pass editor._updateTags to importJSON

nemophrost:
Since `importJSON` is a static method on node classes, there isn't a good way to determine what is going on when `importJSON` is called - for instance are we deserializing the document initially or pasting in some lexical content from the clipboard.

It seems like `editor._updateTags` is a reasonable place to have this kind of information (as it already has `paste` when pasting) so by passing updateTags to `importJSON`, custom logic can be implemented if desired and should be extensible going forward by using additional updateTags when needed.

### DIFF
--- a/packages/lexical-selection/src/__tests__/utils/index.ts
+++ b/packages/lexical-selection/src/__tests__/utils/index.ts
@@ -824,7 +824,7 @@ export async function applySelectionInputs(inputs, update, editor) {
                 {
                   clipboardData: {
                     getData: (type) => {
-                      if (type === 'application/x-lexical-nodes') {
+                      if (type === 'application/x-lexical-editor') {
                         return input.text;
                       }
 
--- a/packages/lexical-utils/src/__tests__/unit/LexicalEventHelpers.test.tsx
+++ b/packages/lexical-utils/src/__tests__/unit/LexicalEventHelpers.test.tsx
@@ -19,9 +19,18 @@ import {HeadingNode, QuoteNode} from '@lexical/rich-text';
 import {
   applySelectionInputs,
   pasteHTML,
+  pasteLexical,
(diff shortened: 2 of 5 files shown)


fantactuka:
Maybe we should we pass editor instance instead as we do for create/export json

acywatson:
> Maybe we should we pass editor instance instead as we do for create/export json

Agree - either this or I think it should be possible to use the existing util method, $hasUpdateTag, inside that node method:

https://github.com/facebook/lexical/blob/e07591e024c585492d9f6a3b83176ccadfb72913/packages/lexical/src/LexicalUtils.ts#L1254

fantactuka:
That's even better, until we have a good reason to pass editor we can keep current api

nemophrost:
> @nemophrost please re-open or tag me or @fantactuka if $hasUpdateTags doesn’t work for you.

@acywatson `$hasUpdateTags` isn't exported in `lexical/src/index.ts` and my build tooling is choking on trying to import directly from `lexical/LexicalUtils` happy to create a new PR to add `$hasUpdateTag` to index file if needed or is there another way to correctly import that?

acywatson:
> > @nemophrost please re-open or tag me or @fantactuka if $hasUpdateTags doesn’t work for you.
> 
> @acywatson `$hasUpdateTags` isn't exported in `lexical/src/index.ts` and my build tooling is choking on trying to import directly from `lexical/LexicalUtils` happy to create a new PR to add `$hasUpdateTag` to index file if needed or is there another way to correctly import that?

Yea go ahead and make a PR if you don’t mind. Not sure why that isn’t exported.

nemophrost:
@acywatson https://github.com/facebook/lexical/pull/4216

PR #4201, opened 2023-03-25
Only allow numbers in TablePlugin row / column number input

AlessioGr:
Previously, you were able to input text for the row / column number count.


fantactuka (commented):
Let's pass type prop instead that will be text by default

AlessioGr:
> Let's pass type prop instead that will be text by default

Would you make the type prop optional & text by default if not specified, or required?

AlessioGr:
> Let's pass type prop instead that will be text by default

Good idea - I've updated it!

zurfyx (approved):
LGTM, thank you!

PR #4204, opened 2023-03-25
Fix Clear Formatting (Fixes #4188)

### LINKED ISSUE #4188: Bug: Clear formatting doesn't respect the selected text
Lexical version: 0.9

## Steps To Reproduce

1. Bold a few words
2. Select one of those words
3. Click "Clear Formatting"

## The current behavior vs the expected behavior

It clears the formatting for the entire editor text vs just the selected text

ivailop7:
Clear formatting respects the selection, also converts Headings and Quotes down to Paragraphs, but doesn't convert Code Blocks or any list types, matching the behaviour of other popular text editors.

Before:

https://user-images.githubusercontent.com/7893468/227743734-7101ae98-10a0-4af2-baca-ccfaf612f58d.mp4

After:

https://user-images.githubusercontent.com/7893468/227743740-8ab0ae0b-2d51-47a1-888b-d30c9f1c8d97.mp4


ivailop7:
@fantactuka could you help me out how to replace a HeadingNode with a TextNode. I tried
```javascript
node.replace(new TextNode('abc'), false);
```
but I get an error that I can only add an Element or a Decorator node to the Root node. What would be the right syntax here?

acywatson:
@ivailop7 you need to wrap the TextNode in a ParagraphNode (or some other element). They can’t exist as direct children of the root.

ivailop7:
> @ivailop7 you need to wrap the TextNode in a ParagraphNode (or some other element). They can’t exist as direct children of the root.

Understood. Thank you @acywatson .

ivailop7:
Updated. The CI failing post the Excalidraw update I believe is because of the parallelization of the tests, thus the sporadic success/failure. I believe it has to be moved from package.json inside the playground to the root one, but needs proper investigation, this is just a theory. Might dig into that one soon.

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -538,14 +538,30 @@ export default function ToolbarPlugin(): JSX.Element {
...
+        nodes.forEach((node, idx) => {
+          // We split the first and last node by the selection
+          // So that we don't format unselected text inside those nodes
           if ($isTextNode(node)) {
+            if (idx === 0 && anchor.offset !== 0) {
+              node = node.splitText(anchor.offset)[1];
+            }
+            if (idx === nodes.length - 1) {
+              node = node.splitText(focus.offset)[0];
+            }
```
zurfyx, line 558: Should we check whether format is already `0` and style is `''`? This will prevent unnecesarily splitting a node

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -538,14 +538,30 @@ export default function ToolbarPlugin(): JSX.Element {
...
             node.setFormat(0);
             node.setStyle('');
             $getNearestBlockElementAncestorOrThrow(node).setFormat('');
-          }
-          if ($isDecoratorBlockNode(node)) {
+          } else if ($isHeadingNode(node) || $isQuoteNode(node)) {
+            node.replace($createParagraphNode(), true);
```
zurfyx, line 563: Is this what's expected out of clear formatting? My hunch is that clear formatting should only clear the above (style + format) but the nodes should be left intact. What makes you think this is the correct behavior? I'm looking at Quip for example

package.json
```diff
@@ -103,6 +103,7 @@
     "@babel/preset-flow": "^7.14.5",
     "@babel/preset-react": "^7.14.5",
     "@babel/preset-typescript": "^7.16.7",
+    "@excalidraw/excalidraw": "^0.14.2",
```
zurfyx, line 106: Let's drop this and package-lock.json, I'm not sure why GH cache is sometimes bad but we can retry when this happens. Definitely something worth investigating but the package should live inside the playground

zurfyx (approved):
Thank you! Adding some minor comments

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
ivailop7: This is the behaviour in Word.

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
ivailop7: This is the behaviour in Word.

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
ivailop7: Very good point, will amend!

ivailop7:
@zurfyx are we good with this one?

zurfyx:
> @zurfyx are we good with this one?

Can you check on the test failures please?

zurfyx:
> Ops, my fault. What's the right order of npm commands to repro these tests locally? When I directly run them, all tests fail, while Unit and Integration pass.

You probably want to run `debug-test-e2e-legacy-chromium` and the other similar variants listed in the package.json. The CI variants work with a different port if that's what you're referring to

ivailop7:
@zurfyx All set here.

ivailop7:
@acywatson @zurfyx I think, this one is ready to merge.

zurfyx:
Thank you Ivaylo, sorry for the delay, I completely missed the updated version

PR #4212, opened 2023-03-27
Unmerge cell

zurfyx:
Unmerge cell feature

https://user-images.githubusercontent.com/193447/227996388-f8ac0da3-09e3-42eb-b440-c5cce778d1de.mov


tylerjbainbridge (approved):

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -90,6 +92,20 @@ function isGridSelectionRectangular(selection: GridSelection): boolean {
   );
 }
 
+function $canUnmerge(): boolean {
+  const selection = $getSelection();
+  if (
+    ($isRangeSelection(selection) && !selection.isCollapsed()) ||
+    (DEPRECATED_$isGridSelection(selection) &&
```
acywatson, line 99: What does our reliance on this here say about our plans to deprecate it?

packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -551,8 +551,67 @@ function $moveSelectionToCell(cell: DEPRECATED_GridCellNode): void {
...
+  const selection = $getSelection();
+  invariant(
+    $isRangeSelection(selection) || DEPRECATED_$isGridSelection(selection),
+    'Expected a RangeSelection or GridSelection',
+  );
+  const anchor = selection.anchor.getNode();
+  const [cell, row, grid] = DEPRECATED_$getNodeTriplet(anchor);
+  const writableCell = cell.getWritable();
+  const colSpan = cell.__colSpan;
+  const rowSpan = cell.__rowSpan;
```
acywatson, line 570: Should we use setters for this outside the core?

I just thought of something else - all the places we use setters and repeatedly call getWriteable in every setter - could we save some overhead by just caching the writeable version of the node for that update cycle? Possible hallucination on my end, just a thought

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
zurfyx: @acywatson - Good point, the GridSelection deprecation plan is still valid. I think we can do the same with NodeSelection. This would have benefits in both terms of size and performance. However, it will take some time to migrate over as there's quite a lot of code that depends on it. I'll experiment with this and share some updates offline, the idea is to have a resolution by end of half.

PR #4216, opened 2023-03-27
Export $hasUpdateTag

nemophrost:


packages/lexical-selection/src/__tests__/utils/index.ts
```diff
@@ -824,7 +824,7 @@ export async function applySelectionInputs(inputs, update, editor) {
                 {
                   clipboardData: {
                     getData: (type) => {
-                      if (type === 'application/x-lexical-nodes') {
+                      if (type === 'application/x-lexical-editor') {
```
zurfyx, line 827: Is this an existing bug? Otherwise, I'd encourage against doing this as it's _kinda_ a breaking change

zurfyx (approved):
LGTM, it seems like it was already explosed in Flow but not exported via TS?

packages/lexical-selection/src/__tests__/utils/index.ts
nemophrost: Yeah @zurfyx this is a bug as far as I can tell. There are no references to `application/x-lexical-nodes` elsewhere in the codebase that I could find and `application/x-lexical-editor` is what is used here:
https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts#L99
https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts#L589

zurfyx: Ah you're right!

Issue #4190, opened 2023-03-24
Bug: Comment popup closes when cursor hovers over table


bencarletonn:
![popupbug](https://user-images.githubusercontent.com/56607084/227593774-b08049e5-5acd-43b1-b0d0-db72348c9f83.gif)

## Steps To Reproduce

1. Select some text, open the comment popup
2.  Hover over table with cursor whilst popup is open

## The current behavior

Popup closes when cursor hovers over table. This is also currently preventing commenting on text embedded inside a table.

## The expected behavior

Popup should remain open when cursor hovers over table.

I did some investigation into this, and it seems that the TableCellResizer mousemove event triggers an editor update when hovering over a table, causing the editor state update listener in the comment plugin to `setShowCommentInput` to false (918 of CommentPlugin.ts). The problem is, TableCellResizer plugin has logic that is dependent on an editor update, i.e. using certain utils functions (`$getNearestNodeFromDOMNode`).

A possible solution would be to refactor the comment plugin so it caches the selection as soon as the comment button is clicked, rather then relying on the `$getSelection()` at each step. I think its important to keep the popup open whilst editor state is updating. Interested to hear what others think

PR #4228, opened 2023-03-28
Bugfix: Comment popup closes when cursor hovers over table

bencarletonn:
Solves the issue raised here: https://github.com/facebook/lexical/issues/4190

- The selection is now temporarily cached inside a ref, `selectionRef` when the comment button is clicked, rather then relying on the current selection using `$getSelection`
- This means changes in selection, caused by editor updates have no effect on the content being wrapped by the Mark node when the comment is being added.
- Able to hover over table with comment popup closing
- Now able to add comments to text inside of tables without the popup closing
- Collapsing selection is now done by the `.selectStart()` and `.selectEnd()` APIs, within the `$wrapSelectionInMarkNode` function

### DIFF
--- a/packages/lexical-mark/src/index.ts
+++ b/packages/lexical-mark/src/index.ts
@@ -124,6 +124,13 @@ export function $wrapSelectionInMarkNode(
       lastCreatedMarkNode = undefined;
     }
   }
+  // Make selection collapsed at the end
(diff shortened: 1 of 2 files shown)


acywatson:
This looks fine to me - sorry for the delay in reviewing

acywatson (approved):

PR #4230, opened 2023-03-28
$cloneNodeDeep

zurfyx:
Copy-paste deep tree hiearchy feature (based on $cloneNode which does shallow node)

_**Also small fix for TextNode which `static clone` only partially clones properties_

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -286,7 +286,12 @@ export class TextNode extends LexicalNode {
   }
 
   static clone(node: TextNode): TextNode {
-    return new TextNode(node.__text, node.__key);
+    const newNode = new TextNode(node.__text, node.__key);
+    newNode.__format = node.__format;
+    newNode.__style = node.__style;
+    newNode.__mode = node.__mode;
+    newNode.__detail = node.__detail;
+    return newNode;
   }
 
   constructor(text: string, key?: NodeKey) {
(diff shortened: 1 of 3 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -286,7 +286,12 @@ export class TextNode extends LexicalNode {
   }
 
   static clone(node: TextNode): TextNode {
-    return new TextNode(node.__text, node.__key);
+    const newNode = new TextNode(node.__text, node.__key);
+    newNode.__format = node.__format;
+    newNode.__style = node.__style;
+    newNode.__mode = node.__mode;
+    newNode.__detail = node.__detail;
```
zurfyx, line 293: Discussed offline with @fantactuka, this needs follow up, this logic is alreay inside `getWritable`. I believe the original idea was to avoid having to deal with unknown internal for nodes like HeadingNode or other custom nodes but `getWritable` isn't really scalable and will also not work when used this way

fantactuka: We can clean up this part: 
https://github.com/facebook/lexical/blob/c0e999d818851c2327a8c04ba089ae85e93f59c8/packages/lexical/src/LexicalNode.ts#L700-L703

Then LexicalNode will only include  core-level props (parent, last, first, size, next, prev), while all other props (that are also serializable into JSON) will be handled within .clone()

thegreatercurve (approved):

mariusplanable:
Could some please share why was this closed?

chrisfrantz:
We were also watching this one. Is this handled in a different issue?

PR #4153, opened 2023-03-19
Bump Excalidraw to v0.14.2

ivailop7:
- The used Excalidraw was over 1 year old (v0.11) and wanted to pick up a number of fixes, including better integration with Vite. The changelog is available here: https://github.com/excalidraw/excalidraw/releases

- Also includes the new Excalidraw UI

Before:
Excalidraw_before_0 11

After:
Excalidraw_after_0 14 2

PR #4231, opened 2023-03-28
Disable CI cache

zurfyx:
Should we disable the cache for now while we investigate the recurrent Excalidraw dependency issue?

```
Error: packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawImage.tsx(9,***7): error TS***307: Cannot find module '@excalidraw/excalidraw' or its corresponding type declarations.
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,13 +32,13 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm i -g npm@8
-      - uses: actions/cache@v3
-        id: cache
-        with:
-          path: |
-            node_modules
-            ~/.cache/ms-playwright
-          key: ${{ runner.os }}-v${{ secrets.CACHE_VERSION }}-${{ hashFiles('package-lock.json') }}
+      # - uses: actions/cache@v3
+      #   id: cache
+      #   with:
+      #     path: |
+      #       node_modules
+      #       ~/.cache/ms-playwright
+      #     key: ${{ runner.os }}-v${{ secrets.CACHE_VERSION }}-${{ hashFiles('package-lock.json') }}
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
@@ -61,13 +61,13 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm i -g npm@8
-      - uses: actions/cache@v3
-        id: cache
-        with:
-          path: |
-            node_modules
-            ~/.cache/ms-playwright
-          key: ${{ runner.os }}-v${{ secrets.CACHE_VERSION }}-${{ hashFiles('package-lock.json') }}
+      # - uses: actions/cache@v3
+      #   id: cache
+      #   with:
+      #     path: |
+      #       node_modules
+      #       ~/.cache/ms-playwright
+      #     key: ${{ runner.os }}-v${{ secrets.CACHE_VERSION }}-${{ hashFiles('package-lock.json') }}
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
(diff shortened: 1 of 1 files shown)


acywatson:
I don't understand how this fixes the problem when changing the cache key doesn't

zurfyx:
> I don't understand how this fixes the problem when changing the cache key doesn't

I believe invalidating the bumping the cache key only works once to have all the subsequent runs to pass (equivalent to deleting all cache entries manually). But as we discussed offline I think there's some post-action storing the bad node_modules after the run

ivailop7:
Since this occurred after my PR, I feel obliged contributing my thoughts on why this occurred after the version bump, but not before. My theory is that there was no clean package-lock.json rebuild, after the excalidraw package was moved to the package.json in the playground folder. 

In my PR, I rebuilt the package-lock.json by deleting the existing package-lock file and node_modules folder and clean npm cache, so the final package-lock.json had a change I found interesting, it moved it from `node_modules/@excalidraw/excalidraw` to `packages/lexical-playground/node_modules/@excalidraw/excalidraw`, if you check the diff file [here](https://github.com/facebook/lexical/pull/4153/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L2370), which is why moving the dependency to the root package.json vs the playground one also fixes it, since it was cached in the right place.

I believe adding a separate cache for the playground node_modules folder would probably solve it too, but not that familiar with the existing testing setup. Hope this is helpful.

acywatson:
> 

Ha! Suspected it was the package-lock all along. Only thing that makes sense. Thanks for chiming in

zurfyx:
> Since this occurred after my PR, I feel obliged contributing my thoughts on why this occurred after the version bump, but not before. My theory is that there was no clean package-lock.json rebuild, after the excalidraw package was moved to the package.json in the playground folder.
> 
> In my PR, I rebuilt the package-lock.json by deleting the existing package-lock file and node_modules folder and clean npm cache, so the final package-lock.json had a change I found interesting, it moved it from `node_modules/@excalidraw/excalidraw` to `packages/lexical-playground/node_modules/@excalidraw/excalidraw`, if you check the diff file [here](https://github.com/facebook/lexical/pull/4153/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L2370), which is why moving the dependency to the root package.json vs the playground one also fixes it, since it was cached in the right place.
> 
> I believe adding a separate cache for the playground node_modules folder would probably solve it too, but not that familiar with the existing testing setup. Hope this is helpful.

Wow, thank you for jumping in! That's actually really useful. I couldn't get my head around it lol - I believe this should work? https://github.com/facebook/lexical/pull/4234

ivailop7:
> > Since this occurred after my PR, I feel obliged contributing my thoughts on why this occurred after the version bump, but not before. My theory is that there was no clean package-lock.json rebuild, after the excalidraw package was moved to the package.json in the playground folder.
> > In my PR, I rebuilt the package-lock.json by deleting the existing package-lock file and node_modules folder and clean npm cache, so the final package-lock.json had a change I found interesting, it moved it from `node_modules/@excalidraw/excalidraw` to `packages/lexical-playground/node_modules/@excalidraw/excalidraw`, if you check the diff file [here](https://github.com/facebook/lexical/pull/4153/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L2370), which is why moving the dependency to the root package.json vs the playground one also fixes it, since it was cached in the right place.
> > I believe adding a separate cache for the playground node_modules folder would probably solve it too, but not that familiar with the existing testing setup. Hope this is helpful.
> 
> Wow, thank you for jumping in! That's actually really useful. I couldn't get my head around it lol - I believe this should work? #4234

Looks good, I think this should solve it and would render this PR unnecessary. Still unsure why a clean npm install moved it.

Issue #4223, opened 2023-03-28
Bug: Floating tool bar not responsive in tables


bencarletonn:
![tablebug](https://user-images.githubusercontent.com/56607084/228257929-c23ad04c-d0f8-45d9-9dca-620a6788104e.gif)

Lexical version: Playground

## Steps To Reproduce

1. Select some text in a table
2. Try to apply a style, or add a comment using the floating toolbar

Link to code example:
I believe this bug was introduced with the release of 0.91. It does not seem to exist in 0.9. 

## The current behavior

- Not able to click on any buttons in the floating toolbar once text is selected inside a table

## The expected behavior

Should be able to apply styles and add comments to selections in tables. I will investigate into this, but I imagine it has been caused due to the large amount of refactoring tables incurred in 0.91. 

After further investigation it seems like the changes in this PR, https://github.com/facebook/lexical/pull/4162/files lead to this problem

PR #4237, opened 2023-03-29
Bugfix: Floating tool bar not responsive in tables

bencarletonn:
Solves the issue raised here: https://github.com/facebook/lexical/issues/4223

- The event in `mouseUpCallback` doesn't bubble up to floating tool bar event handler, so the `.pointerEvent` style is never changed from 'none' to 'auto' meaning the toolbar is not interactive when text is selected within a table cell. 
- When text is selected in table cells, it's treated as a span dom node. We can use this information to check whether we want to propagate the event up to the floating tool bar event handler. 
- If text is not selected in the table cell, but the table cell is clicked on, the event should not bubble up. This is handled in the if statement.

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -145,7 +145,6 @@ export function applyTableHandlers(
     if (event.button !== 0) {
       return;
     }
-
     editor.update(() => {
       const selection = $getSelection();
       const target = event.target;
@@ -176,7 +175,7 @@ export function applyTableHandlers(
   );
 
   const mouseUpCallback = (event: MouseEvent) => {
-    if (isMouseDown) {
+    if (isMouseDown && !doesTargetContainText(event.target as Node)) {
       event.preventDefault();
       event.stopPropagation();
     }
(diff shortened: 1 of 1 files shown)


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -176,7 +176,7 @@ export function applyTableHandlers(
   );
 
   const mouseUpCallback = (event: MouseEvent) => {
-    if (isMouseDown) {
+    if (isMouseDown && !doesTargetContainText(event.target as Node)) {
       event.preventDefault();
       event.stopPropagation();
     }
```
zurfyx, line 182:
```suggestion
```

I wonder what the reason is behind stealing the mouseUp. Or if there's a reason for it it should only be when the selection spans beyond 1 cell (when it's relevant). @tylerjbainbridge, have some context here on whether this would be correct?

zurfyx (commented):
Thank you for taking a stab on this! I think your fix is valid but it might be too specific, if I'm reading this right similar non-text plugins are likely to fail in the future. See the inline comment

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
bencarletonn: > Thank you for taking a stab on this! I think your fix is valid but it might be too specific, if I'm reading this right similar non-text plugins are likely to fail in the future. See the inline comment

Hey @zurfyx , thanks for the review. 

you're right - its definitely specific to tables, as it assumes that text is stored inside a span node. But isn't all the code in LexicalTableSelectionHelpers.ts specific to tables? Similar non-text plugins wont use this set of methods, and wont be affected by the event listeners registered to table elements right?

and, should a table contain a non-text node, wouldn't it be right to assume that we don't want to display the floating tool bar as its very purpose is to apply style to text?

zurfyx (approved):

PR #4242, opened 2023-03-29
Preserve content with merging cells

zurfyx:
https://user-images.githubusercontent.com/193447/228661844-f8b71a49-20c6-415c-ad74-b3c61b40c48e.mov

### DIFF
--- a/packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
@@ -6,7 +6,7 @@
  *
  */
 
-import type {DEPRECATED_GridCellNode} from 'lexical';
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -230,22 +233,37 @@ function TableActionMenu({
...
               lastDescendant.select();
-              isFirstCell = false;
-            } else {
-              nodes[i].remove();
+              firstCell = node;
+              let firstChild;
+              if (
+                node.getChildrenSize() === 1 &&
+                $isParagraphNode((firstChild = node.getFirstChild())) &&
+                firstChild.isEmpty()
```
fantactuka, line 253: Can have some sort of `isEmptyCell` used here and below, we select end of merged cell (not the first one)

fantactuka (approved):

Issue #4243, opened 2023-03-30
Bug: $generateHtmlFromNodes ignores alignment style in case of headings & lists


ionutcirja:
Hello, I have a little issue with generateHtmlFromNodes method.

const html = $generateHtmlFromNodes(editor, null);

Lexical version: 0.9.1

## Steps To Reproduce

1. In an editor I choose a paragraph and a right alignment and the result is: <p dir="ltr" style="text-align: right;"><span>text</span></p>
2. If I change the paragraph to a heading tag or a list item the result is: <h1 class="text-2xl"><span>text</span></h1>.

## The current behavior
As you can observe, the style attribute is missing for headings and lists.

## The expected behavior
The style attribute should be set.

Thank you

khanakia:
I am not able to replicate the issue

https://user-images.githubusercontent.com/2174170/228927874-db4ec447-4496-4e75-b19e-5ad991114ba2.mp4

ionutcirja:
> I am not able to replicate the issue

Inside the editor everything is ok, but try to call $generateHtmlFromNodes(editor, null) and you will be able to see that the style is not added for headings and list.

greeneley:
This issue also happens to me. On the official website’s playground, changing the alignment of heading is fine. However, in my application using the editor, it doesn't work. As indicated by @ionutcirja, inside the editor everything is ok but when calling $generateHtmlFromNodes(editor, null), there is nothing to add the new style for the heading

Currently, I'm using the lexical version 0.8.1

ivailop7:
I fixed the heading one in here: https://github.com/facebook/lexical/pull/4445 and it got released in 0.11. For lists, are you doing whole lists or a list item alignment? Unlikely the list indentation would be modified, I'm looking into supporting alignment for an individual list item, although not sure on what the use case here is.

greeneley:
@ivailop7 Yes, the heading is fixed in 0.11. Thanks for your support.

For list, both whole lists and a list item alignment don't work. 
However, in my opinion, when we align a list, we commonly align whole list instead of a list item alignment

ivailop7:
@greeneley when would you want to align a list? It hasn't come up as a popular use case, thus my questions. This sounds to me more like a triple vertical paragraph (magazine style) alignment thing and having a list in there, vs aligning a list with indentation.

greeneley:
[...]

PR #4250, opened 2023-04-01
heading generateHtmlFromNodes was not adding text-align

khanakia:
generateHtmlFromNodes ignores alignment style in case of headings & lists

Ref.
https://github.com/facebook/lexical/issues/4243


acywatson:
Should be covered here: https://github.com/facebook/lexical/pull/4253

ionutcirja:
When are planning to release this?
Thanks

PR #4252, opened 2023-04-01
Excalidraw Component fixes

ivailop7:
https://user-images.githubusercontent.com/7893468/229314462-892459ed-36d2-4b20-b119-2db9b19fb4b2.mp4

- Clicking outside of the modal was set to close the modal, this meant when you draw and finish the drawing outside of the popup it was closing the popup rather than adding the shape.
- After opening an existing Excalidraw node and drawing something new, if clicking on Discard, then Discard in the popup, it deleted the node, instead of discarding the new changes only. Now it just discards the changes since the last open.


zurfyx (approved):
Thank you!

PR #4253, opened 2023-04-01
Rework generateNodesFromDOM API

acywatson:
I will try to get around to writing up a more extensive explanation later. The main purposes of this are twofold:

1) Improve the out-of-the-box experience for HTML serialization and deserialization in the storage format use-case (as opposed to the transfer/copy-paste use case).

One of the big issues here is that we can't support interpretation of html attributes like background-color or text-align in the core without creating a poor user experience during HTML copy+paste. The idea here is to transfer those semantics via custom attributes on HTML, so we can support HTML-as-a-storage-format (deserializing into an editor with the same configuration) at a higher fidelity while preserving control during copy/paste.

2) Amend the public API to be less cluttered and provide a path for additional options

Ultimately, I'd like to experiment with bringing more of this logic out of the core library and into the editor configuration, possible with some defaults in the rich-text or utils modules. import/exportDOM makes sense on the nodes, but it is honestly pretty clunky to have to do that for just a simple use case like supporting a text style.

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -121,10 +121,27 @@ export type DOMConversion<T extends HTMLElement = HTMLElement> = {
   priority: 0 | 1 | 2 | 3 | 4;
 };
 
+export type DOMConversionOptions = {
+  textStyles?: true;
+  elementFormats?: true;
+  preformatted?: boolean;
+};
+
+export type DOMConversionContext = {
+  editor: LexicalEditor;
+  textStyles?: true;
+  elementFormats?: true;
+  parent?: Node;
+  preformatted?: boolean;
+};
+
 export type DOMConversionFn<T extends HTMLElement = HTMLElement> = (
   element: T,
+  /** @deprecated use context.parent */
   parent?: Node,
+  /** @deprecated use context.preformatted */
   preformatted?: boolean,
+  context?: DOMConversionContext,
 ) => DOMConversionOutput | null;
 
 export type DOMChildConversion = (
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -26,6 +26,7 @@ import type {
   PointType,
   RangeSelection,
 } from './LexicalSelection';
+import type {ElementFormatType} from './nodes/LexicalElementNode';
 import type {RootNode} from './nodes/LexicalRootNode';
 import type {TextFormatType, TextNode} from './nodes/LexicalTextNode';
 
@@ -1590,4 +1591,21 @@ export function $getChildrenRecursively(node: LexicalNode): Array<LexicalNode> {
     }
   }
   return nodes;
+}
+
+export function convertDOMElementLexicalData(
+  element: HTMLElement,
+  lexicalNode: ElementNode,
+): ElementNode {
+  const lexicalDataFormatAlign = element.getAttribute(
+    'lexical-data-format-align',
+  );
+  const lexicalDataIndent = element.getAttribute('lexical-data-indent');
+  if (lexicalDataFormatAlign !== null) {
+    lexicalNode.setFormat(lexicalDataFormatAlign as ElementFormatType);
+  }
+  if (lexicalDataIndent != null) {
+    lexicalNode.setIndent(parseInt(lexicalDataIndent, 10));
+  }
+  return lexicalNode;
 }
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -11,6 +11,7 @@ import type {
   DOMChildConversion,
   DOMConversion,
   DOMConversionFn,
+  DOMConversionOptions,
   GridSelection,
   LexicalEditor,
   LexicalNode,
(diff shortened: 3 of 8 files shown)


fantactuka:
Could you add an example of new API usage?

packages/lexical/src/LexicalNode.ts
```diff
@@ -121,10 +121,27 @@ export type DOMConversion<T extends HTMLElement = HTMLElement> = {
   priority: 0 | 1 | 2 | 3 | 4;
 };
 
+export type DOMConversionOptions = {
+  textStyles?: true;
+  elementFormats?: true;
```
zurfyx, line 126: Does this mean it should be treated like an Element or something?

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1591,3 +1592,20 @@ export function $getChildrenRecursively(node: LexicalNode): Array<LexicalNode> {
...
   return nodes;
 }
+
+export function convertDOMElementLexicalData(
+  element: HTMLElement,
+  lexicalNode: ElementNode,
+): ElementNode {
+  const lexicalDataFormatAlign = element.getAttribute(
+    'lexical-data-format-align',
+  );
```
zurfyx, line 1602: > One of the big issues here is that we can't support interpretation of html attributes like background-color or text-align in the core without creating a poor user experience during HTML copy+paste.

Why is it correct for Lexical to Lexical but not for external users? Did we do some research on how other editors address similar problems? Have we considered an in-between between EditorState and HTML to transfer data between lexical Editors?

zurfyx (commented):
LGTM

> Amend the public API to be less cluttered and provide a path for additional options

I'm unsure how this PR addresses this but I like the long-term vision of moving all the conversion functionality outside of the Core

packages/lexical/src/LexicalUtils.ts
acywatson: Because the semantics of Lexical data storage and transfer shouldn't be coupled to HTML semantics (because they don't map 1:1). This is the root of this issue, imo.

Copy and Paste is one serialization use case, but it's not the only one. Another is simple storage/retrieval. If we rely on HTML semantics to transfer Lexical via imperfect mappings, we end up in a situation where external content can cause issues by inadvertently colliding with those semantics.

For example, an inline background-color style can be added on a span tag on the clipboard in google docs. If Lexical core interprets that as a TextNode style, then every editor will interpret it that way, even if the editor doesn't support those styles. This is why right now, for instance, you can change the alignment of some text in GDocs and then paste it into one of our internal editors and it will be stuck in that alignment.

packages/lexical/src/LexicalNode.ts
acywatson: It controls whether or not you want Lexical to try interpret the element formats and map them to core element formats (indent, align, etc)

acywatson:
> LGTM
> 
> > Amend the public API to be less cluttered and provide a path for additional options
> 
> I'm unsure how this PR addresses this but I like the long-term vision of moving all the conversion functionality outside of the Core

I was thinking because now generateNodesFromDOM will take an options object instead of just continuously adding more arguments.

packages/lexical/src/LexicalUtils.ts
zurfyx: > Because the semantics of Lexical data storage and transfer shouldn't be coupled to HTML semantics

Makes sense, it sounds like we're reusing HTML to build this additional in-between fidelity layer.

> you can change the alignment of some text in GDocs and then paste it into one of our internal editors and it will be stuck in that alignment.

What do you mean by stuck? You can't left-align it afterward?

acywatson: > What do you mean by stuck? You can't left-align it afterward?

Yea, if the editor doesn't support text-alignment (i.e. if it doesn't expose any UI to set the alignment back) then it's just stuck in the center or right alignment. Same with text-color or background-color if those were handled in the core the same way as alignment (they aren't right now).

zurfyx: > Yea, if the editor doesn't support text-alignment (i.e. if it doesn't expose any UI to set the alignment back) then it's just stuck in the center or right alignment. Same with text-color or background-color if those were handled in the core the same way as alignment (they aren't right now).

Have we considered a feature system? https://github.com/facebook/lexical/pull/4254#issuecomment-1495960514

zurfyx (approved):

acywatson:
> I like the concept here.
> 
> QQ, why is `convertDOMElementLexicalData()` only for element nodes?

Good question - it's because all ElementNode subclasses don't need this logic, so I couldn't very easily put it in the base class. Instead, I put it in a util to avoid repetition. Will look closer at this when I have time to take another pass.

acywatson:
> Could you add an example of new API usage?

yes, I will

fantactuka (approved):

ivailop7 (approved):

thegreatercurve:
@acywatson Mind resolving the merge conflicts?

n1215:
I'm eagerly awaiting this feature.
Will it be in the 0.13.0 release?

etrepum:

Issue #3833, opened 2023-02-06
Proposal: Lexical Schemas


tylerjbainbridge:
# Lexical Schemas

### The Problem

A lot of mistakes with Lexical are caused by developers adding nodes where they shouldn’t. 

Schemas are an API proposal that’s designed to add guardrails around the insertion of nodes. They'll check if a given relationship (parent, child, sibling) is valid and allow you to normalize it if possible.

Slate.js solves this elegantly with their [Schema](https://docs.slatejs.org/v/v0.47/guides/schemas) (https://docs.slatejs.org/v/v0.47/guides/schemas) API and developers find it intuitive, so it makes a lot of sense to propose a similar API that’s familiar to users of Slate and other Rich Text solutions.

### The Solution

We’ll stick with our current approach of encapsulating node-related logic on the class itself by introducing a new static method, getSchema, which returns a Lexical Schema.

The schema’s relation methods will be called at different points within the core insertion logic and aim to remove the many çase-specific patches and node methods throughout the codebase while catching bugs before hitting production.

If any of the validations fail it will warn the developer in the console and we’ll provide a normalize method that’ll act as an escape hatch and let you to fix the issue.

### Example

*Take the following Table Row example* 
A row should never not be a child of a table, it should never have a child that isn’t a cell, and it should never be a sibling that isn’t another row. This is expressed almost verbatim in code below.

```js
class TableRowNode extends DEPRECATED_GridRowNode {
  static getSchema(): Schema<TableRowNode> {
    return {
      relations: {
        parent(a: TableRowNode, parent: LexicalNode): boolean {
          return $isTableNode(parent);
        },
        child(a: TableRowNode, child: LexicalNode): boolean {
          return $isTableCellNode(child);
        },
        sibling(a: TableRowNode, sibling: LexicalNode): boolean {
          return $isTableRowNode(sibling);
        },
      },
      normalize(node: TableRowNode, error: LexicalSchemaError, editor: LexicalEditor) {
       if (error.type === 'invalid_parent') {
         // try to repair
       }
      }
    }
  }
}

// For Product specific validations an API like this could also be useful.
$extendSchema(TableRowNode, {
    relations: {
        parent(a: TableRowNode, parent: LexicalNode): boolean {
            // something
        },
        child(a: TableRowNode, child: LexicalNode): boolean {
            // something
        },
        sibling(a: TableRowNode, sibling: LexicalNode): boolean {
            // something
        },
    },
[...]

PR #4254, opened 2023-04-03
Allow overriding HTML serialization behavior from the editor config.

acywatson:
With this change, I'm adding a new path for changing or extending html serialization functionality in the editor. Currently, if you want to change HTML serialization behavior on the core nodes, you generally need to override the node, which I think is an unnecessarily confusing process, given how common the use case is.

### DIFF
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -92,7 +92,15 @@ function $appendNodesToHTML(
     target = clone;
   }
   const children = $isElementNode(target) ? target.getChildren() : [];
-  const {element, after} = target.exportDOM(editor);
+  const registeredNode = editor._nodes.get(target.getType());
+  let exportOutput;
+  if (registeredNode && registeredNode.exportDOM !== undefined) {
+    exportOutput = registeredNode.exportDOM(editor, target);
+  } else {
+    exportOutput = target.exportDOM(editor);
+  }
+
+  const {element, after} = exportOutput;
 
   if (!element) {
     return false;
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -94,6 +94,8 @@ function $appendNodesToHTML(
   const children = $isElementNode(target) ? target.getChildren() : [];
   const registeredNode = editor._nodes.get(target.getType());
   let exportOutput;
+
+  // Use HTMLConfig overrides, if available.
   if (registeredNode && registeredNode.exportDOM !== undefined) {
     exportOutput = registeredNode.exportDOM(editor, target);
   } else {
--- a/packages/lexical-react/flow/LexicalComposer.js.flow
+++ b/packages/lexical-react/flow/LexicalComposer.js.flow
@@ -13,6 +13,7 @@ import type {
   LexicalNode,
   EditorState,
   LexicalNodeReplacement,
+  HTMLConfig,
 } from 'lexical';
 
 export type InitialEditorStateType =
@@ -29,6 +30,7 @@ export type InitialConfigType = $ReadOnly<{
   theme?: EditorThemeClasses,
   editorState?: InitialEditorStateType,
   onError: (error: Error, editor: LexicalEditor) => void,
+  html?: HTMLConfig,
 }>;
 
 type Props = {
--- a/packages/lexical-react/src/LexicalComposer.tsx
+++ b/packages/lexical-react/src/LexicalComposer.tsx
@@ -19,6 +19,7 @@ import {
   createEditor,
   EditorState,
   EditorThemeClasses,
+  HTMLConfig,
   Klass,
   LexicalEditor,
   LexicalNode,
@@ -45,6 +46,7 @@ export type InitialConfigType = Readonly<{
   editable?: boolean;
   theme?: EditorThemeClasses;
   editorState?: InitialEditorStateType;
+  html?: HTMLConfig;
 }>;
 
 type Props = {
@@ -62,6 +64,7 @@ export function LexicalComposer({initialConfig, children}: Props): JSX.Element {
         nodes,
         onError,
         editorState: initialEditorState,
+        html,
       } = initialConfig;
 
       const context: LexicalComposerContextType = createLexicalComposerContext(
@@ -74,6 +77,7 @@ export function LexicalComposer({initialConfig, children}: Props): JSX.Element {
       if (editor === null) {
         const newEditor = createEditor({
           editable: initialConfig.editable,
+          html,
           namespace,
           nodes,
           onError: (error) => onError(error, newEditor),
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -276,6 +276,14 @@ export type LexicalNodeReplacement = {
   replace: Class<LexicalNode>,
   with: (node: LexicalNode) => LexicalNode,
   withKlass?: Class<LexicalNode>,
+}
+
+export type HTMLConfig = {
+  export?: Map<
+    Class<LexicalNode>,
+    (editor: LexicalEditor, target: LexicalNode) => DOMExportOutput,
+  >,
+  import?: DOMConversionMap,
 };
 
 declare export function createEditor(editorConfig?: {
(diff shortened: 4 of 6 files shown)


zurfyx (commented):
Can you elaborate on why we should prefer this over replacement classes. This is diverging from the one way to do things approach and can likely lead into more single individual methods being ovewritten in a similar fashion.

acywatson:
> Can you elaborate on why we should prefer this over replacement classes? The proposed appraoch diverges from the one way to do things approach and can likely lead into more single individual methods being ovewritten in a similar fashion.

Yea, I covered it in the PR description, but if you look at the feedback and questions we have gotten/are getting around HTML serialization, we have a proliferation of issues around minor tweaks people want to do to how core nodes are imported/exported to/from HTML. 

So, to cover the simple use case of adding support for backgroundColor (for instance) in in HTML serialization, you need to 1) learn about node overrides
2) create your node class (which means you need to learn about getType and clone, etc), then 
3) figure out how to change export/importDOM on that node to do what you want. importDOM in particular is not easy to extend - you'd typically need to copy over the logic to a new conversion to add support for a single property.


> the one way to do things approach and can likely lead into more single individual methods being ovewritten in a similar fashion.

I'm not sure that this is really a problem in this context. Can you tell me the practical problem you would see with their being a simpler way for basic extension and a more flexible way for more advanced use cases?

zurfyx:
> I'm not sure that this is really a problem in this context. Can you tell me the practical problem you would see with their being a simpler way for basic extension and a more flexible way for more advanced use cases?

I see what you mean, my first intuition would be to evaluate whether it's possible to build a higher-level API on top of node replacement. Type-safety might not make it easy though.

Second, but that's way beyond this PR, Nodes require heavy configuration, we should simplify them. A 5-liner should be sufficient to get you started with your own Node.

Third, for TextNode and ElementNode which are the most flexible and also built-in into the Core we could consider `features` where users can seamlessly toggle what they need. As basic Nodes are more and more powerful we will encounter more and more cases of users only wanting a subset of features.

That said, I'm pretty sure you've thought more about this space than I have so I've no objections towards merging this, can we get some thoughts from other folks too? @fantactuka @thegreatercurve @tylerjbainbridge

acywatson:
> Second, but that's way beyond this PR, Nodes require heavy configuration, we should simplify them. A 5-liner should be sufficient to get you started with your own Node.

I particularly like this

I was thinking that reviving and merging @EgonBolton's clone and serialization PRs would be a big step in the right direction.

acywatson:
> Third, for TextNode and ElementNode which are the most flexible and also built-in into the Core we could consider `features` where users can seamlessly toggle what they need. As basic Nodes are more and more powerful we will encounter more and more cases of users only wanting a subset of features.

With this one, I wonder what it would mean to have "features" toggled on or off. I thought about this a bit, but it's hard to rationalize what it would mean to turn something "on" or "off" when there are so many other ways to control that thing through other APIs.

acywatson:
> I see what you mean, my first intuition would be to evaluate whether it's possible to build a higher-level API on top of node replacement. Type-safety might not make it easy though.

Interesting - what could this look like?

zurfyx:
> With this one, I wonder what it would mean to have "features" toggled on or off. I thought about this a bit, but it's hard to rationalize what it would mean to turn something "on" or "off" when there are so many other ways to control that thing through other APIs.

I don't think this addresses the entire problem but it can work for a subset. For example, you can define a TextPlugin

`<TextPlugin hasStrikethrough={false} />` or `<TablePlugin hasCellMerging={true} />`

The plugin will prevent and "normalize" bad features. We can even go further and build it into the Node with some invariants but I'm hesitant on introducing additional layer of inheritance (this is one of the flaws of the prototype model).

> Interesting - what could this look like?

Probably not a good idea. I don't think it's wise to hide the implementation details behind the class. Not only they should be well aware of its existance but also they will need a reference to it and a `$isXNode` function. In this case, I'd double down on the idea of simplifying node creation and/or codegen Nodes (since most of it is copy-paste anyway).

acywatson:
> The plugin will prevent and "normalize" bad features.

Via transforms, I guess? How would you implement this?

zurfyx:
> Via transforms, I guess? How would you implement this?

Yes, that's what I had in mind, even though there might be some UX trade-offs at points.

abelsj60:
> > Via transforms, I guess? How would you implement this?
> 
> Yes, that's what I had in mind, even though there might be some UX trade-offs at points.

This is just a small comment: 

1. This is a very interesting conversation.
    - @acywatson, I like the way you're thinking here. I just answered a question on Discord where the tediousness of having to change every single core node made one idea seem "correct," but exhausting. For what that's worth.
3. This reminds me of that schema discussion that was had at some point. Whenever you get around to focusing on these ideas more, maybe that conversation informs this one? Or vice versa? Just a thought.
    - https://github.com/facebook/lexical/issues/3833

GermanJablo:
> Second, but that's way beyond this PR, Nodes require heavy configuration, we should simplify them. A 5-liner should be sufficient to get you started with your own Node.
> 
> > I particularly like this
> >
> > I was thinking that reviving and merging @GermanJablo's clone and serialization PRs would be a big step in the right direct

Thanks for pointing this out. By the way, what's the status of this?

> I see what you mean, my first intuition would be to evaluate whether it's possible to build a higher-level API on top of node replacement. Type-safety might not make it easy though.
> ...
> Third, for TextNode and ElementNode which are the most flexible and also built-in into the Core we could consider features where users can seamlessly toggle what they need. As basic Nodes are more and more powerful we will encounter more and more cases of users only wanting a subset of features.

I'm going to address points 1 and 3 here.

The problem with making features optional is that it increases the surface area of the API and does not solve the problem that customizing something becomes difficult or impossible.

After thinking about it a lot, I've only found two ways to **_flexibly_** extend the behavior of a node: mixins and prototypes.

**Prototypes**: Extending the prototype basically consists of doing what @fantactuka suggested [in this comment](https://github.com/facebook/lexical/issues/1262#issuecomment-1036438842). I've tested it on small examples and it seemed to work. I think it might work well with Typescript if declare global is used, But I don't know how composable it is (in terms of adding more of a "component" or "decorator" functionality to a node).

**Mixins**: The second alternative, which is what I'm using, is to use mixins. It took me a long time to make it typesafe, but I finally did, even making type guards of the type `$isNodeX` (or rather `$hasXMixin`), which is great. I showed an [example on Discord](https://discord.com/channels/953974421008293909/1136747303177306275/1143199226726195362).

The only problem with mixins so far is that static methods have to be repeated. My PRs removing `clone` and `importJSON` would make this easier. In the example I gave I also showed how to override `getType`.

`Indent` and `format` are two use cases that, for example, would bring many benefits to be implemented with this pattern. That is what I wanted to explain in this issue: https://github.com/facebook/lexical/issues/4528

acywatson:
> Thanks for pointing this out. By the way, what's the status of this?

This will be the next thing that I work on in Lexical. I was going to checkout your PR and fix all those conflicts for you because I feel so bad about the delay. Then I think we are good to merge.

> Mixins: The second alternative, which is what I'm using, is to use mixins. It took me a long time to make it typesafe, but I finally did, even making type guards of the type $isNodeX (or rather $hasXMixin), which is great. I showed an [example on Discord](https://discord.com/channels/953974421008293909/1136747303177306275/1143199226726195362).

This is cool...then we could vend mixins for things like this. You'd still have to use node overrides API though, I suppose.

GermanJablo:
> This will be the next thing that I work on in Lexical. I was going to checkout your PR and fix all those conflicts for you because I feel so bad about the delay. Then I think we are good to merge.

haha thanks. If I find time I can update it, although I don't think I can this week.

> This is cool...then we could vend mixins for things like this. You'd still have to use node overrides API though, I suppose.

Exactly. The nodes on which you apply the mixin are the ones that you register in the editor, using the replacement API.

acywatson:
> > This will be the next thing that I work on in Lexical. I was going to checkout your PR and fix all those conflicts for you because I feel so bad about the delay. Then I think we are good to merge.
> 
> haha thanks. If I find time I can update it, although I don't think I can this week.
> 
> > This is cool...then we could vend mixins for things like this. You'd still have to use node overrides API though, I suppose.
> 
> Exactly. The nodes on which you apply the mixin are the ones that you register in the editor, using the replacement API.

I'm gonna merge this, but I really like the mixin idea - let me think about how best to carry that forward.

moy2010:
Just gave it a try on Lexical 0.12.4 by passing an initialConfig to a `LexicalComposer`, and didn't see any transformation on the HTML content.

Is this feature live now?

Issue #4187, opened 2023-03-23
Bug: If the top node is a Collapsible Section, there is no way to append text before it


ivailop7:
https://user-images.githubusercontent.com/7893468/227382869-6f4838f8-da73-4601-afcb-c3d2c36bdf5b.mp4

Lexical version: 0.9

## Steps To Reproduce

1. Clear all the editor contents
2. Append a Collapsible Section (the only node)

## The current behavior

Since the node is at the top, there is no way to position the cursor above it (or on the same line before it) to append text before the section. As a workaround, one can type below it and then use the drag-and-drop plugin to push the paragraph above the section to gain the ability to write above. 

## The expected behavior

Having looked at how we handle this for Images and Polls: we allow to position the cursor on the same line before the image, so the user can do newline, which is fine, since they don't take direct text input and don't occupy the whole row. This is an option, but since the collapsible section takes the whole row, ie we need to squeeze the width of the section if we type text next to it or reflow it to a new line and might also create confusion with typing in the section header.

A few ideas:
1) Append an empty paragraph node before the Collapsible Section when it is the top node in the editor. If someone wants to delete the empty line, they can do so manually.
2) Allow to position the cursor on the same line as the section (at the start of the line) and upon any typed text move the section to the next line.

ivailop7:
After playing with other nodes, I noticed the answer exists in the YouTube Node - when a YouTube node is the top node, the cursor becomes an underline when the above node is not a text one, respectively the same if the node below is not text-based either. This is ideal, I'll replicate this behaviour for the collapsible section too.

Example below:

https://user-images.githubusercontent.com/7893468/227804202-fdd71d28-fad8-495c-a709-b07da37ff5e4.mp4

ivailop7:
I've narrowed it down I need to update `updateDOMBlockCursorElement` but not entirely sure how to solve this in a node-type agnostic way, would need some guidance, since the main wrapper (the container node) is not directly selectable, which I believe is why the existing logic doesn't kick in, beyond the lack of "inInline" and "canBeEmpty" not being defined as False in the node definitions.

khanakia:
I was checking how Notion.js handles this basically there are 2 ways
1. SHIFT + ENTER creates a new node below
2. Clicking on the Empty area at the bottom of the editor simply inserts a new paragraph node

ivailop7:
> I was checking how Notion.js handles this basically there are 2 ways
> 
> 1. SHIFT + ENTER creates a new node below
> 
[...]

PR #4257, opened 2023-04-03
Collapsible enhancements

fantactuka:
Better handling for up/down/left/right arrows to escape collapsible when it's the last/first element in the root/shadowroot

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -309,6 +309,7 @@ test.describe('Selection', () => {
           <span data-lexical-text="true">abc</span>
         </p>
         <p class="PlaygroundEditorTheme__paragraph"><br /></p>
+        <p class="PlaygroundEditorTheme__paragraph"><br /></p>
       `,
     );
   });
--- a/packages/lexical-playground/src/plugins/CollapsiblePlugin/index.ts
+++ b/packages/lexical-playground/src/plugins/CollapsiblePlugin/index.ts
@@ -9,21 +9,28 @@
 import './Collapsible.css';
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
-import {$findMatchingParent, mergeRegister} from '@lexical/utils';
+import {
+  $findMatchingParent,
+  $insertNodeToNearestRoot,
+  mergeRegister,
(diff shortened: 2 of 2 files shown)


fantactuka:
> Hey @fantactuka, great stuff. Wanted to get your thoughts on #4187 on using a block cursor vs adding paragraphs or it's possible I've misunderstood the changes. Feel free to close this issue if you think it makes sense.

As of now block cursor is only applicable to decorators and won't help much with elements since element node that has children won't receive selection (unlike decorators).

package-lock.json
zurfyx, line 1: Delete?

zurfyx (approved):

ivailop7:
I discovered a small bug when playing with those changes, when it's the top node and the empty paragraph node is delete using "delete" key, it deletes the collapsible section, but if "backspace" is used it deletes correctly. Video below:
First deletion is using "delete", second deletion is using "backspace"

https://user-images.githubusercontent.com/7893468/230738957-393a9f38-e212-48cb-ac58-39c87516f69e.mp4

fantactuka:
> I discovered a small bug when playing with those changes, when it's the top node and the empty paragraph node is deleted using "delete" key, it deletes the collapsible section, but if "backspace" is used it deletes the empty paragraph correctly. Video below:

I think it's a general issue, e.g., adding with Heading instead of collapsible and trying to press delete while on line just above it, it'll replace heading with regular paragraph

ivailop7:
> > I discovered a small bug when playing with those changes, when it's the top node and the empty paragraph node is deleted using "delete" key, it deletes the collapsible section, but if "backspace" is used it deletes the empty paragraph correctly. Video below:
> 
> 
> 
> I think it's a general issue, e.g., adding with Heading instead of collapsible and trying to press delete while on line just above it, it'll replace heading with regular paragraph

You are right. To tackle for another day.

ivailop7 (approved):
LGTM

fantactuka:
> > I discovered a small bug when playing with those changes, when it's the top node and the empty paragraph node is deleted using "delete" key, it deletes the collapsible section, but if "backspace" is used it deletes the empty paragraph correctly. Video below:
> 
> I think it's a general issue, e.g., adding with Heading instead of collapsible and trying to press delete while on line just above it, it'll replace heading with regular paragraph

#4380 - addresses the issue

PR #4260, opened 2023-04-03
Add try catch around katex renderer

### LINKED ISSUE #4185: Feature: Better error handling for Katex rendering
Not sure whether to report this as a bug or as a feature because it isn't a problem with the Lexical source code itself. Inserting an equation throws an error on invalid Latex input when writing block-style math—`\begin{\array}`. The correct syntax is `\begin{array}`. The same error occurs for `\begin{\bmatrix}` or `\begin{\pmatrix}`. 

Lexical version: 0.9.0

## Steps To Reproduce

1. Click "Insert" in the Lexical Playground's toolbar.
2. Click "Equation."
3. Uncheck inline. 
4. Type: `\begin{\array}`
5. An error is thrown on the final right-brace. 

https://user-images.githubusercontent.com/87391431/227339979-19b3d558-b1fe-4cb8-af92-8a5a34deb25b.mov

The same error occurs even if the "Inline" option is set. Out of curiosity, I tried it with the Markdown transformer (this is still invalid syntax):

https://user-images.githubusercontent.com/87391431/227342060-3f4d7a12-4deb-45cd-8c06-f7d7eb6f5826.mov

## The current behavior
Error thrown on invalid Latex input.

## The expected behavior
Something less obtrusive. E.g., coloring the input red.

Currently, [Katex has a pending issue](https://github.com/KaTeX/KaTeX/issues/3760) for throwing an error on this particular mistake, even if the rendering option `throwOnError` is set to `false`. A possible fix might be some sort of try-catch block in [KatexRenderer.tsx](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/ui/KatexRenderer.tsx), but I don't know the source code well enough to recommend it.

thegreatercurve:
This is a PR to make sure that Katex doesn't blow up the editor page by throwing an uncaught error. 

Closes #4185

### DIFF
--- a/packages/lexical-playground/src/ui/KatexRenderer.tsx
+++ b/packages/lexical-playground/src/ui/KatexRenderer.tsx
@@ -25,14 +25,18 @@ export default function KatexRenderer({
     const katexElement = katexElementRef.current;
 
     if (katexElement !== null) {
-      katex.render(equation, katexElement, {
-        displayMode: !inline, // true === block display //
-        errorColor: '#cc0000',
-        output: 'html',
-        strict: 'warn',
-        throwOnError: false,
-        trust: false,
-      });
+      try {
+        katex.render(equation, katexElement, {
+          displayMode: !inline, // true === block display //
+          errorColor: '#cc0000',
+          output: 'html',
+          strict: 'warn',
+          throwOnError: false,
+          trust: false,
+        });
+      } catch (_) {
+        // Fail silently
+      }
     }
   }, [equation, inline]);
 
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/ui/KatexRenderer.tsx
```diff
@@ -25,14 +25,18 @@ export default function KatexRenderer({
...
+        katex.render(equation, katexElement, {
+          displayMode: !inline, // true === block display //
+          errorColor: '#cc0000',
+          output: 'html',
+          strict: 'warn',
+          throwOnError: false,
+          trust: false,
+        });
+      } catch (_) {
+        // Fail silently
```
zurfyx, line 38: Should this error be bubbled up somehow, or show some UI saying it errored via ErrorBoundary?

zurfyx (approved):

PR #4263, opened 2023-04-04
fix get text selectRect error

yjhtry:
`getBoundingClientRect`is not a function when`selectionTarget` is of type`Text`, So just by judging `selectionTarget! == null` will not avoid this error

image

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -3084,8 +3084,10 @@ export function updateDOMSelection(
         : domSelection.rangeCount > 0
         ? domSelection.getRangeAt(0)
         : null;
-    if (selectionTarget instanceof HTMLElement) {
-      // @ts-ignore Text nodes do have getBoundingClientRect
+    if (selectionTarget instanceof Text) {
+      const range = document.createRange();
+      range.select(selectionTarget);
+      selectionRect = range.getBoundingClientRect();
       const selectionRect = selectionTarget.getBoundingClientRect();
       scrollIntoViewIfNeeded(editor, selectionRect, rootElement);
     }
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -3084,7 +3084,7 @@ export function updateDOMSelection(
         : domSelection.rangeCount > 0
         ? domSelection.getRangeAt(0)
         : null;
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalSelection.ts
```diff
@@ -3084,7 +3084,7 @@ export function updateDOMSelection(
         : domSelection.rangeCount > 0
         ? domSelection.getRangeAt(0)
         : null;
-    if (selectionTarget !== null) {
+    if (selectionTarget instanceof HTMLElement) {
       // @ts-ignore Text nodes do have getBoundingClientRect
```
zurfyx, line 3088: While ignoring `scrollIntoView` will work I would be inclined to fix it as this logic is intended to run.

1) Range

```suggestion
    if (selectionTarget instanceof Text) {
      const range = document.createRange();
      range.select(selectionTarget);
      selectionRect = range.getBoundingClientRect();
```
2) HTMLElement wrapper assumption

We can assume that each Text is wrapped by an HTMLElement that consist is only one Text. I'm confident this should be true 100% of the cases in Lexical

zurfyx:
Thank you! Can you please expand on your test plan and how you repro'ed the issue? I don't think I've come across this issue before when typing on Lexical

yjhtry:
> Thank you! Can you please expand on your test plan and how you repro'ed the issue? I don't think I've come across this issue before when typing on Lexical

Here's a recurring example [https://codesandbox.io/p/sandbox/relaxed-river-h6oe24](https://codesandbox.io/p/sandbox/relaxed-river-h6oe24)

I'm not sure it has anything to do with this setting
image

Here's the replay video
This problem occurs when I try to select `00:331` and This problem also occurs when the mouse is accidentally crossed 

https://user-images.githubusercontent.com/39440537/229801327-75a9343b-1182-4768-b655-b20cbe11360a.mp4

zurfyx:
The closest we have (built-in) are token nodes but I don't think they're great for your use case either. In any case, it shouldn't crash. The code I provided is pseudocode, you will have to adjust it a bit

yjhtry:
> The closest we have (built-in) are token nodes but I don't think they're great for your use case either. In any case, it shouldn't crash. The code I provided is pseudocode, you will have to adjust it a bit

Yes, I've adjusted

zurfyx (approved):

zurfyx:
Thank you, LGTM!

Issue #4265, opened 2023-04-05
Bug: Deleting a table row causes a crash


michaelsoultanidis:
## Description
Deleting the last table row causes the following crash inside `$deleteTableRow_EXPERIMENTAL`:
```
main.788dd689.js:21 Uncaught Error: Expected GridNode childAtIndex(5) to be RowNode
    at jj.$deleteTableRow__EXPERIMENTAL (main.788dd689.js:21:426261)
    at main.788dd689.js:21:1208697
    at eh (main.788dd689.js:21:189961)
    at th (main.788dd689.js:21:192102)
    at Dh.update (main.788dd689.js:21:213346)
    at main.788dd689.js:21:1208684
    at onClick (main.788dd689.js:21:1211094)
    at Object.Ct (main.788dd689.js:21:16658)
    at Ot (main.788dd689.js:21:16812)
    at main.788dd689.js:21:36721
```
<img width="1430" alt="Screenshot 2023-04-05 at 1 51 44 PM" src="https://user-images.githubusercontent.com/88332746/230060151-28ac05d1-dd05-40b4-a13f-393e88d4a1fc.png">

Lexical version: Latest

## Steps To Reproduce

1. Visit the Lexical Playground app
2. Insert a 5x5 table(not the experimental)
3. Try to delete the first cell of the last row

khanakia:
I am not able to reproduce the issue


https://user-images.githubusercontent.com/2174170/230070779-efe130fd-62fa-4a73-9df7-d3393466e5f6.mov

michaelsoultanidis:
> I am not able to reproduce the issue

The issue is with the non-experimental table

PR #4270, opened 2023-04-05
last table row delete throwing error

khanakia:
Bug: Deleting a table row causes a crash

Fixes https://github.com/facebook/lexical/issues/4265


zurfyx:
Thank you for fixing this!

PR #4272, opened 2023-04-05
Table/TableRow should not be a shadow root

fantactuka:
Shadow root indicates that node can be a parent for "regular" rich text content so it only makes sense for TableCellNode be a shadow root, while table and table row should be regular elements

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -1518,7 +1518,7 @@ test.describe('Tables', () => {
 
     await focusEditor(page);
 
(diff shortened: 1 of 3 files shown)


zurfyx:
>   1 failed
    [chromium] › packages/lexical-playground/__tests__/e***e/Tables.spec.mjs:1463:3 › Tables › Insert column before (with selected cell with rowspan > 1) 


Huh, very interesting 👀 I can help look into this tomorrow if you want

zurfyx (approved):

fantactuka:
Will need more adjustments here, With this change you can forward delete table cell-by-cell, which breaks tables cell structure.

PR #4275, opened 2023-04-06
Allow excluding specific properties from collab syncing via plugin

### LINKED ISSUE #3670: Feature: make excluded properties from yjs sync configurable
I need to be able to exclude specific fields from yjs sync, because i want to centralize logic related to the node in the node class not just the node's data

I see there a Set called `excludedProperties` defined statically in lexical code https://github.com/facebook/lexical/blob/4dae4fb39d1ba6e4cc181b22652bfeed72a67e34/packages/lexical-yjs/src/Utils.ts#L42-L52
I mainly want a way to be able to add more filed in that set

the alternative solution is to keep a map that observe the nodes create through a mutation listener and just keep the data i don't want to sync there

I can make a PR if it is okay to implement this feature

acywatson:
Moves excluded properties to the Binding to avoid sharing them across plugin instances, allows scoping ignored properties to the node, reducing the likelihood of collisions.

Fixes #3670

### DIFF
--- a/packages/lexical-react/src/shared/useYjsCollaboration.tsx
+++ b/packages/lexical-react/src/shared/useYjsCollaboration.tsx
@@ -31,6 +31,7 @@ import {
   REDO_COMMAND,
   UNDO_COMMAND,
 } from 'lexical';
+import {ExcludedProperties} from 'packages/lexical-yjs/src/Bindings';
 import * as React from 'react';
 import {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {createPortal} from 'react-dom';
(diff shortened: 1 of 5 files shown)


packages/lexical-yjs/src/Utils.ts
```diff
@@ -243,6 +244,15 @@ export function syncPropertiesFromYjs(
     if (excludedProperties.has(property)) {
       continue;
     }
+    const additionalExcludedProperties = binding.excludedProperties.get(
+      lexicalNode.constructor as Klass<LexicalNode>,
+    );
+    if (
+      additionalExcludedProperties !== undefined &&
+      additionalExcludedProperties.has(property)
```
fantactuka, line 252: Can be a nice util to be reused below

fantactuka (approved):
🚀

packages/lexical-yjs/src/Utils.ts
```diff
@@ -243,6 +244,15 @@ export function syncPropertiesFromYjs(
...
+    ) {
+      continue;
+    }
```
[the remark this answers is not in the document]
acywatson: Yea, that sort of makes sense - my thought was just that it's best to be resilient to that type of thing by checking in each place. I don't mind taking it out, just concerned about unforeseeable failure states.

packages/lexical-react/src/shared/useYjsCollaboration.tsx
```diff
@@ -31,6 +31,7 @@ import {
   REDO_COMMAND,
   UNDO_COMMAND,
 } from 'lexical';
+import {ExcludedProperties} from 'packages/lexical-yjs/src/Bindings';
```
trueadm, line 34: This references “src”?

acywatson: ugh, auto-import...good catch. Wonder why integrity job passed...

PR #4276, opened 2023-04-06
Replace canExtractContents with ShadowRoot check

zurfyx:
Manually tested deleteForward with a table


fantactuka:
Should we kill the method as well and make it a part of next version bump?

zurfyx:
> Should we kill the method as well and make it a part of next version bump?

Guess I have to wait until 0.10? Is that what you mean?

PR #4284, opened 2023-04-07
Generalize Menu Logic

tylerjbainbridge:
Context-- 
Typeaheads were created and then some parts of the code were reused/adapted for the AutoEmbed menus. This led to a lot of extra code in the `LexicalTypeaheadMenuPlugin.tsx`.

So in preparation for the upcoming `LexicalContextMenuPlugin` I've generalized the core menu logic into `LexicalMenus.ts` and separated the `LexicalNodeMenuPlugin` that's used by Embeds into it's own file.

Nothing functionally changed here. Things were just moved around/renamed to be more clear.
 
You can see this in action on the [`new-context-menu-v1` branch](https://github.com/facebook/lexical/pull/4285) which will have PRs coming soon.


zurfyx:
LGTM, but I don't think we can merge this just yet as removing modules is a breaking change. Do you want to keep the old modules for now or you want to prioritize the next Lexical version?

vignesh-gupta:
LGTM

zurfyx (approved):

leolorenzoluis:
@tylerjbainbridge Great it broke the < .10 version because of this rename no value added besides renames.

PR #4294, opened 2023-04-10
Fix insertBefore/insertAfter types

fantactuka:
Here generic will actually make sense


zurfyx (approved):

acywatson (approved):
Makes sense - update the callsite and should be good

PR #3959, opened 2023-02-23
Fix: Floating toolbar flashes on hover over empty lines when the selection is backwards.

AlessioGr:
This fixes #3727 by removing pointer events from the floating toolbar while you're holding the mouse down button.

PR #4298, opened 2023-04-10
Improve responsiveness of the floating text format toolbar buttons

ChronicLynx:
Introduced a movement threshold so button clicks with slight motion to them aren't filtered out. The current behavior will ignore all clicks to the toolbar with even the slightest of mouse movement, causing the toolbar to feel unresponsive.

This was broken with the fix introduced in #3959. My fix should help to improve the feel of the buttons while still mitigating the toolbar flashes on drag to highlight.

Current Behavior:

https://user-images.githubusercontent.com/29527680/231000297-d6c2cb05-dc8a-4479-b8c7-32e4c3066d45.mov

---

Fixed Behavior:

https://user-images.githubusercontent.com/29527680/231000315-ad7d6c40-2c45-4f8b-88e1-614efdafadc9.mov

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
@@ -67,14 +67,30 @@ function TextFormatFloatingToolbar({
     editor.dispatchCommand(INSERT_INLINE_COMMAND, undefined);
   };
 
+  const movementThreshold = 5; // Set a threshold for minimum movement in pixels
+
+  let initialPosition = {x: 0, y: 0};
+
+  function mouseDownListener(e: MouseEvent) {
+    initialPosition = {x: e.clientX, y: e.clientY};
+  }
+
   function mouseMoveListener(e: MouseEvent) {
     if (
       popupCharStylesEditorRef?.current &&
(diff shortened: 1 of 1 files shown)


ivailop7 (approved):

zurfyx:
Thank you, mind expanding on the reasoning behind a px threshold vs checking whether mouse down + up happen within the same element?

ChronicLynx:
> Thank you, mind expanding on the reasoning behind a px threshold vs checking whether mouse down + up happen within the same element?

Nothing in particular, I'm sure both ways would work just fine.

acywatson:
> checking whether mouse down + up happen within the same element?

Could you hypothetically have different mousedown and mouseup targets though? Like an icon nested in a button? Haven't tested, just wondered if that would work as reliably. @zurfyx

zurfyx:
> > checking whether mouse down + up happen within the same element?
> 
> Could you hypothetically have different mousedown and mouseup targets though? Like an icon nested in a button? Haven't tested, just wondered if that would work as reliably. @zurfyx

My hunch is that the target should be the most realiable approach, but we would check for the a common parent in this case.

That said, I think that what @ChronicLynx is proposing is valid and it fixes an existing problem so I'll proceed to merge this. Thank you!

PR #4299, opened 2023-04-10
Persist files and scene state in Excalidraw plugin

ivailop7:
- When adding images to the Excalidraw scene, the files were not persisted
- When exporting the scene, it was losing some of the scene data, like background color.

Before:

https://user-images.githubusercontent.com/7893468/231016887-7a68c03b-341b-4156-b78c-414b828893b3.mp4

After:

https://user-images.githubusercontent.com/7893468/231016948-61eae2ab-d813-4de5-b588-e5bd7732fbdd.mp4

### DIFF
--- a/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
@@ -9,6 +9,7 @@
 import type {ExcalidrawElementFragment} from './ExcalidrawModal';
 import type {NodeKey} from 'lexical';
 
+import {AppState, BinaryFiles} from '@excalidraw/excalidraw/types/types';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {useLexicalNodeSelection} from '@lexical/react/useLexicalNodeSelection';
 import {mergeRegister} from '@lexical/utils';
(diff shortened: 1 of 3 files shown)


zurfyx (commented):
LGTM, will wait for a bit to see if @acywatson has some thoughts on this since he's been playing with Excalidraw recently

zurfyx (approved):

acywatson:
Sorry, I didn't have a chance to look too closely - is this just storing the image data in the actual excalidraw state? Is it base64 encoded? That might lead to some absolutely massive states with large images the 33% penalty from b64.

Is there support for storing just a reference to the image? Does this PR allow that?

ivailop7:
> Sorry, I didn't have a chance to look too closely - is this just storing the image data in the actual excalidraw state? Is it base64 encoded? That might lead to some absolutely massive states with large images + the 33% penalty from b64.
> 
> 
> 
> Is there support for storing just a reference to the image? Does this PR allow that?

Hey, @acywatson, you are spot on, if someone adds an image, it stores the image as a base64 encoded string in the scene state. Can it lead to massive states and thus a heavy document? Yes, absolutely. That said, I'd like to believe, users are conscious of what adding too many images this way means and I'd take a fully working Excalidraw component with all features working that they can disable (eg. by hiding the "add image" button inside the component), rather than have it miss functionality. I don't have a strong opinion on it, if you want to hide the button in the playground, happy to do it as well.

acywatson:
> > Sorry, I didn't have a chance to look too closely - is this just storing the image data in the actual excalidraw state? Is it base64 encoded? That might lead to some absolutely massive states with large images + the 33% penalty from b64.
> > Is there support for storing just a reference to the image? Does this PR allow that?
> 
> Hey, @acywatson, you are spot on, if someone adds an image, it stores the image as a base64 encoded string in the scene state. Can it lead to massive states and thus a heavy document? Yes, absolutely. That said, I'd like to believe, users are conscious of what adding too many images this way means and I'd take a fully working Excalidraw component with all features working that they can disable (eg. by hiding the "add image" button inside the component), rather than have it miss functionality. I don't have a strong opinion on it, if you want to hide the button in the playground, happy to do it as well.

No, I think we're good - I discussed offline with @zurfyx. it's just playground code, but I had a feeling best practice might be somehow adding a ref to the image and loading it async from storage, I just don't know how that's done in excalidraw. It's a problem for another day - this is good work and thank you for doing it :)

PR #4306, opened 2023-04-11
Table Cell Background Color

zurfyx:
https://user-images.githubusercontent.com/193447/231279672-c23ea8e2-bf45-477f-966d-95f52da9774d.mov


packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -441,12 +491,21 @@ function TableActionMenu({
       onClick={(e) => {
         e.stopPropagation();
       }}>
-      {mergeCellButton !== null && (
-        <>
-          {mergeCellButton}
-          <hr />
-        </>
-      )}
+      {mergeCellButton}
```
amyworrall, line 494: Do we no longer need to nil check the mergeCellButton?

packages/lexical-playground/src/ui/DropdownColorPicker.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type Props = {
+  disabled?: boolean;
+  buttonAriaLabel?: string;
+  buttonClassName: string;
+  buttonIconClassName?: string;
+  buttonLabel?: string;
+  title?: string;
+  stopCloseOnClickSelf?: boolean;
+  color: string;
```
amyworrall, line 22: Would it be appropriate to define a new type for hex colour strings, making it a bit more semantically obvious when they're used as API parameters?

packages/lexical-playground/src/ui/ColorPicker.tsx
```diff
@@ -118,57 +105,51 @@ export default function ColorPicker({
...
+              setInputColor(basicColor);
+              setSelfColor(transformColor('hex', basicColor));
             }}
           />
-        </MoveWrapper>
+        ))}
+      </div>
+      <MoveWrapper
+        className="color-picker-saturation"
+        style={{backgroundColor: `hsl(${selfColor.hsv.h}, 100%, 50%)`}}
```
amyworrall, line 128: you've done this logic (modify an HSV colour to have 100% sat, 50% lightness) a few times; maybe factor into its own function?

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -134,10 +141,30 @@ function $selectLastDescendant(node: ElementNode): void {
...
 }
 
+function currentCellBackgroundColor(editor: LexicalEditor): null | string {
+  return editor.getEditorState().read(() => {
+    const selection = $getSelection();
+    if (
+      $isRangeSelection(selection) ||
+      DEPRECATED_$isGridSelection(selection)
+    ) {
+      const [cell] = DEPRECATED_$getNodeTriplet(selection.anchor);
```
amyworrall, line 151: I assume there isn't a non-deprecated alternative here?

amyworrall (approved):

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -248,6 +270,10 @@ export function convertTableCellNodeElement(
   );
   tableCellNode.__colSpan = domNode_.colSpan;
   tableCellNode.__rowSpan = domNode_.rowSpan;
+  const backgroundColor = domNode_.style.backgroundColor;
+  if (backgroundColor !== '') {
+    tableCellNode.__backgroundColor = backgroundColor;
```
acywatson, line 275: You're going to have the same problem here that all other nodes implementing styling like this do. If the editor you paste into doesn't support this feature, then the user will have no option to remove it. Even if you paste into another Lexical editor, you may end up using HTML because of a namespace mismatch.

packages/lexical-table/src/LexicalTableSelection.ts
```diff
@@ -215,21 +221,22 @@ export class TableSelection {
       this.gridSelection = selection;
       this.isHighlightingCells = true;
       this.disableHighlightStyle();
-      $updateDOMForSelection(this.grid, this.gridSelection);
+      $updateDOMForSelection(this.editor, this.grid, this.gridSelection);
```
acywatson, line 224: can it just be 'editor'

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
zurfyx: Rendering a nullish component in React will not render it. It's a shorthand for `foo != null ? foo : null`

packages/lexical-playground/src/ui/DropdownColorPicker.tsx
zurfyx: What kind of type are you thinking of? Sort of like constant for colors? I didn't change the API in this case, the ColorPicker was already present but in any case we can propose it as a `good-first-issue`

packages/lexical-playground/src/ui/ColorPicker.tsx
zurfyx: Didn't write this file, just moved the files around. Sorry if the blame is confusing

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
zurfyx: The idea is to kill GridSelection eventually by moving it into the Table package itself. That's why it remains as deprecated while still actively using it

packages/lexical-playground/src/ui/DropdownColorPicker.tsx
amyworrall: I meant the JS equivalent of this Swift code:

```
public typealias HexColor = String
```

Just to give a bit more semantics of what the string should be.

PR #4307, opened 2023-04-11
ExportDOM in Debug Tree View

ivailop7:
https://user-images.githubusercontent.com/7893468/231305891-19835d70-b54d-4833-b73b-552a8f1f5326.mp4

I've been meaning to add this for ages - the ability to toggle between node view and an exportDOM view, in order to validate if the a node's exportDOM definition is correct, since the DevTools show the createDOM output, besides manually printing exportDOM inside node methods, there is no intuitive way to check if the exportDOM properties are set correctly during development. 

Happy to hear suggestions for improvement.

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -1368,6 +1368,21 @@ button.action-button:disabled {
   text-decoration: underline;
 }
 
+.debug-treetype-button {
+  border: 0;
+  padding: 0;
+  font-size: 12px;
+  top: 10px;
+  right: 85px;
+  position: absolute;
+  background: none;
+  color: #fff;
+}
+
+.debug-treetype-button:hover {
+  text-decoration: underline;
+}
+
 .connecting {
   font-size: 15px;
   color: #999;
--- a/packages/lexical-playground/src/plugins/TreeViewPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TreeViewPlugin/index.tsx
@@ -15,6 +15,7 @@ export default function TreeViewPlugin(): JSX.Element {
   return (
     <TreeView
       viewClassName="tree-view-output"
+      treeTypeButtonClassName="debug-treetype-button"
       timeTravelPanelClassName="debug-timetravel-panel"
       timeTravelButtonClassName="debug-timetravel-button"
       timeTravelPanelSliderClassName="debug-timetravel-panel-slider"
--- a/packages/lexical-react/src/LexicalTreeView.tsx
+++ b/packages/lexical-react/src/LexicalTreeView.tsx
@@ -7,7 +7,6 @@
  */
 
 import type {
-  EditorConfig,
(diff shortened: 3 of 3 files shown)


Reggaevibes:
> I've been meaning to add this for ages - the ability to toggle between node view and an exportDOM view, in order to validate if the a node's exportDOM definition is correct, since the DevTools show the createDOM output, besides manually printing exportDOM inside node methods, there is no intuitive way to check if the exportDOM properties are set correctly during development.
> 
> Happy to hear suggestions for improvement.

Add me as a friend bro. :) nice informasjon.  I try to se if I  can figuure it out

zurfyx (approved):
LGTM. Nit, should the header show when the tree is larger than 1k nodes?

Also, you may want to know that we're rethinking to retake the extension project as the TreeView has grown out of control. Not only it's big now but we've also got some additional debugging features in mind that it will be hard to fit into this UI in a compact manner.

ivailop7:
> LGTM. Nit, should the header show when the tree is larger than 1k nodes?
> 
> 
> 
> Also, you may want to know that we're rethinking to retake the extension project as the TreeView has grown out of control. Not only it's big now but we've also got some additional debugging features in mind that it will be hard to fit into this UI in a compact manner.

Thanks @zurfyx , Yep, should disappear, will fix it. I've noticed quite a few things I can improve in my PR, in terms of propagating variables. 

On the extension part, that's excellent news. I've always thought the debug should be a separate extension, similar to the one React and Redux have and have the TreeView outside of the core package.

acywatson (approved):
This is sick!

Reggaevibes:
Its been aproved
Have a great day everybody

ons. 12. apr. 2023, 14:42 skrev Gerard Rovira ***@***.***>:

ivailop7:
@acywatson @zurfyx updated the PR to not render the button if the nodes count > 1k. Also changed the variable propagation a bit, here:
https://github.com/facebook/lexical/pull/4307/files#diff-fbd7c10fc2ea36767e9a70524ac63cdecb54b08c0936782358fc252399a9f33aL381-L385
since I was passing the editor as the last parameter anyway, are you happy with this change - consolidating all the variables to just editor. Happy to change it back, doesn't affect the useEffect variables, so should make no performance difference rerendering-wise, I believe.

zurfyx:
> @acywatson @zurfyx updated the PR to not render the button if the nodes count > 1k. Also changed the variable propagation a bit, here: https://github.com/facebook/lexical/pull/4307/files#diff-fbd7c10fc2ea36767e9a70524ac63cdecb54b08c0936782358fc252399a9f33aL381-L385 since I was passing the editor as the last parameter anyway, are you happy with this change - consolidating all the variables to just editor. Happy to change it back, doesn't affect the useEffect variables, so should make no performance difference rerendering-wise, I believe.

Thank you, LGTM!

That1dudeRay (commented):
Yes

PR #4310, opened 2023-04-12
Fix crash on prettier formatting fail

### LINKED ISSUE #4309: Bug: lexical crashes on prettier formatting fail for code block
In the Code Block, there is an option to format the code using Prettier button in the Code Action toolbar. If the code formatting fails, it crashes lexical and the editor becomes unresponsive.

Here is the video demonstrating the issue:
lexical prettier crash bug

Lexical version: 0.9.1

## Steps To Reproduce

1. Insert a block code
2. Type some invalid code which prettier is not able to format. For instance, `function newColl() {let a = 3; let b=5; return a+b::}`
3. Format code using Prettier button

## The current behavior
Editor becomes unresponsive and doesn't respond to anything typed inside the editor. 

## The expected behavior
Prettier formatting fail should not crash the editor.

umaranis:
This pull request resolves #4309


zurfyx (approved):
LGTM, thank you!

PR #4315, opened 2023-04-12
Fix issue where H2 and H3 headings were displayed in uppercase

ensunluer:
Fix the issue where H2 and H3 headings were displayed in uppercase

This commit resolves the problem in which H2 and H3 headings were incorrectly displayed in uppercase letters. The root cause was identified, and the necessary changes were made in the CSS and/or JavaScript files to ensure that the headings are now displayed correctly. This fix improves the readability and consistency of the text editor's headings.

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -38,12 +38,10 @@
   color: rgb(101, 103, 107);
   font-weight: 700;
   margin: 0;
-  text-transform: uppercase;
 }
 .PlaygroundEditorTheme__h3 {
   font-size: 12px;
   margin: 0;
-  text-transform: uppercase;
 }
 .PlaygroundEditorTheme__indent {
   --lexical-indent-base-value: 40px;


acywatson:
I guess I don't mind this, as it's just sample code, but it's worth noting that other editors like GDocs have certain heading styles that transform to upper case. The style of headings is just an opinion that's fully configurable in Lexical.

ensunluer:
> 

I understand that this change is not enough, but I saw it as a problem. Working on the Playground resolves this issue, but a different solution is needed for the point you mentioned. I will continue to work on it

zurfyx:
> I understand that this change is not enough, but I saw it as a problem. Working on the Playground resolves this issue, but a different solution is needed for the point you mentioned. I will continue to work on it

Thank you for the PR. I don't think any further work is needed. I believe that what Acy meant is that this is opinionated. Your editor can choose to have headings in uppercase/lowercase, it is completely up to you. The way it is right now, via CSS, still makes it accessible (by screen readers that will read it as is on HTML) while giving it a _glamorous_ (opinionated) touch.

Issue #4317, opened 2023-04-12
Bug: resizing an image makes the cursor over the text `default` (arrow) instead of `text`


EgonBolton:
Lexical version: 0.9.1

## Steps To Reproduce

1. Resize an image in the playground
2. Move the mouse towards the text.

Link to code example: https://playground.lexical.dev/

## The current behavior
when hovering over the text, `cursor: default`.

## The expected behavior
when hovering over the text, `cursor: text`.

MohitSinghChauhan:
Hello, I would be interested in working on this issue. Could it be assigned to me? My GitHub handle is @MohitSinghChauhan. Thank you for your time and consideration.

PR #4318, opened 2023-04-13
Fix cursor changing to default when resizing image (#4317)

MohitSinghChauhan:
This commit fixes the issue of the cursor changing to default when resizing an image. The change was made in the `setEndCursor()` function, where the `cursor` property of the `editorRootElement` is set to `text`. This ensures that the cursor changes to the expected text cursor when hovering over the text.

No other changes were made in this commit.

### DIFF
--- a/packages/lexical-playground/src/ui/ImageResizer.tsx
+++ b/packages/lexical-playground/src/ui/ImageResizer.tsx
@@ -122,7 +122,7 @@ export default function ImageResizer({
 
   const setEndCursor = () => {
     if (editorRootElement !== null) {
-      editorRootElement.style.setProperty('cursor', 'default');
+      editorRootElement.style.setProperty('cursor', 'text');
     }
     if (document.body !== null) {
       document.body.style.setProperty('cursor', 'default');


zurfyx:
Thank you! Could you explain how your PR makes the existing behavior better (to what it was it before)? I can't repro this #4317 in Chrome, but I can still repro the same issue in Safari even after your fix

https://user-images.githubusercontent.com/193447/231807935-8f23bfc0-5c97-4b01-a531-76fdde8a67ef.mov

MohitSinghChauhan:
Could you please specify the issue you are facing in Safari?

I was able to fix a bug that caused the cursor to change to the default arrow when hovering over text after resizing an image instead of changing to the expected text cursor. I fixed this issue by updating the CSS property of the editor element in the `setEndCursor()` function, which is called just before the image resizing is completed.

Based on the video you provided, it seems that the cursor is correctly set to text while resizing the image. However, if you could provide more details about the issue you are facing in Safari, I would be happy to investigate and help resolve the problem.

zurfyx:
> Could you please specify the issue you are facing in Safari?
> 
> I was able to fix a bug that caused the cursor to change to the default arrow when hovering over text after resizing an image instead of changing to the expected text cursor. I fixed this issue by updating the CSS property of the editor element in the `setEndCursor()` function, which is called just before the image resizing is completed.
> 
> Based on the video you provided, it seems that the cursor is incorrectly set to text while resizing the image. However, if you could provide more details about the issue you are facing in Safari, I would be happy to investigate and help resolve the problem.

Oh yeah, I understand the fix now. Safari looks like another issue. Basically, the resize cursor gets lost while I'm still resizing.

Thank you, will merge the PR now! Feel free to look into this other issue separately

zurfyx (approved):

MohitSinghChauhan:
Great, glad to hear that you understand the fix! And thanks for merging the PR. As for the new issue you mentioned, I can definitely look into it separately.

PR #4325, opened 2023-04-14
expose getActiveEditor

acywatson:


zurfyx:
https://discord.com/channels/953974421008293909/1095768732330107030/1097615143833174037

IMO we shouldn't hint toward the possibility of having transforms inside transform or similar variants. While the `editor` instance is widely available beyond the closures (a powerful object that we never managed to hide despite the efforts to make EditorState more accessible), the utilities inside the closures should be the recommended approach to manipulate the EditorState.

That said, I'm OK to expose this if there's a compelling use case that requires it.

acywatson:
> While the `editor` instance is widely available beyond the closures (a powerful object that we never managed to hide despite the efforts to make EditorState more accessible), the utilities inside the closures should be the recommended approach to manipulate the EditorState.

I'll preface this by saying that I don't really think this is that big of a deal - just think it's good to have a discussion about it.

I would still love to have a deeper understanding of what your reservations are. I see you mentioned transforms inside transforms, but it's not clear to me how exposing this API changes that for better or worse. Can you help me understand?

zurfyx (approved):

PR #4329, opened 2023-04-14
Move link sanitizer into core

acywatson:
This makes the link library more secure by default by sanitizing links using default library logic. Users can provider their own sanitizer overrider or pass explicit null to disable sanitization altogether.

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -398,6 +398,7 @@ export const TOGGLE_LINK_COMMAND: LexicalCommand<
 export function toggleLink(
   url: null | string,
   attributes: LinkAttributes = {},
+  sanitizeUrl?: null | ((_url: string) => string | null),
 ): void {
   const {target, title} = attributes;
   const rel = attributes.rel === undefined ? 'noopener' : attributes.rel;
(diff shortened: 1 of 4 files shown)


acywatson:
@zurfyx 

> Have we considered sanitization rules at a plugin level?

Yes, that means no protection for non-React users. If [React](https://legacy.reactjs.org/blog/2019/08/08/react-v16.9.0.html#deprecating-javascript-urls) can do this in the core, so can we.

> there's repetition every time you want to toggle a link

Sure, good thinking

> Additionally, I think that this function is doing too much, it's a setter and a toggle, perhaps we should revisit this in the future.

Sure

zurfyx:
Having reviewed the internal context around the criticality of having this by default, I can extend my review to say:
1. It's very hard to catch all malicious sites with a sanitizer that works for everybody - internally we use a shimming approach that will kick in when moving to external non-Meta owned sites. Other people may be OK without a shimmer.
2. Sanitizer can be bypassed as it's a client-side module, on Collab this means that all users can be impacted

I believe that the best solution is a curated platform-specific solution around Shimming. However, I also agree that React's link validation (ban `javascript:`) is lightweight and works for 99% of the users. Given that Lexical is heavily inspired in React I think it's good having this built-in.

So I'd suggest something like this

`<LinkPlugin disableJavaScriptURLs={true} />`

And this can work in 2 different ways. 1) URL overwritten 2) Overwritting click handler. Preferably the first.

acywatson:
> `<LinkPlugin disableJavaScriptURLs={true} />`

This won't work for non-React users

acywatson:
We can consider either putting this in LinkNode.createDOM or in the reconciler.

PR #4339, opened 2023-04-17
Draft: Support nested blockquotes in markdown

themagickoala:
A requirement for our markdown needs is to support nested blockquotes. I've created this PR as a starting point to discuss a potential implementation. I'm sure there are things I've missed (@fantactuka mentioned the `$setBlocksType` potentially not behaving as expected on Discord), but wanted to float this here so that you have the full context.

Changes:
 - Update the QUOTE transformer to support nested quotes;
 - Update the export prop to recursively call `$convertToMarkdownString`;
 - Update the `exportTopLevelElements` method to pass transformers to the export prop to support the above;
 - Update the `exportTopLevelElements` method to run text match transformers;
 - Add some test cases for nested quotes.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -63,8 +63,15 @@ function exportTopLevelElements(
   textMatchTransformers: Array<TextMatchTransformer>,
 ): string | null {
   for (const transformer of elementTransformers) {
-    const result = transformer.export(node, (_node) =>
-      exportChildren(_node, textTransformersIndex, textMatchTransformers),
+    const result = transformer.export(
+      node,
+      (_node) =>
+        exportChildren(_node, textTransformersIndex, textMatchTransformers),
+      [
+        ...elementTransformers,
+        ...textTransformersIndex,
+        ...textMatchTransformers,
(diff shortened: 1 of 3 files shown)


fantactuka:
Thanks for the PR!

Few thoughs on it:
By default, Lexical does not nest quotes (neither via toolbar, nor via markdown shortcuts). So the idea that we only do it via markdown import does not sound right. That said, maybe it's better to allow such behaviour, just not have it as a part of core: update markdown export/import to allow recursive element parsing. Then app can define its own quote transformer to support nesting. What do you think?

themagickoala:
@fantactuka  that's the kind of solution I'm going for here. Markdown syntax has the capability to be sort of power user mode, where you can do a little more than you can with a toolbar. That's certainly how it works with our users anyway! I'm not sure the performance of my solution is up to scratch, this is mostly a proof of concept, I'd love to get some pointers though! I didn't want to make too many sweeping changes to the internals of the export/import, but it might end up being necessary to implement the kind of recursive support we might want?

thegreatercurve:
Closing this for the moment due to inactivity, and we're tracking this in a separate issue involving customising the behaviour beyond just markdown

PR #4342, opened 2023-04-17
[0.10] Ban javascript URLs in @lexical/link

acywatson:
More targeted approach to https://github.com/facebook/lexical/pull/4329

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -86,7 +86,7 @@ export class LinkNode extends ElementNode {
 
   createDOM(config: EditorConfig): HTMLAnchorElement {
     const element = document.createElement('a');
-    element.href = this.sanitizeUrl(this.__href);
+    element.href = this.sanitizeUrl(this.__url);
     if (this.__target !== null) {
       element.target = this.__target;
     }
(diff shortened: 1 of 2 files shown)


zurfyx (approved):

packages/lexical-link/src/index.ts
```diff
@@ -154,6 +162,19 @@ export class LinkNode extends ElementNode {
     return node;
   }
 
+  sanitizeUrl(url: string): string {
+    try {
+      const parsedUrl = new URL(url);
+      // eslint-disable-next-line no-script-url
+      if (!SUPPORTED_URL_PROTOCOLS.has(parsedUrl.protocol)) {
+        return 'about:blank';
+      }
```
zurfyx, line 171: What's your opinion on the `file:` protocol? Is it bad to support it? I understand the concerns around `javascript:` but not sure how strict we want to be

acywatson: I don't really have an opinion, tbh. This copies Quill, which seems like a sensible place to start since people presumably have been living with that for a while without any serious problems. My thought is to start with this and then expand. The only issue I have is that there's not really a good way to notify a user when this happens. Like, theoretically the URL could pass the user-supplied validation at the plugin level, and then fail here, which is not ideal, but probably pretty uncommon.

fantactuka (approved):
👍

PR #4343, opened 2023-04-17
TablePlugin feature: cell merge

zurfyx:
Initially, Nodes were simple in terms of "features", so were Tables. Few properties and 100% of the users wanted all of them. Nowadays, a few Nodes have grown, TextNode and TableNode are clear examples.

This PR doesn't address features in a generic manner (yet) but I believe this is an appropriate way to handle features without touching the Nodes, that ultimately should not be smart and should always retain the same type regardless of the feature.

Cell merge is addressed in this PR, later we'll add backgroundColor too.

https://user-images.githubusercontent.com/193447/232581073-c2738256-b93a-4d1e-a899-b0b33b130418.mov

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -62,11 +62,12 @@ async function fillTablePartiallyWithText(page) {
 }
 
 test.describe('Tables', () => {
-  test.beforeEach(({isCollab, page}) => initialize({isCollab, page}));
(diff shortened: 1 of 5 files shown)


acywatson:
this works ok for functionality that's always encapsulated in the same core plugin. Less so for things that are more frequently customized.

zurfyx:
> this works ok for functionality that's always encapsulated in the same core plugin. Less so for things that are more frequently customized.

+1. The idea of a "plugin" is something that I want to explore further too, I think there's something there.

tylerjbainbridge (approved):

PR #4363, opened 2023-04-19
Change `MarkNode::clone` `node` arg type to `this`

potomak:
Subclasses of `MarkNode` can't override `clone` without adding a Flow type error ignore annotation without this change.

### DIFF
--- a/packages/lexical-mark/flow/LexicalMark.js.flow
+++ b/packages/lexical-mark/flow/LexicalMark.js.flow
@@ -22,7 +22,7 @@ export type SerializedMarkNode = SerializedElementNode & {
 declare export class MarkNode extends ElementNode {
   __ids: Array<string>;
 
-  static clone(node: MarkNode): MarkNode;
+  static clone(node: this): MarkNode;
   constructor(ids: Array<string>, key?: NodeKey): void;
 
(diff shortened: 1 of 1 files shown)


packages/lexical-mark/flow/LexicalMark.js.flow
```diff
@@ -22,7 +22,7 @@ export type SerializedMarkNode = SerializedElementNode & {
 declare export class MarkNode extends ElementNode {
   __ids: Array<string>;
 
-  static clone(node: MarkNode): MarkNode;
+  static clone(node: this): MarkNode;
```
fantactuka, line 25: Didn't realize you can do that 😳

potomak:
I signed the CLA, can you rerun that action please?

potomak:
> Once the CLA is signed, our tooling will perform checks and validations. Afterwards, the pull request will be tagged with CLA signed. The tagging process may take up to 1 hour after signing. Please give it that time before contacting us about it.

Ignore my last message about re-running the CLA signature action :)

acywatson (approved):

acywatson:
Seems like we should be doing this for other nodes though, if it's good for one

Issue #3708, opened 2023-01-11
Bug: `DecoratorBlockNode.isInline` should return `false`


DaniGuardiola:
Lexical version: 0.7.6

It is my understanding that this class is a base for *block* nodes, so `isInline()` should return `false`. Currently it returns `true` as inherited by its superclass. Link to code: https://github.dev/facebook/lexical/blob/168f0d43fcbd3d4d3129aff2e57eb9cfab0ba8e4/packages/lexical-react/src/LexicalDecoratorBlockNode.ts#L26

moy2010:
It is not a base for block nodes.

DaniGuardiola:
@moy2010 care to elaborate? What is it for then? I might be misusing them :stuck_out_tongue:

moy2010:
Decorator nodes are a special type of nodes. Check the documentation on
nodes.

DaniGuardiola:
@moy2010 this issue refers to decorator **block** nodes, not just decorator nodes. It directly implies it's supposed to be a block, has formatting (alignment) logic, and renders as a `<div>` (block element). I think you might have missed the "block" part :)

moy2010:
Nope. Check the codebase, these are only used for nodes such as YoutubeNode
and company.

DaniGuardiola:
@moy2010 :facepalm: YoutubeNode is a **block**. Literally [all](https://github.dev/facebook/lexical/blob/168f0d43fcbd3d4d3129aff2e57eb9cfab0ba8e4/packages/lexical-playground/src/nodes/FigmaNode.tsx#L129) [usages](https://github.dev/facebook/lexical/blob/168f0d43fcbd3d4d3129aff2e57eb9cfab0ba8e4/packages/lexical-playground/src/nodes/TweetNode.tsx#L217) [in the codebase](https://github.dev/facebook/lexical/blob/168f0d43fcbd3d4d3129aff2e57eb9cfab0ba8e4/packages/lexical-playground/src/nodes/YouTubeNode.tsx#L131) are blocks, they all return `false` from `isInline`.

You're missing the point. But I digress.

moy2010:
Sorry, you're right. I thought that `DecoratorBlockNode` was `DecoratorNode`, but I realized that it's actually extending it.

The fix is super easy, just override the method.

DaniGuardiola:
[...]

PR #4383, opened 2023-04-21
Make DecoratorBlockNode.isInline return false

EgonBolton:
Fix https://github.com/facebook/lexical/issues/3708

I think that thread was closed because @DaniGuardiola point was not understood, which I think was made clearer in his last comment.

To avoid confusion again, please note that this change is about the DecoratorBlockNode, not the DecoratorNode.

With this change, no one runs the risk of forgetting to override the method.

### DIFF
--- a/packages/lexical-playground/src/nodes/FigmaNode.tsx
+++ b/packages/lexical-playground/src/nodes/FigmaNode.tsx
@@ -122,10 +122,6 @@ export class FigmaNode extends DecoratorBlockNode {
       />
     );
   }
-
-  isInline(): false {
-    return false;
-  }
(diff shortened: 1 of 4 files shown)


DaniGuardiola:
tysm @EgonBolton 🙏

acywatson:
Yea, I think this is right. @thegreatercurve

zurfyx (approved):
LGTM, thank you!

fantactuka (approved):

thegreatercurve (approved):
Lgtm!

RegiByte:
Nice!

Issue #4389, opened 2023-04-22
Undo crashes when Nested editor is empty upon initialization


zurfyx:
By adding

```
imageNode.__caption.update(() => {
  $getRoot().append($createParagraphNode()).selectEnd();
});
```

you can crash the editor when undoing upon creating an image. The Root cause seems to be connected to how we handle the `undoStack` and NestedEditors.

https://user-images.githubusercontent.com/193447/233753676-1a4c1b8a-a753-4432-a212-8523cd163d7c.mov

PR #4390, opened 2023-04-22
Fix HistoryPlugin selection out of sync

zurfyx:
_Need a proper review on this one..._

When the sharedHistory is a shared across parent/nested editors, you can reach a state where the HistoryEntry contains an EditorState and a `undoSelection` that do not point to the editor. This will always cause the editor to crash as it won't be able to find the node to select.

For #4389 in particular it seems like this can ocurr when the you previously had a selection on another editor and you then create the first change onto the new editor. The first change will carry the `editor` and `editorEditor` from the previous parent iteration and the selection will be `previousSelection` of the `nestedEditor`, regardless of what the selection of the `nestedEditor` is it will never be the selection of the (parent) `editor`

What I'm suggesting in this PR is to remove any chances that it can be out-of-sync, by relying on the selection in the same EditorState. 

_But I'm not sure why `undoSelection` was created in the first place so I might be missing something important._

Fixes https://github.com/facebook/lexical/issues/4389

### DIFF
--- a/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
+++ b/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
@@ -6,6 +6,7 @@
  *
  */
 
+import {createEmptyHistoryState, registerHistory} from '@lexical/history';
 import {useLexicalComposerContext} from '@lexical/react/src/LexicalComposerContext';
 import {ContentEditable} from '@lexical/react/src/LexicalContentEditable';
 import LexicalErrorBoundary from '@lexical/react/src/LexicalErrorBoundary';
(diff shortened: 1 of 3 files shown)


zurfyx:
> See #966 summary recording for a case where undo selection was added. Although it seems to work without undoSelection now, probably addressed with a better calculation of history push vs history merge cases

@fantactuka - I can repro the same behavior as the test plan without it 

image


Mind sharing more details on why we need `undoSelection` please?

fantactuka:
> Mind sharing more details on why we need `undoSelection` please?

IIRC it was specifically added for cases when few operations coalesced including caret move, and during undo caret position supposed to be in (N-1). That's all I can remember :) I suppose it has been fixed with history merge algo adjustments for selection-change cases, so as long as it works as in test plan above I'm more than happe to remove undoSelection

zurfyx:
> > Mind sharing more details on why we need `undoSelection` please?
> 
> IIRC it was specifically added for cases when few operations coalesced including caret move, and during undo caret position supposed to be in (N-1). That's all I can remember :) I suppose it has been fixed with history merge algo adjustments for selection-change cases, so as long as it works as in test plan above I'm more than happe to remove undoSelection

That's how it works now which behavior is identical to the previous version. While arguably there's undo nits that don't work like other browsers (line new line) the behavior is the same.


https://user-images.githubusercontent.com/193447/234305039-8eebb5f6-4596-47f4-9fe4-9b883f3faf47.mov

fantactuka (approved):

Issue #3705, opened 2023-01-11
Bug: generateHtmlFromNodes ignores selection argument


SeanBrodyWP:
When I call `$generateHtmlFromNodes` with both an editor and a selection, it only returns the HTML for nodes/text in both the editorState's selection (what the user selected) and the selection object I pass to the call. For example, if I pass a node selection with nodes A,B,C, and the user has only selected B, then it only exports the HTML for B. 

Lexical version:

## Steps To Reproduce

1. Register an update listener
   1. Get all nodes of a certain type
   2. For each node
      1. Create a node selection and add that node's key
      2. Run $generateHtmlFromNodes using that selection
2. Create a few lines/nodes, and change the selection 

When you have multiple nodes, it will always output the currently selected node (or wherever the cursor currently is)

To replicate, replace MyCustomAutoFocusPlugin.js in the [example codesandbox](https://codesandbox.io/s/lexical-plain-text-example-g932e), with this:

```js
import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
import { $nodesOfType, $createNodeSelection, TextNode } from "lexical";
import { useEffect } from "react";
import { $generateHtmlFromNodes } from "@lexical/html";

export default function MyCustomAutoFocusPlugin() {
  const [editor] = useLexicalComposerContext();

  useEffect(() => {
    return editor.registerUpdateListener(() => {
      editor.update(() => {
        const nodes = $nodesOfType(TextNode);
        for (const tNode of nodes) {
          const nodeSelection = $createNodeSelection();
          const key = tNode.getKey();
          nodeSelection.add(key);
          const html = $generateHtmlFromNodes(editor, nodeSelection);
          console.log("html", key, html);
        }
      });
    });
  }, [editor]);

  return null;
}

```

## The current behavior

It always returns the HTML for the user-selected node, not the "selection" passed into the function.
![image](https://user-images.githubusercontent.com/56841458/211694773-a625c8a6-efa2-4811-b027-f0f01a9114db.png)

## The expected behavior
It returns the HTML for the selection I pass into the function, and the user's selection has no effect.

SeanBrodyWP:
Thank you for your response. I was not aware that node selection were exclusivly for decorator nodes. The documentation says that NodeSelections are "selection[s] of multiple arbitrary nodes", so I assumed I could use them for any nodes. In my code, I'm using them to select a custom ElementNode subclass, but for my example code it was easier to work with text nodes.

[...]

PR #4392, opened 2023-04-23
Use provided selection when serializing to HTML

acywatson:
Fixes https://github.com/facebook/lexical/issues/3705

Adds unit tests for HTML serialization

### DIFF
--- a/packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
+++ b/packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
@@ -1,0 +1,153 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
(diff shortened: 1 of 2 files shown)


packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      const editor = createHeadlessEditor({
+        nodes: [
+          HeadingNode,
+          ListNode,
+          ListItemNode,
+          QuoteNode,
+          CodeNode,
+          LinkNode,
+        ],
+      });
```
zurfyx, line 52: Why not `createTestEditor`?

acywatson: Good call - I copied from Markdown where the idea was apparently to skip rendering for faster execution. I can see the benefit of having this config consolidated into one place though. Maybe we need createHeadlessTestEditor or an optional headless param in createTestEditor. Will come back to this.

PR #4395, opened 2023-04-24
Fix text click selection on touch devices #3098

### LINKED ISSUE #3098: Bug: selection changes to null when selecting text after node selection on mobile devices
Lexical version: 0.5.0
<video src="https://user-images.githubusercontent.com/13211445/193463810-8af30292-f54f-455c-9461-56370dc59118.mp4"></video>

IBastawisi:
This PR fixes #3098

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -369,7 +369,7 @@ function onSelectionChange(
 // This results in a tiny selection box that looks buggy/broken. This can
 // also help other browsers when selection might "appear" lost, when it
 // really isn't.
-function onClick(event: MouseEvent, editor: LexicalEditor): void {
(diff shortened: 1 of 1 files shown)


zurfyx (commented):
Thank you, can you expand on why a `click` solution vs `selectionchange`?

IBastawisi:
> Thank you, can you expand on why a `click` solution vs `selectionchange`?

I don't know the exact cause of this, but inside the selectionchange event handler the value of domSelection.anchorNode is null, however in onClick it is not. This happens only with touch event, with mouse it is as expected.

https://user-images.githubusercontent.com/13211445/234132248-13ac54fa-5ee7-4a4b-a9ce-6834437ab8cb.mp4

zurfyx:
Correction: the click handler is likely the best place given that the similar `pointerdown` code will not see selection (`selection.anchor === null`)

PR #4398, opened 2023-04-24
Setting initial state of isSmallWidthViewport

ebads67:
isSmallWidthViewport state is not set correctly when the editor is rendered in a small window. That's why DraggableBlockPlugin is enabled initially when in a small window.

DraggableBlockPlugin-bug


zurfyx (approved):
LGTM, thank you!

Issue #4347, opened 2023-04-18
CMD + K to trigger link dialog


zurfyx:
Ctrl/CMD + K is a common convention to toggle link

AnuragBalhra:
Sounds like a good first issue, I can pick it up if no one else is working on it yet.

Slapbox:
@AnuragBalhra are you planning to work on this? Do you know how the `Ctrl/Cmd+B` are handled for Bold for an example of how to add this?

AnuragBalhra:
Yes I'm starting to work on this. 
I don't already know how to do it but I'll try to investigate how it's done for Ctrl/Cmd+B. If someone already knows any documentation/references about how they are implemented, that would be really helpful.

evskorobogatij:
I do

acywatson:
Closed by #4413

PR #4413, opened 2023-04-26
Updated Toolbar Plugin to Add keyboard shortcut for adding hyperlinks

AnuragBalhra:
Fix for Issue #4347

Registered new event listener in Toolbar plugin to insert link using TOGGLE_LINK_COMMAND when we press Cmd/Ctrl + K

This is my first contribution in the repo so not sure if I did it the right way.

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
@@ -59,10 +59,12 @@ import {
   CAN_REDO_COMMAND,
   CAN_UNDO_COMMAND,
   COMMAND_PRIORITY_CRITICAL,
+  COMMAND_PRIORITY_NORMAL,
   DEPRECATED_$isGridSelection,
   FORMAT_ELEMENT_COMMAND,
   FORMAT_TEXT_COMMAND,
   INDENT_CONTENT_COMMAND,
+  KEY_MODIFIER_COMMAND,
   OUTDENT_CONTENT_COMMAND,
   REDO_COMMAND,
   SELECTION_CHANGE_COMMAND,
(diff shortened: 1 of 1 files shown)


affvalente:
When I press Ctrl+K on windows this change opens the browser's search bar, rather than Lexical's link edit box

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -515,6 +515,16 @@ export default function ToolbarPlugin(): JSX.Element {
       editor.registerEditableListener((editable) => {
         setIsEditable(editable);
       }),
+      activeEditor.registerRootListener((rootElement) => {
+        if (rootElement) {
```
fantactuka, line 519: Can use KEY_MODIFIER_COMMAND handler

AnuragBalhra:
@affvalente Just curious how do other contributors test their changes on different os platforms? I just have Mac. Should I install VM for win?

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
AnuragBalhra: Really appriciate the suggestion. 
I've updated the changes to use KEY_MODIFIER_COMMAND command handler.

fantactuka:
> @affvalente Just curious how do other contributors test their changes on different os platforms? I just have Mac. Should I install VM for win?

e2e tests would run on all platforms, as it's pretty hard to have all of them available to test locally. That said, I think it should work fine now with preventDefault()

fantactuka (approved):
👍

zurfyx (approved):
Thank you, LGTM! Please have a look into https://github.com/facebook/lexical/issues/4439 if you have some time, some small nits but the UX would improve greatly!

PR #4420, opened 2023-04-28
add optional title tag to ImageNode.tsx

nerdess:
A optional title tag on images is useful for accessibility, SEO and UX.

### DIFF
--- a/packages/lexical-playground/src/nodes/ImageNode.tsx
+++ b/packages/lexical-playground/src/nodes/ImageNode.tsx
@@ -36,14 +36,15 @@ export interface ImagePayload {
   maxWidth?: number;
   showCaption?: boolean;
   src: string;
+  title?: string;
   width?: number;
   captionsEnabled?: boolean;
 }
 
 function convertImageElement(domNode: Node): null | DOMConversionOutput {
   if (domNode instanceof HTMLImageElement) {
-    const {alt: altText, src, width, height} = domNode;
-    const node = $createImageNode({altText, height, src, width});
+    const {alt: altText, src, title, width, height} = domNode;
+    const node = $createImageNode({altText, height, src, title, width});
     return {node};
   }
   return null;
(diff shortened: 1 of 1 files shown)


zurfyx (dismissed):
LGTM, thank you!

packages/lexical-playground/src/nodes/ImageNode.tsx
```diff
@@ -52,6 +53,7 @@ function convertImageElement(domNode: Node): null | DOMConversionOutput {
 export type SerializedImageNode = Spread<
   {
     altText: string;
+    title: string;
```
zurfyx, line 56: It's a new field so it's not guaranteed to be there in currently serialized nodes
```suggestion
    title?: string;
```

packages/lexical-playground/src/nodes/ImageNode.tsx
```diff
@@ -202,12 +211,17 @@ export class ImageNode extends DecoratorNode<JSX.Element> {
...
+    return this.__title;
+  }
+
   decorate(): JSX.Element {
     return (
       <Suspense fallback={null}>
         <ImageComponent
           src={this.__src}
           altText={this.__altText}
+          title={this.__title}
```
christhoph, line 224: It would be necessary to add the prop `title` in the `ImageComponent` component

https://github.com/facebook/lexical/blob/e90c28b122b87af19258601c45c3d31afad8bb6e/packages/lexical-playground/src/nodes/ImageComponent.tsx#L120-L131

zurfyx:
Mind checking the failing integrity check please before we can merge? You might have to handle nullish values on the ImageNode

fantactuka:
Although this change makes sense, I wonder if it's needed given ImageNode is not exposed from any package and used purely as a demo in the playground. These playground nodes might be copied as a reference by users and adjusted based on their specific app needs

PR #4422, opened 2023-04-28
Add codemirror to playground

zignis:
A POC implementing codemirror to the playground.

- Line wrap support (with toggle)
- Copy code
- More languages
- Export JSON only contains the code as text content, making the export size and DOM elements size very small
- More performant than @lexical/code and it does not create separate elements for every token, + codemirror features virtualization for very large code contents.
- Code block moved to "Insert" dropdown instead of block type dropdown

Can this be carried forward?


thegreatercurve:
Closing due to inactivity

PR #4423, opened 2023-04-28
feat: playground inline image plugin

58bits:
I thought that perhaps a playground example of an inline image with float and wrapped text might be useful (including `updateDOM`) - and so I've created an inline image node and plugin - all 'namespaced' as a separate plugin and node from the current image node. See the attached video demo. The plugin includes a 'round trip' edit option, and so the inserted image position can be updated.

https://user-images.githubusercontent.com/405612/235197663-7ba3cfe9-60bd-4b75-82b6-75edd1f19d69.mp4

### DIFF
--- a/packages/lexical-playground/src/plugins/InlineImagePlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/InlineImagePlugin/index.tsx
@@ -1,0 +1,342 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+import type {Position} from '../../nodes/InlineImageNode';
+
+import '../../ui/Checkbox.css';
+
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {$wrapNodeInElement, mergeRegister} from '@lexical/utils';
+import {
+  $createParagraphNode,
+  $createRangeSelection,
+  $getSelection,
+  $insertNodes,
+  $isNodeSelection,
+  $isRootOrShadowRoot,
+  $setSelection,
+  COMMAND_PRIORITY_EDITOR,
+  COMMAND_PRIORITY_HIGH,
+  COMMAND_PRIORITY_LOW,
+  createCommand,
+  DRAGOVER_COMMAND,
+  DRAGSTART_COMMAND,
+  DROP_COMMAND,
+  LexicalCommand,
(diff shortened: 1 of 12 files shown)


ivailop7:
This is so cool! I believe it would make sense to extend the existing image plugin vs introducing a new one. Having looked at the changes, this seems quite reasonable. @acywatson and @zurfyx , what do you think?

ivailop7:
> > This is so cool! I believe it would make sense to extend the existing image plugin
> 
> Thanks @ivailop7 !
> 
> I can see how from the Playground perspective that combining this with the existing image plugin would make sense, but... from an editorial user perspective, the dream solution would be to keep them separate and rename the existing plugin 'ResizableImagePlugin' and then include this one as 'InlineImagePlugin'. The reason for this is that from a 'house design style' perspective, we would almost never give an editorial user a 'resizable' image. Images would conform to a design system and style guide - in particular inline floated images.
> 
> Another great thing about Lexical is that since the model elements for paragraphs actually render as `<p>` elements in the editor, inline images actually work! - which is brilliant (not naming any other editors, but this isn't always the case).
> 
> My two cents - although again, I can totally understand how combining them here in the playground would make sense.

Understood, I'm approaching it mostly from reducing the code duplication in the playground, since the node definition overlaps with the existing ImageNode ~85%, so it might be better to extend the existing ImageNode class than the DecoratorNode class, if you prefer to keep them separate, conscious that this creates a dependency, which led to my suggestion to consolidate them into a single node/plugin. 

I think it makes sense to have 1 full-fledged example and then people can remove the things they don't need from the ImageNode/Plugin. Would be nice to get other people's views as well. Not my decision here, just a friendly suggestion. Nonetheless, great work @58bits!

packages/lexical-playground/src/plugins/InlineImagePlugin/index.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+  return data;
+}
+
+declare global {
+  interface DragEvent {
+    rangeOffset?: number;
+    rangeParent?: Node;
+  }
+}
```
thegreatercurve, line 309: I think you're effectively declaring this twice as it's already in the ImagePlugin

thegreatercurve:
It's been a while since I've seen anything to do with floats, but thanks for the work @58bits! I share the concern that there is a lot of duplication with the `ImagePlugin`, but generally, this seems like something that can be addressed in another PR - possibly just unifying the drop and drag events into a common hook. 

Otherwise, this all broadly seems fine to merge.

thegreatercurve (approved):

PR #4424, opened 2023-04-28
Event delegation for NoteEventPlugin

fantactuka:
There's couple of problems with current implementation:
- It relies on mutation listener that does not run for initial state (only for added nodes)
- It attaches event to specific node, so for nodes like TableCell or Paragraph it can add hundreds or thousands of listeners for longer documents

This on attempts to add event delegation with special handling for mouseenter/leave events via capturing since those don't bubble


fantactuka:
@trueadm curious what do you think about it

zurfyx (approved):

PR #4430, opened 2023-04-30
Improve URL regex and separator detection in AutoLinkPlugin

### LINKED ISSUE #3546: Bug: [Playground] autolink escapes on second `.` when typing

NozomiSugiyama:
close #3546

## What this PR does:
This PR updates the URL detection method in the AutoLinkPlugin with the following changes:
- Allow URLs with extensions (e.g., .png, .jpeg)
- Allow URLs that do not start with "www" (e.g., gist.github.com)
- Disallow domains ending with a dot (e.g., example.com.)
- Disallow TLDs with fewer than 2 characters (e.g., example.c)
- Make URL detection case-insensitive (e.g., https://example.com/)

## Before:
Screenshot 2023-04-30 at 16 35 19

## After:
Screenshot 2023-04-30 at 16 35 25

Please review and let me know if you have any suggestions or concerns.


packages/lexical-react/src/LexicalAutoLinkPlugin.ts
```diff
@@ -70,7 +70,7 @@ function findFirstMatch(
   return null;
 }
 
-const PUNCTUATION_OR_SPACE = /[.,;\s]/;
+const PUNCTUATION_OR_SPACE = /[,;\s]/;
```
fantactuka, line 73: I think this part should be improved a bit. Eg having ". " is a punctuation, while just dot is not to allow extensions and subdomains

NozomiSugiyama: Thank you for your comment!
That is certainly true.

I'm reviewing the design myself again, and I am considering whether it is necessary to change the way the LexicalAutoLinkPlugin separator `/[.,;\s]/` is handled. For languages other than English, there are other characters that are used as punctuation marks, so rather than restricting within the LexicalAutoLinkPlugin to specific punctuation characters, it is better to determine if the regular expressions provided in the matchers are satisfied. I think.

Also, I think that it can solve the discomfort of not being judged as a URL even though it matches the REGEX passed in matches.

Eg:  `/(((https?):\/\/)?([\w-]+\.)+[\w-]{1,}(\/([\w()@:%_+.~#?&/=]*)?)?[^.,;\s])/i`

I will try to implement

NozomiSugiyama:
Please check back as we've made corrections.

Changes
1. instead of no longer considering delimiters, AutoLinkNode is now generated or edited by determining whether to combine the TextContent of adjacent TextNode and match it to the URL when generating AutoLinkNode.
2. When converting from Text to Markdown, AutoLinkNode will now work even if it contains an AutoLinkNode.

ivailop7:
@NozomiSugiyama the integration tests are failing, if you are still looking after this PR

ivailop7:
> 

@NozomiSugiyama are you planning to update this one, happy to help on this one if needed.

NozomiSugiyama:
@ivailop7 
My apologies for the delayed response.
I have rebased this onto the latest main branch and fixed the issues with the findLast method that were causing the tests to fail.

ivailop7:
> @ivailop7 My apologies for the delayed response. I have rebased this onto the latest main branch and fixed the issues with the findLast method that were causing the tests to fail.

Reran the tests, but unfortunately only the collab tests are passing, the rest are still failing.

plum-sugiyama:
@ivailop7 
Ah! I missed that, my apologies. I've now fixed the behavior of the auto link node when a LineBreakNode comes at the end in plain-text mode. I have confirmed that the following tests are working correctly on my end:

```
env E2E_EDITOR_MODE=plain-text E2E_EVENTS_MODE=legacy-events npm run test-e2e-chromium
env E2E_EDITOR_MODE=plain-text E2E_EVENTS_MODE=modern-events npm run test-e2e-chromium
env E2E_EDITOR_MODE=rich-text E2E_EVENTS_MODE=legacy-events npm run test-e2e-chromium
env E2E_EDITOR_MODE=rich-text E2E_EVENTS_MODE=modern-events npm run test-e2e-chromium
```
Please check the changes again. Thank you.

NozomiSugiyama:
@ivailop7 
Ah! I missed that, my apologies. I've now fixed the behavior of the auto link node when a LineBreakNode comes at the end in plain-text mode. I have confirmed that the following tests are working correctly on my end:

```
env E2E_EDITOR_MODE=plain-text E2E_EVENTS_MODE=legacy-events npm run test-e2e-chromium
env E2E_EDITOR_MODE=plain-text E2E_EVENTS_MODE=modern-events npm run test-e2e-chromium
env E2E_EDITOR_MODE=rich-text E2E_EVENTS_MODE=legacy-events npm run test-e2e-chromium
env E2E_EDITOR_MODE=rich-text E2E_EVENTS_MODE=modern-events npm run test-e2e-chromium
```
Please check the changes again. Thank you.

ivailop7:
Thank you @NozomiSugiyama for addressing the comments. @fantactuka or @zurfyx  could any of you scan this through again, it looks OK to me, but I think someone a bit more familiar with that part of the code would be nice to sign off as well.

ivailop7 (approved):

ivailop7:
Hey @zurfyx and @fantactuka can you have a look if this looks good to you too, looks mergeable to me.

packages/lexical-react/src/LexicalAutoLinkPlugin.ts
```diff
@@ -70,64 +64,104 @@ function findFirstMatch(
...
+    node.insertAfter(children[j]);
   }
 
-  const contentAfterIsValid =
-    matchEnd < text.length
-      ? isSeparator(text[matchEnd])
-      : isNextNodeValid(node);
-  return contentAfterIsValid;
+  node.remove();
+  return children.map((child) => child.getLatest());
```
acywatson, line 164: Don't return the mutable version of the node here

acywatson (changes requested):

packages/lexical-react/src/LexicalAutoLinkPlugin.ts
```diff
@@ -70,64 +64,104 @@ function findFirstMatch(
   return null;
 }
 
-const PUNCTUATION_OR_SPACE = /[.,;\s]/;
-
-function isSeparator(char: string): boolean {
-  return PUNCTUATION_OR_SPACE.test(char);
-}
+function getPrefixTextSiblings(node: LexicalNode) {
+  const prevSiblings = node.getPreviousSiblings().reverse();
```
acywatson, line 68: Do we need to reverse this twice? Can we just iterate backwards?

packages/lexical-react/src/LexicalAutoLinkPlugin.ts
```diff
@@ -70,64 +64,104 @@ function findFirstMatch(
...
+    nodes: result,
+    textContent: result.map((n) => n.getTextContent()).join(''),
+  };
 }
 
-function isNextNodeValid(node: LexicalNode): boolean {
-  let nextNode = node.getNextSibling();
-  if ($isElementNode(nextNode)) {
-    nextNode = nextNode.getFirstDescendant();
+function splitNodes(
```
acywatson, line 103: What is this supposed to be doing?

packages/lexical-react/src/LexicalAutoLinkPlugin.ts
```diff
@@ -136,138 +170,281 @@ function handleLinkCreation(
...
+    linkNode.splice(0, linkNode.getChildrenSize(), splittedNodes);
+  }
+
+  handleUpdateUrl(linkNode, match, onChange);
+}
+
+function findLast<T, U extends T>(
+  array: T[],
+  predicate: (value: T) => value is U,
+): U | undefined {
```
acywatson, line 319: no explicit undefined - use null

ivailop7:
PR should solve: https://github.com/facebook/lexical/issues/4692

ivailop7:
@NozomiSugiyama have you had a chance to look into Acy's comments?

ivailop7:
Hey @acywatson , @NozomiSugiyama hasn't replied recently, so I addressed your comments and pushed them to their branch. Could you have have a look. I changed the .getLatest() to .getWritable() and fixed the double reverse thing. the splitNodes method is used to split the before and after text away from the text converted to a link.

packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -74,13 +75,86 @@ function runElementTransformers(
...
   transformersByTrigger: Readonly<Record<string, Array<TextMatchTransformer>>>,
 ): boolean {
   let textContent = anchorNode.getTextContent();
-  const lastChar = textContent[anchorOffset - 1];
+
+  const {nodes: prevNodes, textContent: prevTextContent} =
+    getAdjacentPrevTextNodesWithAutoLinkNodes(anchorNode);
+  const nodes = prevNodes.concat(anchorNode);
+  textContent = prevTextContent + textContent;
+  const offset = prevTextContent.length + anchorOffset;
```
fantactuka, line 156: All this logic is not required if no transformer matched by the last char, it should be done below

packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -74,13 +75,86 @@ function runElementTransformers(
...
+    }
+  }
+
+  return {
+    nodes: result.reverse(),
+    textContent: result.map((n) => n.getTextContent()).join(''),
+  };
+}
+
+function extractNodes(
```
fantactuka, line 103: This logic can be replaced with RangeSelection.extract() method:

```js
const selection = $createRangeSelection();
selection.anchor.set(anchorNode.__key, anchorOffset, 'text');
selection.focus.set(focusNode.__key, focusOffset, 'text');

selection.extract() // or selection.getNodes() depending on scenario

```

packages/lexical-markdown/src/MarkdownShortcuts.ts
NozomiSugiyama: Absolutely!
The process for checking whether the last character matches the trigger has been moved forward.

packages/lexical-markdown/src/MarkdownShortcuts.ts
NozomiSugiyama: I appreciate your feedback. To replace the call to the extractNodes function, I have attempted the following approach:

From this:
```
const matchedNodes = extractNodes(nodes, startIndex, match[0].length);
```

to this
```
const selection = $createRangeSelection();
const parent = anchorNode.getParent()!;
selection.anchor.set(parent.__key, startIndex, 'text');
selection.focus.set(parent.__key, startIndex + match[0].length, 'text');
```
Is my understanding correct in this regard?

robfig:
I would love to get this in for the next version. The current AutoLinkPlugin is close but not quite able to be used in production for our use case. This would do it.

packages/lexical-react/src/LexicalAutoLinkPlugin.ts
NozomiSugiyama: The splitNodes function is designed to segment a given TextNode array based on the provided index and length.

This function was implemented in light of the following changes
> instead of no longer considering delimiters, AutoLinkNode is now generated or edited by determining whether to combine the TextContent of adjacent TextNode and match it to the URL when generating AutoLinkNode.
> This change makes it AutoLinkNode even when text like url is input by multiple TextNodes

Please feel free to ask if you have further questions or need additional clarification.

ivailop7:
@NozomiSugiyama let's split this PR into Markdown and non-Markdown, in order to move things forward with it.

NozomiSugiyama:
@ivailop7 
I apologize for the repeated delays in my responses. Regarding splitting the PR into Markdown and non-Markdown components, would it be appropriate to extract the changes from packages/lexical-markdown/src/MarkdownShortcuts.ts into a separate PR?

However, I'd like to point out that due to the changes in the AutoLinkPlugin in this PR, the following E2E tests will not pass. Hence, splitting them in a way that ensures all tests pass might be challenging:

```
[chromium] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:683:5 › Markdown › can convert "hello [world](https://www.test.com/)!" shortcut 
[chromium] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:858:3 › Markdown › can type text with markdown 
[chromium] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:960:3 › Markdown › can adjust selection after text match transformer
```
Screenshot 2023-10-05 at 12 50 32

- The main reason I made modifications to the MarkdownShortcuts is that previously, due to the absence of delimiters like spaces, certain text wasn't recognized as URLs. With the current changes, they are now identified as AutoLinks. Consequently, the MarkdownShortcuts does not function correctly when AutoLinks are included in the evaluated text.

ivailop7:
> @ivailop7 I apologize for the repeated delays in my responses. Regarding splitting the PR into Markdown and non-Markdown components, would it be appropriate to extract the changes from packages/lexical-markdown/src/MarkdownShortcuts.ts into a separate PR?
> 
> However, I'd like to point out that due to the changes in the AutoLinkPlugin in this PR, the following E2E tests will not pass. Hence, splitting them in a way that ensures all tests pass might be challenging:
> 
> ```
> [chromium] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:683:5 › Markdown › can convert "hello [world](https://www.test.com/)!" shortcut 
> [chromium] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:858:3 › Markdown › can type text with markdown 
> [chromium] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:960:3 › Markdown › can adjust selection after text match transformer
> ```
> 
> Screenshot 2023-10-05 at 12 50 32
> * The main reason I made modifications to the MarkdownShortcuts is that previously, due to the absence of delimiters like spaces, certain text wasn't recognized as URLs. With the current changes, they are now identified as AutoLinks. Consequently, the MarkdownShortcuts does not function correctly when AutoLinks are included in the evaluated text.

Hey! Right, that's part of the problem with this PR, we can't couple the Markdown Shortcuts Plugin to the AutoLinks Plugin, the two should be able to operate completely independently of each other an not rely on behaviour of the other as much as possible. Let's make the Markdown Shortcuts mergeable, so extracting stuff to packages/lexical-markdown/src/MarkdownShortcuts.ts is OK and the tests need to pass for the Markdown PR and then we can revisit the logic for the Autolinking.

arp:
@ivailop7 @NozomiSugiyama is there any further progress on this, possibly in a different PR? The problem is still happening on the Playground. Maybe I could help to move this forward?

robfig:
I tested out this PR and it works a lot better than the stock AutoLinkPlugin. 
I did find one issue. URLs include trailing punctuation in the link, which is undesirable. A common usage scenario is putting a URL in a sentence, e.g. 

"Open http://localhost:5560/r/#/."

If you type "Open ", paste the URL, and type ".", the dot is linkified. I think a heuristic of excluding any trailing punctuation from the link would be best. 

Thanks

ivailop7:
> I tested out this PR and it works a lot better than the stock AutoLinkPlugin. I did find one issue. URLs include trailing punctuation in the link, which is undesirable. A common usage scenario is putting a URL in a sentence, e.g.
> 
> "Open http://localhost:5560/r/#/."
> 
> If you type "Open ", paste the URL, and type ".", the dot is linkified. I think a heuristic of excluding any trailing punctuation from the link would be best.
> 
> Thanks

This PR is quite old now and it's unlikely to get merged, so I'm going to close it. If you are interested in taking the AutoLink part-only and cleaning it up and raising a new one, happy to give it another go.

Issue #2452, opened 2022-06-17
Bug: $generateNodesFromDOM does not apply styles properly


ChanhuiSeok:
When using $generateNodesFromDOM to convert HTML to Lexical, styles such as `underline` or `align` are not being applied.

In the case of HTML with styles that _changes tag name_ such as `bold`(`<strong>`) or `Italic`(`<em>`), the style looks well when converted to lexical using $generateNodesFromDOM.

However, for HTML with styles that _only add className(e.g.`editor-text-*`)_, such as `underline` or `strikethrough`, the styles don't seem to be applied when converted to lexical.

I referenced the convert code([HTML -> Lexical - lexical doc](https://lexical.dev/docs/concepts/serialization#html---lexical)).

Lexical version: 0.3.5

## Steps To Reproduce
(* example HTML string - No style applied to text(`STYLE TEST`))
`<div class="editor-input" contenteditable="true" spellcheck="true" data-lexical-editor="true" style="user-select: text; white-space: pre-wrap; word-break: break-word;" role="textbox"><p class="editor-paragraph ltr" dir="ltr"><span data-lexical-text="true">STYLE TEST</span></p></div>`

### case 1. bold
1. select **bold** - it is applied to `<strong>` tag.
![스크린샷 2022-06-17 오전 11 05 53](https://user-images.githubusercontent.com/45116772/174210889-dfaab572-5fcc-4610-bd9f-e8d9629a567a.png)
2. If you convert it to lexical based on the html string obtained in step 1 and insert it, you can check the bold style well as shown below.
![스크린샷 2022-06-17 오전 11 14 31](https://user-images.githubusercontent.com/45116772/174211260-5c3116a4-3cdc-47ab-8447-968624c83f54.png)

### case 2. underline
1. select **underline** - The tagname is _not changed_(still `<span>`) and only `editor-text-underline` is added to the classname.
![스크린샷 2022-06-17 오전 11 18 01](https://user-images.githubusercontent.com/45116772/174211768-ffa9b4dc-9cbb-49ac-8ab8-80bad2815bf8.png)
The underline is also visible in the editor.
![스크린샷 2022-06-17 오전 11 20 30](https://user-images.githubusercontent.com/45116772/174211963-11397b63-147b-4fa6-8e2a-a1c249c0af46.png)

2. However, when convert it to lexical based on html obtained in step 1, converted without underline applied.
[...]

PR #4432, opened 2023-04-30
Add HTML deserialization recipe to docs

ivailop7:
I've used the most approved code snippet from the following issue: https://github.com/facebook/lexical/issues/2452 as a base to address the case for full style serialization/deserialization of HTML styles to preserve the HTML<>JSON fidelity.

Please comment on what to modify, add, remove. I'd like to make this snippet as canonical as possible, since most people will copy-paste without any modifications. I'd like to treat it as a piece of code in the library, we just can't add to the core. Should help solve a lot of the Discord questions by providing a link.

### DIFF
--- a/packages/lexical-website/docs/concepts/serialization.md
+++ b/packages/lexical-website/docs/concepts/serialization.md
@@ -380,11 +380,9 @@ function patchStyleConversion(
 
     const backgroundColor = node.style.backgroundColor;
     const color = node.style.color;
-    const fontStyle = node.style.fontStyle;
+    const fontFamily = node.style.fontFamily;
     const fontWeight = node.style.fontWeight;
     const textDecoration = node.style.textDecoration;
-    const textDecorationLine = node.style.textDecorationLine;
-    const textAlign = node.style.textAlign;
 
     return {
       ...originalOutput,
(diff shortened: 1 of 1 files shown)


packages/lexical-website/docs/concepts/serialization.md
```diff
@@ -298,3 +298,118 @@ export type SerializedTextNodeV2 = Spread<
...
+
+You need to override the base TextNode:
+
+```js
+const initialConfig: InitialConfigType = {
+    namespace: 'editor',
+    theme: editorThemeClasses,
+    onError: (error: any) => console.log(error),
+    nodes: [
+      ExtentedTextNode,
```
zurfyx, line 313:
```suggestion
      ExtendedTextNode,
```

packages/lexical-website/docs/concepts/serialization.md
```diff
@@ -298,3 +298,118 @@ export type SerializedTextNodeV2 = Spread<
...
+  $isTextNode,
+  DOMConversion,
+  DOMConversionMap,
+  DOMConversionOutput,
+  NodeKey,
+  TextNode,
+  SerializedTextNode
+} from 'lexical';
+
+export class ExtentedTextNode extends TextNode {
```
zurfyx, line 334:
```suggestion
export class ExtendedTextNode extends TextNode {
```

packages/lexical-website/docs/concepts/serialization.md
```diff
@@ -298,3 +298,118 @@ export type SerializedTextNodeV2 = Spread<
...
+  constructor(text: string, key?: NodeKey) {
+    super(text, key);
+  }
+
+  static getType(): string {
+    return 'extended-text';
+  }
+
+  static clone(node: ExtentedTextNode): ExtentedTextNode {
+    return new ExtentedTextNode(node.__text, node.__key);
```
zurfyx, line 344:
```suggestion
    return new ExtendedTextNode(node.__text, node.__key);
```

zurfyx (commented):
Thank you, LGTM! Will delegate the stamp to @acywatson since he's worked the most in this space

ivailop7:
@acywatson also wanted to explore what the equivalent for a Heading override would look like and what the difference would be and if there's an alternative to duplication of the logic, besides aggregating all node overrides going to the same node.

ivailop7:
> This seems to work only for spans, not for cases like `<strong style="color: red">foo\</strong\>` (to reproduce, make a piece of text bold and change the color of a part of that bold text).

you are right, didn't realize it changes the span to strong 🤦🏻‍♂️

packages/lexical-website/docs/concepts/serialization.md
```diff
@@ -298,3 +298,114 @@ export type SerializedTextNodeV2 = Spread<
...
+    };
+  }
+
+  static importJSON(serializedNode: SerializedTextNode): TextNode {
+    return TextNode.importJSON(serializedNode);
+  }
+
+  exportJSON(): SerializedTextNode {
+    return super.exportJSON();
+  }
```
thorn0, line 364: These lines look unnecessary.

ivailop7:
@acywatson Should I be doing the bold, italic, etc or just keep the span, strong, em and code ones?

ivailop7:
@fantactuka @acywatson @zurfyx can I get an approval to merge on this one. I've cleaned it up and it's the docs, can always be improved over time.

acywatson (approved):

Issue #4437, opened 2023-05-01
Feature: Add Table Cell writing mode


IBastawisi:
Add Table Cell writing mode (horizontal - vertical)

The `writing-mode` CSS property sets whether lines of text are laid out horizontally or vertically.

https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode 

![Screenshot 2023-05-01 140251](https://user-images.githubusercontent.com/13211445/235443923-38b87556-beb0-4c50-8251-d229b605e45a.png)

PR #4438, opened 2023-05-01
Add TableCell writingMode #4437

IBastawisi:


packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -68,6 +71,7 @@ export class TableCellNode extends DEPRECATED_GridCellNode {
     );
     cellNode.__rowSpan = node.__rowSpan;
     cellNode.__backgroundColor = node.__backgroundColor;
+    cellNode.__writingMode = node.__writingMode;
```
zurfyx, line 74: In retrospective, having a `__backgroundColor` property isn't great. That was my bad, for some reason I thought that cells would never need any more styling properties and that they could be done on child nodes.

Can we copy TextNode `getStyle()`? Maybe the caching part isn't too critical for now but the property is important.

The reason why we have to do this before we merge your PR is backward compatibility. Once we commit to supporting this prop it will be hard to roll it back.

The second (minor) is because we probably don't want to commit on it officially until we address the overflow-x issue somehow:

Screenshot 2023-05-01 at 10 57 45 AM

zurfyx (changes requested):
Thank you! Let me start by saying that this is a nice feature so thank you for working on this

packages/lexical-table/src/LexicalTableCellNode.ts
IBastawisi: I agree that all CSS styles should go in one node property.
regarding the overflow issue we can hide the table cell overflow and also inherit the tr height to allow cell resizing:
```css
.PlaygroundEditorTheme__tableCell {
    height: inherit;
    overflow: hidden;
}
```
I might give it a try soon.

IBastawisi:
The previous commit breaks Cell background color feature disabled test which is expected.
should we deprecate the backgroundColor property or drop it altogether?

IBastawisi:
oh the serialized style property is breaking some unit tests as well, my bad!
image

PR #4445, opened 2023-05-02
exportDOM methods for HeadingNode, QuoteNode

ivailop7:
Adding exportDOM methods for the HeadingNode, QuoteNode (rich-text package). All of the changes are outside of the core, so adding the methods should be safe performance-wise.

Before:

https://user-images.githubusercontent.com/7893468/235786032-a85ee4d8-16e4-4b85-ac3c-583c3d09c000.mp4

After:

https://user-images.githubusercontent.com/7893468/235786043-b8cfe8ae-ecfb-41fb-be27-ae964334d54f.mp4

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -10,9 +10,11 @@ import type {ListNode} from './';
 import type {
   DOMConversionMap,
   DOMConversionOutput,
+  DOMExportOutput,
   EditorConfig,
   EditorThemeClasses,
   GridSelection,
+  LexicalEditor,
   LexicalNode,
   NodeKey,
   NodeSelection,
@@ -120,6 +122,29 @@ export class ListItemNode extends ElementNode {
         conversion: convertListItemElement,
         priority: 0,
       }),
+    };
+  }
+
+  exportDOM(editor: LexicalEditor): DOMExportOutput {
+    const {element} = super.exportDOM(editor);
+
+    if (element) {
+      const formatType = this.getFormatType();
+      element.style.textAlign = formatType;
+
+      const direction = this.getDirection();
+      if (direction) {
+        element.dir = direction;
+      }
+
+      const indent = this.getIndent();
+      if (indent > 0) {
+        element.style.textIndent = `${indent * 20}px`;
+      }
+    }
+
+    return {
+      element,
     };
   }
 
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -10,11 +10,9 @@ import type {ListNode} from './';
 import type {
   DOMConversionMap,
   DOMConversionOutput,
-  DOMExportOutput,
   EditorConfig,
   EditorThemeClasses,
   GridSelection,
-  LexicalEditor,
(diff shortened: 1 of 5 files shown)


packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -123,6 +125,29 @@ export class ListItemNode extends ElementNode {
...
+      element.style.textAlign = formatType;
+
+      const direction = this.getDirection();
+      if (direction) {
+        element.dir = direction;
+      }
+
+      const indent = this.getIndent();
+      if (indent > 0) {
+        element.style.textIndent = `${indent * 20}px`;
```
acywatson, line 142: I don't think any of these apply to list item, but indent especially will not do what we want here, as indent has a different implementation in lists than in elements generally.  Also I'm not sure we should be using textIndent anywhere, as it only indents the first line of the text.

packages/lexical-rich-text/src/index.ts
```diff
@@ -280,6 +306,32 @@ export class HeadingNode extends ElementNode {
...
   }
+
+  exportDOM(editor: LexicalEditor): DOMExportOutput {
+    const {element} = super.exportDOM(editor);
+
+    if (element && this.isEmpty()) {
+      element.append(document.createElement('br'));
+    }
+    if (element) {
+      const formatType = this.getFormatType();
```
acywatson, line 317: This one and quote node seem fine except for the indent thing - I'm not sure why we're using textIndent for this in the core.

packages/lexical-list/src/LexicalListItemNode.ts
fantactuka: Agree that it does not make sense for lists, as for textIndent for other nodes seems like this one is the explanation:
https://github.com/facebook/lexical/blob/aabb5b6ae91193071c1c8b7c53891da49e4c6f83/packages/lexical/src/nodes/LexicalParagraphNode.ts#L82-L85

ivailop7: Thank you @acywatson and @fantactuka will remove the textIndent part, I actually didn't realize it was even a thing outside of lists. What would be the correct way to solve this for lists, where should I be looking?

ivailop7:
@acywatson Could I get a seond look. I've removed the list item part.

packages/lexical-list/src/LexicalListItemNode.ts
acywatson: @fantactuka 

> seems like this one is the explanation:

Yea but if you look at the docs, these do completely different things:

https://developer.mozilla.org/en-US/docs/Web/CSS/text-indent

https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-start

I don't understand how that text-indent property is working the same way as padding-inline-start.

@ivailop7 

> where should I be looking to solve the alignment on an individual list item level.

Alignment in what sense? List indentation is semantic rather than stylistic in Lexical.

acywatson (approved):

packages/lexical-list/src/LexicalListItemNode.ts
ivailop7: as in having a single item from a list have  justification "right", while the others remain with "left" justification.

PR #4447, opened 2023-05-03
Handle a = null in LexicalClickableLinkPlugin

jjjkkkjjj:
I fixed the bug: `TypeError: Cannot read property 'href' of null`.
This bug may be caused by `a` variable that can be `null`. (`domGetParent ` function can return `null`)
That's why `a` variable has no attribute `href`.

I modified `a?.href` and `a?.target` using `?` operator instead of `a.href` and `a.target`.
When the optional `?` operator is used, `a` variable can be `undefined`. So, I modified next condition with `!url` instead of  `url === null || url === ''`.


zurfyx (commented):
Thanks! Can you expand on the issue please? Either add some video on the description or file an issue?

jjjkkkjjj:
@zurfyx 
Sorry for late response...
Thank you for checking my PR! You've already merged the similar PR (#4514), I closed this PR.

PR #4449, opened 2023-05-03
Draft: Implement escape characters for text format transformers

themagickoala:
An attempt at allowing TextMatchTransformers to specify escape characters.

On export, will add in escape characters (if no formatting has been applied already).

On import, will strip out escape characters and not apply the transformer formatting.

I'm almost certain I've missed some behaviours here. I've added a very basic test case, but could probably do with adding more!

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -161,6 +161,11 @@ function exportTextFormat(
       if (!hasFormat(nextNode, format)) {
         output += tag;
       }
+    } else if (transformer.escapeCharacters && applied.size === 0) {
+      output = output.replaceAll(
+        tag,
+        `${transformer.escapeCharacters[0]}${tag}`,
+      );
     }
   }
 
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -242,30 +242,33 @@ function importTextFormatTransformers(
   const transformer = textFormatTransformersIndex.transformersByTag[match[1]];
   let escapedChars = 0;
   for (const char of transformer.escapeCharacters ?? []) {
-    if (
-      leadingNode &&
-      leadingNode.getTextContent().indexOf(char) ===
-        leadingNode.getTextContent().length - char.length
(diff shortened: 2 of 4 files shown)


fantactuka:
I think we should go ahead and split markdown into two separate packages:
@lexical/markdown-shortcuts that will only handle as-you-type conversion
@lexical/commonmark or @lexical/remark for import/export with external library dependency. Attempting to do import/export with regexp will never be 100% precise, and proper tokenization is the only way (although at cost of code size)

acywatson:
> I think we should go ahead and split markdown into two separate packages:
> 
> * `@lexical/markdown-shortcuts` that will only handle as-you-type conversion
> * `@lexical/commonmark` or `@lexical/remark` for import/export with external library dependency. Attempting to do import/export with regexp will never be 100% precise, and proper tokenization is the only way (although at cost of code size)

I wonder if we can keep a single package, just expose different utilities from it. I understand the idea of not having the large 3P dependencies, but structured properly those should be tree-shaken out for anyone not using exports that depend on them.

kevinansfield:
> I think we should go ahead and split markdown into two separate packages:
>
> @lexical/markdown-shortcuts that will only handle as-you-type conversion
> @lexical/commonmark or @lexical/remark for import/export with external library dependency. Attempting to do import/export with regexp will never be 100% precise, and proper tokenization is the only way (although at cost of code size)

FWIW that's pretty much what we ended up doing in our editor because regex based matching wasn't reliable or complete enough.

On [plain text paste events](https://github.com/TryGhost/Koenig/blob/main/packages/koenig-lexical/src/plugins/KoenigBehaviourPlugin.jsx#L976) we first [run the text through markdown-it with the same setup/plug-ins we use elsewhere then effectively "paste" the resulting HTML](https://github.com/TryGhost/Koenig/blob/main/packages/koenig-lexical/src/plugins/MarkdownPastePlugin.jsx#L53-L57) so that it's run through all of our node's `importDOM` methods. Felt like that way we can keep the `importDOM` methods as a more central place for handling conversion too.

themagickoala:
FYI @fantactuka I'm no longer working this PR as I've fully switched over to lexical-remark. Feel free to close it.

PR #4450, opened 2023-05-03
Allow to pass all possible attributes to a content editable div

kazimirchuk:
The issue
----------
Currently the `ContentEditable` component is misleading because the interface for its props makes a union with all possible HTML attributes like so `} & React.AllHTMLAttributes<HTMLDivElement>` but at the same time doesn't pass them to the actual div its creating.

Essentially if you write a code: `<ContentEditable autoCorrect="off" />` typescript won't complain because it's within the interface. However the attribute itself isn't passed down into a `div` so that's where the issue at.

Solution
--------
I assume that it's totally fine to just pass the rest of the attributes for the lack of better context

### DIFF
--- a/packages/lexical-react/src/LexicalContentEditable.tsx
+++ b/packages/lexical-react/src/LexicalContentEditable.tsx
@@ -45,6 +45,7 @@ export function ContentEditable({
   style,
   tabIndex,
   'data-testid': testid,
+  ...rest
 }: Props): JSX.Element {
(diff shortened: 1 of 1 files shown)


thegreatercurve (commented):
Can you put the spread at the start of the doc? I don't want people being able to override the attributes we are controlling in the component.

thegreatercurve (approved):

PR #4453, opened 2023-05-04
Add documentation for TabIndentationPlugin

mrPaintMan:
Simple addition to the documentation for plugins regarding the TabIndentationPlugin, which corresponds to the functionality that was removed on this PR #2855.

### DIFF
--- a/packages/lexical-website/docs/react/plugins.md
+++ b/packages/lexical-website/docs/react/plugins.md
@@ -115,6 +115,14 @@ React wrapper for `@lexical/table` that adds support for tables
 <TablePlugin />
 ```
 
+### `LexicalTabIndentationPlugin`
+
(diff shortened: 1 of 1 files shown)


packages/lexical-website/docs/react/plugins.md
```diff
@@ -115,6 +115,14 @@ React wrapper for `@lexical/table` that adds support for tables
 <TablePlugin />
 ```
 
+### `LexicalTabIndentationPlugin`
+
+Plugin that allows tab indentation in combination with `@lexical/rich-text`.
```
acywatson, line 120: Really appreciate this PR. One small request is to mention @lexical/list here, which is another common use case. Bonus points for mentioning TabIndentationPlugin in lexical-list/README.md

PR #4454, opened 2023-05-04
fix: attach event listeners to initial nodes

### LINKED ISSUE #4213: Bug: NodeEventPlugin does not work for nodes created in initialConfig.initialEditorState
Lexical version: 0.8.1

## Steps To Reproduce

1. Pass in an initialEditorState that creates nodes and appends them to root
2. Use a NodeEventPlugin handler to print to console when a node is selected
3. Observe that the plugin does not fire on the initial editor state -- but does fire on new nodes you create during the editing session

I believe this is happening because the nodes are created before there's an active editor session, so the mutations [here](https://github.com/facebook/lexical/blob/230903f0f536b8d1a84f2ca43dbb390e9a7f9063/packages/lexical-react/src/LexicalNodeEventPlugin.ts#L48) are not triggered to add these nodes to the list of registered elements.

I don't have too much context to explore a fix here -- but maybe a useEffect could sweep through nodes created from the initialConfig and attach the listeners.

OriginalEXE:
fixes #4213 

👋🏻 Some explanation:

- in the previous version listeners were not cleaned up because they would be added only once, and would then be cleaned up when the nodes are removed
- in this new version I've implemented listeners cleanup because `useLayoutEffect` runs twice during development which results in initial nodes getting iterated over twice. Rather than to hack-ily keep track of how many times `useLayoutEffect` ran with the same deps, I've made sure it can run as many times as it wants
- I've cleaned up some comments

### DIFF
--- a/packages/lexical-react/src/LexicalNodeEventPlugin.ts
+++ b/packages/lexical-react/src/LexicalNodeEventPlugin.ts
@@ -12,6 +12,7 @@ import {
   type LexicalEditor,
   type LexicalNode,
   type NodeKey,
+  $nodesOfType,
 } from 'lexical';
 import {useRef} from 'react';
 import useLayoutEffect from 'shared/useLayoutEffect';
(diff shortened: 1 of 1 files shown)


zurfyx (approved):
LGTM, thank you!

PR #4455, opened 2023-05-04
Disable flaky test

zurfyx:
This test is flaky everywhere really, any objections towards disabling it for now?


acywatson (approved):
I'm OK - let's be more aggressive with this for now because we need confidence in the CI. Flaky test doesn't help us much.

PR #4462, opened 2023-05-05
Sanitize floatingLinkEditor

acywatson:
Previously you could cause an anchor to be rendered with javascript: url in this editor via.


zurfyx (approved):
Does it make sense to export this function from @lexical/link since we have the same logic there?

acywatson:
> Does it make sense to export this function from @lexical/link since we have the same logic there?

I just wasn't sure I really wanted a url sanitizer to be part of Lexical's public API.

Issue #4466, opened 2023-05-07
Redundant spaces on Paste


zurfyx:
This is a pretty broad issue for it's easy to repro with Apple notes

```
line 1
 line 2

paragraph 1
paragraph 2
```

Pastes 

```
line 1
 line 2


paragraph 1
paragraph 2
```

Issue #4370, opened 2023-04-20
Bug: Additional spaces appear when pasting bullet list Rich Text from OneNote


olafinpl:
Hi, We've found a weird behavior:

Lexical version: 0.10.0, happens in playground.lexical as well.

## Steps To Reproduce

1. Create text with bullet list in OneNote
2. Paste to Lexical rich text

## The current behavior
When copying text from OneNote, especially bullet list, random spaces appear:
One Note:
![image](https://user-images.githubusercontent.com/131342013/233291082-49cb4d83-4e1c-47d0-9b1b-0ebd007344dc.png)
Lexical playground:
![image](https://user-images.githubusercontent.com/131342013/233291447-966b32e4-0d83-46b5-82df-39350881d8b7.png)

## The expected behavior
No additional white spaces added while pasting.

acywatson:
Thanks for reporting - is there any way that you could help me out by providing the HTML contents of your clipboard when you repro this? I don't have easy access to OneNote to be able to do it myself.

Slapbox:
@olafinpl if you're on Windows I recommend Clipview to explore exactly what is in your clipboard when this occurs - then you can share that here. It will likely be in at least two different formats, eg plaintext and html or plaintext and richtext.

acywatson:
> I used _Free Cliboard Viewer 4.0_ and it gives HTML:
> 
> Version:1.0 StartHTML:0000000105 EndHTML:0000001962 StartFragment:0000000527 EndFragment:0000001922
> 
> `<html xmlns:o="urn:schemas-microsoft-com:office:office" xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882" xmlns="http://www.w3.org/TR/REC-html40">`
> 
> `<head>` `<meta http-equiv=Content-Type content="text/html; charset=utf-8">` `<meta name=ProgId content=OneNote.File>` `<meta name=Generator content="Microsoft OneNote 15">` `</head>`
> 
> `<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>` `<!--StartFragment-->`
> 
[...]

Issue #3677, opened 2023-01-05
Bug: Additional break line when initializing or copy/pasting in the editor with break lines


ErPacheco:
When I initialize the Lexical editor from an DOM like `<p><br></p>`, passing through `generateNodesFromDOM` and inserting the generated nodes to the editor, it imports it with an additional break line like `<p><br><br></p>`. This makes the editor have more break lines than expected. It is also curious that both <br> count as one character if I delete them and count the characters.

@zeitstein mentioned in a discord chat that when trying to pass a single `<p></p>` without break lines, it also imports it with an additional break line like `<p><br></p>`. They also mention that that this PR can be related -> https://github.com/facebook/lexical/pull/3581 (thanks for rising this up in the discord chat, here is the full thread: https://discord.com/channels/953974421008293909/1060331843578249376)

I also managed to reproduce it when copying some text with break lines from the Notes Application from my Mac.

Lexical version: 0.7.6

## Steps To Reproduce

1. Write something like this (with some break lines) in a Note in the Mac Notes Application:
```
First line



Fifth line
```
2. Copy the five lines (or the block of text you have written) and paste it in the lexical playground editor
3. You can find that there are more than 5 lines in this particular case.
<img width="1177" alt="image" src="https://user-images.githubusercontent.com/26458400/210821038-a7e4d114-222c-4034-9cb5-5a73be0fb2d1.png">

-----------------

Here is an example of the InitialState plugin to initialize the editor from a preview HTML string (not sure why is duplicating the text in the codesandbox, in my code only runs once): https://codesandbox.io/s/break-line-example-w1ic2x?file=/src/plugins/InitialState.js

## The current behavior
When copying or initializing some text with break lines, it duplicates them, making the text different from the expected one.

## The expected behavior
When copying or initializing some text with break lines, the number of break lines is exactly the same as the one inserted in the editor.

rabishah:
I'm facing the same issue.

chalecki:
I'm facing the same issue. I've noticed `$generateNodesFromDOM` is adding an extra `<br />` tag for each "empty" paragraph. If you put any character, eg. period or even empty space - the problem does not exist

rabishah:
Facing the same issue, exactly as @chalecki .

willwang888:
Facing the same issue!

chalecki:
I think it was introduced in #590 in commit f38d4d52f24abc26d59c73989fc92c36fab19f8d . 

[...]

PR #2276, opened 2022-05-26
Improve copy + paste logic

trueadm:
This extracts improvements out of https://github.com/facebook/lexical/pull/2241.

PR #4467, opened 2023-05-07
Tame wild white spaces

zurfyx:
# W3-inspired white space handling on paste

This PR attempts to resolve for once and for all the commonly repeated issues around spacing on paste. Turns out there's a well-defined spec (https://www.w3.org/TR/css-text-3/#white-space-processing) for what seemed to be wild-west across various editor implementations.

https://user-images.githubusercontent.com/193447/236697703-827d0e66-2a09-40ce-aaa7-f10cb8be1391.mov

https://user-images.githubusercontent.com/193447/236697707-512b8ee4-ca84-4322-88e6-2769f323a0e3.mov

## W3 summary

https://www.w3.org/TR/css-text-3/#white-space-processing

When white-space is `pre` or similar variants, we don't do any special processing, we assume the incoming text reads as the rendered copy. This is the case for Apple Notes, Chrome, Safari and GDoc.

Otherwise, we transform the text according to collapsible rules. This applies to MS Word, Quip and Notion. Firefox [fails W3 rules](https://codepen.io/zurfyx/pen/KKGvvNV) but it also falls under this bucket.

At a high level, collapsibles mean removing redundant spaces/LF/CRLF/tab:
- At the start/end of the line
- Redundant consequent
- Transformation of LF, CRLF and tab into spaces

The current implementation respects the historical space transform listed in W3 (instead of removal).

## BR rule

This PR also touches BR. The spec (https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element) refers to a "single" element but I'm fairly confident this is after the collapsibles.

## But this PR is not 100% compliant

There is quite a lot CSS to understand to make it 100% compliant (and some of these properties even go beyond what Lexical supports at this point) but the implemented should work in most cases.

For example, `pre`, `pre-wrap` and `pre-line` differences, understanding the tab-size to determine how it should render, display flex and table, or dropping Space Marks. I didn't prioritize any of the aforementioned because no other editor supports any these, they just handle them gracefully.

## Performance

When these rules apply (bucket 2 above), Lexical obviously runs slower as it has to traverse backward and forward accordingly, slice and compute styles. 

~~Pasting the [Moby Dick](https://www.hogbaysoftware.com/posts/moby-dick-workout/) test from Apple Notes results in a 10% regression, from 1.53 to 1.68s on a 2019 MBP.~~

Edit: dropped `getComputedStyles` as we don't expect clipboard users to make use of StyleSheets. And, fixed cache.

Pasting the [Moby Dick](https://www.hogbaysoftware.com/posts/moby-dick-workout/) test from Apple Notes results in a 0.7% regression, from 4.27s to 4.30s on a M1 with x6 slowdown.

| Before | After |
| ----     | ----- |
| Screenshot 2023-05-15 at 2 58 12 PM | Screenshot 2023-05-15 at 2 58 08 PM |

## Bundle size

Arguably this increases the size of TextNode considerably, but eventually (as discussed with @acywatson) all HTML code will be decoupled from Core. Even without this PR, this would have a drastic bundle size decrease for our plain text users.

## Implementation: why on TextNode vs a pre-cleanup?

I've discussed this one offline with @fantactuka but ultimately this one is very opinionated. The alternative to this PR would be a cleanup process before passing the DOM Nodes to the LexicalNode importDOM.

The pre-cleanup assumes that the incoming HTML is bad, but it's actually valid correct HTML, just not in the shape we like. We can add this one-off to handle spacing, but then there will be more, ultimately leading to allowing an array of cleanup steps into the plugin when there was no need to do this in the first place.

Instead, the Nodes themselves can take responsibility just like they are doing now. This might not be as optimal as the one-pass approach but the implementation is still fast considering we never look beyond inlines and that, in most cases, they won't be filled with fragmented white spaces, so average case will still be O(N).

---

Other relevant PRs:

- https://github.com/facebook/lexical/pull/2969
- https://github.com/facebook/lexical/pull/2276

Fixes https://github.com/facebook/lexical/issues/4466
Fixes https://github.com/facebook/lexical/issues/3677
Fixes https://github.com/facebook/lexical/issues/4370


acywatson (approved):

fantactuka (approved):
🫡

birtles:
@zurfyx I hope you don't mind me asking here before filing a new issue. v0.11.0 is breaking some of our copy-paste roundtrip tests where we're using `\xA0` (`&nbsp;`) as a non-collapsible whitespace character when copying.

As far as I can tell, the spec you pointed to here doesn't allow collapsing non-breaking space characters. Specifically it has:

> Except where specified otherwise, white space processing in CSS affects only the document white space characters: spaces (U+0020), tabs (U+0009), and [segment breaks](https://www.w3.org/TR/css-text-3/#white-space-processing).

I couldn't find anywhere that allows collapsing (U+00A0).

Looking at the patch, we're using the `\s` character class in a number of places, e.g.

```js
  textContent = textContent
    .replace(/\r?\n|\t/gm, ' ')
    .replace('\r', '')
    .replace(/\s+/g, ' ');
```

However, `\s` expands to `[\f\n\r\t\v\u0020\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`. Is that intended, or shall I file a bug for this?

PR #4478, opened 2023-05-10
Add text/uri-list support in clipboard.

### LINKED ISSUE #4409: Bug: Support for text/uri-list type when pasting
When opening a link from a mobile app in IOS (specifically I was testing with the linkedin mobile app, but I think text/uri-list is a somewhat common type) and pasting into my app which uses lexical-- the pasted text doesn't appear. The meta type of the pasted text is 'text/uri-list' which seems to be unsupported. I'd expect it to handle this case.

Lexical version: 

## Steps To Reproduce

1. On IOS, go to a post in the linkedin mobile app and click the share button and select 'Copy' from the IOS menu
2. Paste that text into an app that uses lexical 

Link to code example: happens in the playground

## The current behavior
Text does not paste

## The expected behavior
Text pastes a link

xinyuan0801:
## Background
Issue #4409 highlights a problem where users are unable to paste text/uri-list type content. This PR aims to address and resolve this issue.

## Cause of the Problem
The problem is due to discrepancies in the valid MIME types for read/write operations in the clipboard between Chromium and WebKit. 

Chromium supports the read/write operations for the following MIME types [(Chromium)](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/clipboard/clipboard_writer.cc;drc=e882b8e4a8272f65cb14c608d3d2bc4f0512aa20;l=304):
- `text/plain`
- `text/html`
- `image/png`

On the other hand, WebKit supports all the above types as well as `text/uri-list` [(WebKit)](https://webkit.org/blog/10855/async-clipboard-api/).

This inconsistency causes an issue when, for example, a user opens a webpage in Safari on iOS, copies content via the share menu, and tries to paste it elsewhere. In this case, the MIME type of the content in the clipboard is `text/uri-list`.

However, based on the current logic in `$insertDataTransferForRichText` in [`clipboard.ts`](https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts), only `text/plain`, `text/html`, `image/png`, and `'application/x-lexical-editor'` are accepted for pasting. Hence, when a user attempts to paste text of type `text/uri-list`, no content is pasted.

## Proposed Changes
This PR proposes a modification to the `$insertDataTransferForRichText` function in [`clipboard.ts`](https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts) to support pasting of `text/uri-list` MIME type.

## Impact of Changes
This change will allow pasting of `text/uri-list` type content in Safari. However, given Chrome's current lack of support for reading/writing `text/uri-list` in the clipboard, the behaviour of pasting this MIME type in Chrome remains unchanged.

Fixes #4409

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -130,22 +130,35 @@ export function $insertDataTransferForRichText(
   // instead of single paragraph with linebreaks.
   const text = dataTransfer.getData('text/plain');
   if (text != null) {
-    if ($isRangeSelection(selection)) {
-      const parts = text.split(/(\r?\n|\t)/);
-      const partsLength = parts.length;
-      for (let i = 0; i < partsLength; i++) {
-        const part = parts[i];
-        if (part === '\n' || part === '\r\n') {
-          selection.insertParagraph();
-        } else if (part === '\t') {
-          selection.insertNodes([$createTabNode()]);
-        } else {
-          selection.insertText(part);
-        }
+    $pasteMultilineText(selection, text);
+  }
+
+  // Webkit-specific: Supports read 'text/uri-list' in clipboard.
+  const uriListString = dataTransfer.getData('text/uri-list');
+  if (uriListString != null) {
+    $pasteMultilineText(selection, uriListString);
+  }
+}
+
+function $pasteMultilineText(
+  selection: RangeSelection | GridSelection,
+  text: string,
+) {
+  if ($isRangeSelection(selection)) {
+    const parts = text.split(/(\r?\n|\t)/);
+    const partsLength = parts.length;
+    for (let i = 0; i < partsLength; i++) {
+      const part = parts[i];
+      if (part === '\n' || part === '\r\n') {
+        selection.insertParagraph();
+      } else if (part === '\t') {
+        selection.insertNodes([$createTabNode()]);
+      } else {
+        selection.insertText(part);
       }
-    } else {
-      selection.insertRawText(text);
     }
+  } else {
+    selection.insertRawText(text);
   }
 }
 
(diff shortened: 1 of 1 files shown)


zurfyx (commented):
LGTM, thank you! But I'd be curious to know more about the use case. I wasn't able to repro this from my LinkedIn app, Pasteboard shows type `public.url` and it isn't retrievable via `text/uri-list`

You also need to update `$insertDataTransferForPlainText` for plain text users support

xinyuan0801:
> LGTM, thank you! But I'd be curious to know more about the use case. I wasn't able to repro this from my LinkedIn app, Pasteboard shows type `public.url` and it isn't retrievable via `text/uri-list`
> 
> You also need to update `$insertDataTransferForPlainText` for plain text users support

Hi, i believe this has something to do with the copy operation in the share pop-up menu in IOS. i am able to reproduce by using the copy in share pop-up menu of Safari, Linkedin and instagram in IOS.

Below is the screen recording of the repro in Safari and Linkedin app

https://github.com/facebook/lexical/assets/50743260/bc4fec9a-08a5-4f16-8b81-1a4c62c36581

https://github.com/facebook/lexical/assets/50743260/f2269116-3a89-4347-a031-8cd08eaf1183

Sorry my phone's language is Chinese so it might be not so clear, but what i did is just use the copy from ios share pop-up menu and paste it in lexical playground. you can see from the log that there is a Clipboard Event but nothing is pasted. This also happens in instagram when trying to share in the same way and I believe it happens as long as user are using the copy operation provided by IOS native share pop-up menu. 

After the fix, it was able to paste the content from clipboard that are from the share menu pop-up. 


https://github.com/facebook/lexical/assets/50743260/d956bad2-bb76-4cdb-8c8a-c3b5b3fe664e

zurfyx (approved):
Thanks for the videos, appreciate it!

Issue #4474, opened 2023-05-09
Bug: Editor is focused when the initialConfig editorState converts from markdown and registerNodeTransform has been used


darsee:
In the codesandbox linked below, the editor is focused on mount when the `initialConfig` `editorState` converts a markdown heading and `registerNodeTransform` has been used. Is this expected?

Lexical version: 0.10.0

## Steps To Reproduce

1. Open the codesandbox below
2. Observe that the editor is initially focused (cursor is visible)
3. Change `value` to `test` (uncomment line 27)
4. Reload the browser preview, observe that the editor is _not_ focused (cursor is not visible)
6. Change `value` back to `# test` and comment out `<TestPlugin />`
7. Reload the browser preview, observe that the editor is _not_ focused (cursor is not visible)

Link to code example:

https://codesandbox.io/s/lexical-editor-focused-after-node-transform-sfm1wn?file=/src/Editor.js

## The current behavior

On reloading the browser preview, the editor is initially focused (cursor is visible).

## The expected behavior

I expect the editor not to be initially focused unless using `@lexical/react/LexicalAutoFocusPlugin`?

darsee:
Hey @aquarius-wing, thanks for the PR!

I don't think it solves the issue for me though - the problem in the scenario above is that the editor _is_ focused when it loads but I am expecting it _not_ to be focused until I click it. Your PR (if I understand correctly) causes the editor to be focused even if the markdown conversion doesn't match on anything - I'm not sure if this behaviour would be expected.

aquarius-wing:
So, is it possible that I misunderstood your meaning? I thought you wanted the focus to be only when text is inputted. So, you mean the opposite, that even when '# text' is inputted, it should not be focused? I'm quite confused. 😂

aquarius-wing:
I feel that it should be up to the lexical official to determine whether a separate configuration item is needed to control whether to auto-focus.

danchurch-alphasights:
Hi @darsee!

I ran into a very similar problem with `$convertFromMarkdownString` recently.

I was able to work around it by:
```Javascript
useEffect(() => {
  // Set the editor to not editable - it won't take focus while in that state
  editor.setEditable(false);

  const unregister = editor.registerNodeTransform(TextNode, (textNode) => {
    console.log("node transformed", textNode);
  });

  // Set the editor back to editable a frame later
  requestAnimationFrame(() => editor.setEditable(true));

  return unregister;
}, [editor]);
```

<details>
[...]

PR #4482, opened 2023-05-10
Auto focus the text when markdown import got match of none

aquarius-wing:
# Background
Fix the issue #4474 
# Cause of Problem
The text is focused only when the markdown transformer matched
But when it doesn't the auto focus feature was lost after the [commit](https://github.com/aquarius-wing/lexical/commit/e817a55a94522c98103401a2620a179cd110b548) was release
# The logic in deep 
1. the markdown transformer matched
2. the lexical getSelection is not null
3. so in commit link above, it can run `root.selectEnd`

So if I want `root.selectEnd` be run, I must make lexical has selection.
That's the why I wrote the code in this PR


fantactuka:
The point of that [commit](https://github.com/aquarius-wing/lexical/commit/e817a55a94522c98103401a2620a179cd110b548) is to call "select" only if selection was non-null before. With the suggested change, it'll always call select() if some markdown conversion happened.

aquarius-wing:
I believe that regardless of whether the markdown transformer matches or not, the editor should be focused.
Otherwise, it may cause confusion as it sometimes focuses and sometimes does not.

fantactuka:
> I believe that regardless of whether the markdown transformer matches or not, the editor should be focused. Otherwise, it may cause confusion as it sometimes focuses and sometimes does not.

Not really, running markdown import does not require editor to be focused and does not mean user need it to be focused. E.g., imagine page for blog post editing with title (input[type=text]) and content field (lexical). When markdown for content is loaded and populated to lexical, user might still want to have focus on title field

aquarius-wing:
So I should delete the line in 
[node.select(0, 0);](https://github.com/facebook/lexical/blob/a82b7547ea158a5472afda0f702ed844ca25fd0a/packages/lexical-markdown/src/MarkdownTransformers.ts#L93)?

PR #4489, opened 2023-05-12
Add mutation observer to lifecycle diagram

EgonBolton:
I thought it was an important concept of the life cycle.
Without that, there are many things that are not clear. For example, does a Grammarly fix trigger the transforms?
Are transformations as reliable as listeners?

Let me know what you think :)


zurfyx:
Thank you Egon! My opionated thought is that MutationObservers is an advanced topic. Most people don't really need to know about them, in fact, they play a small role into Lexical when you compare them to Event handling. MutationObservers were initially created to stop externals from tampering with the editor content and eventually we tweaked them a bit to understand some of the basic mutations.

[MutationObservers do trigger updates](https://github.com/facebookexternal/Outline//blob/1cf4eada0c414f2747bffad7f47dd0668057554a/packages/lexical/src/LexicalMutations.ts#L127) and they follow the same lifecycle as other events or programatic updates.

Technically wise, we disable them during reconciliation to prevent infinite loops cause by Lexical DOM manipulation but that's just an implementation detail in my opinion.

That said, I think it's a good idea to document this but perhaps in a separate advanced section that covers how Lexical works behind the scenes.

EgonBolton:
Thanks for the explanation @zurfyx.
It's not a priority, but I'd love to get the more detailed explanation of this at some point. I have some ideas and some doubts about it.

You can close this PR :)

zurfyx:
> Thanks for the explanation @zurfyx. It's not a priority, but I'd love to get the more detailed explanation of this at some point. I have some ideas and some doubts about it.
> 
> You can close this PR :)

A deep dive into the core, either in the form of text or video is a good idea, we should definitely do this at some point. In the meantime, feel free to raise any of these specific advanced questions over Discord!

PR #4490, opened 2023-05-12
fix: Only append fragment when element is HTMLElement

lsbyerley:
If merged, this PR will add a check to the `$appendNodesToHTML` function in `lexical-html` to only append the fragment if element is an instanceof HTMLElement.

An error occurs at this line when a TextNode override is created and this override returns a TextNode as opposed to an HTMLElement. 

https://github.com/facebook/lexical/blob/d10c4e6e55261b2fdd7d1845aed46151d0f06a8c/packages/lexical-html/src/index.ts#L125

Example CustomTextNode override:

```
class CustomTextNode extends LexicalTextNode {
  static getType() {
    return 'custom-text';
  }

  static clone(node) {
    return new CustomTextNode(node.__text);
  }

  exportDOM() {
    const text = this.__text;
    const element = document.createTextNode(text);
    return { element };
  }
}
```


acywatson:
This is good - we’d also need to change the type of the element property in DOMExportOutput to be HTMLElement | Text

lsbyerley:
> This is good - we’d also need to change the type of the element property in DOMExportOutput to be HTMLElement | Text

Keep the null or replace with text? `element: HTMLElement | Text | null;`

acywatson:
> > This is good - we’d also need to change the type of the element property in DOMExportOutput to be HTMLElement | Text
> 
> Keep the null or replace with text? `element: HTMLElement | Text | null;`

Keep null, good call

acywatson:
@lsbyerley can you sign the CLA?

zurfyx:
Please, do also check the failing tests, it seems like Units are not passing

lsbyerley:
@zurfyx Looking into the unit tests more and I'm not entirely sure the best way to solve this. It appears `element: HTMLElement | Text | null;` on `DOMExportOutput` has created quite the ripple effect. The `LexicalHtml.test.ts` unit test is below, however, the same error appears in all unit tests. pointing to the same 4 lines. Any insight are appreciated

Screenshot 2023-05-19 at 12 59 40 PM

Screenshot 2023-05-19 at 1 02 43 PM

lsbyerley:
I'm not entirely sure what else is failing. With my most recent changes, all unit tests are passing locally for me

lsbyerley:
@acywatson do you have any insight as to why the majority of the e2e tests except a few? https://github.com/facebook/lexical/actions/runs/5026862456/

acywatson:
> @acywatson do you have any insight as to why the majority of the e2e tests are passing except a few? https://github.com/facebook/lexical/actions/runs/5026862456/

Possibly just flakiness - do they fail locally?

acywatson:
@lsbyerley are we still just waiting on the CLA here?

lsbyerley:
@acywatson Correct. Hoping to have CLA approval today or tomorrow. Will update the thread when ready

derekhubbard:
@acywatson The CLA has been signed, and looks like the CLA github bot already picked up on that - nice! 

Let us know if we can help resolve the Vercel deployment issues. It doesn't look like I have access, so not sure how to troubleshoot.

acywatson:
> @acywatson The CLA has been signed, and looks like the CLA github bot already picked up on that - nice!
> 
> Let us know if we can help resolve the Vercel deployment issues. It doesn't look like I have access, so not sure how to troubleshoot.

Hm, looks like there are integrity and unit tests failures - might start with that + a rebase and see if that helps?

lsbyerley:
@acywatson the unit tests are now failing because of a typescript error at the `exportDOM` function in `LexicalTableNode`. I don't recall seeing this before and I'm not sure why it doesn't like it now. Still looking into it

https://github.com/facebook/lexical/blob/b5d73eb2eb715890a6f262b263eeadb239d3137c/packages/lexical-table/src/LexicalTableNode.ts#L93C50-L93C50

Screenshot 2023-08-10 at 12 31 48 PM

lsbyerley:
@acywatson unit tests should be passing now

acywatson:
Awesome - I'm sorry to do this, but I just remembered that we are trying avoid using instanceof because it interferes with SSR in some cases. Instead, for this, we export isHTMLElement from @lexical/utils.

Can we use that?

lsbyerley:
> Awesome - I'm sorry to do this, but I just remembered that we are trying avoid using instanceof because it interferes with SSR in some cases. Instead, for this, we export isHTMLElement from @lexical/utils.
> 
> Can we use that?

No problem! I've updated the PR to use `isHTMLElement` instead

lsbyerley:
@acywatson Looks like there are only a few e2e tests failing.. i can't seem to run them locally so I'm not exactly sure why they are.

derekhubbard:
@acywatson Any update on this? Anything we can do to help get this in?

acywatson (approved):

PR #4492, opened 2023-05-13
fix: Typo in DecoratorNode example

JanCizmar:


zurfyx (changes requested):
Thank you, the current version seems correct, the param should always be provided. Alternatively, `LexicalNode | null | undefined` which is what we use now with TS but it's probably worth revising all examples if we go down this route

JanCizmar:
Oh, so it's flow syntax. I was confused.

PR #4504, opened 2023-05-16
Add Copy & Paste to Context Menu

tylerjbainbridge:


packages/lexical/src/LexicalCommands.ts
```diff
@@ -90,10 +90,12 @@ export const DRAGOVER_COMMAND: LexicalCommand<DragEvent> =
   createCommand('DRAGOVER_COMMAND');
 export const DRAGEND_COMMAND: LexicalCommand<DragEvent> =
   createCommand('DRAGEND_COMMAND');
-export const COPY_COMMAND: LexicalCommand<ClipboardEvent | KeyboardEvent> =
-  createCommand('COPY_COMMAND');
-export const CUT_COMMAND: LexicalCommand<ClipboardEvent | KeyboardEvent> =
-  createCommand('CUT_COMMAND');
+export const COPY_COMMAND: LexicalCommand<
+  ClipboardEvent | KeyboardEvent | null
```
zurfyx, line 94: I remember we already handled the null case on the clipboard code, can you double check this is case? If that's the case this is fine without additional changes
```suggestion
  null | ClipboardEvent | KeyboardEvent
```

packages/lexical-rich-text/src/index.ts
```diff
@@ -417,14 +417,17 @@ function onPasteForRichText(
...
   editor: LexicalEditor,
 ): void {
-  event.preventDefault();
   editor.update(
     () => {
       const selection = $getSelection();
-      const clipboardData =
-        event instanceof InputEvent || event instanceof KeyboardEvent
-          ? null
-          : event.clipboardData;
```
zurfyx, line 427: Do we need these changes here?

packages/lexical-rich-text/src/index.ts
```diff
@@ -999,6 +1003,7 @@ export function registerRichText(editor: LexicalEditor): () => void {
       PASTE_COMMAND,
       (event) => {
         const [, files, hasTextContent] = eventFiles(event);
+        // debugger;
```
zurfyx, line 1006:
```suggestion

```

ivailop7:
I think adding "Paste as plain text" would be a really cool thing to have. Can be added later as well, just sharing thoughts.

thegreatercurve:
> I think adding "Paste as plain text" would be a really cool thing to have. Can be added later as well, just sharing thoughts.

@ivailop7 We have something like this already with the "Clear formatting" button in the drop down which removes all of the BIUS formatting from pasted text.

thegreatercurve:
Closing this for the moment due to inactivity but happy to re-open if the changes can be addressed.

ivailop7:
True, but the request comes more from the aspect that when we paste content from outside, it would override the node, for example a blockquote rather than append text inside, which is where the 'paste as plain text' option is required.

ivailop7:
We'll raise a new PR on this soon.

PR #4507, opened 2023-05-17
Allow arbitrary awareness fields in Collaboration plugin

attnov:
### Summary
This change allows arbitrary data (for example 3rd party UUID) to be added to awareness. Right now user name and color is hard coded in Collaboration plugin. The new parameter `awarenessData` is optional.

### Usage example
We can specify our arbitrary data like this:
```
const awarenessData = {
  uuid: 42,
  foo: 'bar',
);
```

With the new parameter, we can send down our arbitrary data:
```
<CollaborationPlugin
  id="id"
  providerFactory={providerFactory}
  shouldBootstrap={shouldBootstrap}
  awarenessData={awarenessData}
/>
```

### DIFF
--- a/packages/lexical-yjs/src/index.ts
+++ b/packages/lexical-yjs/src/index.ts
@@ -78,7 +78,7 @@ export function initLocalState(
 ): void {
   provider.awareness.setLocalState({
     anchorPos: null,
-    awarenessData: awarenessData,
+    awarenessData,
     color,
     focusPos: null,
     focusing: focusing,
@@ -99,7 +99,7 @@ export function setLocalStateFocus(
   if (localState === null) {
     localState = {
       anchorPos: null,
-      awarenessData: awarenessData,
+      awarenessData,
       color,
       focusPos: null,
       focusing: focusing,


ivailop7 (approved):
Thank you, Attila. @acywatson and @zurfyx . Could you give it a look as well.

fantactuka:
I don't think works since map is not json serializeable, should be plain object like remaining of awareness data. Personal opinion: naming looks a bit odd, as usage will be awareness.getLocalState().awarenessFields, maybe data or extraData would do better, no strong opinion tho

fantactuka (changes requested):
~

packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
```diff
@@ -281,4 +281,35 @@ describe('Collaboration', () => {
...
+    };
+
+    client1.start(container, awarenessFields1);
+    client2.start(container, awarenessFields2);
+
+    await expectCorrectInitialContent(client1, client2);
+
+    expect(client1.awareness.getLocalState().awarenessFields).toEqual(
+      awarenessFields1,
+    );
```
fantactuka, line 307: Could you use awareness.getStates() instead, to ensure that client1 has proper awareness from client2 and vise versa?

packages/lexical-react/src/shared/useYjsCollaboration.tsx
```diff
@@ -225,27 +229,41 @@ export function useYjsFocusTracking(
   provider: Provider,
   name: string,
   color: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  awarenessFields?: object,
```
fantactuka, line 233: I wonder if we can use `{ ... }` here and remove eslint comment above

fantactuka:
Thanks for updates! 

@acywatson @zurfyx thoughts on naming?

acywatson:
> Thanks for updates!
> 
> @acywatson @zurfyx thoughts on naming?

I like having awareness in the name, but fields is probably wrong. "awarenessData" or "extraAwaresnessData" seems fine.

doc block on the interface would help people know what it's for.

packages/lexical-react/src/shared/useYjsCollaboration.tsx
attnov: Actually, we don't need the eslint comments anymore - `any` is not used anymore. Removed all of them.

attnov:
> > Thanks for updates!
> > @acywatson @zurfyx thoughts on naming?
> 
> I like having awareness in the name, but fields is probably wrong. "awarenessData" or "extraAwaresnessData" seems fine.
> 
> doc block on the interface would help people know what it's for.

Renamed the parameter to `awarenessData` and added documentation on the inferface.

packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
attnov: It looks like [here ](https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/__tests__/unit/utils.tsx#L94)the mock for getStates() only inserts the current client's awareness state.

packages/lexical-react/src/__tests__/unit/utils.tsx
```diff
@@ -62,7 +63,8 @@ class Client {
   _listeners: Map<string, Set<(data: unknown) => void>>;
   _updates: Uint8Array[];
   awareness: {
-    getLocalState(): void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    getLocalState(): any;
```
zurfyx, line 67: Can we make this better? Besides the `object` prop what other types we don't know?

zurfyx (approved):

packages/lexical-react/src/__tests__/unit/utils.tsx
attnov: I have added types for getLocalState(), getStates(), setLocalState(), etc.
Thanks for checking!

packages/lexical-yjs/src/index.ts
```diff
@@ -73,9 +74,11 @@ export function initLocalState(
   name: string,
   color: string,
   focusing: boolean,
+  awarenessData: object,
 ): void {
   provider.awareness.setLocalState({
     anchorPos: null,
+    awarenessData: awarenessData,
```
fantactuka, line 81:
```suggestion
   awarenessData,
```

Similar one below

fantactuka (approved):

PR #4510, opened 2023-05-17
Fix GC memory leak

### LINKED ISSUE #4509: Does $nodesOfType access a stale state?
TL;DR: working example with the weird behaviour: https://codesandbox.io/s/stale-nodes-of-type-lwdg3z?file=/src/UpdateTrackerPlugin.tsx

---

In my React app at work I have a plugin that register an update listener and extract all nodes of a certain type to process them outside the editor. Thing is, I noticed that $nodesOfType looks like it's accessing some stale state, while $dfs always reads from the latest available state.

As a minimal reproduction example, in the linked CodeSandbox I have a sample plugin (the UpdateTrackerPlugin) that listens for state updates and collects the CustomNode nodes using both $nodesOfType and $dfs.

Collected nodes texts are then shown in the sections below the text editor.

Is it something I'm not handling on my own? Or a concept I'm missing? Or it's simply some unintended behaviour?

---

Source: morgandam @ https://discord.com/channels/953974421008293909/1103258229988929627

zurfyx:
Turns out we introduced a memory leak on the GC module when we launched doubly linked lists.

The root cause is that now order is important
1. We have to start with ElementNodes first, if a child is removed from the Map there's no way to traverse (next) children later (unless we keep a separate copy).
2. The attached assumption on children node  in `$garbageCollectDetachedDeepChildNodes` was not correct. It will  never be true because the parent is not attached in the first place.

Fixes #4509

### DIFF
--- a/packages/lexical/src/LexicalGC.ts
+++ b/packages/lexical/src/LexicalGC.ts
@@ -47,6 +47,7 @@ function $garbageCollectDetachedDeepChildNodes(
   let child = node.getFirstChild();
 
   while (child !== null) {
+    const nextChild = child.getNextSibling();
     const childKey = child.__key;
     if (child !== undefined && child.__parent === parentKey) {
       if ($isElementNode(child)) {
@@ -66,7 +67,7 @@ function $garbageCollectDetachedDeepChildNodes(
       }
       nodeMap.delete(childKey);
     }
-    child = child.isAttached() ? child.getNextSibling() : null;
+    child = nextChild;
   }
 }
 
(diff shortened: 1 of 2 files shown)


fantactuka (approved):

packages/lexical/src/LexicalGC.ts
```diff
@@ -47,6 +47,7 @@ function $garbageCollectDetachedDeepChildNodes(
   let child = node.getFirstChild();
 
   while (child !== null) {
+    const nextChild = child.getNextSibling();
     const childKey = child.__key;
     if (child !== undefined && child.__parent === parentKey) {
```
fantactuka, line 52: Not related to PR changes, but wondering why would we have undefined check here, it would throw one line above when accessing __key

PR #4511, opened 2023-05-17
Allow more granular priorities

acywatson:
This provides more flexibility in the command priority system by providing more granularity in the priority values. This is useful for cases where core functionality uses a low command priority, but you need that core functionality to run first and only then delegate to your handler. 

A concrete example is the core typeahead plugin, which intercepts Enter events at LOW priority and selects the option if the menu is open. In a case where you also want Enter to submit the form if the menu is NOT open, the only option is to use EDITOR priority, which is reserved for the editor and may cause other collisions. 

I think the most robust solution here is to introduce more granularity. I made the ranges large because we only want to do this one time - if we do it again, we'd risk breaking people who are relying on numbers that fall between the values represented by the constants.

Another option might be to make 10 or 20 constants, instead of 5, and we can do it that way if that's preferred.

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -212,13 +212,56 @@ export type CommandListener<P> = (payload: P, editor: LexicalEditor) => boolean;
 
 export type EditableListener = (editable: boolean) => void;
 
-export type CommandListenerPriority = number;
+export type CommandListenerPriority =
+  | 0
+  | 1
+  | 2
+  | 3
+  | 4
+  | 5
+  | 6
+  | 7
+  | 8
+  | 9
+  | 10
+  | 11
+  | 12
+  | 13
+  | 14
+  | 15
+  | 16
+  | 17
+  | 18
+  | 19
+  | 20;
 
-export const COMMAND_PRIORITY_EDITOR = 0;
-export const COMMAND_PRIORITY_LOW = 100;
-export const COMMAND_PRIORITY_NORMAL = 200;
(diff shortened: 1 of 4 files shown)


packages/lexical/src/LexicalEditor.ts
```diff
@@ -743,13 +789,18 @@ export class LexicalEditor {
       );
     }
 
-    const listeners = listenersInPriorityOrder[priority];
+    if (!listenersInPriorityOrder.has(priority)) {
```
zurfyx, line 792: Given that the focus is on performance on use, rather than registration, should we build a 20-sized array we can access faster?

packages/lexical/src/LexicalEditor.ts
```diff
@@ -212,13 +212,56 @@ export type CommandListener<P> = (payload: P, editor: LexicalEditor) => boolean;
...
 export type EditableListener = (editable: boolean) => void;
 
-export type CommandListenerPriority = 0 | 1 | 2 | 3 | 4;
-
-export const COMMAND_PRIORITY_EDITOR = 0;
-export const COMMAND_PRIORITY_LOW = 1;
-export const COMMAND_PRIORITY_NORMAL = 2;
-export const COMMAND_PRIORITY_HIGH = 3;
-export const COMMAND_PRIORITY_CRITICAL = 4;
+export type CommandListenerPriority =
```
zurfyx, line 215: I believe we have to ship this as a breaking change given that the numbers are not necessarily opaque

acywatson:
We'll revisit this later

Issue #3784, opened 2023-01-26
Bug: Triple-click selections have incorrect format


Awjin:
https://github.com/facebook/lexical/issues/2648 has related discussion around how triple-click selection should behave, but this is a separate bug.

Lexical version: 0.7.6

## Steps To Reproduce

1. Create two paragraphs.
2. Triple-click to select the first paragraph.
3. Set the format to "bold".

## Current behavior

```ts
selection.hasFormat("bold") === false
```

## Expected behavior

```ts
selection.hasFormat("bold") === true
```

https://user-images.githubusercontent.com/3211874/214970056-cf024ecb-ec0f-4f76-b9bc-14c300b7ecce.mov

Awjin:
Lexical includes the trailing newline when you triple-click. This is correct behavior selection behavior.

The issue is, there are 2 differing user flows after triple-clicking:

1. Text insertion expects the newline to be included (e.g. deleting a triple-selected line deletes the trailing newline).

2. Text formatting does not expect the newline to be included.

trueadm:
I'm unsure if there's one discrete fix here that wouldn't seem like a hack. This is actually how the browser intends it to work AFAIK.

Awjin:
> I'm unsure if there's one discrete fix here that wouldn't seem like a hack. This is actually how the browser intends it to work AFAIK.

Are you referring to the selection behavior? If so, agreed; the selection behavior is correct.

But, determining whether that selection has a format shouldn't be influenced by the trailing newline.

madhusudhand:
There is another case where, double click to select a work should highlight right format.
It works good if the selection made doesn't include first letter.

In the following scenario, `selection.hasFormat` returned `true` when second work is selected (double click to select), and the same returned `false` when the first word is selected

![lex-formatting](https://user-images.githubusercontent.com/1935113/230641990-14380f59-9100-4b12-8560-614523ebec4e.gif)

PR #4512, opened 2023-05-17
Adjust selection after triple click

fantactuka:
Triple click expands selection to cover whole element, but also moves focus node to the next element sibling. So even though visually it looks like single text block is selected, selection is also expanded to the next line. If we try to modify such selection two blocks are affected. 

Easiest way to reproduce: have two paragraphs, tripple click on first, convert into heading, then both paragraphs will be converted into headings. Another example is tripple click within table cell, and pressing delete: it selects current cell and beginning of the next one. Hitting delete will completely delete one of the cells

Before:

https://github.com/facebook/lexical/assets/132642/1bf6284c-4b72-4e5e-9d62-a8368e6e732c

After:

https://github.com/facebook/lexical/assets/132642/0c9fc3ad-911c-4bd9-a3d6-f2d0ce830699

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -440,4 +440,39 @@ test.describe('Selection', () => {
       `,
     );
   });
+
+  test('Can adjust tripple click selection', async ({
+    page,
+    isPlainText,
+    isCollab,
+  }) => {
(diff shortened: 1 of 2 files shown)


zurfyx (approved):
Uh, nice! This one has been there forever. Doesn't close any issue?

fantactuka:
Fixes #2648 and #3784

LuciNyan:
Incredible!

PR #4513, opened 2023-05-17
Fix GC memory leak (2)

zurfyx:
The hardcoded condition on the test (`const node = $getNodeByKey('1');`) was hiding that the fix in #4510 was actually not good enough.

Subtrees fail. For example, if a subtree is deleted and we visit the parent later, their children might not have deleted (previously deleted)

This PR stores the dirtyElements keys to delete and does the deletion later, without interfering with the loop itself.

### DIFF
--- a/packages/lexical/src/LexicalGC.ts
+++ b/packages/lexical/src/LexicalGC.ts
@@ -42,6 +42,7 @@ function $garbageCollectDetachedDeepChildNodes(
   parentKey: NodeKey,
   prevNodeMap: NodeMap,
   nodeMap: NodeMap,
+  nodeMapDelete: Array<NodeKey>,
   dirtyNodes: Map<NodeKey, IntentionallyMarkedAsDirtyElement>,
 ): void {
   let child = node.getFirstChild();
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalGC.ts
```diff
@@ -99,10 +105,14 @@ export function $garbageCollectDetachedNodes(
...
         }
-        nodeMap.delete(nodeKey);
+        nodeMapDelete.push(nodeKey);
       }
     }
   }
+  const nodeMapDeleteLength = nodeMapDelete.length;
+  for (let i = 0; i < nodeMapDeleteLength; i++) {
+    nodeMap.delete(nodeMapDelete[i]);
+  }
```
fantactuka, line 115: IIRC, `for .. of` is cooler 😎

fantactuka (approved):

packages/lexical/src/LexicalGC.ts
zurfyx: @fantactuka - Is it also faster for arrays? afaik we started using for..of as a replacement for `Array.from(<iterable = Set|Map)`

Issue #4514, opened 2023-05-18
Bug: Got error `Cannot read property of null` when using `LexicalClickableLinkPlugin` and clicking on a non-link element


vimutti77:
Got error 'Cannot read property of null' when using 'LexicalClickableLinkPlugin' and clicking on a non-link element

Lexical version: 0.10.0

## Steps To Reproduce

1. Go to playground https://playground.lexical.dev
2. Click on read-only mode.
3. Click on any element in the editor that is not a link.
4. See the error in console.

## The current behavior
```
TypeError: Cannot read properties of null (reading 'href')
    at LexicalClickableLinkPlugin.dev.js:68:21
    at beginUpdate (Lexical.dev.js:8013:5)
    at updateEditor (Lexical.dev.js:8104:5)
    at LexicalEditor.update (Lexical.dev.js:10810:5)
    at HTMLDivElement.onClick (LexicalClickableLinkPlugin.dev.js:57:21)
```

## The expected behavior
No error in console

PR #4519, opened 2023-05-19
Fix clickable link plugin #4514

fantactuka:
Add missing null check, adjust dom traversal util


zurfyx (approved):
Is https://github.com/facebook/lexical/pull/4447 relevant after this one?

thegreatercurve (approved):

Issue #4518, opened 2023-05-19
Bug: TableRowNode and TableCellNode's `importDom` methods don't take into account the element's height and width respectively


GMchris:
Every transformation (exportDOM, importJSON, exportJSON, createDOM) take into account a TableRowNode or TableCellNode's height or width, however trying to parse HTML (even if it was outputted by Lexical itself) doesn't set the height/width of the node. This seems like a mistake, however it's specifically only these two things that don't get imported.

Lexical version: 0.10.0

## Steps To Reproduce

1. Create a lexical editor with the `TablePlugin` and the `Table`, `TableRow` and `TableCell` nodes
2. Implement some functionality for resizing `TableCell`s or `TableRow`s (I more or less replicated the logic provided in the playground.
3. Using `$generateNodesFromDOM` on some DOM that features a table, with rows that have `style="height: xx"` or cells that have `style="width: xx"`

## The current behavior

The width/height in the DOM is ignored. Cell/Row nodes are added without these properties set

## The expected behavior

These properties should get set, especially because they're already part of their corresponding classes.

## Proposed solution

The constructors of these nodes already support receiving this sizing property, let's look at some code within the cell node. This code is ran to turn a dom node into a lexical table cell node.

```typescript
export function convertTableCellNodeElement(
  domNode: Node,
): DOMConversionOutput {
  const domNode_ = domNode as HTMLTableCellElement;
  const nodeName = domNode.nodeName.toLowerCase();

  const tableCellNode = $createTableCellNode(
    nodeName === 'th'
      ? TableCellHeaderStates.ROW
      : TableCellHeaderStates.NO_STATUS,
  );
  tableCellNode.__colSpan = domNode_.colSpan;
  tableCellNode.__rowSpan = domNode_.rowSpan;
  const backgroundColor = domNode_.style.backgroundColor;
  if (backgroundColor !== '') {
    tableCellNode.__backgroundColor = backgroundColor;
  }

  return {
    forChild: (lexicalNode, parentLexicalNode) => {
      if ($isTableCellNode(parentLexicalNode) && !$isElementNode(lexicalNode)) {
        const paragraphNode = $createParagraphNode();
        if (
          $isLineBreakNode(lexicalNode) &&
          lexicalNode.getTextContent() === '\n'
        ) {
          return null;
        }
        paragraphNode.append(lexicalNode);
        return paragraphNode;
      }

      return lexicalNode;
    },
    node: tableCellNode,
  };
}
```

[...]

PR #4521, opened 2023-05-19
Allow importing width and height from table cell and DOM nodes

GMchris:
Addresses #4518

### DIFF
--- a/packages/lexical-table/src/LexicalTableCellNode.ts
+++ b/packages/lexical-table/src/LexicalTableCellNode.ts
@@ -27,6 +27,8 @@ import {
   DEPRECATED_GridCellNode,
 } from 'lexical';
 
+import {PIXEL_VALUE_REG_EXP} from './constants';
+
 export const TableCellHeaderStates = {
   BOTH: 3,
   COLUMN: 2,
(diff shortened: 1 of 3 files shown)


acywatson:
Historically, the problem with doing things like this is that it can cause something like table cell height to be imported into an editor that doesn't otherwise support row resizing. This would mean the row is sort of "stuck" with the height and the user wouldn't be able to change it in the destination editor. Typically we'd ask users to implement this in a TableCellNode override, but I will defer to @zurfyx if he think this is functionality that should be baked into core tables logic.

acywatson:
> > 
> 
> I understand this concern, but as far as I can tell this data is also passed with the serialized node, so a copy and paste of a resized column or row from an editor that allows resizing into one that doesn't presents the same "problem" - the cell/row is now "stuck" at this size. In fact `importDOM` is the only place in these nodes that doesn't attempt to transfer the width/height information.
> 
> Just as a counter-point, if resizing should be a custom operation and editor-builders are expected to implement it themselves fully, why do the base Row and Cell Nodes even contain width and height properties.

Yes, valid point. Let's do this then.

PR #4527, opened 2023-05-22
Preserve indentation when converting to listItems

GermanJablo:
This PR causes indents to be preserved when the selection is converted to list-items.

https://www.loom.com/share/9ce5fca77eb94819b20d5ff10128d48e

Also, to avoid strange displacement, the getIndent property of listItem is modified so that the initial value is 1, since a listItem without nesting matches paragraphs or headings with indent 1.

This follows the reasoning of https://github.com/facebook/lexical/pull/4025


packages/lexical-rich-text/src/index.ts
```diff
@@ -838,13 +838,13 @@ export function registerRichText(editor: LexicalEditor): () => void {
...
-          }
+          // we trigger DELETE_CHARACTER_COMMAND if we can't outdent the element
+          const indent = element.getIndent();
+          const outdentCommand = editor.dispatchCommand(
+            OUTDENT_CONTENT_COMMAND,
+            undefined,
+          );
+          if (indent !== element.getIndent()) return outdentCommand;
         }
         return editor.dispatchCommand(DELETE_CHARACTER_COMMAND, true);
```
GermanJablo: I have modified this part again.
Delete is triggered in case the element cannot be outdented.
This makes the code more robust and prevents unwanted behavior without having to depend on a particular implementation.

packages/lexical-rich-text/src/index.ts
```diff
@@ -849,9 +849,13 @@ export function registerRichText(editor: LexicalEditor): () => void {
...
-          if (element.getIndent() > 0) {
-            return editor.dispatchCommand(OUTDENT_CONTENT_COMMAND, undefined);
-          }
+          // we trigger DELETE_CHARACTER_COMMAND if we can't outdent the element
+          const indent = element.getIndent();
+          const outdentCommand = editor.dispatchCommand(
+            OUTDENT_CONTENT_COMMAND,
+            undefined,
+          );
+          if (indent !== element.getIndent()) return outdentCommand;
```
acywatson, line 858: nit: we typically avoid these one-liner conditionals in favor of curly braces. I don't really mind it, but IIRC this would be sort of inconsistent.

acywatson (commented):
My one concern is moving from zero to 1-based indentation, but I think this is safe given we don't serialize the value of the indent property and instead rely on the semantic structure of the list to control indentation level.

packages/lexical-playground/__tests__/e2e/List.spec.mjs
```diff
@@ -1149,7 +1149,9 @@ test.describe('Nested List', () => {
     await assertHTML(
       page,
       html`
-        <p dir="ltr"><span data-lexical-text="true">a</span></p>
+        <p dir="ltr" style="padding-inline-start: calc(40px)">
```
acywatson, line 1152: So does this mean that now converting a single-level list to Paragraphs will make the resulting paragraphs indented one level?

GermanJablo: Yes I thought so too. But think of it this way:

If you wanted to do the reverse of what I did in the loom video above (convert `2, 3, 4, 5, 6` to paragraphs again), what would you expect to happen? I suppose to return to the original state.

Converting paragraphs to lists and lists to paragraphs would now be reverse operations. The only exception to this would be paragraphs with indent 0, which become items with indent 1.

That could be avoided by making the first level list-items and first level paragraphs aligned. That's the behavior OneNote chose. However, it looks weird and doesn't follow standard HTML (it would require CSS which I can see many users having difficulty with).

Having looked at many editors, this seemed to me to be the most balanced behavior. But if you have another suggestion I'd love to hear it.

GermanJablo:
I personally don't need this anymore and it's outdated. If there is demand I can work on this again.

PR #4540, opened 2023-05-23
Add lexical menu command priority prop

kbkmn:
keyboardPriority prop added to LexicalMenu and LexicalTypeaheadMenuPlugin so you can, for example, use KEY_ENTER_COMMAND overriding creation of new line, but not when selecting item in menu

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -18,6 +18,8 @@ import {
   $getSelection,
   $isRangeSelection,
   $isTextNode,
+  COMMAND_PRIORITY_LOW,
+  CommandListenerPriority,
   createCommand,
   LexicalCommand,
   LexicalEditor,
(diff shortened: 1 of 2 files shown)


acywatson:
I actually ran into this problem recently and this is one of the solutions discussed. The problem is it doesn't scale very well as a general pattern, as you could potentially need to add the same prop for all the other commands registered in this plugin (or other plugins).

What I ended up doing instead is registering a new command like "LEXICAL_TYPEAHEAD_OPEN" with a boolean value. Then dispatch it with true/false in the onOpen/onClose handlers. Then you can keep track of that state in a variable and condition your other handlers on that value.

We can do this without changing the existing API or introducing a new pattern. Would that work for your use case?

kbkmn:
> I actually ran into this problem recently and this is one of the solutions discussed. The problem is it doesn't scale very well as a general pattern, as you could potentially need to add the same prop for all the other commands registered in this plugin (or other plugins).
> 
> What I ended up doing instead is registering a new command like "LEXICAL_TYPEAHEAD_OPEN" with a boolean value. Then dispatch it with true/false in the onOpen/onClose handlers. Then you can keep track of that state in a variable and condition your other handlers on that value.
> 
> We can do this without changing the existing API or introducing a new pattern. Would that work for your use case?

Do you mean i need to register command in every plugin thay uses that specific key and dispatch it from my LexicalMenu wrapper or am i missing something from docs?

thegreatercurve:
I think Acy's approach is the simplest and avoids adding in new props to the module.

PR #4544, opened 2023-05-24
Arrow up selection fix #4536

### LINKED ISSUE #4536: Bug: Using `shift+up` for selection breaks on going from image to list item node
If you add an image to the bottom of the Lexical playground text and use `shift+up` keys to select the image, continuing the selection will break once it hits the list item subnode. It works fine in the opposite direction. See gif:

First works in the downward direction, then fails in the upward.
CleanShot 2023-05-23 at 15 15 15

Lexical version: `0.10.0`

## Steps To Reproduce

1. Create list with listitems
2. Add an image below
3. Put the cursor after the image and use `shift+up` to select text up to the listitems

Link to code example: use the playground

## The current behavior
Selection breaks when moving up, but not down

## The expected behavior
Selection should continue

fantactuka:
That condition looks a bit odd and is likely redundant or missing extra logic. Basically any shift + arrow-up selection will be broken if going into { canBeEmpty: false } (tables, lists). Original issue uses image as an example, but even list + paragraph case is broken:
1. Add list + paragraph, press shift + up in attempt to select
2. Note that cursor just moves to the list, without selecting anything

Before: 

https://github.com/facebook/lexical/assets/132642/80813c6d-8373-407d-a7c3-dda0aaa3802d

After:

https://github.com/facebook/lexical/assets/132642/94b8cd75-1358-4b66-8afa-88b77988a512

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/ListsHTMLCopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/ListsHTMLCopyAndPaste.spec.mjs
@@ -6,7 +6,6 @@
  *
  */
 
-import {moveLeft} from '../../../keyboardShortcuts/index.mjs';
 import {
   assertHTML,
   assertSelection,
@@ -375,7 +374,6 @@ test.describe('HTML Lists CopyAndPaste', () => {
     await page.keyboard.press('Enter');
     await page.keyboard.press('Enter');
     await page.keyboard.press('ArrowUp');
-    await moveLeft(page, 4);
     await page.keyboard.press('ArrowUp');
     await page.keyboard.press('ArrowUp');
     await pasteFromClipboard(page, {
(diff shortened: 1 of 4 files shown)


packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/ListsHTMLCopyAndPaste.spec.mjs
```diff
@@ -375,7 +374,6 @@ test.describe('HTML Lists CopyAndPaste', () => {
     await page.keyboard.press('Enter');
     await page.keyboard.press('Enter');
     await page.keyboard.press('ArrowUp');
-    await moveLeft(page, 4);
```
fantactuka, line 378: This test was wrong as moving upward was broken

packages/lexical-playground/src/ui/ContentEditable.css
```diff
@@ -13,6 +13,7 @@
   position: relative;
   outline: 0;
   padding: 8px 28px 40px;
+  min-height: 150px;
```
fantactuka, line 16: Minor adjustments for mobile screen on a playground, not really a separate PR-worthy

zurfyx (approved):

Issue #4547, opened 2023-05-25
Bug: After use 'Tab' botton,  and then typing Chinese character, the whole editor will be broken and can't do anything


gaozhijie94:
Lexical version: v0.11.0 play ground

## Steps To Reproduce

1. typing any character.
2. then use the 'Tab'
3. after the space '    ' , typing any Chinese character.(not copy paste)
4. and the type 'Enter' to add a new line
5. then the editor can't do anything, it just broken.

## The current behavior
<img width="1141" alt="截屏2023-05-25 14 04 53" src="https://github.com/facebook/lexical/assets/76510265/cae40079-12f0-42cf-b05d-3a85a98d90b4">

acywatson:
@zurfyx did tab changes break composition?

PR #4553, opened 2023-05-25
Mutations to respect canInsertBeforeAfter

zurfyx:
We're usually very lenient when it comes to composition (as it can break easily) and we don't touch it unless it's imprescindible.

However, Mutations as they're now are breaking the `canInsert{Before/After}` rules which directly impacts things like Mentions or TabNode. This PR forbids Mutations that violate this rule and will instead fallback to `input`.

The other question is whether TabNode has to implement `canInsert{Before/After}` or we can do something smarter with it.

This PR also adjusts the behavior for MentionNode, which also makes me think whether this `canInsert` rule should be automatically applied when there's segmentation as I can't think why it would make sense to allow characters at the end when segmentation implies that the Node will be recreated as soon as you type on it.

Fixes https://github.com/facebook/lexical/issues/4547

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -635,6 +635,7 @@ export function $updateTextNodeFromDOMContent(
       }
       const parent = node.getParent();
       const prevSelection = $getPreviousSelection();
+      const prevTextContentSize = node.getTextContentSize();
       const compositionKey = $getCompositionKey();
       const nodeKey = node.getKey();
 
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalUtils.ts
```diff
@@ -635,6 +635,7 @@ export function $updateTextNodeFromDOMContent(
       }
       const parent = node.getParent();
       const prevSelection = $getPreviousSelection();
+      const prevTextContentSize = node.getTextContentSize();
```
fantactuka, line 638: This could've been called from the condition given it's only used once and might not even needed to be executed

PR #4555, opened 2023-05-25
Fire selection change on decorator delete

### LINKED ISSUE #4402: Bug: Deleting a DecoratorNode does not trigger a SelectionChange event, unless it has isIsolated() set to true
Lexical version: 0.9.2

## Steps To Reproduce

1. Create a DecoratorNode
2. Delete it
3. Observe that no SelectionChange event was triggered

Link to code example: https://codesandbox.io/s/vibrant-rgb-gf8rc9?file=/src/nodes/node.tsx
- In the above Sandbox, type in anything and hit ENTER to create a DecoratorNode
- Then delete it and observe no console.log() happened for that deletion event (there is a SelectionChangePlugin listening and printing)
- The bug fixes if you set the node to isIsolated(): true. This is commented out in node.tsx in the below example.

## The current behavior

https://user-images.githubusercontent.com/116599321/234130020-eb0f0af6-dfc5-4f30-9a75-d7b746b9f564.mov

## The expected behavior

You can get the expected behavior by setting `isIsolated(): true` on the decorator node. I've put that in the Code sandbox, commented out, in node.tsx

https://user-images.githubusercontent.com/116599321/234130090-4f2a3f2c-8488-4cda-8821-a4d6aed24155.mov

acywatson:
Fixes #4402 

For some reason, in this case,  the native selectionchange event doesn't fire. The best I can tell, it's because DOM Selection already matches what we try to set it to in the reconciler when we call setBaseAndExtent. I don't see where we're explicitly changing the native selection in this code path, so I guess it's the browser. In any case, this explicitly dispatches selection change to cover this case.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -32,6 +32,7 @@ import {
   DEPRECATED_GridCellNode,
   DEPRECATED_GridNode,
   DEPRECATED_GridRowNode,
+  SELECTION_CHANGE_COMMAND,
   TextNode,
 } from '.';
 import {DOM_ELEMENT_TYPE, TEXT_TYPE_TO_FORMAT} from './LexicalConstants';
@@ -2080,6 +2081,8 @@ export class RangeSelection implements BaseSelection {
           $setSelection(nodeSelection);
         } else {
           possibleNode.remove();
+          const editor = getActiveEditor();
+          editor.dispatchCommand(SELECTION_CHANGE_COMMAND, undefined);
         }
         return;
       } else if (


packages/lexical/src/LexicalSelection.ts
```diff
@@ -2080,6 +2081,8 @@ export class RangeSelection implements BaseSelection {
           $setSelection(nodeSelection);
         } else {
           possibleNode.remove();
+          const editor = getActiveEditor();
+          editor.dispatchCommand(SELECTION_CHANGE_COMMAND, undefined);
```
fantactuka, line 2085: Can we mark selection as dirty instead?

acywatson: I don’t think that will fix the problem. The reconciler is trying to reconcile selection, but when we call setBaseAndExtent DOM API, it doesn’t seem to trigger a native selectionchange, which is what we depend on to fire our own command in most cases.

fantactuka: We can have it as a follow up discussion, but my thoughts around SELECTION_CHANGE_COMMAND:

I wonder if we should ever rely on DOM's `selectionchange` event, and instead use lexical selection diff for 100% of cases: when Lexical selection is reconciled (either after sync'ing with DOM or via API0 we can diff it with the previous one, and if not the same - dispatch the command. Similar to

```js
editor.registerUpdateListener(({editorState, prevEditorState}) => {
    const prevSelection = prevEditorState.read($getSelection);
    const nextSelection = editorState.read($getSelection);

    const hasChanged = prevSelection === null 
            ? nextSelection !== null 
            : nextSelection === null || !prevSelection.is(nextSelection)

    ...
});
```

fantactuka: @zurfyx thoughts?

acywatson: That actually might be a good idea - we also have some weird patches in other modules around SELECTION_CHANGE. I'm not sure if it would cover all the cases or what load-bearing bugs it might break. Conceptually it makes sense to me that a Lexical SELECTION_CHANGE event should fire only when Lexical selection changes. Possible edge cases around focus/blur?

zurfyx: Heads up, haven't deeply thought about this. What is our stance on normalizing browser behavior? Do other browsers trigger selection change? The SELECTION_CHANGE command is a browser event so unless other browsers trigger it I'm unsure we should as well. However, to fulfill the case of notifications we can also introduce a separate command that does something very similar to this one but is not tied to selection change. WDYT?

acywatson: > Haven't deeply thought about this, but what is our stance on normalizing browser behavior? Do other browsers trigger selection change? The SELECTION_CHANGE command is a browser event so unless other browsers trigger it I'm unsure we should as well. However, to fulfill the case of notifications (is this all we need?) we can also introduce a separate command that does something very similar to this one but is not tied to selection change. WDYT?

I mean, from a semantic standpoint I just don't understand how it would make any sense to backspace delete a node and not have a SELECTION_CHANGE fire. Like, obviously the selection has changed in that scenario. Not having the corresponding event fire is always going to raise questions and create bugs.

I feel that one of the features we've touted repeatedly is that Lexical *does* normalize browser behavior (particularly with respect to selection) and smooth out the inconsistencies. I don't see a compelling reason to change that now.

PR #4556, opened 2023-05-26
Better Error Message for exportJSON type

howlowck:
When creating a custom Node that extends an existing Node, and `exportJSON()` is implemented but uses the output of `super.exportJSON()`, the current error message `LexicalNode: Node %s does not implement .exportJSON().` is misleading.


fantactuka:
Could you run `npm run extract-codes` and commit `codes.json` file as well?

fantactuka (approved):

howlowck:
> Could you run `npm run extract-codes` and commit `codes.json` file as well?

Thank you! Updated (not sure why it added more messages in `codes.json` than just my change.)

Issue #4565, opened 2023-05-27
Feature: LexicalClickableLinkPlugin should support ctrl click mode


fuchen:
In editing mode, when clicking at a link text, I expect move the cursor position rather than open the link. Usually `Ctrl+click` is expected to open the link.

arnabrahman:
I want to take a look @ivailop7

ivailop7:
Go for it!

arnabrahman:
@ivailop7 i have submitted a PR

arnabrahman:
I got the issue wrong. @fuchen is it possible to provide a screenshot of the issue? When in the editing mode, if I click on a link it doesn't open anything. The link only opens in a new tab from the editor during the read mode. I am not sure if I am doing anything wrong or not.

arnabrahman:
@ivailop7 If you have any insight it will help too

ivailop7:
@arnabrahman I agree the `Ctrl + Click` on a link inside the editor should open the link in a new tab, it's an industry-wide behaviour that makes sense to me. It should be implemented in the Playground. The current behaviour on the mouse click inside a link is to move the cursor to that position and not open the link, which is what's described at the original comment, so I don't see the need for any further changes there.

arnabrahman:
[Lexical Playground.webm](https://github.com/facebook/lexical/assets/12469224/a6517fa5-0edf-45c1-a2e6-b21170c7ea63)

Sorry to ping you again, but I just need to confirm the requirement of this issue. So here when I click on the link with `ctrl+click` it should open in a new tab? Attached is the video @ivailop7

ivailop7:
> Lexical.Playground.webm 
> Sorry to ping you again, but I just need to confirm the requirement of this issue. So here when I click on the link with `ctrl+click` it should open in a new tab? Attached is the video @ivailop7

Yep, that's all.

utkarsh-shrivastav77:
is this issue open

sathishsridhar:
is this issue still open, I was able to find a fix for this, can I open a PR ?

davlet61:
> is this issue still open, I was able to find a fix for this, can I open a PR ?

Hey, what's your fix? I am planning to implement the same in my app, maybe I can get some inspiration from you :)

sathishsridhar:
Kindly check the above PR for the fix @davlet61

tarngerine:
fwiw just using the plugin provided in `@lexical/react` worked for me: 

[...]

PR #4567, opened 2023-05-28
feat: open a link in a new tab with ctrl+click/command+click

arnabrahman:
## Description of your changes
The issue description is not much. What I have learned from that is, right now when we click on a link in editing mode it opens in a new tab. 

Changes have been made so one can only open a link in a new tab with `ctrl+click`. And for Mac users, it will be `command+click`. I have written an e2e test to check this. 
### How to verify this change
An e2e test for this new feature has been written to verify the changes.

### Related issue
 #4565

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
@@ -20,6 +20,7 @@ import {
   assertHTML,
   assertSelection,
   click,
+  expect,
   focus,
   focusEditor,
   html,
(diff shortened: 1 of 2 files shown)


arnabrahman:
Let me know if I have guessed it right or if something needs to be changed

fantactuka:
Isn't the original issue about the link within editor, not inside floating menu?

arnabrahman:
@fantactuka Sorry for understanding the issue incorrectly. I will confirm with the op & get back.

fantactuka:
Closing based on https://github.com/facebook/lexical/issues/4565#issuecomment-1567097377

PR #4580, opened 2023-05-30
Fix a11y for ContentEditable

maximshipko:
This PR fix accessibility of ContentEditable in the non editable mode.

When Lexical editor is not editable, there is no `role` attribute, but `aria-autocomplete` has the value 'none'. That breaks our pipeline during a11y check with the critical warning:
image

As I understand the issue, we can not have aria attributes without a valid role. So I keep the role 'textbox' for non-editable case, but at the same time set `aria-readonly="true"` as suggested in [doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role)

### DIFF
--- a/packages/lexical-react/src/LexicalContentEditable.tsx
+++ b/packages/lexical-react/src/LexicalContentEditable.tsx
@@ -82,6 +82,7 @@ export function ContentEditable({
       aria-labelledby={ariaLabelledBy}
       aria-multiline={ariaMultiline}
       aria-owns={!isEditable ? undefined : ariaOwns}
+      aria-readonly={!isEditable ? true : undefined}
       aria-required={ariaRequired}
       autoCapitalize={autoCapitalize}
       className={className}
@@ -89,7 +90,7 @@ export function ContentEditable({
       data-testid={testid}
       id={id}
       ref={ref}
-      role={!isEditable ? undefined : role}
+      role={role}
       spellCheck={spellCheck}
       style={style}
       tabIndex={tabIndex}


packages/lexical-react/src/LexicalContentEditable.tsx
```diff
@@ -82,14 +82,15 @@ export function ContentEditable({
...
+      {...(!isEditable ? {'aria-readonly': true} : null)}
       aria-required={ariaRequired}
       autoCapitalize={autoCapitalize}
       className={className}
       contentEditable={isEditable}
       data-testid={testid}
       id={id}
       ref={ref}
-      role={!isEditable ? undefined : role}
+      role={role}
```
zurfyx, line 93: I'd be inclined to say that a non-editable textbox does not require to be marked as a textbox. Lexical can be used in a transparent manner to provide the reading the experience. However, given the other `aria` and the fact that this is still customizable makes me think that this change is reasonable with what we have now

packages/lexical-react/src/LexicalContentEditable.tsx
```diff
@@ -82,14 +82,15 @@ export function ContentEditable({
       aria-labelledby={ariaLabelledBy}
       aria-multiline={ariaMultiline}
       aria-owns={!isEditable ? undefined : ariaOwns}
+      {...(!isEditable ? {'aria-readonly': true} : null)}
```
zurfyx, line 85: Nit, I believe the use of undefined should do the same
```suggestion
      aria-readonly={!isEditable ? true : undefined}
```

zurfyx (approved):

moy2010:
> As I understand the issue, we can not have aria attributes without a valid role. So I keep the role 'textbox' for non-editable case, but at the same time set `aria-readonly="true"` as suggested in [doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role)

FYI, the documentation actually suggests to use a `readonly` attribute:

```
[aria-readonly](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-readonly) attribute

Indicates that the user cannot modify the value of the text field. Instead of using aria-readonly, use the semantic <input type="text"> or <textarea> with a readonly attribute.
```

packages/lexical-react/src/LexicalContentEditable.tsx
maximshipko: Thanks for the review @zurfyx

packages/lexical-react/src/LexicalContentEditable.tsx
maximshipko: Good point! That is more readable 👍

maximshipko:
@moy2010,
`readonly` is for form elements (input, select, textarea etc), but Lexical is using editable DIV, that's why `aira-readonly`

PR #4581, opened 2023-05-30
Fix theme mutation with separate cache

acywatson:
The "cache" for getCachedClassNameArray is the theme itself. We mutate the theme object provided in the initial config, which is problematic because it can cause runtime type errors when we're looking for a string when accessing the theme properties directly, but instead we get an array. It looks like this is possible because we have an explicit *any* in the EditorTheme class types in TS, which we also probably need to fix.

### DIFF
--- a/packages/lexical-code/src/CodeHighlightNode.ts
+++ b/packages/lexical-code/src/CodeHighlightNode.ts
@@ -9,6 +9,7 @@
 import type {
   EditorConfig,
   EditorThemeClasses,
+  LexicalEditor,
   LexicalNode,
   LineBreakNode,
   NodeKey,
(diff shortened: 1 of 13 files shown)


fantactuka (commented):
This should be on editor instance, not a global variable

acywatson:
> This should be on editor instance, not a global variable (otherwise risking conflicts with multiple instances and different themes)

Yea I just realized that...wondering if that's causing UT failures. Some sort of pollution across test cases maybe

acywatson:
I'm wondering if we even need this caching at all, tbh.

acywatson:
> I'm wondering if we even need this caching at all, tbh.

@fantactuka we can try this instead: https://github.com/facebook/lexical/pull/4582

PR #4582, opened 2023-05-30
Another approach to fixing the className cache

acywatson:
The "cache" for getCachedClassNameArray is the theme itself. We mutate the theme object provided in the initial config, which is problematic because it can cause runtime type errors when we're looking for a string when accessing the theme properties directly, but instead we get an array.

This approach (vs #4581 ) has the advantage of leveraging the theme to keep the cache scoped to the Editor, disadvantage of still mutating the theme object, although collision probability seems low.

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -1014,6 +1014,14 @@ export function getCachedClassNameArray(
   classNamesTheme: EditorThemeClasses,
   classNameThemeType: string,
 ): Array<string> {
+  if (classNamesTheme.__lexicalClassNameCache === undefined) {
+    classNamesTheme.__lexicalClassNameCache = {};
+  }
+  const classNamesCache = classNamesTheme.__lexicalClassNameCache;
+  const cachedClassNames = classNamesCache[classNameThemeType];
+  if (cachedClassNames !== undefined) {
+    return cachedClassNames;
+  }
   const classNames = classNamesTheme[classNameThemeType];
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalUtils.ts
```diff
@@ -1014,6 +1014,14 @@ export function getCachedClassNameArray(
   classNamesTheme: EditorThemeClasses,
   classNameThemeType: string,
 ): Array<string> {
+  if (classNamesTheme.__lexicalClassNameCache === undefined) {
+    classNamesTheme.__lexicalClassNameCache = {};
```
zurfyx, line 1018: Nit, why an object vs a Map?

zurfyx (approved):
LGTM, I wonder to what extent we want to promote immutability for user's input such as a theme (but I know this was already there)

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1022,7 +1030,7 @@ export function getCachedClassNameArray(
   // applied to classList.add()/remove().
   if (typeof classNames === 'string') {
```
zurfyx, line 1031: How does cache work for the nested ones? Or we don't really support this?

acywatson: That’s a good point. I’m gonna try to just get rid of this caching. It’s only here to make dealing with DOMTokenList APIs easier and still performant. It might have outlived it’s usefulness.

PR #4588, opened 2023-06-01
mouseenter event typo fix

agentraghav:
# Descripton 
updated mouseenter event typo in injectedScript.js file


zurfyx (commented):
Makes sense, thank you. That said, this is a package we forked from playwright, I wonder whether it'd make more sense to fix it there and then update our copy. @acywatson wdyt?

agentraghav:
> Is this fixing a problem?

i have to check but definitely this might be causing bug somewhere

acywatson:
> Makes sense, thank you. That said, this is a package we forked from playwright, I wonder whether it'd make more sense to fix it there and then update our copy. @acywatson wdyt?

We do need to update our playwright anyway, but this means updating @trueadm's fork from the upstream, which may or may not already contain this fix.

thegreatercurve (approved):

PR #4592, opened 2023-06-02
Fix uuid bug in `AutocompleteNode.clone()`

### LINKED ISSUE #4589: Bug: Appears to be typo/bug in Playground autocomplete plugin
[On this line](https://github.com/facebook/lexical/blame/73ac48864cdd1021502a39421022ccafc2d98648/packages/lexical-playground/src/nodes/AutocompleteNode.tsx#LL41C15-L41C15), it seems that 
```tsx
export class AutocompleteNode extends DecoratorNode<JSX.Element | null> {
  // TODO add comment
  __uuid: string;

  static clone(node: AutocompleteNode): AutocompleteNode {
    return new AutocompleteNode(node.__key);
  }

  constructor(uuid: string, key?: NodeKey) {
      super(key);
      this.__uuid = uuid;
  }
```
should be changed to
```diff
export class AutocompleteNode extends DecoratorNode<JSX.Element | null> {
  // TODO add comment
  __uuid: string;

  static clone(node: AutocompleteNode): AutocompleteNode {
-    return new AutocompleteNode(node.__key);
+    return new AutocompleteNode(node.__uuid);
  }
  constructor(uuid: string, key?: NodeKey) {
      super(key);
      this.__uuid = uuid;
  }
```

Lexical version: 11.1

## The current behavior

Currently upon cloning, the uuid will change (see that the node key is used instead).

The UUID is meant to prevent multiple autocomplete options in the same view, but if the uuid can change, this functionality doesn't work. I noticed this when creating a modified autocomplete plugin based off this one. After changing selection, there could be several nodes working independently trying to update the editor concurrently.

## The expected behavior

The UUID should be stable across all instances of the plugin and all nodes tracked by the plugin. This way the autocomplete node transform can clean up old ones.

## Questions

Should I ping the original author? I may be misunderstanding the intended behavior here

msdrigg:
UUID and key should be stable when the node is cloned, but currently, the old key becomes the new UUID and a new key is regenerated. 

This feels like a simple typo that arose in the v2 rewrite a few months ago.

Fixes  #4589

### DIFF
--- a/packages/lexical-playground/src/nodes/AutocompleteNode.tsx
+++ b/packages/lexical-playground/src/nodes/AutocompleteNode.tsx
@@ -39,7 +39,7 @@ export class AutocompleteNode extends DecoratorNode<JSX.Element | null> {
   __uuid: string;
 
   static clone(node: AutocompleteNode): AutocompleteNode {
-    return new AutocompleteNode(node.__key);
+    return new AutocompleteNode(node.__uuid, node.__key);
   }
 
   static getType(): 'autocomplete' {


zurfyx (approved):
Thank you, LGTM! Merging as soon as the bot replies with the signed CLA

Issue #4548, opened 2023-05-25
Bug: cant delete list next sibling table


yaojiu19:
Lexical version: 0.11.0

## The current behavior
![image](https://github.com/facebook/lexical/assets/18085334/b32f145b-dc23-4820-8b59-c2dd59288aa0)

## The expected behavior
delete list and append paragraph

ivailop7:
This works now. Closing.

PR #4593, opened 2023-06-02
[WIP] Fix internalResolveSelectionPoint resolved offset

zurfyx:
The resolved offset was not correct with ElementNodes. We were propagating the offset when the offset is only valid for the immediate child.

Putting this PR on hold for now because as we fix this other bad behavior around deletion arise. It seems like this bug was preventing other bigger bugs.

Fixes https://github.com/facebook/lexical/issues/4548


ivailop7:
Legitimately stumbled upon this behaviour by coincidence today, and was contemplating if it was working as designed or a bug. Great stuff!

ivailop7:
Hey @zurfyx , is this PR still a thing?

potatowagon:

PR #4597, opened 2023-06-03
Fix MenuTypeahead position

### LINKED ISSUE #4545: Bug: Typeahead menu opens in the wrong position
Lexical version: 0.11.0

## Steps To Reproduce

1. Open Lexical Playground
2. Type some sample text so that a Typeahead menu would overlap the edge of the editor a bit.
3. Type `@a` to open the Typeahead menu - the menu is in the correct position
4. Remove and type `@a` again - the menu is now in the wrong position

## The current behavior
The Typeahead menu is in the wrong position as soon as it is opened more than once:

## The expected behavior
The Typeahead menu should open to the right of the current cursor position, like in v0.10.0.
Or (ideally) the Typeahead menu always ends at the end of the editor and never goes over the edge:
Screenshot 2023-05-24 at 20 03 02

aleksandr-kiliushin:
Closes #4545.

Demo (before and after):

https://github.com/facebook/lexical/assets/65129988/4855d4f6-fc19-4f8b-9332-41af475f68ce

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -498,7 +498,7 @@ export function useMenuAnchorRef(
 
         if (left + menuWidth > rootElementRect.right) {
           containerDiv.style.left = `${
-            left - menuWidth + window.pageXOffset
+            rootElementRect.right - menuWidth + window.pageXOffset
           }px`;
         }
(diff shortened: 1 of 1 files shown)


ivailop7 (approved):

zurfyx (approved):

zurfyx:
Thank you!

PR #4598, opened 2023-06-03
Update nodes.md

navanshu:
Fixed the incorrect typescript in the code examples

### DIFF
--- a/packages/lexical-website/docs/concepts/nodes.md
+++ b/packages/lexical-website/docs/concepts/nodes.md
@@ -183,7 +183,7 @@ export function $createCustomParagraphNode(): ParagraphNode {
   return new CustomParagraph();
 }
 
-export function $isCustomParagraphNode(node?: LexicalNode): boolean {
+export function $isCustomParagraphNode(node: LexicalNode | null | undefined): boolean {
   return node instanceof CustomParagraph;
 }
 ```
@@ -230,7 +230,7 @@ export function $createColoredNode(text: string, color: string): ColoredNode {
   return new ColoredNode(text, color);
 }
 
-export function $isColoredNode(node?: LexicalNode): boolean {
+export function $isColoredNode(node: LexicalNode | null | undefined): boolean {
   return node instanceof ColoredNode;
 }
 ```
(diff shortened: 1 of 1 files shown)


zurfyx (changes requested):
Thank you, that's Flow typing, these examples were never updated. If you want to go ahead with these corrections you probably want to type `null | undefined | LexicalNode` instead

See an example
https://github.com/facebook/Lexical//blob/d05ae76e5ab452ef8ee977f74ed4e890ac7c5db0/packages/lexical/src/nodes/LexicalTextNode.ts#L1125-L1129

navanshu:
> Thank you, that's Flow typing, these examples were never updated. If you want to go ahead with these corrections you probably want to type `null | undefined | LexicalNode` instead
> 
> See an example https://github.com/facebook/Lexical//blob/d05ae76e5ab452ef8ee977f74ed4e890ac7c5db0/packages/lexical/src/nodes/LexicalTextNode.ts#L1125-L1129

Yeah, did think of that, just started getting into lexical.

packages/lexical-website/docs/concepts/nodes.md
```diff
@@ -183,7 +183,7 @@ export function $createCustomParagraphNode(): ParagraphNode {
   return new CustomParagraph();
 }
 
-export function $isCustomParagraphNode(node: ?LexicalNode): boolean {
+export function $isCustomParagraphNode(node: LexicalNode | null | undefined): boolean {
```
zurfyx, line 186: Mind updating the return type too on the 3 of them please?
```suggestion
export function $isCustomParagraphNode(node: LexicalNode | null | undefined): node is CustomParagraph  {
```

zurfyx (approved):

packages/lexical-website/docs/concepts/nodes.md
navanshu: Instance of will always return a boolean, I think the return type is correct, it is upto typescript to the type of Node

zurfyx: > Instance of will always return a boolean, I think the return type is correct, it is upto typescript to assert the type of Node

While `boolean` is correct, TS can't infer that the type is CustomParagraph unless it's specifically marked like it, which can come useful in many ocassions. See predicates examples - https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

navanshu: :D :O
This is interesting, I had no idea about this.

zurfyx:
LGTM, thank you!

PR #4603, opened 2023-06-05
Fix iOS newline insertion

### LINKED ISSUE #4266: Bug: cannot escape lists
On the playground, on iOS, when I add 2 bullets and then hit enter to add a new bullet and then enter to try to escape the list, it doesn't work so it's not possible to escape a list with double enter 

I’m getting a line break, which keeps the bullet, empty line, new line indented under the bullet. New enters just keep this going

Thanks
Andre

### LINKED ISSUE #4335: Bug: new line at start of heading
When pressing enter at the start of a heading, Lexical creates a <br> after the ##, rather than add an empty line before the heading which is the expected behaviour

The "##" are not visible so the enter is pressed with the cursor at the start of the visible line

Current outcome:
```
## 
Heading
```

Expected / desired outcome:
```
<empty line>
## Heading
```

Don't think anyone pressing enter for a new line at the start of a heading wants to create a two line heading but rather wants to create more space from the prior paragraph

Thanks

fantactuka:
#1943 introduced a special code path for Safari to handle performance issues around newlines, which altered the way iOS handlese newlines as well. E.g., when user hits enter iOS inserts new paragraph and auto-presses Shift (so that newline started with capital letter). If user hits enter again, we'd treat it as Shift+Enter which is insterting linebreak instead of another paragraph. This removes that alteration for iOS and only keeps for Desktop safari to allow native handling for newlines

Fix #4335 
Fix #4266

### DIFF
--- a/packages/lexical-plain-text/src/index.ts
+++ b/packages/lexical-plain-text/src/index.ts
@@ -44,6 +44,7 @@ import {
   CAN_USE_BEFORE_INPUT,
   IS_APPLE_WEBKIT,
   IS_IOS,
+  IS_SAFARI,
 } from 'shared/environment';
 
 function onCopyForPlainText(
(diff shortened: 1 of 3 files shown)


thegreatercurve (approved):

fantactuka:
From offline discussion: we might be able to clean up some changes from #1943 instead of patching it (given #1938 was addressed with repainting fix with `outline: none`).

fantactuka:
So it appears Safari does not trigger "insertLineBreak" neither for mobile, nor desktop. So the fix would be ignoring shift key flag on mobile in favor of creating new paragraphs instead of line breaks. 

Code from #1943 is still valuable as it fixes lag when press and hold Shift + Enter in Safari to insert multiple lines.

PR #4611, opened 2023-06-06
Don't throw error on format element

### LINKED ISSUE #4601: Bug: Editor crashes when trying to align a Horizontal Rule
Lexical version: 0.11.1

## Steps To Reproduce

1. Go to the [playground](https://playground.lexical.dev/) 
2. Insert a horizontal rule anywhere on the editor
3. Select the horizontal rule, either alone or as a multiple-lines selection
4. Click on any of the alignment or out/indent action items in the toolbar
5. Check the error message in the console, which in dev is something like this:
```bash
Uncaught Error: Expected node N to have closest block element node.
from $getNearestBlockElementAncestorOrThrow at LexicalUtils.dev.js
```

The editor stops working after the error.

## The current behavior
When you try to align a selection containing a horizontal rule the editor "crashes" and the action is not completed.
## The expected behavior
The action completes and the editor keeps working.

thegreatercurve:
Fixes #4601

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -35,6 +35,7 @@ import {
   $shouldOverrideDefaultCharacterSelection,
 } from '@lexical/selection';
 import {
+  $findMatchingParent,
   $getNearestBlockElementAncestorOrThrow,
   addClassNamesToElement,
   mergeRegister,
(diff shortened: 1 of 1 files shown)


fantactuka (approved):

fantactuka:
🚢

PR #4613, opened 2023-06-06
pared down lexical playground removing unnecessary stuff but still works

sidhire:


sidhire:
Oops, sorry about that. Was intending to make the PR on my fork.

PR #4616, opened 2023-06-07
fix: freeze get style object in DEV env to avoid mutations

### LINKED ISSUE #3946: Bug: getStyleObjectFromCSS should return a read-only object
Lexical version: 0.8.0

The object returned by `getStyleObjectFromCSS` is the same object stored in the style cache. Mutating it will have undesired effects since the cache key will no longer correspond to the contents of the object. This causes bugs like #3944.

We should, at a minimum, return a `Readonly<>` object so that TS users get an error if they try to mutate the object. We could possibly freeze the object too if we wanted the extra runtime guarantee.

Happy to make up a PR if you agree with this change. It would be a breaking change to the current API but any code broken by the change is possibly already broken in practice.

JatinRanka:
closes #3946

Currently, the `getStyleObjectFromCSS` returns the object which is directly mutable. So, added logic to freeze the object before returning to avoid accidental mutations in __DEV__.

### DIFF
--- a/packages/lexical-selection/src/utils.ts
+++ b/packages/lexical-selection/src/utils.ts
@@ -199,6 +199,12 @@ export function getStyleObjectFromCSS(css: string): Record<string, string> {
     value = getStyleObjectFromRawCSS(css);
     CSS_TO_STYLES.set(css, value);
   }
+
+  if (__DEV__) {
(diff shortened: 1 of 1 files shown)


fantactuka:
Looks good! 
Please accept CLA so that we can merge

fantactuka (approved):

PR #4620, opened 2023-06-07
fix issue with dead keys (i.e. backticks) when converting markdown

mbongard:
On keyboards that use [dead keys](https://en.wikipedia.org/wiki/Dead_key) for backticks the markdown converter does not work properly. The second backtick (for inline code) triggers the conversion, before the user has confirmed the key.
See here:

https://github.com/facebook/lexical/assets/32648275/f3535af7-4450-4dd8-b74a-0fe88d40ca3e

The event should be ignored, if the editor is still composing.

Further below I adjusted the offset check slightly, to allow the conversion to markdown when the user confirms the key. This is required, because the cursor will not move during these two operations.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/MarkdownShortcuts.ts
@@ -387,6 +387,11 @@ export function registerMarkdownShortcuts(
         return;
       }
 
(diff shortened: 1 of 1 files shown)


fantactuka (approved):
Thank you!

PR #4627, opened 2023-06-09
Add error formating params

### LINKED ISSUE #4624: Feature: Improve prod error message
We receive a lot of lexical errors in production, some of those errors have `%s` like https://lexical.dev/docs/error?code=67.
It would be nice to have https://lexical.dev/docs/error?code=67&v=MentionNode.

image

As I understand from this comment formatProdErrorMessage can receive 3 args

https://github.com/facebook/lexical/blob/c3a392e2ba977e1e1581b787eb83bde26c6da1dd/scripts/error-codes/transform-error-messages.js#L31-L37

but in formating only the code is used.

https://github.com/facebook/lexical/blob/c3a392e2ba977e1e1581b787eb83bde26c6da1dd/scripts/error-codes/formatProdErrorMessage.js#L15

I tried reproducing prod error message locally in lexical playground, but no success, so I am not sure if the comment above is right.

```diff
// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    replaceCodePlugin({
      replacements: [
        {
          from: /__DEV__/g,
-          to: 'true',
+          to: 'false',
        },
      ],
    }),
```

Does the idea make sense, can we do this?

fantactuka:
Add adj/noon params into error url to support error formatting:

1. Trigger error by registrering mutation on non-existing node
Screenshot 2023-06-09 at 10 20 04 AM

2. Use website preview to get formatted error

https://lexical-czmewj31g-fbopensource.vercel.app/docs/error?code=37&a=MentionNode
Screenshot 2023-06-09 at 10 21 43 AM

Fixes #4624

### DIFF
--- a/scripts/error-codes/formatProdErrorMessage.js
+++ b/scripts/error-codes/formatProdErrorMessage.js
@@ -12,9 +12,16 @@
 // template literal strings. The messages will be replaced with error codes
 // during build.
 
-function formatProdErrorMessage(code) {
+function formatProdErrorMessage(code, adj, noon) {
+  let url = `https://lexical.dev/docs/error?code=${code}`;
+  if (adj != null) {
+    url += `&a=${decodeURIComponent(adj)}`;
+  }
+  if (noon != null) {
+    url += `&n=${decodeURIComponent(noon)}`;
+  }
   throw Error(
-    `Minified Lexical error #${code}; visit https://lexical.dev/docs/error?code=${code} for the full message or ` +
+    `Minified Lexical error #${code}; visit ${url} for the full message or ` +
       'use the non-minified dev environment for full errors and additional ' +
       'helpful warnings.',
   );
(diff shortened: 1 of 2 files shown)


scripts/error-codes/formatProdErrorMessage.js
```diff
@@ -12,9 +12,17 @@
 // template literal strings. The messages will be replaced with error codes
 // during build.
 
-function formatProdErrorMessage(code) {
+function formatProdErrorMessage(code, adj, noon) {
```
zurfyx, line 15:
```suggestion
function formatProdErrorMessage(code, adj, noun) {
```

`noun` is how it's called inside `transform-error-messages.js`

zurfyx (approved):
Is this flexible enough to cater future use cases? I feel like two hardcoded params is too random.

fantactuka:
> Is this flexible enough to cater future use cases? My first impression is that two hardcoded params is too random

Ugh, missed this comment before merging 🤦‍♂️
I'll probably rework it to support arbitrary params array instead

PR #4628, opened 2023-06-09
Support custom link protocols

ivailop7:
After introducing URL sanitiziation in the LinkPlugin in 0.10, this broke some use cases that relied on custom link protocols. Introducing an argument that allows users to specify custom protocols that should be considered valid.

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -71,22 +71,29 @@ export class LinkNode extends ElementNode {
     return new LinkNode(
       node.__url,
       {rel: node.__rel, target: node.__target, title: node.__title},
+      node.__extraProtocols,
       node.__key,
     );
   }
 
-  constructor(url: string, attributes: LinkAttributes = {}, key?: NodeKey) {
+  constructor(
+    url: string,
+    attributes: LinkAttributes = {},
+    extraProtocols: Set<string> = new Set(),
+    key?: NodeKey,
+  ) {
     super(key);
     const {target = null, rel = null, title = null} = attributes;
     this.__url = url;
     this.__target = target;
     this.__rel = rel;
     this.__title = title;
+    this.__extraProtocols = extraProtocols;
   }
 
   createDOM(config: EditorConfig): HTMLAnchorElement {
     const element = document.createElement('a');
-    element.href = this.sanitizeUrl(this.__url);
+    element.href = this.sanitizeUrl(this.__url, this.__extraProtocols);
     if (this.__target !== null) {
       element.target = this.__target;
     }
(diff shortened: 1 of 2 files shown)


fantactuka:
I don't think extra protocols should be a node property. That data is stored and if you decide to support more protocols, existing content won't have it

fantactuka (changes requested):
~

ivailop7:
Sure, there are 2 sanitizeUrl methods, one in utils used in FloatingLink editor, editable from Playground-level and one inside the LinkNode class, which is used inside this class. I do think these should be consolidated. Is your suggestion, passing sanitizeUrl method as an argument to the LinkPlugin and consolidating them?

acywatson:
> Sure, there are 2 sanitizeUrl methods, one in utils used in FloatingLink editor, editable from Playground-level and one inside the LinkNode class, which is used inside this class. I do think these should be consolidated. Is your suggestion, passing sanitizeUrl method as an argument to the LinkPlugin and consolidating them?

Even though the logic is the same, these are actually two different things. The one in LinkNode is a security mitigation in the core library rendering logic. The one in FloatingLinkToolbar is a similar mitigation, but the reason it is separate is because it is not part of the public API and I don't think it should be. We can't share the logic across playground and core(@lexical/link) implementation without creating a public API (which is as it should be - the playground is a consumer app and not part of Lexical itself).

Note that you can override supported protocols right now by overriding LinkNode (as Maks pointed out) with a new sanitizeUrl method. I'm not opposed to something that makes it simpler to adjust only the supported protocols, but I haven't thought of a way that's better than just allowing the method override.

fantactuka:
We can make sanitizeURL method static on LinkNode, so that it can be called from floating dialog UI (or elsewhere, if needed)

acywatson:
> We can make sanitizeURL method static on LinkNode, so that it can be called from floating dialog UI (or elsewhere, if needed)

Yea, it's just a weird pattern - do we do that anywhere else? Utilities as static methods?

I'm not opposed, as it does accomplish the consolidation, in practice.

ivailop7:
Understood. I would still love to avoid overriding the node for just adding a single string to a set and since LinkPlugin already takes validateURL method as an argument, it felt suitable to extend this paradigm. I think making it a static class, would just create confusion for new comers to the project.

acywatson:
> Understood. I would still love to avoid overriding the node for just adding a single string to a set

Yea, I agree with this. I wonder if we can revive this idea that never really saw the light of day - it was a sort of context object that you could specify in the initial editor config and then access in the nodes in createDOM and updateDOM. What if we had that and then every node had a reference to it somehow, maybe via a property on the base LexicalNode class. Then you could add and access any kind of centralized configuration you want as a replacement for some node override use cases.

I also find it onerous to have to override nodes for simple things like this, particularly with the amount of boilerplate currently required.

ivailop7:
Will close this one, I hope a solution like what Acy described makes it in the future. We'll override the node for now on our end.

PR #4635, opened 2023-06-11
Make Excalidraw Modal render interacting user specific

### LINKED ISSUE #1788: Excalidraw modal shows on collab when the other person interacts with it
Go to split view, click Excalidraw, the modal shows on both sides. Should only display the result on both sides

ivailop7:
Fixes #1788 

Before when an Excalidraw node is added, the popup will show up for all collaborators, not just the creating user. Which would lead to a number of other issues, as shown in the Before video. Skipping the syncing for a showModal property on the ExcalidrawNode, allows for independent editing.

Before:

https://github.com/facebook/lexical/assets/7893468/817b2e48-c4ec-41ce-982d-44b60d3bb997

After:

https://github.com/facebook/lexical/assets/7893468/1bc7e901-26e7-4bc3-a601-eec8f15d2df3

### DIFF
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -29,6 +29,7 @@ import {CAN_USE_DOM} from 'shared/canUseDOM';
 import {createWebsocketProvider} from './collaboration';
 import {useSettings} from './context/SettingsContext';
 import {useSharedHistoryContext} from './context/SharedHistoryContext';
+import {ExcalidrawNode} from './nodes/ExcalidrawNode';
 import TableCellNodes from './nodes/TableCellNodes';
 import ActionsPlugin from './plugins/ActionsPlugin';
 import AutocompletePlugin from './plugins/AutocompletePlugin';
@@ -74,6 +75,12 @@ import Placeholder from './ui/Placeholder';
 const skipCollaborationInit =
   // @ts-ignore
   window.parent != null && window.parent.frames.right === window;
+
+const excludedCollaborationProps = new Map();
+excludedCollaborationProps.set(
+  ExcalidrawNode,
+  new Set<string>(['__showModal']),
+);
 
 export default function Editor(): JSX.Element {
   const {historyState} = useSharedHistoryContext();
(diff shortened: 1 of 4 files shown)


fantactuka:
Wouldn't advice storing ui state on node itself as it might backfire. Eg here you filter it out from collab but it'll be stored in undo stack for non collab. Maybe we can come up with a better approach

fantactuka:
Internally we did following:
Node only stores drawing state, plug-in is responsible for rendering drawings dialog and registering INSERT/UPDATE_EXCALIDRAW_COMMAND. Insert command inserts node into editor and opens dialog (isModalOpen is internal state of plugin) and sends nodeKey to dialog. When user hits save it'll use node key and update drawing state on node

ivailop7:
> Internally we did following: Node only stores drawing state, plug-in is responsible for rendering drawings dialog and registering INSERT/UPDATE_EXCALIDRAW_COMMAND. Insert command inserts node into editor and opens dialog (isModalOpen is internal state of plugin) and sends nodeKey to dialog. When user hits save it'll use node key and update drawing state on node

Gerard offered the excludedProperties as an approach, which seemed straightforward to me. Your suggestion sounds better, but I'm not sure how to skip propagation of the show modal event to the other collaborators. If you have this solved already, might be easier to just port the changes over.

fantactuka:
> @fantactuka, @acywatson, I think it's worth a discussion on when to use offline properties and the tooling we want to provide around them. It's nice that we added the ability to customize what's synced over the wire but:

Do you recall what was initial usecase when it was added? With non-collab history and persisted content (when lexical json is stored), I wonder if it's useful or whether it can be replaced by plugin-stored state

zurfyx:
> > @fantactuka, @acywatson, I think it's worth a discussion on when to use offline properties and the tooling we want to provide around them. It's nice that we added the ability to customize what's synced over the wire but:
> 
> Do you recall what was initial usecase when it was added? With non-collab history and persisted content (when lexical json is stored), I wonder if it's useful or whether it can be replaced by plugin-stored state

AFAIK the public `excluded` property was introduced to solve Excalidraw as well 😛  (@acywatson might be able to share more details on this first iteration)

ivailop7:
Waiting on @fantactuka for best practice around handling local state

StyleT:
Still reproducible :P

Issue #2108, opened 2022-05-08
Feature: Ability to render the editor in an iframe


cmmartin:
I'm trying to render the editor inside an iframe (via a react portal), but it's not working and I suspect this is due to the fact that it's referencing the global window object for selection state, etc. I'd like to be able to pass a custom window object (the iframe window)

One solution is to provide the window object in a react context. It can be passed in via an optional prop to the `RichTextPlugin` or `PlainTextPlugin` (or maybe even `LexicalComposer`) and would default to `window`. 

Alternatively, [Slate](https://github.com/ianstormtaylor/slate) reads the window from the contentEditable DOM node. So if you have a ref to that DOM node, you can look up it's parent window with something like...

```
contentEditableNode.current.ownerDocument.defaultView
```

We can expose this as a helper and replace all references to the global `window` with the helper like Slate...

https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/plugin/react-editor.ts#L54-L64

trueadm:
I think we can extract it from the contenteditable like Slate does.

CarlosBalladares:
this would be useful

trueadm:
Thanks for reporting this. Fixed in #2918 :)

darabun:
I'm trying to get this working but can't seem to get it to work when it's in the iframe.

I have it reproduced here: https://codesandbox.io/s/aged-frog-iu8ff7

It works when outside of the iframe but not when inside it.
Is there something obvious that I'm doing wrong?

cmmartin:
@trueadm can we reopen this? I'm facing the same issue as @darabun. 

Interestingly, I can almost make it work if I provide an initial editor state with some text content like this: https://codesandbox.io/s/modern-wave-1ehl2o?file=/src/Lexical.js

But you'll notice the Enter/Return character doesn't work in the iframe, but works outside it. It seems like there are some event listeners missing in the iframe.

cmmartin:
@darabun See https://github.com/facebook/lexical/issues/3534

trueadm:
Let me know if this PR helps: https://github.com/facebook/lexical/pull/3573

cmmartin:
Definitely an improvement, but there are still selection issues. It seems like we're getting the proper selection part of the time now. Try inserting a line break in the iframe here...

https://codesandbox.io/s/sweet-fire-e9ry1d

trueadm:
How about https://github.com/facebook/lexical/pull/3578?

[...]

PR #4649, opened 2023-06-13
Fix clipboard if content editable is in an iframe.

imdreamrunner:
This is also my first contribution attempt to the project. Sorry in advance if something doesn't make sense!

In #2108  we support for iframe (use React.Portal to render the content editable area in an iframe). It works well.

Recently I notice that some of the clipboard logic doesn't work in iframe. More specifically, I found that `DecoratorNode` won't be exported on copy.

After digging into it, I found that is caused by global class variables having different instances in different window. There is more detail about this behaviour [in MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms).

In this diff, the issue I personally encounter has been fixed. But there are wider usage of `instanceof` in the codebase.

Overall, I want to hear from maintainers about

- Whether it's okay to directly use internal field `_window` from `LexicalEditor`
- Should I add another example / another mode of `<Editor />` for usage with iframe

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -12,7 +12,7 @@ import {
   $cloneWithProperties,
   $sliceSelectedTextNodeContent,
 } from '@lexical/selection';
-import {$findMatchingParent} from '@lexical/utils';
+import {$findMatchingParent, objectKlassEquals} from '@lexical/utils';
 import {
   $createParagraphNode,
   $createTabNode,
@@ -585,13 +585,15 @@ export async function copyToClipboard(
   }
 
   const rootElement = editor.getRootElement();
-  const domSelection = document.getSelection();
+  const windowDocument =
+    editor._window == null ? window.document : editor._window.document;
+  const domSelection = windowDocument.getSelection();
   if (rootElement === null || domSelection === null) {
     return false;
   }
-  const element = document.createElement('span');
+  const element = windowDocument.createElement('span');
   element.style.cssText = 'position: fixed; top: -1000px;';
-  element.append(document.createTextNode('#'));
+  element.append(windowDocument.createTextNode('#'));
   rootElement.append(element);
   const range = new Range();
   range.setStart(element, 0);
@@ -602,13 +604,13 @@ export async function copyToClipboard(
     const removeListener = editor.registerCommand(
       COPY_COMMAND,
       (secondEvent) => {
-        if (secondEvent instanceof ClipboardEvent) {
+        if (objectKlassEquals(secondEvent, ClipboardEvent)) {
           removeListener();
           if (clipboardEventTimeout !== null) {
             window.clearTimeout(clipboardEventTimeout);
             clipboardEventTimeout = null;
           }
-          resolve($copyToClipboardEvent(editor, secondEvent));
+          resolve($copyToClipboardEvent(editor, secondEvent as ClipboardEvent));
         }
         // Block the entire copy flow while we wait for the next ClipboardEvent
         return true;
@@ -622,7 +624,7 @@ export async function copyToClipboard(
       clipboardEventTimeout = null;
       resolve(false);
     }, EVENT_LATENCY);
-    document.execCommand('copy');
+    windowDocument.execCommand('copy');
     element.remove();
   });
 }
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -39,6 +39,7 @@ import {
   $getNearestBlockElementAncestorOrThrow,
   addClassNamesToElement,
   mergeRegister,
+  objectKlassEquals,
 } from '@lexical/utils';
 import {
   $applyNodeReplacement,
@@ -449,7 +450,10 @@ async function onCutForRichText(
   event: CommandPayloadType<typeof CUT_COMMAND>,
   editor: LexicalEditor,
 ): Promise<void> {
-  await copyToClipboard(editor, event instanceof ClipboardEvent ? event : null);
+  await copyToClipboard(
+    editor,
+    objectKlassEquals(event, ClipboardEvent) ? (event as ClipboardEvent) : null,
+  );
   editor.update(() => {
     const selection = $getSelection();
     if ($isRangeSelection(selection)) {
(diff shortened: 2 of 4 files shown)


packages/lexical-rich-text/src/index.ts
```diff
@@ -983,7 +989,12 @@ export function registerRichText(editor: LexicalEditor): () => void {
     editor.registerCommand(
       COPY_COMMAND,
       (event) => {
-        copyToClipboard(editor, event instanceof ClipboardEvent ? event : null);
+        copyToClipboard(
+          editor,
+          event instanceof getClassFromEditorWindow(editor, ClipboardEvent)
+            ? event
+            : null,
```
zurfyx, line 996:
```suggestion
	 Object.getPrototypeOf(event).constructor.name === ClipboardEvent.name ? event : null,
```

where ClipboardEvent.name would be hoisted

Given that we're also using this in the Clipboard module we can also abstract this on a clipboard exported function.

zurfyx (commented):
Thank you for addressing this, left a small comment regarding how we do the `instanceof` comparison. We have be careful it works fine for headless mode, including JSDOM which is quite popular. I proposed a similar approach to what we use for `importDOM`, where instead of instance checks we look for specific properties instead. It's not as type-safe but it will work for non-browser environments.

packages/lexical-rich-text/src/index.ts
imdreamrunner: That's a good idea! I updated the PR using the new approach and not relying on the window object. I've created a util function that at least makes TypeScript happy about it.

packages/lexical-utils/src/index.ts
```diff
@@ -511,3 +511,29 @@ export function isHTMLElement(x: Node | EventTarget): x is HTMLElement {
   // @ts-ignore-next-line - strict check on nodeType here should filter out non-Element EventTarget implementors
   return x.nodeType === 1;
 }
+
+type IsInstanceOfSupportedType = Event | Node;
+export type ConstructorOf<C> = {new (...args: never[]): C};
```
zurfyx, line 516:
```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
type Klass<T> = new (...args: any[]) => C;
```

packages/lexical-utils/src/index.ts
```diff
@@ -511,3 +511,29 @@ export function isHTMLElement(x: Node | EventTarget): x is HTMLElement {
   // @ts-ignore-next-line - strict check on nodeType here should filter out non-Element EventTarget implementors
   return x.nodeType === 1;
 }
+
+type IsInstanceOfSupportedType = Event | Node;
```
zurfyx, line 515:
```suggestion
```

packages/lexical-utils/src/index.ts
```diff
@@ -511,3 +511,29 @@ export function isHTMLElement(x: Node | EventTarget): x is HTMLElement {
...
+
+/**
+ * @param object = The instance of the global type
+ * @param globalClass = The class of the global type
+ * @returns Whether the object is an instance of the global class, ignoring the difference across window (e.g. different iframs)
+ */
+export function isInstanceOfCrossWindow<T extends IsInstanceOfSupportedType>(
+  object: unknown,
+  globalClass: ConstructorOf<T>,
+): object is T {
```
zurfyx, line 526:
```suggestion
): boolean {
```

This is not correct given that we're just checking equality. The object Class might not be the same as the Class you're comparing against

packages/lexical-utils/src/index.ts
```diff
@@ -511,3 +511,29 @@ export function isHTMLElement(x: Node | EventTarget): x is HTMLElement {
...
+/**
+ * @param object = The instance of the global type
+ * @param globalClass = The class of the global type
+ * @returns Returns the instance if it is an instance of the global type
+ */
+export function coerceInstanceOfCrossWindow<
+  T extends IsInstanceOfSupportedType,
+>(object: unknown, globalClass: ConstructorOf<T>): T | null {
+  return isInstanceOfCrossWindow(object, globalClass) ? object : null;
+}
```
zurfyx, line 539:
```suggestion

```

Remove, see previous comment

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -585,6 +585,8 @@ export async function copyToClipboard(
   }
 
   const rootElement = editor.getRootElement();
+  const document =
+    editor._window == null ? window.document : editor._window.document;
```
zurfyx, line 589:
```suggestion
```

Not used? Or shadowing? If it's relevant let's use another name

packages/lexical-rich-text/src/index.ts
```diff
@@ -983,7 +987,10 @@ export function registerRichText(editor: LexicalEditor): () => void {
     editor.registerCommand(
       COPY_COMMAND,
       (event) => {
-        copyToClipboard(editor, event instanceof ClipboardEvent ? event : null);
+        copyToClipboard(
+          editor,
+          coerceInstanceOfCrossWindow(event, ClipboardEvent),
```
zurfyx, line 992:
```suggestion
          objectKlassEquals(event, ClipboardEvent) ? event as ClipboardEvent : null,
```

See comments below

packages/lexical-utils/src/__tests__/unit/LexicalUtilsInstanceOf.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  MyEventShadow = MyEvent;
+}
+
+describe('LexicalUtilsInstanceOf tests', () => {
+  initializeUnitTest((testEnv) => {
+    it('isInstanceOfCrossWindow', async () => {
+      const eventInstance = new MyEvent('');
+      expect(eventInstance instanceof MyEvent).toBeTruthy();
+      expect(isInstanceOfCrossWindow(eventInstance, MyEvent)).toBeTruthy();
+      expect(isInstanceOfCrossWindow(eventInstance, MyEvent2)).toBeFalsy();
```
zurfyx, line 33:
```suggestion
      expect(eventInstance instanceof MyEvent2).toBeFalsy();
      expect(isInstanceOfCrossWindow(eventInstance, MyEvent2)).toBeFalsy();
```

zurfyx (commented):
Thank you! Some final nits

packages/lexical-clipboard/src/clipboard.ts
imdreamrunner: It's shadowing. Let me rename

zurfyx (approved):
LGTM, thank you for iterating on this!

PR #4654, opened 2023-06-13
Fix reconciler recovery listeners

fantactuka:
When reconciler fails we try to recover editor by reseting it to empty state and then doing full reconciliation for `pendingEditorState`. If that goes through, we trigger listeners but the problem is that since we reset editor during recovery `prevEditorState` will be empty, while it's expected to be whatever state was before changes that caused reconciler failre. This PR adjusts `commitPendingUpdates` to ensure it uses proper prevEditorState for those listeners.

This issue affecting collab in its worst way since collab heavily relies on `prevEditorState` from update listener to calculate require Yjs updates for mutated nodes. And since in reconciler recovery scenario prev state is empty, collab assumes that all mutated nodes didn't exist before and re-inserts them causing content duplication. Here's an example of MentionNode that has error thrown inside `updateDOM` method:

https://github.com/facebook/lexical/assets/132642/f1a39631-5526-49c7-b330-00071f7618ee

### DIFF
--- a/packages/lexical/src/LexicalUpdates.ts
+++ b/packages/lexical/src/LexicalUpdates.ts
@@ -433,7 +433,10 @@ function handleDEVOnlyPendingUpdateGuarantees(
   };
 }
 
-export function commitPendingUpdates(editor: LexicalEditor): void {
+export function commitPendingUpdates(
+  editor: LexicalEditor,
+  recoveryEditorState?: EditorState,
+): void {
   const pendingEditorState = editor._pendingEditorState;
   const rootElement = editor._rootElement;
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalUpdates.ts
```diff
@@ -645,8 +650,6 @@ function triggerTextContentListeners(
 
 function triggerMutationListeners(
   editor: LexicalEditor,
-  currentEditorState: EditorState,
-  pendingEditorState: EditorState,
```
fantactuka, line 649: This is just cleanup, we didn't use these args

zurfyx (approved):

PR #4655, opened 2023-06-13
Check for specific excluded props

fantactuka:
`excludedProperties` is a global list of properties that are not synced in collab mode, which includes node-specific props like `__size`, `__first`, `__last` (that should be filtered for elements only), or `__text` which is relevant for text nodes. That introduces possible issue where a decorator node (let's say Emoji) can have perfectly valid `__size` property, but it won't be sync'd over collab because it's filtered out for all nodes.

This PR checks if property should be excluded based on node type (element, text, root) and any extra excluded props from plugin to avoid it

Ref #3670

### DIFF
--- a/packages/lexical-yjs/src/Utils.ts
+++ b/packages/lexical-yjs/src/Utils.ts
@@ -79,7 +79,7 @@ function isExcludedProperty(
 
   const nodeKlass = node.constructor as Klass<LexicalNode>;
   const excludedProperties = binding.excludedProperties.get(nodeKlass);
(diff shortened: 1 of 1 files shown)


acywatson:
https://github.com/facebook/lexical/pull/4275

fantactuka:
> #4275

That one is still in place, it's just core properties are filtered out more precisely (e.g., __size only filtered from element nodes, but not text/decorators)

ivailop7 (approved):
Nice addition!

PR #4664, opened 2023-06-15
Normalize select all on NodeSelection

### LINKED ISSUE #4658: Bug: `RangeSelection` using root splices text from last node

zurfyx:
The following selection is invalid, it should point to the corresponding child nodes

Screenshot 2023-06-15 at 5 09 33 PM

Fixes #4658


9larsons:
Heads up, if a `DecoratorNode` (guessing any empty node w/r to text) is first in the document, when you Select All you won't get the first node selected. (ie starting with a `RangeSelection`)

This fix makes it so that if you come from a `NodeSelection` that Select All does work, including when the first node is a `Decorator`/empty.

fantactuka (approved):

zurfyx:
> Heads up, if a `DecoratorNode` (guessing any empty node w/r to text) is first in the document, when you Select All you won't get the first node selected. (ie starting with a `RangeSelection`)
> 
> This fix makes it so that if you come from a `NodeSelection` that Select All does work, including when the first node is a `Decorator`/empty.

https://github.com/facebook/lexical/pull/4671

Issue #4517, opened 2023-05-18
Bug: Listeners fire when the initial editorState is changed


EgonBolton:
Lexical version: 0.9.1

## Steps To Reproduce

1. In the code-sandbox, use the buttons to switch between note 1 and note 2.

Link to code example: https://codesandbox.io/s/green-fog-l5wu7r?file=/src/Editor.js

## The current behavior
The listeners are fired, as seen in the console.

## The expected behavior
I think the listeners should not fire in this case. If someone wants to use multiple editors, going to another editor and back would trigger all registered listeners unnecessarily.

EgonBolton:
Ok, I found that what was causing the error in my project and in the codesandbox were different reasons. In the codesandbox, it was due to the AutoFocusPlugin and in my project, a dynamic attribute on the CollaborationPlugin.

The AutoFocusPlugin thing shouldn't be a problem, as it only fires listeners that somehow listen to the selection, which makes sense. The problem is when CollaborationPlugin is passed a dynamic "id" attribute, every time the editor/document is changed, all the listeners are fired, including the text listener for example.

I've updated the codesandbox to replicate this error, but I couldn't get the y-websocket server to work. I think in order to replicate this bug, you would have to copy the code locally and run `npx y-websocket-server` in a terminal.

EgonBolton:
UPDATE: merged the two messages above, improved the problem explanation and got the repro to work (can't be codesandbox for collab server to work)

PR #4666, opened 2023-06-16
Fix: Listeners fire on initialization with CollaborationPlugin

GermanJablo:
this fix https://github.com/facebook/lexical/issues/4517

There are some updates in the Collaboration plugin that fire listeners when an editor is initialized.

This PR prevents the update in `syncYjsChangesToLexical` from firing the listeners when that function is first executed using a tag.

I know that there are other listeners in CollabPlugin that also fire listeners on initialization and that I have not been able to solve it with the same strategy, although it is solved by remounting the LexicalComposer completely: (`<LexicalComposer key={noteId} >`).

The PR only contains the code that fixes the bug. If you want to experiment with the reproducible code you can do it in my `collab` branch.


fantactuka:
I'd want to understand a bit more about your case. I think it's collab and non-collab initial states and on change are completely different things and it's ok to behave differently. 
- For non-collab we know initial state in advance (let's say it's fetched from server or local storage), with that we can filter out listeners when initial state is applied (if we want to).
- For collab, you don't really know whether the state you've received is initial state, or initial doc was empty and it's just another user who opened the same document and added new content to it. In that scenario you wouldn't really want to ignore the change

GermanJablo:
@fantactuka I am going to expand the explanation:

The reason why updating non-collab when initializing the initialEditorState doesn't fire the listeners, is because they haven't been registered yet (because the lexicalComposer is executed before the plugins).

However, in Collab that does not happen. If you refresh the page or change the document, all the listeners fire.

I don't think anyone wants to fire listeners on editor initialization, even if there were changes made by the user or a third party on another device.

In fact, the principle is the same for collab and non-collab. The normal thing is to make a call to the DB to receive the initialEditorState that is passed to LexicalComposer. Why wouldn't it be of interest there to trigger the listeners and in collab to do so?

> it can be done with the same ref-based flag approach on app level

Not really. You would have to opt-in each update to add the flag to it, and then check the same on each listener. I don't see much point in it, and it seems very easy to forget to do that.

Is it better understood now? Please let me know if you still have questions

ivailop7:
@ebads67 to have a look as well

fantactuka:
Related to OnChange plugin I think it's worth exploring `history-merge` tag. Here's some background for it:

When you register node transformer it will immediately mark all existing nodes of that type as dirty and cause transformers to run upon registration. That will ensure that any pre existing content of the editor will go through newly added node transformer.

Quite often (mainly depending on whether your content is initialized sync or async) plugins are initialized after composer has set its content. When plugin includes node transformers, it would run those and potentially update state based on transformers logic triggering corresponding listeners (update, text content, mutations, etc). This updates could be counted as part of content initialization (and are often so), and we mark those with `history-merge` tag (mainly to avoid adding it into history stack). 

That's the reason #2706 introduced `history-merge` tag check for onChange plugin as well to filter out changes potentially triggered by such plugins during initialization. That same tag used in few other places, including initalizeEditorState for both plain and rich text plugins.

With all that said, I think it might be a better option to use `history-merge` tag for initial received state for the collab. It'd make it behave the same way as non-collab with async content (when composer is initialized with editorState: null, and then setEditorState called from async collab, e.g., once data loaded from server)

I'd be very careful with cutting off all listeners from collab initialization, as there's a high chance some app plugins would rely on those.

GermanJablo:
Let me see if I understand you:

Are you saying that in initialization history should be merged but the listeners should be executed?

If so, I have 2 questions:

1. Why should this be the behavior for async updates, node transforms, or collab initialization, but not for sync updates after composer initialization?
2. I admit that with a little imagination, I can think of some scenarios in which some listeners should be executed. However, in my experience, this is not the case with 100% of my listeners and has caused bugs (which I didn't describe so as not to make this too long). It would be nice to allow each listener to choose whether to be executed in this scenario (something my current PR doesn't allow). But, **_wouldn't it be better if the default is that they are not executed_**? I believe that on initialization the listeners should opt-in their execution since this is the exception rather than the rule.

fantactuka:
> wouldn't it be better if the default is that they are not executed

No, because then you leave no option to have it if app needs it. E.g., let's assume there's an app that has comments within content, where CommentNode stores comment id, and comments are stored separately. So when content is initialized update listener (or mutation listener) could be used to fetch comments related data based on those IDs.

GermanJablo:
> No, because then you leave no option to have it if app needs it

I mean, my point is to allow the user to choose for each listener, which is doable. The question was about the default behavior.

And about question 1?

GermanJablo:
And a third question:
3. If I want to display a loading before the provider syncs the initial content, how should I do it?

It is currently not possible to achieve this without having to pass state from a child to a parent (CollabPlugin to Editor).

One idea is that CollabPlugin instead of receiving a providerFactory receives a provider already created. This way you could do something like:

```typescript
export default function Editor() {
  const [isLoading, setIsLoading] = useState(true);
  // ...
  provider.on("sync", () => setIsLoading(false));
  
  if (isLoading) return <p>loading...</p>;
  return (
       // ...
      <CollaborationPlugin
        id={id}
        provider={provider}
      />
    </LexicalComposer>
  );
}
```

fantactuka:
> 3. If I want to display a loading before the provider syncs the initial content, how should I do it?
> 
> It is currently not possible to achieve this without having to pass state from a child to a parent (CollabPlugin to Editor).
> 
> One idea is that CollabPlugin instead of receiving a providerFactory receives a provider already created. This way you could do something like:

Similar approach with factory:

```js
const providerFactory = useCallback(() => {
  const provider = createProvider();
  provider.once('sync', onLoad);
  return provider;
}, []);

  <CollaborationPlugin id={id} providerFactory={providerFactory} />
  {isLoading && <Loading />}
  ...
</LexicalComposer>;
```


> Why should this be the behavior for async updates, node transforms, or collab initialization, but not for sync updates after composer initialization?

I would say it's rather flaw of current sync initial states, than issue with async. E.g., transformers running on register mainly done to fight that flaw. Similar story with mutation listener not being fired for initial content which requires adding separate logic with $nodeOfType for plugins.

GermanJablo:
> Similar approach with factory:

That is exactly what I tried. But to not show both the editor AND the loading, I should conditionally render the RichTextPlugin at least, and that didn't work for me.

> I would say it's rather flaw of current sync initial states, than issue with async. E.g., transformers running on register mainly done to fight that flaw. Similar story with mutation listener not being fired for initial content which requires adding separate logic with $nodeOfType for plugins.

processing

PR #4671, opened 2023-06-16
Override selectall on RangeSelection

### LINKED ISSUE #4670: Bug: CMD + A doesn't capture first DecoratorNode
https://github.com/facebook/lexical/assets/193447/fa43bc01-5c2d-428b-aa25-74f7b434014b

For DecoratorNodes (because of contenteditable = false) select all doesn't capture those elements when at its boundaries. Surprisingly, it works well for FF

zurfyx:
See issue description. Given that it works well on FF, I've decided not to overide FF's behavior for now. Safari's behavior is sometimes worse than Chrome, caret jumping around instead of selecting so Chrome + Safari are a must-do override.

I didn't narrow down to checking whether there's DecoratorNodes at its boundaries because the root cause seems to be `contenteditable=false` but it doesn't necessarily have to sit at the top

```
<div contenteditable="true"><div contenteditable="false">...</div></div>
```

^ This will not work on Chrome

Might be that the pattern is first line of descendants not contenteditable (given the correct valid Nodes that can take this properties (i.e. iframe can't take it)) but since I'm not too confident on what the pattern is and the logic might be overly complicated, I'm overriding them all.

Fixes #4670

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -28,7 +28,6 @@ import {
   $isRangeSelection,
   $isRootNode,
   $isTextNode,
-  $normalizeSelection__EXPERIMENTAL,
   $setCompositionKey,
   BLUR_COMMAND,
   CLICK_COMMAND,
@@ -80,6 +79,7 @@ import {
   $getNodeByKey,
   $isSelectionCapturedInDecorator,
   $isTokenOrSegmented,
+  $selectAll,
   $setSelection,
   $shouldInsertTextAfterOrBeforeTextNode,
   $updateSelectedTextFromDOM,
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -1019,11 +1019,15 @@ function onKeyDown(event: KeyboardEvent, editor: LexicalEditor): void {
       } else if (isSelectAll(keyCode, metaKey, ctrlKey)) {
         event.preventDefault();
         editor.update(() => {
-          const root = $getRoot();
-          const selection = root.select(0, root.getChildrenSize());
-          $setSelection($normalizeSelection__EXPERIMENTAL(selection));
+          $selectAll();
```
fantactuka, line 1022: Do we even need to check what kind of selection we have to run selectAll? Would be nice to have cmd+a when have grid selection too

fantactuka (approved):

PR #4672, opened 2023-06-16
Paste decorators before empty text nodes

9larsons:
Right now if you paste a decorator node (like the Youtubenode) on an empty text node, it'll insert the decorator node after the empty text. This feels a bit unintuitive, as then the user has a blank line.

Instead, if there's no text, we should insert the decorator before the empty text node so the user can go on typing, adding other content, etc.

New behavior:
CleanShot 2023-06-16 at 10 36 46

Old/current behavior (on main):
CleanShot 2023-06-16 at 10 37 41


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1550,7 +1550,11 @@ export class RangeSelection implements BaseSelection {
       if ($isElementNode(target) && !target.isInline()) {
         lastNode = node;
         if ($isDecoratorNode(node) && !node.isInline()) {
-          target = target.insertAfter(node, false);
+          if (target.isEmpty()) {
```
zurfyx, line 1553:
```suggestion
          if (target.canBeEmpty() && target.isEmpty()) {
```

zurfyx (commented):
I can't recall whether this was actually intentionally to avoid the impossible to move cursor (before the block cursor was implemented). Otherwise, LGTM! Can we write a test for it please?

@fantactuka Mind having a pass to it since you were looking into insertNodes?

fantactuka (approved):
👍

9larsons:
Added a test, although I'm not sure if that's the best home for it.

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
```diff
@@ -855,4 +856,52 @@ test.describe('CopyAndPaste', () => {
...
+    test.skip(isPlainText);
+
+    // copying and pasting the node is easier than creating the clipboard data
+    const TEST_URL = 'https://www.youtube-nocookie.com/embed/jNQXAC9IVRw';
+    await focusEditor(page);
+    await insertYouTubeEmbed(page, TEST_URL);
+    await page.keyboard.press('ArrowLeft');
+    await page.keyboard.press('Meta+C');
+    await page.keyboard.press('ArrowRight');
+    await page.keyboard.press('Meta+V');
```
zurfyx, line 873:
```suggestion
    await copyToClipboard(page);
```

The Meta key only works for Mac

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
```diff
@@ -855,4 +856,52 @@ test.describe('CopyAndPaste', () => {
...
+  test('Pasting a decorator node on a blank line inserts before the line', async ({
+    page,
+    isPlainText,
+  }) => {
+    test.skip(isPlainText);
+
+    // copying and pasting the node is easier than creating the clipboard data
+    const TEST_URL = 'https://www.youtube-nocookie.com/embed/jNQXAC9IVRw';
+    await focusEditor(page);
+    await insertYouTubeEmbed(page, TEST_URL);
```
zurfyx, line 869:
```suggestion
    await insertYouTubeEmbed(page, YOUTUBE_SAMPLE_URL);
```

You can import this from utils

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
```diff
@@ -855,4 +856,52 @@ test.describe('CopyAndPaste', () => {
...
   });
+
+  test('Pasting a decorator node on a blank line inserts before the line', async ({
+    page,
+    isPlainText,
+  }) => {
+    test.skip(isPlainText);
+
+    // copying and pasting the node is easier than creating the clipboard data
+    const TEST_URL = 'https://www.youtube-nocookie.com/embed/jNQXAC9IVRw';
```
zurfyx, line 867:
```suggestion
```

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
zurfyx, line 873: This isn't really copy-pasting, you didn't do a range selection or click, there's a `click` function, that might work

Also, you can do `await page.pause()` with `debug-test-e2e-chromium` to understand what the E2E test is seeing

zurfyx (commented):
Thank you for the test! Adding some final nits on the test itself

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
9larsons: Had to sync with `main` to get it 👍 .

9larsons:
@zurfyx not sure why but the test is now failing with the insert YouTube embed locally.. any idea? All I changed was how the url was added.

9larsons:
There's a couple tests not passing on `main` that are also failing here.

zurfyx:
> There's a couple tests not passing on `main` that are also failing here.

Do you mind rebasing? Main should be fine now. We're aware of some heavy flakiness for a couple Linux tests but otherwise the CI should pass

zurfyx:
The below tests seem to be consistently failing, can you revise these please?

[1]   2 failed
[1]     [webkit] › packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/HTMLCopyAndPaste.spec.mjs:195:3 › HTML CopyAndPaste › Copy + paste a paragraph element between horizontal rules 
[1]     [webkit] › packages/lexical-playground/__tests__/e2e/HorizontalRule.spec.mjs:286:3 › HorizontalRule › Can copy and paste a horizontal rule

9larsons:
Hey @zurfyx sorry for the delay - tests were actually surfacing a likely unintended consequence of this update. I changed the logic to only use `insertBefore` if pasting only one node. Otherwise what you'd end up with out of order pasted contents. Let me know if that looks all right to you.

acywatson (approved):

9larsons:
@zurfyx hey - would you be able to take another pass at this?

9larsons:
@zurfyx - is this something you're still willing to merge?

PR #4689, opened 2023-06-22
Table selection keyboard controls

fantactuka:
This diff addresses some table selection around keyboard (arrow/tab) controls:

- Allow shift+arrows to select content within cells (we used to select whole cell right away without possibility of simple text selection). This part might require more polishing as there's some range bounding rects calculations to predict whether cursor will move to the next cell, and for now it's intentionally simplified to see if we actually need anything more complicated
- Allow escaping grid selection with arrows or escape key
- `hasTabHandler` to enable tab key switching between cells. Although handle itself makes sense, when app tends to have more complex content within cells like lists, code, or inventible paragraphs tab override to jump through cells does not make much sense
- Refactored/simplified drag-to-select event handlers
- Moved table selection hijacking css into theme class instead of hardcoded class

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/TablesHTMLCopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/TablesHTMLCopyAndPaste.spec.mjs
@@ -264,7 +264,8 @@ test.describe('HTML Tables CopyAndPaste', () => {
       page,
       html`
         <p class="PlaygroundEditorTheme__paragraph"><br /></p>
-        <table class="PlaygroundEditorTheme__table disable-selection">
+        <table
+          class="PlaygroundEditorTheme__table PlaygroundEditorTheme__tableSelection">
           <tr>
             <th
               class="PlaygroundEditorTheme__tableCell PlaygroundEditorTheme__tableCellHeader"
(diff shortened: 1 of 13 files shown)


ivailop7 (approved):

zurfyx (approved):
Wow, this looks great!

Some feedback:
- Should the tab key select the entire content of the next cell?
- Why doesn't the tab key work with a range selection inside a cell? Is this intended?
- I believe shift+Arrow at a boundary should first select the entire cell instead of two?
- I don't agree with the Escape key behavior, it should mimic a browser RangeSelection

I think the current approach is good even though (as discussed offline) we might want to change some bits to rely on computed position instead of boundaries but this one will certainly perform much better for now

fantactuka:
Some feedback:

> Should the tab key select the entire content of the next cell?
> Why doesn't the tab key work with a range selection inside a cell? Is this intended?

For both above: I didn't change tab key for now, I think it might need this changes plus some tweaks to work with tabIndent plugin

> I believe shift+Arrow at a boundary should first select the entire cell instead of two?

Talked about it a bit offline, tldr: boundary calculation are intentionally simplified for this version to test a bit, can make it more precise if need (might cover some edge cases)

> I don't agree with the Escape key behavior, it should mimic a browser RangeSelection

Yeah, it's probably a better approach. Might be a bit tricky to unhighlight cells on blur and re-highlight on focusing, will have it as a follow up

PR #4690, opened 2023-06-22
Swap noopener for noreferrer

acconrad:
The use of `noreferrer` [supersedes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) `noopener` and it [implies](https://github.com/jsx-eslint/eslint-plugin-react/issues/2022) its use. It is safer, more widely-supported, and expands the range of supported browsers from XSS `target=_blank` [vulnerabilities](https://security.stackexchange.com/questions/216135/xss-with-a-tag-with-target-blank). `noopener` should therefore be deprecated in favor of `noreferrer` whenever `target=_blank` and `rel=noopener` is used. Instances of `rel="noopener noreferrer"` are safe (at worst, redundant) and were not included in this PR.


zurfyx (approved):
LGTM, thank you!

Issue #4697, opened 2023-06-23
Bug: New table selection breaks when table cells are already selected


PeterBul:
Trying to select table cells when cells are already selected leads to a range selection instead of a grid selection. In Chrome, this leads to a range selection that has the correct anchor and focus nodes. In Firefox the range selection is completely wrong, with the focus node being the anchor node and the selection going in the wrong direction.

Lexical version: 0.11.1

## Steps To Reproduce

1. Add table with 4 rows and 4 columns
2. Select the 4 center cells.
3. Try to select the 2 cells to the right in the group of already selected cells (col 2, row 1-2 (0 indexed))
4. This should result in a range selection going through the cells in Chrome and a different kind of edge selection in Firefox. The visual selection in Firefox does not match the underlying selection held by Lexical. (This can be inspected with the debugger in Lexical Playground), Lexical has the selection (col 2-3, row 2) while the visual selection is (col 2, row 1-2).

https://github.com/facebook/lexical/assets/23295934/68a8f9d8-cc44-4d6a-b2e9-800066a45b2b

Here, the cells marked with blue are the ones that are actually selected in Lexical
![table-selection-firefox](https://github.com/facebook/lexical/assets/23295934/069ef375-d6f5-46c5-bf62-971f8f387c73)

https://github.com/facebook/lexical/assets/23295934/d55e26a1-bbcb-4fc3-8d5d-9b87beb4805b

Link to code example:
https://codesandbox.io/s/lexical-table-plugin-example-with-experimental-forked-trcmsx

## The current behavior
Selecting cells in a table with cells already selected results in a range selection. This selection has the right anchor and focus nodes in Chrome but not in Firefox.

## The expected behavior
Selecting cells in a table with cells already selected should result in a new grid selection.

ivailop7:
This should work correctly now. @PeterBul can you verify it works correctly for you now?

PR #4699, opened 2023-06-23
Allow table selection when cells are already selected

PeterBul:
Earlier on table selection,  previous selection was cleared before immediately returning on mousedown if table cells were already selected. By not returning, but still clearing selection, we are able to make a proper new table selection instead of ending up in bad states.

Fixes issue #4697

See issue for more details.

Before:

https://github.com/facebook/lexical/assets/23295934/d9eec6bc-6829-4dca-a4d8-023acf3ed8fa

https://github.com/facebook/lexical/assets/23295934/e101c75e-6f2c-4369-84a7-3cda2bc72817

After:

https://github.com/facebook/lexical/assets/23295934/3d1832a2-d105-422c-b005-107c3afcc4bf

https://github.com/facebook/lexical/assets/23295934/c39419ed-08d6-4cb1-97d2-04655ab0c4f8

### DIFF
--- a/packages/lexical-playground/__tests__/regression/4697-repeated-table-selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/regression/4697-repeated-table-selection.spec.mjs
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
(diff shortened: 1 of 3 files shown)


fantactuka:
Started some rework on table selection in #4689 which seem to address your case, but regardless, having more tests is great. Thank you!

fantactuka (approved):

PR #4700, opened 2023-06-23
Fix empty editor focus

GermanJablo:
Currently editor.focus did not work on empty editors.
To use AutofocusPlugin it was necessary to initialize the editor with even one paragraph.
Correct me if I'm wrong, but it looks like a bug, right?


acywatson:
I guess you aren't the first person to try this:

https://github.com/facebook/lexical/pull/3254

PR #4682, opened 2023-06-20
Preserving language and indentation of code block when serialized/deserialized

ebads67:
- The language and indentation of code block is preserved when serialized/deserialized

PR #4702, opened 2023-06-23
Fix code block export

### LINKED ISSUE #4677: Bug: code node not parsed correctly when converting from html to lexical
I noticed sometimes, code block is not parsed correctly when converting html back to lexical state. See the screen recording for reproduction. It seems that in this case, the code block was parsed into a textNode instead of the codeNode. 

https://github.com/facebook/lexical/assets/1328543/bf904f67-5338-4a54-9f65-4da0d5a975e9

Lexical version: 0.11.1

## Steps To Reproduce
see above screen recording 

Link to code example:

```
export function RichTextEditor({
    initialEditorStateHtml,
}: EditorProps) {
    const editorConfig = {
        ...DEFAULT_EDITOR_CONFIG,
        theme: editorTheme,
        editable: !readonly,
        editorState: (editor: LexicalEditor) => {
            const dom = parser.parseFromString(
                initialEditorStateHtml ?? '',
                'text/html',
            );
            const rootNode = $getRoot();
            for (const node of nodes) {
                try {
                    rootNode.append(node);
                } catch (err) {
                    continue;
                }
            }
        },
    };
    return (
         ....
        </LexicalComposer>
    );
}

```

## The current behavior
Code block is parsed into textNode
## The expected behavior
Code block should be parsed into codeNode and render as code block correctly.

fantactuka:
We should export multi-line code as pre, same as other editors do

Fixes #4677

### DIFF
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -129,7 +129,7 @@ export class CodeNode extends ElementNode {
   }
 
   exportDOM(): DOMExportOutput {
-    const element = document.createElement('code');
+    const element = document.createElement('pre');
     element.setAttribute('spellcheck', 'false');
     const language = this.getLanguage();
     if (language) {
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -140,16 +140,12 @@ describe('Markdown', () => {
       skipExport: true,
     },
     {
-      // Import only: export will have extra <pre> wrapper
-      html: '<code spellcheck="false"><span>Code</span></code>',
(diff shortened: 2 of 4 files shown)


thegreatercurve (approved):

ivailop7:
very cool, we were thinking of suggesting this change. Somewhat unrelated, but curious, would this have any implications on the markdown plugin?

zurfyx (approved):
I like that, can we revert the line I pointed out here please (Textnode -> `isNodePre` -> `node.nodeName === 'CODE' ||`)? https://github.com/facebook/lexical/pull/4682

ivailop7:
> I like that, can we revert the line I pointed out here please (Textnode -> `isNodePre` -> `node.nodeName === 'CODE' ||`)? #4682

Yes, we can delete it now. I'll do a quick PR.

fantactuka:
NP, I'll add to this one

ivailop7 (approved):

Issue #4708, opened 2023-06-26
conversion div as ParagraphNode


yaojiu19:
importDOM for ParagraphNode only return p tag
![image](https://github.com/facebook/lexical/assets/18085334/d2825db6-bac7-40d4-b8f0-fd4da8540079)

```
<div>1</div>
<div>2</div>
```
use $generateNodesFromDOM(), will return [TextNode, TextNode]

can you add conversion div as ParagraphNode,  like:
```
importDOM() {
  return {
    p: () => ({
      conversion: convertParagraphElement,
      priority: 0,
    }),
    div: () => ({
      conversion: convertParagraphElement,
      priority: 0,
    }),
  }
}
```
use $generateNodesFromDOM(), will return [ParagraphNode, ParagraphNode]

PR #4709, opened 2023-06-26
conversion div as ParagraphNode

yaojiu19:
feat by #4708

### DIFF
--- a/packages/lexical/src/nodes/LexicalParagraphNode.ts
+++ b/packages/lexical/src/nodes/LexicalParagraphNode.ts
@@ -60,6 +60,10 @@ export class ParagraphNode extends ElementNode {
         conversion: convertParagraphElement,
         priority: 0,
       }),
+      div: (node: Node) => ({
+        conversion: convertParagraphElement,
+        priority: 0,
+      }),
     };
   }
 


ivailop7:
This can't be merged, this will break any other div-based custom node, since this doesn't have any condition when to failover the div tag parsing to other nodes. The reason this doesn't break things for you is probably the lack of custom nodes in your editor instance. My advice is to override the text node with the behaviour you are after in this case. An example on how to override the text node is available here: https://lexical.dev/docs/concepts/serialization#handling-extended-html-styling

yaojiu19:
> 这无法合并，这将破坏任何其他基于 div 的自定义节点，因为这没有任何条件何时将 div 标签解析故障转移到其他节点。这不会对您造成破坏的原因可能是您的编辑器实例中缺少自定义节点。我的建议是用您在这种情况下所追求的行为覆盖文本节点。有关如何覆盖文本节点的示例是可用的这里：[https://lexical.dev/docs/concepts/serialization#handling-extend-html-styling](https://lexical.dev/docs/concepts/serialization#handling-extended-html-styling)

maybe can higher priority of failover the div tag parsing to other nodes

yaojiu19:
> This can't be merged, this will break any other div-based custom node, since this doesn't have any condition when to failover the div tag parsing to other nodes. The reason this doesn't break things for you is probably the lack of custom nodes in your editor instance. My advice is to override the text node with the behaviour you are after in this case. An example on how to override the text node is available(可用) here: https://lexical.dev/docs/concepts/serialization#handling-extended-html-styling

maybe can higher priority of failover the div tag parsing to other nodes

zurfyx:
Thank you for the PR, closing it for now, unfortunately (as @ivailop7 mentioned) divs don't tend to follow a predictable structure, unlike `p` or `pre`. They can be nested and wrapped multiple times and it is valid HTML to wrap other elements such as `p` (for which we already have specific logic).

While this is a known bug, the solution would have to understand the parent nodes to determine when a div can be converted to a block.

PR #4710, opened 2023-06-26
Improve md error message

acywatson:
Also simplifies the hasNode(s) API. I don't think using Array.prototype.every should be a problem here, but willing to listen to opposing viewpoints.

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -862,7 +862,7 @@ export class LexicalEditor {
    * @returns True if the editor has registered all of the provided node types, false otherwise.
    */
   hasNodes<T extends Klass<LexicalNode>>(nodes: Array<T>): boolean {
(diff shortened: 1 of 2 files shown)


fantactuka (approved):

packages/lexical/src/LexicalEditor.ts
```diff
@@ -847,22 +847,22 @@ export class LexicalEditor {
...
-      const klass = nodes[i];
-      const type = klass.getType();
-
-      if (!this._nodes.has(type)) {
-        return false;
-      }
-    }
-
-    return true;
+    return nodes.every(this.hasNode.bind(this));
```
zurfyx, line 865: Mind expanding on why an arrow function instead of the previous imperative approach since we usually prefer the later?

acywatson: Where do you see an arrow function?

PR #4713, opened 2023-06-27
Context Menu Options Added

Shubhankerism:
This PR adds cut, copy, paste, paste as plain text and delete node options to the context menu.

Demo:
context-menu

### DIFF
--- a/packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
@@ -6,13 +6,20 @@
  *
  */
 
-import type {LexicalNode} from 'lexical';
-
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {
   LexicalContextMenuPlugin,
   MenuOption,
 } from '@lexical/react/LexicalContextMenuPlugin';
+import {$findMatchingParent} from '@lexical/utils';
+import {
+  type LexicalNode,
+  $getSelection,
+  $isRangeSelection,
+  COPY_COMMAND,
+  CUT_COMMAND,
+  PASTE_COMMAND,
+} from 'lexical';
 import {useCallback, useMemo} from 'react';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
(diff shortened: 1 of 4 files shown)


packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
```diff
@@ -99,14 +106,85 @@ export default function ContextMenuPlugin(): JSX.Element {
...
+          const selection = $getSelection();
+          if ($isRangeSelection(selection)) {
+            const currentNode = selection.anchor.getNode();
+
+            const parentNode = $findMatchingParent(currentNode, (node) => {
+              return node.__parent === 'root';
+            });
+            parentNode?.remove();
+          }
+        },
```
zurfyx, line 184: Will this work for ImageNode which is an inline node? I presume that in this case it will still delete the entire paragraph?

packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
```diff
@@ -99,14 +106,85 @@ export default function ContextMenuPlugin(): JSX.Element {
...
+      }),
+      new ContextMenuOption(`Delete Node`, {
+        onSelect: (_node) => {
+          const selection = $getSelection();
+          if ($isRangeSelection(selection)) {
+            const currentNode = selection.anchor.getNode();
+
+            const parentNode = $findMatchingParent(currentNode, (node) => {
+              return node.__parent === 'root';
```
zurfyx, line 180: Nit; I would try to avoid the use of private properties beyond the core, so I'd probably use a combination of `getParent` and `$isRootNode` in this case. It's not the most performant but it's the recommended consumer-wise

packages/lexical-utils/src/index.ts
```diff
@@ -524,5 +524,7 @@ export function objectKlassEquals<T>(
   object: unknown,
   objectClass: ObjectKlass<T>,
 ): boolean {
-  return Object.getPrototypeOf(object).constructor.name === objectClass.name;
+  return object
```
zurfyx, line 527: I presume you're checking null values here. Can we use triple equality?
 
```suggestion
  return typeof object === 'object'
```

packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
```diff
@@ -99,14 +106,85 @@ export default function ContextMenuPlugin(): JSX.Element {
...
+            });
+            if (permission.state === 'denied') {
+              alert('Not allowed to paste from clipboard.');
+              return;
+            }
+
+            for (const type of item.types) {
+              const dataString = await (await item.getType(type)).text();
+              data.setData(type, dataString);
+            }
```
zurfyx, line 139: Does this cover images? Can also be done in a follow-up PR

zurfyx (commented):
Looking great, thanks for working on this! Adding some minor comments

packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
Shubhankerism: Hi, thanks for the review! :) 
Yes, it will delete the entire paragraph, the current logic removes the node that has root as its parent.

packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
Shubhankerism: I have updated the logic to now use `getParents()`.

packages/lexical-utils/src/index.ts
Shubhankerism: Using `object !== null` as `typeof null` returned `object`.

ivailop7 (approved):
LGTM

ivailop7:
Waiting on @zurfyx to approve and merge.

PR #4717, opened 2023-06-28
Fix: Use $createListItemNode within importJSON to override on deserialization

kazukinagata:
This PR fixes an issue with node deserialization. Currently, when importing from JSON, the `importJSON` method directly returns an instance of ListItemNode instead of an overridden one. I believe that just like other nodes, ListItemNode should also go through `$applyNodeReplacement`.

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -124,9 +124,7 @@ export class ListItemNode extends ElementNode {
   }
 
   static importJSON(serializedNode: SerializedListItemNode): ListItemNode {
-    const node = $applyNodeReplacement<ListItemNode>(
-      new ListItemNode(serializedNode.value, serializedNode.checked),
-    );
+    const node = new ListItemNode(serializedNode.value, serializedNode.checked);
     node.setFormat(serializedNode.format);
     node.setDirection(serializedNode.direction);
     return node;
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -124,7 +124,9 @@ export class ListItemNode extends ElementNode {
   }
 
   static importJSON(serializedNode: SerializedListItemNode): ListItemNode {
-    const node = new ListItemNode(serializedNode.value, serializedNode.checked);
+    const node = $createListItemNode();
+    node.setChecked(serializedNode.checked);
+    node.setValue(serializedNode.value);
     node.setFormat(serializedNode.format);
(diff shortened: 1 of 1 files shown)


acywatson (commented):
You should just implement importJSON and exportJSON on the node override class.

kazukinagata:
@acywatson 
In response to your feedback, I'd like to point out that all nodes, excluding ExcalidrawNode and StickyNode, go through `$applyNodeReplacement`. The need to implement importJSON depending on what is being overridden could potentially confuse users. I believe maintaining a consistent method across nodes could help avoid this confusion.

acywatson:
> @acywatson
> In response to your feedback, I'd like to point out that all nodes, excluding ExcalidrawNode and StickyNode, go through `$applyNodeReplacement`. The need to implement importJSON depending on what is being overridden could potentially confuse users. I believe maintaining a consistent method across nodes could help avoid this confusion.

I’m not sure I’m following your concern, but you are correct that those other nodes are using applyNodeReplacement but not directly. I’m not sure this is an intentional pattern and it’s probably wrong everywhere, but for now I’d accept a PR that uses $createListItemNode, but I don’t think we should use applyNodeReplcement directly here

kazukinagata:
@acywatson 
Yes, I agree that using $createListItemNode would offer better consistency in theory. However, looking at the implementation of $createListItemNode, I see that while we can pass 'checked' as an argument, we cannot pass 'value'.

https://github.com/facebook/lexical/blob/main/packages/lexical-list/src/LexicalListItemNode.ts#L535

Modifying the interface of $createListItemNode would affect a broad scope of the project, which led me to consider using $applyNodeReplacement directly as a more appropriate solution in this case.

acywatson:
> @acywatson Yes, I agree that using $createListItemNode would offer better consistency in theory. However, looking at the implementation of $createListItemNode, I see that while we can pass 'checked' as an argument, we cannot pass 'value'.
> 
> https://github.com/facebook/lexical/blob/main/packages/lexical-list/src/LexicalListItemNode.ts#L535
> 
> Modifying the interface of $createListItemNode would affect a broad scope of the project, which led me to consider using $applyNodeReplacement directly as a more appropriate solution in this case.

Yea but why not create the node and use the setter methods like we do elsewhere?

kazukinagata:
Ah, indeed you're right! 😄  I'd like to proceed with your suggestion. Should I create a new commit on this PR, or would it be better to open a new PR?

acywatson:
> 

You can commit to this PR - I’ll reopen

kazukinagata:
I've finished making the changes. Could you please check it?

acywatson (approved):

PR #4721, opened 2023-06-29
Keeping local comments and sharedCommentsArray orders in sync

ebads67:
The order of [CommentStore local comments](https://github.com/facebook/lexical/blob/1ce51a8b6e4bec14601829f5503022015dc05403/packages/lexical-playground/src/commenting/index.ts#L106C3-L106C12) and the [shared comments array](https://github.com/facebook/lexical/blob/1ce51a8b6e4bec14601829f5503022015dc05403/packages/lexical-playground/src/commenting/index.ts#L250) must be exactly the same. Otherwise, the order of comments may be different for different users, and most importantly the [delete operation](https://github.com/facebook/lexical/blob/1ce51a8b6e4bec14601829f5503022015dc05403/packages/lexical-playground/src/commenting/index.ts#L187C11-L187C24) depends on it.

If the delta contains more than one insert (happens during the initial load of comments) the order of the comments is reversed. This PR fixes this issue by applying the inserts to the local comments in the reverse order.

This is a one line change. Turn on the ignore-whitespaces feature to review.

commentsWrongOrder

### DIFF
--- a/packages/lexical-playground/src/commenting/index.ts
+++ b/packages/lexical-playground/src/commenting/index.ts
@@ -351,50 +351,53 @@ export class CommentStore {
 
               if (Array.isArray(insert)) {
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                insert.forEach((map: YMap<any>) => {
-                  const id = map.get('id');
-                  const type = map.get('type');
+                insert
+                  .slice()
+                  .reverse()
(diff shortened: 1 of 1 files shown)


fantactuka (approved):
Good catch, thank you!

fantactuka (approved):
Good catch, thank you!

PR #4727, opened 2023-06-30
Fixed drop down buttons to prevent sumbit

Ou7law007:


fantactuka (approved):
Thanks!

fantactuka:
You'll need to change single to double quotes to pass linter

PR #4731, opened 2023-07-01
Show excalidraw edit button

IBastawisi:
The current way to edit the Excalidraw node is to double click the image which is not super accessible.
This pull request adds a button to open the modal if the node is selected.


packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -161,6 +161,10 @@ export default function ExcalidrawComponent({
     }, 200);
   };
 
+  const openModal = useCallback(() => {
```
zurfyx, line 164: Nit; we might not need `useCallback` in this case as it's just an event handler

zurfyx (approved):
LGTM, thank you!

PR #4735, opened 2023-07-03
Set data-gutter on exported code element

IBastawisi:
The exported code element is missing line numbers, This PR adds the `data-gutter` attribute to the `exportDOM()` method of `CodeNode` to fix this issue.

### DIFF
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -12,6 +12,7 @@ import type {
   DOMConversionOutput,
   DOMExportOutput,
   EditorConfig,
+  LexicalEditor,
   LexicalNode,
   NodeKey,
   ParagraphNode,
@@ -45,6 +46,7 @@ import {
   ElementNode,
   $isTabNode,
   $createTabNode,
+  $isLineBreakNode,
 } from 'lexical';
 import {
   $isCodeHighlightNode,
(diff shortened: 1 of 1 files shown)


fantactuka:
This data attribute is only used internally to render line numbers, what's the reason for it to be exported?

IBastawisi:
> This data attribute is only used internally to render line numbers, what's the reason for it to be exported?

The line numbers are displayed via `content: attr(data-gutter);` style declaration, if the attribute is missing no content is displayed.

image

fantactuka:
In this case it should be fixed in plugin to make it generate gutter data for newly created code node

IBastawisi:
> In this case it should be fixed in plugin to make it generate gutter data for newly created code node

everything works as expected, but the exported html element doesn't have the data-gutter attribute.
It seems obvious to add it to the `exportDOM()` method.

acywatson:
> > In this case it should be fixed in plugin to make it generate gutter data for newly created code node
> 
> everything works as expected, but the exported html element doesn't have the data-gutter attribute. It seems obvious to add it to the `exportDOM()` method.

As Maks said, this property is not part of the public API - on paste, these attributes can be regenerated - they don't need to be transported as part of the paste payload. They'd mean nothing to any other editor because they're an implementation detail.

2wheeh:
@ibastawisi Hi. I'm on same issue and found this PR. How did you make it use `data-gutter` on html when exporting without this PR?

PR #4743, opened 2023-07-05
[RFC] Use css counter for line numbers

fantactuka:
Alternative approach for line numbers. Current implementation adds `data-gutter` attribute that includes line numbers that are rendered via `code_block:before` content. It has issues with line-wraps as data-gutter does not calculate amount of visible lines taken by one line of code. On top of it, it uses caching logic to throttle gutter recalculation and its value might become invalid.

This approach uses [css counters](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_counter_styles/Using_CSS_counters). Spans or line breaks that are preceded by a line break are line starters, so we could use those to count lines and style their `:before` to include counter content. It requires zero javascript and its only downside is it can't render line numbers for empty lines because br:before is not accepting much of styles:

Screenshot 2023-07-04 at 8 27 20 PM


IBastawisi:
incredible! btw you can display content before the `<br>` tag; this worked for me 😊
```css
.PlaygroundTheme__code>br+br {
  display: block;
  content: counter(line-number);
  height: 20px;
}

.PlaygroundTheme__code>br+br:before {
  color: #777;
  content: counter(line-number);
  user-select: none;
  position: absolute;
  left: 5px;
  width: 50px;
}
```

HarshithMjh:
Hi @fantactuka , I too needed this fix. Since you closed this, Are you planning on any alternate fix for code block line numbers issue?

thorn0:
@ibastawisi That doesn't work in Firefox and Safari, unfortunately. Also in Chrome the caret starts behaving funny around empty lines. If you move it with the left/right arrow keys, it needs two key presses instead of one.

2wheeh:
For me, I just needed this to style exported html without `data-gutter`. It helped me a lot thanks !

ibastawisi:
> For me, I just needed this to style exported html without `data-gutter`. It helped me a lot thanks !

It's quite buggy as @thorn0 said, though I'm still using it on my project. But I plan to set `data-gutter` to the exported markup in  my custom code node.

PR #4748, opened 2023-07-05
Added function to generate html from root

rmindo:
Added a function that really helpful when working with the root like modifying children, adding and removing and that needs to convert to HTML for other purposes before updating it to the editor.


acywatson:
Isn't this what the $generateHtmlFromNodes does if the second argument (selection) is undefined?

rmindo:
> Isn't this what the $generateHtmlFromNodes does if the second argument (selection) is undefined?

Ah, yeah it is. Now that this feature actually exists, the docs should have examples showing how this work.

ivailop7:
Are we good to close this one? @acywatson @rmindo

PR #4755, opened 2023-07-07
Allow nodesOfType to accept a list of types

acywatson:


zurfyx (approved):

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1076,16 +1076,19 @@ export function setMutatedNode(
...
+  const klassList = Array.isArray(klass) ? klass : [klass];
+  const klassTypes = new Set(klassList.map((k) => k.getType()));
   const nodes = editorState._nodeMap;
   const nodesOfType: Array<T> = [];
   for (const [, node] of nodes) {
     if (
-      node instanceof klass &&
-      node.__type === klassType &&
+      klassList.some((k) => node instanceof k) &&
+      klassTypes.has(node.__type) &&
```
thorn0, line 1091: Why not `klassList.some((k) => node instanceof k && node.__type === k.getType())`?
What's the point of the `klassTypes` set if `some` iterates the list anyway?

acywatson: Yea good point, My original plan was to get rid of the instanceof check, so adding this back in was an afterthought. I need to think more about this.

acywatson:
This significantly slows down the base case (a single node type) in my testing and it's easy enough to do direct iteration through the node map in rare cases where we need to deal with more than one node type., so I'm closing this for now. I also tested just removing the instanceof check itself, and there's no meaningful difference in execution speed (for relatively short prototype chains, anyway).

```
function testPerf(editor: LexicalEditor): void {
  editor.update(() => {
    for (let i=0; i<1000; i++) {
      $createHeadingNode('h1');
      $createParagraphNode().append($createTextNode('Hello World'));
    }
    const start = performance.now();
    for (let i=0; i<10000; i++) {
      $nodesOfType(TextNode);
    }
    const end = performance.now();
    console.log(end - start);
  })
}
```

PR #4756, opened 2023-07-07
Fix $transferStartingElementPointToTextPoint()

amyworrall:
Fix $transferStartingElementPointToTextPoint()


amyworrall:
Previously there was an issue if both start and end points were of type .element and referred to the same element. When the new text node was inserted, the offset of the end point of the selection was not modified.

zurfyx (approved):

zurfyx:
Let's remove package-lock.json and should be good to go. We don't usually want to include this unless we're intentionally upgrading packages

Issue #4668, opened 2023-06-16
Bug: Pasting position is wrong if a line ends with DecoratorNode


imdreamrunner:
Lexical version: v0.11.1

## Steps To Reproduce

1. Insert an inline image in the middle of a paragraph
2. Select the image, copy it. And paste it in the end of the same paragraph
3. Keep the cursor at the end of the paragraph, paste the same image again
    - Expectation: another image paste at the end of the paragraph
    - Actual: another image paste before the first inline image
4. We can repeat step 3 with text in the clipboard. The text will be pasted before the first inline image as well.

https://github.com/facebook/lexical/assets/2252473/131b95b5-027f-423c-a6d8-f8d56c5a5463

It seems that the insert position of the clipboard content will be wrong if there's an inline DecoratorNode at the end of a paragraph.

Link to code example: NA

## The current behavior

See steps above

## The expected behavior

See steps above

Infinity8sailor:
Hi @zurfyx , I was working on this issue and found these insights,
/---
![image](https://github.com/facebook/lexical/assets/43951027/4f063797-587c-45ce-b9fd-6e5052ab4e07)
target with DecoratorNode is entering into the else block in selection.InsertNode [ LexicalSelection.ts], where nexttarget is is being set to its parent thats why its pasting any content at the end of first node of selection paragraph.
/---
![image](https://github.com/facebook/lexical/assets/43951027/53a76143-91cb-4ae0-a3a2-4448ec0d057e)
Also, Decorator node is set to return isInline always true so this 1678 actually blocking it where it can be sorted. 
/---
![image](https://github.com/facebook/lexical/assets/43951027/df585a7c-2a26-4ea7-8796-95bd118ac1a8)
Removing line 1540 can solve this issue as well, I can understand there might be different reasons why its there if you guys can give some inputs, I would like to build the logic for same.
/---
There are couple of ways we can sort this and would like to work on this if some insights I can get from you guys.
Thank you.

etrepum:
I can't reproduce this, sounds like it was fixed by the insertNodes rewrite some time ago https://github.com/facebook/lexical/pull/4758#issuecomment-1900593550

PR #4758, opened 2023-07-09
Fix: Pasting position is wrong if a line ends with DecoratorNode

Infinity8sailor:
#4668
Removed the line checking for target should not be decoratorNode. DecoratorNode is inline so adding textNode or another decoratorNode should not be a problem.
### Current Behavior
https://github.com/facebook/lexical/assets/43951027/214b0b8f-b554-4182-bf24-357692c5a06f
### With Updated code
https://github.com/facebook/lexical/assets/43951027/ada6337d-256b-4c85-a321-4ea319184dda

Current code code changes fixes the bug. I can understand that line could be for some cases, I would like to know the same.
Thank you.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1542,7 +1542,6 @@ export class RangeSelection implements BaseSelection {
       const node = nodes[i];
       if (
         !$isRootOrShadowRoot(target) &&
-        !$isDecoratorNode(target) &&
         $isElementNode(node) &&
         !node.isInline()
       ) {


Infinity8sailor:
Any Update on this or comments.

acywatson:
> Author

It fails integrity checks and it's breaking 2 end to end tests. The logic in insert nodes is very crufty and most of it has been built up to address specific edge cases over time. Highly unlikely that you can just delete part of a condition and not break something.

777PolarFox777:
We're using emojis as decorators to render them as images and faced the same issue. Are there any chances it will be merged?

zurfyx:
Closing as obsolete, `insertNodes` was rewritten from scratch.

PR #4762, opened 2023-07-10
Reset editor state when setting null root

fantactuka:
We missed resetting editor state back to its original (or pending value) when unset root element (by either unmounting contenteditable or directly calling `setRootElement(null)`). Which caused editor with unset root element to always return empty editor state


ivailop7 (approved):
I guess this is with regards to the splice error for collab node?

PR #4763, opened 2023-07-10
Fixes broken trunk

fantactuka:
Trunk seems broken, this one fixes it with some adjustments to #4756

### DIFF
--- a/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
+++ b/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
@@ -14,9 +14,9 @@ import {
   LexicalEditor,
   RangeSelection,
 } from 'lexical';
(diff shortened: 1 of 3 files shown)


packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -14,9 +14,9 @@ import {
   LexicalEditor,
   RangeSelection,
 } from 'lexical';
+import {$createRangeSelection} from 'lexical/src';
```
fantactuka, line 17: This file didn't change, only formatting changes to pass linter

zurfyx: True but this import looks bad, `lexical` doesn't work?

zurfyx (approved):

PR #4769, opened 2023-07-11
package.lock

zurfyx:
I always miss the same command 🤦‍♂️ We should make it into the publish script


ivailop7 (approved):

acywatson:
> I always miss `npm install` 🤦‍♂️ We should make it into the publish script

lol or just read the runbook instead of trying to memorize the steps :)

PR #4770, opened 2023-07-12
Stop collapsing non-collapsible whitespace on paste

### LINKED ISSUE #4564: Bug: Lexical collapses non-collapsible whitespace on paste [regression]
(Originally posted as a question at https://github.com/facebook/lexical/pull/4467#issuecomment-1560524576)

Lexical version: 0.11.0 and 0.11.1

## Steps To Reproduce

1. Open the lexical playground in one browser (e.g. Firefox)
2. Type "A⍽⍽⍽B" where ⍽ is a non-collapsible space such as full-width space "　" (U+3000) or a non-breaking space (U+00A0).

   (A full-width space is easy to type using, e.g. a Japanese IME. For a non-breaking space, on Windows I can it into notepad using <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Space</kbd> and then paste into Lexical Playground).
   
3. Select the "A⍽⍽⍽B" text and copy it to the clipboard
4. Open the lexical playground in another browser (e.g. Chrome)
5. Paste it

Assuming GitHub doesn't normalize whitespace here, an even simpler STR is:

1. Copy this string here: "A　　　B"
2. Paste into Lexical Playground

## The current behavior

"A⍽⍽⍽B" gets converted into "A B" (i.e. a single U+0020 separating A and B)

## The expected behavior

"A⍽⍽⍽B" gets pasted as-is

## Notes

(For the first STR you need to copy and paste between different browsers so that we follow the "paste HTML" codepath, not the "paste Lexical" one.)

Appears to be a regression from #4467.

The spec mentioned in that PR doesn't appear to allow collapsing full-width or non-breaking space characters. Specifically it has:

> Except where specified otherwise, white space processing in CSS affects only the document white space characters: spaces (U+0020), tabs (U+0009), and [segment breaks](https://www.w3.org/TR/css-text-3/#white-space-processing).

I couldn't find anywhere that allows collapsing other whitespace characters..

Looking at the patch, it seems to be using the `\s` character class in a number of places, e.g.

```js
  textContent = textContent
    .replace(/\r?\n|\t/gm, ' ')
    .replace('\r', '')
    .replace(/\s+/g, ' ');
```

However, `\s` expands to `[\f\n\r\t\v\u0020\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]` whereas I think we are only interested in matching `[\t\u0020]` here?

birtles:
Fixes #4564.

### DIFF
--- a/packages/lexical/src/nodes/LexicalLineBreakNode.ts
+++ b/packages/lexical/src/nodes/LexicalLineBreakNode.ts
@@ -58,13 +58,13 @@ export class LineBreakNode extends LexicalNode {
             ((firstChild as Text).nextSibling === node &&
               (firstChild as Text).nodeType === DOM_TEXT_TYPE &&
               ((firstChild as Text).textContent || '').match(
-                /^[\s|\r?\n|\t]+$/,
+                /^( |\t|\r?\n)+$/,
               ) !== null)) &&
           ((lastChild = parentElement.lastChild) === node ||
             ((lastChild as Text).previousSibling === node &&
               (lastChild as Text).nodeType === DOM_TEXT_TYPE &&
               ((lastChild as Text).textContent || '').match(
-                /^[\s|\r?\n|\t]+$/,
+                /^( |\t|\r?\n)+$/,
               ) !== null))
         ) {
           return null;
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -1177,10 +1177,7 @@ function convertTextDOMNode(domNode: Node): DOMConversionOutput {
     }
     return {node: nodes};
   }
-  textContent = textContent
-    .replace(/\r?\n|\t/gm, ' ')
-    .replace('\r', '')
-    .replace(/\s+/g, ' ');
+  textContent = textContent.replace(/\r/g, '').replace(/[ \t\n]+/g, ' ');
   if (textContent === '') {
     return {node: null};
   }
@@ -1196,7 +1193,7 @@ function convertTextDOMNode(domNode: Node): DOMConversionOutput {
     ) {
       const previousTextContent = previousText.textContent || '';
       if (previousTextContent.length > 0) {
-        if (previousTextContent.match(/(?:\s|\r?\n|\t)$/)) {
+        if (/[ \t\n]$/.test(previousTextContent)) {
           textContent = textContent.slice(1);
         }
         isStartOfLine = false;
@@ -1216,7 +1213,7 @@ function convertTextDOMNode(domNode: Node): DOMConversionOutput {
       (nextText = findTextInLine(nextText, true)) !== null
     ) {
       const nextTextContent = (nextText.textContent || '').replace(
-        /^[\s|\r?\n|\t]+/,
+        /^( |\t|\r?\n)+/,
         '',
       );
       if (nextTextContent.length > 0) {
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/LinksHTMLCopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/LinksHTMLCopyAndPaste.spec.mjs
@@ -201,7 +201,7 @@ test.describe('HTML Links CopyAndPaste', () => {
             class="PlaygroundEditorTheme__listItem PlaygroundEditorTheme__ltr"
             dir="ltr"
             value="1">
-            <span data-lexical-text="true">â..ï¸.Â Line 1Â</span>
+            <span data-lexical-text="true">â..ï¸.Â&nbsp;Line 1Â&nbsp;</span>
             <a
               class="PlaygroundEditorTheme__link PlaygroundEditorTheme__ltr"
               dir="ltr"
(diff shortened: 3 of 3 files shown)


packages/lexical/src/nodes/LexicalLineBreakNode.ts
```diff
@@ -58,13 +58,13 @@ export class LineBreakNode extends LexicalNode {
             ((firstChild as Text).nextSibling === node &&
               (firstChild as Text).nodeType === DOM_TEXT_TYPE &&
               ((firstChild as Text).textContent || '').match(
-                /^[\s|\r?\n|\t]+$/,
```
thorn0, line 61: Not directly related to this PR, but this regexp doesn't look right. `|` doesn't work as a separator inside character classes (`[...]`), it's interpreted literally there.

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -1180,7 +1180,7 @@ function convertTextDOMNode(domNode: Node): DOMConversionOutput {
   textContent = textContent
     .replace(/\r?\n|\t/gm, ' ')
     .replace('\r', '')
-    .replace(/\s+/g, ' ');
```
thorn0, line 1183: Changes like this might introduce bugs because `\s` stands for many other whitespace characters, not only space.
Specifically, it's equivalent to `[\f\n\r\t\v\u0020\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]` (source: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes)).
These other characters (in particular, `U+2028` and `U+2029`) are known to be used by various text editors for whatever reasons, so content copied from those editors sometimes contains them. That's why replacing `\s` with just a space isn't a good idea.

birtles: The fact that `\s` expands to many (non-collapsible) whitespace characters is precisely the regression this PR fixes. That is, we are trying to preserve these non-collapsible whitespace characters. Please see the description in #4564.

thorn0: I'm worried specifically about U+2028 and U+2029 used by some editors. E.g. TextMate generates U+2028 on ctrl+enter. But I've done some tests with those characters and not been able to find any discrepancies in behavior between the playgrounds on lexical.dev and in this PR, so I guess it's okay to proceed with these changes.

birtles: Thanks. Yes, the spec that the regressing change (#4467) sought to implement only allows collapsing `\t`, ` ` (U+0020) and segment breaks (≈newlines) so I think it should be safe to revert the collapsing of all other whitespace characters.

If there are other characters that, contrary to the spec, we need to collapse for compatibility with other editors then I think they should probably be introduced as separate changes with corresponding tests.

packages/lexical/src/nodes/LexicalLineBreakNode.ts
birtles: Thanks for calling that out. That surprised me too. It's probably worth fixing that at the same time.

birtles: Ok I've fixed up these regexes and the ones in `LexicalTextNode.ts` too. Should be ready to review now.

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -1196,7 +1196,7 @@ function convertTextDOMNode(domNode: Node): DOMConversionOutput {
     ) {
       const previousTextContent = previousText.textContent || '';
       if (previousTextContent.length > 0) {
-        if (previousTextContent.match(/(?:\s|\r?\n|\t)$/)) {
+        if (previousTextContent.match(/( |\t|\r?\n)$/)) {
```
birtles, line 1199: I deliberately dropped the non-capturing specifier here since we don't appear to need it.

thorn0: It doesn't matter if `\n` is preceded with `\r`. So it can be rewritten this way:
```suggestion
        if (previousTextContent.test(/[ \t\n]$/)) {
```

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -1196,7 +1196,7 @@ function convertTextDOMNode(domNode: Node): DOMConversionOutput {
     ) {
       const previousTextContent = previousText.textContent || '';
       if (previousTextContent.length > 0) {
-        if (previousTextContent.match(/(?:\s|\r?\n|\t)$/)) {
+        if (previousTextContent.test(/[ \t\n]$/)) {
```
thorn0, line 1199: Doh! It should be
```suggestion
        if (/[ \t\n]$/.test(previousTextContent)) {
```

packages/lexical/src/nodes/LexicalTextNode.ts
thorn0: `\t` can't be found here as it's replaced by the first `replace` call in the chain. I propose to simplify the chain this way:
```js
  textContent = textContent.replace(/\r/g, '').replace(/[ \t\n]+/g, ' ');
```

birtles: Ok, should be fixed now.

thorn0 (commented):
Looks good

acywatson:
@zurfyx  as our resident whitespace aficionado

birtles:
> @zurfyx as our resident whitespace aficionado

@zurfyx review ping?

zurfyx (approved):
Thank you, that makes sense to me! While there's more than the ` ` type of space listed in the W3 to handle, this is more correct than the wildcard `\s`

PR #4772, opened 2023-07-12
Remove duplicate declaration

aMoRRoMa:
I faced it after the latest release. A quick fix that removes duplication.

### DIFF
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -81,8 +81,8 @@ type DecoratorListener = (decorator: {
   [NodeKey]: any,
 }) => void;
 type RootListener = (
-  element: null | HTMLElement,
-  element: null | HTMLElement,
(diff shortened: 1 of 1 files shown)


thorn0:
It really has two parameters. E.g. see https://lexical.dev/docs/concepts/dom-events#1-event-delegation
I think they just should be renamed.

acywatson:
> It really has two parameters. E.g. see https://lexical.dev/docs/concepts/dom-events#1-event-delegation I think they just should be renamed.

Yea, this is right. currentRoot, prevRoot should work.

aMoRRoMa:
> It really has two parameters. E.g. see https://lexical.dev/docs/concepts/dom-events#1-event-delegation I think they just should be renamed.

Oh, I see. So, I'll rename them to make it not confusing.

aMoRRoMa:
@acywatson @thorn0 Please take a look again

Issue #4572, opened 2023-05-29
Bug: Unable to input any CJK language with google input tool


kalvin807:
Lexical version: 0.10.0

## Steps To Reproduce

1. Install Google input tool at google chrome store https://chrome.google.com/webstore/detail/google-input-tools/mclkkofklkfljcocdinagocijmpgbhab
2. Open playground
3. Enable a CJK language (e.g Cantonese)
4. Enter any word (e.g neihou for 你好)
5. Caret flashes but cannot input any word

https://github.com/facebook/lexical/assets/46393179/5c802f68-c28e-484c-81b4-a1645812d2f4

Link to code example:

## The current behavior
It seems like that any input made by using the input suggestion will bugged. I wonder if this may relate to https://github.com/facebook/lexical/pull/3846 because the bug visually look the same

## The expected behavior
The word can be entered properly

zurfyx:
Your code pointer is spot on, it always happens at start of line which is when we have to run additional logic to fix inconsistencies and also create the first TextNode when this doesn't exist.

Unfortunately, I can't think of an easy way to target this event, besides maybe doing `document.getElementByID('GOOGLE_INPUT_CHEXT_FLAG')`

<img width="426" alt="Screenshot 2023-06-01 at 10 03 59 am" src="https://github.com/facebook/lexical/assets/193447/3925321c-9e26-4d52-99d8-b733ece3a08d">

Perhaps we should invert the logic to have `if (IS_ANDROID)` instead but we'd have to test this logic carefully on Windows and Linux not to cause a major regression.

kalvin807:
submitted a fix at https://github.com/facebook/lexical/pull/4783

kalvin807:
@zurfyx bumping on this issue because I can't type with googleIME in all product that uses lexical :(
Please check the PR if this is good enough to patch for now 🙏

dragonman225:
Hi @kalvin807, do you still encounter CJK input problems with Google Input Tools?

It seems to work for me now, at least on Google Chrome 125.0.6422.176 on macOS.

https://github.com/facebook/lexical/assets/25148955/d4101dfc-a92a-4e37-8d9c-e1d19eadc842

mayrang:
PR #4783 (the proposed fix) was closed without merging, and a commenter confirmed this works on Chrome 125 (June 2024). Tested again with Google Input Tools on the latest Chrome — CJK input works correctly.

[...]

PR #4783, opened 2023-07-15
skip node removal if google ime exist

kalvin807:
Fix: https://github.com/facebook/lexical/issues/4572

By skipping the node.removal() if the dom has a element with id `GOOGLE_INPUT_CHEXT_FLAG` which indicate google ime is loaded.

**Questions**
1. Not sure if calling is function every-time will degrade the performance of lexical, but I don't have a good method to detect google ime

As zurfyx said in issue, maybe it is better to have a method to detect android, but I am not very familiar with that. Happy to help if it can push this fix faster. This bug is very annoying and appear in all meta product :(


potatowagon:

Issue #4360, opened 2023-04-19
Feature: keyboard shortcuts to move nodes up, down, indent, unindent


affvalente:
Ctrl+shift up / ctrl+shift down to move node up or down
ctrl+shift left / right to indent / unindent

affvalente:
Any news on this one?

Infinity8sailor:
@zurfyx Can you assign this to me? I would like to work.

PR #4787, opened 2023-07-16
Feature: Move Block up or down with alt key

Infinity8sailor:
Feature: #4360 
I tried working with the short-cuts mentioned in issue but shift was selecting the other blocks as well with up/down key. Since, I use this alt key to move code lines in vs-code, I tried with this alt key to move the block level elements up or down.

In this ( #4360  ) issue another point was to add indent or unindent list elements feature but I think It is already Implementated as tab key.
Demo:
https://github.com/facebook/lexical/assets/43951027/87a7f496-74d1-47b5-b0a3-c7f6d0b49955
Thanks.


fantactuka:
Should this be a plug-in, I'm not sure it's a core feature

Infinity8sailor:
Even I had Created it as a plugin only initially, but it was just a Command and related to blocks only so added it to rich text.

ivailop7:
> Even I had Created it as a plugin only initially, but it was just a Command and related to blocks only so added it to rich text.

I side with @fantactuka here, I imagine this more within the DraggableBlocks plugin rather than the RichText Plugin itself.

Infinity8sailor:
`
[1]   1) [chromium] › packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs:695:3 › CodeBlock › Can move around lines with option+arrow keys 
`

Is ALT + UpArrow command fine for moving blocks? 
There is already a test case for Option+arrow and its failing due to alt key used in this commit. 
Looking into test code I found that Alt + up/down is already implemented in codeBlock for line shifting.
Any suggestions for other key combination for commands?

ivailop7:
> `[1] 1) [chromium] › packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs:695:3 › CodeBlock › Can move around lines with option+arrow keys `
> 
> Is ALT + UpArrow command fine for moving blocks? There is already a test case for Option+arrow and its failing due to alt key used in this commit. Looking into test code I found that Alt + up/down is already implemented in codeBlock for line shifting. Any suggestions for other key combination for commands?

I think using Ctrl (or Cmd)+Shift+Up/Down only after a block is already selected to move it up/down is the consistent behaviour

axhamre:
Was this ever implemented?

PR #4788, opened 2023-07-16
[RFC] Find and replace

ivailop7:
This is a Proof of concept Find & Replace that finds text if there are is no format mismatch, eg. not spread across multiple text nodes. Right now, it highlights the found text, next step is to replace it with a MarkNode that is local to the user searching.

The current logic is not efficient. If someone has an efficient scrolling window algo for the text spread across spans, please reach out. Have to clean up CSS as well.

https://github.com/facebook/lexical/assets/7893468/eff62bd1-c64a-46f8-8e5b-0ea8919b6b22


ivailop7:
Sadly don't have enough time to invest in moving this PR forward, if someone else is more keen on creating a proper one should go ahead.

Issue #3305, opened 2022-11-02
Bug: last character from previous sentence being mixed with current typed character


developaul-k:
Lexical version: 0.5.1-next.1
OS: iOS 15.6.1 - Safari 

https://user-images.githubusercontent.com/25634263/199491639-eae111f3-6df6-4dbc-be0b-f70bf75de029.MP4

## Steps To Reproduce

1. Set up your keyboard language to Korean
2. Write any sentence in Korean, for example: "안녕하세요"
3. Using Backspace delete all the characters you just wrote
4. Type any character, for example: "ㅇ"

## The current behavior
You will see the last character from previous sentence being combined with the new character you just typed -> "용"

## The expected behavior
It should only show a single character : 'ㅇ'

2wheeh:
### Workaround:
make a plugin to apply how #4814 fixes this

```ts
import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext';
import { $getNearestBlockElementAncestorOrThrow } from '@lexical/utils';
import {
  $getNearestNodeFromDOMNode,
  $getSelection,
  $isDecoratorNode,
  $isRangeSelection,
  $isRootNode,
  COMMAND_PRIORITY_CRITICAL,
  KEY_BACKSPACE_COMMAND,
  OUTDENT_CONTENT_COMMAND,
} from 'lexical';
import { useEffect } from 'react';

// https://github.com/facebook/lexical/pull/4814#issuecomment-2037716048
export default function FixIOSKoreanIssuePlugin() {
  const [editor] = useLexicalComposerContext();

  useEffect(
    () =>
      editor.registerCommand<KeyboardEvent>(
        KEY_BACKSPACE_COMMAND,
        (event) => {
          if ($isTargetWithinDecorator(event.target as HTMLElement)) {
            return false;
          }
          const selection = $getSelection();
          if (!$isRangeSelection(selection)) {
            return false;
          }

          const { anchor } = selection;
          const anchorNode = anchor.getNode();

          if (selection.isCollapsed() && anchor.offset === 0 && !$isRootNode(anchorNode)) {
            const element = $getNearestBlockElementAncestorOrThrow(anchorNode);
            if (element.getIndent() > 0) {
              event.preventDefault();
              return editor.dispatchCommand(OUTDENT_CONTENT_COMMAND, undefined);
            }
          }

          // let the browser handle backspace on iOS Korean keyboards,
          // which clears composing and delete the character w/o composition events
[...]

PR #4814, opened 2023-07-24
Fix: Handling Korean character clearing exceptions - issue 3305

DeadIntegral:
https://github.com/facebook/lexical/issues/3305 is the handling of the bug.

This issue only occurs with combination characters, and on iOS, the
backspace in the character "가나다" expects "가낟".

On other devices, a backspace in "가나다" expects "가나ㄷ".

When using lexical on iOS devices, the backspace in "가나다" becomes "가나" and looks like two letters have been removed. However, the actual event in the browser is a single occurrence of keyCode 8. (The other device is in the isComposing state of keyCode 229.) This means that iOS is expecting "가낟" and it doesn't match the lexical, and we think this is causing it.

Since it only happens in Korean, we excluded it with navigator.language and navigator.userAgent to reduce side effects as much as possible.

Before

https://github.com/facebook/lexical/assets/16383168/b1f64eec-143d-4cd8-8465-b5757a9ec169

After

https://github.com/facebook/lexical/assets/16383168/86e80d68-1277-4365-bfe7-cfbb5ef54264


2wheeh:
@acywatson @ivailop7 @zurfyx  May I kindly ask this receive some attention?

2wheeh:
I guess it's not an rich text only problem so it should be handled in `LexicalEvent` or somewhere in the core

DeadIntegral:
> Since It's not a rich text specific problem, I guess this should be handled in `LexicalEvent` or somewhere in the core.

I agree. I wanted to find the source of the problem and fix it, but it was too much work.
I know Lexical handles multibyte characters independently on IOS, and it was hard to track down. If we can fix the root problem, all the better.

Thank you for your interest in this issue!

2wheeh:
> > Since It's not a rich text specific problem, I guess this should be handled in LexicalEvent or somewhere in the core.
I agree. I wanted to find the source of the problem and fix it, but it was too much work.
I know Lexical handles multibyte characters independently on IOS, and it was hard to track down. If we can fix the root problem, all the better.
>
> Thank you for your interest in this issue!

@DeadIntegral 
Let me try to clarify this.

on IOS:
- Korean Keyboard inputs don't emit `CompositionEvents` (any of `compositionstart`, `compositionend`, `compositionupdate`) keeping `isComposing` as `false`. That’s same context as keyCode is not 229.
- Instead, They are handled in `onBeforeInput`  with `deleteContentBackward` and `insertText` of `InputType`. When we type `ㄱ` => `가`, `ㄱ` is deleted and `가` is inserted using beforeinput events. While doing this, it clears composing, which is not accessible due to above reason, at least I don't know what API can help.
- on backspace key input, its default behavior would be clearing composing and deleting the character. But it reaches `KEY_BACKSPACE_COMMAND` handler on keyDown where we prevent default and call deletion handler. Here we delete character but do not clear composing. 
- So this leads to push remnants when next key inputs come.

To fix this, this PR returns earlier than preventing default behavior in `KEY_BACKSPACE_COMMAND` handler.
But currently this change spawns a new bug due to earlier returning before `isCollapsed` handling:

https://github.com/facebook/lexical/assets/40269597/b6d7c1a8-5284-4ce3-9f45-fd956918036c

 So, I guess It should return later than `isCollapsed` handling and also line for preventing default should move:

```ts
 editor.registerCommand<KeyboardEvent>(
      KEY_BACKSPACE_COMMAND,
      (event) => {
        if ($isTargetWithinDecorator(event.target as HTMLElement)) {
          return false;
        }
        if (!$isRangeSelection(selection)) {
          return false;
        }

        const {anchor} = selection;

        if (
          selection.isCollapsed() &&
          anchor.offset === 0 &&
          !$isRootNode(anchorNode)
        ) {
          const element = $getNearestBlockElementAncestorOrThrow(anchorNode);
          if (element.getIndent() > 0) {
            return editor.dispatchCommand(OUTDENT_CONTENT_COMMAND, undefined);
          }
        }

        // let the browser handle backspace on iOS Korean keyboards,
        // which clears composing and delete the character w/o composition events
        if (IS_IOS && navigator.language === 'ko-KR') {
          return false;
        }

        event.preventDefault();
        return editor.dispatchCommand(DELETE_CHARACTER_COMMAND, true);
      },
      COMMAND_PRIORITY_EDITOR,
    ),
```

and we should add same fix to `registerPlainText`:
```ts
    editor.registerCommand<KeyboardEvent>(
      KEY_BACKSPACE_COMMAND,
      (event) => {

        if (!$isRangeSelection(selection)) {
          return false;
        }

        if (IS_IOS && navigator.language === 'ko-KR') {
          return false;
        }

        event.preventDefault();
        return editor.dispatchCommand(DELETE_CHARACTER_COMMAND, true);
      },
      COMMAND_PRIORITY_EDITOR,
    ),
```
We might need to check:
1. the effect of moving preventDefault to later than `isCollapsed` handling.
2. any other missing cases caused by not calling `onCompositionEndEmpl` of `onCompositionStart`.

2.could be next step though. I believe this fix as is better than nothing.

@ivailop7 @acywatson @zurfyx May I kindly ask this receive some attention? It would be so great if you advise the proper way to fix this. The issue is quite critical for Korean apps.

packages/lexical-rich-text/src/index.ts
```diff
@@ -860,6 +860,17 @@ export function registerRichText(editor: LexicalEditor): () => void {
...
+        // Exception handling for iOS native behavior instead of Lexical's behavior when using Korean on iOS devices.
+        // On other devices, if you backspace during composing, you get a keyCode of 229 (composing state), but on iOS, the behavior is different: it outputs a keyCode of 8 and appends the remnants of the composing character to the previous character.
+         if (
+          event.keyCode === 8 &&
+          navigator.language === 'ko-KR' &&
+          /iPad|iPhone|iPod/.test(navigator.userAgent)
+        ) {
+          return false;
+        }
+
```
potatowagon, line 873: looks low risk to me. if e2e tests pass, ill be merging this to unblock

2wheeh: as is would introduce a different issue recorded in my comment on backspace keydown when is collapsed

potatowagon (approved):

packages/lexical-rich-text/src/index.ts
potatowagon: i see is this safe to merge?

potatowagon: if its not, ill close this PR

2wheeh:
@potatowagon I've opened a pr #6819 based on this branch since the author seems to have been busy :)

PR #4818, opened 2023-07-24
Extract "select all" behavior to new `SELECT_ALL_COMMAND`

montlebalm:
Move the "select all" behavior from `LexicalEvents` into a new `SELECT_ALL_COMMAND`. This will make overriding the default behavior easier for developers and matches the way that commands like copy, paste, arrow up, etc are handled.

Overriding "select all" is not common, but several applications have introduced context-aware versions of the behavior that are difficult to change with lexical's current implementation. For example, an editor may choose to have the first invocation of `⌘ + a` select the current paragraph and a subsequent `⌘ + a` select the entire document.

### DIFF
--- a/packages/lexical-plain-text/src/index.ts
+++ b/packages/lexical-plain-text/src/index.ts
@@ -21,6 +21,7 @@ import {mergeRegister} from '@lexical/utils';
 import {
   $getSelection,
   $isRangeSelection,
+  $selectAll,
   COMMAND_PRIORITY_EDITOR,
   CONTROLLED_TEXT_INSERTION_COMMAND,
   COPY_COMMAND,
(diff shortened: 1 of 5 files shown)


fantactuka (approved):
👍

acywatson:
I'm good with this, but I think if someone is using LexicalEvents directly (looks like this can be called via setRootElement) this could break them. I don't think this is a common case, but might be more common outside in vanilla js implementations.

I think we might want to version bump unless I'm wrong about this.

acdcjunior:
Sorry for hijacking the thread, but on the off chance anyone also gets issues after upgrading lexical: in my case dependabot created a PR that updated lexical from 0.11.3 to 0.12.0 and suddenly CTRL+A stopped working, even though I used `RichTextPlugin`.

What was causing this? In my case `@lexical/react` was still pointing to `0.11.3`. Updating it to `0.12.0` (so that both `lexical` and `@lexical/react` now point to  `0.12.0`) fixed the issue.

Which makes sense since this PR changes multiple packages.

Anyway, HTH.

PR #4823, opened 2023-07-25
feat(playground): add support for collapsed margins in draggable block

mathieumetral:
Add support for collapsed margins for draggable blocks in the playground.

**Before**

https://github.com/facebook/lexical/assets/46543621/0472a2ba-5171-4e6f-bd3b-b40d875a4d1f

**After**

https://github.com/facebook/lexical/assets/46543621/76d9c86d-ca60-4067-9711-582e58f4c937


ivailop7 (approved):
LGTM, but a second pair of eyes would be nice.

PR #4831, opened 2023-07-26
Fix div paste

### LINKED ISSUE #4465: Support div on paste
```
<div>
a
</div>
<div>
b
</div>
```

Should return 
```
a
b
```

instead of `       a              b       `

karam-qaoud:
Fixes #4465

### DIFF
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -342,7 +342,7 @@ function convertDivElement(domNode: Node): DOMConversionOutput {
   const isCode = isCodeElement(div);
   if (!isCode && !isCodeChildElement(div)) {
     return {
-      node: null,
+      node: $createParagraphNode(),
     };
   }
   return {
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
@@ -829,6 +829,29 @@ test.describe('CopyAndPaste', () => {
     );
   });
 
+  test('Copy paste html divs', async ({page, isPlainText}) => {
+    test.skip(isPlainText);
+    await focusEditor(page);
+    await pasteFromClipboard(page, {
+      'text/html': '<div>a</div><div>b</div>',
+    });
+    await assertHTML(
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">a</span>
+        </p>
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">b</span>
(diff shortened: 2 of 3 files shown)


karam-qaoud:
The current behavior is that div elements are completely ignored by Lexical. This fix simply deals with divs as paragraph nodes, and that solves the issue. But I am not sure that's optimal. I think we should come up with a more primitive Lexical node to handle divs (something more lightweight that just groups elements together - like React Fragments). Also, would be cool if we could squash unnecessary divs as an optimization in case there is lots of nesting in the input html.

packages/lexical-code/src/CodeNode.ts
```diff
@@ -342,7 +342,7 @@ function convertDivElement(domNode: Node): DOMConversionOutput {
   const isCode = isCodeElement(div);
   if (!isCode && !isCodeChildElement(div)) {
     return {
-      node: null,
+      node: $createParagraphNode(),
```
acywatson, line 345: This will only work if CodeNode happens to be registered on the editor. What we might want to do is add some logic in ParagraphNode to handle divs, but this needs to be lower priority than the Code conversion and it could have some unintended effects (this is not uncommon with HTML copy + paste, though, given how all text editors do it differently).

acywatson:
> The current behavior is that div elements are completely ignored by Lexical. This fix simply deals with divs as paragraph nodes, and that solves the issue. But I am not sure that's optimal. I think we should come up with a more primitive Lexical node to handle divs (something more lightweight that just groups elements together - like React Fragments). Also, would be cool if we could squash unnecessary divs as an optimization in case there is lots of nesting in the input html.

One of the issues here is that divs can have different semantics in different context. For instance, some code editors put divs on the clipboard for each line of code (I can't remember which do this, but definitely some popular ones).

packages/lexical-code/src/CodeNode.ts
karam-qaoud: Gotcha, I just added a case to handle divs in ParagraphNode

karam-qaoud:
> One of the issues here is that divs can have different semantics in different context. For instance, some code editors put divs on the clipboard for each line of code (I can't remember which do this, but definitely some popular ones).

Ah, I see! It makes sense to prioritize consistency over optimization in this case.

acywatson:
I think this should be ok - can we fix the unit tests?

Thanks for your recent contributions @karam-qaoud!

karam-qaoud:
> Thanks for your recent contributions @karam-qaoud!

Glad I could help!

(after looking into the failing unit tests) It seems like paragraph tags are not the best replacement of div tags. If we try to include a `table` inside a `p` tag, the html becomes invalid. The reason is that browsers try to be smart and automatically close the `p` tag when they encounter another block-level element inside it, and that makes the html invalid. The same issue happens when you try to nest `p` tags inside each other. 

That said, I am considering whether we should introduce `LexicalDivNode`? But I have mixed feelings about it, just like you said divs can have many different semantics. What do you think?

zurfyx:
Inactivity cleanup. This is still an issue but unfortunately the correct approach moves around the ability to be able to detect when content is a block vs inline and divs and play both roles depending on their styling. Additionally, divs tend to be redundant containers and should be excluded during the conversion when this is the case.

PR #4836, opened 2023-07-27
Playground: refactor component picker plugin

thorn0:


acywatson:
Seems fine - what's the aim here?

thorn0:
@acywatson I was reading through this code and noticed a bunch of small things.
Commit 1: local simplifications without moving code. E.g. getting rid of `ts-ignore` comments and redundancies like `new RegExp(/foo/)`. Also it fixes a bug with filtering options by keywords.
Commit 2 extracts functions out of the component because shorter functions are easier to read and because `useCallback` isn't really needed when the function isn't passed to a prop.

packages/lexical-playground/src/plugins/ComponentPickerPlugin/index.tsx
```diff
@@ -108,245 +109,237 @@ function ComponentPickerMenuItem({
...
-              option.keywords != null
-              ? option.keywords.some((keyword) =>
-                  new RegExp(queryString, 'gi').exec(keyword),
-                )
-              : false;
+            const regex = new RegExp(queryString, 'i');
+            return (
+              regex.test(option.title) ||
+              option.keywords?.some((keyword) => regex.test(keyword))
+            );
```
thorn0, line 338: This is a bug fix.

packages/lexical-playground/src/plugins/ComponentPickerPlugin/index.tsx
```diff
@@ -108,245 +109,237 @@ function ComponentPickerMenuItem({
...
-            return new RegExp(queryString, 'gi').exec(option.title) ||
-              option.keywords != null
-              ? option.keywords.some((keyword) =>
-                  new RegExp(queryString, 'gi').exec(keyword),
-                )
-              : false;
+            const regex = new RegExp(queryString, 'i');
+            return (
+              regex.test(option.title) ||
+              option.keywords?.some((keyword) => regex.test(keyword))
```
fantactuka, line 337: `keywords` is not nullable so ?. is redundant

packages/lexical-playground/src/plugins/ComponentPickerPlugin/index.tsx
```diff
@@ -108,245 +109,237 @@ function ComponentPickerMenuItem({
...
-      new ComponentPickerOption('Code', {
-        icon: <i className="icon code" />,
-        keywords: ['javascript', 'python', 'js', 'codeblock'],
-        onSelect: () =>
-          editor.update(() => {
-            const selection = $getSelection();
+    );
+  }
+
+  return options;
```
fantactuka, line 149: Full and partial table matchers can be replaced with a single one:

```
const tableMatch = queryString.match(/^(\d{1,2})x?(\d{1,2})?$/);

if (tableMatch !== null) {
  const row = Number(tableMatch[1]);
  const cols = tableMatch[2]
    ? [Number(tableMatch[2])]
    : [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];

  return cols.map(col => ({
    icon: <i className="icon table" />,
    label: `${row}x${col} Table`,
    keywords: ['table'],
    onSelect: (_: string) => {
      ...
    },
  }));
}
```

fantactuka (approved):
LG!

PR #4842, opened 2023-07-28
Added "Page Break" to Lexical Playground

Dias1c:
image

# New Page Break Plugin
Page Break - breaks page. You can see it when you are exporting this document to PDF. 
Its helpfull thing if you are working with PDF document.

### On exporting created HTML to PDF
image

## How to create page break?
1. Using toolbar (Click to `+ insert` button in toolbar, and select `Page Break` option)
2. Using slash `/` menu (Start writing `/page` and select `Page Break`)

## Additional information
I have loaded `scissors` icon from [there](https://www.svgrepo.com/svg/40147/scissors)
image
For icon using [CC0 License](https://www.svgrepo.com/page/licensing#CC0)


ivailop7:
Looking great! For the icon, we've been using the bootstrap icons pack, it has a scissors icon as well. Would be nice to keep it consistent. 

https://icons.getbootstrap.com/

Dias1c:
Ah, ok. I will change it tomorrow soon as possible :)

packages/lexical-playground/src/nodes/PageBreakNode/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  exportJSON(): SerializedLexicalNode {
+    return {
+      type: this.getType(),
+      version: 1,
+    };
+  }
+
+  exportDOM(): DOMExportOutput {
+    return {element: this.createDOM()};
```
acywatson, line 132: I think this is what the super class does (so omitting the method would delegate to that implementation) but I may be wrong.

acywatson (approved):
Nice!

Dias1c:
> Looking great! For the icon, we've been using the bootstrap icons pack, it has a scissors icon as well. Would be nice to keep it consistent.
> 
> https://icons.getbootstrap.com/

I have replaced scissors icon to `bootstraps` scissors icon

packages/lexical-playground/src/nodes/PageBreakNode/index.tsx
Dias1c: Is it planned to omit the "createDOM" method?

In order to eliminate the human factor when changing the class, and in order not to repeat the same code, I decided to use this method.

In the future, if you need to delete this method, I think it will not be difficult to insert the entire contents of the "createDOM" method into "exportDOM".

If you want, I can write an external function that does the same thing, and I will use it in these methods.

The **only reason visible to me why it is worth separating** this dependency, I see if we would have to draw in the editor, and show different styles/properties of the same HTML when exporting.

*Perhaps I misunderstood you, so I would be glad for a detailed explanation.*

acywatson: PageBreakNode extends DecoratorNode, which extends LexicalNode.

DecoratorNode does not implement exportDOM, but [LexicalNode does](https://github.com/facebook/lexical/blob/cc2522e587a8ee0f3520d19fdd9995bc76161804/packages/lexical/src/LexicalNode.ts#L780). The implementation of exportDOM in LexicalNode is the same as your implementation in PageBreakNode.

Therefore, if you remove your implementation, calls to PageBreakNode.exportDOM will use the implementation of LexicalNode.exportDOM, which will do the same thing. So I think you can just delete PageBreakNode.exportDOM.

Dias1c: Thanks for explained response. I have made commit which removes `exportDOM` method from `PageBreakNode`.

ivailop7:
Just formatting left to pass integrity checks and it's ready to go

PR #4850, opened 2023-07-31
fix(playground-text-format-toolbar): don't hide toolbar if current node is a paragraph

mathieumetral:
Improve the toolbar to format text and avoid hiding it unnecessarily

**Before**

https://github.com/facebook/lexical/assets/46543621/399604a8-b6ae-4a42-bf38-e8beba6cbca3

**After**

https://github.com/facebook/lexical/assets/46543621/d1297907-ec54-4ba1-b848-c4cae92af5b0


mathieumetral:
Any updates on this PR? Maybe @ivailop7 ?

PR #4860, opened 2023-08-07
Quote can be entered in multiple lines

fannifanni:
### In the quote, click enter to exit the quote, and the quote text cannot be entered in multiple lines

quote-only-input-one-line

## Implementing Line Break in Quote Node

quote-line-break

## Additional Information

- In the quote, click enter, and the number of quote lines will increase by one.
- Press two consecutive enters and there are two lines of empty text to exit quote，same as code exit mechanism.


fantactuka:
You can do it with Shift+Enter to have multiline quote, or you can override ENTER_COMMAND in app-specific plugin without changing core node.

fannifanni:
> You can do it with Shift+Enter to have multiline quote, or you can override ENTER_COMMAND in app-specific plugin without changing core node.

OK，thank you!

Issue #4196, opened 2023-03-25
Bug: Max length plugin breaks when max length is reached and all contents are deleted


harish-sethuraman:
Lexical version: Latest

## Steps To Reproduce

1. Enable max length in settings 
2. type until max length is reached
3. select all and delete it
4. repeat 2 and 3 until the bug unveils

https://user-images.githubusercontent.com/32865581/227710291-c6d0850d-0164-4536-bebe-814a23924a58.mov

Link to code example:

playground.lexical.dev

## The current behavior
https://user-images.githubusercontent.com/32865581/227710291-c6d0850d-0164-4536-bebe-814a23924a58.mov
## The expected behavior
Max length plugin should Ideally not break the editor

harish-sethuraman:
Anything else is required? is it reproducible for anyone else as well?

SalmanWasiKhan:
I think it is related to #4075

takuyakanbr:
Closing as this was fixed in #4871.

Issue #4075, opened 2023-03-09
Bug: MaxLengthPlugin crashes when the limit is reached and enter is pressed


luo-git:
Lexical version:
0.8.0

## Steps To Reproduce

1. Go to lexical playground and turn on Max Length Plugin (https://playground.lexical.dev/?isMaxLength=true)
2. Delete everything.
3. Enter random text until limit is reached.
4. Press enter.
5. Crashes.
![image](https://user-images.githubusercontent.com/51979232/224071753-81f1e68b-3bab-4fcf-8b62-d51fb62fb64a.png)

Link to code example:
https://playground.lexical.dev/?isMaxLength=true

## The current behavior
Crashes. Further input has no effect.

## The expected behavior
Enter should do nothing.

SalmanWasiKhan:
I am also facing the same issue (using 0.9.1). 

To add a bit more context I am seeing this error on the console while using it in another project and also in the clone
![image](https://user-images.githubusercontent.com/71233371/228931684-537c362b-afc4-47e2-8938-87c2cacf6acf.png)

kirandash:
I've tested this issue with the latest version of Lexical and it appears to be fixed. The MaxLengthPlugin no longer crashes when the limit is reached and enter is pressed. Seems like this was addressed in PR #4871 which has been merged.

For anyone still experiencing this issue, please make sure you're using the latest version of Lexical. If you're still encountering the problem with the latest version, please provide:
1. Your exact Lexical version
2. A minimal reproduction case
3. Any console errors you're seeing

Since this has been fixed and merged, I suggest we close this issue.

PR #4871, opened 2023-08-09
Fix: MaxLengthPlugin crashes when the limit is reached and enter is pressed

khsily:
This is an attempt to fix two main issues I found:
1. When cloning node, the activeEditorState will have some unreachable node keys which will be remove later. It causes `Uncaught Error: createChildrenArray: node does not exist in nodeMap` error.
2. Pressing enter on its maxLength deletes previous node because the node has 0 content size.

For more details please see: #4075  #4196


fantactuka:
It makes sense to me, as we're trying to restore specified editor state but relying on `getLatest()` which would attempt to take a node from active editor state instead.

@zurfyx WDYT?

fantactuka (approved):

PR #4877, opened 2023-08-10
Table unmerge

### LINKED ISSUE #4876: Bug: $unmergeCell adds to wrong rows when cell is two the left and there are more than 2 rows
$unmergeCell adds cells correctly to the first and second row, but if the row span is more than two, the cells that are supposed to go in the following rows are all added to the second row.

Lexical version: 0.11.3

## Steps To Reproduce

1. Add table
2. Merge cells on the left of the table with a row span of 3 or more
3. Unmerge newly merged cell

https://github.com/facebook/lexical/assets/23295934/74ce9053-5e42-4f24-b4e4-721d1ee802a8

Link to code example:

## The current behavior
From row 3 in merged cell, all cells are added to the second row.
## The expected behavior
Cells should be added to their respective row

PeterBul:
Fixes #4876

When adding cells to rows when unmerging cells, $unmergeCell used `row.getNextSibling()` with the same row an all iterations  of the row span except the first to get current row node. This resulted in the same row node being used on all iterations.

https://github.com/facebook/lexical/assets/23295934/5f23b3ed-c467-4b55-b595-67bfb3ab2a12

### DIFF
--- a/packages/lexical-playground/__tests__/regression/4876-unmerge-cell.spec.mjs
+++ b/packages/lexical-playground/__tests__/regression/4876-unmerge-cell.spec.mjs
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+import {
+  click,
+  expect,
+  focusEditor,
+  initialize,
+  insertTable,
+  mergeTableCells,
+  selectCellsFromTableCords,
+  test,
+  unmergeTableCell,
+} from '../utils/index.mjs';
+
+test.describe('Regression test #4876', () => {
+  test.beforeEach(({isCollab, page}) => initialize({isCollab, page}));
+  test('unmerging cells should add cells to correct rows', async ({
+    page,
+    isPlainText,
+    isCollab,
+  }) => {
+    test.skip(isPlainText);
+
+    await focusEditor(page);
+
+    await insertTable(page, 4, 4);
+
+    await click(page, '.PlaygroundEditorTheme__tableCell');
+    await selectCellsFromTableCords(
+      page,
+      {x: 0, y: 1},
+      {x: 1, y: 3},
+      true,
+      false,
+    );
+
+    await mergeTableCells(page);
+
+    await unmergeTableCell(page);
+
+    const tableRow = await page.locator('tr');
+    expect(await tableRow.count()).toBe(4);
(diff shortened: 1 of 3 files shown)


PeterBul:
I rebased the PR onto main.
The integration tests seemed to fail because of `??`. I have changed it to logical or `||`.

zurfyx (approved):

packages/lexical-playground/__tests__/regression/4876-unmerge-cell.spec.mjs
```diff
@@ -0,0 +1,53 @@
...
+      {x: 1, y: 3},
+      true,
+      false,
+    );
+
+    await mergeTableCells(page);
+
+    await unmergeTableCell(page);
+
+    const tableRow = await page.locator('tr');
```
zurfyx, line 46: `page.locator` will work fine in offline mode but not when there's a split view on collab, that's why the test is failing there

You probably want to do 
```
let leftFrame = page;
  if (IS_COLLAB) {
    leftFrame = await page.frame('left');
  }
```

And additionally consider adding a utility function in utils/index.mjs near `waitForSelector`

zurfyx (commented):
Thank you for the fix! Adding one last comment

packages/lexical-playground/__tests__/regression/4876-unmerge-cell.spec.mjs
PeterBul: I have now made a utility function `locate` in the same style as the other utility functions.

lininu:
Hi, it still happen now 🥺 https://playground.lexical.dev/#doc=H4sIAAAAAAAAE-2YTWvDMAyG_4vOpjgpbTZfBzsPchw7uLbWhHlWUJx2peS_j37BMlhYyxil9skYW7KeVx8HbwFtHYjLoAOC2gIThd1qqtpZRg_qebB5EWBrRhNq8qB855yAV-J3HUABCKi9RR9ASQFh0yAoaDTrJeumAgEr5HZvmAkI-BEej5ayFz8_-f_BnO836IVDg85987vQ5m3J1Hn7QI745MyQKxt9uFKhtnhKwFQA0_p4NibK7aFnCT2hx4t-YRxM60EYvx6kt6dnHm8pyYSe0ONFT7PzXD3zkdk5jbmUEnqM6GmApFJK6KmLrkvPiEspYvS_6qIL_Yy14vVIv_-mHkTaC3C6DaVeoQWVFdnsPp_L4q6YFwJa6tjs7J6c3hxS89Ua5CSbTST0nxnqL9MPFwAA

PR #4881, opened 2023-08-10
Preserve Text Color on Import DOM

ChanManChan:
Text styles (color) are not applied when calling importDOM method. This PR will fix that issue


acywatson:
You can do this in userland by overriding TextNode:

https://lexical.dev/docs/concepts/serialization#handling-extended-html-styling

PR #4900, opened 2023-08-18
Modularize the paste process with a new INSERT_NODES_COMMAND

GermanJablo:
Through a small refactoring, this PR splits the paste process into 2 stages: `PASTE_COMMAND`, and `INSERT_NODES_COMMAND`.

The reason for this is that I need to do some transformations on the pasted nodes, and I would end up copying a huge amount of code just to modify a couple of lines.

This allows users to intervene only in the final part of the process.


GermanJablo:
I comment to notify that I left an edit in the previous message last Friday.

___

As for me, my only reservation about this PR is to make it future proof, and not follow a specific implementation.

INSERT_NODES command does some normalization, so the current paste process is:
1. Convert clipboard to an array of nodes.
2. Normalize the array to be inserted.
3. Insert the normalized array.

My question is whether the normalization from step 2 should be part of PASTE_COMMAND or INSERT_NODES (or a third command?). On the other hand... Is this 3-stage process one implementation among other possible ones? Personally, the only thing that occurs to me is that the normalization could be done at the insertion itself or post insertion.
Maybe these are not critical questions, but worth sharing. Let me know what you think.

acywatson:
> I comment to notify that I left an edit in the previous message last Friday.
> 
> As for me, my only reservation about this PR is to make it future proof, and not follow a specific implementation.
> 
> INSERT_NODES command does some normalization, so the current paste process is:
> 
> 1. Convert clipboard to an array of nodes.
> 2. Normalize the array to be inserted.
> 3. Insert the normalized array.
> 
> My question is whether the normalization from step 2 should be part of PASTE_COMMAND or INSERT_NODES (or a third command?). On the other hand... Is this 3-stage process one implementation among other possible ones? Personally, the only thing that occurs to me is that the normalization could be done at the insertion itself or post insertion. Maybe these are not critical questions, but worth sharing. Let me know what you think.

This is a good question. The semantics of pasting in an editor typically couple transformation and insertion, which makes it hard to align with an external convention or example and still support this use case.

My opinion is that normalization here is built around the editors handling of insertion. We make it safe to insert at the root and rely on the insertNodes algorithm or others to handle any merging or whatever if the paste destination is somewhere else. If we're going to make this generic, I would put the override hook **before** normalization, so that whatever the user is going to do isn't impacted by the normalization logic. So I guess that means it needs to be part of a paste-specific INSERT_NODES command.

GermanJablo:
If I am understanding you correctly, then I should leave it as it is now in this PR, right?

And regarding the name of the command? `INSERT_PASTED_NODES_COMMAND` or `INSERT_NODES_COMMAND`?

acywatson:
> INSERT_PASTED_NODES_COMMAND

This, I think.

acywatson:
Also some failing Unit tests, apparently

GermanJablo:
Something pending with this?

Is it possible that it will be in the next release?

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -172,18 +176,22 @@ export function $insertDataTransferForRichText(
     dataTransfer.getData('text/plain') || dataTransfer.getData('text/uri-list');
   if (text != null) {
     if ($isRangeSelection(selection)) {
+      let lastParagraphNode = $createParagraphNode();
```
acywatson, line 179: There are some e2e tests failing, possibly because of this

nerdess:
@GermanJablo I need to also cleanup the nodes after a user has pasted content (e.g. from MS Word) into Lexical. Your PR looks helpful for that! 

Can you maybe post some sample code how you utilized your PR and achieved a transformation on your pasted nodes? I am struggeling to intercept into the pasting-process.

GermanJablo:
I'm working right now on fixing the tests. As some entire suites passed I thought those who failed were flaky.

@nerdess you have tried the new command on this PR, right?

acywatson:
> @GermanJablo Not yet, I was first trying to get the already existing `PASTE_COMMAND` to work but it does not get triggered upon pasting and I am wondering why. So I thought I could peek at your code and get some inspiration.
> 
> ```
> useEffect(() => {
>       return mergeRegister(
>             editor.registerCommand(
>                 PASTE_COMMAND,
>                 (payload) => {
>                     console.log('paste', payload);  //This never happens!!! :(
>                     return true;
>                 },
>                 COMMAND_PRIORITY_EDITOR,
>               ),
>        );	  
> }, [editor]);
> ```

Did you try a higher priority? COMMAND_PRIORITY_EDITOR is what Lexical uses internally, generally, so if there's another listener at that priority it could run first and stop propagation to your listener.

acywatson:
> I'm working right now on fixing the tests. As some entire suites passed I thought those who failed were flaky.

Yea, not an unreasonable assumption to make, given condition of the CI right now.

nerdess:
@acywatson ok that is some new learning....didn't know that listeners on the same priority can cancel the remaining ones...the command works now when using `COMMAND_PRIORITY_HIGH`

@GermanJablo wanted to use your fork instead of the lexical repo but somehow it always turns out

`"@lexical/monorepo": "GermanJablo/lexical",`
    
in package.json?!
    
I expect
    
`"lexical": "GermanJablo/lexical"`
        
Might just wait for the PR to be merged...

GermanJablo:
@nerdess The way to test the code is to checkout the PR. If you want to incorporate the changes to your project, yes, you should wait for it to merge.

Trying to fix the tests in this PR, I ended up redoing the `insertNodes` method in another PR. But I currently have a blocker, [described here](https://github.com/facebook/lexical/pull/5002#discussion_r1326155278). Any help is welcome.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -1972,7 +1972,7 @@ describe('LexicalSelectionHelpers tests', () => {
         });
 
         expect(element.innerHTML).toBe(
-          '<h1 dir="ltr"><span data-lexical-text="true">foobar</span></h1>',
+          '<h1 dir="ltr"><span data-lexical-text="true">foo</span></h1><p dir="ltr"><span data-lexical-text="true">bar</span></p>',
```
GermanJablo, line 1975: changed to maintain consistency with content copied from GDocs (they don't wrap the last line in a <p> in the clipboard

GermanJablo:
I found a better way to achieve this.

Even with a command to insert nodes the user would have to modify a piece of code that is very large and complex like `insertNodes`.

https://github.com/facebook/lexical/pull/5201 makes this PR obsolete, as it allows you to "modify" pasted nodes simply by overriding the `insertRangeAfter` method.

PR #4904, opened 2023-08-20
Adds ElementFormatDropdown to the Toolbar plugin

itaquito:
Hello! 

I noticed that most of the buttons on the playground are reactive depending on the current selection of the editor, with the exception of the element format dropdown (aka, the align paragraph dropdown).

I added a new element with the corresponding update logic to enable this dropdown to change with the current selection. Hope it meets the standards of the project.

Thanks for creating such an awesome project!


fantactuka (approved):
Looks good!

PR #4915, opened 2023-08-22
Update editor payload type

cksal0805:
Hello,
Thank you for developing such an excellent framework.

I have a question regarding the use of "dispatchCommand".

I often find myself passing undefined when using dispatchCommand payload.

In fact, the actual playground examples are set up that way too. However, this seems like a cumbersome step every time.

If it's not mandatory to always include undefined, it would be nice if the type could be updated accordingly.

example
```ts
const formatBulletList = () => {
    if (blockType !== "ul") {
      editor.dispatchCommand(INSERT_UNORDERED_LIST_COMMAND, undefined);
    } else {
      editor.dispatchCommand(REMOVE_LIST_COMMAND, undefined);
    }
    setShowBlockOptionsDropDown(false);
};
```

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -878,7 +878,7 @@ export class LexicalEditor {
    */
   dispatchCommand<TCommand extends LexicalCommand<unknown>>(
     type: TCommand,
-    payload: CommandPayloadType<TCommand>,
+    payload?: CommandPayloadType<TCommand>,
   ): boolean {
     return dispatchCommand(this, type, payload);
   }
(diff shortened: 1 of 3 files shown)


fantactuka:
LGTM, cc @thegreatercurve as commands expert

acywatson (commented):
Might need to update Flow types also. 

This seems like it would require us to check for the existence of payload before accessing it in every command listener...

zurfyx:
Sorry for the late review. This was intentional as there is no way (to my knowledge) to validate the payload when it's not passed. I.e. it works well when it's undefined `dispatchCommand(FOO)` but it won't catch `dispatchCommand(BAR)` where `T` in `Command<T>` is not `undefined`

PR #4927, opened 2023-08-28
fix: disable spellcheck for code format

zignis:
Disables spellcheck for text nodes with code format.

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -451,7 +451,7 @@ export class TextNode extends LexicalNode {
     const dom = document.createElement(tag);
     let innerDOM = dom;
     if (this.hasFormat('code')) {
-      dom.spellcheck = false;
+      dom.setAttribute('spellcheck', 'false');
     }
     if (outerTag !== null) {
       innerDOM = document.createElement(innerTag);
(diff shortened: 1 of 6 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -450,6 +450,9 @@ export class TextNode extends LexicalNode {
     const tag = outerTag === null ? innerTag : outerTag;
     const dom = document.createElement(tag);
     let innerDOM = dom;
+    if (outerTag === 'code') {
```
zurfyx, line 453: Probably safer to compare against the property value 
```suggestion
    if (this.hasFormat('code')) {
```

zurfyx (commented):
Thanks! Should we also add a condition on LexicalCodeNode to handle code blocks?

You'll also need to update the E2E tests to reflect this new property

zignis:
> Thanks! Should we also add a condition on LexicalCodeNode to handle code blocks?
> 
> You'll also need to update the E2E tests to reflect this new property

I think LexicalCodeNode already handles this:

https://github.com/facebook/lexical/blob/0e7977ef15fb007153088a1bdc52d827e893d5ae/packages/lexical-code/src/CodeNode.ts#L106

zurfyx:
Thank you! I'll merge this. Any chance you can push another PR with the code block (CodeBlockNode)?

Screenshot 2023-08-30 at 5 55 22 PM

The first case should also be marked as non spellcheckable but it won't fall under the TextNode condition you added

zurfyx (approved):

zignis:
> Thank you! I'll merge this. Any chance you can push another PR with the code block (CodeBlockNode)?
> 
> Screenshot 2023-08-30 at 5 55 22 PM
> 
> The first case should also be marked as non spellcheckable but it won't fall under the TextNode condition you added
> 
> 

I'll look into it

zignis:
@zurfyx the code element already has `spellcheck=false`:

Snapshot_2023-08-31_19-19-15

Am i missing something here?

Issue #4908, opened 2023-08-21
Bug: Reverts to previos styles when you click in an empty editor


asaAnnabelle:
When you change a style when the text field is empty, and click into the editor box instead of just starting to type, the styles are reverted to the initial styles. 

Lexical version: v0.12.0

## Steps To Reproduce

1. Go to the Lexical Playground and remove all the text 
2. Change some styles
3. Click in the editor box (before typing)
4. Now, the styles are reverted to the initial styles

Link to code example: https://playground.lexical.dev/ 

The styles seems to be removed in the `LexicalEvents.ts -> onSelectionChange -> updateEditor` 
at the last section of this code blob `else if (anchor.type === 'element')`
```javascript
if (
          currentTimeStamp < timeStamp + 200 &&
          anchor.offset === lastOffset &&
          anchor.key === lastKey
        ) {
          selection.format = lastFormat;
          selection.style = lastStyle;
        } else {
          if (anchor.type === 'text') {
            selection.format = anchorNode.getFormat();
            selection.style = anchorNode.getStyle();
          } else if (anchor.type === 'element') {
            selection.format = 0;
            selection.style = '';
          }
        }
```
This problem can be fixed with two solutions.
1. Remove the `currentTimeStamp < timeStamp + 200`. However, I am not sure why this time check is added. 
2. Change the `if(anchor.type === 'text')` to get values. E.g. the `lastFormat` and `lastStyle`.

## The current behavior
![behavior_now](https://github.com/facebook/lexical/assets/59730495/a1c0dfb7-6a90-4bdb-92a1-0fff5113b26b)

## The expected behavior
![behavior_wanted](https://github.com/facebook/lexical/assets/59730495/4e044ecc-0826-4d4f-838c-9740a1b38079)

PR #4928, opened 2023-08-28
fix: remove revert to previos styles on empty editor click

asaAnnabelle:
Fixes the problem where the styles get reverted when you click on a collapsed selection.

Fix for Issue: [#4908](https://github.com/facebook/lexical/issues/4908)

## The current behavior
behavior_now

## The new expected behavior
behavior_wanted

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -334,7 +334,7 @@ function onSelectionChange(
             selection.format = anchorNode.getFormat();
             selection.style = anchorNode.getStyle();
           } else if (anchor.type === 'element') {
-            selection.format = lastFormat;
+            selection.format = anchorNode.getFormat();
             selection.style = lastStyle;
           }
         }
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -334,8 +334,8 @@ function onSelectionChange(
             selection.format = anchorNode.getFormat();
             selection.style = anchorNode.getStyle();
           } else if (anchor.type === 'element') {
-            selection.format = 0;
-            selection.style = '';
+            selection.format = anchorNode.getFormat();
```
acywatson, line 337: I guess we could lift this expression up, since it now applies to both cases.

acywatson (dismissed):

packages/lexical/src/LexicalEvents.ts
asaAnnabelle: Seems like getFormat() will not get any format when clicking, and styles does not exist in the element node. We can move it into the first if block, to minimize duplication?
Screenshot 2023-09-12 at 09 19 26

acywatson:
Several tests failing here

zurfyx:
Thank you, seems like there's another 4 failing unit tests to go

potatowagon:

Issue #4880, opened 2023-08-10
Bug: Font Colors do not apply to text within cells


ivailop7:
Bold, Italic, Underline work by selecting multiple cells and applying those styles, but color and backgroundColor do not get applied:

https://github.com/facebook/lexical/assets/7893468/a09963c6-945c-492d-9d0d-4cd266d22ce9

tahir606:
Hi guys, a question regarding this issue. From what I have been able to capture from going through the code is that gridSelection is being used for table selections. But it has been marked as deprecated throughout the code. Can somebody explain that?

@ivailop7

ivailop7:
The tables guru is @fantactuka

tahir606:
@ivailop7 I have resolved this issue, the functionality for this issue wasn't built, but I have added relevant compatibility.

Here's a [loom video](https://www.loom.com/share/85ac1ef1a3f9434dbff476b6545a1f6c?sid=089343f4-c146-449b-a079-6b52c958c709)


I will be cleaning up the code and opening a PR shortly.

tahir606:
Hi! @ivailop7, I have created a PR for this issue and it is linked above. Please review and let me know if any changes are required. This is my first PR on this project so I might be missing something but I have tested lint. 

Thanks!

ivailop7:
Great! Having a look.

PR #4934, opened 2023-08-29
Fix: font colors do not apply to text within table cells

tahir606:
Resolves [issue 4880](https://github.com/facebook/lexical/issues/4880).

I have added handling for applying font color and background color through the toolbar to selected table cells in a table.


ivailop7:
@fantactuka What's the right way to do this if the goal is to move away from the grid selection.

tahir606:
I can see that we are moving away from GridSelection and focusing on RangeSelection, which is why I have reused as much of the existing code as possible. The only thing that I have added is parsing the grid into range. In the future, if we want to remove usage we will just need to take care of the parsing. :D 

@fantactuka @ivailop7

ivailop7:
@tahir606 the unit tests fail, otherwise LGTM

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -304,6 +303,37 @@ function $patchStyle(
...
+  selection.getNodes().forEach((cellNode) => {
+    if (
+      cellNode.getType() === 'tablecell' &&
+      cellNode.getTextContentSize() !== 0
+    ) {
+      anchor.set(cellNode.getKey(), 0, 'element');
+      focus.set(cellNode.getKey(), cellNode.getChildrenSize(), 'element');
+      $patchStyleText(formatSelection, patch);
+    }
+  });
```
zurfyx, line 332: If the limitation is just simply the fact that GridSelection needs to be converted to a RangeSelection, should we just have a RangeSelection that spans the entire table? I believe this would perform better as the iteration would only be done once and also simplify this logic.

I.e. RootNode `element` selection from index 0 to 1 where RootNode children[0] is a TableNode

We can then even change the signature of the function to accept either RangeSelection or GridSelection

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -668,6 +670,9 @@ export default function ToolbarPlugin(): JSX.Element {
         const selection = $getSelection();
         if ($isRangeSelection(selection)) {
           $patchStyleText(selection, styles);
+        } else if (DEPRECATED_$isGridSelection(selection)) {
+          $patchStyleTable(selection, styles);
+          activeEditor.dispatchCommand(SELECTION_CHANGE_COMMAND, undefined);
```
zurfyx, line 675: Let's use `$setSelection(null)`, the dispached command was never envision to trigger programatic selection changes, it was more like a way to forward external events or similar

zurfyx:
Thank you Tahir! added some comments

packages/lexical-selection/src/lexical-node.ts
tahir606: Hi @zurfyx, I understand a bit, but not completely. Do you mean that we should change the table implementation to use the RangeSelection?

Can you elaborate a bit? 

FYI: I am just getting started on this project, which is why what you're saying is not completely clear to me.

Thanks!

zurfyx: Nono, what I'm suggesting is that instead of range selecting each of the nodes we range select the entire table and then apply the function, the same anchor/focus logic you have in 328-329L but applied to the entire table. This selection is sort of equivalent to the GridSelection and will enable you to call $patchStyleText

tahir606: @zurfyx I will go through the code in more detail to understand the concept, but I am passing only the selected cells in the table. If I pass the whole table, won't that apply the style changes to the whole table?

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
tahir606: Resolved. @zurfyx

packages/lexical-selection/src/lexical-node.ts
tahir606: Hey @zurfyx, I have looked into the code to see if what you are recommending can be applied, I believe it isn't completely feasible because:
1- If I pass in the whole table it will apply the changes to all the table cells instead of just the selection.
2- The current $getSelection function is using the EditorState to return the selection and it is built to return the gridSelection right now dynamically in the case of the table. 

In my opinion, I think we should go forward with this structure for the moment. What do you think? @zurfyx

tahir606: Hi @zurfyx! How should we proceed here?

tahir606: Hey @zurfyx, reaching out again. :D

zurfyx (approved):
Sorry for the delay, I commited some minor edits on top and it's good to go now, thank you! Would you mind publishing another PR with an E2E test that covers this case please?

PR #4935, opened 2023-08-29
docs: fix example in getting-started/react

mmollick:
I noticed the examples in [Getting Started with React](https://lexical.dev/docs/getting-started/react) weren't working. Updated it to reflect the recommendation from the [Listeners](https://lexical.dev/docs/concepts/listeners#registerupdatelistener) page.


acywatson (approved):
Nice catch, thanks :)

PR #4939, opened 2023-08-30
feat: allow forwarding ref to the content editable component

zignis:
Allows ref to be passed to the ContentEditable component


zignis:
@acywatson fixed

PR #4940, opened 2023-08-30
fix: respect whitelisted style css attributes on text node when importing AND exporting the DOM

Amnesthesia:
We noticed strange behaviour with `TextNode` that's not easy to pick up on the Lexical playground, because Lexical playground doesn't demonstrate exporting to HTML and then importing that HTML again.

TextNode correctly sets **bold**, _italic_, and other style attributes when it *exports* the DOM, but this is lost if you export/import this HTML

This PR defines a small set of "allowed" CSS style attributes (font-family, color, and float) that should be imported, and exported again, when parsing html (instead of discarding these styles).

Originally we used our own node replacement for this, creating a custom `TextNode` that extended LexicalTextNode, but this lead to strange problems with `@lexical/clipboard`, where (for reasons I haven't been able to figure out) text could not be copied without copying the entire nodes, even if `$isTextNode` returned true for our custom nodes. 

We've applied this same patch on the prototype of the builtin Lexical TextNode to solve our issues, and this PR seeks to get that upstream :)

### DIFF
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -80,11 +80,11 @@ describe('Markdown', () => {
       md: '*Hello* world',
     },
     {
-      html: '<p><b><strong>Hello</strong></b><span> world</span></p>',
+      html: '<p><strong>Hello</strong><span> world</span></p>',
       md: '**Hello** world',
     },
     {
-      html: '<p><i><b><strong>Hello</strong></b></i><span> world</span></p>',
+      html: '<p><i><strong>Hello</strong></i><span> world</span></p>',
       md: '***Hello*** world',
     },
     {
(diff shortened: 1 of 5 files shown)


Amnesthesia:
> The problem is that adding it to the core TextNode means it's added everywhere, even for plain text editors.

Right, that makes sense. Is there a way we can know whether we're in a plaintext editor or not? Because when you extend `TextNode`, a bunch of strange behaviours arise — custom text nodes are sometimes being merged together correctly when deleting a newline for example, and the use case mentioned in the OP here about styles set on the text node arent actually being exported (and, what's the point of those styles then?)

When it comes to plaintext editors, if they only export the plaintext and no markup, do they still need to get their output from `exportDOM` (I mean, they don't need a DOM?), or could they just `getTextContent()`?

acywatson:
> Right, that makes sense. Is there a way we can know whether we're in a plaintext editor or not

This wouldn't fix the problem, as every rich text editor doesn't support these features. You'd need some sort of feature system to turn this on/off, which is way clunkier (imo) than just allowing some sort of generic override.

> Because when you extend TextNode, a bunch of strange behaviours arise — custom text nodes are sometimes being merged together correctly when deleting a newline for example, and the use case mentioned in the OP here about styles set on the text node arent actually being exported (and, what's the point of those styles then?)

I think for the first thing (merging) it'd be best to have a separate issue so we can repro and fix it.

For the second thing, the export* and import* methods exist so you can represent those styles with whatever semantics you want. So if you want to add an attribute to the HTML like "my-special-styles={color: 'red'}" you can do that and then just check for it in import and apply whatever styles you want.

> When it comes to plaintext editors, if they only export the plaintext and no markup, do they still need to get their output from exportDOM (I mean, they don't need a DOM?), or could they just getTextContent()?

PlainText is sort of a misnomer here. It includes things like, for example, the FB comment composer, which is more a plaintext editor than a rich one, but only supports a couple of features like mentions or hashtags. If you want a pure plaint text editor, that's an HTML textarea, which means you don't need Lexical.

Amnesthesia:
Ah ... Shoot, well ... Not being able to export styled text in a HTML editor is quite an issue, so what would be the best way to approach this then?

PR #4942, opened 2023-08-30
List node export dom direction

ibastawisi:
The list node exported dom is missing `dir` attribute, this PR adds it.

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -10,9 +10,11 @@ import type {ListNode} from './';
 import type {
   DOMConversionMap,
   DOMConversionOutput,
+  DOMExportOutput,
   EditorConfig,
   EditorThemeClasses,
   GridSelection,
+  LexicalEditor,
   LexicalNode,
   NodeKey,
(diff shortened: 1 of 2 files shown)


ibastawisi:
For anybody facing the same issue, you can add these attributes by setting your editor config html key as follows:
```ts
export const htmlConfig: HTMLConfig = {
  export: new Map<Klass<LexicalNode>, (editor: LexicalEditor, target: LexicalNode) => DOMExportOutput>([
    [
      ListNode,
      (editor, node) => {
        const listNode = node as ListNode;
        const output = listNode.exportDOM(editor);
        const element = output.element;
        if (!element || !isHTMLElement(element)) return output;
        const direction = listNode.getDirection();
        if (direction) { element.dir = direction; }
        return { element };
      },
    ],
    [
      ListItemNode,
      (editor, node) => {
        const listItemNode = node as ListItemNode;
        const output = listItemNode.exportDOM(editor);
        const element = output.element;
        if (!element || !isHTMLElement(element)) return output;
        const direction = listItemNode.getDirection();
        if (direction) { element.dir = direction; }
        return { element };
      },
    ],
  ]),
};
```

Issue #4954, opened 2023-08-31
Bug w/Repro Repo: Custom node and plugin causes TextNode to strip trailing whitespace on importDOM


JakeHadley:
tl;dr: The issue is when the $generateNodesFromDOM function runs, the Lexical TextNode has it's trailing whitespace stripped out when followed by my custom node.

I'm trying to build out my own extended formatting text node. Currently this example only has a bold option, but I'm looking to utilize this to provide different font-weights, among other things. I have steps recorded via Windows Steps Recorder, but I essentially created a few plugins to create a custom text node, load existing html that is formatted for our requirements, and a toolbar. When I format my custom node via one of the buttons on the toolbar, it adds a class to the node. On exportDOM, I create my own custom HTML tag for our internal storage. On importDOM, I translate that custom tag to be a span with the correct class. 

Lexical version: ^0.12.0

## Steps To Reproduce

1. Hit "Primary" button to enable the editor
2. Highlight "something" and hit the bold icon button. Text is bolded
3. Hit the check button icon to close the editor. Text is still bolded.
4. Hit "Primary" button again to enable the editor. Unbolded text now has it's trailing space removed

![lexical-repro](https://github.com/facebook/lexical/assets/35585841/f03963f7-b541-4e76-bf64-2e16ca7f7c38)

It's a nextjs app. Clone the repo, and from the root folder run `npm run dev`.
Link to code example: https://github.com/JakeHadley/lexical-test

## The current behavior
Lexical TextNode whitespace is stripped out on importDOM

## The expected behavior
The text should appear in the editor the same way as it appears outside the editor

acywatson:
@zurfyx I think this is related to the whitespace changes. Second issue about this, I believe.

acywatson:
Here's the other issue: https://github.com/facebook/lexical/issues/4919

JakeHadley:
@acywatson I'm considering opening this back up, but I wanted to confirm first. I have this same issue happening again, but it's with a custom bare bones text node that I'm using to override TextNode so that I can control the export of the html that happens from `generateHtmlFromNodes`. importDOM is getting called in the node, but I guess I don't know for sure if the super call actually works the way it should or not. I'm essentially trying to just get the text exported from the regular text nodes, which is why I'm overriding and exporting `{ element: new Text(this.__text) }` from `exportDOM`.

This is the node, and it's having the whitespace issue again:
```
[...]

Issue #4919, opened 2023-08-25
Bug: HTML serialization and deseralization are not symmetric


augustobmoura:
Lexical version: 0.12.0

## Steps To Reproduce

1. Generate an HTML output with content with multiple spaces together
2. Import same HTML on another editor, spaces will be collapsed

Link to code example:
https://codesandbox.io/s/nameless-bash-dx6p2c?file=/src/App.js

## The current behavior

Since #4467, `$generateNodesFromDOM` collapses all spaces into a single one, the
original reasoning is for fixing a problem with pasting text from different
places. But for our use case the HTML comes from a truthful source (database)
and can be deserialized as it is, and we intend to keep the spaces the same way
they were written.

## The expected behavior

We should also be able to generate nodes from HTML as it is when it comes from
trustful source (a backend, for example). This can be done by either removing
the logic of white space collapsing on `$generateNodesFromDOM`, or adding an
option to not try to change spaces, or split the implementation in two, one for
pasting and for regular use, as documented on [Serialization and De-serialization](https://lexical.dev/docs/concepts/serialization#html)

liaobinbin:
I think the problem should be in the part of the code 
```
const parser = new DOMParser(); 
const dom = parser.parseFromString(html, "text/html");
```
because at this point, the generated DOM has already removed any extra spaces.

augustobmoura:
The generated DOM doesn't collapse the spaces, they are still there, if you try to preview the HTML they will look collapsed, but that's just the default styling for it. If you have `white-space: pre-wrap`, like the editor already has it, they will show up.

liaobinbin:
I'm sorry, I saw that the spaces were missing in the debugger. This is a display error, the spaces still exist in the DOM object.

https://github.com/facebook/lexical/blob/dcebd0a72d595da2d3e6fe30533b864f505f1bcc/packages/lexical/src/nodes/LexicalTextNode.ts#L1166-L1182

I found this.

https://github.com/facebook/lexical/blob/dcebd0a72d595da2d3e6fe30533b864f505f1bcc/packages/lexical/src/nodes/LexicalTextNode.ts#L1131-L1137

augustobmoura:
@zurfyx, I didn't check on master yet, but I don't see how #4969 would fix this issue. HTML is being exported correctly, the only problem we have is with import, where we are stripping contiguous spaces, basically mutating the input for no reason. This issue should be re-opened.

augustobmoura:
[...]

PR #4969, opened 2023-09-05
Retain TextNode whitespace on export

zurfyx:
#4467 introduced an HTML import regression for trailing spaces. The root cause is that our exported HTML is not correct, the whitespaces should never be at the end or should be marked with the white-space property. This diff does the later which is equivalent to the top-most `white-space: pre-wrap`

It also fixes export for these other surfaces that implement whitespace parsing correctly (i.e. Apple Pages):

Before

https://github.com/facebook/lexical/assets/193447/9d5bb067-18f0-419b-b458-f29e84340fa8

After

https://github.com/facebook/lexical/assets/193447/4279c985-08b2-4020-8e7f-db0d204d96f0

Closes https://github.com/facebook/lexical/issues/4919
Closes https://github.com/facebook/lexical/issues/4954

### DIFF
--- a/packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
+++ b/packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
@@ -105,7 +105,7 @@ describe('HTML', () => {
       html = $generateHtmlFromNodes(editor, selection);
     });
 
-    expect(html).toBe('<span>World</span>');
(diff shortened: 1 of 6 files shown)


acywatson:
UTs failing.

>The root cause is that our exported HTML is not correct, the whitespaces should never be at the end or should be marked with the white-space property

Presumably this is because we chose to use whitespace: pre-wrap to convey these semantics on import in #4467

acywatson (approved):

PR #4977, opened 2023-09-07
Move isHTMLElement into core

zurfyx:
Adding this dependency onto core adds another 6.9K to core unnecessarily. Since it's just a couple very lightweight functions I'm moving them into core for now while retaining the export onto @lexical/utils for backward compatibility.

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -23,13 +23,15 @@ import {
   DEPRECATED_$isGridSelection,
   EditorState,
   ElementNode,
+  isHTMLAnchorElement,
+  isHTMLElement,
   Klass,
   LexicalEditor,
   LexicalNode,
 } from 'lexical';
 import invariant from 'shared/invariant';
 
-export {$splitNode};
+export {$splitNode, isHTMLAnchorElement, isHTMLElement};
 
 export type DFSNode = Readonly<{
   depth: number;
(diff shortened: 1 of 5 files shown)


acywatson:
> Adding this dependency onto core adds another 6.9K to core unnecessarily

If this is true, then our tree-shaking is broken. There's no reason why we should be bundling the entire utils package into the core because of two imports. I don't mind if we do this but I want to:

1. verify that we've checked that this actually reduces core bundle size (not sure if I trust the above size limit checker).

2. understand why tree-shaking is not working in this scenario, as this impacts every other package we bundle.

acywatson:
Actually two more thoughts:

1) the core should probably stay dependency-free anyway

2) could the tree-shaking issue have to do with not specifying @lexical/utils as a dependency in the lexical core package.json?

acywatson (approved):

zurfyx:
> > Adding this dependency onto core adds another 6.9K to core unnecessarily
> 
> If this is true, then our tree-shaking is broken. There's no reason why we should be bundling the entire utils package into the core because of two imports. I don't mind if we do this but I want to:
> 
> 1. verify that we've checked that this actually reduces core bundle size (not sure if I trust the above size limit checker).
> 2. understand why tree-shaking is not working in this scenario, as this impacts every other package we bundle.

I'm not sure tree-shaking will work on a per-file basis (not what I've seen so far) but internally this is not ideal, see https://fburl.com/diff/o2jlrdx2

ajax-fam-d:
Hi, after this pr LexicalRichText is failing `utils.isHTMLElement is not a function HeadingNode.exportDOM`

valdemarua:
@ajax-fam-d most likely you haven't updated all @lexical packages to the latest 0.12.2 version.

PR #4982, opened 2023-09-09
fix(lexical): ElementNode can not be resolved

### LINKED ISSUE #4981: Bug: ElementNode can not resolve in vscode
`const parentNode = node.getParentOrThrow()` then `parentNode.xxx` doesn't have autocomplete feature.

Lexical version: 0.12.2

## The current behavior
vscode can not resolve ElementNode  in LexicalNode.d.ts
动画

## The expected behavior
ElementNode  can be correctly resolved

liuweiGL:
close #4981

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -23,7 +23,7 @@ import {
   $isRootNode,
   $isTextNode,
   ElementNode,
-} from '.';
+} from './index';
 import {
   $getSelection,
   $isNodeSelection,


zurfyx:
Thanks for the PR! I'm unsure why this is needed (`.` is perfectly valid and can be resolved correctly by tsc), @acywatson do you have some thoughts on this since you commented on the issue?

acywatson:
> 

I commented my only thought lol. it works fine for me.

liuweiGL:
> Thanks for the PR! I'm unsure why this is needed (`.` is perfectly valid and can be resolved correctly by tsc), @acywatson do you have some thoughts on this since you commented on the issue?

When it used as dependency, the package.json#main attribute specifies the package entry file as `lexical.js`.

liuweiGL:
I found that this pr is incorrect and the package.json#types property should be configured

PR #4988, opened 2023-09-11
Fix comment plugin selection rect top position

vonsa:
Reproduction steps:
1. Open the lexical playground
2. Add more content so you can scroll down the page
3. Select some text after scrolling
4. Press the comment icon

Result:
Selection rect is not positioned correctly

Fix:
Take the scroll offset into account

### DIFF
--- a/packages/lexical-playground/src/plugins/CommentPlugin/index.css
+++ b/packages/lexical-playground/src/plugins/CommentPlugin/index.css
@@ -51,7 +51,7 @@ i.add-comment {
 
 .CommentPlugin_CommentInputBox {
   display: block;
-  position: fixed;
+  position: absolute;
   width: 250px;
   min-height: 80px;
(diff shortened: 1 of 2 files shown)


zurfyx:
Thanks for the PR! I can't seem to repro it, where/how does it repro for you? 

https://github.com/facebook/lexical/assets/193447/42b6ea76-0073-458e-9ad4-4d8b019d4c41

vonsa:
Sure, you can see it here:

https://github.com/facebook/lexical/assets/60012821/e690d23d-e87a-4b65-be8f-0a9136840e90

vonsa:
I've also added a fix for the fixed position comment box, as it seems weird to stay in the same position on scroll?

zurfyx (approved):
Ah, thanks for the video and the fix! LGTM

PR #5001, opened 2023-09-14
Add highlight to format types

zurfyx:
Missing type for a format we already support/store


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -91,7 +91,8 @@ export type TextFormatType =
   | 'highlight'
   | 'code'
   | 'subscript'
-  | 'superscript';
+  | 'superscript'
+  | 'highlight';
```
acywatson, line 95: Looks like you have it twice in this union

acywatson (approved):

PR #5002, opened 2023-09-14
Fix `insertNodes` and `insertParagraph`

### LINKED ISSUE #4295: Bug: Pasting text immediately after an inline element (e.g. link) causes the text and its following siblings to be moved into the inline element
Lexical version: 0.9.1

## Steps To Reproduce

1. Go to [Lexical Playground](https://playground.lexical.dev/)
2. Select some text, e.g. the word "playground"
3. Copy it to the clipboard
4. Position the cursor after the "Lexical website" link
5. Press Paste

## The current behavior

The pasted word ("playground") _and_ all the following text on that line end up being link-ified

https://user-images.githubusercontent.com/1232595/230831265-67ba23ab-c2fd-4767-ac23-93b56c5f762b.mp4

## The expected behavior

The pasted word should _not_ be link-ified and, at very least, the following text should not be link-ified either.

Note that this behavior reproduces with a non-collapsed selection too, provided the selection starts/ends at the end of an inline element (since for a non-collapsed selection we'll end up dropping the selection contents through a call to `removeText()`).

## Analysis

A bit of debugging points to the code here:

https://github.com/facebook/lexical/blob/933ec6ada2b2f25db807cb125de75952ea0f7fd3/packages/lexical/src/LexicalSelection.ts#L1409-L1412

Specifically, if the selection is at the start of a text run (anchor offset is zero) then we choose the previous sibling as the target and append the pasted text and its siblings into the previous sibling.

I'm not sure what the reasoning is for this logic but it produces surprising results for this particular case.

GermanJablo:
- This PR greatly simplifies the `insertNodes` and `insertParagraph` methods and resolves #4295
- Due to design errors in those methods that were modified, `insertNewAfter` in `LinkNode` was not "inserting a new link after", but rather a new block. This hack has been fixed and removed as it is no longer needed. ~~I've also added a to-do comment that `insertNewAfter` should not return `LexicalNode | null` but `this | null`. This would be easier to achieve now~~. Change of mind. Actually what would have been better is for `insertNewAfter` to be called `handleEnter`, since it is used only for that, and leads to confusion.
- I have moved the `$isBlock` function from `range-selection.ts` to `@lexical/utils`. I've marked it as internal use because I think maybe later some nodes should expose an isBlock() method. Topic for another PR.
- I have moved the `$getAncestor` function from `lexical-link/src/index.ts` to `@lexical/utils`. I have corrected a minor error that this function had and have also made it return a more precise type. I have used this function in a couple of places where you can see how it simplifies the code. Surely there is potential to apply it in many more places. A tip: often when using the `node.isInline` method what you are actually trying to achieve is `$getAncestor(node, $isBlock)`.


packages/lexical-link/src/index.ts
```diff
@@ -224,23 +228,16 @@ export class LinkNode extends ElementNode {
...
-      return linkNode;
-    }
-    return null;
+    const linkNode = $createLinkNode(this.__url, {
+      rel: this.__rel,
+      target: this.__target,
+      title: this.__title,
+    });
+    this.insertAfter(linkNode, restoreSelection);
+    return linkNode;
```
GermanJablo, line 240: Quoting from the PR description:

> Due to design errors in those methods that were modified, insertNewAfter in LinkNode was not "inserting a new link after", but rather a new block. This hack has been fixed and removed as it is no longer needed. I've also added a to-do comment that insertNewAfter should not return LexicalNode | null but this | null. This would be easier to achieve now.

packages/lexical-link/src/index.ts
```diff
@@ -450,9 +447,7 @@ export function toggleLink(
       const firstNode = nodes[0];
       // if the first node is a LinkNode or if its
       // parent is a LinkNode, we update the URL, target and rel.
-      const linkNode = $isLinkNode(firstNode)
-        ? firstNode
-        : $getLinkAncestor(firstNode);
+      const linkNode = $getAncestor(firstNode, $isLinkNode);
```
GermanJablo, line 450: One detail that I noticed the new `$getAncestor` function improved

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1041,19 +1038,19 @@ export class RangeSelection implements BaseSelection {
...
       } else if (part === '\t') {
-        nodes.push($createTabNode());
+        paragraph.append($createTabNode());
       } else {
-        nodes.push($createTextNode(part));
+        paragraph.append($createTextNode(part));
       }
     }
-    this.insertNodes(nodes);
+    this.insertNodes([paragraph]);
```
GermanJablo, line 1053: This is required by modifications to the `insertNodes` method

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1531,494 +1528,32 @@ export class RangeSelection implements BaseSelection {
...
+      if ($isElementNode(node)) {
+        currentBlock.insertAfter(node);
+        currentBlock = node;
+      }
+    });
+    if (!$isElementNode(nodes[0])) return false;
+    const prevLast =
+      nodes.length > 1 || firstBlock.isEmpty() ? currentBlock : firstBlock;
+    mergeBlocks(firstBlock, nodes[0]);
+    mergeBlocks(prevLast, lastBlock);
```
GermanJablo, line 1544: The algorithm consists of the following:
- Inserting a paragraph splits the selection in the block of the `anchor` (firsBlock) and in the inserted block (lastBlock).
- Paste all the paragraphs or blocks of the array in the middle of those 2 divided paragraphs.
- Finally I combine the first pasted block with `firstBlock` and the last one with `lastBlock`.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -3472,3 +3007,49 @@ export function DEPRECATED_$getNodeTriplet(
...
+
+  if (pointNode.getParentOrThrow().isInline()) {
+    splitElement(pointNode.getParentOrThrow());
+  }
+
+  pointNode = point.getNode();
+  const block = $getAncestor(pointNode, INTERNAL_$isBlock);
+  if (block) return splitElement(block);
+  return null;
+}
```
GermanJablo, line 3044: Basically the splitElement function is executed once, unless the anchor is inside an inline element (in which case it is executed twice).

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1531,494 +1528,32 @@ export class RangeSelection implements BaseSelection {
...
+    const firstBlock = $getAncestor(this.anchor.getNode(), INTERNAL_$isBlock)!;
+    const lastBlock = this.insertParagraph()!;
+    let currentBlock = firstBlock;
+    nodes.forEach((node) => {
+      if ($isElementNode(node)) {
+        currentBlock.insertAfter(node);
+        currentBlock = node;
+      }
+    });
+    if (!$isElementNode(nodes[0])) return false;
```
GermanJablo, line 1540: This is my current blocker.
If I put `if (!$isElementNode(nodes[0])) return false;` the function continues to execute even if the condition is true.
If I wrap the statement in brackets:
```
if (!$isElementNode(nodes[0])) {
   return false;
}
```
The function stops when the condition is true, but Jest gives me the following error:

 

>  packages/lexical/src/LexicalSelection.ts:1466:149 - error TS2345: Argument of type 'LexicalNode' is not assignable to parameter of type 'ElementNode'.
> 
>      1466 /* eslint-disable *//* eslint-disable-line */;oo_cv(`e18d1569_28`);/* eslint-disable *//* eslint-disable-line */oo_al(mergeBlocks(firstBlock, nodes[0 ]), `e18d1569_11`);

Aren't both options supposed to be equivalent?

packages/lexical/src/LexicalSelection.ts
fantactuka: Will it handle case when link is inside mark node (inline within inline)?

fantactuka:
Great PR, @GermanJablo! `insertNodes` needed rework for a while now :) 

I've had some tests for it in-progress on my local with attempt to get most of common permutations, I'll spend a bit more time on it to add other scenarios but in case you want to run it with your implementation, here're some examples:

https://gist.github.com/fantactuka/cf83eb3ce3999941baf6f31207a25559

There're annoying failures due to text-wrap inline style mismatch, so for the sake of experiment I [commented this out](https://github.com/facebook/lexical/blob/ad18d11b3768877ee8fa2926debb895617b97d27/packages/lexical/src/nodes/LexicalTextNode.ts#L604).

GermanJablo:
Great Maksim. I've been trying to find a way for all the tests to pass for 5 days straight. Right now I'm still with that.

Let's hope new edge cases don't appear with your tests 🤞 haha

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/ListsCopyAndPaste.spec.mjs
```diff
@@ -262,13 +262,13 @@ test.describe('Lists CopyAndPaste', () => {
...
             dir="ltr">
-            <span data-lexical-text="true">three</span>
+            <span data-lexical-text="true">two</span>
           </li>
           <li
             value="2"
             class="PlaygroundEditorTheme__listItem PlaygroundEditorTheme__ltr"
             dir="ltr">
-            <span data-lexical-text="true">two</span>
+            <span data-lexical-text="true">three</span>
```
GermanJablo, line 271: The expect was incorrect. This PR detected the bug and fixed it.

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/ListsCopyAndPaste.spec.mjs
```diff
@@ -534,10 +534,10 @@ test.describe('Lists CopyAndPaste', () => {
             dir="ltr">
             <span data-lexical-text="true">five</span>
           </li>
-          <li value="5" class="PlaygroundEditorTheme__listItem">
-            <span data-lexical-text="true">45</span>
-          </li>
         </ul>
+        <p class="PlaygroundEditorTheme__paragraph">
+          <span data-lexical-text="true">45</span>
+        </p>
```
GermanJablo, line 540: The result here could be three ways:
```js
// Option 1
- five45

// Option 2: Google Docs, Word, this PR
- five
45

// Option 3: Lexical current behavior
- five
- 45
```

Options 1 and 2 seem reasonable to me. I don't see much sense in the 3rd one. I chose to follow the behavior of Word and GDocs.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -194,20 +194,20 @@ describe('LexicalSelectionHelpers tests', () => {
...
       // insertLineBreak
       setupTestCase((selection, element) => {
-        selection.insertLineBreak(true);
+        selection.insertLineBreak();
 
         expect(selection.anchor).toEqual(
           expect.objectContaining({
             key: element.getKey(),
-            offset: 0,
+            offset: 1,
```
GermanJablo, line 202: All anchor and focus changes in this file are fixes for incorrect expects.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -1970,13 +1967,11 @@ describe('LexicalSelectionHelpers tests', () => {
             return;
           }
 
-          expect(() => {
-            selection.insertNodes([heading, $createTextNode('bar')]);
-          }).toThrow();
+          selection.insertNodes([heading, $createTextNode('bar')]);
         });
```
GermanJablo, line 1971: The previous insertNodes expected to receive an array with top-level elements with their descendants inside, or to receive the top-level + the descendants as elements of the array that were being inserted in the process. However, it did not work with a combination of both structures.

The new insertNodes is 'smarter', since each node in the array knows where to insert it even in mixed situations.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -1970,13 +1967,11 @@ describe('LexicalSelectionHelpers tests', () => {
...
 
         expect(element.innerHTML).toBe(
-          '<h1 dir="ltr"><span data-lexical-text="true">foo</span></h1>',
+          '<h1 dir="ltr"><span data-lexical-text="true">foobar</span></h1>',
         );
       });
     });
```
GermanJablo, line 1977: The previous insertNodes expected to receive an array with top-level elements with their descendants inside, or to receive the top-level + the descendants as elements of the array that were being inserted in the process. However, it did not work with a combination of both structures.

The new insertNodes is 'smarter', since each node in the array knows where to insert it even in mixed situations.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -2154,13 +2149,11 @@ describe('LexicalSelectionHelpers tests', () => {
...
-            selection.insertNodes([heading, $createTextNode('bar')]);
-          }).toThrow();
+          selection.insertNodes([heading, $createTextNode('bar')]);
         });
 
         expect(element.innerHTML).toBe(
-          '<p dir="ltr"><span data-lexical-text="true">Existing text...foo</span></p>',
+          '<p dir="ltr"><span data-lexical-text="true">Existing text...foobar</span></p>',
         );
       });
```
GermanJablo, line 2159: The previous insertNodes expected to receive an array with top-level elements with their descendants inside, or to receive the top-level + the descendants as elements of the array that were being inserted in the process. However, it did not work with a combination of both structures.

The new insertNodes is 'smarter', since each node in the array knows where to insert it even in mixed situations.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -2574,7 +2567,10 @@ describe('LexicalSelectionHelpers tests', () => {
         const element2 = $createTestElementNode();
         $insertNodes([element1, element2]);
       });
-      expect(element.innerHTML).toBe('<div><br></div><div><br></div>');
+      expect([
+        '<div><br></div><div><br></div>',
+        '<div><br></div><p><br></p>',
+      ]).toContain(element.innerHTML);
```
GermanJablo, line 2573: The decision of which block has priority when inserting nodes when one of the two (the one inserted or the one on which it is inserted) is empty, is arbitrary. For the purposes of this test, neither of these two results interferes with its objective.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1527,522 +1522,107 @@ export class RangeSelection implements BaseSelection {
...
-    }
-    const siblings = [];
-
-    // Get all remaining text node siblings in this element so we can
-    // append them after the last node we're inserting.
-    const nextSiblings = anchorNode.getNextSiblings();
-    const topLevelElement = $isRootOrShadowRoot(anchorNode)
-      ? null
-      : anchorNode.getTopLevelElementOrThrow();
+  insertNodes(nodes: Array<LexicalNode>) {
```
GermanJablo, line 1526: selectStart is no longer needed. Note that insertNodes is a selection method, which the first thing it does is delete the selected content. Where the selection should end up can only be one, and it is now handled automatically.

On the other hand, it is not necessary to return a boolean either. Insertion is expected to always be possible (see comment on the versatility of insertNodes in LexicalSelectionHelpers)

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1527,522 +1522,107 @@ export class RangeSelection implements BaseSelection {
...
-    // Time to insert the nodes!
-    for (let i = 0; i < nodes.length; i++) {
-      const node = nodes[i];
+    function mergeBlocks(
+      first: ElementNode,
+      second: ElementNode,
+      selection: RangeSelection,
+    ) {
+      // TO-DO: we have to reason generic behavior that is
+      // not hacks around list-items and codeblock.
```
GermanJablo, line 1556: Lists and code blocks have exceptions.

Ideally this can be improved. I temporarily used `'__language'` instead of `$isCodeNode` so that `@lexical/selection` does not depend on `@lexical/code`.

I think there are A LOT of tests and code that wouldn't be necessary if codeNode were a decoratorBlock. If that is the best solution or not I don't know.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1527,522 +1522,107 @@ export class RangeSelection implements BaseSelection {
...
-      }
-    }
-    if (selectStart) {
-      this.insertNodes([lineBreakNode], true);
-    } else {
-      if (this.insertNodes([lineBreakNode])) {
-        lineBreakNode.selectNext(0, 0);
-      }
-    }
+  insertLineBreak(): void {
```
GermanJablo, line 1624: see comment about selectStart above

packages/lexical/src/LexicalSelection.ts
GermanJablo: Good question. No. But right now it's not working on Lexical either. You can confirm it in the playground by pressing enter on a comment that involves a link. That's why it should be a job for another PR.

GermanJablo:
I've made quite a bit of progress on this, it's almost finished. All the unit tests have already passed, and only about 9 e2e tests fail. Of those 9, approximately half have to do with Prettier, I don't know what's wrong with him. I also just realized in CI as well that there is no support for ES2020 so integrity fails.

As you can see in the comments, this PR resolves a handful of bugs.

GermanJablo:
Only 4 failed tests locally. I'm almost there

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -173,8 +173,8 @@ export function $insertDataTransferForRichText(
   if (text != null) {
     if ($isRangeSelection(selection)) {
       const parts = text.split(/(\r?\n|\t)/);
-      const partsLength = parts.length;
-      for (let i = 0; i < partsLength; i++) {
+      if (parts.at(-1) === '') parts.pop();
+      for (let i = 0; i < parts.length; i++) {
```
GermanJablo, line 177: At one point there was a test that failed if an empty (unnecessary) string was added last.

packages/lexical/src/nodes/LexicalTabNode.ts
```diff
@@ -65,15 +65,6 @@ export class TabNode extends TextNode {
...
 
-  spliceText(
-    _offset: number,
-    _delCount: number,
-    _newText: string,
-    _moveSelection?: boolean | undefined,
-  ): TextNode {
-    invariant(false, 'TabNode does not support spliceText');
-  }
-
```
GermanJablo, line 76: @zurfyx Can you check if this causes a regression of PR #5066?
I'm not sure what that PR does, but this one is very much based on spliceText.

packages/lexical-code/src/CodeNode.ts
```diff
@@ -285,10 +282,21 @@ export class CodeNode extends ElementNode {
...
+      codeNode.splice(index, 0, nodesToInsert);
+      const last = insertNodes.at(-1);
+      if (last) last.select();
+      else if (anchor.offset === 0) split.selectPrevious();
+      else split.getNextSibling()!.selectNext(0, 0);
+    }
+    if ($isCodeNode(firstSelectionNode)) {
+      const {offset} = selection.anchor;
+      firstSelectionNode.splice(offset, 0, [$createLineBreakNode()]);
+      firstSelectionNode.select(offset + 1, offset + 1);
```
GermanJablo, line 299: `insertParagraph` calls `insertNewAfter`, and `insertNodes` calls `insertParagraph`. To simplify things and avoid bugs and circular references (very common while refactoring), CodeNode's `insertNewAfter` no longer depends on `insertNodes`.

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs
```diff
@@ -860,8 +860,13 @@ test.describe('CopyAndPaste', () => {
 
   test('Pasting a decorator node on a blank line inserts before the line', async ({
     page,
+    isCollab,
     isPlainText,
   }) => {
+    test.fixme(
+      isCollab,
+      'Seems to be failing on collab, at least since PR 5002',
+    );
```
GermanJablo, line 869: In my previous push all versions in collab failed. I don't know if it was already flaky from before or if I did it now.

packages/lexical-rich-text/src/index.ts
```diff
@@ -354,12 +354,15 @@ export class HeadingNode extends ElementNode {
...
-        : $createParagraphNode();
+      anchorOffet === this.getTextContentSize() || !selection
+        ? $createParagraphNode()
+        : $createHeadingNode(this.getTag());
     const direction = this.getDirection();
     newElement.setDirection(direction);
     this.insertAfter(newElement, restoreSelection);
+    if (anchorOffet === 0 && !this.isEmpty() && selection) {
+      this.replace($createParagraphNode(), restoreSelection);
+    }
```
GermanJablo, line 365: Previously `insertNewAfter` delegated to `insertParagraph` or `insertNodes` the responsibility of replacing the heading with a paragraph. By moving that responsibility here we can further simplify `insertNodes` which is already too complex on its own.

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/HTMLCopyAndPaste.spec.mjs
```diff
@@ -241,9 +239,9 @@ test.describe('HTML CopyAndPaste', () => {
     );
     await assertSelection(page, {
       anchorOffset: 16,
-      anchorPath: [2, 0, 0],
+      anchorPath: [1, 0, 0],
       focusOffset: 16,
-      focusPath: [2, 0, 0],
+      focusPath: [1, 0, 0],
     });
   });
```
GermanJablo, line 247: This could be considered a bug fixed, since there is no need to insert an empty paragraph

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/ListsHTMLCopyAndPaste.spec.mjs
```diff
@@ -390,27 +390,23 @@ test.describe('HTML Lists CopyAndPaste', () => {
...
           </li>
+        </ul>
+        <hr class="" contenteditable="false" data-lexical-decorator="true" />
+        <ul class="PlaygroundEditorTheme__ul">
           <li
-            value="2"
             class="PlaygroundEditorTheme__listItem PlaygroundEditorTheme__ltr"
-            dir="ltr">
+            dir="ltr"
+            value="1">
```
GermanJablo, line 399: another improvement to insertNodes.
It doesn't make sense for the hr to be inserted after the list-item the cursor is over, if the cursor is at the beginning.

GermanJablo:
Ok I did it!
My only doubt is that perhaps I caused a regression in #5066, as I asked in one of the comments.
I can confirm that this PR resolves #4295. I will try to add a test of this case in another PR.
Several bugs that were not reported were also solved, as can be seen in the modified tests.

Additionally, this PR lays the groundwork for greatly simplifying `insertText` and `removeText` (I already have an idea of how to do it in a few lines of code using my new `RemoveTextAndSplitBlock` primitive). I also think that with the new `insertNodes` many of the things that are done to normalize the nodes before pasting them are not going to be necessary. I will try to do all this in another PR.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1525,521 +1521,118 @@ export class RangeSelection implements BaseSelection {
...
       } else {
-        target = placementNode;
+        const index = RemoveTextAndSplitBlock(this);
+        firstBlock.splice(index, 0, nodes);
+        const last = nodes.at(-1)!;
+        if (last.select) last.select();
+        else last.selectNext(0, 0);
       }
+      return;
     }
```
GermanJablo, line 1542: The node insertion logic is different in code.

I used `'__language' in node` instead of `isCodeNode(node)` to prevent core from depending on @lexical/code. Something that may be more elegant is to create an `INSERT_NODES_COMMAND` to which `lexical/code` is hooked.
I am going to address this at https://github.com/facebook/lexical/pull/4900

zurfyx:
Thanks @GermanJablo, appreciate your hard work on this! `insertNodes` has always been a hotspot for bugs and incredibly difficult to debug, arguably the hardest function to follow in the entire Lexical codebase. For reference, we discussed this PR on our last sync and we have intention to merge it.

GermanJablo:
I'm glad to be of help!

I would appreciate any priority that can be given to this PR since keeping it updated with main can be difficult. If there's anything else I can do to move this forward, just let me know.

Thank you!

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1525,521 +1521,118 @@ export class RangeSelection implements BaseSelection {
...
-          target = anchorNode.getParentOrThrow();
-        }
-        siblings.push(anchorNode);
-      } else if (anchorOffset === textContentLength) {
-        target = anchorNode;
-      } else if (anchorNode.isToken()) {
-        // Do nothing if we're inside a token node
-        return false;
+    const notInline = (node: LexicalNode) =>
+      ($isElementNode(node) || $isDecoratorNode(node)) && !node.isInline();
```
zurfyx, line 1545: We have an inline utility already, should we use this?

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1525,521 +1521,118 @@ export class RangeSelection implements BaseSelection {
...
+    const insertedParagraph = shouldInsert ? this.insertParagraph() : null;
+
+    const last = nodes.at(-1)!;
+    const nodeToSelect = $isElementNode(last)
+      ? last.getLastDescendant() || last
+      : last;
+    const nodeToSelectSize = nodeToSelect.getTextContentSize();
+
+    let currentBlock = firstBlock;
+    nodes.forEach((node) => {
```
zurfyx, line 1566: Nit, let's not use `.forEach` in core, prefer `for of` or a classic `for (let..`

packages/lexical/src/LexicalSelection.ts
```diff
@@ -3470,3 +3060,35 @@ export function DEPRECATED_$getNodeTriplet(
   );
   return [cell, row, grid];
 }
+
+function RemoveTextAndSplitBlock(selection: RangeSelection) {
```
zurfyx, line 3064:
```suggestion
function removeTextAndSplitBlock(selection: RangeSelection): number {
```

packages/lexical/src/nodes/LexicalTabNode.ts
zurfyx: Do we need to remove this? Basically TabNode is a one-off node that it's `__text` should never be modified, it should be a static `\t`. Other text should be preprended or inserted after, this node can only render `\t`

zurfyx (approved):
Thanks, great work GermanJablo! Left some minor comments

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -173,8 +173,8 @@ export function $insertDataTransferForRichText(
   if (text != null) {
     if ($isRangeSelection(selection)) {
       const parts = text.split(/(\r?\n|\t)/);
-      const partsLength = parts.length;
-      for (let i = 0; i < partsLength; i++) {
+      if (parts.at(-1) === '') parts.pop();
```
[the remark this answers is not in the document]
GermanJablo: It seems to transpile well. The latest ES spec incorporated Array.last(), but I didn't try that one just in case.

I immediately add the braces, I forget out of habit and personal taste 😅

packages/lexical/src/LexicalSelection.ts
GermanJablo: If you mean $isInlineElementOrDecoratorNode it is different. If I use negation, TextNodes would be valid. Here I am looking for nodes that are not inline, but are of type Element or Decorator.

packages/lexical/src/nodes/LexicalTabNode.ts
GermanJablo: By adding this code the insertNodes logic does not work, and you would have to add an additional patch around TabNode. This does not cause any errors since in the case of TabNode the cursor will always be at the beginning or at the end. Maybe we can (1) throw an error if spliceText happens with offset === 1 here, or (2) tokenize tabnode?

packages/lexical-rich-text/src/index.ts
GermanJablo: I think the `code` case is different from this one. There you avoid insertNodes. Here you avoid burdening insertNodes with the responsibility of replacing or transforming nodes. 

Actually, everything would be clearer if `insertNewAfter` was called `handleEnter`.

packages/lexical-code/src/CodeNode.ts
GermanJablo: Remember that in insertNodes I am temporarily using the `__language in node` hack.

What I would like to do in another PR is an "INSERT_NODES_COMMAND" that is used in @lexical/code. Then this code would not be necessary either.

GermanJablo:
Thanks to you @zurfyx.  I already answered/resolved the comments!

GermanJablo:
Bug detected: you cannot paste a heading at the beginning of the document if it starts with another heading.
I'm working on fixing it

PR #5010, opened 2023-09-16
fix: LinkNode.importDOM should support all node content

### LINKED ISSUE #4533: Bug: LinkNode.importDOM doesn't support image content
loading the following HTML:
`<a href="https://facebook.com"></a>`
drops the A element.

this happens because of this:
https://github.com/facebook/lexical/blob/5595519202f9241e3e2b8ab3b240c787c58484e2/packages/lexical-link/src/index.ts#L286

Lexical version: 0.10

if I may suggest the fix could look like this:
`if ((content !== null && content !== '') || domNode.querySelector('img') !== null)`

k-rajat19:
fix #4533


zurfyx (approved):
Thank you! While there might false positives (i.e. empty Text nodes), I think that's a reasonable approach and we can iterate on this later when we encounter them

PR #5013, opened 2023-09-17
fix(Playground): CMD +K improvements

### LINKED ISSUE #4439: Playground CMD + K improvements
Follow up o f https://github.com/facebook/lexical/issues/4347 

CMD + K on a link should prompt you to edit it, CMD + K on a non-link should take you directly to editing it instead of just typing `https://`, as it's still not ideal for keyboard users.

k-rajat19:
improved UX :
CMD +K on a link prompts users to edit it and on a non-link takes them directly to edit it.
resolves #4439

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
@@ -36,17 +36,20 @@ function FloatingLinkEditor({
   isLink,
   setIsLink,
   anchorElem,
+  isLinkEditMode,
+  setIsLinkEditMode,
 }: {
   editor: LexicalEditor;
   isLink: boolean;
   setIsLink: Dispatch<boolean>;
   anchorElem: HTMLElement;
+  isLinkEditMode: boolean;
+  setIsLinkEditMode: Dispatch<boolean>;
 }): JSX.Element {
   const editorRef = useRef<HTMLDivElement | null>(null);
   const inputRef = useRef<HTMLInputElement>(null);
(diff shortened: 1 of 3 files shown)


zurfyx (changes requested):
Thanks for the PR! The way I envisioned this to work is to have CMD + K showing the floating menu, then focusing the link input so that you're able to type right away like GDocs or Microsoft Word where you don't have to lift your hands from the keyboard. Would you mind iterating on this?

k-rajat19:
> Thanks for the PR! The way we envisioned this to work in #4439 is to have CMD + K showing the floating menu, then focusing the link input so that you're able to type the link right away like GDocs or Microsoft Word where you don't have to lift your hands from the keyboard. Would you mind iterating on this?

yup that makes sense

packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -161,10 +164,10 @@ function FloatingLinkEditor({
   }, [editor, updateLinkEditor]);
 
   useEffect(() => {
-    if (isEditMode && inputRef.current) {
+    if (isLinkEditMode && inputRef.current) {
       inputRef.current.focus();
     }
-  }, [isEditMode]);
+  }, [isLinkEditMode, isLink]);
```
k-rajat19, line 170: added a fix here, now it's directly  focusing on the link input  when a user presses the button

k-rajat19:
@zurfyx  is there any thing to add?

zurfyx (approved):
LGTM, thank you!

PR #5019, opened 2023-09-19
Fix: ToolbarPlugin paste error with start, end format

khsily:
https://github.com/facebook/lexical/assets/30749436/80d97912-55e6-4a14-8df4-d192fe1fe802

This error was made in PR #4904, which caused an 'Uncaught TypeError: Cannot read properties of undefined (reading 'name')' error due to missing format types (start and end).

This PR will add start, end format options to alignment dropdown.
스크린샷 2023-09-19 오후 3 16 57


zurfyx (approved):
Thank you, LGTM!

PR #5024, opened 2023-09-20
Fix event handler removal in useDynamicPositioning, remove dupe code

thorn0:
If a handler was added as capturing, it must be removed as such too.


zurfyx (approved):
LGTM, thank you!

PR #5030, opened 2023-09-20
Port List e2e test to unit tests

acywatson:
WIP


packages/lexical/src/__tests__/unit/LexicalListPlugin.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+import * as ReactTestUtils from 'react-dom/test-utils';
+
+import {
+  INSERT_UNORDERED_LIST_COMMAND,
+  REMOVE_LIST_COMMAND,
+} from '../../../../lexical-list/src/index';
+
+describe('@lexical/list tests', () => {
+  let container: HTMLDivElement | null = null;
+  let reactRoot;
```
fantactuka, line 35: Not a blocker, but I wonder if we even need react here? I guess list plugin should have `registerList` or similar exposed, then there's just less of react-related noise in the code

acywatson: I thought about this, but the reason I ultimately kept it is because I want to be able to easily compose the plugins, many of which don't expose neat handles in the same way as @lexical/list. Most things in @lexical/react are exposed as react components, and those are the bulk of what I want to test so we can get rid of some e2es

zurfyx (approved):
That's very nice! ✨

fantactuka (approved):

Issue #3616, opened 2022-12-22
Can't be able to use `generateHtmlFromNodes` correctly.


eashish93:
When using `generateHtmlFromNodes`, I encountered an error: 
`To use $generateHtmlFromNodes in headless mode please initialize a headless browser implementation such as JSDom before calling this function.` 
which is fine as JSDOM is required, but this error doesn't mention which variables this function needs, so I dig into the code and found that we need to declare something like this first to make it work:
```js
const { JSDOM } = jsdom;

const window = new JSDOM(`...`).window;
global.document = window.document;
// @ts-expect-error
global.window = window;

// Sometimes `global.navigator` also needs to be defined with `global.window `

// @ts-expect-error
global.DocumentFragment = () => JSDOM.fragment(``);
```
This function is checking for both `document` and `window`. This should only check for `document` and `DocumentFragment` only, as these are the only methods `generateHtmlFromNodes` using.

Lexical version: 0.7.4

Multiply:
Same issue here. Overriding document, window etc. also seems to trip up NextJS's API runtime, so an option to pass in JSDOM as an override would be amazing.

sneko:
> Same issue here. Overriding document, window etc. also seems to trip up NextJS's API runtime, so an option to pass in JSDOM as an override would be amazing.

@thegreatercurve @zurfyx any plan to add this? Would avoid polluting the backend environment as mentioned in https://github.com/facebook/lexical/issues/3097#issuecomment-1511496254

Bunkerbewohner:
I also just stumbled over this. For context: I'm persisting the editor's collaboration state in form of the Ydoc using Hocuspocus and a custom persistence extension. On the server-side I'm storing/loading the Ydoc, and in addition I want to generate HTML from the Ydoc / editorstate so that I can server-side render an HTML preview of the editor before it's fully loaded.

Fritiof:
I can't even get the JSDOM solution to work, happy to get an example to look at if anyone can share!

Chigozie-Gerald:
> I also just stumbled over this. For context: I'm persisting the editor's collaboration state in form of the Ydoc using Hocuspocus and a custom persistence extension. On the server-side I'm storing/loading the Ydoc, and in addition I want to generate HTML from the Ydoc / editorstate so that I can server-side render an HTML preview of the editor before it's fully loaded.

[...]

Issue #3097, opened 2022-10-02
Bug: unable to run $generateHtmlFromNodes with headless editor


alexluong:
Lexical version: 0.5.0

## Steps To Reproduce

1. Install the necessary dependencies (lexical, @lexical/headless, @lexical/html, etc)
2. In a Node environment, generate HTML string from a lexical editor state JSON string. See reproduction link below.

Link to code example: https://stackblitz.com/edit/node-fsv6np

## The current behavior

With the above reproduction, when running the script this is the log from the terminal:

```
$ node index.js
markdown: hello world
```

## The expected behavior

```
$ node index.js
markdown: hello world
html: <p>hello world</p>
this should be logged
```

fantactuka:
Did you add any dom-like environment to handle dom creation? We [do not provide any built-in replacement](https://github.com/facebook/lexical/blob/main/packages/lexical-html/src/index.ts#L61) for node.js, so you'll need to include [jsDom](https://github.com/jsdom/jsdom) or similar to generate html on server

fantactuka:
To be fair, that document/window check does not work correctly, hence you don't see proper explanation. I've created #3102 to adjust that

agustaf9:
Adding this line worked for me.

`function DocumentFragment() {
  return JSDOM.fragment('')
}`

Negan1911:
@fantactuka It is possible to pass the window / document to headless so we don't check on the global objects?

Negan1911:
Not quite sure if y'all saw it, but on the HTML documentation there's a way to pass the dom as a 2nd parameter: https://github.com/jsdom/jsdom

konhi:
[This implementation ](https://github.com/2wheeh/lexical-nextjs-ssr/blob/6d97581def739cb5763b5a3a2d611f8873a3013e/src/lib/index.ts)solves the issue for Next

PR #5035, opened 2023-09-21
fix(lexical-html): allow dom to be passed to generateHtmlFromNodes

Fritiof:
This solves using this method in Next.js for example, that does not handle JSDOM well when on the server side. 

issue #3616 & #3097


Fritiof:
requires a lot more work to make it work than expected

PR #5041, opened 2023-09-24
Fix/floating formatter #5021

### LINKED ISSUE #5021: Bug: When I add a link I am unable to access the other floating menu options
When I add a link, I am unable to display the other options unless I remove the link. So, in this case, first, I need to apply stylings, and then I can add a link to see the formatted way of the link. Is there a way to display the other options with a link?

https://github.com/facebook/lexical/assets/127859730/581ba3e8-99a0-4757-b998-0ae88769fd65

Lexical version: "^0.12.0"

## Steps To Reproduce

1. Highlight a text
2. You will see the floating options. Select Link option.
3. Add a link to that text
4. When you try to add other options using floating toolbar. You are unable to do it

## The current behavior

When you try to add other options using floating toolbar to a link, you are unable to do it.

## The expected behavior

It should show other options of the floating toolbar.

shubham168:
fixes #5021 issue, format popup stays even after text is changed to valid link format

fix

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
@@ -124,9 +124,14 @@ function TextFormatFloatingToolbar({
     ) {
       const rangeRect = getDOMRangeRect(nativeSelection, rootElement);
 
-      setFloatingElemPosition(rangeRect, popupCharStylesEditorElem, anchorElem);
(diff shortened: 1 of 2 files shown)


shubham168:
Hey @zurfyx @fantactuka @acywatson please review this PR

zurfyx (approved):
Thank you, stamping for now but we might have to revise the UX in the future as this is unlikely to play well with a11y. What we have now is not a11y either so your proposal is a win.

PR #5044, opened 2023-09-25
Allow LexicalTypeaheadMenuPlugin to work when inside an iframe

lestertay:
Hi Lexical team, I'm making some changes to the `LexicalTypeaheadMenuPlugin` such that when we're using the ContentEditable in an iframe but rendering the dropdown menu of the typeahead outside (iframe's parent), we'll still calculate the position of the dropdown menu correctly.

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -46,7 +46,7 @@ function getTextUpToAnchor(selection: RangeSelection): string | null {
 }
 
 function tryToPositionRange(leadOffset: number, range: Range, editorWindow: Window): boolean {
-  const domSelection = _window.getSelection();
+  const domSelection = editorWindow.getSelection();
   if (domSelection === null || !domSelection.isCollapsed) {
     return false;
   }
@@ -236,8 +236,8 @@ export function LexicalTypeaheadMenuPlugin<TOption extends MenuOption>({
   useEffect(() => {
     const updateListener = () => {
       editor.getEditorState().read(() => {
-        const _window = editor._window ?? window;
-        const range = _window.document.createRange();
+        const editorWindow = editor._window ?? window;
+        const range = editorWindow.document.createRange();
         const selection = $getSelection();
         const text = getQueryTextForSearch(editor);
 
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -45,8 +45,8 @@ function getTextUpToAnchor(selection: RangeSelection): string | null {
   return anchorNode.getTextContent().slice(0, anchorOffset);
 }
 
-function tryToPositionRange(leadOffset: number, range: Range): boolean {
-  const domSelection = window.getSelection();
+function tryToPositionRange(leadOffset: number, range: Range, _window: Window): boolean {
```
fantactuka, line 48:
```suggestion
function tryToPositionRange(leadOffset: number, range: Range, editorWindow: Window): boolean {
```

fantactuka (approved):
Unrelated to this PR, but maybe we should add editor.getWindow() / editor.getDocument() with fallback to globals. It might simplify swapping globals with jsdom (e.g., for server rendering)

ivailop7:
> Unrelated to this PR, but maybe we should add editor.getWindow() / editor.getDocument() with fallback to globals. It might simplify swapping globals with jsdom (e.g., for server rendering)

I think, this is a great idea.

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -236,7 +236,8 @@ export function LexicalTypeaheadMenuPlugin<TOption extends MenuOption>({
   useEffect(() => {
     const updateListener = () => {
       editor.getEditorState().read(() => {
-        const range = document.createRange();
+        const _window = editor._window ?? window;
```
zurfyx, line 239: @lestertay @fantactuka - I wonder if `editor._window` should always be set to avoid the fallback to the `?? window` in getDefaultView that powers this setter https://github.com/facebookexternal/Outline//blob/02f01f160e8894f133e076ab5e38e7a3fd6a5f7a/packages/lexical/src/LexicalEditor.ts#L961-L967

lestertay: @fantactuka i think that'll be useful and a sensible fallback. I'll be happy to make another PR to enable this since i've seen quite a few places within Lexical using `editor._window`.

fantactuka: Sure, that was an idea for a follow up 👍

Issue #5037, opened 2023-09-22
Bug: Crash on older browsers


michaelsoultanidis:
<img width="1251" alt="Screenshot 2023-09-22 at 4 02 16 PM" src="https://github.com/facebook/lexical/assets/88332746/e4fcb6dd-c3bd-474f-a713-e715d1cce1f7">

Our project that uses Lexical as well Lexical Playground appears to crash when using older browsers such as Chromium 65 from 2018 and Firefox 57. I believe the reason of the crash is the lack of support for Optional Catch Binding in that particular browser. 

[The docs](https://lexical.dev/docs/getting-started/supported-browsers) state that Lexical should support these browsers.
Lexical version: Latest

## Steps To Reproduce

1. Download an older browser such as Chromium 65 or Firefox 57
2. Visit Lexical Playground

PR #5048, opened 2023-09-25
Transpile optional catch binding

zurfyx:
My understanding is that `"target": "ES2019",` should do the trick, but it doesn't.

This additional babel config does it but unsure why I need to import separately..

Closes https://github.com/facebook/lexical/issues/5037

Screenshot 2023-09-25 at 3 56 53 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
       },
       "devDependencies": {
         "@ampproject/rollup-plugin-closure-compiler": "^0.27.0",
+        "@babel/plugin-proposal-optional-catch-binding": "^7.18.6",
         "@babel/preset-flow": "^7.14.5",
         "@babel/preset-react": "^7.14.5",
         "@babel/preset-typescript": "^7.16.7",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -999,6 +1000,7 @@
       "version": "7.18.6",
       "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-optional-catch-binding/-/plugin-proposal-optional-catch-binding-7.18.6.tgz",
       "integrity": "sha512-Q40HEhs9DJQyaZfUjjn6vE8Cv4GmMHCYuMGIWUnlxH6400VGxOuwWsPt4FxXxJkC/5eOzgn0z21M9gMT4MOhbw==",
+      "deprecated": "This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-optional-catch-binding instead.",
```
acywatson, line 1003: Can we use the not deprecated one?

acywatson (approved):

PR #5051, opened 2023-09-25
Support for nullish & coalescent

zurfyx:
Nullish
Before: 94739 bytes
After: 94898 bytes

Coalescent
Before: 94739 bytes
After: 94924 bytes

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -89,9 +89,6 @@ module.exports = {
         'packages/lexical-playground/**',
         'packages/lexical-devtools/**',
       ],
-      rules: {
-        'lexical/no-optional-chaining': OFF,
-      },
     },
   ],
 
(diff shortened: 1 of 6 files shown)


fantactuka (commented):
So we got +0.2kb for 6 usages?

ivailop7:
I thought ?? made sense only when comparing with numbers where 0 would be a valid value and not consider it false, in the case with empty objects || is perfectly valid, no?

zurfyx:
Discussed this offline with @acywatson and @fantactuka. It seemed like a fine idea at first given it unblocks a much more ergonomic and safer API but the considerable size increase makes it unappealing when we strive for low bundle size. @ivailop7 let's retake the conversation on our next sync, parking this for now

Issue #5045, opened 2023-09-25
Bug: Triple-click selects link rather than paragraph when link is first child


kevinansfield:
When triple-clicking on a paragraph to select the whole paragraph if it starts with a link the selection collapses from the full paragraph down to just the link.

Lexical version: v0.12.2

## Steps To Reproduce

In the playground:

1. Add a link to the beginning of a paragraph
2. Triple-click anywhere in the paragraph after the link

https://github.com/facebook/lexical/assets/415/ca5cc954-b6e6-4ac9-8f2b-5789191fb476

## The current behavior

Selection collapses to just the link.

## The expected behavior

Whole paragraph is selected, matching the behaviour when there's no link at the beginning of the paragraph.

flavorjones:
I cannot reproduce this in Lexical v0.44.

https://github.com/user-attachments/assets/11e66ac2-7c21-4579-a4b2-51de8333bdb5

PR #5053, opened 2023-09-26
Fixed triple-click only selecting link when paragraph starts with a link

kevinansfield:
closes https://github.com/facebook/lexical/issues/5045

- modified the triple-click selection adjustment that removes unwanted trailing element to be aware of links as they create a deeper child/parent relationship than the previous code expected
- for now this commit uses a workaround of modifying the DOM selection directly because attempts to detect the paragraph element and use `paragraphNode.select(0)` resulted in an odd state where visually the selection looked correct and text-level formatting could be applied as expected but block-level changes like changing to a header had no effect


packages/lexical/src/LexicalEvents.ts
```diff
@@ -393,17 +393,35 @@ function onClick(event: PointerEvent, editor: LexicalEditor): void {
...
+              normalizedAnchorElement &&
+              normalizedAnchorElement !== normalizedFocusElement
+            ) {
+              // use DOM range because paragraphNode.select(0) results in a selection that
+              // visually appears to cover the link+text but can't have block-level formatting applied.
+              // FIXME: needs a deeper fix?
+              const range = document.createRange();
+              range.selectNodeContents(normalizedAnchorElement);
+              domSelection.removeAllRanges();
+              domSelection.addRange(range);
```
fantactuka, line 424: I wonder if we could instead look up for nearest non-inline element:

```
const selectableNode = $findMatchingParent(
  anchorNode,
  (node) => $isElementNode(node) && !node.isInline(),
);

if (selectableNode !== null) {
  selectableNode.select(0);
}
```

kevinansfield: For comparison...

State when creating selection via DOM:
```
 root
  ├ (11) paragraph  
> | ├ (8) link  "https://" { rel: noreferrer }
> | | └ (9) text  "One"
> | |              ^^^
> | └ (10) text  " two"
> |               ^^^^
  └ (4) paragraph  
    └ (5) text  "three"

 selection: range  
  ├ anchor { key: 9, offset: 0, type: text }
  └ focus { key: 10, offset: 4, type: text }
```

State when creating selection via `paragraphNode.select(0)`:

```
 root
  ├ (2) paragraph  
> | ├ (7) link  "https://" { rel: noreferrer }
> | | └ (3) text  "One"
> | └ (6) text  " two"
  └ (4) paragraph  
    └ (5) text  "three"

 selection: range  
  ├ anchor { key: 2, offset: 0, type: element }
  └ focus { key: 6, offset: 4, type: text }
```

Key difference being the anchor where it's the paragraph node with `.select(0)` but the beginning of the link's text node with dom selection. The latter works with block modifications where the former doesn't.

kevinansfield: What's the expectation here regarding selection - should `.select(0)` create something that resembles the DOM-derived selection, or should both of those selections be effectively the same in terms of further manipulation?

ivailop7: @fantactuka can you have second look here, when you have a chance

ivailop7:
This issue is still outstanding, would still be nice to fix.

PR #5056, opened 2023-09-26
Fix: right click select images

bencarletonn:
- When I right click on an image, I would expect for it to be selected, similar to how MS word and google docs behave. 
- Because the image is not currently selected when right clicking, none of the context menu commands work correctly. 

### **The current behavior** 
rightclickselect

### **The new expected behavior** 
rightclickselectfix


ivailop7 (approved):
Fantastic! Thank you, Ben! @zurfyx , @fantactuka , @acywatson , you happy with this one, before I merge this one?

fantactuka:
I would try avoiding adding new commands into core that are not used by core itself or plain/rich text plugins. We recently spotted few commands (e.g., CLEAR_HISTORY_COMMAND) that take bundle size, but not necessarily used (which is even more likely with context menu command).

bencarletonn:
> I would try avoiding adding new commands into core that are not used by core itself or plain/rich text plugins. We recently spotted few commands (e.g., CLEAR_HISTORY_COMMAND) that take bundle size, but not necessarily used (which is even more likely with context menu command).

I'm happy to move the context menu command away from the core into the images plugin, however, I'd argue it belongs in the core. There's many situations where it may be useful for a node (not just an images node) to execute some logic in response to a right click. Leaving it in the core increases this flexibility, at the expense of a few bytes.

ivailop7:
I agree with Ben, the core already has the dragover, dragend, etc. events included, in my opinion it makes sense contextmenu to be alongside them, the tradeoff is worth it.

acywatson:
I'm not clear on why we can't do this in userland by just listening for this event. Maybe we should think about removing more of these commands from the core. Listening directly to DOM events is not a bad pattern, imo.

ivailop7:
Ok, we'll move it to outside of core...

bencarletonn:
moved away from the core, apologies for the delay

acywatson (approved):

ivailop7 (approved):
Thanks, Ben! @fantactuka , @zurfyx happy with this one?

Issue #4857, opened 2023-08-03
Bug: Issue in Add Column shortcut in Experimental table


sathishsridhar:
On Hovering a table created using the Experimental version of the table plugin, shortcuts for adding the column and row appears. I am able to hover over the row shortcut and click it to add a new row but unable to do the same for column because the add column shortcut disappears when the mouse pointer leaves the table.

This issue can be reproduced in the Lexical Playground itself.

## The current behavior

https://github.com/facebook/lexical/assets/21175404/a3700168-441d-4dd6-811b-e497278b0d28

acywatson:
This is experimental. We don't support it and won't be addressing any bugs related to it at this time.

PR #5058, opened 2023-09-27
fix : Add column shortcut gets hidden when mouse leave the table #4857

sathishsridhar:
Though the feature is experimental, incase if you need a fix for this, kindly check out the PR. Please close them if its not necessary. 

 https://github.com/facebook/lexical/issues/4857#issuecomment-1664242477
 
Before Fix:

https://github.com/facebook/lexical/assets/21175404/54834861-7d0f-4dec-8e06-bca5ff25db64

After Fix:

https://github.com/facebook/lexical/assets/21175404/d53c9548-12b6-4078-910b-8e8929b07975

### DIFF
--- a/packages/lexical-playground/src/nodes/TableComponent.tsx
+++ b/packages/lexical-playground/src/nodes/TableComponent.tsx
@@ -1717,10 +1717,10 @@ export default function TableComponent({
 
   return (
     <div
-    style={{
-      position: 'relative',
-      width: (tableRef.current?.clientWidth || 0) + 'px',
-    }}>
+      style={{
+        position: 'relative',
+        width: (tableRef.current?.offsetWidth || 0) + 'px',
+      }}>
       <table
         className={`${theme.table} ${isSelected ? theme.tableSelected : ''}`}
         ref={tableRef}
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/nodes/TableComponent.tsx
```diff
@@ -1716,7 +1716,11 @@ export default function TableComponent({
   }
 
   return (
-    <div style={{position: 'relative'}}>
+    <div
+    style={{
+      position: 'relative',
+      width: (tableRef.current?.clientWidth || 0) + 'px',
```
fantactuka, line 1722: Not a big difference here, but `offsetWidth` is slightly better to ensure width includes border width. Otherwise looks good!

Please run prettier, as integrity test is failing

fantactuka (approved):

packages/lexical-playground/src/nodes/TableComponent.tsx
sathishsridhar: Thanks for the suggestion, changed it to offsetWidth and ran prettier.

PR #5070, opened 2023-09-29
support multiple windows

robfig:
This commit fixes 3 issues

1. The lastTextInputTimeStamp was not being updated on
   popout windows because the listener was not being added.

2. The time origin of event.timestamp and performance.now
   was not the same without using the specific window.

3. Event classes (KeyboardEvent, ClipboardEvent) are defined
   on each window, so instanceof checks need to check against
   the one defined on the window where the event originates.
   I converted these to use e.g. "clipboardData" in event instead.

I enountered these bugs in our Electron app.  It opens a new window
via window.open() and renders to it from the parent window. As a
result, there are multiple window objects that can host Lexical
editors and one JS environment.

I haven't found much documentation on multi-window setups online
besides this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms


packages/lexical/src/LexicalMutations.ts
```diff
@@ -39,19 +39,27 @@ import {
 const TEXT_MUTATION_VARIANCE = 100;
 
 let isProcessingMutations = false;
-let lastTextEntryTimeStamp = 0;
+const lastTextEntryTimeStamp_ = new WeakMap();
```
zurfyx, line 42: What are the practical use cases of tracking entry timestamp by window?

robfig: `event.timeStamp` is milliseconds since the time origin of the window, as is `performance.now()`, it is NOT a unix timestamp or wall time.. I found that surprising too.

Different windows have different time origins, so the timestamp of an events in different windows are not comparable. As a result, we need separate timestamps per window.

robfig: Hi @zurfyx do you mind reviewing this PR? It's not so many lines of code and fixes severe bugs for us. I have our version patched but would be great to remove the patch and be able to upgrade easily again. Thanks

robfig:
@zurfyx @acywatson bump on this PR. Any feedback? Maintaining patches like this makes it impossible for us to routinely update Lexical

packages/lexical/src/LexicalEvents.ts
```diff
@@ -162,7 +162,7 @@ let lastKeyDownTimeStamp = 0;
 let lastKeyCode = 0;
 let lastBeforeInputInsertTextTimeStamp = 0;
 let unprocessedBeforeInputData: null | string = null;
-let rootElementsRegistered = 0;
+const rootElementsRegistered = new WeakMap();
```
zurfyx, line 165:
```suggestion
const rootElementsRegistered = new WeakMap<Document, number>();
```

packages/lexical/src/LexicalEvents.ts
```diff
@@ -1136,12 +1136,13 @@ export function addRootElementEvents(
 ): void {
   // We only want to have a single global selectionchange event handler, shared
   // between all editor instances.
-  if (rootElementsRegistered === 0) {
-    const doc = rootElement.ownerDocument;
+  const doc = rootElement.ownerDocument;
+  const refs = rootElementsRegistered.get(doc) ?? 0;
+  if (!refs) {
```
zurfyx, line 1141:
```suggestion
  if (documentRootElementsCount === undefined) {
```

packages/lexical/src/LexicalEvents.ts
```diff
@@ -1136,12 +1136,13 @@ export function addRootElementEvents(
 ): void {
   // We only want to have a single global selectionchange event handler, shared
   // between all editor instances.
-  if (rootElementsRegistered === 0) {
-    const doc = rootElement.ownerDocument;
+  const doc = rootElement.ownerDocument;
+  const refs = rootElementsRegistered.get(doc) ?? 0;
```
zurfyx, line 1140:
```suggestion
  const documentRootElementsCount = rootElementsRegistered.get(doc) || 0;
```

packages/lexical/src/LexicalEvents.ts
```diff
@@ -1240,13 +1241,13 @@ export function addRootElementEvents(
 }
 
 export function removeRootElementEvents(rootElement: HTMLElement): void {
-  if (rootElementsRegistered !== 0) {
-    rootElementsRegistered--;
-
+  const doc = rootElement.ownerDocument;
+  const refs = rootElementsRegistered.get(doc) ?? 0;
+  if (refs) {
```
zurfyx, line 1246:
```suggestion
  const documentRootElementsCount = rootElementsRegistered.get(doc);
  invariant(...);
```

zurfyx (commented):
Thanks for the PR! Nits aside, I'd like to understand your use case better we merge this PR as this can have implications in other core modules that rely on window and document.

My understanding is that each Electron window runs in a separate renderer process with its own DOM. If that's the case, Lexical is initialized once per window, when does Lexical need to know about differents documents? Do you mind sharing some additional context and code snippets?

robfig:
Thank you for the review. I will look at merging in your comments and rebasing to get it current.

We have a multi-window Electron app, and we began by having each window be its own renderer process. That turned out to cause performance issues, and we found that single process + multiple windows was much faster / lighter. I'm afraid I don't have more information to share about it beyond what was in the description ^. The app I'm referring to is https://ro.am/

packages/lexical/src/LexicalEvents.ts
robfig: These two suggestions seem at odds^. The resulting condition would never occur

```
const documentRootElementsCount = rootElementsRegistered.get(doc) || 0;
if (documentRootElementsCount === undefined) {
   ...
```

robfig:
@zurfyx I rebased the PR on main and applied your comments in https://github.com/facebook/lexical/pull/5070/commits/0e141f9a19d7626422fd43f2f100a1f042584be0

zurfyx:
Thank you for iterating on this! Seems like this test is consistently failing? `[chromium] › packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/TablesHTMLCopyAndPaste.spec.mjs:243:3 › HTML Tables CopyAndPaste › Copy + paste - Merge Grids`

robfig:
Thanks @zurfyx , it was because of a mistake I made when rebasing. Just pushed a commit which should fix the test. Thanks

packages/lexical-rich-text/src/index.ts
```diff
@@ -435,10 +436,14 @@ function onPasteForRichText(
...
-        event instanceof InputEvent || event instanceof KeyboardEvent
+        objectKlassEquals(event, InputEvent) ||
+        objectKlassEquals(event, KeyboardEvent)
           ? null
-          : event.clipboardData;
-      if (clipboardData != null && selection !== null) {
+          : (event as ClipboardEvent).clipboardData;
+      if (
+        clipboardData != null &&
+        ($isRangeSelection(selection) || $isTableSelection(selection))
```
zurfyx, line 445: Why do we need to specify the type of selection now? We don't want to import `@lexical/table` here, we intentionally isolated this module so that you can only pay the bundle size cost when the product needs tables

robfig:
Hi @zurfyx , gentle ping that this PR is fully updated for all comments you've made

acywatson (approved):
I talked to Gerard and he's good with this

robfig:
Woo hoo! Thanks guys

PR #5071, opened 2023-09-29
chore: add typedefs for collab nodes and remove unused members

zignis:
Augment typedefs for yjs (collab nodes), and removed unused `_unobservers` member from the collab decorator node


zurfyx (approved):
LGTM, but the unobservers thing looks like a thing that could've been useful for some product use cases, @fantactuka do you have more context on that?

fantactuka:
> LGTM, but the unobservers thing looks like a thing that could've been useful for some product use cases, @fantactuka do you have more context on that?

Should be good, never seen it used

fantactuka (approved):

Issue #4340, opened 2023-04-17
Bug: When deleting text with the keyboard of a device it's also removing text to the right of the pointer.


dist159:
In the editor when doing a long press of the delete key in an android device it removes the text to the right of the text pointer.

Lexical version: v0.9.2

## Steps To Reproduce

Tested device: Samgung Galaxy S22 Ultra with default keyboard

1. Go to the lexical playground https://playground.lexical.dev in the browser of an android device
2. Delete the predefined content and write a random paragraph 
3. Select in the text any point where there is still text to the right
4. keep pressing the delete key in the keyboard

## The current behavior

It deletes text to the left and right of the pointer

https://user-images.githubusercontent.com/59893365/232548029-4b6f358a-7955-475d-a83b-15948bb577cc.mp4

## The expected behavior

It should only delete text to the left of the pointer.

https://user-images.githubusercontent.com/59893365/232548647-52290174-df6f-411a-93ef-a035e819d340.mp4

Reza-tm:
we are using lexical in production and we have the same problem 🥲 .
For now we decided to use html textarea instead of lexical editor.
We love lexical and I hope this problem will be solved soon so that we can return to lexical sooner

kevinansfield:
We're seeing the same issue too, no long-press on the backspace key needed.

cristopher-iov:
@trueadm I did a little research on this issue and found the reason why it happens.

In this [PR](https://github.com/facebook/lexical/pull/2940) that line now returns `false`, I don't really understand why it could have changed but that causes it to not enter that if and go to else.

I was testing a bit and giving a value to the `_compositionKey` property fixes it.

```ts
if ($isRangeSelection(selection)) {
  // Used for handling backspace in Android.
  const isAndroidDevice = !!navigator.userAgent.match(/Android/i);

  if (isAndroidDevice) {
    $setCompositionKey(selection.anchor.key);
  }

  if (
    isPossiblyAndroidKeyPress(event.timeStamp) &&
    editor.isComposing() &&
    selection.anchor.key === selection.focus.key
  ) {
    $setCompositionKey(null);
    lastKeyDownTimeStamp = 0;
    // Fixes an Android bug where selection flickers when backspacing
    setTimeout(() => {
      updateEditor(editor, () => {
        $setCompositionKey(null);
      });
    }, ANDROID_COMPOSITION_LATENCY);
    if ($isRangeSelection(selection)) {
      const anchorNode = selection.anchor.getNode();
[...]

Issue #4941, opened 2023-08-30
Bug: Hitting backspace deletes two characters on Android 


fredthebuilder:
Lexical version: 0.11.3 & 0.12.0 (also on https://playground.lexical.dev/)

## Steps To Reproduce

1. Open https://playground.lexical.dev/ on Chrome on Android
2. Write a word
3. Put the cursor somewhere in the middle
4. hit backspace - two characters get deleted

## The current behavior

https://github.com/facebook/lexical/assets/143589809/3660f9cc-19eb-4764-8b03-3c9a2549e089

(results in HelloWorld -> Hellorld instead of HelloWorld -> HellWorld)

Two characters get deleted

## The expected behavior

Only the character before the cursor gets deleted

moy2010:
This is a known issue: https://github.com/facebook/lexical/issues/4340

kirandash:
> In this thread there's a solution which is not perfect but it kinda works https://github.com/facebook/lexical/issues/3896
> 
> 
> 
> I created a patch using `patch-package` where I commented out the `else` block as described in one of the comments. However it produces couple of really weird side effects (which I believe are connected):
> 
> - When you enter just one character and try to delete it - it simply doesn't do it on Android, cursor moves to the beginning of the string leaving you with that letter being intact
> 
> - When you try do select all and delete - again doesn't do anything, except of moving cursor to the beginning of the string
> 
> - When you try to remove mention nodes and have a custom handling for this scenario - it doesn't do so correctly, hard to explain but maybe it's just my case
> 
> 
> 
> I'm currently working on finding a workaround, but so far no progress. I'm thinking of just adding a "trash icon" near the editor on all Android devices to help users clear up data in case they can't achieve it with backspace

Hi @kazimirchuk did you manage to find a fix for this? The current fix from lexical produces these weird side effects as you mentioned which is a bad UX at the moment

ivailop7:
Yes, it's merged, but we haven't published a lexical version since, so the fix is not out yet. That said, this is more of a short-term patch, until the actual Chromium mobile bug gets sorted out.

kirandash:
Thanks for your response @kazimirchuk and @ivailop7 🙏🏻 Much appreciated 🙂

Unfortunately this issue is a blocker for me now since the editor on my project is for end users especially on mobile device. I am curious how the facebook team managed to patch this for the post editor.

kirandash:
[...]

Issue #1973, opened 2022-04-25
Bug: Hitting backspace on android removes both characters which are adjacent to the caret


StormVanDerPol:
When you have a collapsed selection, aren't in composition mode, and you hit backspace in the middle of a word using Android Keyboard apps like Gboard or Microsoft SwiftKey, both characters adjacent to the caret will be removed, instead of just the character prior to the caret.

Video for demonstration:

https://user-images.githubusercontent.com/56631072/165106914-f5655cdc-9079-4c87-928c-7bbd742a294d.mp4

Lexical version: 0.2.4
Android: 12

## Steps To Reproduce

1. go to https://playground.lexical.dev/
2. place your caret in any word, e.g. "Playground", between letter "o" and "u"
3. Hit backspace
4. It should now say "Playgrnd" instead of "Playgrund"

## The current behavior
Hitting backspace removes both characters adjacent to the caret.

## The expected behavior
Hitting backspace removes only the character prior to the caret.

aearly:
I can reproduce this using the standard Android keyboard in Android Chrome.  Android Firefox does not have the same behavior.

StormVanDerPol:
@trueadm seems like a recent change caused a regression. Just had some people complain describing this bug. Tested on my pixel 5 and sadly it's happening again.
Can we re-open this?

trueadm:
@StormVanDerPol Would you mind helping me find what version of Lexical was the culprit? Was it 0.7? I wonder if it might have been https://github.com/facebook/lexical/pull/3434.

trueadm:
I wonder if a change in Chromium may have caused this then? :O

StormVanDerPol:
@trueadm we're still on 0.6.4. I can reproduce it in the latest playground however.

I went around and tested a bunch of older vercel previews and seems this regression happened quite some time ago, sorry for not catching it earlier.

Everything works as intended in PR #2412, which was your fix. However it already didn't work in the relevant release 0.3.4 (#2443).

I also tried #2416 which edits the same bit of code and here too it is broken, maybe it's related

trueadm:
This was definitely working as intended until recently. I have a feeling that https://github.com/facebook/lexical/pull/2416 may have "broken" the fix in the latest Chromium changes relating to the selection. 

With the duplication occurring, I'm guessing the logic to call `event.preventDefault();` isn't working?

trueadm:
So what I think might be happening, is that a recent change in Chromium has caused this logic:

[...]

PR #5077, opened 2023-10-03
fix: fix android backspace bug

cristopher-iov:
**Context:**

When this [line](https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalEvents.ts#L520) `editor.isComposing()` is evaluated, the value returned is always `false`, so the code within the `if` is not executed and continues to the `else`.

This causes deleting a character on an Android device to delete the character on the right.

I was testing a bit and giving a value to the `_compositionKey` property fixes it.

I don't know if it's the best solution, but this could help or be a guide.

**Issue:** https://github.com/facebook/lexical/issues/4340

**Evidence:** https://www.loom.com/share/24898783cffd4023bf85460966702898

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -515,6 +515,12 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
 
       if ($isRangeSelection(selection)) {
         // Used for handling backspace in Android.
+        const isAndroidDevice = !!navigator.userAgent.match(/Android/i);
+
+        if (isAndroidDevice) {
+          $setCompositionKey(selection.anchor.key);
+        }
+
         if (
           isPossiblyAndroidKeyPress(event.timeStamp) &&
           editor.isComposing() &&
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -13,6 +13,7 @@ import type {TextNode} from './nodes/LexicalTextNode';
 
 import {
   CAN_USE_BEFORE_INPUT,
+  IS_ANDROID,
   IS_APPLE_WEBKIT,
   IS_FIREFOX,
   IS_IOS,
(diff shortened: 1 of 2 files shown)


ivailop7:
These platform specific bugs always worry me, I wonder if a deeper bug is happening here, but don't know enough about mobile browsers and why this is not happening for iOS for example. Anyhow, please create IS_ANDROID in here: https://github.com/facebook/lexical/blob/main/packages/shared/src/environment.ts and use the variable from there in here.

alexgleason:
Thank you for creating this PR @cristopher-iov 

My users are annoyed by this bug. Merge?

ivailop7 (approved):

bilobom:
@ivailop7 @alexgleason 
This PR is not the optimal solution , when there is one character left at any line it wont delete it and it will jump to the begining of line without deleting the character.

Evidence:
https://github.com/facebook/lexical/assets/19841175/50d93d4c-f6db-474a-bfb0-b63280e587da

alexgleason:
The actual underlying issue here is a bug in Chromium Android, right?

Your comment here is very helpful: https://github.com/facebook/lexical/issues/1973#issuecomment-1753996678

Chromium bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1403455#c2

This MR is still mostly an improvement, but frustrating that mobile Chromium sucks at this. It's a big portion of my userbase.

ivailop7:
Not great, let's see when this bug Chromium Android bug gets fixed and we'll revisit this. Don't know what the right answer is here.

kirandash:
> @ivailop7 @alexgleason This PR is not the optimal solution , when there is one character left at any line it wont delete it and it will jump to the begining of line without deleting the character.
> 
> Evidence: [facebook/lexical/assets/19841175/50d93d4c-f6db-474a-bfb0-b63280e587da](https://github.com/facebook/lexical/assets/19841175/50d93d4c-f6db-474a-bfb0-b63280e587da)

I was able to fix it with these code changes: https://github.com/facebook/lexical/issues/4941#issuecomment-1818305883
Hope it helps. If anyone has any feedback I will be happy to investigate further 🙏🏻

PR #5078, opened 2023-10-03
Set up GA4 for website

zpao:
Summary:
This was never set up. It's easy enough and we should have some analytics.

Test Plan:
`npm run build` & diff before & after output. The only change was adding the gtag code.


acywatson:
image

acywatson (approved):

Issue #4491, opened 2023-05-12
Bug: `$getSelectionStyleValueForProperty(selection, 'font-size')` not updating correctly


jonesguy14:
When using `$getSelectionStyleValueForProperty(selection, 'font-size')`, it does not seem to update if user changes font size on some text which has previously modified its font size. The `font-family` property also exhibits the same behavior.

Lexical version: 0.10.0

## Steps To Reproduce

This can be reproduced in the https://playground.lexical.dev/:

1. Open lexical playground
2. Change font size for some text
3. Click away from the text
4. Highlight that same text again, and change the font size

I've also created a sandbox, see FontSizePlugin: https://codesandbox.io/s/dazzling-visvesvaraya-8pjcv5

Repro in sandbox by:
1. Type some text and highlight it
2. Change font size to 24px
3. Deselect text and reselect
4. Change font size to 48px

See that the text font size does change, but the `getSelectionStyleValueForProperty` result still says '24px'.

## The current behavior

Toolbar does not update, but text font size does change. `getSelectionStyleValueForProperty` seems to still return the original value.

## The expected behavior

`getSelectionStyleValueForProperty` will return the new font size, so that the toolbar can update to new font size.

This only happens for text that has font size changed, the initial change from no font-size to some value does seem to update correctly.

GAURAV1-ui:
Can I work on this issue.

acywatson:
> Can I work on this issue.

Sure - assigned to you

jonesguy14:
This also reproduces for the `font-family` property, updated in description. Hoping the fix will be for all `$getSelectionStyleValueForProperty` property types.

greeneley:
This issue also happens with me. Select a text -> change the font-size -> de-select text -> re-select this text -> change the font-size -> I see the toolbar doens't update the according font-size although the text font size in editor does change.
My using lexical version is 0.11.1

greeneley:
Hi @GAURAV1-ui,
How about the pull request ?

MaximMukhanov:
@GAURAV1-ui 
It seems to be a problem with style property on selection, it seems to not update on style change.
Hope it helps!

XFG16:
Is there any fix to this as of now?

jonesguy14:
@GAURAV1-ui any progress here?

@acywatson can this get unassigned/reassigned if not?

jonesguy14:
[...]

PR #5082, opened 2023-10-04
Fix getting style via $getSelectionStyleValueForProperty when selection is collapsed

Piliuta:
Fixes https://github.com/facebook/lexical/issues/4491

**Note:** This PR focuses on a fix inside `$getSelectionStyleValueForProperty` and **not** on the RangeSelection style property (which I believe should be a long-term fix).

## Context:

`style` property on the RangeSelection was implemented in https://github.com/facebook/lexical/pull/3863

however, it was not implemented for the case when selection is **not** collapsed

https://github.com/facebook/lexical/blob/c09ab7314fba32f0246a24df35c5d64061541978/packages/lexical/src/LexicalEvents.ts#L306-L350

Since `$getSelectionStyleValueForProperty` relies on this property, it does not return the correct value in cases when the selection is not collapsed.

### Problem
This method is often used for reflecting current values on toolbars, but it does not work in half (range selection) cases.

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -538,7 +538,7 @@ export function $getSelectionStyleValueForProperty(
   const endOffset = isBackward ? focus.offset : anchor.offset;
   const endNode = isBackward ? focus.getNode() : anchor.getNode();
 
-  if (selection.style !== '') {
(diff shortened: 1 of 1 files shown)


acywatson (approved):
Thanks for the fix! Sorry I missed all the comments on that issue.

PR #5083, opened 2023-10-04
MarkdownExport: fix export for overlapping styles

### LINKED ISSUE #4895: Bug: Markdown export produces invalid markdown for overlapping formats
## Steps To Reproduce

1. In Lexical Playground, enter a string with overlapping formats, such as "helloworld!" formatted in normal, bold, bold+italic, italic, normal as shown here:

```
 root
  └ (53) paragraph  
    ├ (48) text  "he"
    ├ (49) text  "llo" { format: bold }
    ├ (50) text  "wor" { format: bold, italic }
    ├ (51) text  "ld" { format: italic }
>   └ (52) text  "!"
```

2. Click Export to markdown. 

Actual: `he**llo*wor**ld*!`
Expected: `he**llo**___wor___*ld*!`, or any other valid encoding according to CommonMark

Copy those strings here to see the difference
https://spec.commonmark.org/dingus/

I'm not sure if any markdown parser supports overlapping formats as they are produced by the Markdown export. 
CommonMark says they are not valid: 

> When two potential emphasis or strong emphasis spans overlap, so that the second begins before the first ends and ends after the first ends, the first takes precedence. Thus, for example, `*foo _bar* baz_` is parsed as `<em>foo _bar</em> baz_` rather than `*foo <em>bar* baz</em>`.

Perhaps Markdown export should produce markdown for each TextNode at a time, or each set of consecutive TextNodes that share a common format, instead of treating the individual formats separately.

robfig:
Previously, exporting text with "layered" formats resulted in layered
markdown. For example, this string:

     root
      └ (53) paragraph
        ├ (48) text  "he"
        ├ (49) text  "llo" { format: bold }
        ├ (50) text  "wor" { format: bold, italic }
        ├ (51) text  "ld" { format: italic }
    >   └ (52) text  "!"

Would be exported like this:

    he**llo*wor**ld*!

The export incrementally applied the italic to a portion of a bold span.
This seems to not be valid according to Commonmark, although micromark
supports it to some extent (but not fully -- it breaks on strikethroughs
for me).

This change closes and opens markdown tags every time the format
changes between text nodes. It now exports that example like this:

    he**llo*****wor****ld*!

While this uses a few more characters, it will work in more
circumstances, it fixes the specific issue I was having, and it matches
Draft.js's export behavior, anecdotally.

Fixes #4895

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -12,7 +12,7 @@ import type {
   TextMatchTransformer,
   Transformer,
 } from '@lexical/markdown';
-import type {ElementNode, LexicalNode, TextFormatType, TextNode} from 'lexical';
+import type {ElementNode, LexicalNode, TextNode} from 'lexical';
 
 import {
   $getRoot,
@@ -22,6 +22,7 @@ import {
   $isTextNode,
 } from 'lexical';
 
+import {TEXT_TYPE_TO_FORMAT} from '../../lexical/src/LexicalConstants';
 import {transformersByType} from './utils';
 
 export function createMarkdownExport(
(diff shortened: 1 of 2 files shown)


robfig:
Oh, I didn't realize, but this change to export produces markdown that fails to import. Hm

potatowagon:

PR #5087, opened 2023-10-05
Update anchor selection if split by patching style

Piliuta:
## Problem

Setting styles using `$patchStyleText` does not update the selection when the anchor node is split. This causes the selection to contain a node that is not part of the selection. As a result, the toolbar/`$getSelectionStyleValueForProperty` is not able to determine the correct selection style.

https://github.com/facebook/lexical/assets/1575198/e438e605-4d4d-4c78-9967-e3176ed531f8

## Solution

Update the selection anchor with the new text node after splitting.

https://github.com/facebook/lexical/assets/1575198/22c177ae-fd17-4b30-bb68-2adab21a9b78


Piliuta:
Failed tests seem spurious 😏

ivailop7 (approved):

PR #5100, opened 2023-10-08
Consolidate Image and InlineImage Plugin refactor

ivailop7:
It never sat right with me that we had 2 plugins that had 90% of the code duplicated, so I went ahead and added the extra property to the main ImagePlugin togglable form the toolbar and deleted the second plugin and its styling.

https://github.com/facebook/lexical/assets/7893468/b334f06b-0a2e-4d0c-88d8-c5dc1bf1ab8c

After doing this, I discovered a bug in the original implementation of inline images, but didn't want to increase the scope of the existing PR. If the image is inlined it doesn't get included if lexical contents are copied and then pasted externally, while when it's in block mode it does work correctly when pasted.


acywatson:
e2es a little weird here but love the idea!

ivailop7:
Rebasing after the insertNodes rewrite, there's a bug when you add an image and start typing and then press Enter, it throws 'Cannot read properties of null (reading 'getChildAtIndex')' which I can't pin down. It was working prior to the insertNodes change.

ivailop7:
Closing this for now, to reevaluate in the future, this got too far behind main and troubleshooting has become too painful. Would be easier to redo from scatch.

PR #5103, opened 2023-10-09
fix(collab): allow editor to not default to a selection for better user experience

nadine-nguyen:
## Issue

Sometimes the editor in collaboration mode is unable to recover the selection (which typically happens after an `undo`/`redo`). This results in the editor selection to fallback to the end of the document. 

This becomes a poor user experience for long documents (having the editor scroll to the very end when they undo/redo)

https://github.com/facebook/lexical/assets/19589943/5997cd2e-63c2-41ef-ac39-3cf7e92b4960

## Scope of work

As a quick solution for better user experience, while the selection undo/redo logic is properly addressed, I've added an `option` property to `CollaborationPlugin` so consumers can opt out of the default fallback. 

---

Please let me know if there are any other better ways to resolve this UX issue 🙏


nadine-nguyen:
Only the chromium e2e collab that isn't passing - test seems to be a bit flaky?

nadine-nguyen:
@zurfyx @fantactuka @acywatson @thegreatercurve Bump! Any opinions about adding this `options` property?

nadine-nguyen:
@zurfyx @fantactuka @acywatson @thegreatercurve Bump

zurfyx (changes requested):
Sorry for the late review, I'm hesitant to stamp this because 1) as you pointed out is a workaround 2) introduces a new API. Deprecating an API method causes churn and should be avoided whenever possible.

I wonder how hard it is to mimic the logic we already do on `removeNode`, basically if we detect the node was deleted, try to find the nearest up/down node on the previous EditorState.

PR #5107, opened 2023-10-10
markdown: use "  \n" as the representation for LineBreakNode

robfig:
In Markdown, single newlines are irrelevant for presentation.
They are only for the editing convenience of the author.

In HTML and the Lexical composer, a LineBreakNode / `<br>` tag
represent a line break in the presentation of the content.
Currently, the equivalence with single newlines causes the markdown
representation to not be accurate to the composer view or HTML
generated or imported.

In Markdown, you can create a line break in the presentation and
achieve the desired effect by ending a line with two spaces
(followed by a newline). Update the import/export to do this.


potatowagon:

PR #5113, opened 2023-10-11
docs: fix typo in decorators.md

CanRau:
> because it's docs only I hope it's fine I just open a PR against main 😬 

alternative is to name both payload I guess


zurfyx (approved):
Thank you!

PR #5122, opened 2023-10-16
add process logic when paste on a QuoteNode

funbinhe1996:
**Context**
When we use `>` and then type whitespace, the Lexical will create a QuoteNode, if we continue to typing, all text will be inside the QuoteNote.
By right, when we copy and paste in this case, the text should be pasted inside the QuoteNode as well. 
Currently, paste normal text does not work as above, so we try to create this Diff  to fix it.


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -239,6 +240,10 @@ function $basicInsertStrategy(
 
     const isLineBreakNode = $isLineBreakNode(node);
 
+    if (i === 0 && $isQuoteNode(selection.anchor.getNode())) {
+      currentBlock = selection.anchor.getNode();
```
acywatson, line 244: Typically with generic modules like this, we try to avoid conditional logic based on specific node types. Instead, we try to condition things on one of the node base classes (ElementNode, DecoratorNode, TextNode) or on "characteristics" (isInline, isSimpleText, etc.)

funbinhe1996: Thanks for your review. 
Actually I tried with more general check before, like `$isElementNode(selection.anchor.getNode())`, but in this case, one test case will be failed.
Here I have some questions:
Refer to this test case, may I know why the `\n` before `a` is omitted in expectedHTML. As when I checked our QuoteNote, it only supports single line, if we enter any `enter` inside QuoteNode, it will jump out from QuoteNode. so I feel this is not correct.

```js
{
          expectedHTML:
            '<p class="editor-paragraph" dir="ltr"><span data-lexical-text="true">a</span><br><span data-lexical-text="true">b</span><br><br></p>',
          inputs: [
            pasteHTML(`<span style="white-space: pre">\na\r\nb\r\n</span>`),
          ],
          name: 'white-space: pre (no touchy) (2)',
},
```
```html
<p class=\"editor-paragraph\" dir=\"ltr\">
<span data-lexical-text=\"true\">a</span><br><span data-lexical-text=\"true\">b</span><br><br></p>
```
```html
<p class=\"editor-paragraph\" dir=\"ltr\">
<br>
<span data-lexical-text=\"true\">a</span><br><span data-lexical-text=\"true\">b</span><br><br></p>
```

zurfyx: @funbinhe1996 - Thank you! You're spot on with the source of the problem. The issue is definitely connected to the fact that we incorrectly create a block level node in this case.

The reason why we introduced this logic was to fix the most basic invalid structures (i.e. pasting a TextNode to the Root), in this case the most obviously answer is to wrap with a ParagraphNode but there seems to be some nodes that don't need wrapping, i.e. QuoteNode or HeadingNode. This logic should perhaps traverse upwards to see if the anchor needs a blocknode or there already exists one that we can reuse.

Ultimately, I think that some sort of Schemas should handle this automatically and this logic should be removed completely altogether.

PR #5123, opened 2023-10-16
Add `commandPriority` option to `LexicalMenu` and dependent components

huw:
Resolves #3714, which is a long-standing issue that multiple users seem to have faced. Previously, this repo has solved it by:

1. Setting all of the commands to `COMMAND_PRIORITY_NORMAL` (#2885)
2. Setting all of the commands to `COMMAND_PRIORITY_CRITICAL` (#3106)
3. Setting all of the commands to `COMMAND_PRIORITY_LOW` (#3441)

It is pretty clear that a more generic solution is needed. This PR adds a `commandPriority` prop to `LexicalMenu` and its dependent components, including:

- `LexicalTypeaheadMenuPlugin` (most often called out)
- `LexicalNodeMenuPlugin`
- `LexicalContextMenuPlugin`
- `LexicalAutoEmbedPlugin` (under `menuCommandPriority`, because this plugin registers an insertion command that should remain at `COMMAND_PRIORITY_EDITOR`)

Future changes should address:

1. Customising priority per-command (this is hard to implement today since commands don’t have string IDs in production builds)
2. Customising priority for all components in `@lexical/react`

However, since `LexicalMenu` has been repeatedly and specifically changed, IMO this PR has a higher cost-benefit than doing those immediately (although I’d love a maintainer’s thoughts!).

### DIFF
--- a/packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
+++ b/packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
@@ -5,7 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-import type {LexicalNode, MutationListener} from 'lexical';
+import type {
+  CommandListenerPriority,
+  LexicalNode,
+  MutationListener,
+} from 'lexical';
 
 import {$isLinkNode, AutoLinkNode, LinkNode} from '@lexical/link';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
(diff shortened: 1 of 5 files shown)


acywatson:
Yea this makes sense to me. Not sure why we put this off so long.

huw:
Do I need to get the build to green to merge this?

acywatson:
> Do I need to get the build to green to merge this?

Nope! All good. Thank you

PR #5126, opened 2023-10-17
 Fixed the toolbar style activation state is incorrect in certain cases

WarrenLee19:
Lexical version:

## Steps To Reproduce

1. Create three paragraphs.
2. Set the text style of the second and third lines to be bold and underlined, 
3. Drag from the end of the first line to the third line, the style activation state of the toolbar is incorrect.

### before:
The style activation state of the toolbar is not activated

https://github.com/facebook/lexical/assets/48719861/b5aecdc4-f8e9-4e2d-83fc-31582a6243ce

### after

The toolbar's style activation state should be bold and underlined

https://github.com/facebook/lexical/assets/48719861/862f1245-5643-4ff5-8866-07ebb9260c88


zurfyx (approved):
Sorry for the late review, thank you for working on this! I don't think it was necessarily incorrect (the current behavior resembles Google Docs but this behavior is probably the most desirable, especially considering that we don't highlight the end part of the first selected line.

Issue #4415, opened 2023-04-26
Bug: Crash when resize a table cell in a table with merged cells


Jianxq12:
Lexical version: 0.10.0

## Steps To Reproduce

1. Create a 5x5 table (not the experimental) in the Lexical Playground.
2. Merge two cells
3. move this mouse cursor to the end cell line of the table 
4. try to resize this last column of the table

**it will caught error, then a  blue line will always  follow  this mouse cursor, it can't disappeared.  after,  unable to input  anything in the editor.**

caught error: 
> caught Error: Expected table cell to be inside of table row.
    at main.0714caa5.js:21:1223100
    at eh (main.0714caa5.js:21:189954)
    at th (main.0714caa5.js:21:192095)
    at Dh.update (main.0714caa5.js:21:213339)
    at main.0714caa5.js:21:1222815
    at main.0714caa5.js:21:1223660
    at Object.Ct (main.0714caa5.js:21:16658)
    at Ot (main.0714caa5.js:21:16812)
    at main.0714caa5.js:21:36721
    at To (main.0714caa5.js:21:36815)

## The current behavior
![image](https://user-images.githubusercontent.com/24584911/234588401-95e78216-bbc3-4022-adf9-e27cb36a891a.png)
![image](https://user-images.githubusercontent.com/24584911/234588794-c1a33ac9-882e-43e6-a6ee-e1f0450492aa.png)

Link to code example:

## The expected behavior
after merged cells, can resize table column  and  input something normally.

icrosil:
With landed commit above, issue should be resolved now.

PR #5141, opened 2023-10-18
Updated cell calculation based on cellspans

icrosil:
## Why
Based on this [issue](https://github.com/facebook/lexical/issues/4415) with Tables current algorithm for resizing is out of shape for cases where cells are merged. What happens is it tries to map index from given active cell and make sure that it exists on each row, if not - throwing exception.

#4415 

## What changed
I've updated TableCellResizer to acomodate that - now it claculates length of row based on children and their cell span. From this we are finding which column corresponds to given active cell, that now always should be less or equal to active cell. After this we are using this column to set width.

Test:

https://github.com/facebook/lexical/assets/5208565/b31e9159-4df7-40cb-a9bc-b7e4f2cb4344

E2E test - couldn't find a way to resize in unit/e2e test.


ivailop7 (approved):
Thank you for this fantastic fix!

Issue #4660, opened 2023-06-14
Bug: Non-experimental $insertTableColumn crashes the editor when trying to add multiple columns


PeterBul:
When calling $insertTableColumn with multiple columns, cells are added to the rows while looping. This causes the second iteration of the loop to select the newly created cell node as target node. This does not seem like it is the intended functionality, since getCordsFromCellNode is called on the target node, which causes Lexical to crash because the newly created node does not exist in the DOM yet.

$insertTableColumn__EXPERIMENTAL is also available, but since this is experimental, and also does not work with multiple columns (tested in playground), I don't see this as a valid replacement yet. It would therefore be preferable to to fix the current $insertTableColumn.

Lexical version: 0.11.1

## Steps To Reproduce

1. Add a table with 2 or more columns.
2. Mark cells in table spanning at least 2 columns.
3. Open table cell action menu and click "Insert x columns left"

Link to code example:
https://codesandbox.io/s/lexical-table-plugin-example-forked-qd40pm

## The current behavior
The editor crashes with the message: "Cell not found in table."

## The expected behavior
The editor should not crash, and two columns should be added.

ivailop7:
Nice find! I reproduced it after a few attempts.

PeterBul:
I managed to fix it by moving some code out of the inner loop of $insertTableColumn. I'm not sure how one would test this though, since the playground uses $insertTableColumn__EXPERIMENTAL by now

zurfyx:
> I managed to fix it by moving some code out of the inner loop of $insertTableColumn. I'm not sure how one would test this though, since the playground uses $insertTableColumn__EXPERIMENTAL by now

Thank you for looking into this! Feel free to publish the PR without test plan. You can temporarily drop the experimental call and run the table tests locally (`npm run test-e2e-chromium -- Tables.spec`)

PR #5145, opened 2023-10-19
Updated $insertTableColumn with batching

icrosil:
## Why
In #4660 we found an issue with $insertTableColumn function that used inital set for a table elements to insert new cells for column 1 by 1. In case you have selected 2+ columns we would insert first, and then for inserting second we would reference newly created cell that is not exist in given grid yet.

## What
There are multiple ways to fix this issue, but batching actions into 1 array seems to be elegant enough.

| before | after |

https://github.com/facebook/lexical/assets/5208565/73083226-c074-46eb-a0eb-abdcade86a52 

https://github.com/facebook/lexical/assets/5208565/8abc7737-ac89-468c-bef7-06ad32aa0bce

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -1598,25 +1598,20 @@ test.describe('Tables', () => {
       page,
       html`
         <p class="PlaygroundEditorTheme__paragraph"><br /></p>
-        <table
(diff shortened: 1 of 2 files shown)


zurfyx (approved):
Thank you Illia! It seems like your own test is failing, can you check this before merging?

icrosil:
Remaining issues on CI seems to be unrelated to this PR, mergin branch.

Issue #5144, opened 2023-10-19
Bug: Pressing Delete when on a LineBreakNode with a LinkNode following removes part of the link's text


9larsons:
Lexical version: v0.12.2

## Steps To Reproduce
https://github.com/facebook/lexical/assets/21961100/205505ba-68e2-4d19-b94b-194b6368a55f
1. Enter multiple line breaks using shift+enter
2. Create a link with at least a few characters of text in the textnode
3. Press up arrow
4. Hit delete

## The current behavior
Removes part of the text in the `LinkNode`'s child `TextNode`.

## The expected behavior
Only remove the line break.

9larsons:
https://github.com/facebook/lexical/pull/5146

PR #5146, opened 2023-10-19
Fixed delete removing link node text when on linebreak

9larsons:
closes https://github.com/facebook/lexical/issues/5144
- mutations to the end text node were too permissive
- splicing of text on removing an element (insert empty) should only impact text selections


thegreatercurve:
LGTM! Thanks for the contribution @9larsons!

thegreatercurve (approved):

PR #5149, opened 2023-10-22
markdown: support formatted link text

### LINKED ISSUE #5148: Bug: Markdown import/export of formatted link text
Lexical version: 0.12.2

## Steps To Reproduce

1. In Lexical playground: linked text with inline code format:

```
 root
  └ (58) paragraph  
    └ (59) link  "https://lexical.dev"
>     └ (60) text  "XXX" { format: code }
```

2. Export to markdown. 

## The current behavior

```markdown
`[XXX](https://lexical.dev)`
```

## The expected behavior

```markdown
[`XXX`](https://lexical.dev)
```

Within an inline code block, markdown is not interpreted, so the current result prints as the literal markdown for the link. However, Markdown tags are valid inside the text content of a Link, so the valid export of that document is with inline code inside the text content of the link.

## Preliminary Investigation

I added these two failing test cases to LexicalMarkdown.test.ts
```js
    {
      html: '<p><a href="https://lexical.dev"><code spellcheck="false" style="white-space: pre-wrap;"><span>XXX</span></code></a><span style="white-space: pre-wrap;"> world</span></p>',
      md: '[`XXX`](https://lexical.dev) world',
    },
    {
      html: '<p><span style="white-space: pre-wrap;">Hello </span><a href="https://lexical.dev"><code spellcheck="false" style="white-space: pre-wrap;"><span>XXX</span></code></a><span style="white-space: pre-wrap;"> world</span></p>',
      md: 'Hello [`XXY`](https://lexical.dev) world',
    },
```

I was able to get the export working very easily, with this diff:

```diff
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -344,16 +344,17 @@ export const LINK: TextMatchTransformer = {
       return null;
     }
     const title = node.getTitle();
-    const linkContent = title
-      ? `[${node.getTextContent()}](${node.getURL()} "${title}")`
-      : `[${node.getTextContent()}](${node.getURL()})`;
+    const linkHref = title ? `${node.getURL()} "${title}"` : node.getURL();
     const firstChild = node.getFirstChild();
     // Add text styles only if link has single text node inside. If it's more
     // then one we ignore it as markdown does not support nested styles for links
     if (node.getChildrenSize() === 1 && $isTextNode(firstChild)) {
-      return exportFormat(firstChild, linkContent);
+      return `[${exportFormat(
+        firstChild,
+        node.getTextContent(),
+      )}](${linkHref})`;
     } else {
-      return linkContent;
+      return `[${node.getTextContent()}](${linkHref})`;
     }
   },
   importRegExp:
```

I got stuck on making the import work. TextFormatTransformers (INLINE_CODE) are processed before TextMatchTransformers (LINK). As a result, the `XXX` ends up as a standalone text node, and TextMatchTransformers only apply to a single text node at a time, so the LINK does not match. 

I don't see how to fix that within the current structure of the code. Considering LINK is builtin and there are special cases for other builtins, I think it would be reasonable to add handling for it, but I was hoping to get guidance / buy-in before going that route. IMO being able to style link text is a valid use case and lexical-markdown should support it.

robfig:
Fixes #5148

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -14,6 +14,7 @@ import type {
 } from '@lexical/markdown';
 import type {ElementNode, LexicalNode, TextFormatType, TextNode} from 'lexical';
 
+import {$isLinkNode} from '@lexical/link';
 import {
   $getRoot,
   $isDecoratorNode,
@@ -176,7 +177,7 @@ function getTextSibling(node: TextNode, backward: boolean): TextNode | null {
   if (!sibling) {
     const parent = node.getParentOrThrow();
 
-    if (parent.isInline()) {
+    if (parent.isInline() && !$isLinkNode(parent)) {
       sibling = backward
         ? parent.getPreviousSibling()
         : parent.getNextSibling();
@@ -185,7 +186,7 @@ function getTextSibling(node: TextNode, backward: boolean): TextNode | null {
 
   while (sibling) {
     if ($isElementNode(sibling)) {
-      if (!sibling.isInline()) {
+      if (!sibling.isInline() || $isLinkNode(sibling)) {
         break;
       }
 
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -17,6 +17,8 @@ import type {LexicalNode, TextNode} from 'lexical';
 
 import {$createCodeNode} from '@lexical/code';
 import {$isListItemNode, $isListNode, ListItemNode} from '@lexical/list';
+import {$createLinkNode} from '@lexical/link';
+import {LINK} from '@lexical/markdown';
 import {$isQuoteNode} from '@lexical/rich-text';
 import {$findMatchingParent} from '@lexical/utils';
 import {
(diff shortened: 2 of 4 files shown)


acywatson (approved):
Nice, thanк  you!

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -176,7 +177,7 @@ function getTextSibling(node: TextNode, backward: boolean): TextNode | null {
   if (!sibling) {
     const parent = node.getParentOrThrow();
 
-    if (parent.isInline()) {
+    if (parent.isInline() && !$isLinkNode(parent)) {
```
acywatson, line 180: On second thought, is there a better way than special-casing LinkNode?

This creates a dependency on @lexical/link, which is less than ideal. it might be necessary here though.

@fantactuka thoughts?

robfig: It would still be a special case, but I can break the dependency by inlining the test of `parent.getType() === "link"` or similar. 

Thinking about expanding beyond that.. would it be ok to remove the code that continues formatting spans across Element nodes & always create boundaries on either side of consecutive spans of Text Nodes?

Eomm:
What is missing to this PR to get it merged?

robfig:
Rebased on main

packages/lexical-markdown/src/MarkdownExport.ts
robfig: We've been using this change in production for 6 months and been happy with it. What needs to be done to get this upstream @acywatson ?

acywatson: I have no objection to this now. If we get the CI passing again, I'll merge it

GermanJablo:
Hi @robfig, seems like there are failing tests. Do you think you could review this?

robfig:
Hi, I looked at the failing tests and realized that the approach I took does lead to that functionality being changed/broken (I forget which exactly). It didn't seem like something which would be easy to fix, so I lost steam

potatowagon:

Issue #5159, opened 2023-10-24
Bug: Align list item do not change the export DOM


trinhvinhtruong96:
Lexical version:
V0.12.2
## Steps To Reproduce

1. Click Bulleted List button
![image](https://github.com/facebook/lexical/assets/48130078/d60ed991-541d-45b1-859c-32085a58422d)
2. Type something
![image](https://github.com/facebook/lexical/assets/48130078/aca8b27e-235b-43bd-8c4d-88f01c2e982b)
3. Click export DOM button, look at the DOM
![image](https://github.com/facebook/lexical/assets/48130078/480d3128-56d1-420c-861d-29f5b04d77d2)
4. Click on the text and then click Right Align button
![image](https://github.com/facebook/lexical/assets/48130078/92e94a45-211b-48a3-a688-ed32b6ff214c)
5. Look at the DOM
![image](https://github.com/facebook/lexical/assets/48130078/b4f1c5db-8d6e-48c9-8705-4aed2a55239e)

Link to code example:
https://playground.lexical.dev/

## The current behavior
The DOM do not change after click Right Align button
## The expected behavior
The right align style should be applied into the list item DOM

PR #5160, opened 2023-10-24
Bug: Align list item do not change the export DOM

trinhvinhtruong96:
This pull request related to #5159
Lexical version:
V0.12.2
## Steps To Reproduce

1. Click Bulleted List button
image
2. Type something
image
3. Click export DOM button, look at the DOM
image
4. Click on the text and then click Right Align button
image
5. Look at the DOM
image

Link to code example:

## The current behavior
The DOM do not change after click Right Align button
## The expected behavior
The right align style should be applied into the list item DOM

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -10,6 +10,7 @@ import type {ListNode} from './';
 import type {
   DOMConversionMap,
   DOMConversionOutput,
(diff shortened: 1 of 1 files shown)


ivailop7:
For @acywatson to confirm if happy with this approach

thegreatercurve:
This approach should be fine. We do something similar in ParagraphNode

thegreatercurve (approved):

Issue #5096, opened 2023-10-06
Bug: useCharacterLimit doesn't work if adding a character before the overflow


francistogram:
Lexical version:

## Steps To Reproduce
1. Go to https://playground.lexical.dev/?isCharLimit=true the character limit is set to 5
2. Type "abcde" and see the 0 there correctly
<img width="184" alt="image" src="https://github.com/facebook/lexical/assets/3721449/c44c0e25-f81f-46e6-a9cf-9782fb7f8c0d">

3. Type "f"  and see the -1 
<img width="164" alt="image" src="https://github.com/facebook/lexical/assets/3721449/37aa2d8e-3d9f-464d-82b3-2fde8da9ed9f">

4. Add a "1" to the front and see a -4 even when it should be -2
<img width="251" alt="image" src="https://github.com/facebook/lexical/assets/3721449/4902e234-fa6c-4f92-a7d4-8701d2d2e02a">

You also see that the tree structure is 
```
 root
  └ (47) paragraph  
>   ├ (59) text  "1abcd"
    ├ (65) overflow  
    | └ (64) text  "e"
    └ (63) overflow  
      └ (62) text  "f"
```
when the two overflow nodes should be merged and I think the root cause of the bug

Link to code example: https://playground.lexical.dev/?isCharLimit=true

## The current behavior
Does not merge overflow nodes when characters are added before the current overflow and overcounts the characters

## The expected behavior
Should act the same adding characters before or after the overflow

JohnyKapo:
Hello may I work on this issue and get assigned?

francistogram:
If you think you can resolve go for it!

My suspicion is that the fix is somewhere in here 

https://github.com/facebook/lexical/blob/9a679ecb2fa3ca999772e2a170f4ad15d19d4635/packages/lexical-react/src/shared/useCharacterLimit.ts#L143

francistogram:
For posterity I figured out a fix

Take `mergePrevious` https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/shared/useCharacterLimit.ts#L244 and instead write `mergeNext` using `const nextNode = overflowNode.getNextSibling();` 

and then replace the call [here](https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/shared/useCharacterLimit.ts#L219)

JohnyKapo:
> For posterity I figured out a fix
> 
[...]

PR #5161, opened 2023-10-24
FIX - Bug: useCharacterLimit doesn't work if adding a character before the overflow

JohnyKapo:
Bug: useCharacterLimit doesn't work if adding a character before the overflow [#5096 ]

 - fixed useCharacterLimit(), instead of merging previous oveflow node it now merges next overflow node (fixed creating unnecessary overflow nodes)
 - updated useLexicalCharacterLimit.test.ts to support new method mergeNext() instead of mergePrevious()
 - updated CharacterLimit.spec.mjs to also support new overflow nodes merging


2wheeh:
> Thank you, it seems like some tests are not passing. I appreciate the fixes and I'm happy to merge them as soon as the CI passes but I also wanted to bring to your attention that we're planning to release a brand new version of the CharacterLimitPlugin in the upcoming weeks.

@zurfyx Hi! any news for the brand new CharacterLimitPlugin? 
The gap between remaining characters and overflow nodes still happens.

o-alexandrov:
@zurfyx would you merge this PR, if tests are fixed?

Please kindly reply, so I would work on wrapping up this PR.

PR #5165, opened 2023-10-25
Update index.tsx

dirheimerb:
.keycode is depreciated, so I updated the method to the recommended method of .key


zurfyx (approved):

zurfyx:
LGTM, thank you!

Issue #4647, opened 2023-06-13
Bug: Adding a table inside of a table and selecting the inner table is breaking the page. 


HitaishLokeshEntr:
I created a Table 5 * 5 and added a table inside one of the table cells as shown in the image 

<img width="1375" alt="Screenshot 2023-06-13 at 3 31 06 PM" src="https://github.com/facebook/lexical/assets/120470917/14c899bd-1ce9-48d2-9df4-53849ce6b2ef">

When trying to select the inner child table the code is breaks and table does not respond post that. The console throws the below error. 

main.79db4264.js:21 Uncaught Error: Minified Lexical error #111; visit https://lexical.dev/docs/error?code=111 for the full message or use the non-minified dev environment for full errors and additional helpful warnings. - ( this is saying "focus is not on grid" )

Lexical version: 0.10.0
 
This is present in the https://playground.lexical.dev/ example as well. 

Can somebody please help me out with this. 

This is a zip of the video. Please check : 

[Screen Recording 2023-06-13 at 3.35.52 PM.zip](https://github.com/facebook/lexical/files/11733574/Screen.Recording.2023-06-13.at.3.35.52.PM.zip)

I would really appreciate any help that i get. Thank you

PR #5166, opened 2023-10-25
Fixes grid selection to be aware of nested tables

icrosil:
## Why
As pointed in #4647 Table selection behaved with issue on moving selection between nested tables. What happened is that we expect nodes anchor/focus (from/to) to be in the same grid by default, but it isn't true for nested tables.

## What
Here after finding focus and anchor cells we check if grids for both of them are equal - that would mean we are on the same level, if not we are moving one selection up. This change makes selection behaviour as next:
- If we move focus from higher level to lower - same higher cell to be selected.
- If we move focus from nested level to level above (any cell) that would mean we make whole original cell to be focused.
- If we move focus from higher to same level including some of nested tables - they will be skipped to focus higher level cells.

Before:

https://github.com/facebook/lexical/assets/5208565/59dcb11c-c267-4b29-947f-2ef1840ea61f

After:

https://github.com/facebook/lexical/assets/5208565/2f526ffa-4cec-4ba0-bfbb-5d7b53690ca8


ivailop7:
A bit on the nit picking side, but if I do 3-level table in table nesting, selection still breaks.

icrosil:
> A bit on the nit picking side, but if I do 3-level table in table nesting, selection still breaks.

yeah, wasn't sure if that's something lexical want to support or not, let me see if could generalize it to cover it.

ivailop7:
Great, however, when I use the / menu and select '5x5 table' and then in any cell I do it again '/' and then '5x5 table' it crashes the selection directly even when using 2 tables nested within each other.

icrosil:
> Great, however, when I use the / menu and select '5x5 table' and then in any cell I do it again '/' and then '5x5 table' it crashes the selection directly even when using 2 tables nested within each other.

I cannot reproduce it locally on this branch, could you please share video steps?

ivailop7:
https://github.com/facebook/lexical/assets/7893468/f37b5364-7bfe-4f88-810c-15fdc04d2662

So I managed to replicate the issue, it appears when I use the keyboard arrows to navigate, it happens 50% of the time. If I do point and click, it works fine, but if I navigate with the arrows upon inserting the second table, selection freezes.

icrosil:
> broken_tables.mp4 
> So I managed to replicate the issue, it appears when I use the keyboard arrows to navigate, it happens 50% of the time. If I do point and click, it works fine, but if I navigate with the arrows upon inserting the second table, selection freezes.

Thanks for the details, was able to replicate now, it seems that selection and grid actions are located in multiple places and not the same for mouse and keyboard events, I'm putting down all possible use cases with keyboard and going to update PR with fixes for all edge cases I could find there.

ivailop7:
Hey @icrosil , is this good for review? I see it working well in the playground preview, but the tests you've added fail often, so unsure if you are planning on more change for this PR.

icrosil:
> Hey @icrosil , is this good for review? I see it working well in the playground preview, but the tests you've added fail often, so unsure if you are planning on more change for this PR.

Hey I wanted to fix tests first, to make sure I'm not breaking anything, but if that would take too long I'd probably slim them down as functionality now is in better state.

I wouldn't claim that it fixed all possible scenarios because of the fragmented selection control I think that would be extremely difficult, but at least we fixed highlighted use cases. Let me try update fixes first, will ping you here on thread when ready for a review. Thanks!

icrosil:
> Hey @icrosil , is this good for review? I see it working well in the playground preview, but the tests you've added fail often, so unsure if you are planning on more change for this PR.

I've decided to remove flaky tests from diff to see if CI will pass on this, doing more manual checks locally please feel free to review code, especially if CI is green :)

ivailop7 (approved):
LGTM

PR #5169, opened 2023-10-26
Table resize fix

AlexanderReznik:
This fixes the bug of being stuck in the resizing mode for the table
Before:
table resize
After:
table resize fixed

Also changed the calculation of cell width for the new cell width to match the user selection.
Before: 
table resize no move
After:
table resize no move fixed


ivailop7 (approved):
Thank you Alex for fixing this!

PR #5180, opened 2023-10-31
Fix at() transpilation

zurfyx:
We can't use `at` until we transpile it. This is not the case now.

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -173,10 +173,11 @@ export function $insertDataTransferForRichText(
   if (text != null) {
     if ($isRangeSelection(selection)) {
       const parts = text.split(/(\r?\n|\t)/);
-      if (parts.at(-1) === '') {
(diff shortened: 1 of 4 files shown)


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -173,10 +173,11 @@ export function $insertDataTransferForRichText(
   if (text != null) {
     if ($isRangeSelection(selection)) {
       const parts = text.split(/(\r?\n|\t)/);
-      if (parts.at(-1) === '') {
+      const partsLength = parts.length;
+      if (parts[partsLength - 1] === '') {
         parts.pop();
       }
-      for (let i = 0; i < parts.length; i++) {
+      for (let i = 0; i < partsLength; i++) {
```
amyworrall, line 180: Is this correct? You may have just popped something off `parts`, and then you're using the old `partsLength` to iterate. Won't you get an index out of bounds?

amyworrall (approved):

Issue #4851, opened 2023-08-01
Feature: Support __style for ListItemNode


maliuta-oleksandr:
I'm experience an issue when I style text inside Lists - pseudo selectors remain with default color, font size... And I can't style parent list item with node transform, because it doesn't support __style as TextNode. Then I have or override the node like this, but it's looks crazy ) And would be awesome to export all functions from @lexical/list to reuse it

```
import type { NodeKey, EditorConfig, EditorThemeClasses } from 'lexical';
import { ListItemNode, ListNode, $isListNode } from '@lexical/list';
import {
    addClassNamesToElement,
    removeClassNamesFromElement,
} from '@lexical/utils';

function $setListItemThemeClassNames(
    dom: HTMLElement,
    editorThemeClasses: EditorThemeClasses,
    node: ListItemNode,
): void {
    const classesToAdd = [];
    const classesToRemove = [];
    const listTheme = editorThemeClasses.list;
    const listItemClassName = listTheme ? listTheme.listitem : undefined;
    let nestedListItemClassName;

    if (listTheme && listTheme.nested) {
        nestedListItemClassName = listTheme.nested.listitem;
    }

    if (listItemClassName !== undefined) {
        const listItemClasses = listItemClassName.split(' ');
        classesToAdd.push(...listItemClasses);
    }

    if (listTheme) {
        const parentNode = node.getParent();
        const isCheckList =
            $isListNode(parentNode) && parentNode.getListType() === 'check';
        const checked = node.getChecked();

        if (!isCheckList || checked) {
            classesToRemove.push(listTheme.listitemUnchecked);
        }

        if (!isCheckList || !checked) {
            classesToRemove.push(listTheme.listitemChecked);
        }

        if (isCheckList) {
            classesToAdd.push(
                checked ? listTheme.listitemChecked : listTheme.listitemUnchecked,
            );
        }
    }

    if (nestedListItemClassName !== undefined) {
        const nestedListItemClasses = nestedListItemClassName.split(' ');

        if (node.getChildren().some((child) => $isListNode(child))) {
            classesToAdd.push(...nestedListItemClasses);
        } else {
            classesToRemove.push(...nestedListItemClasses);
        }
    }

    if (classesToRemove.length > 0) {
        removeClassNamesFromElement(dom, ...classesToRemove);
    }

    if (classesToAdd.length > 0) {
        addClassNamesToElement(dom, ...classesToAdd);
    }
}

function updateListItemChecked(
    dom: HTMLElement,
[...]

PR #5184, opened 2023-11-01
Add __style to ListItemNode

alex-malyita:
**Context** https://github.com/facebook/lexical/issues/4851


acywatson:
Feel free to re-open if we can get the unit tests passing.

Issue #5186, opened 2023-11-01
Bug: LexicalMenu positioning can sit on top of large text


9larsons:
Lexical version: v0.12.2

https://github.com/facebook/lexical/assets/21961100/108fdedf-6fff-44fc-977b-45028552cc2b

## Steps To Reproduce
1. Use a large font size to type some text
2. Type a ':' to open the EmojiPickerPlugin typeahead menu
3. Observe the menu overlaid on the text

## The current behavior
Positioning is based on a fixed margin.

## The expected behavior
Positioning should be based on the bottom of the anchor element.

PR #5187, opened 2023-11-01
Improved LexicalMenu positioning relative to text

9larsons:
closes Facebook/lexical#5186
- positioning of the menu was done with a static margin
- menu could easily end up on top of text when using large fonts (20px+)
- positioning changed to be underneath the anchor element

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -483,17 +483,22 @@ export function useMenuAnchorRef(
   const [editor] = useLexicalComposerContext();
   const anchorElementRef = useRef<HTMLElement>(document.createElement('div'));
   const positionMenu = useCallback(() => {
+    anchorElementRef.current.style.top = anchorElementRef.current.style.bottom;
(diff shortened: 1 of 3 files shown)


ivailop7 (approved):
LGTM

packages/lexical-react/src/shared/LexicalMenu.ts
```diff
@@ -483,17 +483,22 @@ export function useMenuAnchorRef(
...
-      containerDiv.style.top = `${top + window.pageYOffset}px`;
+      const anchorHeight = anchorElementRef.current.offsetHeight; // use to position under anchor
+      containerDiv.style.top = `${
+        top + window.pageYOffset + anchorHeight + 3
+      }px`;
       containerDiv.style.left = `${left + window.pageXOffset}px`;
       containerDiv.style.height = `${height}px`;
       containerDiv.style.width = `${width}px`;
       if (menuEle !== null) {
+        menuEle.style.top = `${top}`;
```
thorn0, line 501: Was it meant to be `` `${top}px` ``? On the other hand, if things work as expected without fixing this, is this line really needed?

PR #5188, opened 2023-11-01
fixed issue 5181 where link address is not shown if comment is set in…

### LINKED ISSUE #5181: Bug: Can't see link url if comment is set inside link node.
If comment is added inside link, and cursor is placed inside comment node, link url is shown empty.

Lexical version:
0.12.2 (playground)

## Steps To Reproduce
1. Add link
2. Inside link add comment
3. Select comment
4. Input is shown with no URL

Video: 
https://github.com/facebook/lexical/assets/50548714/0065f9ac-47a0-462d-bab8-f8020eef817f

Image:
image

## The current behavior
After adding comment inside link, link url is not shown in input.

## The expected behavior
After adding comment inside link, link url should be visible in input.

btezzxxt:
as title fix #5181 

https://github.com/facebook/lexical/assets/25192107/d4fa572b-673f-4ef4-a9b2-7a2c9aedf45c

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
@@ -60,12 +60,9 @@ function FloatingLinkEditor({
     if ($isRangeSelection(selection)) {
       const node = getSelectedNode(selection);
       const linkParent = $findMatchingParent(node, $isLinkNode);
-      const autoLinkParent = $findMatchingParent(node, $isAutoLinkNode);
 
       if (linkParent) {
         setLinkUrl(linkParent.getURL());
-      } else if (autoLinkParent) {
-        setLinkUrl(autoLinkParent.getURL());
       } else if ($isLinkNode(node)) {
         setLinkUrl(node.getURL());
       } else {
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -59,9 +59,13 @@ function FloatingLinkEditor({
     const selection = $getSelection();
     if ($isRangeSelection(selection)) {
       const node = getSelectedNode(selection);
-      const parent = node.getParent();
-      if ($isLinkNode(parent)) {
-        setLinkUrl(parent.getURL());
+      const linkParent = $findMatchingParent(node, $isLinkNode);
+      const autoLinkParent = $findMatchingParent(node, $isAutoLinkNode);
```
zurfyx, line 63: Do we need this? `$isLinkNode` should match for AutoLink no?

zurfyx (approved):
See the comment but otherwise LGTM, thank you!

packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
btezzxxt: oh you are right, I don't think we'd like to match for autolinks

PR #5196, opened 2023-11-02
Issue paste in quote

funbinhe1996:


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -249,15 +249,21 @@ function $basicInsertStrategy(
       node.isParentRequired()
     ) {
       if (currentBlock === null) {
-        currentBlock = node.createParentElementNode();
-        topLevelBlocks.push(currentBlock);
-        // In the case of LineBreakNode, we just need to
-        // add an empty ParagraphNode to the topLevelBlocks.
```
zurfyx, line 255: Should we preserve the comment, I think the logic for LineBreak is still the same?

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -249,15 +249,21 @@ function $basicInsertStrategy(
...
           continue;
         }
+        if (
+          $isElementNode(selection.anchor.getNode()) &&
+          !$isParagraphNode(selection.anchor.getNode())
+        ) {
+          topLevelBlocks.push(node);
+        } else {
+          currentBlock = node.createParentElementNode();
+          topLevelBlocks.push(currentBlock);
```
zurfyx, line 264: Should we consolidate this logic with the linebreak logic, `continue` seems to be the only different bit. Maybe we can reorganize the two ifs and the else better

zurfyx (commented):
Thank you! Looking good to me

zurfyx:
@GermanJablo - this PR revealed we've got an important flaw in the current `insertNodes` logic, this node should never take over the paragraph

https://github.com/facebook/lexical/assets/193447/26abca57-5c88-43c2-977f-06282d5fece6

hoobean1996:
I think you already resolved this issue, I will close it.

hoobean1996:
@zurfyx  No worry, actually I feel that's correct approach, as when I fix it, I do feel my solution is really ad-hoc. You provided better approach that's very awesome

Issue #4134, opened 2023-03-15
Feature: Add HISTORY_STATE_CHANGED_COMMAND for when HistoryState changes


TheSwilliam:
There doesn't seem to be a super straightforward way to achieve this currently. Obviously you can just check for it on editor.registerUpdateListener but it would be nice to have a command that passes in a true "state like" object. Currently the HistoryState is the same object and it makes it difficult to work with. 

Something along the lines of would be nice
```
export const HISTORY_STATE_CHANGED_COMMAND: LexicalCommand<HistoryState> = createCommand(
  "HISTORY_STATE_CHANGED_COMMAND"
);

```
Where the history state is a new object each time.

thegreatercurve:
@TheSwilliam I like this idea, but I think we probably already expose the building blocks to enable this functionality. Similar to what we do in the playground, you can create a custom history instance using `createEmptyHistoryState`, and then listen to the below history commands to get the current `HistoryState`.

Rather than just having one `HISTORY_STATE_CHANGED_COMMAND` command,  it would also give the developer some awareness of where the history state change comes from, whether or not it's an undo or redo:
```js
 const {historyState} = useSharedHistoryContext();

  useEffect(() => {
    return mergeRegister(
      editor.registerCommand(
        CAN_UNDO_COMMAND,
        () => {
          console.log(historyState);
          return false;
        },
        COMMAND_PRIORITY_LOW,
      ),
      editor.registerCommand(
        CAN_REDO_COMMAND,
        () => {
          console.log(historyState);
          return false;
        },
        COMMAND_PRIORITY_LOW,
      ),
    );
  }, [activeEditor, editor, historyState, updateToolbar]);
```

PR #5197, opened 2023-11-02
Add ability to track history changes

golovachruslan:
This commit is adding ability to track updates for undo/redo history stack in `@lexical/history` module:

Currently it's hard to track such updates for undo/redo history stack. To simplify this process - new `HISTORY_CHANGE_COMMAND` command was added. It will be dispatched on each `applyChange` call, except changes with mergeAction === `DISCARD_HISTORY_CANDIDATE`

### DIFF
--- a/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
+++ b/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
@@ -6,7 +6,12 @@
  *
  */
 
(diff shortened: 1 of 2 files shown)


thegreatercurve:
Thanks for the PR @golovachruslan! Unfortunately, I think this probably exposes too much of our internal logic, especially when we merge onto the stack, to really be useful. 

You can use the `CAN_UNDO` and `CAN_REDO` though to achieve the effect you're looking for - tracking when undo or redo states get pushed onto the stack. 

See example in similar issue in #4134

Issue #5194, opened 2023-11-02
Feature: Redo methods that use arrays and indexes and rely more on centralized schema normalization


GermanJablo:
After (1) having worked for more than three weeks on refactoring `insertNodes` and derived PRs and (2) working on my codebase with a centralized schema normalization that I described [here](https://github.com/facebook/lexical/issues/3833#issuecomment-1753600466); I want to share my reflections and conclusions.

The 3 main conclusions I reached are the following:
1. Lexical should rely as little as possible on methods or functions whose arguments are arrays of nodes.
2. Lexical should rely as little as possible on methods or functions whose arguments are the position index of a node within its parent.
3. Lexical should rely as much as possible on centralized schema normalization.

### 1. Rely as little as possible on arrays of nodes as arguments

The `ElementNode.splice` method is clearly inspired by `Array.splice`. However, siblings in Lexical are not ([anymore](https://github.com/facebook/lexical/issues/1860)) arrays, but double linked lists. 

What I see happening frequently, is that one does O(n) iterations to "form the array", and then iterates over that array again to do something [O(n) again] with them somewhere else (in methods like `splice`, `append` or `insertNodes`.

### 2. Rely as little as possible on position indexes

If you put `O(n)` in the search engine, you will find 3 comments `TODO: this is O(n), can we improve?`. All 3 have to do with indexes. 

### 3. Lexical should rely as much as possible on centralized schema normalization.

`insertNodes` did a lot, and still does a lot. The reason for this is that it has implicit responsibility for normalizing nodes. And the same is true for many other functions.

After I used in my codebase the `SchemaPlugin` I suggested in https://github.com/facebook/lexical/issues/3833#issuecomment-1753600466, I found that I could magically eliminate a LOT of unnecessary code and tests. Also, it was only then that I realized how I was implementing the same normalizations in different places (e.g., wrapping an orphaned textNode in a paragraph).

If Lexical provided an official schema (as a set of node transformations) that is easily customizable, I think a lot of code would be redundant, mainly in `@lexical/list`, `@lexical/table` and `@lexical/code`. 

___

## Concrete proposals:

- [ ] deprecate* `ElementNode.splice` in favor of a new function `$replaceRange:(start, end, replaceStart, replaceEnd)`. The method should have an invariant in case `end` is not found as "later sibling of `start`", or `replaceEnd` is not found as "later sibling of `replaceStart`".
[...]

PR #5201, opened 2023-11-03
Improvements in insertNodes

GermanJablo:
- In `insertNodes` I divided the process into two cases: All elements of the array are inline, or at least one element is not inline. That makes `insertNodes` about 10 lines longer, but much easier to reason with and refactor.
- I renamed `$basicInsertStrategy` to `$wrapInlineNodes` and moved it from `clipboard` to `insertNodes`. Now `insertNodes` and `paste` should not have different behaviors, which also made refactoring more difficult. Thanks to this, an incorrect test was detected and corrected: "'white-space: pre (no touchy) (2)".
- Tests about "disjoint sibling text node" were removed. Before the previous PR where I changed `insertNodes`, these tests verified that an error was thrown in that case. This does not happen now. However, the case with "disjoint sibling text node" has already been tested in the G Docs pasting suite.
- Instead of iterating over each node and inserting it, I defined an `insertRangeAfter` method in LexicalNode that can be used to let users define how the nodes should be pasted. This PR makes https://github.com/facebook/lexical/pull/4900 obsolete as it offers a simpler solution.


packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -721,6 +721,8 @@ export function prettifyHTML(string, {ignoreClasses, ignoreInlineStyles} = {}) {
     output = output.replace(/\sstyle="([^"]*)"/g, '');
   }
 
+  output = output.replace(/\s__playwright_target__="[^"]+"/, '');
+
```
GermanJablo, line 725: locally sometimes some tests fail me due to this attribute that playwright uses. I don't know if it happens to any of you too.
It seems to be a bug:
https://github.com/microsoft/playwright/issues/22004

packages/lexical/src/LexicalNode.ts
```diff
@@ -898,6 +898,37 @@ export class LexicalNode {
...
+      current = current.getNextSibling()!;
+      nodestToInsert.push(current);
+    }
+
+    let currentNode: LexicalNode = this;
+    for (const node of nodestToInsert) {
+      currentNode = currentNode.insertAfter(node);
+    }
+  }
+
```
GermanJablo, line 931: TO-DO: This can be optimized to only modify `firstToinsert` and `lastToInsert`.

The reason I didn't do it in the form `insertRangeAfter(nodes: LexicalNode[])` is because this method can be optimized for double linked lists. See https://github.com/facebook/lexical/issues/5194

GermanJablo:
update: I just moved the insertAfter logic inside insertRangeAfter, so that insertAfter is a specific case of insertRangeAfter. Thus, overriding the behavior of `insertRangeAfter` would also cover `insertRange`.

acywatson (approved):

acywatson:
I'm good here - gonna let @zurfyx take a pass since he's more familiar with the new $insertNodes logic at the moment.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -1934,50 +1934,6 @@ describe('LexicalSelectionHelpers tests', () => {
...
-
-          heading.append(child);
-
-          const selection = $getSelection();
-
-          if (!$isRangeSelection(selection)) {
-            return;
-          }
-
-          selection.insertNodes([heading, $createTextNode('bar')]);
```
zurfyx, line 1974: Do we have tests for the correct behavior of this? I think it's important to ensure that block + block will remain on future iterations.

packages/lexical/src/LexicalNode.ts
```diff
@@ -898,6 +898,37 @@ export class LexicalNode {
...
+  /**
+   * Insert a series of nodes after this LexicalNode (as next siblings)
+   *
+   * @param firstToInsert - The first node to insert after this one.
+   * @param lastToInsert - The last node to insert after this one. Must be a
+   * later sibling of FirstNode. If not provided, it will be its last sibling.
+   *
+   *
+   * */
+  insertRangeAfter(firstToInsert: LexicalNode, lastToInsert?: LexicalNode) {
```
zurfyx, line 910: Can you share some more thoughts on why we need this additional method on the API? I feel like `insertAfter` already covers most use cases and otherwise a spread (array) params like we have for `append` will do. The signature is also a bit confusing in my opinion, I understand the performance aspect of it but it's a hard trade-off.

For the sake of insertNodes I would personally have this function as a module level.

```
insertRangeAfter(node: LexicalNode, ...other params)
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1552,60 +1551,72 @@ export class RangeSelection implements BaseSelection {
...
     const nodeToSelectSize = nodeToSelect.getTextContentSize();
-
-    let currentBlock = firstBlock;
-    for (const node of nodes) {
-      if (node === firstNotInline && isMergeable(node)) {
-        currentBlock.append(...node.getChildren());
-      } else if (notInline(node)) {
-        currentBlock = currentBlock.insertAfter(node) as ElementNode;
+    function restoreSelection() {
+      if (nodeToSelect.select) {
```
zurfyx, line 1564: I would be more inclined to keep against `$isTextNode` and `$isElementNode` as it would give us type-safetiness, it's currently typed as `any`

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1552,60 +1551,72 @@ export class RangeSelection implements BaseSelection {
...
+      const index = removeTextAndSplitBlock(this);
+      firstBlock.splice(index, 0, nodes);
+      restoreSelection();
+      return;
+    }
+
+    // CASE 3: At least 1 element of the array is not inline
+    const blocks = $wrapInlineNodes(nodes).getChildren();
+    const isLI = (node: LexicalNode) =>
+      '__value' in node && '__checked' in node;
```
zurfyx, line 1580: Is this part of the command abstraction part? Like code blocks?

zurfyx (approved):
Thank you GermanJablo! Sorry for the late review

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
GermanJablo: Yes, this case (block + text) and also the one you mention (block + block) are covered in the copy paste suite

packages/lexical/src/LexicalNode.ts
GermanJablo: Yes, two reasons.

1. First, it provides a way to modify the behavior of multiple nodes inserted after another. For example, in my case I want that when several blocks are pasted, the indent level of the initial block is added to all of them. This is not currently possible. Using PASTE_COMMAND would mean having to repeat hundreds of lines of code. See https://github.com/facebook/lexical/pull/4900

2. Secondly, these methods offer a much more performant API than `splice` and `append`, optimized for the double linked list structure that uses lexical. See https://github.com/facebook/lexical/issues/5194

I don't recommend using a spread array param, since you almost always insert nodes that are already siblings anyway, and that would require an additional iteration loop.

I also wouldn't "hide" it inside the `insertNodes` module for internal use. This use case is not only useful for copy-paste. For example, in my application you can drag and drop a series of blocks to another position.

packages/lexical/src/LexicalSelection.ts
GermanJablo: Or better yet: https://github.com/facebook/lexical/pull/5205
? 😅

packages/lexical/src/LexicalSelection.ts
GermanJablo: In the current version of `insertNodes` it was pretending to be `lexical/list` agnostic, but it wasn't.
That's why one of the current conditions of `isMergeable` is `(!firstBlock.isEmpty() || !$isRootOrShadowRoot(firstBlock.getParentOrThrow()));`.

Surely anyone who reviews that will not understand that condition (in fact it's still hard for me lol), so I came clean with a more declarative condition.

packages/lexical/src/LexicalNode.ts
zurfyx: @GermanJablo - The two points you made are totally valid, but I also think that having an ergonomic API and backward compatibility to certain extent is very important.

In my opinion, linked lists are a fairly advanced CS concept and even with a good understanding of them, having `firstToInsert` and `lastToInsert` would still lead me to some trial and error to fully understand how the function works.

As for performance, I would take into account real-life use cases. 	`insertNodes` may be worth optimizing, a plugin which handles clipboard paste once every 5 min can probably do with a function that takes slightly longer to initialize. At the end of the day the cost is a linear `O(n)`.

I propose the following to unblock this PR. We hide this function from the public API and I bring this and #5194 up for team discussion to gather more opinions.

GermanJablo: This change is backward compatible.

`...other params` would be O(n), but `lastToInsert` is O(1). Yes, in `insertNodes` you have to iterate all the nodes anyway. But this may not always be the case. A real life example of this is dragging consecutive blocks. `firstToInsert` and `lastToInsert` are obtained immediately from the anchor and focus of the selection.

I understand that `...other params` may be a little easier to understand, but do you think that even with the JSDOC comments `lastToInsert` is difficult to understand?

zurfyx: Isn't the function implementation `O(2n)`? The while loop + for loop. I believe the additional for loop isn't necessary but in any case we're now talking about `O(3n)` which is the same order of magnitude.

IMO the function is not immediately obvious, it's a sort of slice based function and the preconditions are that the first and the last have to be children of the same parent. To me having `insertAfter(node1, node2, node3)` is more intuitive and **iff** optimal performance a must have I'd probably consider a utility that does this for you

```
$insertChildrenAfter(refNode: LexicalNode, parent: ElementNode);
```

GermanJablo: Oh I see where the misunderstanding is. The function is not finished. I'm simply iterating and applying insertAfter on each node to save time. But the idea is to correct the implementation later. There will probably be a loop to verify that `lastToInsert` is a later sibling of `firstToInsert`. But only properties of those two nodes need to be modified, not of all the intermediate ones.

PR #5207, opened 2023-11-05
Fix out of order events during table selection

umk:
With current implementation of [applyTableHandlers](https://github.com/facebook/lexical/blob/ed8f85fb5e5106dba237e2a166e57acaa28e8fbe/packages/lexical-table/src/LexicalTableSelectionHelpers.ts#L58), when user taps a touchpad fast enough (or, presumably, clicks a mouse button), a `mouseup` event may sneak into event loop before `setTimeout` adds the actual implementation of `mousedown` handler there, so `mouseup` remains unhandled and `mousemove` is not removed until the next click. This corrupts the user experience as the table remains in "selecting" state when mouse is already up.

The issue is reproduced in lexical 0.12.2, as well as ed8f85fb5e5106dba237e2a166e57acaa28e8fbe; Chrome 118.0.5993.117 (Official Build) (arm64) 

The PR fixes this behavior by canceling `mousedown` handler upon `mouseup` if it wasn't called yet.

Before:

(logging the balance between adds and removes of `mouseup` events; the problem is reproduced when balance remains 1 after tap)

https://github.com/facebook/lexical/assets/1835333/4b9deab4-8c5b-4519-bf49-f5c4c7c049c9

After:

(few taps, then showing that selection still works for a table)

https://github.com/facebook/lexical/assets/1835333/98af19e8-9d5e-4687-b7af-aa56cbbe715e

### DIFF
--- a/packages/lexical-playground/__tests__/utils/index.mjs
+++ b/packages/lexical-playground/__tests__/utils/index.mjs
@@ -678,6 +678,10 @@ export async function dragMouse(
   await page.mouse.move(fromX, fromY);
   await page.mouse.down();
 
+  // Defer dragMouse continuation past the mouse down
+  // event handler, deferred via setTimeout.
+  await sleep(0);
+
   let toX = toBoundingBox.x;
   let toY = toBoundingBox.y;
   if (positionEnd === 'middle') {
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -79,11 +79,13 @@ export function applyTableHandlers(
   attachTableSelectionToTableElement(tableElement, tableSelection);
 
   tableElement.addEventListener('mousedown', (event: MouseEvent) => {
-    setTimeout(() => {
-      if (event.button !== 0) {
-        return;
-      }
+    if (event.button !== 0) {
+      return;
+    }
 
+    let onMouseMove: ((moveEvent: MouseEvent) => void) | undefined;
+
+    const timeoutId = setTimeout(() => {
       if (!editorWindow) {
         return;
       }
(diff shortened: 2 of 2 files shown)


ivailop7 (approved):
Great catch!

ivailop7 (dismissed):
Great catch!

ivailop7:
I've reran the tests a few times they seem to fail for a few table test cases - adding a row, etc. Could you have a look, if your change requires updating the table tests.

umk:
@ivailop7 That's correct, the most of the table selection tests started to fail in this PR. I have implemented a quick fix to give the `setTimeout` handler a chance to execute before tests continue with [`page.mouse.move`](https://github.com/facebook/lexical/blob/8c7cc81c0f434d7e4c7586069e1a3805643233d7/packages/lexical-playground/__tests__/utils/index.mjs#L691), but it seems that the bug, addressed by this PR, contributes to successful execution of the tests in `main` branch. This video demonstrates that `mouseup` handler is not removed after focusing on the first cell.

[chromium main.webm](https://github.com/facebook/lexical/assets/1835333/579dcd2f-1e87-4791-bb45-2e845a06f0d3)

Because of that, fixing the bug introduces other sort of errors, which root cause I couldn't figure out, unfortunately... It's related to Chromium tests (Webkit and Firefox are OK), which clear table selection upon `page.mouse.up`. The `selectionchange` handler invocations in the video below seem to be responsible for this.

[chromium vs. firefox.webm](https://github.com/facebook/lexical/assets/1835333/0132786a-cc12-4443-9498-60d8b58b20e5)

So I need an input from an expert here. By your preference I can leave this PR as is or replace it with an issue, so someone can help if there is a chance.

ivailop7:
@fantactuka @zurfyx @acywatson hey guys, could you help out here? Your input will be quite valuable here.

zurfyx (approved):
The timeout cleanup is a great catch! We have a similar one for the core event system that we use to normalize events but never heard of any problem there so that's potentially the complete fix, thank you!

Mind double checking on the failing tests because we can merge please?

ivailop7:
Rebased the PR into latest main, since it was about 100 commits behind and to retrigger tests.

PR #5209, opened 2023-11-06
Add conditional format setting, if the node is link fetch the parent paragraph node

### LINKED ISSUE #5089: Bug: After link input the formatting is lost
Hello everyone,

there is a formatting bug that happens when a link is added. This bug already exists in the Lexical-playground.

image

image

does anyone have a solution?

tahir606:
Closes #5089


ivailop7 (changes requested):
Extract the setElementFormat outside of the if/else, this will remove the code duplication and keep only one if statement that if true just passes the parent, so your call to setElementFormat becomes agnostic to the node passed.

ivailop7 (approved):

PR #5221, opened 2023-11-09
Fix lexical-link typo s/_rel/__rel/

etrepum:
I was investigating the use of `[x: string]: any;` in LexicalNode under the assumption that it would hide bugs and I found one.


zurfyx (approved):
Thank you!

Issue #4378, opened 2023-04-21
Bug: Crash when adding a column header to a table with merged cells


michaelsoultanidis:
Adding a column header to a table with merged cells fails with a crash.

Lexical version: 0.10.0 (Latest)

## Steps To Reproduce

1. Create a 4x4 table (**not the experimental**) in the Lexical Playground.  
2. Merge cells (crash is consistently reproducible by merging cells as shown in the video)
3. Add two new rows at the end of the table
4. Click on one of the newly created cells and try to add a column header.

Stack trace:
```
Uncaught Error: Expected table cell to be inside of table row.
    at main.7d82f19f.js:21:1216271
    at eh (main.7d82f19f.js:21:189954)
    at th (main.7d82f19f.js:21:192095)
    at Dh.update (main.7d82f19f.js:21:213339)
    at main.7d82f19f.js:21:1216079
    at onClick (main.7d82f19f.js:21:1218689)
    at Object.Ct (main.7d82f19f.js:21:16658)
    at Ot (main.7d82f19f.js:21:16812)
    at main.7d82f19f.js:21:36721
    at To (main.7d82f19f.js:21:36815)
```

https://user-images.githubusercontent.com/88332746/233641686-0f892a60-a265-4091-a594-da63bc3fdfa9.mov

icrosil:
Hey @zurfyx, going to claim this issue as per tables fixing project.

PR #5230, opened 2023-11-15
Fixed column header on merged cells

icrosil:
## Why
in #4378 tables had an issue to make column headers in rows that had less cells than we wanted to highlight.
When we merge cells across rows and columns we are making single cell to most up/left cell that makes all other columns and rows be unaware of how many cells they actually have in a given row/column, as we checked against it - error was thrown.

## What
To fix this issue I've moved check on min/max cell selected up from the loop - as it doesn't make sense to check it on each row, and if we have a row where we don't have needed column - we simply don't highlight it.

## PS
This is not the first problem like this in a codebase, for example similar issue was fixed here https://github.com/facebook/lexical/pull/5141

Considering that we had ~15 issues with tables here and there - I wouldn't say that this is most critical problem, but some helper files to iterate over rows/columns/cells that would include knowledge about spans would help decrease amount of issues like this. I don't think it should be focus of this PR but to keep in mind.

https://github.com/facebook/lexical/assets/5208565/58b97f2f-18ff-4371-94a0-955397c335c4

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -38,6 +38,7 @@ import {
   selectFromAdditionalStylesDropdown,
   setBackgroundColor,
   test,
+  toggleColumnHeader,
   unmergeTableCell,
 } from '../utils/index.mjs';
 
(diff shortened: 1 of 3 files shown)


zurfyx (approved):
Thank you!

ivailop7 (approved):
LGTM

PR #5237, opened 2023-11-16
Delete experimental table from playground

Fetz:
Deleting the experimental table from the playground for the following reasons:
- The experimental table was not kept up to date with the non version
- Having 2 tables it only confuse new users
- Create noise (issues being raised for experimental table etc)
- If anyone is dependent on the experimental table they can still create a plugin based on the deleted code from this Pull Request or check the git history

Manual testing the playground and creating a table:
https://github.com/facebook/lexical/assets/1071113/92212a2d-0047-4915-8e01-3c7fb57849ac


packages/lexical-playground/src/plugins/ComponentPickerPlugin/index.tsx
```diff
@@ -171,22 +170,6 @@ function getBaseOptions(editor: LexicalEditor, showModal: ShowModal) {
             }),
         }),
     ),
-    new ComponentPickerOption('Table', {
```
ivailop7, line 174: we still want to keep the standard table one

ivailop7 (commented):
This will need indeed need some untangling.

ivailop7:
@Fetz deployment is still failing, let me know when it's ready for review.

GermanJablo:
I would greatly appreciate if you could provide context about this PR.

I opened a thread on discord about this:
https://discord.com/channels/953974421008293909/1177620481533284463

Fetz:
@ivailop7 sorry for the delay, fixed the deployment and tested the playground.

ivailop7 (approved):
LGTM

agriffis:
This missed `packages/lexical-playground/src/nodes/TableNode.tsx` which should also be deleted.

agriffis:
I just added the removal of `TableNode.tsx` to #5342 because otherwise `tsc` fails

Issue #5255, opened 2023-11-20
Bug: Background color only applied to the first cell after dragging table cells 


paigekim29:
Lexical version: v0.12.4

## Steps To Reproduce
1. Visit Lexical Playground: https://playground.lexical.dev/
2. Create a table.
3. Drag the table cells as needed.
4. Change the background color.
5. Note that the background color is applied only to the first cell, not the entire selection.

## The current behavior
After dragging table cells, the background color is only applied to the first cell.
https://github.com/facebook/lexical/assets/70982342/9b3446d7-1202-43fa-8686-281619cffce5

## The expected behavior
When multiple cells are selected, the background color should be applied to each cell.
https://github.com/facebook/lexical/assets/70982342/2789479f-e30b-4c3c-a719-86a957cf463e

ivailop7:
This has been fixed

PR #5256, opened 2023-11-20
Bug: Background color only applied to the first cell after dragging table cells

paigekim29:
This pull request is related to #5255 
- update the handleCellBackgroundColor function to apply the background color to each selected cell


paigekim29:
I made the base branch wrong. I will send the new PR. Thank you :)

Issue #5105, opened 2023-10-10
Bug: The first letter took different format typing after selecting 2 different style text


trinhvinhtruong96:
Lexical version:
 v0.12.0
## Steps To Reproduce

1. Active bold button and type some bold text
2. Inactive bold button and type some normal text
3. Select all bold and normal text
4. type new text

Link to code example:
https://playground.lexical.dev/

## The current behavior
Only the first letter of new text get bold
![chrome-capture-2023-9-10](https://github.com/facebook/lexical/assets/48130078/536d754b-f34c-4051-8c24-b72bb88a2325)
## The expected behavior
The first letter of the new text should have the same format as the other

mayrang:
The original symptom (only the first character inheriting bold) no longer reproduces on main — PR #8069 and PR #8162 fixed this by ensuring all replacement text inherits the first node's format consistently. This issue can be closed as fixed.

PR #5263, opened 2023-11-21
Bug: The first letter took different format when typing after selecting 2 different style text

trinhvinhtruong96:
This Pull Request is related to #5105

Update selection insertText function:

- Update inserted text format to selection format
- Update inserted text style to selection style

It was a bit confusing to me for the first time to decide to keep the style/format of the inserted text the same as the first node or set it to selection style/format. However, after diving into insertText function, i decided to set it to selection style/format as the other case inside insertText function does the same. Otherwise, the toolbar displays the status of selection style/format so I think the inserted text should folow it


potatowagon:

Issue #5260, opened 2023-11-20
Bug: problem with emojies in rtl texts


tasadophi:
hi, I have a problem with emojies. When text starts with emoji editor behave to it like ltr but the text is an rtl text

is there any way to fix this?

tasadophi:
@fantactuka @acywatson hi good job, could you check this please? I opened an pr for it [here](https://github.com/facebook/lexical/pull/5269)

PR #5269, opened 2023-11-22
fix: check pure text by regex (without emojis)

tasadophi:
This pr related to this [issue](https://github.com/facebook/lexical/issues/5260). The regex was wrong and return rtl when any emoji was in the text. I added an extra regex to replace emojis with empty string then main regex check the pure text and return real direction

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -162,7 +162,15 @@ export function getNearestEditorFromDOMNode(
 }
 
 export function getTextDirection(text: string): 'ltr' | 'rtl' | null {
-  const textWithoutEmojis = text.replace(/[^\p{L}\p{N}\p{P}\p{Z}^$\n]/gu, '');
+  const stripEmojis = (str: string) =>
+    str
+      .replace(
+        /([\u2700-\u27BF]|[\uE000-\uF8FF]|\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDFFF]|[\u2011-\u26FF]|\uD83E[\uDD10-\uDDFF])/g,
+        '',
+      )
+      .replace(/\s+/g, ' ')
+      .trim();
+  const textWithoutEmojis = stripEmojis(text);
   if (RTL_REGEX.test(textWithoutEmojis)) {
     return 'rtl';
   }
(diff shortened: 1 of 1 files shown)


tasadophi:
This bug occurs when write emoji and after that is rtl text and editor make whole text ltr @fantactuka @zurfyx @acywatson

tasadophi:
@fantactuka @zurfyx @acywatson Hi, good job I fixed this issue and pass tests now. You make a lot of users happy by fix this bug

packages/lexical/src/LexicalUtils.ts
```diff
@@ -162,10 +162,19 @@ export function getNearestEditorFromDOMNode(
...
 export function getTextDirection(text: string): 'ltr' | 'rtl' | null {
-  if (RTL_REGEX.test(text)) {
+  const stripEmojis = (str: string) =>
+    str
+      .replace(
+        /([\u2700-\u27BF]|[\uE000-\uF8FF]|\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDFFF]|[\u2011-\u26FF]|\uD83E[\uDD10-\uDDFF])/g,
+        '',
+      )
+      .replace(/\s+/g, ' ')
+      .trim();
```
zurfyx, line 172: While this is correct, this might hit performance hard, let's rework this regex. You want a Regex that starts with `^` and looks up at most a O(1) number of characters. Spacing might be a bit of an edge case here since it can be up to O(n) but at the same time it's rare to have them at the start of the text.

In fact, I would say that you can bundle this Regex with the LTR and RTL regexes so that you have something like `^(skipSpacesAndEmojis)LTR or RTL`

I would also move it outside of this function so that it's compiled once.

zurfyx (changes requested):
Thank you for fixing this, that's a good one! Added a comment

packages/lexical/src/LexicalUtils.ts
tasadophi: Yes, everything in programming world can be better and better. could you please do this changes? Because Im not perfect in regex and its performance problems. Just I found this bug and I wanted help to lexical devs and say thanks to them that make our works easy !

potatowagon:

Issue #5270, opened 2023-11-22
Bug: Cannot read properties of undefined (reading 'trim') in $getSelectionStyleValueForProperty


yeion7:
We've encountered a recurring error in our custom editor developed using Lexical. Multiple Sentry reports have highlighted the error message:

```
Cannot read properties of undefined (reading 'trim')
```

After extensive debugging, the issue has been traced to a [specific line](https://github.com/facebook/lexical/blob/main/packages/lexical-selection/src/utils.ts#L184) in the codebase.

This error seems to originate from the utilization of $getSelectionStyleValueForProperty within the toolbar, [similar to its usage in the playground](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx#L640).

Lexical version: `0.11.2`

## Steps To Reproduce

No reproducing steps, most likely is due some malformed css

Link to code example:

PR #5271, opened 2023-11-22
Fix: Resolve "Cannot read properties of undefined (reading 'trim')" issue in $getSelectionStyleValueForProperty

yeion7:
Addresses #5270

Changes Made:

* Modified `getStyleObjectFromRawCSS` to validated key and value before calling `trim`


packages/lexical-selection/src/utils.ts
```diff
@@ -181,7 +181,9 @@ export function getStyleObjectFromRawCSS(css: string): Record<string, string> {
   for (const style of styles) {
     if (style !== '') {
       const [key, value] = style.split(/:([^]+)/); // split on first colon
-      styleObject[key.trim()] = value.trim();
+      if(key && value) {
```
yeion7, line 184:
```suggestion
      if (key && value) {
```

acywatson (approved):

PR #5272, opened 2023-11-22
Add debounceTime parameter to onChangePlugin

kazimirchuk:
Suggestion 🤓 
==========

Let's add a `debounceTime` parameter to OnChangePlugin so that nobody needs to implement it on their side?

Arguments in favour ➕
---------------------
- Lodash's `debounce` is already in dependencies
- I assume it's rare occasion when people need to process every key stroke
- I assume it's common to debounce onChange because it may lead to components been re-rendered and async requests been fired

Arguments against ➖
-------------------
- Anybody could implement it from scratch without much issues

Let me know if it makes sense and if it seems reasonable to add this 🙏


acywatson:
I very much appreciate the PR, but I think it's trivial for someone to add this on their own. I'm hesitant to expand our API surface at this point, on principle.

Issue #5254, opened 2023-11-20
Bug: Problem recognizing input URLs in the autolink plugin.


kim-gilbert:
Lexical version:

## Steps To Reproduce

1. Type https://www.google.com
2. Check linked parts.

<img width="656" alt="스크린샷 2023-11-20 오후 2 14 49" src="https://github.com/facebook/lexical/assets/6689758/b8435e6d-6e0a-43f1-a42b-13c934559618">

Link to code example: https://playground.lexical.dev/

## The current behavior
Only part of the URL is linked.

## The expected behavior
The URL should be fully linked.

PR #5275, opened 2023-11-22
Fix: Autolink plugin URL recognition failures

disalechinmay:
This PR fixes https://github.com/facebook/lexical/issues/5254 
The above issue was happening because period (`.`) was being considered as a separator sequence (`PUNCTUATION_OR_SPACE` in `packages/lexical-react/src/LexicalAutoLinkPlugin.ts`) which is wrong.

Additionally, this PR introduces an enhancement to the existing regular expression used for URL validation. The existing regex for URL validation was effective but limited in scope, failing to recognize several common URL formats. With the growing need for more comprehensive URL validation in modern web applications, it became imperative to update our regex to handle a broader range of URLs, including those with user credentials and special domains like 'localhost', as well as IP addresses. To ensure the robustness of the updated regex, a series of tests were conducted with various URL formats. The test results show a comparison between the old variant of the regex and the new enhanced variant. The new variant successfully matches a wider range of URLs, making it a more versatile choice for URL validation.

| Test URL                                                          | Old Variant Match | New Variant Match |
|-------------------------------------------------------------------|:-----------------:|:-----------------:|
| `http://example.com`                                              | ✅                | ✅                |
| `https://www.example.com`                                         | ✅                | ✅                |
| `www.example.com`                                                 | ✅                | ✅                |
| `example.com`                                                     | ❌                | ✅                |
| `http://localhost:8000`                                           | ❌                | ✅                |
| `ftp://example.com`                                               | ❌                | ❌                |
| `http://example.com/path/to/resource?query=string#fragment`       | ✅                | ✅                |
| `example`                                                         | ❌                | ❌                |
| `http://192.168.1.1`                                              | ✅                | ✅                |
| `https://example.com:3000/path`                                   | ✅                | ✅                |
| `http://example.com:8080/path`                                    | ✅                | ✅                |
| `https://subdomain.example.com`                                   | ✅                | ✅                |
| `http://127.0.0.1`                                                | ✅                | ✅                |
| `http://example.com/path/to/page.html`                            | ✅                | ✅                |
| `https://example.com?param=value`                                 | ✅                | ✅                |
| `http://example.co.uk`                                            | ✅                | ✅                |
| `https://255.255.255.255`                                         | ✅                | ✅                |
| `https://example.com#anchor`                                      | ✅                | ✅                |
| `http://example.com/path/to/page.html?query=string#fragment`      | ✅                | ✅                |
| `http://[::1]` (IPv6 address)                                     | ❌                | ❌                |
| `https://username:password@example.com`                           | ✅                | ✅                |
| `ftp://192.168.0.1/resource`                                      | ❌                | ❌                |
| `http://example` (no TLD)                                         | ❌                | ❌                |
| `http://.com` (no domain name)                                    | ❌                | ❌                |
| `https://example.com:invalidport` (invalid port)                  | ✅                | ❌                |
| `http:///path/without/domain` (missing domain)                    | ❌                | ❌                |
| `http://-example.com` (invalid domain start)                      | ✅                | ❌                |
| `http://example-.com` (invalid domain end)                        | ✅                | ❌                |
| `http://example..com` (double dot in domain)                      | ✅                | ❌                |
| `http://ex..ample.com` (double dot in domain)                     | ✅                | ❌                |
| `http://abcdefghij.com` (long domain name)                        | ✅                | ✅                |


ivailop7:
This looks great, could you update the AutoLink tests to reflect your changes and I'll merge it.

disalechinmay:
@ivailop7 - I've updated the tests 🧪 
Also updated the regex a bit.

ivailop7 (approved):
LGTM

thorn0:
The complexity of this regexp seems problematic. Just open the playground and start typing to see what I mean. My browser starts dying when the input is around 30 characters long and freezes at 40.

acywatson:
> The complexity of this regexp seems problematic. Just open the playground and start typing to see what I mean. My browser starts dying when the input is around 30 characters long and freezes at 40.

It was, in fact, problematic. Reverted it - thanks for the catch.

moy2010:
@acywatson , @thorn0 , the performance issue wasn't caused by this PR. See [here](https://github.com/facebook/lexical/issues/5287) for more info.

acywatson:
> @acywatson , @thorn0 , the performance issue wasn't caused by this PR. See [here](https://github.com/facebook/lexical/issues/5287) for more info.

I'm pretty certain that it was. If you go back literally one commit and check the preview deployment, the issue isn't there. There may be multiple issues here, but I think this PR is part of the problem at least.

Issue #5276, opened 2023-11-22
Feature: GridSelection separation from lexical core


icrosil:
## Why
GridSelection creates unnecessary complexities within code by increasing bundle size and providing functionality that is not required from core use cases. In addition current Selection architecture (BaseSelection, RangeSelection, NodeSelection, GridSelection) used regularly with if conditions and multiple actions based on the selection type.

All of the reasons above are concluding that Selection approach should be refactored with goal of removing GridSelection from core package.

## Approach
On internal disucssion with @zurfyx and @fantactuka we've aligned on next steps:
- gradually moving code away from practices of "| RangeSelection | NodeSelection | GridSelection"
- removing / deprecating pieces of code that are outdated , or should be covered by one of the selections directly
- Move GridSelection to a Tables package

## Notes
- Created issue for visibility purpose, so all of upcoming discussions can be done here.
- All the PRs will be linked here as work will be completed in a several stages.

HoaPhamThi:
i am using lexical version 0.12.5, why can't i find GridSelection and $isGridSelection in lexical/table

PR #5280, opened 2023-11-23
[PR 1] Swap instance selection for BaseSelection in most cases

icrosil:
## Why
In #5276 we proposed a solution on how to migrate GridSelection out of the core of the bundle, here is the first step by removing type usage to the BaseSelection where all type of selections could be added

## What
- Updated `| GridSelection | RangeSelection | NodeSelection` for BaseSelection
- Added 3 methods to BaseSelection interface - insertNodes, getCachedNodes, setCachedNodes that probably should be private but have no preference personally.
- Update TS and Flow files
- Updated tests from $getSelection to $getRangeSelection where logically we expect Range selection


packages/lexical-react/src/useLexicalNodeSelection.ts
```diff
@@ -62,11 +62,12 @@ export function useLexicalNodeSelection(
           selection = $createNodeSelection();
           $setSelection(selection);
         }
-
-        if (selected) {
-          selection.add(key);
-        } else {
-          selection.delete(key);
+        if ($isNodeSelection(selection)) {
```
fantactuka, line 65: It should already be node selection at this point, right?

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
```diff
@@ -308,9 +309,9 @@ describe('LexicalSelectionHelpers tests', () => {
       await Promise.resolve().then();
 
       editor.getEditorState().read(() => {
-        const selection = $getSelection();
+        const selection = $getRangeSelection();
 
-        if ($isNodeSelection(selection)) {
+        if ($isNodeSelection(selection) || selection == null) {
```
fantactuka, line 314: Can clean up $isNodeSelection here and below

packages/lexical/src/nodes/__tests__/unit/LexicalElementNode.test.tsx
```diff
@@ -96,7 +95,8 @@ describe('LexicalElementNode tests', () => {
         // If you broke this test, you changed the public interface of a
         // serialized Lexical Core Node. Please ensure the correct adapter
         // logic is in place in the corresponding importJSON  method
-        // to accomodate these changes.
+        // to accomodate these changes.import { $getRangeSelection } from 'lexical/src';
```
fantactuka, line 98: VSCode's auto imports are pain sometimes :)

fantactuka (approved):
Nice one 👍

packages/lexical-react/src/useLexicalNodeSelection.ts
icrosil: yes, although flow was still complaining at this point.

packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
icrosil: yes, although flow was still complaining at this point so couldn't delete it :(

packages/lexical-react/src/__tests__/unit/PlainRichTextPlugin.test.tsx
```diff
@@ -170,9 +169,9 @@ describe('LexicalNodeHelpers tests', () => {
       await editor.getEditorState().read(() => {
         expect($rootTextContent()).toBe('foo');
 
-        const selection = $getSelection();
+        const selection = $getRangeSelection();
 
-        if ($isNodeSelection(selection)) {
+        if (selection == null) {
```
zurfyx, line 174: Should we just use `!$isRangeSelection(selection)` like we use elsewhere? Also, prefer `===` in Lexical codebase

packages/lexical-selection/src/__tests__/utils/index.ts
```diff
@@ -13,6 +13,7 @@ import {
   $isRangeSelection,
   $isTextNode,
 } from 'lexical';
+import {$getRangeSelection} from 'lexical/src';
```
zurfyx, line 16: Should be imported from `'lexical'`

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2597,20 +2619,18 @@ export function internalCreateRangeSelection(
...
-export function $getPreviousSelection():
-  | null
-  | RangeSelection
-  | NodeSelection
-  | GridSelection {
+export function $getRangeSelection(): null | RangeSelection {
+  const selection = $getSelection();
+  const isRange = $isRangeSelection(selection);
+  return isRange ? selection : null;
+}
```
zurfyx, line 2631: I would be inclined not to have this function in the API, for two reasons:
1. It doesn't save lines of code. In all your ocurences above the LoC are the same
2. Acknowledging that there are other types of selection is helpful for the developer to consider other possible use cases

The end result is still a two liner

```
if ($isRangeSelection(selection))
```
vs
```
const selection = $getRangeSelection();
if (selection !== null)
```

packages/lexical/src/nodes/__tests__/unit/LexicalTextNode.test.tsx
```diff
@@ -121,7 +121,8 @@ describe('LexicalTextNode tests', () => {
         // If you broke this test, you changed the public interface of a
         // serialized Lexical Core Node. Please ensure the correct adapter
         // logic is in place in the corresponding importJSON  method
-        // to accomodate these changes.
+        // to accomodate these changes.import { $getRangeSelection } from 'lexical/src';
```
zurfyx, line 124:
```suggestion
        // to accomodate these changes.
```

zurfyx (approved):
Great work, added some nit comments but I love what I'm seeing here! Thanks for taking the lead on this on this major refactor

PR #5281, opened 2023-11-24
[PR 2] 5276 PointSelection instead of Grid/RangeSelections

icrosil:
## Why
In #5276 we proposed a solution on how to migrate GridSelection out of the core of the bundle, here is the second step by removing most usage of GridSelection where we can safely use PointSelection

## What
- introduced PointSelection that is interface for Grid and Range becuase they both have anchor and focus, but implemented as class because we want runtime validations of class.


packages/lexical/src/LexicalSelection.ts
```diff
@@ -268,12 +268,93 @@ export interface BaseSelection {
...
-  insertRawText(text: string): void;
   insertText(text: string): void;
+  insertRawText(text: string): void;
   is(selection: null | BaseSelection): boolean;
   insertNodes(nodes: Array<LexicalNode>): void;
-  getCachedNodes(): null | Array<LexicalNode>;
-  setCachedNodes(nodes: null | Array<LexicalNode>): void;
+}
+
+export class PointSelection implements BaseSelection {
```
thorn0, line 277: Seems to make sense to mark this class as `abstract`.

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -349,7 +350,7 @@ export function $patchStyleText(
   let lastNode = selectedNodes[lastIndex];
 
   if (selection.isCollapsed()) {
-    $patchStyle(selection, patch);
+    $patchStyle(selection as RangeSelection, patch);
```
thorn0, line 353: If non-core code is going to be allowed to create subtypes of `PointSelection`, this type assertion is unsafe.

packages/lexical/src/LexicalSelection.ts
icrosil: actually abstract is exactly what I was looking for, thanks @thorn0 for mentioning

Issue #5274, opened 2023-11-22
Bug: unable to delete the first character on android chrome


kirandash:
when there is one character left at any line it wont delete it and it will jump to the beginning of line without deleting the character

Lexical version: 0.12.4

## Evidence

https://github-production-user-asset-6210df.s3.amazonaws.com/19841175/276697867-50d93d4c-f6db-474a-bfb0-b63280e587da.mp4

kirandash:
I will be creating a PR for this based on my solution shared here: https://github.com/facebook/lexical/issues/4941#issuecomment-1818305883

bnol:
+1

rohini21:
@kirandash @acywatson Thanks for this fix. When will this be published?

kirandash:
> @kirandash @acywatson Thanks for this fix. When will this be published?

@rohini21 Not sure when it will be released next. In the meantime, you can use a patch ex: patch-package for npm or pnpm patch etc to patch Lexical.dev.js and Lexical.prod.js files on your project.

How to generate the Lexical.dev.js and Lexical.prod.js file with the fix?
Go to LexicalEvent.ts file: add the fix code and run `npm run build-release`. Once the build is ready, use those files from dist folder on your project.

Hope that helps.

kirandash:
@rohini21 This is now merged to main and available in release version `v0.12.5`

PR #5282, opened 2023-11-25
fix: backspace in Android not deleting first character

kirandash:
Fix for issue: https://github.com/facebook/lexical/issues/5274

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -539,6 +539,11 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
             selection.format = anchorNode.getFormat();
             selection.style = anchorNode.getStyle();
           }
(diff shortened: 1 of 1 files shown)


acywatson:
Thanks for submitting this - running the CI now.

acywatson:
Can you provide a quick explanation as to how this fixes the issue/how you arrived at this solution?

kirandash:
> Can you provide a quick explanation as to how this fixes the issue/how you arrived at this solution?

After some debugging on android I found that the deletion was not happening when the number of characters on selected node text content is <= 1 probably because of the current issue on android chrome as pointed out here: https://github.com/facebook/lexical/issues/4941#issuecomment-1793877373. 

Hence I decided to do override the default android behaviour only when the text length <=1 and do a manual trigger for the deletion similar to what was implemented in the else block.

Not sure if it's the best solution. But for now this is the only solution I could think of. Please do let me know if this can be improved..

acywatson (approved):

PR #5283, opened 2023-11-25
Remove style attributes for text/html before paste

FroeMic:
Update $insertDataTransferForRichText to remove style attributes for text/html before paste

Situation: When using the RichTextPlugin, common use cases involve controlling the styles with which the content is rendered to ensure a consistent experience. 

Problem: When pasting content from external sources (Example VS Code) the style attributes are inserted as well, creating an uncontrolled experience of how pasted content is rendered. 

Proposed Solution: Automatically remove all style attributes before pasting "text/html" content.

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -158,6 +158,19 @@ export function $insertDataTransferForRichText(
     try {
       const parser = new DOMParser();
       const dom = parser.parseFromString(htmlString, 'text/html');
+
+      // strip any style attributes from pasted text/html content
+      const removeStyleAttributes = (element: HTMLElement) => {
+        element.removeAttribute('style');
+        for(const child of el.childNodes) {
+          if (child instanceof HTMLElement) {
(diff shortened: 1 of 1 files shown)


FroeMic:
Not sure whether this is the right place to raise this here: In finding a way to fix this issue for me, I figured that there isn't really a way to intercept paste command. 

Implementing lexical as a developer, I would find it "cleaner" to be able to hook into certain events and write custom middleware for them. 

Using the provided callbacks I could only listen for the "paste" event after its content had already be inserted by the editor, but not before to check or even conditionally intercept the event to avoid lexicals default behavior.

zurfyx:
Thank you, can you expand further on the use cases and the inconsistent experience you pointed out? Style tags ensure the correct translation of HTML to Lexical when there's no tag equivalent (or when the source editor decides to do so). For example: https://github.com/facebookexternal/Outline//blob/dbb081359c4c2f883f9e68ab9d81d69aa2f1c978/packages/lexical/src/nodes/LexicalTextNode.ts#L1118-L1133

potatowagon:

PR #5284, opened 2023-11-26
docs: improve DecoratorNode docs

petrovmiroslav:
Add description for DecoratorNode.prototype.decorate method, add description to the docs website "Extending DecoratorNode" section.

### DIFF
--- a/packages/lexical-website/docs/concepts/nodes.md
+++ b/packages/lexical-website/docs/concepts/nodes.md
@@ -1,5 +1,3 @@
-
-
 # Nodes
 
 ## Base Nodes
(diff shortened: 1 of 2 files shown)


acywatson:
This is good, thanks. One small typo that actually muddles the semantics. Otherwise, good to go.

packages/lexical-website/docs/concepts/nodes.md
```diff
@@ -271,7 +269,13 @@ export function $createVideoNode(id: string): VideoNode {
...
 
-export function $isVideoNode(node: LexicalNode | null | undefined): node is VideoNode {
+export function $isVideoNode(
+  node: LexicalNode | null | undefined,
+): node is VideoNode {
   return node instanceof VideoNode;
 }
 ```
+
+Using `useDecorators`, `PlainTextPlugin` and `RichTextPlugin` execute `React.createPortal(reactDecorator, element)` for each `DecoratorNode`,
```
acywatson, line 279: I think we mean to say "executes" instead of "execute" here. The latter makes it sound like an instruction, which I don't think is the intent. It's an explanation.

petrovmiroslav: Thank you for pointing it out to me. I meant that both `PlainTextPlugin` and `RichTextPlugin` execute `React.createPortal(reactDecorator, element)` for each `DecoratorNode`, using `useDecorators` hook.
 Could you help me to phrase that correctly, please?

acywatson: Yea just change "execute" to "executes" and we're good

acywatson (approved):

PR #5288, opened 2023-11-27
Allow exporting a document fragment from the exportDOM function

### LINKED ISSUE #5238: Feature: Allow `exportDOM` to return a `DocumentFragment`
When exporting an element as DOM, sometimes I want to remove that element and only export it's children.

My current use-case for this is the `LinkNode`, which I have extended as `CustomLinkNode` in order to change the default `exportDOM` behaviour. If the `href` of a link is set to `https://`, I want to consider that link invalid and remove it from the exported HTML, but if it contains other elements like `<em>`, those should remain.

Here is an example of the code I would like to write:

```ts
exportDOM(editor: LexicalEditor): DOMExportOutput {
  const { element } = super.exportDOM(editor)

  if (element && isHTMLElement(element)) {
    element.removeAttribute('class')

    if (element.getAttribute('href') === 'https://') {
      const content = new DocumentFragment()
      // Here we are appending all the children of the link to the fragment we created above
      content.append(...element.childNodes)
      // Lexical currently does not allow a fragment as the element
      return { element: content }
    }
  }

  return {
    element,
  }
}
```

From what I can tell, this should work fine, except for this section of this file which only allows `HTMLElement`:
https://github.com/facebook/lexical/blob/83b831784dea38677775482953b89987cdd6969d/packages/lexical-html/src/index.ts#L133-L135

I propose that the line above be changed to something like

```ts
if (isHTMLElement(element) || isHTMLFragment(element)) {
  element.append(fragment);
}
```

And the return type of `DOMExportOutput` to accept a `DocumentFragment` as the `element`.

GRA0007:
Closes #5238 

This PR allows a node's `exportDOM` function to return a [`DocumentFragment`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) in addition to a HTMLElement or TextNode, which allows exporting a collection of elements with no wrapping element from that node.

See the linked issue for more details, but my basic use case is being able to remove a wrapping anchor tag from the HTML output if the `href` property is invalid, but keep any formatting (such as <em> tags) within that anchor tag.

### DIFF
--- a/packages/lexical-html/src/index.ts
+++ b/packages/lexical-html/src/index.ts
@@ -130,14 +130,20 @@ function $appendNodesToHTML(
   }
 
   if (shouldInclude && !shouldExclude) {
-    if (isHTMLElement(element)) {
+    if (isHTMLElement(element) || element instanceof DocumentFragment) {
       element.append(fragment);
     }
     parentElement.append(element);
 
     if (after) {
       const newElement = after.call(target, element);
-      if (newElement) element.replaceWith(newElement);
+      if (newElement) {
+        if (element instanceof DocumentFragment) {
+          element.replaceChildren(newElement);
+        } else {
+          element.replaceWith(newElement);
+        }
+      }
     }
   } else {
     parentElement.append(fragment);
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -342,8 +342,8 @@ export type DOMConversionOutput = {
   node: null | LexicalNode | Array<LexicalNode>,
 };
 export type DOMExportOutput = {
-  after?: (generatedElement: ?HTMLElement) => ?HTMLElement,
-  element?: HTMLElement | null,
+  after?: (generatedElement: ?HTMLElement) => ?HTMLElement | DocumentFragment,
+  element?: HTMLElement | DocumentFragment | null,
 };
 export type NodeKey = string;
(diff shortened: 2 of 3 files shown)


packages/lexical-html/src/index.ts
```diff
@@ -130,14 +130,20 @@ function $appendNodesToHTML(
   }
 
   if (shouldInclude && !shouldExclude) {
-    if (isHTMLElement(element)) {
+    if (isHTMLElement(element) || element instanceof DocumentFragment) {
```
acywatson, line 133: I think we avoid using instanceof because it complicates server-side use cases somehow. I can't recall exactly. It's hard to polyfill or something? 

You can just do the same thing we do in isHTMLElement here.

acywatson (changes requested):

packages/lexical-html/src/index.ts
GRA0007: All good, I've added a util for that. Let me know if it should just be defined in this file instead of exported from `@lexical/utils` like `isHTMLElement`.

skopz356:
@acywatson  What is the reason that this PR was closed?

etrepum:
I think it was closed because it was stale and had failing tests, if this feature were implemented with test coverage and passing tests it would be merged

GRA0007:
Ah apologies @skopz356 I didn't have the time to fix it, thanks for picking it back up in #6641!

PR #5291, opened 2023-11-27
[PR 3] Removed edge-cases of GridSelection

icrosil:
## Why
In #5276 we proposed a solution on how to migrate GridSelection out of the core of the bundle, here is the third step by removing edge-case usage of GridSelection to decrease dependencies.

## What
- insertGeneratedNodes - function that inserts into table nodes from another tables - refactored to use only core functions, without any knowledge of the gridSelection
- mergeNodeStrategy - related to funciton above, now uses RangeSelection to select newly added nodes.
- patchStyleText - can handle current logic without knowledge of GridSelection


icrosil:
Hey @zurfyx , @fantactuka could you please help me review this PR?

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -340,9 +339,10 @@ function $mergeGridNodesStrategy(
   }
 
   if (newAnchorCellKey && newFocusCellKey) {
-    const newGridSelection = DEPRECATED_$createGridSelection();
-    newGridSelection.set(gridNode.getKey(), newAnchorCellKey, newFocusCellKey);
-    $setSelection(newGridSelection);
+    const rangeSelection = $createRangeSelection();
+    rangeSelection.anchor.set(newAnchorCellKey, 0, 'element');
+    rangeSelection.focus.set(newFocusCellKey, 0, 'element');
+    $setSelection(rangeSelection);
```
zurfyx, line 345: This is a tricky one, I remember that a selection like this can break the table because you're essentially selecting the cells so operations like delete can make it inconsistent. I think that this entire function should be moved to the tables plugin and we can have a command to delegate handling this logic there.

We don't necessarily have to do it in this PR so you can leave GridSelection for now.

zurfyx (approved):

PR #5292, opened 2023-11-27
keep selection style when clicking empty editor

### LINKED ISSUE #5202: Bug: Lexical Playground bug & Lexical internal logic bug in every lexical version
Lexical version: every version

## Steps To Reproduce

https://github.com/facebook/lexical/assets/78121870/1ec772c8-4511-4d07-8e4b-e1cb685e59e0

1. select `20px` in toolbar 
2. click editor 

Link to code example:

## The current behavior
- reset default fontSize

## The expected behavior
- keep current style

I found what makes bug.  [onSelectionChange](https://github.com/facebook/lexical/blob/ed8f85fb5e5106dba237e2a166e57acaa28e8fbe/packages/lexical/src/LexicalEvents.ts#L258) has below logic.

```typescript
        if (
          currentTimeStamp < timeStamp + 200 &&
          anchor.offset === lastOffset &&
          anchor.key === lastKey
        ) {
          selection.format = lastFormat;
          selection.style = lastStyle;
      }
```

If conditional statement is true, It works expected. but `currentTimeStamp < timeStamp + 200` is false. 
What is the intention of this code??

youngkyo0504:
resolve #5202 

change code to keep selection style when clicking editor which has empty content.

`before`

https://github.com/facebook/lexical/assets/78121870/c2198c28-e190-4efe-886c-03ed20e4be80

`after`

https://github.com/facebook/lexical/assets/78121870/01f0deb7-5b65-4e01-b77a-cdaf16a85bc1

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -11,6 +11,7 @@ import type {NodeKey} from './LexicalNode';
 import type {ElementNode} from './nodes/LexicalElementNode';
 import type {TextNode} from './nodes/LexicalTextNode';
 
+import {$isRootTextContentEmpty} from '@lexical/text';
 import {
   CAN_USE_BEFORE_INPUT,
   IS_ANDROID,
@@ -334,7 +335,10 @@ function onSelectionChange(
           if (anchor.type === 'text') {
             selection.format = anchorNode.getFormat();
             selection.style = anchorNode.getStyle();
-          } else if (anchor.type === 'element') {
+          } else if (
+            anchor.type === 'element' &&
+            !$isRootTextContentEmpty(editor.isComposing(), false)
+          ) {
             selection.format = 0;
             selection.style = '';
           }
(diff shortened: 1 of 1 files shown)


ivailop7:
@youngkyo0504 can you sign the CLA and I'm OK with merging this.

youngkyo0504:
> @youngkyo0504 can you sign the CLA and I'm OK with merging this.

Thank you! I have signed the CLA 😁

packages/lexical/src/LexicalEvents.ts
```diff
@@ -11,6 +11,7 @@ import type {NodeKey} from './LexicalNode';
 import type {ElementNode} from './nodes/LexicalElementNode';
 import type {TextNode} from './nodes/LexicalTextNode';
 
+import {$isRootTextContentEmpty} from '@lexical/text';
```
acywatson, line 14: I don't think we should create this dependency between the core and @lexical/text

acywatson (changes requested):

packages/lexical/src/LexicalEvents.ts
youngkyo0504: @acywatson 

how about using this instead of delete dependency? 

```ts
const root = $getRoot();
isRootEmpty  root.getTextContent() === ''
```

acywatson (approved):

PR #5293, opened 2023-11-27
[PR 4] remove GridSelection from lexical core

icrosil:
## Why
In #5276 we proposed a solution on how to migrate GridSelection out of the core of the bundle, here is the fourth step by removing GridSelection into lexical-table package.

## What
Moved GridSelection to lexical-table, updated all references accordingly.


zurfyx (commented):
That's looking great! Thank you

Issue #4692, opened 2023-06-22
Bug: AutoLinkPlugin fails simple examples like https://www.google.com


cranium:
Lexical version: 0.11.1

## Steps To Reproduce

1. Go to https://playground.lexical.dev/
2. Type https://www.google.com
3. https://www.google is linked but not .com

I have an example here which appears to match the problem in the playground:
```javascript
import { AutoLinkNode } from "@lexical/link";
import { AutoLinkPlugin } from "@lexical/react/LexicalAutoLinkPlugin";
import {
  InitialConfigType,
  LexicalComposer,
} from "@lexical/react/LexicalComposer";
import { ContentEditable } from "@lexical/react/LexicalContentEditable";
import { HistoryPlugin } from "@lexical/react/LexicalHistoryPlugin";
import { OnChangePlugin } from "@lexical/react/LexicalOnChangePlugin";
import { PlainTextPlugin } from "@lexical/react/LexicalPlainTextPlugin";
import { $getRoot, $getSelection, EditorState, ParagraphNode } from "lexical";

// When the editor changes, you can get notified via the
// LexicalOnChangePlugin!
function onChange(editorState: EditorState) {
  editorState.read(() => {
    // Read the contents of the EditorState here.
    const root = $getRoot();
    const selection = $getSelection();

    console.log(root, selection);
  });
}

// Catch any errors that occur during Lexical updates and log them
// or throw them as needed. If you don't throw them, Lexical will
// try to recover gracefully without losing user data.
function onError(error) {
  throw error;
}

const domainList = [
    "com",
    "google",
]

const Editor = () => {
  const URL_MATCHER = new RegExp(
    `(?:https?:\\/\\/)?[-a-zA-Z0-9@:%\\._\\+~#=]{1,256}\\.(?:${domainList.join(
      "|"
    )})\\b(?:[-a-zA-Z0-9()@:%_\\+.~#?&//=]*)`,
    "gi"
  );

  const MATCHERS = [
    (text: string) => {
      const match = URL_MATCHER.exec(text);

      console.log(text, match);

      if (match === null) {
        return null;
      }
      const fullMatch = match[0];
      return {
        index: match.index,
        length: fullMatch.length,
        text: fullMatch,
        url: fullMatch.startsWith("http") ? fullMatch : `https://${fullMatch}`,
        attributes: { rel: "noreferrer", target: "_blank" }, // Optional link attributes
      };
    },
  ];

  const initialConfig: InitialConfigType = {
[...]

PR #5318, opened 2023-12-01
fixed issue 4692 where 'www' would be accepted as domain name

mbongard:
Fixed issue https://github.com/facebook/lexical/issues/4692.

The problem occurred when manually typing a link with `www`, ie: `https://www.12345.com` instead of `https://12345.com` for domains with up to 6 characters.
In this case the Regex would match `www` as the domain and the actual domain as the TLD, because the TLD matches up to 6 characters.

image

We can get around this problem by not allowing `www` as the domain part using a negative lookahead.


zurfyx (commented):
Thank you, I believe the root cause is the Autolink plugin and its sibling lookup. The regex isn't perfect but it should still work based on the match.

We copy-pasted it from the top upvoted SO answer https://stackoverflow.com/a/3809435/2013580. Do you think we should borrow a different Regex?

Also, custom subdomains are very common, would this still work with your proposal or should we find a better Regex that can cater these better?

mbongard:
Hi,
You are right in that this does not address the root problem. The original regex does match both URLs in the example I gave.
This was supposed to be a work around. However, I had not considered other subdomains, which this hot fix does not address. I think you should close this PR, my apologies.

PR #5325, opened 2023-12-02
Fix insertNodes bugs

GermanJablo:
From Discord:

> there's 1 or 2 potential issues that we to iterate on asap or revert the commit and take a look at carefully.
> 
> This error has been reported multiple times among employees on clipboard paste before we reverted:
> 
> Cannot read properties of null (reading 'append')
> 
> Based on the trace this is pointing to:
> lastInsertedBlock.append(...insertedParagraph.getChildren());
> 
> Second and not yet sure if connected, when there's a new line at the end it crashes before the selection is lost (the line break node is removed before so we can't restore to it).

cc @zurfyx 

this replace https://github.com/facebook/lexical/pull/5324

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1596,19 +1596,17 @@ export class RangeSelection implements BaseSelection {
     const notInline = (node: LexicalNode) =>
       ($isElementNode(node) || $isDecoratorNode(node)) && !node.isInline();
 
-    let nodeToSelect = $isElementNode(last)
-      ? last.getLastDescendant() || last
-      : last;
     if (!nodes.some(notInline)) {
       const index = removeTextAndSplitBlock(this);
       firstBlock.splice(index, 0, nodes);
-      nodeToSelect.selectEnd();
+      last.selectEnd();
       return;
     }
 
     // CASE 3: At least 1 element of the array is not inline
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1596,18 +1596,17 @@ export class RangeSelection implements BaseSelection {
...
       firstBlock.splice(index, 0, nodes);
-      nodeToSelect.selectEnd();
+      last.selectEnd();
       return;
     }
 
     // CASE 3: At least 1 element of the array is not inline
-    const blocks = $wrapInlineNodes(nodes).getChildren();
+    const blocksParent = $wrapInlineNodes(nodes);
+    const nodeToSelect = blocksParent.getLastDescendant()!;
```
zurfyx, line 1608: Will `blocksParent` always have children?

zurfyx (approved):
Appreciate the quick iteration on this!

packages/lexical/src/LexicalSelection.ts
GermanJablo: In the current implementation yes

zurfyx: Oh you're right, I missed the fact that our starting point now is the parent virtual ParagraphNode instead of their children 🤦‍♂️

PR #5327, opened 2023-12-03
Refactor classname splitting from `' '` to `/\s+/`

sakib412:


zurfyx (commented):
Thanks, any real life example when this is needed?

sakib412:
> Thanks, any real life example when this is needed?

https://discord.com/channels/953974421008293909/1180070470935117884/1180070470935117884

image

acywatson:
> Thanks, any real life example when this is needed?

Yea if you have more than one space between classnames, the original implementation will fail. There's a thread about it on Discord.

sakib412:
> Almost feels like this could be extracted out to a util so we don't risk inconsistent class splitting logic in different places.

I have added a util function and testing

packages/lexical/src/LexicalUtils.ts
```diff
@@ -28,6 +28,7 @@ import type {
 import type {RootNode} from './nodes/LexicalRootNode';
 import type {TextFormatType, TextNode} from './nodes/LexicalTextNode';
 
+import {splitClasses} from '@lexical/utils';
```
acywatson, line 31: Thanks for making this change. However, we don't want the core to have any dependencies, including on other lexical packages, like @lexical/utils.

This means split classes needs to live in the core if it's used there, unfortunately.

acywatson (changes requested):

zurfyx:
Inactivity cleanup. Feel free to reopen

Issue #5330, opened 2023-12-04
Bug: Forward Line Deletion deletes two rows


ivailop7:
Lexical version: 0.12.4

When doing CMD + Delete (keycode 46, available on full-size keyboards, not on macbook laptops), it triggers a DELETE_LINE_COMMAND which is a forward line deletion, and it deletes 2 lines instead of 1 and having the cursor at the start of the line. Backwards line deletion (CMD + Backspace, keycode 8) behaves correctly.
https://github.com/facebook/lexical/assets/7893468/a825fb21-6c96-410e-acff-2697294d7613

PR #5331, opened 2023-12-04
Fix forward line deletion

ivailop7:
Fix for: https://github.com/facebook/lexical/issues/5330
Before video inside the Issue.

After:
https://github.com/facebook/lexical/assets/7893468/41a46b60-5c5f-4531-8a2e-af3af767958b

It does delete line by line, low FPS in the recording shows as if it deletes 2 at once.
I'm fairly sure, I'm not solving this correctly. Because after the line deletion, the cursor should be at the start of the next line. Which is currently not the behaviour.

@zurfyx @fantactuka @acywatson I need your help here for a proper fix, since it's in the core.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -21,6 +21,7 @@ import {
   $isDecoratorNode,
   $isElementNode,
   $isLineBreakNode,
+  $isParagraphNode,
   $isRootNode,
   $isTextNode,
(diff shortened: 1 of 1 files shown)


zurfyx:
Somehow the behavior seems correct to me, is this the correct shortcut? 

https://github.com/facebook/lexical/assets/193447/75dd0330-6580-46ac-9878-ec3bd8746910

PR #5334, opened 2023-12-05
[PR-5] Remove references of PointSelection

icrosil:
## Why
In #5276  we proposed a solution on how to migrate GridSelection out of the core of the bundle, here is the fourth step by removing GridSelection into lexical-table package.

## What
Removed all intermediate references of PointSelection.

- Replaced usages of PointSelection with BaseSelection and extra check if it's NodeSelection. If it's not Node then it's likely to be PointSelection.
- $insertGeneratedNodes function updates so if it's range in Grid - call mergeGrid function that is moved to RangeSelection, if it is otherwise but have anchor and focus treat it like pointSelection and create RangeSelection that will be transformed into GridSeleciton on a command dispatch.
- BaseSelection extended with next methods:
  - getStartEndPoints - which is used to return anchor and focus where available
  - back to interface to not break external dependencies
  - isCollapsed(): boolean;
  - isBackward(): boolean;
  - getCachedNodes(): LexicalNode[] | null;
  - setCachedNodes(nodes: LexicalNode[] | null): void;

## Discussion topic
At this moment we can observe a bit too much of code duplication between Selections, considering that we had to move most of differences into BaseSelection interface. From example in the stack with PointSelection it would've made sense to have BaseSelection as abstract class going forward.


icrosil:
@zurfyx, @fantactuka if you are still around do you mind reviewing this last piece?

acywatson (approved):

packages/lexical-list/src/formatList.ts
```diff
@@ -94,9 +93,10 @@ export function insertList(editor: LexicalEditor, listType: ListType): void {
   editor.update(() => {
     const selection = $getSelection();
 
-    if ($INTERNAL_isPointSelection(selection)) {
+    if (selection != null) {
```
zurfyx, line 96:
```suggestion
    if (selection !== null) {
```

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -270,7 +263,7 @@ function BlockFormatDropDown({
       editor.update(() => {
         let selection = $getSelection();
 
-        if ($INTERNAL_isPointSelection(selection)) {
+        if (selection != null) {
```
zurfyx, line 266:
```suggestion
        if (selection !== null) {
```

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -370,7 +363,7 @@ function FontDropDown({
     (option: string) => {
       editor.update(() => {
         const selection = $getSelection();
-        if ($INTERNAL_isPointSelection(selection)) {
+        if (selection != null) {
```
zurfyx, line 366:
```suggestion
        if (selection !== null) {
```

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -727,7 +720,7 @@ export default function ToolbarPlugin({
     (styles: Record<string, string>) => {
       activeEditor.update(() => {
         const selection = $getSelection();
-        if ($INTERNAL_isPointSelection(selection)) {
+        if (selection != null) {
```
zurfyx, line 723:
```suggestion
        if (selection !== null) {
```

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -321,9 +323,12 @@ function $patchStyle(
  * @param patch - The patch to apply, which can include multiple styles. { CSSProperty: value }
  */
 export function $patchStyleText(
-  selection: INTERNAL_PointSelection,
+  selection: BaseSelection,
```
zurfyx, line 326: Shouldn't this be RangeSelection? Once you get rid of GridSelection here you will have a similar problem and having the function do nothing is not a great behavior. How are you thinking about Grid?

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1825,11 +1793,13 @@ function getCharacterOffset(point: PointType): number {
...
 }
 
-function getCharacterOffsets(
-  selection: INTERNAL_PointSelection,
+export function $getCharacterOffsets(
+  selection: BaseSelection,
 ): [number, number] {
-  const anchor = selection.anchor;
-  const focus = selection.focus;
+  const [anchor, focus] = selection.getStartEndPoints();
```
zurfyx, line 1799: I'd be keen to leave the previous version, in this one you're constructing and deconstructing an array unnecessarily. Won't make a difference performance wise but I think the previous version was fine

Or you can change it to `const {anchor, focus} = ..`

packages/lexical-selection/src/range-selection.ts
```diff
@@ -110,13 +114,16 @@ function $removeParentEmptyElements(startingNode: ElementNode): void {
  * @param wrappingElement - An element to append the wrapped selection and its children to.
  */
 export function $wrapNodes(
-  selection: INTERNAL_PointSelection,
+  selection: BaseSelection,
   createElement: () => ElementNode,
   wrappingElement: null | ElementNode = null,
 ): void {
+  const [anchor, focus] = selection.getStartEndPoints();
+  if ($isNodeSelection(selection) || anchor == null || focus == null) {
```
zurfyx, line 122: I think we can wrap NodeSelection nodes just fine? I'd say either we implement it or we make the signature more strict

packages/lexical-selection/src/range-selection.ts
```diff
@@ -42,10 +43,16 @@ import {getStyleObjectFromCSS} from './utils';
...
  */
 export function $setBlocksType(
-  selection: INTERNAL_PointSelection,
+  selection: BaseSelection | null,
   createElement: () => ElementNode,
 ): void {
-  if (selection.anchor.key === 'root') {
+  if (selection == null || $isNodeSelection(selection)) {
+    return;
+  }
```
zurfyx, line 51: I think we can implement it for NodeSelection? I'd say either we implement it or we make the signature more strict

packages/lexical-rich-text/src/index.ts
```diff
@@ -582,11 +581,11 @@ export function registerRichText(editor: LexicalEditor): () => void {
         const selection = $getSelection();
 
         if (typeof eventOrText === 'string') {
-          if ($INTERNAL_isPointSelection(selection)) {
```
zurfyx, line 585: I wonder if we can remove this, because this hardcoded function will be very hard to override by anyone else who does custom selection, but we can revisit later

packages/lexical-rich-text/src/index.ts
```diff
@@ -439,7 +438,7 @@ function onPasteForRichText(
         event instanceof InputEvent || event instanceof KeyboardEvent
           ? null
           : event.clipboardData;
-      if (clipboardData != null && $INTERNAL_isPointSelection(selection)) {
+      if (clipboardData != null && selection != null) {
```
zurfyx, line 441: Can you give more details on the reasoning here? This is not the same as the condition on the left

zurfyx (changes requested):
Thanks, great work, this is a complicated PR, added some comments.

Couple of opinionated Lexical syntax stuff we do here:
- Prefer `===` to `==`
- Prefer `null | X` over `X | null`

Also, don't forget to run `npm run extract-codes` if you change invariants

packages/lexical-selection/src/lexical-node.ts
icrosil: Right now this function performs as single function to patch styles for any selection
Node -> noop
Range -> do
Not Range (technically Grid) -> create Range

personally I think we should leave it like this unless we want to get rid of all functions like this from the code, but that would be a great deal of refactoring.

packages/lexical/src/LexicalSelection.ts
icrosil: the problem here and reason why we need this new function is that Lexical usually tries to get anchors/focus from BaseSelection - like here and it could be nulls for NodeSelection, but proper in Range or Grid. So basically we had 2 ways of dealing with them:

make new class to get specialized types - PointSelection which we decided to get rid of.
or put some sort of anchor/focus knowledge in baseSelection - getStartEndPoints.

it's also similar to problem above because functions we cannot specify type and have to work backwards from BaseSelection, which seems like best approach here if we are allowing Selections to be created by users or package specific selections like Grid.

packages/lexical-selection/src/range-selection.ts
icrosil: ooh this is a good spot!

packages/lexical-rich-text/src/index.ts
icrosil: yeah so now clipboard data will try to insert text on NodeSelection too that I thought should be alright or caught by tests. But what we can do - check if selection is node and not allow it, wdyt?

icrosil:
Hey @zurfyx I've resolved most of the feedback and left comments for other.

ivailop7 (approved):
Great effort!

icrosil:
Hey @zurfyx I've updated the patchStyle that we discussed offline, changes - I was able to remove !isNodeSelection and !isRangeSelection parts, returning early if no anchor/focus and treating GridSelection same as Range, that was already working with 1 simple change.

Will update more if CI is red, please let me know if there is anything else you'd like to revisit.

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -326,35 +323,16 @@ function $patchStyle(
...
   patch: Record<
     string,
     string | null | ((currentStyleValue: string | null) => string)
   >,
 ): void {
   const selectedNodes = selection.getNodes();
```
zurfyx, line 332: `getNodes` will not capture all nodes except for RangeSelection, do we need to traverse them all to find at least all the TextNodes? We can exclude RangeSelection for the sake of performance.

I can help with this but let's at least document this in an issue and leave the GridSelection logic around so that we don't regress

zurfyx (commented):
Thanks for iterating on this! Added some comments (mostly minor stuff)

Also, you may want to check on why the playground errors upon click
Screenshot 2024-01-17 at 4 29 34 PM

packages/lexical-selection/src/lexical-node.ts
icrosil: So the only question here is how current method works for GridSelection, because NodeSelection is noop and RangeSelection works as expected.

I found that current patchStyle is actually working almost out of the box for Grid too. Here is how:

let's take in consideration 2 cases how GridSelection could be potentially involved:
- text selected inside of grid cell - A
- text selected across many grid cells - B

in case A - that's pure RangeSelection, which already will work as expected. For case B - we can safely assume that  we don't have case where not all grid cell is selected, because when we expand selection beyond 1 cell - it will set selection level for cells only. in this case GridSelection will have anchor -> cell start, focus -> cell end. Both elements would include all text inside without offsets.

Given that we need to make sure that pathStyle would work for case B. in the PR I've updated some logic in the end to make sure we are highlighting all cells, but answering your question:
getNodes will return Grid, Row, Cell, Paragraph and Text nodes (all of them) and current styling mechanism already filtering and applying style only to text blocks.

PS and only at this point I read your edit saying its alright xD

zurfyx (approved):
Thank you for iterating on this 🙏

PR #5335, opened 2023-12-05
Bug Fix: added a fix for discarding multiple color picker entries on drag

bhavyakaria:
fix: https://github.com/facebook/lexical/issues/3890

Before:

https://github.com/facebook/lexical/assets/16178833/04d1e8d9-dabb-481b-9a33-c54d9363e7ad

After:

https://github.com/facebook/lexical/assets/16178833/455e9222-7da8-448e-bcc1-67641b4639df


bhavyakaria:
PR is ready for review.

packages/lexical-playground/src/ui/ColorPicker.tsx
```diff
@@ -186,14 +203,21 @@ function MoveWrapper({className, style, onChange, children}: MoveWrapperProps) {
     move(e);
 
     const onMouseMove = (_e: MouseEvent): void => {
+      setDragged(true);
+      updateSkipHistoryFlag(true);
```
zurfyx, line 207: Do we need a `useState` here or can we do with `useRef`? It might be easier to debug later without the side effects of having this variable is the useEffect dependency. I believe even a module variable would suffice for this case since there can only be one event in total in the entire page

zurfyx (commented):
Thank you for the fix!

packages/lexical-playground/src/ui/ColorPicker.tsx
bhavyakaria: Thanks for the review @zurfyx 

Agreed, `useState` is not required, we can use `useRef`, just to clarify we are talking about just the `dragged` variable, right?

bhavyakaria: @zurfyx Thanks for the review.

Agreed, `useState` was not necessary, have made he skipFlag to be module variable and `dragged` to be `useRef`

bhavyakaria:
PR is ready for review

ivailop7 (approved):
LGTM

PR #5342, opened 2023-12-06
refactor(playground): kill ts-ignore

agriffis:
Replace with ts-expect-error where still needed.


thegreatercurve:
Thanks for this @agriffis! Would you mind fixing the conflict and I can merge this in.

thegreatercurve (approved):

agriffis:
@thegreatercurve Done. Note that this commit affects a file that should actually be deleted, see https://github.com/facebook/lexical/pull/5237#issuecomment-1847165683

but merging this doesn't prevent that, it's just a question of ordering. :shrug:

agriffis:
In fact, the reason the other PR went through cleanly was because `ts-ignore` hid the problem.

agriffis:
Blech, I'll just update this PR to delete the file

agriffis:
Ok all set

agriffis:
@thegreatercurve This has two commits now so it should be clear what's going on

zurfyx (approved):
LGTM, thank you, commits gets squashed once the PR is landed!

PR #5348, opened 2023-12-07
Fix collapseAtStart on ParagraphNode, HeadingNode

### LINKED ISSUE #5350: Bug: DELETE_CHARACTER_COMMAND pushes meaningless histories
Lexical version: latest

## Steps To Reproduce

1. press `DELETE` or `BACKSPACE` on the very first position (offset === 0) with collapsed selection, without any content and sibling.
2. then it pushes additional histories per each `DELETE_CHARACTER_COMMAND` for doing nothing.

Link to code example: can reproduce it on playground.

## The current behavior

https://github.com/facebook/lexical/assets/40269597/44603988-7473-4cf3-b74f-e019b8d6e43c

## The expected behavior

https://github.com/facebook/lexical/assets/40269597/05785732-bb78-4e50-94a5-9ebade2c195a

2wheeh:
fix #5350

### 1. ParagraphNode on empty editor, `DELETE`, `BACKSPACE`

before: 

https://github.com/facebook/lexical/assets/40269597/276b33d1-aaef-4a08-9d1c-d8b5cd46086d

after:

https://github.com/facebook/lexical/assets/40269597/3ed27818-48a2-48be-a72b-7d765c2faf0b

### 2. HeadingNode when it has text content, `BACKSPACE`

before:

https://github.com/facebook/lexical/assets/40269597/0a1b2c3c-da8d-4025-9638-b52678bfbdc3

after:

https://github.com/facebook/lexical/assets/40269597/86506853-88e2-4f7c-99ac-2e92daca8908

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -2048,7 +2048,7 @@ export class RangeSelection extends INTERNAL_PointSelection {
           }
         }
         $updateCaretSelectionForUnicodeCharacter(this, isBackward);
-      } else if (isBackward && anchor.offset === 0) {
+      } else if (anchor.offset === 0) {
         // Special handling around rich text nodes
         const element =
           anchor.type === 'element'
--- a/packages/lexical/src/nodes/LexicalParagraphNode.ts
+++ b/packages/lexical/src/nodes/LexicalParagraphNode.ts
@@ -131,14 +131,14 @@ export class ParagraphNode extends ElementNode {
       if (nextSibling !== null) {
         this.selectNext();
         this.remove();
-        return true;
+      } else {
+        const prevSibling = this.getPreviousSibling();
+        if (prevSibling !== null) {
+          this.selectPrevious();
+          this.remove();
+        }
       }
-      const prevSibling = this.getPreviousSibling();
-      if (prevSibling !== null) {
-        this.selectPrevious();
-        this.remove();
-        return true;
-      }
+      return true;
     }
     return false;
   }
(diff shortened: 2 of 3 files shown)


ivailop7:
Could we get a description on the issue this is fixing?

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -131,14 +131,14 @@ export class ParagraphNode extends ElementNode {
...
-        this.remove();
-        return true;
+      } else {
+        const prevSibling = this.getPreviousSibling();
+        if (prevSibling !== null) {
+          this.selectPrevious();
+          this.remove();
+        }
       }
+      return true;
```
2wheeh, line 141: `collpaseAtStart` should return `true` not to call `removeText()` even through it has no siblings, because here editor is empty.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2048,7 +2048,7 @@ export class RangeSelection extends INTERNAL_PointSelection {
           }
         }
         $updateCaretSelectionForUnicodeCharacter(this, isBackward);
-      } else if (isBackward && anchor.offset === 0) {
+      } else if (anchor.offset === 0) {
```
2wheeh, line 2051: This change prevents the addition of unnecessary history on a forward deletion on an empty editor.

But with this, a forward deletion acts different from now:
when there is no text content, forward deletion breaks the type (forward deletion does this on notion editor) and I think `isBackward` here is intended to block doing this ...🤔

PR #5354, opened 2023-12-08
Clear link using key shortcut bug

### LINKED ISSUE #5353: Bug: Toggle hyperlink using keyboard shortcut doesn't clear the link
Lexical version: v0.12.5 (latest)

## Steps To Reproduce

1. Open lexical playground
2. Select some text
3. Press Ctrl+K (or Cmd+K) to create hyperlink
4. Type in a URL in the floating link editor
5. Press the `enter` key to confirm the URL
6. Select the same text again and press Ctrl+K 
7. This should clear the URL, but instead URL changes to `https://`

This bug is only there when using the keyboard shortcut (Ctrl+K). The toolbar buttons for toggling hyperlink doesn't have this bug.

## The current behavior
Here is a hyperlink in the playground:

image

Press Ctrl+K to toggle:

image

## The expected behavior
The expected behavior is for the URL to be cleared:

image

umaranis:
This PR fixes #5353

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
@@ -66,6 +66,9 @@ function FloatingLinkEditor({
       } else {
         setLinkUrl('');
       }
+      if (isLinkEditMode) {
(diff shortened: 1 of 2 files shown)


zurfyx (approved):
LGTM, thank you!

Issue #5162, opened 2023-10-24
Bug: 


matsuyama-k1:
Lexical version:
V0.12.2

## Steps To Reproduce

1. Create a Node that inherits TextNode and canInsertTextBefore is false. (ex. HashtagNode)
![Screenshot 2023-10-24 at 22 21 41](https://github.com/facebook/lexical/assets/111737064/b686b941-b654-48ee-89e3-b4fc54a2b036)

2. Select text from the beginning of the corresponding node
![Screenshot 2023-10-24 at 22 30 48](https://github.com/facebook/lexical/assets/111737064/a2fc309a-1d4f-464e-a4b1-5c422f059328)

3. Type Asian language such as Japanese and Chinese. (Languages that trigger node.isComposing() to be true)

4. In addition to not displaying text input, it becomes impossible to change the selection using mouse clicks, etc.
![Screenshot 2023-10-24 at 22 38 12](https://github.com/facebook/lexical/assets/111737064/9be328ef-3141-40af-a5d6-64ef30f3cc0c)

6. You can return Selection to selectable by pressing enter, but the design will collapse and the entered text will not be reflected.
![Screenshot 2023-10-24 at 22 46 03](https://github.com/facebook/lexical/assets/111737064/6019bc08-4b5d-4c74-ac4f-78a555dc6701)

Link to code example:
https://playground.lexical.dev/

## The current behavior
If you select a Node for which canInsertTextBefore is False from the beginning and enter an Asian language that sets node.isComposing() to True, the text will not be displayed and the selection will not be able to be changed.

## The expected behavior
Even if you select a Node for which canInsertTextBefore is False from the beginning and enter an Asian language that sets node.isComposing() to True, act as same with English input.

matsuyama-k1:
Maybe the situation is related to this function.
https://github.com/facebook/lexical/blob/df2a50bc88e0778af26e109502cfcfb9cbe245d5/packages/lexical/src/LexicalUtils.ts#L598

mayrang:
This was fixed by PR #5363 (merged December 2023), which added an `!isComposing` guard to the `canInsertTextBefore` check in `$updateTextNodeFromDOMContent`. The fix is included in v0.12.6+. Does anyone still reproduce this on a recent version?

PR #5363, opened 2023-12-10
Fixed: for #5162, The beginning of a TextNode with canInsertTextBefore false in …

matsuyama-k1:
Fix for #5162

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -19,18 +19,18 @@ import type {
   RegisteredNodes,
   Spread,
 } from './LexicalEditor';
-import type { EditorState } from './LexicalEditorState';
-import type { LexicalNode, NodeKey, NodeMap } from './LexicalNode';
+import type {EditorState} from './LexicalEditorState';
+import type {LexicalNode, NodeKey, NodeMap} from './LexicalNode';
 import type {
   BaseSelection,
   PointType,
   RangeSelection,
 } from './LexicalSelection';
-import type { RootNode } from './nodes/LexicalRootNode';
-import type { TextFormatType, TextNode } from './nodes/LexicalTextNode';
+import type {RootNode} from './nodes/LexicalRootNode';
+import type {TextFormatType, TextNode} from './nodes/LexicalTextNode';
 
-import { CAN_USE_DOM } from 'shared/canUseDOM';
-import { IS_APPLE, IS_APPLE_WEBKIT, IS_IOS, IS_SAFARI } from 'shared/environment';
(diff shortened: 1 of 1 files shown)


matsuyama-k1:
When multiple strings are selected from the beginning of a TextNode with $canInsertTextBefore false and input in languages where composing is performed, such as Japanese and Korean, the text input is not reflected and the editor will not accept input until enter is pressed.
I addressed this problem.
Perhaps the cause of the problem is that in languages where composing occurs, COMPOSITION_START_CHAR is entered by default, so two characters are input.
In my understanding, the onInput function is executed when a node is created and the second character is input, but in Japanese and other languages, it works when a single character is input.
Therefore, canInsertTextBefore malfunctions.
However, it was not possible to directly prevent the problem when two characters are input, so I added an if statement condition to solve the most recent problem.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -650,7 +650,8 @@ export function $updateTextNodeFromDOMContent(
             prevSelection.anchor.offset === 0) ||
             (prevSelection.anchor.key === textNode.__key &&
               prevSelection.anchor.offset === 0 &&
-              !node.canInsertTextBefore()) ||
+              !node.canInsertTextBefore() &&
+              !isComposing) ||
```
zurfyx, line 654: Can we do the same on L657 for `canInsertTextAfter`?

zurfyx (approved):
Thanks!

From a correctness perspective this PR is not great, the `canInsertBefore` was meant to prevent any input at boundaries which is what you're doing here.

Unfortunately, composition is not flexible, and we can't reposition the cursor when that happens, and with the MutationObserver logic combined is the reason why you're not seeing any text.

I believe the best solution would be to swap DOM nodes but that is something we don't support now and might get tricky with custom extended nodes.

Given that this change improves a11y, I'm inclined to accept it despite the correctness trade-off.

packages/lexical/src/LexicalUtils.ts
matsuyama-k1: I have inserted the line within the `canInsertTextAfter` context and confirmed that the jest tests run correctly. Additionally, I attempted to run `the npm run test-e2e-chromium` command for end-to-end testing, but the execution time was excessively long, preventing the tests from completing. I haven't encountered any specific scenarios that cause bugs when the line is not inserted within canInsertTextAfter, but it appears to function properly when the line is included.

zurfyx: Makes sense, yeah, some E2E tests are flaky and take multiple times to pass unfortunately

PR #5368, opened 2023-12-11
Fix MarkdownShortcut to transfrom for heading with HashtagNode

### LINKED ISSUE #5366: Bug: Hashtag blocks transformation to Headings
Lexical version: 0.12.5 (Latest)

## Steps To Reproduce

1. go to the Playground
2. write some text (or heading, ... )
3. put selection on the very first position
4. enter #, ##, ...  + ' ' to transform to heading

Link to code example: it can be reproduced on Playground

## The current behavior

https://github.com/facebook/lexical/assets/40269597/3bf0e8e5-335a-4739-83f6-36624d997420

## The expected behavior
It transforms to heading by `HEADING` transformer

2wheeh:
Fix #5366 

before: https://github.com/facebook/lexical/issues/5366#:~:text=The%20current%20behavior-,before,-%2Dheading%2D01.mov

after: 

https://github.com/facebook/lexical/assets/40269597/361433c4-2717-4aac-b152-3cb3f1918a13

### DIFF
--- a/packages/lexical-markdown/src/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/MarkdownShortcuts.ts
@@ -15,6 +15,7 @@ import type {
 import type {ElementNode, LexicalEditor, TextNode} from 'lexical';
 
 import {$isCodeNode} from '@lexical/code';
+import {$isHashtagNode} from '@lexical/hashtag';
 import {
   $createRangeSelection,
   $getSelection,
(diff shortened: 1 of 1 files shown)


2wheeh:
So it wasn't a root cause and this kind of coupling wouldn't get welcomed :( 
I think the bug is around hashtag regex (https://github.com/facebook/lexical/issues/5366#issuecomment-1851436578)

Issue #5369, opened 2023-12-11
Bug: `INTERNAL_$isBlock` treats decorator nodes incorrectly


9larsons:
Lexical version: v0.12.3+

## Steps To Reproduce

1. Select a range of text that includes a decorator node, e.g. line break.
2. Attempt to apply a block change like the Header 1 option.
3. Observe that nothing happens. This appears to also break selection in the Playground and produce an error.

https://github.com/facebook/lexical/assets/21961100/35db0154-5bed-4811-a5d5-a9f022aecb2e

## The current behavior
Block types cannot be set if the selection includes a decorator node.

## The expected behavior
Block types skip decorator nodes.

---

This is a result of the larger `insertNodes` changes in v0.12.3+.

Causal code:
https://github.com/facebook/lexical/pull/5002/files#diff-33353db27f56a5d74e7281df0a8a2fc223eef41e9dc2efc723a412560a0148c9R561-R563

Previously, `isBlock` skipped over decorator nodes (returned false) without specifically calling them out. I'm not sure what the reasoning was behind adding the check in `INTERNAL_$isBlock` for decorator nodes, but we should be skipping these entirely as the remaining logic in `$setBlocksType` is particular to element nodes.

PR #5371, opened 2023-12-11
Fix `DecoratorNode`s being treated as block elements

9larsons:
closes https://github.com/facebook/lexical/issues/5369

This is a result of the larger insertNodes changes in v0.12.3+.

Causal code:
https://github.com/facebook/lexical/pull/5002/files#diff-33353db27f56a5d74e7281df0a8a2fc223eef41e9dc2efc723a412560a0148c9R561-R563

`INTERNAL_$isBlock` is treating (non-inline) `DecoratorNode`s as block elements, causing `ElementNode` methods being called on decorator nodes that result in broken functionality.

While someone could extend the decorator node class to have the indent and format properties, they are generally not elements and should not be treated as such.

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -571,8 +571,8 @@ export function $getSelectionStyleValueForProperty(
 export function INTERNAL_$isBlock(
   node: LexicalNode,
 ): node is ElementNode | DecoratorNode<unknown> {
-  if ($isDecoratorNode(node) && !node.isInline()) {
-    return true;
+  if ($isDecoratorNode(node)) {
+    return false;
   }
   if (!$isElementNode(node) || $isRootOrShadowRoot(node)) {
     return false;
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -7,7 +7,6 @@
  */
 
 import type {
-  DecoratorNode,
   ElementNode,
   INTERNAL_PointSelection,
   LexicalNode,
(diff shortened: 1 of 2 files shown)


packages/lexical-selection/src/range-selection.ts
```diff
@@ -571,8 +571,8 @@ export function $getSelectionStyleValueForProperty(
 export function INTERNAL_$isBlock(
   node: LexicalNode,
 ): node is ElementNode | DecoratorNode<unknown> {
```
GermanJablo, line 573: Then the return type should be `node is ElementNode`, right?

GermanJablo (commented):
I'm surprised that no tests are breaking, since I made `insertNodes` considering the `decorators` (which in fact before were not included as blocks and I had to include them for it to work).

I haven't reviewed `insertNodes` in detail again, so for my part the only thing I can point out at a glance is a comment about the return type.

packages/lexical-selection/src/range-selection.ts
acywatson: I think this is right. @9larsons can you make the change and we can merge?

9larsons: Done! Sorry for the delay, I was out for a few weeks. Thanks for taking a look.

zurfyx (approved):
Thanks for looking into this! The terminology for this function is not correct. We have originally referred to blocks as a piece of content that ocuppies the entire horizontal space, resembling `display: block` on the browser. Decorator nodes can be blocks so I'd be inclined the say that we have to fix this instead https://github.com/facebook/lexical/blob/main/packages/lexical-selection/src/range-selection.ts#L77

Stamping to unblock since it's been about a month since this was open but ideally we want to iterate on this

PR #5376, opened 2023-12-12
Fix hashtag regex blocking transform to heading

### LINKED ISSUE #5366: Bug: Hashtag blocks transformation to Headings

2wheeh:
Fixes #5366

This prevents multiple hash chars on hashtag, which blocks transformation to heading.

before:

https://github.com/facebook/lexical/issues/5366#:~:text=The%20current%20behavior-,before,-%2Dheading%2D01.mov

after:

https://github.com/facebook/lexical/assets/40269597/d7804367-57fe-4188-b21c-8ea906f21921

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs
@@ -302,4 +302,46 @@ test.describe('Hashtags', () => {
       `,
     );
   });
+
+  test('Should break when additional # added in front of it', async ({
+    page,
+  }) => {
+    await focusEditor(page);
+    await page.keyboard.type('#hello');
+
+    await waitForSelector(page, '.PlaygroundEditorTheme__hashtag');
+
+    await assertHTML(
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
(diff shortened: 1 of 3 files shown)


2wheeh:
I thought additional coupling between `@lexical/markdown` and `@lexical/hashtag` wouldn't be welcomed so I closed #5368.
But now on my second thought, It could be better than modifying regex when considering the existing services depending on lexical.

I would greatly appreciate it if you could advise me which is correct approach @acywatson

acywatson:
> I thought additional coupling between `@lexical/markdown` and `@lexical/hashtag` wouldn't be welcomed so I closed #5368. But now on my second thought, It could be better than modifying regex when considering the existing services depending on lexical.
> 
> I would greatly appreciate it if you could advise me which is correct approach @acywatson

Sorry for the delay....this seems right to me!

acywatson (approved):

PR #5382, opened 2023-12-13
dropdown text visibility problem fixed for medium and smaller screens

lSelectral:
When screen size gets smaller, dropdown text becomes invisible so, editor toolbar buttons stay visible.
But dropdown input element text become invincible too.

I fixed this problem by checking if it is a dropdown that isn't in editor and make it visible again.

BEFORE
BEFORE

---

AFTER
AFTER

Chrome: 119.0.6045.200 
Windows: 22H2 OS Build 19045.3758

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -802,6 +802,10 @@ i.page-break,
     display: none !important;
   }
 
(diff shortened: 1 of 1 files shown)


lSelectral:
@ivailop7 Can you please take a look at that?

ivailop7 (approved):

PR #5383, opened 2023-12-13
Feature/excalidraw resize

lSelectral:
Resizing excalidraw figures was already exist. But width and height hasn't saved to anywhere.
So I store the width and height in Excalidraw node then apply style to span element and make the span element, **inline-block** so it can change size while still being span element.


ivailop7 (approved):
LGTM

PR #5389, opened 2023-12-14
fix: Reset compositionKey when backspacing a selection on Android where anchor and focus keys are different

### LINKED ISSUE #5259: Bug: When using Gboard on Android, doing "Select all" then deleting the text doesn't seem to correctly stop the select all action
Lexical version: 0.12.4

## Steps To Reproduce

1. Open the playground on an Android browser (I've tested on both Chrome and Firefox)
2. Make sure you're using Gboard as the keyboard
3. "Select all" text
4. Delete the text using the backspace button in Gboard

## The current behavior

Once you "Select all" and delete the text, the text will correctly get deleted. However, if you start typing again, you'll notice that the select all action doesn't seem to have stopped and as you keep typing all the text will keep getting selected and replaced by the next character, making it very hard to actually type more than a character.

Here's a screen recording of the behavior:

https://github.com/facebook/lexical/assets/8348101/e5204cc3-c95a-43c9-96ac-6de86a90eada

## The expected behavior

Once you delete all the text selected using the "Select all" action, it should stop automatically selecting text when you start typing again.

## Additional context

This issue seems to only happen when using Gboard and not other keyboards. I tested with Samsung Keyboard as well.
Also, this issue is not present in v0.12.2 or earlier.

amanharwara:
fixes #5259

PR #5077 added a fix for the android backspace bug where pressing backspace delete two characters at once, by setting the `compositionKey` to `selection.anchor.key` when on Android. However, in cases where the `anchor.key` and `focus.key` are different, the `compositionKey` isn't cleared causing an issue where the selection wouldn't go away as described in #5259.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -558,6 +558,7 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
             dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
           }
         } else {
+          $setCompositionKey(null);
           event.preventDefault();
           dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
         }


acywatson:
Looks right, thank you!

acywatson (approved):

PR #5397, opened 2023-12-16
fix: Text format getting lost when using suggestion/autocorrect on iOS

### LINKED ISSUE #4801: Bug: Text format lost after auto-correcting on iOS
On an IOS device, if the user typed a misspelled word with formats like Bold or Italic, after clicking the autocorrected work on the native keyboard, it replaces the word with the correct one but the formats are lost.

Example from lexical playground website and an IOS simulator (note: it also happens on physical devices)
autocorrect-lexical

Lexical version: 0.11.2

## Steps To Reproduce

1. On a lexical editor type a word with formatting applied
2. Make sure it is a misspelled word so the native keyboard prompts you to correct it
3. Tap on the native keyboard suggestion and the word will be corrected but the formats are lost

Link to code example: It happens in the playground https://github.com/facebook/lexical/tree/main/packages/lexical-playground (see above screen record)

## The current behavior
After autocorrecting a word on IOS the text formats are lost

## The expected behavior
After autocorrecting a word on IOS the text formats should be kept

amanharwara:
fixes #4801 

Auto-correct/suggestions on iOS (and macOS?) trigger the `insertReplacementText` input type event, and inserting the nodes generated from the `text/html` data cause the formatting to be lost.
That isn't an issue if only the `text/plain` data is used to insert the text. With the `insertReplacementText` there's probably no need to be using the HTML data anyways.

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -124,6 +124,7 @@ export function $insertDataTransferForRichText(
   dataTransfer: DataTransfer,
   selection: BaseSelection,
   editor: LexicalEditor,
+  event?: InputEvent,
 ): void {
   const lexicalString = dataTransfer.getData('application/x-lexical-editor');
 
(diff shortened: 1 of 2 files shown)


zurfyx (commented):
Thanks for digging into this! Have we explored the option of merging the HTML with the previously written content? What is the HTML value we're getting from the iPhone? The reason why I'm saying this is because while Apple might always be consistent, password managers, spellcheckers and others may rely on the HTML passed through this event

amanharwara:
That's a good point. I'll need to look into it more.

BTW, this [comment](https://github.com/facebook/lexical/issues/4801#issuecomment-1858764112) I made on the issue has more info about the stacktrace of what happens when you press a suggestion

amanharwara:
> What is the HTML value we're getting from the iPhone?

The HTML value seems to be just plaintext, which then leads to a `TextNode` being generated from it.

redstar504:
I have also been facing this issue and proposed another solution in https://github.com/facebook/lexical/pull/5789

etrepum:
According to #5789 this has been fixed

PR #5407, opened 2023-12-20
docs: correct custom node examples

inaseem:
I noticed the examples for CustomParagraph had incorrect types on constructor and few imports were also missing in [Nodes](https://lexical.dev/docs/concepts/nodes) page. Since those are copy paste sample, the examples should just work without any modification. Corrected the types and added the missing import in them.


acywatson (approved):

acywatson:
Thank you!

Issue #5333, opened 2023-12-05
Feature: Export more functions from Lexical core package


yf-yang:
This feature request comes from https://github.com/facebook/lexical/issues/5313

# Description
1. export all exports of LexicalUtils.ts to lexical's users
2. export `getActiveEditor` and `getActiveEditorStates` from LexicalUpdate.ts

# Motivation
1. Utility functions benefit both internal library maintainers and library users. For example, `$getNodeByKeyOrThrow`, as a variant of `$getNodeByKey` with type guard, is useful for TypeScript users. Currently I have to copy paste the source again somewhere because it is not exported from the core package. More annoyingly, those functions appear in auto completion, but they are actually not accessible.
2. `getActiveEditor` and `getActiveEditorStates` are not commonly used, but they are pretty useful when trying to access the editor instance within a `LexicalNode`'s scope, especially in `importJSON` method. For my personal use case, that would be useful to map a node to editor-specific contextual data. That would also be useful when using Lexical along with other state management libraries since Lexical may not be the only one. Currently, there is no way to backlink a node to its containing editor.

# Considerations
Some considerations specifically related to 2:
- Export `getActiveEditorStates` may not be so useful since `$getRoot` is also available, it is enough to access the whole tree from a node.
- I am not sure by design `importJSON` should be treated as a pure function. The use case is that I do need to interact with external stuff when moving nodes between nested editors (which needs serialization and deserialization), but Lexical does not offer an api to deal with side effects.

acywatson:
I don't have a problem with this - I opened a PR to do it a long time ago, but @zurfyx had some light reservations:

https://github.com/facebook/lexical/pull/4325

PR #5409, opened 2023-12-21
feat: Expose getActiveEditor and getActiveEditorState

yf-yang:
Partially implements feature request #5333

Submit a pull request and waiting for the team's opinion.

By the way, do we have any bot that builds this commit online so that I can directly install this commit via npm? Need this one to access parent editor during `importJSON` for my own purpose.

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -1615,3 +1615,19 @@ export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
   }
   return predicate(parent) ? parent : null;
 }
+
+/**
+ * Utility function for accessing current active editor instance.
+ * @returns Current active editor
+ */
+export function $getEditor(): LexicalEditor {
+  return getActiveEditor();
+}
+
+/**
+ * Utility function for accessing current active editor's internal state.
+ * @returns Current active editor's state
+ */
+export function $getEditorState(): EditorState {
+  return getActiveEditorState();
+}
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -1623,11 +1623,3 @@ export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
 export function $getEditor(): LexicalEditor {
   return getActiveEditor();
 }
-
-/**
- * Utility function for accessing current active editor's internal state.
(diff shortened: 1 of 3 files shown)


acywatson:
If we do expose them, I'd like to add typedoc doc blocks to them so they show up in the API docs and people have a sense of how they work (e.g., what is the concept of "active" editor)

yf-yang:
@acywatson I've added two $-prefixed functions. Do you have any comments on the naming/doc?

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1615,3 +1615,19 @@ export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
...
+  return getActiveEditor();
+}
+
+/**
+ * Utility function for accessing current active editor's internal state.
+ * @returns Current active editor's state
+ */
+export function $getEditorState(): EditorState {
+  return getActiveEditorState();
+}
```
zurfyx, line 1633: The pending editor state was always meant to be kept private (by design), that's why we provide a rich set of utilities to modify its content and should be treated as an implementation detail. It's also one of the mutable parts of Lexical. Hence, it is dangerous to expose externally.

Can you provide more details on why you think it's important to expose this please?

yf-yang: Yes, I agree with you. Let me try to summarize my thoughts:
`$getEditorState` is almost equivalent to `$getRoot` and `$getSelection`, so it is not necessary.
However, by exposing `$getEditor`, then one can always do `$getEditor().getEditorState()`, which is equivalent (isn't it? not so sure), so I just added it here without too much thoughts.
I am OK to remove this API, but your comment makes me consider if exposing `$getEditor` is a good idea. The only reason to expose `$getEditor` is to access the editor from within a node because you can always do so when calling `editor.update` or `editorState.read` by accessing the parent scope. What's your opinion?

zurfyx: `$getEditor().getEditorState()` always returns the rendered EditorState (as opposed to the pending). `$getRoot` allows you to read or manipulate the rendered or pending editor state without any knowledge on whether the EditorState or nodes have been cloned. That is the reason why we're inclined to think that pending editor state should always be kept private!

zurfyx (approved):
Thanks for iterating on this, LGTM!

PR #5418, opened 2023-12-25
Fix inserting redundant paragraph on enter in MarkNode

strdr4605:
Related to
- f15a17564b6267531732ba5cc824b8aaf45f5b54 (these changes introduced the bug)

I am wondering if we can find a general solution, to fix the issue for any node.

Before

https://github.com/facebook/lexical/assets/16056918/da3c99d6-45b6-4a6f-9db9-390210faf879

After

https://github.com/facebook/lexical/assets/16056918/780cda16-cd1e-4e2c-b267-3ed2f220bd61

### DIFF
--- a/packages/lexical-mark/src/MarkNode.ts
+++ b/packages/lexical-mark/src/MarkNode.ts
@@ -20,12 +20,7 @@ import {
   addClassNamesToElement,
   removeClassNamesFromElement,
 } from '@lexical/utils';
-import {
-  $applyNodeReplacement,
-  $isElementNode,
-  $isRangeSelection,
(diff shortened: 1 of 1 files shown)


acywatson:
Interesting. Will let @GermanJablo comment.

GermanJablo:
> I am wondering if we can find a general solution, to fix the issue for any node.

`insertParagraph` works with links, which are also inlines. I don't know what was intended with the previous version of `MarkNode.insertNewAfter`, but apparently that was the problem.

LGTM.

zurfyx (approved):

PR #5431, opened 2023-12-29
docs: fix outdated importDOM types in docs

thorn0:


packages/lexical-website/docs/concepts/serialization.md
```diff
@@ -145,27 +143,7 @@ static importDOM(): DOMConversionMap | null {
 
 If the imported ```<table>``` doesn't align with the expected GitHub code HTML, then we return null and allow the node to be handled by lower priority conversions.
 
-Much like `exportDOM`, `importDOM` exposes APIs to allow for post-processing of converted Nodes. The conversion function returns a `DOMConversionOutput` which can specify a function to run for each converted child (forChild) or on all the child nodes after the conversion is complete (after). The key difference here is that ```forChild``` runs for every deeply nested child node of the current node, whereas ```after``` will run only once after the transformation of the node and all its children is complete. Finally, `preformatted` flag indicates that nested text content is preformatted (similar to `<pre>` tag) and all newlines and spaces should be preserved as is.
```
acywatson, line 148: Did we delete the part explaining the "preformatted" flag?

thorn0: Such a flag doesn't exist anywhere in the code

acywatson: huh.

acywatson (approved):

PR #5436, opened 2024-01-02
fix: convertTextDOMNode now correctly displays colors

SenaHuang:


ivailop7:
Is there an open Issue for this PR?

SenaHuang:
@ivailop7 I created one for this bug #5441

ivailop7:
I'm afraid this won't be merged, because this is a rich-text specific feature and not including it in the core was a deliberate choice some time ago to keep the core library minimal because of people who want to use the framework for plain text and other use cases that don't include rich text. This is why the example in the documentation was provided on how to address the missing HTML styles, color and backgroundColor are just 2 of a very large number of possible style attributes, so even this won't address the core issue at hand. This was discussed a few times in previous GitHub issues.

zurfyx (changes requested):
Thank you, this is intentional but not ideal (to fit within our product code). I believe we'll want to introduce a sort of feature flags or an easy way to do replacement and patch this logic in. @acywatson introduced https://github.com/facebook/lexical/pull/4254 recently

I'll bring this up to the rest of the team but unfortunately I don't think we can merge this PR as-is and we'll have to find another way :(

Fetz:
@SenaHuang thanks for your contribution, like others said, supporting paste of colors doesn't fit most of use-cases.

I think the best solution is to either improve our documentation on how to create custom paste and/or make it easier and more flexible for devs to create their own custom paste to fit use cases like this without requiring modifying this code.

Issue #5422, opened 2023-12-26
Bug: drop-down menu will not be loaded  in a sub-application loaded by qiankun


ninawangyimei:
Problem
---
When using Lexical to develop a rich text editor in a sub-application loaded by qiankun, if the node is added to document.body, the style of the drop-down menu will not be loaded, and the function will eventually be abnormal.

Solution
packages/lexical-react/src/shared/LexicalMenu.ts
useMenuAnchorRef function add a parameter
```
export function useMenuAnchorRef(
  resolution: MenuResolution | null,
  setResolution: (r: MenuResolution | null) => void,
  className?: string,
  // add a parameter
  anchorElementRefContainer?: HTMLElement,
}) {
   ...
    if (!containerDiv.isConnected) {
        if (className != null) {
          containerDiv.className = className;
        }
        containerDiv.setAttribute('aria-label', 'Typeahead menu');
        containerDiv.setAttribute('id', 'typeahead-menu');
        containerDiv.setAttribute('role', 'listbox');
        containerDiv.style.display = 'block';
        containerDiv.style.position = 'absolute';
       // added to a specific container
        (anchorElementRefContainer || document.body).append(containerDiv);
      }
   ...
}
```

ivailop7:
@ebads67 I believe https://github.com/facebook/lexical/pull/5481 fixes this, right?

ninawangyimei:
yes, #5481 fixed this problem.thanks!

PR #5437, opened 2024-01-02
fix: allow anchorElementRef node to be added to a specific container

ninawangyimei:
Resolves: [#5422](https://github.com/facebook/lexical/issues/5422)

@zurfyx  @fantactuka  @acywatson


zurfyx (changes requested):
Thank you, while the proposal looks reasonable to me there's a few consitently failing tests, would you mind looking into this please?

ninawangyimei:
> Thank you, while the proposal looks reasonable to me there's a few consitently failing tests, would you mind looking into this please?

Thanks for the review，The changes in [#5481](https://github.com/facebook/lexical/pull/5481)  are consistent with this commit, so I closed the merge directly.

PR #5445, opened 2024-01-03
Allow setting className for checklist in theme

amanharwara:
Right now there's no easy way to style the list element of a checklist specifically since it uses the same className as a `ul` and doesn't have something like a data attribute that could be used as a selector. This PR allows setting a className to a checklist specifically.

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -221,6 +221,7 @@ function setListThemeClassNames(
     const listClassName = listTheme[node.__tag];
     let nestedListClassName;
     const nestedListTheme = listTheme.nested;
+    const checklistClassName = listTheme.checklist;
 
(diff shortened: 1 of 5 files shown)


zurfyx (approved):
LGTM, thank you! I believe the workaround with CSS is `ul:has(> .PlaygroundEditorTheme__listItemUnchecked)` but it's complicated and somewhat frown upon with frameworks like Stylex

amanharwara:
Oh no, not sure how this happened

amanharwara:
@zurfyx Tests should be good now. The ones that have failed seem to be unrelated to the changes of this PR. Could you take a look?

zurfyx:
> @zurfyx Tests should be good now. The ones that have failed seem to be unrelated to the changes of this PR. Could you take a look?

You're right, main branch is failing :X

PR #5446, opened 2024-01-03
fix: wrong keyboard key in e2e tests

zignis:


zignis:
> Thank you, I believe the `alt` key is to move within words? In which case what we have now is accurate?

Sorry for the late reply, I can confirm that the `alt` (`option`) key on mac is used to move to the end of a paragraph (tested on macos sonoma 14.3).

zignis:
> @zignis are you still planning on updating this PR?

I'll try fixing the tests next week

potatowagon:

Issue #5449, opened 2024-01-04
Bug: logic issue in LexicalSelection.ts file that make collapseAtStart never run


nsnl-coder:
Lexical version: 0.12.5

## Steps To Reproduce

1. add collapseAtStart function inside LayoutItemNode
2. delete at start of item node and collapseAtStart will not run.

I found the issue in LexicalSelection.ts file.

```typescript

  deleteCharacter(isBackward: boolean): void {
    const wasCollapsed = this.isCollapsed();

     // delete logic here but i removed it

    if (
      isBackward &&
      !wasCollapsed && // it should be wasCollapsed here
      this.isCollapsed() &&
      this.anchor.type === 'element' &&
      this.anchor.offset === 0
    ) {
      const anchorNode = this.anchor.getNode();
      if (
        anchorNode.isEmpty() &&
        $isRootNode(anchorNode.getParent()) &&
        anchorNode.getIndexWithinParent() === 0
      ) {
        anchorNode.collapseAtStart(this);
      }
    }
  }
```

After the deletion logic, the code is checking whether it should run collapsedAtStart function.
The collapsedAtStart function should run when wasCollapsed is true, not when it is false.

Also, collapseAtStart should be working for node which is not top level node.
I believe  $isRootNode(anchorNode.getParent()) check can be removed so i can easily handle event when **nested element node** collapsed at start.

## The current behavior
collapseAtStart function is not running.
## The expected behavior
collapseAtStart function should run.

etrepum:
The use case of this is not very clear, and it sounds like it would not be backwards compatible. If you can more clearly explain what you're trying to do in a new issue, rather than focus on some specific implementation details, I think you would have a better chance of getting support for whatever it is you're trying to accomplish with this change.

PR #5450, opened 2024-01-04
bug: add a fix for collapseAtStart never run for nested element node

nsnl-coder:
Run collapseAtStart function of parent node  if:
- child element is only child.
- child element is empty (does not contain any node)

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1754,9 +1754,10 @@ export class RangeSelection extends INTERNAL_PointSelection {
       }
     }
     this.removeText();
+
     if (
       isBackward &&
-      !wasCollapsed &&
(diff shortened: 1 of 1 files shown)


ivailop7:
Can we have the return in curly braces and what issue is this actually fixing, do we have outstanding issues raised?

nsnl-coder:
Its related to this issue:
https://github.com/facebook/lexical/issues/5449

To summarize, the issue is the condition inside if check is wrong lead to collapseAtStart function of custom node never run.
Its is not a big issue since i can register an event to handle it but i expect it to work out of the box.

ivailop7 (approved):
please put the return in curly braces to keep the formatting consistency

nsnl-coder:
I changed the code to match the format.

I'm sorry because i accidentally close the pull request.

PR #5451, opened 2024-01-04
Replace font size dropdown with font size entry component

Shubhankerism:
This PR replaces the font size drop down with an font size entry component.  This will allow the user to add custom font sizes ranging from `8`to `72`. There are also increment and decrement buttons, which allow font size update relative to the current size.
For font sizes:
- 8-12: Step size is 1
- 12-20: Step size is 2
- 20-36: Step size is 4
- 36-72: Step size is 12

This PR also fixes the issue where the user can format the codeblock if the selection begins outside the codeblock. The issue is illustrated below:

code-block-isssue

### Font size entry component:
font-entry

### DIFF
--- a/packages/lexical-code/src/CodeHighlightNode.ts
+++ b/packages/lexical-code/src/CodeHighlightNode.ts
@@ -136,10 +136,6 @@ export class CodeHighlightNode extends TextNode {
     return self.__highlightType;
   }
 
-  /**
-   *
-   * @returns true if the codehighlight node can be formatted, false otherwise.
-   */
   isStylable(): boolean {
     return false;
   }
--- a/packages/lexical-code/src/CodeHighlightNode.ts
+++ b/packages/lexical-code/src/CodeHighlightNode.ts
@@ -136,7 +136,7 @@ export class CodeHighlightNode extends TextNode {
     return self.__highlightType;
   }
 
-  isStylable(): boolean {
+  canHaveFormat(): boolean {
     return false;
   }
 
--- a/packages/lexical-code/src/CodeHighlightNode.ts
+++ b/packages/lexical-code/src/CodeHighlightNode.ts
@@ -136,6 +136,14 @@ export class CodeHighlightNode extends TextNode {
     return self.__highlightType;
   }
 
+  /**
+   *
+   * @returns true if the codehighlight node can be formatted, false otherwise.
+   */
+  isStylable(): boolean {
+    return false;
+  }
+
   createDOM(config: EditorConfig): HTMLElement {
     const element = super.createDOM(config);
     const className = getHighlightThemeClass(
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -395,7 +395,7 @@ export function $patchStyleText(
 
   // This is the case where we only selected a single node
   if (selectedNodes.length === 1) {
-    if ($isTextNode(firstNode) && firstNode.isStylable()) {
+    if ($isTextNode(firstNode) && firstNode.canHaveFormat()) {
       startOffset =
         startType === 'element'
           ? 0
@@ -431,7 +431,7 @@ export function $patchStyleText(
     if (
       $isTextNode(firstNode) &&
       startOffset < firstNode.getTextContentSize() &&
-      firstNode.isStylable()
+      firstNode.canHaveFormat()
     ) {
       if (startOffset !== 0) {
         // the entire first node isn't selected, so split it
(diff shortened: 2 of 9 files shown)


packages/lexical-selection/src/lexical-node.ts
```diff
@@ -292,14 +293,19 @@ export function $addNodeStyle(node: TextNode): void {
 
 function $patchStyle(
   target: TextNode | RangeSelection,
-  patch: Record<string, string | null>,
+  patch: Record<
+    string,
+    string | null | ((currentStyleValue: string | null) => string)
```
Shubhankerism, line 298: The `$patchStyle` method can now accept a function that fetches the next value for the property, instead of just absolute values. This also separates the concern of calculating the value for a property away from the selection package.

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
+import {$isCodeHighlightNode} from '@lexical/code';
```
acywatson, line 8: This is sort of problematic, as it creates a dependency between these two packages. If someone were to implement code highlighting differently, they wouldn't be able to do this for their node. Typically, in these situations, we try to abstract the behavior out somehow so that's it's a heuristic on the node class, like "isUnstylable" or whatever the case is here.

ivailop7:
Once, you address the package inter-dependency:
Also, could we fix up the styling a bit:
- Add a separator before the minus button
- Either the hover over the buttons is misaligned or the icons inside are not centered
- Let's remove this 'shadow' inside border effect to a something lighter fitting the toolbar a bit better

packages/lexical-selection/src/lexical-node.ts
Shubhankerism: Hi, thanks for the review. I removed the package interdependency by exposing `isStylable` method from the node. 

I also tried implementing by adding a class "isUnstylable" to the node DOM, but this approach required more modification to `$patchStyleText` method for checking the classList for each node, like using `getElementByKey` for each node. Also, required changing several exisiting tests as the codeblock DOM would change.

The first approach seemed like a more cleaner way to get the task done with minimal code changes to existing `$patchStyleText` method, so went ahead with it. Please let me know if this works. :)

Shubhankerism:
> Once, you address the package inter-dependency: Also, could we fix up the styling a bit:
> 
> * Add a separator before the minus button
> * Either the hover over the buttons is misaligned or the icons inside are not centered
> * Let's remove this 'shadow' inside border effect to a something lighter fitting the toolbar a bit better

Thanks for the review, fixed it.

packages/lexical-code/src/CodeHighlightNode.ts
```diff
@@ -136,6 +136,14 @@ export class CodeHighlightNode extends TextNode {
     return self.__highlightType;
   }
 
+  /**
```
ivailop7, line 139: I don't think you need this comment in here

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -440,6 +440,14 @@ export class TextNode extends LexicalNode {
     return toggleTextFormatType(format, type, alignWithFormat);
   }
 
+  /**
+   *
+   * @returns true if the text node can be formatted, false otherwise.
```
ivailop7, line 445: `@returns true if the text node supports font styling`

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -440,6 +440,14 @@ export class TextNode extends LexicalNode {
     return toggleTextFormatType(format, type, alignWithFormat);
   }
 
+  /**
+   *
+   * @returns true if the text node supports font styling, false otherwise.
+   */
+  isStylable(): boolean {
+    return true;
+  }
```
zurfyx, line 449: What about `canHaveFormat` and `canHaveStyle` so that it follows the functions and properties we already have in the Node?

zurfyx (approved):
LGTM, thank you! Added a comment on the new API method that might need some small additional tweaks to the existing code

packages/lexical/src/nodes/LexicalTextNode.ts
ivailop7: canHaveFormat it is 🙂 Thanks, Gerard!

Shubhankerism: Thanks for the suggestion, will make the change! :)

zurfyx: Sorry, I meant, should we split the functionality into two? So `canHaveFormat` controls the format property whereas the `canHaveStyle` controls the style property

ivailop7 (approved):
LGTM

placeba:
Does `canHaveFormat` have to spread on FORMAT_TEXT_COMMAND? 
If a text node doesn't support formatting (`canHaveFormat` returns `false`) we cannot apply bold, italic and other types.

PR #5453, opened 2024-01-06
bug: collapseAtStart never run for custom element node

nsnl-coder:
Its related to this issue:
https://github.com/facebook/lexical/issues/5449

To summarize, the issue is the condition inside if check is wrong lead to collapseAtStart function of custom node never run.
Its is not a big issue since i can register an event to handle it but i expect it to work out of the box.

I'm sorry to create this pull request again since i accidentally closed the old pull request.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1754,9 +1754,10 @@ export class RangeSelection extends INTERNAL_PointSelection {
       }
     }
     this.removeText();
+
     if (
       isBackward &&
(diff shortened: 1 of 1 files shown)


ivailop7 (approved):
LGTM, but I need a second opinion here. @zurfyx @acywatson @fantactuka

zurfyx (changes requested):
Do you mind sharing more details about what you're trying to do here? This changes the original behavior and is not backward compatible. For reference, the original method only handles non-collapsed selection at the start of the document.

PR #5458, opened 2024-01-07
Fix inserting link open editor in edit mode

### LINKED ISSUE #5355: Bug: Inserting a new hyperlink using toolbar should open the editor edtiable mode
Lexical version: v0.12.5 (latest)

## Steps To Reproduce

1. Open lexical playground
2. Select some text
3. Insert a hyperlink using the toolbar button (using the Ctrl+K shortcut doesn't reproduce this issue)
4. The floating editor opens in read-only mode

This issue only exists when the hyperlink is inserted using the toolbar button (The shortcut ctrl+K works fine).

Link to code example:

## The current behavior

image

## The expected behavior

image

2wheeh:
fixes #5355

The issue seemed to be fixed in #5372 but it got reverted in #5456 because this broke some links, autolinks e2e tests.
It was because #5372 changed scenario of link editor a bit without modifying it in the tests accordingly.

So I added the e2e test modification in this PR.

And plus, #5372 missed the case of floating toolbar, which could be some other issue. (which needs `setIsLinkEditMode` to be passed to `InlineImageComponent` on current implementation so it might need to refactor `InlineImageComponent`)

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
@@ -165,6 +165,7 @@ test.describe('Auto Links', () => {
 
     await selectAll(page);
     await click(page, '.link');
+    await click(page, '.link-confirm');
 
     await assertHTML(
       page,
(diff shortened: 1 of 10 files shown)


2wheeh:
I don't really know much about different selection behaviors per browsers so please let me know if i'v done something wrong here : [fix selection assertion](https://github.com/facebook/lexical/pull/5458/commits/5613e8454ce4823612c3c0f566dc09445695fda2)

acywatson (approved):

PR #5460, opened 2024-01-07
Fix floating link editor on inline image caption

2wheeh:
`InlineImageComponent` makes two of floating link editor toolbar element and it's triggering some bugs.

before:

https://github.com/facebook/lexical/assets/40269597/f71da3f7-3d69-4eb8-99bb-81296ffc9587

after:

https://github.com/facebook/lexical/assets/40269597/2f5487bd-3a18-472f-8a36-6f6ded32478d


ivailop7 (approved):
LGTM

PR #5462, opened 2024-01-07
Fix typechecking issues in unit tests

thorn0:
~~I noticed the unit tests are excluded from typechecking. Tried including them and saw multiple issues. I fixed some of them and this might already be mergeable, but it's a bit too early to enable typechecking for the tests. 1153 errors in 34 files =)~~

Okay. Fixed only the tests in the `lexical` package and enabled typechecking for them. I'm not going to fix other packages in this PR.

Also noticed:
- Prettier needs to be updated to use newer TS features like `satisfies`
- ts-jest displays warnings about the TS version. Needs to be updated too or probably replaced with a faster transform like esbuild.


acywatson:
Wow, thanks! This is especially helpful as we're trying to lean more into unit tests instead of e2e.

acywatson (approved):

PR #5472, opened 2024-01-10
Fix multiline conversion to codeblock

Shubhankerism:
This PR fixes the issue with the lexical editor crashing when we do a multi line backward selection and try to convert it to a codeblock.

Before:

codeblock-backward-selection

After:

codeblock-backward-selection-fix

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1319,10 +1319,10 @@ export class RangeSelection extends INTERNAL_PointSelection {
       );
       return selection.insertNodes(nodes);
     }
-    let firstBlock = $getAncestor(this.anchor.getNode(), INTERNAL_$isBlock)!;
-    if ($isRangeSelection(this) && this.isBackward()) {
-      firstBlock = $getAncestor(this.focus.getNode(), INTERNAL_$isBlock)!;
-    }
+
+    const firstPoint = this.isBackward() ? this.focus : this.anchor;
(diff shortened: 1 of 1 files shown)


ivailop7 (approved):
LGTM

ivailop7:
@GermanJablo does this change look sensible?

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1319,7 +1319,10 @@ export class RangeSelection extends INTERNAL_PointSelection {
       );
       return selection.insertNodes(nodes);
     }
-    const firstBlock = $getAncestor(this.anchor.getNode(), INTERNAL_$isBlock)!;
+    let firstBlock = $getAncestor(this.anchor.getNode(), INTERNAL_$isBlock)!;
+    if ($isRangeSelection(this) && this.isBackward()) {
+      firstBlock = $getAncestor(this.focus.getNode(), INTERNAL_$isBlock)!;
+    }
```
GermanJablo, line 1325: I would prefer:
```ts
const firstPoint = this.isBackward() ? this.focus : this.anchor;
const firstBlock = $getAncestor(firstPoint.getNode(), INTERNAL_$isBlock)!;
```

ivailop7: Agreed

Shubhankerism: Thanks for the review, made the changes :)

PR #5480, opened 2024-01-12
Added new line options on import and export markdown

nmuren:
With these updates, users can choose to keep extra empty lines they see on editor updates.
Default params ensure existing usages are not affected.


zurfyx (commented):
Thanks for the PR, we discussed as a team as it's a fairly opinionated one. Given that you can do the pre/post-processing with an independent handler on your end and the lack of use cases to support this additional flag, we are inclined to preserve the function signature as is. Feel free to reopen the PR if you want to continue the discussion or submit further proposals.

PR #5481, opened 2024-01-12
Adding ability to mount LexicalMenus to arbitrary divs

ebads67:
When Lexical is embedded in other web apps mounting the menus on `document.body` can be problematic. This PR allows passing a parent element on which the menus are mounted.


zurfyx (approved):
LGTM, thank you!

ivailop7 (approved):

Issue #5344, opened 2023-12-06
Bug: Broken Command Type in the Debug View in the Playground


ivailop7:
At some point we've broken the TreeView, showing UNKNOWN for all dispatched commands.
Will look into fixing this.

![image](https://github.com/facebook/lexical/assets/7893468/4b0740ee-ebb0-4f26-9d01-e31528dfcec1)

yf-yang:
I think it is a desired behavior. When self hosting the playground, the type could be correctly shown.

https://github.com/facebook/lexical/blob/2c825792956f219f97da88c455d53855ddb1d9f5/packages/lexical/src/LexicalCommands.ts#L13-L15

ivailop7:
@acywatson @zurfyx @fantactuka do you know why this was made to work only in DEV ? This seems very strange to me.

ivailop7:
Not sure the increased bundle size is worth it, even though it would be nice to see the events in the playground for easier reproduction of issues. Closing.

PR #5482, opened 2024-01-14
Return event name in Production

ivailop7:
Fix for #5344

Confirmed with @fantactuka and @zurfyx this is a good fix, since it's unlikely to remove any minification benefits.

@thegreatercurve the original conditional was introduced by you in https://github.com/facebook/lexical/pull/2942

Are you fine with this change?


ivailop7:
Still unsure if the change is worth it, to see the event names in the playground to bug reproductions, at the cost of increasing the bundle size.

ivailop7:
Maybe it's better to keep it as it is for now, will close the issue I raised originally

PR #5484, opened 2024-01-14
Fix columns layout update command

ivailop7:
The current update command for the columns layout is broken when going to a lower column count. The current method instead of deleting, actually appends and gets the node in an invalid state. This fixes the update operation, so when going from 3 columns to 2, it will delete the last column and preserve the content as is in the first 2.


zurfyx (approved):
No objections but note that this changes the behavior - instead of transferring the children we now delete them

ivailop7:
> No objections but note that this changes the behavior - instead of transferring the children we now delete them

That's fine, this is the deletion method, we don't want to keep them.

PR #5485, opened 2024-01-14
Fix escape down for collapsible section when collapsed

ivailop7:
At the moment there's a logic that when the collapsible section is the very last node, pressing ArrowDown will append a paragraph below to escape the section. This only worked when the section is expanded, since the logic only applied on exiting the content node of the container, but when the section was collapsed, you exit from the title node, so the logic didn't kick in. 
Extending the logic to check for either fixes the issue.

This was only an issue for the "Escape down", since "Escape up" is always going from the title, irrespective if expanded or collapsed.


zurfyx (approved):
LGTM, are we running this iff it's collapsed?

ivailop7:
right now, it runs in either case, collapsed or expanded. Didn't want to tie it to the node state, when it has a clean solution.

PR #5507, opened 2024-01-18
[lexical] Add tests for HTMLConfig

2wheeh:
**EDIT**:
This adds tests for HTMLConfig


etrepum (commented):
I like that this adds tests but it does let you write incorrect code that passes the type checker because really what we would want to say is that the key and value of the map are related (something like `forall T extends LexicalNode. [K: Klass<T>, V: (editor: LexicalEditor, target: T) => DOMExportOutput]` but I don't think it would be possible to express something like that directly with Map's type

2wheeh:
> I don't think it would be possible to express something like that directly with Map's type

@etrepum You are right. I reverted that and changed this PR to focus on adding unit tests for HTMLConfig.

etrepum (approved):
Thank you!

PR #5517, opened 2024-01-19
Allow an empty paragraph child on converting from markdown

### LINKED ISSUE #5516: Bug: an empty table cell breaks the editor on converting from markdown string
Lexical version: the latest (v0.12.6)

## Steps To Reproduce

1.  go to the playground
2. make table in markdown with empty cells:
    ```md
    | foo | |
    ```
4. convert it
5. then empty table cell node doesn't have paragraph node child:
    ```
    root
   └ (141) table  
        └ (142) tablerow  
          ├ (135) tablecell  
          | └ (137) paragraph  
          |   └ (136) text  "foo"
          └ (138) tablecell  <- this one should keep the paragraph child.

    ```
7. once you try to enter new line by pressing `ENTER` in the empty cell, it breaks the app even without any error logging and response.

Link to code example: can reproduce in the playground

## The current behavior

https://github.com/facebook/lexical/assets/40269597/4866802e-1499-4303-89c1-107401378cc1

## The expected behavior
1. keep the paragraph node as a child even when the cell is empty.
2. At least, it should escape the context and throw a proper error.

2wheeh:
fixes #5516

after:

https://github.com/facebook/lexical/assets/40269597/0e1e970f-f91f-499a-a33c-9ebb6a81979b

### DIFF
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -77,21 +77,13 @@ export function createMarkdownImport(
       );
     }
 
-    const wasRootNotEmpty = root.isEmpty() === false;
-
     // Removing empty paragraphs as md does not really
     // allow empty lines and uses them as delimiter
     const children = root.getChildren();
     for (const child of children) {
-      if (isEmptyParagraph(child)) {
+      if (isEmptyParagraph(child) && root.getChildrenSize() > 1) {
         child.remove();
       }
-    }
-
-    // When removing empty paragraphs makes the nodes empty that
-    // can not be empty, an empty paragraph should be appended.
-    if (wasRootNotEmpty && root.isEmpty() && !root.canBeEmpty()) {
-      root.append($createParagraphNode());
     }
 
     if ($getSelection() !== null) {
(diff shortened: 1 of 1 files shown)


packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -77,15 +77,23 @@ export function createMarkdownImport(
...
     }
 
+    const wasRootNotEmpty = root.isEmpty() === false;
+
     // Removing empty paragraphs as md does not really
-    // allow empty lines and uses them as dilimiter
+    // allow empty lines and uses them as delimiter
     const children = root.getChildren();
     for (const child of children) {
       if (isEmptyParagraph(child)) {
```
zurfyx, line 86:
```suggestion
      if (isEmptyParagraph(child) && root.getChildrenSize() > 1) {
```

zurfyx (approved):
Thanks, the proposal is reasonable, should we just compare whether clearing the last one leads to an empty state? Also, root `canBeEmpty` is just a product of inheriting ElementNode, the only case where this is valid is when starting the editor and the expectation is that soon after it will be populated with the bootstrapping content (the case where this is the most obvious is in collab)

2wheeh:
@zurfyx Thank you for kind explanation. I misunderstood some methods. I applied your suggestion: 036b878
Btw, while debugging this, I found some other cases that table node looses table row and cell node. I will open a new issue about them.

zurfyx (approved):
Thank you for iterating on this!

PR #5521, opened 2024-01-19
Combined selection format to ignore empty text nodes

zurfyx:
Extending #5126 to exclude all empty text nodes as well as the resulting from the user's selection (which may not span the entire TextNode).

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -304,8 +304,6 @@ function onSelectionChange(
     if ($isRangeSelection(selection)) {
       const anchor = selection.anchor;
       const anchorNode = anchor.getNode();
-      const focus = selection.focus;
-      const focusNode = focus.getNode();
 
       if (selection.isCollapsed()) {
         // Badly interpreted range selection when collapsed - #1482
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -351,30 +349,34 @@ function onSelectionChange(
...
-          selection.getTextContent().endsWith('\n') &&
-          (anchor.offset === 0 || focus.offset === 0) &&
-          $isTextNode(nodes[nodes.length - 1])
-        ) {
-          nodes.pop();
-        }
-
         const nodesLength = nodes.length;
+        const isBackward = selection.isBackward();
+        const startOffset = isBackward ? focusOffset : anchorOffset;
```
ivailop7, line 358: I think it would be better to do an old-fashioned if/else here, since it will end up with 1 comparison and 4 assignments vs 4 comparisons and 4 assignments, or does this have some minification benefits?

ivailop7 (approved):

packages/lexical/src/LexicalEvents.ts
ivailop7: Nevermind, you'll need to do it with closure outside the if, so you'd end up with 1 comparison but 4 declarations and 4 reassignments. Doubt it will make any difference in bundle size or minification

PR #5528, opened 2024-01-22
Move Grid to @lexical/table

zurfyx:
We don't want Grid modules inside the Core either, move these ones to the @lexical/table package


ivailop7 (approved):
🚀

PR #5533, opened 2024-01-23
Rename LexicalTableSelection to LexicalTableObserver

zurfyx:
Leveraging the fact that we're doing a minor release to fix the naming mess we have here.

LexicalSelection is not a selection, it's merely a sort of helper. When you look at the origin of it, you can see it's an observer, it leverages observability to avoid recomputing everything from scratch.

This PR does some renaming:
1. LexicalTableSelection -> LexicalTableObserver
2. Cell -> LexicalDOMCell (to avoid confusion with the actual CellNode)
3. Cells -> LexicalDOMRows
4. Grid -> LexicalDOMTable
5. LexicalTableSelection.grid -> LexicalTableSelection.table
6. Rremoved `grid` in the helper functions that had it, the term `grid` should not longer be used


ivailop7 (approved):

ivailop7:
You have integration tests failing, because of incorrect flow definition.

PR #5534, opened 2024-01-23
Rename GridSelection to TableSelection

zurfyx:
Having renamed the badly named TableSelection, we can now rename GridSelection to TableSelection and this concludes the existance of GridSelection.


ivailop7 (approved):
LGTM

PR #5540, opened 2024-01-24
v0.13.0

zurfyx:
# v0.13.0 (2024-01-24)

This version wraps up a major tables refactor, and includes some more bugfixes!

## TableSelection

This version gets rid of `GridSelection` and `Grid{Cell/Row/Table}Node`, you should now (exclusively) use the Table equivalents.

Upgrade instructions (rename these in your codebase):
- All table files now live in `@lexical/table` as opposed to a mix between `lexical` and `@lexical/table`.
- `GridSelection` -> `TableSelection`
- `$isGridSelection` -> `$isTableSelection`
- `LexicalTableSelection` -> `LexicalTableObserver`
- `Cell` (type) -> `TableDOMCell`
- `Cells` (type) -> `TableDOMRows`
- `Grid` (type) -> `TableDOMTable`
- `LexicalTableSelection.grid` (the property) -> `LexicalTableSelection.table`
- (helper function that contained `grid` in their name no longer do)

**Why?**

When we started GridSelection, we wanted GridSelection to cater a variety of use cases rather than remain specific to the Tables implementation. Fast forward, 1+ years we still haven't found any additional use case that fits GridSelection (internally or within the community). Hence, we think this is an opportunity to trim down the Core bundle size by making GridSelection specific to Tables.

**Custom selection**

But it's not just a cleanup! We leveraged this opportunity to revise the basics of selection to enable anyone create their own custom selection (just like we did for tables). You can now build on top of BaseSelection and construct a selection of any type (i.e. a px based selection).

(...) append other stuff at the bottom


ivailop7 (approved):
🎉

PR #5541, opened 2024-01-24
Add type literal for serialized nodes types

strdr4605:
There are use cases when you don't have the editor and operate only on the serialized state. I think it makes sense to have the ts type of `type` prop be the literal of the node.

```ts
// example.ts
export const extractCommentEditorData = (
  nodes: (SerializedLineBreakNode | SerializedAutoLinkNode | SerializedTextNode)[],
): CommentEditorData => {
  const editorData = nodes.map(element => {
    if (element.type === 'linebreak') {
      return {
        type: 'text',
        text: '\n'
      };
    }
    if (element.type === 'autolink') {
      const node = element as SerializedAutoLinkNode;
      return {
        type: 'text',
        text: node.url ?? ''
      };
    }
    return {
      type: 'text',
      text: (element as SerializedTextNode).text
    };
  }) as CommentEditorData;

  return editorData;
};
```

### DIFF
--- a/packages/lexical-code/flow/LexicalCode.js.flow
+++ b/packages/lexical-code/flow/LexicalCode.js.flow
@@ -109,6 +109,7 @@ declare export function normalizeCodeLang(lang: string): string;
 
 export type SerializedCodeNode = {
   ...SerializedElementNode,
+  type: 'code',
   language: string | null | void,
   ...
 };
--- a/packages/lexical-code/src/CodeHighlightNode.ts
+++ b/packages/lexical-code/src/CodeHighlightNode.ts
@@ -50,6 +50,7 @@ export const DEFAULT_CODE_LANGUAGE = 'javascript';
 
 type SerializedCodeHighlightNode = Spread<
   {
+    type: 'code-highlight';
     highlightType: string | null | undefined;
   },
   SerializedTextNode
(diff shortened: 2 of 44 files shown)


zurfyx:
Thanks, isn't type already included in the `.toJSON()` serialization? I.e. this test below

https://github.com/facebookexternal/Outline//blob/8a4ac97b761f6bb8de28aceb731cee97612cc465/packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx#L872

strdr4605:
> Thanks, isn't type already included in the `.toJSON()` serialization? I.e. this test below
> 
> 
> 
> https://github.com/facebookexternal/Outline//blob/8a4ac97b761f6bb8de28aceb731cee97612cc465/packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx#L872


The type is of type string always. 

these changes make the type to be the literal like "text" or "mark".

strdr4605:
> Thanks, isn't type already included in the `.toJSON()` serialization? I.e. this test below
> 
> 
> 
> https://github.com/facebookexternal/Outline//blob/8a4ac97b761f6bb8de28aceb731cee97612cc465/packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx#L872


The type is of type string always. 

these changes make the type to be the literal like "text" or "mark".

zurfyx:
Makes sense, iirc we used to do this and we dropped it because of some inheritance trouble; TextNode -> `text` can't be extended into MentionNode -> `mention` but TextNode -> string can be extended into MentionNode -> string

Issue #1400, opened 2022-03-02
window.Promise breaks in a Node environment


zurfyx:
> in Lexical.dev.js in the top-level scope, NativePromise is defined as window.Promise which makes things throw ("window is not defined") in a node environment

PR #5547, opened 2024-01-25
Check global.window on DOM availability check

2wheeh:
There are some use cases `@lexical/headless` and `$generateHtmlFromNodes` in the server side and `nextjs` is probably the most common framework to implement this.

I found some issue and discussions related to this:
- #3616
- #3097 
- #5266
- #2880
- https://github.com/facebook/lexical/discussions/3765#discussioncomment-5815254

I've met same situation as well, which is that `@lexical/headless` is not compatible with `nextjs` servserside even with `jsdom` (or `happy-dom`) since #3102 (v.0.6.0)

I don't know why but anyway in `nextjs`, `typeof window` is still `undefined` after assigning `global.window`.
So it fails type assertion fixed in #3102.

Can reproduce here:
- https://stackblitz.com/~/github.com/2wheeh/lexical-nextjs-ssr
-> you can see it works with `lexical@0.5.0` and `nextjs` but in the console you also can see `typeof window` is still `undefined` while `global.window` is changed as we assign.

So could we check `typeof global.window` with `typeof window`?


zurfyx (approved):
Thanks, I remember we had issues with `window` in the past https://github.com/facebook/lexical/issues/1400 ; even if this works fine I believe this PR might not be complete (iframe support and other occurences, we introduced this window utility for this https://github.com/facebook/lexical/blob/a5b20c03fa494d6fa0d22f6dd8e5bd2db7b342ea/packages/lexical/src/LexicalUtils.ts#L1296-L1302). In any case, I'm happy to merge as-is and we can iterate on this later

2wheeh:
@zurfyx thank you for detailed comment ! I agree that this PR is not complete for sure 🥲 even after I see how many `CAN_USE_DOM` is called everywhere and there is no use of `global.window`. I just hope this can help some guys struggling with `nextjs`

PR #5557, opened 2024-01-27
Add `@lexical/html` to docs

2wheeh:
Somehow `@lexical/html` is not on the docs. So this PR adds `@lexical/html` to docs.
and I tried to supplement `Exporting` in headless mode as much as `Importing` covers but it needs some checks:
1. `@ts-expect-error` for assigning `window` seems inevitable at the moment:
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/51276
https://github.com/capricorn86/happy-dom/issues/1227
but I'm not sure this should be included in instruction here.

2. I've seen that `DocumentFrament` used to be needed with `nextjs` which seems not anymore with the latest lexical (https://stackblitz.com/~/github.com/2wheeh/lexical-nextjs-ssr) and apparently `navigator` is needed sometimes(#3616). But I couldn't address the reason and when :( Would this comment be enough for now?

3. Is `sidebar_position` effective? there are some duplicative numbers already. Should this be same as order in the sidebar?

While at this, I also fixed some typos and added `@lexical/mark`, `@lexical/overflow` (they are practically empty though).


zurfyx (approved):
Thank you! I think 1 and 2 are reasonable, 3 maybe @acywatson has thoughts

acywatson:
> Is sidebar_position effective? there are some duplicative numbers already. Should this be same as order in the sidebar?

Effective in what sense? I think it's functional, though perhaps not the best design I've ever seen for managing the order of items in a sidebar.

At the end of the day, I don't think it matters too much what order these are in, so what you have is probably fine.

acywatson (approved):

2wheeh:
> Effective in what sense? I think it's functional, though perhaps not the best design I've ever seen for managing the order of items in a sidebar.

I was just curious whether it determines the actual order in rendering and accordingly what i did would be a problem. Sorry for my bad English. Thank you for the explanation though!

PR #5558, opened 2024-01-28
Fix: add class to elements exported from CodeNode

2wheeh:
On exporting, `CodeHighlightNode` and other many `ElementNode` have same class as on created. However, `CodeNode` skips to add class to it, so on the page that uses exported HTML, the CSS for theming `code` should be duplicated for `pre`. So I'm utilizing `htmlConfig` as a workaround for now. 
before:
image

I'm aware of `exportDOM` and `createDOM` of `CodeNode` return different semantic but still they could have same class.

### DIFF
--- a/packages/lexical-code/src/CodeNode.ts
+++ b/packages/lexical-code/src/CodeNode.ts
@@ -12,6 +12,7 @@ import type {
   DOMConversionOutput,
   DOMExportOutput,
   EditorConfig,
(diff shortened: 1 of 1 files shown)


zurfyx (approved):
LGTM! In the future, we'll need to think of a better way to expand types that doesn't involve mixing the core with ecosystem

Issue #5560, opened 2024-01-28
Bug: Right Align in Toolbar for RTL direction shows Left Align Icon


fahad12398:
The icon for Right Align, in case of RTL direction, should be Right Align icon in Toolbar in Lexical Playground. Currently, it is Left Align icon. The right align will always keep your text to the right irrespective of text direction (LTR, RTL).

![image](https://github.com/facebook/lexical/assets/48366592/6f531516-eac2-4633-9e27-b0ba5ba371ae)

In `packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx`, `iconRTL` for `right` ELEMENT_FORMAT_OPTIONS should be `right-align`.

Lexical version:
v0.13.1

## Steps To Reproduce

1. Go to Lexical Playground.
2. Select Right Align.
3. type some text.
4. Change Direction to RTL.
5. The icon for Right Align should stay as it was before changing direction to RTL. 

Link to code example:
https://playground.lexical.dev/

## The current behavior
The icon for Right Align, in case of RTL direction, is Left Align icon.

## The expected behavior
The icon for Right Align, in case of RTL direction, should be Right Align icon.

CRIMSON-CORP:
Im sorry pls how were you able to change the text Direction on lexical? setDirection() does not do anything for me

fahad12398:
Sorry for late reply, i changed the direction in Lexical Playground using context menu. The steps are below:

1. right click on the mouse.
2. go to writing direction menu.
3. select Right to Left.

PR #5561, opened 2024-01-28
Bug: Right Align in Toolbar for RTL direction shows Left Align Icon

fahad12398:
This pull request is related to #5560.

In `packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx`, `iconRTL` for `right` ELEMENT_FORMAT_OPTIONS should be `right-align`.

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
@@ -184,7 +184,7 @@ const ELEMENT_FORMAT_OPTIONS: {
   },
   right: {
     icon: 'right-align',
-    iconRTL: 'left-align',
+    iconRTL: 'right-align',
     name: 'Right Align',
   },
   start: {


zurfyx (commented):
Thanks, it visually it looks correct to me, is it just the classname naming the one that's wrong?

fahad12398:
> Thanks, it visually it looks correct to me, is it just the classname naming the one that's wrong?

Yup, The classname naming is wrong.
Also, i have tried changing the direction but the icons don't change. Although, the alignment changes in case of start and end, remains same in case of right and left after changing direction. Which is correct behavior. But, the icons of start and end doesn't change.
I was thinking of raising the issue on that later. After further investigating and finding the issue.

zurfyx:
Thank you, in this case, I believe the correct approach is to rename the underlying icons.

zurfyx (approved):
Completely misread the code, sorry :X

fahad12398:
> Completely misread the code, sorry :X

Thank you for your comments. No worries at all! We all have those moments where we misread code. I'm grateful for your proactive approach in engaging and clarifying any misunderstandings. I apologize for not responding sooner; things have been quite busy on my end.

PR #5568, opened 2024-01-30
Update react documentation

pavel-shermayster-sage:
Added missing `return null` and custom plugin rename.
The documentation aligns with the tutorial https://www.youtube.com/watch?v=qIqxvk2qcmo&t=1s.v=qIqxvk2qcmo&t=1s


zurfyx (approved):
The `null` return is optional. In React no return means `undefined` and you can still render `undefined` components. Likewise, for components that return a boolean.

In any case, LGTM for the sake of consistency

PR #5571, opened 2024-01-31
Added Link Type check and simplified floating link plugin

HarrySIV:
Extension of #5551 
@zurfyx is this closer to what you were looking for? I certainly made it simple for the product side, but I'm not sure if there's a more optimal solution, or if this is overkill. 
Also I noticed $getAncestor and $findMatchingParent are similar but I'm not sure if there's a distinct difference I'm not seeing.

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -29,6 +29,7 @@ import {
   $getSelection,
   $isElementNode,
   $isLeafNode,
+  $isLineBreakNode,
   $isRangeSelection,
   createCommand,
   ElementNode,
@@ -542,14 +543,13 @@ function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
 }
 
 /**
- * Checks selection to ensure only a single LinkNode or AutoLinkNode is being selected,
- * selecting any Node that doesn't have a some Link as a parent will return false.
+ * Checks selection to ensure only an AutoLinkNode or LinkNode is being selected and returns it.
  * @param selection - The current range selection.
- * @returns - true if selection only contains a single LinkNode or AutoLinkNode, false otherwise.
+ * @returns - the (Auto)LinkNode or null.
  */
-export function $onlyIncludesAParentLink(
+export function $getSelectedLinkNode(
   selection: RangeSelection,
-): boolean | null {
+): LinkNode | null {
   const anchor = selection.anchor;
   const focus = selection.focus;
   const isBefore = anchor.isBefore(focus);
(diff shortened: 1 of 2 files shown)


packages/lexical-link/src/index.ts
```diff
@@ -535,3 +540,69 @@ function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
...
+  const firstPoint = isBefore ? anchor : focus;
+  const lastPoint = isBefore ? focus : anchor;
+  const firstNode = firstPoint.getNode();
+  const lastNode = lastPoint.getNode();
+  const firstLinkNode = getParentLink(firstNode);
+  const lastLinkNode = getParentLink(lastNode);
+  if (
+    !firstLinkNode ||
+    !lastLinkNode ||
+    (firstLinkNode && !firstLinkNode.is(lastLinkNode))
```
zurfyx, line 565: Doesn't this lead to the same problem? The first node can be a link while the last node points to a LineBreakNode

packages/lexical-link/src/index.ts
```diff
@@ -535,3 +540,69 @@ function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
...
+        } else {
+          continue;
+        }
+      } else {
+        matchedParent = node;
+        continue;
+      }
+    }
+    const parent = getParentLink(node);
+    if (!parent) {
```
zurfyx, line 588:
```suggestion
    if (parent === null) {
```

packages/lexical-link/src/index.ts
```diff
@@ -535,3 +540,69 @@ function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
...
+      return false;
+    } else if (!matchedParent) {
```
zurfyx, line 590:
```suggestion
    } else if (matchedParent === null) {
```

packages/lexical-link/src/index.ts
```diff
@@ -535,3 +540,69 @@ function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
...
+
+/**
+ * Checks selection to ensure only a single LinkNode or AutoLinkNode is being selected,
+ * selecting any Node that doesn't have a some Link as a parent will return false.
+ * @param selection - The current range selection.
+ * @returns - true if selection only contains a single LinkNode or AutoLinkNode, false otherwise.
+ */
+export function $onlyIncludesAParentLink(
+  selection: RangeSelection,
+): boolean | null {
```
zurfyx, line 552:
```suggestion
export function $getSelectedLinkNode(
  selection: RangeSelection,
): null | LinkNode {
```

packages/lexical-link/src/index.ts
```diff
@@ -535,3 +540,69 @@ function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
...
+    const parent = getParentLink(node);
+    if (!parent) {
+      return false;
+    } else if (!matchedParent) {
+      matchedParent = parent;
+    } else if (!parent.is(matchedParent)) {
+      return false;
+    }
+  }
+  return true;
```
zurfyx, line 596: As suggested above, let's change this for the actual matched LinkNode

zurfyx (commented):
Added some minor comments but for the most part LGTM, thank you!

HarrySIV:
So I ended up having to delve further into `getNodesBetween` to get around Lexical being incapable of selecting LineBreakNodes. The big issue I couldn't solve is when you select a LineBreakNode on the beginning or end of selection, lexical selects the entire paragraph node anyway. 

So there's a weird case where you have a linebreak and a link, and if you select just the linebreak only, lexical thinks you're selecting the whole paragraph and, therefore, the link. I'm not sure how to work around this.

etrepum:
The selection issues should be solved by #6883, but this PR is quite stale and has conflicts. Please re-open if you intend to update it to work with the current version.

Issue #3807, opened 2023-02-02
Bug: Text alignment doesn't work properly in lists


nickschinestzki:
Lexical version: 0.7.6

## Steps To Reproduce

1. Create any list with 3 items 
2. Set each row to a different alignment (left, center and right)

Link to code example:

## The current behavior
Both in Safari and Chrome browsers the list item markers don't follow the content, staying always to the left. When tested in Mozilla Firefox, the markers follow the content. 

> Another bug is that when parsing the JSON to HTML, the alignment of the list items is ignored and all items are aligned to the left. 

Here's some screenshots of the current behavior in this 3 browsers (taken from the [Playground](https://playground.lexical.dev)).

### Safari
<img width="628" alt="Captura de Tela 2023-02-01 às 15 49 15" src="https://user-images.githubusercontent.com/33634787/216202590-2d32de99-6b4c-46da-afd3-e8cdf1792771.png">

### Chrome
<img width="537" alt="Captura de Tela 2023-02-01 às 15 49 09" src="https://user-images.githubusercontent.com/33634787/216202667-cb677a92-0901-4767-8fa5-5947171f6075.png">

### Firefox
<img width="571" alt="Captura de Tela 2023-02-01 às 15 49 57" src="https://user-images.githubusercontent.com/33634787/216202692-e1840a16-a33e-4ff5-8973-ea3c0546c3d5.png">

## The expected behavior
I want every browser to follow the current behavior found in the last example, the one from **Firefox**. I'm searching for some way, some style adjustment that could provide me that. Also searching for the solution of the other bug, that JSON -> HTML is ignoring the applied alignment to list items.

PR #5578, opened 2024-02-03
Preserve bullet item indent on newline

ivailop7:
Before:

https://github.com/facebook/lexical/assets/7893468/6f3a3463-dbef-4109-9ccf-ddf84f51ff9b

After:

https://github.com/facebook/lexical/assets/7893468/c1b852bf-5877-447e-b73a-9c63c6e8fb74

This is a fix for https://github.com/facebook/lexical/issues/3999 technically.

Applying this on 'ul' level, I verified it behaves correctly in Safari as well.

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -282,36 +282,36 @@
 .PlaygroundEditorTheme__ol1 {
   padding: 0;
   margin: 0;
-  list-style-position: inside;
+  list-style-position: outside;
 }
 .PlaygroundEditorTheme__ol2 {
   padding: 0;
   margin: 0;
   list-style-type: upper-alpha;
-  list-style-position: inside;
+  list-style-position: outside;
 }
 .PlaygroundEditorTheme__ol3 {
   padding: 0;
   margin: 0;
   list-style-type: lower-alpha;
-  list-style-position: inside;
+  list-style-position: outside;
 }
 .PlaygroundEditorTheme__ol4 {
   padding: 0;
   margin: 0;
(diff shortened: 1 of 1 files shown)


ivailop7:
@GermanJablo is this one OK with you?

GermanJablo:
check this one: https://github.com/facebook/lexical/issues/3999

ivailop7:
> check this one: #3999

I have, I've mentioned it in the description. 'inside' was already applied on list level, so changing to 'outside' at list-level fixes both indent for newline and the cursor position starts to work correctly in Safari on both Mac and iOS

zurfyx (approved):
Based on the description, LGTM! 

@GermanJablo, your research seems to conclude that `inside` should not be required, but what about `outside`?

GermanJablo:
Sorry, I hadn't seen the mention of the issue.

What I'm saying there is that:
- on the one hand, `list-style-position` should be in `outside`, or directly omitted since [outside is the default value](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position#formal_definition).
- on the other hand, apply the `format` property that is responsible for the alignment of the text on the ListNode (`ol` and `ul`), and not on the ListItem (`li`), to avoid the problem described in [#2334](https://github.com/facebook/lexical/pull/2334) and [#3807](https://github.com/facebook/lexical/issues/3807).

acywatson:
Browsers are known to behave differently with these types of things - have we checked Firefox, Safari, and Edge, at least?

ivailop7:
> Browsers are known to behave differently with these types of things - have we checked Firefox, Safari, and Edge, at least?

I have tested in Chrome, Safari and Firefox on Mac and it works consistently fine.

Issue #5581, opened 2024-02-03
Bug: Adding a horizontal line under a list and selecting leads to a crash


ivailop7:
Lexical 0.13.1

Adding a horizontal line under a list and selecting, crashes the editor. This happens only when the checklist is the very first node in the editor, if there is a paragraph before the checklist and select from it, and the checklist is a middle node, selection works correctly.

https://github.com/facebook/lexical/assets/7893468/c42aefe5-9842-44db-b72c-9787a7a45766

PR #5583, opened 2024-02-03
Fix list selection when first node followed by an element node

ivailop7:
Before:

https://github.com/facebook/lexical/assets/7893468/c42aefe5-9842-44db-b72c-9787a7a45766

After:

https://github.com/facebook/lexical/assets/7893468/8e295d84-ebf0-4fb4-952c-7377cb37c14c

Fixes: https://github.com/facebook/lexical/issues/5581


zurfyx (dismissed):
Good one, any chance we can have a test on this before merging?

ivailop7:
> Good one, any chance we can have a test on this before merging? 

Of course

zurfyx (approved):

PR #5584, opened 2024-02-03
IME Input via CDP with latest Playwright

ivailop7:
We are 13 versions or so behind on Playwright, because we are using a custom fork due to IME Input in tests that only supported Chromium and WebKit. Since the PR from that fork didn't get merged in Playwright and we need to move forward with Playwright versions and latest features and stability. The current workaround for IME events is sending them via CDP, so I migrated those tests using that method. Also deleted the custom fork from the repo.


zurfyx (approved):
🚀

PR #5586, opened 2024-02-05
fix: update docs example

jpulec:
Fix this example to be proper JSX and use the correct variable name.


zurfyx (approved):
LGTM, can merge as soon as the CLA is signed!

Issue #5589, opened 2024-02-06
Loading a lot of list items in the initial value when you have any node transform registered causes UI to freeze


amanharwara:
Not sure if this can necessarily be considered a bug, but I think this does need to be looked into.

If you have a lot of list items in your initial value and have any node transform registered, even the ones not registered for ListItemNode, it will block for a long time. The more transforms you have, the worse it is.

Here is a profile of the playground running locally with a large number of list items in the initial value:

![image](https://github.com/facebook/lexical/assets/8348101/011a30da-0641-4d19-a7f2-eacf7cc56a70)

Removing most of the transforms brings the time taken much lower:

![image](https://github.com/facebook/lexical/assets/8348101/cad1d9a5-63da-48a3-a430-c7630f15c394)

From what I can tell, it has something to do with the [`transform()`](https://github.com/facebook/lexical/blob/main/packages/lexical-list/src/LexicalListItemNode.ts#L106) method on `ListItemNode` (initially added here https://github.com/facebook/lexical/pull/4030)

amanharwara:
Fixed by #5590

PR #5590, opened 2024-02-06
fix: Improve performance of updateChildrenListItemValue

etrepum:
The algorithm used to update `ListItemNode` values was quadratic when called in a loop as it interrogated all previous siblings. Calculating all expected values from the `ListNode` ahead of time is simpler and much more performant.

Since a transform was already in place for `ListItemNode` it seemed reasonable to implement one for `ListNode` and consolidate all of the `updateChildrenListItemValue` usage there to avoid repeating the same list traversals for each item. This also lead the way to removing code from other places that attempted to renumber lists at each mutation where that might be necessary.

RE #5589

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -39,12 +39,7 @@ import {
 import invariant from 'shared/invariant';
 
 import {$createListNode, $isListNode} from './';
-import {
-  $handleIndent,
-  $handleOutdent,
-  mergeLists,
-  updateChildrenListItemValue,
-} from './formatList';
+import {$handleIndent, $handleOutdent, mergeLists} from './formatList';
 import {isNestedListNode} from './utils';
 
 export type SerializedListItemNode = Spread<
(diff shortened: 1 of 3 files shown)


etrepum:
The test-e2e-ci-webkit test failures are the same ones I can reproduce locally on main, they do not appear to be regressions:

```
    7 failed
    [webkit] › packages/lexical-playground/__tests__/e2e/HorizontalRule.spec.mjs:30:3 › HorizontalRule › Can create a horizontal rule and move selection around it 
    [webkit] › packages/lexical-playground/__tests__/e2e/Keywords.spec.mjs:231:3 › Keywords › Can type "congrats Bob!" where " Bob!" is bold 
    [webkit] › packages/lexical-playground/__tests__/e2e/Links.spec.mjs:1605:3 › Links › Can convert part of a text node into a link with forwards selection 
    [webkit] › packages/lexical-playground/__tests__/e2e/Links.spec.mjs:1721:3 › Links › Can convert part of a text node into a link with backwards selection 
    [webkit] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:1026:3 › Markdown › can adjust selection after text match transformer 
    [webkit] › packages/lexical-playground/__tests__/e2e/Tab.spec.mjs:20:3 › Tab › can tab + IME ───
    [webkit] › packages/lexical-playground/__tests__/e2e/TextFormatting.spec.mjs:733:3 › TextFormatting › Can select multiple text parts and format them with shortcuts 
  29 skipped
  379 passed (2.3m)
```

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -105,10 +100,12 @@ export class ListItemNode extends ElementNode {
 
   static transform(): (node: LexicalNode) => void {
```
etrepum, line 101: Not sure if this is still worth doing? Just about any operation that would put checked in an inconsistent state (other than directly calling setChecked on an existing connected ListItemNode) would naturally trigger the ListNode transform, which is why I gave it a fast bail-out if __checked is null

zurfyx (approved):
Thank you, great work! (same feedback shared by other members of team) In the future, we might incorporate a way to encapsulate the plugin/nodes logic but what you did is probably the best it can get right now

PR #5602, opened 2024-02-09
Fixed position for lexical menu

ebads67:
LexicalMenu position is calculated using coordinates relative to the view port. Then its container div should have fixed position property.


ivailop7 (approved):
LGTM

PR #5614, opened 2024-02-15
Revert "Fixed position for lexical menu (#5602)"

zurfyx:
Reverting as the menu does not work correctly when the page is scrolled. If you scroll down the entire viewport, the menu doesn't even show.

https://github.com/facebook/lexical/assets/193447/018b9e18-5b0f-4b8b-9647-fc12e7850472

@ebads67 can you share more details on your use case and we can follow up in a separate PR?

This reverts commit 0ebb41307624af16cb2cdd48e9804ccb96edd32c.

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -530,7 +530,7 @@ export function useMenuAnchorRef(
         containerDiv.setAttribute('id', 'typeahead-menu');
         containerDiv.setAttribute('role', 'listbox');
         containerDiv.style.display = 'block';
-        containerDiv.style.position = 'fixed';
+        containerDiv.style.position = 'absolute';
         parent.append(containerDiv);
       }
       anchorElementRef.current = containerDiv;


thegreatercurve (approved):

ebads67:
> Reverting as the menu does not work correctly when the page is scrolled. If you scroll down the entire viewport, the menu doesn't even show.
> 
>  Screen.Recording.2024-02-15.at.9.11.55.AM.mov 
> @ebads67 can you share more details on your use case and we can follow up in a separate PR?
> 
> This reverts commit [0ebb413](https://github.com/facebook/lexical/commit/0ebb41307624af16cb2cdd48e9804ccb96edd32c).

This issue doesn't have anything to do with my PR, and existed before my change. I can fix it in a separate PR.
I think `Fixed` position for the anchor is correct as its position calculated relative to the viewport. `absolute` position also works in playground only because the anchor div is inserted inside the `document.body`.

2wheeh:
The wrong positioning of typeahead menu on scroll started from #5187

PR #5621, opened 2024-02-17
Unbullet one list item not the whole list

### LINKED ISSUE #5039: Bug: Toggling the list item is not working properly
When the text is highlighted and the user tries to remove the numbered/ bullet points from any specific word it automatically removes the numbered/ bullet points from other points too

https://github.com/facebook/lexical/assets/127859730/0da13777-55f4-416a-bd32-82d0d33f9ebc

Lexical version:

## Steps To Reproduce

1.Create a list with items
2. Try to remove the bullet or number list from the middle item
3. It will remove all the bullet or number list items

Link to code example: [lexical.](https://playground.lexical.dev/)

## The current behavior

When the text is highlighted and the user tries to remove the numbered/ bullet points from any specific word it automatically removes the numbered/ bullet points from other points too

## The expected behavior

It should only remove the selected bullet/number element

ivailop7:
Before

https://github.com/facebook/lexical/assets/7893468/65dd9a8c-45f6-435a-a002-1c6232e41150

After

https://github.com/facebook/lexical/assets/7893468/654e5111-c84b-4b2e-8fd1-3f95690f984f

Fixes #5039

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/List.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/List.spec.mjs
@@ -415,28 +415,33 @@ test.describe('Nested List', () => {
       page,
       html`
         <p
-          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
-          dir="ltr">
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__indent PlaygroundEditorTheme__ltr"
+          dir="ltr"
+          style="padding-inline-start: calc(120px)">
           <span data-lexical-text="true">Hello</span>
         </p>
         <p
-          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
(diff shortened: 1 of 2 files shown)


acywatson:
Nice! I'm really surprised this works, as we have a bunch of logic in the lists module to split the lists in various nesting situations. I wonder if this means we can get rid of that stuff, if setBlocksType is handling it well...

zurfyx (dismissed):
Nice! As Acy indicated, I wonder if we can simplify the lists code.

Besides that, would be nice to have an external module to share toolbar code so that we don't have to replicate the fixes elsewhere now, I'll put it on our shared list

acywatson:
Looks like we still do have 3 lists tests consistently failing

ivailop7:
will look into this again over the weekend

ivailop7:
@acywatson tests fixed

acywatson (approved):

PR #5625, opened 2024-02-18
Fix forward word deletion destructing element nodes

ivailop7:
Forward word deletion (Ctrl + Del on Windows) would destroy composite element nodes (tables, columns layout).
The logic to handle this was already implement inside Delete Character, so extracted it to helper and reused it inside Delete Word as well.

Before: 

https://github.com/facebook/lexical/assets/7893468/2419b44b-2b76-4256-b185-581c739dc8e3

https://github.com/facebook/lexical/assets/7893468/ebdace55-1fb3-447c-8c4b-cd8c5c488909

After:

https://github.com/facebook/lexical/assets/7893468/1e8ebefb-49ed-4e70-99c3-4e1454bf234c

https://github.com/facebook/lexical/assets/7893468/d9cf00f3-ec10-490d-97bb-2f5503968855

### DIFF
--- a/packages/lexical-playground/src/nodes/LayoutContainerNode.ts
+++ b/packages/lexical-playground/src/nodes/LayoutContainerNode.ts
@@ -65,6 +65,10 @@ export class LayoutContainerNode extends ElementNode {
     return $createLayoutContainerNode(json.templateColumns);
   }
 
+  isShadowRoot(): boolean {
+    return true;
+  }
+
   canBeEmpty(): boolean {
     return false;
   }
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1582,7 +1582,9 @@ export class RangeSelection implements BaseSelection {
     if (this.isCollapsed()) {
       const anchor = this.anchor;
       let anchorNode: TextNode | ElementNode | null = anchor.getNode();
-      if (this.forwardDeletion(anchor, anchorNode, isBackward)) return;
+      if (this.forwardDeletion(anchor, anchorNode, isBackward)) {
+        return;
+      }
 
       // Handle the deletion around decorators.
       const focus = this.focus;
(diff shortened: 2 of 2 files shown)


packages/lexical-playground/src/nodes/LayoutContainerNode.ts
```diff
@@ -65,6 +65,10 @@ export class LayoutContainerNode extends ElementNode {
     return $createLayoutContainerNode(json.templateColumns);
   }
 
+  isShadowRoot(): boolean {
+    return true;
+  }
```
zurfyx, line 70: This shouldn't be needed, as @fantactuka pointed out for tables, a shadow root shouldn't be needed to encapsulate another shadow root. (we left it for tables because it was fixing some bugs...)

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1548,27 +1581,11 @@ export class RangeSelection implements BaseSelection {
...
-          $isElementNode(anchorNode) &&
-          anchor.offset === anchorNode.getChildrenSize()) ||
-          (anchor.type === 'text' &&
-            anchor.offset === anchorNode.getTextContentSize()))
-      ) {
-        const parent = anchorNode.getParent();
-        const nextSibling =
-          anchorNode.getNextSibling() ||
-          (parent === null ? null : parent.getNextSibling());
+      if (this.forwardDeletion(anchor, anchorNode, isBackward)) return;
```
zurfyx, line 1585: 😇
```suggestion
      if (this.forwardDeletion(anchor, anchorNode, isBackward)) {
        return;
      }
```

zurfyx (approved):
That's a clean fix!

packages/lexical-playground/src/nodes/LayoutContainerNode.ts
ivailop7: Keeping this one for now, until we do we generic solution for all 'non-inlined element nodes' escaping when last or first nodes, since it fixes the forward deletion for the layout plugin

ivailop7:
@zurfyx this one is ready

zurfyx (approved):

PR #5626, opened 2024-02-18
Escape Layout Plugin

### LINKED ISSUE #5375: Bug: Unable to delete LayoutItemNode and LayoutContainerNode when there are no paragraph before and after the node
Lexical version: 0.12.5

## Steps To Reproduce

2. Insert a "Columns Layout"
3. Remove the paragraph before and after the node
4. try to delete the columns layout

## The current behavior
- Unable to delete the node

## The expected behavior
- Should be able to delete the node on hitting backspace/delete
- Also it would be nice if we can add a delete icon on hover which on click deletes the node.

Currently we don't have any example on the lexical-playground which demonstrates better deleting experience for an element node

ivailop7:
Fixes #5375

Up/Left Arrow when no node above Columns layout add a paragraph before or Down/Right Arrow add a paragraph below if such doesn't exist. I've used the Collapsible section logic, just simplified it a bit.

### DIFF
--- a/packages/lexical-playground/src/plugins/LayoutPlugin/LayoutPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/LayoutPlugin/LayoutPlugin.tsx
@@ -9,12 +9,23 @@
 import type {ElementNode, LexicalCommand, LexicalNode, NodeKey} from 'lexical';
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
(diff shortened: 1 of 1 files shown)


zurfyx (approved):
Stamping to unblock but I think it's worth discussing on our next sync. Block cursors were supposed to fix this problem in a generic manner. The fact that we need this code in place probably means we need a better generic solution

PR #5628, opened 2024-02-18
[0.14] Remove unused canExtractContents

ivailop7:
Was meant to be since 0.10, the method was defined only on TableNode and unused anywhere else. Should be safe to remove, since nobody should be relying on this externally.


acywatson (approved):

acywatson:
> nobody should be relying on this externally

How do we know? Probably safest to do a "major" version bump with this.

PR #5636, opened 2024-02-20
Revert "Fix hashtag regex blocking transform to heading"

acywatson:
@2wheeh 

Unfortunately, I made a mistake here - some Meta surfaces support multiple leading "#" characters in a hashtag, so this would break them. Instead, we might need a generic way to override the regex.

Reverts facebook/lexical#5376


zurfyx (approved):

2wheeh:
> some Meta surfaces support multiple leading "#" characters in a hashtag, so this would break them.

This was I worried... Thanks for check.
Could you reopen #5366 ?

PR #5643, opened 2024-02-21
fix(iframe): use window of current view in checking event instances

bilobom:
## conext
In checking for event prototype ancestry , lexical assumes the default window is the browser window.
example:
```js
event instanceof DragEvent
```  
 this is like : 
```js
event instanceof window.DragEvent
```  
This will fail and return false even thought the event is indeed a `DragEvent` because the event is an instance of the iframe window's DragEvent :
```js
event instanceof iframe.contentWindow.DragEvent
```

instanceof checks for Prototype ancestry. When you ask is some event  instanceof DragEvent it translate to 
```js
event.prototype === DragEvent
```
but 
```js
window.DragEvent !== iframe.contentWindow.DragEvent
```
## PR 

This pr tries to do the following:
1- Export `getWindow` function as utility throught `@lexical/utils`
2- Uses `getWindow` to access window whenever instanceof is used on constructor that belongs to window where editor resides

## PoC maybe ??
Yes here is a link to code demonstating this :
https://stackblitz.com/edit/vitejs-vite-sfz7ft?file=src%2FApp.tsx

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -288,6 +288,21 @@ export const $findMatchingParent: {
 };
 
 /**
+ * get the window of current default view .
+ * @param editor - The lexical editor
+ * @returns window of current default view
+ */
+export function getWindow(
+  editor: LexicalEditor,
+): NonNullable<LexicalEditor['_window']> {
+  const windowObj = editor._window;
+  if (windowObj === null) {
+    invariant(false, 'window object not found');
+  }
+  return windowObj;
+}
+
+/**
  * Attempts to resolve nested element nodes of the same type into a single node of that type.
  * It is generally used for marks/commenting
  * @param editor - The lexical editor
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -288,21 +288,6 @@ export const $findMatchingParent: {
 };
 
 /**
- * get the window of current default view .
- * @param editor - The lexical editor
- * @returns window of current default view
- */
-export function getWindow(
-  editor: LexicalEditor,
-): NonNullable<LexicalEditor['_window']> {
-  const windowObj = editor._window;
-  if (windowObj === null) {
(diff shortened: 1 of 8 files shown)


packages/lexical-utils/src/index.ts
```diff
@@ -287,6 +287,21 @@ export const $findMatchingParent: {
   return null;
 };
 
+/**
+ * get the window of current default view .
+ * @param editor - The lexical editor
+ * @returns window of current default view
+ */
+export function getWindow(
```
acywatson, line 295: Maybe instead of (or in addition to) this, we should just have something like our own implementation of instanceof:

```
function instanceOf(instance, target): boolean
```

which just gets the correct window in it's implementation.

bilobom: I'm not sure how we can get the window from this function, we would need to get the `target` that blongs to window of current view beforehand : 
```js
function instanceOf(instance, target): boolean
```
usage:
```js
const windowOfCurrentView =  getWindow(editor)
inctanceof(event, windowOfCurrentView.DrageEvent)
```

bilobom: In retrospect, exporting the `getWindow` might help in other situations where properties of window/document are accessed .
examples:
```js
window.getSelection()
document.caretRangeFromPoint
document.createRange()
window.createElement
```   
we can see a number of PRs that addresses this issue and this function might help
see discussions here #4649
here also @zurfyx comment here https://github.com/facebook/lexical/pull/5547#pullrequestreview-1844055998

I'm willing to go around and address this iframe issues and i believe this utility is usefull

bilobom: @acywatson I have used `objectKlassEquals` function that was already implemented better than getWindow each time
can you review it again please ?

bilobom:
@acywatson , this is a gentle reminder of the fix, can you review it please. thanks in advance

etrepum:

PR #5644, opened 2024-02-21
Fix deletion issues when using Chromium Android

### LINKED ISSUE #5538: Bug: Choosing word suggestion on Android will sometimes delete parts or whole of the next word
Lexical version: 0.12.6

## Steps To Reproduce / The current behavior

https://github.com/facebook/lexical/assets/8348101/a08ed7e8-d3a7-4098-9136-0b07d0f98306

It's slightly harder to reproduce consistently in real-life usage, but I've been able to consistently reproduce the issue using the words `Rins place` as you can see in the video. When I move my cursor inside the word `Rins` and then select the suggestion `Runs`, it deletes part of the next word.

It seems to depend on where the cursor is in the word. In the video, you'll see that if I have the cursor between `Ri_ns` then it deletes 3 characters from the next word. If I have it between `Rin_s` then it deletes 2 characters and 1 character if I have it at the end of the word `Rins_`. If I have the cursor after the R, `R_ins`, then it doesn't delete any character from the next word.

## The expected behavior

Choosing a word suggestion should not delete anything from the next word

amanharwara:
Because `preventDefault` doesn't actually do anything on the `beforeinput` event on Chromium Android, `deleteContentBackward` will actually delete things twice, once by Lexical and once by the browser. This is the cause of issues like #4941 and #5538.

The issue with backspace deleting two characters at once was patched in #5077 by pretending to be in composition mode so that the else statement with the `DELETE_CHARACTER_COMMAND` line wouldn't be triggered. This patched the issue but in turn caused issues #5274 and #5259.

#5274 was fixed in #5282 by adding new logic to actually trigger the `DELETE_CHARACTER_COMMAND` if the current node had the length `<= 1`. #5259 was fixed in #5389 by resetting the compositionKey if the keys of the anchor and focus nodes is different.

The changes in this PR were originally made to fix #5538, but I realized that it also fixes the other Chromium Android deletion issues. Doing it this way, we don't have to pretend to be in composition mode and the code is also more clear about why that logic is required.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -13,7 +13,7 @@ import type {TextNode} from './nodes/LexicalTextNode';
 
 import {
   CAN_USE_BEFORE_INPUT,
-  IS_ANDROID,
+  IS_ANDROID_CHROME,
   IS_APPLE_WEBKIT,
   IS_FIREFOX,
   IS_IOS,
@@ -545,15 +545,13 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
       }
 
       if ($isRangeSelection(selection)) {
-        // Used for handling backspace in Android.
-        if (IS_ANDROID) {
-          $setCompositionKey(selection.anchor.key);
-        }
+        const isSelectionAnchorSameAsFocus =
+          selection.anchor.key === selection.focus.key;
 
         if (
           isPossiblyAndroidKeyPress(event.timeStamp) &&
           editor.isComposing() &&
-          selection.anchor.key === selection.focus.key
+          isSelectionAnchorSameAsFocus
         ) {
           $setCompositionKey(null);
           lastKeyDownTimeStamp = 0;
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -545,15 +545,13 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
...
-          $setCompositionKey(selection.anchor.key);
-        }
+        const isSelectionAnchorSameAsFocus =
+          selection.anchor.key === selection.focus.key;
 
         if (
           isPossiblyAndroidKeyPress(event.timeStamp) &&
           editor.isComposing() &&
-          selection.anchor.key === selection.focus.key
+          isSelectionAnchorSameAsFocus
```
ivailop7, line 554: you can use 'selection.isCollapsed()'

packages/lexical/src/LexicalEvents.ts
```diff
@@ -573,15 +571,23 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
...
-          }
         } else {
           $setCompositionKey(null);
           event.preventDefault();
-          dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
+          // Chromium Android at the moment seems to ignore the preventDefault
+          // on 'deleteContentBackward' and still deletes the content. Which leads
+          // to multiple deletions. So we let the browser handle the deletion in this case.
+          const selectedNodeText = selection.anchor.getNode().getTextContent();
+          const hasSelectedAllTextInNode =
```
ivailop7, line 581: I believe there was a method that already does that but don't remember the name. It's possible it was for all the text on the line. Worth checking some of the $ methods in the Selection package

quanvan1999:
There are numerous problems with Android. Can you merge the issue?

PR #5652, opened 2024-02-23
Add support for grid selection in text alignment.

### LINKED ISSUE #5638: Bug: Change text alignment in multiple cells
It's not possible to change text alignment in a table when multiple cells are selected.

Lexical version: 0.13.1

## Steps To Reproduce

2. Add a table 5x5.
3. Add any text to the cells.
4. Select multiple cells.
5. Change text alignment.

## The current behavior
The text in the cells is not aligned.

## The expected behavior
The text alignment in the cells is changed.

https://github.com/facebook/lexical/assets/5062807/e055cab8-9d35-400a-be00-99cde948f438

serey-roth:
The added support allows text alignment across multiple cells simultaneously by selecting those cells. Fixes #5638.

Before:

[before-text-alignment.webm](https://github.com/facebook/lexical/assets/88986106/74ee321e-b55e-4834-aa7d-0bc9613055c9)

After:

[after-text-alignment.webm](https://github.com/facebook/lexical/assets/88986106/629324cb-4619-4b35-af7a-bb9c6a05a6ff)

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -632,7 +632,11 @@ export function registerRichText(editor: LexicalEditor): () => void {
       FORMAT_ELEMENT_COMMAND,
       (format) => {
         const selection = $getSelection();
-        if (!$isRangeSelection(selection) && !$isNodeSelection(selection)) {
+        const validSelection =
+          $isRangeSelection(selection) ||
+          $isNodeSelection(selection) ||
+          DEPRECATED_$isGridSelection(selection);
+        if (!validSelection) {
           return false;
         }
         const nodes = selection.getNodes();
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -34,6 +34,7 @@ import {
   $moveCharacter,
   $shouldOverrideDefaultCharacterSelection,
 } from '@lexical/selection';
+import {$isTableSelection} from '@lexical/table';
 import {
   $findMatchingParent,
   $getNearestBlockElementAncestorOrThrow,
@@ -635,7 +636,7 @@ export function registerRichText(editor: LexicalEditor): () => void {
         const validSelection =
           $isRangeSelection(selection) ||
           $isNodeSelection(selection) ||
-          DEPRECATED_$isGridSelection(selection);
+          $isTableSelection(selection);
         if (!validSelection) {
           return false;
         }
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -34,7 +34,6 @@ import {
   $moveCharacter,
   $shouldOverrideDefaultCharacterSelection,
 } from '@lexical/selection';
-import {$isTableSelection} from '@lexical/table';
 import {
   $findMatchingParent,
   $getNearestBlockElementAncestorOrThrow,
@@ -633,11 +632,7 @@ export function registerRichText(editor: LexicalEditor): () => void {
       FORMAT_ELEMENT_COMMAND,
       (format) => {
         const selection = $getSelection();
-        const validSelection =
-          $isRangeSelection(selection) ||
-          $isNodeSelection(selection) ||
-          $isTableSelection(selection);
-        if (!validSelection) {
+        if (!$isRangeSelection(selection) && !$isNodeSelection(selection)) {
           return false;
         }
         const nodes = selection.getNodes();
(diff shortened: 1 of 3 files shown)


packages/lexical-rich-text/src/index.ts
```diff
@@ -641,7 +641,11 @@ export function registerRichText(editor: LexicalEditor): () => void {
       FORMAT_ELEMENT_COMMAND,
       (format) => {
         const selection = $getSelection();
-        if (!$isRangeSelection(selection) && !$isNodeSelection(selection)) {
+        const validSelection =
+          $isRangeSelection(selection) ||
+          $isNodeSelection(selection) ||
+          DEPRECATED_$isGridSelection(selection);
```
ivailop7, line 647: I believe you should be using isTableSelection, but @zurfyx should be able to confirm.

serey-roth: Updated!

packages/lexical-rich-text/src/index.ts
```diff
@@ -632,7 +633,11 @@ export function registerRichText(editor: LexicalEditor): () => void {
       FORMAT_ELEMENT_COMMAND,
       (format) => {
         const selection = $getSelection();
-        if (!$isRangeSelection(selection) && !$isNodeSelection(selection)) {
+        const validSelection =
+          $isRangeSelection(selection) ||
+          $isNodeSelection(selection) ||
+          $isTableSelection(selection);
+        if (!validSelection) {
```
ivailop7, line 640: This is not the right approach to fix this, because it couples the 'rich-text' package to the 'tables' package, because of the isTableSelection import. The way to fix this correctly is to register FORMAT_ELEMENT_COMMAND inside the table plugin (LexicalTableSelectionHelpers.ts), similar to: https://github.com/facebook/lexical/blob/main/packages/lexical-table/src/LexicalTableSelectionHelpers.ts#L339-L369

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -367,6 +370,33 @@ export function applyTableHandlers(
...
+        formatSelection.focus.set(
+          currentFocus.key,
+          currentFocus.offset,
+          currentFocus.type,
+        );
+        $setSelection(formatSelection);
+        return false;
+      },
+      COMMAND_PRIORITY_CRITICAL,
+    ),
```
serey-roth, line 397: @ivailop7 I'm transforming the selection to a `RangeSelection` and let the command in https://github.com/facebook/lexical/blob/69754ac50d5e025daacc35cd0cd79a50e76a3ab6/packages/lexical-rich-text/src/index.ts#L631-L651 handles the rest. I've found this to be a better solution than duplicating the existing format command. What are your thoughts?

ivailop7 (approved):
LGTM, could @zurfyx , @acywatson or @fantactuka have a look as well.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
zurfyx: 1. While it works for this particular case, the selection you've expressed above is not valid. Table selection can't be directly converted to a RangeSelection. Both anchor and focus are ElementNodes so their offset is unlikely to be 0. Similarly, you can't have a selection that goes from inside a ShadowRoot to another ShadowRoot.
2. A RangeSelection is linear and traverses the TreeNodes in order. Notice that table selection is a visually "rectangular" selection. For this reason, your implementation may target nodes that are not even selection, see video below.
3. For best UX, we want to restore selection to the previous value.

https://github.com/facebook/lexical/assets/193447/3ee92660-a955-4385-8a5b-76aca02c534f

For 2, you can leverage  `$computeTableMap` to extract all the (relevant) table cells.

zurfyx (changes requested):
Thank you for taking a stab at this! Added a comment

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -367,6 +370,59 @@ export function applyTableHandlers(
...
+        for (let i = minRow; i <= maxRow; i++) {
+          for (let j = minColumn; j <= maxColumn; j++) {
+            const cell = tableMap[i][j].cell;
+            cell.setFormat(formatType);
+
+            const cellChildren = cell.getChildren();
+            for (let k = 0; k < cellChildren.length; k++) {
+              const child = cellChildren[k];
+              if ($isElementNode(child) && !child.isInline()) {
+                child.setFormat(formatType);
```
serey-roth, line 415: https://www.loom.com/share/9bacba8043234c5f8013652d5c010573?sid=3f6684e5-201d-43d9-9c65-035a1370d439

zurfyx: @serey-roth - Yes, that looks like another bug. In fact, I'd expect that pressing Enter on a centered paragraph would center the next one.

zurfyx (dismissed):
LGTM, thanks for iterating on this! Do you mind writing an E2E test for it please?

zurfyx (approved):
Thank you!

PR #5653, opened 2024-02-23
Refactor classname splitting from ' ' to /\s+/

sakib412:
To fix this issue: https://discord.com/channels/953974421008293909/1180070470935117884/1180070470935117884

image


acywatson:
Doesn't this address the same issue?

https://github.com/facebook/lexical/pull/5649

etrepum:
@acywatson Looks like it's the same sort of solution as #5649, feel free to close that one if this approach chosen instead

Issue #4457, opened 2023-05-05
Bug: Multiple line selection presence in collaboration is not display correctly


TehAwol:
Lexical version: v0.10.0 (latest)

## Steps To Reproduce

1. Copy paste some text from a lorem ipsum [generator](https://www.lipsum.com/) (Plain text paste works as expected)
2. Make a selection across multiple lines

Link to code example: [Reproducible on playground](https://playground.lexical.dev/split/?isCollab=true)

![Screenshot 2023-05-05 at 08 50 00](https://user-images.githubusercontent.com/56039404/236394399-98505f82-ced7-4843-8740-237c8eefc698.png)

## The current behavior
Multiple line selection presence is only shown on the initial and last line. Lines in between are not highlighted at all except a tiny slit at the end of the line.

## The expected behavior
All lines should be highlighted the same as the initial and the last line.

fantactuka:
Which browser/OS did you use. Works well on Mac/Chrome:

<img width="1728" alt="Screenshot 2023-05-05 at 7 30 21 PM" src="https://user-images.githubusercontent.com/132642/236584588-9a2b078a-30c4-49f9-8c4b-badfb2fc72aa.png">

TehAwol:
Have had same issue on Brave (Chromium) and Firefox. Seems to be intimately linked with copying and pasting several paragraphs.

Image below of behaviour on Firefox
![image](https://user-images.githubusercontent.com/56039404/236620923-201a996d-8192-4f7d-90f0-ff094c7d3b92.png)

fantactuka:
Can you do a recording?

TehAwol:
Here you go. This is with text copied and pasted directly from a lorem ipsum generator :
https://user-images.githubusercontent.com/56039404/236800610-48a92e7d-f041-4a86-8bda-1444fed5ac41.mov

hamza512b:
The issue is happening if the text has justify format. I am working on a fix https://github.com/facebook/lexical/pull/5657

https://github.com/facebook/lexical/assets/63897375/09a60c66-032c-46c1-9dbe-47f590f96653

hamza512b:
I am just commenting on this in case someone come across this issue, to fix this behavior. New users should use `selectionHighlight` prop on the  `CollaborationPlugin`. This has been decided to be opt in to preserve existing style sheets rules. Since this new API.

PR #5657, opened 2024-02-25
Fixing problems with collab cursors

hamza512b:
While I was using Lexical for my side project. I come  accross these problems with cursors in collabaration.

- [x]  The cursors are unstylable.
- [x]  They do not change direction on rtl locale.
~~They do not work with line mutli line selection.~~

In this pr I intend to fix these issues or at least some of them.


packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -254,7 +370,50 @@ function updateCursor(
...
+
+    // Calculating selection is forward or backward since it not saved on the selection
+    const isForwardSelection =
+      anchorElement.compareDocumentPosition(
+        editor.getElementByKey(focusKey)!,
+      ) === Node.DOCUMENT_POSITION_FOLLOWING;
+
+    // Running this calculation on all elements is expensive,
+    // we could do virtualization to only calculate the elements in view
+    // or cap the number of elements to calculate
```
hamza512b, line 389: Calculating rectangules could be slow if the user for example selects the whole document and document is very large. To mitagte the problem we could solve in different ways.

hamza512b:
Let give background:

The [https://github.com/facebook/lexical/issues/4457](https://github.com/facebook/lexical/issues/4457) (multi line selection is broken) is happening because the  [createRectsFromDOMRange](https://github.com/hamza512b/lexical/blob/16ed8d55cd9a5b6cc53f8adca7bd208e23cdd050/packages/lexical-selection/src/utils.ts#L124) function is using `range.getClientRects`.  The function has different behivour on each browser. I tested accross different browsers and each browser is giving different output. 


This might be related: https://github.com/w3c/csswg-drafts/issues/7155

fantactuka:
Hi @hamza512b, I've tried to address issue with overlapping collab selection, and looks like it can be fixed by adjusting this part:

https://github.com/facebook/lexical/blob/7ec15be43b988e9742947f29dbfd362f90b71f7f/packages/lexical-selection/src/utils.ts#L150-L169

to 

```
  const singleElementRects = [];
  const selectionRectsLength = selectionRects.length;
  for (let i = 0; i < selectionRectsLength; i++) {
    const selectionRect = selectionRects[i];
    if (selectionRect.width + rootPadding < rootRect.width) {
      singleElementRects.push(selectionRect);
    }
  }

  let prevRect = null;
  const nonOverlappingRects = [];
  const singleElementRectsLength = singleElementRects.length;
  for (let i = 0; i < singleElementRectsLength; i++) {
    const selectionRect = singleElementRects[i];
    const isOverlappingRect =
      prevRect != null &&
      prevRect.top <= selectionRect.top &&
      prevRect.top + prevRect.height > selectionRect.top &&
      prevRect.left + prevRect.width > selectionRect.left;

    prevRect = selectionRect;

    if (!isOverlappingRect) {
      nonOverlappingRects.push(selectionRect);
    }
  }

  return nonOverlappingRects;
```

Where we should filter out rects that span across multiple elements first, and only after that remove overlapping ones

packages/lexical-yjs/src/SyncCursors.ts
hamza512b: This is not as bad as I first thought. Techniqually in production app there should be limit on how large the document should be. And performance issue is because we drawing many dom elements as once if document is too large (200 000 words)

hamza512b:
> Hi @hamza512b, I've tried to address issue with overlapping collab selection, and looks like it can be fixed by adjusting this part:
> 
> https://github.com/facebook/lexical/blob/7ec15be43b988e9742947f29dbfd362f90b71f7f/packages/lexical-selection/src/utils.ts#L150-L169
> 
> to
> 
> ```js
>   const singleElementRects = [];
>   const selectionRectsLength = selectionRects.length;
>   for (let i = 0; i < selectionRectsLength; i++) {
>     const selectionRect = selectionRects[i];
>     if (selectionRect.width + rootPadding < rootRect.width) {
>       singleElementRects.push(selectionRect);
>     }
>   }
> 
>   let prevRect = null;
>   const nonOverlappingRects = [];
>   const singleElementRectsLength = singleElementRects.length;
>   for (let i = 0; i < singleElementRectsLength; i++) {
>     const selectionRect = singleElementRects[i];
>     const isOverlappingRect =
>       prevRect != null &&
>       prevRect.top <= selectionRect.top &&
>       prevRect.top + prevRect.height > selectionRect.top &&
>       prevRect.left + prevRect.width > selectionRect.left;
> 
>     prevRect = selectionRect;
> 
>     if (!isOverlappingRect) {
>       nonOverlappingRects.push(selectionRect);
>     }
>   }
> 
>   return nonOverlappingRects;
> ```
> 
> Where we should filter out rects that span across multiple elements first, and only after that remove overlapping ones


This wont work, maybe it could work on chrome but it will not work on firefox because the https://github.com/facebook/lexical/pull/5657#issuecomment-1965388205 

`getClientRects` in chrome we getting output that we might be able to deal with in some way. But in firefox the output we are getting are simpily not enough. 

I think we need to aproach this in different way completly without using Range.getClientRects()

Log the getClientRects output to see for yourself in firefox, chrome vs safari. It is behaving differently.

hamza512b:
I have pushed new commit. Before there was issues where it would show the selection incorrectly if anchor or focus node were on end of paragraph node. 

Related to the optional classnames, I will make these changes in the next commit.

hamza512b:
I am thinking about the optional classnames, It might not be very good idea because we have by default inline styles and it is hard to override inlines styles. We would have to write !important on the properities we would like to override. Just letting you know.

But maybe your way is better for backward compatibility too.

amanharwara:
Yeah overriding inline styles can be a bit hard. Maybe what can be done for that is that each of the property that is currently added in the inline style by default can be added as a CSS variable instead of an actual property & value. And then there can be a default className which implements those in the theme CSS file.
This way it can be overriden easily if required, and is still backwards compatible for people who don't want to do that.

What do you all think?

hamza512b:
Hi @amanharwara !
I am not familiar how do you define a default className in Lexical? I thought Lexical does not import any styles, and developers have to import the style files themselves.

Unless you mean the playground CSS file which is not part of the actual library. In that case I have already done that until I changed as requested.

amanharwara:
> I thought Lexical does not import any styles, and developers have to import the style files themselves.
> 
> Unless you mean the playground CSS file which is not part of the actual library. In that case I have already done that until I changed as requested.

Ah yes, you are right, my bad. That was basically what I was suggesting, too. But I guess that wouldn't actually be backwards-compatible, since you'd need to have those styles defined outside the core.

Maybe if a className is passed, then the inline styles shouldn't get added? That would allow having easy control over the styling of those elements.

That being said, I think overall having to use `!important` for some properties might not be the worst thing if having the inline styles is absolutely required. @fantactuka is there a reason why we can't omit the cssText if a theme class is provided?

hamza512b:
As I see right now, the first part of the PR is far more complicated than imagined. I tried finding solution, implementing an alternative for `Range.getClientRects`, but with no luck (It works paritially but with still some problemes).  I think what I am trying to do is a work arround for something that is initially flawed. So I came to this conclusion:

It is not pratical having cursors as absolutely positioned. Instead this could have been much simpiler if we had a way to make selections inline within the text.

So I have reverted the first part of the part and keep the second part.

*Note I update the Pull request details so it is more relavent

PR #5664, opened 2024-02-28
docs: add a tip about discrete state update

### LINKED ISSUE #5660: Bug: "HTML -> Lexical" code snippet from the documentation does not work
Lexical version: 0.13.1

## Steps To Reproduce

1. Go to the [docs page](https://lexical.dev/docs/concepts/serialization#html---lexical) describing HTML -> Lexical conversion in headless environment.
2. Try to run the attached snippet.

Link to code example: https://stackblitz.com/edit/vitest-dev-vitest-t95w44

## The current behavior

Editor state is empty (no children). `$generateNodesFromDOM` seems to work properly though, so the problem is related to the insertion phase.

## The expected behavior

Editor state consists of the DOM elements imported using the snippet.

Dzieni:
Fixes #5660 

Also, I think it might be reasonable to copy/adapt the [discrete update section of @DaniGuardiola's article](https://dio.la/article/lexical-state-updates#discrete-updates) into the docs. Applying `discrete: true` sounds like a common use case in a headless context.

### DIFF
--- a/packages/lexical-website/docs/concepts/serialization.md
+++ b/packages/lexical-website/docs/concepts/serialization.md
@@ -84,6 +84,12 @@ editor.update(() => {
 });
 ```
 
+:::tip
+
+Remember that state updates are asynchronous, so executing `editor.getEditorState()` immediately afterwards might not return the expected content. To avoid it, [pass `discrete: true` in the `editor.update` method](https://dio.la/article/lexical-state-updates#discrete-updates).
+
+:::
+
 #### `LexicalNode.importDOM()`
 You can control how an `HTMLElement` is represented in `Lexical` by adding an `importDOM()` method to your `LexicalNode`.
 


packages/lexical-website/docs/concepts/serialization.md
```diff
@@ -84,6 +84,12 @@ editor.update(() => {
 });
 ```
 
+:::tip
+
+Remember that state updates are asynchronous, so executing `editor.getEditorState()` immediately afterwards might not return the expected content. To avoid it, [pass `discrete: true` in the `editor.update` method](https://dio.la/article/lexical-state-updates#discrete-updates).
```
acywatson, line 89: Thanks for this PR!

While I do think it's good to mention the "discrete" tag, I don't want to encourage people to do this. In most cases, people should just be using the standard listener pattern to interact with the state. I think leaning on discrete too heavily can encourage inefficient patterns.

My suggestion here is to also include the consequences of adding the discrete tag, one of which is that the update won't be batched, which can lead to unnecessary update cycles.

Another suggestion would be to add a new section about tags in the "update" section of the docs, because this is really not an issue specific to serialization.

acywatson (changes requested):

packages/lexical-website/docs/concepts/serialization.md
Dzieni: @acywatson Just to be clear - is there a better way than `discrete: true` to solve the problem mentioned in #5660? Maybe `htmlToLexical` should return a promise that gets resolved by the editor update listener?

acywatson: > @acywatson Just to be clear - is there a better way than discrete: true to solve the problem mentioned in https://github.com/facebook/lexical/issues/5660? Maybe htmlToLexical should return a promise that gets resolved by the editor update listener?

Ah, I see - I didn't actually look at the issue - I'm sorry about that.

So, yes: discrete is one option here. Another option is to pass an onUpdate callback:

```
editor.update(() => {

}, {
  onUpdate: () => {  editor.getEditorState().read(() => {})... }
})
```

I think your change is fine, then. You can also add something about onUpdate, if you want.

acywatson (approved):

packages/lexical-website/docs/concepts/serialization.md
Dzieni: OK, I think I'll add a few words in the "Updating state" chapter about possible waits to await the update.

acywatson:
I'll merge this for now, please feel free to add the additional docs later.

PR #5665, opened 2024-02-28
Refactor: simplify getFirstDescendant and getLastDescendant

thorn0:


packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -147,29 +147,19 @@ export class ElementNode extends LexicalNode {
...
-        const child = node.getFirstChild<T>();
-        if (child !== null) {
-          node = child;
-          continue;
-        }
-      }
-      break;
+    while ($isElementNode(node)) {
+      const child = node.getFirstChild<T>();
+      if (child === null) break;
```
fantactuka, line 152: nit: I believe we always add `{}` for conditions

fantactuka (approved):

acywatson (approved):

acywatson (approved):

PR #5666, opened 2024-02-28
Configure ESLint: enable curly rule, fix on save

thorn0:


.vscode/settings.json
```diff
@@ -1,8 +1,7 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "eslint.alwaysShowStatus": true,
```
fantactuka, line 3: Do we have to remove it?

packages/lexical-code/src/CodeNode.ts
```diff
@@ -272,7 +272,9 @@ export class CodeNode extends ElementNode {
           let spaces = 0;
           const text = node.getTextContent();
           const textSize = node.getTextContentSize();
-          for (; spaces < textSize && text[spaces] === ' '; spaces++);
```
thorn0, line 275: ESLint adds empty braces here and then complains about an empty block, so I changed this to `while`.

acywatson (approved):
lgtm

PR #5667, opened 2024-02-28
docs: Added visualization of the Lexical modular architecture

StyleT:
This highlights few things:
- Lexical is a framework for editor creation, not a ready to use editor
- What are the most vital plugins user may need to create own editor

As I was exploring codebase & documentation it took me several hours to draw proposed picture in my mind, hope it will save some time for others!

Note: Diagram created via draw.io and it's source is embedded into svg file


acywatson:
Even if this isn't perfect, I think it certainly adds value over what we have now.

Thanks Vlad!

acywatson (approved):

PR #5668, opened 2024-02-28
docs: Interactive Vanilla JS example

### LINKED ISSUE #5671: Bug: CodeSandbox is broken at https://lexical.dev/
## Steps To Reproduce

1. Open https://lexical.dev/
2. Navigate to "Simple Setup" / "Powerful Features" / "Built to Extend" tabs
3. See that "Simple Setup" example is broken

I can fix this easy peasy in #5668 and we'll start storing codebase of these examples within repo, so such issue doesn't happen again
Screenshot 2024-02-29 at 2 03 16 PM

StyleT:
At this moment Vanilla JS documentation lacks complete example that user can simply "copy-paste" or fork and run locally within seconds. This is especially important considering that Lexical can't be bootstrapped with just a few lines of code, while playground example we have for React is quite overwhelming (especially for less experienced engineers).

Additionally this fixes #5671

Here the proposal is to utilize StackBlitz that allow to run dev environment locally in user browser. For example `react-router` also [uses it](https://github.com/remix-run/react-router/blob/main/examples/basic/README.md).

If this diff goes through - I'll do same minimalistic example for React.

Additionally I didn't create any tests for it intentionally for 2 reasons:
1. I don't want to spend extra time on tests before getting endorsement from core team on the approach
2. Let's discuss if we even need them as since dependencies versions here are locked - there is not much things that can go wrong. And tests wouldn't cover StackBlitz integration (most probable cause of issues). And if we want to have them - let's discuss if they can live alongside Lexical E2E tests.

P.S.: At this moment StackBlitz pointed to forked repo intentionally to allow functionality preview, this is to be fixed before merge

### DIFF
--- a/examples/vanilla-js/README.md
+++ b/examples/vanilla-js/README.md
@@ -1,0 +1,7 @@
+# Basic Vanilla JS example
+
+Here we have simplest Lexical setup in rich text configuration (`@lexical/rich-text`) with history (`@lexical/history`) and accessibility (`@lexical/dragon`) features enabled.
+
+**Run it locally:** `npm i && npm run dev`
+
+[![Open in StackBlitz](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/github/facebook/lexical/tree/main/examples/vanilla-js?file=src/main.ts)
--- a/examples/vanilla-js/index.html
+++ b/examples/vanilla-js/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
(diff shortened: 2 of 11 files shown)


acywatson:
Looks good! These tests seem to be failing locally on main, so I'll fix that separately.

StyleT:
@acywatson I'm on it ;) It's unrelated to this PR though

MathiasWP:
Would it be possible to get a fully playground Vanilla JS example?

It's quite hard to use Lexical in other frameworks that React for more advanced rich editing like tables, slash-menu, latex etc.

Issue #5662, opened 2024-02-28
Bug: Inserting a table from external source can lead to Lexical crash


vect-spasquereau:
Hello there. I'm working with lexical tables and I was trying to implement background color control. I was testing with tables copied from external resources (google doc, google spreadsheet). Lexical is able to import the tables, but is crashing when I trying to leave the table.

Lexical version: 0.13.1

## Steps To Reproduce

1. Go in Google doc, insert a table and copy it
2. Paste it into Lexical
3. Go to last cell and press the right arrow to focus the table end
4. Press Enter to insert a paragraph underneath the table
5. Lexical is crashing

https://www.loom.com/share/dc72b967351a49afa1de3e5f0d1d7b66?sid=c3b11039-a4fd-4877-a176-74204fd36c4a

Link to code example: reproductible in Lexical playground

After investigation, I think Lexical is going in an infinite loop in LexicalSelection -> removeTextAndSplitBlock
(@birtles I think you've been working on this 👀)

birtles:
@vect-spasquereau Sorry I'm on leave at the moment so I haven't had a chance to dig through old issues but did I mention working on something like that? Or did I cause a regression here?

vect-spasquereau:
@birtles sorry I didn't notice you are on leave. I just checked the file history and saw you were the last one implementing/refactoring the LexicalSelection -> removeTextAndSplit and LexicalSelection -> splitNodeAtPoint functions. I assumed it could mean something for you 🙂
Nevermind, it's an edge case, I'll try to come up with a fix to break this infinite loop

vect-spasquereau:
well, that was a wrong lead. Fixing the `convertTableCellNodeElement` when copying tables from external sources.

vect-spasquereau:
https://github.com/facebook/lexical/pull/5670

birtles:
> @birtles sorry I didn't notice you are on leave.

Not at all! GitHub doesn't make that information very visible.

> I just checked the file history and saw you were the last one implementing/refactoring the LexicalSelection -> removeTextAndSplit and LexicalSelection -> splitNodeAtPoint functions. I assumed it could mean something for you 🙂 Nevermind, it's an edge case, I'll try to come up with a fix to break this infinite loop

I'm afraid I haven't touched lexical code since that patch. It sounds like you worked it out though!

PR #5670, opened 2024-02-29
fix(tables): Add missing paragraph node when copying empty cells from external resources

vect-spasquereau:
Fixes https://github.com/facebook/lexical/issues/5662

### Issue

When copy/pasting a table from external sources (google doc, google spreadshet, e.g.), an empty cell is created with no child node :

image

### Change

Return `after` callback in `convertTableCellNodeElement` to insert paragraph if the cell is empty :

image

### DIFF
--- a/packages/lexical-table/src/LexicalTableCellNode.ts
+++ b/packages/lexical-table/src/LexicalTableCellNode.ts
@@ -316,6 +316,12 @@ export function convertTableCellNodeElement(
   }
 
   return {
+    after: (childLexicalNodes) => {
+      if (childLexicalNodes.length === 0) {
+        childLexicalNodes.push($createParagraphNode());
+      }
+      return childLexicalNodes;
+    },
     forChild: (lexicalNode, parentLexicalNode) => {
       if ($isTableCellNode(parentLexicalNode) && !$isElementNode(lexicalNode)) {
         const paragraphNode = $createParagraphNode();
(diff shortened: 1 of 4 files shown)


StyleT:
Hi! Do you think this can be covered with test?

vect-spasquereau:
hey @StyleT , I implemented a test for my fix. As table nodes are not massively covered, I added a bunch of tests for rows and cells.

zurfyx (commented):
Thank you. However, the paragraph will be automatically created upon typing, which resembles the same behavior as when you type on an empty editor (only a RootNode). Do you mind expanding on the problem you're trying to solve here?

vect-spasquereau:
@zurfyx unfortunately no, the paragraph node is not created upon typing in an empty cell 😕 It creates a text node, not a paragraph node:

image

The issue is when you focus an empty table cell node and that you press "Enter". There are some methods involved in `LexicalSelection` that are supposed to find the right position where the new paragraph should be inserted. These methods (`removeTextAndSplitBlock` and `splitNodeAtPoint`) create an infinite loop in this specific case that crashed lexical. I don't want to modify these functions because I'm not aware of the impacts it could have. Having a paragraph inside empty cells is fixing this issue.

Also, I noticed that a paragraph node was explicitely created in cells in methods like `$createTableNodeWithDimensions`, `$insertTableRow`, `$insertTableColumn`, etc. So I assumed it was the right thing to do 🙂

acywatson (approved):
I think this is reasonable. Will let @zurfyx confirm that he doesn't have further objections.

zurfyx (approved):

vect-spasquereau:
thx 🙌

PR #5677, opened 2024-03-02
docs: Added conceptual view of the Lexical Core Design

StyleT:
This allows the viewer to have a simple visual representation of what Lexical Core actually does and how User Application shall interact with it.

Any suggestions are welcomed!


acywatson (approved):
I think this works, but will give others a chance to review.

It doesn't really cover the asynchronous nature of state updates (i.e., batching), but I'm not sure that's strictly necessary. Another gap would be update listeners, which are also triggered at the end of the update cycle.

There are a few charts in here that may be helpful, though I think your approach works better for written documentation.

StyleT:
> I think this works, but will give others a chance to review.
> 
> It doesn't really cover the asynchronous nature of state updates (i.e., batching), but I'm not sure that's strictly necessary. Another gap would be update listeners, which are also triggered at the end of the update cycle.
> 
> There are a few charts in here that may be helpful, though I think your approach works better for written documentation.

Thanks for the feedback! I’ll add listeners for sure! 
Talking about the lack of async nature of the state update demonstration - it’s intentional to keep this diagram small and easy. More will follow on state lifecycle as I dig deeper into architecture here.

ivailop7 (approved):

acywatson:
> > I think this works, but will give others a chance to review.
> > It doesn't really cover the asynchronous nature of state updates (i.e., batching), but I'm not sure that's strictly necessary. Another gap would be update listeners, which are also triggered at the end of the update cycle.
> > There are a few charts in here that may be helpful, though I think your approach works better for written documentation.
> 
> Thanks for the feedback! I’ll add listeners for sure! Talking about the lack of async nature of the state update demonstration - it’s intentional to keep this diagram small and easy. More will follow on state lifecycle as I dig deeper into architecture here.

I just realized I forgot to add the link:

https://docs.google.com/presentation/d/1IanPm9qzLP4Oiqht83weMCZalCaCJpH4QvFqeD29-uQ/edit#slide=id.p

I meant to say "there are a few charts in here (above link) that may be helpful".

StyleT:
@acywatson @ivailop7 PTAL and merge, updated diagram a bit. 

I'll use info from https://docs.google.com/presentation/d/1IanPm9qzLP4Oiqht83weMCZalCaCJpH4QvFqeD29-uQ/edit#slide=id.p within other pages as within conceptual view we're treating `lexical` core as black box, while those examples showcase inner design.

acywatson (approved):

PR #5678, opened 2024-03-02
docs: Added reasoning on why State is necessary

StyleT:
I wasn't able to find any reasoning on why the State concept is necessary in Lexical so I added a few simple examples that helped me understand it.

It's a good practice to justify such architectural decisions.


acywatson:
Good enough, thanks!

acywatson (approved):

Issue #5620, opened 2024-02-17
How to keep formatting between lines?


nickschinestzki:
I'm trying to achieve a different behavior regarding the formatting between lines in the Lexical Editor.

### The current behavior
1. In an empty editor, the user starts adding formatting and styles in the editor (font-family, font-size, color, bold, italic, etc)
2. When you press `Enter` once, it will insert a new line, which will have an "unstyled cursor" but upon new typing, the styles and formatting are applied again to the text.
3. When you press `Enter` twice, the new line, separated by the other ones, will lose all of the previous formatting and styling, returning to the default. 

https://github.com/facebook/lexical/assets/33634787/b7fc7e60-0ce6-4d5c-be79-bea582a9a675

### The desired behavior
1. In an empty editor, the user starts adding formatting and styles in the editor (font-family, font-size, color, bold, italic, etc)
2. When pressing `Enter` regarding of how much times/lines, the styles and formats are preserved. The user doesn't have to select the font-family and font-size again (for example) every time he creates a new a paragraph.

How can I achieve this?

_Originally posted by @nickschinestzki in https://github.com/facebook/lexical/discussions/5619_

moy2010:
It sound like you'll need to prevent the stop the propagation of the `Enter` button event (check Lexical's command pattern), and handle yourself the creation of new paragraph nodes to conserve the existing state.

ivailop7:
@acywatson @zurfyx do we agree, this should be the default behaviour?

acywatson:
> @acywatson @zurfyx do we agree, this should be the default behaviour?

I mean the trade off is now that you have the opposite problem. If a user wants to just bold one line or add several formats to one segment of text, they need to then clear all those formats off when they start typing on the next line.

I suspect the case of adding formats to a small portion of otherwise non-formatted text (or even typing the text non-formatted, then going back to apply formats to specific sections) is far more common than the case of setting a bunch of formats and typing multiple lines with those formats engaged.

Probably the perspective we should come from here, though, is what's the easiest way to set it up so a developer can easily customize it for their own editor. From that angle, I think it might be better to make the formats sticky, and allow people to write a plugin that clears the formatting on enter or on double-enter, as they see fit.

Additionally, though, this would be a breaking change.

nickschinestzki:
I have discovered that using `Shift` + `Enter` results in the exact behavior I'm searching for. I'm having difficulties to formulate this into a Plugin, that would not require the user to keep pressing `Shift` to make it work.

[...]

PR #5692, opened 2024-03-09
Preserve style after double line break

ivailop7:
Fixes: #5620

The issue: When you press Enter twice, the styling gets lost in the skipped paragraph, same when you type inside a table cells.
Why change this: consistent behaviour with other text editors.

This essentially undoes the changes done in here: https://github.com/facebook/lexical/pull/5292 while still fixing the issue that PR officially addressed.

Before:

https://github.com/facebook/lexical/assets/7893468/328f5197-fa2d-457a-a4dd-7b2a4802ceb1

After:

https://github.com/facebook/lexical/assets/7893468/ae34f9f5-8146-4f0b-8580-181fe2bff74b

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -324,10 +324,6 @@ function onSelectionChange(
         const [lastFormat, lastStyle, lastOffset, lastKey, timeStamp] =
           collapsedSelectionFormat;
 
-        const root = $getRoot();
-        const isRootTextContentEmpty =
-          editor.isComposing() === false && root.getTextContent() === '';
-
         if (
           currentTimeStamp < timeStamp + 200 &&
           anchor.offset === lastOffset &&
@@ -343,9 +339,6 @@ function onSelectionChange(
             );
             selection.format = anchorNode.getFormat();
             selection.style = anchorNode.getStyle();
-          } else if (anchor.type === 'element' && !isRootTextContentEmpty) {
-            selection.format = 0;
-            selection.style = '';
           }
         }
       } else {


zurfyx (commented):
Hmm, which editors are you checking against for the table case? What I can see on GDocs and Word is that they retain the format regardless of whether you click again, but only for that specific position.

For the create paragraph, my hunch is that this logic is a one-off and they simply replicate the format when you insert a new paragraph.

Also, this PR seems to be preserving the style when clicking on a element position.

https://github.com/facebook/lexical/assets/193447/0d7ccb9b-0ac1-4311-8a78-20c10a19512c

zurfyx:
^ We should really be writing some tests for all this discussion we're having above 😛

ivailop7:
> Hmm, which editors are you checking against for the table case? What I can see on GDocs and Word is that they retain the format regardless of whether you click again, but only for that specific position.
> 
> For the create paragraph, my hunch is that this logic is a one-off and they simply replicate the format when you insert a new paragraph.
> 
> Also, this PR seems to be preserving the style when clicking on a element position.
> 
>  Screen.Recording.2024-03-12.at.9.42.24.AM.mov

Fair enough on tables, happy to keep it losing the current style between cells.

Losing the current format after a double line break, should be fixed though. If you have better suggestion on the approach, open to suggestions, since this is a super annoying problem when using the editor extensively. 
I went with cleaning up the element node style nullifying because you can apply a style to a link node already and I couldn't find a use case when no style made sense, could be wrong. Happy to go for an alternative approach.

zurfyx:
> > Hmm, which editors are you checking against for the table case? What I can see on GDocs and Word is that they retain the format regardless of whether you click again, but only for that specific position.
> > For the create paragraph, my hunch is that this logic is a one-off and they simply replicate the format when you insert a new paragraph.
> > Also, this PR seems to be preserving the style when clicking on a element position.
> > Screen.Recording.2024-03-12.at.9.42.24.AM.mov
> 
> Fair enough on tables, happy to keep it losing the current style between cells.
> 
> Losing the current format after a double line break, should be fixed though. If you have better suggestion on the approach, open to suggestions, since this is a super annoying problem when using the editor extensively. I went with cleaning up the element node style nullifying because you can apply a style to a link node already and I couldn't find a use case when no style made sense, could be wrong. Happy to go for an alternative approach.

Gave it a stab since I wasn't sure either, happy to follow up with a PR if you want - https://github.com/facebook/lexical/compare/main...eventsftw


https://github.com/facebook/lexical/assets/193447/b03a04fa-988c-41c7-bcc6-8dac2e5984d0

This should be very close to Word and GDoc. They seem to reset format but they leverage the empty "TextNode" to store this information when there's no text (this is a guess, I have no idea about their codebase).

ivailop7:
> > > Hmm, which editors are you checking against for the table case? What I can see on GDocs and Word is that they retain the format regardless of whether you click again, but only for that specific position.
> > > For the create paragraph, my hunch is that this logic is a one-off and they simply replicate the format when you insert a new paragraph.
> > > Also, this PR seems to be preserving the style when clicking on a element position.
> > > Screen.Recording.2024-03-12.at.9.42.24.AM.mov
> > 
> > 
> > Fair enough on tables, happy to keep it losing the current style between cells.
> > Losing the current format after a double line break, should be fixed though. If you have better suggestion on the approach, open to suggestions, since this is a super annoying problem when using the editor extensively. I went with cleaning up the element node style nullifying because you can apply a style to a link node already and I couldn't find a use case when no style made sense, could be wrong. Happy to go for an alternative approach.
> 
> Gave it a stab since I wasn't sure either, happy to follow up with a polished PR if you want - [main...eventsftw](https://github.com/facebook/lexical/compare/main...eventsftw)
> 
>  Screen.Recording.2024-03-12.at.2.37.16.PM.mov 
> This should be very close to Word and GDoc. They seem to reset format but they leverage the empty "TextNode" to store this information when there's no text (this is a guess, I have no idea about their codebase).

Looking good, you can use $getPreviousSelection I think, the rest makes sense. Go for a proper PR :)

PR #5697, opened 2024-03-10
Mention node as token

### LINKED ISSUE #5687: Bug:  mention deleting one by one
On the desktop, if the mention has two words, deleting them with backspace deletes mentions one by one. In Android, by removing the mentioned word, it behaves like deleting the text character by character. Usually, mentions don't allow cursor in it. By deleting the word, it removes the word with just one click. How do I improve this one?

Lexical version: 0.12.6

## Steps To Reproduce

1. Open lexical playground
2. check the mentioned words

ivailop7:
Fixes #5687


packages/lexical-playground/src/nodes/MentionNode.ts
```diff
@@ -120,7 +120,7 @@ export class MentionNode extends TextNode {
 
 export function $createMentionNode(mentionName: string): MentionNode {
   const mentionNode = new MentionNode(mentionName);
-  mentionNode.setMode('segmented').toggleDirectionless();
```
zurfyx, line 123: This was intentional, tokens are hardly accessible. The idea behind segmented is that screen readers still perceive it as text and they can edit naturally (breaking the mention when needed)

ivailop7: Ok, fair enough. Will close the PR. Why could token behave the same was for screen readers, just have a different deletion logic?

zurfyx: @ivailop7 Actually let's reopen the tagged issue, the behavior described in the issue for Android doesn't seem correct. Maybe there's something wrong in Android composition mode.

Issue #5698, opened 2024-03-10
Bug: Bulleting an image when image is selected crashes


ivailop7:
Version 0.13.1

https://github.com/facebook/lexical/assets/7893468/315afaaf-520f-4faa-86ea-108c1e59849e

**Expected Behaviour**
Should just bullet/unbullet respectively.

ivailop7:
I fixed the bulleting, but unbulleting still doesn't work.

etrepum:
Confirmed that unbulleting still does not work when on a NodeSelection

[repro link](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE52RUUvDMBSF_0o5z9nWbo7NvPrio7CBiPgQ27s2cpuM5LbdGP3v0hVFhyjuKdx7cs6XQ06gwooPGzFC0CcE72U488pyEchBP_82GJYtHQQaT8Tsu2THvqOQWJeIZUliZXeSMLkIhdzsxXo3xP9NfVEobKB8dLiGWWHnQ20GGBSsK8gJdKogxz1BjyEKLYV49mR93ytUZMtqvFebw6MtpIJepqlCrHx39_GkneFICjHk0JiZGEni7HjuNBk7TVrT3t7HlqZv-xKfVFubkr5hFbqRkvb_L8E2ihWqLxJbww0Nla5LvEgbVttRfW2YadCjmCBnVUwJjYZxBeyHP1BgE2VjWiqgs9ViMU_n6_XqZpkpRN-EfPA9sDmWwTeu-OpGOp1n0xT9Oz1cUBqnAgAA)

* Click on bulleted image node
* Press cmd-opt-4 or try and toggle the bullet list from the block section of the toolbar (also notice that it says "Normal" rather than "Bullet List")

It works fine when the selection is a RangeSelection before or after the image node, the problem only occurs for a NodeSelection

PR #5699, opened 2024-03-10
Fix Bulleting a Node Selection

ivailop7:
Partial fix for: #5698

When selection is NodeSelection, not a range one nodes like Image, Poll, etc. are not bulleted correctly.
 
Unbulleting doesn't work, but as the full proper fix will be quite involved - introducing a proper REMOVE_LIST_ITEM_COMMAND that unbulleting any node, just doing this part for the first half.

Review with hiding the whitespace, it's only 3-4 lines changes, but deindenting, makes it look as larger change.


zurfyx (approved):
LGTM!

Issue #5646, opened 2024-02-22
Bug: Header becomes a numbered list


ronaldlangeveld:
Lexical version: Latest

## Steps To Reproduce

1. Go to the Lexical Playground
2. Modify the **Welcome to the playground** header and make it **1. Welcome to the playground**
3. Notice it converts to a numbered list item instead of remaining a header as soon as you make a space between _1. and Welcome_

<img width="456" alt="Screenshot 2024-02-22 at 10 37 49" src="https://github.com/facebook/lexical/assets/19263291/474af925-03ab-46a4-8a8b-3dc9af34d4aa">

<img width="554" alt="Screenshot 2024-02-22 at 10 38 12" src="https://github.com/facebook/lexical/assets/19263291/24a41248-a771-4a4b-8832-14742683e370">

acywatson:
Thanks for reporting - what is the expected outcome here?

ronaldlangeveld:
> Thanks for reporting - what is the expected outcome here?

To be able to type a header, with a number in front, without automatically converting it to numbered list, eg:

### 1. Header title

ivailop7:
Ah, have been meaning to look into this one.

ivailop7:
@acywatson you know the list plugin the most, can you help me on this one?

ivailop7:
This needs a generic fix, allowing for bulleting/unbulleting of code block and quote block as well.

ronaldlangeveld:
Thanks for looking into it @ivailop7 🙏🏻

WoodenPC:
up

acywatson:
I think this is actually an issue with the markdown shortcut plugin, as opposed to the list plugin. If you just don't use the markdown shortcut plugin, this won't happen.

It does probably also make sense to change the markdown plugin somehow to not to the list transformation in certain situations (like when the parent is a heading). I'm not sure if this should be the default or a config option, though.

ivailop7:
Tried it with MarkdownShortcuts plugin disabled, it's still not behaving correctly. I had look at this some time ago. It's an issue for anything that doesn't have a paragraph as a parent, which is why bulleting of links and images works (they are children of a paragraph), while Quote, Heading, Collapsible Section are not.

acywatson:
[...]

PR #5705, opened 2024-03-13
fixed:#5646 Bug: Header becomes a numbered list

muskan2622:
Lexical version: 0.13.1

Steps To Reproduce
Go to the Lexical Playground
Modify the Welcome to the playground header and make it 1. Welcome to the playground
Notice it converts to a numbered list item instead of remaining a header as soon as you make a space between 1. and Welcome
image


ivailop7:
Hi, Is this meant to be a draft PR? I'm not seeing an actual fix for the referenced mention in the title.

ivailop7:
But this is just a change to the test file, not addressing the actual issue, which is keeping the Heading format inside the bullet item.

PR #5707, opened 2024-03-13
Retain format on programatic updates

### LINKED ISSUE #5620: How to keep formatting between lines?

zurfyx:
Follow up on https://github.com/facebook/lexical/pull/5692 

https://github.com/facebook/lexical/assets/193447/b03a04fa-988c-41c7-bcc6-8dac2e5984d0

Closes #5620

Potential follow up https://github.com/facebook/lexical/issues/5708

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -306,6 +306,7 @@ function onSelectionChange(
     if ($isRangeSelection(selection)) {
       const anchor = selection.anchor;
       const anchorNode = anchor.getNode();
+      const focus = selection.focus;
 
       if (selection.isCollapsed()) {
         // Badly interpreted range selection when collapsed - #1482
@@ -326,10 +327,6 @@ function onSelectionChange(
         const [lastFormat, lastStyle, lastOffset, lastKey, timeStamp] =
           collapsedSelectionFormat;
 
-        const root = $getRoot();
-        const isRootTextContentEmpty =
-          editor.isComposing() === false && root.getTextContent() === '';
-
         if (
           currentTimeStamp < timeStamp + 200 &&
           anchor.offset === lastOffset &&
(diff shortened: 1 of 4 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -286,6 +286,8 @@ function onSelectionChange(
     }
   }
 
+  const previousSelection = editor._editorState._selection;
```
zurfyx, line 289: @ivailop7 to use `$getPreviousSelection` I'd have to do `.getEditorState.read`, do you think it's worth it, particularly when this is run at every single selection change?

ivailop7: Fair enough, can skip.

ivailop7 (approved):

ivailop7:
We no longer need this one because of https://github.com/facebook/lexical/pull/5822 , right?

PR #5714, opened 2024-03-14
Send click event in read-only editor

ivailop7:
Right now only a SELECTION_CHANGE_COMMAND is passed when the editor is in read-only mode. Which makes sense, since we don't want to allow copy, paste, etc when contenteditable doesn't have true 'readonly' mode. 

However, I stumbled upon a situation where this logic disabled the 'click' event not firing the CLICK_COMMAND at all when in read-only, which is needed if you want to have any interactivity even in read-only when a non-editing user needs to interact with the widget, eg open a popup on a double click to view something for example. So passing this event through irrespective of editor editability.

Before:

https://github.com/facebook/lexical/assets/7893468/92e832d1-6b61-4a93-a91d-3e866930ceb3

After:

https://github.com/facebook/lexical/assets/7893468/fe741a9c-9ba8-4923-ba75-3a09af6d1781

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -1208,7 +1208,7 @@ export function addRootElementEvents(
               return;
             }
             stopLexicalPropagation(event);
-            if (editor.isEditable()) {
+            if (editor.isEditable() || eventName === 'click') {
               onEvent(event, editor);
             }
           }


zurfyx (approved):
Does this work for mobile? I think we need to define official rules for what editable means. Since this was introduced it feels like everyone has a slightly different impression. You may still remember but initially it was called read-only but given that we still allowed it to be editable via plugins and the like we renamed it to editable 😛

ivailop7:
> Does this work for mobile? I think we need to define official rules for what editable means. Since this was introduced it feels like everyone has a slightly different impression. You may still remember but initially it was called read-only but given that we still allowed it to be editable via plugins and the like we renamed it to editable 😛

This does indeed ring a bell, but I think I was too new to the project at the time and wasn't aware of the context change at the time. Just tried it, it doesn't change anything for mobile behaviour, since I believe it differentiates between a 'mouse click' and a 'pointer click'.

PR #5716, opened 2024-03-15
Support moduleResolution bundler

### LINKED ISSUE #4160: Support for `bundler` `moduleResolution` in TypeScript 5.0
This is a request for updating `lexical` to TS 5.0, and making sure it supports the new `bundler` `moduleResolution`.

After upgrading a project to [TS5](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/), we switched to using the new [`bundler` `moduleResolution` strategy](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler).

I noticed that running `tsc` resulted in the following errors in `lexical` (which is currently on TS `^4.6.4`):

```bash
$ npm run typecheck

> typecheck
> tsc

node_modules/@lexical/list/formatList.d.ts:9:40 - error TS7016: Could not find a declaration file for module './'. '~/exampleProject/node_modules/@lexical/list/LexicalList.js' implicitly has an 'any' type.
9 import { ListItemNode, ListNode } from './';
                                         ~~~~

node_modules/@lexical/list/utils.d.ts:9:40 - error TS7016: Could not find a declaration file for module './'. '~/exampleProject/node_modules/@lexical/list/LexicalList.js' implicitly has an 'any' type.
9 import { ListItemNode, ListNode } from './';
                                         ~~~~

node_modules/@lexical/table/LexicalTableUtils.d.ts:10:50 - error TS7016: Could not find a declaration file for module '.'. '~/exampleProject/node_modules/@lexical/table/LexicalTable.js' implicitly has an 'any' type.
10 import { InsertTableCommandPayloadHeaders } from '.';
                                                    ~~~

node_modules/lexical/LexicalNode.d.ts:10:29 - error TS7016: Could not find a declaration file for module '.'. '~/exampleProject/node_modules/lexical/Lexical.js' implicitly has an 'any' type.
10 import { ElementNode } from '.';
                               ~~~

node_modules/lexical/LexicalNormalization.d.ts:8:47 - error TS7016: Could not find a declaration file for module '.'. '~/exampleProject/node_modules/lexical/Lexical.js' implicitly has an 'any' type.
8 import type { RangeSelection, TextNode } from '.';
                                                ~~~

node_modules/lexical/LexicalSelection.d.ts:13:26 - error TS7016: Could not find a declaration file for module '.'. '~/exampleProject/node_modules/lexical/Lexical.js' implicitly has an 'any' type.
13 import { TextNode } from '.';
                            ~~~

node_modules/lexical/LexicalUtils.d.ts:14:59 - error TS7016: Could not find a declaration file for module '.'. '~/exampleProject/node_modules/lexical/Lexical.js' implicitly has an 'any' type.
14 import { DecoratorNode, ElementNode, LineBreakNode } from '.';
                                                             ~~~

node_modules/lexical/nodes/LexicalElementNode.d.ts:11:26 - error TS7016: Could not find a declaration file for module '../'. '~/exampleProject/node_modules/lexical/Lexical.js' implicitly has an 'any' type.
11 import { TextNode } from '../';
                            ~~~~~

Found 8 errors in 8 files.

Errors  Files
        1  node_modules/@lexical/list/formatList.d.ts:9
        1  node_modules/@lexical/list/utils.d.ts:9
        1  node_modules/@lexical/table/LexicalTableUtils.d.ts:10
        1  node_modules/lexical/LexicalNode.d.ts:10
        1  node_modules/lexical/LexicalNormalization.d.ts:8
        1  node_modules/lexical/LexicalSelection.d.ts:13
        1  node_modules/lexical/LexicalUtils.d.ts:14
        1  node_modules/lexical/nodes/LexicalElementNode.d.ts:11
```

We found that adding `"types": "./index.d.ts",` to the package.json of the libraries used worked as a temporary solution for most errors except one.

This error wasn't fixed by that, and reverting `moduleResolution` to `node` didn't help either. Deleting `/` in the path did:

```bash
node_modules/lexical/nodes/LexicalElementNode.d.ts:11:26 - error TS7016: Could not find a declaration file for module '../'. '~/exampleProject/node_modules/lexical/Lexical.js' implicitly has an 'any' type.
11 import { TextNode } from '../';
```

https://github.com/facebook/lexical/blob/52af9c426ba8e6c8832254b116b6ee32be3cc26d/packages/lexical/src/nodes/LexicalElementNode.ts#L20

Lexical version: 0.9.0

## The current behavior
An error output is thrown by `tsc` when using `lexical` in a TS5 project.

## The expected behavior
All type checks to be successful with no errors.

### LINKED ISSUE #5117: Bug: return types of methods are not resolved by typescript
Lexical version: 0.12.2

## Steps To Reproduce

1. Set `moduleResolution` to something other than "node" ("Bundler" in my case) in tsconfig.json's compilerOptions

Link to code example: [stackblitz](https://stackblitz.com/edit/vitejs-vite-h2jpdu?file=src%2FApp.tsx)

## The current behavior

Some of the methods on lexical objects have return type `any`.

image

I know it is possible to cast it but I believe it is better to get to the root cause of this.

## The expected behavior

Those methods should have proper return type.

image

### LINKED ISSUE #5710: Bug: some types are imported as any
Lexical version: 0.13.1 the latest

## Steps To Reproduce

1. type assertion to `PointType` from `lexical`
2. call getNode() method.
3. typescript fails to parse correct return type, instead it shows `any`

It happens when import the lexical from npm.
It works well on the code base this repo.

## The current behavior
image

## The expected behavior
image

2wheeh:
fixes #5710
fixes #5117 
fixes #4160

This will support correct type inference with `moduleResolution: bundler` option of tsconfig, which is default on creat next app, create vite.

Currently, when importing `lexcial` or `@lexical/...` with `moduleResolution: bundler`, `.d.ts` files resolves `'.'` to the main entry on package.json.
So if some types are imported with `'.'`, this ends up being inferred as `any`.

For example, ElementNode here is imported from `'.'`.

image

```ts
import { LexicalNode } from 'lexical'
const node = {} as LexicalNode;

node.getParent() // this should be inferred as `ElementNode | null` 
```

image

This should have inferred like this. LexicalNode is not imported from `'.'` so it works:

image

With adding entry for types, it can infer correct type with `moduleResolution: bundler`.

image


2wheeh:
I tried to convert lexical's `moduleResolution` to `bundler`(392322a) but somehow it fails to deploy on Vercel.
So i just reverted it. I believe now is enough to make possible to use `bundler` in the project importing lexical.

2wheeh:
There was one exception that uses `'../'` which has to be `'../index'`  (fe2bbac)
https://github.com/facebook/lexical/issues/4160#issuecomment-1478396843

2wheeh:
Converted `moduleResolution` to `bundler`

- updated y-websocket to 1.5.4 (https://github.com/yjs/y-websocket/issues/133#issuecomment-1464868312)

- changed some lines for overriding yjs type.
- https://github.com/facebook/lexical/issues/4160#issuecomment-1478396843: `'../'` seems not a problem on current TS. It infers correct type even without adding `types` entry but when I go to `.d.ts` it seems not resolving the path correctly🤔. so I just changed it to make it sure.

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawModal.tsx
```diff
@@ -235,6 +235,7 @@ export default function ExcalidrawModal({
             excalidrawAPI={excalidrawAPIRefCallback}
             initialData={{
               appState: initialAppState || {isLoading: false},
+              // @ts-expect-error
               elements: initialElements,
```
2wheeh, line 239: I couldn't resolve this one...

2wheeh:
Now, no tsc error with either `node` or `bundler`

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawModal.tsx
```diff
@@ -21,16 +22,14 @@ import {createPortal} from 'react-dom';
 import Button from '../../ui/Button';
 import Modal from '../../ui/Modal';
 
-export type ExcalidrawElementFragment = {
-  isDeleted?: boolean;
-};
+export type InitialElements = ExcalidrawInitialDataState['elements'];
```
2wheeh, line 25: Looks like a part of the typeof Excalidraw props internally acted differently with `node` or `bundler`. Let's just rely on the types from the upstream package instead of declaring it ourselves.

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawModal.tsx
```diff
@@ -21,16 +22,14 @@ import {createPortal} from 'react-dom';
 import Button from '../../ui/Button';
 import Modal from '../../ui/Modal';
 
-export type ExcalidrawElementFragment = {
-  isDeleted?: boolean;
-};
+export type ExcalidrawInitialElements = ExcalidrawInitialDataState['elements'];
```
2wheeh, line 25: Looks like a part of the typeof Excalidraw props internally acted differently with `node` or `bundler`. Let's just rely on the types from the upstream package instead of declaring it ourselves.

2wheeh:
this needs to upgrade react, @types/react too

2wheeh:
rebased and resolved new errors. I guess It's ready unless failing `e2e-collab-windows (18.18.0, firefox)` is due to this change.

packages/lexical-playground/package.json
```diff
@@ -32,7 +32,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-error-boundary": "^3.1.4",
-    "y-websocket": ">=1.3.x",
+    "y-websocket": "^1.5.4",
```
acywatson, line 35: What’s this about?

acywatson:
What do you expect this to break (if anything) and what’s the upgrade path?

2wheeh:
@acywatson 
> what’s the upgrade path?

With `bundler`, 
```jsonc
// jest.config.js

'^@lexical/react/src/(.*)$': '<rootDir>/packages/lexical-react/src/$1',
```
 this fails to resolve the paths somehow, leading unit tests to fail.

And also just aligning the path with the way we do in tsconfig path alias seems more easy to maintain. Different imports for same path were easily allowed. There are some more in unit test codes.

```ts
// i.e. both are used in unit tests

import {useLexicalComposerContext} from '@lexical/react/src/LexicalComposerContext';
```

So I updated path matchers in jest config for unifying imports in unit tests and make it work with `bundler`.
 
 > What do you expect this to break (if anything)

I think  it's fine with both `node` and `bundler` now. but not sure for the future for the other one since we develop with one. maybe we need a strategy to support both appropriately. (new test on CI ? or just supporting bundler covers both ?)

or 

In certain apps using lexical with `bundler`, if they have put `@ts-expect-error` to address type inference failures, it must be removed with a lexical version update.

2wheeh:
Close this since it is incorporating to #5774 .

PR #5718, opened 2024-03-15
chore(@lexical/text): Split into several files for better readability and to simplify further expansion

StyleT:
It's time to do some housekeeping if we want to add more helpers there :)


acywatson:
I don't mind this but let's hold off for a bit on merging until we get this ESM support release working.

acywatson (approved):

StyleT:
@acywatson Hi! Do you think we can merge it now? Or if not - what is the blocking issue?

PR #5719, opened 2024-03-15
docs: Added links to API docs for respective packages

StyleT:
Currently many package dedicated pages are quite useless (ex: https://lexical.dev/docs/packages/lexical-text) and there is no way to get to https://lexical.dev/docs/api/modules/lexical_text that is actually helpful.

This change aims to fix this.

Note that use of the absolute links here is deliberate to make buttons work from Github / NPM / UI as well as https://lexical.dev

How it shall look like after website deployment:

https://github.com/facebook/lexical/assets/1125900/562b7ac6-d10d-4be0-82ab-e59a4de94f39

**_Note that failed tests here aren't related to the changes in this PR_**


acywatson:
tight!

acywatson (approved):

PR #5724, opened 2024-03-17
feat: add a cleanup method to be called during garbage collection for LexicalNode 

yf-yang:
When a LexicalNode uses an external resource, there is no way to release the resource because the lifecycle methods of LexicalNode is incomplete. This PR adds a new method which should be called when a node is removed from the editorState.

### DIFF
--- a/packages/lexical/src/LexicalGC.ts
+++ b/packages/lexical/src/LexicalGC.ts
@@ -112,7 +112,7 @@ export function $garbageCollectDetachedNodes(
   for (const nodeKey of nodeMapDelete) {
     const node = nodeMap.get(nodeKey);
     if (node !== undefined) {
-      node[Symbol.dispose]();
+      node.cleanUp();
     }
     nodeMap.delete(nodeKey);
   }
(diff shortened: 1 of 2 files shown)


yf-yang:
Alright, the ci is already broken due to typescript version issue. Waiting for the maintenance team's opinion.

yf-yang:
MutationListener can also deal with a node's lifecycle. It requires users to externally maintain a data structure to maintain resources and subscribe to the mutations to keep the data structure in sync with Lexical. By design, can Lexical node internally hold those resources? If that is a bad pattern, I'll close the PR.

etrepum:
The MutationListener approach doesn't really require any additional data structures, only the registration of the listener. You could implement this pattern in terms of a MutationListener and a plug-in that implements this protocol.

```typescript
useEffect(() =>
  editor.registerMutationListener(ResourceNode, (nodes, { prevEditorState })) => {
    for (const [key, mutation] of nodes.entries()) {
      if (mutation === "destroyed") {
        $getNodeByKey(key, prevEditorState)?.cleanUp();
      }
    }
  }),
  [editor],
);
```

yf-yang:
I see, that's right

PR #5726, opened 2024-03-17
Fix: don't call importDOM methods more than once

thorn0:
I didn't add this check, only fixed it.


ivailop7 (approved):
That's a nice one! Good catch!

Issue #5233, opened 2023-11-15
Bug: Backspace in Firefox duplicates entire line


AlessioGr:
https://github.com/facebook/lexical/assets/70709113/50e4b047-ff90-4b2e-9e63-1fc9a52477e8

This only happens in Firefox. It does not happen in Chrome. I'm on macOS.

**To reproduce:**
1. Create multi-line paragraph
2. Select last word in the paragraph
3. Press backspace

Original issue (from lexical editor used in payload): https://github.com/payloadcms/payload/issues/3928

patricsgamma:
Found this issue originally in Payload's implementation, now also in Lattice an HR software that is also using lexical editor.
Would love to see some movement on it, so not every customer support is just recommending chrome as default to solve any browser related problem ;)

kalgynirae:
Just in case it's useful, I've found several slightly different methods for reproducing this:

**Common setup for all methods below**
1. Use Firefox
2. Type "This is a test"
3. Press Shift+Enter (to enter a linebreak in the current paragraph)

**Up/Down arrow key method**
1. Position the cursor at the end of the first line
2. Press Down, Up (so that the cursor is once again positioned at the end of the first line, but you got there via an Up arrow keypress)
3. Press Backspace
--> See the text "his is a test" appear after the cursor

**Double-click selection method**
1. Double-click on the word "test" to select it
2. Press Backspace
--> See the text "his is a test" appear after the cursor

**Ctrl+Shift+Right arrow selection method**
1. Position the cursor just before the word "test"
2. Press Ctrl+Shift+Right (macOS: Option+Shift+Right) to select the whole word "test"
--> See the text "his is a test" appear after the cursor

**Ctrl+Delete method**
1. Position the cursor just before the word "test"
2. Press Ctrl+Delete (macOS: Option+Fn+Backspace) to delete the whole word "test"
--> See the text "his is a test" appear after the cursor

The common pattern in all of these methods seems to be that the cursor jumps to the position immediately before the linebreak via some "unusual" method.

agriffis:
I'm confused! The fix for this was merged, then reverted. The bug still reproduces on the playground. Should this issue be reopened?

PR #5734, opened 2024-03-18
fixed: #5233 firefox related issue for backspace deletion  where "focus" offset is of type element instead of text

Sahejkm:
https://github.com/facebook/lexical/assets/163521239/944ae9bf-1eed-49af-94fe-bc37305624c7

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -564,7 +564,6 @@ test.describe('Selection', () => {
     page,
     isPlainText,
   }) => {
-    test.skip(isPlainText);
     await focusEditor(page);
     await page.keyboard.type('This is testing one');
     // enter linebreak element
@@ -577,9 +576,7 @@ test.describe('Selection', () => {
     await pressShiftEnter(page);
     await page.keyboard.press('ArrowLeft');
     await page.keyboard.press('ArrowUp');
-    await page.keyboard.press('ArrowLeft');
-    await page.keyboard.press('ArrowLeft');
-    await page.keyboard.press('ArrowLeft');
+    await moveLeft(page, 3);
     await extendToNextWord(page);
     await page.keyboard.press('Backspace');
     await assertHTML(
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -562,7 +562,6 @@ test.describe('Selection', () => {
 
   test('can delete word at end of line in a multi paragraph with linebreak by backspace', async ({
     page,
-    isPlainText,
   }) => {
     await focusEditor(page);
     await page.keyboard.type('This is testing one');
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -9,6 +9,7 @@
 import {
   deleteBackward,
   deleteForward,
+  extendToNextWord,
   moveLeft,
   moveRight,
   moveToEditorBeginning,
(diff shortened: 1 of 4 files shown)


zurfyx (commented):
Thank you!

First of, the selection is not necessarily incorrect. It is a valid selection. Can we do better? Probably, the result you've proposed seems reasonable to me but I would inclined to do a bit more research, particularly since this is a core function.

My hunch is that the normalization "error" we want to address is that the node next to it is a TextNode and this selection becomes redundant, so by moving to the end of the TextNode this selection is better. This already happens in the opposite case LTR (L108-109) but unfortunately for the other case we cut short.

However, before doing any of this, I would be inclined to dig deeper into the actual root cause. Why is this selection causing an issue with deleteCharacter? Can we look into `deleteCharacter(isBackward: boolean)`?

Finally, I would add an E2E test to cover this, we have E2E tests that automatically run in the Chrome, FF, and Safari so hopefully Firefox on Playwright behaves just like the browser version and we can test against it.

zurfyx:
> Fixes #5233
> 
> @zurfyx what is $normalizeSelection and should it still be experimental?

We can probably make it official, initially the idea was to keep this sort of private but I've found that it is a handy shortcut at times when implementing plugins.

Sahejkm:
> However, before doing any of this, I would be inclined to dig deeper into the actual root cause. Why is this selection causing an issue with deleteCharacter? Can we look into `deleteCharacter(isBackward: boolean)`?

i did look into deleteCharacter method
it calls "this.removeText();" by doing "insertText()"
so other approach is we could keep the change local to that delete function, something like

```
   if(this.anchor.isBefore(this.focus) && this.focus.type === 'element' && $isLineBreakNode(this.focus.getNode().getChildAtIndex(this.focus.offset)) ){
      const offset = this.focus.offset;
      const node = this.focus.getNode(); 
      const nextNode = node.getChildAtIndex(offset - 1);
      if(nextNode != null){
      this.focus.set(
        nextNode.__key,
        nextNode.getTextContentSize(),
        'text',
      )
      }
    }
    this.removeText();
```

This also works as original issue came from line 
> firstNode = firstNode.spliceText(startOffset, delCount, text, true);

the issue is that endOffset and delCount in the method get calculated wrongly given the endOffset is from focusNodeOffset which is of value 3 as its of type "element" instead of type TextNode

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -558,4 +559,44 @@ test.describe('Selection', () => {
       `,
     );
   });
+
+  test('can delete word at end of line in a multi paragraph with linebreak by backspace', async ({
+    page,
+    isPlainText,
+  }) => {
+    test.skip(isPlainText);
```
zurfyx, line 567: It reproes on plain text too `?https://playground.lexical.dev/?isRichText=false`, should we delete this? 
```suggestion
```

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -558,4 +559,44 @@ test.describe('Selection', () => {
...
+    // enter linebreak element
+    await pressShiftEnter(page);
+    await page.keyboard.type('This is testing three');
+    // enter linebreak element
+    await pressShiftEnter(page);
+    await page.keyboard.press('ArrowLeft');
+    await page.keyboard.press('ArrowUp');
+    await page.keyboard.press('ArrowLeft');
+    await page.keyboard.press('ArrowLeft');
+    await page.keyboard.press('ArrowLeft');
```
zurfyx, line 582:
```suggestion
    await moveLeft(page, 3);
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -822,7 +822,7 @@ export class RangeSelection implements BaseSelection {
       }
     }
 
-    if (selectedNodesLength === 1) {
+    if (anchor.key === focus.key && selectedNodesLength === 1) {
```
zurfyx, line 825: Nit, resolving properties at the end of the condition
```suggestion
    if (selectedNodesLength === 1 && anchor.key === focus.key) {
```

zurfyx (approved):
Added some nits but overall LGTM, thank you Sahej!

zurfyx (approved):

Issue #5731, opened 2024-03-18
Bug: failure to build with Webpack due to "Module not found: Error: Default condition should be last one" [regression]


birtles:
Lexical version: 0.14.0 and 0.14.1

## Steps To Reproduce

1. Attempt to build a project that uses @lexical/react with Webpack

## The current behavior

You get a bunch of errors of the form:

```
Module not found: Error: Default condition should be last one
```

## The expected behavior

It should build without error.

## Analysis

This appears to be a regression from #5618 which changed `lexical-react/package.json` adding a bunch of exports of the form:

```json
  "exports": {
    "./LexicalAutoEmbedPlugin": {
      "import": {
        "default": "./LexicalAutoEmbedPlugin.esm.js",
        "types": "./LexicalAutoEmbedPlugin.d.ts"
      },
      "require": "./LexicalAutoEmbedPlugin.js"
    },
```

Webpack wants the `default` there to come last, i.e. after `types`.

I've confirmed locally that just switching the order of those two lines in each instance allows the build to pass again.

I'll make up a PR for this as soon as I get a spare moment (I'm currently on leave so feel free to steal this from me).

ivailop7:
Waiting to see if there are other issues with the version before we publish a 0.14.2

2wheeh:
This is breaking nextjs app as well

<img width="908" alt="image" src="https://github.com/facebook/lexical/assets/40269597/6b30686c-0e0c-4f36-8c87-1d27af074dc9">

PR #5618, opened 2024-02-16
ESM build

etrepum:
This adds a parallel ESM build, in addition to the current CJS build.

Code was added to `npm run update-version` to maintain the package.json metadata for each package (declaring all of the exports & setting `"sideEffect": false`).

The deepest rabbit hole was the prod build, which uses @ampproject/rollup-plugin-closure-compiler. I could not find an obvious issue about it but when it tries to optimize `LexicalClickableLinkPlugin` it ends up transforming `export default function LexicalClickableLinkPlugin(…)` to `export function default(…)` which is an error. I couldn't figure out how to fix it without also fixing the upstream package, but it seemed that terser was more popular for this use case so I gave that a shot and it compiled straight away.

The second deepest rabbit hole was figuring out how to do a fork module for the ESM build. It seems that a lot of the ecosystem is not ready for dynamic imports and top-level await, so it currently uses the pattern of importing both the dev and prod builds and exporting one of them based on `process.env.NODE_ENV`. This appears to work correctly with tree-shaking, at least with `"sideEffect": false` in the package.json files using vite/esbuild.

```typescript
import * as modDev from './Lexical.dev.esm.js';
import * as modProd from './Lexical.prod.esm.js';
const mod = process.env.NODE_ENV === 'development' ? modDev : modProd;
export const $addUpdateTag = mod.$addUpdateTag;
// …
```

Adds an [/esm/](https://lexical-playground-cf4gxxxxt-fbopensource.vercel.app/esm/) endpoint to the playground to demonstrate an ESM native build using static html with an importmap and no bundler at all. 

I did not add these to /examples/ because I have currently vendored release builds of lexical packages so they can use the esm build, but I have demonstrated that it works with SvelteKit and Astro (with some vite configuration):

* https://stackblitz.com/~/github.com/etrepum/lexical-esm-sveltekit-vanilla-js
* https://stackblitz.com/~/github.com/etrepum/lexical-esm-astro-react

The required vite configuration looks like this:

```typescript
import { sveltekit } from '@sveltejs/kit/vite';
import { defineConfig } from 'vitest/config';

export default defineConfig({
	plugins: [sveltekit()],
	test: {
		include: ['src/**/*.{test,spec}.{js,ts}']
	},
	// This ssr config is the part that is lexical-specific
	ssr: {
		noExternal: [/^(lexical|@lexical\/.*)$/]
	}
});
```

Resolves [#1707](https://github.com/facebook/lexical/issues/1707).

PR #5737, opened 2024-03-18
Generate default entry in package.json exports last & use .mjs for esm modules

etrepum:
Follow-up to #5732 hot fix (for the #5731 regression due to #5618) - ensure that metadata will be generated correctly when new modules are added.

Also renames .esm.js files to .mjs to improve compatibility with experimental Node.js flags.

Closes #5741 - I saw this PR which had missed a few things and touched the same code so I went ahead and cherry-picked it and finished it (but moved the programmatic package.json updates to their own commit for easier review).


ivailop7 (dismissed):
LGTM

PR #5741, opened 2024-03-20
fix: use mjs as file extension

jason89521:
execute `NODE_OPTIONS=--experimental-vm-modules jest`

image

only execute `jest`

image

It seems that Node.js would treat files that don't have a .mjs extension as CommonJS.


etrepum:
I incorporated the changes from this into #5737, nice catch!

Issue #5740, opened 2024-03-20
Bug: ComponentPickerPlugin scrolling is buggy - dropdown disappears onScroll


JingkangZhang:
In some edge cases, the component dropdown plugin cannot be scrolled correctly - if you scroll down, the dropdown disappears. Fairly frustrating user experience.

https://github.com/facebook/lexical/assets/30300487/67a496fc-2cb1-4ea8-ba17-74959612bb2b

Lexical version: live on the [playground](https://playground.lexical.dev/) as of now

## Steps To Reproduce

1. input the text as shown in the video
2. type '/' to bring up the component picker and scroll down 

Link to code example:

## The current behavior
see above

## The expected behavior
dropdown shouldn't disappear

JingkangZhang:
From my preliminary investigation, in this unhappy case, isInView [here](https://github.com/facebook/lexical/blob/6e4f072ea194e7d3e631e488ad1ca1be79add50e/packages/lexical-react/src/shared/LexicalMenu.ts#L225) becomes false when the scroll happens and the menu gets closed. 

So two monkey patches came to me, that I tested and confirmed would work:
1. delete the [logic to position the dropdown on top](https://github.com/facebook/lexical/blob/6e4f072ea194e7d3e631e488ad1ca1be79add50e/packages/lexical-react/src/shared/LexicalMenu.ts#L517-L525) - never place it on top. This is not ideal when it's opening up at the very bottom. Well, actually, it just might be that the top px calculation is off - perhaps the calculation is supposed to make sure it doesn't overlap, but it does visually.
<img width="498" alt="Screenshot 2024-03-20 at 3 14 15 AM" src="https://github.com/facebook/lexical/assets/30300487/10d2098d-6c03-4467-9398-b8a65cbe9cbd">
<img width="391" alt="Screenshot 2024-03-20 at 3 14 23 AM" src="https://github.com/facebook/lexical/assets/30300487/a01f7b82-24af-44a9-9f33-39a37a1a4fb8">

2. could be something wrong with this [isTriggerVisibleInNearestScrollContainer function](https://github.com/facebook/lexical/blob/6e4f072ea194e7d3e631e488ad1ca1be79add50e/packages/lexical-react/src/shared/LexicalMenu.ts#L225-L234) - not clear what is not visible here, and then, why close the dropdown at all on scrolling? So getting rid of the logic entirely seems to work for me.

[...]

PR #5743, opened 2024-03-20
Fix check for menu cannot fit on top

JingkangZhang:
issue https://github.com/facebook/lexical/issues/5740 

Originally, the check doesn't consider the resolution height, but the resolution height is being used below to determine the style.top. This causes the element to overlap with the parent box, and in turn causes the menu to close onScroll, because of the check [here](https://github.com/facebook/lexical/blob/6e4f072ea194e7d3e631e488ad1ca1be79add50e/packages/lexical-react/src/shared/LexicalMenu.ts#L225).

Screenshot 2024-03-20 at 3 14 23 AM
Screenshot 2024-03-20 at 3 31 46 AM

Add the height check to the RHS of the > , so the menu won't be placed on the top if it cannot fit.

Taking a step back, I'm not sure why when the menu is placed on top of the cursor, it's `height` length away from the cursor, whereas when the menu is placed below the cursor, it's 0 length away from the cursor. In other words, we could have removed both `+ height` on line 519 and `- height` on line 522, but I didn't implement that to preserve the original style.

## Results

Before:
Screenshot 2024-03-20 at 4 33 28 AM

After:
Screenshot 2024-03-20 at 4 33 46 AM

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -516,7 +516,7 @@ export function useMenuAnchorRef(
         if (
           (top + menuHeight > window.innerHeight ||
             top + menuHeight > rootElementRect.bottom) &&
-          top - rootElementRect.top > menuHeight
+          top - rootElementRect.top > menuHeight + height
         ) {
           containerDiv.style.top = `${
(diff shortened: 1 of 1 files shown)


StyleT (approved):
LGTM! Thanks for contribution

PR #5744, opened 2024-03-20
@lexical/playground: Moved Vite config to TS and removed code duplication

StyleT:
Here we extract module resolution config to `viteModuleResolution.ts` and move both Vite configs from `.js` to `.ts`


ivailop7 (approved):
LGTM, I do think we can have the vite config as .mts (https://vitejs.dev/guide/troubleshooting#config) if we want to move more towards ESM practices.

ivailop7:
Bumping the vite to v5 in package.json has been in my do list for a while. In case you want to give it a go.

ivailop7 (approved):
@StyleT I think this is good to be merged

packages/lexical-playground/vite.prod.config.ts
```diff
@@ -0,0 +1,60 @@
...
+  },
+  define: {
+    'process.env.IS_PREACT': process.env.IS_PREACT,
+  },
+  plugins: [
+    replaceCodePlugin({
+      replacements: [
+        {
+          from: /__DEV__/g,
+          to: 'true',
```
canastro, line 42: should this be false?

etrepum:
This reverted the /esm/ example in the playground from #5618, I'll create another PR to restore it

PR #5747, opened 2024-03-20
feat(@lexical/devtools): Added basic extension scaffolding

StyleT:
Scaffolding Includes:
- Extension with TS/React support
- HMR for extension panels and fast reload for other configs
- Dev mode with browser auto-start with dedicated profile
- Store to reactively sync data cross different extension surfaces: window, content script, background SW, popup
- Messaging API for RPC between extension surfaces: window, content script, background SW, popup
- Chrome and FF support (after https://github.com/serversideup/webext-bridge/pull/83)

**Out of scope:**
- Actual Tree View component
- Publishing pipeline
- Tests

**Demo in Chrome:**

https://github.com/facebook/lexical/assets/1125900/b5f4c272-f2db-4e48-b9db-d94eee15ce36

**Demo in Firefox:**

https://github.com/facebook/lexical/assets/1125900/19974599-d1e0-47e9-9ca3-bc75582221d5

**DevX demo:**

https://github.com/facebook/lexical/assets/1125900/963a611f-53ce-487b-8a58-d8445d3ddffa


packages/lexical-devtools/src/entrypoints/injected/main.ts
```diff
@@ -0,0 +1,57 @@
...
+    tabID,
+    Object.fromEntries(editors.map((e) => [e._key, e.getEditorState()])),
+  );
+
+  editors.forEach((editor) => {
+    if (states[editor._key] !== undefined) {
+      // already registered
+      return;
+    }
+    editor.registerUpdateListener((event) => {
```
fantactuka, line 49: Is there any unmounting handler for the extension so that it could unsubscribe from editors? Similar to what we do with in `useEffect` when adding listeners in plugins

packages/lexical-devtools/src/entrypoints/injected/main.ts
```diff
@@ -0,0 +1,57 @@
+/**
```
fantactuka, line 1: Not a blocker, but there's a lot of main.ts and index.ts files, I wonder if those are updated often. If so - imagine having 5 index.ts files open and trying to find which tab you actually wanted in VS Code :)

packages/lexical-devtools/src/entrypoints/injected/main.ts
```diff
@@ -0,0 +1,57 @@
...
+function scanAndListenForEditors(
+  tabID: number,
+  extensionStore: StoreApi<ExtensionState>,
+) {
+  const {setStatesForTab, lexicalState} = extensionStore.getState();
+  const states = lexicalState[tabID] ?? {};
+
+  // TODO: refresh editors present of the page
+  const lexicalNodes = Array.from(
+    document.querySelectorAll('div[data-lexical-editor]').values(),
```
fantactuka, line 35: I believe `.values()` is not needed and it'll convert into array just fine

fantactuka (dismissed):
Can't provide much expertise in wxt, but end result looking great. Some nits in general, but I'm happy to stamp as is. The only cons so far is unmounting and listeners clean up

packages/lexical-devtools/src/entrypoints/injected/main.ts
StyleT: Yeah, but it's dictated by WXT framework: https://wxt.dev/entrypoints/background.html#filenames

I don't think that at the end of this project those files will contain a lot of code. They shall only bootstrap respective entry point and pass control over to other files that contain actual business logic

packages/lexical-devtools/src/entrypoints/injected/main.ts
StyleT: This is a great point! 

As far as I understand this there is no need to do so because:
- "unmount" for injected script happens when user closes this browser window or navigates to a different URI (and we don't care about this case as all JS stops execution there)
- When DOM node with corresponding editor is being deleted - all event listeners shall be cleaned up with it. There is no way to watch for it except for MutationObserver

So unless Lexical has any hook for destruction you can watch for (I believe there is none) - we shall be good with this approach

packages/lexical-devtools/src/entrypoints/injected/main.ts
StyleT: Nice catch! I'll accommodate this in the next PR

package.json
```diff
@@ -27,6 +27,7 @@
     "extract-codes": "node scripts/build.js --codes",
     "flow": "node ./scripts/check-flow-types.js",
     "tsc": "tsc",
+    "tsc-extension": "cd ./packages/lexical-devtools && npm run compile",
```
acywatson, line 30: Does this work? The script is in a different package JSON. I'm rusty on workspaces tho

StyleT: It's basic linux cd (change current dir) :P 
It will change dir to the `./packages/lexical-devtools` and execute `npm run compile` from the child package
And current dir will be "rolled back" as this command ends execution

packages/lexical-devtools/src/entrypoints/injected/main.ts
fantactuka: Can try doing it when scanning page for editors (and call unmount() those are no longer available). E.g., if you're using extension on fb home feed it'll unmount comment boxes as you scroll and you don't want to keep it all in memory

package.json
acywatson: Lol I didn't even see the cd command

StyleT: I'll actually change it to `npm run compile -w @lexical/devtools` in the follow up PR

packages/lexical-devtools/src/store-sync/internal/pages.ts
```diff
@@ -0,0 +1,67 @@
...
+  // wait to be ready
+  await proxyStore.ready();
+  // initial state
+  store.setState(proxyStore.getState());
+
+  const callback = (state: T, oldState: T) => {
+    (proxyStore.dispatch({state}) as unknown as Promise<T>)
+      .then((syncedState) => {
+        if (syncedState) {
+          // success
```
acywatson, line 28: I guess you can invert this condition, move the logic up and then remove the else block.

packages/lexical-devtools/wxt.config.ts
```diff
@@ -0,0 +1,72 @@
...
+      // duplicate installations of DevTools).
+      // By specifying a key in the built manifest.json file,
+      // we can make it so the generated extension ID is stable.
+      // For more details see the docs here: https://developer.chrome.com/docs/extensions/reference/manifest/key
+      // Generated via:
+      // $ openssl genrsa 2048 | openssl pkcs8 -topk8 -nocrypt -out key.pem # private key
+      // $ openssl rsa -in key.pem -pubout -outform DER | openssl base64 -A # this key below (strip % at the end)
+      // $ openssl rsa -in key.pem -pubout -outform DER | shasum -a 256 | head -c32 | tr 0-9a-f a-p # extension ID
+      // @ts-expect-error https://github.com/wxt-dev/wxt/issues/521#issuecomment-1978147707
+      manifestConf.key =
```
acywatson, line 46: Where's the PK stored?

acywatson (approved):
nice

packages/lexical-devtools/wxt.config.ts
StyleT: Nowhere as it's not needed for dev purposes. We'll use GitHub Secrets for publishing

packages/lexical-devtools/src/entrypoints/injected/main.ts
```diff
@@ -0,0 +1,57 @@
...
+  ) as LexicalHTMLElement[];
```
etrepum, line 36: Might be worth a filter here (maybe with a type guard that validates everything at runtime instead of casting) since a SSR-rendered editor (and possibly other scenarios) could leave a residual data prop but no attached LexicalEditor?

StyleT: Updated!

packages/lexical-devtools/src/store-sync/internal/pages.ts
StyleT: Updated!

packages/lexical-devtools/src/entrypoints/injected/main.ts
StyleT: > Can try doing it when scanning page for editors (and call unmount() those are no longer available). E.g., if you're using extension on fb home feed it'll unmount comment boxes as you scroll and you don't want to keep it all in memory

Will try this, but it shall be fine. I was checking for it by manually removing DOM nodes via devtools

acywatson (approved):

Issue #2632, opened 2022-07-12
Bug: `@lexical/markdown` `text-format` combination


hanford:
When hydrating a lexical editor with the following markdown `**Bold *Italic***` calling `$convertFromMarkdownString` should properly format this as bold & and bold italic text. However, for some reason, it doesn't, and the rendered result is **`Bold *Italic*`**

This seems to work when switching between editor content and markdown in the playground, but it does **not** work when initializing the playground with this content.

Lexical version: 3.8

## Steps To Reproduce

1. Have a lexical instance with `initialEditorState` that invokes `$convertFromMarkdownString('**Bold *Italic***')`
2. See the content be rendered in editor as **`Bold *Italic*`** rather than the expected **`Bold` *`Italic`***

Link to code example:

https://codesandbox.io/s/many-lexical-autofocus-forked-efq8sq?file=/src/App.js

## The current behavior
Content is Bold and asterisks are rendered in editor

## The expected behavior
Content should be Bold and Italic text should be italicized

--

Happy to share any more information if it's useful

hanford:
Narrowed the issue down to a bug in [this function](https://github.com/facebook/lexical/blob/main/packages/lexical-markdown/src/v2/MarkdownImport.ts#L201), where when splitText is called, it can break up markdown format syntax so it's not formatted...

i.e. the following markdown syntax will never be imported correctly:

```
*Foo **Bar***
**Foo *Bar***
_Foo __bar___
__Foo _bar___
``` 

When `findOutermostMatch` is called on the above pieces of markdown, it wrongly identifies the outer most pieces of markdown, causing the other contents to then be "split" and subsequent formatting is never formatted correctly

In the first case, ```*Foo **Bar***```, `findOutermostMatch` wrongly returns `**Bar**`, then `splitText` is called, and the `leadingNode` and `remainingNode` are `*Foo` and `*` respectively, causing lexical to then render

<img width="645" alt="image" src="https://user-images.githubusercontent.com/2148168/181378254-51dfc953-426d-4549-b999-9eea1a53826c.png">

fantactuka:
@hanford Thanks for digging into it! Was suspecting findOutermostMatch is not working correctly but didn't get a chance to play with regExp. The question is if it's even possible with regExp only. Especially for cases like above when `***` is a closing tag for separate `*` and `**`. So the logic would be if we have `*` and `**` opened, then the next `***` is their combined closing tag, and we don't need to look for any `*` or `**` further. 

[...]

PR #5758, opened 2024-03-23
[lexical-markdown] Bug Fix: preserve the order of markdown tags for markdown combinations, and close the tags when the outmost tag is closed

yhw5:
Issue: https://github.com/facebook/lexical/issues/2303, #2632
Related PR: https://github.com/facebook/lexical/pull/2304

**This PR aims to fix the following:**
Closing tags were applied in a different order than opening tags.
Unclosed tags remain in the markdown for complex markdown combinations.

And this change should not add any unnecessary duplicate markdown tags

Examples:
Note: underline is a custom markdown here, the tag is "++". You can imagine it as strikethrough "~~"
1. bold, italic and underline combination + italic
image
markdown:
before: `++***BoldplusItalicplusUnderline**++Italic*` (unclosed italic inside the first block)
after: `***++BoldplusItalicplusUnderline++****Italic*` (properly closed all the opening tags if the outmost one is closed)

2. bold + underline and bold combination
image
markdown output:
before: `**bold++underline**++` (markdown tag order is wrong)
after: `**bold++underline++**`


yhw5:
Hi @fantactuka I noticed that you have mainly been involved in the referenced issues and PRs above. Could you please take a look at this proposed change? If this could be merged, that would be really great! Thanks

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -141,7 +141,11 @@ describe('Markdown', () => {
     },
     {
       html: '<p><span style="white-space: pre-wrap;">Hello </span><s><i><b><strong style="white-space: pre-wrap;">world</strong></b></i></s><span style="white-space: pre-wrap;">!</span></p>',
-      md: 'Hello ~~***world***~~!',
+      md: 'Hello ***~~world~~***!',
```
yhw5, line 144: I changed the expected markdown here because I rearranged the order of tags in the markdownExport. The order change was to address edge cases, such as the one from L146-148, where the very first opening tag needs to be closed as the last one.

yhw5:
The unit tests and integrity should be OK now. Please help review, thanks

yhw5:
@ivailop7 @fantactuka @acywatson @zurfyx @Fetz May I get review for this one? Appreciate your time 🙏

yhw5:
QVQ any concerns regarding the fix? Really like to get the issue fixed especially for the current broken markdown combinations

SaintPepsi:
Does this also fix cases where multiple formats are being added/removed?
Github doesn't support it natively (first line is adding bold, then italic, and then removing bold from the word "that":
```
**Example _Text **that** is wrapped in_ bold**
**Example _Text _**that**_ is wrapped in_ bold**
**Example _Text_** _that_ **_is wrapped in_ bold**
```
**Example _Text **that** is wrapped in_ bold**
**Example _Text _**that**_ is wrapped in_ bold**
**Example _Text_** _that_ **_is wrapped in_ bold**

yhw5:
> Does this also fix cases where multiple formats are being added/removed? Github doesn't support it natively (first line is adding bold, then italic, and then removing bold from the word "that":
> 
> ```
> **Example _Text **that** is wrapped in_ bold**
> **Example _Text _**that**_ is wrapped in_ bold**
> **Example _Text_** _that_ **_is wrapped in_ bold**
> ```
> 
> **Example _Text **that** is wrapped in_ bold** **Example _Text _**that**_ is wrapped in_ bold** **Example _Text_** _that_ **_is wrapped in_ bold**

yes exactly!

etrepum (commented):
I think this looks good, not sure how it fell off the radar! Will approve and merge if the test suite is still in a good state.

PR #5762, opened 2024-03-25
fix: Double input of text including composing, such as Japanese, at TextNode Boundaries with in a MarkNode #5761

### LINKED ISSUE #5761: Bug: Double input of text including composing, such as Japanese, at TextNode Boundaries with in a MarkNode
Lexical version: v0.14.2

## Steps To Reproduce

1. Create CommentNode
2. Inside the LinkNode, create TextNode that is separated into at least two Nodes. (e.g., by partially formatting text)
3. Input texts with composition between boundary of TextNodes

https://github.com/facebook/lexical/assets/111737064/8abc5dc5-d538-4be6-a091-4610856768f4

Link to code example:

## The current behavior
Under certain conditions, text that includes composing (such as Japanese) is inputted twice.

This happens inside an ElementNode, like a MarkNode or LinkNode, where `canInsertTextAfter` or `canInsertTextBefore` is set to false. If a part of the internal TextNode is formatted to split it, and text that includes composing is inputted at the boundary, the composition is released, and the text is inputted twice.

## The expected behavior

The text should not be inputted twice at the boundary, and composing should not be released.

matsuyama-k1:
fixed #5761

## After
https://github.com/facebook/lexical/assets/111737064/4d6f94da-a90a-4810-b79f-7e463e8101cb

## Bug
This happens inside an ElementNode, like MarkNode or LinkNode, where `canInsertTextAfter` or `canInsertTextBefore` is set to false. If a part of the internal TextNode is formatted to split it, and text that includes composing is inputted at the boundary, the composition is released, and the text is inputted twice.

## Cause of the bug
The bug arise when the `CONTROLLED_TEXT_INSERTION_COMMAND` is unnecessarily triggered, even when `isComposition` is true at the node boundary.
Despite the fact that text insertion into the Node is already being behind elsewhere during composing, the `selection.insertText` function called by the command cause a repeated input.
This leads to the text being doubly entered.

## Solution
To resolve this, I modified the behavior so that `CONTROLLED_TEXT_INSERTION_COMMAND` does not get triggered while composing at the boundary.
`$shouldInsertTextAfterOrBeforeTextNode` function now returns false during composing because it was previously causing the command to dispatch at boundary inputs.

Since text input during composing is a provisional state, the default behavior is sufficient.
The strategy now ensure that the behavior at the boundary is confirmed only after composing has concluded.

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -713,13 +713,15 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
   if (offset === 0) {
     return (
       !node.canInsertTextBefore() ||
-      !parent.canInsertTextBefore() ||
+      (!parent.canInsertTextBefore() && !node.isComposing()) ||
       isToken ||
       $previousSiblingDoesNotAcceptText(node)
     );
   } else if (offset === node.getTextContentSize()) {
     return (
-      !node.canInsertTextAfter() || !parent.canInsertTextAfter() || isToken
+      !node.canInsertTextAfter() ||
+      (!parent.canInsertTextAfter() && !node.isComposing()) ||
+      isToken
     );
   } else {
     return false;


ivailop7 (approved):

ivailop7:
Could we add a test for this please, there are quite a few tests for IME and hashtags, which you can use to build one for this use case.

matsuyama-k1:
@ivailop7 
Ok!
I want to know where I write to write test code for IME.

If we were to name a test for this case, it would be 'Testing IME input on the boundary between two TextNodes contained within an ElementNode where canInsertTextAfter or canInsertBefore is false'. 
Thus, it might be appropriate to write this as a test for MarkNode and LinkNode that meet the above conditions, or as a test for core parts like LexicalSelection.

ivailop7:
> @ivailop7 
> 
> Ok!
> 
> I want to know where I write to write test code for IME.
> 
> 
> 
> If we were to name a test for this case, it would be 'Testing IME input on the boundary between two TextNodes contained within an ElementNode where canInsertTextAfter or canInsertBefore is false'. 
> 
> Thus, it might be appropriate to write this as a test for MarkNode and LinkNode that meet the above conditions, or as a test for core parts like LexicalSelection.

Sure, a test under the ime Composition file should be better

matsuyama-k1:
Is this a test file you mention?
https://github.com/facebook/lexical/blob/111117d0013917d0481cf418430db75a8afbb497/packages/lexical-playground/__tests__/e2e/Composition.spec.mjs#L184

ivailop7:
> @ivailop7 
> 
> Is this a test file you mention?
> 
> https://github.com/facebook/lexical/blob/111117d0013917d0481cf418430db75a8afbb497/packages/lexical-playground/__tests__/e2e/Composition.spec.mjs#L184

Yes

PR #5765, opened 2024-03-26
Fix clear formatting in table

AlexanderReznik:
Before:
table clear formatting before

After:
table clear formatting after


ivailop7 (approved):
Thank you, Alex!

PR #5766, opened 2024-03-26
Fix selection started from inside of the table

AlexanderReznik:
Making it possible to select content beyond the table when starting the selection from inside of the table. In this case the whole table becomes selected similar to the case when selection starts from outside of the table and goes inside. Table selection is replaced with the Range selection
Before:
table selection before
After:
table selection after

Allowing to drag a selection from one table to the other without crashing
Fixes the following error:
2 tables crash


ivailop7:
Thank you, Alex! @zurfyx @acywatson @fantactuka can you give this one a look?

zurfyx (approved):

zurfyx:
LGTM!

PR #5767, opened 2024-03-26
Improve table selection and navigation

### LINKED ISSUE #5627: Selection breaks when table is the last node
If we add a table and it's the **last node**, it creates 2 issues
1. When you press 'Up' to go back to the last table cell, gets stuck and doesn't move, even though the event gets fired (Video 1)
2. When you have the cursor right after the table on the same line, to press Enter to add a paragraph below, it goes back to the last table cell and adds a new paragraph inside that cell, rather than under the table. (Video 2)
I do believe these 2 are connected to the same selection issue, thus raising in a single issue rather than 2.

https://github.com/facebook/lexical/assets/7893468/5263ef66-99b5-4e18-a4d1-d2fe006d7bc5

https://github.com/facebook/lexical/assets/7893468/4ac5aef5-a591-4bfc-9f31-c5a1278f9af3

serey-roth:
This PR addresses several issues related to table selection and navigation within the editor. The following changes have been made:

- Pressing the "arrow left" key to exit the first cell moves selection directly to the previous sibling node, if exists. Before, the selection is placed after the table and then at the new node. (Video 1)
- Pressing the "arrow right" key to exit the last cell moves selection directly to the next sibling node, if exists. Before, the selection is placed before the table and then at the new node. (Video 2)
- Pressing the"arrow left" key at the start of a paragraph under the table moves selection to the last cell of the table. (Fix #5627) (Video 3)
- Pressing the "Enter" key, when the cursor is before the table on the same line, moves selection to a new preceding paragraph. (Video 4)
- Pressing the "Enter" key, when the cursor is after the table on the same line, moves selection to the next paragraph (if exists), or otherwise,  a new paragraph. (Fix #5627) (Video 5)

https://www.loom.com/share/f16a323757294da7b767ba870fa18421?sid=e7ee42d6-2e59-4ffb-87db-00d8fed838d8

https://www.loom.com/share/b461840c78c443d4816cbb94cad938f7?sid=aa3406b5-b50b-4071-b1ef-b235e57e9154

https://www.loom.com/share/107d1bcac82f497085231858235fe3f3?sid=8dfc4d27-68af-465c-87e6-1e274e48e2bb

https://www.loom.com/share/454f7f63c2b34a3a838c3b0e6ccc6ca4?sid=0b00f765-360c-438e-a8b0-0b47f78ce4d9

https://www.loom.com/share/fed0547e1e894953811a70648c0c3751?sid=b75d789b-db7d-4ddf-a62f-331488bcdb0c

Updated 03/27/2024:
- Can now enter a nested table from the start of a paragraph in a table cell via "arrow left" key. (Video 6)
- Can now exit a nested table into the parent table cell via horizontal arrow keys. (Video 6)
- If a table is the last node and not a nested table, and the cursor is on the same line as the table, typing will insert the text in a new paragraph following the new table. (Video 7)

https://www.loom.com/share/9eedbba40dae4dc4b57090bf9404d0c9?sid=784bd357-fdc9-4206-895e-365714c65680

https://www.loom.com/share/e179a5c87235409fa8d1945636e5ba03?sid=009ab1ed-70b2-47d0-8cf4-891ccf26e62e

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -1160,12 +1160,6 @@ function $handleArrowKey(
   }
 
   if ($isRangeSelection(selection) && selection.isCollapsed()) {
-    // Horizontal move between cels seem to work well without interruption
-    // so just exit early, and handle vertical moves
-    if (direction === 'backward' || direction === 'forward') {
-      return false;
-    }
-
     const {anchor, focus} = selection;
     const anchorCellNode = $findMatchingParent(
       anchor.getNode(),
@@ -1196,6 +1190,48 @@ function $handleArrowKey(
           tableObserver,
         );
       }
+    }
+
+    if (direction === 'backward' || direction === 'forward') {
+      // Hitting horizontal arrow keys on an empty first or last cell should move the selection out of the table.
+      if (anchor.type !== 'element') {
+        return false;
+      }
+
+      const [tableMap, cellValue] = $computeTableMap(
+        tableNode,
+        anchorCellNode,
+        anchorCellNode,
+      );
+      const firstCell = tableMap[0][0];
+      const lastCell = tableMap[tableMap.length - 1][tableMap[0].length - 1];
+      const {startColumn, startRow} = cellValue;
+
+      const isExiting =
+        direction === 'backward'
+          ? startColumn === firstCell.startColumn &&
+            startRow === firstCell.startRow
+          : startColumn === lastCell.startColumn &&
+            startRow === lastCell.startRow;
+      if (!isExiting) {
+        return false;
+      }
+
+      const nextNode =
+        direction === 'backward'
+          ? tableNode.getPreviousSibling()
+          : tableNode.getNextSibling();
+      if (!nextNode) {
+        return false;
+      }
+
+      stopEvent(event);
+      if (direction === 'backward') {
+        nextNode.selectEnd();
+      } else {
+        nextNode.selectStart();
+      }
+      return true;
     }
 
     const anchorCellDom = editor.getElementByKey(anchorCellNode.__key);
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -28,7 +28,6 @@ import {
   $getSelection,
   $isElementNode,
   $isRangeSelection,
-  $isRootNode,
   $isTextNode,
   $setSelection,
   COMMAND_PRIORITY_CRITICAL,
@@ -1243,39 +1242,31 @@ function $handleArrowKey(
       $isRangeSelection(selection) &&
       selection.isCollapsed()
     ) {
-      // Hitting arrow left at the start of a paragraph right under a table should move the selection
-      // to the last cell of the table.
       const anchorType = selection.anchor.type;
+      const anchorOffset = selection.anchor.offset;
       if (
         anchorType !== 'element' &&
-        !(anchorType === 'text' && selection.anchor.offset !== 0)
+        !(anchorType === 'text' && anchorOffset === 0)
       ) {
         return false;
       }
-
       const anchorNode = selection.anchor.getNode();
       if (!anchorNode) {
         return false;
       }
-
-      const topLevelNode = $findMatchingParent(anchorNode, (node) => {
-        return (
-          $isElementNode(node) &&
-          !!node.getParent() &&
-          $isRootNode(node.getParent())
-        );
-      });
-      if (!topLevelNode) {
+      const parentNode = $findMatchingParent(
+        anchorNode,
+        (n) => $isElementNode(n) && !n.isInline(),
+      );
+      if (!parentNode) {
         return false;
       }
-
-      const previousNode = topLevelNode.getPreviousSibling();
-      if (!previousNode || !$isTableNode(previousNode)) {
+      const siblingNode = parentNode.getPreviousSibling();
+      if (!siblingNode || !$isTableNode(siblingNode)) {
         return false;
       }
-
       stopEvent(event);
-      tableNode.selectEnd();
+      siblingNode.selectEnd();
       return true;
     }
     return false;
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -27,6 +27,7 @@ import type {
 import {$findMatchingParent} from '@lexical/utils';
 import {
   $createParagraphNode,
+  $createTextNode,
   $getNearestNodeFromDOMNode,
   $getPreviousSelection,
   $getSelection,
@@ -451,6 +452,27 @@ export function applyTableHandlers(
           if (!$isTableCellNode(tableCellNode)) {
             return false;
           }
+
+          if (typeof payload === 'string') {
+            const edgePosition = $getTableEdgeCursorPosition(
+              editor,
+              selection,
+              tableNode,
+            );
+            if (!edgePosition) {
+              return false;
+            }
+            // TODO: Add support for nested tables
+            let paragraphNode;
+            if (edgePosition === 'first') {
+              paragraphNode = $insertParagraphAtTableEdge(tableNode, 'before');
+            } else {
+              paragraphNode = $insertParagraphAtTableEdge(tableNode, 'after');
+            }
+            paragraphNode.append($createTextNode(payload));
+            paragraphNode.selectEnd();
+            return true;
+          }
         }
 
         return false;
@@ -755,21 +777,6 @@ export function applyTableHandlers(
     editor.registerCommand(
       INSERT_PARAGRAPH_COMMAND,
       () => {
-        const nativeSelection = window.getSelection();
-        if (nativeSelection === null) {
-          return false;
-        }
-        const nativeAnchorNode = nativeSelection.anchorNode;
-        if (nativeSelection.anchorNode === null) {
-          return false;
-        }
-        if (nativeAnchorNode !== editor.getRootElement()) {
-          return false;
-        }
-
-        // When the cursor is right after the table on the same line,
-        // there's no following node, the internal selection is within the last table cell
-        // while the native selection is at the root.
         const selection = $getSelection();
         if (
           !$isRangeSelection(selection) ||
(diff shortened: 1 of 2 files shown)


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1198,6 +1314,48 @@ function $handleArrowKey(
       }
     }
 
+    if (direction === 'backward' || direction === 'forward') {
```
ivailop7, line 1317: I think this if statement becomes redundant, since it's always true? Unless there's another value for direction I'm unaware of.

serey-roth:
> This looks VERY VERY good, there's still a small bug related to what you are fixing, if you put your cursor on the line after the table, when you press Enter, it goes to the next row correctly now, but if you type 'abc' it doesn't go to the line below, but to the last cell in the table and puts the 'abc' in there.

This happens because the selection is still in the last cell of the table and the table is the last node. In this scenario, should we create a new paragraph and insert the text there? In my opinion, this is the most sensible solution since the cursor is technically outside. @ivailop7 What do you think?

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
serey-roth: The `$handleArrowKey` is called for every arrow key event.

ivailop7:
> > This looks VERY VERY good, there's still a small bug related to what you are fixing, if you put your cursor on the line after the table, when you press Enter, it goes to the next row correctly now, but if you type 'abc' it doesn't go to the line below, but to the last cell in the table and puts the 'abc' in there.
> 
> This happens because the selection is still in the last cell of the table and the table is the last node. In this scenario, should we create a new paragraph and insert the text there? I think this is the most sensible solution since the cursor is technically outside. @ivailop7 What do you think?

Yes, I think this is the most sensible approach. Looking at how other editors behaves for this, they don't even allow the cursor to be on the same line as the table, but move it to a paragraph below the table directly when you click after the row after the table. So let's go with the same. Add a paragraph below and type in there.

serey-roth:
> > > This looks VERY VERY good, there's still a small bug related to what you are fixing, if you put your cursor on the line after the table, when you press Enter, it goes to the next row correctly now, but if you type 'abc' it doesn't go to the line below, but to the last cell in the table and puts the 'abc' in there.
> > 
> > 
> > This happens because the selection is still in the last cell of the table and the table is the last node. In this scenario, should we create a new paragraph and insert the text there? I think this is the most sensible solution since the cursor is technically outside. @ivailop7 What do you think?
> 
> Yes, I think this is the most sensible approach. Looking at how other editors behave for this, they don't even allow the cursor to be on the same line as the table, but move it to a paragraph below the table directly when you click after the row after the table. So let's go with the same. Add a paragraph below and type in there.

@ivailop7 Currently, the cursor will be on the same line as the table if you click on the table's right edge. Do we want to remove this as well? In that case, clicking outside the table, regardless of the click position, will move the selection to the next line.

ivailop7:
Let's keep it for now next to the table and upon typing to append the paragraph below and have the text there. If we need to change it later, should be easy.

serey-roth:
> Let's keep it for now next to the table and upon typing to append the paragraph below and have the text there. If we need to change it later, should be easy.

@ivailop7 I've added this, but it only works when the table doesn't have a nested table as the last child in the last cell. I've noticed it's because of the selection; the cursor will still be on the last cell of the nested table when it appears to be on the same line as the outer table (see video below).

https://www.loom.com/share/3c655aead08e4286a7d3f6646a3c2498?sid=aadcfb4e-897f-4c8c-9d83-0d4c5e6276df

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
ivailop7: Got it, thank you!

ivailop7:
> > Let's keep it for now next to the table and upon typing to append the paragraph below and have the text there. If we need to change it later, should be easy.
> 
> @ivailop7 I've added this, but it only works when the table doesn't have a nested table as the last child in the last cell. I've noticed it's because of the selection; the cursor will still be on the last cell of the nested table when it appears to be on the same line as the outer table (see video below). The same thing happens when pressing the "Enter" key.
> 
> https://www.loom.com/share/3c655aead08e4286a7d3f6646a3c2498?sid=aadcfb4e-897f-4c8c-9d83-0d4c5e6276df

The nested tables strike again... to be honest I'm fine with this limitation, since it's still miles better than what we had before. Random speculation, but maybe a check if the parent node is TableCellNode could be a way to become aware of the nested tables and behave accordingly. Either way, happy with the direction!

ivailop7:
Once you are happy with it, would you mind adding tests :)

serey-roth:
> Once you are happy with it, would you mind adding tests :)

Will do! As you noticed, I didn't include the ability to insert a new paragraph by clicking under the last row of the table. I'll tackle this on later, but I'm thinking it should only work for the outermost table? @ivailop7 What do you think?

ivailop7:
Doing this after this PR should be fine! 👍🏻

ivailop7:
@serey-roth one of your own tests is failing, once ready, will get the others to have a look at the PR too.

serey-roth:
> @serey-roth one of your own tests is failing, once ready, will get the others to have a look at the PR too.

The tests are failing in collab mode. I've noticed that no paragraphs are inserted along with the table, which ends up changing the selection path. @ivailop7  Is this the correct behavior?

ivailop7:
> > @serey-roth one of your own tests is failing, once ready, will get the others to have a look at the PR too.
> 
> The tests are failing in collab mode. I've noticed that no paragraphs are inserted along with the table, which ends up changing the selection path. @ivailop7 Is this the correct behavior?

Not sure I get this one. Is this an issue that happens only when in collab, but fine in standalone? When you say no paragraph are inserted with the table, do you mean, a paragraph after the table upon 'table insert' or something else?

serey-roth:
> > > @serey-roth one of your own tests is failing, once ready, will get the others to have a look at the PR too.
> > 
> > 
> > The tests are failing in collab mode. I've noticed that no paragraphs are inserted along with the table, which ends up changing the selection path. @ivailop7 Is this the correct behavior?
> 
> Not sure I get this one. Is this an issue that happens only when in collab, but fine in standalone? When you say no paragraph are inserted with the table, do you mean, a paragraph after the table upon 'table insert' or something else?

It only happens in collab, and only when you insert the table via the dropdown after the initial load. Currently, it looks like we don't automatically insert a paragraph after editor loads. This causes the selection to be null. (See video below)

https://www.loom.com/share/6990dd65bb4c4f83876444ebf23e7f8e

https://www.loom.com/share/bf61832280834c8799e63887c8231573?sid=f7ffca54-d2da-4b29-afa2-ec5a3d55b404

ivailop7:
> > > > @serey-roth one of your own tests is failing, once ready, will get the others to have a look at the PR too.
> > > 
> > > 
> > > The tests are failing in collab mode. I've noticed that no paragraphs are inserted along with the table, which ends up changing the selection path. @ivailop7 Is this the correct behavior?
> > 
> > 
> > Not sure I get this one. Is this an issue that happens only when in collab, but fine in standalone? When you say no paragraph are inserted with the table, do you mean, a paragraph after the table upon 'table insert' or something else?
> 
> @ivailop7 It only happens in collab, and only when you insert the table via the dropdown after the initial load. Currently, it looks like we don't automatically insert a paragraph after editor loads. This causes the selection to be null. (See videos below)
> 
> https://www.loom.com/share/6990dd65bb4c4f83876444ebf23e7f8e
> 
> https://www.loom.com/share/bf61832280834c8799e63887c8231573?sid=f7ffca54-d2da-4b29-afa2-ec5a3d55b404

This is interesting, didn't realize the difference in behaviour that the default paragraph is not insert when in collab, which makes sense. I do still find the failures strange, since other test would be affected. In this case, try just insert a paragraph after focusing the editor so that you never have an editor without a paragraph node, if you don't get them to work do a `test.skip(isPlainText || isCollab);` as a last resort. Nonetheless, great work here!

ivailop7:
Also, don't forget to sort the imports for integrity to pass as well

ivailop7:
@serey-roth I pushed a commit to your branch to help with the tests, let me know when you want to do more changes or you are done with this PR, so I can give this a proper read.

serey-roth:
> @serey-roth I pushed a commit to your branch to help with the tests, let me know when you want to do more changes or you are done with this PR, so I can give this a proper read.

@ivailop7 Thanks for taking care of the failing tests! I rewrote some of them last night, but it looks like I missed a few. I'm happy with the current changes. Please let me know if there's anything to adjust.

ivailop7:
> > @serey-roth I pushed a commit to your branch to help with the tests, let me know when you want to do more changes or you are done with this PR, so I can give this a proper read.
> 
> @ivailop7 Thanks for taking care of the failing checks! I rewrote all my tests last night, but it looks like they weren't exactly correct. I'm happy with the current changes so please review them and let me know if there are anything I need to adjust.
> 
> Out of curiosity, how did you handle the missing paragraph in the collab editor?

Nothing really, removed all the extra code, I found it suspicious in a first place, since all other table tests should fail by the same logic and they didn't. I only added 'undefined' in the assertHTML methods, to not compare the left editor state with the right editor state, as it was done for the other table tests and things worked. I'll start reviewing now.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -806,6 +833,40 @@ export function applyTableHandlers(
...
+          return false;
+        }
+        // TODO: Add support for nested tables
+        let paragraphNode;
+        if (edgePosition === 'first') {
+          paragraphNode = $insertParagraphAtTableEdge(tableNode, 'before');
+        } else {
+          paragraphNode = $insertParagraphAtTableEdge(tableNode, 'after');
+        }
+        paragraphNode.selectEnd();
```
ivailop7, line 863: This chunk of code is the same as in the block under 'if (typeof payload === 'string')' above, worth exporting to a reusable method

ivailop7 (changes requested):
Only the one method duplication comment, the rest is LGTM to me. @zurfyx can I get a second pair of eyes as well.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
serey-roth: All done!

ivailop7 (approved):
LGTM

PR #5772, opened 2024-03-27
Respect CSS zoom

ebads67:
Css `zoom` property affects the mouse position in `event`s. However, functions like `getBoundingClientRect` do not care about the css `zoom`. Hence, the calculations that use mouse positions and `getBoundingClientRect` return wrong values. In this PR those calculations are fixed to return the correct value when css `zoom` property is used.

Css `zoom` has only been implemented in Nightly version of Firefox. So the fix does not apply to Firefox.

In all the examples below the css `zoom` property is set to 2 on the body of the document..

Issue with checklist interactions:
checklist-zoom
Issue with drag handle:
drag-handle-zoom
Issue with context menu:
context-menu-zoom

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -530,12 +530,11 @@ export function $insertFirst(parent: ElementNode, node: LexicalNode): void {
  * css zoom property.
  * @param element
  */
-export function calculateZoomLevel(element: Element): number {
+export function calculateZoomLevel(element: Element | null): number {
   let zoom = 1;
-  let elem: Element | null = element;
-  while (elem) {
-    zoom *= Number(window.getComputedStyle(elem).getPropertyValue('zoom'));
-    elem = elem.parentElement;
+  while (element) {
+    zoom *= Number(window.getComputedStyle(element).getPropertyValue('zoom'));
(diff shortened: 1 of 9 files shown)


ivailop7:
Looking very good! Thanks for fixing! Integrity fails because of prettier formatting.

ivailop7 (approved):
Fantastic fix!

zurfyx (approved):

packages/lexical-utils/src/index.ts
```diff
@@ -524,3 +525,20 @@ export function $insertFirst(parent: ElementNode, node: LexicalNode): void {
...
   }
 }
+
+/**
+ * Calculates the zoom level of an element as a result of using
+ * css zoom property.
+ * @param element
+ */
+export function calculateZoomLevel(element: Element | null): number {
+  if (IS_FIREFOX) {
```
Daniel15, line 535: Should this do feature detection rather than browser detection?

Daniel15:
> Css zoom has only been implemented in Nightly version of Firefox.

Interesting, I didn't realise they finally implemented it. The bug tracking its implementation is 17 years old: https://bugzilla.mozilla.org/show_bug.cgi?id=390936

PR #5778, opened 2024-03-28
Add the ability to discard history more flexibly

hanselabreu:
Currently the only way to discard history is via the `historic` tag. One problem with this approach is that sometimes you do not have control over the update being sent to the editor, thus making it impossible to append any `tags` to the change.

My proposal is to allow a `discardHistory` option to be passed to `registerHistory` which allows a more dynamic way to prevent history from being pushed.

Example below:
https://github.com/facebook/lexical/assets/27902567/1cacdc9a-0dec-4244-80c5-9f2c30361e4e

### DIFF
--- a/packages/lexical-history/src/index.ts
+++ b/packages/lexical-history/src/index.ts
@@ -44,6 +44,10 @@ export type HistoryState = {
   redoStack: Array<HistoryStateEntry>;
   undoStack: Array<HistoryStateEntry>;
 };
+export type HistoryOptions = {
+  delay: number;
+  discardHistory: boolean;
+};
 
 type IntentionallyMarkedAsDirtyElement = boolean;
 
@@ -220,7 +224,7 @@ function isTextNodeUnchanged(
 
 function createMergeActionGetter(
   editor: LexicalEditor,
-  delay: number,
+  options: HistoryOptions,
 ): (
   prevEditorState: null | EditorState,
   nextEditorState: EditorState,
@@ -244,7 +248,7 @@ function createMergeActionGetter(
 
     // If applying changes from history stack there's no need
     // to run history logic again, as history entries already calculated
-    if (tags.has('historic')) {
+    if (options.discardHistory || tags.has('historic')) {
       prevChangeType = OTHER;
       prevChangeTime = changeTime;
       return DISCARD_HISTORY_CANDIDATE;
@@ -288,7 +292,7 @@ function createMergeActionGetter(
         shouldPushHistory === false &&
         changeType !== OTHER &&
         changeType === prevChangeType &&
-        changeTime < prevChangeTime + delay &&
+        changeTime < prevChangeTime + options.delay &&
         isSameEditor
       ) {
         return HISTORY_MERGE;
@@ -389,9 +393,9 @@ function clearHistory(historyState: HistoryState) {
 export function registerHistory(
   editor: LexicalEditor,
   historyState: HistoryState,
-  delay: number,
+  options: HistoryOptions,
 ): () => void {
-  const getMergeAction = createMergeActionGetter(editor, delay);
+  const getMergeAction = createMergeActionGetter(editor, options);
 
   const applyChange = ({
     editorState,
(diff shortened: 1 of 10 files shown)


etrepum:
I don't think you need a PR to do this, I haven't tested this or thought too much about what the current or redoStack should be but something along these lines ought to work

```typescript
import { useState, useRef } from 'react';
import {createEmptyHistoryState, HistoryPlugin} from '@lexical/history';

export default function DiscardHistoryPlugin({ discard = false }: { discard: boolean }) {
  const [history, setHistory] = useState(createEmptyHistoryState);
  const historyRef = useRef(history);
  useEffect(() => {
    if (discard) {
      // save a copy of the history from before discard=true
      historyRef.current = {...history};
    } else {
      // restore the history redoStack & undoStack unless the initial state was discard=false
      setHistory((curHistory) => curHistory === historyRef.current ? curHistory : { ...historyRef.current, current: curHistory.current });
    }
  }, [discard]);
  return (<HistoryPlugin history={history} />);
}
```

hanselninja:
> I don't think you need a PR to do this, I haven't tested this or thought too much about what the current or redoStack should be but something along these lines ought to work
> 
> ```ts
> import { useState, useRef } from 'react';
> import {createEmptyHistoryState, HistoryPlugin} from '@lexical/history';
> 
> export default function DiscardHistoryPlugin({ discard = false }: { discard: boolean }) {
>   const [history, setHistory] = useState(createEmptyHistoryState);
>   const historyRef = useRef(history);
>   useEffect(() => {
>     if (discard) {
>       // save a copy of the history from before discard=true
>       historyRef.current = {...history};
>     } else {
>       // restore the history redoStack & undoStack unless the initial state was discard=false
>       setHistory((curHistory) => curHistory === historyRef.current ? curHistory : { ...historyRef.current, current: curHistory.current });
>     }
>   }, [discard]);
>   return (<HistoryPlugin history={history} />);
> }
> ```

Honestly, not sure I follow you here. That snippet is not accomplishing the goal of this PR. Also, it should probably trigger an infinite loop once you pass `history` as a dependency.

hanselabreu:
> I don't think you need a PR to do this, I haven't tested this or thought too much about what the current or redoStack should be but something along these lines ought to work
> 
> ```ts
> import { useState, useRef } from 'react';
> import {createEmptyHistoryState, HistoryPlugin} from '@lexical/history';
> 
> export default function DiscardHistoryPlugin({ discard = false }: { discard: boolean }) {
>   const [history, setHistory] = useState(createEmptyHistoryState);
>   const historyRef = useRef(history);
>   useEffect(() => {
>     if (discard) {
>       // save a copy of the history from before discard=true
>       historyRef.current = {...history};
>     } else {
>       // restore the history redoStack & undoStack unless the initial state was discard=false
>       setHistory((curHistory) => curHistory === historyRef.current ? curHistory : { ...historyRef.current, current: curHistory.current });
>     }
>   }, [discard]);
>   return (<HistoryPlugin history={history} />);
> }
> ```

Honestly, not sure I follow you here. That snippet is not accomplishing the goal of this PR. Also, it should probably trigger an infinite loop once you pass `history` as a dependency.

The problem is that there is no way to intercept editor changes when `discard = true` to prevent them from being pushed to the `undoStack`.

packages/lexical-history/src/index.ts
```diff
@@ -244,7 +248,7 @@ function createMergeActionGetter(
 
     // If applying changes from history stack there's no need
     // to run history logic again, as history entries already calculated
-    if (tags.has('historic')) {
+    if (options.discardHistory || tags.has('historic')) {
       prevChangeType = OTHER;
       prevChangeTime = changeTime;
       return DISCARD_HISTORY_CANDIDATE;
```
hanselninja, line 254: Got the `discardHistory` name from here.

hanselabreu, line 254: Got the `discardHistory` name from here.

etrepum:

PR #5780, opened 2024-03-29
fix: markdown import code block with prefix

samuelint:
## Problem
When triple backtick bloc code is prefix with tabs or spaces, the parser does not recognize text as a block code.

Not parsed:
```markdown
   \```typescript
   const a = 'hello';
   \```
```

A concret use case, is when code block is under a list.
Example of markdown string to be loaded to lexical:
```
1. First step
   \```bash
   echo "hello"
   \```
2. Second step
   \```bash
   echo "world"
   \```
```
**Note:** `\` is only there to escape code block rendering. The real use case does not have the `\`

## Solution
Change CODE Transformer & MarkdownImport regex to accept tabs or spaces in prefix.

### Current regex
image

### Updated regex
image


acywatson (approved):
I think this is OK. Thanks for the PR.

I will ask @fantactuka to take a look and merge if he has no concerns.

samuelint:
Any update @acywatson @fantactuka, can this MR be merged?

StyleT:
Merging this as apparently it got stuck :P

PR #5781, opened 2024-03-29
fix: restore playground esm build

etrepum:
The /esm/ build of the playground added in #5618 was reverted by the #5744 vite config refactor. This adds it back using the same strategy of a sibling module to avoid code duplication.

### DIFF
--- a/packages/lexical-playground/viteCopyEsm.ts
+++ b/packages/lexical-playground/viteCopyEsm.ts
@@ -27,7 +27,7 @@ export default function viteCopyEsm() {
       {dest: './build/esm/', src: './esm/*'},
       {dest: './build/', src: ['./*.png', './*.ico']},
       ...parseImportMapImportEntries().map(([mod, fn]) => ({
-        dest: path.join('./build/esm/dist', fn),
+        dest: './build/esm/dist/',
         src: path.join(
           `../${mod.replace(/^@/, '').replace(/\//g, '-')}`,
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/viteCopyEsm.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+import fs from 'fs';
```
ivailop7, line 8: I think this should be 'node:fs' and 'node:path' below

ivailop7:
@StyleT can you have a look at this one?

packages/lexical-playground/viteCopyEsm.ts
etrepum: I haven't really seen a case where this makes a difference, although I have not spent any time with the less popular runtimes (at least in a build context). Doesn't hurt though, so I've changed this in [fd47be0](https://github.com/facebook/lexical/pull/5781/commits/fd47be0fce46a47280115dd7e61162a098432326)

etrepum:
Should be working again at https://lexical-playground-7i9hj307l-fbopensource.vercel.app/esm/

ivailop7 (approved):
LGTM

PR #5782, opened 2024-03-29
Deprecate unused methods on ElementNode

thorn0:


ivailop7:
This clean up seems correct, since it's only used in ListItemNode and can be defined there and taken outside of 'Core', however ultimately leaving to @acywatson and @zurfyx to decide if happy with this clean up or there was a plan to use this method in some fashion beyond lists.

thorn0:
Just noticed a similar situation with `canReplaceWith` and `canInsertAfter`. They aren't even used anywhere, only defined on both `ElementNode` and `ListItemNode`. Probably, there is more. It'd be good to clean up this stuff. The `ElementNode` [API page](https://lexical.dev/docs/api/classes/lexical.ElementNode) looks a bit intimidating because of all these methods that don't have any description. They look especially suspicious when you subclass `ElementNode` and have to wonder whether you need to override any of those.

acywatson:
Wow, I didn't realize we aren't using these. Great to clean this up.

I wonder when the callsites were removed...

acywatson:
As these are part of the public API, it's probably best to formally deprecate them, then remove them in a later version.

thorn0:
@acywatson is it good now?

acywatson (approved):

PR #5788, opened 2024-03-30
fix: Address more flaky collab tests

etrepum:
General improvements to the reliability of collab tests

* Fix a few statements where promises were not returned or awaited (particularly in beforeEach)
* Moved the yjs top-level paragraph conflict logic outside of the selection recovery condition
* Used more of the playwright locator methods that retry automatically (may not be strictly necessary but it was part of my exploration)
* Wait for both frames to be connected to collab during initialization
* Clean up a few instances where await and async functions were used for no reason
* Use a bit more of the helper functions to clean things up

### DIFF
--- a/packages/lexical-yjs/src/SyncEditorStates.ts
+++ b/packages/lexical-yjs/src/SyncEditorStates.ts
@@ -103,6 +103,11 @@ export function syncYjsChangesToLexical(
         const event = events[i];
         syncEvent(binding, event);
       }
+      // If there was a collision on the top level paragraph
+      // we need to re-add a paragraph
+      if ($getRoot().getChildrenSize() === 0) {
+        $getRoot().append($createParagraphNode());
+      }
 
       const selection = $getSelection();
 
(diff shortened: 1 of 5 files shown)


packages/lexical-yjs/src/SyncEditorStates.ts
```diff
@@ -103,6 +103,11 @@ export function syncYjsChangesToLexical(
         const event = events[i];
         syncEvent(binding, event);
       }
+      // If there was a collision on the top level paragraph
+      // we need to re-add a paragraph
+      if ($getRoot().getChildrenSize() === 0) {
+        $getRoot().append($createParagraphNode());
+      }
```
etrepum, line 110: I had to bang my head against the wall for a very long time to figure out that this would fix the issue. I haven't read enough yjs and syncEvent code to figure out *why* it's doing the wrong thing (and that it's timing dependent!), but I don't see the harm in making sure that the editor doesn't become completely empty.

ivailop7 (approved):

acywatson (approved):

PR #5789, opened 2024-03-31
fix: iOS Autocorrect strips formatting by reporting wrong dataType

redstar504:
fixes https://github.com/facebook/lexical/issues/4801

A troubling issue has been identified with the iOS keyboard's autocorrect behavior and how it affects text formatting.

When a user selects a word prediction, or, when the operating system injects an autocorrected word (such as when typing `its` and not including the apostrophe), the data transfer type is incorrectly reported as `text/html`.  This does not reflect the actual content, as the payload only contains plain text.

Addressing this issue involves a very straightforward validation to ensure the presence of element nodes within the parsed DOM when the data type is reported as `text/html`, before proceeding with inserting the DOM nodes.

My solution simply checks for the presence of element nodes in the parsed DOM using `querySelector('*')`.  An alternative condition, `Array.from(dom.body.childNodes).some(node => node.nodeType === 1)` could be considered; however `querySelector` offers a more concise expression.

Since iOS reports both `text/plain` and `text/html` data types for it's keyboard corrections, this case falls back to `text/plain` and works properly.  My only concern is if there could be some data transfers that might contain only plain text while reporting only `text/html`.  Perhaps we could condition for those and fall through to the `text/plain` handler if no DOM nodes exist as a result of the predicate that checks for elements.

In my opinion, this simple fix takes the editor from unusable to usable on iOS devices. See the before and after videos. 

Before:

https://github.com/facebook/lexical/assets/1311325/e411b7a5-4e90-4743-be32-346fa4eb4771

After:

https://github.com/facebook/lexical/assets/1311325/54adad94-3e53-4c44-84ca-edfa3bff1543

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -153,7 +153,7 @@ export function $insertDataTransferForRichText(
   // Skip HTML handling if it matches the plain text representation.
   // This avoids unnecessary processing for plain text strings created by
   // iOS Safari autocorrect, which incorrectly includes a `text/html` type.
-  if (htmlString && !(plainString && plainString === htmlString)) {
+  if (htmlString && plainString !== htmlString) {
     try {
       const parser = new DOMParser();
       const dom = parser.parseFromString(
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -148,7 +148,12 @@ export function $insertDataTransferForRichText(
   }
 
   const htmlString = dataTransfer.getData('text/html');
-  if (htmlString) {
+  const plainString = dataTransfer.getData('text/plain');
+
+  // Skip HTML handling if it matches the plain text representation.
+  // This avoids unnecessary processing for plain text strings created by
+  // iOS Safari autocorrect, which incorrectly includes a `text/html` type.
+  if (htmlString && !(plainString && plainString === htmlString)) {
     try {
       const parser = new DOMParser();
       const dom = parser.parseFromString(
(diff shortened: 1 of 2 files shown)


ivailop7:
If this the assumed correct fix, could we please add a test and fix the tests, since they appear to be failing across the board.

ivailop7:
This seems to work correctly now.

BPreisner:
@ivailop7 this still doesn’t work for me on iOS 18:


https://github.com/user-attachments/assets/1c4ec4cd-27b8-495c-b246-13a1933b3f59

amanharwara:
Yeah, this is still an issue. Very easily reproducible using the playground on iOS (i tried with a iOS 16.7 device and also a iOS 18.2 simulator). Can this PR be re-opened?

etrepum:
This PR still needs to be fixed before it can be accepted. If no progress is made on the failing tests and lack of new tests in the next few weeks then it will be closed again.

BPreisner:
@redstar504 are you still around to take a look on these failing tests? 🙏🏼

redstar504:
I updated my solution with a different approach that looks to pass the tests.

If the paste includes both the `text/html` and `text/plain` meta types, as it does when an autocorrect option on Safari is chosen, it simply checks to make sure the content of both are actually different before proceeding with HTML handling.  The content never includes HTML when an autocorrect option is chosen, so I'm not sure why they include that meta type.

BPreisner:
Awesome @redstar504 🚀 @etrepum could you please review?

etrepum (commented):
The code looks fine but it would be better if there was a test that shows that it does what it is trying to do

redstar504:
Added a test.

etrepum (approved):
Thank you for taking the time to write a test!

BPreisner:
Awesome guys, thanks for fixing this @redstar504!

PR #5791, opened 2024-03-31
Node-aware Context Menu Items

ivailop7:
Expose a 'onWillOpen' method that passes the 'contextmenu' mouse event, so users can do their logic what items should end up in the context menu before it's rendered. Have added an example for 'Link node', that appends an item to 'Remove Link' when a link node is right-clicked. Also fixed wrong positioning of the context menu when the page is scrolled.

### DIFF
--- a/packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
@@ -6,6 +6,7 @@
  *
  */
 
+import {$isLinkNode, TOGGLE_LINK_COMMAND} from '@lexical/link';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {
   LexicalContextMenuPlugin,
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
```diff
@@ -199,10 +203,32 @@ export default function ContextMenuPlugin(): JSX.Element {
     [editor],
   );
 
+  const onWillOpen = (event: MouseEvent) => {
+    let newOptions = defaultOptions;
+    editor.update(() => {
```
acywatson, line 208: This should probably be a read that returns the node type, then call an update onClick

packages/lexical-react/src/LexicalContextMenuPlugin.tsx
```diff
@@ -97,6 +99,9 @@ export function LexicalContextMenuPlugin<TOption extends MenuOption>({
   const handleContextMenu = useCallback(
     (event: MouseEvent) => {
       event.preventDefault();
+      if (onWillOpen) {
```
acywatson, line 102: Can we be strict with the check here, like for onOpen and onClose?

packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
ivailop7: $getNearestNodeFromDOMNode works only inside editor.update(), not unsure on why this is the case.

packages/lexical-react/src/LexicalContextMenuPlugin.tsx
ivailop7: updated

acywatson (approved):

Issue #5754, opened 2024-03-22
Bug: Numbered list does not increment in ImageNode caption


jcayne:
Lexical version: v0.14.2 (2024-03-18)

## Steps To Reproduce

1. Navigate to the Lexical playground (https://playground.lexical.dev/) and clear the contents.
2. Create a Numbered list and add an entry.
3. From the toolbar, insert an image (ImageNode). Select the sample.
4. Add a caption to the image.
5. Copy the Numbered list and paste it into the caption.
6. Press enter to add a new entry to the list. The numbered list does not increment.
7. Outside the caption, press enter on the Numbered list. The numbered list increments.

<img width="219" alt="image" src="https://github.com/facebook/lexical/assets/20030168/e8e7fa1f-3f0c-49da-baa7-21f11133c390">

## The current behavior
The numbered list in an ImageNode caption does not increment, for example:
```
1. Item 1
1. Item 2
```

## The expected behavior
The numbered list in an ImageNode caption increments, for example:
```
1. Item 1
2. Item 2
```

PR #5798, opened 2024-04-01
Quick fix for #5754: Numbered list does not increment in ImageNode caption #5754

potatowagon:
**Before**:
see #5646 

**After**:

https://github.com/facebook/lexical/assets/19604232/d48f4880-364e-446b-91e1-44f47b0f2d3c

---

reason for failure: nested editor does not set the transforms of the nodes when creating the editor. This lead to updateChildrenListItemValue, which increments the value of `<li>` (ie, `1.`, `2`. ...), to not be called for nested editors.

- this is a short term fix
- did not add list plugin or markdown plugin to caption nested editor to keep size small
- the long term fix is something like https://github.com/facebook/lexical/tree/nestedsomething

### DIFF
--- a/packages/lexical-react/src/LexicalNestedComposer.tsx
+++ b/packages/lexical-react/src/LexicalNestedComposer.tsx
@@ -20,6 +20,7 @@ import {
   LexicalNode,
   LexicalNodeReplacement,
 } from 'lexical';
+import {Transform} from 'packages/lexical/src/LexicalEditor';
 import * as React from 'react';
 import {ReactNode, useContext, useEffect, useMemo, useRef} from 'react';
 import invariant from 'shared/invariant';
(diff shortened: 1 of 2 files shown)


packages/lexical-react/src/LexicalNestedComposer.tsx
```diff
@@ -20,10 +20,20 @@ import {
...
 
+function getTransformSetFromKlass(
+  klass: KlassConstructor<typeof LexicalNode>,
+): Set<Transform<LexicalNode>> {
+  const transform = klass.transform();
+  return transform !== null
+    ? new Set<Transform<LexicalNode>>([transform])
+    : new Set<Transform<LexicalNode>>();
+}
+
```
potatowagon, line 36: this is from https://github.com/facebook/lexical/blob/5d61ab28edcd8ffe44e5411dcdf089e44dae4bc9/packages/lexical/src/LexicalEditor.ts#L493

packages/lexical-react/src/LexicalNestedComposer.tsx
```diff
@@ -20,10 +20,20 @@ import {
   LexicalNode,
   LexicalNodeReplacement,
 } from 'lexical';
+import {KlassConstructor, Transform} from 'packages/lexical/src/LexicalEditor';
```
zurfyx, line 23: We can't use relative paths, these will not work well when we import these packages outside the playground. You want to use the name of the package instead

```suggestion
import type {KlassConstructor, Transform} from 'lexical';
```

zurfyx (approved):

zurfyx:
Fix LGTM, thank you! Let's address the comment and feel free to merge

packages/lexical-react/src/LexicalNestedComposer.tsx
potatowagon: thankyou good feedback :)

zurfyx (approved):

zurfyx (approved):

PR #5800, opened 2024-04-01
Refactor ClickableLinkPlugin

ivailop7:
The ClickableLinkPlugin needed a small refactor, because in its current implementation it cannot disable the clickability of links in read-only mode. Moving the activeness state to a prop and that prop replacing the internal isEditable, allows for full control on links behaviour in either editor state.

### DIFF
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -191,7 +191,7 @@ export default function Editor(): JSX.Element {
             <TwitterPlugin />
             <YouTubePlugin />
             <FigmaPlugin />
-            <LexicalClickableLinkPlugin active={!isEditable} />
+            <LexicalClickableLinkPlugin disabled={isEditable} />
             <HorizontalRulePlugin />
             <EquationsPlugin />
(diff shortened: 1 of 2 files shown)


zurfyx:
Why should link be disabled in read-only mode? And if that's the case, why not having this logic built-in into the plugin?

ivailop7:
> Why should link be disabled in read-only mode? And if that's the case, why not having this logic built-in into the plugin?

I want to replicate the link behaviour we have in edit mode, in read-only, which is on click to not open the link, but show the floating toolbar only without edit capabilities. I think makes sense even for the Playground, but thought this difference in behaviour was intentional so I kept it. I'd rather give more control to users via props, since this file is in the react package, not in the Playground.

zurfyx (approved):
Based on the behavior and the code it takes to pass the editable hook inside the Plugin I'd be keen to have a DisabledLinkPlugin. The reason why I'm saying this is because it requires the editable subscription and the newTab property is redundant when this is enable.

But this is an opinionated thought and got no strong opinions on! Could also be different outside the playground

PR #5804, opened 2024-04-02
fix: avoid auto-embed menu overflowing off right side of page

donotlb:
**Before fix**: 

image

**After fix**:

image


thegreatercurve:
Seems like a reasonable fix! Thanks.

thegreatercurve (approved):

PR #5807, opened 2024-04-02
Table selection stuck fix

AlexanderReznik:
Starting selection in front of the table didn't trigger adding event listeners for mouseMove and mouseUp. Fixing this.

Before:
Selection stuck before

After:
Selection stuck after

### DIFF
--- a/packages/lexical-table/src/LexicalTableObserver.ts
+++ b/packages/lexical-table/src/LexicalTableObserver.ts
@@ -72,7 +72,7 @@ export class TableObserver {
   editor: LexicalEditor;
   tableSelection: TableSelection | null;
   hasHijackedSelectionStyles: boolean;
-  inSelectionProgress: boolean;
+  isSelecting: boolean;
 
   constructor(editor: LexicalEditor, tableNodeKey: string) {
     this.isHighlightingCells = false;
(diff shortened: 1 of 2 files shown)


packages/lexical-table/src/LexicalTableObserver.ts
```diff
@@ -72,6 +72,7 @@ export class TableObserver {
   editor: LexicalEditor;
   tableSelection: TableSelection | null;
   hasHijackedSelectionStyles: boolean;
+  inSelectionProgress: boolean;
```
ivailop7, line 75: does it make sense to be called 'isSelecting' maybe?

ivailop7: since there's a method called 'isHighlightingCells' to keep things consistent

AlexanderReznik: Yes, I didn't like the previous name either
https://github.com/facebook/lexical/pull/5807/commits/60215b1dd2426293403d86a97dcd10ae00e4aa87

ivailop7 (commented):
LGTM

ivailop7 (approved):

PR #5811, opened 2024-04-03
fix: removed overwrite of copyright

svendjanis:
The copyright object in footer.json was overriding the copyright value set in docusaurus.config.js, where the year is correctly utilized. The removed copyright object in footer.json also had the year 2022 hardcoded.


thegreatercurve:
Thanks @svendjanis!

thegreatercurve (approved):

PR #5822, opened 2024-04-04
Fix #5620 Persist format at paragraph level for new line

### LINKED ISSUE #5620: How to keep formatting between lines?

Sahejkm:
- The selection looses the format (bold,italic,etc) on tap of a new line
- Update the paragraph node to persist the format which could be used to update the selection format on new paragraph creation
- This fixes #5620
Before 

https://github.com/facebook/lexical/assets/163521239/389b5f78-154a-4c5d-9e2b-d523415e2789

After:

https://github.com/facebook/lexical/assets/163521239/9ccdc2ec-676d-4d02-b005-773dd8381bfd


packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -6,35 +6,67 @@
...
   LexicalNode,
+  NodeKey,
 } from '../LexicalNode';
 import type {
   ElementFormatType,
   SerializedElementNode,
 } from './LexicalElementNode';
-import type {RangeSelection} from 'lexical';
 
+import {type RangeSelection} from 'lexical';
```
ivailop7, line 26: I believe 'type' should be outside of the curly braces

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -6,35 +6,67 @@
...
+
+  setTextFormat(type: number): this {
+    const self = this.getWritable();
+    self.__textFormat = type;
+    return self;
+  }
+
+  hasTextFormat(type: TextFormatType): boolean {
+    const formatFlag = TEXT_TYPE_TO_FORMAT[type];
+    return (this.getTextFormat() & formatFlag) !== 0;
```
ivailop7, line 67: Is this deliberately bitwise 'and' ?

packages/lexical/src/nodes/__tests__/unit/LexicalParagraphNode.test.ts
```diff
@@ -113,17 +113,16 @@ describe('LexicalParagraphNode tests', () => {
...
       await editor.update(() => {
-        const result = paragraphNode.insertNewAfter(
-          $getSelection() as RangeSelection,
-          false,
-        );
-
-        expect(result).toBeInstanceOf(ParagraphNode);
-        expect(result.getDirection()).toEqual(paragraphNode.getDirection());
+        const selection = $getSelection();
+        if (selection != null && $isRangeSelection(selection)) {
```
zurfyx, line 117:
```suggestion
        if ($isRangeSelection(selection)) {
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
zurfyx: I believe we're mimicking the same format behavior from ElementNode, multiple values on a the same int

packages/lexical/src/LexicalSelection.ts
```diff
@@ -837,7 +837,6 @@ export class RangeSelection implements BaseSelection {
         (firstNodeFormat !== format || firstNodeStyle !== style)
       ) {
         if (firstNode.getTextContent() === '') {
-          firstNode.setFormat(format);
```
zurfyx, line 840: Q: Do you know where this node format is being set now?

packages/lexical/src/nodes/LexicalParagraphNode.ts
Sahejkm: yeah its referenced from LexicalTextNode and used for logging/printing the format in the Tree View

packages/lexical/src/LexicalEvents.ts
```diff
@@ -344,7 +344,15 @@ function onSelectionChange(
             selection.format = anchorNode.getFormat();
             selection.style = anchorNode.getStyle();
           } else if (anchor.type === 'element' && !isRootTextContentEmpty) {
-            selection.format = 0;
+            const lastNode = anchor.getNode();
+            if (
+              lastNode instanceof ParagraphNode &&
+              lastNode.getChildrenSize() === 0
+            ) {
+              selection.format = lastNode.getTextFormat();
```
zurfyx, line 352: We can leave this condition as children size 0 for now but we should revise it later. For example, if you start typing a paragraph with an image it wouldn't follow under this rule

zurfyx (commented):
We need to look into this crash


https://github.com/facebook/lexical/assets/193447/02a1bad7-842b-4d21-a931-a10675808061

packages/lexical/src/LexicalSelection.ts
Sahejkm: yeah, for text node its always from selection, same as before

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -581,4 +582,49 @@ test.describe('Selection', () => {
       `,
     );
   });
+
+  test('Can persist the text format from the paragraph', async ({page}) => {
```
zurfyx, line 586: You probably want to skip plain text here

zurfyx (commented):
LGTM, thanks! Can you add a couple more tests please?

1. Toggle format with CMD + B at the start of a paragraph of different format, toggled format persists
2. Removing all nodes of a paragraph with format and typing (I know that behavior doesn't resemble GDocs/Word but let's at least make sure it doesn't regress).

Can we follow up on the table formatting (in a separate PR)? Particularly persisting the format on creation and adding/removing rows and columns.

zurfyx (approved):

nickschinestzki:
Does this fix also applies for styling? Like text color, font family, font size, etc?

PR #5824, opened 2024-04-04
Fix copying html table with unequal number of cells in a row

KatsiarynaDzibrova:
Issue: when copying html table with not equal number of cells in row the pasted table is broken. The number of cells in each row is different and it breaks on selection.
Fix: adding empty cells on paste

Before:
old-copy-partial

After:
new-copy-partial

### DIFF
--- a/packages/lexical-react/src/LexicalTablePlugin.ts
+++ b/packages/lexical-react/src/LexicalTablePlugin.ts
@@ -84,7 +84,7 @@ export function TablePlugin({
         COMMAND_PRIORITY_EDITOR,
       ),
       editor.registerNodeTransform(TableNode, (node) => {
-        const gridMap = $computeTableMapSkipCellCheck(node);
+        const [gridMap] = $computeTableMapSkipCellCheck(node, null, null);
         const maxRowLength = gridMap.reduce((curLength, row) => {
           return Math.max(curLength, row.length);
         }, 0);
--- a/packages/lexical-react/src/LexicalTablePlugin.ts
+++ b/packages/lexical-react/src/LexicalTablePlugin.ts
@@ -13,12 +13,11 @@ import type {
 } from '@lexical/table';
 import type {NodeKey} from 'lexical';
 
-import {mergeRegister} from '@lexical/utils';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {
   $computeTableMap,
-  $createTableCellNode,
   $computeTableMapSkipCellCheck,
+  $createTableCellNode,
   $createTableNodeWithDimensions,
   $getNodeTriplet,
   $isTableCellNode,
@@ -30,7 +29,11 @@ import {
   TableNode,
   TableRowNode,
 } from '@lexical/table';
-import {$insertFirst, $insertNodeToNearestRoot} from '@lexical/utils';
+import {
+  $insertFirst,
+  $insertNodeToNearestRoot,
+  mergeRegister,
+} from '@lexical/utils';
 import {
   $createParagraphNode,
   $getNodeByKey,
--- a/packages/lexical-react/src/LexicalTablePlugin.ts
+++ b/packages/lexical-react/src/LexicalTablePlugin.ts
@@ -96,6 +96,7 @@ export function TablePlugin({
           const lastCellMap = gridMap[i][rowLength - 1];
           const lastRowCell = lastCellMap.cell;
           for (let j = rowLength; j < maxRowLength; ++j) {
+            // TODO: inherit header state from another header or body
             const newCell = $createTableCellNode(0);
             newCell.append($createParagraphNode());
             if (lastRowCell !== null) {
--- a/packages/lexical-react/src/LexicalTablePlugin.ts
+++ b/packages/lexical-react/src/LexicalTablePlugin.ts
@@ -13,10 +13,12 @@ import type {
 } from '@lexical/table';
 import type {NodeKey} from 'lexical';
 
+import {mergeRegister} from '@lexical/utils';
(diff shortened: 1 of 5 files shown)


ivailop7 (dismissed):
Fantastic! Thank you, Katsia!

ivailop7:
@KatsiarynaDzibrova tests are still failing

KatsiarynaDzibrova:
> @KatsiarynaDzibrova tests are still failing

Looking into it

zurfyx (commented):
Thank you, have you considered having a transform for this at the plugin level? While I think this approach is correct ultimately we want to lean toward a $normalizer (aka Schemas) which would help mitigate bad structures originating from text input and plugins.

This might also help simplify the additional exported utilities introduced in this PR. WDYT?

ivailop7:
ping @zurfyx

zurfyx:
The normalizer was an example, we definitely want to think about it further before we can propose something generic.

But you can still have the table plugin do these checks for you.

```
editor.registerTransform(TableCell, node => {
  // validate correct number of cells and rows
});
```

Because of the current implementation which relies on a full scan of rows and column this might not be very optimal so we may want to restrict to the paste event for now, you can control this via commands. In the future, we would ideally keep track of the previously explored columns and rows and resume the check from the position instead of scanning the full table.

KatsiarynaDzibrova:
> The normalizer was an example, we definitely want to think about it further before we can propose something generic.
> 
> But you can still have the table plugin do these checks for you.
> 
> ```
> editor.registerTransform(TableCell, node => {
>   // validate correct number of cells and rows
> });
> ```
> 
> Because of the current implementation which relies on a full scan of rows and column this might not be very efficient for long tables so we may want to restrict to the paste event for now, you can control this via commands. In the future, we would ideally keep track of the previously explored columns and rows and resume the check from the position instead of scanning the full table.

I agree that it makes sense to trigger this logic only on paste.
But does it mean creating a new command, which will be dispatched in `convertTableElement`? Because `PASTE_COMMAND` can't be used here as it's called before the node is created and I don't think there is another existing command that can be used here.
I don't think it's worth creating a new command if we plan to remove it later. Maybe it's better to leave the logic on TableNode level and move it later if the normalizer is created?

packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -714,12 +714,32 @@ export function $unmergeCell(): void {
...
+  );
+  invariant(cellAValue !== null, 'Anchor not found in Grid');
+  invariant(cellBValue !== null, 'Focus not found in Grid');
+  return [tableMap, cellAValue, cellBValue];
+}
+
+function $computeTableMapHelper(
+  grid: TableNode,
+  cellA: TableCellNode,
+  cellB: TableCellNode,
```
zurfyx, line 741: Let's make this null-ish and we avoid exposing another function externally. This is also backward-compatible safe.

```suggestion
  cellA: null | TableCellNode,
  cellB: null | TableCellNode,
```

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -57,24 +63,49 @@ export function TablePlugin({
...
+        }, 0);
+        for (let i = 0; i < gridMap.length; ++i) {
+          const rowLength = gridMap[i].length;
+          if (rowLength === maxRowLength) {
+            continue;
+          }
+          const lastCellMap = gridMap[i][rowLength - 1];
+          const lastRowCell = lastCellMap.cell;
+          for (let j = rowLength; j < maxRowLength; ++j) {
+            const newCell = $createTableCellNode(0);
```
zurfyx, line 99: This should be inherited from somewhere, either another header or body. We can skip for now but we should track with a todo or something

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -57,24 +63,49 @@ export function TablePlugin({
...
+          }
 
-        const firstDescendant = tableNode.getFirstDescendant();
-        if ($isTextNode(firstDescendant)) {
-          firstDescendant.select();
+          return true;
+        },
+        COMMAND_PRIORITY_EDITOR,
+      ),
+      editor.registerNodeTransform(TableNode, (node) => {
```
zurfyx, line 86: Let's see how this perform, otherwise we may want to add a check that this only happens strictly on clipboard paste.

@Sahejkm this is the case we discussed offline, ideally this map is cached and future checks are O(1).

zurfyx (approved):
Thanks for iterating on this and sorry for keeping you blocked for so long, stamping but please address the last comment before merging

packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -714,12 +714,31 @@ export function $unmergeCell(): void {
     cell.setRowSpan(1);
   }
 }
+export function $computeTableMapSkipCellCheck(grid: TableNode) {
+  const [tableMap] = $computeTableMapHelper(grid, null, null);
+  return tableMap;
+}
```
zurfyx, line 720:
```suggestion
```

zurfyx (dismissed):

packages/lexical-table/src/LexicalTableUtils.ts
KatsiarynaDzibrova: I don't see any changes here

ivailop7: @zurfyx I don't see any changes either in the suggestion, did you mean something else here?

zurfyx: This function becomes redundant now that the other takes nullable values, doesn't it? Should we use the other function instead?

ivailop7 (approved):

PR #5833, opened 2024-04-05
fix: webpack needs its own export condition

### LINKED ISSUE #5823: Bug: fails to compile on nextjs Client Component SSR - Error: Element type is invalid: expected a string
Lexical version: 0.14.3

## Steps To Reproduce

1. clone https://github.com/2wheeh/nextjs-lexical-build-error
2. try build (or compile dynamic route pages by accessing `localhost:3000/not-work/hi`)

Link to code example: https://github.com/2wheeh/nextjs-lexical-build-error

## The current behavior

Call apis or anything imported from Lexical in **_ClientComponent_** (with `'use client'` directive).
Rendering this component in nextjs server side fails.

```ts
// client-component.tsx
'use client';

import { COMMAND_PRIORITY_CRITICAL } from 'lexical'; // just a simple constant. 

export default function ClientComponent() {
  COMMAND_PRIORITY_CRITICAL;

  return <>hi</>;
}
```

```ts
// not-work/page.tsx
import ClientComponent from '../client-component';

export default function Page() {
  return <ClientComponent />;
}
```

```
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.
```

build time: fails to generate static pages.

image

run time: fails to compile dynamic ssr pages.

image

Now It only works with dynamic imports with no ssr.
```ts
// work/page.tsx
import dynamic from 'next/dynamic';

const ClientComponent = dynamic(() => import('../client-component'), { ssr: false });

export default function Page() {
  return <ClientComponent />;
}
```

## The expected behavior

works well as it was til 0.14.2

etrepum:
Webpack (and SWC?) produce some truly inscrutable error messages when you provide it with ESM that uses top-level await. Unfortunately, it also prefers to load "node" condition code over "default". In order to support both node ESM imports and webpack at the same time we need to add a separate "webpack" condition at a higher priority than "node".

Related reading: https://webpack.js.org/guides/package-exports/#conditions-preprocessor-and-runtimes

🤦 

Fix #5823

### DIFF
--- a/packages/lexical-clipboard/package.json
+++ b/packages/lexical-clipboard/package.json
@@ -30,6 +30,7 @@
     ".": {
       "import": {
         "types": "./index.d.ts",
+        "webpack": "./LexicalClipboard.mjs",
         "node": "./LexicalClipboard.node.mjs",
         "default": "./LexicalClipboard.mjs"
       },
--- a/packages/lexical-code/package.json
+++ b/packages/lexical-code/package.json
@@ -30,6 +30,7 @@
     ".": {
       "import": {
         "types": "./index.d.ts",
+        "webpack": "./LexicalCode.mjs",
         "node": "./LexicalCode.node.mjs",
         "default": "./LexicalCode.mjs"
       },
--- a/packages/lexical-dragon/package.json
+++ b/packages/lexical-dragon/package.json
@@ -23,6 +23,7 @@
     ".": {
       "import": {
         "types": "./index.d.ts",
+        "webpack": "./LexicalDragon.mjs",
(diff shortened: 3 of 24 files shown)


2wheeh:
You are amazing !

etrepum:
It doesn't always feel that way after weeks of fixing one thing and breaking something else in the process 😅 At least everyone is nice here!

acywatson (approved):

ivailop7 (approved):

ivailop7:
Seeing all these webpack-related issues, is a good reminder why everyone embraced vite and esbuild so fast.

PR #5809, opened 2024-04-02
docs: clarify read/update language in intro, readme & editor state

etrepum:
Add some clarifying documentation to the introductory pages:

* [README.md](https://github.com/etrepum/lexical/blob/doc-updates/README.md)
* [Intro](https://lexical-git-fork-etrepum-doc-updates-fbopensource.vercel.app/docs/intro)
* [Editor State](https://lexical-git-fork-etrepum-doc-updates-fbopensource.vercel.app/docs/concepts/editor-state)

PR #5838, opened 2024-04-05
chore: upgrade to docusaurus 3

etrepum:
In #5809 I added some gfm table syntax which docusaurus 2 doesn't render by default. I tried including the remark-gfm plugin with docusaurus 2 but it didn't work for esm reasons (my villain arc) so here is an upgrade to docusaurus 3 where it does work (and remark-gfm is enabled by default).

I did have to change broken links from throw to warn mostly because there *are* broken links and it just wasn't catching them before. I've noticed at least one misspelling and otherwise a systemic problem is that typedoc labels must be `[a-z_][0-9a-z_]*` but JavaScript and TypeScript names can have other characters, like our beloved `$`. typedoc seems perfectly happy to strip these from the labels it generates, but it doesn't do the same kind of normalization for something like `{@link lexical!$getRoot}` so it will generate an anchor `#$getroot` but it should be `#getroot`. I'm looking into a workaround for this but that will likely take a little longer.


zurfyx (approved):

zurfyx (commented):
LGTM, thanks!

PR #5839, opened 2024-04-05
Fixed playground draggable block while scrolled

Legonois:
When using the playground in a smaller window and scrolling down, it is impossible to move a node to the top node.

This fix adjusts the draggableBlockPlugin to allow the placement of a node at the root while scrolled. It was improperly calculated, as the mouseY variable is tied to the location relative to the page and not the screen. 

I encountered this bug in my own build of this plugin.

### DIFF
--- a/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
@@ -231,7 +231,8 @@ function setTargetLine(
     anchorElem.getBoundingClientRect();
   const {marginTop, marginBottom} = getCollapsedMargins(targetBlockElem);
   let lineTop = targetBlockElemTop;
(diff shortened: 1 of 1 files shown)


ivailop7:
Hi, could you add a short gif of the issue, I'm trying to reproduce the original problem that's solved. The fix does make sense to me. You would also need to sign the CLA.

Legonois:
Here is the GIF of the issue:
Lexical-Playground-Bug

I also signed the agreement.

Tell me if there's anything else I need to do!

Legonois:
Wait, it still doesn't solve the underlying issue. The target line gets set correctly, but it does not set what element gets moved. I will be pushing a change later today

Sahejkm:
Hi @Legonois , should we add tests as well for the fix ?

ivailop7:
Reopening until the changes from this PR are ported over to the other one.

PR #5844, opened 2024-04-08
Fixed playground draggable block while scrolled

Legonois:
Last pull request did not include the fix for the actual on drop event, only for the placement of the target line. This adjusts the drop event to account for scrolling


ivailop7:
I guess this PR is meant to supersede the #5839 ? So 5839 is closable?

ivailop7:
Still not able to reproduce this issue in the current Playground

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
```diff
@@ -361,7 +361,8 @@ function useDraggableBlockMenu(
         return true;
       }
       const targetBlockElemTop = targetBlockElem.getBoundingClientRect().top;
-      if (pageY / calculateZoomLevel(target) >= targetBlockElemTop) {
+      const adjsutedTop = targetBlockElemTop + window.scrollY;
+      if (pageY / calculateZoomLevel(target) >= adjsutedTop) {
```
potatowagon, line 365: nit: typo in `adjsutedTop`

ivailop7: @Legonois are you planning on fixing the typo, so we can proceed forward with the merge

Legonois:
Just checked the deployments for the bug and it seems fixed! 

Tell me if there is anything else I need to do before this can be pulled to main

PR #5850, opened 2024-04-09
docs: fix typedoc links and add missing packages

etrepum:
* Added lexical-typedoc-module-name plugin to determine the module name from the path
  * Removed `@module` from all index.ts files
  * Enabled header/header eslint rule everywhere
  * Fixed header/header lint mistakes from modules that formerly contained `@module`
* Added lexical-remark-slugify-anchors plugin to fix typedoc-generated links to symbols that start with `$` so that they work in docusaurus, and also fixes up the text so the links don't include the package scope, e.g. `lexical/html!$generateHtmlFromNodes` now displays as `$generateHtmlFromNodes` and should also link to the correct place
* Added `@lexical/react` and `@lexical-yjs` package docs to the sidebar
* Added entrypoints for `@lexical/plain-text`, `@lexical/rich-text`, `@lexical/react`  to the typedoc config
* Configures typedoc to generate github links to the main tree (it works by default locally, but wherever this is building doesn't have a proper git checkout with remote, so this will override it)
* Removes unnecessary `sidebar_position`, `cache_reset` frontmatter from package docs
* Fixed `@link` typo in LexicalNode typedoc comment
* Fixed URL typo in @lexical/devtools-core package docs


StyleT:
I will give others opportunity to review and merge

acywatson (approved):

PR #5851, opened 2024-04-09
Documentation: add PR template

potatowagon:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,16 @@
+## Description
+
+*Describe the changes in this diff*
+
+Closes #<issue>
+
+## Test plan
+
+### Before
+
+*Insert relevant screenshots/recordings/automated-tests*
+
+
+### After
+
+*Insert relevant screenshots/recordings/automated-tests*
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 ## Description
 
-*Describe the changes in this diff*
+*Describe the changes in this pull request*
 
 Closes #<issue>
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,22 @@
-## Description
+<!-- 
+Title format should be:
+[PR Type][Affected Packages] title
 
+Example:
+[Feature][lexical-playground] Add more emojis 
+
+Choose from the following PR Types:
+Breaking change / Refactor / Feature / Bug Fix / Documentation Update / Chore
+-->
+
+## Description
+<!-- 
+- What is the current behavior that you are modifying? 
+- What are the behavior or changes that are being added by this PR?
+-->
 *Describe the changes in this pull request*
 
-Closes #<issue>
+**Closes:** #<!-- issue number -->
 
 ## Test plan
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,9 @@
 <!-- 
 Title format should be:
-[PR Type][Affected Packages] title
+[Affected Packages] PR Type: title
 
 Example:
-[Feature][lexical-playground] Add more emojis 
+[lexical-playground][lexical-link] Feature: Add more emojis 
 
 Choose from the following PR Types:
 Breaking change / Refactor / Feature / Bug Fix / Documentation Update / Chore


.github/pull_request_template.md
```diff
@@ -0,0 +1,20 @@
+## Issue
+
+*Please reference an issue...*
+
+#1234
+
+## Description
+
+*Describe the changes in this diff*
```
zurfyx, line 9: [Opinionated] Let's use the closes tag, I also don't think we need an entire section to reference the issue, too much clutter.

```suggestion
## Description

*Describe the changes in this diff*

Closes #<issue>
```

zurfyx (commented):
LGTM, but I'll let this PR sit for a bit to see if the rest have some opinions

.github/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
...
```
fantactuka, line 3: nit: this pull request

StyleT:
Hi!

Let's add selection of the affected modules to the PR template, so we can sort them easier. Ex: [package name] (@lexical/core). This would allow people that recently worked on the respective packages to review PRs faster.

Also sorting by type would help: chore, docs, feat, bug...

potatowagon:
> Let's add selection of the affected modules to the PR template, so we can sort them easier. Ex: [package name] (@lexical/core). 
> Also sorting by type would help: chore, docs, feat, bug...

do u have an example of how this would fit into the PR template? youre welcomed to suggest a code change directly onto .github/pull_request_template.md

.github/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
+## Description
```
StyleT, line 1:
```suggestion
## What type of PR is this?

<!-- Please try to limit your pull request to one type (bugfix, feature, etc). Submit multiple pull requests if needed. -->
<!-- Leave 1 option from the list: -->
**Breaking change / Refactor / Feature / Bug Fix / Documentation Update / Chore**

## Description

<!-- Describe the changes in this pull request -->

**Current behavior:**
<!-- Please describe the current behavior that you are modifying. -->

**New behavior:**
<!-- Please describe the behavior or changes that are being added by this PR. -->

**Closes:** #<!-- issue number -->

**Affected packages:** <!-- List names of the affected packages within lexical monorepo. If none affected - put "infra" here.

## Test plan

### Before

*Insert relevant screenshots/recordings/automated-tests*

### After

*Insert relevant screenshots/recordings/automated-tests*

## Added/updated automated tests?
_We encourage you to keep the code coverage percentage at 80% and above._

- [ ] Yes
- [ ] No, and this is why: _please replace this line with details on why tests
      have not been included_
- [ ] I need help with writing tests
```

StyleT (changes requested):

.github/pull_request_template.md
zurfyx: I wonder how much this will impact velocity. I like the detailed template as a means to write down as much information as possible but a too verbose summary will impact both the writers and the readers (I often find myself re-reading old PRs to understand the context).

IMO the package name should be part of the title, the close part can be part of the description and the automated tests don't need a checkbox list.

I'd advocate for a balance between @potatowagon 's initial PR and this.

potatowagon: I added a title guidance, which prompts to list PR Type and lexical packages. reason: those info would be helpful when screening the list of pr titles of a sync diff.

For the description ive updated the guidance to include @StyleT suggestion. leaving that as guidance instead as the before/after behavior is covered in the test plan

.github/workflows/test.yml
```diff
@@ -129,7 +129,7 @@ jobs:
             ~/Library/Caches/ms-playwright
           key: ${{ runner.os }}-v${{ secrets.CACHE_VERSION }}-${{ hashFiles('package-lock.json') }}
       - name: Install dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+        # if: steps.cache.outputs.cache-hit != 'true'
```
potatowagon, line 132: not planning to commit this, temp fix to workflows error

>failed to load config from /Users/runner/work/lexical/lexical/packages/lexical-playground/vite.prod.config.ts
error during build:
Error: 
You installed esbuild on another platform than the one you're currently using.
This won't work because esbuild is written with native code and needs to
install a platform-specific binary executable.

Specifically the "esbuild-darwin-[64](https://github.com/facebook/lexical/actions/runs/8844961076/job/24287908130?pr=5851#step:8:65)" package is present but this platform
needs the "esbuild-darwin-arm64" package instead. People often get into this
situation by installing esbuild on Windows or macOS and copying "node_modules"
into a Docker image that runs Linux, or by copying "node_modules" between
Windows and WSL environments.

StyleT (approved):

PR #5855, opened 2024-04-09
Fix #5663 styling from td not persisted on copy paste

### LINKED ISSUE #5663: Bug: copy and pasting table cell does not contains individual cell text style
If I have copied a table from any sheet and pasted it here, it's not showing any styling on any cell. Instead, it is displaying a normal table with no style.

Lexical version:0.12.5

## Steps To Reproduce

[Screencast from 28-02-24 05:48:11 PM IST.webm](https://github.com/facebook/lexical/assets/107410323/74246ddd-4072-438d-858e-5f54b531530a)

Sahejkm:
Before:

https://github.com/facebook/lexical/assets/163521239/5d32198b-aa6b-4fd0-9d3c-b6f03b0eaa6f

After:

https://github.com/facebook/lexical/assets/163521239/bbddae10-bef6-472d-9dc5-26ab34bee24c

closes #5663

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs
@@ -980,6 +980,64 @@ test.describe('CodeBlock', () => {
       </span>
     </code>
   `;
+  const EXPECTED_HTML_GOOGLE_SPREADSHEET = html`
+    <table class="PlaygroundEditorTheme__table">
+      <tr style="height: 21px">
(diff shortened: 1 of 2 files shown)


zurfyx (approved):
LGTM, we might eventually want to move most of the E2E tests in this file to unit, there is little value in a full E2E run -- like we did in here https://github.com/facebookexternal/Outline//blob/9f8419195176e44f5755ef087d989dabb524c5a4/packages/lexical-table/src/__tests__/unit/LexicalTableNode.test.ts#L71

PR #5856, opened 2024-04-09
Fix inserting text from Grammarly when full node is selected with forward selection

Piliuta:
## Problem
Inserting text from Grammarly fails to save the state when a full-text node is selected in the forward direction (from start to end).

https://github.com/facebook/lexical/assets/1575198/bc336023-2447-4524-9dda-a60530667d64

## Solution
Check the selection direction when comparing inserted content.

https://github.com/facebook/lexical/assets/1575198/74d15054-e6c4-4707-9209-8e9dc453b867

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Extensions.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Extensions.spec.mjs
@@ -7,6 +7,10 @@
  */
 
 import {
+  moveToEditorBeginning,
+  moveToEditorEnd,
+} from '../keyboardShortcuts/index.mjs';
+import {
   assertHTML,
   assertSelection,
   evaluate,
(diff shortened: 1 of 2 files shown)


StyleT:
Do you think it's possible to cover this with test? So it wouldn't reoccur later on

Piliuta:
> Do you think it's possible to cover this with test? So it wouldn't reoccur later on

This is a fair request. I was looking at the original PR and saw this comment that says it might be hard to test https://github.com/facebook/lexical/pull/3429#issuecomment-1328035465 so I did not try.

But I can take another look and see if it is possible to cover at least this logic with a test.

Piliuta:
> Do you think it's possible to cover this with test? So it wouldn't reoccur later on

Added a couple of tests https://github.com/facebook/lexical/pull/5856/commits/e39f762dd31f13e985cb8871f1e2aa6b87ec7573

Double double-checked them on the current maser and the second failed there that confirms the bug
image

StyleT (approved):

StyleT:
Approved! Thanks for this fantastic PR. 

I'll give others a chance to review and merge it. If no one will take an action - expect it to be merged this week.

Piliuta:
I see a new version was released but this PR wasn't merged. Is this missing anything or just waiting for the next version? If this is okay, any ETA for the next release?

cc @acywatson

Issue #5812, opened 2024-04-03
Bug: Inserting a table from  "Clickup" can lead to Lexical crash


denindka:
Lexical version: - 0.14.2

## Steps To Reproduce

1. Go in "Clickup", insert a table and copy it
2. Paste it into Lexical
3. Go to some cell with text
4. Press Enter(new line)
5. Lexical is crashing

https://github.com/facebook/lexical/assets/20279408/048c3643-91bb-4e6a-af7c-d96e91408d58

StyleT:
Hi! 
Can you provide any other way to reproduce this issue for those who doesn't have clickup.com account?

denindka:
Hi! of course here an attachment
[Playground 2024-04-08T07_26_04.935Z.lexical.zip](https://github.com/facebook/lexical/files/14902295/Playground.2024-04-08T07_26_04.935Z.lexical.zip)

Sahejkm:
closing this as it should be fixed now

PR #5857, opened 2024-04-09
[lexical-html] Feature: Support copy pasting block and inline nodes properly

### LINKED ISSUE #4465: Support div on paste

potatowagon:
## Description

Closes #4465 #5812

hasBlockAncestorElement means if a node has an ancestor element somewhere up the tree between itself and the root (or shadow root).

Paragraphs being block elements should not be nested in another block element, and the semantically right way to do so is to insert line breaks.

If hasBlockAncestorElement, wrap continuous child nodes in an Artificial Node, else, wrap continuous child nodes in paragraph node. 

The implementation for wrapping continuous child nodes is reused from #5834

Afterwards insert line breaks (`<br>`) in between all Artificial nodes 
```
[A,A] => [A <br> A] 
```
and unwrap the contents of the Artificial nodes, by replacing the Artificial node with its children at the parent node.
```
[A <br> A] => [123 <br> 456] 
```
This would hopefully insert paragraphs and line breaks correctly.

Big thanks to @zurfyx brainstorming this approach! 

## Test Plan

Case: nested divs
```
<div>
  a
  <div>
    b
    <div>
      c
      <div>
        <div>
        </div>
        z
      </div>
    </div>
    d
    e
  </div>
  fg
</div>
```

Screenshot 2024-04-09 at 10 20 22 PM

### Before
Screenshot 2024-04-09 at 10 20 35 PM

### After
Screenshot 2024-04-09 at 10 20 45 PM

---
Case: nested divs in list
```
<ol>
  <li>
    1
    <div>
      2
    </div>
    3
  </li>
  <li>
    A
    <div>
      B
    </div>
    3
  </li>
</ol>
```

Screenshot 2024-04-09 at 10 16 33 PM

### Before

Screenshot 2024-04-09 at 10 16 20 PM

### After

Screenshot 2024-04-09 at 10 19 35 PM


potatowagon:
Case: nested spans and nested divs
```
<div>
  a b
  <span>
    c d
    <span>
      e
    </span>
  </span>
  <div>
    f
    <span>
      g h
    </span>
  </div>
</div>
```

Screenshot 2024-04-15 at 3 35 40 PM

## before

Screenshot 2024-04-15 at 3 35 21 PM

## after

Screenshot 2024-04-15 at 3 35 35 PM

potatowagon:
Case: simple span in a div

```
<div>
  a
  <span>b</span>
</div>
```
Screenshot 2024-04-15 at 3 39 50 PM


## Before

Screenshot 2024-04-15 at 3 40 26 PM



## After

Screenshot 2024-04-15 at 3 40 49 PM

packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs
```diff
@@ -1118,7 +1143,7 @@ test.describe('CodeBlock', () => {
...
...3px 2px 3px;vertical-align:bottom;font-style:italic;" data-sheets-value="{&quot;1&quot;:2,&quot;2&quot;:&quot;MWP_WORK_LS_COMPOSER&quot;}">MWP_WORK_LS_COMPOSER</td><td style="overflow:hidden;padding:2px 3px 2px 3px;vertical-align:bottom;text-decoration:underline;text-align:right;" data-sheets-value="{&quot;1&quot;:3,&quot;3&quot;:77349}">77349</td></tr><tr style="height:21px;"><td style="overflow:hidden;padding:2px 3px 2px 3px;vertical-align:bottom;" data-sheets-value="{&quot;1&quot;:2,&quot;2&quot;:&quot;Lexical&quot;}">Lexical</td><td style="overflow:hidden;padding:2px 3px 2px 3px;vertical-align:bottom;text-decoration:line-through;" data-sheets-value="{&quot;1&quot;:2,&quot;2&quot;:&quot;XDS_RICH_TEXT_AREA&quot;}">XDS_RICH_TEXT_AREA</td><td style="overflow:hidden;padding:2px 3px 2px 3px;vertical-align:bottom;" data-sheets-value="{&quot;1&quot;:2,&quot;2&quot;:&quot;sdvd sdfvsfs&quot;}" data-sheets-textstyleruns="{&quot;1&quot;:0}{&quot;1&quot;:5,&quot;2&quot;:{&quot;5&quot;:1}}"><span style="font-size:10pt;font-family:Arial;font-style:normal;">sdvd </span><span style="font-size:10pt;font-family:Arial;font-weight:bold;font-style:normal;">sdfvsfs</span></td></tr></tbody></table>`,
```
potatowagon, line 1146: looks like this:
Screenshot 2024-04-15 at 11 05 43 PM
---

had to update the gsheets test case because this fix wraps a text node cell that is a direct decendent of table cell with a para, which is the correct behavior

## Before (prod)

Screenshot 2024-04-15 at 11 14 10 PM
^ no para in last cell

## After (this PR)

Screenshot 2024-04-15 at 11 13 51 PM

^ has para in last cell

potatowagon:
Case: github markdown tablecopypaste

A | B | C |
---------- | ----------- | -----------
1 | 2 | 3
4 | `5` | `6`
new <br/>line | **bold** | *italic*
new <br/> line <br/>  new <br/> line | 10 | 11

^ copy directly into lexical

## Before

Screenshot 2024-04-15 at 10 05 45 PM

## After

Screenshot 2024-04-15 at 10 06 16 PM

packages/lexical-html/src/index.ts
```diff
@@ -249,6 +276,22 @@ function $createNodesFromDOM(
     childLexicalNodes = postTransform(childLexicalNodes);
   }
 
+  if (!isInlineDomNode(node)) {
```
zurfyx, line 279: I don't think we can safely assume that if it's not in the inline list then it's a block. The list is not exhaustive, there's the possibility for custom elements and there's other display types (i.e. `inline-block`). I would look up for another list that's block specific

packages/lexical-html/src/index.ts
```diff
@@ -180,6 +196,8 @@ const IGNORE_TAGS = new Set(['STYLE', 'SCRIPT']);
 function $createNodesFromDOM(
   node: Node,
   editor: LexicalEditor,
+  allArtificialNodes: Array<ArtificialNode>,
+  hasBlockAncesterLexicalNode: boolean,
```
zurfyx, line 200:
```suggestion
  hasBlockAncestorLexicalNode: boolean,
```

packages/lexical-html/src/index.ts
```diff
@@ -263,3 +306,48 @@ function $createNodesFromDOM(
...
+      node.insertAfter($createLineBreakNode());
+    }
+  });
+  allArtificialNodes.forEach(function (node, i) {
+    // unwrap children of Artificial Node and attach to parent at Artificial Node's position
+    const parent = node.getParent();
+    if (parent != null) {
+      const index = parent.getChildren().indexOf(node);
+      if (index != null) {
+        parent.splice(index, 1, node.getChildren());
```
zurfyx, line 349: Let's do a combination of `artificialNode.getPrevious` and `insertBefore`. `indexOf` and `splice` are O(n) operations with linked lists

packages/lexical/src/nodes/ArtificialNode.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+import type {EditorConfig} from 'lexical';
+
+import {ElementNode} from './LexicalElementNode';
+
+export class ArtificialNode extends ElementNode {
```
zurfyx, line 12: Let's rename this to __DO_NOT_USE. I would also add a reconciler check (at least a development mode check) that would throw or delete when an ArtificialNode is detected. We definitely want to prevent an ArtificialNode ever getting stored. Also, can you file an issue to remove this? Eventually we want to remove this

zurfyx (commented):
Great work! Do you mind adding the test cases we whiteboarded please?

packages/lexical-html/src/index.ts
potatowagon: do u mean 
```
const prevNode = artificialNode.getPreviousSibling();
// just pseudo code this wont work since insertAfter only takes 1 lexical // node and not an array of it
prevNode.insertAfter(artificialNode.getChildren())
artificialNode.remove()
```
cause artificialNode.getPrevious and insertBefore prev would be:
[A's children][Prev][A]

unless u mean iteratively inserting A's children before A, which would work too:
[Prev][A's children][A]

packages/lexical-html/src/index.ts
```diff
@@ -263,3 +306,48 @@ function $createNodesFromDOM(
...
+
+function wrapContinuousInlines(
+  domNode: Node,
+  nodes: Array<LexicalNode>,
+  createWrapperFn: () => ElementNode,
+): Array<LexicalNode> {
+  const out: Array<LexicalNode> = [];
+  let continuousInlines: Array<LexicalNode> = [];
+  // wrap contiguous inline child nodes in para
+  nodes.forEach(function (node, i) {
```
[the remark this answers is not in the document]
potatowagon: updated to for loop

packages/lexical/src/nodes/ArtificialNode.ts
potatowagon: updated to __DO_NOT_USE

packages/lexical-html/src/index.ts
potatowagon: updated to iteratively inserting A's children before A

.github/workflows/test.yml
```diff
@@ -390,7 +390,7 @@ jobs:
             ~/Library/Caches/ms-playwright
           key: ${{ runner.os }}-v${{ secrets.CACHE_VERSION }}-${{ hashFiles('package-lock.json') }}
       - name: Install dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
+        # if: steps.cache.outputs.cache-hit != 'true'
```
potatowagon, line 393: temp commit to fix ci error

potatowagon:
> > Paragraphs being block elements should not be nested in another block element, and the semantically right way to do so is to insert line breaks.
> 
> Are we aligned on these semantics? Where did this invariant come from?

i found this from:
https://stackoverflow.com/questions/10086912/why-is-table-not-allowed-inside-p

this is one example of block element (table) being nested in another block element (para) not being semantically correct. in general from http://www.w3.org/TR/html401/struct/text.html#h-9.3.1

> The P element represents a paragraph. It cannot contain block-level elements (including P itself).

packages/lexical-html/src/index.ts
potatowagon: https://github.com/facebook/lexical/blob/d04eac093ac50d37085f3b5a9f9237c988b875bd/packages/lexical/src/LexicalUtils.ts#L1605-L1611

have added this, but facing the problem where i had to add `<td>` to the list else table cell children wont be processed here:

https://github.com/facebook/lexical/blob/d04eac093ac50d37085f3b5a9f9237c988b875bd/packages/lexical-html/src/index.ts#L279-L293

if the above happens, table cell children not having missing linebreaks, this test case fails:

https://github.com/facebook/lexical/blob/d04eac093ac50d37085f3b5a9f9237c988b875bd/packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs#L1085

from this source,

https://stackoverflow.com/questions/21840505/td-element-is-a-inline-element-or-block-level-element#:~:text=From%20a%20style%20perspective%2C%20td,they're%20different%20from%20both.

`td` is neither an inline nor block element. its an inline-block. should we have another function to check for inline block? and include that in the condition, aka 

```
if (isBlockDomNode(node) || isInlineBlockDomNode) {
```

potatowagon:
> Great work! Do you mind adding the test cases we whiteboarded please?

added!

packages/lexical-html/src/index.ts
potatowagon: fixed typo

packages/lexical-html/src/index.ts
potatowagon: or we could stick to `!sInlineDomNode` since that would mean any node that is either block or neither block/inline like td

ivailop7 (approved):
LGTM! We'll revert #5951 once this is merged. @Shubhankerism

packages/lexical-html/src/index.ts
```diff
@@ -249,6 +276,22 @@ function $createNodesFromDOM(
     childLexicalNodes = postTransform(childLexicalNodes);
   }
 
+  if (isBlockDomNode(node)) {
```
zurfyx, line 279: a

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/HTMLCopyAndPaste.spec.mjs
```diff
@@ -275,4 +275,229 @@ test.describe('HTML CopyAndPaste', () => {
...
+
+    await focusEditor(page);
+
+    const clipboard = {
+      'text/html': `
+    123
+    <div>
+      456
+    </div>`,
+    };
```
zurfyx, line 290: Not a blocker but an interested follow up would be to refactor this file tests to have input / output expectation, this would reduce a lot of duplicated code.

I.e.
```
const TEST_CASES = [
  {
    input: 
    expectedOutput:
    expectedSelection
  }
]

for (const ... of TEST_CASES) {
  test(...) 
}
```

You might also want to consider creating a `good first` issue for this so that someone else can pick it up

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1584,6 +1584,32 @@ export function isHTMLElement(x: Node | EventTarget): x is HTMLElement {
...
+}
+
+/**
+ *
+ * @param node - the Dom Node to check
+ * @returns if the Dom Node is a block node
+ */
+export function isBlockDomNode(node: Node) {
+  const blockNodes = new RegExp(
+    /^(address|article|aside|blockquote|canvas|dd|div|dl|dt|fieldset|figcaption|figure|footer|form|h1|h2|h3|h4|h5|h6|header|hr|li|main|nav|noscript|ol|p|pre|section|table|td|tfoot|ul|video)$/,
```
zurfyx, line 1607: Out of curiosity, where did you get this list from?

zurfyx (approved):
Great work! Thank you for the comprehensive set of tests

packages/lexical/src/LexicalUtils.ts
potatowagon: i got it from https://www.w3schools.com/html/html_blocks.asp

Issue #5840, opened 2024-04-05
Quick start guide that requires HTML file only


StyleT:
See background here: #1845

Would be nice to have something like https://github.com/quilljs/quill#quickstart - a small, self-contained vanilla HTML/JS snippet loading the editor from CDN + also a way to grab the release tar and serve the JS contents from a static hosting (such as GitHub Pages).

Lexical already has [quick start guide](https://lexical.dev/docs/getting-started/quick-start), however it requires typescript/nodejs/npm installed locally as well as TS knowledge.

This, even more minimalistic approach would allow Lexical to be run on the websites that doesn't have any compiler/bundler.

CC @vadimkantorov - idea creator

vadimkantorov:
A somewhat working attempt of building a markdown editor (stripped off live collaboration features and made playground-derived UI much simpler and embeddable): https://github.com/vadimkantorov/lexical-playground-only

Problems with it:
1. There is only a version for ES Modules, requires usage as `<script type="module" ...>`. There should also be a simpler synchronous variant `<scrip ...>`, optionally defining some (potentially awaitable) init function
2. Markdown plugin of lexical has some issues: https://github.com/facebook/lexical/issues/5473
3. React optimizes away Lexical's classes and method names. Ideally, the LexicalComposer object model should be properly available even without React. Issue example: https://github.com/facebook/lexical/issues/5473
4. assets files are too many. There should be a variant with only lexical.js and lexical.css with all other resources embedded.
```
3-columns.4734495c.svg
EquationComponent.04e5f40a.css
EquationComponent.207a82bc.js
EquationComponent.2ae445bf.css
EquationComponent.7c5710a9.js
ExcalidrawComponent.0962d1ef.js
ExcalidrawComponent.09670137.css
ExcalidrawComponent.0c3dc28f.css
ExcalidrawComponent.d2cb4bc2.js
ImageComponent.210226de.css
ImageComponent.5c3ed33d.js
ImageComponent.8314e11b.js
ImageComponent.fc90b599.css
ImageResizer.2ff261bf.js
ImageResizer.c2987cf7.js
InlineImageComponent.72081c3a.js
InlineImageComponent.b2a19f6f.js
InlineImageComponent.b9345ede.css
InlineImageComponent.bbf49c2e.css
KaTeX_AMS-Regular.0cdd387c.woff2
KaTeX_AMS-Regular.30da91e8.woff
KaTeX_AMS-Regular.68534840.ttf
[...]

PR #5858, opened 2024-04-09
feat: use dev/prod conditions instead of webpack in exports

etrepum:
Replace the `"webpack"` condition in package.json exports with `"development"` and `"production"` to make it easier to skip consideration of the fork module. Before this change, there was no metadata that pointed directly to the dev or prod artifacts.

See also: https://webpack.js.org/guides/package-exports/#optimizations

Not quite sure if rollup, vite, etc. support this condition without configuration but they have been demonstrated to work with the fork module as-is.

I was also thinking that this would be useful for CDN usage (as one would want with #5840) since [esm.sh](https://esm.sh/#development-mode) supports conditions (Development Mode)

I also noticed the size-limit action was failing so I added some custom configuration that I think should be easier to maintain. I also changed it to run build-prod instead of build so we are measuring the artifact that matters.


etrepum:
As far as I can tell the failure is a flaky collab test. It didn't repro locally on my mac

.size-limit.js
```diff
@@ -0,0 +1,45 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+'use strict';
+const glob = require('glob');
+const path = require('node:path');
+const fs = require('fs-extra');
+
+const alias = Object.fromEntries(
```
StyleT, line 13: Pls add comment that explains purpose of this file. 
Also can you pls elaborate on why we replace simple config from `package.json` with this fairly complex file?

.size-limit.js
```diff
@@ -0,0 +1,45 @@
...
+const modifyWebpackConfig = (config) => Object.assign(config, extendConfig);
+
+function sizeLimitConfig(pkg) {
+  return {
+    path: alias[pkg],
+    modifyWebpackConfig,
+  };
+}
+
+module.exports = ['lexical', '@lexical/rich-text', '@lexical/plain-text'].map(
```
StyleT, line 43: Why only these 3 packages? Can we discover them based on the config property in their dirs/package.json? We already do glob here

scripts/updateVersion.js
```diff
@@ -64,12 +64,23 @@ function updateVersion() {
   });
 }
 
+function withExtension(fileName, ext) {
```
StyleT, line 67: Pls add JSDoc to the function. Also maybe naming `replaceExtension` would fit better...

.size-limit.js
etrepum: I think that would be an @fantactuka question (size-limit was introduced in #3600), these were the only modules measured in the previous configuration. Now that we do have more infrastructure to support this size-limit tool we have more options but the measurements are only useful if people care about them and I don't know what the team cares about especially because there is no configured limits such that the build would fail.

scripts/updateVersion.js
etrepum: I've added more comments to both modules and renamed this function accordingly

package.json
```diff
@@ -98,7 +98,7 @@
     "update-version": "node ./scripts/updateVersion",
     "postversion": "git checkout -b ${npm_package_version}__release && npm install && npm run update-version && npm run update-changelog && git add -A && git commit -m v${npm_package_version} && git tag -a v${npm_package_version} -m v${npm_package_version}",
     "release": "npm run prepare-release && node ./scripts/npm/release.js",
-    "size": "npm run build && size-limit"
+    "size": "npm run build-prod && size-limit"
```
etrepum, line 101: Previously, the size measurements were done using the dev build! If we wanted to measure both dev and prod we would change this to build-release. If we use build-release here as-is it would take slightly longer and still use the prod build (through the fork module) so there's no reason to do that now unless we want to start measuring both.

StyleT (approved):
LGTM! I will give others opportunity to review and merge

PR #5859, opened 2024-04-09
Fix bug with typeahead positioning

thegreatercurve:


acywatson (approved):

ivailop7:
That's unfortunate, because my change fixed the context menu being rendered off screen when the playground was scrolled. The context menu is now broken again when the page is scrolled. So I guess we can't have it both ways 🤷🏻‍♂️

thegreatercurve:
> That's unfortunate, because my change fixed the context menu being rendered off screen when the playground was scrolled. The context menu is now broken again when the page is scrolled. So I guess we can't have it both ways 🤷🏻‍♂️

I'll try to set aside some time and look into abstracting out `useMenuAnchorRef` into a separate hook, so we can have a separate internal version, if needed.

PR #5864, opened 2024-04-10
Revert "Revert "feat(@lexical/devtools): Added TreeView rendering instead of a simple textarea""

thegreatercurve:
Re-adds #5830 with logic for building to WWW


thegreatercurve:
CC @StyleT

zurfyx (approved):

PR #5870, opened 2024-04-11
Replace async arrow with function

zurfyx:
Intern isn't happy with async arrows, we should follow up with a lint rule like we did with coalescent.


Sahejkm (commented):
LGTM

agriffis:
What's intern?

PR #5871, opened 2024-04-11
fix: inserting a table from external resources lead Lexical crash

IAmWangyixin:
Fix https://github.com/facebook/lexical/issues/5812
**Before:**
see https://github.com/facebook/lexical/issues/5812

**After:**
tableCell tree

---
reason for crash: When press Enter the program gets stuck in a dead loop while checking for INTERNAL_$isBlock(node).

I haven't thought of a suitable way to fix the problem in lexical or lexical/table, but I think I can avoid the Lexical crash by wrapping Paragraph around elements other than paragraphs, lists, code blocks, Quote in the playground.

Please let me know if there have a better solution.


zurfyx (commented):
> but I think I can avoid the Lexical crash by wrapping Paragraph around elements other than paragraphs

This is correct, text should never be a top-level element, which does apply in this case because TableCell is a ShadowRoot. However, instead of adding a custom table cell, which will be hard to replicate by others, I would explore why this happens and whether insertNodes can do this for us as it doesn't happen for the RootNode.

packages/lexical-playground/src/nodes/TableCellNode.tsx
```diff
@@ -0,0 +1,196 @@
...
+    width?: number;
+    backgroundColor?: null | string;
+  },
+  SerializedElementNode
+>;
+
+const CustomTableCell = 'customTableCell';
+const PIXEL_VALUE_REG_EXP = /^(\d+(?:\.\d+)?)px$/;
+
+export class CustomTableCellNode extends TableCellNode {
```
potatowagon, line 68: should this file be named CustomTableCellNode instead of TableCellNode?

potatowagon:
i wonder what html is being copied onto the clipboard. i would be curious to know what (or if theres any) dom element wrapping the text node that results in it being a top level element to table cell

potatowagon:
i wonder what html is being copied onto the clipboard. i would be curious to know what (or if theres any) dom element wrapping the text node that results in it being a top level element to table cell. 

do u mind sharing what html is being copied to clipboard? u may inspect that by downloading pasteboard: https://github.com/sindresorhus/Pasteboard-Viewer 

(i learnt this from @zurfyx , thanks for sharing this useful tool!)

lastly good job debugging and finding the root cause!

IAmWangyixin:
> However, instead of adding a custom table cell, which will be hard to replicate by others, I would explore why this happens and whether insertNodes can do this for us as it doesn't happen for the RootNode.

Thanks for the review and for suggesting a better solution. I'll give it a try.

IAmWangyixin:
> do u mind sharing what html is being copied to clipboard

@potatowagon of course here the html
```
<table class="PlaygroundEditorTheme__table">
  <colgroup>
    <col>
    <col>
    <col>
    <col>
  </colgroup>
  <tbody>
    <tr>
      <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
        <span style="white-space: pre-wrap;">text</span>
      </td>
      <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
        <span style="white-space: pre-wrap;">something wrong when print enter.</span>
      </td>
      <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
        <span style="white-space: pre-wrap;">text</span>
      </td>
      <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
        <span style="white-space: pre-wrap;">5.5,5.6</span>
      </td>
    </tr>
  </tbody>
</table>
```

potatowagon:
> > do u mind sharing what html is being copied to clipboard
> 
> @potatowagon of course here the html
> ```
> <table class="PlaygroundEditorTheme__table">
>   <colgroup>
>     <col>
>     <col>
>     <col>
>     <col>
>   </colgroup>
>   <tbody>
>     <tr>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">text</span>
>       </td>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">something wrong when print enter.</span>
>       </td>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">text</span>
>       </td>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">5.5,5.6</span>
>       </td>
>     </tr>
>   </tbody>
> </table>
> ```

https://github.com/facebook/lexical/pull/5857 may fix this as tgat PR would wrap inline text nodes in a paragraph if theres nothing between the inline node and the root (or shadow root in this case for `td`

will be testing your case on that pr

potatowagon:
> > do u mind sharing what html is being copied to clipboard
> 
> @potatowagon of course here the html
> ```
> <table class="PlaygroundEditorTheme__table">
>   <colgroup>
>     <col>
>     <col>
>     <col>
>     <col>
>   </colgroup>
>   <tbody>
>     <tr>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">text</span>
>       </td>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">something wrong when print enter.</span>
>       </td>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">text</span>
>       </td>
>       <td class="PlaygroundEditorTheme__tableCell" style="width: 175px; border: 1px solid black; vertical-align: top; text-align: start;">
>         <span style="white-space: pre-wrap;">5.5,5.6</span>
>       </td>
>     </tr>
>   </tbody>
> </table>
> ```

just realised this is the copied html of the table in lexical editor, could u share instead the copied html after copy from the external app?

IAmWangyixin:
> just realised this is the copied html of the table in lexical editor, could u share instead the copied html after copy from the external app?

You're right, what I got before was the html after saving on lexical, I think the following is the original pasted html.

```
<table style="width: auto; text-align: start">
  <tbody>
    <tr>
      <td
        colspan="1"
        rowspan="1"
        width="auto"
        style="text-align: start; line-height: 1.5">
        text
      </td>
      <td
        colspan="1"
        rowspan="1"
        width="auto"
        style="text-align: start; line-height: 1.5">
        something wrong when print enter.
      </td>
      <td
        colspan="1"
        rowspan="1"
        width="auto"
        style="text-align: start; line-height: 1.5">
        text
      </td>
      <td
        colspan="1"
        rowspan="1"
        width="auto"
        style="text-align: start; line-height: 1.5">
        5.5,5.6
      </td>
    </tr>
  </tbody>
</table>
```

ivailop7:
I wonder if the crash is fixed by https://github.com/facebook/lexical/pull/6139 since the example here all have TBODY tag, which we didn't respect in the TableObserver. Thoughts @potatowagon @zurfyx ?

potatowagon:
> I wonder if the crash is fixed by #6139 since the example here all have TBODY tag, which we didn't respect in the TableObserver. Thoughts @potatowagon @zurfyx ?

afaik this issue is solved already

what could have fixed the crash was text nodes inside of a <td> now being wraped in a <p> on pasting, which wont lead to crash

potatowagon:
closing since this issue is solved

Issue #5869, opened 2024-04-10
Improve & document NPM package addition process


StyleT:
At this moment addition of the new NPM package to Lexical monorepo requires repetitive updates to ~10 different files. On top of that Meta specific WWW build requires separate changes done to internal codebase in order to keep it running.

**Example:**
In #5861 we reverted #5830 that added `@lexical/devtools-core` package. Later it was re-added in #5864 with fixes done to the build system.

**Scope:**
- Document steps necessary to add new NPM package to Lexical monorepo. Specifically:
  - PR description format to highlight new package addition
  - Changes to the shared files necessary
  - Merge procedure
- Reduce number of changes needed to add new package

etrepum:
I'd be happy to work on this if you want to delegate, I have been touching the build and documentation a lot lately. See also #5858 and #5850

etrepum:
I think if we pull out the code I have in those two PRs into shared modules we can use that across the board to handle basically all of the hard-coded mess across the build scripts

* [Discover entry points for all non-private packages](https://github.com/facebook/lexical/pull/5858/files#diff-f0bf0e4e8091545d3336d832dffef6bed18bef7bd9c3e2a71b9fa44c9e1c67daR13-R31) - this would need some extra stuff to deal with @lexical/react which has multiple entry points and no barrel file (which is basically already in the below script)
* [Discover entry points for lexical/react](https://github.com/facebook/lexical/pull/5850/files#diff-26dec96b8d49038db55d3673e9422c82c00e67e7c819da365390db4ef697e468R32-R49)

StyleT:
@etrepum feel free to jump on it! I can later add Meta specific WWW build steps

etrepum:
@StyleT will do, if you can get those PRs approved/merged (or at least one of them?) it would make it a little easier for me since AFAIK you can only stack on top of at most one of them as a base branch with the github workflow

PR #5876, opened 2024-04-11
Refactor build scripts and npm package process

etrepum:
Addresses refactoring of npm/build processes per #5869 

- [x] Adds some tests to sanity check packages (their package.json, the existence of flow and docs)
- [x] Refactors basically all of scripts/* to use a generated model for the available packages
- [x] Update scripts/www/rewriteImports.js to use the new infrastructure
- [x] Generate tsconfig.json paths (npm run update-tsconfig)
- [x] Generate .flowconfig paths (npm run update-flowconfig)
- [x] Generate boilerplate `README.md` and `docs/packages/{package}.md` files (npm run update-docs)
- [x] jest config
- [x] viteModuleResolution
- [x] docusaurus config
- [x] Started a [Maintainers' Guide](https://lexical-mzg4hwe1m-fbopensource.vercel.app/docs/maintainers-guide) section in the docs (currently fairly hidden, only referenced in the repo README)
- [x] Adds build integration tests to make sure all of the examples and several other framework scenarios (astro, svelte, next.js) build and run with release artifacts

RE #5869 this mostly addresses the "Reduce number of changes needed to add new package" - since I have not added a new package myself yet I think another maintainer (@StyleT?) could add those docs to the guide using this as a baseline?

I was also thinking about maybe starting on an eslint plugin in the next week or two to help with the "rules of $functions" so if someone else doesn't get to it first I could add that documentation then when I am more familiar with all of those details.

Closes #5940 (parts of this PR were extracted there to maybe expedite review)


etrepum:
@StyleT I wanted to put this up earlier on in case you had any feedback on this approach

scripts/www/rewriteImports.js
```diff
@@ -8,72 +8,78 @@
 
 'use strict';
 
+const fs = require('fs-extra');
 const glob = require('glob');
-const {readFile, writeFile} = require('fs');
+const path = require('node:path');
+const {packagesManager} = require('../shared/packagesManager');
+const npmToWwwName = require('./npmToWwwName');
+const {t, transform} = require('hermes-transform');
```
etrepum, line 16: Oh boy it was a journey to find a way to correctly transform modern flow files. ast-types is not up to date with flow's features so jscodeshift/recast can't print transformed code. I was lucky to find this used by react-native.

scripts/www/rewriteImports.js
```diff
@@ -8,72 +8,78 @@
...
+          context.replaceNode(node.source, t.StringLiteral({value}));
+        }
+      },
+      Program(node) {
+        if (
+          node.docblock &&
+          node.docblock.comment &&
+          node.docblock.comment.value.includes('@flow strict')
+        ) {
+          node.docblock.comment.value = node.docblock.comment.value.replace(
```
etrepum, line 54: I don't think it's correct to mutate this in-place, but I couldn't quickly figure out a mutator incantation that didn't fail… and this produced the correct output. Can't be worse than some regexes right? 😆 

Maybe someone inside meta could refactor this to do it the right way.

tsconfig.json
```diff
@@ -12,212 +12,207 @@
...
+      "@lexical/text/src": ["./packages/lexical-text/src"],
+      "@lexical/utils/src": ["./packages/lexical-utils/src"],
+      "@lexical/yjs/src": ["./packages/lexical-yjs/src"],
+      "shared/src": ["./packages/shared/src"]
     }
   },
   "include": ["./libdefs", "./packages"],
   "exclude": [
+    "./libdefs/*.js",
     "**/lexical-*/src/__tests__/**",
```
etrepum, line 210: I'd like to turn these on but the tests have all sorts of broken types and fixing them would be some big scope creep!

packages/lexical-website/docs/maintainers-guide.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Package naming conventions
+
+### Overall
+
+| Usage | Convention |
+| -- | -- |
+| Directory name | `packages/lexical-package-name` |
+| Entrypoint | `packages/lexical-package-name/index.ts` |
+| Flow types | `packages/lexical-package/flow/LexicalPackage.js.flow` |
```
etrepum, line 44: Does anyone happen to know if those `packages/lexical-package/LexicalPackage.js` files are used for anything? Is that for www tooling? They would be easy to generate but I think it would be even better to get rid of them :)

They are the files that boil down to `module.exports = require('./dist/LexicalPackage.js');`

StyleT (commented):
Hi! And thanks for this work!

1. Do you think we can split this PR onto a few smaller ones? I see changes here that are seemingly unrelated (like creation of the maintainers' guide). Dunno how we can merge it all at one w/o taking a leap of faith ;P 
2. I would prefer to keep all that is related to the Lexical contribution within `[CONTRIBUTING.md](https://github.com/facebook/lexical/blob/main/CONTRIBUTING.md)` Unless we have a strong reason to do it differently
3. This PR seemingly misses TODO steps on how to add a new package

etrepum:
@StyleT

1. In a way, the maintainers guide is documentation about the tools that this PR either updates or adds. I could add that file separately, but I'm not sure what the point would be. I could certainly split up the work into multiple commits, e.g. "all of the tools" and "automated modifications to configuration" could be separate, but otherwise this is all rather intertwined.
2. The primary reason I chose to put it in lexical-website was that the CONTRIBUTING.md isn't rendered anywhere on lexical.dev, and the intended audience for this document are people working a lot closer to the build infrastructure than almost all contributors would need to worry about. It seemed like a better idea to keep CONTRIBUTING.md short so it's not intimidating.
3. Per creating a new module, that I would either like someone else to work on, or I will handle that separately in the course of creating a new module. I haven't added a new module yet, so I am not confident that I have all of the required steps in order, especially things that are outside of the source tree like the PR template and other meta concerns. There is some commentary about that in the description of this PR.

scripts/__tests__/unit/build.test.js
```diff
@@ -0,0 +1,116 @@
...
+          ),
+        ).toEqual([]);
+      });
+      it('monorepo dependencies must use the exact monorepo version', () => {
+        Object.entries(dependencies).forEach(([dep, version]) => {
+          if (publicNpmNames.has(dep)) {
+            expect([dep, version]).toEqual([dep, monorepoPackageJson.version]);
+          }
+        });
+      });
```
etrepum, line 66: This is a test that would've caught the version issue of #5901

scripts/npm/prepare-release.js
```diff
@@ -10,46 +10,48 @@
...
-    await exec(`cp -R LICENSE ./packages/${pkg}/npm`);
-    await exec(`cp -R ./packages/${pkg}/README.md ./packages/${pkg}/npm`);
-  }
+/**
+ * Prepares the npm directories for all public packages. If a package should
+ * not be published, make sure to set `"private": true` in its package.json.
+ */
+function preparePublicPackages() {
+  packagesManager.getPublicPackages().forEach(preparePackage);
 }
```
etrepum, line 55: This is the part that would've published @lexical/devtools-core (re #5901)

etrepum:
happy to rebase and fix conflicts on this one when someone gives me a heads-up (on here or discord) that they're ready to review and consider for merge, I don't want to have to do it too many times

packages/lexical-website/docs/maintainers-guide.md
```diff
@@ -0,0 +1,244 @@
...
+| -- | -- |
+| Directory name | `packages/lexical-package-name` |
+| Entrypoint | `packages/lexical-package-name/src/index.{ts,tsx}` |
+| Flow types | `packages/lexical-package/flow/LexicalPackageName.js.flow` |
+| package.json name | `@lexical/package-name` |
+| Documentation | `packages/lexical-package-name/README.md` |
+| Unit Tests | `packages/lexical-package-name/src/__tests__/unit/LexicalPackageName.test.{ts,tsx}` |
+| dist (gitignore'd build product) | `packages/lexical-package-name/dist` |
+| npm (gitignore'd prerelease product) | `packages/lexical-package-name/npm` |
+| www entrypoint? | `packages/lexical-package-name/LexicalPackageName.js` |
```
acywatson, line 50: yes

scripts/clean.js
```diff
@@ -9,12 +9,13 @@
...
-  fs.removeSync(path.resolve(`./.ts-temp`));
-  fs.removeSync(path.resolve(`./packages/${pkg}/dist`));
-  fs.removeSync(path.resolve(`./packages/${pkg}/npm`));
-});
+fs.removeSync(path.resolve(`./npm`));
+fs.removeSync(path.resolve(`./.ts-temp`));
+packagesManager
+  .getPublicPackages()
+  .forEach((pkg) =>
+    ['dist', 'npm'].forEach((subdir) => fs.removeSync(pkg.resolve(subdir))),
```
acywatson, line 20: don't we cover npm on line 15?

acywatson:
This looks good to me - huge improvement.

I honestly don't know if all the WWW and flow stuff is right, but I think the fastest way to tell is to merge and fix any issues that come up in the sync.

I'll run a sync internally ASAP, then we can merge this and do a sync with just this PR so we can catch any issues.

acywatson (approved):

packages/lexical-website/docs/maintainers-guide.md
etrepum: great, just added a test to check for them and a script to create the stubs

scripts/clean.js
etrepum: no, line 15 covers the monorepo's npm. This is something I added in order to facilitate the build integration tests. It hosts the `npm pack` output from each of the public packages

```
❯ ls npm
lexical-0.14.5.tgz               lexical-hashtag-0.14.5.tgz       lexical-mark-0.14.5.tgz          lexical-rich-text-0.14.5.tgz
lexical-clipboard-0.14.5.tgz     lexical-headless-0.14.5.tgz      lexical-markdown-0.14.5.tgz      lexical-selection-0.14.5.tgz
lexical-code-0.14.5.tgz          lexical-history-0.14.5.tgz       lexical-offset-0.14.5.tgz        lexical-table-0.14.5.tgz
lexical-devtools-core-0.14.5.tgz lexical-html-0.14.5.tgz          lexical-overflow-0.14.5.tgz      lexical-text-0.14.5.tgz
lexical-dragon-0.14.5.tgz        lexical-link-0.14.5.tgz          lexical-plain-text-0.14.5.tgz    lexical-utils-0.14.5.tgz
lexical-file-0.14.5.tgz          lexical-list-0.14.5.tgz          lexical-react-0.14.5.tgz         lexical-yjs-0.14.5.tgz
```

packages/lexical-website/docs/maintainers-guide.md
acywatson: Oh yea but do you need the question mark in there now that you know lol

etrepum: it got removed but github always shows comments in the context of the commit they were written about https://github.com/facebook/lexical/pull/5876/commits/1e2a4960c330fd14491f2f09756023c3f15dfeb5#diff-d575e613cfb9cf9256293e9cb0f838e5b4ddcf3890f11aed370bbc03f15cd3bcR50

scripts/update-docs.js
```diff
@@ -0,0 +1,79 @@
...
+    # \`${npmName}\`
+
+[![See API Documentation](https://lexical.dev/img/see-api-documentation.svg)](https://lexical.dev/docs/api/modules/${apiModuleName})
+
+${description}
+`.trim() + '\n'
+  );
+}
+
+function updateDocs() {
```
StyleT, line 43: So as far as I understand this script doesn't update anything but rather creates files only. I see the following issues here:

1. It can't keep existing docs consistent
2. We run it at a wrong time. `update-packages` is a bit too late as "invalid" package already got into `main` branch. And it wouldn't help package creator to see that this script shall be run (if I understand it correctly)

I think it's better to opt for Docusaurus plugin and generate it on the fly (same approach we use for API docs). Plus extend `scripts/__tests__/unit/build.test.js` to verify that all public packages have README

Here is a crude example of how this can be done: https://github.com/facebook/lexical/compare/main...feature/docs_packages_autogen
It lacks watch mode + ability to set package order + custom sidebar name w/o affecting original README.md
It was inspired by https://github.com/tgreyuk/typedoc-plugin-markdown/blob/master/packages/docusaurus-plugin-typedoc/src/plugin.ts#L18

StyleT: btw if we keep it as is - may be better to rename it to "create-docs"

scripts/update-tsconfig.js
```diff
@@ -0,0 +1,110 @@
...
+    filepath: jsonFileName,
+  });
+  if (prevTsconfigContents !== nextTsconfigContents) {
+    fs.writeFileSync(jsonFileName, nextTsconfigContents);
+  }
+}
+
+async function updateAllTsconfig() {
+  const prettierConfig = (await prettier.resolveConfig('./')) || {};
+  await updateTsconfig({
```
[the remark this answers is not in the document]
etrepum: That's fine, we can add more files if that's the preferred style. I was just following the existing convention.

scripts/__tests__/integration/prepare-release.test.js
```diff
@@ -0,0 +1,31 @@
...
+        'npm',
+        `${pkg.getDirectoryName()}-${pkg.packageJson.version}.tgz`,
+      );
+      test(`${tgzPath} exists`, () =>
+        expect(fs.existsSync(tgzPath)).toBe(true));
+    });
+  }
+});
+['examples', 'scripts/__tests__/integration/fixtures']
+  .flatMap((packagesDir) => glob.sync(`${packagesDir}/*/package.json`))
```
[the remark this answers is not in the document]
etrepum: They are real application packages, but not published to npm. Adding them as zip files would be something that could be done but then it is even harder to see why and how the tests broke. We can add them to examples but I did not do that here because I didn't want to have the discussions about whether these are officially endorsed or not since there are third party packages for some of these frameworks that offer better integration.

scripts/__tests__/integration/fixtures/lexical-esm-nextjs/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "test": "playwright test"
+  },
+  "dependencies": {
+    "@lexical/plain-text": "^0.14.5",
```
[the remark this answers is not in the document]
etrepum: I did not take that approach because the same test code is used to verify that the examples work with build artifacts, which are necessarily projects that can be extracted from the monorepo and run elsewhere.

.prettierignore
```diff
@@ -1,14 +1,16 @@
...
 
-packages/**/dist/**
-packages/**/build/**
-packages/**/npm/**
-packages/**/config/**
+**/dist/**
+**/build/**
+**/npm/**
+**/__tests__/integration/fixtures/**
+**/config/**
```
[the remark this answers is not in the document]
etrepum: I have no opinion on config folders, this was just keeping the ignore files in sync between eslint and prettier

scripts/update-docs.js
etrepum: update-packages is supposed to be run before anything is merged to main, it is not too late, when creating a new project you would also need it to generate other configuration. If you see the WIP eslint PR it has some documentation about this https://github.com/facebook/lexical/blob/690f73920a9314805bef9f9ceb096e3eda5f4e00/packages/lexical-website/docs/maintainers-guide.md#creating-a-new-package

We could add a docs autogen plugin but I was just going with the existing conventions.

Either way, watch doesn't work when editing those README files because the markdown transclusion doesn't support watch of the transcluded files (this predated the work that I have done).

etrepum: Oh and there are tests that will fail if you try and add a new package without somehow creating all of the required boilerplate files, so it would be hard to land something on main before running update-packages or equivalent.

.prettierignore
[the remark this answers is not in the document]
etrepum: Removed from both in 1614a6047575ce7a980ace2ca8cc8209bfcd7165 since there are no config folders to begin with

scripts/update-tsconfig.js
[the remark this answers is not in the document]
etrepum: Extracted to tsconfig*.paths.json 224e201808b701337bad2a2dedfec4b4ca645be3

scripts/__tests__/integration/prepare-release.test.js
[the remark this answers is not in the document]
StyleT: > They are real application packages, but not published to npm. Adding them as zip files would be something that could be done but then it is even harder to see why and how the tests broke. We can add them to examples but I did not do that here because I didn't want to have the discussions about whether these are officially endorsed or not since there are third party packages for some of these frameworks that offer better integration.

Makes sense. Then can we give them more of a semantic naming so it's clear what exactly we test with every one of them. I'm puzzled why exactly svelte/astro/etc is important here.

scripts/update-tsconfig.js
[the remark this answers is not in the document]
StyleT: > 224e201

Awesome! I would still suggest to name it `tsconfig*.generated.json` to emphasize that they are not manually written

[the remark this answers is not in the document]
etrepum: Had to do a follow-up fix since the jest config expected a specific layout of tsconfig.json 262fed01f88c4d7e3bbc525c115016f9eea259b2

scripts/update-docs.js
etrepum: renamed update-docs and update-www-stubs to create-docs and create-www-stubs in ddf3d3fb73561a408dde6e9dd9fdcbc78abee739

StyleT: > Either way, watch doesn't work when editing those README files because the markdown transclusion doesn't support watch of the transcluded files (this predated the work that I have done).

So if you use approach similar to https://github.com/facebook/lexical/compare/main...feature/docs_packages_autogen with watch implemented in a way similar to https://github.com/tgreyuk/typedoc-plugin-markdown/blob/master/packages/docusaurus-plugin-typedoc/src/plugin.ts#L97 - it will work. As then you don't need markdown transclusion.

Btw this lack of watch support caused me to lose few hours on debugging few weeks ago. But I wasn't paying attention to it.

P.S.: I'm ok to do it as follow up for sure

scripts/__tests__/integration/prepare-release.test.js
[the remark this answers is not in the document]
acywatson: > I'm puzzled why exactly svelte/astro/etc is important here.

Tons of people use these frameworks in open source

scripts/update-tsconfig.js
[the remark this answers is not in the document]
etrepum: it seems that esbuild does not support having an array for the extends argument so we can't have the build json extend both the main tsconfig and a generated file, so I will roll back those changes and someone can revisit splitting up the generated tsconfig separately.

```
[vite:esbuild] The "path" argument must be of type string. Received an instance of Array
file: /Users/bob/src/lexical/packages/lexical-playground/src/index.tsx
error during build:
TypeError: The "path" argument must be of type string. Received an instance of Array
    at validateString (node:internal/validators:162:11)
    at Object.isAbsolute (node:path:1160:5)
    at resolveExtends (/Users/bob/src/lexical/node_modules/vite/dist/node/chunks/dep-6e2fe41e.js:19612:22)
    at parseExtends (/Users/bob/src/lexical/node_modules/vite/dist/node/chunks/dep-6e2fe41e.js:19593:34)
    at parse$f (/Users/bob/src/lexical/node_modules/vite/dist/node/chunks/dep-6e2fe41e.js:19544:24)
    at async loadTsconfigJsonForFile (/Users/bob/src/lexical/node_modules/vite/dist/node/chunks/dep-6e2fe41e.js:19933:24)
    at async transformWithEsbuild (/Users/bob/src/lexical/node_modules/vite/dist/node/chunks/dep-6e2fe41e.js:19736:36)
    at async Object.transform (/Users/bob/src/lexical/node_modules/vite/dist/node/chunks/dep-6e2fe41e.js:19821:32)
    at async transform (/Users/bob/src/lexical/node_modules/rollup/dist/shared/rollup.js:22085:16)
    at async ModuleLoader.addModuleSource (/Users/bob/src/lexical/node_modules/rollup/dist/shared/rollup.js:22311:30)
```

scripts/__tests__/integration/fixtures/lexical-esm-nextjs/package.json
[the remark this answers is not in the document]
etrepum: The integration tests do verify that the a correct version of lexical was installed, and updating the package.json or package-lock.json are not really necessary in this scenario (unless updating the other dependencies, e.g. the frameworks) since we are essentially linking our versions of the monorepo package artifacts into node_modules.

https://github.com/facebook/lexical/pull/5876/files#diff-19bab06b607c228fd4b4b2a97b16cbc72e6ad4b656771cdf497172152339ee64R124-R140

scripts/update-docs.js
etrepum: I also lost some time to the lack of watch support for README.md files and resolving it is something I would work on if you don't get to it first :)

scripts/__tests__/integration/prepare-release.test.js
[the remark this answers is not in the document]
etrepum: I would be open to cleaning up the test fixtures for these frameworks and making them "full" examples, but the goal of adding them here was to prevent future compatibility regressions since I've caused and fixed so many along the way to getting esm support. It is really hard to test so many combinations of frameworks that all have their own special way of using webpack/swc/esbuild/vite/etc. unless it's already in the monorepo test suite.

Angular would be a good one to do as a follow-up. The choice of next.js, sveltekit and astro were made because we are not testing them with examples in-tree (we are only testing a basic vite config with and without react), and we have gotten issues or discord problem reports about those frameworks specifically. I had them sitting around for my own testing.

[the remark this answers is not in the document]
etrepum: What we are testing with them right now is simple:
* does lexical build with this framework
* did lexical render (tested by starting the build product & checking with playwright)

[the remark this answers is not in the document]
StyleT: Ok, got it. Sounds good to me.

Out of curiosity:
Can you give some example of any change to Lexical that won’t break vanilla js integration but will break any of these “non-React” frameworks

StyleT (approved):

scripts/__tests__/integration/prepare-release.test.js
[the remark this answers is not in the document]
etrepum: The regressions that I noticed were mostly around changes to the package.json files and the fork modules. Rather than vanilla js vs. non-react frameworks it's usually of a question of vite vs swr vs webpack plus whatever configuration to those bundlers come with the framework.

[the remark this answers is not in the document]
etrepum: #5823 is a concrete next.js issue that this test would've caught

PR #5881, opened 2024-04-13
Add h4 h5 h6 support to playground

newproplus:
Add h4 h5 h6 support to ToolbarPlugin and TableOfContentsPlugin

### DIFF
--- a/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
+++ b/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
@@ -1,9 +1,17 @@
-.table-of-contents .heading2 {
+.table-of-contents .heading3 {
   margin-left: 10px;
 }
 
-.table-of-contents .heading3 {
+.table-of-contents .heading4 {
   margin-left: 20px;
+}
+
+.table-of-contents .heading5 {
+  margin-left: 30px;
+}
+
+.table-of-contents .heading6 {
+  margin-left: 40px;
 }
 
 .selected-heading {
--- a/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
+++ b/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
@@ -57,6 +57,7 @@
   color: black;
   font-weight: bold;
   cursor: pointer;
+  text-transform: uppercase;
 }
 
 .headings {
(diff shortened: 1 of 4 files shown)


potatowagon:
hi! could u add screenshots of the before and after results? before could show what h1 to h3, and after could show h1 to h6

packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
```diff
@@ -1,11 +1,19 @@
-.table-of-contents .heading2 {
```
potatowagon, line 1: wont this break styling for heading 2?

style for heading2 deleted, but it is still used here: https://github.com/facebook/lexical/blob/616679a48175881c2a0f7970bbb438c59cbb8168/packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.tsx#L24

potatowagon (changes requested):

packages/lexical-playground/src/plugins/TableOfContentsPlugin/index.css
newproplus: The style of heading 2 here only has a 10px indentation. In order to shrink the indentation of all headings, especially for h6, I made heading 2 without any indentation.
I have standardized the styles of h1 to h6, making h1 have a larger font size. Additionally, the font sizes of h1 to h6 decrease gradually, ensuring that even h6 is still larger than regular text.


The previous image:



The image of my version:

newproplus:
I have standardized the styles of h1 to h6, making h1 have a larger font size. Additionally, the font sizes of h1 to h6 decrease gradually, ensuring that even h6 is still larger than regular text.

In order to shrink the indentation of all headings, especially for h6, I made heading 2 without any indentation.

The before image:



The after image:

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -28,20 +28,38 @@
   padding-left: 16px;
 }
 .PlaygroundEditorTheme__h1 {
-  font-size: 24px;
-  color: rgb(5, 5, 5);
```
potatowagon, line 32: this changes all the existing playground heading styles. the headings have been styled that way since https://github.com/facebook/lexical/commit/8b3945ef899b2dfe5abe849eca8e96ee3b9be30b

will defer to others who have more context if changing the current playground heading styles is ok. there could have been reasons why the current heading styles are as such

newproplus: Due to uncertainty about whether the code here can be modified. I closed this PR.

PR #5884, opened 2024-04-13
Remove unnecessary type assertions

2wheeh:
I believe this is better for type-safety

### DIFF
--- a/packages/lexical-playground/src/hooks/useReport.ts
+++ b/packages/lexical-playground/src/hooks/useReport.ts
@@ -36,8 +36,8 @@ export default function useReport(): (
 ) => ReturnType<typeof setTimeout> {
   const timer = useRef<ReturnType<typeof setTimeout> | null>(null);
   const cleanup = useCallback(() => {
-    if (timer !== null) {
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/hooks/useReport.ts
```diff
@@ -36,8 +36,8 @@ export default function useReport(): (
 ) => ReturnType<typeof setTimeout> {
   const timer = useRef<ReturnType<typeof setTimeout> | null>(null);
   const cleanup = useCallback(() => {
-    if (timer !== null) {
-      clearTimeout(timer.current as ReturnType<typeof setTimeout>);
+    if (timer.current !== null) {
+      clearTimeout(timer.current);
```
etrepum, line 40: While you're in here, semantically it would make sense to also set `timer.current = null;` after clearing, although the `clearTimeout` API doesn't really care if it's called with invalid ids

2wheeh: Thanks ! I added on 153cd4c

potatowagon (approved):
lgtm

PR #5886, opened 2024-04-13
Minor refactor on ListMaxIndentLevelPlugin

2wheeh:
Minor refactors for better readability:
- `type Props` doesn't need to be this verbose. It might have been auto-generated (?) during a migration to TS.
- `shouldPreventIndent` without negation is even more intuitive than `!isIndentPermitted`.

### DIFF
--- a/packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
+++ b/packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
@@ -6,7 +6,7 @@
  *
  */
 
-import type {RangeSelection} from 'lexical';
+import type {ElementNode, RangeSelection} from 'lexical';
 
 import {$getListDepth, $isListItemNode, $isListNode} from '@lexical/list';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
@@ -15,14 +15,9 @@ import {
   $isElementNode,
   $isRangeSelection,
   COMMAND_PRIORITY_CRITICAL,
-  ElementNode,
   INDENT_CONTENT_COMMAND,
 } from 'lexical';
 import {useEffect} from 'react';
-
-type Props = Readonly<{
-  maxDepth: number | null | undefined;
-}>;
 
 function getElementNodesInSelection(
   selection: RangeSelection,
@@ -41,7 +36,7 @@ function getElementNodesInSelection(
   );
 }
 
-function isIndentPermitted(maxDepth: number): boolean {
+function shouldPreventIndent(maxDepth: number): boolean {
   const selection = $getSelection();
 
   if (!$isRangeSelection(selection)) {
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
```diff
@@ -6,7 +6,7 @@
  *
  */
 
-import type {RangeSelection} from 'lexical';
+import type {ElementNode, RangeSelection} from 'lexical';
```
potatowagon, line 9: looks good

packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
```diff
@@ -69,16 +64,20 @@ function isIndentPermitted(maxDepth: number): boolean {
...
-  return totalDepth <= maxDepth;
+  return totalDepth > maxDepth;
 }
 
-export default function ListMaxIndentLevelPlugin({maxDepth}: Props): null {
+export default function ListMaxIndentLevelPlugin({
+  maxDepth = 7,
+}: {
+  maxDepth?: number;
+}): null {
```
potatowagon, line 74: looks good

packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
```diff
@@ -69,16 +64,20 @@ function isIndentPermitted(maxDepth: number): boolean {
...
   const [editor] = useLexicalComposerContext();
 
   useEffect(() => {
     return editor.registerCommand(
       INDENT_CONTENT_COMMAND,
-      () => !isIndentPermitted(maxDepth ?? 7),
+      () => shouldPreventIndent(maxDepth),
```
potatowagon, line 80: [nit + opinion] im leaning towards sticking to isIndentPermitted, as  negating isIndentPermitted will not lead to double negation. for shouldPreventIndent, since the negation is now baked into the function name : (should prevent) -> the negation, Indent -> the action, this opens up the opportunity for !shouldPreventIndent which is a double negation and looks confusing

potatowagon (approved):
stamping as most of the changes in this diff makes sense

packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
2wheeh: I understand your concern about potential double negation with `shouldPreventIndent`. 
But It's not just changing the name but the return value as well. 
```diff
_  return totalDepth <= maxDepth;
+  return totalDepth > maxDepth;
```

With the current `isIndentPermitted`, we're essentially negating the returned value to decide whether to propagate the action. 

FWIW, This seems a form of double negation already. 
When I read current code in my mental:
 - it's not `isIndentPermitted` -> action 
 - But it's `!isIndentPermitted` === `false` -> action

```ts
if ((!isIndentPermitted()) === true) {
  // prevent action
else {
  // do action
}
```
On the other hand, with `shouldPreventIndent`, the returned value is passed as is, without the need for negation.
This directly communicates the intent of the function - should the action be prevented or not. 

```ts
if (shouldPreventIndent() === true) {
  // prevent action
} 
else {
  // do action
}
```

potatowagon: sure feel free to merge! this is mostly a linguistic preference thing and can be pretty subjective, so dont let it be a blocker. appreciate the refactoring in this PR

PR #5889, opened 2024-04-15
fix(lexical/createEditor): avoid implicit any type

jason89521:
## Why

We should avoid implicit `any` type because it may cause problems in the future.

## What

Explicitly setup the type of the `replace` and `replaceWithKlass` when creating them.

## Additional information

before
image


potatowagon (approved):
👍

StyleT:
LGTM!

PR #5890, opened 2024-04-15
Fix broken style for node overrides example on docs

2wheeh:
Styling for the example on [node-replacement](https://lexical.dev/docs/concepts/node-replacement) is broken.

### Before
image

### After
image


potatowagon (approved):
very nice thanks for this fix 👍

PR #5891, opened 2024-04-15
Remove a special case in table deleteTextHandler

AlexanderReznik:
This was introduced in https://github.com/facebook/lexical/pull/3213 but wasn't cleaned up after a proper fix.
Currently this does not allow the removal of empty lines in a collapsible section inside of a table.

Before:
table-collapsible-paragraph

After:
table-collapsible-paragraph-after

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -26,6 +26,7 @@ import {
   focusEditor,
   html,
   initialize,
+  insertCollapsible,
   insertHorizontalRule,
   insertSampleImage,
   insertTable,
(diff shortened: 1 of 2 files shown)


StyleT:
Hi! Thanks for this fix! Do you think we can write a test for it?

zurfyx (approved):

AlexanderReznik:
> Hi! Thanks for this fix! Do you think we can write a test for it?

Sure! https://github.com/facebook/lexical/pull/5891/commits/d95d99f6bed4a8d60775400b712bb96c01d49891

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -1301,6 +1302,95 @@ test.describe('Tables', () => {
...
+              class="PlaygroundEditorTheme__tableCell PlaygroundEditorTheme__tableCellHeader">
+              <p class="PlaygroundEditorTheme__paragraph"><br /></p>
+            </th>
+          </tr>
+        </table>
+        <p class="PlaygroundEditorTheme__paragraph"><br /></p>
+      `,
+    );
+  });
+
```
potatowagon, line 1393: 👍

potatowagon (approved):

ivailop7 (approved):

PR #5894, opened 2024-04-15
feat(@lexical/devtools): Extracted store/messaging implementation to the separate packages

StyleT:
Reasoning:
- This adds strongly types RPC API for communication between extension surfaces (devtools / injected script / popup / background)
- It moves store synchronization implementation to the unified transport layer (instead of relying on ad-hoc patches of the existing libraries)
- We remove code that is unrelated to the Lexical DevTools from the repo to streamline communication stack development and keep examples/tests/infra outside of Lexical. So errors like this one: https://github.com/serversideup/webext-bridge/pull/83 can be fixed outside of Lexical.


.eslintrc.js
```diff
@@ -126,7 +126,6 @@ module.exports = {
   rules: {
     'accessor-pairs': OFF,
 
-    'brace-style': [ERROR, '1tbs'],
```
StyleT, line 129: Prettier already enforces this. Removal of this rules prevents eslint and prettier from contradicting each other

fantactuka (approved):
Looks great, nice that you don't have to drag messaging methods through components

Issue #5901, opened 2024-04-17
Bug: @lexical/react 0.14.4 brings in @lexical/devtools-core dependency without declaring it


etrepum:
Lexical version: 0.14.4

## Steps To Reproduce

1. Update a project to @lexical/react 0.14.4 without adding @lexical/devtools-core as a dependency
2. Try and build or run the project

Link to code example:

https://github.com/etrepum/lexical-esm-nextjs but anything that imports @lexical/react/LexicalTreeView should demonstrate the problem

## The current behavior

Tries to import from @lexical/devtools-core which is not published yet

## The expected behavior

It should not import @lexical/devtools-core, or that package should be published and @lexical/react should have a dependency on it

## Root cause

[blame of packages/lexical-react/src/LexicalTreeView.tsx](https://github.com/facebook/lexical/blame/main/packages/lexical-react/src/LexicalTreeView.tsx) shows that this was last added in this commit: https://github.com/facebook/lexical/commit/c0cca39759172d5fa51f49842f39ab75c7adc767

StyleT:
It's a bummer :( Let me take a look

etrepum:
#5876 should take care of the build and publish scripts (and the fix to update-version to handle devDependencies is in there too). There is of course some risk in making large changes to them, but the build is broken or at least fragile anyway…

etrepum:
The v0.14.5 release fixes this with next.js, I will verify that the other minimal framework examples still work and confirm shortly

etrepum:
Confirmed that v0.14.5 is working with my sveltekit and astro examples as well! Closing as fixed 💯

PR #5902, opened 2024-04-17
fix(@lexical/react): @lexical/react 0.14.4 brings in @lexical/devtools-core dependency without declaring it

StyleT:
See #5901

### DIFF
--- a/packages/lexical-devtools/package.json
+++ b/packages/lexical-devtools/package.json
@@ -2,7 +2,7 @@
   "name": "@lexical/devtools",
   "description": "Lexical DevTools browser extension",
   "private": true,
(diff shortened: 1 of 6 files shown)


packages/lexical-devtools/package.json
```diff
@@ -28,7 +28,7 @@
     "@vitejs/plugin-react": "^4.2.1",
     "typescript": "^5.3.3",
     "lexical": "0.14.2",
```
etrepum, line 30: this is also out of date but not a public package so maybe not important

StyleT: updated

Sahejkm (approved):
LGTM

PR #5915, opened 2024-04-17
feat(@lexical/devtools): Added Edge support

StyleT:
This will allow us to test Chrome build steps while we're waiting for the Chrome Web Store registration (on the final stretch). And since Edge is Chromium based - it costs kinda nothing.

Screenshot 2024-04-17 at 4 40 37 PM


Sahejkm (approved):
LGTM

PR #5917, opened 2024-04-18
Fix selecting table selects an image after

KatsiarynaDzibrova:
Issue: selection of table by cursor dragging also selects the image after the table
Fix: resolve selection focus to the beginning of the paragraph with an image, not the root element with offset

Before:
old-table-picture-select

After:
new-table-picture-select

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -17,6 +17,7 @@ import type {
 import type {
   BaseSelection,
   ElementFormatType,
+  ElementNode,
   LexicalCommand,
   LexicalEditor,
   LexicalNode,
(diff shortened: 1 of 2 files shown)


ivailop7 (dismissed):

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -702,13 +714,16 @@ export function applyTableHandlers(
           if (isPartialyWithinTable) {
             const newSelection = selection.clone();
             if (isFocusInside) {
-              newSelection.focus.set(
-                tableNode.getParentOrThrow().getKey(),
-                isBackward
-                  ? tableNode.getIndexWithinParent()
-                  : tableNode.getIndexWithinParent() + 1,
```
zurfyx, line 709: Let's check why this number is incorrect, that's the root cause of the issue. Notice that it will work if you remove the `+1`.

While the solution you have suggested is valid, it is equivalent of introducing a custom selection normalizer which I don't think that is what we want. In the hypothetical case we wanted this, we would probably prefer to improve `$normalizeSelection` (the common core module).

zurfyx (changes requested):

ivailop7 (approved):
LGTM

ivailop7:
I'm happy here, any thoughts before I go for merge here, @zurfyx

ivailop7 (approved):

PR #5919, opened 2024-04-18
Add setTimeout for mousemove handler

AlexanderReznik:
Fixes
table-selection-scroll
Selection is getting stuck and the page is not getting scrolled automatically

Happens in Chrome, not in Firefox
The problem is that the order of events `mousemove` and 'selectionchange` is not deterministic. When `mousemove` is handled before it prevents `selectionchange` handler from executing fully resulting in this bug.

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -102,15 +102,17 @@ export function applyTableHandlers(
     };
 
     const onMouseMove = (moveEvent: MouseEvent) => {
-      const focusCell = getDOMCellFromTarget(moveEvent.target as Node);
-      if (
-        focusCell !== null &&
-        (tableObserver.anchorX !== focusCell.x ||
-          tableObserver.anchorY !== focusCell.y)
-      ) {
-        moveEvent.preventDefault();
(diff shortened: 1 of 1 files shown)


ivailop7 (approved):
LGTM

AlexanderReznik:
> Do you think it's possible to add test here to prevent future regression?

I've tried, couldn't add a test for the mouse leaving the window to cause a scroll

StyleT:
> > Do you think it's possible to add test here to prevent future regression?
> 
> I've tried, couldn't add a test for the mouse leaving the window to cause a scroll

No worries, can you pls then add a code comment to explain `setTimeout` presence and `TODO` with newly created issue number that shall address lack of test here. So someone else (maybe me), can take a look at the test.

I'll merge it right after you have comment + follow up issue to write test.

zurfyx (approved):

AlexanderReznik:
> No worries, can you pls then add a code comment to explain `setTimeout` presence and `TODO` with newly created issue number that shall address lack of test here. So someone else (maybe me), can take a look at the test.

https://github.com/facebook/lexical/issues/5928
https://github.com/facebook/lexical/pull/5919/commits/c626b1b8cbeb7daa1a42d474881c4fc790bdd6c1

ivailop7 (approved):

PR #5924, opened 2024-04-18
feat(@lexical/devtools): Publish pipeline added

StyleT:
This adds publish pipeline draft (as I'm still waiting for stores approval for FF/Edge + license update for Apple/Chrome).

All secrets were added to repo and I'll publish access request procedure in the follow up diff that adds Chrome/Safari


.github/workflows/devtools-extension-publish.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish DevTools extension to stores
+
+on:
+  workflow_dispatch:
+
+jobs:
+  release:
+    if: github.repository_owner == 'facebook'
```
acywatson, line 8: what's the purpose of this condition?

acywatson (approved):
This looks right - I expected a checksum or signing step of some sort, though?

StyleT:
> This looks right - I expected a checksum or signing step of some sort, though?

So signing happens at the store side after version reviewal. We simply upload zip there and they do review + signing.

PR #5932, opened 2024-04-21
chore(@lexical/history): Remove redundant register call

2wheeh:
```ts
 editor.registerUpdateListener(applyChange)
```

It's called redundantly although there hasn't been any issues as update listeners are managed in a `Set`.


StyleT (approved):
LGTM! Thanks 🙏

PR #5938, opened 2024-04-22
Maintain text decoration styles on paste

Shubhankerism:
Currently pasting text content (eg, from gDocs) with both underline and strikethrough formatting loses both these formatting styles. This PR fixes that issue.

The `text-decoration` property's value can be a string that's a combination of different values like underline and line-through (eg: `text-decoration: underline line-through;`), hence modified the logic accordingly.

### Before
text-decor-before

### After
text-decor-after


ivailop7 (approved):
LGTM

PR #5941, opened 2024-04-22
Set eqeqeq rule to always and clarify type checks

2wheeh:
I tried to improve readability and type safety.

### Changes

- Set `eqeqeq` eslint rule to [`always`](https://eslint.org/docs/latest/rules/eqeqeq#always), which forces `===`, `!==` instead of `==`, `!=`. 
- Simplified some assertions which has double negation around `invariant`.
- Tweaked some types and removed unnecessary type checks.

### Further steps
- Unifying some unnecessary `undefined` and `null` where one of both would suffice.
- ~~Developing a general method to infer correct types from arrays or objects considering when index or key is not valid.~~ -> `--noUncheckedIndexedAccess` introduced in TS4.1 is for this.
- Implementing proper handling for `__lexical**` fields on the global namespace (mostly with `@ts-expect-error: internal field`), similar to [the way in devtools](https://github.com/facebook/lexical/blob/main/packages/lexical-devtools/src/types.ts).

### DIFF
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -194,10 +194,9 @@ export function getEndOfCodeInLine(
   anchor: CodeHighlightNode | TabNode,
 ): CodeHighlightNode | TabNode {
   const lastNode = getLastCodeNodeOfLine(anchor);
-  invariant(
-    !$isLineBreakNode(lastNode),
-    'Unexpected lineBreakNode in getEndOfCodeInLine',
-  );
+  if ($isLineBreakNode(lastNode)) {
+    invariant(false, 'Unexpected lineBreakNode in getEndOfCodeInLine');
+  }
   return lastNode;
 }
 
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -695,7 +695,7 @@ function handleShiftLines(
     start = getFirstCodeNodeOfLine(focusNode);
     end = getLastCodeNodeOfLine(anchorNode);
   }
-  if (start == null || end == null) {
+  if (start === null || end === null) {
     return false;
   }
 
@@ -726,7 +726,7 @@ function handleShiftLines(
   const sibling = arrowIsUp
     ? linebreak.getPreviousSibling()
     : linebreak.getNextSibling();
-  if (sibling == null) {
+  if (sibling === null) {
     return true;
   }
 
@@ -739,7 +739,7 @@ function handleShiftLines(
         : getLastCodeNodeOfLine(sibling)
       : null;
   let insertionPoint =
-    maybeInsertionPoint != null ? maybeInsertionPoint : sibling;
+    maybeInsertionPoint !== null ? maybeInsertionPoint : sibling;
   linebreak.remove();
   range.forEach((node) => node.remove());
   if (type === KEY_ARROW_UP_COMMAND) {
@@ -813,7 +813,7 @@ export function registerCodeHighlighting(
     );
   }
 
-  if (tokenizer == null) {
+  if (tokenizer === undefined) {
     tokenizer = PrismTokenizer;
   }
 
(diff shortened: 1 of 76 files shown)


packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -36,9 +36,9 @@ import {PUNCTUATION_OR_SPACE, transformersByType} from './utils';
 const MARKDOWN_EMPTY_LINE_REG_EXP = /^\s{0,3}$/;
 const CODE_BLOCK_REG_EXP = /^```(\w{1,10})?\s?$/;
 type TextFormatTransformersIndex = Readonly<{
-  fullMatchRegExpByTag: Readonly<Record<string, RegExp>>;
+  fullMatchRegExpByTag: Readonly<Partial<Record<string, RegExp>>>;
   openTagsRegExp: RegExp;
-  transformersByTag: Readonly<Record<string, TextFormatTransformer>>;
+  transformersByTag: Readonly<Partial<Record<string, TextFormatTransformer>>>;
```
2wheeh, line 41: When the key is not a valid one, it should infer the value as `undefined`.
`Partial` wrapping `Record` enables it.

For instance:
```ts
const transformer = textFormatTransformersIndex.transformersByTag[match[1]];

  if (transformer) {
    for (const format of transformer.format) {
      if (!currentNode.hasFormat(format)) {
        currentNode.toggleFormat(format);
      }
    }
  }
```
`transformer` should be inferred as `undefined` when `match[1]` is not a valid key.

Otherwise, there is a potential risk of removing `if(transformer)` check.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1915,7 +1913,10 @@ function internalResolveSelectionPoint(
     } else if (editor._blockCursorElement !== null) {
       resolvedOffset--;
     }
-    resolvedNode = getNodeFromDOM(childDOM);
+
+    if (childDOM) {
+      resolvedNode = getNodeFromDOM(childDOM);
+    }
```
2wheeh, line 1919:
```ts
 let childDOM = childNodes[resolvedOffset];
```
When `resolvedOffset` is not a valid index, `childDOM` is assigned as `undefined`.
This check was missed since TS infers it as `ChildNode`.

Getting value from array or object (like this and `splitText`) might need safer way in the end.

packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -194,10 +194,9 @@ export function getEndOfCodeInLine(
   anchor: CodeHighlightNode | TabNode,
 ): CodeHighlightNode | TabNode {
   const lastNode = getLastCodeNodeOfLine(anchor);
-  invariant(
-    !$isLineBreakNode(lastNode),
-    'Unexpected lineBreakNode in getEndOfCodeInLine',
-  );
+  if ($isLineBreakNode(lastNode)) {
```
acywatson, line 197: Is this an improvement?

Why not leverage the invariant call itself here? Why a separate conditional?

packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -349,7 +349,7 @@ function printTargetProperties(node: LinkNode) {
 function printRelProperties(node: LinkNode) {
   let str = node.getRel();
   // TODO Fix nullish on LinkNode
```
acywatson, line 351: Is this still relevant?

packages/lexical-devtools/src/entrypoints/injected/InjectedPegasusService.ts
```diff
@@ -50,7 +50,7 @@ export class InjectedPegasusService
     isReadonly: boolean,
   ): void {
     const editorNode = queryLexicalNodeByKey(editorKey);
-    if (editorNode == null) {
+    if (editorNode === undefined) {
```
acywatson, line 53: I think ideally queryLexicalNodeByKey would return null and never undefined. We try to avoid dealing with undefined anywhere within the library. I don't want to expand the scope of this PR, but maybe a comment.

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -1185,10 +1183,10 @@ export function findParentPreDOMNode(node: Node) {
 function convertTextDOMNode(domNode: Node): DOMConversionOutput {
   const domNode_ = domNode as Text;
   const parentDom = domNode.parentElement;
-  invariant(
-    parentDom !== null,
-    'Expected parentElement of Text not to be null',
-  );
+
+  if (parentDom === null) {
```
acywatson, line 1187: Same thing here - it seems inconsistent with how this was done everywhere else.

acywatson:
Feel free to reopen if we can get conflicts resolved.

PR #5946, opened 2024-04-23
Fix table of contents for headings in tables and collapsible sections

KatsiarynaDzibrova:
Fixes following issues with table of contents:
- The order in table of contents is wrong for heading inside tables and collapsible sections.
- The heading is duplicated when resizing parent table cell

Now on each heading update dfs is used to get all headings and create a new TOC from scratch. It might take more time, but the change shouldn't be noticeable

Before:
toc-before

After:
toc-after

### DIFF
--- a/packages/lexical-react/src/LexicalTableOfContents.tsx
+++ b/packages/lexical-react/src/LexicalTableOfContents.tsx
@@ -10,7 +10,7 @@ import type {LexicalEditor, NodeKey, NodeMutation} from 'lexical';
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {$isHeadingNode, HeadingNode, HeadingTagType} from '@lexical/rich-text';
-import {$dfs} from '@lexical/utils';
+import {$getPreviousNode} from '@lexical/utils';
 import {$getNodeByKey, $getRoot, TextNode} from 'lexical';
 import {useEffect, useState} from 'react';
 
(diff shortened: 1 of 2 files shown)


packages/lexical-react/src/LexicalTableOfContents.tsx
```diff
@@ -138,42 +86,15 @@ export default function LexicalTableOfContentsPlugin({
...
-                );
-              }
-            }
-          }
+          const dfsNodes = $dfs().filter((dfsNode: DFSNode) =>
+            $isHeadingNode(dfsNode.node),
+          );
+          const headingNodes = dfsNodes.map(
+            (dfsNode) => dfsNode.node as HeadingNode,
+          );
```
ivailop7, line 96: You can reduce from 2 loops to 1 loop using reduce or flatmap. Syntax should be something along those lines:

```typescript
const headingNodes = $dfs().reduce((headingNodesList, node) => $isHeadingNode(dfsNode.node) ? [...acc, value] : acc, [])
```

ivailop7 (commented):
A small speedup comment from me. @zurfyx @acy, happy with this approach?

packages/lexical-react/src/LexicalTableOfContents.tsx
KatsiarynaDzibrova: Thank you. Added the change

ivailop7 (dismissed):

packages/lexical-react/src/LexicalTableOfContents.tsx
```diff
@@ -138,42 +86,13 @@ export default function LexicalTableOfContentsPlugin({
...
-                  currentTableOfContents,
-                );
-              }
-            }
-          }
+          const headingNodes: HeadingNode[] = $dfs().reduce(
+            (acc, {node}) => ($isHeadingNode(node) ? [...acc, node] : acc),
+            [] as HeadingNode[],
+          );
+          currentTableOfContents = $newTableOfContents(headingNodes);
```
ivailop7, line 95: I think we can do an extra performance optimization - we store the list, and we only replace it with the output from dfs on create, update for a headingNode, but on delete, we just delete from the map. What do you thing?

zurfyx (commented):
Have we measured performance? Running a full `$dfs` every time that a HeadingNode is mutated might be fairly inefficient, that why the previous version uses mutated nodes. For reference, the reconciler doesn't look at all nodes either, instead it traverses through the dirty nodes, I would take the reconciler as an example for optimizing this plugin.

KatsiarynaDzibrova:
Made the following changes:
1. Explicitly checking if the heading is already in TOC to avoid duplicates
2. Replaced dfs with listener on root updates, which updates all child headings
3. Find previous heading for a new heading only when it's updated
The performance should be much better now as we only visit updated nodes

@zurfyx feel free to take a look

packages/lexical-utils/src/index.ts
```diff
@@ -222,6 +222,29 @@ function $getDepth(node: LexicalNode): number {
   return depth;
 }
 
+export function $getPreviousNode(
```
ivailop7, line 225: It lacks a comment description, also confused by the naming getPreviousNode, yet in the first if statement, it gets the last child, which is not 'previous' to my understanding.

ivailop7: Still unsure if this should be in utils or encapsulated in the TOC plugin file. Thoughts @zurfyx ?

zurfyx (approved):

PR #5948, opened 2024-04-24
fix(LexicalNode): fix inline decorator isSelected

jason89521:
## Why

I pressed `shift + ArrowLeft` at the beginning of the video, and then pressed `BackSpace` to delete the selection. You can see the image node is selected, but when I deleted the selection, it didn't been deleted. Actually, the image node was not selected.

At the second half of the video, I pressed `shift + ArrowRight`, the first image also showed that it was selected. However, the image node didn't been selected actually, so when I deleted the selection, the image node didn't been deleted.

https://github.com/facebook/lexical/assets/47748132/8b33a6fb-f6e1-41b1-b5cd-3aded298e3d4

## What

Beside checking whether the selection is collapsed, we should also check non-collapsed situation.

## How

First, check whether the node is an inline decorator node.

Then check whether the anchor is before the focus.

- If so
  - check whether the anchor at the element node's last offset
  - check whether the anchor's node is the decorator node's parent
  - check whether the decorator node is the last child
  - if the above 3 conditions are satisfied, return false
- else
  - check whether the focus at the element node's last offset
  - check whether the focus's node is the decorator node's parent
  - check whether the decorator node is the last child
  - if the above 3 conditions are satisfied, return false

## Additional Information

After this pr:

https://github.com/facebook/lexical/assets/47748132/f3e3257b-f26d-45d4-8322-672dda184d9e


StyleT:
Hi! Thanks for the contribution! 

Can you pls write tests to cover this use case?

jason89521:
@StyleT added in [c1d6ada](https://github.com/facebook/lexical/pull/5948/commits/c1d6adaa560fb8eb720bb05a7a0bbf8740fa1960)!

jason89521:
@StyleT  Hi, just a gentle nudge to see if you could take a moment to review this pull request when you have a chance. Thanks!

ivailop7 (approved):

PR #5951, opened 2024-04-24
Fix TableCell child nodes on paste

Shubhankerism:
This PR fixes pasting tables that have the copied html in such format:
```
...
<td>
<span> ... </span>
</td>
...
```
Pasting this results in the following node structure:

image

which causes the editor to freeze and eventually crash on enter key press, as the text node is not wrapped in a paragraph node.
This issue is similar to the one described [here](https://github.com/facebook/lexical/pull/5670#issuecomment-1978455980) and was fixed for the case when pasting empty cells.

This PR addresses the case when on paste the children of the `TableCell` node are not within a block element node. 

Another case to reproduce this issue is when we copy something with such html:
```
...
<td>
<a> ... </a>
</td>
...
```
This results in a link node inside a tablecell node which again crashes on enter key press. 

Also, this PR fixes the issue with alignment of text inside the cell as well, as this formatting can now be applied on the wrapping block element node.
So, copying from google sheets will now have the correct alignment.

### Before
table-align-before

### After
table-align-after

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs
@@ -1006,7 +1006,7 @@ test.describe('CodeBlock', () => {
           </p>
         </td>
         <td class="PlaygroundEditorTheme__tableCell">
-          <p class="PlaygroundEditorTheme__paragraph">
+          <p class="PlaygroundEditorTheme__paragraph" style="text-align: right">
             <span
               class="PlaygroundEditorTheme__textUnderline"
               data-lexical-text="true">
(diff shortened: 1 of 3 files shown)


ivailop7 (approved):

zurfyx:
Thank you, we can merge this, but I'd be inclined to roll it back as soon as https://github.com/facebook/lexical/pull/5857 is ready as it fixes the same issue in a more generic manner cc @potatowagon

ivailop7:
Thanks Gerard! Happy to revert once the larger one lands. Do we have an ETA on the generic solution? If it's a week, we can wait it out.

potatowagon:
#5857 is ready, tests and comments addressed @ivailop7 @zurfyx

PR #5956, opened 2024-04-24
Fix CI invariant transpile

zurfyx:


packages/lexical/src/LexicalSelection.ts
```diff
@@ -2681,10 +2681,17 @@ export function $getTextContent(): string {
...
 
 function removeTextAndSplitBlock(selection: RangeSelection): number {
+  let selection_ = selection;
   if (!selection.isCollapsed()) {
-    selection.removeText();
+    selection_.removeText();
   }
-  const selection_ = $getSelection();
+  // A new selection can originate as a result of node replacement, in which case is registered via
+  // $setSelection
```
Sahejkm, line 2689:
```suggestion
  // $getSelection
```

Sahejkm (approved):

packages/lexical/src/LexicalSelection.ts
zurfyx: @Sahejkm - I meant the value is set via `$setSelection` which as a result can be retrieved via `$getSelection`

PR #5957, opened 2024-04-24
Refactor table resizer rendering logic

AlexanderReznik:
Disabling table resizer when mouse button is pressed, instead of when the selection is of type `TableSelection`.

This makes ui slightly better in some cases:
1. Cursor doesn't change when starting the table selection
table-selection-cursor

2. Cursor doesn't change when during range selection the cursor is over the table
range-selection-cursor

### DIFF
--- a/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
@@ -18,17 +18,11 @@ import {
   $getTableRowIndexFromTableCellNode,
   $isTableCellNode,
   $isTableRowNode,
-  $isTableSelection,
   getDOMCellFromTarget,
   TableCellNode,
 } from '@lexical/table';
 import {calculateZoomLevel} from '@lexical/utils';
-import {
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
```diff
@@ -102,7 +83,7 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
           });
           return;
         }
-
+        updateIsMouseDown(isMouseDownOnEvent(event));
```
AlexanderReznik, line 86: This is a safety measure in case `mouseup` is not registered by the browser. I'm not sure if this can ever happen and I'll be happy to hear a second opinion, if this check is required

ivailop7:
@AlexanderReznik, in the preview here, I'm still getting the selection stuck

AlexanderReznik:
> @AlexanderReznik, in the preview here, I'm still getting the selection stuck

This is not getting fixed in this pr, but I'll try to fix it later

ivailop7 (approved):

PR #5959, opened 2024-04-24
Fix insertText with element selection

zurfyx:
It's valid to have an text + element selection that `selection.getNodes()` will only recognize as one node. This happens fairly often in FF when trying to span the entire text node selection to the entire line and is followed by a line break. See test to understand the problem at hand better.

I believe this PR fixes the entire problem and we don't need to understand further combinations, as `insertText` relies on normalized selection.

Closes https://github.com/facebook/lexical/issues/5233


acywatson:
fascinating

acywatson (approved):

ivailop7 (approved):

PR #5978, opened 2024-04-27
When creating a new check list, set the `checked` value of the list item to `false` instead of `undefined`

amanharwara:
At the moment, when creating a new checklist, the newly created list item doesn't have the `checked` set until the user manually toggles it. This PR makes sure that the `checked` value is set to `false` and not `undefined`.
This also makes the behavior in line with the behavior when creating a check list using the markdown transformer.

https://github.com/facebook/lexical/assets/8348101/a065966a-ceae-4ce4-bf11-250a041efd0f


packages/lexical-list/src/formatList.ts
```diff
@@ -156,7 +158,9 @@ function createListOrMerge(node: ElementNode, listType: ListType): ListNode {
 
   const previousSibling = node.getPreviousSibling();
   const nextSibling = node.getNextSibling();
-  const listItem = $createListItemNode();
+  const listItem = $createListItemNode(
+    listType === 'check' ? false : undefined,
+  );
```
potatowagon, line 163: wondering if its more appropriate to make the handling of this behaviour clearer inside packages/lexical-list/src/LexicalListItemNode.ts

potatowagon (commented):
could u please update the description with recordings/screenshots to show no regressions to checklist behavior?

packages/lexical-list/src/formatList.ts
amanharwara: how about now? i've refactored `LexicalListItemNode.getChecked()` to always return a boolean if the parent list type is `"check"` and undefined otherwise.

potatowagon:
packages/lexical-playground/__tests__/e2e/List.spec.mjs:1170:3 › Nested List › Can create check list, toggle it to bullet-list and back 

is breaking

amanharwara:
Sorry for the long delay. Should be fixed now.

etrepum (approved):
LGTM, seems reasonable to enforce this and the node transform should keep it consistent

Issue #5930, opened 2024-04-20
Proposal: DecoratorElementNode


yf-yang:
I've wanted this feature for a long long time, today I get some ideas to implement it, so I'd like to discuss it here.

# People need something like "ReactElementNode"

I've seen such questions again and again. People not only need a plain ElementNode which is a single DOM node, they want to also use other framework's generated DOM node as the parent of a set of Lexical nodes. However, this is not natively supported.

Let's first check other frameworks. Slate.js seamlessly [supports it](https://docs.slatejs.org/walkthroughs/03-defining-custom-elements), TipTap breaks the element and the child into `NodeViewWrapper` and `NodeViewComponent`, check [doc](https://tiptap.dev/docs/editor/guide/node-views/react#adding-a-content-editable) and [example](https://tiptap.dev/docs/editor/guide/node-views/examples#drag-handles). All of them are handle those cases in one editorState.

Now let's go back to Lexical. Lexical offers nested editors and LexicalNestedComposer. However, after heavily develop use cases with it, I find it has a lot to improve.

# Why nested editor is not a good idea?
1. Moving contents between editors is painful.
You can't directly move nodes between editors. Serialization is mandatory and I have to deal with at least two `editor.update`s.

2. Create a plugin that supports multiple editors is painful.
Due to previous reason, when I created a drag and drop plugin that supports nested editors, I have to consider if the editor to be dropped is the same with the dragging node's editor. There are two cases, the first one is one `editor.update`, the second one needs two `update` call and serialization.

3. Modifying multiple editors SIMULTANEOUSLY is poorly supported.
Suppose I move a node from one editor to the other. This is a transaction that involves two editors. However, both HistoryPlugin and CollaborationPlugin treats them as two separate transactions. I've spent lots of time modifying them.

I've been going deeper with those questions. I think, as an editor framework whose idea originates from React, the editorState should also be singleton, just like how React does. To be more precise, for a single LexicalComposer, all the nodes within it, including those nested editors, should share the same editorState.

# Implementation Proposal
## What nested editor is doing
[...]

PR #5981, opened 2024-04-28
feat(core): add experimental decorator element node and nested root node

yf-yang:
# DecoratorElementNode
This is a VERY EARLY draft of DecoratorElementNode (#5930).

I submit this PR because I want to:
1. Confirm the API design.
2. Allow the maintenance team and the community to review the codes as early as possible, because I am not an expert of reconciliation, your guidance/your bug report may help me work faster to deal with the reconciliation correctly.

## Naive Demo
- CardNode (Fixed children)

https://github.com/facebook/lexical/assets/36890796/61a83c26-02cc-47d4-ab9e-08755512b988

- ReactListNode(Variable children)

https://github.com/facebook/lexical/assets/36890796/812dd021-22a5-4a8a-bdc2-155d71c0dc76

## Tasks to do before this PR is submitted
- [ ] Confirm the API design is good enough (I don't like `editor.setNestedRootElement()`, could it be better?)
- [ ] Fix the most conspicuous reconciliation bugs
- [x] Provide a demo including two fixed nested roots (CardNode)
- [x] Provide a demo of a node containing arbitrary number of child nested roots (ReactListNode)
- [x] Provide a demo that the nested root could be mounted and unmounted with a button (Add a show/hide button to CardNode)

## Tasks I don't plan to do in this PR
Since I don't have so much spare time, I plan to deal with those issues later, maybe the community can offer some help.

- Add enough test cases
- Document the node well
- Deal with IO and copy/paste
- Deal with corner cases of the selection change (there are many many bugs with it)
- Guarantee deeply nested decorator element nodes are working correctly
- Make UI of the demo nodes beautiful

# API Documentation Draft
## DecoratorElementNode
To combine external framework and Lexical node trees, decorator element node could be an option. It offers a `decorate()` method to allow external framework to render the DOM, but the external can also have child Lexical nodes handled by Lexical.

``` typescript
export class CardNode extends EXPERIMENTAL_DecoratorElementNode<JSX.Element> {
  static getType(): string {
    return 'card';
  }

  static clone(node: CardNode): CardNode {
    return new CardNode(node.__key);
  }

  constructor(key?: NodeKey) {
    super(key);

    // ElementNode will automatically clone children
    // So we only need to set the children if the node is new
    if (key === undefined) {
      this.append($createNestedRootNode());
      this.append($createNestedRootNode());
    }
  }

  createDOM(): HTMLElement {
    return document.createElement('div');
  }

  updateDOM(): false {
    return false;
  }

  decorate(editor): JSX.Element {
    const onTitleRef = (element) => {
      editor.setNestedRootElement(this.getChildAtIndex(0).getKey(), element);
    }
    const onBodyRef = (element) => {
      editor.setNestedRootElement(this.getChildAtIndex(1).getKey(), element);
    }
   
    return (
      <div>
        <div ref={onTitleRef} />
        <div ref={onBodyRef} />
      </div>
    );
  }
}
```
Take the `CardNode` as an example:
- DecoratorElementNode is a subclass of ElementNode, so it can have an array of LexicalNodes. However, for DecoratorElementNode, the only valid type of its children is `NestedRootNode`. To use other nodes, they should be NestedRootNode's children.
- Like DecoratorNode, DecoratorElementNode provides `decorate()` method, where developers can implement a render function to be handled by external frameworks.
- To allow Lexical handle external framework's DOMs and mount Lexical's dom to them, `LexicalEditor` provides an `setNestedRootElement` method. When a new HTMLElement is set, Lexical will render its children DOMs.

If a DecoratorElementNode has a fixed set of children, developers are responsible to record each child's index and meaning. If it has a variable array of children, then just make sure the `onRef`'s order in `decorate` function maps to each child correctly.

### DIFF
--- a/packages/lexical-playground/src/nodes/CardNode.tsx
+++ b/packages/lexical-playground/src/nodes/CardNode.tsx
@@ -1,0 +1,110 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {
+  $applyNodeReplacement,
+  $createNestedRootNode,
+  $createParagraphNode,
+  $createTextNode,
+  EXPERIMENTAL_DecoratorElementNode,
+  EXPERIMENTAL_NestedRootNode,
+  LexicalNode,
+  NodeKey,
+} from 'lexical';
+import * as React from 'react';
+
+export class CardNode extends EXPERIMENTAL_DecoratorElementNode<JSX.Element> {
+  static getType(): string {
+    return 'card';
+  }
+
+  static clone(node: CardNode): CardNode {
+    return new CardNode(node.__key);
+  }
+
+  constructor(key?: NodeKey) {
+    super(key);
+
+    // ElementNode will automatically clone children
+    // So we only need to set the children if the node is new
+    if (key === undefined) {
+      const title = $createNestedRootNode();
+      const titleParagraph = $createParagraphNode();
+      titleParagraph.append($createTextNode('Title sample text'));
+      title.append(titleParagraph);
+      this.append(title);
+      const body = $createNestedRootNode();
+      const bodyParagraph = $createParagraphNode();
+      bodyParagraph.append($createTextNode('Content sample text'));
+      body.append(bodyParagraph);
+      this.append(body);
+    }
+  }
+
+  get title(): EXPERIMENTAL_NestedRootNode {
+    return this.getChildAtIndex(0) as EXPERIMENTAL_NestedRootNode;
+  }
+
+  get body(): EXPERIMENTAL_NestedRootNode {
+    return this.getChildAtIndex(1) as EXPERIMENTAL_NestedRootNode;
+  }
+
+  // View
+
+  createDOM(): HTMLElement {
+    return document.createElement('div');
+  }
+
+  updateDOM(): false {
+    return false;
+  }
+
+  decorate(): JSX.Element {
+    return (
+      <div
+        style={{
+          borderColor: 'black',
+          borderStyle: 'solid',
+          borderWidth: 1,
+        }}>
+        <div>Title</div>
+        <div
+          style={{
+            borderColor: 'red',
+            borderStyle: 'solid',
+            borderWidth: 1,
+            minHeight: 100,
+            minWidth: 300,
+          }}
+          ref={this.title.onRef}
+        />
+        <div>Content</div>
+        <div
+          style={{
+            borderColor: 'blue',
+            borderStyle: 'solid',
+            borderWidth: 1,
+            minHeight: 100,
+            minWidth: 300,
+          }}
+          ref={this.body.onRef}
+        />
+      </div>
+    );
+  }
+}
+
+export function $createCardNode(): CardNode {
+  return $applyNodeReplacement(new CardNode());
+}
+
+export function $isCardNode(
+  node: LexicalNode | null | undefined,
+): node is CardNode {
+  return node instanceof CardNode;
+}
--- a/packages/lexical-playground/src/nodes/CardNode.tsx
+++ b/packages/lexical-playground/src/nodes/CardNode.tsx
@@ -6,29 +6,35 @@
  *
  */
 
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {
   $applyNodeReplacement,
   $createNestedRootNode,
   $createParagraphNode,
   $createTextNode,
+  $getNodeByKey,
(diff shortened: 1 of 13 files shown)


GermanJablo:
What are the problems you currently encounter with DecoratorNodes in your use case?

In my case they are:
- different undo managers with the editor.
- fine grained real time collaboration

I believe that these problems are solvable and that they should be addressed even if this PR goes ahead.
Perhaps by talking about these problems and yours we can think of other alternatives.

In the past I tried to do something similar to what you do in this PR (on a local branch), but I ran into a lot of complications. The 1 LexicalNode = 1 DOM node model is something that is deeply rooted from the roots of Lexical.

yf-yang:
@GermanJablo #5930

GermanJablo:
Still, may I know how you would describe what specific problems this solves?

yf-yang:
Decorate an ElementNode.

Consider things like that: https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/nodes/LayoutContainerNode.ts

You have to know vanilla JS well to make a complex element node, or you can do very little stuff. Alternatively, NestedEditor is needed and there exists multiple editor states.

LvChengbin:
I'm all for adding support for the `decorate` method to `ElementNode`.

1. We often rely on React component libraries like Material-UI to keep our element nodes, such as ParagraphNode and QuoteNode, looking consistent. But without the decorate method, we end up writing more CSS just to match the styles of components in the library.

2. Trying to do things across nested editors comes with its fair share of headaches—think wonky history management, wonky selection handling, tangled data structures, plugin clashes, you name it.

vans37:
> Still, may I know how you would describe what specific problems this solves?

I'm new to Lexical, but I struggle with my use case:
I develop basic image gallery plugin. There's a gallery node which is decorator and it has a property __children that are builtin ImageNode from playground. 
The issues arises when  new images inserted into gallery - I create new ImageNode and append them as children prop to my GalleryNode, but doing that causes image to be non interactive(can't edit captions, no visible focus and selection etc.) - because it is not registered in the editor state which leads to $getNodeByKey to return null and all the troubles(internally ImageComponent uses $getNodeByKey to get image that is being interacted with). 
If I insert image nodes into editor via insertNodes() method, images will become registered in the editor and are interactive, but it leads to node duplication, because image nodes will be part of __children property of gallery and at the same time they will be appended to the root node.
I don't exactly know how to go about it for now, maybe use element node which is parent for other nodes? But then user can insert any node type into my gallery including text, and I need to disable all interactions etc.

CanRau:
Am also dreaming of something like an ElementNode with decorate capability 🥹
A separate `DecoratorElementNode` as proposed here would be very welcome to not risk breaking anything existing but allowing for this additional feature where desired 🙌🏼 

Agree with most if not all reasons pointed out here, where the "simplest" is being able to just re-use existing ui components as @LvChengbin has mentioned

Also nested editor situation isn't (yet) as intuitive to use imho, especially, as already mentioned, when it comes to selections, history etc.

vans37:
> > Still, may I know how you would describe what specific problems this solves?
> 
> I'm new to Lexical, but I struggle with my use case: I develop basic image gallery plugin. There's a gallery node which is decorator and it has a property __children that are builtin ImageNode from playground. The issues arises when new images inserted into gallery - I create new ImageNode and append them as children prop to my GalleryNode, but doing that causes image to be non interactive(can't edit captions, no visible focus and selection etc.) - because it is not registered in the editor state which leads to $getNodeByKey to return null and all the troubles(internally ImageComponent uses $getNodeByKey to get image that is being interacted with). If I insert image nodes into editor via insertNodes() method, images will become registered in the editor and are interactive, but it leads to node duplication, because image nodes will be part of __children property of gallery and at the same time they will be appended to the root node. I don't exactly know how to go about it for now, maybe use element node which is parent for other nodes? But then user can insert any node type into my gallery including text, and I need to disable all interactions etc.

Answering my own question: use regular ElementNode as a parent for other nodes, i.e GalleryNode that is a wrapper(parent) for already existing ImageNode in the lexical. Create new GalleryNode from user interaction (toolbar, slash menu etc.), and handle adding new images to gallery - create ImageNode and append it as a child to GalleryNode.
The problem was to create a separate gallery handler (mostly buttons, and inputs for user to select new images, set amount of columns and other settings) that  is a react component and needed to be appended as a child into each gallery that existed in the editor (there can be multiple galleries in one editor).
At first I used a method of document.createElement and setAttribute to set a node key on each gallery controls and createDOM which created html stucture. Then in separate react component via useEffect I found all the galleries there were in the editor and attached event listeners to it. It was tedious and proned to errors.
As it later turned out I was wrong and there is a way easier method: create a separate DecoratorNode that renderes react component(decorates) and then append it as a child to GalleryNode, and voila you can use regular react without vanilla js in createDOM.

PR #5982, opened 2024-04-28
fix: allow tsc to typecheck tests, fix type issues in those tests

etrepum:
Mostly just adding non-null assertions and adding types where inference fails. I did refactor a few list reduces that do a concat (which is also quadratic time) to use flatMap which is easier for TypeScript to infer.

I added some docs to the FAQ relevant to reconciliation and how the tests are written https://lexical-git-fork-etrepum-fix-test-types-fbopensource.vercel.app/docs/faq#when-does-reconciliation-happen

### DIFF
--- a/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
+++ b/packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
@@ -34,12 +34,12 @@ import {
 } from 'lexical/src';
 import {createTestEditor, TestComposer} from 'lexical/src/__tests__/utils';
 import React from 'react';
-import {createRoot} from 'react-dom/client';
+import {createRoot, Root} from 'react-dom/client';
 import * as ReactTestUtils from 'react-dom/test-utils';
 
 describe('LexicalHistory tests', () => {
   let container: HTMLDivElement | null = null;
-  let reactRoot;
+  let reactRoot: Root;
 
   beforeEach(() => {
     container = document.createElement('div');
--- a/packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
+++ b/packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
@@ -8,10 +8,10 @@
 
 import {$createTextNode, $getRoot, ParagraphNode, TextNode} from 'lexical';
 
-import {createTestConnection, waitForReact} from './utils';
+import {Client, createTestConnection, waitForReact} from './utils';
 
 describe('Collaboration', () => {
-  let container = null;
+  let container: null | HTMLDivElement = null;
 
   beforeEach(() => {
     container = document.createElement('div');
@@ -19,11 +19,11 @@ describe('Collaboration', () => {
   });
 
   afterEach(() => {
-    document.body.removeChild(container);
+    document.body.removeChild(container!);
     container = null;
   });
 
-  async function expectCorrectInitialContent(client1, client2) {
+  async function expectCorrectInitialContent(client1: Client, client2: Client) {
     // Should be empty, as client has not yet updated
     expect(client1.getHTML()).toEqual('');
     expect(client1.getHTML()).toEqual(client2.getHTML());
@@ -42,8 +42,8 @@ describe('Collaboration', () => {
     const client1 = connector.createClient('1');
     const client2 = connector.createClient('2');
 
-    client1.start(container);
-    client2.start(container);
+    client1.start(container!);
+    client2.start(container!);
 
     await expectCorrectInitialContent(client1, client2);
 
@@ -56,7 +56,7 @@ describe('Collaboration', () => {
 
         const text = $createTextNode('Hello world');
 
-        paragraph.append(text);
+        paragraph!.append(text);
       });
     });
 
@@ -71,8 +71,8 @@ describe('Collaboration', () => {
       client2.update(() => {
         const root = $getRoot();
 
-        const paragraph = root.getFirstChild<ParagraphNode>();
-        const text = paragraph.getFirstChild<TextNode>();
+        const paragraph = root.getFirstChild<ParagraphNode>()!;
+        const text = paragraph.getFirstChild<TextNode>()!;
 
         text.spliceText(6, 5, 'metaverse');
       });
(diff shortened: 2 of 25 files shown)


packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
```diff
@@ -34,12 +34,12 @@ import {
 } from 'lexical/src';
 import {createTestEditor, TestComposer} from 'lexical/src/__tests__/utils';
 import React from 'react';
-import {createRoot} from 'react-dom/client';
+import {createRoot, Root} from 'react-dom/client';
```
potatowagon, line 37: nit: we could import Root as type instead
```suggestion
import type {Root} from 'react-dom/client';
import {createRoot} from 'react-dom/client';
```

packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
```diff
@@ -8,22 +8,22 @@
 
 import {$createTextNode, $getRoot, ParagraphNode, TextNode} from 'lexical';
 
-import {createTestConnection, waitForReact} from './utils';
+import {Client, createTestConnection, waitForReact} from './utils';
```
potatowagon, line 11:
```suggestion
import type {Client} from './utils';
import {createTestConnection, waitForReact} from './utils';
```

packages/lexical-react/src/__tests__/unit/LexicalComposer.test.tsx
```diff
@@ -8,14 +8,14 @@
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import * as React from 'react';
-import {createRoot} from 'react-dom/client';
+import {createRoot, Root} from 'react-dom/client';
```
potatowagon, line 11:
```suggestion
import type {Root} from 'react-dom/client';
import {createRoot} from 'react-dom/client';
```

packages/lexical-react/src/__tests__/unit/PlainRichTextPlugin.test.tsx
```diff
@@ -22,9 +22,10 @@ import {
   $getRoot,
   $getSelection,
   $isRangeSelection,
+  LexicalEditor,
 } from 'lexical';
 import * as React from 'react';
-import {createRoot} from 'react-dom/client';
+import {createRoot, Root} from 'react-dom/client';
```
potatowagon, line 28:
```suggestion
import type {Root} from 'react-dom/client';
import {createRoot} from 'react-dom/client';
```

packages/lexical-react/src/__tests__/unit/PlainRichTextPlugin.test.tsx
```diff
@@ -22,9 +22,10 @@ import {
   $getRoot,
   $getSelection,
   $isRangeSelection,
+  LexicalEditor,
 } from 'lexical';
```
potatowagon, line 26:
```suggestion
} from 'lexical';
import type {LexicalEditor} from 'lexical';
```

packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
```diff
@@ -30,7 +30,13 @@ import {
   $isElementNode,
   $isRangeSelection,
   $setSelection,
+  DecoratorNode,
+  ElementNode,
+  LexicalEditor,
+  LexicalNode,
```
potatowagon, line 36: ditto nit: could be imported as type instead

packages/lexical-website/docs/faq.md
```diff
@@ -37,6 +37,55 @@ If you've used React Hooks before, you can think of `$` functions as being somet
...
+```
+
+At a high level, very roughly, the order of operations looks like this:
+
+1. `editor.update()` is called
+2. `updateA()` is called and updates the editor state
+3. `editor.update()` schedules a reconciliation microtask and returns
+4. `await` schedules a resume microtask and yields control to the task executor
+5. the reconciliation microtask runs, reconciling the editor state with the DOM
+6. the resume microtask runs
```
potatowagon, line 87: 👍 
thanks for adding this! will leave to @zurfyx and other reconciler experts to vet

packages/lexical-selection/src/__tests__/utils/index.ts
```diff
@@ -12,6 +12,8 @@ import {
   $isNodeSelection,
   $isRangeSelection,
   $isTextNode,
+  LexicalEditor,
+  PointType,
 } from 'lexical';
```
potatowagon, line 17:
```suggestion
} from 'lexical';

import type {LexicalEditor, PointType} from 'lexical';
```

potatowagon (commented):
- the only non test file modified is `packages/lexical-selection/src/__tests__/utils/index.ts`. adds typing only so low risk change
- rest of code changes to test files and adds typing -> very low risk change
- left some minor nits about importing as type

looks good! thankyou for this :)

potatowagon (dismissed):

etrepum:
I'd be happy to fix up those imports if you want, I was just doing whatever the language server suggested since I was fixing a few hundred type errors! Perhaps it would be better to do that cleanup separately in a more systematic way? There are other bad import patterns that we could get rid of in the tests (e.g. importing public code from 'lexical/src' instead of 'lexical', and we could get rid of those tsconfig hacks too). It seems likely that we are also importing values as types elsewhere in there… if that's even something we care about? Type imports are usually done as a performance thing or to prevent import side-effects or circular dependencies if values from the module are not otherwise imported.

etrepum:
If we do clean up the type-only imports there are linters and typescript flags that should be configured to enforce whatever style is chosen, e.g. https://typescript-eslint.io/rules/consistent-type-imports/  and https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax are relevant for that

potatowagon:
yes addressing the type imports in a seperate PR makes sense

would be great if a lint could autofix this

StyleT (approved):

etrepum:
Latest commit just fixes the conflicts with main

potatowagon (approved):

StyleT (approved):

Issue #5922, opened 2024-04-18
Feature: escape key handler for plain text plugin


jeffreyclu:
The rich text plugin has an [escape key handler built in](https://github.com/facebook/lexical/blob/main/packages/lexical-rich-text/src/index.ts#L906-L916), but the rich text plugin does not. Can we have the same for the plain text editor?

ivailop7:
Was added and then reverted, since it broke a number of Escape behaviours. @StyleT are we still planning on doing this one, or just closing the issue?

Rakesh-git-2:
I'd like to work on this

etrepum:
This is the current intended behavior, it is straightforward to register your own escape key handler to blur the plain text editor if desired.

It’s not clear that this is expected, e.g. a textarea doesn’t blur when you press escape.

PR #5991, opened 2024-04-30
[lexical-plaintext] Feature: add escape key handler

jeffreyclu:
## Description

Adds the same escape key handler to the plain text package as exists in lexical/richtext.

**Closes:** #5922

### Before
In the plaintext playground, pressing `Esc` doesn't blur the editor.

https://github.com/facebook/lexical/assets/55211974/a34cdd6e-d0a4-483e-ae46-f80759b39340

### After
In the plaintext playground, pressing `Esc` blurs the editor.

https://github.com/facebook/lexical/assets/55211974/3a219753-03c5-4c92-9eeb-d44ce421d112

### DIFF
--- a/packages/lexical-plain-text/src/index.ts
+++ b/packages/lexical-plain-text/src/index.ts
@@ -37,6 +37,7 @@ import {
   KEY_BACKSPACE_COMMAND,
   KEY_DELETE_COMMAND,
   KEY_ENTER_COMMAND,
+  KEY_ESCAPE_COMMAND,
   PASTE_COMMAND,
   REMOVE_TEXT_COMMAND,
   SELECT_ALL_COMMAND,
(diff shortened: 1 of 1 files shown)


StyleT (approved):
LGTM! Thanks for the contribution 🙌

sqs:
(First, thank you for the Lexical editor. It's the best!)

Why is it desirable for <kbd>Esc</kbd> to blur the editor? That is not how most other text editors work. It makes sense for web browser location bars and modal text inputs, but not for multi-line document editor fields.

Here is an example of a situation where <kbd>Esc</kbd> blurring can be frustrating to the user:

1. Type <kbd>@</kbd> and then realize you don't want to add an @-mention.
2. Press <kbd>Esc</kbd> a few times because usually <kbd>Esc</kbd> is idempotent.
3. Oops! Now you have not only exited the @-mention menu, but you've also blurred the editor and can't resume typing.


https://github.com/facebook/lexical/assets/1976/87df1161-f390-4e6e-96a4-5382023be542

PR #5993, opened 2024-04-30
[@lexical/website] Documentation: Added FAQ page around Lexical collaborative mode

StyleT:
## Description

This adds FAQ page with to address common pitfalls around Yjs integration. This is a first PR in a series that shall improve documentation in this area.

**Closes:** #4442

## Test plan

See Vercel bot preview

### DIFF
--- a/packages/lexical-website/docs/collaboration/faq.md
+++ b/packages/lexical-website/docs/collaboration/faq.md
@@ -1,0 +1,158 @@
+---
+sidebar_position: 2
+---
+
+# Collaboration FAQ
+
+## Source of truth: Lexical State, Yjs and App's DB
+
+It's recommended to treat the Yjs model as the source of truth. You can store the document to a database for indexing.
+But, if possible, you should never forget the Yjs model, as this is the only way clients without internet access can reliably join and sync with the server.
+
+You can also treat the database as the source of truth. This is how it could be achieved:
+
+- Clients receive a `sessionId` when they connect to the server
+- When a client connects without an existing `sessionId`, get the content from the database and create a `sessionId`
+- When all clients disconnect, forget the room content and `sessionId` on the server after some timeout (e.g. 1 hour)
+- When a client reconnects, use that content on the server. Furthermore, get the `sessionId` from the client
+- When two clients with different `sessionId` reconnect, one of the clients should forget the room content. _In this case the client will lose content_ - although it is very unlikely if you set the forget timeout (see point 2) very high.
+
+Or, there is an ever simpler approach:
+
+- When a client connects to the server, the server populates the room content if empty
+- When all clients disconnect, the server forgets the room content after some timeout (e.g. 1 hour)
+- When a client was not able to reconnect for 40 minutes, the client must forget its local updates and start fresh (this should be enforced by the server)
+
+When the database is the source of truth, and if you want to be able to forget the Yjs model, you will always run into cases where clients are not able to commit changes. That's not too bad in most projects. It somehow limits you, because you can't cache the document on the client using y-indexeddb. On the other hand, it is much easier to maintain, and do Yjs upgrades. Furthermore, most people would say that SQL is a bit more reliable than Yjs.
+
+_* Based on the advice of the Yjs author - [Kevin Jahns](https://github.com/yjs/yjs/issues/82#issuecomment-328365015)_
+
+
+## Initializing `EditorState` from Yjs Document
+
+It's achievable by leveraging headless Lexical and no-op provider for Yjs:
+
+<details>
+  <summary>createHeadlessCollaborativeEditor.ts</summary>
+
+  ```typescript
+  import type {Binding, Provider} from '@lexical/yjs';
+  import type {
+    Klass,
+    LexicalEditor,
+    LexicalNode,
+    LexicalNodeReplacement,
+    SerializedEditorState,
+    SerializedLexicalNode,
+  } from 'lexical';
+
+  import {createHeadlessEditor} from '@lexical/headless';
+  import {
+    createBinding,
+    syncLexicalUpdateToYjs,
+    syncYjsChangesToLexical,
+  } from '@lexical/yjs';
+  import {type YEvent, applyUpdate, Doc, Transaction} from 'yjs';
+
+  export default function headlessConvertYDocStateToLexicalJSON(
+    nodes: ReadonlyArray<Klass<LexicalNode> | LexicalNodeReplacement>,
+    yDocState: Uint8Array,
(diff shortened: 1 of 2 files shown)


packages/lexical-website/docs/collaboration/faq.md
```diff
@@ -0,0 +1,146 @@
...
+It's recommended to treat the Yjs model as the source of truth. You can store the document to a database for indexing.
+But, if possible, you should never forget the Yjs model, as this is the only way clients without internet access can reliably join and sync with the server.
+
+You can also treat the database as the source of truth. This is how it could be achieved:
+
+- Clients receive a `sessionID` when they connect to the server
+- When a client connects without an existing `sessionID`, get the content from the database and create a `sessionID`
+- When all clients disconnect, forget the room content and `sessionID` on the server after some timeout (e.g. 1 hour)
+- When a client reconnects, use that content on the server. Furthermore, get the `sessionID` from the client
+- When two clients with different `sessionID` reconnect, one of the clients should forget the room content. _In this case the client will lose content_ - although it is very unlikely if you set the forget timeout (see point 2) very high.
```
etrepum, line 18: nit: s/sessionID/sessionId/g to match the typical case style in lexical

packages/lexical-website/docs/collaboration/faq.md
```diff
@@ -0,0 +1,146 @@
...
+          );
+        }
+      }
+    );
+
+    binding.root.getSharedType().observeDeep((events, transaction) => {
+      if (transaction?.origin !== binding) {
+        syncYjsChangesToLexical(binding, provider, events);
+      }
+    });
```
etrepum, line 125: I think this should return an `() => void` that unregisters and unobserves otherwise it could possibly leak

packages/lexical-website/docs/collaboration/faq.md
```diff
@@ -0,0 +1,146 @@
...
+      nodes,
+    });
+
+    const id = 'main';
+    const doc = new Doc();
+    const docMap = new Map([[id, doc]]);
+    const provider = createNoOpProvider();
+    const binding = createBinding(editor, provider, id, doc, docMap);
+
+    registerCollaborationListeners(editor, provider, binding);
```
etrepum, line 83: this should be done somewhere else where it can be more easily unregistered, the unregistration function should be in the return type, or perhaps unregistration should be scheduled as a microtask. another option would be to make this a HOF (e.g. `withHeadlessCollabEditor`) that provides an editor to a callback function

packages/lexical-website/docs/collaboration/faq.md
```diff
@@ -0,0 +1,146 @@
...
+- When a client connects to the server, the server populates the room content if empty
+- When all clients disconnect, the server forgets the room content after some timeout (e.g. 1 hour)
+- When a client was not able to reconnect for 40 minutes, the client must forget its local updates and start fresh (this should be enforced by the server)
+
+When the database is the source of truth, and if you want to be able to forget the Yjs model, you will always run into cases where clients are not able to commit changes. That's not too bad in most projects. It somehow limits you, because you can't cache the document on the client using y-indexeddb. On the other hand, it is much easier to maintain, and do Yjs upgrades. Furthermore, most people would say that SQL is a bit more reliable than Yjs.
+
+_* Based on the advice of the Yjs author - [Kevin Jahns](https://github.com/yjs/yjs/issues/82#issuecomment-328365015)_
+
+
+## Initializing `EditorState` from Yjs Document
```
etrepum, line 31: Maybe we should add a simple collab editor example project?

packages/lexical-website/docs/collaboration/faq.md
```diff
@@ -0,0 +1,146 @@
...
+
+It's achievable by leveraging headless Lexical and no-op provider for Yjs:
+
+<details>
+  <summary>createHeadlessCollaborativeEditor.js</summary>
```
etrepum, line 36:
```suggestion
  <summary>createHeadlessCollaborativeEditor.ts</summary>
```

packages/lexical-website/docs/collaboration/faq.md
StyleT: Working on it ;) I have a branch already

etrepum (commented):
Looks great!

packages/lexical-website/docs/collaboration/faq.md
```diff
@@ -0,0 +1,158 @@
+---
+sidebar_position: 2
+---
+
+# Collaboration FAQ
+
+## Source of truth: Lexical State, Yjs and App's DB
+
+It's recommended to treat the Yjs model as the source of truth. You can store the document to a database for indexing.
+But, if possible, you should never forget the Yjs model, as this is the only way clients without internet access can reliably join and sync with the server.
```
acywatson, line 10: I don't understand the meaning of this statement - why would they forget it? Maybe we can make it more clear somehow.

acywatson (approved):

acywatson:
I think this is a great start. My overall feedback is to start even simpler when we write deeper docs about this, as the overall suggested architecture is unclear (what's the Yjs model vs the document, how does one map onto the other, etc)

packages/lexical-website/docs/collaboration/faq.md
etrepum: I think it emphasizes the first sentence. Could perhaps be worded like this:

> We recommend that you treat the Yjs model as the source of truth. Yjs allows clients that are offline (or have unstable connections) to make local edits and have the document reliably synchronized with your server and other clients when connection is re-established.

etrepum:
I think the best supplement for this kind of documentation would be a working example that shows best practices, e.g. persistent yjs truth + database for indexing/snapshots. Possibly even have two models of the headless collab support since it's likely that people will want to use a SaaS collab channel like liveblocks.io or reflect.net so they may want a sync example that is a headless client rather than being directly coupled to a y-websocket-server service.

The helper functions here also seem like a great candidate to be included in @lexical/yjs (which also simplifies the docs and examples). Since @lexical/headless is so small I doubt adding it as a dependency would be a problem?

StyleT:
@acywatson @etrepum Thanks for feedback! I'm merging this as is for now and will do follow up later next week or sooner. 

My plan is to make at least:
- More docs
- Dedicated Yjs example, based on WSS/WebRTC + storage on a file system + awareness 
- Extract Yjs from React package so it can be used effectively with or w/o React
- Add collab mode to https://playground.lexical.dev/ (I have it working w/o server via WebRTC). Now it's only accessible when run locally
- (stretch) Add public WSS server for https://playground.lexical.dev/ so it works cross browser

PR #5996, opened 2024-04-30
[lexical] Chore: make unable to find active editor state error more descriptive about potential dependency issue

AlessioGr:
This error message can be kinda similar to the React hooks error message:

> Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
> See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.

E.g. If people have multiple versions of lexical installed, where the lexical editor is initialized by one version of lexical, but their own custom nodes are initialized by a different version, this error message will occur.

This happens to SO many people after they upgrade my package and forget to make sure their lexical version matches the one which my package is using.

Adding this hint to the error message will give them a chance to figure out why this might be happening


scripts/error-codes/codes.json
```diff
@@ -14,7 +14,7 @@
   "12": "Expected table cell",
   "13": "Cannot use method in read-only mode.",
   "14": "One or more transforms are endlessly triggering additional transforms. May have encountered infinite recursion caused by transforms that have their preconditions too lose and/or conflict with each other.",
-  "15": "Unable to find an active editor state. State helpers or node methods can only be used synchronously during the callback of editor.update() or editorState.read().",
+  "15": "Unable to find an active editor state. State helpers or node methods can only be used synchronously during the callback of editor.update() or editorState.read(). This issue can also occur if you have a mismatching or more than one copy of lexical in the same app.",
```
acywatson, line 17: If I'm not mistaken, this file is actually generated and you have to change this at the invariant callsite, then run npm run extract-codes

etrepum: that is correct, this is a generated file

AlessioGr: @etrepum @acywatson check it out again! Had to fix the extract-codes script first, as it was sending files like package.json to babel through the ast parser which then threw an error. Now it only handles ts/js files.

codes.json got a lot smaller now, strings like "getLatest() on clone node" are not present anymore. But I also cannot find them anywhere else in the codebase, so that's probably expected

packages/lexical-devtools/src/lexicalForExtension.ts
```diff
@@ -28,7 +28,9 @@ function getActiveEditorState(): lexical.EditorState {
       'Unable to find an active editor state. ' +
         'State helpers or node methods can only be used ' +
         'synchronously during the callback of ' +
-        'editor.update() or editorState.read().',
+        'editor.update() or editorState.read(). ' +
+        'This issue can also occur if you have a mismatching ' +
```
StyleT, line 32: IMO, this is more technically correct:

```suggestion
        'This issue can also occur if you have more than one copy of lexical in the same JS bundle ' +
        'or you're trying to update/read state of lexical that belongs to another JS bundle.'
```

StyleT:
If it happens often to the users of your library - may I ask if you place Lexical as dependency or peerDependency in package.json? It shall be the latter with range of versions supported

etrepum (commented):
Based on the output of codes.json I don't think this script is working correctly (probably not your fault). I would expect only new codes to be added to the map, and old codes should have the same index.

The reason for this is that the website's lookup tool should work with any prod version of lexical. In order for that to be the case, the ids for any given error message must be stable, and old entries should persist even if they are no longer present in the source code.

I think in this case the best approach for a PR would be to only change the error messages in the source code, and leave the matter of updating the error-codes script (and the json file) to work correctly can be handled separately.

PR #5999, opened 2024-05-01
fix(build): clean up of build-www rewriteImports (follow-up to #5995)

etrepum:
## Description

Follow-up to #5995, while this fix worked it was not as clean as it should be and probably shouldn't pass flow type checks if it was analyzed. I did a little more digging and I found a more correct incantation, but it could probably still use some improvement if there is a hermes-transform expert around.

### Before

node ./scripts/www/rewriteImports.js rewrites flow files correctly (using sketchy code)

### After

node ./scripts/www/rewriteImports.js rewrites flow files correctly (using slightly better code)

### DIFF
--- a/scripts/www/__tests__/unit/transformFlowFileContents.test.js
+++ b/scripts/www/__tests__/unit/transformFlowFileContents.test.js
@@ -1,0 +1,107 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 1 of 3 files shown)


acywatson (approved):

etrepum:
I'll take a look at that approach and write some unit tests

acywatson (approved):

PR #6001, opened 2024-05-01
Clamp number of registered editor above 0

thegreatercurve:
## Description
Since #5070, there are certain circumstances in which the document counts for `rootElementsRegistered` can fall below 0, especially if `removeRootElementEvents` is called more than `addRemoveRootElementEvents`. 

As number of registered root elements should never be fewer than 1 after `addRemoveRootElementEvents` is called, and because the number of root elements should never be fewer than 0 when `removeRootElementEvents` is called, we should clamp these values.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -1212,7 +1212,10 @@ export function addRootElementEvents(
   }
   rootElementsRegistered.set(
     doc,
-    Math.max(documentRootElementsCount ?? 0, 0) + 1,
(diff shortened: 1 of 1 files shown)


acywatson (approved):

acywatson (approved):

Daniel15:
We should probably have an invariant rather than just clamping, so we can more easily find any bad code that causes this condition.

thegreatercurve:
Closing in favour of https://github.com/facebook/lexical/pull/6003

PR #6002, opened 2024-05-01
Fix add root element count

zurfyx:
Element precendence https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence

Additive > logical or

For reference, the second time

```
> 1 || 0 + 1
1
```

PR #6003, opened 2024-05-01
Root element count invariant & test

zurfyx:
Follow up to https://github.com/facebook/lexical/pull/6002


thegreatercurve (approved):

wonkyum-kim:
Hi. I found that there is no `expect` in the test code you wrote. Is this intended?

PR #6005, opened 2024-05-02
[build] Refactor: move error-codes generation to existing babel plugin

etrepum:
## Description

Previously we had a rollup plugin and separate a babel plugin for managing error codes, which are a map of sequential integers to error message strings that are used in minified prod builds so that we are not shipping verbose error message strings.

Both plugins essentially did the same work, but the updating of the map was only done in the rollup plugin. This consolidates all of the functionality to the babel plugin which is now able to conditionally update the error map if the `--codes` argument was present.

This PR also updates the postversion script to ensure that codes are generated before a release and ensures that the versions and package-lock.json are updated accordingly.

Code extraction is now integrated into `npm run build-release` (which is a dependency of `npm run prepare-release` and `npm run release`) as a failsafe to ensure codes are updated before release.

I chose not to update the filename for codes.json in this PR. I think we can do that separately in a follow-up to address the concern that the filename should let people know that it is generated. I think we should do a release with this first before changing too many more things.

### Before

`npm run error-codes` was supposedly in a state where it no longer worked (re #5996)

It was also easy to forget to do this before creating a release, which would leave error messages unextracted (larger prod build).

### After

`npm run increment-version` and `npm run build-release` (and anything downstream such as `npm run prepare-release` or `npm run release`) will generate the scripts/error-codes/codes.json file.

The new unit tests should also pass!


etrepum:
@ivailop7 this one would be nice to have in the release as well, especially if there is some issue running extract-codes without it

acywatson (approved):

ivailop7 (approved):

PR #6007, opened 2024-05-02
[lexical-playground][TableCellRecizer] Bug Fix: Correctly take zoom into account when calculating table drag zones

apella:
## Description
In our implementation of this editor, the user can modify the zoom of the editor area independently of the rest of the page. This skews the selection areas of the table cell resizer. This commit fixes that.

It also slightly shifts the rectangles to *center* on the edge, instead of being outside the cell, and scales these rectangles proportionally. This makes for a more intuitive user experience as the user will try to select the edge, not the outside.

**Closes:** #

### Before
lexical-zoom-original

cell-resizer-old-regions

### After
lexical-zoom-fixed

(selection zones were temporarily coloured yellow to display the zones. Notice how they now overlap with the cell border)
cell-resizer-new


apella:
Closing this in preference of a more succinct solution  suggested by @ebads67

PR #6014, opened 2024-05-02
[lexical] Feature: error checking for node key re-use with type mismatch in __DEV__

etrepum:
## Description

Based on issues that several people have run into, it seems useful to add some additional invariant checking around node keys.

See also: [discord message](https://discord.com/channels/953974421008293909/953974421486436393/1235675644726083745)

- Adds a stub implementation of shared/invariant for use under jest to make development easier
- Adds a `__DEV__` check to make sure that if a new Node is created with an explicit existing key, that the type of the node did not change

### Before

Bad things will silently happen if you accidentally re-use node keys, such as with a subtly incorrect node replacement function:

```js
const config = {
  // ...
  nodes: [
    EdifactTextNode,
    {
      replace: TextNode,
      with: (node) => {
        return new EdifactTextNode(node.text, node.__key)
      },
    },
  ],
}
```

### After

An error message is shows up now, in `__DEV__.` There are unit tests that verify that the specific error message comes up when creating a new node in this way (tested in isolation and with a node replacer).

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -15,12 +15,17 @@ const common = {
 };
 
 // Use tsconfig's paths to configure jest's module name mapper
-const moduleNameMapper = Object.fromEntries(
-  Object.entries(tsconfig.compilerOptions.paths).map(([name, [firstPath]]) => [
-    `^${name}$`,
-    firstPath.replace(/^\./, '<rootDir>'),
-  ]),
-);
+const moduleNameMapper = {
+  ...Object.fromEntries(
+    Object.entries(tsconfig.compilerOptions.paths).map(
+      ([name, [firstPath]]) => [
+        `^${name}$`,
+        firstPath.replace(/^\./, '<rootDir>'),
+      ],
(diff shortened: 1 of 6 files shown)


potatowagon (approved):
looks good 👍

acywatson:
I have a small concern about the size here - would be nice to be able to tree-shake this stuff out of the prod build.

acywatson:
> I have a small concern about the size here - would be nice to be able to tree-shake this stuff out of the prod build.

Discussed offline - we'll follow up and ensure that we're doing what we can to shake out these DEV things in rollup

etrepum:
Just for posterity's sake I did confirm that unused functions that do not get exported are tree-shaken correctly in both the `.mjs` and `.js` prod builds at this time. Had this functionality not worked, it would also show up elsewhere (e.g. handleDEVOnlyPendingUpdateGuarantees is only used under `__DEV__` in LexicalUpdates.ts)

Issue #6012, opened 2024-05-02
Chore: Update vite, rollup, and associated build-related packages


etrepum:
The build currently uses rollup@2.75.7 and vite@2.9.16 which are probably about a year behind (both have had multiple major version number increments since then).

PR #6018, opened 2024-05-03
[build] Chore: Upgrade rollup and vite to latest versions

etrepum:
## Description

rollup and vite have not been upgraded in some time, it makes sense to keep them up to date

Packages upgraded:
* rollup (and all plugins)
* vite (and all plugins)
* eslint (and all plugins) - but only up to 8
  * Switched from flowtype to ft-flow due to change in fb preset.
  * Switched from deprecated babel-eslint to @babel/eslint-parser
* docusaurus

I also upgraded most of the workflow actions because I was getting a spurious cache failure. I did not update to actions/upload-artifact@v4 which seemed like it might need some other config changes https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes

Other changes:

* Lint fixes for the new eslint-plugin-simple-import-sort heuristics
* playground's `setupEnv.ts` was getting tree shaken in prod because it was only performing side-effects so I refactored it and the settings panel which didn't really work correctly previously
* playground now loads the dev (vercel, most tests) or prod (specific e2e tests) version of lexical from built artifacts rather than directly compiling the typescript (except when in serve mode, e.g. npm run start, so you can use hmr while developing locally)

**Closes:** #6012

### Before

Builds work and pass all tests

### After

Builds still work and still pass all the tests


ivailop7:
Thank you for doing this!

etrepum:
@ivailop7 @zurfyx one thing I noticed is that we were not currently deploying the prod build to the playground (probably after the esm refactor), but I just restored that behavior here. We were using a module resolution strategy that pointed vite directly to the typescript source files instead of the build products in dist.

I think maybe we should consider using the dev build of lexical on vercel, since it's much friendlier (in particular with regard to the tree view command history and potential error messages) and leave the prod verification to CI tasks… what do you all think? (RE #2460)

ivailop7:
Yeah, I noticed we had the command names in the Playground which are lost when it's a prod build. It was after the ESM change. I agree, I think Dev build for PRs and Prod for the Playground makes sense.

potatowagon (approved):
great work thankyou for this upgrade!

zurfyx:
> @ivailop7 @zurfyx one thing I noticed is that we were not currently deploying the prod build to the playground (probably after the esm refactor), but I just restored that behavior here. We were using a module resolution strategy that pointed vite directly to the typescript source files instead of the build products in dist.
> 
> I think maybe we should consider using the dev build of lexical on vercel, since it's much friendlier (in particular with regard to the tree view command history and potential error messages) and leave the prod verification to CI tasks… what do you all think? (RE #2460)

Thank you for doing this Bob! I think that's reasonable, the point of the prod build since we first introduced in the CI has always been to have some degree of certainty on a build that is slightly different than dev. We had some issues in the past with the rollup bundling that we didn't catch until it broke in prod. Currently, I think that the CI step that runs all E2E on prod already gives us this confidence so we can move on with your proposal.

zurfyx (approved):

scripts/npm/packages.js
```diff
@@ -1,45 +0,0 @@
-#!/usr/bin/env node
```
zurfyx, line 1: I assume this is an unrelated cleanup

packages/lexical-utils/src/__tests__/unit/LexicalNodeHelpers.test.ts
```diff
@@ -207,7 +207,7 @@ describe('LexicalNodeHelpers tests', () => {
     test('DFS of empty ParagraphNode returns only itself', async () => {
       const editor: LexicalEditor = testEnv.editor;
 
-      let paragraphKey;
+      let paragraphKey: string;
```
zurfyx, line 210:
```suggestion
      let paragraphKey: NodeKey;
```

packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -26,11 +36,9 @@ export const LEGACY_EVENTS = process.env.E2E_EVENTS_MODE === 'legacy-events';
 export const SAMPLE_IMAGE_URL =
   E2E_PORT === 3000
     ? '/src/images/yellow-flower.jpg'
-    : '/assets/yellow-flower.a2a7c7a2.jpg';
+    : findAsset('yellow-flower*.jpg');
```
zurfyx, line 39: So much better, finding the suffix was always time-consuming

zurfyx (approved):
I noticed that the port is 5173 now, I wonder how hard it is to make port 3000 back, no strong opinions though. Merging now to avoid the rebase churn.

scripts/npm/packages.js
etrepum: Yes, this was supposed to be deleted in #5876

packages/lexical-utils/src/__tests__/unit/LexicalNodeHelpers.test.ts
etrepum: honestly I only did it this way because there's a plethora of other tests that use string for keys, might be a good lint for us to find variables named `/[Kk]ey$/` with a string type!

PR #6020, opened 2024-05-03
[lexical-markdown][lexical-playground] Feature: Option to include blanklines in markdown render

potatowagon:
## Description

option to make the number of blank lines consistent between markdown and the rendered markdown

**Closes:** #

To enable:
```
$convertFromMarkdownString(
  ...,
  true // shouldPreserveNewLines -> set to true,
)

$convertToMarkdownString(
   ...,
   true // shouldPreserveNewLines -> set to true,
);
```

These flags are optional. Defaults to old behavior of removing newlines on import and delimiting with newlines on export.

### Before

https://github.com/facebook/lexical/assets/19604232/108fdd80-0cde-4678-b4dd-b301132082fa

blank lines are ignored and removed

### After

https://github.com/facebook/lexical/assets/19604232/b8ce7e72-e1c2-4b60-9b96-c92a95791faf

number of blank lines consistent between markdown and rendered markdown

---
### Added settings in playground to toggle newline preservation

https://github.com/facebook/lexical/assets/19604232/3e4d6c1e-e25a-45d7-8cdc-5a77a4afce08

---

Unit and e2e tests pass locally

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -26,9 +26,10 @@ import {isEmptyParagraph, transformersByType} from './utils';
 
 export function createMarkdownExport(
   transformers: Array<Transformer>,
-  isNewlineDelimited: boolean = true,
+  shouldPreserveNewLines: boolean = false,
 ): (node?: ElementNode) => string {
   const byType = transformersByType(transformers);
+  const isNewlineDelimited = !shouldPreserveNewLines;
 
   // Export only uses text formats that are responsible for single format
   // e.g. it will filter out *** (bold, italic) and instead use separate ** and *
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -26,6 +26,7 @@ import {isEmptyParagraph, transformersByType} from './utils';
 
 export function createMarkdownExport(
   transformers: Array<Transformer>,
+  isNewlineDelimited: boolean = true,
 ): (node?: ElementNode) => string {
   const byType = transformersByType(transformers);
 
@@ -51,7 +52,8 @@ export function createMarkdownExport(
       if (result != null) {
         output.push(
           // seperate consecutive group of texts with a line break: eg. ["hello", "world"] -> ["hello", "/nworld"]
-          i > 0 &&
+          isNewlineDelimited &&
+            i > 0 &&
             !isEmptyParagraph(child) &&
             !isEmptyParagraph(children[i - 1])
             ? '\n'.concat(result)
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -128,7 +128,7 @@ function exportChildren(
         exportTextFormat(child, child.getTextContent(), textTransformersIndex),
       );
     } else if ($isElementNode(child)) {
-      // empty para returns ""
+      // empty paragraph returns ""
       output.push(
         exportChildren(child, textTransformersIndex, textMatchTransformers),
       );
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -22,7 +22,7 @@ import {
   $isTextNode,
 } from 'lexical';
 
-import {transformersByType} from './utils';
+import {isEmptyParagraph, transformersByType} from './utils';
 
 export function createMarkdownExport(
   transformers: Array<Transformer>,
@@ -39,7 +39,8 @@ export function createMarkdownExport(
     const output = [];
     const children = (node || $getRoot()).getChildren();
 
-    for (const child of children) {
+    for (let i = 0; i < children.length; i++) {
+      const child = children[i];
       const result = exportTopLevelElements(
         child,
         byType.element,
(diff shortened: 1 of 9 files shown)


fantactuka:
LGTM, I believe there were some PRs in past with similar idea. More broader discussion (covered it a bit offline with Gerard & folks): it seems like best route for MD plugin is to eliminate out of the box import and export parsing and instead start relying on external library (e.g., remark), while plugin will only responsible to connect Lexical and 3rd party lib. That would allow delegating this kind of configuration to those libs.

Shortcuts is a bit more lightweight functionality (and handles simple formats only) so it could still remain a part of the plugin (that way we're not bloating editor with full MD parsing library for simple shortcuts).

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -22,10 +22,11 @@ import {
   $isTextNode,
 } from 'lexical';
 
-import {transformersByType} from './utils';
+import {isEmptyParagraph, transformersByType} from './utils';
 
 export function createMarkdownExport(
   transformers: Array<Transformer>,
+  isNewlineDelimited: boolean = true,
```
zurfyx, line 29: I believe the two are the same feature, do we need a different flag name for these?

I would probably call it `preserveNewLines`

packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -92,20 +100,6 @@ export function createMarkdownImport(
...
-  if (!$isParagraphNode(node)) {
-    return false;
-  }
-
-  const firstChild = node.getFirstChild();
-  return (
-    firstChild == null ||
-    (node.getChildrenSize() === 1 &&
-      $isTextNode(firstChild) &&
-      MARKDOWN_EMPTY_LINE_REG_EXP.test(firstChild.getTextContent()))
```
zurfyx, line 105: Is this intentional? I believe this regex is particularly important on top of the default Node isEmpty

packages/lexical-markdown/src/MarkdownExport.ts
potatowagon: that makes sense to standardise the flag name for both import and export, will change!

packages/lexical-markdown/src/MarkdownImport.ts
potatowagon: i moved this to packages/lexical-markdown/src/utils.ts for reuse

packages/lexical-markdown/src/utils.ts
```diff
@@ -429,3 +435,19 @@ export function transformersByType(transformers: Array<Transformer>): Readonly<{
...
+  }
+
+  const firstChild = node.getFirstChild();
+  return (
+    firstChild == null ||
+    (node.getChildrenSize() === 1 &&
+      $isTextNode(firstChild) &&
+      MARKDOWN_EMPTY_LINE_REG_EXP.test(firstChild.getTextContent()))
+  );
+}
```
potatowagon, line 453: @zurfyx here it is

GermanJablo:
> it seems like best route for MD plugin is to eliminate out of the box import and export parsing and instead start relying on external library (e.g., remark) @fantactuka 

I'm not very familiar with Remark, but I have a hard time seeing how a generic library could correctly parse Lexical's editorState.

Separate topic, I do like this PR. In my app I use markdown for the tests, and to make the conversion idempotent I do something quite dirty: I add a specific string to the empty paragraphs, and after the conversion is done I delete them.

packages/lexical-markdown/src/MarkdownExport.ts
potatowagon: addressed in https://github.com/facebook/lexical/pull/6020/commits/89a7b30eb41f9922355f5d19bebe0abc98d981b3

potatowagon:
note: there is a change to blockquote behavior.

with shouldPreserveNewLines = false (old behavior)

>hello

text

becomes

```
>hello

text
```
because of newline delimeter

---
with shouldPreserveNewLines = true (new optional behavior)

>hello

text

^imagine no newlines seperating them

becomes

```
>hello
>text
```

tho this is consistent with other markdown surfaces such as github & [dillinger](https://dillinger.io/) where texts right below blockquotes get included in the blockquote

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -116,6 +128,7 @@ function exportChildren(
         exportTextFormat(child, child.getTextContent(), textTransformersIndex),
       );
     } else if ($isElementNode(child)) {
+      // empty para returns ""
```
zurfyx, line 131: If you think this comment adds some value, let's write "paragraph" instead

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -22,12 +22,14 @@ import {
   $isTextNode,
 } from 'lexical';
 
-import {transformersByType} from './utils';
+import {isEmptyParagraph, transformersByType} from './utils';
 
 export function createMarkdownExport(
   transformers: Array<Transformer>,
+  shouldPreserveNewLines: boolean = false,
```
zurfyx, line 29: Do you think the `preserveNewLines` comment we discussed applies here?

zurfyx (dismissed):

packages/lexical-markdown/src/MarkdownExport.ts
potatowagon: quite confused what u mean as its already been applied in https://github.com/facebook/lexical/commit/89a7b30eb41f9922355f5d19bebe0abc98d981b3 ,

`isNewlineDelimited` replaced with `shouldPreserveNewLines`,

`preserveNewLines` is named as `shouldPreserveNewLines` to be consistent with boolean naming conventions

packages/lexical-markdown/src/MarkdownExport.ts
potatowagon: addressed in https://github.com/facebook/lexical/pull/6020/commits/6a4285398df3599ec04602bb6e3290d5e00137d0

Sahejkm (approved):
LGTM

PR #6024, opened 2024-05-04
[lexical-playground] Chore: add test for pasting over mentions

potatowagon:
## Description
add an e2e test for pasting over mentions. part of a sev followup task for lexical crashing when pasting over mentions. fix was in #5954

## Test plan

npm run start & npm run test-e2e-chromium

Screenshot 2024-05-23 at 12 33 36 PM

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Mentions.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Mentions.spec.mjs
@@ -6,12 +6,11 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


zurfyx:
Thank you for following up on this 🙏

potatowagon:
@zurfyx im good to close this if unit test is prefered, we discuss which unit test file would be best to put copy paste tests

zurfyx (dismissed):

Sahejkm (approved):

PR #6026, opened 2024-05-04
[playground] Bug Fix: change default dev port to 3000

etrepum:
## Description

`npm run dev` (and associated `npm run start`) used to start on port 3000, which was the old vite default. The default changed to 5173, but no reason we can't keep the previous behavior https://github.com/facebook/lexical/pull/6018#pullrequestreview-2039349437

### Before

`npm run dev` listens on port 5173 (after #6018 was merged)

### After

* `npm run dev` listens on port 3000
* tests and other commands that also specify a port still get the desired choice

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "start": "cross-env NODE_ENV=development concurrently \"npm:collab\" \"npm run dev --prefix packages/lexical-playground\"",
     "start:website": "npm run start --prefix packages/lexical-website -- --port 3001",
     "start:playground": "npm run start-test-server",
-    "dev": "npm run dev --prefix packages/lexical-playground",
+    "dev": "npm run dev --prefix packages/lexical-playground --",
     "start-test-server": "npm run preview --prefix packages/lexical-playground -- --port 4000",
     "build": "node scripts/build.js",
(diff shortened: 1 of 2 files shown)


ivailop7 (approved):

ivailop7:
Does it make sense to put the port in the vite config files, rather than extend the npm scripts. I think it's cleaner, no strong opinion on this. Approved.

zurfyx (approved):
Thank you for quick follow-up!

etrepum:
@ivailop7 I also don't have a strong opinion about it either way. `server: {port: 3000}` in the vite.config.ts does that. If you specify multiple ports on the command line it takes the last one, so practically speaking other than where the setting is it's not a very observable difference. We manually specify the port elsewhere in the package.json files so I just did it in the same place.

ivailop7:
> @ivailop7 I also don't have a strong opinion about it either way. `server: {port: 3000}` in the vite.config.ts does that. If you specify multiple ports on the command line it takes the last one, so practically speaking other than where the setting is it's not a very observable difference. We manually specify the port elsewhere in the package.json files so I just did it in the same place.

Fair enough

PR #6027, opened 2024-05-04
Fix OverflowNode configuration

zurfyx:
OverflowNode was badly configured, this was revealed by https://github.com/facebook/lexical/pull/6025. It is always inline and can't be empty. Surprisingly, the current configuration was passing all tests fine but the video below show some surprisingly bad behavior when we only fix isInline, we can't have dangling empty span.

Before:

https://github.com/facebook/lexical/assets/193447/7915df5e-944d-49d6-80ca-b9c6061a3434

After:

https://github.com/facebook/lexical/assets/193447/0b39fd28-794d-4a15-bafa-a98d0d76876a


ivailop7 (approved):

etrepum:
#6456 may also fix this incidentally because it adds a transform to prevent empty OverflowNode

etrepum (approved):

PR #6028, opened 2024-05-04
[lexical-playground][file] Feature: Playground link sharing

etrepum:
## Description

Adds a share button to the playground's ActionToolbar which will create a #doc= URL that includes a compressed and serialized document. The plugin also has an effect that will set the editor state from this document and clear history on initial load. It changes the URL when the share button is clicked as well, I just copied the [typescript playground](https://www.typescriptlang.org/play/) UX for this, I think it's done as a fallback for if the clipboard.writeText fails.

I added the following new exports to @lexical/file to support this feature:

* type SerializedDocument
* serializedDocumentFromEditorState
* editorStateFromSerializedDocument

Here's an [example link with #doc= in the preview](https://lexical-playground-git-fork-etrepum-playgro-fc2d4a-fbopensource.vercel.app/#doc=H4sIAAAAAAAAE91X22ocRxD9laL94Jdh1_IFOwOGGBSDIOAQrUAQ8lDbXTPdUW330FWjWcXoX_Qt-jLTM44ua-OEzUS28zDsdHX3qeuprXlvyAVN-VhRydTvTU5Jy6_1gV2maOrfdhaOFAOb-kllmpQ3qOPrJjkytYlFwqYyohdcBKYySls1tTkksTl0GlIswouubI97lTmnLEVeH1z-XhkXMtnxXG1Ys7lVZEQxlwshOoqT5o9IntCF2N4Dq4xiW_aemstqJjfeOCeAIB4zwbpXTRE0gXqCjvGizamP7rHAm9GDVUq8xgyDD9bDEJjBZkIlQPhSFKr9DXzkkr2F9ii-BGEu9Ndw8uvPoB4VQrTcOyrRsGnTZRIhBxgdCOWAHP4kBy7ZfkNRF7AaI9S3IQKyJPAogBGoacjqhDjGR0jHaE6FCVIqE5qcNqA-yA3gqMgyYQYfRFO-gBQhxKABGTihW8CRgvUYW5IRsVg-eIrj4l4Cg4DlYM-K_QIDMVdwBH_0omBTF8iVK9dXf5OxWeqrKJiIcqee9ifMZ7nCIZ7tECUTj3Y1iTkNRR_mltTUsWeujAYtdk6LPpezXrWTerkchmFxazRjbBcpt8ti_NL8m1K7vjo5hSblMe8lH-pDPIOgj0sVRCqpQmiQeY32bDwZmjG3lkO3TpjdYshBaUVbhQYDy-K_6DwdZmwzdn6399BW3_7l8Gz1cQTo3FSQMCUrxBYiDUDbLmWV0o1-ZNoGi7xsAlMRSN-VzYlCDaH2mepvJxizMQeOb1rP4cdGMTt3RIPSZse_c-Se5uwD8okjb3Pa_HTn3_oh_Xo6m1935o3i0FfO17N9IXfgimg17a57ZtLR90KS2ymk5_mmkNW7w3f_N_5eXymJynfK1-srhC6njjI0jOJhQyLYEpycPjBTv9OKJhknnm_i42A2jj5wMX_N1O-h-581mz2Ax6_Ye8G-rAyj6DGekzP1wcuD569e_PDqWXkqI6nPttz75e7EfXPbPFkcPF-8MJcfAE3nMT0uDwAA)

## Test plan

* Change some content from the editor's default state
* Click the share button
* Expect the URL to change with a `#doc=...` hash, the URL to be on the clipboard, and the "URL copied to clipboard" flash message to show
* Paste it into another browser (or make some changes and reload) and expect that the previously shared state to be restored (only the state, no history)

There's some unit tests that ensure round-trip of the doc<->hash and an e2e test that runs the above test plan


etrepum:
@zurfyx any thoughts on this before I finish up the UI and tests?

zurfyx (commented):
Brilliant, thank you for putting this together!

acywatson:
this is pretty cool - I'm not sure how useful it will be for real-life debugging, but I might be wrong and there's certainly no harm. Curious if you had particular use cases/debugging friction points in mind.

Another area of exploration might be the test generator - I think it's defunct and never worked super-well, but the idea was to be able to start "recording mode" and it would generate an e2e test for you as you interacted with the playground editor.

etrepum:
@acywatson I picked it off the roadmap 🤷‍♂️ Thinking about debugging use cases I think it is actually pretty useful because you can make a link to the playground from an issue with the state required to repro an issue (so long as history isn't involved in the bug). You can also use it to demonstrate a playground state that shows off a new node or feature. It's also in just about every other playground site.

acywatson:
> @acywatson I picked it off the roadmap 🤷‍♂️ Thinking about debugging use cases I think it is actually pretty useful because you can make a link to the playground from an issue with the state required to repro an issue (so long as history isn't involved in the bug). You can also use it to demonstrate a playground state that shows off a new node or feature. It's also in just about every other playground site.

Makes sense!

> so long as history isn't involved in the bug

I suspect most issues actually involve a sequence of interactions against a particular initial state, but at least it gets us that initial state.

Once again, no objections to this at all, more just thinking about what could be a next step to making it even easier to create shareable repros and further reduce triage/maintenance overhead.

acywatson (approved):

PR #6029, opened 2024-05-05
[CI] Refactor: clean up github workflows for cache correctness

etrepum:
## Description

We have a lot of test workflow. The workflows could get in various states where the cache was poisoned. 

The cache correctness related changes:
* Use the actions/setup-node@v4 `cache: npm` option which also considers the node version, and should in theory do the right thing (the actions/cache docs recommend to [not cache node_modules](https://github.com/actions/cache/blob/main/examples.md#node---npm) as we were doing)
* Use `npm ci`. Maybe a little slower than caching node_modules, takes about 30s, but the global artifacts should be cached and we have a postinstall action to run for wxt with the lexical-devtools package which was the cause of some of the problems.
* Move playwright caching to its own step which a cache key that includes all the things
* Add a reusable e2e-test workflow to make the multiple matrix jobs easier to manage

### Before

Sometimes get strange build failures due to poisoned cache (hard to intentionally repro)

### After

* All expected matrix combinations are still tested
* Tests pass
* Hopefully cache poisoning won't happen anymore

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,7 +1,8 @@
 name: Nightly Release Branch
 on:
   # remove the [] when this is turned on
-  schedule: []
+  schedule:
+    []
     # Run daily at 2:30am UTC
     # - cron: '30 2 * * 1-5'
 jobs:
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,10 +1,10 @@
 name: Nightly Release Branch
 on:
   # remove the [] when this is turned on
-  schedule:
-    []
-    # Run daily at 2:30am UTC
-    # - cron: '30 2 * * 1-5'
+  []
+  # Run daily at 2:30am UTC
+  # schedule:
+  #  - cron: '30 2 * * 1-5'
 jobs:
   release:
     # prevents this action from running on forks
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,6 +1,7 @@
 name: Nightly Release Branch
 on:
-  schedule:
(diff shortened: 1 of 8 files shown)


packages/lexical-devtools/wxt.config.ts
```diff
@@ -127,7 +127,7 @@ export default defineConfig({
           find: 'lexicalOriginal',
           replacement: path.resolve('../lexical/src/index.ts'),
         },
-        ...moduleResolution('preview'),
+        ...moduleResolution('source'),
```
etrepum, line 130: I don't know how this didn't get caught by the integrity check when it was merged?

.github/workflows/nightly-release.yml
```diff
@@ -1,8 +1,10 @@
 name: Nightly Release Branch
 on:
-  schedule:
-    # Run daily at 2:30am UTC
-    # - cron: '30 2 * * 1-5'
+  # remove the workflow_dispatch when this is turned on
+  workflow_dispatch
```
etrepum, line 4: I did this so I would stop getting emails about it on every push

.size-limit.js
```diff
@@ -8,14 +8,13 @@
...
 /**
- * We use this file to configure the size-lmit tool, rather than their simpler
+ * We use this file to configure the size-limit tool, rather than their simpler
  * yaml package.json configuration, because we need to override the resolution
  * of modules to ensure we are pulling in monorepo build products as
  * dependencies rather than trying to use something stale from node_modules.
  */
 const glob = require('glob');
 const path = require('node:path');
-const fs = require('fs-extra');
```
etrepum, line 18: I noticed this unused import and typo when I was auditing the workflows

ivailop7 (approved):

Issue #6040, opened 2024-05-06
Bug: Cannot toggle editor isEditable using react + react-dom v19 beta


acusti:
Lexical version: v0.14.5

## Steps To Reproduce

1. install [react + react-dom v19 beta](https://www.npmjs.com/package/react?activeTab=versions) (latest dependency string is `9.0.0-beta-5d29478716-20240506`)
2. use `editor.setEditable()` API to toggle editable state

Link to code example:

https://stackblitz.com/~/github.com/acusti/lexical-rich-toggle-editable

if you switch the initial isEditable value from `false` to `true`, the editor’s content editable DOM element will remain always `contenteditable="true"`

if you swap the react + react-dom dependency strings from the v19 beta to v18.3.1, the example behaves correctly

(note that i had to update the `npm install` script to `npm install --legacy-peer-deps` in order to prevent issues with lexical’s peer dependencies that specify react 18.x, but that configuration change is already applied in the stackblitz code example above)

## The current behavior

the contenteditable state of the editor’s `<ContentEditable />` component never changes from its initial value, even when `editor.setEditable()` is called with a changed value. this is because the [registered editable listener](https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalContentEditable.tsx#L71) is never invoked, though when i create a local listener via initialConfig’s `editorState` callback, that listener is invoked every time the editor’s editable state changes.

## The expected behavior

i expect the `ContentEditable` component to get the update editable state, as occurs when not using the v19 beta.

acusti:
note that i believe this isn’t specific to changes to the editor’s `editable` but is actually a more fundamental issue where a number of the editor closure utilities don’t work as expected. for example:
```
import { $selectAll } from 'lexical';
// …
editor.update(() => {
   $selectAll();
});
```
doesn’t work using the react v19 beta but does work using v18.3.1.

etrepum:
I suspect what's happening here is due to some behavior change in React.StrictMode and the order of effects since it runs them all twice. If you build it and preview it works. In your listener it's not registering for updates during an effect so it's not subject to the double-running.

etrepum:
To be clear what I think is happening is this:

1. Effect A.dev is called, registering an editable listener
2. Effect A.dev listener is unregistered
3. Effect B.dev is called, setting the editor state, triggering any editable listeners (but not A.dev because it already unregistered)
4. Effect A.prod is called, registering an editable listener
[...]

PR #6041, opened 2024-05-06
[WIP][lexical-react] Bug Fix: React 19 StrictMode correctness

etrepum:
## Description

React 19's StrictMode re-uses the results of useMemo between renders ([notes](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#strict-mode-improvements)). We have code in the composers that uses useMemo to create LexicalEditor which are a mutable state nightmare if shared, causing all sorts of hard to understand bugs and makes us look pretty bad in dev modes everywhere. `useState(…)[0]` should not have this issue.

We can't fix the fact that you can of course make similar mistakes in your own components. We could also add documentation to the FAQ or somewhere else that shows best practices. I'm not sure exactly what practice we would recommend if you want to do some sort of editor mutation that the cleanup function doesn't take care of (e.g. set up some initial editor state with a plugin).

**Closes:** #6040

### Before

See #6040 for a full example

### After

Unit tests pass. Not sure what the best way to run the test suite with React 19 in CI would be?

### DIFF
--- a/packages/lexical-react/src/LexicalComposer.tsx
+++ b/packages/lexical-react/src/LexicalComposer.tsx
@@ -6,7 +6,10 @@
  *
  */
 
-import type {LexicalComposerContextType} from '@lexical/react/LexicalComposerContext';
+import type {
+  LexicalComposerContextType,
+  LexicalComposerContextWithEditor,
+} from '@lexical/react/LexicalComposerContext';
 
 import {
   createLexicalComposerContext,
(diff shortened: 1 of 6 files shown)


etrepum:
After thinking about this some more I think the approach we have is probably correct enough, but we have some things to address regarding React 19:

* CI should run a separate suite of tests against react@beta
* Track down deprecated APIs and replace them with React 17,18,19 compatible approaches
* Write some documentation about all the ways you can muck things up when effects are not used correctly, especially in StrictMode (e.g. trying to exfiltrate data from the editorState callback) - promoting functionality like useLexicalSubscription instead of error-prone DIY

PR #6043, opened 2024-05-06
[examples/react-rich-collab] Feature: Added collaboration focused example

StyleT:
## Description
**Changes:**
- Adds example that is focused on collaborative capabilities of Lexical 
- Updates documentation significantly in this context

## Test plan

https://github.com/facebook/lexical/assets/1125900/8bb70d72-8996-4256-9c53-ac02974b3148


examples/react-rich-collab/src/App.tsx
```diff
@@ -0,0 +1,165 @@
...
+};
+
+export default function App() {
+  const providerName =
+    new URLSearchParams(window.location.search).get('provider') ?? 'webrtc';
+  const [userProfile, setUserProfile] = useState(() => getRandomUserProfile());
+  const containerRef = useRef<HTMLDivElement | null>(null);
+  const [yjsProvider, setYjsProvider] = useState<null | Provider>(null);
+  const [connected, setConnected] = useState(false);
+  const [activeUsers, setActiveUsers] = useState<[string, string][]>([
```
etrepum, line 43: I find an object to be a little nicer than using tuples for this sort of thing. It makes the intent a lot clearer and much harder to mix them up (plus inference works better, sometimes with fixed length tuples you need to add annotations so it doesn't generalize to `T[]`).

It also makes sense to keep the clientID around (the key in getStates()) since that's the only true unique identifier

examples/react-rich-collab/src/App.tsx
```diff
@@ -0,0 +1,165 @@
...
+        <input
+          type="color"
+          value={userProfile[1]}
+          onChange={(e) => setUserProfile([userProfile[0], e.target.value])}
+        />
+      </p>
+      <p>
+        <b>Active users:</b>{' '}
+        {activeUsers.map(([name, color], idx) => (
+          <Fragment key={name + color}>
```
etrepum, line 140: there's no real guarantee that this is unique, if we don't keep the clientID around it should just be idx

examples/react-rich-collab/src/App.tsx
```diff
@@ -0,0 +1,165 @@
...
+        {/* With CollaborationPlugin - we MUST NOT use @lexical/react/LexicalHistoryPlugin */}
+        <CollaborationPlugin
+          id="lexical/react-rich-collab"
+          providerFactory={providerFactory}
+          // Optional initial editor state in case collaborative Y.Doc won't
+          // have any existing data on server. Then it'll user this value to populate editor.
+          // It accepts same type of values as LexicalComposer editorState
+          // prop (json string, state object, or a function)
+          initialEditorState={$initialEditorState}
+          shouldBootstrap={true}
```
etrepum, line 156: This is a bit of a tangent but  I would love to see some documentation about what `shouldBootstrap` is for. Without reading through a lot of code I don't think the semantics are obvious, e.g. is there some sort of race condition when two clients try and bootstrap at the same time, and this is a workaround for when you are intentionally creating this scenario with two frames that load in the same page? It feels like the sort of option that should be optional (default true I guess) and only used in weird situations. If there is a race condition, is there some way to fix that and get rid of the option? e.g. a leader election or write winning strategy based on clientID ordering?

etrepum (commented):
This looks fantastic! I think this kind of example will really help people get started with collab.

examples/react-rich-collab/src/App.tsx
StyleT: Long story short - you should do bootstrap on server, not at the client (to exactly avoid race condition). Let me elaborate on it here.

zurfyx (approved):

Issue #3468, opened 2022-11-30
Bug: TextEntity infinite loop with provided HashtagPlugin


pro-vi:
Lexical version: 0.6.4

## Steps To Reproduce

Type `#aaa #bbb` which is parsed into two `HashtagNode`s, copy, paste right after `#bbb` no space, `TextEntity` will start to loop infinitely. See recording.

https://github.com/facebook/lexical/blob/67b159ee9275e4a8f67c53b9aad97d0d97cddb8b/packages/lexical-react/src/LexicalHashtagPlugin.ts#L280-L284

In `LexicalText`, could have something to do with `HashtagNode` being an `instanceof TextNode`, leaving no break point when two hashtag nodes are adjacent, infinite parsing.

https://github.com/facebook/lexical/blob/67b159ee9275e4a8f67c53b9aad97d0d97cddb8b/packages/lexical-text/src/index.ts#L183-L187

Curiously, I wasn't able to reproduce this in the playground, while I don't see any specific logic handling this behavior, and that it uses the same Hashtag Regex.

Link to code example: https://codesandbox.io/s/lexical-plain-text-example-forked-4rspdy?file=/src/Editor.js

https://user-images.githubusercontent.com/9211198/204885829-5598d7b0-460f-4b31-8f5f-9bca9493f05f.mov

alisnic:
This is a workaround for the issue in case anyone else needs it:

```typescript
  useEffect(() => {
    return editor.registerCommand(
      PASTE_COMMAND,
      event => {
        if (event instanceof ClipboardEvent && event.clipboardData?.types.includes('text/plain')) {
          const pastedText = event.clipboardData.getData('text');

          if (pastedText[0] !== '#') {
            return false;
          }

          editor.update(() => {
            const selection = $getSelection();
            if (selection) {
              const nodes = selection.getNodes();
              const targetNode = nodes.length === 1 ? nodes[0] : nodes[0]?.getPreviousSibling();
              const finalText = targetNode?.getType() === 'hashtag' ? ` ${pastedText}` : pastedText;
              selection.insertText(finalText);
            }
          });

          return true;
        }

        return false;
      },
      COMMAND_PRIORITY_HIGH
    );
  }, [editor]);

```

PR #6045, opened 2024-05-07
Fix #5703 Infinite loop on hashtag transform

### LINKED ISSUE #5703: Bug: Typing an invalid or partially valid hashtag followed by a valid hashtag causes infinite transform execution
Hi!  Typing `#eggs#eggs #e` in a Lexical editor causes infinite transform execution, sometimes followed by the editor no longer responding properly to input.  Additionally, hashtags seem like they should either be valid when directly adjacent, or they should never be valid when adjacent, as in the first `#eggs` should not be a hashtag once the second `#` is appended.

This also happens in a plaintext editor.

Screenshot 2024-03-12 at 10 24 22 AM

Lexical version: 0.13.1

## Steps To Reproduce

1. Open the [Lexical playground](https://playground.lexical.dev/)
2. Type `#eggs#eggs #e`, observe the console error.  On mac at least, cmd+A no longer selects all text.  Backspace also does not remove text.

Link to code example:

## The current behavior

- The first `#eggs` instance is marked as a hashtag despite being invalid
- `#e` is not marked as a hashtag
- An infinite transform error is thrown
- The editor may or may not respond to select all via keyboard
- The editor may respond to backspaces by moving the beginning of the selection range backward instead of deleting characters, or it may do nothing

## The expected behavior

- The two instances of `#eggs` are both either valid or invalid.
- `#e` is marked as a hashtag
- An infinite transform error is not thrown
- The editor responds to key input properly

Sahejkm:
## Description
- while transforming the hashtag text entity, 
- if we are skipping a text entity we are not adding the skip length to the next hash tag entity found in the transform logic for the text entity resulting into wrong split and infinite loop
- Fix to add the previous match length to the next match start and end offset before splitting the text node for the next match 

Closes #5703 and #3468 as well

### Before

https://github.com/facebook/lexical/assets/163521239/3482f35c-5f71-4142-b746-79486aa21f3b

### After

https://github.com/facebook/lexical/assets/163521239/56130eb3-3cc9-4905-bc0e-6be4922bc630

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs
@@ -345,4 +345,42 @@ test.describe('Hashtags', () => {
       `,
     );
   });
+
+  test('Should not break while skipping invalid hashtags #5703', async ({
+    page,
+  }) => {
+    await focusEditor(page);
+    await page.keyboard.type('#hello');
+
+    await page.keyboard.press('Space');
+
+    await page.keyboard.type('#world');
+    await page.keyboard.type('#invalid');
+
+    await page.keyboard.press('Space');
+    await page.keyboard.type('#next');
+
+    await waitForSelector(page, '.PlaygroundEditorTheme__hashtag');
+
+    await assertHTML(
+      page,
+      html`
+        <p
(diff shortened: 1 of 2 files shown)


zurfyx (approved):

zurfyx:
Wow, I'm surprised this has been in there so long! Can we add a test for this please?

ivailop7 (approved):

Issue #6044, opened 2024-05-07
Feature: React 19 testing and documentation


etrepum:
Some of the behavior changes in React 19, particularly the ones relevant to StrictMode, can have observable effects when people upgrade (#6040 for example). After spending a little time looking into it, I think the library usage is correct enough, but we should try and write some tests that exercise the issues that StrictMode can cause.

Here are some things I think we should address for React 19:

* Add some jobs to CI that run tests against react@beta react-dom@beta (probably unit tests and one platform+browser set of e2e). I think the workspaces make this a little tricky but locally it seemed to work if I upgraded it in each workspace that has a direct dependency on react
```
$ grep version node_modules/react/package.json
  "version": "18.2.0",
$ for x in website devtools playground; do npm -w packages/lexical-$x i react@beta react-dom@beta; done
…
$ grep version node_modules/react/package.json
  "version": "19.0.0-beta-b498834eab-20240506",
```

* Scan the library for React usage that should change in the [React 19 Upgrade Guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide), particularly the [StrictMode improvements](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#strict-mode-improvements)… but also our tests would need workarounds because act moved with a deprecation warning so we'll need a wrapper for that.
* Write new tests that try and exercise potential issues with StrictMode, particularly around LexicalComposer and the function version of editorState, but also other situations where callbacks might be expected and not happen if certain things were initialized in the first pass of the double-render (e.g. editable or rootListener, update, etc.)
* Add FAQ entries and/or a usage guide to the lexical react docs that covers how to use (and not use) effects, useLexicalSubscription, etc.

Issue #5911, opened 2024-04-17
Docs: Add FAQs about hot module reloading 


etrepum:
Since Lexical relies on object identity quite a lot, it can run into trouble when developing apps using a hot module reloader. We should have some docs that cover which error messages are likely to happen in these scenarios, and also the sorts of workarounds you can use to ensure that the Lexical editor is reloaded when its dependencies change identity.

These docs could possibly also cover the scenario when two different versions of Lexical (or a plugin) are loaded into the same app, since some of the error messages may be similar.

acywatson:
One thing that has bitten me before with this is COMMAND identity. Sometimes, for some reason, the references used by the listeners and the ones defined in the core diverge. I think it has to do with HMR somehow, but not sure. it can be really confusing and hard to debug when it happens.

PR #6048, opened 2024-05-07
[lexical-react] Feature: React 19 unit tests

etrepum:
## Description

React 19 has several [behavior changes to StrictMode](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#strict-mode-improvements) that could easily be thought to be bugs in either React or Lexical but at this point it seems that all cases are "simply" user error. This PR takes the following approach to mitigate this potential issue:

- Use the React 17+ automatic JSX runtime (required for React 19, this was merged in #6088)
- Modernize Lexical code, tests, and doc examples to be (hopefully) correct and warning-free with React 18 and 19 (based on a thorough reading of the [React 19 Upgrade Guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide)) - some of this was cleaning up existing tests to be more correct, not necessarily React 19 related but important for validation. No codemods were run, the only large change was to the tests (adding an act shim), the codemod wouldn't have helped us in a React 18 compatible way.
- Change CI workflow to run a full set of the unit and e2e tests in CI with React 19
- New tests to validate our expectations of how React should interact with Lexical (both in and out of StrictMode)
- New documentation covering @lexical/react best practices, and common mistakes that can be made with useEffect that surface under StrictMode (which are usually real bugs in concurrent mode and/or suspense that are hard to reproduce otherwise)
- Workaround for vite config so that the playground builds with React 19

Other inclusions:
- Fix up some overlapping names in package.json files
- Fix all unit test warnings (including silencing `ArtificialNode__DO_NOT_USE` warning in `__DEV__` everywhere), audit `@ts-ignore` usage and remove unnecessary usage
- Fix a window mutation leak in one of the unit tests (updated `queueMicrotask` without restoration)
- Updates TypeScript because we were installing two versions (one for lexical-devtools and one elsewhere)
- Moves integration tests behind extended-tests label

NOTE:  @lexical/react does claim compatibility with react >=17.x in its peerDependencies but only ^18.2.0 and beta are tested as of this PR. It would probably be a better use of resources to raise the bar to >=18.x (or just leave it as-is until someone tells us it is broken) than to test further backwards, unless there are significant users that are stuck on 17.x for some reason.

Closes: #6044 
Closes: #6126
Closes: #6132
Closes: #5911

### Before

* The tests had some hard to debug warnings (due to microtask concurrency and in one case just a partially broken test) with any version of React
* We had no automated way of confirming that everything works with React 19 
* Users would commonly run into tricky StrictMode issues that typically required expert help to diagnose

### After

* All tests pass without warning in React 18 and 19
* CI confirm that everything works with React 18 and React 19 (technically not 19, but whichever react and react-dom versions are tagged beta at the time of CI)
* Users will probably have the same issues, but we'll have some documentation to point them to

## Follow-up Considerations

We could help mitigate more user issues by:

* More aggressively detecting common pitfalls with `__DEV__` warnings or invariants (e.g. write a hook that gets eliminated in prod that warns if a particular argument does not have a stable identity)
* Adding lint rules to @lexical/eslint. A class of lint rule that would be useful in this scenario would be detecting functions without stable identity (useCallback, useMemo, useRef, etc. or being defined out of component scope) that are passed as component props or hook arguments that would be problematic like CollaborationPlugin providerFactory or useLexicalSubscription


.github/workflows/test.yml
```diff
@@ -55,6 +56,12 @@ jobs:
           cache: npm
       - name: Install dependencies
         run: npm ci
+      - name: Override React ${{ matrix.react-override }}
+        if: matrix.react-override != ''
+        # This should be safe since we are caching ~/.npm and not node_modules
+        run: |
+          npm i --no-save react@${{ matrix.react-override }} react-dom@${{ matrix.react-override }}
+          grep version node_modules/{react,react-dom}/package.json
```
etrepum, line 64: I'm not entirely sure this is the best way but it does appear to work (the grep is left there so that it can be confirmed in the output, we could consider sending it to GITHUB_OUTPUT and display the actual version tested in a more visible place)

packages/lexical/src/__tests__/unit/LexicalNormalization.test.tsx
```diff
@@ -125,7 +125,7 @@ describe('LexicalNormalization tests', () => {
 
             const selection = paragraph.select();
             getAnchor(selection).set(text1.__key, 0, 'text');
-            getFocus(selection).set(text2.__key, 2, 'text');
+            getFocus(selection).set(text2.__key, 1, 'text');
```
etrepum, line 128: This fixes a bug in the test, the values here were wrong (text2 is only 1 character long, so an offset of 2 does not make sense and jest-dom correctly throws an error later, but that error gets turned into a warning as a firefox-mitigation)

etrepum:
This shouldn't be merged until after #6080 as I will need to revisit the workflow additions

StyleT:
Hi! Looks awesome. Can you pls rebase it and I'll take a look asap.

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -60,10 +60,6 @@ import {
   TestComposer,
   TestTextNode,
 } from '../utils';
-// No idea why we suddenly need to do this, but it fixes the tests
-// with latest experimental React version.
-// @ts-ignore
-global.IS_REACT_ACT_ENVIRONMENT = true;
```
etrepum, line 66: This was already set globally and I added a comment linking to the reason

etrepum:
@StyleT ready to go, you will need to add the `extended-tests` label if you want the full e2e suite to run

.github/workflows/call-core-tests.yml
```diff
@@ -30,6 +30,7 @@ jobs:
     strategy:
       matrix:
         node-version: [18.18.0]
+        react-override: ['', 'beta']
```
necolas, line 33: If this were called 'react-tag' or 'react-version' that might be cleared

.github/workflows/call-e2e-all-tests.yml
```diff
@@ -121,3 +121,18 @@ jobs:
       browser: ${{ matrix.browser }}
       editor-mode: ${{ matrix.editor-mode }}
       events-mode: ${{ matrix.events-mode }}
+
+  react-beta:
```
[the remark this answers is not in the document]
etrepum: It is adding a e2e but with a reduced matrix with 4 outputs, I'll move it to the extensive test suite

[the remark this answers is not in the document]
etrepum: Looking more closely it was already in the extensive tests. I will reduce the matrix from 4 to 1, picking the cheapest options of rich-text (not collab) and prod: false (not knowing which platform is cheapest) and if someone cares later we can expand it. I think this should increase the cost by <3% since we have ~40 e2e tests and many of them are going to be slower (e.g. firefox, collab).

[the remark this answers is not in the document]
necolas: > I'm not sure that cron is the ideal unless there is also a way to surface the results so that more people will see them

Workflow dispatches and crons are mostly for maintainers. The maintainers will get notified when a cron-based workflow fails. Typically this is how we periodically test against pre-releases of dependencies, because you don't usually want that to block a merge.

For example, the react-strict-dom integration test with `react@next` started failing a while ago, which surfaced the change in React's testing packages but doesn't block any incoming PRs with a test failure. You might get test failures against `beta` in the future too, but generally those failures are not something you'd want to block 
a PR merge.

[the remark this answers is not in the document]
etrepum: I'll leave it up to maintainers to decide if they want to move it to another place. From my experience in the past few months they seem quite busy and a fair amount of the contributions and fixes are coming from outside the team that would be notified. I don't believe that there's any rule that requires all checks to pass before merge with the current configuration, so I don't think that failures would necessarily mean that an incoming PR is blocked.

packages/lexical/src/LexicalEditor.ts
```diff
@@ -447,7 +447,7 @@ export function createEditor(editorConfig?: CreateEditorArgs): LexicalEditor {
       // Ensure custom nodes implement required methods.
       if (__DEV__) {
         const name = klass.name;
-        if (name !== 'RootNode') {
+        if (name !== 'RootNode' && name !== 'ArtificialNode__DO_NOT_USE') {
```
potatowagon, line 450: thanks, we need this fix quite urgently

potatowagon (dismissed):

thegreatercurve (approved):

packages/lexical/src/LexicalEditor.ts
etrepum: I don't think there's anything controversial in here, the right person just needs to look at it. It's mostly improvements to tests and documentation, although this one in particular also fixes normal runtime behavior.

etrepum:
The test failure look like a github fault

https://github.com/facebook/lexical/actions/runs/9106181327/job/25032956714?pr=6048

```
Download action repository 'actions/checkout@v4' (SHA:0ad4b8fadaa221de15dcec353f45205ec38ea70b)
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/0ad4b8fadaa221de15dcec353f45205ec38ea70b'. Error: nodename nor servname provided, or not known (api.github.com:443)
Warning: Back off 12.403 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/0ad4b8fadaa221de15dcec[35](https://github.com/facebook/lexical/actions/runs/9106181327/job/25032956714?pr=6048#step:1:39)3f45205ec38ea70b'. Error: nodename nor servname provided, or not known (api.github.com:443)
Warning: Back off 15.559 seconds before retry.
Error: nodename nor servname provided, or not known (api.github.com:443)
```

etrepum:
Closing and re-opening to kick off another test run

etrepum:
Doesn't look like that fired off another extended test run

etrepum:
Hard to tell which test hung but I'll take a look, I suspect restarting the test suite would pass though because they passed before.

necolas:
The react-beta tests are hanging at the moment. You might want to iterate on that test without running the extended test suite any more times. I'll remove the label for now

etrepum:
Removing the label wouldn't really help me confirm when it's fixed it unless I move it out of the extended test suite, and then it would get run for all builds if I didn't do another commit after approval (dismissing the approval) to move the workflow back.

etrepum:
I think something must've changed in the react beta tag that's throwing off rollup's heuristics which causes an issue with the playground build. Taking longer to resolve than I expected.

etrepum:
So far all I know is that there seems to be some sort of issue with the latest build of react beta and the heuristics that @rollup/plugin-commonjs uses to determine the named exports (another fun fork module day). I'll have to look into it more tomorrow.

```
RollupError: src/ui/FlashMessage.tsx (12:8): "createPortal" is not exported by "../../node_modules/react-dom/index.js", imported by "src/ui/FlashMessage.tsx".
file: /Users/bob/src/lexical/packages/lexical-playground/src/ui/FlashMessage.tsx:12:8
10: 
11: import {ReactNode} from 'react';
12: import {createPortal} from 'react-dom';
            ^
```

necolas:
Yes move it out of extended, get it passing, move it back into extended, get approval. This is also why it shouldn't be running on every PR IMO

etrepum:
I've disabled that one new e2e test suite (https://github.com/facebook/lexical/pull/6048/commits/774a2979da8969ec064d11f336ac9f35e68adfba), since it is not a regression and this PR includes other important fixes that should be released sooner than later. Every other workflow was passing before this commit, so up to the reviewer to decide to add extended-tests back before merging.

etrepum:
> Yes move it out of extended, get it passing, move it back into extended, get approval. This is also why it shouldn't be running on every PR IMO

In this particular case it was passing, until it wasn't. I think something changed upstream during that time. Anyway, it does not need to be resolved in this PR because it is not a regression, so it is disabled and all tests that run *should* be passing now.

etrepum:
Giving it another go since I have it all passing locally. Hopefully CI behaves the same way.

etrepum:
Test run is green, removing the workflow hack. https://github.com/facebook/lexical/actions/runs/9131368689/job/25110422301?pr=6048

packages/lexical-playground/vite.prod.config.ts
```diff
@@ -66,7 +66,11 @@ export default defineConfig({
     }),
     react(),
     viteCopyEsm(),
-    commonjs(),
+    commonjs({
+      // This is required for React 19 (at least 19.0.0-beta-26f2496093-20240514)
+      // because @rollup/plugin-commonjs does not analyze it correctly
+      strictRequires: [/\/node_modules\/(react-dom|react)\/[^/]\.js$/],
+    }),
```
etrepum, line 73: This took a maddening amount of time in the debugger to figure out what was going on and come up with a workaround that did not require a custom plugin!

Related:
* https://github.com/facebook/react/issues/10021
* https://github.com/facebook/react/issues/11503

etrepum:
Merged with main and waiting for review.

packages/lexical/src/__tests__/unit/LexicalUtils.test.ts
```diff
@@ -46,20 +46,22 @@ describe('LexicalUtils tests', () => {
 
     test('scheduleMicroTask(): promise', async () => {
       jest.resetModules();
-      // @ts-ignore
-      window.queueMicrotask = undefined;
```
etrepum, line 50: This was never unset, so I rewrote this test

etrepum:
Up to date with main, waiting for review.

etrepum:
Since the last time it was reviewed and ready to merge (but was blocked due to the jsx runtime regression & the react 19 beta rollup regression that has a workaround now), not much has changed:

* [Typescript upgrade](https://github.com/facebook/lexical/pull/6048/commits/75f5c9894dcab0214a7d779386a06bab2a642a53) - we had two versions installed because lexical-devtools was requesting a version that wasn't satisfied by other constraints
* [A better react override strategy](https://github.com/facebook/lexical/pull/6048/commits/90a98b1e4106e4cb14342906eb05d9544fceb4e3) - this also moves the integration tests to extended-tests since we should not need to run them constantly
* [A fix to the react beta e2e test](https://github.com/facebook/lexical/pull/6048/commits/5b810917485fbdf123c8bbc01b7b01a7948f228b) - something changed in React 19 beta from when I started to when we got this back to the approval stage last week and I had to change the rollup config accordingly
* [A more robust way to silence the artificial node warnings](https://github.com/facebook/lexical/pull/6048/commits/b21494ee77ca7511e16eadb542b928068fcdfe17)

Commits that are here that cancel out:

* [Adding](https://github.com/facebook/lexical/pull/6048/commits/d378edb5859bed246dc6d65bae890a7f20e8fe5e) then [removing](https://github.com/facebook/lexical/pull/6048/commits/a48687c64ada807ef015569e47105072e924ba10) the react beta e2e tests from the core suite to validate that everything still works (no net change)
* [A fix for the jsx build size regression](https://github.com/facebook/lexical/pull/6048/commits/492816e0e9512f59976965be93190bc61bb28d08) - this was cherry-picked and merged to main so its contents are not relevant anymore (no net change)
* Merge commits without any manual conflict resolution.

etrepum:
I would really like to see this included before the next release, it fixes all of the noise in the tests (and in `__DEV__`). It's annoying to work on the tests when they are spewing false positive to the console. Plus it makes everything React 19 ready, of course.

etrepum:
All tests are currently passing, the failure is the label-on-approval workflow issue being debugged in #6153

ivailop7:
Sure, I'm OK with it, but someone from Meta needs to confirm they are good with this one.

acywatson (approved):

PR #6049, opened 2024-05-07
[lexical-table][lexical-playground] Bug Fix: Table selection stuck 

AlexanderReznik:
## Description
Prevents table selection from being stuck.
The problem is wrong order of handling the mouse events. `mouseup` can happen before `mousedown` handler added a handler for  `mouseup`. This leaves the editor in a state of selecting. 
This PR adds clean up logic into `mousemove` to prevent selection being stuck

The modification of `dragMouse` is required for tests to pass, as it seems that this bug contributed to the successful execution of some test

### Before
table-selection-stuck

### DIFF
--- a/packages/lexical-playground/__tests__/utils/index.mjs
+++ b/packages/lexical-playground/__tests__/utils/index.mjs
@@ -612,6 +612,7 @@ export async function dragMouse(
   positionStart = 'middle',
   positionEnd = 'middle',
   mouseUp = true,
+  slow = false,
 ) {
   let fromX = fromBoundingBox.x;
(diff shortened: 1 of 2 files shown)


ivailop7:
@AlexanderReznik on manually testing, it does work in Safari on a Mac as well, but the issue seems to be indeed a test thing with WebKit, since it consistently fails on the same DraggableBlock, it's  possible it's indeed a correct failure. Maybe worth taking inspiration from the DraggableBlock tests to implement the ones here.

ivailop7 (approved):
LGTM

PR #6053, opened 2024-05-07
[lexical-playground][lexical-text] Bug Fix: handling hashtag following multiple invalid matches

2wheeh:
## Description

`prevMatchLengthToSkip` is introduced in #6045 to handle invalid previous match.
So I believe we should let `prevMatchLengthToSkip` do its job instead of early returning `$textNodeTransform` when the next match is followed.

### Before

https://github.com/facebook/lexical/assets/40269597/f5f4637d-4d29-4a23-a294-017f61b5e531

### After

https://github.com/facebook/lexical/assets/40269597/faa609ab-aa7e-4492-bbdf-d0b600c46e8a

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs
@@ -383,4 +383,61 @@ test.describe('Hashtags', () => {
       `,
     );
   });
+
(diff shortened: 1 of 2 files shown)


StyleT (approved):

Sahejkm (approved):
Awesome thanks for adding this so quick, i was also looking into fixing this test case next.

PR #6054, opened 2024-05-08
Fix #6051 Deprecate keycode to code

### LINKED ISSUE #6051: Feature: Switch from deprecated keyCode to key in internal onKeyDown handler
`keyCode` is deprecated (see https://w3c.github.io/uievents/#legacy-KeyboardEvent). As "keyCode" is officially deprecated, lexical should be updated to replace all references to keyCode with either `KeyboardEvent.code` or `KeyboardEvent.key` where applicable.

One of the functions using the deprecated keyCode: https://github.com/facebook/lexical/blob/ac97dfa9e14a73ea2d6934ff566282d7f758e8bb/packages/lexical/src/LexicalEvents.ts#L992

**What's the impact?** The popular `@testing-library/user-event` package no longer sets "keyCode" on their synthetic keyboard events.

Sahejkm:
## Description

Replace .keyCode with .code to follow new MDN API

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#browser_compatibility

**Closes:** #6051

### DIFF
--- a/packages/lexical-playground/src/plugins/TestRecorderPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TestRecorderPlugin/index.tsx
@@ -95,7 +95,7 @@ const formatStep = (step: Step) => {
 };
 
 export function isSelectAll(event: KeyboardEvent): boolean {
-  return event.keyCode === 65 && (IS_APPLE ? event.metaKey : event.ctrlKey);
+  return event.code === 'KeyA' && (IS_APPLE ? event.metaKey : event.ctrlKey);
 }
 
 // stolen from LexicalSelection-test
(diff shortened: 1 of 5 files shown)


zurfyx (commented):
I feel like this is correct but the size regression of including all the strings as opposed to numbers is noticeable (comparable to the discussion we had around coalescent operator). I also don't think the deprecated API is going away anytime soon, just like everyone uses `document.execCommand`. But it would also help toward the task assigned...

zurfyx (approved):

Daniel15:
This breaks non-QWERTY layouts (Workplace: https://fb.workplace.com/groups/lexicaldiscussion/posts/1950653992035669/) and the numpad enter key (https://fb.workplace.com/groups/lexicaldiscussion/posts/1950753535359048/), because the web platform assumes a QWERTY layout:

> For example, the code returned is "KeyQ" for the Q key on a QWERTY layout keyboard, but the same code value also represents the ' key on Dvorak keyboards and the A key on AZERTY keyboards. That makes it impossible to use the value of code to determine what the name of the key is to users if they're not using an anticipated keyboard layout.

Probably safer to use `key` instead of `code` (e.g. https://github.com/Daniel15/Website/blob/master/Daniel15.Web/wwwroot/Content/not_an_easter_egg/secret.ts#L20)

PR #6056, opened 2024-05-08
[lexical-text] Bug Fix: for handling multiple matches on hashtags

2wheeh:
## Description

~~#6053 deleted escape on adjacent next matches. but it turns out it was needed in different place (probably after replacement for current match has done ?)~~

~~without it, editor breaks on pasting or importing text that has adjacent matches.~~
 => #6067

And `nodeToReplace` can be `undefined` implicitly. Now it's the case. So I added null check for it. 
This could occur in other places especially around `TextNode.splitText()` unless we use `--noUncheckedIndexedAccess`.

### Before

https://github.com/facebook/lexical/assets/40269597/4f5b1756-a80a-428a-a919-b6075c754c9d

https://github.com/facebook/lexical/assets/40269597/362b301d-1793-4e4d-9903-cb634c08964e

https://github.com/facebook/lexical/assets/40269597/8763cd71-60f9-4d05-a490-050ec6cbdec1

### After

https://github.com/facebook/lexical/assets/40269597/5946d471-2a5f-4e47-9e40-96bc6ea66c9b

https://github.com/facebook/lexical/assets/40269597/66c68640-9f6b-4d38-aada-7ba9497ba0c5

### DIFF
--- a/packages/lexical-text/src/registerLexicalTextEntity.ts
+++ b/packages/lexical-text/src/registerLexicalTextEntity.ts
@@ -155,7 +155,7 @@ export function registerLexicalTextEntity<T extends TextNode>(
 
       invariant(
         nodeToReplace !== undefined,
-        'nodeToReplace should not be undefined.',
+        'nodeToReplace should not be undefined. You may want to check splitOffsets passed to the splitText.',
       );
 
       const replacementNode = createNode(nodeToReplace);
--- a/packages/lexical-text/src/registerLexicalTextEntity.ts
+++ b/packages/lexical-text/src/registerLexicalTextEntity.ts
@@ -155,7 +155,8 @@ export function registerLexicalTextEntity<T extends TextNode>(
 
       invariant(
         nodeToReplace !== undefined,
-        'nodeToReplace should not be undefined. You may want to check splitOffsets passed to the splitText.',
+        '%s should not be undefined. You may want to check splitOffsets passed to the splitText.',
+        'nodeToReplace',
       );
 
       const replacementNode = createNode(nodeToReplace);
--- a/packages/lexical-text/src/registerLexicalTextEntity.ts
+++ b/packages/lexical-text/src/registerLexicalTextEntity.ts
@@ -14,6 +14,7 @@ import {
   LexicalNode,
   TextNode,
 } from 'lexical';
+import invariant from 'shared/invariant';
 
 export type EntityMatch = {end: number; start: number};
 
@@ -152,9 +153,10 @@ export function registerLexicalTextEntity<T extends TextNode>(
         );
       }
 
-      if (nodeToReplace == null) {
-        return;
-      }
+      invariant(
+        nodeToReplace !== undefined,
+        'nodeToReplace should not be undefined.',
+      );
 
       const replacementNode = createNode(nodeToReplace);
       replacementNode.setFormat(nodeToReplace.getFormat());
--- a/packages/lexical-text/src/registerLexicalTextEntity.ts
+++ b/packages/lexical-text/src/registerLexicalTextEntity.ts
@@ -151,6 +151,11 @@ export function registerLexicalTextEntity<T extends TextNode>(
           match.end + prevMatchLengthToSkip,
         );
       }
+
+      if (nodeToReplace == null) {
+        return;
+      }
+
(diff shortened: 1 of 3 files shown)


2wheeh:
https://github.com/facebook/lexical/assets/40269597/56409695-7ef8-4a49-a205-4809bb98cd3c

Now it seems to work well.

cc @Sahejkm

packages/lexical-text/src/registerLexicalTextEntity.ts
```diff
@@ -128,12 +127,6 @@ export function registerLexicalTextEntity<T extends TextNode>(
             return;
           }
         }
-      } else {
-        const nextMatch = getMatch(nextText);
-
-        if (nextMatch !== null && nextMatch.start === 0) {
```
Sahejkm, line 134: - i think this early return condition is right in itself but it should only happen with "if match === null" check, 

- the real issue is after the 1st match is found for the example #hello#world, it doesn't update the prevSibling to replacementNode, added a fix here to address this to update prevSibling for next loop comparison: #6067

packages/lexical-text/src/registerLexicalTextEntity.ts
```diff
@@ -158,13 +151,23 @@ export function registerLexicalTextEntity<T extends TextNode>(
...
       nodeToReplace.replace(replacementNode);
 
       if (currentNode == null) {
         return;
       }
+
+      const nextMatch = getMatch(nextText);
+      if (nextMatch !== null && nextMatch.start === 0) {
+        return;
+      }
```
Sahejkm, line 170: - i think the real issue was previousSibling was not getting updated that's why it couldn't execute this condition above correctly and skip the loops correctly.
```
if (
        match.start === 0 &&
        $isTextNode(prevSibling) &&
        prevSibling.isTextEntity()
      ) {
        prevMatchLengthToSkip += match.end;
        continue;
      }
```

- updated here https://github.com/facebook/lexical/pull/6067/files, 
- these extra conditions won't be needed after we fix that check

2wheeh: Thank you for explanation! 
I reverted here. Failing tests added in the PR will be okay with #6067 merged.

packages/lexical-text/src/registerLexicalTextEntity.ts
```diff
@@ -158,6 +158,11 @@ export function registerLexicalTextEntity<T extends TextNode>(
           match.end + prevMatchLengthToSkip,
         );
       }
+
+      if (nodeToReplace == null) {
+        return;
+      }
+
```
Sahejkm, line 165: these lines are not needed as well now right ?

Sahejkm (commented):
Hi thanks @2wheeh, we could pull now for the tests to pass, then will approve and close this.

packages/lexical-text/src/registerLexicalTextEntity.ts
2wheeh: No you are right. I made these to invariant to prevent mistake in splitOffsets handling. is it okay?

Sahejkm: ok, could we add the invariant error code in codes.json similar to https://github.com/facebook/lexical/commit/d6cdef134f6f00b7b4b5e00e017a8922cf086961 ? Thanks!

Sahejkm: or i can add it in next PR and unblock this 1st

Sahejkm (approved):
LGTM, thanks for adding extensive tests for this!

packages/lexical-text/src/registerLexicalTextEntity.ts
2wheeh: oh i added on the last commit !

2wheeh: Test is not running though

Sahejkm (approved):

packages/lexical-text/src/registerLexicalTextEntity.ts
Sahejkm: this has happened 2-3 times recently i think, might need to look into the git workflows, we could run locally and merge, will monitor if it happen again and open an issue

PR #6061, opened 2024-05-08
Remove collapseAtStart

zurfyx:
Not used?


ivailop7:
Tests failing, but I like the direction of the PR :)

acywatson:
To borrow a phrase from @trueadm: "what're you on about?" lol

This is definitely used: 

https://github.com/facebook/lexical/blob/83fd6fa2bb48c5efa3c9c96e8c52dfa8faf4ae51/packages/lexical/src/LexicalSelection.ts#L1686

Issue #6059, opened 2024-05-08
Text formatting of token text nodes


lacroixdavid1:
The `RangeSelection.formatText` function does not format token text nodes between the range selection. Is there a reason why? I wish I could have text nodes that support rich-text formatting and still behave like text nodes. Any solutions?
 
Reference: https://github.com/facebook/lexical/blob/1154413a2c251146a4efc0658d58e74956193b95/packages/lexical/src/LexicalSelection.ts#L1168 

Thank you!

lacroixdavid1:
Closing as PR was merged.

kemsky:
This works:

```ts
    public createDOM(config: EditorConfig, editor?: LexicalEditor): HTMLElement {
        const dom = super.createDOM(config);

        dom.style.cssText = 'background-color: yellow';
        dom.classList.add((config.theme as PlaceholderNodeTheme).placeholder ?? 'placeholder');
        dom.spellcheck = false;

        return dom;
    }
```

This does not work:

```ts
    public createDOM(config: EditorConfig, editor?: LexicalEditor): HTMLElement {
        const dom = document.createElement('span');
        
        dom.classList.add((config.theme as PlaceholderNodeTheme).placeholder ?? 'placeholder');
        dom.spellcheck = false;
        dom.textContent = this.__placeholderName;
        
        return dom;
    }
```

PR #6062, opened 2024-05-08
[lexical-selection] Bug Fix  /  Fixes text formatting with segmented and token nodes #6059

lacroixdavid1:
## Description

It is currently only possible to format token nodes by selecting them entirely. They won't be formatted if they are in the middle of the range selection. Also, if the selection offset is in the middle of the node, it will split the node.

This PR ensures middle nodes are formated, just like when selecting them entirely. Also, it prevents breaking segmented or token nodes in the middle of the text where they shouldn't be allowed.

https://github.com/facebook/lexical/issues/6059

## Test plan

Added unit tests to the PR. See LexicalSelection.test.tsx.

### Before

- Formatting text in the middle of a segmented or token text node was splitting the node.
- Formatting range selection with a segmented or token node in the middle was not formatting the node.

### After

- Formatting text in the middle of a segmented or token text node is not splitting the node anymore. It is formatting the whole node.
- Formatting range selection with a segmented or token node in the middle now formats the node.

### DIFF
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
@@ -698,7 +698,6 @@ describe('LexicalSelection tests', () => {
         insertParagraph(),
         moveNativeSelection([1, 0, 0], 2, [1, 0, 0], 3),
         formatBold(),
-        // insertText('A')
       ],
       name: 'Format token node if it is the single one selected',
     },
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
@@ -623,6 +623,113 @@ describe('LexicalSelection tests', () => {
       ],
       name: 'Format selection that starts on element and ends on text and retain selection',
     },
+
+    {
+      expectedHTML:
+        '<div contenteditable="true" style="user-select: text; white-space: pre-wrap; word-break: break-word;" data-lexical-editor="true">' +
+        '<p class="editor-paragraph"><br></p>' +
+        '<p class="editor-paragraph" dir="ltr">' +
+        '<strong class="editor-text-bold" data-lexical-text="true">Hello</strong><strong class="editor-text-bold" data-lexical-text="true"> world</strong>' +
+        '</p>' +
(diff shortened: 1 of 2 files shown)


ivailop7:
This looks interesting, never stumbled upon this issue. For @Sahejkm @zurfyx and @potatowagon to comment,

acywatson:
> @lacroixdavid1 could you help explain this scenario via a video recording ? Thanks!

It's pretty easy to repro the issue on the playground by just inserting a mention, selecting it partially, and applying bold format.

acywatson:
I think these are actually probably the correct semantics for token/segmented nodes. The whole point of these is that they are a single "unit" and are interacted with as such from a creation/deletion standpoint.

The only other way I could think to deal with this would be to treat the styled portion of the node as and additional segment, rather than splitting the node and removing the token/segmented characteristics entirely.

acywatson (commented):
I'm good with this, but would like a second pair of eyes given that core selection is particularly prone to cause widespread issues.

lacroixdavid1:
> Formatting range selection with a segmented or token node in the middle was not formatting the node. @lacroixdavid1 could you help explain this scenario via a video recording ? Thanks!

**For the following example, the [MentionNode](https://github.com/lacroixdavid1/lexical/blob/main/packages/lexical-playground/src/nodes/MentionNode.ts#L122) was changed from `segmented` to `token`.**

### Before
before

### After
after

julienmouyal:
Any progress on this one? Seems like a great improvement. Thanks!

ivailop7 (approved):
LGTM, @zurfyx to sign off as well

lacroixdavid1:
Any chances of getting this one merged into 0.15.1? :pray: pls

zurfyx (approved):
Thank you, this makes a lot of sense! This actually makes me realize that we should recheck if there's any other codepath that can potentially mangle a segmented node like this. Segmented nodes should only segment in one way, separators, you shouldn't be able to split a Mention node arbritrary (at least not via user-input).

lacroixdavid1:
Thank you, guys. Lexical is very solid and has a bright future.

Patrascu-Lucian:
Thank you for this change!

It works very well, I found only one issue regarding:
`Formatting range selection with a segmented or token node in the middle now formats the node.`
This doesn't work when formatting text color

PR #6067, opened 2024-05-09
[Leixcal][Hashtag]  Fix prevSibling not getting updated after replacement of match as followup #6053

Sahejkm:
- followup on #6053 #6056 

https://github.com/facebook/lexical/assets/163521239/e308dbce-1a40-4b43-93eb-661b8cb5afce

### DIFF
--- a/packages/lexical-text/src/registerLexicalTextEntity.ts
+++ b/packages/lexical-text/src/registerLexicalTextEntity.ts
@@ -100,7 +100,6 @@ export function registerLexicalTextEntity<T extends TextNode>(
       }
     }
 
-    // eslint-disable-next-line no-constant-condition
     let prevMatchLengthToSkip = 0;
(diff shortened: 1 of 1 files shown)


zurfyx (dismissed):

packages/lexical-text/src/registerLexicalTextEntity.ts
```diff
@@ -165,6 +158,8 @@ export function registerLexicalTextEntity<T extends TextNode>(
       if (currentNode == null) {
         return;
       }
+      prevMatchLengthToSkip = 0;
```
Sahejkm, line 161: Missed a scenario to reset prevMatchLengthToSkip after match found for successive iterations


Uploading Screen Recording 2024-05-10 at 12.14.12 AM.mov…

zurfyx (approved):

PR #5979, opened 2024-04-27
[*] Chore: Fix all new lint warnings, plus manual exports clean-up

etrepum:
## Description

* Runs @lexical/eslint-plugin over the source to resolve all new warnings
* Manually fix up names that were shadowed
* Manually fix up invariants, flow types, and docs for renamed and deprecated public API
* Manually fix up internal usage of newly deprecated APIs

Newly deprecated APIs (exports still available):

* @lexical/link toggleLink - renamed to $toggleLink
* @lexical/offset createChildrenArray - renamed to $createChildrenArray
* @lexical/selection trimTextContentFromAnchor - renamed to $trimTextContentFromAnchor

## Test Plan

* Manually verify that the public API is stable in code review, either through no change or through a @deprecated alias
* Manually verify that the flow types for any renamed or added public APIs are changed accordingly
* All tests still pass

## Review notes:

This is a really big PR if you read through the automated stuff. It could be reviewed in pieces:

* [Configuration change](https://github.com/facebook/lexical/pull/5979/commits/bac16e6face61d40be785b3447fe5757d697a47f) (first commit) - updateEditor is basically editor.update and should be considered a lexical provider
* [Automated `npm run lint -- --fix`](https://github.com/facebook/lexical/pull/5979/commits/1498f7db134672dae265a6cb780cc39db8a05ed5) (second commit) - this is completely automated and probably does not need close review
* [Manual fix-ups](https://github.com/facebook/lexical/pull/5979/files/1498f7db134672dae265a6cb780cc39db8a05ed5..1c60e7bf04cd06e8d465c460f34105881e632e0d) (almost everything after the second commit) - worth a close look
* [Doc improvements](https://github.com/facebook/lexical/pull/5979/commits/d5b6ea8f01aa229b1d63ef20049da423f56ac641) (last relevant commit) - adds `{@link $newName}` around deprecations for better typedoc ux, and removes unused markdown partial plugin
* [Flaky collab test mitigation?](https://github.com/facebook/lexical/pull/5979/commits/58a94e6e8d59343d194bf48d794558a5c6b8ae01) (not so relevant commit) adds some stuff to maybe help address collab tests that seem flaky. It seemed to help a bit locally, hard to tell whether it it's an solid improvement in CI or not until a lot of test runs have happened.

PR #6069, opened 2024-05-09
Fix autolink styles on creation

Shubhankerism:
This PR fixes the issue with autolink nodes not having the same styles as the original text node.

### Before
autolink-style-before

### After
autolink-style-after

### DIFF
--- a/packages/lexical-react/src/LexicalAutoLinkPlugin.ts
+++ b/packages/lexical-react/src/LexicalAutoLinkPlugin.ts
@@ -179,7 +179,7 @@ function extractMatchingNodes(
   ];
 }
 
-function createAutoLinkNode(
+function $createAutoLinkNode_(
   nodes: TextNode[],
   startIndex: number,
   endIndex: number,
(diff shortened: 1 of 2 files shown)


packages/lexical-react/src/LexicalAutoLinkPlugin.ts
```diff
@@ -179,7 +179,7 @@ function extractMatchingNodes(
   ];
 }
 
-function createAutoLinkNode(
+function $createAutoLinkNode_(
```
ivailop7, line 182: @etrepum this was autogenerated from the eslint stuff, I don't think the _ is deliberate? Can we fix the rules?

ivailop7: nevermind, it's a name clash...

etrepum: This is why I did #5979 to look at these in one go, but nobody has reviewed it yet

PR #6071, opened 2024-05-09
[lexical-playground][TableCellResizer] Feature: Center the drag zone over the edge of the cell

apella:
## Description
Shift the selection rectangles to the center of the edge, instead of being outside the cell. This makes for a more intuitive user experience as the user will try to select the edge, not the outside of the table cell.

### Before
(briefly visualized the areas so that the change can be seen)

cell-resizer-old-regions

### After
cell-resizer-new


ivailop7:
@ebads67 is this the final version we settled on?

ivailop7 (approved):

ebads67:
> @ebads67 is this the final version we settled on?

Yes, this is just centering the cell resizer drag zone.

PR #6076, opened 2024-05-10
[lexical] Bug Fix: Mutation listeners not fired for replacement nodes

aleksandr-lapushkin:
## Description
- Currently, mutation listeners will not fire for replacement nodes registered with `withKlass` option.
- This PR modifies how mutation listeners are registered to respect that option. Mutation listeners registered for the base node will also fire for the replacement.

### Before
- Use playground project
- Create custom Table Node and register it
- Set up node replacement a-la `{replace: TableNode: withKlass: CustomTableNode, ... }`
- Try resizing the table columns
- No cursor appears

https://github.com/facebook/lexical/assets/12970405/f733f4b5-c14c-4bf0-baf2-21e89c6a8690

### After
- Use playground project
- Create custom Table Node and register it
- Set up node replacement a-la `{replace: TableNode: withKlass: CustomTableNode, ... }`
- Try resizing the table columns
- Columns can be resized

https://github.com/facebook/lexical/assets/12970405/db4cdee7-9df0-48b2-bdab-c8d69f331c72

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -204,7 +204,10 @@ export type Transform<T extends LexicalNode> = (node: T) => void;
 
 export type ErrorHandler = (error: Error) => void;
 
-export type MutationListeners = Map<MutationListener, Klass<LexicalNode>>;
+export type MutationListeners = Map<
+  MutationListener,
+  ReadonlyArray<Klass<LexicalNode>>
+>;
 
 export type MutatedNodes = Map<Klass<LexicalNode>, Map<NodeKey, NodeMutation>>;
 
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalEditor.ts
```diff
@@ -204,7 +204,10 @@ export type Transform<T extends LexicalNode> = (node: T) => void;
 
 export type ErrorHandler = (error: Error) => void;
 
-export type MutationListeners = Map<MutationListener, Klass<LexicalNode>>;
+export type MutationListeners = Map<
+  MutationListener,
+  ReadonlyArray<Klass<LexicalNode>>
```
aleksandr-lapushkin, line 209: I don't _quite_ like this. I was thinking of migrating this to nodes like with `transforms`, but this is a relatively inexpensive fix 🤷🏻

etrepum (commented):
This part of the code isn't something I have a lot of expertise with yet, but my concerns are:

* The docs and tests for the `withKlass` feature seem to be nearly nonexistent so it's hard to tell what exactly what the semantics are supposed to be
* What if there's a chain of node replacements? I think with a large enough ecosystem people will be plugins that extend plugins.
* Should there also be an `invariant(withKlass.prototype instanceof replace, '…')` or similar so that there's some confidence that they're substitutable in this kind of way? (when the editor is created not when listeners are registered of course). I guess it seems that way looking at #3639 when they tried to implement this in the types.

aleksandr-lapushkin:
Closing in favour of https://github.com/facebook/lexical/pull/6089

Issue #6068, opened 2024-05-09
Feature: picker active state on devtool UI 


2wheeh:
Currently, devtool  doesn't reflect picker state on `EditorInspectorButton`.

I think It would be better if it reflects pickerActive state on UI.

For example:

https://github.com/facebook/lexical/assets/40269597/cf114ea3-1cb0-4ee5-a0c9-4c9eaaba4405

H0onnn:
@StyleT Hello ! Can I solve this problem ?

2wheeh:
@H0onnn Hi I worked this on #6077 😿

StyleT:
Closed with https://github.com/facebook/lexical/pull/6077 will be released with the next extension update

PR #6077, opened 2024-05-10
[lexical-devtools] Feature: Reflect picker state on inspector button ui

2wheeh:
## Description

`EditorInspectorButton` reflects pickerActive state on UI.

**Closes:** #6068

### Before

https://github.com/facebook/lexical/assets/40269597/dec0b200-d955-4c04-8d3a-d04096f583bb

### After

https://github.com/facebook/lexical/assets/40269597/9ec3152e-75be-4722-9257-e6d5eee087f8

### DIFF
--- a/packages/lexical-devtools/src/store.ts
+++ b/packages/lexical-devtools/src/store.ts
@@ -25,7 +25,6 @@ export interface ExtensionState {
   isSelecting: {
     [tabID: number]: boolean;
   };
-  initTab: (tabID: number) => void;
   markTabAsRestricted: (tabID: number) => void;
   setStatesForTab: (
     id: number,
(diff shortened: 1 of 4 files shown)


StyleT (changes requested):
Hi! Thanks for this contribution!

Do you think it's possible to implements this via `ExtensionStore` instead of a bit quirky way when we try to keep sync via pending promise?

Also if done via store - we can sync that state to popup window as well easy peasy, while current approach makes it impossible

2wheeh:
@StyleT Thanks ! I reverted and changed to utilize store. Please check the last commit

packages/lexical-devtools/src/store.ts
```diff
@@ -28,10 +28,13 @@ export interface ExtensionState {
     states: {[editorKey: string]: SerializedRawEditorState},
   ) => void;
   setSelectedEditorKey: (tabID: number, editorKey: string | null) => void;
+  isSelecting: boolean;
```
StyleT, line 31: It should be per tab, as other values. As you may have 2 windows/tabs where you do selection "in parallel".

I have it in plans to make store work "per tab/window" by default, but I'm not there yet :P

StyleT (commented):
Looks much better now!

packages/lexical-devtools/src/store.ts
2wheeh: reverted the last commit because manipulating externalStore before reactDOM createRoot makes page restricted somehow and was also over the scope of this PR.

StyleT (approved):
Awesome work!

PR #6080, opened 2024-05-10
Only run CI e2e tests on approved PRs

necolas:
Avoid running the entire suite of (expensive) e2e tests on every PR that is opened or updated. Only run if PR is approved.

Reference: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved

### DIFF
--- a/.github/workflows/call-e2e-test.yml
+++ b/.github/workflows/call-e2e-test.yml
@@ -1,4 +1,4 @@
-name: Lexical e2e Tests
+name: Lexical e2e test runner
 
 on:
   workflow_call:
@@ -11,7 +11,7 @@ on:
       prod: {required: false, type: boolean}
 
 jobs:
-  e2e-tests:
+  e2e-test:
     runs-on: ${{ inputs.os }}
     if: (inputs.browser != 'webkit' || inputs.os == 'macos-latest') && (inputs.editor-mode != 'rich-text-with-collab' || inputs.events-mode != 'legacy-events')
     env:
--- a/.github/workflows/tests-extended.yml
+++ b/.github/workflows/tests-extended.yml
@@ -1,0 +1,16 @@
+name: Lexical Tests (Extended)
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
+    paths-ignore:
+      - 'packages/lexical-website/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e-tests:
+    if: github.repository_owner == 'facebook' && contains(github.event.pull_request.labels.*.name, 'extended-tests')
+    uses: ./.github/workflows/call-e2e-all-tests.yml
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -1,4 +1,4 @@
 module.exports = {
-  '*.(js|mjs|jsx|css|html|d.ts|ts|tsx)': 'prettier --write',
+  '*.(js|mjs|jsx|css|html|d.ts|ts|tsx|yml)': 'prettier --write',
   '*.(js|mjs|jsx|ts|tsx)': ['eslint --fix'],
 };
--- a/.github/workflows/call-core-tests.yml
+++ b/.github/workflows/call-core-tests.yml
@@ -1,0 +1,62 @@
+name: Lexical Core Tests
+
+on:
+  workflow_call:
+
+jobs:
+  integrity:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [20.11.0]
+    env:
+      CI: true
+      GITHUB_TOKEN: ${{ secrets.DANGER_GITHUB_API_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+      - name: Install dependencies
+        run: npm ci
+      - run: npm run ci-check
+      - run: npm run build
+      - run: npm run build-www
+
+  unit:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [18.18.0]
+    env:
+      CI: true
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+      - name: Install dependencies
+        run: npm ci
+      - run: npm run test-unit
+
+  integration:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [18.18.0]
+    env:
+      CI: true
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+      - name: Install dependencies
+        run: npm ci
+      - run: npm run test-integration
--- a/.github/workflows/call-e2e-all-tests.yml
+++ b/.github/workflows/call-e2e-all-tests.yml
@@ -1,90 +1,17 @@
-name: Lexical Tests
+name: Lexical e2e tests
 
 on:
-  push:
-    branches:
-      - main
-    paths-ignore:
-      - 'packages/lexical-website/**'
-  pull_request:
-    types: [opened, synchronize, reopened]
-    paths-ignore:
-      - 'packages/lexical-website/**'
-
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  workflow_call:
 
 jobs:
-  integrity:
-    if: github.repository_owner == 'facebook'
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [20.11.0]
-    env:
-      CI: true
-      GITHUB_TOKEN: ${{ secrets.DANGER_GITHUB_API_TOKEN }}
-    steps:
-      - uses: actions/checkout@v4
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ matrix.node-version }}
-          cache: npm
-      - name: Install dependencies
-        run: npm ci
-      - run: npm run ci-check
-      - run: npm run build
-      - run: npm run build-www
-
-  unit:
-    if: github.repository_owner == 'facebook'
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [18.18.0]
-    env:
-      CI: true
-    steps:
-      - uses: actions/checkout@v4
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ matrix.node-version }}
-          cache: npm
-      - name: Install dependencies
-        run: npm ci
-      - run: npm run test-unit
-
-  integration:
-    if: github.repository_owner == 'facebook'
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [18.18.0]
-    env:
-      CI: true
-    steps:
-      - uses: actions/checkout@v4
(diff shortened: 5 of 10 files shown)


etrepum:
It might be nicer to move the e2e tests to a separate workflow file, and maybe also do a single run of e2e tests in the test.yml workflow and a more comprehensive matrix on approval

necolas:
Yes, I floated some similar ideas as well. At the moment the e2e tests are being run every time a PR is updated, and every time a merge to main happens. This has put Lexical way over its CI budget. As a first step, I'd like to validate that a relatively small and quick change can demonstrate reduction in CI costs. If we can do that, I think further organization and optimization sounds great.

ivailop7 (approved):

etrepum:
I suggested the separate workflow file because it would be a smaller logical change since the approval stuff could be floated out to one place instead of N different places

etrepum:
but maybe my understanding of the schema is a bit off, I haven't tried to do it. looks good either way!

necolas:
Haven't seen this before; cool feature.

image

It looks like some open PRs haven't run the e2e tests. Hard to tell why some open PR have or haven't run all tests.

acywatson:
I agree with this in spirit, but in practice it's gonna slow us down significantly to have to get a stamp before we can see the CI run outcome.

Is there a way to make an exception for codeowners or specific users? (i.e., something like: if it's a maintainer, run on submit, otherwise, run on approve.)

acywatson:
> It looks like some open PRs haven't run the e2e tests. Hard to tell why some open PR have or haven't run all tests.

I believe this is because of the budget issue

etrepum:
I think what would be budget and velocity friendly would be to run a very limited matrix of e2e before approval (but not in draft), e.g. mac chromium or whatever is cheapest, and then do the full suite later on approval or some other condition that doesn't happen on every commit for every contributor. If there was some parameters around how much these things cost and what the budget are the scope could be widened a bit to cover more combinations (e.g. mac+chromium and mac+chromium+collab). We could even consider ditching some things like legacy events altogether unless that's still important enough to worry about.

Breaking the build merely by contributing too much isn't something I anticipated 😆

necolas:
> Is there a way to make an exception for codeowners or specific users?

There is a way to do that. I looked through recent commits and PRs, and it looks like most of them are created by Meta employees or Lexical collaborators - it would limit the budget impact to exclude maintainers.

I think what @etrepum suggested could be a good balance, e.g., you decide one or two e2e suites to run always, and the rest are deferred until PR approval. Later we could try to exclude more PRs (like this one) from running e2e tests at all; that would help allocate more budget to the PRs that benefit most.

acywatson:
> I think what @etrepum suggested could be a good balance

Yea, this is a good idea.

necolas:
Once the existing CI workflow is back up and running, I'm happy to keep iterating on feedback until you're happy it doesn't negatively impact development.

.github/workflows/test.yml
```diff
@@ -109,7 +109,7 @@ jobs:
       events-mode: ${{ matrix.events-mode }}
 
   e2e-windows:
-    if: github.repository_owner == 'facebook'
+    if: github.repository_owner == 'facebook' && (github.event.review.state == 'APPROVED' || contains(github.event.repository.collaborators, github.actor))
```
necolas, line 112: Apparently this is how to check if the PR author is a collaborator

necolas: I picked windows here because:
1) Assumed it's the most used OS by end-users
2) Assumed most contributors are developing on mac and would have less visibility into windows regressions
3) Has the smallest browser Matrix (no Safari)

Let me know if you'd like the opt-in test(s) to be changed.

etrepum:
looks like the ci-check failure is prettier, it is picky about the formatting of those yml files

.github/workflows/test.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       - run: npm run test-integration
 
   e2e-mac:
-    if: github.repository_owner == 'facebook'
+    if: github.repository_owner == 'facebook' && github.event.review.state == 'APPROVED'
```
necolas, line 80: Everything except `e2e-windows` will run only after a PR is approved.

.lintstagedrc.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
-  '*.(js|mjs|jsx|css|html|d.ts|ts|tsx)': 'prettier --write',
+  '*.(js|mjs|jsx|css|html|d.ts|ts|tsx|yml)': 'prettier --write',
```
necolas, line 2: Auto-format `yml` files on commit

necolas:
This diff successfully ran only the windows e2e matrix.

.github/workflows/test.yml
necolas: e2e-windows is still 8 jobs out of 37 total e2e jobs (if I counted right). By my maths that's still not enough to keep you comfortably under budget over time. I'll get that down to 1 job

acywatson:
> This diff successfully ran only the windows e2e matrix.

I think this PR is correct, but I'm a little confused that you're saying that the windows e2e matrix is running. It looks to me like the macOS matrix is running via e2e-canary.

My understanding of the intent is that we run this smaller batch of tests againt macOS/chromium commit, then run everything else on approval. Is that right?

acywatson (approved):

acywatson:
> I think this PR is correct

I may have spoken too soon - if we want the rest of the test suite to run upon approval, we probably need to change this line to trigger the test.yml workflow on approval:

 https://github.com/facebook/lexical/blob/83fd6fa2bb48c5efa3c9c96e8c52dfa8faf4ae51/.github/workflows/test.yml#L10

acywatson (changes requested):
See comments re: triggers

etrepum:
> I may have spoken too soon - if we want the rest of the test suite to run upon approval, we probably need to change this line to trigger the test.yml workflow on approval:
> 
> https://github.com/facebook/lexical/blob/83fd6fa2bb48c5efa3c9c96e8c52dfa8faf4ae51/.github/workflows/test.yml#L10
> 
> Probably just need to add "submitted"
> 
> https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved

I think this also means that the cheaper tests and canary will run again when the button is clicked. IMO a separate workflow file to contain the tests we don't want to run unconditionally really does make sense if we're going to keep iterating on this before merge

acywatson:
> I think this also means that the cheaper tests and canary will run again when the button is clicked.

Yea, this also occurred to me. I think it's an improvement either way, but your suggestion does seem optimal here (i.e., just run the stuff we didn't run and don't re-run stuff that already passed).

necolas:
[As I mentioned](https://github.com/facebook/lexical/pull/6080/files/5b67cdff9c444057cb82a24c3a8cda3ef6b124be#r1597312608), I switched to only run a single e2e test by default to keep costs down. All of them except the canary should run on approval - if you approve the PR we can check that. Can't tell if that worked because you requested changes.

I'm happy to iterate on how this is organized and optimized once we get the CI costs down from 10x where they should be. Some of the suggestions from reviewers probably aren't right because commits after approval still need tests to run. Let's try to land a first pass that works in a general sense and go from there

acywatson:
> if you approve the PR we can check that. Can't tell if it works because you requested changes.

I did and they didn't. I requested changes to dismiss my approval so I can test it again once it's fixed.

Screenshot 2024-05-11 at 4 54 03 PM

> I'm happy to iterate on how this is organized and optimized once we get the CI costs down from 10x where they should be.

Sure, if you can fix the fact that all the tests don't run on approval, then I'm happy to merge this and iterate whenever you want. Right now, it's not mergeable because the tests don't run on approval as they should - I provided [guidance](https://github.com/facebook/lexical/pull/6080#issuecomment-2105956306) on how to make that happen.

etrepum:
With regard to my previous comment, I am saying that I believe that once this is fixed, if it's done in the same workflow file, the tests will probably run again after the approval button is clicked because it's a new *event* for the same commit and the workflow file will catch all of them and run through all of the jobs. My understanding is that it will run everything it did before, plus the things it did not run before. In that scenario we are running the 10% (before approval) + 100% (after approval) = 110% of the tests, instead of the ideal 10% (before approval) + 90% (after approval) = 100% of the tests.

I think it would be fine to get rid of the contributor conditions entirely and have a set of more expensive e2e tests in one workflow file that run only on approval (and probably not bother with that expensive workflow on push to main altogether), and the cheaper canary/unit run on push and PR.

necolas:
I separated the core tests and e2e tests into additional workflow calls. In this iteration, the core tests and canary e2e test will always run on PRs and pushes to `main`. If you'd prefer a different workflow for `main` pushes, let me know. The extended e2e test suite is only run if a specific label (pick a name) is added to the PR by someone with triage access. 

Main / PR
* core-tests
  * unit, integrity, integration
* e2e-tests
  * canary

Label
 * e2e-tests
  * all
 
It was hard to test using *only* reviewer-approval as a trigger - you can't approve your own PRs. This diff also highlighted that it's a bit silly that reviewers have to approve a PR just to find out if it's mergable. Using the approach suggested by GitHub might also have some [issues cancelling workflows](https://github.com/orgs/community/discussions/25372#discussioncomment-6363007) if other reviews occur after approval.

Using a label as a trigger is a bit more flexible I think. Adding the label could still be automated too, with another workflow.

necolas:
I mocked the e2e test workflow to make iterating / reviewing faster and cheaper.

PR skips extended tests by default:

image

Added "[approved](https://github.com/facebook/lexical/labels/approved)" label (name TBD) and extended tests run:

image

Removed label and updated diff, extended tests don't run:

image

.github/workflows/tests-extended.yml
```diff
@@ -0,0 +1,16 @@
...
+    paths-ignore:
+      - 'packages/lexical-website/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e-tests:
+    if: github.repository_owner == 'facebook' && contains(github.event.pull_request.labels.*.name, 'approved')
```
etrepum, line 15: To bike-shed slightly on the name I would recommend changing the allow label to something like 'extended-tests' or 'e2e-tests' instead of 'approved'. My reasoning behind this is:

* We may also want to have a deny label like 'no-extended-tests' or 'no-e2e-tests' for PRs that a reviewer knows won't benefit from the full suite
* A contributor with sufficient privileges working on something that they know would benefit from the full suite could self-label without looking like they are rubber-stamping their own PR

.github/workflows/call-e2e-test.yml
```diff
@@ -0,0 +1,72 @@
...
+  e2e-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+      - name: Mock e2e test
```
etrepum, line 23: We of course need to un-mock before merge but great idea to make this quick, cheap and easy to verify!

etrepum (commented):
This looks like a great approach to me, I think it makes sense (as a follow-up) to add a workflow that automatically adds the label so reviewers don't have to remember to do a second action after approving a PR

.github/workflows/tests-extended.yml
necolas: Thanks. Renamed to 'extended-tests'.

An alternative / addition to a "deny label" would be to expand on the `paths-ignore` to help automatically exclude PRs that only change files like READMEs, etc., which don't need functional testing.

I was also looking at the [patch Dominic introduced](https://github.com/facebook/lexical/pull/750) to run e2e tests with beforeinput disabled - those of you more familiar with the status of "legacy events" will be better placed to determine if they still need as much testing.

acywatson (dismissed):
> I think it makes sense (as a follow-up) to add a workflow that automatically adds the label so reviewers don't have to remember to do a second action after approving a PR

Makes sense. This lgtm now.

Thanks for working on this, @necolas

acywatson:
We're waiting on a commit to unmock before we merge, right?

necolas:
> We're waiting on a commit to unmock before we merge, right?

Mock removed.

> Thanks for working on this, @necolas

No worries. The Lexical OSS project has some cool features I learnt about too. Very impressed with what you've all done here

Sahejkm (approved):
LGTM, thanks @necolas for optimising the workflows for budget control.

StyleT (approved):

Issue #6082, opened 2024-05-10
Feature: paragraphNode should apply configured indent value from theme on exportDOM


2wheeh:
### from exportDOM

![image](https://github.com/facebook/lexical/assets/40269597/bca898ac-a611-490b-9dc8-64addb47575a)

### from reconciler

![image](https://github.com/facebook/lexical/assets/40269597/b17b54a8-0c44-4683-82c3-946fddc3ae77)

element returned from `exportDOM` of `paragraphNode` doesn't have indent class from theme and doesn't apply configured indentation base value as well.

instead, it always has 20px as indentation value which makes difference between generated html and contenteditable.

Should this exported element have indent class from theme and also same indent value as configured in indent class ?

2wheeh:
What i understand:

1. When importing, 20px seems to be set considering to pasting HTML with a small base indent size. https://github.com/facebook/lexical/blob/157b77c2d84fb47aa85f98171e4285b284c26abc/packages/lexical/src/nodes/LexicalParagraphNode.ts#L197
2. For consistency between exporting and importing, we are setting 20px `text-indent` on exporting as well. https://github.com/facebook/lexical/blob/157b77c2d84fb47aa85f98171e4285b284c26abc/packages/lexical/src/nodes/LexicalParagraphNode.ts#L126
3. However, the default indent size when creating DOM is 40px. https://github.com/facebook/lexical/blob/157b77c2d84fb47aa85f98171e4285b284c26abc/packages/lexical/src/LexicalReconciler.ts#L122
4. Pasting HTML copied from the same editor (same namespace) doesn't trigger `importDOM`, which is utilized for HTML strings copied from outside sources. https://github.com/facebook/lexical/blob/157b77c2d84fb47aa85f98171e4285b284c26abc/packages/lexical-clipboard/src/clipboard.ts#L134
5. Thus, the situation here: ensuring (A) consistency between exporting and importing for outside the editor vs (B) consistency between creating and exporting from within the editor.

Personally, I lean towards (B) since it ensures the same result for the same editor state source but (A) is for outside sources.

2wheeh:
For someone who needs this, you can utilize HTML config as a workaround.

for example:

```ts
const htmlConfig: HTMLConfig = {
  export: new Map([
    [
      ParagraphNode,
      (editor, node) => {
        const targetNode = node as ParagraphNode;
        const element = document.createElement('p');

        if (element && isHTMLElement(element)) {
[...]

PR #6083, opened 2024-05-11
[WIP] Feature: Apply indent value from theme on exporting paragraphNode

2wheeh:
## Description

Now, `exportDOM` of `paragraphNode` applies indent value that is configured from theme and also has indent class.
So we get more similar html generated to dom created by reconciler.

**Closes:** #6082

### Before

### After


2wheeh:
It seems to be set considering to pasting html which has 20px of base indent size...
It's not as simple as i expected.
Closing this for now.

PR #6087, opened 2024-05-11
[lexical][TextNode] Feature: improve supporting multiple format on importDOM of TextNode

2wheeh:
## Description

Currently, when importing textNode from outside sources (or read-only mode), `convertTextFormatElement` function is utilized for `i`, `em`, `strong`, ... 

However, this doesn't reflect their textFormat by style attributes unlike the handling of `span` in `convertSpanElement`.

So this PR extracts some logic from `convertSpanElement` and applies it to `convertTextFormatElement`.

### Before

https://github.com/facebook/lexical/assets/40269597/e389e072-a5f1-4731-82b2-1dd9ba9f60dd

### After

https://github.com/facebook/lexical/assets/40269597/4ab55ca5-75a9-4ad0-8230-f5994b2af59c

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -1313,22 +1313,25 @@ function applyTextFormatFromStyle(
     if (!$isTextNode(lexicalNode)) {
       return lexicalNode;
     }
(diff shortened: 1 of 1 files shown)


acywatson (approved):
This makes sense. Thanks

Issue #6079, opened 2024-05-10
Feature: Deprecate default exports from public entrypoints (or at least give every module named exports)


etrepum:
The following modules use default exports in their entrypoints, which is inconsistent with the rest of the monorepo:

* packages/lexical-react/src/LexicalClickableLinkPlugin.tsx
* packages/lexical-react/src/LexicalErrorBoundary.tsx
* packages/lexical-react/src/LexicalTableOfContents.tsx
* packages/lexical-react/src/useLexicalEditable.ts
* packages/lexical-react/src/useLexicalSubscription.ts

There are other reasons not to use default exports (e.g. the types are wrong in cjs, the module names are not always consistent with what we call them in OSS docs for www reasons, etc.) but I think inconsistency is a good enough reason.

I propose that we, at minimum:

- Add named exports in all of these cases

Bonus points if we can agree to:

- Mark default exports as `@deprecated`
- Add a lint rule to warn when new default exports are added to public entrypoints (at least for packages/lexical-react/ where it has happened several times and has the most impact, everything else is a barrel file or not really API code like lexical-eslint)

StyleT:
@etrepum Hi! Can we close this?

etrepum:
@StyleT yes, it did not get closed because of a bug in the issue template that's fixed by #6124

PR #6088, opened 2024-05-11
[lexical-react] Breaking change: Deprecate public default exports

etrepum:
## Description

There are a handful of reasons to avoid default exports described briefly in #6079. A compelling reason not to use them is in our case is that they are rarely and inconsistently used in monorepo entrypoints, but there are other more technical reasons that they are not ideal (they generate confusing API docs for example)

* Adds named exports for all public entrypoints (only the @lexical/react package had modules with default exports)
* Marks default exports with `@deprecated` comments
* Updates monorepo code and docs to use named exports
* Adds lint rule to prevent future inconsistencies
* Uses the [automatic jsx runtime](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) so we don't have errors if an `import * as React from 'react'` is forgotten. I wonder why this isn't the default preset yet, given that it's been available for a few years.
* Switches rollup to [named exports](https://rollupjs.org/configuration-options/#output-exports)
* Removes a console.log from a unit test 
* Fixes scripts/update-flowconfig to inject the `@flow strict` into the boilerplate template header correctly

This is isolated to the `@lexical/rect` package and affects the following modules:

* `@lexical/react/LexicalClickableLinkPlugin` (now exports name `ClickableLinkPlugin`)
* `@lexical/react/LexicalErrorBoundary` `(now exports name `LexicalErrorBoundary`)
* `@lexical/react/LexicalTableOfContents` -> `@lexical/react/LexicalTableOfContentsPlugin`(which exports name `TableOfContentsPlugin`) - the old module still exists with only a default export
* `@lexical/react/useLexicalEditable` (now exports name `useLexicalEditable`)
* `@lexical/react/useLexicalSubscription` (now exports name `useLexicalSubscription`)

**Closes:** #6079

## Test plan

* All test suites should still pass
* Reviewer(s) should manually verify that default exports are still available in all of the entrypoints that were changed
* Someone with knowledge of www should verify that this will not cause problems, I believe I updated the flow types correctly although one case seemed wrong to begin with (will note in review comment) but the change to rollup's outputOptions [export: 'named'](https://rollupjs.org/configuration-options/#output-exports) could be problematic depending on what haste (or whatever that's called these days) does with exports named default. I don't think it's possible to make rollup output something like `module.exports = function defaultExport() {}; module.exports.defaultExport = module.exports;` which is pretty much the only thing that would have the same cjs semantics. It could also of course be fixed with a codemod fairly easily as there are only five renames here.


packages/lexical-react/flow/LexicalClickableLinkPlugin.js.flow
```diff
@@ -7,6 +7,8 @@
  * @flow strict
  */
 
-declare export function MLCClickableLinkPlugin({
+declare export function ClickableLinkPlugin({
```
etrepum, line 10: This seemed wrong, since no such name was present. Not sure what was going on here?

packages/lexical-website/docs/react/plugins.md
```diff
@@ -169,18 +169,20 @@ Adds markdown shortcut support: headings, lists, code blocks, quotes, links and
 <MarkdownShortcutPlugin />
 ```
 
-### `TableOfContentsPlugin`
-This plugin allows you to navigate to certain sections of the page by clicking on headings that exist inside these sections. Once you load the plugin, it automatically collects and injects the headings of the page inside the table of contents, then it listens to any deletions or modifications to those headings and updates the table of contents. Additionally, it's able to track any newly added headings and inserts them in the table of contents once they are created. This plugin also supports lazy loading - so you can defer adding the plugin until when the user needs it.
-```jsx
-<TableOfContentsPlugin />
-```
-You can alternatively leverage the use of `LexicalTableOfContents` API, which provides you with all the functionality that `TableOfContentsPlugin` provides, but without any styling.
+### `LexicalTableOfContents`
```
etrepum, line 172: I changed these docs because I believe they describe the playground `TableOfContentsPlugin`. I don't love that `LexicalTableOfContents` has an inconsistent name with most other plugins but I wanted to focus this mostly on just swapping the default with the name that they are documented to have.

StyleT (commented):
This change (and the lint rule) really really good in a sense that it allows to unify approach to exports. 

However breaking changes here seems to be unnecessary... This will be especially painful for larger codebases. 

Better approach would be to do both named + default export while marking default as deprecated, announce change, let people migrate at their pace, deprecate defaults.

But let's hear other opinions too!

etrepum:
This PR does that, default exports are still there, but it's breaking in the sense that working with the newer docs/examples won't be compatible with older lexical

etrepum:
Also technically in cjs this is a breaking change if the bundler doesn't do the right thing with exports named default with these rollup settings

packages/lexical-react/flow/LexicalClickableLinkPlugin.js.flow
StyleT: Let's keep it as is and mark deprecated

StyleT (approved):
LGTM with 1 comment!

StyleT:
> This PR does that, default exports are still there, but it's breaking in the sense that working with the newer docs/examples won't be compatible with older lexical

yep, I see now, by bad. Still asleep here :P

etrepum:
@StyleT we could keep that flow type as-is, but it references a symbol that didn't exist in the first place? Would probably be better to remove it if using a name that exists isn't the proper fix

StyleT:
> @StyleT we could keep that flow type as-is, but it references a symbol that didn't exist in the first place? Would probably be better to remove it if using a name that exists isn't the proper fix

yep, you're right. It's just a flow file. Good to be merged. Let's allow others at least a day to look at it

Issue #4498, opened 2023-05-15
Bug: registerMutationListener on replaced node


bilobom:
I have a CustomTableNode that replaces the internal TableNode, the TableSelection isn't working on this CustomTableNode. I had to rewrite the mutation listener to listen for CustomTableNode mutation . Is there a way to support mutation listeners for node override ?

link to reproduce : https://codesandbox.io/s/infallible-mendeleev-lsn5fz?file=/src/Editor.js

here is my custom table node:
```ts
import { TableNode } from '@lexical/table'

import {
  DOMConversionMap,
  DOMExportOutput,
  EditorConfig,
  LexicalEditor,
  NodeKey,
  SerializedElementNode
} from 'lexical'

export type SerializedTableNode = SerializedElementNode & {
  type: 'custom-table'
  version: 1
}

export class CustomTableNode extends TableNode {
  static getType(): string {
    return 'custom-table'
  }
  constructor(key?: NodeKey) {
    super(key)
  }
  static clone(node: TableNode): CustomTableNode {
    return new CustomTableNode(node.__key)
  }
  createDOM(config: EditorConfig, editor?: LexicalEditor): HTMLElement {
    return super.createDOM(config, editor)
  }
  static importDOM(): DOMConversionMap | null {
    return super.importDOM()
  }
  static importJSON(_serializedNode: SerializedTableNode): CustomTableNode {
    return new CustomTableNode()
  }
  exportJSON(): SerializedTableNode {
    return {
      ...super.exportJSON(),
      type: 'custom-table',
      version: 1
    }
  }
  exportDOM(editor: LexicalEditor): DOMExportOutput {
    const { element, after } = super.exportDOM(editor)
    return {
      element,
      after(generatedElement) {
        const tableElement = after?.(generatedElement) || ''
        const container = document.createElement('div')
        container.classList.add('editor__table-container')
        container.append(tableElement)
        return container
      }
    }
  }
}

```
I had to rewrite the same logic that attaches TableSelection for CustomTableNode so that selection works.
```js
registerMutationListener(CustomTableNode, (nodeMutations) => {
  // initializeTableNode and applyHandlers for selection
}
```
Lexical version:1.10.0

## Steps To Reproduce

1. create a custom node and replace it with node that has a mutation listener builtin . 
2. this listeners won't take effect.

## The current behavior
registerMutationObserver for original node won't work for the replaced node
## The expected behavior
registerMutationObserver for original node should work for the replaced node

PR #6089, opened 2024-05-11
fix mutate with replace klass

Piechota:
When registering mutation check for Node that replaced original one.
_mutation_ are a map so only one entry.

Should fix  [#4498](https://github.com/facebook/lexical/issues/4498#issue-1709961845)


etrepum:
This looks like it solves the same problem as #6076 so it should close that as well if this PR is chosen. My review comments would be basically the same: https://github.com/facebook/lexical/pull/6076#pullrequestreview-2051214716

etrepum:
I think this approach is probably better, but without really any docs or test coverage it's hard to say if there are any additional caveats to this simpler solution

Piechota:
In that other PR you have a good point about chain of node replacements.
Currently this code doesn't support it, I will fix it when I have a second.

Piechota:
So I have added loop to support multilevel replacement.
But I have problem testing it, it seems like right now it isn't supported really well. 
I think we should also loop over _replace_ in $applyNodeReplacement()
But that's for different change.

etrepum:
Good idea, it seems that there is a lack of test coverage all around for this corner of the API!

StyleT:
If no-one picks this up - I'll look at it later to add some tests. I touched klasses already in the devtools

aleksandr-lapushkin:
> If no-one picks this up - I'll look at it later to add some tests. I touched klasses already in the devtools

That would be very welcome! I'll close https://github.com/facebook/lexical/pull/6076 since this one seems more in line with expectations.

StyleT:
Closed in favor of https://github.com/facebook/lexical/pull/6189 (tests added)

PR #6091, opened 2024-05-12
[lexical][TextNode] Feature: Add test for #6087

2wheeh:
## Description

This PR adds test for #6087

And for the further improvement, It would be nicer if importDOM supports font size or colors. So I added todo about this.

## Test plan

recording of the manual reproduction of added test:

https://github.com/facebook/lexical/assets/40269597/1cd61730-892b-4590-bb75-42e61c0c4829


zurfyx (approved):
Thank you for the test! Colors might have to be implemented at a product level for now since they're not backward compatible (you can override HTML or Node)

Issue #6097, opened 2024-05-13
Feature: Add extended-tests label on approval


etrepum:
When a PR is approved, if the label no-extended-tests is not present, add the extended-tests label.

Could be done with an workflow that uses the [gh CLI](https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues) directly or call an action such as [labeler](https://github.com/actions/labeler) and use a condition to run it only when the PR is approved (and maybe if it touches any of public packages source file, the playground, or a package/package-lock file)

Follow-up to #6080

StyleT:
I think it may better to utilize GH merge queues, in addition to manual triggers/labels:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue

It’s also discussed here: https://github.com/orgs/community/discussions/53922

etrepum:
Given that I don't have the authorization to use labels or approvals, I don't have a strong opinion either way. I just thought I should add the follow-up issue based on what was discussed on the workflow PR so it doesn't get forgotten, since at the moment there is an extra step in reviews. As long as it's easy and efficient for the people who are doing the reviews (and not error-prone), I will be happy with it.

We should probably also add a few words to the maintainers guide with how the review process works, to communicate the process change to existing maintainers and to make onboarding a new maintainer more efficient.

StyleT:
Thanks for filing this anyway! We shall solve it either way

necolas:
A merge queue is what I first suggested internally too. But didn't realize GitHub now has a built-in version of it!

Sahejkm:
following up with OSS support to enable Merge Queue options in for internal meta settings as well, its not visible right now.

etrepum:
It seems that this didn't work in #6129 when @ivailop7 approved it: https://github.com/facebook/lexical/actions/runs/9133904177/job/25118355291?pr=6129

Sahejkm:
I think it's the same issue as the size limit job, it doesn't run from forks, and need to use the pull request target event 
:(

PR #6100, opened 2024-05-14
[Lexical] #6097 run extended tests on merge_group event

Sahejkm:
- Follow up on #6097  run extended tests on merge group event
- This is to ensure :
   - No branch is merged without running e2e tests successfully
   - e2e tests are run every time on merge not just if a label is added manually by facebook owner.
- will need the Merge Queue settings to be enabled 1st before we we could test this


StyleT:
Wow! 🤩

thegreatercurve (approved):

Issue #5968, opened 2024-04-26
Bug: Should not focus on root element


aquarius-wing:
Lexical version:

## Steps To Reproduce

1. Open [Playground](https://playground.lexical.dev/)
2. Remove all text and type ``` and space key
3. Type some text like 11111
4. Press Arrow Up Key multiple times

https://github.com/facebook/lexical/assets/16633962/25e84327-c799-4e6a-a178-0455c8520f82

Link to code example:

## The current behavior

You'll notice that the selection switches back and forth between the text node and the root node

## The expected behavior

Just focus the text node only

StyleT:
Hi! 
Thanks for noticing this. IMO it actually shall stick to `root` and stay there. Why? How else can you insert something before that code block? :) 
Also cursor shall be placed before code block

aquarius-wing:
It's turely work when we want to add text before code block, but it doesn't clear to tell user "now we can insert text before code block!" when focus the root element.
Because it seems exactly the same.

In my situation, my solution is then user press up, it will insert a paragraph before directly.

https://github.com/facebook/lexical/assets/16633962/f577fb46-cbef-4a32-a017-9307aa96c3ab

Speaking of which, why do I raise this question? It's to ensure that the focus can be correctly disengaged at the beginning of my editor (not the application shown in the video above).

## Example
It's like we have a form, which has input , and a lexical editor, I want the user to be able to navigate between them by pressing the up arrow key.

However, the behavior differs when we focus on the code block, quote and offset is set to 0, it will make the code I write become complex, I believe this is an area that could be improved.

So I suggest, the selection can not be switch to the root element.

levensta:
Closed by https://github.com/facebook/lexical/pull/8393

@etrepum fyi

PR #6103, opened 2024-05-14
[lexical-core][code-node] Bug Fix: Adding an empty paragraph before code node

MaxPlav:
## Description
Adding an empty paragraph before a code node when a user clicks Arrow Up from the code block and there are no other nodes before the code node.

**Closes:** #5968

### Before

https://github.com/facebook/lexical/assets/5062807/395ad864-807f-4a3e-ab3e-e2dc933c04dd

### After

https://github.com/facebook/lexical/assets/5062807/428af10e-d3c0-41fb-843b-404bb974a990

### DIFF
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -21,6 +21,7 @@ import './CodeHighlighterPrism';
 import {mergeRegister} from '@lexical/utils';
 import {
   $createLineBreakNode,
+  $createParagraphNode,
   $createTabNode,
   $createTextNode,
   $getNodeByKey,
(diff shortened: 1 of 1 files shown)


StyleT:
For the context: we have similar issue with QuoteNode (although selection there never jumps to the root node), so maybe we shall try to fix both :) 


https://github.com/facebook/lexical/assets/1125900/fee500cf-9ddb-4884-806d-ce3e577dc429

MaxPlav:
The issue is reproducible for other ElementNodes and should be fixed not only for the playground editor. Checked on ImageNode and it's possible to set the cursor before the node for adding a new element.

StyleT:
Discussed with @MaxPlav that more comprehensive solution needed here that accounts for more nodes and use cases (ex: two "column layout" nodes in the middle of the document)

PR #6107, opened 2024-05-15
Add exportJSON to ArtificialNode

thegreatercurve:
Fix browser warnings for missing method on ArtificialNode.

### DIFF
--- a/packages/lexical/src/nodes/ArtificialNode.ts
+++ b/packages/lexical/src/nodes/ArtificialNode.ts
@@ -20,4 +20,8 @@ export class ArtificialNode__DO_NOT_USE extends ElementNode {
     const dom = document.createElement('div');
     return dom;
   }
+
+  exportJSON() {
+    return super.exportJSON();
+  }
 }


potatowagon (approved):
thanks lgtm

packages/lexical/src/nodes/ArtificialNode.ts
```diff
@@ -20,4 +20,8 @@ export class ArtificialNode__DO_NOT_USE extends ElementNode {
     const dom = document.createElement('div');
     return dom;
   }
+
+  exportJSON() {
+    return super.exportJSON();
```
zurfyx, line 25: Let's make it throw an invariant instead please. This Node should not exist and should never land on any database.

zurfyx (changes requested):

packages/lexical/src/nodes/ArtificialNode.ts
etrepum: I have a commit in #6048 that just ignores the warnings for these nodes, which is another approach that would preserve expected runtime errors without sacrificing DX https://github.com/facebook/lexical/pull/6048/commits/8cd0664de7613d7a9cf6b0ed86cac1d56f2fd48e

potatowagon: im not sure if changing to invariant would continue to raise errors. the browser warnings was reported to break the serialisation of the editor state on KP's usecase (T189214869), hence the goal of the PR to silence the warnings.

potatowagon: > I have a commit in #6048 that just ignores the warnings for these nodes, which is another approach that would preserve expected runtime errors without sacrificing DX [8cd0664](https://github.com/facebook/lexical/commit/8cd0664de7613d7a9cf6b0ed86cac1d56f2fd48e)

thanks for this, lgtm!

thegreatercurve:
Closing in favour of #6048

PR #6110, opened 2024-05-15
[Lexical] Replace code with key to check KeyboardEvents

Sahejkm:
Replace code with key to check KeyboardEvents

Before:

## NumPad Return

https://github.com/facebook/lexical/assets/163521239/dc711375-2e55-4115-80a5-366af49e4129

## Dvorak Copy

https://github.com/facebook/lexical/assets/163521239/b4976cc2-4010-4550-bb09-2c4f03472694

After:

## NumPad Return

https://github.com/facebook/lexical/assets/163521239/355ec910-b3fb-4b04-b071-67f48d0e611a

## Dvorak Copy

https://github.com/facebook/lexical/assets/163521239/15c95677-ad6f-498d-8928-b2eea351950d

## Android Backspace

https://github.com/facebook/lexical/assets/163521239/0d22af12-8af0-4b4e-81dc-9993377b4e52

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -516,7 +516,7 @@ function $canRemoveText(
 
 function isPossiblyAndroidKeyPress(timeStamp: number): boolean {
   return (
-    lastKeyCode === '229' &&
+    lastKeyCode === 'PageUp' &&
     timeStamp < lastKeyDownTimeStamp + ANDROID_COMPOSITION_LATENCY
   );
 }
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -516,7 +516,7 @@ function $canRemoveText(
 
 function isPossiblyAndroidKeyPress(timeStamp: number): boolean {
   return (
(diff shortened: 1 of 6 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -516,7 +516,7 @@ function $canRemoveText(
 
 function isPossiblyAndroidKeyPress(timeStamp: number): boolean {
   return (
-    lastKeyCode === 'Delete' &&
+    lastKeyCode === 'Dead' &&
```
Daniel15, line 519: Does this still work on Android?

Sahejkm: i will test the issue reported https://github.com/facebook/lexical/issues/1973 for which this was introduced in https://github.com/facebook/lexical/pull/2412 , but looks like there were still lot of android backspace issues reported even after this fix and check was introduced 

i updated the value 1:1 to be as per MDN reference https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values  ( earlier various sources suggest different values for this keycode 229)

Screenshot 2024-05-17 at 10 28 12 AM

potatowagon (approved):
lgtm

Issue #5564, opened 2024-01-29
Load img error UI


zurfyx:
Currently, if an image fails to load because of a network error, page is down etc. the Node is silently inserted but no UI is shown. Ideally, we show some piece of UI that shows that the image failed to load somehow and that it's also easy to remove.

I.e.
![image](https://github.com/facebook/lexical/assets/193447/70957baa-b280-4f20-8d08-50a0317f679c)

Shiv-2712:
Hello, I am enthusiastic about tackling this issue. Given my experience in handling UI issues, I would appreciate it if you could assign this task to me.

Shiv-2712:
Hello again,

After thorough investigation, I identified the cause behind the UI going blank when the user faces network issues or page loading problems. The main issue stems from the lazy loading of a crucial component, leading to the error message: ```Uncaught TypeError: Failed to fetch dynamically imported module: http://localhost:3000/src/nodes/ImageComponent.tsx.``` This error occurs because the component wasn't loaded initially, and when the network connection is lost, it triggers this error, resulting in a blank UI.

To tackle this issue, I made adjustments to the code by replacing the dynamic import statement with a synchronous import for the ***ImageComponent***. This alteration ensures that the component is loaded at the beginning, resolving the error that arises when the network is disconnected. Consequently, even in offline scenarios, the UI remains functional, and the uploaded image is displayed successfully.

P.S.: After implementing these changes, the bundle size increased slightly from 8.8 MB to 8.9 MB.

Shiv-2712:
I've identified a potential solution to the problem, but unfortunately, the ```Suspense``` fallback doesn't seem to be functioning as expected. Despite implementing the suspense fallback in both the ```ImageComponent``` and ```ImageNode``` components within the node directory, the issue persists, and I'm having difficulty identifying their parent component.

Upon further investigation, it appears that the parent component should also implement the suspense library fallback. However, I'm struggling to locate the parent component that is utilizing the suspense library, or perhaps this is beyond my current understanding. If you notice and comprehend the issue, I would appreciate your assistance in resolving it.

etrepum:
[...]

PR #6111, opened 2024-05-15
[lexical-playground][image-node] Bug Fix: Load image error UI

MaxPlav:
Currently, if an image fails to load because of a network error, the page is down, etc. the Node is silently inserted but no UI is shown.

## Description

Adding a fallback image that would substitute an original image if it fails to load due to a network failure etc. 

**Closes:** #5564

## Test plan
https://github.com/facebook/lexical/issues/5564#issuecomment-2040647027

1. Open https://playground.lexical.dev/ in a browser
2. Insert an image and block the image load request via the dev tools.
3. Try to insert the image once again.
4. See that a fallback image is loaded instead.

### Before

https://github.com/facebook/lexical/assets/5062807/8501bb84-ba14-4f63-be25-dc44d0fc65c9

### After

https://github.com/facebook/lexical/assets/5062807/a35980d7-4358-4eca-8806-626b36e10be5


packages/lexical-playground/src/images/broken-image.png
StyleT, line 1: May I suggest using SVG instead of PNG? You can take this one (MIT license) for example: https://www.svgrepo.com/svg/377374/image-broken

StyleT (commented):
LGTM overall!

packages/lexical-playground/src/images/broken-image.png
MaxPlav: thanks, replaced the fallback image with that
Screenshot 2024-05-15 at 16 14 43

StyleT (approved):

PR #6117, opened 2024-05-16
[Lexical] Add label and run extended test on approve

### LINKED ISSUE #6097: Feature: Add extended-tests label on approval

Sahejkm:
- Add label and run extended test on approve

- followup on #6080 

- Fixes #6097

### DIFF
--- a/.github/workflows/add-label-run-extended-tests-approve.yml
+++ b/.github/workflows/add-label-run-extended-tests-approve.yml
@@ -1,4 +1,4 @@
-name: Add Label to run extended tests on approve
+name: Extended tests on approve
 
 on:
   pull_request_review:
--- a/.github/workflows/add-label-run-extended-tests-approve.yml
+++ b/.github/workflows/add-label-run-extended-tests-approve.yml
@@ -9,7 +9,7 @@ concurrency:
 
 jobs:
   label_approved_pr:
-    if: github.event.review.state == 'commented' && !contains( github.event.pull_request.labels.*.name, 'extended-tests')
+    if: github.event.review.state == 'approved' && !contains( github.event.pull_request.labels.*.name, 'extended-tests')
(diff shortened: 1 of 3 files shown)


.github/workflows/add-label-run-extended-tests-approve.yml
```diff
@@ -0,0 +1,24 @@
...
+      pull-requests: write
+    steps:
+      - run: gh pr edit "$NUMBER" --add-label "extended-tests"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          NUMBER: ${{ github.event.pull_request.number }}
+  e2e-tests:
+    needs: label_approved_pr
+    uses: ./.github/workflows/call-e2e-all-tests.yml
```
Sahejkm, line 24: - initially i was thinking on "pull_request" "labeled" event would auto trigger after label_approved_pr job finishes, but it wasn't triggering from the tests, 
- so i added this with a dependency on label_approved_pr if label_approved_pr executes then only this job will execute, giving same results

Sahejkm: i think this is known issue, there is a solution for this https://github.com/orgs/community/discussions/25565 , lemme try

potatowagon (approved):
can try this out, if it works gotta update the docs about the extended-tests label

PR #6120, opened 2024-05-16
[Lexical] Add label to run e2e tests on approve

Sahejkm:
- Add label and run extended test on approve

- followup on https://github.com/facebook/lexical/pull/6080

- Fixes https://github.com/facebook/lexical/issues/6097

### DIFF
--- a/.github/workflows/add-label-run-extended-tests-approve.yml
+++ b/.github/workflows/add-label-run-extended-tests-approve.yml
@@ -1,4 +1,4 @@
-name: Add Label to run extended tests on approve
+name: Extended tests on approve
 
 on:
   pull_request_review:
--- a/.github/workflows/add-label-run-extended-tests-approve.yml
+++ b/.github/workflows/add-label-run-extended-tests-approve.yml
@@ -9,7 +9,7 @@ concurrency:
 
 jobs:
   label_pull_requests:
-    if: github.event.review.state == 'commented' && !contains( github.event.pull_request.labels.*.name, 'extended-tests')
+    if: github.event.review.state == 'approved' && !contains( github.event.pull_request.labels.*.name, 'extended-tests')
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
--- a/.github/workflows/add-label-run-extended-tests-approve.yml
+++ b/.github/workflows/add-label-run-extended-tests-approve.yml
@@ -9,7 +9,7 @@ concurrency:
 
 jobs:
   label_approved_pr:
-    if: github.event.review.state == 'commented' && !contains( github.event.pull_request.labels.*.name, 'extended-tests')
+    if: github.event.review.state == 'approved' && !contains( github.event.pull_request.labels.*.name, 'extended-tests')
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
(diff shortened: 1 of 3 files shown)


potatowagon (approved):

etrepum (commented):
LGTM, only concern is that we might want another tag for which this would not happen (e.g. an existing no-extended-tests label). We don't have anything like that yet, so I don't think it should be a blocker.

.github/workflows/add-label-run-extended-tests-approve.yml
```diff
@@ -0,0 +1,24 @@
...
+      pull-requests: write
+    steps:
+      - run: gh pr edit "$NUMBER" --add-label "extended-tests"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          NUMBER: ${{ github.event.pull_request.number }}
+  e2e-tests:
+    needs: label_pull_requests
+    uses: ./.github/workflows/call-e2e-all-tests.yml
```
necolas, line 24: I don't think this is right. Once the label is added, it should kick off the other `tests-extended` workflow automatically, as that already gets triggered when labels are added.

.github/workflows/add-label-run-extended-tests-approve.yml
```diff
@@ -0,0 +1,24 @@
+name: Add Label to run extended tests on approve
+
+on:
+  pull_request_review:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
necolas, line 8: Why cancel in progress?

.github/workflows/add-label-run-extended-tests-approve.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  label_pull_requests:
```
necolas, line 11: The naming format used for all other workflows is `foo-bar`. So this one should probably be called something like `label-after-approved`

.github/workflows/add-label-run-extended-tests-approve.yml
```diff
@@ -0,0 +1,24 @@
+name: Add Label to run extended tests on approve
```
necolas, line 1: The file name of this workflow should be 'add-label-after-approved'. This workflow shouldn't be running e2e tests, just adding a label

necolas:
I think the workflow file merged in this PR needs revisiting

etrepum:
It's too bad that GitHub doesn't really offer any way to simulate these things, even to see what a matrix would expand to

.github/workflows/add-label-run-extended-tests-approve.yml
Sahejkm: I tested this and found out that 1 GitHub workflow action can't trigger another workflow action and hence this was needed, with a "needs" check , the second job won't run unless 1 job is not skipped. Without this tests don't trigger on approve

.github/workflows/add-label-run-extended-tests-approve.yml
Sahejkm: Thanks for the feedback, will update the naming conventions

Issue #6132, opened 2024-05-17
Bug: Investigate size-limit regression caused in theory by automatic jsx runtime


etrepum:
I've received some reports that after #6088 was merged there was a size-limit regression. I wasn't given very concrete details about the issue so I am centralizing what I know so far into an issue.

* #6131 size-limit does not produce comments from forks, so this was not detected during the PR review cycle
* I've been told that 'it's JSX. It's visible in the package bundle diff' but that seems strange since at least in the OSS repo the size-limit action is not measuring any bundle that includes React!

/cc @Sahejkm @StyleT

etrepum:
I am not seeing any significant difference in size.
| Module | v0.14.5 | v0.15.0 | #6048 | #6048/v0.14.5 |
|-|-|-|-|-|
| Lexical.js | 22.27 kB | 22.44 kB | 22.44 kB | 0.76% |
| LexicalRichText.js | 32.47 kB | 33.08 kB | 33.14 kB | 2.06% | 
| LexicalPlainText.js | 32.47 kB | 33.09 kB | 33.15 kB | 2.09% |

Previous v0.14.5 release
```
$ git checkout v0.14.5 && npm ci && npm run size
[…]
HEAD is now at 32841422 v0.14.5
[…]
  packages/lexical/dist/Lexical.js
  Size:         22.27 kB with all dependencies, minified and brotlied
  Loading time: 436 ms   on slow 3G
  Running time: 191 ms   on Snapdragon 410
  Total time:   626 ms
  
  packages/lexical-rich-text/dist/LexicalRichText.js
  Size:         32.47 kB with all dependencies, minified and brotlied
  Loading time: 635 ms   on slow 3G
  Running time: 500 ms   on Snapdragon 410
  Total time:   1.2 s
  
  packages/lexical-plain-text/dist/LexicalPlainText.js
  Size:         32.47 kB with all dependencies, minified and brotlied
  Loading time: 635 ms   on slow 3G
  Running time: 748 ms   on Snapdragon 410
  Total time:   1.4 s
```

Latest v0.15.0 release:
```
$ git checkout v0.15.0 && npm ci && npm run size
[…]
HEAD is now at ea8ca9b2 v0.15.0
[…]
  
  packages/lexical/dist/Lexical.js
  Size:         22.44 kB with all dependencies, minified and brotlied
  Loading time: 439 ms   on slow 3G
  Running time: 290 ms   on Snapdragon 410
  Total time:   728 ms
  
[...]

PR #6133, opened 2024-05-17
fix(lexical): #6132 react build size regression from #6088

StyleT:
## Description
Changes are based on the https://github.com/facebook/lexical/pull/6048/commits/492816e0e9512f59976965be93190bc61bb28d08 by @etrepum

**Closes:** #6132 

## Before
Screenshot 2024-05-17 at 3 42 25 PM

## After

Screenshot 2024-05-17 at 3 47 11 PM

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -94,13 +94,11 @@ function resolveExternalEsm(id) {
  * party dependencies or peerDependencies that we do not want to include
  * in the bundles.
  */
-const externals = [
-  ...Object.entries(wwwMappings).flat(),
-  'react-dom',
-  'react',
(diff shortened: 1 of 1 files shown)


necolas:
Any idea why this doesn't show up in the size workflow report? There are no scripts run in that workflow, it all seems buried in some external action

etrepum:
@necolas The size workflow report only measures three modules, none of them have any React dependencies. It never measured anything else. I did a full investigation in #6132.

https://github.com/facebook/lexical/blob/main/.size-limit.js#L55-L68

zurfyx (approved):

PR #6134, opened 2024-05-18
[Lexical] Use classic mode for jsx runtime for meta build

Sahejkm:
- Use classic mode for jsx runtime for meta build
- fix build getting exception for react/jsx-runtime module doesn't exist, use classic mode for the build


StyleT (approved):

etrepum:
Test failure looks like a flaky collab run

PR #6135, opened 2024-05-18
Remove 'legacy-events' from mac & linux e2e matrix

necolas:
## Description

Remove `legacy-events` from the e2e test matrix for all OS and browsers except `chromium` on `windows`. I think this removes 12 e2e test runs.

The `legacy-events` version of e2e tests was [adding in Oct 2021](https://github.com/facebook/lexical/pull/750) and tests for when `beforeinput` is not available. But the browser support for that event is quite broad these days, mostly in line with the lowest versions of Chrome Meta supports (in practice) internally. So the need to test not-`beforeinput` doesn't appear to be significant. I thought we'd keep chromium-windows for now, because Windows is the only OS where we see the oldest versions of Chrome (where `beforeinput` is not supported) being used.

image

### DIFF
--- a/.github/workflows/call-e2e-all-tests.yml
+++ b/.github/workflows/call-e2e-all-tests.yml
@@ -10,7 +10,7 @@ jobs:
         node-version: [18.18.0]
         browser: ['chromium', 'firefox', 'webkit']
         editor-mode: ['rich-text', 'plain-text']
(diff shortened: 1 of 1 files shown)


etrepum (commented):
Looks great to me, I was hoping that someone in Meta would look into this

ivailop7 (approved):

PR #6136, opened 2024-05-18
Fix the label-on-approval workflow

necolas:
Resolves issues with https://github.com/facebook/lexical/pull/6120 (probably)


Sahejkm:
- I don't think this works and auto triggers e2e tests, 1 hit workflow can't trigger another hit workflow as per the docs u less we use a pat token to run e2e tests

- 6120 is working fine for meta employees but not for external reviews becoz the GitHub actions don't have permissions for fork PRs and need pull request target event to work

.github/workflows/add-label-after-approved.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  pull_request_review:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  label-after-approved:
+    if: github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests')
```
Sahejkm, line 12: I tested this change locally by changing approved to commented, it doesn't auto trigger e2e tests unless some manual change on the PR

necolas:
Got it. Updated to rename jobs for clarity (since it's less about labels and more about what happens after approval), and adopted the alternative trigger that you said would work for PRs from forks

Sahejkm (dismissed):

Sahejkm:
- I think it broke :( , I will test and create a pr for this as it was ald on my mind 😀

- we have to give both pull request and pull request target 

- we should specify branch also when using pull request target

necolas:
Updated

necolas (commented):
test review

Sahejkm (approved):
Thanks @necolas for enhancing this!

PR #6139, opened 2024-05-19
Add tbody and thead tags to TableObserver

Piechota:
I ran into an issue when I tried to create table like:
```html
<table>
   <colgroup>
   .....
   </colgroup>
   <tbody>
   ....
   </tbody>
</table>
```

I created one ElementNode with <table>
one TextNode with <colgroup>
and TableNode with <tbody>

but it didn't update values in TableObserver and table didn't work correctly.
This PR fixes it.

### DIFF
--- a/packages/lexical-table/src/LexicalTableObserver.ts
+++ b/packages/lexical-table/src/LexicalTableObserver.ts
@@ -118,7 +118,11 @@ export class TableObserver {
           const target = record.target;
           const nodeName = target.nodeName;
 
-          if (nodeName === 'TABLE' || nodeName === 'TR') {
+          if (
+            nodeName === 'TABLE' ||
+            nodeName === 'TBODY' ||
+            nodeName === 'TR'
+          ) {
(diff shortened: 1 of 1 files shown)


ivailop7 (dismissed):

packages/lexical-table/src/LexicalTableObserver.ts
```diff
@@ -118,7 +118,11 @@ export class TableObserver {
           const target = record.target;
           const nodeName = target.nodeName;
 
-          if (nodeName === 'TABLE' || nodeName === 'TR') {
+          if (
+            nodeName === 'TABLE' ||
+            nodeName === 'TBODY' ||
+            nodeName === 'TR'
+          ) {
```
ivailop7, line 125: by this logic, shouldn't THEAD be included as well?

Piechota: good point
added it

ivailop7 (approved):

PR #6140, opened 2024-05-20
[Lexical] Fix meta sync build failing due to recent changes in LexicalErrorBoundary.tsx

Sahejkm:
- #6088 removes 

> import * as React from 'react'; 

from LexicalErrorBoundary.tsx 

- but in generated file LexicalErrorBoundary.dev.js / LexicalErrorBoundary.prod.js  this import was needed,
-  would need further investigation whats going wrong, but this should fix the WWW sync causing some e2e tests to fail with the issue replicable on trying the steps manually as well

Screenshot 2024-05-20 at 12 19 39 PM

### DIFF
--- a/packages/lexical-react/src/LexicalErrorBoundary.tsx
+++ b/packages/lexical-react/src/LexicalErrorBoundary.tsx
@@ -6,6 +6,7 @@
  *
  */
 
+import * as React from 'react';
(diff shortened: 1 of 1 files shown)


freddymeta (approved):
lgtm

potatowagon (approved):
Lgtm

PR #6141, opened 2024-05-20
Update react.md

Vishnukr1101:
fixing error : export 'LexicalErrorBoundary' (imported as 'LexicalErrorBoundary') was not found in '@lexical/react/LexicalErrorBoundary' (possible exports: default)

## Description

**Closes:** #

### Before

### After


etrepum:
This is because the docs are generated from main. The example does work, but only in the next release of lexical (after 0.15.0) where the default exports are deprecated.

Vishnukr1101:
I had a export error on version version 0.15.0, while using the example from docs.

etrepum:
Yes, what I'm saying is that this documentation is for the version that comes *after* 0.15.0. It does work, just not yet in any released version.

PR #6143, opened 2024-05-20
[lexical-react] Bug Fix: Use automatic jsx runtime with react/jsx-runtime -> react alias in www

etrepum:
## Description

Changes www builds to use automatic jsx runtime with the react/jsx-runtime module aliased to react.

As far as I can tell react's [index.fb.js](https://github.com/facebook/react/blob/c325aec1ee5ae970ea70efc9b19574aa0e72c9a1/packages/react/index.fb.js#L60) is used as the module's entrypoint for fb builds and it includes the JSX runtime (Fragment is already present in the react module).

This should solve the problems that #6134 and #6140 are trying to work around in a more systematic and future-compatible way.

## Test plan

Someone with access to intern will have to do a sync and make sure that this does indeed work.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -72,6 +72,8 @@ const wwwMappings = {
   'prismjs/components/prism-swift': 'prism-swift',
   'prismjs/components/prism-typescript': 'prism-typescript',
   'react-dom': 'ReactDOMComet',
+  // The react entrypoint in fb includes the jsx runtime
+  'react/jsx-runtime': 'react',
 };
 
 /**
@@ -192,7 +194,7 @@ async function build(name, inputFile, outputPath, outputFile, isProd, format) {
               tsconfig: path.resolve('./tsconfig.build.json'),
             },
           ],
-          ['@babel/preset-react', {runtime: isWWW ? 'classic' : 'automatic'}],
+          ['@babel/preset-react', {runtime: 'automatic'}],
         ],
       }),
       {


Sahejkm (approved):
LGTM

etrepum:
Note that there are no failing tests here, the approval labeling workflow still isn't working under some conditions #6136 #6120

Sahejkm:
Sorry about this, i am asking the OSS support for the next steps.

Looks like "pull_request_review" event was triggered , i missed while reviewing Nicolas added "pull_request_review" back, but this doesn't help with PRs from fork again :(

i have listed the problem as below:

**PROBLEM WE ARE TRYING TO SOLVE**  
We are trying to add a workflow in Lexical to
Add Label to a PR if it doesn't have the label
Run e2e tests for the PR if label added
We want to add label as we don't want to have multiple runs of e2e tests as we are trying to solve the Repo Over budget issue

**CURRENT APPROACH**   
we wrote this workflow file for the same:  
[https://github.com/.../actions/runs/9160647827/workflow](https://github.com/facebook/lexical/actions/runs/9160647827/workflow)
This workflow works well for PRs from Meta employees but it fails for PRs from Forks / external contributors with the error
GraphQL: Resource not accessible by integration (addLabelsToLabelable)
[https://github.com/.../runs/9160647827/job/25183782894](https://github.com/facebook/lexical/actions/runs/9160647827/job/25183782894)


**NEXT STEPS TO FIX FAILURE IN CURRENT APPROACH**  
We need help on resolving this issue. For Github sources we learnt following things:
- pull_request_target : this action gives the github workflow actions permissions for read/write from forked PR. But this action doesn't help us as we want to trigger this action on "Approval" , so only option we have is "pull_request_review" which doesn't have permissions to add Labels on PRs  from forks
- leverage PAT token for this workflow to add labels ?
- settings to allow GithubWorkflow Actions Access to PRs from fork (followed this OSS WP group, looks like this option is disabled by default and is not recommended to use this for security reasons)
- If this option is not possible other closest option we have is to run e2e tests on merge queue.

Sahejkm:
Sorry about this, i am asking the OSS support for the next steps.

Looks like "pull_request_review" event was triggered , i missed while reviewing Nicolas added "pull_request_review" back, but this doesn't help with PRs from fork again :(

i have listed the problem as below:

PROBLEM WE ARE TRYING TO SOLVE
We are trying to add a workflow in Lexical to
Add Label to a PR if it doesn't have the label
Run e2e tests for the PR if label added
We want to add label as we don't want to have multiple runs of e2e tests as we are trying to solve the Repo Over budget issue

CURRENT APPROACH
we wrote this workflow file for the same:
[https://github.com/.../actions/runs/9160647827/workflow](https://github.com/facebook/lexical/actions/runs/9160647827/workflow)
This workflow works well for PRs from Meta employees but it fails for PRs from Forks / external contributors with the error
GraphQL: Resource not accessible by integration (addLabelsToLabelable)
[https://github.com/.../runs/9160647827/job/25183782894](https://github.com/facebook/lexical/actions/runs/9160647827/job/25183782894)

NEXT STEPS TO FIX FAILURE IN CURRENT APPROACH
We need help on resolving this issue. For Github sources we learnt following things:

pull_request_target : this action gives the github workflow actions permissions for read/write from forked PR. But this action doesn't help us as we want to trigger this action on "Approval" , so only option we have is "pull_request_review" which doesn't have permissions to add Labels on PRs from forks
leverage PAT token for this workflow to add labels ?
settings to allow GithubWorkflow Actions Access to PRs from fork (followed this OSS WP group, looks like this option is disabled by default and is not recommended to use this for security reasons)
If this option is not possible other closest option we have is to run e2e tests on merge queue.

PR #6146, opened 2024-05-20
[lexical][auto-link] Fix auto link node escapes on second "."

MaxPlav:
## Description
The period in the separators list `PUNCTUATION_OR_SPACE ` leads to an auto-link creation prematurely before the user completes typing. Removing the "." from the list. 
Inspired by https://github.com/facebook/lexical/pull/5275 which was reverted because of Regexp changes.

**Closes:** #5254 #3546

## Test plan
2. Type "https://www.foo.com"
3. See that the link is created for the full URL.

### Before

https://github.com/facebook/lexical/assets/5062807/032f41a5-fd2c-441b-81ba-fbba7aa3aa10

### After

https://github.com/facebook/lexical/assets/5062807/92d2f557-2a45-4954-857a-88a4d428e4b6


StyleT:
I think it's not that easy :P See how it fails now:


https://github.com/facebook/lexical/assets/1125900/4ddc826d-3e0a-4982-bd88-9ce8e3f08880

MaxPlav:
> I think it's not that easy :P See how it fails now:
> 
>  Screen.Recording.2024-05-20.at.4.34.33.PM.mov

updated the regexp with a negative lookbehind to exclude special characters at the end

StyleT:
Hi! So now links doesn't work correctly when you have a dot at the end of the sentence:


https://github.com/facebook/lexical/assets/1125900/e22ece88-8305-4cfd-8894-8fbb8cc41662

MaxPlav:
> Hi! So now links doesn't work correctly when you have a dot at the end of the sentence:
> 
>  Screen.Recording.2024-05-21.at.1.01.03.PM.mov

thanks for checking that, should be fixed now

StyleT (approved):

PR #6153, opened 2024-05-21
[Lexical] Attempt to fix after approval workflow

Sahejkm:
- #6136  adds ```pull_request_target``` action but it doesn't get triggered on approve so its not helpful right now to use as is, hence could be removed to avoid confusions.

using my fine grain access PAT to trigger the workflow for add label
- it only has read/write access to PR and
- read access to content

Test Plan
- tested in https://github.com/facebook/lexical/pull/6168

TODO
- will replace my PAT with lexical-bot PAT once setup is done

### DIFF
--- a/.github/workflows/after-approval.yml
+++ b/.github/workflows/after-approval.yml
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   label-pr:
-    if: github.event.review.state == 'commented' && !contains(github.event.pull_request.labels.*.name, 'extended-tests')
+    if: github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests')
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -25,6 +25,6 @@ jobs:
         run: |
           echo "Adding label 'extended-tests' to PR $NUMBER"
           gh pr edit "$NUMBER" --add-label "extended-tests" || (echo "Failed to add label" && exit 1)
-  # e2e-tests:
-  #   needs: label-pr
-  #   uses: ./.github/workflows/call-e2e-all-tests.yml
+  e2e-tests:
+    needs: label-pr
+    uses: ./.github/workflows/call-e2e-all-tests.yml
--- a/.github/workflows/after-approval.yml
+++ b/.github/workflows/after-approval.yml
@@ -24,6 +24,3 @@ jobs:
         run: |
           echo "Adding label 'extended-tests' to PR $NUMBER"
           gh pr edit "$NUMBER" --add-label "extended-tests" || (echo "Failed to add label" && exit 1)
-  e2e-tests:
-    needs: label-pr
-    uses: ./.github/workflows/call-e2e-all-tests.yml
--- a/.github/workflows/after-approval.yml
+++ b/.github/workflows/after-approval.yml
@@ -3,10 +3,6 @@ name: After Approval
 on:
   pull_request_review:
     types: [submitted]
-  pull_request_target:
-    types: [review_requested, opened, reopened, synchronize]
-    branches:
-      - 'main'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


Sahejkm (commented):
test review

necolas:
Why did leaving a review (not approval) add the extended-tests label?

.github/workflows/after-approval.yml
```diff
@@ -3,10 +3,6 @@ name: After Approval
 on:
   pull_request_review:
     types: [submitted]
-  pull_request_target:
```
zpao, line 6: I think this is the one you want to keep. `pull_request_target` does not downgrade access of the token. `pull_request_review` does.

.github/workflows/after-approval.yml
```diff
@@ -17,8 +13,10 @@ jobs:
     if: github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests')
     runs-on: ubuntu-latest
     permissions:
+      contents: read
       pull-requests: write
     steps:
+      - uses: actions/checkout@v4
```
zpao, line 19: I don't think you need these changes and actually the permissions docs say explicitly not to checkout in this case.

> Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event.
> — https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

.github/workflows/after-approval.yml
zpao: At least that's my reading of the permissions docs where they discuss downgrading permission. `pull_request_target` is the only PR related event called out as not downgrading.

> When a workflow is triggered by the [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event, the GITHUB_TOKEN is granted read/write repository permission, even when it is triggered from a public fork.
> https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Sahejkm:
That was done to just test the changes.

Sahejkm:
> Why did leaving a review (not approval) add the extended-tests label?

That was done to test the changes only.

.github/workflows/after-approval.yml
Sahejkm: > I think this is the one you want to keep. `pull_request_target` does not downgrade access of the token. `pull_request_review` does.

This is not working for us as pull_request_target only triggers on any changes to PR, but doesn't trigger on review

.github/workflows/after-approval.yml
Sahejkm: I think I was removing pull_request_target already in the PR as it's not helping us in this case

Sahejkm:
I am thinking of exploring the custom GitHub bot for fixing the problem we are facing with privacy as it would have the required permissions in this case as we want to trigger the add label event on approval and pull_request_target doesn't help us with this.

.github/workflows/after-approval.yml
```diff
@@ -17,16 +13,14 @@ jobs:
...
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.ADD_LABEL_PAT }}
           GH_REPO: ${{ github.repository }}
           NUMBER: ${{ github.event.pull_request.number }}
         run: |
           echo "Adding label 'extended-tests' to PR $NUMBER"
           gh pr edit "$NUMBER" --add-label "extended-tests" || (echo "Failed to add label" && exit 1)
-  e2e-tests:
-    needs: label-pr
-    uses: ./.github/workflows/call-e2e-all-tests.yml
```
Sahejkm, line 32: these won't be needed as these will be triggered from github app workflow actions in https://github.com/facebook/lexical/blob/main/.github/workflows/tests-extended.yml after label is added via this workflow using PAT.

StyleT (approved):

PR #6171, opened 2024-05-24
Make placeholder accessible

zurfyx:
Placeholder won't read to screen readers, it technically can be read but it is not in the expected position in the flow. Instead, it should use ~`placeholder`~ `aria-placeholder` on the contenteditable.

Placeholder is now part of the ContentEditable (backward compatible for now), for two reasons:
1. It makes sense to couple the `contenteditable` with the placeholder. That's how native `placeholder` and `input` work.
2. This is a common pitfall. By coupling them we can ensure that either these properties are provided or none (when there's no placeholder).

Before, doesn't read, interferes with the flow:

https://github.com/facebook/lexical/assets/193447/f56858f7-95d6-4d19-b5ff-baa954e1bdea

After:

https://github.com/facebook/lexical/assets/193447/3ebd395e-aca1-4c34-9e8e-b14b0ecaf6df


StyleT (approved):

Fetz (approved):

etrepum:
I can't repro that integration test failure locally (likely has something to do with the github node cache behavior) but perhaps removing the `--no-save --no-package-lock` from the suite might fix it
https://github.com/facebook/lexical/blob/main/scripts/__tests__/integration/utils.js#L92

ivailop7 (approved):

potatowagon (approved):

potatowagon (approved):

Sahejkm (approved):

Issue #5582, opened 2024-02-03
Bug: Selection over collapsible section gets lost


ivailop7:
Lexical: 0.13.1

In the below video selecting backwards, always selects the contents inside a collapsible section, but when I select forward, sometimes, the offset becomes -1. As shown in the below video:

https://github.com/facebook/lexical/assets/7893468/c1b5d348-c67e-4de4-9616-a82d36cf3ce4

ivailop7:
I found what causes it, not sure how to fix it.
If the element in the createDOM method is changed from 'details' to 'div' everything works correctly, however, I think it makes sense to preserve the details tag irrespectively. This seems to be a very low-level thing, around how selection work. I do think, details should be treated as a div for all intended purposes, but unsure where in the selection logic, there is differentiation between the div and non-divs. Interestingly, it works fine for 'summary' which is a child of the div.

ivailop7:
Also think, details and summary should be added to:

https://github.com/facebook/lexical/blob/8a57b66abceda34aee403bccfbb809b57ed29af5/packages/lexical/src/LexicalUtils.ts#L1644-L1650

ivailop7:
I can confirm this works fine in Safari and Firefox, and in Chrome only when the line before is empty. We have a selection issue in Chrome, which we don't in Firefox, compare the number of calls two recordings below in different browsers:

Firefox - OK (<50 hits):

https://github.com/facebook/lexical/assets/7893468/d468c211-0e44-4b28-922a-addf4ef3e278

Chrome - Problematic (1k+ hits):

https://github.com/facebook/lexical/assets/7893468/68870bdb-4cf9-4a85-84d4-7faaedda2891

ivailop7:
Continued investigating, the root of the issue appears to be the discrepancy of the DOM selection returned by the browser (Correct in Firefox, Safari - Range Selection, Chrome returns caret collapsed selection)

<img width="1496" alt="Screenshot 2024-05-26 at 23 14 31" src="https://github.com/facebook/lexical/assets/7893468/ba36a13f-97ac-4150-b1c5-b0b7d4634405">

zurfyx:
I can't repro this exact issue (is it Windows?) but `section` seems buggy on Chrome, maybe we should consider a `div` implementation as you suggested above as the least intrusive way to mitigate the issue.

https://codepen.io/zurfyx/pen/mdYOmbv?editors=1010

Notice 2 issues with selection:
[...]

PR #6179, opened 2024-05-27
Collapsible as div

zurfyx:
_Memorial day pair coding with @ivailop7 👯_

### Why `div` vs the natively supported `details`?

Turns out that `details` is broken on Chrome. More details on the issue https://github.com/facebook/lexical/issues/5582

Selection now works well on Chrome:

https://github.com/facebook/lexical/assets/193447/e5e15246-0ef9-4853-aa5e-df018a9df373

Find still works

https://github.com/facebook/lexical/assets/193447/840b20e1-6a43-42eb-b72c-bb32212d10f2

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -1452,7 +1452,7 @@ test.describe('Tables', () => {
             <th
               class="PlaygroundEditorTheme__tableCell PlaygroundEditorTheme__tableCellHeader">
               <p class="PlaygroundEditorTheme__paragraph"><br /></p>
-              <details class="Collapsible__container" open="">
+              <div class="Collapsible__container">
                 <summary class="Collapsible__title">
                   <p class="PlaygroundEditorTheme__paragraph">
                     <span data-lexical-text="true">123</span>
@@ -1466,7 +1466,7 @@ test.describe('Tables', () => {
                   <p class="PlaygroundEditorTheme__paragraph"><br /></p>
                   <p class="PlaygroundEditorTheme__paragraph"><br /></p>
                 </div>
-              </details>
+              </div>
               <p class="PlaygroundEditorTheme__paragraph"><br /></p>
             </th>
             <th
@@ -1489,7 +1489,7 @@ test.describe('Tables', () => {
             <th
               class="PlaygroundEditorTheme__tableCell PlaygroundEditorTheme__tableCellHeader">
               <p class="PlaygroundEditorTheme__paragraph"><br /></p>
-              <details class="Collapsible__container" open="">
+              <div class="Collapsible__container">
                 <summary class="Collapsible__title">
                   <p class="PlaygroundEditorTheme__paragraph">
                     <span data-lexical-text="true">123</span>
@@ -1498,7 +1498,7 @@ test.describe('Tables', () => {
                 <div class="Collapsible__content">
                   <p class="PlaygroundEditorTheme__paragraph"><br /></p>
                 </div>
-              </details>
+              </div>
               <p class="PlaygroundEditorTheme__paragraph"><br /></p>
             </th>
             <th
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -1427,6 +1427,7 @@ test.describe('Tables', () => {
     page,
     isPlainText,
     isCollab,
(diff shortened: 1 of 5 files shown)


ivailop7 (approved):
Thanks Gerard!

fantactuka:
Sadly, auto expanding on matching content the using cmd+f will no more 🥲

zurfyx:
> Sadly, auto expanding on matching content when using cmd+f will no more 🥲

@fantactuka Good callout but fear not! Look at this API that @ivailop7 discovered [hidden until-found](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)

I'll follow up with an else case for other browsers since is Chromium™️

zurfyx:
Turns out neither FF nor Safari support CMD + F on `details`. I'm proposing to keep `section` anyway for correctness since it's likely that eventually all browsers will properly support it and that's the best we can offer for now.

packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleContentNode.ts
```diff
@@ -36,9 +42,33 @@ export class CollapsibleContentNode extends ElementNode {
     return new CollapsibleContentNode(node.__key);
   }
 
-  createDOM(config: EditorConfig): HTMLElement {
+  createDOM(config: EditorConfig, editor: LexicalEditor): HTMLElement {
     const dom = document.createElement('div');
     dom.classList.add('Collapsible__content');
+    if (IS_CHROME) {
+      editor.getEditorState().read(() => {
```
etrepum, line 49: Is createDOM ever expected to be called without activeEditorState set correctly?

potatowagon:
cool, i did not know lexical had a collapsible container until this PR! 

unrelated topic: im alittle surprised the > on the far left isnt rotating 90 degrees to show the collapse/expand state, something like


https://github.com/facebook/lexical/assets/19604232/c1d54e4d-820e-493d-8603-7a45008e719d

shall i file a feature request for this if it makes sense for lexical?

packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleContainerNode.ts
```diff
@@ -53,24 +58,47 @@ export class CollapsibleContainerNode extends ElementNode {
   }
 
   createDOM(config: EditorConfig, editor: LexicalEditor): HTMLElement {
-    const dom = document.createElement('details');
+    // details is not well supported in Chrome #5582
+    let dom: HTMLElement;
+    if (IS_CHROME) {
+      dom = document.createElement('div');
```
potatowagon, line 64: oh the `<div Collapsible__container>` isnt being toggled open (`<div Collapsible__container open>`) for chrome thats why the > doesnt rotate

potatowagon:
> cool, i did not know lexical had a collapsible container until this PR!
> 
> tho the > on the far left isnt rotating 90 degrees to show the collapse/expand state,
> 
> before (on prod):
> 
>  Screen.Recording.2024-05-28.at.9.39.10.AM.mov 
> after (thi PR):
> 
>  Screen.Recording.2024-05-28.at.9.39.25.AM.mov

resolved in https://github.com/facebook/lexical/pull/6179/commits/5721f337efc93446e344f68ed35a0a9b03832177

packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleContainerNode.ts
potatowagon: resolved in https://github.com/facebook/lexical/pull/6179/commits/5721f337efc93446e344f68ed35a0a9b03832177

potatowagon: sorry, pushed by accident, didnt know it will add on as a commit to this PR

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -1443,6 +1444,13 @@ test.describe('Tables', () => {
     await page.keyboard.press('Enter');
     await page.keyboard.press('Enter');
 
+    const collapsibleOpeningTag =
+      browserName === 'chromium'
+        ? '<div class="Collapsible__container">'
```
potatowagon, line 1449:
```suggestion
        ? '<details class="Collapsible__container" open="true">'
```

potatowagon (approved):
👍 nice

Issue #5675, opened 2024-03-01
Feature: proposal for Lexical Devtools


Aemulatio:
Hi. First of all thank for your framework, its great!
I'm working with lexical and when creating some intricate logic I want to see current state of the lexical (e.g. json format) for that I create a temporary button which help me to get it.. Sooo, it would be amazing to see devtools browser extension in the future. 

As I think the most hard part is get all instances on the page. 
Maybe somehow get all namespaces (if they are unique) and getting data for each of them.

StyleT:
Hello! I think it's on the roadmap, and personally I wanted to work on it if no one takes it on first.

StyleT:
This one is in progress: #5747  #5830

StyleT:
**Development plan (v1, MVP):**

- [x] Scaffold a new [cross-browser web extension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Build_a_cross_browser_extension) which works with Firefox, Chrome, Edge, and Safari.
    #5747 
- [x] Port the existing inline ASCI Lexical state tree representation to the extension.
    #5830
- [ ] @StyleT Register (& share access with Core Team) developer accounts in:
  - [ ] Chrome Web Store ($5 one time fee)
  - [ ] Apple Developer Program ($100/mo)
  - [ ] [Firefox AMO](https://extensionworkshop.com/documentation/publish/submitting-an-add-on/#listing-on-amo) (free)
- [ ] @StyleT Build a CI pipeline within our Github Actions which handles publishing new extension versions to their respective web extension stores and publish v1 to the respective stores

**Development plan (v2):**
- [ ] _[Up for grabs]_ Create a more interactive representation of the Lexical state tree, akin to the Component view within [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en).
  - [ ] Respective tree nodes should be collapsible or expandable.
  - [ ] Respective tree nodes should highlight the relevant DOMNode on hover of the Lexical Node.
  - [ ] Add slide out panel with all properties on the node.
  - [ ] Add a view for the current selection.
  - [ ] Add the ability to time-travel between different editor states.
  - [ ] Add a log of all Lexical commands and their respective payloads.
  - [ ] View should be performant even for larger Lexical state trees.
  - [ ] View should have a suite of unit and E2E tests which test for basic Lexical states.
  - [ ] Add details of editor config.
[...]

PR #6180, opened 2024-05-28
[lexical-react][lexical-dev-tools-core] Feature: Allow TreeView custom print output

irahopkinson:
## Description
- Previously, the TreeView output couldn't be customized.
- This PR allows you to customize existing output and support custom node output.
  - added optional `customPrintNode` function property to `TreeView`.
    - If the function is not defined then the built-in output is used in the TreeView.
    - If the function returns nothing then the built-in output is used.
    - If the function returns text then that text is used in the output for that node.
  - moved output of mark IDs to `printNode` so it can be customized (it didn't seem to belong where it was anyway).
  - made appropriate changes to flow.
  - Note I didn't know how to update the API documentation as requested in `https://github.com/facebook/lexical/blob/main/CONTRIBUTING.md` but it looks like it is auto-generated anyway.

### Before

Previously in the Playground the TreeView output couldn't be customized:
image

### After

To test, I changed `packages\lexical-playground\src\plugins\TreeViewPlugin\index.tsx` to:
```typescript
/**
 *
 *
 */

import {$isMarkNode} from '@lexical/mark';
import {TreeView} from '@lexical/react/LexicalTreeView';
import {LexicalNode} from 'lexical';

function $customPrintNode(node: LexicalNode) {
  if ($isMarkNode(node)) {
    return `uids: [ ${node.getIDs().join(', ')} ]`;
  }
  return '';
}

export default function TreeViewPlugin(): JSX.Element {
  return (
    <TreeView
      viewClassName="tree-view-output"
      treeTypeButtonClassName="debug-treetype-button"
      timeTravelPanelClassName="debug-timetravel-panel"
      timeTravelButtonClassName="debug-timetravel-button"
      timeTravelPanelSliderClassName="debug-timetravel-panel-slider"
      timeTravelPanelButtonClassName="debug-timetravel-panel-button"
      customPrintNode={$customPrintNode}
      editor={editor}
    />
  );
}
```

And then the Playground displays:
image

Note in the TreeView the mark IDs array label is now "uids" (instead of "id").


packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -245,9 +249,17 @@ function normalize(text: string, obfuscateText: boolean = false) {
   return textToPrint;
 }
 
-// TODO Pass via props to allow customizability
```
potatowagon, line 248: this comment may still be valid, perhaps we shouldnt delete it

packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -245,9 +249,17 @@ function normalize(text: string, obfuscateText: boolean = false) {
...
+function printNode(
+  node: LexicalNode,
+  customPrintNode?: CustomPrintNode,
+  obfuscateText: boolean = false,
+) {
+  const customPrint = customPrintNode
+    ? customPrintNode(node, obfuscateText)
+    : undefined;
+  if (customPrint) {
+    return customPrint;
```
potatowagon, line 261:
```suggestion
  if (customPrintNode !== undefined) {
    return customPrintNode(node, obfuscateText);
```

potatowagon:
what are some example usecases of custom print node?

packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -245,9 +249,17 @@ function normalize(text: string, obfuscateText: boolean = false) {
...
+  } else if ($isTextNode(node)) {
```
potatowagon, line 262: since the if block above returns we dont need to wrap this in an else

```suggestion
 if ($isTextNode(node)) {
```

packages/lexical-devtools-core/src/generateContent.ts
irahopkinson: This PR allows full customizability for the TreeView print output for any existing node or any custom node and it it does it by passing a function in through the props. Perhaps I'm missing something but what part of that comment still needs implementing?

packages/lexical-devtools-core/src/generateContent.ts
irahopkinson: We only want the custom print output if `customPrintNode(...)` produces some text, i.e. if it returns an empty string because it doesn't handle a node then the existing code below would run and we would get the built-in print output. Your suggested change would not do that but return no output. That's not what we want here.

packages/lexical-devtools-core/src/generateContent.ts
irahopkinson: I was just following the existing pattern as all the following `else if` blocks all return. Strictly they could all be changed to `if` blocks but I think that would make the code harder to read. I'm happy to change them all to `if` blocks if you prefer that?

irahopkinson (commented):
Thanks for your speedy response.

irahopkinson:
> what are some example usecases of custom print node?

The test code in the PR description above is one example where we override the existing output for a `MarkNode`. So existing output can be changed when debugging, e.g. I could include an attribute's value in the output.

Also you can add handling of TreeView output for custom nodes, e.g. if I created a Custom `MarkNode` that grouped IDs according to type then I could display the output of those grouped IDs when debugging.

packages/lexical-devtools-core/src/generateContent.ts
potatowagon: `if (customPrint)` isnt clear because customPrint isnt a boolean. preferably be more explicit about the comparison, such as
```
if (customPrint !== undefined && customPrint.length > 0) {
    return customPrint;
  }
```

packages/lexical-devtools-core/src/generateContent.ts
potatowagon: the custom print case is a seperate case from the default print so its clearer to have it in its own if block rather then lump it tgt with the default print conditions

packages/lexical-devtools-core/src/generateContent.ts
irahopkinson: That makes sense, done.

packages/lexical-devtools-core/src/generateContent.ts
irahopkinson: That makes sense also, done.

irahopkinson (commented):
I also rebased since `generateContent.ts` file changed on the **main** branch but I made my changes in a new commit for you to review.

StyleT:
Hi! 

Thanks for contributing! I think that overall strategy here is to eventually kill TreeView (see V2 plan here https://github.com/facebook/lexical/issues/5675#issuecomment-2043312215)... So I'm not sure it's a good idea to expand it now.

@potatowagon wdyt?

irahopkinson:
@StyleT yep totally get that and have seen that plan (and I'm excited for those changes ASAP) but there is no timing on that plan. So even if this change is only useful for a few months that would be helpful for me and hopefully others. Also given that the `customPrintNode` property is optional I wouldn't think adding it would hurt anything. However, I'm happy to let this PR go if it doesn't fit.

packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -32,6 +32,11 @@ import {
 
 import {LexicalCommandLog} from './useLexicalCommandsLog';
 
+export type CustomPrintNode = (
```
potatowagon, line 35: nit: preferably rename CustomPrintNode to CustomPrintNodeFn to not confuse this with other lexical nodes such as TextNode

potatowagon:
the re

> Hi!
> 
> Thanks for contributing! I think that overall strategy here is to eventually kill TreeView (see V2 plan here [#5675 (comment)](https://github.com/facebook/lexical/issues/5675#issuecomment-2043312215))... So I'm not sure it's a good idea to expand it now.
> 
> @potatowagon wdyt?

the example usecases listed by @irahopkinson makes sense to me so im good to have this merge

packages/lexical-devtools-core/src/generateContent.ts
irahopkinson: Done.

potatowagon (approved):
👍

PR #6182, opened 2024-05-28
[lexical-dev-tools-core] Feature: Index tree view cmds

potatowagon:
## Description
I always get confused which way the commands list is flowing in the debug view, so adding an index to show that clearly

it also helps show that commands listed are jumped if > 10 commands get updated in the list

### Before

https://github.com/facebook/lexical/assets/19604232/003f0876-027c-40ba-9b21-188f816712bc

^ not clear which direction commands are flowing

### After

https://github.com/facebook/lexical/assets/19604232/7a04cab5-13f9-4af7-a0b7-27f4c444b7b1

^ with index the list is clearer

### DIFF
--- a/packages/lexical-devtools-core/src/generateContent.ts
+++ b/packages/lexical-devtools-core/src/generateContent.ts
@@ -28,8 +28,9 @@ import {
   $isParagraphNode,
   $isRangeSelection,
   $isTextNode,
-  LexicalCommand,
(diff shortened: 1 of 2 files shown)


thegreatercurve (approved):

potatowagon:
broken e2e tests dont seem related 

Tables › Can copy + paste (internal) using Table

PR #6187, opened 2024-05-28
[docs][examples] Update plain text and rich text examples

### LINKED ISSUE #5989: Feature: bring all examples into the monorepo
The homepage examples are all in the examples folder of the monorepo which makes it easy for anyone on the team to keep them maintained and up to date.

There are other examples in the documentation that point to codesandbox that are owned by individual developers which are harder to manage.

- https://codesandbox.io/s/lexical-plain-text-example-g932e
- https://codesandbox.io/s/lexical-rich-text-example-5tncvy
- https://codesandbox.io/embed/ecstatic-maxwell-kw5utu?fontsize=14&hidenavigation=1&module=/src/Editor.js,/src/plugins/CollapsiblePlugin.ts,/src/nodes/CollapsibleContainerNode.ts&theme=dark&view=split

I suggest we copy these to the examples folder and change the links to use StackBlitz as we are using on the homepage.

We could also update the PR template to include StackBlitz as an example instead of CodeSandbox.

CodeSandbox is a bit of a mess because they limit free accounts to 20 sandboxes, so it will make it much more likely that a sandbox on an older issue, discord discussion, or even the docs ends up getting deleted.

MaxPlav:
1. Created the plain text example project `react-plain-text`.
2. Updated the rich text example project with several plugins.
3. Update the examples links in the documentation to point StackBlitz instead of CodeSandbox.

## Description
Closes #5989

## Test plan
Rich text example:

https://github.com/facebook/lexical/assets/5062807/b67bc447-3269-4a90-855c-20da085361d9

Plain text example:

https://github.com/facebook/lexical/assets/5062807/cf7cb51e-868e-477a-9622-a4234972033f


examples/react-rich/src/plugins/ToolbarPlugin.tsx
StyleT, line 1: IMO, we need to keep `react-rich` as simple as possible... I can kinda justify having `AutoLinkPlugin` and `FloatingLinkEditorPlugin`, but why do we need all the other things here? 

Idea behind `react-rich` example is to serve as a good starting point for anyone who aims to create it's own Lexical baked editor. So people would add (not remove) features.

examples/react-plain-text/src/App.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  // The editor theme
+  theme: ExampleTheme,
+};
+
+export default function App() {
+  return (
+    <LexicalComposer initialConfig={editorConfig}>
+      <div className="editor-container">
+        <ToolbarPlugin />
```
StyleT, line 38: I am really wondering if we need toolbar here... Shouldn't it have same features as `vanilla-js`?

The more complex examples are - the harder is to support them

StyleT (commented):
LGTM overall! Awesome work

examples/react-plain-text/src/App.tsx
StyleT: I think each feature in the example shall serve an educational purpose. So it should be there for a reason.

And the fact that it was there in the old codesandbox isn’t a good enough reason to me :) 

So unless we know why those features are needed there (ex: it’s an implementation example for particular documentation article) - I would remove them.

MaxPlav: > I think each feature in the example shall serve an educational purpose. So it should be there for a reason.
> 
> And the fact that it was there in the old codesandbox isn’t a good enough reason to me :)
> 
> So unless we know why those features are needed there (ex: it’s an implementation example for particular documentation article) - I would remove them.

hi, thanks for the explanation! I've just removed extra plugins from both examples to keep the examples as simple as possible, and updated the screenshot videos.

StyleT (approved):

PR #6189, opened 2024-05-28
[lexical] Bug Fix: mutation listener set for original node should work with the replaced node

### LINKED ISSUE #4498: Bug: registerMutationListener on replaced node

StyleT:
## Description

Based on the https://github.com/facebook/lexical/pull/6089 (this one adds tests)

Closes #4498

## Test plan

see tests!

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -445,13 +445,23 @@ export function createEditor(editorConfig?: CreateEditorArgs): LexicalEditor {
         replace = options.with;
         replaceWithKlass = options.withKlass || null;
       }
-      // Ensure custom nodes implement required methods.
+      // Ensure custom nodes implement required methods and replaceWithKlass is instance of base klass.
       if (__DEV__) {
         // ArtificialNode__DO_NOT_USE can get renamed, so we use the type
         const nodeType =
           Object.prototype.hasOwnProperty.call(klass, 'getType') &&
           klass.getType();
         const name = klass.name;
+
+        if (replaceWithKlass) {
+          invariant(
+            replaceWithKlass.prototype instanceof klass,
+            "%s doesn't extend the %s",
+            replaceWithKlass.name,
+            name,
+          );
+        }
+
         if (
           name !== 'RootNode' &&
           nodeType !== 'root' &&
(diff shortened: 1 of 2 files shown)


potatowagon (approved):
looks good thankyou

richardaum:
Is this `withKlass` property used for replace nodes documented?

etrepum:
withKlass is used in the documentation examples now

PR #6191, opened 2024-05-29
[lexical-list][lexical-playground] Bug fix: support pasting google doc checklist

### LINKED ISSUE #6022: Bug: Google Docs checklist does not paste correctly
Lexical version:

## Steps To Reproduce

1. copy a checklist from google docs
2. paste into lexical

the html copied to clipboard for checkbox is 
```
<li dir="ltr" role="checkbox" aria-checked="false" style="list-style-type:none;font-size:11pt;font-family:'Optimistic Text',sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;" aria-level="1"><img src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEgAAABICAYAAABV7bNHAAAA1ElEQVR4Ae3bMQ4BURSFYY2xBuwQ7BIkTGxFRj9Oo9RdkXn5TvL3L19u+2ZmZmZmZhVbpH26pFcaJ9IrndMudb/CWadHGiden1bll9MIzqd79SUd0thY20qga4NA50qgoUGgoRJo/NL/V/N+QIAAAQIECBAgQIAAAQIECBAgQIAAAQIECBAgQIAAAQIECBAgQIAAAQIECBAgQIAAAQIEyFeEZyXQpUGgUyXQrkGgTSVQl/qGcG5pnkq3Sn0jOMv0k3Vpm05pmNjfsGPalFyOmZmZmdkbSS9cKbtzhxMAAAAASUVORK5CYII=" width="17.599999999999998px" height="17.599999999999998px" alt="unchecked" aria-roledescription="checkbox" style="margin-right:3px;" />
```

could add an import conversion map on listnode to handle this case on the condition `<li role="checkbox" aria-checked="false">` or `unchecked` 

## The current behavior

- the check box gets pasted as a block image
Screenshot 2024-05-04 at 2 51 19 PM

Screenshot 2024-05-04 at 2 52 12 PM

## The expected behavior

converts to a lexical checklist

potatowagon:
## Description
support pasting google doc checklist

Closes #6022
## Test plan
added unit test 

### Before

Screenshot 2024-05-29 at 1 17 57 PM

^ the checkboxes are img so unusable

### After

Screenshot 2024-05-29 at 1 14 01 PM


potatowagon:
- [ ] scsdcs
- [ ] scsdcs

ivailop7 (dismissed):

packages/lexical-playground/src/nodes/ImageNode.tsx
```diff
@@ -37,9 +37,18 @@ export interface ImagePayload {
   captionsEnabled?: boolean;
 }
 
+function isGoogleDocCheckboxImg(img: HTMLImageElement): boolean {
+  return (
+    img.parentElement != null &&
+    img.parentElement.tagName === 'LI' &&
+    img.previousSibling === null &&
+    img.src.startsWith('data:image/png;base64,iVBOR')
```
StyleT, line 45: this image node has `roledescription="checkbox"` IMO it's better to check for it rather then performing base64 match

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -495,7 +495,12 @@ function updateListItemChecked(
 
 function $convertListItemElement(domNode: Node): DOMConversionOutput {
   const checked =
-    isHTMLElement(domNode) && domNode.getAttribute('aria-checked') === 'true';
+    isHTMLElement(domNode) &&
+    (domNode.getAttribute('aria-checked') === 'true'
+      ? true
+      : domNode.getAttribute('aria-checked') === 'false'
+      ? false
+      : undefined);
```
zurfyx, line 503: Why is the `undefined` important here? Nit, can we simplify this nested ternary? We can even retrieve the `getAttribute` once which saves us from a second `getAttribute` call

zurfyx (approved):

packages/lexical-list/src/LexicalListItemNode.ts
potatowagon: the undefined is for non checkbox cases, eg. bullet list. sure will save the getAttribute as var

StyleT (approved):

potatowagon:
test failures are from flaky test & unrelated to PR changes. double check on main the same tests fail

PR #6198, opened 2024-05-29
[Lexical ] run tests on approval if there is no extended-tests label manually added

Sahejkm:
## WHAT:
- run tests on approval if there is no extended-tests label manually added

## WHY:
- Sadly even after using PAT, the add label step fails for PRs from fork, latest example https://github.com/facebook/lexical/pull/6180

- Now the latest error is the github workflow is not able to read the secret and hence can't get the token for PRs from fork, as as per the github wiki the Secrets cant be passed on to the workflows running on PRs from fork, exception being GITHUB_TOKEN
- there are solutions to fix this https://jiminbyun.medium.com/github-actions-workflow-run-vs-workflow-call-3f1a5c6e19d4 but again as per the Meta internal posts, this option is again restricted from passing secrets to running workflow_runs on PRs from fork

## Options: 
now the remaining options are:

- Run tests on Approve action, without adding any label (this PR does that) : 
  - Problems with this flow : need to avoid duplicate runs for duplicate approvals ? Could refer https://github.com/marketplace/actions/skip-duplicate-actions

- Run tests on merge/push:
  - this gives the signals/confidence to the Approver/Contributor very late in the cycle when merging the PR,  
  - we can't stop the merge on e2e failures as  if there is a flaky e2e test, it makes the need to have a very stable CI

- Run e2e tests with cron job:
  - this  might be even more riskier than Option 2 as we have to bisect which pr cause e2e failure and would be very less efficient


potatowagon (dismissed):

potatowagon (commented):
oops press wrongly

potatowagon:
did not notice its in draft will re-review when ready

potatowagon (dismissed):

potatowagon (approved):

PR #6201, opened 2024-05-29
[LexicalTextNode] Refactor: remove unnecessary type casting

potatowagon:
## Description

noticed that [`DOMConversionFn<HTMLElement>`](https://github.com/facebook/lexical/blob/9a73b2486f1fbd99b1ce7c9657f4990a5a7967fb/packages/lexical/src/LexicalNode.ts#L124) -> it takes in a `HTMLElement` as input, but quite alot of conversion fn implementations have the input type as `Node`, and then later cast it as a `HTMLElement` (or its child classes).

seems redundant as we can directly have the input type as the HTMLElement

## Test plan

npm run ci-check

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -1088,9 +1088,9 @@ export class TextNode extends LexicalNode {
   }
 }
 
-function convertSpanElement(domNode: Node): DOMConversionOutput {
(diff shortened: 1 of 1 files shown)


StyleT (approved):

potatowagon:
failing is known flaky test & dont look related

https://github.com/facebook/lexical/actions/runs/9294132717/job/25578693971?pr=6201

Issue #6144, opened 2024-05-20
Feature: Plan for versioned documentation


etrepum:
Currently, all documentation is generated whenever main is deployed. This can be confusing for users when there are new APIs or deprecations such that the documentation does not match the behavior or features of the latest released version. 

A short-term plan might be to simply mark the documentation website (and playground) with some information that lets users know that the documentation reflects main and may not be available in a released version. Docusaurus does this with a note:

> This is unreleased documentation for Docusaurus Canary 🚧 version.
> For up-to-date documentation, see the [latest version](https://docusaurus.io/docs) (3.3.2).

Longer term, whenever the API is considered stable, it would make sense to have multiple doc website available (e.g. 19.react.dev, 18.react.dev) or to use [docusaurus versioning](https://docusaurus.io/docs/versioning). Some possibilities for interim middle-ground solutions:

* Start marking methods/features with a [@since](https://jsdoc.app/tags-since) tag, although that wouldn't necessarily cover deprecations where the example code changes to match main in a backwards incompatible way (as in #6141)
* Always deploy two doc websites, one for main (maybe call this next or canary) and one that reflects the latest released version.

etrepum:
Current consensus is that we should have two doc sites, we didn't go into detail about using @since tags, I don't think that's as important at this 0.x stage.

Logistically the following things should happen:

- Decide on a name for the doc site (e.g. next.lexical.dev or canary.lexical.dev)
- Add the code to display the version the doc site corresponds to (even just next or latest), could even be done client-side based on the URL, so that people know they are looking at docs that may include unreleased features
- Change the release workflow to maintain a release branch (latest? release?) that tracks the most recent manually published version (like the latest tag in npm)
- Provision DNS and web hosting (Vercel?) for this new doc site deployed from that branch

etrepum:
Now that #6204 is released, we have the necessary build infrastructure to manage multiple documentation websites, but any further progress on that is blocked on other people with elevated privileges doing things:

1. Another release on the latest channel needs to happen to create the latest__release branch (v0.16.0 was released before #6204)
2. Someone at Meta needs to change the Vercel config for lexical.dev to build from the latest__release branch
3. (optional) Someone at Meta needs to set up a new Vercel site (and probably add Cloudflare DNS) for the domain that tracks a more current build (e.g. next.lexical.dev, nightly.lexical.dev, canary.lexical.dev, etc.) from an appropriate branch (e.g. main or nightly__release)

[...]

PR #6204, opened 2024-05-30
[*] Feature: Automated nightly releases

### LINKED ISSUE #6169: Feature: Automated nightly releases
We have basically all of the build infrastructure to do automated nightly releases, but the workflow to do it isn't finished.

Since the release schedule is generally on the order of 2-8 weeks, the pace of development is rapid, and it is very difficult to use lexical as a dependency from git, it would make a lot of sense to have an automated release channel for users who need or want access to the latest fixes and features.

Scope:

* Refactor release workflow to a workflow_call so it can be dispatched from the manual release workflow and the automated nightly workflow
* Change the version/postversion scripts (or the release workflow) to maintain a branch that tracks the latest commit published to that channel (e.g. latest__release and nightly__release) (also useful for #6144)
* Finish the nightly-release workflow

etrepum:
## Description

Refactor release workflow to use reusable workflows so the work can be shared

Add a nightly workflow to create a release every night in the nightly channel

Maintain release channel branches (`latest__release`, `nightly__release`) so that we will have the option to build the docs site from the latest released version instead of or in addition to main (#6144)

Closes #6169

## Test plan

* Try creating a release to make sure that still works
* See if it works when the cron job runs?
* I created very similar workflows and scripts in this repo to see if it *should* work and it did https://github.com/etrepum/lexical-builder

## Future Considerations

* Fix release process to use the same tag for both GitHub and NPM releases (the nightly release does not have this issue, only manual workflow_dispatch releases with a PR)
* Do not do a nightly release when there are no changes since the last one

### DIFF
--- a/packages/lexical-website/docs/maintainers-guide.md
+++ b/packages/lexical-website/docs/maintainers-guide.md
@@ -258,7 +258,7 @@ The postversion script will:
 - Create a version commit and tag from the branch
 
 This is typically executed through the `version.yml` GitHub Workflow which
-will also push the tag.
+will also push the tag and branch.
 
 ### npm run changelog
 
@@ -270,3 +270,20 @@ Update the changelog from git history.
 plus creating a tag in git, and likely other steps.
 
 Runs prepare-release to do a full build and then uploads to npm.
+
+## Release Procedure
+
+This is the current release procedure for public releases, at least as of
+May 2024 (~0.15.0).
+
+The main branch should be "frozen" during this procedure (no other PRs should
+be merged during this time). This avoids a mismatch between the contents of
+the GitHub release (created from main in step 1) and the NPM release (created
+from main in step 4).
+
+1. Create a new version with the Github Actions "Create New Release Branch" workflow (`version.yml`)
+2. Raise a PR against version branch created by that action
+3. After PR is approved with passing tests, merge PR
+4. After PR is merged to main, publish to NPM with the Github Actions "Publish to NPM" workflow (`pre-release.yml`)
+5. Create a GitHub release from the tag created in step 1, manually editing the release notes
+6. Announce the release in #announcements on Discord
--- a/.github/workflows/call-increment-version.yml
+++ b/.github/workflows/call-increment-version.yml
@@ -1,0 +1,51 @@
+name: Create New Release Branch
+on:
+  workflow_call:
+    inputs:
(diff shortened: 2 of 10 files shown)


packages/lexical-website/docs/maintainers-guide.md
```diff
@@ -270,3 +270,20 @@ Update the changelog from git history.
 plus creating a tag in git, and likely other steps.
 
 Runs prepare-release to do a full build and then uploads to npm.
+
+## Release Procedure
+
+This is the current release procedure for public releases, at least as of
+May 2024 (~0.15.0).
+
+The main branch should be "frozen" during this procedure (no other PRs should
```
potatowagon, line 279: hello thanks for this! pardon my qns:

1. what are the signals for when the release is happening? do we monitor in the actions tab for when the release workflows (im assuming call-npm-publish.yml) is running?

2. is there a way to automate the freezing? so PRs dont accidentally get merged while the release is happening.

potatowagon (approved):

packages/lexical-website/docs/maintainers-guide.md
etrepum: Those are process questions that I don't have answers for, I'm merely documenting the way I've been told it works, but I did add some notes about the risk. @ivailop7 or @acywatson may be able to provide more detail. I think the answer might be that releases happen fairly quickly and PRs don't get merged that often when several core maintainers are working on a release so the potential inconsistency is just ignored.

I think we should change it so that this isn't a concern, but the scope of this PR is to enable nightly releases and not to change the existing release process.

etrepum:
Just FYI the repository that I was testing this with is correctly doing nightly releases by cron job

* https://github.com/etrepum/lexical-builder/actions/runs/9312104213 ran the nightly release branch action
* https://github.com/etrepum/lexical-builder/branches has a nightly__release branch that is up to date
* https://github.com/etrepum/lexical-builder/tags has a v0.0.3-nightly.20240531.0 tag
* https://www.npmjs.com/package/@etrepum/lexical-builder?activeTab=versions has a [0.0.3-nightly.20240531.0](https://www.npmjs.com/package/@etrepum/lexical-builder/v/0.0.3-nightly.20240531.0) release

ivailop7:
We'll need @acywatson for this one

PR #6214, opened 2024-05-30
refactor: refactored InlineImageNode - spans for correct HTML

58bits:
[lexical-playground][inline-image-plugin] Refactor: Spans for correct HTML

Refactor

## Description

Refactored InlineImageNode into its own directory, and changed  inline image containing divs to spans for technically 'more correct' HTML for inline and float elements. Note that in theory there should be an 'inline' version of ContentEditable too.


packages/lexical-playground/src/nodes/InlineImageNode/InlineImageComponent.tsx
```diff
@@ -352,7 +352,7 @@ export default function InlineImageComponent({
   return (
     <Suspense fallback={null}>
       <>
-        <div draggable={draggable}>
+        <span draggable={draggable}>
```
potatowagon, line 355: since inline image is in a span container it does make sense for its children to be span instead of div

currently:
Screenshot 2024-05-31 at 11 11 38 AM

potatowagon (approved):

packages/lexical-playground/src/nodes/InlineImageNode/InlineImageComponent.tsx
58bits: Thanks @potatowago.  I'm actually the author of this plugin and thought I would tidy this up a little ;-)

Sahejkm:
Hi @58bits  , sorry for the inconvenience caused in the CI do you mind rebasing against Main before we try remerging this PR to main, to make sure all e2e tests are green. Thanks a lot!

58bits:
> Hi @58bits , sorry for the inconvenience caused in the CI do you mind merging latest Main to this branch before we try remerging this PR to main, to make sure all e2e tests are green. Thanks a lot!

Hi @Sahejkm  - okay sure. Done (I think).

Sahejkm:
CI pass apart from flaky collab tests which will be addressed in https://github.com/facebook/lexical/pull/6238

PR #6215, opened 2024-05-30
[lexical-website] Documentation Update: Validate stackblitz URLs in documentation

etrepum:
## Description

Sometimes we publish stackblitz URLs that point to a branch and don't change it back to main after review, which means examples can go stale or break if that branch is deleted.

To make PR review of stackblitz URLs easier, and to make the examples stable (relevant to #6144), this also adds docusaurus markdown preprocessor to rewrite these URLs on build to point to a specific commit (for PRs) or release branch.

### Before

stackblitz URLs could point to branches other than main

### After

Tests fail if a stackblitz URL does not point to main

Stackblitz URLs reference a specific commit for PRs so they can be tested in situ

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "eslint-plugin-sort-keys-fix": "^1.1.2",
     "flow-bin": "^0.226.0",
     "fs-extra": "^10.0.0",
-    "glob": "^7.2.0",
+    "glob": "^10.4.1",
     "google-closure-compiler": "^20220202.0.0",
     "gzip-size": "^6.0.0",
     "hermes-parser": "^0.20.1",
--- a/packages/lexical-website/docs/collaboration/react.md
+++ b/packages/lexical-website/docs/collaboration/react.md
@@ -100,7 +100,7 @@ function Editor() {
 
 Source code: [examples/react-rich-collab](https://github.com/facebook/lexical/tree/main/examples/react-rich-collab)
 
-<iframe width="100%" height="600" src="https://stackblitz.com/github/facebook/lexical/tree/fix/collab_example/examples/react-rich-collab?embed=1&file=src%2FApp.tsx&terminalHeight=0&ctl=1&showSidebar=0&devtoolsheight=0&view=preview" sandbox="allow-forms allow-modals allow-popups allow-popups-to-escape-sandbox allow-presentation allow-same-origin allow-scripts"></iframe>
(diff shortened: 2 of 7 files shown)


packages/lexical-website/docs/collaboration/react.md
```diff
@@ -100,7 +100,7 @@ function Editor() {
 
 Source code: [examples/react-rich-collab](https://github.com/facebook/lexical/tree/main/examples/react-rich-collab)
 
-<iframe width="100%" height="600" src="https://stackblitz.com/github/facebook/lexical/tree/fix/collab_example/examples/react-rich-collab?embed=1&file=src%2FApp.tsx&terminalHeight=0&ctl=1&showSidebar=0&devtoolsheight=0&view=preview" sandbox="allow-forms allow-modals allow-popups allow-popups-to-escape-sandbox allow-presentation allow-same-origin allow-scripts"></iframe>
+<iframe width="100%" height="600" src="https://stackblitz.com/github/facebook/lexical/tree/main/examples/react-rich-collab?embed=1&file=src%2FApp.tsx&terminalHeight=0&ctl=1&showSidebar=0&devtoolsheight=0&view=preview" sandbox="allow-forms allow-modals allow-popups allow-popups-to-escape-sandbox allow-presentation allow-same-origin allow-scripts"></iframe>
```
etrepum, line 103: This is an example of a case where we have left a non-main branch reference in an example (this isn't an isolated incident, I have fixed this class of inconsistency several times in various other PRs)

package.json
```diff
@@ -161,7 +161,7 @@
     "eslint-plugin-sort-keys-fix": "^1.1.2",
     "flow-bin": "^0.226.0",
     "fs-extra": "^10.0.0",
-    "glob": "^7.2.0",
+    "glob": "^10.4.1",
```
etrepum, line 164: I think this should be safe, if not we'll find out pretty quickly in tests. The upgrade is required to use an array of glob patterns, if it causes problems I will revert this part and use flatMap.

StyleT (dismissed):
Nice! Thanks for the fix

etrepum:
I went ahead and added the follow-up considerations to this PR to save some time in creating another one. I'm done now :)

potatowagon (approved):
thankyou for this

StyleT (approved):

PR #6216, opened 2024-05-30
[lexical-react] Bug Fix: checklist cannot be toggled in sub-editors

AlessioGr:
## Description

Checklists within sub-editors do not work. When toggling them, both the click event listeners for the sub-editor, as well as the one for the parent-editor, are triggered. Say the checklist item is untoggled, this would cause the checklist item to be toggled-on and then immediately toggled-off. Visually, it will look like nothing happened at all, as seen in the video.

Adding `event.stopPropagation()` stops the event from being incorrectly propagated up to the parent

### Before

https://github.com/facebook/lexical/assets/70709113/0fada556-a4de-4fb2-aef1-0d9b56f09bcc

### After

https://github.com/facebook/lexical/assets/70709113/a4b95553-7522-4cce-b9af-3de2aef876e9


potatowagon (approved):
👍 thanks!

Sahejkm:
hi @AlessioGr, sorry for the inconvenience caused in the CI do you mind rebasing against Main before we try remerging this PR to main, to make sure all e2e tests are green. Thanks a lot!

zurfyx:
@AlessioGr We have to revert this as it's breaking internally but feel free to post a follow up PR. While `event.stopPropagation()` might be correct for this case, it is currently being triggered for every single click on the editor, the scope needs to be revised.

Issue #6137, opened 2024-05-18
Feature: Change size-limit action to measure something more meaningful than the core modules


etrepum:
The size-limit workflow, as currently configured, has the following deficiencies:

* It only reports changes, it does not enforce any hard limits
* There is no documentation for what these limits should be, currently just vibes based(?) other than the intro docs which say "The core package of Lexical is only 22kb in file size (min+gzip)" which is no longer true, at least w/o further processing it's 30kb (see below)
* It only measures the prod+cjs builds of three modules: `lexical`, `@lexical/plain-text`, and `@lexical/rich-text`
* Notably this does not consider any relationship with dependencies that are meant to be external (yjs, y-websocket, react, react-dom, etc.), dependencies that are meant to be bundled (prism in `@lexical/code`), or anything that involves a significant code transform (other than dead-code elimination of `__DEV__`)

```
$ cat node_modules/lexical/Lexical.prod.js | gzip | wc -c
   30746
```

See also:
* #6132
* https://github.com/facebook/lexical/pull/6133#issuecomment-2118312061

etrepum:
It might also be worth thinking about removing the running time measurements because they are so incredibly noisy in this CI environment. See https://github.com/facebook/lexical/pull/6143#issuecomment-2120500275 which made no change to any code that size-limit can see (the www build is not used in this scenario) but is reporting >20% changes to runtime on two of the three modules. I think errors of this magnitude makes the numbers counter-productive because people may spend time investigating phantom performance regressions.

Sahejkm:
Hi @etrepum do you think we can close this issue right now ? Or do we need to have hard limits of failing the build once it cross a threshold ? Or any other package you think we should include in the size check before closing this ?

PR #6219, opened 2024-05-31
[Lexical][Size-Checks] Measure both cjs/esm builds for regression checks

Sahejkm:
## WHAT

follow up on #6137 

- Measure both cjs/esm build combinations for size checks
- Also ignore size-limit file from triggering extended tests run
## WHY
- Current regression checks meaure only cjs checks, enhance to check both cjs/esm changes.
- Not needed to run e2e tests for size-limit workflow updates

## Test
```
sahejk@sahejk-mbp lexical % npx size-limit
✔ Adding to empty webpack project
✔ Running JS in headless Chrome
  
  lexical - cjs
  Size:         22.5 kB with all dependencies, minified and brotlied
  Loading time: 440 ms  on slow 3G
  
  lexical - esm
  Size:         22.97 kB with all dependencies, minified and brotlied
  Loading time: 449 ms   on slow 3G
  
  @lexical/rich-text - cjs
  Size:         33.25 kB with all dependencies, minified and brotlied
  Loading time: 650 ms   on slow 3G
  
  @lexical/rich-text - esm
  Size:         23.05 kB with all dependencies, minified and brotlied
  Loading time: 451 ms   on slow 3G
  
  @lexical/plain-text - cjs
  Size:         33.21 kB with all dependencies, minified and brotlied
  Loading time: 649 ms   on slow 3G
  
  @lexical/plain-text - esm
  Size:         225 B with all dependencies, minified and brotlied
  Loading time: 10 ms on slow 3G
  
  @lexical/react - cjs
  Size:         150.4 kB with all dependencies, minified and brotlied
  Loading time: 3 s      on slow 3G
  
  @lexical/react - esm
  Size:         83.86 kB with all dependencies, minified and brotlied
  Loading time: 1.7 s    on slow 3G
  
sahejk@sahejk-mbp lexical % 
```

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -37,10 +37,11 @@ const getAliasType = (type) =>
     packagesManager.getPublicPackages().flatMap((pkg) =>
       pkg.getNormalizedNpmModuleExportEntries().map(([k, v]) => {
         switch (type) {
-          case 'development':
-            return [k, pkg.resolve('dist', v.require.development)];
-          case 'production':
-            return [k, pkg.resolve('dist', v.require.production)];
+          case 'esm':
+            return [
+              k,
+              pkg.resolve('dist', v.require.default.replace('.js', '.mjs')),
+            ];
           default:
             return [k, pkg.resolve('dist', v.require.default)];
         }
@@ -52,7 +53,7 @@ const modifyWebpackConfigForType = (config, alias) =>
   Object.assign(config, {resolve: {alias}});
 
 function sizeLimitConfig(pkg) {
-  return ['development', 'production'].map((type) => {
+  return ['cjs', 'esm'].map((type) => {
     const aliasType = getAliasType(type);
     return {
       path:
@@ -72,9 +73,9 @@ function sizeLimitConfig(pkg) {
 /**
  * These are the packages that were measured previously in #3600
  * We could consider adding more packages and/or also measuring
- * other build combinations such as esbuild/webpack, mjs/cjs etc.
+ * other build combinations such as esbuild/webpack.
  *
- * The current configuration measures only: webpack + cjs + dev/prod.
+ * The current configuration measures only: webpack + esm/cjs.
  *
  */
 module.exports = [
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -34,26 +34,20 @@ const path = require('node:path');
 const {packagesManager} = require('./scripts/shared/packagesManager');
 const getAliasType = (type) =>
   Object.fromEntries(
-    packagesManager.getPublicPackages().flatMap((pkg) =>
-      pkg.getNormalizedNpmModuleExportEntries().map(([k, v]) => {
-        switch (type) {
-          case 'esm':
-            return [
-              k,
-              pkg.resolve('dist', v.require.default.replace('.js', '.mjs')),
-            ];
-          default:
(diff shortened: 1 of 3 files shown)


etrepum:
Is there a reason to track dev size? It's not really clear what that would help with, since size is presumably not a primary concern for dev. It might be more useful to track the cjs and esm prod builds (probably not at Meta, but most other places are likely using esm or soon will be)

Sahejkm:
thanks for reviewing, Makes sense to me. Updated

.size-limit.js
```diff
@@ -32,47 +32,55 @@ const path = require('node:path');
...
  */
 const {packagesManager} = require('./scripts/shared/packagesManager');
-const alias = Object.fromEntries(
-  packagesManager
-    .getPublicPackages()
-    .flatMap((pkg) =>
-      pkg
-        .getNormalizedNpmModuleExportEntries()
-        .map(([k, v]) => [k, pkg.resolve('dist', v.require.default)]),
+const getAliasType = (type) =>
```
etrepum, line 35: If you use `import` and `require` as the possible values for type, you can simplify this to roughly the same as before using `pkg.resolve('dist', v[type].default)`

.size-limit.js
```diff
@@ -32,47 +32,55 @@ const path = require('node:path');
...
+  Object.fromEntries(
+    packagesManager.getPublicPackages().flatMap((pkg) =>
+      pkg.getNormalizedNpmModuleExportEntries().map(([k, v]) => {
+        switch (type) {
+          case 'esm':
+            return [
+              k,
+              pkg.resolve('dist', v.require.default.replace('.js', '.mjs')),
```
etrepum, line 43:
```suggestion
              pkg.resolve('dist', v.import.default),
```

.size-limit.js
```diff
@@ -32,47 +32,55 @@ const path = require('node:path');
...
 }
 
 /**
  * These are the packages that were measured previously in #3600
  * We could consider adding more packages and/or also measuring
- * other build combinations such as esbuild/webpack, mjs/cjs, dev/prod, etc.
+ * other build combinations such as esbuild/webpack.
  *
- * The current configuration measures only: webpack + cjs + prod.
+ * The current configuration measures only: webpack + esm/cjs.
```
etrepum, line 78: It's probably still worth noting that we are measuring the prod artifacts

.size-limit.js
```diff
@@ -32,47 +32,55 @@ const path = require('node:path');
...
+      modifyWebpackConfig: (config) =>
+        modifyWebpackConfigForType(config, aliasType),
+      running: false,
+      name: pkg + ' - ' + type,
+    };
+  });
 }
 
 /**
  * These are the packages that were measured previously in #3600
```
etrepum, line 74: This is no longer accurate since more packages are being measured now

etrepum (commented):
Looks like it would work, but could be a bit simpler by using the exports for lookup instead of manipulating the paths directly

Sahejkm:
Resolved comments, thanks for rereview!

potatowagon (approved):
lgtm

Sahejkm:
test failures look unrelated, will check separately. Seems to impact other builds as well like https://github.com/facebook/lexical/pull/6216/checks

PR #6220, opened 2024-05-31
[lexical-markdown] Bug Fix: standardise selection to move to start after markdown toggle 

### LINKED ISSUE #6177: Bug: `$convertFromMarkdownString` places caret at the end of the document (instead there should be a default way of pasting long markdown without scrol-downs due to caret at the end)
Even at https://playground.lexical.dev/ the caret on load is at the end of the document.

When a document is long, this forces overflow to happen (especially if a hyperlink is at the last line of the document, so a tooltip is rendered and overflow/scrolling is happening), and in general, on document load (from markdown as well) we expect caret to be at the top

In the meanwhile, would you have an advise how to work around this and force the cursor to not jump to the end?

Thanks!

image

potatowagon:
## Description

standardise the cursor behavior after toggling markdown, it is observed that `$convertToMarkdownString` moves cursor to the start, so make `$convertFromMarkdownString` follow suit. see https://github.com/facebook/lexical/pull/6220#discussion_r1627137045

also, executing @etrepum 's suggestion to add documentation to highlight that the selection (aka cursor) moves after the markdown operation, since it may not be an expected behavior

Closes #6177

### Before

see video in 

https://github.com/facebook/lexical/pull/6220/files#r1627137045

### After

https://github.com/facebook/lexical/assets/19604232/3871c3af-fc08-4515-9466-5ccd8ea7bc90


packages/lexical-playground/src/plugins/ActionsPlugin/index.tsx
```diff
@@ -189,7 +189,6 @@ export default function ActionsPlugin({
             $createCodeNode('markdown').append($createTextNode(markdown)),
           );
       }
-      root.selectEnd();
```
potatowagon, line 192: after removing this, we expose the selection behavior of `$convertFromMarkdownString` and `$convertToMarkdownString`.

`$convertToMarkdownString` moves the selector (aka cursor) to the start after the markdown -> string conversion.

in this PR adjust `$convertFromMarkdownString` to do the same. Also the attached issue has raised a preference for having the cursor at the start instead of the end

Sahejkm (approved):

PR #6224, opened 2024-05-31
Revert PR 1st causing some new added tests fail in collab mode and webkit mode

Sahejkm:
- Revert PR #6200 1st causing some new added tests fail in collab mode and webkit mode
- Sorry @AlexanderReznik for this we didn't add extended-tests label before merging this PR. Could you help fix the e2e tests in the PR and republish it ? Thank you !

- Example failing tests observed in https://github.com/facebook/lexical/actions/runs/9311543149/job/25630875667?pr=6216

## NOTE
Breaking tests are passing after revert but some test run builds are failing due to vite config error
```
> > vite build
failed to load config from D:\a\lexical\lexical\packages\lexical-playground\vite.config.ts
error during build:
Error: Missing package with directory name 'shared'
```


potatowagon (approved):

potatowagon:
maybe it would be faster to raise a pr fixing the test? lemme see if i can raise something tonight

potatowagon (changes requested):
PR6228 fixes the test so we can close this

etrepum:
I think there are still test failures in #6228 but we would want to check this after #6227 (or merge that code in here) to make sure that the combination covers all of the test regressions

StyleT (approved):

StyleT:
> PR6228 fixes the test so we can close this

#6228 still fails tests after rebase on top of https://github.com/facebook/lexical/pull/6227 - merging this

PR #6227, opened 2024-05-31
[*] Fix: Resolve Windows regression caused by upgrading glob

etrepum:
## Description

Upgrading glob in #6215 changed its behavior on Windows, since it now expects you to specify paths that use '/' as the path separator so that '\\' can be used for escaping glob characters unless the windowsPathsNoEscape option is specified which disables escaping of glob characters (a feature we do not use).

Hopefully #6198 works and lands soon so that this type of regression can be avoided in the future

## Test Plan

Please add the extended-tests label to this PR so that we can confirm that it works before merging

### Before

Windows e2e tests failed

### After

Windows e2e tests should succeed

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,6 +3,7 @@
 **/dist/**
 **/build/**
 **/npm/**
+!scripts/npm/**
 **/.output/**
 **/.browser-profiles/**
 **/__tests__/integration/fixtures/**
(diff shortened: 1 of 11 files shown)


StyleT (approved):

etrepum:
One of the failures was unrelated to #6228, https://github.com/facebook/lexical/pull/6227/commits/db1eedd1bc70715f9611634b1c032c981399ccc3 should fix it

etrepum:
I believe that the only remaining test failures will be fixed by #6228, the glob related issues should be solved

StyleT (approved):

PR #6228, opened 2024-05-31
[lexical-playground] Bug Fix: Fix tests

AlexanderReznik:
## Description
Fixing tests from https://github.com/facebook/lexical/pull/6200

I have them passing locally in WSL so it's a bit of a guess, but I'm unifying how `click` is made

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -1622,7 +1622,7 @@ test.describe('Tables', () => {
       false,
     );
     await mergeTableCells(page);
-    await page.locator('th').first().click();
+    await click(page, 'th');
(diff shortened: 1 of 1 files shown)


potatowagon:
thanks for the quick fix

etrepum:
I think we should probably add extended-tests to #6227 first so that the windows e2e suite runs correctly again, update this PR with main after that lands to bring in the fix, and then add the extended-tests label to this PR so we can see the full suite running and confirm that the fix works cross-platform

StyleT (approved):

etrepum:
Based on the mac failures I don't think this is resolved just yet, it appears to be slight differences in size metrics (height) https://github.com/facebook/lexical/actions/runs/9322094913/job/25662413222?pr=6228

StyleT:
Rejected in favor of #6224 

Pls fix tests and re-submit your changes.  🙌

PR #6230, opened 2024-05-31
[lexical-playground] Fix bug: Insert an image inside another image's caption (#6109)

### LINKED ISSUE #6109: Bug: Insert an image inside another image's caption
Currently, a user can insert an image inside another image's caption which may cover up or deface the original image.
Probably a variety of elements that can be inserted in an image's caption should be limited.

Lexical version:

## Steps To Reproduce

1. Open https://playground.lexical.dev/ in Chrome
2. Insert a sample image
3. Click on the image caption
4. Go to the Toolbar and insert "Image"
5. Select from "URL".
6. Insert a new image URI.
7. See that the new image covers the original one.

## The current behavior

https://github.com/facebook/lexical/assets/5062807/efb8e972-6d74-4f86-b2b1-d61cd4ed3794

## The expected behavior
A user is not able to insert an image into the caption.

MaxPlav:
Context:
Now users are able to insert an image/GIF/table etc, into another image's caption. 

Solution:
This PR limits what the users can insert into the image caption and its nested editor. Excluding some controls from the toolbar if the selection is inside the nested editor that belongs to the image caption.

Closes #6109

## Test plan
2. Insert a sample image with a caption.
3. Focus on the caption.
4. Insert a sample image/table/GIF etc.

### Before

https://github.com/facebook/lexical/assets/5062807/5e0ad65e-00ee-4fbf-9f2c-22359f37a7f2

### After

https://github.com/facebook/lexical/assets/5062807/63b175ec-976e-4ab0-b05f-3ec18be8b93e

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
@@ -43,6 +43,7 @@ import {
   $findMatchingParent,
   $getNearestBlockElementAncestorOrThrow,
   $getNearestNodeOfType,
+  $isEditorIsNestedEditor,
   mergeRegister,
 } from '@lexical/utils';
 import {
@@ -545,10 +546,22 @@ export default function ToolbarPlugin({
   const [isRTL, setIsRTL] = useState(false);
   const [codeLanguage, setCodeLanguage] = useState<string>('');
   const [isEditable, setIsEditable] = useState(() => editor.isEditable());
+  const [isImageCaption, setIsImageCaption] = useState(false);
 
   const $updateToolbar = useCallback(() => {
     const selection = $getSelection();
     if ($isRangeSelection(selection)) {
+      if (activeEditor !== editor && $isEditorIsNestedEditor(activeEditor)) {
+        const rootElement = activeEditor.getRootElement();
+        setIsImageCaption(
+          !!rootElement?.parentElement?.classList.contains(
+            'image-caption-container',
+          ),
+        );
+      } else {
+        setIsImageCaption(false);
+      }
+
       const anchorNode = selection.anchor.getNode();
       let element =
         anchorNode.getKey() === 'root'
@@ -656,19 +669,25 @@ export default function ToolbarPlugin({
         $getSelectionStyleValueForProperty(selection, 'font-size', '15px'),
       );
     }
-  }, [activeEditor]);
+  }, [activeEditor, editor]);
 
   useEffect(() => {
     return editor.registerCommand(
       SELECTION_CHANGE_COMMAND,
       (_payload, newEditor) => {
-        $updateToolbar();
         setActiveEditor(newEditor);
+        $updateToolbar();
         return false;
       },
       COMMAND_PRIORITY_CRITICAL,
     );
   }, [editor, $updateToolbar]);
+
+  useEffect(() => {
+    activeEditor.getEditorState().read(() => {
+      $updateToolbar();
+    });
+  }, [activeEditor, $updateToolbar]);
 
   useEffect(() => {
     return mergeRegister(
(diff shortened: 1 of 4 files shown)


MaxPlav:
update: the image caption nested editor uses PlainTextPlugin so it can paste the content as plain text - similar behavior as for a Sticky Note node. Fixed the issue so that an image/GIF/table etc. can be pasted to an image caption.

MaxPlav:
update: resolved an issue with non-functional text formatting and alignment in the image caption nested editor.

https://github.com/facebook/lexical/assets/5062807/ca4a8d5c-cd35-4ea0-912b-ed662e06aa53

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -545,125 +546,141 @@ export default function ToolbarPlugin({
...
+    [editor],
+  );
 
   useEffect(() => {
     return editor.registerCommand(
       SELECTION_CHANGE_COMMAND,
       (_payload, newEditor) => {
-        $updateToolbar();
         setActiveEditor(newEditor);
+        $updateToolbar(newEditor);
```
StyleT, line 683: I think you don't need to pass `newEditor` to `$updateToolbar()` as you can grab it from `activeEditor` var from the component scope

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -545,125 +546,141 @@ export default function ToolbarPlugin({
...
-          ? anchorNode
-          : $findMatchingParent(anchorNode, (e) => {
-              const parent = e.getParent();
-              return parent !== null && $isRootOrShadowRoot(parent);
-            });
+  const $updateToolbar = useCallback(
+    (newEditor: LexicalEditor) => {
+      const selection = $getSelection();
+      if ($isRangeSelection(selection)) {
+        if (newEditor !== editor && $isEditorIsNestedEditor(newEditor)) {
```
StyleT, line 555:
```suggestion
        if (activeEditor !== editor && $isEditorIsNestedEditor(activeEditor)) {
```

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1691,3 +1691,14 @@ export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
 export function $getEditor(): LexicalEditor {
   return getActiveEditor();
 }
+
+/**
+ * Checks if the editor is a nested editor created by LexicalNestedComposer
+ */
+export function $isEditorIsNestedEditor(editor: LexicalEditor): boolean {
```
StyleT, line 1698: Instead of having this function you can simply check if `editor._parentEditor` is non `null` :) 

Although maybe we still can have helper...

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
MaxPlav: not sure why but the function has a stale `activeEditor` value sometimes, so passing as an argument

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -74,6 +74,7 @@ import {Dispatch, useCallback, useEffect, useState} from 'react';
 import * as React from 'react';
 import {IS_APPLE} from 'shared/environment';
 
+import {$isEditorIsNestedEditor} from '../../../../lexical/src/LexicalUtils';
```
StyleT, line 77: You can't really import non-exported file from `lexical` package here

But you can move it to `@lexical/utils`

StyleT (approved):
Excellent!

PR #6232, opened 2024-05-31
Fix programmatic clipboard copy on non-editable

### LINKED ISSUE #6231: Programmatic copy fails on non-editable
Any copy that's triggered programmaticaly (not via a browser event like cmd + c) is failing on non-editable mode.

For reference, the context menu triggers it programmatically https://github.com/facebookexternal/Outline//blob/3e117f8fff6d35efcb0f63d4832f76fb15dbcfed/packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx#L110-L114

```
new ContextMenuOption(`Copy`, {
        onSelect: (_node) => {
          editor.dispatchCommand(COPY_COMMAND, null);
        },
      }),
```

https://github.com/facebook/lexical/assets/193447/6982db20-a9b1-43c0-ba60-02365a69abd9

zurfyx:
Clipboard copy uses a double execCommand hack. It turns out that we weren't listening for the copy event on non-editable. This PR allowlists this event which will subsequently dispatch the command after, I think this list should be revised after, potentially adding `focus` and `blur` but will do for now.

https://github.com/facebook/lexical/assets/193447/381fdb47-8e27-42d5-900f-ea55ae793ab2

Closes #6231


packages/lexical/src/LexicalEvents.ts
```diff
@@ -1241,72 +1241,68 @@ export function addRootElementEvents(
...
-                    COPY_COMMAND,
-                    event as ClipboardEvent,
-                  );
-
-                case 'paste':
-                  return dispatchCommand(
+            const isEditable = editor.isEditable();
+            switch (eventName) {
+              case 'cut':
+                return dispatchCommand(
```
etrepum, line 1247: Shouldn't cut also be conditional when `isEditable`?

zurfyx: @etrepum  You're right 🤦‍♂️

etrepum:
Looks like the current failure is prettier for packages/lexical/src/LexicalEvents.ts

packages/lexical/src/LexicalEvents.ts
```diff
@@ -1241,72 +1241,71 @@ export function addRootElementEvents(
               return;
             }
             stopLexicalPropagation(event);
-            if (editor.isEditable()) {
-              switch (eventName) {
-                case 'cut':
-                  return dispatchCommand(
+            const isEditable = editor.isEditable();
+            switch (eventName) {
```
etrepum, line 1245: Do you think this would be easier to maintain as a table, e.g. 

[ts playground example](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcAyoCWBjAhgGwGEIBbE3AOwBMAeAFQAVdF8JcqA+OAXjgG8AUHATJgAfgBccAM4womCgHMA3AIC+q0JFhwAZgFcK2GJggU42KMFwxgxMpVp0OACiQpJMuQsUBKfkJwVjD6UOZ87sAa6gJa0PDYZrJwhACqdAD6hADyALK5AIIAcgAiPBZWNnak5NQ0gYT4mGAARmxQVACiAG7AFPAAPnAA0sCIbbgdPX2DcBT6+PgCrgBEaZk5+cUlK76a4PEWSfA5DACaWXmFpeWW1rb2tbQNTa3tXb39cEOj4+-TXyG80Wyxca2y50uW1Ku1UAgUtigulw2FQGBAOAInSomBg0ACwkw0mxuNwLXwwBcvikbQgFMoqjUAgMRhMZjgOOkYBs2AAFo9HPRXMAcXioFJ0Zj8CSxQAaI4OagSrB4Ig1QXOeXclhsKhSOj+QRMgSJCjJABKnQKJWyRTQF06ADVOkU6ABlcqCYSJMCIKSnC6ba4lWXqOC4aRHM0wVSm5KdEoASToBQAQmhOhknS73Z7AgA6QuW622+1Z52ut2h736GD+9JQ4OhtThyNxmMCTtxHQs4ymcyc7kwPkAmAuEW46DKjGqmXQeXAT4wIq4EjAKSyeRKBdLqSjw2BdtwRczXK4MDlCdi-NEmVkilUuBiOAJ5NpjPlnMeqTFm12h0Vu6yhwIexwKk85TjkuZ4XhGcDmtYVBmPgiA0AhiQdDQm4+PKhhUMAugKCK3zoCqBACnUo4cNRvgANonv0K5rgAuqohK6HALiJIqVAHsIwiDjy-LqtQ46ivO4GODuMx7IETJMkAA)
```typescript
const READONLY_EVENTS = {
  copy: COPY_COMMAND,
  // more commands here
} as const;
const EDITABLE_EVENTS = {
  ...READONLY_EVENTS,
  cut: CUT_COMMAND,
  // more commands here
} as const;

export function dispatchEvent(editor: LexicalEditor, eventName: string, event: Event) {
  const eventMap = editor.isEditable() ? EDITABLE_EVENTS : READONLY_EVENTS; 
  const command = (eventMap as Readonly<Record<string, undefined | LexicalCommand<Event>>>)[eventName];
  if (command) {
    dispatchCommand(editor, command, event);
  }
}
```

ivailop7 (approved):

PR #6209, opened 2024-05-30
[Lexical][CI] Add more packages for size limit regressions check

Sahejkm:
## WHAT

Add @lexical/react package for size limit regressions check, follows up on #6137

## WHY

- While supporting changes to making Lexical compatible to React 19, there was a size regression in the Meta internal build but wasn't detected in the external build.

- adding @lexical/react package to size limit check to capture the size limit changes on each PR

## TEST PLAN

```
sahejk@sahejk-mbp lexical % npx size-limit                           
✔ Adding to empty webpack project
✔ Running JS in headless Chrome
  
  lexical
  Size:         22.5 kB with all dependencies, minified and brotlied
  Loading time: 440 ms  on slow 3G
  
  @lexical/rich-text
  Size:         33.25 kB with all dependencies, minified and brotlied
  Loading time: 650 ms   on slow 3G
  
  @lexical/plain-text
  Size:         33.21 kB with all dependencies, minified and brotlied
  Loading time: 649 ms   on slow 3G
  
  @lexical/react
  Size:         150.38 kB with all dependencies, minified and brotlied
  Loading time: 3 s       on slow 3G
  
sahejk@sahejk-mbp lexical % 

```

PR #6234, opened 2024-06-01
[*] Fix: size-limit report for esm modules

### LINKED ISSUE #6233: Bug: [*] size-limit report for @lexical/plain-text esm seems wrong
## Steps To Reproduce

1. Create a PR
2. See a size-limit measurement for @lexical/plain-text of 225 bytes (e.g. https://github.com/facebook/lexical/pull/6232#issuecomment-2142963336)

## The current behavior

https://github.com/facebook/lexical/pull/6232#issuecomment-2142963336

```
@lexical/plain-text - esm | 225 B (0%)
```

## The expected behavior

A larger number.

I suspect that it's measuring the fork module or something like that, although with build-prod there shouldn't be a fork module so more investigation is necessary.

etrepum:
## Description

For some reason the default configuration of size-limit for @lexical/plain-text caused nearly the whole module to be tree shaken, so nothing of interest was measured. This makes sure to test the full size of every package we are measuring.

Closes #6233

### Before

size-limit would report a 225 bytes for @lexical/plain-text esm

```
  @lexical/plain-text - esm
  Size:         225 B with all dependencies, minified and brotlied
  Loading time: 10 ms on slow 3G
```

### After

size-limit should report a meaningful number for @lexical/plain-text esm

```
  @lexical/plain-text - esm
  Size:         24.44 kB with all dependencies, minified and brotlied
  Loading time: 478 ms   on slow 3G
```

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -50,6 +50,7 @@ function sizeLimitConfig(pkg) {
   return ['require', 'import'].map((type) => {
     const aliasType = getAliasType(type);
     return {
+      import: '*',
       path:
         aliasType[pkg] != null
           ? aliasType[pkg]


etrepum:
The github action for size-limit does some wonky stuff to base it from main for comparison, but I did confirm that it is doing the right thing locally. You can see this in other size-limit related PRs such as #6219 and #6209 and see that their PR comments are not based on the incoming config.

Sahejkm (approved):

PR #6238, opened 2024-06-03
[lexical-playground] CI: skip flaky collab e2e test 

### LINKED ISSUE #6241: Stabilise CI: resolve flaky e2e tests

potatowagon:
## Description
eg. https://github.com/facebook/lexical/actions/runs/9322462619/job/25663598465

^ flaky collab tests failing on ci signal

disable flaky collab tests so we dont waste resources running them + reduce noise in the ci

Closes #6241

## Test plan
```
npm run test-e2e-collab-chromium   
```

run this until all flaky test skipped

>  Consider splitting slow test files to speed up parallel execution
  73 skipped
  361 passed (3.7m)

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
@@ -1963,7 +1963,7 @@ test.describe('Links', () => {
     page,
     isCollab,
   }) => {
-    test.fixme(isCollab && IS_MAC && IS_LINUX, 'Flaky on Collab, Mac + Linux');
+    test.fixme(isCollab && (IS_MAC || IS_LINUX), 'Flaky on Collab, Mac/Linux');
     await focusEditor(page);
     await page.keyboard.type('Hello ');
     await toggleBold(page);
--- a/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
@@ -25,6 +25,7 @@ import {
   html,
   initialize,
   IS_LINUX,
+  IS_MAC,
   keyDownCtrlOrMeta,
   keyUpCtrlOrMeta,
   pasteFromClipboard,
@@ -428,7 +429,9 @@ test.describe('Links', () => {
 
   test(`Can create a link with some text after, insert paragraph, then backspace, it should merge correctly`, async ({
     page,
+    isCollab,
   }) => {
+    test.fixme(isCollab, 'Flaky on Collab');
     await focusEditor(page);
     await page.keyboard.type(' abc def ');
     await moveLeft(page, 5);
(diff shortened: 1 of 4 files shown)


packages/lexical-playground/__tests__/e2e/Links.spec.mjs
```diff
@@ -1960,7 +1963,7 @@ test.describe('Links', () => {
     page,
     isCollab,
   }) => {
-    test.fixme(isCollab && IS_LINUX, 'Flaky on Linux + Collab');
+    test.fixme(isCollab && IS_MAC && IS_LINUX, 'Flaky on Collab, Mac + Linux');
```
potatowagon, line 1966: will see if this is flaky on windows, if yes will disable for all envs

Sahejkm: this should be ```IS_MAC || IS_LINUX``  ?

Sahejkm: i see 1 more flaky test in the file from https://github.com/facebook/lexical/actions/runs/9348305142/job/25727177122?pr=6214 will fix that as well :(

Sahejkm: should this be && or || ?

potatowagon: > should this be && or || ?

Oh yea should be ||

Sahejkm:
would like to check how we are marking so many tests as flaky, in the past 5-6 PR merges i ran i only see 1 test out of these fail 1 time, marking too many tests as flaky might be risky

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -474,8 +474,13 @@ test.describe('Selection', () => {
...
   });
 
-  test('Can delete sibling elements forward', async ({page, isPlainText}) => {
+  test('Can delete sibling elements forward', async ({
+    page,
+    isPlainText,
+    isCollab,
+  }) => {
     test.skip(isPlainText);
+    test.fixme(isCollab, 'Flaky on Collab');
```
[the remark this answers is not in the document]
etrepum: flaky tests currently cause failures in CI if they fail 4 times in a row. I think the chances of this happening for an individual test are pretty rare but the more flaky tests there are then the more likely a CI failure is.

For example, if there's a 25% chance of a flaky test failing then there's roughly an 0.4% chance that the particular test will cause a CI failure. If we have 10 of those flaky tests (with the same failure rate), then nearly 4% of CI runs will fail.

It would make sense to at least identify them with comments when a test is suspected to be flaky so that the first course of action would be to re-run the tests rather than try and debug something that has no relevance to the PR under test.

PR #6242, opened 2024-06-03
[test][Lexical][CI] Increase workers to 6 to improve run time

Sahejkm:
## What
Increase workers to 6 to improve run time

## Why
- locally doing this does give a performance boost in overall run time for firefox tests


Sahejkm:
closing this as increasing to 6 doesn't see impact in overall usage on CI, some tests also fail tho on CI on increasing the ||ism, might wanna check again if those are flaky tests

PR #6245, opened 2024-06-04
[lexical-playground] CI: fix flaky headings e2e test

### LINKED ISSUE #6241: Stabilise CI: resolve flaky e2e tests

potatowagon:
## Description
test failing at random because enter key is pressed too soon after the last move right key press. add a 500 ms delay to fix that

Closes #6241

## Test plan

 npm run start & npm run test-e2e-collab-chromium  

### Before

```
    - Expected  - 2
    + Received  + 2

      <h1 class="PlaygroundEditorTheme__h1 PlaygroundEditorTheme__ltr" dir="ltr">
    -   <span data-lexical-text="true">Welco</span>
    +   <span data-lexical-text="true">Welc</span>
      </h1>
      <h1 class="PlaygroundEditorTheme__h1 PlaygroundEditorTheme__ltr" dir="ltr">
    -   <span data-lexical-text="true">me to the playground</span>
    +   <span data-lexical-text="true">ome to the playground</span>
      </h1>

  1 failed
    [chromium] › packages/lexical-playground/__tests__/e2e/Headings/HeadingsEnterAtEnd.spec.mjs:22:6 › Headings - stays as a heading when you press enter in the middle of a heading 
```

### After

  ✓  1 …adings/HeadingsEnterAtEnd.spec.mjs:22:6 › Headings - stays as a heading when you press enter in the middle of a heading (4.4s)

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Headings/HeadingsEnterAtEnd.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Headings/HeadingsEnterAtEnd.spec.mjs
@@ -46,7 +46,7 @@ test(`Headings - stays as a heading when you press enter in the middle of a head
 
   await moveToEditorBeginning(page);
 
-  await moveRight(page, 5);
+  await moveRight(page, 5, 500);
 
   await page.keyboard.press('Enter');
 


packages/lexical-playground/__tests__/e2e/Headings/HeadingsEnterAtEnd.spec.mjs
```diff
@@ -46,7 +46,7 @@ test(`Headings - stays as a heading when you press enter in the middle of a head
 
   await moveToEditorBeginning(page);
 
-  await moveRight(page, 5);
+  await moveRight(page, 5, 500);
```
potatowagon, line 49: some thoughts: 

i considered adding a default delay to moveRight fn defintion instead but that might slow down all e2e tests calling moveRight so did not do that

fantactuka (approved):

packages/lexical-playground/__tests__/e2e/Headings/HeadingsEnterAtEnd.spec.mjs
Sahejkm: is 500ms choosen randomly or after testing ? wanted to make sure it doesn;t make it too slow ?

Other Example using a value 100ms : https://github.com/facebook/lexical/blob/2e59e48dbbf1872e8dfb2f1a0ff62ef3ee428fde/packages/lexical-playground/__tests__/e2e/Links.spec.mjs#L1129

Sahejkm: - i guess this will be followed up in #6246 , 
- would also like to get more data about this:  the flakiness of this test in the past real CI runs, time saving we get in fixing flakynees by adding sleeps vs actual retries

potatowagon: @Sahejkm 

>  the flakiness of this test in the past real CI runs

im not too sure whats an efficient way to get this value except for combing through past PRs and checking on the failure signal one by one. it sounds like a painful process. u did bring up a nice point on exploring playright reports. ill explore that to see if it can give this value for future runs

> time saving we get in fixing flakynees by adding sleeps vs actual retries

after https://github.com/facebook/lexical/pull/6246, the test run takes 2.5s. in worse case scenerio where all 4 playwright retries fail, time taken is >10s. so potentially >7.5s for every CI run

zurfyx:
@potatowagon LGTM! Do you know if it's a generalized issue? We might want to consider a value between 200-500ms as a default

potatowagon:
> @potatowagon LGTM! Do you know if it's a generalized issue? We might want to consider a value between 200-500ms as a default

i have to experiment to confirm if adding general delays will help reduce flakiness in general

PR #6257, opened 2024-06-05
[lexical-playground] CI: fix flaky MaxLength emoji e2e test

potatowagon:
## Description
`&nbsp;` appears randomly thus making test flaky, add keypress delays when typing to prevent that

## Test plan

add `--reporter=html --repeat-each=10` to "test-e2e-collab-firefox": "cross-env E2E_BROWSER=firefox E2E_EDITOR_MODE=rich-text-with-collab playwright test --project=\"firefox\"", in build script

then run `npm run start & npm run test-e2e-collab-firefox`

### Before

Screenshot 2024-06-05 at 6 29 26 PM

### After
```
npm run start & npm run cross-env E2E_BROWSER=firefox E2E_EDITOR_MODE=rich-text-with-collab playwright test --project="firefox" --reporter=html --repeat-each=10


> @lexical/monorepo@0.16.0 test-e2e-collab-firefox
> cross-env E2E_BROWSER=firefox E2E_EDITOR_MODE=rich-text-with-collab playwright test --project="firefox" --reporter=html --repeat-each=10


> @lexical/monorepo@0.16.0 start
> cross-env NODE_ENV=development concurrently "npm:collab" "npm run dev --prefix packages/lexical-playground"

[collab] 
[collab] > @lexical/monorepo@0.16.0 collab
[collab] > cross-env HOST=localhost PORT=1234 npx y-websocket-server
[collab] 
[1] 
[1] > lexical-playground@0.16.0 dev
[1] > vite --host --port 3000
[1] 

Running 10 tests using 4 workers
[4/10] [firefox] › packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs:77:8 › MaxLength › can restrict emojis on boundaries
[1] 
[1]   VITE v5.2.11  ready in 620 ms
[1] 
[1]   ➜  Local:   http://localhost:3000/
[1]   ➜  Network: http://172.19.82.250:3000/
  Slow test file: [firefox] › packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs (53.0s)
  Consider splitting slow test files to speed up parallel execution
  10 passed (19.9s)

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
@@ -79,7 +79,7 @@ test.describe('MaxLength', () => {
     await pasteFromClipboard(page, {
       'text/plain': 'lorem ipsum dolor sit amet, consectetur adipiscing elit',
     });
-    await pressBackspace(page, 1, STANDARD_KEYPRESS_DELAY_MS);
+    await pressBackspace(page, 1);
 
     await assertHTML(
       page,
@@ -105,7 +105,7 @@ test.describe('MaxLength', () => {
       `,
     );
 
-    await pressBackspace(page, 1, STANDARD_KEYPRESS_DELAY_MS);
+    await pressBackspace(page, 1);
     await page.keyboard.type('💏');
 
     await assertHTML(
--- a/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
@@ -79,7 +79,7 @@ test.describe('MaxLength', () => {
     await pasteFromClipboard(page, {
       'text/plain': 'lorem ipsum dolor sit amet, consectetur adipiscing elit',
     });
-    await pressBackspace(page, 1);
+    await pressBackspace(page);
 
     await assertHTML(
       page,
@@ -105,7 +105,7 @@ test.describe('MaxLength', () => {
       `,
     );
 
-    await pressBackspace(page, 1);
+    await pressBackspace(page);
     await page.keyboard.type('💏');
 
     await assertHTML(
--- a/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
@@ -7,6 +7,10 @@
  */
 
 import {
+  pressBackspace,
+  STANDARD_KEYPRESS_DELAY_MS,
+} from '../keyboardShortcuts/index.mjs';
+import {
   assertHTML,
   assertSelection,
   clearEditor,
@@ -75,7 +79,7 @@ test.describe('MaxLength', () => {
     await pasteFromClipboard(page, {
       'text/plain': 'lorem ipsum dolor sit amet, consectetur adipiscing elit',
     });
-    await page.keyboard.press('Backspace');
+    await pressBackspace(page, 1, STANDARD_KEYPRESS_DELAY_MS);
 
     await assertHTML(
       page,
@@ -88,7 +92,7 @@ test.describe('MaxLength', () => {
       `,
     );
 
-    await page.keyboard.type('💏');
+    await page.keyboard.type('💏', {delay: STANDARD_KEYPRESS_DELAY_MS});
 
     await assertHTML(
       page,
@@ -101,7 +105,7 @@ test.describe('MaxLength', () => {
       `,
     );
 
-    await page.keyboard.press('Backspace');
+    await pressBackspace(page, 1, STANDARD_KEYPRESS_DELAY_MS);
     await page.keyboard.type('💏');
 
     await assertHTML(

### CODE REVIEW

File: packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs [hunk 1]
```diff
@@ -88,7 +92,7 @@ test.describe('MaxLength', () => {
       `,
     );
 
-    await page.keyboard.type('💏');
+    await page.keyboard.type('💏', {delay: STANDARD_KEYPRESS_DELAY_MS});
```
Sahejkm, line 95: also is this test failing in CI in past runs for all environments ? or we wanna add delays based on environments where it fails like firefox or iscollab etc

potatowagon:
followup thoughts: 
to scale solving flakiness due to lack of keypress delays
- could look into setting the standard kepress delay as the default delay after a set of consecutive keypresses
- as long as the total delay added per test on avg does not cross 2 s
- seems feasible since each tests dont usually have more then 10 typing actions
- adding a default delay after a set of consecutive key presses wouldnt add more then 1s per test given a standard delay of 100ms

packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
potatowagon: feels abit overkill to guard the delays based on enviroment given at most a total of 300ms (1 third of a second) js added.

the benefit is the test is stable across all enviroments

Sahejkm: i just feel our e2e test suite will grow and scale overtime, we should try to add whats needed, overusing sleeps is also not great experience, making e2e test suite slow. unless this test is flaky over all environments and needs sleeps

Sahejkm: Also, if we see the run https://github.com/facebook/lexical/actions/runs/9364683650/job/25778198911?pr=6246 for #6246 it actually shows Heading test is still flaky after the fix. How are we verifying the fix ? As its not consistent and still appears after this change

Screenshot 2024-06-05 at 7 33 45 PM

potatowagon: thanks for pointing out a still flaky test result. ill investigate more tmr. today i discovered the repeat param which allows me to stress test to make sure the test is no longer flaky.

also, u have a valid point to reduce unnecessary sleeps so that devx and cost is not affected. i plan to experiment how much total delay would be added for x amount of default delay, and see if such a value is an acceptable value interms of devx and cost. and also, the projectes value if our test suite grows

potatowagon: for this test, it is consistenly flaky on my machine when i run the suite of firefix collab test. when ran isolated, it is also consistently flaky. after this fix and stress testing it 10 times, it passes consistently on all 10 runs

etrepum:
I wonder if there is an approach we could use where we attach an updateListener to the editor under test and wait for the update to be reconciled (with a timeout of 100ms or whatever the constant should be) after a keypress instead of always sleeping 100ms. Of course it wouldn't generalize to all keypresses but it might help remove a lot of sleeps and let us both speed up the tests and make them more reliable.

Sahejkm:
+1 to Bob's point, if we could build a waitFor() kinda API which waits for the necessary step to complete with some Max timeout to Wait, it would be more reliable over sleeps which add constant delays without any underlying reason for the delays , was reading this yesterday

potatowagon:
> I wonder if there is an approach we could use where we attach an updateListener to the editor under test and wait for the update to be reconciled (with a timeout of 100ms or whatever the constant should be) after a keypress instead of always sleeping 100ms. Of course it wouldn't generalize to all keypresses but it might help remove a lot of sleeps and let us both speed up the tests and make them more reliable.

thanks @etrepum , that sounds like a good idea! will give it a try

packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
```diff
@@ -75,7 +79,7 @@ test.describe('MaxLength', () => {
     await pasteFromClipboard(page, {
       'text/plain': 'lorem ipsum dolor sit amet, consectetur adipiscing elit',
     });
-    await page.keyboard.press('Backspace');
+    await pressBackspace(page, 1, STANDARD_KEYPRESS_DELAY_MS);
```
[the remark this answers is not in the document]
potatowagon: strangely today i wasnt able to repro the failures on my machine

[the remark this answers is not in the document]
potatowagon: ah ok if i run the whole test suit, able to repro it again

packages/lexical-playground/__tests__/e2e/MaxLength.spec.mjs
```diff
@@ -75,7 +79,7 @@ test.describe('MaxLength', () => {
     await pasteFromClipboard(page, {
       'text/plain': 'lorem ipsum dolor sit amet, consectetur adipiscing elit',
     });
-    await page.keyboard.press('Backspace');
+    await pressBackspace(page, 1);
```
Sahejkm, line 82: default is 1 right ? adding this again gives any value ?

Sahejkm (approved):

potatowagon:
tests failing due to falky test

PR #6259, opened 2024-06-05
[lexical-markdown] Add test to keep code language

Minhir:
## [lexical-markdown] Bug Fix: preserve language tag for Markdown code

Right now, PR contains only one commit with tests to highlight the problem.  

If the markdown file contains a code block with an unsupported language (a language for which the project doesn't have syntax highlighter) the lexical will remove any information about that language. 

Let's say we have markdown with the following content:

~~~
```rust
fn main() {
    println!("Hello World!");
}
```
~~~

If we don't have `prism-rust` in the project, lexical will strip all the info about the language and saved file will look like:

~~~
```
fn main() {
    println!("Hello World!");
}
```
~~~

That behavior seems to me like a bug. 

## Possible solution

I think `CodeNode` should always keep `__language` field, no matter what the `mapToPrismLanguage()` returns. And we can create another field called `__isSyntaxHighlightSupported` for highlighting check. 

What do you think? I'll be happy to create another commit with a fix if provided solution works for the lexical team.


StyleT:
Makes sense! Commit with the fix would really help here :P

packages/lexical-code/src/CodeNode.ts
```diff
@@ -48,13 +48,11 @@ export type SerializedCodeNode = Spread<
...
-const mapToPrismLanguage = (
+const isLanguageSupportedByPrism = (
   language: string | null | undefined,
-): string | null | undefined => {
+): boolean => {
   // eslint-disable-next-line no-prototype-builtins
-  return language != null && window.Prism.languages.hasOwnProperty(language)
-    ? language
-    : undefined;
+  return language ? window.Prism.languages.hasOwnProperty(language) : false;
```
mayneyao, line 55:
```js
try {
    return language ? window.Prism.languages.hasOwnProperty(language) : false;
} catch (error) {
    return false
}
```

this may help the headless editor work in a Web Worker

Minhir: Good point!

StyleT (approved):

PR #6265, opened 2024-06-07
[Lexical][CI] Skip extended tests workflow run on multiple approvals

Sahejkm:
## WHAT
Followup on #6198 , to skip extended tests workflow run on multiple approvals

## WHY
Running e2e tests twice on approved PRs doesn't make sense and add to the overal CI cost, hence skip duplicate workflow run

### DIFF
--- a/.github/workflows/after-approval.yml
+++ b/.github/workflows/after-approval.yml
@@ -18,14 +18,10 @@ jobs:
           do_not_skip: '["pull_request", "merge_group"]'
   e2e-tests:
     needs: pre_job
-    if: needs.pre_job.outputs.should_skip != 'true' && (github.event.review.state == 'commented' && !contains(github.event.pull_request.labels.*.name, 'extended-tests'))
-    runs-on: ubuntu-latest
-    steps:
-      - name: Tesing workflow executed or skipped ${{needs.pre_job.outputs.should_skip}}
-        run: |
-          echo "Test executed ...  ${{needs.pre_job.outputs.should_skip}}"
+    if: needs.pre_job.outputs.should_skip != 'true' && (github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests'))
+    uses: ./.github/workflows/call-e2e-all-tests.yml
 
-#   integration-tests:
-#     needs: pre_job
-#     if: needs.pre_job.outputs.should_skip != true && github.event.review.state == 'commented' && !contains(github.event.pull_request.labels.*.name, 'extended-tests')
-#     run: |
+  integration-tests:
+    needs: pre_job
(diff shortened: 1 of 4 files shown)


Sahejkm (commented):
test1

Sahejkm (commented):
test2

Sahejkm (commented):
test3

Sahejkm (commented):
test1

Sahejkm (commented):
test

Sahejkm (commented):
test

Sahejkm (commented):
comment 1

Sahejkm (commented):
comment 2

.github/workflows/after-approval.yml
```diff
@@ -0,0 +1,27 @@
+name: After Approval
+
+on:
+  pull_request_review:
+    types: [submitted]
+
```
potatowagon, line 6: should we also add 

```
 paths-ignore:
      - 'packages/lexical-website/**'
      - 'packages/*/README.md'
      - '.size-limit.js'
```

here?

.github/workflows/tests-extended.yml
```diff
@@ -7,18 +7,16 @@ on:
...
-    types: [submitted]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   e2e-tests:
-    if: (github.repository_owner == 'facebook' && contains(github.event.pull_request.labels.*.name, 'extended-tests') && github.event.review.state != 'approved') || (github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests'))
+    if: github.repository_owner == 'facebook' && contains(github.event.pull_request.labels.*.name, 'extended-tests')
```
potatowagon, line 17: cmiiw, my understanding is this workflow is only triggered by extended-test label, and .github/workflows/after-approval.yml is triggered only after approval, and both workflows are independent of each other

.github/workflows/after-approval.yml
Sahejkm: - pull_request_review doesn't support this actually and it wasn't supporting it before this change as well, this only works for pull_request workflow

However, the new pre-job i added for https://github.com/fkirc/skip-duplicate-actions actually has this feature so it can be tested later https://github.com/fkirc/skip-duplicate-actions?tab=readme-ov-file#paths_ignore

.github/workflows/tests-extended.yml
Sahejkm: yeah, i needed to add "skip for multiple approval check" hence need to decouple into separate workflow so that this skip is independent of the pull_request workflow which is used for running e2e based on "labels"

potatowagon (approved):
lgtm

Issue #5420, opened 2023-12-25
Bug: build prerelease on macOS


johnnyoshika:
Lexical version: 0.12.5
OS: macOS Sonoma 14.2 (Apple Silicon M2)

## Steps To Reproduce

1. Run `npm run build-release` at the root of this project

Link to code example: not applicable

## The current behavior

Running `npm run build-release` results in:

```
> node scripts/build.js --prod --release

node:internal/errors:496
    ErrorCaptureStackTrace(err);
    ^

Error: spawn Unknown system error -86
    at ChildProcess.spawn (node:internal/child_process:414:11)
    at spawn (node:child_process:761:9)
    at Compiler.run (/Users/johnny/projects/lexical/lexical/node_modules/@ampproject/rollup-plugin-closure-compiler/node_modules/google-closure-compiler/lib/node/closure-compiler.js:76:26)
    at /Users/johnny/projects/lexical/lexical/node_modules/@ampproject/rollup-plugin-closure-compiler/dist/index.js:1564:18
    at new Promise (<anonymous>)
    at compiler (/Users/johnny/projects/lexical/lexical/node_modules/@ampproject/rollup-plugin-closure-compiler/dist/index.js:1554:12)
    at Object.renderChunk (/Users/johnny/projects/lexical/lexical/node_modules/@ampproject/rollup-plugin-closure-compiler/dist/index.js:1837:33)
    at async Chunk.render (/Users/johnny/projects/lexical/lexical/node_modules/rollup/dist/shared/rollup.js:15005:20)
    at async /Users/johnny/projects/lexical/lexical/node_modules/rollup/dist/shared/rollup.js:16132:40
    at async Promise.all (index 0) {
  errno: -86,
  code: 'PLUGIN_ERROR',
  syscall: 'spawn',
  pluginCode: 'Unknown system error -86',
  plugin: 'closure-compiler',
  hook: 'renderChunk'
}
```

## The expected behavior

`dist` folders to be generated as described [here](https://github.com/facebook/lexical/issues/3896#issuecomment-1429593096) and [here](https://github.com/facebook/lexical/issues/3958#issue-1596671396).

acywatson:
Weird - thanks for reporting. Might be time for a gcc upgrade.

etrepum:
Can't repro

PR #6271, opened 2024-06-07
fix: yjs collaboration plugin in react strict mode

meronogbai:
## Description

The collaboration plugin doesn't work well in nextjs 14.2 with strict mode enabled. See videos below to see how it breaks. I fixed the issues by relying on a useEffect instead of a useMemo to initialize yjs binding and provider. 

## Relevant thread

https://discord.com/channels/953974421008293909/1233102329931366531/1233102329931366531

## Test plan

- Clone [test repo](https://github.com/meronogbai/nextjs-yjs-lexical)
- cd into the cloned test repo
- Sign up for [liveblocks](https://liveblocks.io/) and add `LIVEBLOCKS_SECRET_KEY` to the root `.env.local` file. Alternatively, I can send you a secret key as well.
- Run `npm run dev` and test the editor on multiple tabs/windows/browsers.
- To fix the issue, copy the updated `LexicalCollaborationPlugin` file into node_modules

```bash
cp LexicalCollaborationPlugin.dev.mjs node_modules/@lexical/react/LexicalCollaborationPlugin.dev.mjs
```
- Stop the previous development server and start it again. Keep in mind that you have to remove the `.next` folder because nextjs caches built modules.
```bash
rm -r .next && npm run dev
```

### Before

https://www.loom.com/share/460727ea35b04e26a311d55d5283be3

### After

https://www.loom.com/share/d3a2701d4ef6456a9bd7c9a202d38ebe

## Note

~I tried to run `npm run build-release`, but it didn't work due to https://github.com/facebook/lexical/issues/5420 so I had to compile "manually".~ Never mind, it was just an error with my changes.


StyleT (changes requested):
Hi! Thanks for the fix but `npm run tsc` is failing. Can you pls fix it?

fantactuka:
Could you use `git mv` for renaming to retain file history

meronogbai:
@fantactuka I believe I did that already

image

The blame info hasn't changed as you can see in https://github.com/facebook/lexical/blame/e6982e727b59eddaf69b25ba8fd32a9e837d969b/packages/lexical-react/src/LexicalCollaborationPlugin.tsx

StyleT:
@meronogbai Hi! Took another look at this.. So we still call `providerFactory` here twice, although we now do proper destruction handling, which is great! However I would say we still may do 1 step forward and make this code actually call `providerFactory` only once.

Why? From my experience `y-webrtc` would throw on you even if you call `provider.disconnect()`.

Here is simple test that you can add to this PR and make it pass it:

```typescript
/**
 *
 *
 */

import { CollaborationPlugin } from '@lexical/react/LexicalCollaborationPlugin';
import { LexicalComposer } from '@lexical/react/LexicalComposer';
import { ContentEditable } from '@lexical/react/LexicalContentEditable';
import {LexicalErrorBoundary} from '@lexical/react/LexicalErrorBoundary';
import { RichTextPlugin } from '@lexical/react/LexicalRichTextPlugin';
import {createRoot, Root} from 'react-dom/client';
import * as ReactTestUtils from 'shared/react-test-utils';
import * as Y from 'yjs';

describe(`LexicalCollaborationPlugin`, () => {
  let container: HTMLDivElement;
  let reactRoot: Root;

  const editorConfig = Object.freeze({
    // NOTE: This is critical for collaboration plugin to set editor state to null. It
    // would indicate that the editor should not try to set any default state
    // (not even empty one), and let collaboration plugin do it instead
    editorState: null,
    namespace: 'Test editor',
    nodes: [],
    // Handling of errors during update
    onError(error: Error) {
      throw error;
    },
  });

  beforeEach(() => {
    container = document.createElement('div');
    reactRoot = createRoot(container);
    document.body.appendChild(container);
  });

  test(`providerFactory called only once`, () => {
    const providerFactory = jest
      .fn((id: string, yjsDocMap: Map<string, Y.Doc>) => {
        const doc = new Y.Doc();
        yjsDocMap.set(id, doc);

        return {
          awareness: {
            getLocalState: () => null,
            getStates: () => new Map(),
            off: () => {},
            on: () => {},
            setLocalState: () => {},
          }, 
          connect: () => {}, 
          disconnect: () => {},
          off: () => {},
          on: () => {},
        }
      });
    function MemoComponent() {
      return (<LexicalComposer initialConfig={editorConfig}>
        {/* With CollaborationPlugin - we MUST NOT use @lexical/react/LexicalHistoryPlugin */}
        <CollaborationPlugin
          id="lexical/react-rich-collab"
          providerFactory={providerFactory}
          // Unless you have a way to avoid race condition between 2+ users trying to do bootstrap simultaneously
          // you should never try to bootstrap on client. It's better to perform bootstrap within Yjs server.
          shouldBootstrap={false}
        />
        <RichTextPlugin
          contentEditable={<ContentEditable className="editor-input" />}
          placeholder={<div className="editor-placeholder">Enter some rich text...</div>}
        />
      </LexicalComposer>);
    }
    ReactTestUtils.act(() => {
      reactRoot.render(
        <React.StrictMode>
          <MemoComponent />
        </React.StrictMode>,
      );
    });
    
    expect(providerFactory).toHaveBeenCalledTimes(1);
  });
});
```

How to make code pass this test? Use "dirty" hack from here https://taig.medium.com/prevent-react-from-triggering-useeffect-twice-307a475714d7

meronogbai:
Thank you so much @StyleT! I pushed a commit with your suggested changes 🫡

StyleT (approved):

StyleT:
I'm good to merge this, seems like test failures are unrelated

PR #6274, opened 2024-06-07
Bug Fix: Fix selection when hitting "shift+arrowdown" into a table.

### LINKED ISSUE #6273: Bug: Shift+ArrowDown into a table selects only up to the first cell and its children
Lexical version: 0.15.0

Pressing "Shift+ArrowDown" into a table selects only up to the first table-cell node and its children, even though the whole table appears selected. This is different from "Shift+ArrowUp," which correctly selects the entire table and all of its children.

[screencast-playground.lexical.dev-2024.06.07-14_50_45.webm](https://github.com/facebook/lexical/assets/88986106/da356bd2-e8f3-48b6-b328-ff37281c83a0)

## Steps To Reproduce

1. Insert a 2x2 table.
2. Select the first paragraph.
3. Hit `Shift+↓`.

## The current behavior
The focus is on the root, and the selection includes only up to the first cell and its children. 

## The expected behavior
The entire table and its children should be included in the selection.

serey-roth:
Pressing "Shift+ArrowDown" into a table selects only up to the first table-cell node and its children, even though the whole table appears selected. 

## Description

Previously, the new focus would be set on the table's parent if the current selection was partially within a table. Now, it will be on the first/last cell of the table. 

Closes #6273

### Before

[screencast-playground.lexical.dev-2024.06.07-14_50_45.webm](https://github.com/facebook/lexical/assets/88986106/95423213-f2ee-4c90-b673-fd723d9b5b5d)

### After

[screencast-localhost_3000-2024.06.07-15_27_25.webm](https://github.com/facebook/lexical/assets/88986106/ad8b3df1-879e-443c-acfe-61b744ad5a26)


ivailop7 (approved):

ivailop7:
@serey-roth can you recheck the tests, something seems off, I've reran them a couple of times, but fail consistently.

Sahejkm:
sorry, removing the tag to prevent e2e test on every run, the tests should be triggered on next Approval again automatically.

ivailop7 (approved):

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1284,37 +1314,112 @@ function $handleArrowKey(
...
-      );
-      if (!parentNode) {
-        return false;
-      }
-      const siblingNode = parentNode.getPreviousSibling();
-      if (!siblingNode || !$isTableNode(siblingNode)) {
-        return false;
+        const anchorNode = selection.anchor.getNode();
+        if ($isRootOrShadowRoot(anchorNode)) {
+          const selectedNode = selection.getNodes()[0];
```
serey-roth, line 1351: In all my tests, I noticed that when the selection is on the top-level `ElementNode`, the paragraph in the first table-cell will be selected.

serey-roth:
@ivailop7 This PR has grown beyond just fixing the "shift+arrowdown" keyboard combination for tables. I've also updated it to handle the "shift+arrowup" combination and various scenarios for these combinations:

- When the table is a node
- When the table is the first node/last node
- When the table is the only node

The changes include a fix for table deletion after using shift+arrow into the table. I've moved the selection logic into the key arrow command listener, which is more logical than the selection change command listener since the latter covers more than just key arrow events.

One thing I've noticed and would like your thoughts on: with my changes, when we shift+arrow down into a table, it also selects the following paragraph. This isn't new behavior—it also happens if you select the table with a mouse. If you think this shouldn't happen, the problem goes beyond the scope of the keyboard events.

ivailop7:
> @ivailop7 This PR has grown beyond just fixing the "shift+arrowdown" keyboard combination for tables. I've also updated it to handle the "shift+arrowup" combination and various scenarios for these combinations:
> 
> 
> 
> - When the table is a node
> 
> - When the table is the first node/last node
> 
> - When the table is the only node
> 
> 
> 
> The changes include a fix for table deletion after using shift+arrow into the table. I've moved the selection logic into the key arrow command listener, which is more logical than the selection change command listener since the latter covers more than just key arrow events.
> 
> 
> 
> One thing I've noticed and would like your thoughts on: with my changes, when we shift+arrow down into a table, it also selects the following paragraph. This isn't new behavior—it also happens if you select the table with a mouse. If you think this shouldn't happen, the problem goes beyond the scope of the keyboard events.

This is fine. Will review and merge as is. If you want to fix the selection of the following paragraph, which I think we should, let's do it in a separate PR.

ivailop7 (approved):

ivailop7:
It seems it's failing your new tests, could you check?

ivailop7 (approved):

PR #6291, opened 2024-06-10
[lexical-playground][lexical-devtools] Fix: Add missing explicit devDependencies for vite config

### LINKED ISSUE #6280: Bug: missing NPM `packages/lexical-playground/package.json`'s `devDependencies`: `@rollup/plugin-babel @babel/plugin-transform-flow-strip-types @babel/preset-react`
Hi, I have a project with extracted playground into a separate repo for tag v0.16.0:

When doing `npm run --prefix packages/lexical-playground build-prod`, I'm getting an error `Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@rollup/plugin-babel'`

Then by adding it to the prior `npm install` run, I'm getting a similar errors for `@babel/plugin-transform-flow-strip-types @babel/preset-react`

Adding all three to `npm install` makes it work:
```shell
npm install --force --prefix packages/lexical-playground @rollup/plugin-babel @babel/plugin-transform-flow-strip-types @babel/preset-react
npm run --prefix packages/lexical-playground build-prod
```

So I assume this is indeed only a matter of adding these to `devDependencies`?

Removing `--force` produces this error:
```
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: lexical-playground@0.16.0
npm error Found: vite@5.2.[13](https://github.com/vadimkantorov/lexical-playground-only/actions/runs/9437767079/job/25994006725#step:4:14)
npm error node_modules/vite
npm error   dev vite@"^5.2.11" from the root project
npm error
npm error Could not resolve dependency:
npm error peer vite@"^2" from vite-plugin-replace@0.1.1
npm error node_modules/vite-plugin-replace
npm error   dev vite-plugin-replace@"^0.1.1" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

etrepum:
## Description

A few modules used in the vite/wxt builds were missing as explicit devDependencies in lexical-playground and lexical-devtools. This adds them to package.json. There shouldn't be any observable difference in behavior, unless you are trying to extract these projects out of the monorepo or maybe trying to use more strict dependency management tool than npm.

Closes #6280

### Before

I don't think you could run into any problems with the status quo without doing something unsupported, e.g. #6280, because npm/node will happily just use whatever happens to be present in the workspace root's node_modules.

### After

All known dependencies are in the respective package.json files. This probably means that these packages are a bit easier to extract (but there are other "unique" build dependencies like the shared package and invariant that would still make it a bit non-trivial)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -479,11 +479,11 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.24.2",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.2.tgz",
-      "integrity": "sha512-y5+tLQyV8pg3fsiln67BVLD1P13Eg4lh5RW9mF0zUuvLrv9uIQ4MCL+CRT+FTsBlBjcIan6PGsLcBN0m3ClUyQ==",
+      "version": "7.24.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.7.tgz",
+      "integrity": "sha512-BcYH1CVJBO9tvyIZ2jVeXgSIMvGZ2FDRvDdOIVQyuklNKSsx+eppDEBq/g47Ayw+RqNFE+URvOShmf+f/qwAlA==",
       "dependencies": {
-        "@babel/highlight": "^7.24.2",
+        "@babel/highlight": "^7.24.7",
         "picocolors": "^1.0.0"
       },
       "engines": {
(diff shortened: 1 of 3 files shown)


Sahejkm (approved):

StyleT (approved):

vadimkantorov:
thank you for fixing this :) I'll try with the next release if we still need `--force` :)

PR #6298, opened 2024-06-12
CI: Only send flaky test notice once

potatowagon:
## Description
Previously flaky test notice is sent on each call-e2e-test.yml failure. we just need to send once for each e2e test job. so moving the notice trigger after the main e2e test job completes. there are 2 entrypoints to that, in tests-extended.yml & after-approval.yml

### Before

flaky test notice posts on every e2e test failure

### After

flaky test notice only post once per e2e test job failure

https://github.com/facebook/lexical/pull/6298#issuecomment-2162444144

Screenshot 2024-06-12 at 4 56 06 PM

tested
1. e2e job has pass and fail runs -> trigger notice
2. e2e job all fail runs -> trigger notice

^ tested using the extended-test flow

will wait for approval to test after approval flow


.github/workflows/after-approval.yml
```diff
@@ -22,6 +22,18 @@ jobs:
     if: needs.pre_job.outputs.should_skip != 'true' && (github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests'))
     uses: ./.github/workflows/call-e2e-all-tests.yml
 
+  flaky-test-notice:
+    needs: e2e-tests
+    runs-on: ubuntu-latest
+    if: always()
```
Sahejkm, line 28: is this needed ? if its always gonna execute

.github/workflows/tests-extended.yml
```diff
@@ -17,6 +17,18 @@ jobs:
     if: github.repository_owner == 'facebook' && contains(github.event.pull_request.labels.*.name, 'extended-tests')
     uses: ./.github/workflows/call-e2e-all-tests.yml
 
+  flaky-test-notice:
+    needs: e2e-tests
+    runs-on: ubuntu-latest
+    if: always()
```
Sahejkm, line 23: ^

.github/workflows/after-approval.yml
potatowagon: yup its needed, because if e2e-tests job is not completed successfully, this job skips entirely unless if: always() is specified

potatowagon: this is the commit that tested the above observation: https://github.com/facebook/lexical/pull/6298/commits/363203a487ae549f877ceffda4a3a1b5c6218c5d

Sahejkm (approved):

Sahejkm (approved):

PR #6302, opened 2024-06-12
[Lexical][Meta] Add meta intern static docs config

Sahejkm:
## WHAT
- Add meta intern static docs config
- Fix broken JSX rendering in MD files caused due to preprocessor added for markdown validations

## WHY
- needed to sync and build intern static docs in meta for lexical


packages/lexical-website/docusaurus.config.js
```diff
@@ -225,15 +225,6 @@ const config = {
...
 
-  markdown: {
-    format: 'md',
-    preprocessor: ({fileContent}) =>
-      fileContent.replaceAll(
-        'https://stackblitz.com/github/facebook/lexical/tree/main/',
-        STACKBLITZ_PREFIX,
-      ),
-  },
-
```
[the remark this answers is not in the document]
etrepum: have you tried just removing the `format: 'md',`?

etrepum, line 236: it looks like this update to docusaurus-plugin-internaldocs-fb is switching from a markdown to mdx, so removing that configuration override should resolve the problem without throwing away the preprocessor which is useful for testing new examples in preview environments.

```js
  markdown: {
    preprocessor: ({fileContent}) =>
      fileContent.replaceAll(
        'https://stackblitz.com/github/facebook/lexical/tree/main/',
        STACKBLITZ_PREFIX,
      ),
  },
```

Sahejkm: this does the job indeed. Thanks @etrepum Bob!

packages/lexical-utils/src/index.ts
```diff
@@ -131,9 +131,9 @@ export function isMimeType(
  *  3. Order aware (respects the order when multiple Files are passed)
  *
  * const filesResult = await mediaFileReader(files, ['image/']);
- * filesResult.forEach(file => editor.dispatchCommand('INSERT_IMAGE', {
+ * filesResult.forEach(file => editor.dispatchCommand('INSERT_IMAGE', \\{
  *   src: file.result,
- * }));
+ * \\}));
```
potatowagon, line 136: what is \\ this for?

potatowagon (approved):

packages/lexical-utils/src/index.ts
Sahejkm: Fixing parsing error for mdx , it misunderstands it as code

etrepum: MDX is implicitly JSX anywhere that could possibly make sense (it's a weird hybrid of markdown and jsx), so wherever you have `{` or `<` outside of some other markdown context like backticks or a code block it starts parsing as a JSX child expression or a JSX tag.

potatowagon: i see

PR #6306, opened 2024-06-13
[lexical-link] Fix bug when can't remove link formatting from autolink

### LINKED ISSUE #5607: Bug: Can't remove link formatting from links added by AutoLinkPlugin
When link is typed manually and converted to link by AutoLink plugin
It is recognised as link but I cant remove link formatting when clicking on link format button

Lexical version: 0.13.1

## Steps To Reproduce

1. Type link "www.customlink.com"
2. Select link for editing with FloatingLinkToolbar
3. Click on link formatting button

https://github.com/facebook/lexical/assets/23315851/6b7e63e9-6993-4373-8693-c107058922f5

## The current behavior
Link formatting isn't removed

## The expected behavior
Link formatting is removed

MaxPlav:
This PR fixes a bug where a user cannot remove link formatting (unlink) from links created by AutoLinkNode. However, it works for links created by LinkNode. The proposal is to add a flag property in AutoLinkNode to indicate if the link was unlinked. If so, render the link as regular text, keeping styles. The hidden auto-link can become a link again via the Toolbar Link button.

Closes #5607

## Test plan
2. Input to the editor "www.facebook.com".
3. Check the auto-link is created.
4. Focus on the link.
5. Click the delete button on the popup.
6. The link should transform to the regular text "www.facebook.com".

### Before

https://github.com/facebook/lexical/assets/5062807/4687fd1e-b469-4546-9602-361023a9602c

### After

https://github.com/facebook/lexical/assets/5062807/37492852-bc42-4dc3-90a8-d77a3bb0fae3

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -35,12 +35,18 @@ export type LinkAttributes = {
   title?: null | string;
 };
 
+export type AutoLinkAttributes = Partial<
+  Spread<LinkAttributes, {isUnlinked?: boolean}>
+>;
+
 export type SerializedLinkNode = Spread<
   {
     url: string;
   },
   Spread<LinkAttributes, SerializedElementNode>
 >;
+
+type LinkHTMLElementType = HTMLAnchorElement | HTMLSpanElement;
 
 const SUPPORTED_URL_PROTOCOLS = new Set([
   'http:',
@@ -82,7 +88,7 @@ export class LinkNode extends ElementNode {
     this.__title = title;
   }
 
-  createDOM(config: EditorConfig): HTMLAnchorElement {
+  createDOM(config: EditorConfig): LinkHTMLElementType {
     const element = document.createElement('a');
     element.href = this.sanitizeUrl(this.__url);
     if (this.__target !== null) {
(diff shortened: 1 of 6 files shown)


packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -89,6 +89,11 @@
   text-decoration: underline;
   cursor: pointer;
 }
+.PlaygroundEditorTheme__hiddenLink {
```
StyleT, line 92: I think that instead of masking `<a>` tag as "regular" text - we shall simply render "hidden link" as `<span>`

packages/lexical-link/src/index.ts
```diff
@@ -309,25 +317,94 @@ export function $isLinkNode(
...
+  setIsUnlinked(value: boolean) {
+    const self = this.getWritable();
+    self.__isUnlinked = value;
+    return self;
+  }
+
+  createDOM(config: EditorConfig): HTMLAnchorElement {
+    if (this.__isUnlinked) {
+      // render a link that behaves like a span
+      const element = document.createElement('a');
```
StyleT, line 372: Simply use `span` here? Or no wrapper all together?

StyleT (commented):
LGTM overall! Can you also add some unit & E2E tests pls?

MaxPlav:
> LGTM overall! Can you also add some unit & E2E tests pls?

hi, sure I've added an e2e test and unit tests. Also now the unlinked AutoLink node is rendered as "span" element. For that I had to update several types of the `LinkNode` methods, as a return type of the overridden method must be a subtype of the return type in the superclass.

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
MaxPlav: removed that

StyleT (approved):
LGTM!

PR #6307, opened 2024-06-14
[Lexical][CI] Fix issues in after_approval workflow

Sahejkm:
## WHAT
Fix some issues in Lexical CI job  after observing past runs :
- path-ignore for e2e tests not working due to warning on wrong keyword, should be replaced by path_ignore 
Screenshot 2024-06-14 at 11 05 57 AM

- flaky-test notice not working on "PRs from forks"
https://github.com/facebook/lexical/actions/runs/9506494216/job/26204471038?pr=6274

- concurrent-workflow skipping all concurrent workflows running on multiple PRs rather just for the PR in focus, - change settings from 'always' to "same_content_newer".

## Test plan

#6308

### DIFF
--- a/.github/workflows/after-approval.yml
+++ b/.github/workflows/after-approval.yml
@@ -29,7 +29,7 @@ jobs:
     if: always()
     steps:
       - name: Flaky test notice PR comment
-        if: (${{ github.event.pull_request.head.repo.full_name == 'facebook/lexical' }} && (needs.e2e-tests.result == 'failure' || needs.e2e-tests.result == 'cancelled'))
+        if: (needs.e2e-tests.result == 'failure' && ${{ github.event.pull_request.head.repo.full_name == 'facebook/lexical' }}) || (needs.e2e-tests.result == 'cancelled' && ${{ github.event.pull_request.head.repo.full_name == 'facebook/lexical' }})
         uses: thollander/actions-comment-pull-request@v2
         with:
           message: |
--- a/.github/workflows/after-approval.yml
+++ b/.github/workflows/after-approval.yml
@@ -23,26 +23,6 @@ jobs:
     if: needs.pre_job.outputs.should_skip != 'true' && (github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests'))
     uses: ./.github/workflows/call-e2e-all-tests.yml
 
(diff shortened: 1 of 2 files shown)


.github/workflows/after-approval.yml
```diff
@@ -6,34 +6,23 @@ on:
...
           skip_after_successful_duplicate: 'true'
           do_not_skip: '["pull_request", "merge_group"]'
-          paths-ignore: '["packages/lexical-website/**", "packages/*/README.md", ".size-limit.js"]'
+          paths_ignore: '["packages/lexical-website/**", "packages/*/README.md", ".size-limit.js"]'
   e2e-tests:
     needs: pre_job
     if: needs.pre_job.outputs.should_skip != 'true' && (github.event.review.state == 'approved' && !contains(github.event.pull_request.labels.*.name, 'extended-tests'))
     uses: ./.github/workflows/call-e2e-all-tests.yml
 
-  flaky-test-notice:
```
potatowagon, line 25: can we please leave this here first? it still works on branches on this repo. ill like to try and fix instead of deleting it so quickly

Sahejkm: - this keeps the pipeline red and its not a new issue we are trying to fix, its kinda wrong signal to PRs from forks, lemme check if we can figure out if PR is from fork else i would like to not make the CI red

- we can readd this once we have the fix ready

potatowagon: >lemme check if we can figure out if PR is from fork else i would like to not make the CI red

this would be great, if its failing the ci and cant figure something out for the above then im ok to remove first.

Sahejkm: @potatowagon  i tried couple if approaches but looks like "always()" is causing this step to always run regardless of whatever if else check i put related to ${{ github.event.pull_request.head.repo.full_name != 'facebook/lexical' }} anywhere in code.

would prefer removing this 1st till we get a better fix

Sahejkm: check this weird run

https://github.com/facebook/lexical/actions/runs/9510521267/job/26215150424

Screenshot 2024-06-14 at 12 55 08 PM

it evaluating (true && (needs.e2e)) as true

potatowagon (approved):

PR #6318, opened 2024-06-17
Few missing exports that are useful

Piechota:
Just added bunch of exports that were useful to me.
Might help others.


Sahejkm (approved):
LGTM

PR #6335, opened 2024-06-24
Fix undo

Piechota:
When doing undo in document where some text had format != 0
Error was thrown:
```
TypeError: Cannot assign to read only property '_cachedNodes' of object '#<RangeSelection>'
    at RangeSelection.setCachedNodes 
    at ParagraphNode.getWritable 
    at ParagraphNode.setTextFormat 
    at reconcileParagraphFormat 
    at $reconcileChildrenWithDirection 
 ```
Basically editorState is readOnly at this point, but we try to write to ParagraphNode when text's format is different from paragraph.
I have copied solution from `reconcileBlockDirection`

### DIFF
--- a/packages/lexical/src/LexicalReconciler.ts
+++ b/packages/lexical/src/LexicalReconciler.ts
@@ -355,7 +355,10 @@ function reconcileParagraphFormat(element: ElementNode): void {
     subTreeTextFormat !== element.__textFormat &&
     !activeEditorStateReadOnly
   ) {
-    element.setTextFormat(subTreeTextFormat);
+    if (!activeEditorStateReadOnly) {
+      const writableNode = element.getWritable();
+      writableNode.__textFormat = subTreeTextFormat;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


Sahejkm:
Hi [Piechota](https://github.com/Piechota), 
- could we add the screenshots / videos to replicate this issue ? Thanks !
- we should also add a test for this

Sahejkm (approved):

Piechota:
@Sahejkm I will add tests later this week
as for screenshot/video 
there isn't much to show with those. 
Set content with editor.setEditorState() (I missed that part in description and this might be the reason why I don't see this on playground) -> type something -> ctrl+z = error if there was TextNode with format != 0

Tbh I don't understand exactly where this error is coming from, but there is readOnly check in `reconcileBlockDirection` too and my guess is it was for the same error.

Sahejkm:
> Tbh I don't understand exactly where this error is coming from (what makes editorState to be readOnly), but there is readOnly check in `reconcileBlockDirection` too and my guess is it was for the same error.

It would be great to get some steps to replicate the crash to understand the issue and fix better

Piechota:
Small update.
I couldn't reproduce this with simple unit test.
So either I am doing something wrong (most likely) or it has something to do with input method.
I will have to investigate further.

etrepum:

PR #6340, opened 2024-06-26
[lexical-html] Bug Fix: Preserve text alignment of nested block elements, when converting HTML to Lexical Nodes

### LINKED ISSUE #6162: Bug: Converting HTML to Lexical Nodes strips text alignment from paragraphs
Lexical version: 0.15.0

## Steps To Reproduce

1. Create a new block element lexical node and register an importDOM on "div".
2. Using the HTML: `<div><p style="text-align: right;">Hello world!</p></div>`, pass it through `$generateNodesFromDOM`
3. You'll see the format property of our paragraph node will be set incorrectly to "0".

This is where the bug is coming from: https://github.com/facebook/lexical/blob/2dac7f5c4751bf772bf573ad7e3665e1482a5bfd/packages/lexical-html/src/index.ts#L324

```
if ($isBlockElementNode(node)) {
  node.setFormat(textAlign);
  out.push(node);
}
```

Is the `node.setFormat` line necessary? If it is, perhaps something like:

```
if (!node.getFormat()) node.setFormat(textAlign);
```

## The current behavior

The paragraph node's alignment is being overwritten by the div's unset alignment.

## The expected behavior

If alignment (i.e. format) is set on the paragraph, it should take precedence over its parent block format property.

nzigopis:
## Description
When an HTML block element, contained in another element, has text alighment  set, 
preserve the alignment, and don't replace from parent's
Closes #6162 

### Before

An html text like this: 
`<div><p style="text-align: center;">Hello world!</p></div>`
would result in paragraph with no text align,
while an html like this:
`<div style="text-align: right;"><p style="text-align: center;">Hello world!</p></div>`
would set the paragraph's text align to _right_

### After

In both cases above, paragraph preserves it's _center_ alignment


ivailop7 (approved):

jiangtutu0529:
> ## Description
> When an HTML block element, contained in another element, has text alighment set, preserve the alignment, and don't replace from parent's Closes #6162
> 
> ## Test plan
> ### Before
> An html text like this: `<div><p style="text-align: center;">Hello world!</p></div>` would result in paragraph with no text align, while an html like this: `<div style="text-align: right;"><p style="text-align: center;">Hello world!</p></div>` would set the paragraph's text align to _right_
> 
> ### After
> In both cases above, paragraph preserves it's _center_ alignment

>An html text like this: `<p class="editor-paragraph"><span style="white-space: pre-wrap;">4534</span><span style="color: rgb(255, 185, 140); white-space: pre-wrap;">53</span></p> ` would result in paragraph with no color,how can i fix this bug?

cccccroge:
> > ## Description
> > When an HTML block element, contained in another element, has text alighment set, preserve the alignment, and don't replace from parent's Closes #6162
> > ## Test plan
> > ### Before
> > An html text like this: `<div><p style="text-align: center;">Hello world!</p></div>` would result in paragraph with no text align, while an html like this: `<div style="text-align: right;"><p style="text-align: center;">Hello world!</p></div>` would set the paragraph's text align to _right_
> > ### After
> > In both cases above, paragraph preserves it's _center_ alignment
> 
> > An html text like this: `<p class="editor-paragraph"><span style="white-space: pre-wrap;">4534</span><span style="color: rgb(255, 185, 140); white-space: pre-wrap;">53</span></p> ` would result in paragraph with no color,how can i fix this bug?

Yes, this is actually annoying...

PR #6347, opened 2024-06-27
[lexical] Feature: Implement Editor.read and EditorState.read with editor argument

### LINKED ISSUE #6346: Feature: Provide a variant of EditorState.read that sets activeEditor
## Description

`EditorState.read` can only use $functions that don't depend on `activeEditor` which ends up leading users to use `Editor.update` even for fundamentally read-only operations. A common use case would be `$getNodeFromDOMNode` or `$getNodeFromDOM` or to avoid having explicit `LexicalEditor` parameters for $functions (`$generateHtmlFromNodes`, `$getHtmlContent`, etc.).

We should have a way to set the `activeEditor` with an additional argument to `EditorState.read`.

There should also be a version of `read` directly from the `LexicalEditor` for symmetry with `update`. This would also provide an opportunity to optionally get the `_pendingEditorState` so that users can "read their own writes" from an `update` before reconciliation has occurred.

The public API changes would be as follows:

```ts
export interface EditorReadOptions {
  // Not sure whether the best default would be to
  // use pending or the latest reconciled state
  pending?: boolean;
}

class LexicalEditor {
  // …
  read<V>(callbackFn: () => V, options: EditorReadOptions = {): V {
    const editorState = (options.pending && this._pendingEditorState) || this.getEditorState();
    return editorState.read(callbackFn, { activeEditor: this });
  }
}
```

```ts
export interface EditorStateReadOptions {
  editor?: LexicalEditor | null;
}

class EditorState {
  // …
  read<V>(callbackFn: () => V, options?: EditorStateReadOptions): V {
    // Changes would be required to internal readEditorState to support this
    // Using editor as first argument for symmetry with lexical-devtools lexicalForExtension internal API
    return readEditorState(options && options.editor || null, this, callbackFn);
  }
}
```

## Impact

Users would have a simpler to explain model for what a $function can do, as most would transition from using `editor.update` for read-only operations to `editor.read` and people will also likely prefer `editor.read` to `editor.getEditorState().read`.

The extra argument to `EditorState.read` would mostly be an implementation detail that I don't think most people would use, but it keeps the code path very similar to the current paradigm.

etrepum:
## Description

Adds an `options` argument to `EditorState.read` allowing the `activeEditor` to be set.

Adds a new `Editor.read` method as a convenience that can call `EditorState.read` with the editor set.

Closes #6346


packages/lexical/src/LexicalEditor.ts
```diff
@@ -1113,6 +1117,24 @@ export class LexicalEditor {
...
+  /**
+   * Executes a read of the editor's state, with the
+   * editor context available (useful for exporting and read-only DOM
+   * operations). Much like update, but prevents any mutation of the
+   * editor's state.
+   * @param callbackFn - A function that has access to read-only editor state.
+   * @param options - A bag of options to control the behavior of the read.
+   * @param options.pending - Use the pending editorState. Use this only when
+   * it is necessary to read the state that has not yet been reconciled (this
+   * is the state that you would be working with from editor.update).
```
zurfyx, line 1129: IMO this behavior is confusing, `.read` should always grab the `pending` state just like `update`. You would not expect an update to be done on top of the reconciler, but rather on top of your previous changes, likewise for 

```
editor.update(() => {
  // append paragraph
});
editor.read(() => {
 // paragraph should be here
});
```

packages/lexical/src/LexicalEditorState.ts
```diff
@@ -108,8 +112,12 @@ export class EditorState {
     return this._nodeMap.size === 1 && this._selection === null;
   }
 
-  read<V>(callbackFn: () => V): V {
-    return readEditorState(this, callbackFn);
+  read<V>(callbackFn: () => V, options?: EditorStateReadOptions): V {
```
zurfyx, line 115: I'm not a fan of this option like I stated in https://github.com/facebook/lexical/issues/6346#issuecomment-2197367721

While the intent is valid, I feel like it can easily lead to pitfalls where the editor is no longer compatible with the EditorState.

For reference, @ivailop7 mentioned a real-use case around DOM (for tables) and this can potentially fail and be hard to debug with this API, whereas `editor.read` is intuitive and always does what expected.

zurfyx (commented):
Thank you Bob! Added two opinionated comments, I wonder what the rest think, particularly the folks who were involved in the discussion @fantactuka @ivailop7 @StyleT

packages/lexical/src/LexicalEditor.ts
etrepum: I don't hold any strong opinions on the API, so long as the option exists. I implemented it this way mostly based on the loudest opinions at the time 😆

packages/lexical/src/LexicalEditorState.ts
etrepum: I'm totally open to leaving `EditorState.read`'s signature alone and have `Editor.read` call `readEditorState` directly. I'll wait until there seems to be some consensus before changing it.

etrepum: I only implemented it this way because it seemed to be the stated preference of @StyleT on the call and @fantactuka in #6346

StyleT: My point was that we shouldn't have 2 ways of reading the state. But we may sugar coat some API. As mentioned in https://github.com/facebook/lexical/issues/6346#issuecomment-2195583861 the idea is the following:
1. `editor.read` calls `EditorState.read` while passing correct `activeEditor` as an option
2. We promote `editor.read` in the documentation, while keeping `EditorState.read` for backward compatibility reasons and for more advanced use cases


This allows to:
a. Achieve API consistency
b. Avoid confusing "regular users" with 2 similar APIs
c. Reduce code duplication and establish relation between related APIs "in code"

packages/lexical/src/LexicalEditor.ts
ivailop7: I'm with Gerard on this one.

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -22,6 +22,7 @@ import {
   $createNodeSelection,
   $createParagraphNode,
   $createTextNode,
+  $getEditor,
```
ivailop7, line 25: Could I add a small request to add a test, that checks if '$getNearestNodeFromDOMNode' can be called within the new "read" method.

packages/lexical/src/LexicalEditor.ts
etrepum: I'm surprised that this would be your preference since the DOM use case would really be best handled with the latest reconciled editor state rather than the pending editor state since the DOM is not necessarily in sync with that yet.

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
etrepum: Sure, it does work, just added tests. Noticed that there's a small inconsistency in that the root node is never set in the `_keyToDOMMap`. Not sure if is really intentional or not.

It does only run the tests with the latest reconciled state (the current default) because that is really what probably makes the most sense for this use case.

etrepum:
It seems there's not yet any consensus as to whether `Editor.read` should choose the pending or previous state.

pending: @zurfyx, @ivailop7 
previous: @StyleT, @fantactuka 

I don't have any strong preference, as long as both are possible. I lean slightly towards previous state because it matches what's rendered (and after transforms) which is possibly a more generally useful scenario. I think users can be surprised either way ("read doesn't have my latest writes!" or "the dom doesn't match what I'm reading! shouldn't my transforms prevent the state from looking like this?!"). For my personal use cases I mostly just wanted to be able to get a handle to the editor from a read to get access to information that's independent of the current state (config, theme, builder, etc.).

I also don't have any strong preference as to whether we add options to `EditorState.read` (the current implementation, suggested by @StyleT and @fantactuka), or have Editor.read directly call into the lower-level internal `readEditorState` (presumably what @zurfyx would prefer?).

Poking around in the implementation of `$getNearestNodeFromDOMNode` I noticed that we don't ever consider the `RootNode` in any of those calls. Not sure if that was an intentional decision or just an accident based on the implementation detail that only the children of `RootNode` are reconciled in the standard way so `storeDOMWithKey` is never called (but it could be in `setRootElement`). I think it would mostly only matter in the nested editor case?

zurfyx:
> I don't have any strong preference, as long as both are possible. I lean slightly towards previous state because it matches what's rendered (and after transforms) which is possibly a more generally useful scenario. I think users can be surprised either way ("read doesn't have my latest writes!" or "the dom doesn't match what I'm reading! shouldn't my transforms prevent the state from looking like this?!"). For my personal use cases I mostly just wanted to be able to get a handle to the editor from a read to get access to information that's independent of the current state (config, theme, builder, etc.).

IMO the delay in the reconcilation should be transparent to the user. We only introduced it because of performance, as it was unreasonable to have each update trigger reconciliation (the most expensive part). The `discrete` option was intentionally named this way so that only the most expert users would need to understand the implications of batching.

I would be keen on destroying any potential link to `pending` or `latest` on product code unless it's explictly marked by advanced flags.

For example, `editor.getEditorState()` can lead lead to a potential obsolete state. In this case, we may want to complete reconciliation and return the newly created EditorState instead.

Likewise, `.read` should return the same predictable instance.

> Poking around in the implementation of `$getNearestNodeFromDOMNode` I noticed that we don't ever consider the `RootNode` in any of those calls. Not sure if that was an intentional decision or just an accident based on the implementation detail that only the children of `RootNode` are reconciled in the standard way so `storeDOMWithKey` is never called (but it could be in `setRootElement`). I think it would mostly only matter in the nested editor case?

We never considered the `contenteditable` itself to be the RootNode, even though the CSS classname says otherwise. The RootNode technically has not DOM.

etrepum:
Having editor.read flush any pending reconciliation sounds like a nice plan. I think the only real downside there is if it ends up nested in an update that has already made changes and will make more after the read, that could get a bit ugly (maybe raise an error for nesting in that scenario?). Anyone else have thoughts on this API change?

zurfyx:
> Having editor.read flush any pending reconciliation sounds like a nice plan. I think the only real downside there is if it ends up nested in an update that has already made changes and will make more after the read, that could get a bit ugly (maybe raise an error for nesting in that scenario?). Anyone else have thoughts on this API change?

I don't think you necessarily need to flush on `read` because users will never have access to the EditorState. The reason why I suggested flushing in `getEditorState()` is because that's the time when they export a copy of the EditorState and has to stay equivalent with the state you would access in other operations.

But I do agree that nested updates can lead to unexpected behavior, they're not immediately obvious. Unfortunately, I don't think we can throw runtime errors as they won't work well with React, we've seen this before where you can get in a state where the next React render happens inside an update (the well-known antipattern and the reason we have to be lenient with it).

etrepum:
I think we'd have to add a different method from `getEditorState()` because flushing there would be a breaking change, it's pretty easy to find situations where it's expected to return a previous state (e.g. `editorStateHasDirtySelection` is the first one).

Adding a flush-on-read (or flush-on-whatever-read-calls) and having it work nested at the end of an update is probably necessary because command listeners are wrapped in an implicit update but I think a lot of code doesn't really consider that and sets up their own read contexts.

I've implemented flush-on-read in the latest commit.

StyleT (approved):
LGTM!

etrepum:
I updated the branch and the tests are passing again (I guess main was broken earlier this week)

ivailop7 (approved):

zurfyx (approved):

PR #6355, opened 2024-06-30
[lexical-playground] Table Hover Action Buttons

ivailop7:
Recreate the Experimental Table Hover Action Buttons for the Classic Table as a standalone playground plugin.
Clean up some of the remnants from the original implementation that was deleted a year ago.

https://github.com/facebook/lexical/assets/7893468/cf12e22b-687d-4988-a2f5-2ec251178b9c

### DIFF
--- a/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
@@ -14,6 +14,7 @@ import {
   $insertTableRow__EXPERIMENTAL,
   $isTableCellNode,
   $isTableNode,
+  TableCellNode,
   TableNode,
   TableRowNode,
 } from '@lexical/table';
@@ -25,7 +26,6 @@ import {createPortal} from 'react-dom';
 
 import {useDebounce} from '../CodeActionMenuPlugin/utils';
 
-const BUTTON_SPACE_PX = 25;
 const BUTTON_WIDTH_PX = 20;
 
 function TableHoverActionsContainer({
@@ -44,7 +44,13 @@ function TableHoverActionsContainer({
 
   const debouncedOnMouseMove = useDebounce(
     (event: MouseEvent) => {
-      const {tableDOMNode} = getMouseInfo(event);
+      const {isOutside, tableDOMNode} = getMouseInfo(event);
+
+      if (isOutside) {
+        setShownRow(false);
+        setShownColumn(false);
+        return;
+      }
 
       if (!tableDOMNode) {
         return;
@@ -52,9 +58,10 @@ function TableHoverActionsContainer({
 
       tableDOMNodeRef.current = tableDOMNode;
 
-      let hoveredRowNode: TableNode | null = null;
-      let hoveredColumnNode: TableNode | null = null;
-      let tableDOMElement;
+      let hoveredRowNode: TableCellNode | null = null;
+      let hoveredColumnNode: TableCellNode | null = null;
+      let tableDOMElement: HTMLElement | null = null;
+
       editor.update(() => {
         const maybeTableCell = $getNearestNodeFromDOMNode(tableDOMNode);
 
@@ -63,11 +70,9 @@ function TableHoverActionsContainer({
             $isTableNode(node),
           );
           if (!table) {
-            setShownRow(false);
-            setShownColumn(false);
-            hoveredRowNode = null;
             return;
           }
+
           tableDOMElement = editor.getElementByKey(table?.getKey());
 
           if (tableDOMElement) {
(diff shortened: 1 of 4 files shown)


packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+      let tableDOMElement: HTMLElement | null = null;
+
+      editor.update(() => {
+        const maybeTableCell = $getNearestNodeFromDOMNode(tableDOMNode);
+
+        if ($isTableCellNode(maybeTableCell)) {
+          const table = $findMatchingParent(maybeTableCell, (node) =>
+            $isTableNode(node),
+          );
+          if (!table) {
```
zurfyx, line 72: ⭐️ ...
```suggestion
          if (!$isTableNode(table)) {
```

packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+            return;
+          }
+
+          tableDOMElement = editor.getElementByKey(table?.getKey());
+
+          if (tableDOMElement) {
+            const rowCount = (table as TableNode).getChildrenSize();
```
zurfyx, line 79: ... ⭐️ some other cases below as well (mostly a nit tho)
```suggestion
            const rowCount = table.getChildrenSize();
```

packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+    if (!shouldListenMouseMove) {
+      return;
+    }
+
+    document.addEventListener('mousemove', debouncedOnMouseMove);
+
+    return () => {
+      setShownRow(false);
+      setShownColumn(false);
+      debouncedOnMouseMove.cancel();
```
zurfyx, line 142: [Opinionated] debounce is fine, I would've chosen throttle

packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+      target.closest<HTMLElement>(
+        'button.PlaygroundEditorTheme__tableAddRows',
+      ) ||
+      target.closest<HTMLElement>(
+        'button.PlaygroundEditorTheme__tableAddColumns',
+      ) ||
+      target.closest<HTMLElement>('div.TableCellResizer__resizer')
+    );
+
+    return {isOutside, tableDOMNode};
```
zurfyx, line 228: I would've considered using `getNearestNodeFromDOMNode`, nothing beats the performance of your implementation but this method will already do the editor checks for you and will work regardless of the markup.

packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+    return () => {
+      setShownRow(false);
+      setShownColumn(false);
+      debouncedOnMouseMove.cancel();
+      document.removeEventListener('mousemove', debouncedOnMouseMove);
+    };
+  }, [shouldListenMouseMove, debouncedOnMouseMove]);
+
+  editor.registerMutationListener(TableNode, (mutations) => {
```
zurfyx, line 147: Should we wrap this with a `useEffect` and `return` it to make sure there's no memory leaks?

packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+          tableDOMNodeRef.current,
+        );
+        maybeTableNode?.selectEnd();
+        if (insertRow) {
+          $insertTableRow__EXPERIMENTAL();
+          setShownRow(false);
+        } else {
+          $insertTableColumn__EXPERIMENTAL();
+          setShownColumn(false);
+        }
```
zurfyx, line 181: No strong opinions. Your version is easy to read
```suggestion
        $insertTableColumn__EXPERIMENTAL();
        setShownColumn(insertRow);
```

zurfyx (commented):
Looks beautiful! Left some nits

Also, we may want to look into the case where you move from last column to last row or viceversa, it doesn't seem to show in this case

https://github.com/facebook/lexical/assets/193447/14d63701-4296-4fb7-83e6-db8dff3ca942

packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
ivailop7: This doesn't work since it's different methods it's setShownRow and setShownColumn, not setShownColumn at all times.

ivailop7:
> Looks beautiful! Left some nits
> 
> Also, we may want to look into the case where you move from last column to last row or viceversa, it doesn't seem to show in this case
> 
>  Screen.Recording.2024-07-01.at.10.41.01.AM.mov

yeah, hovering over the bottom right cell, should show both, but right now only does add row, as a higher priority, running on the assumption that 'add row' would be much more popular than 'add column' in a every day use. It's good enough IMO in the current state, but open for people polishing the nits.

packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
ivailop7: done

ivailop7:
@zurfyx for a second look post addressing comments

zurfyx (approved):

PR #6357, opened 2024-07-01
[lexical] Feature: registerMutationListener should initialize its existing nodes

### LINKED ISSUE #6349: Bug: The extended TableNode does not work correctly.
Lexical version: 0.16.0

## Steps To Reproduce

1. create a new table node like:
```ts
export class MyTableNode extends TableNode {
  static getType(): string {
    return 'my-table';
  }
...
}
```
2. it will break all table functions since the `LexicalTablePlugin` not work correct at the line 135 in bottom image, `tableNodes` will be empty, so that the TableObserver will not be initialized.
image

### LINKED ISSUE #6356: Feature: registerMutationListener should be called with pre-existing DOM
## Description

See also https://github.com/facebook/lexical/issues/6349#issuecomment-2197536502

Typically when `registerMutationListener` is called, code is not written to check for any initial state. However, it's possible and sometimes desirable to do this lazily (e.g. lazy initialization of plug-ins, plug-ins registered after the initial state is rendered, etc.). Other listener features have this type of support, e.g. `registerNodeTransform` marks all nodes as dirty, `registerRootListener` immediately calls its listener, etc.

The rare case when you may not want this to happen can be handled with an additional option. One such case might be the `TableOfContentsPlugin`, which already handles lazy initialization correctly and has an efficient implementation. I haven't done any benchmarking on the pair of `HeadingNode` and `TextNode` mutation listeners that implement the continuous updates, but it's clearly much worse than a single traversal of the root's children.

## Impact

It will be easier to correctly implement mutation listeners correctly. You can be certain that every lexical node in the DOM has a corresponding `'created`' either when the mutation listener was registered, or when it was initially reconciled.

For performance sensitive use cases where this is not desired, a `skipInitialization` option will be available.

etrepum:
## Description

When `registerMutationListener` is called, if there are any existing nodes with that type, then the listener should immediately be called once with any existing nodes in the DOM from the previously reconciled state.

This makes it easier to implement mutation listeners correctly, especially for plug-ins that may be loaded lazily, and mirrors the behavior for `registerRootListener` and `registerNodeTransform` where it is expected that the code will be run for existing content (though there are other types listeners that do not have this behavior, I think it really makes the most sense for mutation listeners to behave this way). Some existing mutation listeners either did not support lazy loading, or did this incorrectly (e.g. by checking for an exact node type match instead of considering node replacement as in #6349).

This new "synthetic" mutation can be detected from the listener by checking for the `registerMutationListener` update tag, and it can be skipped entirely with the new `{skipInitialization: true}` option argument (mutation listeners registered in this way will have the same semantics as before this PR).

Closes #6356
Closes #6349

## Test plan

Unit tests are included. Only one existing unit test needed to be changed to use the new `{skipInitialization: true}` option.

I did have to fix an egregious react rules violation in [CodeActionMenuPlugin](https://github.com/facebook/lexical/pull/6357/commits/c73a70a6f526f4be6cee43081e3eb2560d3bffef) that caused an infinite rendering loop with this change, the registration needed to be wrapped in `useEffect`. The previous behavior was basically a memory & performance leak, fortunately only in playground code.


zurfyx (commented):
Thanks Bob! This is intentional:
1. Conceptually Nodes aren't "created" when you register the function.
2. The listeners weren't meant to be called immediately upon registration (the Root listener was meant to be a one-off).

Note that this is also a dangerous breaking change, the API is backward compatible but the behavior is not. Internally, we commonly pair this function with a Map to extend the behavior of the Node independently and these now repeated events can cause trouble with the initialization.

https://github.com/facebook/lexical/pull/4099 is a potential option to mitigate the issue. I think it is reasonable to have this build-in via a property (don't have strong opinions on 2.) but at the same time I'd advocate for a more flat API instead of yet another configuration object.

Would like to hear what other think about this as well.

etrepum:
`$nodesOfType` is not good for this use case because it only considers exact type, not node replacement by subclass, so it creates a different initial set than this version of `registerMutationListener`. If `$nodesOfType` worked I wouldn't really have been motivated to fix `registerMutationListener` this way.

etrepum:
Other than the number of calls, the breaking API change is unlikely to cause problems because it will just emit an initial created event where previously you had to do a separate read to discover them and this was only implemented correctly one time in the monorepo (it was implemented wrong in two places, and not implemented at all in several other places). Implementing `registerMutationListener` in this way removes the need for #4099 altogether, making the API simpler, at the cost of a mostly harmless breaking API change.

Alternatively the new behavior could be added with an option, or a separate function/method, but I would prefer the breaking change because it makes it much easier to write correct code and will probably even fix a few latent bugs.

fantactuka:
> 1. Conceptually Nodes aren't "created" when you register the function.
> 2. The listeners weren't meant to be called immediately upon registration (the Root listener was meant to be a one-off).

It depends on how editor's state is initialized. If `initialEditorState` is async then mutation listener will be initialized before it's executed and called for initial notes. It's likely less common use case though

> Note that this is also a dangerous breaking change, the API is backward compatible but the behavior is not. Internally, we commonly pair this function with a Map to extend the behavior of the Node independently and these now repeated events can cause trouble with the initialization.

+1, I did check registerMutationListener + $nodesOfType (commonly used to handle initial nodes) internally and it relies on nodeID which should avoid double initialization, but there could be usages that just go and mutate some DOM for 'created' mutation without tracking nodeID at all

fantactuka:
Re: `registerMutationListener` vs `registerNodesOfType`, I do agree with Bob that keeping simpler API is better here. Can think about rollout plan to simplify migration out of nodesOfType+registerMutationListener (e.g., making it opt-in for patch version bumb, and opt-out for minor)

etrepum:
An opt-in migration path sounds fine to me, we can use the existing option argument for that and explicitly use `{skipInitialization: false}` to get the new behavior now. The only place I was motivated to use `{skipInitialization: true}` in the monorepo was the TOC plugin (the one correct init-before-register implementation) and only because it is tracking state for multiple mutation listeners in one place so it would've been awkward to initialize it efficiently with the pair of initial calls.

zurfyx:
> `$nodesOfType` is not good for this use case because it only considers exact type, not node replacement by subclass, so it creates a different initial set than this version of `registerMutationListener`. If `$nodesOfType` worked I wouldn't really have been motivated to fix `registerMutationListener` this way.

This is not intentional, this function should look into replacements like we do elsewhere.

> Re: `registerMutationListener` vs `registerNodesOfType`, I do agree with Bob that keeping simpler API is better here. Can think about rollout plan to simplify migration out of nodesOfType+registerMutationListener (e.g., making it opt-in for patch version bumb, and opt-out for minor (or v1 if we do it soon enough))

@fantactuka Can you expand on what a simpler API means for you? In either case, it would be a one-liner. 

The concern I raised about the current implementation is that previously created nodes are also reported as `created`, and the (not-so-important) fact that we need to introduce another object to carry more conditionals once more.

etrepum:
The `registerNodesOfType` utility from #4099 is useful, but for different use cases (sometimes you do care about update or destroy, rather then tracking a single set). In fact, it would be even simpler to implement `registerNodesOfTypeListener` with this version of `registerMutationListener` because you can get rid of the separate initialization code.

```ts
export function registerNodesOfTypeListener(
  editor: LexicalEditor,
  klass: Klass<LexicalNode>,
  listener: (nodes: Set<NodeKey>) => void,
  fireImmediately: void | boolean = false,
): () => void {
  let nodes = new Set<NodeKey>();
  return editor.registerMutationListener(klass, (mutations, {updateTags}) => {
    let newNodes: Set<NodeKey> | null = null;
    for (const [key, mutation] of mutations) {
      const created = mutation === 'created';
      const destroyed = mutation === 'destroyed';
      if (created || destroyed) {
        if (newNodes === null) {
          newNodes = new Set(nodes);
        }
        if (created) {
          newNodes.add(key);
        }
        if (destroyed) {
          newNodes.delete(key);
        }
      }
    }
    if (newNodes !== null) {
      nodes = newNodes;
      if (fireImmediately || !updateTags.has('registerMutationListener')) {
        listener(nodes);
      }
    }
  }, {skipInitialization: false});
}
```

Other than a potential complexity/performance trade-off of tracking previously created nodes separately (which is opt-in here), what are the use cases for having a mutation listener that ignores initial creates? I think in most cases the existing code is just wrong (doesn't support lazy loading at all or doesn't correctly track existing nodes), and if it is right then it is probably already has "defenses" to work around the fact that it is not seeing a complete picture of the DOM nodes (e.g. tracking nodes in a separate map so that an update can be considered to be a create) so the worst case is usually just doing extra work (which would've happened in the TOC plug-in if I didn't opt-in to `skipInitialization`) or breaking a very sensitive test (e.g. tracking the number of calls). There was one case where it exposed a bug (`CodeActionsMenu` calling register outside of an effect with no clean-up!).

etrepum:
FWIW I don't think `$nodesOfType` can be fixed without taking an `editor` argument or depending on an `activeEditor` from `editor.update` (or #6347's `editor.read`) since the transform information isn't available on the `EditorState`.

etrepum:
I've added an internal cache (`getCachedTypeToNodeMap`) for the expensive calculation so that unnecessary work is not repeated when multiple mutation listeners are added in sequence without a reconciliation between them. This cache is also used by `$nodesOfType` when called with a read-only `activeEditorState`.

I've also changed the default to `skipInitialization: true` for backwards compatibility with notes in the documentation that this default will change in 0.17.0.

packages/lexical-playground/src/plugins/CodeActionMenuPlugin/index.tsx
```diff
@@ -109,26 +109,32 @@ function CodeActionMenuContainer({
...
+                break;
+
+              default:
+                break;
+            }
+          }
+        });
+        setShouldListenMouseMove(codeSetRef.current.size > 0);
+      },
+      {skipInitialization: false},
```
etrepum, line 134: CodeActionMenuPlugin did not previously handle initialization correctly

packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -818,18 +818,22 @@ export function registerCodeHighlighting(
...
+                updateCodeGutter(node as CodeNode, editor);
+              }
             }
           }
-        }
-      });
-    }),
+        });
+      },
+      {skipInitialization: false},
```
etrepum, line 835: CodeHighlighter did not previously handle initialization correctly

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -183,17 +183,21 @@ function TableActionMenu({
...
+          nodeMutations.get(tableCellNode.getKey()) === 'updated';
+
+        if (nodeUpdated) {
+          editor.getEditorState().read(() => {
+            updateTableCellNode(tableCellNode.getLatest());
+          });
+          setBackgroundColor(currentCellBackgroundColor(editor) || '');
+        }
+      },
+      {skipInitialization: true},
```
etrepum, line 199: TableActionMenuPlugin did handle initialization correctly (it only listens for updates)

packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
```diff
@@ -105,24 +105,23 @@ export function LexicalAutoEmbedPlugin<TEmbedConfig extends EmbedConfig>({
   }, []);
 
   const checkIfLinkNodeIsEmbeddable = useCallback(
-    (key: NodeKey) => {
-      editor.getEditorState().read(async function () {
+    async (key: NodeKey) => {
+      const url = editor.getEditorState().read(function () {
```
etrepum, line 109: This code was very sketchy previously, and only worked at all because it used `linkNode.__url`. We should probably have a lint that tells you not to use async for read or update

packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
```diff
@@ -146,7 +145,9 @@ export function LexicalAutoEmbedPlugin<TEmbedConfig extends EmbedConfig>({
     };
     return mergeRegister(
       ...[LinkNode, AutoLinkNode].map((Klass) =>
-        editor.registerMutationListener(Klass, (...args) => listener(...args)),
+        editor.registerMutationListener(Klass, (...args) => listener(...args), {
+          skipInitialization: true,
```
etrepum, line 149: This implementation is not compatible with `skipInitialization: false` because it resets if it sees `key === nodeKey` which would fire with `'created'`, could also be fixed by checking for `mutation !== 'created' && key === nodeKey`

packages/lexical-react/src/LexicalTableOfContentsPlugin.tsx
```diff
@@ -230,6 +230,8 @@ export function TableOfContentsPlugin({children}: Props): JSX.Element {
           setTableOfContents(currentTableOfContents);
         });
       },
+      // Initialization is handled separately
+      {skipInitialization: true},
```
etrepum, line 234: LexicalTableOfContentsPlugin already had an efficient and correct initialization (though this wouldn't break it)

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -129,17 +128,6 @@ export function TablePlugin({
       }
     };
 
-    // Plugins might be loaded _after_ initial content is set, hence existing table nodes
-    // won't be initialized from mutation[create] listener. Instead doing it here,
-    editor.getEditorState().read(() => {
-      const tableNodes = $nodesOfType(TableNode);
```
etrepum, line 135: This was not correct because `$nodesOfType` can't consider node replacement without depending on the editor

packages/lexical-website/docs/concepts/dom-events.md
```diff
@@ -30,20 +30,20 @@ This can be a simple, efficient way to handle some use cases, since it's not nec
 
 ## 2. Directly Attach Handlers
 
-In some cases, it may be better to attach an event handler directly to the underlying DOM node of each specific node. With this approach, you generally don't need to filter the event target in the handler, which can make it a bit simpler. It will also guarantee that you're handler isn't running for events that you don't care about. This approach is implemented via a [Mutation Listener](https://lexical.dev/docs/concepts/listeners).
+In some cases, it may be better to attach an event handler directly to the underlying DOM node of each specific node. With this approach, you generally don't need to filter the event target in the handler, which can make it a bit simpler. It will also guarantee that your handler isn't running for events that you don't care about. This approach is implemented via a [Mutation Listener](https://lexical.dev/docs/concepts/listeners).
 
 ```js
+const registeredElements: WeakSet<HTMLElement> = new WeakSet();
```
etrepum, line 36: Just fixing the correctness and formatting of this relevant example, the state needs to be outside of the listener implementation!

packages/lexical/src/LexicalEditor.ts
```diff
@@ -825,15 +838,43 @@ export class LexicalEditor {
...
+      this.getRegisteredNode(klass),
+    ).klass;
+    const mutations = this._listeners.mutation;
+    mutations.set(listener, klassToMutate);
+    const skipInitialization = options && options.skipInitialization;
+    if (
+      !(skipInitialization === undefined
+        ? DEFAULT_SKIP_INITIALIZATION
+        : skipInitialization)
+    ) {
```
etrepum, line 866: Would be `const skipInitialization = options?.skipInitialization ?? DEFAULT_SKIP_INITIALIZATION;` if we could use null coalescing

StyleT (approved):
LGTM!

etrepum:
I updated the branch and the tests are passing again (I guess main was broken earlier this week)

packages/lexical-code/src/CodeHighlighter.ts
ivailop7: nice catch

ivailop7:
> I updated the branch and the tests are passing again (I guess main was broken earlier this week)

LGTM, could you check the new TableHoverActions plugin I added as well for rules violation as well, I copied a lot of from the CodeActions plugin and I'm on Mobile right now.

etrepum:
@ivailop7 good catch, the new `TableHoverActions` plugin did require `{skipInitialization: false}` for correct lazy initialization.

ivailop7:
> @ivailop7 good catch, the new `TableHoverActions` plugin did require `{skipInitialization: false}` for correct lazy initialization.

Thanks!

ivailop7:
Leaving to @zurfyx to merge

packages/lexical/src/LexicalEditor.ts
```diff
@@ -842,49 +883,51 @@ export class LexicalEditor {
...
+      klass.getType(),
+    );
+    if (!nodeMap) {
+      return;
+    }
+    const nodeMutationMap = new Map<string, NodeMutation>();
+    for (const k of nodeMap.keys()) {
+      nodeMutationMap.set(k, 'created');
+    }
+    if (nodeMutationMap.size > 0) {
```
zurfyx, line 915: For the sake of predictability I wonder if we should always call it regardless. This is sort of inline with my [previous comment](https://github.com/facebook/lexical/pull/6357#issuecomment-2200445030) but less radical, I understand that libraries like Rxjs follow similar patterns.

⭐️ This predicatibility also gives users more flexibility and the option flag becomes more of a convenience rather than a must-have.

packages/lexical-website/docs/concepts/listeners.md
```diff
@@ -81,15 +81,21 @@ Get notified when a specific type of Lexical node has been mutated. There are th
...
+If any existing nodes are in the DOM, and skipInitialization is not true, the listener
+will be called immediately with an updateTag of 'registerMutationListener' where all
+nodes have the 'created' NodeMutation. This can be controlled with the skipInitialization option
+(default is currently true for backwards compatibility in 0.16.x but will change to false in 0.17.0).
+
 ```js
 const removeMutationListener = editor.registerMutationListener(
   MyCustomNode,
-  (mutatedNodes) => {
+  (mutatedNodes, { updateTags, dirtyLeaves, prevEditorState }) => {
```
zurfyx, line 92: Is this correct?

packages/lexical/src/LexicalEditor.ts
```diff
@@ -211,6 +212,18 @@ export type MutatedNodes = Map<Klass<LexicalNode>, Map<NodeKey, NodeMutation>>;
...
 export type NodeMutation = 'created' | 'updated' | 'destroyed';
 
+export interface MutationListenerOptions {
+  /**
+   * Skip the initial call of the listener with pre-existing DOM nodes.
+   *
+   * The default is currently true for backwards compatibility with <= 0.16.1
+   * but this default is expected to change to false in 0.17.0.
+   */
+  skipInitialization?: boolean;
```
zurfyx, line 222: Is this the best term for users? Initialization describes the work done behind the users, for users it's merely skipping the first one. Do we even need? ⭐️

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1118,16 +1118,21 @@ export function setMutatedNode(
...
+    return nodes ? [...nodes.values()] : [];
+  }
   const nodes = editorState._nodeMap;
   const nodesOfType: Array<T> = [];
   for (const [, node] of nodes) {
     if (
       node instanceof klass &&
       node.__type === klassType &&
-      (readOnly || node.isAttached())
+      node.isAttached()
```
zurfyx, line 1135: I think this is correct. I believe I added the `readOnly` flag before the GC fix

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1691,3 +1696,34 @@ export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
 export function $getEditor(): LexicalEditor {
   return getActiveEditor();
 }
+
+/** @internal */
+export type TypeTodNodeMap = Map<string, NodeMap>;
```
zurfyx, line 1701: Nit, typo
```suggestion
export type TypedToNodeMap = Map<string, NodeMap>;
```

packages/lexical/src/LexicalEditor.ts
etrepum: I don't have a strong preference one way or the other, correct code won't notice a difference. The current zero check is "more" backwards compatible in that the behavior only changes if the mutation is registered after the initial reconciliation, so I think more tests will end up having to change a bit (the ones based on call count rather than meaningful substance). Since we ended up with an option and a deprecation phase I think either is fine. Anyone else have an opinion about this?

packages/lexical-website/docs/concepts/listeners.md
etrepum: Yes, as far as I can tell
```ts
export type MutationListener = (
  nodes: Map<NodeKey, NodeMutation>,
  payload: {
    updateTags: Set<string>;
    dirtyLeaves: Set<string>;
    prevEditorState: EditorState;
  },
) => void;
```

packages/lexical/src/LexicalEditor.ts
etrepum: We need an option because on the call it was discussed that we need the default behavior to be exactly the current behavior. When the default changes, it should be very rare to have to specify this option, which is why it should have a name that implies that the default is false. I don't have a strong preference as to whether it's called "skipInitialization" or "doNotSendCreateEventForExistingDOM" or whatever.

packages/lexical/src/LexicalUtils.ts
etrepum: `readOnly` can never happen here anymore because there's an early exit for that case above

```ts
  if (editorState._readOnly) {
    const nodes = getCachedTypeToNodeMap(editorState).get(klassType) as
      | undefined
      | Map<string, T>;
    return nodes ? [...nodes.values()] : [];
  }
```

packages/lexical/src/LexicalUtils.ts
etrepum: Good catch! Looks like I didn't finish a refactor from one naming convention to another *and* had a typo in there.

zurfyx (approved):
Stamping, sorry for the back and forth, let's maybe take some time next session to discuss the nits that are relevant for the version bump. Merging as soon as the tests pass since @ivailop7 and @StyleT have already given their thumb up

PR #6358, opened 2024-07-01
[Lexical] Fix broken sync due to deprecation of ReactDOMComet to ReactDOM

Sahejkm:
## WHAT
replace mapping for react-dom to ReactDOM from ReactDOMComet

## WHY
ReactDOMComet was deprecated 5 days ago to ReactDOM in Meta intern builds


etrepum (commented):
LGTM - shouldn't have any effect on OSS

potatowagon (approved):

PR #6361, opened 2024-07-01
[lexical-playground][lexical-poll] Bug Fix: Fixes undefined context inside Poll add option

romanWebiya:
When lexical parse PollNode, you get error `Cannot get getWritable of undefined`.
The reason is that `this` is equal to undefined because of passing addOption function as an argument.

## Test plan
Add a poll to initial state
Get error on init, no poll content inside


fantactuka (approved):

ivailop7:
@romanWebiya could you sign the CLA please. Thanks!

PR #6367, opened 2024-07-02
[lexical] Bug Fix: Merge pasted paragraph into empty quote

### LINKED ISSUE #5108: Bug: Pasting into an empty quote node converts it to a paragraph
Lexical version: v0.12.2

## Steps To Reproduce

1. Open https://playground.lexical.dev
2. Copy some text from elsewhere in the playground editor contents
3. Start a new quote using either <code>> </code> or the menu
4. Paste your copied text

## The current behavior

The quote formatting is lost and replaced with a plain paragraph.

https://github.com/facebook/lexical/assets/415/7d8b7703-a7c6-49f5-bb3a-52303d272f99

## The expected behavior

The quote formatting is kept when pasting.

2wheeh:
## Description

It adds handling for pasting paragraph into an empty quote node.

Closes #5108

### Before

https://github.com/facebook/lexical/assets/40269597/c0f404db-f816-4d28-b0ae-db48e4985a9d

### After

https://github.com/facebook/lexical/assets/40269597/a0c90ce2-239a-49cf-aa76-17bc2c1f1f2a


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1257,12 +1257,13 @@ export class RangeSelection implements BaseSelection {
     const blocks = blocksParent.getChildren();
     const isLI = (node: LexicalNode) =>
       '__value' in node && '__checked' in node;
+    const isQuote = (node: LexicalNode) => node.__type === 'quote';
     const isMergeable = (node: LexicalNode): node is ElementNode =>
```
etrepum, line 1261: It seems like this would probably be better off as a method of ElementNode so that it can be implemented correctly without these fragile checks (the `__type` check will fail under node replacement)

2wheeh: Thank you and sorry for the late reply.

What I understood:

Using `getType()` or `__type` check is going to fail under node replacement since the replacement node might have some thing like `custom-quote` on `__type`.


> as a method of ElementNode

Currently, `QuoteNode` doesn't have unique internal field such as `__value` or `__checked` of `ListItemNode`.

So, you suggested to use a method of ElementNode.

does it mean one of these?

1. adding new method on `ElementNode`:
```ts
export class ElementNode extends LexicalNode {
  isQuote(): boolean {
    return false;
  }
}

export class QuoteNode extends ElementNode {
  isQuote(): true {
    return true;
  }
}

const isQuote = (node: ElementNode) => node.isQuote() === true;
```

2. adding new method (or internal field) on `QuoteNode`:
```ts

export class QuoteNode extends ElementNode {
 /** @internal*/
 __quoteFlag = true;

 // or
  isQuote(): true {
    return true;
  }
}

const isQuote =  (node: LexicalNode) => '__quoteFlag' in node;
// or
const isQuote = (node: LexicalNode) => 'isQuote' in node && node.isQuote() === true;
```

if none of these are what you meant, would you elaborate some more?

2wheeh: Thank you for kind explanation. I added `canMergeWhenEmpty` method.

2wheeh:
integration test failure is not related to this PR: https://github.com/facebook/lexical/issues/6413

StyleT:
Hi! 

I think we lack documentation here. Feel free to add JSDoc + section to https://lexical.dev/docs/concepts/nodes about this new method

StyleT:
@GermanJablo PTAL as discussed.

Afterthought: seems like call of the `canMergeWhenEmpty` shall not happen within `CASE 3: At least 1 element of the array is not inline` to solve issue with quote (which is not a list).

So more information & research is needed here...

acywatson (approved):

acywatson (changes requested):
Sorry I approved the wrong PR

2wheeh:
@StyleT 

> Afterthought: seems like call of the canMergeWhenEmpty shall not happen within CASE 3: At least 1 element of the array is not inline to solve issue with quote (which is not a list).

Based on my understanding, `CASE 3: At least 1 element of the array is not inline` handles cases where inserted nodes contain non-inline elements such as Paragraphs.

The call of `canMergeWhenEmpty` determines whether an inserted non-inline node is mergeable with `firstBlock` when `firstBlock` (which can be a ListItem, Quote, etc.) is empty.

`isMergeable` returns true when `firstBlock` is not empty or when `firstBlock` can merge even if it's empty:
```ts
const isMergeable = (node: LexicalNode): node is ElementNode =>
  $isElementNode(node) &&
  INTERNAL_$isBlock(node) &&
  !node.isEmpty() &&
  $isElementNode(firstBlock) &&
  (!firstBlock.isEmpty() || firstBlock.canMergeWhenEmpty());

if (isMergeable(firstToInsert)) {
  // merge
}
```

Currently, we only check if `firstBlock` is a ListItem with the call of `isLI`. To fix this issue with empty Quote, we need Quote (or other elements) to also merge inserted nodes when it is empty.

That's why `canMergeWhenEmpty` is introduced as a generalized method. I don't see any reason why Case 3 should be limited to ListItems only.

If I've misunderstood your comment, could you please elaborate further?

GermanJablo:
- I'm ok with a solution like canMergeWhenEmpty. My only observation is that isLi is still used in one more place. If it is replaced by canMergeWhenEmpty, does any test fail?
- Do we agree that the correct behavior should be that of the video "after" (or that of the new test)? I mean, there is also the possibility that the 2 paragraphs are inserted within the quote. I don't know how GDocs and Word do it.

2wheeh:
@GermanJablo 

> My only observation is that isLi is still used in one more place. If it is replaced by canMergeWhenEmpty, does any test fail?

It doesn't fail.

But the other `isLI` call is to remove the ListItem inserted by `insertNewAfter` at the end. I'm not sure replacing this with `canMergeWhenEmpty` would be semantically correct.

Instead, we could check if `insertedParagraph` has same type as `lastInsertedBlock`:
```diff
- isLI(insertedParagraph) || INTERNAL_$isBlock(lastToInsert)
+ insertedParagraph.__type === lastInsertedBlock.__type || INTERNAL_$isBlock(lastToInsert)
```

or we could introduce new generalized method for this behavior, which could be useful for the nodes that might want to keep the inserted Paragraph from `insertNewAfter`.

> I don't know how GDocs and Word do it.

AFAIK, GDocs and Word don't have a dedicated "block quote" format in the way we (or other block based editors such as Notion) do. Instead, they just visually mimic the appearance of a block quote by a combination of formatting options.

At least, Notion behaves like the video "after" but yes, it needs to agree.

GermanJablo:
1. It seems fine to me.
2. My intuition tells me that both should be included in the quote, although I don't have strong opinions on the matter. Maybe someone else wants to contribute their opinion.

2wheeh:
@GermanJablo 

1. done it.
2. If it's just about quote, more dedicated fix would be needed here.
    Otherwise ListItem insertion behavior changes as well:

https://github.com/user-attachments/assets/5783fae3-256c-4454-a10c-ff13fe257f2a

GermanJablo:
LGTM

acywatson (dismissed):
Looks to be in order - thank you!

StyleT (dismissed):
LGTM! 

Although I would really ask to add documentation as follow up here

etrepum:
I created #6443 to track the missing docs

2wheeh:
@etrepum @StyleT Thank you. I added jsdoc here.

2wheeh:
Failure of `e2e-tests / prod (macos-latest) ...` is due to #6463

etrepum (approved):
LGTM!

PR #6368, opened 2024-07-04
[lexical-yjs] init collabNode in nested composer with existing state

ololoken:
## Description
When nested composer has state and connects to collab binding's collabNodeMap stays empty. As result synchronization with remote is not fired.


StyleT:
Hi! 

Can you pls elaborate on the issue? How to reproduce it now?

And would it be possible to cover fix with tests to prevent regression?

etrepum:
Without tests and/or some kind of demonstration of the issue that this intends to solve it's hard to accept. Please re-open when and if you intend to provide enough details for this to be properly reviewed.

PR #6375, opened 2024-07-06
[lexical-playground] Fix Table Hover Actions Noclick Bug

ivailop7:
Had an annoying bug in the Table Hover Actions plugin, where when one hovers over one of the buttons but then clicks the other one, the state was not correct. Fixed.

Before:

https://github.com/facebook/lexical/assets/7893468/2e7f6948-2bf0-4122-ac4f-7ad5374166ef

After: 

https://github.com/facebook/lexical/assets/7893468/5b74cee0-ba2b-407c-b6d9-24803c79403a


potatowagon (approved):
lgtm

Issue #5592, opened 2024-02-07
Bug: Pasting plain text into RichTextPlugin does not preserve line breaks


chrysb:
Lexical version: 0.13.1

## Steps To Reproduce

1. Copy text from a normal textarea that has line multiple line breaks
2. Paste into Lexical with a RichTextPlugin

```
    <LexicalComposer
      key='editor'
      initialConfig={{
        ...editorConfig,
        editorState: defaultValue,
      }}
    >
      <RichTextPlugin
        contentEditable={
          <ContentEditable className='outline-none w-full leading-7 h-full dark:text-gray-300 text-gray-800' />
        }
        placeholder={<Placeholder />}
        ErrorBoundary={LexicalErrorBoundary}
      />
      <HistoryPlugin />
      <ListPlugin />
      <LinkPlugin />
      <ListMaxIndentLevelPlugin maxDepth={2} />
      <MarkdownShortcutPlugin transformers={TRANSFORMERS} />
      <AutoLinkPlugin />
    </LexicalComposer>
```

## The current behavior

```
Sentence once. Sentence two.
```

## The expected behavior
```
Sentence one.

Sentence two.
```

Liakhov:
I have the same issue.

And also have an issue, during pasting HTML, for example:

**const htmlData = dataTransfer.getData('text/html');**
**htmlData:**
`<meta charset='utf-8'><div>first paragraph</div><div>first paragraph</div>`. - in this case we will have 1 paragraph without line breaks.

The correct line breaks I can get only if `htmlData` consists of  `p` and `span` tags for the HTML.

lukasb:
I just reproduced this bug with 0.14.2. Updated to 0.16.0 and it went away.

jnnwnk:
I've got a similar issue. I copy text from google docs (which results in data transfer of type `text/html`). The HTML looks like this (I removed styling attributes for better overview):

```html
<html>
<body>
    <!--StartFragment-->
    <meta charset="utf-8">
    <b>
    	<br />
        <p dir="ltr">Test1</span><span><br /></span><span>Test2</span></p>
        <p dir="ltr">Test3</span></p>
    </b><br class="Apple-interchange-newline">
    <!--EndFragment-->
</body>
</html>
```

The first line break (before `p` tag) is recognized correctly.
The second line break (which is embedded between Test1 and Test2 in the first paragraph) gets lost.

Any idea how to fix this? I just updated to version 0.16.0 but the issue still exists.
[...]

PR #6376, opened 2024-07-07
[lexical][lexical-playground] Bug Fix: Create line break on paste of content type text/html

jnnwnk:
### Current behaviour

On copy past of text with content type `text/html` (e.g. from google docs), line breaks inside of paragraphs are ignored. The HTML looks like this (I removed styling attributes for better overview):

```html
<html>
<body>
    <!--StartFragment-->
    <meta charset="utf-8">
    <b>
    	<br />
        <p dir="ltr">Test1</span><span><br /></span><span>Test2</span></p>
        <p dir="ltr">Test3</span></p>
    </b><br class="Apple-interchange-newline">
    <!--EndFragment-->
</body>
</html>
```

The first line break (before p tag) is recognized correctly.
The second line break (which is embedded between Test1 and Test2 in the first paragraph) gets lost.

### Expected / Fixed behaviour

The line break is inserted correctly, no matter where it was placed.

Closes [#5592](https://github.com/facebook/lexical/issues/5592)

### Before

https://github.com/facebook/lexical/assets/10629407/f1482eec-b412-4308-b441-5d386d82a807

### After

https://github.com/facebook/lexical/assets/10629407/d537a8f3-a04e-4912-9785-3eea963e12e4

### DIFF
--- a/packages/lexical/src/nodes/LexicalLineBreakNode.ts
+++ b/packages/lexical/src/nodes/LexicalLineBreakNode.ts
@@ -14,6 +14,7 @@ import type {
   SerializedLexicalNode,
 } from '../LexicalNode';
 
+import {DOM_TEXT_TYPE} from '../LexicalConstants';
 import {LexicalNode} from '../LexicalNode';
 import {$applyNodeReplacement} from '../LexicalUtils';
 
@@ -48,7 +49,10 @@ export class LineBreakNode extends LexicalNode {
 
   static importDOM(): DOMConversionMap | null {
     return {
-      br: () => {
+      br: (node: Node) => {
+        if (isOnlyChild(node)) {
+          return null;
+        }
         return {
           conversion: $convertLineBreakElement,
           priority: 0,
@@ -71,7 +75,7 @@ export class LineBreakNode extends LexicalNode {
   }
 }
 
-function $convertLineBreakElement(): DOMConversionOutput {
+function $convertLineBreakElement(node: Node): DOMConversionOutput {
   return {node: $createLineBreakNode()};
 }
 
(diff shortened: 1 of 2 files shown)


jnnwnk:
Hi guys, this is my first contribution to an open source project of this size.
Please let me know if I need to add anything further or what needs to be done so that the change can be merged.

potatowagon:
tested with these cases
```
<ol>
  <li>
    1<br />3
  </li>
</ol>
```
```
 1<br />3
```
they are working

potatowagon (approved):

packages/lexical/src/nodes/LexicalLineBreakNode.ts
```diff
@@ -89,9 +89,9 @@ export function $isLineBreakNode(
   return node instanceof LineBreakNode;
 }
 
-function isOnlyChild(node: Node): boolean {
+function isOnlyChildInParagraph(node: Node): boolean {
   const parentElement = node.parentElement;
-  if (parentElement !== null) {
+  if (parentElement !== null && parentElement.tagName === 'P') {
```
potatowagon, line 94: maybe we could generalise this further to not convert all only child line breaks in block nodes, cause eg.

```
1
<div>
<br />
</div>
3
```
becomes

```
1


3
```
when it should be 
```
1

3
```

doing on all block nodes would need more testing, ill look into it in a seperate PR.

meanwhile this change is an upgrade for lexical so i would be happy to merge once all the tests pass :)

jnnwnk: @potatowagon Are there any tests I need to fix/add?

The checks look good for me, except for `After Approval` which was skipped for any reason? If there is anything I can do to make this mergable, let me know :)

PR #6384, opened 2024-07-09
Prettier sort test attributes

zurfyx:
Turns out our `format` utility for tests was never sorting the attributes, which is needed when comparing against the output produced by the Jest test.

For some reason the test environment doesn't see `.prettierrc.js` and I have no idea why, but there's an alternatively way where you can just import the same config programatically.

_Solves the test failures in https://github.com/facebook/lexical/pull/6381_


fantactuka (dismissed):

acywatson:
lint?

since when do we fail on warnings?

zurfyx:
> lint?
> 
> since when do we fail on warnings?

You're right, I misread that, it's a prettier issue and it's obvious from the error message...

Sahejkm (approved):

PR #6389, opened 2024-07-10
Fix clear rootElement on React

zurfyx:
When we introduced the `if` condition in https://github.com/facebook/lexical/pull/5070 we missed the `null` case.

### DIFF
--- a/packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
+++ b/packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
@@ -972,7 +972,7 @@ describe('LexicalEditor tests', () => {
           [editor] = useLexicalComposerContext();
 
           useEffect(() => {
-            editor.registerRootListener(listener);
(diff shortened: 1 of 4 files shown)


packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -1011,7 +1011,7 @@ describe('LexicalEditor tests', () => {
         await Promise.resolve().then();
       });
 
-      expect(listener).toHaveBeenCalledTimes(3);
+      expect(listener).toHaveBeenCalledTimes(4);
```
zurfyx, line 1014: No idea why it's calling the listener one more time with the HTML reference, and for some reason the useEffect without dependencies will also be unmounted defeating the purpose of this test. If anyone knows more about `ReactTestUtils.act` I'm all ears

StyleT (approved):
LGTM!

ivailop7 (approved):

PR #6392, opened 2024-07-11
[lexical-html] Feature: support pasting empty block nodes

potatowagon:
## Description
An empty block node (eg. `<div></div>`) in between 2 inline nodes in html places the next inline node on the next line, acting as a linebreak. lexical currently doesnt handle pasting empty block nodes, so there is a missing line break where the empty block node should be.

see test plan for demo.

test cases are added as unit tests 

## Test plan
```
<div>
  1
  <div></div>
  2
</div>
```
Screenshot 2024-07-12 at 11 56 58 AM

### Before

Screenshot 2024-07-12 at 11 56 48 AM

### After

Screenshot 2024-07-12 at 11 58 12 AM

---

```
<ol>
  <li>1
    <div></div>
    2
  </li>
  <li>
  </li>
  <li>3</li>
</ol>
```
Screenshot 2024-07-12 at 12 05 57 PM

### Before

Screenshot 2024-07-12 at 12 09 06 PM

### After

Screenshot 2024-07-12 at 12 09 00 PM

### DIFF
--- a/packages/lexical/src/__tests__/unit/HTMLCopyAndPaste.test.ts
+++ b/packages/lexical/src/__tests__/unit/HTMLCopyAndPaste.test.ts
@@ -52,25 +52,6 @@ describe('HTMLCopyAndPaste tests', () => {
             </div>`,
         },
         {
-          expectedHTML: `<p dir="ltr"><span data-lexical-text="true">a</span></p><p dir="ltr"><span data-lexical-text="true">b b</span></p><p dir="ltr"><span data-lexical-text="true">c</span></p><p dir="ltr"><span data-lexical-text="true">z </span></p><p dir="ltr"><span data-lexical-text="true">d e </span></p><p dir="ltr"><span data-lexical-text="true">fg</span></p>`,
-          name: 'nested divs',
-          pastedHTML: `<div>
-            a
-            <div>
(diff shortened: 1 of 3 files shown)


packages/lexical/src/__tests__/unit/HTMLCopyAndPaste.test.ts
```diff
@@ -51,25 +51,6 @@ describe('HTMLCopyAndPaste tests', () => {
...
-                <div>
-                  <div></div>
-                  z
-                </div>
-              </div>
-              d e
-            </div>
-            fg
-          </div>`,
-        },
```
potatowagon, line 72: moved this to packages/lexical-utils/src/__tests__/unit/LexicalEventHelpers.test.tsx onPasteFromRichText

because an extra `<br />` is introduced in this test setup which i cant repro on the playground

potatowagon: codepen
Screenshot 2024-07-12 at 11 30 30 AM

lexical playground

Screenshot 2024-07-12 at 11 30 51 AM

freddymeta (approved):
lgtm

PR #6397, opened 2024-07-12
[lexical-react][lexical-playground] sync draggable block plugin to www

potatowagon:
## Description

- abstract out the ui components from draggable block plugin so we can customise the ui for menu and target line

- currently sticking with original implementation which depends on the ui component to be a html element to be compatible with ref. will explore (in another PR) if its possible to make this more flexible so we can pass in any react component without having to maintain the ref

Closes #

### Before

Draggable block plugin works

### After

https://github.com/user-attachments/assets/4ee44fad-e30d-4440-8857-a5cce476153e

no regressions, draggable block plugin still works


packages/lexical-react/src/shared/rect.ts
```diff
@@ -17,7 +17,7 @@ type ContainsPointReturn = {
   };
 };
 
-export class Rect {
+export class Rectangle {
```
potatowagon, line 20: rename because w3c bundle has a Rect class already and clash in naming cause build to fail

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
```diff
@@ -7,431 +7,37 @@
...
-        <div className={isEditable ? 'icon' : ''} />
-      </div>
-      <div className="draggable-block-target-line" ref={targetLineRef} />
-    </>,
-    anchorElem,
-  );
-}
-
-export default function DraggableBlockPlugin({
+export default function PlaygroundDraggableBlockPlugin({
```
zurfyx, line 19:
```suggestion
export default function DraggableBlockPlugin({
```

zurfyx (approved):
LGTM, I would consider suffixing the component to `__EXPERIMENTAL` while we polish the UI since it will give us more flexibility to change the API

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
potatowagon: cant rename that to DraggableBlockPlugin as thats the name of the base plugin

Screenshot 2024-07-17 at 12 06 46 PM

potatowagon:
> LGTM, I would consider suffixing the component to `__EXPERIMENTAL` while we polish the UI since it will give us more flexibility to change the API

ok added

Sahejkm (approved):

potatowagon:
failure from flaky test

PR #6403, opened 2024-07-15
Restore registerRootListener null call

zurfyx:
Restores the null call I removed in #6389, unfortunately it's a breaking change. Products are using this to do the teardown. For example:

```
if (prevElement !== null) {
    prevElement.removeEventListener('click', handleClick);
    prevElement.removeEventListener('pointerdown', handlePointerDown);
  }
```

We may need to introduce a new function/signature and always migrate to it. If we were to mimic React this would be intuitive and wouldn't need to return a fake `null` `rootElement`:

```
registerRootListener(prevElement, nextElement) {
  
  return () => {
    if (prevElement !== null) {
      prevElement.removeEventListener(...);
  }
```

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -768,6 +768,7 @@ export class LexicalEditor {
     listener(this._rootElement, null);
     listenerSetOrMap.add(listener);
     return () => {
+      listener(null, this._rootElement);
       listenerSetOrMap.delete(listener);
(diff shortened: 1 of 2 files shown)


ivailop7 (approved):
that didn't last long

zurfyx:
> that didn't last long

@ivailop7 haha right, but it's worth pointing out I didn't revert the other PR fully, the nullable condition still remains

PR #6411, opened 2024-07-17
CI: run flaky tests on firefox browsers

potatowagon:
## Description

to see if putting flaky test in a less crowded run helps with the flakiness since most of the flaky tests are on firefox

## Test plan

wait for CI to run, there should be a FF job for flaky test

### DIFF
--- a/.github/workflows/call-e2e-all-tests.yml
+++ b/.github/workflows/call-e2e-all-tests.yml
@@ -146,8 +146,7 @@ jobs:
     strategy:
       matrix:
         node-version: [18.18.0]
-        # Currently using single browser & os combination for flaky tests to reduce cost impact
-        browser: ['chromium']
+        browser: ['chromium', 'firefox']
         editor-mode: ['rich-text', 'plain-text', 'rich-text-with-collab']
         events-mode: ['modern-events']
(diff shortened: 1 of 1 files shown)


Sahejkm (approved):

.github/workflows/call-e2e-all-tests.yml
```diff
@@ -146,8 +146,7 @@ jobs:
     strategy:
       matrix:
         node-version: [18.18.0]
-        # Currently using single browser & os combination for flaky tests to reduce cost impact
-        browser: ['chromium']
+        browser: ['chromium', 'firefox']
```
Sahejkm, line 149: Should we use all browsers ? Not just Firefox

potatowagon: sorry just saw this, the other is webkit, we can see how it goes with adding firefox. just alittle weary there will be alot of flaky jobs (since its 3 per browser)

Sahejkm: I think the workflow anyways run separately for all browsers after isolation, so running 2 or 3 browser will not make a difference, but the objective was to see if flakyness appears in isolation or not and fix it ..

PR #6419, opened 2024-07-18
Fix discrete nested updates

zurfyx:
If the `discrete` flag was part of a nested update it was being ignored. Instead, the correct behavior is that this update and the previously computed updates should be flushed.

### DIFF
--- a/packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
+++ b/packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
@@ -21,6 +21,7 @@ import {
   $createLineBreakNode,
   $createNodeSelection,
   $createParagraphNode,
+  $createRangeSelection,
   $createTextNode,
   $getEditor,
   $getNearestNodeFromDOMNode,
@@ -62,6 +63,7 @@ import {
   $createTestElementNode,
   $createTestInlineElementNode,
   createTestEditor,
+  createTestHeadlessEditor,
   TestComposer,
   TestTextNode,
 } from '../utils';
@@ -307,9 +309,11 @@ describe('LexicalEditor tests', () => {
 
   it('Should handle nested updates in the correct sequence', async () => {
     init();
+    const onUpdate = jest.fn();
 
     let log: Array<string> = [];
 
+    editor.registerUpdateListener(onUpdate);
     editor.update(() => {
       const root = $getRoot();
       const paragraph = $createParagraphNode();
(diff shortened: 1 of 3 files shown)


packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -2373,6 +2400,90 @@ describe('LexicalEditor tests', () => {
...
+    );
+
+    const textContent = headless
+      .getEditorState()
+      .read(() => $getRoot().getTextContent());
+    expect(textContent).toBe('Async update\n\nSync update');
+    expect(onUpdate).toHaveBeenCalledTimes(1);
+  });
+
+  it('can use discrete after a non-discrete setEditorState to flush the entire queue', () => {
```
acywatson, line 2430: Is flushing the entire queue correct or should we flush the pending updates, then flush the new update separately? I'm worried about leaking flags and such.

zurfyx: Ah no, the labeling is incorrect, the behavior is correct

acywatson (approved):

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
zurfyx: @acywatson fair point, this is correct as far as the current behavior goes. If you had a `discrete` update with nested non-discrete updates, the nested updates would also be bundled into the discrete. I think this is worth an offline discussion, clearly nested updates and discrete are very unpredictable.

etrepum: Last time I looked at this I couldn't think of a way to do it correctly with the current API.

✅ probably ok, because the outer update doesn't do anything on its own (common in a command listener)
```ts
editor.update(() => {
  editor.update(() => $changeSomeStuff(), { discrete: true });
  // this is fine assuming reconciliation already happened, in which case the rest of the update is more like a read
});
```

✅ probably ok, because the discrete update is a "tail call" so it would be equivalent to floating the discrete flag upwards
```ts
editor.update(() => {
  $changeSomeStuff();
  editor.update(() => $changeSomeStuff(), { discrete: true });
  // this is fine assuming reconciliation already happened, in which case the rest of the update is more like a read
});
```

❌ this breaks assumptions about update atomicity
```ts
editor.update(() => {
  $changeSomeStuff();
  editor.update(() => $changeSomeStuff(), { discrete: true });
  // oh no! this will be a separate logical update or the discrete update actually hasn't been reconciled yet
  $changeSomeStuff();
});
```

etrepum: I think the "most correct" workaround for this would be to have some way to mark the outer update as read-only after flush. I think any solution (or non-solution) here is going to be a bit convoluted.

PR #6420, opened 2024-07-18
[lexical-history][lexical-selection][lexical-react] Fix: #6409 TextNode change detection

### LINKED ISSUE #6409: Bug: history should work with custom TextNodes containing non-standard properties
I created my own `TextNode` to have random css class support via the added property `__className` and some methods like `addClass`. While trying it out via a button in my toolbar I realized that the editor wouldn't allow me to undo this change, while all other changes work as expected, checked lexical source code, playground code, issues and searched discord until I asked directly on [Discord](https://discord.com/channels/953974421008293909/1262877639471923272) and after poking around a little @etrepum discovered the hardcoded [`isTextNodeUnchanged`](https://github.com/facebook/lexical/blob/cad2861343c1cffdfdb27bd5aed59198f43dc32f/packages/lexical-history/src/index.ts#L186-L218) helper function which won't consider custom properties.

I think it would be nice to move `isTextNodeUnchanged` as a method maybe `isUnchanged` (or maybe `isChanged` or `hasChanged` or `hasChanges`) into TextNode so subclasses can override it when they need to support custom properties. Then the history plugin would call that method [here](https://github.com/facebook/lexical/blob/cad2861343c1cffdfdb27bd5aed59198f43dc32f/packages/lexical-history/src/index.ts#L301) instead of the hardcoded function.

This method might even live in `LexicalNode` so other nodes could also make use of it, tho I don't have a direct need for this right now I think.

Lexical version: 0.16.1

## Steps To Reproduce

1. Create a `CustomTextNode` extending `TextNode`
2. Add a non-standard property like `__className`
3. Duplicate a method like `setFormat` from TextNode to set your property from 2
4. trigger this method somehow
5. Realize that undo won't work 😅

My `addClass` method
```tsx
addClass(className: string): this {
	const classes = self.__className?.split(" ").map((s) => s.trim());
	classes.push(className);
	self.__className = classes.join(" ");
	return self;
}
```

The button onClick that calls it (my actual code is a little longer and inspired by [`$patchStyleText`](https://github.com/facebook/lexical/blob/cad2861343c1cffdfdb27bd5aed59198f43dc32f/packages/lexical-selection/src/lexical-node.ts#L338-L478), tho this is enough to reproduce the problem)
```tsx
onClick={() => {
    activeEditor.update(() => {
            const selectedNodes = selection.getNodes();
            if ($isTextNode(selectedNodes[0])) {
                selectedNodes[0].addClass("text-red-500");
            }
        }
    });
}}
```

etrepum:
## Description

* Targeted fix for lexical-history's `TextNode` change detection when only one text leaf is dirty but unchanged
* Add delay option to HistoryPlugin
* Fix `$cloneWithProperties` - the previous implementation did not correctly handle `__textFormat` from `ParagraphNode`. Also updated the very incorrect doc string.
* Clean up lint and test warning issues elsewhere
* Move `$cloneWithProperties` to the lexical package because the correct implementation was already inlined in `LexicalNode.getWritable()`

A more general fix may be possible, it's unclear whether this hack is really necessary for MaxLengthPlugin but other text node transforms may have similar issues so removing the hack could be backwards incompatible.

Closes #6409

### Before

History's TextNode change detection did not work correctly for subclasses where only properties that did not exist on TextNode changed.

### After

The tests show that the behavior is consistent with expectations

### DIFF
--- a/packages/lexical-react/src/LexicalContentEditable.tsx
+++ b/packages/lexical-react/src/LexicalContentEditable.tsx
@@ -15,7 +15,6 @@ import {forwardRef, Ref, useLayoutEffect, useState} from 'react';
 import {ContentEditableElement} from './shared/LexicalContentEditableElement';
 import {useCanShowPlaceholder} from './shared/useCanShowPlaceholder';
 
-/* eslint-disable @typescript-eslint/ban-types */
 export type Props = Omit<ElementProps, 'editor'> & {
   editor__DEPRECATED?: LexicalEditor;
 } & (
@@ -30,8 +29,6 @@ export type Props = Omit<ElementProps, 'editor'> & {
           | JSX.Element;
       }
   );
-
-/* eslint-enable @typescript-eslint/ban-types */
 
 export const ContentEditable = forwardRef(ContentEditableImpl);
 
--- a/packages/lexical-react/src/LexicalHistoryPlugin.ts
+++ b/packages/lexical-react/src/LexicalHistoryPlugin.ts
@@ -17,13 +17,15 @@ export {createEmptyHistoryState} from '@lexical/history';
 export type {HistoryState};
 
 export function HistoryPlugin({
+  delay,
   externalHistoryState,
 }: {
+  delay?: number;
   externalHistoryState?: HistoryState;
 }): null {
   const [editor] = useLexicalComposerContext();
 
-  useHistory(editor, externalHistoryState);
+  useHistory(editor, externalHistoryState, delay);
 
   return null;
 }
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
@@ -2272,42 +2272,44 @@ describe('LexicalSelection tests', () => {
     it('adjust offset for inline elements text formatting', async () => {
       await init();
 
-      await editor!.update(() => {
-        const root = $getRoot();
+      await ReactTestUtils.act(async () => {
+        await editor!.update(() => {
+          const root = $getRoot();
 
-        const text1 = $createTextNode('--');
-        const text2 = $createTextNode('abc');
-        const text3 = $createTextNode('--');
+          const text1 = $createTextNode('--');
+          const text2 = $createTextNode('abc');
+          const text3 = $createTextNode('--');
 
-        root.append(
-          $createParagraphNode().append(
-            text1,
-            $createLinkNode('https://lexical.dev').append(text2),
-            text3,
-          ),
-        );
+          root.append(
+            $createParagraphNode().append(
+              text1,
+              $createLinkNode('https://lexical.dev').append(text2),
+              text3,
+            ),
+          );
 
-        $setAnchorPoint({
-          key: text1.getKey(),
-          offset: 2,
-          type: 'text',
+          $setAnchorPoint({
+            key: text1.getKey(),
+            offset: 2,
+            type: 'text',
+          });
+
+          $setFocusPoint({
+            key: text3.getKey(),
+            offset: 0,
+            type: 'text',
+          });
+
+          const selection = $getSelection();
+
+          if (!$isRangeSelection(selection)) {
+            return;
+          }
+
+          selection.formatText('bold');
+
+          expect(text2.hasFormat('bold')).toBe(true);
         });
-
-        $setFocusPoint({
-          key: text3.getKey(),
-          offset: 0,
-          type: 'text',
-        });
-
-        const selection = $getSelection();
-
-        if (!$isRangeSelection(selection)) {
-          return;
-        }
-
-        selection.formatText('bold');
-
-        expect(text2.hasFormat('bold')).toBe(true);
       });
     });
   });
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -36,37 +36,35 @@ import {
 function $updateElementNodeProperties<T extends ElementNode>(
   target: T,
   source: ElementNode,
-): T {
+): void {
   target.__first = source.__first;
   target.__last = source.__last;
   target.__size = source.__size;
   target.__format = source.__format;
   target.__indent = source.__indent;
   target.__dir = source.__dir;
-  return target;
 }
 
 function $updateTextNodeProperties<T extends TextNode>(
   target: T,
   source: TextNode,
-): T {
+): void {
   target.__format = source.__format;
   target.__style = source.__style;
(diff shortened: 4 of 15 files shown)


packages/lexical-selection/src/lexical-node.ts
```diff
@@ -33,65 +30,6 @@ import {
...
-    return $updateElementNodeProperties(clone, node);
-  }
-
-  if ($isTextNode(node) && $isTextNode(clone)) {
-    return $updateTextNodeProperties(clone, node);
-  }
-
-  if ($isParagraphNode(node) && $isParagraphNode(clone)) {
-    return $updateParagraphNodeProperties(clone, node);
-  }
```
etrepum, line 91: This was wrong because ParagraphNode is always an ElementNode and the exit already happened. This was implemented correctly in LexicalNode.getWritable() so I extracted and exported that implementation instead

packages/lexical-utils/src/index.ts
```diff
@@ -446,12 +446,7 @@ export function $restoreEditorState(
   const activeEditorState = editor._pendingEditorState;
 
   for (const [key, node] of editorState._nodeMap) {
-    const clone = $cloneWithProperties(node);
-    if ($isTextNode(clone)) {
-      invariant($isTextNode(node), 'Expected node be a TextNode');
-      clone.__text = node.__text;
-    }
```
etrepum, line 453: I could not find any possible reason for this code to be here, TextNode.clone already guarantees `clone.__text === node.__text`

packages/lexical-website/src/components/Gallery/utils.tsx
```diff
@@ -51,6 +51,6 @@ export function useFilteredExamples(examples: Array<Example>) {
         searchName,
         tags,
       }),
-    [examples, searchName],
+    [examples, searchName, tags],
```
etrepum, line 54: eslint fix

packages/lexical/src/LexicalNode.ts
```diff
@@ -697,34 +696,12 @@ export class LexicalNode {
...
-      mutableNode.__dir = latestNode.__dir;
-    } else if ($isTextNode(latestNode) && $isTextNode(mutableNode)) {
-      mutableNode.__format = latestNode.__format;
-      mutableNode.__style = latestNode.__style;
-      mutableNode.__mode = latestNode.__mode;
-      mutableNode.__detail = latestNode.__detail;
-    }
+    const mutableNode = $cloneWithProperties(latestNode);
     cloneNotNeeded.add(key);
-    mutableNode.__key = key;
```
etrepum, line 722: This was not necessary because Node.clone must already do this, I added an invariant to guarantee it

packages/lexical-selection/src/__tests__/unit/LexicalSelection.test.tsx
```diff
@@ -2272,42 +2272,44 @@ describe('LexicalSelection tests', () => {
     it('adjust offset for inline elements text formatting', async () => {
       await init();
 
-      await editor!.update(() => {
-        const root = $getRoot();
+      await ReactTestUtils.act(async () => {
```
etrepum, line 2275: Wrapping this is necessary because the update causes a react update

packages/lexical-react/src/LexicalHistoryPlugin.ts
```diff
@@ -17,13 +17,15 @@ export {createEmptyHistoryState} from '@lexical/history';
 export type {HistoryState};
 
 export function HistoryPlugin({
+  delay,
```
etrepum, line 20: Seemed reasonable to fix this missing option while I was in here

packages/lexical-react/src/LexicalContentEditable.tsx
```diff
@@ -15,7 +15,6 @@ import {forwardRef, Ref, useLayoutEffect, useState} from 'react';
 import {ContentEditableElement} from './shared/LexicalContentEditableElement';
 import {useCanShowPlaceholder} from './shared/useCanShowPlaceholder';
 
-/* eslint-disable @typescript-eslint/ban-types */
```
etrepum, line 18: Could not find a reason for this, if there's an internal meta reason we should sync the eslint config to match whatever that rule is

etrepum:
I don't see the typical Lexical Tests checks running on this PR? /cc @potatowagon @Sahejkm

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1727,3 +1733,44 @@ export function getCachedTypeToNodeMap(
...
+    mutableNode.__indent = latestNode.__indent;
+    mutableNode.__format = latestNode.__format;
+    mutableNode.__dir = latestNode.__dir;
+  } else if ($isTextNode(latestNode) && $isTextNode(mutableNode)) {
+    mutableNode.__format = latestNode.__format;
+    mutableNode.__style = latestNode.__style;
+    mutableNode.__mode = latestNode.__mode;
+    mutableNode.__detail = latestNode.__detail;
+  }
+  if (__DEV__) {
```
etrepum, line 1767: I moved this to a `__DEV__` because it's a new invariant in a relatively hot code path but we might want to run it unconditionally?

potatowagon:
> I don't see the typical Lexical Tests checks running on this PR? /cc @potatowagon @Sahejkm

thats strange, why are the jobs skipped, will investigate

potatowagon:
it seems like the core test runs are missing

potatowagon:
> it seems like the core test runs are missing

think its because of this: https://github.com/facebook/lexical/blob/666ccb2e4a20cd7d453d5b255f52426a954242df/.github/workflows/tests.yml#L12

could be its ignoring core test trigger because this PR modifies the lexical website path. will fix it

potatowagon:
> > it seems like the core test runs are missing
> 
> think its because of this:
> 
> https://github.com/facebook/lexical/blob/666ccb2e4a20cd7d453d5b255f52426a954242df/.github/workflows/tests.yml#L12
> 
> could be its ignoring core test trigger because this PR modifies the lexical website path. will fix it

fixed in https://github.com/facebook/lexical/pull/6423

potatowagon:
> > > it seems like the core test runs are missing
> > 
> > 
> > think its because of this:
> > https://github.com/facebook/lexical/blob/666ccb2e4a20cd7d453d5b255f52426a954242df/.github/workflows/tests.yml#L12
> > 
> > could be its ignoring core test trigger because this PR modifies the lexical website path. will fix it
> 
> fixed in #6423

prolly not the reason since core tests are triggered on https://github.com/facebook/lexical/pull/6425

potatowagon (approved):
thankyou for the fix!

ivailop7 (approved):

PR #6423, opened 2024-07-19
CI: remove paths-ignore

potatowagon:
## Description
when a PR modifies any path in paths-ignore, even if the pr has other changes from paths, the whole PR gets ignored from core test runs

see example: https://github.com/facebook/lexical/pull/6420

### Before

PR does not get test runs

### After

PR gets test runs

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,12 +4,8 @@ on:
   push:
     branches:
       - main
-    paths-ignore:
-      - 'packages/lexical-website/**'
   pull_request:
     types: [opened, synchronize, reopened]
-    paths-ignore:
-      - 'packages/lexical-website/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


Sahejkm:
it actually worked fine in https://github.com/facebook/lexical/pull/6357 , are we sure about the reason

potatowagon:
> * it actually worked fine in [[lexical] Feature: registerMutationListener should initialize its existing nodes #6357](https://github.com/facebook/lexical/pull/6357) , are we sure about the reason
> 
> the check to ignore lexical-website was before change from nicholas and seemed fair as website changes shudn't trigger test

m not sure, need to test, based on the docs

Screenshot 2024-07-19 at 4 21 34 PM

it maybe ?

i think no harm landing this first just to see if it unblocks. i have another wip PR to test out using the `paths:` option and specify paths to include so hopefully PR that modify both website and other paths will trigger the tests

potatowagon:
prolly not the reason since: https://github.com/facebook/lexical/pull/6425

PR #6426, opened 2024-07-19
[lexical-react] update flow typing for draggable block plugin

potatowagon:
## Description
update flow typing

## Test plan

works fine

Screenshot 2024-07-19 at 4 50 25 PM


packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
```diff
@@ -16,7 +16,7 @@ function isOnMenu(element: HTMLElement): boolean {
   return !!element.closest(`.${DRAGGABLE_BLOCK_MENU_CLASSNAME}`);
 }
 
-export default function PlaygroundDraggableBlockPlugin({
+export default function DraggableBlockPlugin({
```
potatowagon, line 19: to match file name

zurfyx (approved):

PR #6428, opened 2024-07-19
[lexical][lexical-selection] Bug Fix: Respect mode when patching text style

### LINKED ISSUE #6427: Bug: Text mode is not respected when patching text style
Lexical version: 0.16.1

## Steps To Reproduce

1. In the playground, mention somebody `@Droidbait` (basically, create a MentionNode, this will be in `segmented` mode)
2. Select some text of the node partially partially Dr<ins>**oid**</ins>bait.
3. Change the font size
4. The text node gets split up, even though it is in segmented mode. Expected behaviour would be to increase the font size of the entire node. 
 

## The current behavior
Token and segmented nodes are improperly split, causing inconsistencies in the text appearance and structure.

## The expected behavior
The entire token/segmented node should be patched correctly without being split, maintaining the integrity of the node.

adrianmxb:
## Description

There is an issue in Lexical where token and segmented text nodes are not correctly handled during text styling operations. This results in improper formatting / node splitting, leading to unexpected behaviour.

After some debugging, I saw that it is handled correctly in the `RangeSelection` class, but not in the helper method used to patch the text style.
https://github.com/facebook/lexical/blob/2c90ee08e6cb9a073ed4cb7a9fca3c8f273d9f47/packages/lexical/src/LexicalSelection.ts#L1136-L1182

Closes #6427 

1. Added `$isTokenOrSegmented` to the list of exported utilities.
2. Updated `$patchStyleText` function: Adjusted logic to prevent splitting of token/segmented nodes

### DIFF
--- a/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
+++ b/packages/lexical-selection/src/__tests__/unit/LexicalSelectionHelpers.test.ts
@@ -30,6 +30,7 @@ import {
   LexicalNode,
   ParagraphNode,
   RangeSelection,
+  TextModeType,
   TextNode,
 } from 'lexical';
 import {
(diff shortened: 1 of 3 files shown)


ivailop7:
Could we get a test for this please

adrianmxb:
@ivailop7 I added a unit test for it

ivailop7:
> @ivailop7 I added a unit test for it

Thanks, have ran all the e2e once they finish, will give it a final look before merging.

PR #6433, opened 2024-07-19
[lexical][auto-link] Fix auto link crash editor

### LINKED ISSUE #6430: Bug: AutoLink crashes the editor
Type in the playground.lexical.dev: 
abc@abc.abc.abc and it crashes the editor

MaxPlav:
Fixed bug introduced in https://github.com/facebook/lexical/pull/6146. Improving the way a potential top-level domain is parsed for email autolinks. Adding more e2e tests, apparently the email auto-link use case wasn't covered by tests.

Closes #6430

### Before

https://github.com/user-attachments/assets/41384f51-a2c5-489b-b95d-d2d9e13dce55

### After

https://github.com/user-attachments/assets/bcbfd089-a103-40db-bd50-45a3bc3be3ac

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -276,6 +276,16 @@ export class LinkNode extends ElementNode {
       selection.getTextContent().length > 0
     );
   }
+
+  isEmailURI(): boolean {
+    return this.__url.startsWith('mailto:');
+  }
+
+  isWebSiteURI(): boolean {
+    return (
+      this.__url.startsWith('https://') || this.__url.startsWith('http://')
+    );
+  }
 }
 
 function $convertAnchorElement(domNode: Node): DOMConversionOutput {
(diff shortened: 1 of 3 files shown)


etrepum:
It would probably be a good idea to add a unit test suite for this sort of thing, e2e is a bit overkill to test a regex

ivailop7:
I agree with Bob here, these should be perfectly fine to be unit tests, otherwise LGTM.

MaxPlav:
hi @etrepum @ivailop7 thanks for the review, the e2e tests for the Autolink node have existed for quite a while (added probably here https://github.com/facebook/lexical/pull/1156) but they didn’t cover *all use cases* specifically for the Autolink node created for a recognized email pattern. The way the text is transformed to an Autolink node is spread in multiple functions, it’s a bit complicated and not easy to troubleshoot. In this diff (besides the bug fix) I’m just covering all possible positive and negative use cases for the whole autolink creation/update flow transformed from a text *email* pattern. So I would suggest keeping the e2e tests as it might be not sufficient to cover the functions with unit tests. :) wdyt?

PR #6434, opened 2024-07-20
[lexical-playground] Feature: Store lexical editor on the window for quick inspection

tom-sherman:
## Description

I've frequently found myself hunting through React Dev Tools to find the editor in state and store it on a global so that I can dig into the internals.

This would save me a few clicks 😅

## Test plan

Tested locally in dev mode.


etrepum:
```ts
lexicalEditor = document.querySelector('[data-lexical-editor]').__lexicalEditor
```

tom-sherman:
Oh, TIL! Is that documented?

etrepum:
Documented? No. Reliable? Yes, this is what the lexical dev tools does.

```ts
export default function queryLexicalNodes(): LexicalHTMLElement[] {
  return Array.from(
    document.querySelectorAll('div[data-lexical-editor]'),
  ).filter(isLexicalNode);
}

export function isLexicalNode(
  node: LexicalHTMLElement | Element,
): node is LexicalHTMLElement {
  return (node as LexicalHTMLElement).__lexicalEditor !== undefined;
}
```

PR #6435, opened 2024-07-21
[lexical-rich-text] Bug Fix: HeadingNode.insertNewAfter

GermanJablo:
## Description

If a paragraph is inserted when the selection is at the end of a heading, a paragraph should be inserted.

However, the logic for detecting whether the selection was at the end of the heading was incorrect. It did not work if, for example, the heading had multiple TextNodes.


GermanJablo:
Before and After: https://www.loom.com/share/522f8f3bd808464bb07c7019c3abdf36

potatowagon:
i can reproduce the issue, had to create multiple text nodes in heading, eg. bold "Hello" in "Hello world"

packages/lexical-rich-text/src/__tests__/unit/LexicalHeadingNode.test.ts
```diff
@@ -106,6 +106,56 @@ describe('LexicalHeadingNode tests', () => {
...
+
+    test('HeadingNode.insertNewAfter() end', async () => {
+      const {editor} = testEnv;
+      let headingNode: HeadingNode;
+      await editor.update(() => {
+        const root = $getRoot();
+        headingNode = new HeadingNode('h1');
+        const headingTextNode1 = $createTextNode('hello');
+        const headingTextNode2 = $createTextNode(' world');
+        headingTextNode2.setFormat('bold');
```
potatowagon, line 141: 👍

potatowagon (approved):
thanks for the fix!

Issue #5958, opened 2024-04-24
How to keep styles between lines? 


nickschinestzki:
Lexical version: 0.14.5

## The current behavior

As mentioned in #5620, I am looking for a way to keep both formats **and styles** between lines. There was an update in #5822 with logic to persist formats (bold, italic, underline) between paragraphs and it works great. One part of the issue remains that I can't persist CSS styles in new paragraphs.

So in my use case, if the user selects a "body" font family different than the default one, or maybe a different text color, he needs to reapply them in every new paragraph after an empty line (See #5620 to understand this).

## The expected behavior

When user breaks no matter how much lines, the CSS styles applied to a paragraph persist between lines.

## Attempt to fix the issue 

As an attempt to address this, I have created a CustomParagraphNode which is generating a new performance problem. 

```
import { $applyNodeReplacement, ParagraphNode } from 'lexical';
import { $patchStyleText } from '@lexical/selection';

export class CustomParagraphNode extends ParagraphNode {
  constructor(cssStyle, key) {
    super(key);
    this.__cssStyle = cssStyle;
  }
  static getType() {
    return 'custom-paragraph';
  }
  getCssStyle() {
    const self = this.getLatest();
    return self.__cssStyle;
  }
  setCssStyle(style) {
    const self = this.getWritable();
    self.__cssStyle = style;
    return self;
  }

  static clone(node) {
    return new CustomParagraphNode(node.__cssStyle, node.__key);
  }

  // View
  static importJSON(serializedNode) {
    const node = $createCustomParagraphNode(serializedNode.cssStyle);
    node.setFormat(serializedNode.format);
    node.setIndent(serializedNode.indent);
    node.setDirection(serializedNode.direction);
    node.setTextFormat(serializedNode.textFormat);
    node.setCssStyle(serializedNode.cssStyle);
    return node;
  }

  createDOM(config) {
    const element = super.createDOM(config);
    element.style.cssText = this.__cssStyle;
    return element;
  }

  exportJSON() {
    return {
      ...super.exportJSON(),
      cssStyle: this.getCssStyle(),
      textFormat: this.getTextFormat(),
      type: 'custom-paragraph',
      version: 1,
    };
  }

  // Mutation

  insertNewAfter(rangeSelection, restoreSelection) {
    const newElement = $createCustomParagraphNode(rangeSelection.style);
    newElement.setTextFormat(rangeSelection.format);
    newElement.setCssStyle(rangeSelection.style);
    const direction = this.getDirection();
[...]

PR #6437, opened 2024-07-21
[lexical] [lexical-selection] Preserve paragraph styles between lines

ivailop7:
This fixes the bug where we don't preserve style attributes like color, font size, background color or any custom styles when creating a new paragraph (press Enter twice).
Heavily inspired by the solution for the formatting: https://github.com/facebook/lexical/pull/5822
I've just extended in the same fashion for the styling.
Closes https://github.com/facebook/lexical/issues/5958

Before:

https://github.com/user-attachments/assets/cea09e27-d979-4017-b36f-8b3f36518fd8

After:

https://github.com/user-attachments/assets/738cc9c9-0f2f-423d-bb97-9f6a0b0148ab


zurfyx (approved):

anasteele206:
You should be careful of making getStyle and setStyle methods on LexicalElementNode. This can cause conflicts for users who want to upgrade their lexical dependency when they are extending nodes to support importing inline style to them.

PR #6439, opened 2024-07-22
[lexical] Bug Fix(keydown): trigger copy command for non range selection

### LINKED ISSUE #6438: Bug: table can't be copied when only table is present
Lexical version: 0.16.1

A table doesn't get copied to the clipboard when there is only one table as content.

## Steps To Reproduce

1. Visit https://playground.lexical.dev/
2. Insert a table using the `Insert` dropdown.
3. Delete the paragraph nodes, above and below the table.
4. Press `Cmd+A` to select the table.
5. Press `Cmd+C` to copy the table.
6. Press `Backspace` to delete the table.
7. Press `Cmd+V` to trigger the paste action.

Expected: I should see a table on the editor.
Actual: I see random text pasted from the clipboard that I had copied before.

https://github.com/user-attachments/assets/9c0e1447-2fb2-428a-a478-6b08927e0fbd

mrdivyansh:
## Description
The copy and cut command isn't triggered when the selection is `TableSelection` type. Currently, the code checks for `NodeSelection`; however, it should have been supporting custom selection as well. 

Closes #6438

## Test plan
~~I have added an e2e test but the test passes even without my changes. I am not sure why I can't reproduce the bug by automated interaction but manual interaction. I am looking for help on this.~~
[The added test works for Firefox.](https://github.com/facebook/lexical/pull/6439#discussion_r1693946553)

### Before
https://github.com/user-attachments/assets/4d6c10f1-7e68-488f-be9c-1b9b1899e0e7

### After
https://github.com/user-attachments/assets/aa7844a6-2bd4-4dde-92b2-8c393c8771c5


potatowagon:
similar to the test, im unable to reproduce the issue described in https://github.com/facebook/lexical/issues/6438 

https://github.com/user-attachments/assets/fba79ba7-90ad-4af3-85c1-21a4b6436882

followed the same steps listed in the issue

are u using any special setup?

mrdivyansh:
@potatowagon, you missed the 3rd step. I can see a paragraph node (409) before the table node (410) in your demo. You need to ensure there is no other nodes except a table node.

potatowagon:
ok i can repro the issue after including the third step

packages/lexical/src/LexicalEvents.ts
```diff
@@ -1080,7 +1079,7 @@ function onKeyDown(event: KeyboardEvent, editor: LexicalEditor): void {
     dispatchCommand(editor, REDO_COMMAND, undefined);
   } else {
     const prevSelection = editor._editorState._selection;
-    if ($isNodeSelection(prevSelection)) {
+    if (!$isRangeSelection(prevSelection)) {
```
potatowagon, line 1082: have checked that this includes a solo table into the copy cmd, tho whats missing is that the table doesnt get deleted on cut (which is also happening in prod so not an issue introed in this PR)


https://github.com/user-attachments/assets/93812c3e-c141-4914-a6b8-04dc646e0984

packages/lexical-playground/__tests__/regression/6438-copy-table-selection.spec.mjs
```diff
@@ -0,0 +1,58 @@
...
+    await insertTable(page, 5, 5);
+
+    // Ensure no elements are present except the table.
+    // Delete the first empty paragraph
+    await moveToEditorBeginning(page);
+    await page.keyboard.press('Backspace');
+    // Delete the last empty paragraph
+    await moveToEditorEnd(page);
+    await page.keyboard.press('Backspace');
+
```
potatowagon, line 39: >I am not sure why I can't reproduce the bug by automated interaction but manual interaction. 

try adding an assert here to assert that the table isnt wrapped by 2 para nodes

potatowagon: if the assert fails it could be need to add keypress delays before hitting backspace

mrdivyansh: I observed a weird behaviour with the Chrome browser. The copy function works when I intermittently switch to another tab and then return to Lexical to trigger the copy. Perhaps this is the reason the test doesn't fail for Chromium browser.

https://github.com/user-attachments/assets/f0b05423-d49e-4362-a467-481f88c58abc

So, I decided to switch to Firefox with Playwright. 💥! the test failed 🎉.

https://github.com/user-attachments/assets/875d20d2-833c-4bc1-9bbb-19ff33849233

potatowagon: this test is failing alot on the collab e2e suite, could u please help to fix it? if not landing this change will cause the CI to be unstable

possible fixes include adding asserts and keypress delays at backspace

packages/lexical/src/LexicalEvents.ts
potatowagon: hi @zurfyx! checking with u is this one line change safe?

packages/lexical-playground/__tests__/regression/6438-copy-table-selection.spec.mjs
mrdivyansh: I tried with assertion. They were passing. I will check the failing tests.

mrdivyansh:
None of the tests fail on my local machine. Added delays and assertions. It might help.

etrepum:
Pretty sure that this was fixed in #6834

PR #6444, opened 2024-07-22
[lexical] Bug Fix: getCachedTypeToNodeMap should handle a empty and writable EditorState

etrepum:
## Description

Follow-up to #6357

When an editor is still initializing the this._editorState may be writable but will only have the 'root' entry. We need to special case this to prevent an exception if a mutation listener with initialization is registered before the first reconciliation.

## Test plan

The unit test covers this case and will fail without the change

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -903,7 +903,7 @@ export class LexicalEditor {
     klass: Klass<LexicalNode>,
   ): void {
     const prevEditorState = this._editorState;
-    const nodeMap = getCachedTypeToNodeMap(this._editorState).get(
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalEditor.ts
```diff
@@ -903,7 +903,7 @@ export class LexicalEditor {
     klass: Klass<LexicalNode>,
   ): void {
     const prevEditorState = this._editorState;
-    const nodeMap = getCachedTypeToNodeMap(this._editorState).get(
+    const nodeMap = getCachedTypeToNodeMap(prevEditorState).get(
```
etrepum, line 906: not a meaningful change but should optimize better

zurfyx (approved):

PR #6446, opened 2024-07-23
[Lexical][CI] ignore running unit/integerity/e2e tests on examples folder code

Sahejkm:
## Description

- ignore running unit/integerity/e2e tests on examples folder code
- examples are not impacting the real package code and shouldn't be triggering running of e2e/unit tests, similar to website


potatowagon (approved):
hmm iirc integrity runs the npm run ci-check which runs prettier, would that be relevant for website and example code?

Sahejkm:
> hmm iirc integrity runs the npm run ci-check which runs prettier, would that be relevant for website and example code?

yea that would make sense, but right now looks like the ```pretty``` check is coupled with tests workflow, might need to take it out in a separate PR

PR #6456, opened 2024-07-24
[lexical][lexical-overflow] Refactor: simplified removeText and insertText rewrite (part 1)

GermanJablo:
This is something I've been wanting to do since I did the insertNodes rewrite. These are also methods with long and very complex implementations.

Currently, removeText depends on insertText, since removeText is defined as inserting an empty string (`insertText("")`). This PR seeks to reverse the dependency of these methods, making insertText dependent on removeText.

This can be achieved in 2 steps:
1. Rewrite `removeText` without making it dependent on insertText.
2. Rewrite `insertText` with the new removeText.

So far in this PR, I have managed to complete step 1 and 95% of step 2. I decided to leave step 2 for a later PR because there are about 4 tests that fail and, after having dedicated so much time to it, I want to do a pause on this.


packages/lexical-overflow/src/index.ts
```diff
@@ -72,6 +73,19 @@ export class OverflowNode extends ElementNode {
...
+  }
+
+  static transform(): (node: LexicalNode) => void {
+    return (node: LexicalNode) => {
+      invariant($isOverflowNode(node), 'node is not a OverflowNode');
+      if (node.isEmpty()) {
+        node.remove();
+      }
+    };
+  }
```
GermanJablo, line 88: It's likely that `insert/removeText` currently has some logic that prevents this. If it works in all cases, I don't know.

Ensuring that an overFlowNode cannot be empty in lexical-overflow is a more robust and less error-prone solution, and it frees up the load on other packages.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -709,12 +709,56 @@ export class RangeSelection implements BaseSelection {
...
+    // // When composing, we need to adjust the anchor offset so that
+    // // we correctly replace that right range.
+    // if (
+    //   textNode.isComposing() &&
+    //   this.anchor.type === 'text' &&
+    //   anchorNode.getTextContent() !== ''
+    // ) {
+    //   this.anchor.offset -= text.length;
+    // }
+
```
GermanJablo, line 761: Once the failing tests are fixed and this implementation is activated, 364 lines of code can be removed (`$transferStartingElementPointToTextPoint` is only used in the current implementation of `insertText`).

packages/lexical-playground/__tests__/e2e/Mentions.spec.mjs
```diff
@@ -773,55 +741,30 @@ test.describe('Mentions', () => {
...
+            class="mention"
+            style="background-color: rgba(24, 119, 232, 0.2);"
+            data-lexical-text="true">
+            Luke Skywalker
+          </span>
+        </p>
+      `,
+    );
+
     await assertSelection(page, {
```
GermanJablo, line 768: I coincidentally noticed that the `if` and `else` blocks are the same here

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1061,7 +1105,68 @@ export class RangeSelection implements BaseSelection {
...
+    }
+    if ($isTextNode(lastNode)) {
+      lastNode.spliceText(0, lastPoint.offset, '');
+      fixText(lastNode, lastPoint, lastPoint.offset);
+    }
+    if (firstNode.isAttached() && $isTextNode(firstNode)) {
+      firstNode.selectEnd();
+    } else if (lastNode.isAttached() && $isTextNode(lastNode)) {
+      lastNode.selectStart();
+    }
```
etrepum, line 1161: I think this fixes the bugs in here related to fixing up the selection

```suggestion
    const fixText = (
      node: TextNode,
      point: PointType,
      delCount: number,
    ): TextNode | null => {
      if (node.getTextContent() === '') {
        node.remove();
        return null;
      } else if (delCount !== 0 && $isTokenOrSegmented(node)) {
        const textNode = $createTextNode(node.getTextContent());
        textNode.setFormat(node.getFormat());
        textNode.setStyle(node.getStyle());
        point.set(textNode.getKey(), point.offset, 'text');
        return node.replace(textNode);
      } else {
        return node;
      }
    };
    if (firstNode === lastNode && $isTextNode(firstNode)) {
      const delCount = Math.abs(focus.offset - anchor.offset);
      firstNode.spliceText(firstPoint.offset, delCount, '');
      const attachedNode = fixText(firstNode, firstPoint, delCount);
      if (attachedNode) {
        attachedNode.select(firstPoint.offset, firstPoint.offset);
      }
      return;
    }
    if ($isTextNode(lastNode)) {
      lastNode.spliceText(0, lastPoint.offset, '');
      const attachedNode = fixText(lastNode, lastPoint, lastPoint.offset);
      if (attachedNode) {
        attachedNode.selectStart();
      }
    }
    if ($isTextNode(firstNode)) {
      const delCount = firstNode.getTextContentSize() - firstPoint.offset;
      firstNode.spliceText(firstPoint.offset, delCount, '');
      const attachedNode = fixText(firstNode, firstPoint, delCount);
      if (attachedNode) {
        attachedNode.selectEnd();
      }
    }
```

etrepum: Although there's probably another case to handle when both nodes are removed and you need to select the parent

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1061,7 +1105,67 @@ export class RangeSelection implements BaseSelection {
...
+        const textNode = $createTextNode(node.getTextContent());
+        textNode.setFormat(node.getFormat());
+        textNode.setStyle(node.getStyle());
+        return node.replace(textNode);
+      }
+    };
+    if (firstNode === lastNode && $isTextNode(firstNode)) {
+      const del = Math.abs(focus.offset - anchor.offset);
+      firstNode.spliceText(firstPoint.offset, del, '', true);
+      lastNode = fixText(firstNode, del) || lastNode;
```
etrepum, line 1144: An explicit selection here might make sense, since right now this depends on node.replace doing the right thing with the selection under certain circumstances

GermanJablo: Look at lines 1156-9 :)

etrepum: Look at line 1145 😉

packages/lexical/src/LexicalSelection.ts
GermanJablo: I saw your suggestion after I submitted my last fix. I haven't tested that snippet but the current version works. Thanks a lot!

packages/lexical/src/LexicalSelection.ts
GermanJablo: Ah yes, I see what you mean. The variable reassignment here is unnecessary, I've removed it.

Nothing needs to be done with the selection here as replace handles selection correction just fine (I wrote that code).

As a side note, I think fixText should actually be part of spliceText, just like how replace handles selection normalization. Maybe I should do that in another PR.

etrepum: I wrote some unit tests to verify that it does seem to be doing what I expected and they all passed

```diff
diff --git a/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts b/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
index 7055f361a..de1e6cd8e 100644
--- a/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
+++ b/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
@@ -9,15 +9,20 @@
 import {$createLinkNode, $isLinkNode} from '@lexical/link';
 import {
   $createParagraphNode,
+  $createRangeSelection,
   $createTextNode,
   $getRoot,
+  $getSelection,
   $isParagraphNode,
   $isTextNode,
+  $setSelection,
   LexicalEditor,
+  ParagraphNode,
   RangeSelection,
+  TextNode,
 } from 'lexical';
 
-import {initializeUnitTest, invariant} from '../utils';
+import {$assertRangeSelection, initializeUnitTest, invariant} from '../utils';
 
 describe('LexicalSelection tests', () => {
   initializeUnitTest((testEnv) => {
@@ -338,5 +343,156 @@ describe('LexicalSelection tests', () => {
         });
       });
     });
+    describe('removeText', () => {
+      describe('with a leading TextNode and a trailing segmented TextNode', () => {
+        let leadingText: TextNode;
+        let trailingSegmentedText: TextNode;
+        let paragraph: ParagraphNode;
+        beforeEach(() => {
+          testEnv.editor.update(
+            () => {
+              leadingText = $createTextNode('leading text');
+              trailingSegmentedText =
+                $createTextNode('segmented text').setMode('segmented');
+              paragraph = $createParagraphNode().append(
+                leadingText,
+                trailingSegmentedText,
+              );
+              $getRoot().clear().append(paragraph);
+            },
+            {discrete: true},
+          );
+        });
+        test('remove all text', () => {
+          testEnv.editor.update(
+            () => {
+              const sel = $createRangeSelection();
+              sel.anchor.set(leadingText.getKey(), 0, 'text');
+              sel.focus.set(
+                trailingSegmentedText.getKey(),
+                trailingSegmentedText.getTextContentSize(),
+                'text',
+              );
+              $setSelection(sel);
+              sel.removeText();
+              expect(leadingText.isAttached()).toBe(false);
+              expect(trailingSegmentedText.isAttached()).toBe(false);
+              expect($getRoot().getAllTextNodes()).toHaveLength(0);
+              const selection = $assertRangeSelection($getSelection());
+              expect(selection.isCollapsed()).toBe(true);
+              expect(selection.anchor.key).toBe(paragraph.getKey());
+              expect(selection.anchor.offset).toBe(0);
+            },
+            {discrete: true},
+          );
+        });
+        test('remove initial TextNode', () => {
+          testEnv.editor.update(
+            () => {
+              const sel = $createRangeSelection();
+              sel.anchor.set(leadingText.getKey(), 0, 'text');
+              sel.focus.set(
+                leadingText.getKey(),
+                leadingText.getTextContentSize(),
+                'text',
+              );
+              $setSelection(sel);
+              sel.removeText();
+              expect(leadingText.isAttached()).toBe(false);
+              expect(trailingSegmentedText.isAttached()).toBe(true);
+              expect($getRoot().getAllTextNodes()).toHaveLength(1);
+              const selection = $assertRangeSelection($getSelection());
+              expect(selection.isCollapsed()).toBe(true);
+              expect(selection.anchor.key).toBe(trailingSegmentedText.getKey());
+              expect(selection.anchor.offset).toBe(0);
+            },
+            {discrete: true},
+          );
+        });
+        test('remove trailing segmented TextNode', () => {
+          testEnv.editor.update(
+            () => {
+              const sel = $createRangeSelection();
+              sel.anchor.set(trailingSegmentedText.getKey(), 0, 'text');
+              sel.focus.set(
+                trailingSegmentedText.getKey(),
+                trailingSegmentedText.getTextContentSize(),
+                'text',
+              );
+              $setSelection(sel);
+              sel.removeText();
+              expect(leadingText.isAttached()).toBe(true);
+              expect(trailingSegmentedText.isAttached()).toBe(false);
+              expect($getRoot().getAllTextNodes()).toHaveLength(1);
+              const selection = $assertRangeSelection($getSelection());
+              expect(selection.isCollapsed()).toBe(true);
+              expect(selection.anchor.key).toBe(leadingText.getKey());
+              expect(selection.anchor.offset).toBe(
+                leadingText.getTextContentSize(),
+              );
+            },
+            {discrete: true},
+          );
+        });
+        test('remove initial TextNode and partial segmented TextNode', () => {
+          testEnv.editor.update(
+            () => {
+              const sel = $createRangeSelection();
+              sel.anchor.set(leadingText.getKey(), 0, 'text');
+              sel.focus.set(
+                trailingSegmentedText.getKey(),
+                'segmented '.length,
+                'text',
+              );
+              $setSelection(sel);
+              sel.removeText();
+              expect(leadingText.isAttached()).toBe(false);
+              // expecting a new node since it was segmented
+              expect(trailingSegmentedText.isAttached()).toBe(false);
+              const allTextNodes = $getRoot().getAllTextNodes();
+              expect(allTextNodes.map((node) => node.getTextContent())).toEqual(
+                ['text'],
+              );
+              const selection = $assertRangeSelection($getSelection());
+              expect(selection.isCollapsed()).toBe(true);
+              expect(selection.anchor.key).toBe(allTextNodes[0].getKey());
+              expect(selection.anchor.offset).toBe(0);
+            },
+            {discrete: true},
+          );
+        });
+        test('remove partial initial TextNode and partial segmented TextNode', () => {
+          testEnv.editor.update(
+            () => {
+              const sel = $createRangeSelection();
+              sel.anchor.set(leadingText.getKey(), 'lead'.length, 'text');
+              sel.focus.set(
+                trailingSegmentedText.getKey(),
+                'segmented '.length,
+                'text',
+              );
+              $setSelection(sel);
+              sel.removeText();
+              expect(leadingText.isAttached()).toBe(true);
+              expect(trailingSegmentedText.isAttached()).toBe(false);
+              const allTextNodes = $getRoot().getAllTextNodes();
+              // These should get merged in reconciliation
+              expect(allTextNodes.map((node) => node.getTextContent())).toEqual(['lead', 'text']);
+              const selection = $assertRangeSelection($getSelection());
+              expect(selection.isCollapsed()).toBe(true);
+              expect(selection.anchor.key).toBe(leadingText.getKey());
+              expect(selection.anchor.offset).toBe('lead'.length);
+            },
+            {discrete: true},
+          );
+          testEnv.editor.getEditorState().read(() => {
+            const allTextNodes = $getRoot().getAllTextNodes();
+            // These should get merged in reconciliation
+            expect(allTextNodes.map((node) => node.getTextContent())).toEqual(['leadtext']);
+            expect(leadingText.isAttached()).toBe(true);
+          })
+        });
+      });
+    });
   });
 });
```

GermanJablo: Thanks @etrepum. I've already added it. Adding you as a PR contributor is something only maintainers can do, right? If so, please "CC @mantainers"

packages/lexical/src/LexicalSelection.ts
etrepum: Do you expect to add that to this PR, or after this one lands?

etrepum (approved):
LGTM - not sure if you were planning to extend this PR or follow-up afterwards

GermanJablo:
Yes, to be continued in another PR

etrepum:
I think some of the E2E tests are failing, I was hoping that it was just flakiness but it has consistently failed since I turned on extended tests. #6488 is passing (after a flaky test failure) which is based on the same main commit.

potatowagon:
> I think some of the E2E tests are failing, I was hoping that it was just flakiness but it has consistently failed since I turned on extended tests. #6488 is passing (after a flaky test failure) which is based on the same main commit.

yea quite abit of e2e plaintext tests failures that are not known to be flaky: https://github.com/facebook/lexical/actions/runs/10251936357/job/28361149775?pr=6456

etrepum (changes requested):
Waiting for e2e test fixes before merge

GermanJablo:
done @etrepum

etrepum (approved):
Could use another set of eyes on this one since it's such a fundamental change to removeText

ivailop7:
Let's do 0.17.1 first and then do this one as a 0.18 with the defaults change.

packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleTitleNode.ts
```diff
@@ -96,6 +96,18 @@ export class CollapsibleTitleNode extends ElementNode {
     return true;
   }
 
+  static transform(): (node: LexicalNode) => void {
```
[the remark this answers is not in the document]
GermanJablo: Not always. If you remove this you will see that some tests fail because there may be empty CollapsibleTitleNodes.
Apparently, it wasn't necessary because it was one of the many things that `insertText` did.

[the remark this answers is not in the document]
ivailop7: OK, let's update those tests and remove the transform part from the node, since when people copy-paste the Playground it would create more confusion, if it's purely there for tests to pass.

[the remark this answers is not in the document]
etrepum: I think both may be necessary for correct behavior, not just tests. The container won't be empty if it has an empty title since it's measured by children and not text, so the empty title is removed first and then the next pass removes the container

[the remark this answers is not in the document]
GermanJablo: Yes, it's not just about the tests. I don't see anything wrong or confusing here. It's a pattern that other nodes use.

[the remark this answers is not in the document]
ivailop7: That's the part that confuses me, the Layout plugin should be in the same boat by that rule, since it's near identical implementation to the Collapsible Plugin, unless it's purely because we are missing tests for it and the transform method should be added to the Layout plugin as well?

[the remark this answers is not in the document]
GermanJablo: As a general rule I think it's a good idea to use transforms to make sure that nodes that should be used together can't be orphaned.
I don't know what the layout plugin is but I think improving that would be out of scope of this PR.

PS: I updated the branch and there is a flaky test that failed. Can you please restart it?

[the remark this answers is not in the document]
etrepum: I think in general we are missing transforms for a handful of nodes to guarantee correct behavior, e.g. the merge-unmerge issue with TableCellNode not having paragraph children. The static transform method is experimental and new-ish, previously you had to have the correct editor and/or plugin configuration to have all of the required transforms. I would agree that adding other transforms is out of scope for this PR, unless the previous removeText implementation also handled them.

ivailop7 (approved):
Approving, trusting you both you have this planned out. Haven't spent much time getting the whole context here.

adrianmxb:
I am a bit late to the party but the changes here break removal of TextNodes in `token` and `segmented` mode.

CleanShot 2024-10-01 at 15 58 03

GermanJablo:
Isn't that supposedly the correct behavior, or am I missing something? Are you sure it worked differently before? There were a lot of those tests I had to pass.

PR #6458, opened 2024-07-25
[lexical] Bug Fix: Allow getTopLevelElement to return a DecoratorNode

### LINKED ISSUE #6451: Bug: DecoratorNodes no longer allowed at the root?
Previously in the docs (~v0.12) I read that the root node can have either element nodes or decorator nodes. I can't find that in the docs now. I'm now on the latest Lexical v0.16.1, when I click on a decorator node I have at the root I get this error: 
```console
Error: Children of root nodes must be elements
    at <MyDecoratorNode>.getTopLevelElement
```  

Indeed `getTopLevelElement` in `LexicalNode` gives that error  https://github.com/facebook/lexical/blob/ebc8d669fcd7ccc9f0e550be360056787e6722a0/packages/lexical/src/LexicalNode.ts#L356-L359

Lexical version: v0.16.1

## Steps To Reproduce

1. Use the CodeSandbox link below.
2. Click on the chapter marked as '1' in the editor.
3. The error above is displayed.

Link to code example:

https://codesandbox.io/p/sandbox/decorator-in-root-h46skq?file=%2Fsrc%2FEditor.tsx

## The current behavior
The error above is displayed.

## The expected behavior
No error is displayed.

## Impact of fix
It's currently confusing that you can have a `DecoratorNode` at the root but when you try to do something with it produces an error.

If it's really intended that a `DecoratorNode` can't be at the root then when loading editor state it should error if there is a `DecoratorNode` at the root.

etrepum:
## Description

Allows `getTopLevelElement` to return a `DecoratorNode` from `LexicalNode`, and uses declaration merging to return a more specific type when called from an `ElementNode`, `TextNode`, or `DecoratorNode`. I couldn't get it to work in a "cleaner" way than declaration merging.

With TypeScript I was able to get the type to be exactly `this | ElementNode | null` for `DecoratorNode.getTopLevelElement` but with Flow it's just `DecoratorNode<mixed> | ElementNode | null`. Flow claimed that `DecoratorNode<mixed>` wasn't compatible with `DecoratorNode<T>` but it really should be in return position like that, and I don't have enough recent flow experience to track down why or if it's possible to get it to return the more precise result.

This PR also applies the fix at the value and type level to `getTopLevelElementOrThrow`.

Closes #6451

### Before

See #6451

### After

Unit test coverage is included

### DIFF
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -19,6 +19,7 @@ import {
   $isElementNode,
   $isRootNode,
   $isTextNode,
+  type DecoratorNode,
   ElementNode,
 } from '.';
 import {
@@ -376,14 +377,14 @@ export class LexicalNode {
    * non-root ancestor of this node, or null if none is found. See {@link lexical!$isRootOrShadowRoot}
    * for more information on which Elements comprise "roots".
    */
-  getTopLevelElement(): ElementNode | null {
+  getTopLevelElement(): ElementNode | DecoratorNode<unknown> | null {
     let node: ElementNode | this | null = this;
     while (node !== null) {
       const parent: ElementNode | null = node.getParent();
       if ($isRootOrShadowRoot(parent)) {
         invariant(
-          $isElementNode(node),
-          'Children of root nodes must be elements',
+          $isElementNode(node) || (node === this && $isDecoratorNode(node)),
+          'Children of root nodes must be elements or decorators',
         );
         return node;
       }
@@ -397,7 +398,7 @@ export class LexicalNode {
    * non-root ancestor of this node, or throws if none is found. See {@link lexical!$isRootOrShadowRoot}
    * for more information on which Elements comprise "roots".
    */
-  getTopLevelElementOrThrow(): ElementNode {
+  getTopLevelElementOrThrow(): ElementNode | DecoratorNode<unknown> {
     const parent = this.getTopLevelElement();
     if (parent === null) {
       invariant(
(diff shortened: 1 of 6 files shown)


packages/lexical/src/LexicalNode.ts
```diff
@@ -376,14 +377,14 @@ export class LexicalNode {
    * non-root ancestor of this node, or null if none is found. See {@link lexical!$isRootOrShadowRoot}
    * for more information on which Elements comprise "roots".
    */
-  getTopLevelElement(): ElementNode | null {
+  getTopLevelElement(): ElementNode | DecoratorNode<unknown> | null {
```
potatowagon, line 380: https://www.totaltypescript.com/concepts/any-type#any-vs-unknown

new to this, was curious about the choice to use unknown

potatowagon: would unkown in ts be the best mapping to mixed type in flow?

potatowagon (approved):
looks good! will see how this plays out with flow during the sync

packages/lexical/src/LexicalNode.ts
etrepum: In most cases when you use `any` instead of `unknown` you're just saying "I give up and I don't want type safety and/or TypeScript doesn't have the features I need". In this case we don't care what the type parameter of `DecoratorNode` is but we have to specify one so it makes sense to use the least specific but most safe type.

etrepum: but to answer your question, flow's mixed is almost exactly the same thing as typescript's unknown. If you write unknown in flow, the language server will recommend that you use mixed instead.

```
The equivalent of TypeScript's `unknown` type in Flow is `mixed`. Flow(ts-syntax)
```

PR #6464, opened 2024-07-26
[lexical-devtools] The new collapsible node tree

MaxPlav:
This diff is inspired by the react dev tools extension and suggests a lexical node tree for the dev tools extension mentioned in 
https://github.com/facebook/lexical/issues/5675. The nodes tree is built on top of https://github.com/dgreene1/react-accessible-treeview. Currently, the tree can expand and collapse on a user click and hightlight the appropriate node in the lexical editor. The selected lexical node details and lexical selection details are displayed on the right panel. 
The availability of the new tree is controlled by the env variable `VITE_COLLAPSIBLE_TREE=1` in `lexical-devtools` package.

Testing on the dev environment:
1. Go to `lexical-devtools` package.
2. Set `VITE_COLLAPSIBLE_TREE=1` in `.env.development` or `.env.development.local`.
3. Run `npm run dev`.

Screenshot 2024-07-26 at 15 14 50

https://github.com/user-attachments/assets/3699139c-0151-482d-bb93-531a99577378


etrepum:
@MaxPlav Is there a reason this is still in draft? What's left to do? It seems a bit odd to make it an environment toggle instead of either using it by default or having a setting in the UI somewhere to turn it on or off, but otherwise at a glance it looks like this might be usable in its current state.

PR #6470, opened 2024-07-27
Hide Table Add Buttons on KeyDown

ivailop7:
Hide the table hover add buttons when the user starts typing, to prevent the button overlap.

Before:

https://github.com/user-attachments/assets/1fcbe68b-9c67-415f-a399-801c06505a26

After:

https://github.com/user-attachments/assets/2a939d15-cab4-4b96-a11f-935ca473740c


ivailop7:
Will come up with a better solution. Closing for now.

PR #6472, opened 2024-07-29
[@lexical/selection] Feature: yield target to style patch fn

mrdivyansh:
## Description
$patchStyleText doesn't expose the target to the CSS property `patch` function. However, it can be very helpful when the next style value may depend on the target node/selection.

## Test plan
Added unit tests

### Before
```ts
function $patchStyleText(
  selection: BaseSelection,
  patch: Record<
    string,
    | string
    | null
    | (( currentStyleValue: string | null) => string)
  >,
): void
```

### After
```ts
function $patchStyleText(
  selection: BaseSelection,
  patch: Record<
    string,
    | string
    | null
    | ((currentStyleValue: string | null, target: TextNode | RangeSelection ) => string)
  >,
): void
```


packages/lexical-selection/src/lexical-node.ts
```diff
@@ -243,16 +243,18 @@ function $patchStyle(
   target: TextNode | RangeSelection,
   patch: Record<
     string,
-    string | null | ((currentStyleValue: string | null) => string)
+    | string
+    | null
+    | ((currentStyleValue: string | null, _target: typeof target) => string)
```
etrepum, line 248: Is there a need to underscore this parameter name? Does it get caught by the eslint shadow rule?

```suggestion
    | ((currentStyleValue: string | null, target: typeof target) => string)
```

mrdivyansh: Yes, shadow rule complains.

Screenshot 2024-08-01 at 5 24 59 PM

etrepum (approved):
LGTM, it seems reasonable to pass this information along and it's great that you've also added tests for this feature! One thing that might be a bit counter-intuitive is that the target is not necessarily the same target that is passed to `$patchStyle` since it will only be a `RangeSelection` when the given selection is collapsed

Issue #6459, opened 2024-07-25
Bug: HorizontalRuleNode breaks TableCell in some case


turnerguo:
Lexical version:

## Steps To Reproduce

1. insert a HorizontalRuleNode in table.
2. select above blank paragraph before HorizontalRuleNode and press delete key
3. the empty paragraph will deleted, keep press delete key the TableCell will deleted too, see the video on detail.

https://github.com/user-attachments/assets/decd08b6-0fe7-4897-9ce2-852f11d45e69

etrepum:
This no longer reproduces in the same way, but there are still some edge cases here because once the divider is deleted you can enter text directly into the TableCellNode, it does not create a wrapping ParagraphNode.

ivailop7:
> This no longer reproduces in the same way, but there are still some edge cases here because once the divider is deleted you can enter text directly into the TableCellNode, it does not create a wrapping ParagraphNode.

I'm no longer able to reproduce this issue, I see even after deleting the horizontal line, it does insert a paragraph. I guess you've added a transform in the big tables rewrite?

etrepum:
yeah https://github.com/facebook/lexical/blob/main/packages/lexical-table/src/LexicalTablePluginHelpers.ts#L68-L76

PR #6486, opened 2024-08-01
[@lexical/playground] fix: block cursor show horizontal

keiseiTi:
## Description

fix lexical PlaygroundEditor Theme  blockCursor show

RE #6459

iShot_2024-08-01_23 11 31

### Before

not need test

### After

not need test

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -1753,8 +1753,8 @@ button.item.dropdown-item-active i {
   display: block;
   position: absolute;
   top: -2px;
-  width: 20px;
-  border-top: 1px solid black;
+  height: 18px;
+  border-left: 1px solid black;
   animation: CursorBlink 1.1s steps(2, start) infinite;
 }
 


potatowagon:
could we have some screenshots to highlight the differences in the before & after please?

keiseiTi:
> could we have some screenshots to highlight the differences in the before & after please?

Before

iShot_2024-08-03_10 47 42

After

iShot_2024-08-01_23 11 31

etrepum (approved):
I changed the PR description because this does not close #6459 (the table cell is still deleted) but it does fix the cursor display

ivailop7:
@etrepum, let's revert this one. The block cursor is intended to be an underline. This now brakes when you have a table as first cell and move the cursor before the table. The reason it only works for the horizontal line is because of the CSS margins.

etrepum:
@ivailop7 It looks like it works to me? This is playground css so if a user has their own css that would make this hard to see (e.g. table border of the same color) they could change it accordingly

https://github.com/user-attachments/assets/1b4895f8-eb9a-4c41-878a-cd2b97b3224f

etrepum:
Another RTE that I could find that supported this state had similar tall vertical cursor behavior next to a table: https://richtexteditor.com/Demos/table-support.aspx

Although this one has an underline cursor above the table
https://prosemirror-tables.netlify.app/

The other editors I could find wouldn't let you get into that selection state

ivailop7:
> @ivailop7 It looks like it works to me? This is playground css so if a user has their own css that would make this hard to see (e.g. table border of the same color) they could change it accordingly
> 
> 
> 
> https://github.com/user-attachments/assets/1b4895f8-eb9a-4c41-878a-cd2b97b3224f
> 
> 
> 
> 

Press up from that position.

keiseiTi:
This is the editor of Yuque (a Chinese document website), and the cursor of the horizontal line in its table looks like this.

I also tried other editor interactions and they all seemed to behave differently. So it depends on how you define the interaction.

iShot_2024-08-05_09 51 08

turnerguo:
https://prosemirror-tables.netlify.app/
I think ProseMirror's cursor intension is much more clear.
For Horizontal Line it should be a horizontal cursor their.

Issue #6481, opened 2024-07-31
Bug: Uncaught Error: Unable to find an active editor state. State helpers or node methods can only be used synchronously during the callback of editor.update() or editorState.read().


aanavadiya03:
Lexical version: 0.16.1
Browser: Chrome latest version
Device: mac 

## Problem 
I'm using the Lexical editor in two separate micro frontend applications. When I attempt to load both applications into the parent frontend, one RTE functions normally, and another one throw errors.

## Steps To Reproduce
It is difficult to produce micro frontend environment in the codesandbox that we're working on, please let me know if any easier way to create micro frontend sandbox.

## The current behavior
When I attempt to load both applications into the parent frontend, one RTE functions normally, while the other produces the following errors when attempting to access it:
```
Uncaught Error: Unable to find an active editor state. State helpers or node methods can only be used synchronously during the callback of editor.update() or editorState.read().
    at getActiveEditorState (Lexical.dev.mjs:7544:13)
    at $getNodeByKey (Lexical.dev.mjs:800:39)
    at _ParagraphNode.getLatest (Lexical.dev.mjs:3774:20)
    at _ParagraphNode.getParent (Lexical.dev.mjs:3445:25)
    at _ParagraphNode.getIndexWithinParent (Lexical.dev.mjs:3425:25)
    at $internalResolveSelectionPoint (Lexical.dev.mjs:6951:39)
    at $internalResolveSelectionPoints (Lexical.dev.mjs:7041:31)
    at $internalCreateRangeSelection (Lexical.dev.mjs:7139:35)
    at $internalCreateSelection (Lexical.dev.mjs:7092:12)
    at $beginUpdate (Lexical.dev.mjs:8088:41)
```

```
Lexical.dev.mjs:7544 Uncaught Error: Unable to find an active editor state. State helpers or node methods can only be used synchronously during the callback of editor.update() or editorState.read().
    at getActiveEditorState (Lexical.dev.mjs:7544:13)
    at $getNodeByKey (Lexical.dev.mjs:800:39)
    at _TextNode.isAttached (Lexical.dev.mjs:3380:20)
    at $isNodeValidForTransform (Lexical.dev.mjs:7582:41)
    at $applyAllTransforms (Lexical.dev.mjs:7624:35)
    at $beginUpdate (Lexical.dev.mjs:8099:9)
    at updateEditor (Lexical.dev.mjs:8168:5)
    at onSelectionChange (Lexical.dev.mjs:2424:3)
[...]

PR #6488, opened 2024-08-03
[lexical] Feature: Add version identifier to LexicalEditor constructor

### LINKED ISSUE #6145: Feature: Add version/build identifier to LexicalEditor
It's useful when debugging, and will presumably be useful for devtools, if details of the lexical build are present in the editor. The `__DEV__` checks could also use this introspection capability to help people debug the dreaded `LexicalComposerContext.useLexicalComposerContext: cannot find a LexicalComposerContext` when there is a bundler configuration issue causing multiple builds of lexical to be present. Possibly something like '0.15.0-cjs-dev' or '0.15.0-esm-prod'. I don't have any strong opinions on the format of the version or build string that is included.

It could be present on the editor as a static, so it would be accessible with `lexicalEditor.constructor.BUILD_IDENTIFIER` or whatever naming convention is decided upon.

I think the three important things are:
* Version
* CJS/ESM
* prod/dev

It wouldn't prevent all possible problems (e.g. a dependency could bundle the same build of lexical not as an external) but I think it would be helpful and have negligible impact to build size or runtime.

etrepum:
## Description

New `LexicalEditor.version` static to help users diagnose installation issues.

Also refactors other `__lexicalEditor` code to avoid conflicts with multiple lexical builds on the page (RE #6481)

Inspecting the lexical editor's version can be done as follows (example assumes exactly one editor on the page):

```
> document.querySelector('[contenteditable]').__lexicalEditor.constructor.version
'0.17.0+dev.esm'
```

It does not have support for coming up with special build identifiers for the versions of lexical we build in CI and for the website that do not correspond to npm versions.

Closes #6145
Closes #5996

## Test plan

There's an integration test that checks to make sure the version is set and accessible from the editor instance

### Before

```
> oldLexical.createEditor().constructor.version
undefined
> lexical.createEditor().constructor.version
'0.17.0+dev.esm'
```

### After

Using https://lexical-playground-git-fork-etrepum-version-a6c103-fbopensource.vercel.app/esm/

Screenshot 2024-08-03 at 4 08 56 PM

### DIFF
--- a/packages/lexical-playground/vite.config.ts
+++ b/packages/lexical-playground/vite.config.ts
@@ -52,6 +52,10 @@ export default defineConfig(({command}) => {
             from: /__DEV__/g,
             to: 'true',
           },
+          {
+            from: 'process.env.LEXICAL_VERSION',
+            to: JSON.stringify(`${process.env.npm_package_version}+git`),
+          },
         ],
       }),
       babel({
--- a/packages/lexical-website/docs/concepts/listeners.md
+++ b/packages/lexical-website/docs/concepts/listeners.md
@@ -84,7 +84,7 @@ handle external UI state and UI features relating to specific types of node.
 If any existing nodes are in the DOM, and skipInitialization is not true, the listener
 will be called immediately with an updateTag of 'registerMutationListener' where all
 nodes have the 'created' NodeMutation. This can be controlled with the skipInitialization option
-(default is currently true for backwards compatibility in 0.16.x but will change to false in 0.17.0).
+(default is currently true for backwards compatibility in 0.17.x but will change to false in 0.18.0).
 
 ```js
 const removeMutationListener = editor.registerMutationListener(
--- a/packages/lexical/src/LexicalUpdates.ts
+++ b/packages/lexical/src/LexicalUpdates.ts
@@ -6,16 +6,6 @@
  *
  */
 
-import type {
-  CommandPayloadType,
-  EditorUpdateOptions,
-  LexicalCommand,
-  LexicalEditor,
-  Listener,
-  MutatedNodes,
-  RegisteredNodes,
-  Transform,
-} from './LexicalEditor';
 import type {SerializedEditorState} from './LexicalEditorState';
 import type {LexicalNode, SerializedLexicalNode} from './LexicalNode';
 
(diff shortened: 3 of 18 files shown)


packages/lexical/src/LexicalUpdates.ts
```diff
@@ -110,13 +113,55 @@ export function getActiveEditor(): LexicalEditor {
...
+        ` and incompatible editors with versions ${Array.from(
+          incompatibleEditors,
+        ).join(', ')}`,
+      );
+    }
+    return output.join('');
+  } else {
+    return '';
+  }
+}
```
etrepum, line 163: For DX reasons it might be worth including this troubleshooting in the prod build, but it's difficult to eliminate the strings

etrepum:
I think these are flaky collab test failures

packages/lexical-website/docs/concepts/listeners.md
```diff
@@ -84,7 +84,7 @@ handle external UI state and UI features relating to specific types of node.
 If any existing nodes are in the DOM, and skipInitialization is not true, the listener
 will be called immediately with an updateTag of 'registerMutationListener' where all
 nodes have the 'created' NodeMutation. This can be controlled with the skipInitialization option
-(default is currently true for backwards compatibility in 0.16.x but will change to false in 0.17.0).
+(default is currently true for backwards compatibility in 0.17.x but will change to false in 0.18.0).
```
etrepum, line 87: Changed the deprecation policy here since 0.17.0 was the first release with the changes to `registerMutationListener` (#6357)

packages/lexical-playground/vite.config.ts
```diff
@@ -52,6 +52,10 @@ export default defineConfig(({command}) => {
             from: /__DEV__/g,
             to: 'true',
           },
+          {
+            from: 'process.env.LEXICAL_VERSION',
+            to: JSON.stringify(`${process.env.npm_package_version}+git`),
```
etrepum, line 57: The playground is never built with an npm release of lexical since it directly includes all of the source

potatowagon (approved):
👍

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -110,13 +113,46 @@ export function getActiveEditor(): LexicalEditor {
...
-        'editor.update() or editor.read().',
+        'editor.update() or editor.read().%s',
+      collectBuildInformation(),
     );
   }
-
   return activeEditor;
 }
 
+function collectBuildInformation(): string {
```
zurfyx, line 123: @etrepum nitpicking but such detailed error is something I'd consider gating under the `__DEV__` build or developer extension. I don't think it's very common to have multiple editors on the same page and it's less common to make this mistake, and there is quite some logic in here 679 bytes minimized.

Issue #4986, opened 2023-09-09
Feature: Pass `LexicalEditor` to `LexicalComposer`


montlebalm:
It would be useful to be able to pass an instance of `LexicalEditor` into `LexicalComposer` just like you're able to do with `LexicalNestedComposer`.

I'm working with a setup where a `DecoratorNode` needs to generate serialized versions of 2 editors rendered with a React component. It's simpler if the `DecoratorNode` controls the editor instances and passes them down to the component. The alternative is having the `DecoratorNode` prop drill the params necessary to render the editors, let the React component instantiate the editors, and then access the instances via multiple levels of refs.

The `LexicalComposer` config still allows for a `editor__DEPRECATED` param that makes my proposal possible. My feature request is to drop the `__DEPRECATED` label and support it as a config option.

To avoid confusion, the initial config could support both `editor` and `editorState`, but only allow one at a time.

PR that deprecated the `editor` prop: https://github.com/facebook/lexical/pull/1407

yf-yang:
After examine the code, I found you can create your own provider of LexicalComposerContext, it is exported.

etrepum:
If you're not using it to create the editor, then you should create a LexicalComposerContext directly

PR #6494, opened 2024-08-06
[lexical-react] remove editor__DEPRECATED that has been deprecated for two years

etrepum:
## Description

The `editor__DEPRECATED` prop has been deprecated for two years (#1407), this PR removes it.

## Test plan

There were no tests for this feature, and its absence does not need new tests.

### DIFF
--- a/packages/lexical-react/flow/LexicalComposer.js.flow
+++ b/packages/lexical-react/flow/LexicalComposer.js.flow
@@ -23,7 +23,6 @@ export type InitialEditorStateType =
   | ((editor: LexicalEditor) => void);
 
 export type InitialConfigType = $ReadOnly<{
-  editor__DEPRECATED?: LexicalEditor | null,
   editable?: boolean,
   namespace: string,
   nodes?: $ReadOnlyArray<Class<LexicalNode> | LexicalNodeReplacement>,
--- a/.eslintignore
+++ b/.eslintignore
@@ -14,3 +14,4 @@
 .ts-temp
 **/.docusaurus
 /playwright-report
+test-results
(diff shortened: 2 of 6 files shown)


packages/lexical-react/flow/LexicalComposer.js.flow
```diff
@@ -23,7 +23,6 @@ export type InitialEditorStateType =
   | ((editor: LexicalEditor) => void);
 
 export type InitialConfigType = $ReadOnly<{
-  editor__DEPRECATED?: LexicalEditor | null,
```
potatowagon, line 26: editor__DEPRECATED is used some places on www,

https://www.internalfb.com/code/search?q=repo%3Awww%20editor__DEPRECATED

 im not sure how backwards compatible this would be
 
 @zurfyx should we try cleaning up editor__DEPRECATED in www first before syncing this over?

zurfyx: @potatowagon Good point, we need to fix these occurrences first. I'll connect with you offline but unfortunately we should put this PR on hold for now

fantactuka:
I just recalled Chriss @montlebalm was mentioning use case with reusing editor instance via this prop: https://github.com/facebook/lexical/issues/4986

Wonder if it's used by other folks too and (although it was marked as deprecated forever) it worth having explicit exception when it's passed

zurfyx:
> use
 
@fantactuka FWIW we're also using the same workaround described in the issue internally and that's not going away anytime soon

etrepum:
@fantactuka I can't think of a good reason to render a `LexicalComposer` when you already have the editor instance, `LexicalComposerContext.Provider` can be used instead

zurfyx (approved):
We can merge now! Thank you

PR #6501, opened 2024-08-07
Fix splitText when detached

zurfyx:
When `splitText` is run on a Node with no parent it crashes. I don't see any obvious reason why this function should only work on attached nodes and there are some valid cases when you'd want to attach the resulting nodes later and leverage `splitText` to preserve the styling and format.

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -935,7 +935,7 @@ export class TextNode extends LexicalNode {
       return [self];
     }
     const firstPart = parts[0];
-    const parent = self.getParentOrThrow();
+    const parent = self.getParent();
     let writableNode;
     const format = self.getFormat();
     const style = self.getStyle();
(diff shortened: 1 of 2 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -1005,23 +1005,25 @@ export class TextNode extends LexicalNode {
...
-      writableParent.splice(insertionIndex, 1, splitNodes);
-    }
+    if (parent !== null) {
+      internalMarkSiblingsAsDirty(this);
+      const writableParent = parent.getWritable();
+      const insertionIndex = this.getIndexWithinParent();
+      if (hasReplacedSelf) {
+        writableParent.splice(insertionIndex, 0, splitNodes);
+        this.remove();
+      } else {
```
etrepum, line 1015: I know this isn't the code that's changing in this PR but it seems like there might be a hasReplacedSelf selection bug, the above code to fixup selections only handles parts>0 and the splice selection fixup code won't do anything either because the node removal is happening separately. It also seems a bit odd that the special case only happens with IS_SEGMENTED and not IS_TOKEN

etrepum (approved):
Looks like a net improvement, but I think there may be other edge cases left in this method

PR #6507, opened 2024-08-09
Multiple update tags

zurfyx:
With `history-merge` being a sort of imposed tag to follow certain behavior, you're forced to trigger another `editor.update` to pass your own tag. This is a patchy workaround. What's more, this workaround doesn't work for `discrete` updates.

This diff adds support for an array of tags, I kept the `tag` name singular for backward compatibility.


etrepum:
FWIW there is `$addUpdateTag` which can be used instead of triggering multiple `editor.update` calls (or using the tag property at all)

etrepum (approved):
LGTM

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -219,6 +219,20 @@ function $normalizeAllDirtyTextNodes(
   }
 }
 
+function addTags(editor: LexicalEditor, tags: undefined | string | string[]) {
+  if (!tags) {
```
acywatson, line 223: be strict here

acywatson (approved):

packages/lexical/src/LexicalUpdates.ts
etrepum, line 223:
```suggestion
  if (tags === undefined || tags.length === 0) {
```

etrepum (approved):

PR #6509, opened 2024-08-10
[lexical-playground] Feat: Cannot insert table into table

keiseiTi:
## Description

In notion or other editors, cannot insert a table into a table

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -24,6 +24,7 @@ import {
   copyToClipboard,
   deleteTableColumns,
   deleteTableRows,
+  expect,
   focusEditor,
   html,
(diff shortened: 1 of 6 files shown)


ivailop7:
I proposed this functionality some time ago, but was rejected. Someone even fixed the table in table selection logic. Unlikely to be merged. It's also the type of feature that can be implemented on playground-level for those that want it.

etrepum:
@ivailop7 this PR prevents tables from being inserted into tables, seems like a win unless someone is planning to fix all of the bugs with nested tables?

ivailop7:
I'd rather have the merged cells issues fixed first. They are largest weak point of tables right now. I've labelled all the tables issues, in case someone wants to have a go at them. I'm in favour of disabling table withing table, but need the rest of the core folks to chime in as well @zurfyx and @StyleT, @fantactuka

PR #6510, opened 2024-08-10
[lexical-playground] Fix: in playground show component-menu when scroll

keiseiTi:
## Description

in playground show component-menu when scroll

### Before

iShot_2024-08-10_15 19 16

### After

iShot_2024-08-10_15 27 15


ivailop7:
This is essentially undoing: https://github.com/facebook/lexical/pull/5859 

In the preview for this PR, you can reproduce the issue YOffset was fixing, if you do a very long text, scroll down and do / for the menu to show up, it's rendered incorrectly offscreen. This logic is also used by the Lexical context menu, which you essential fix this for correctly, but break the typeahead when triggered at a lower scrolled page position.

etrepum (approved):
This seems to work great for me, I am unable to reproduce any positioning issues with the menu in the scenarios I could find

PR #6511, opened 2024-08-12
[lexical] surface more error details in reconciler

potatowagon:
## Description

since 25 jul, observing a high number of `The node to be replaced/removed is not a child of this node` errors, at the following 2 lines where im adding logging. These errors spiked on 6 Aug but have since died down (on its own) to ~ 2 errors a daily

i've not been able to reproduce those errors so surfacing more details in the error msg to get more clues

i made sure to only surface privacy safe details in the error

fix attempt: https://github.com/facebook/lexical/pull/6513

## Test plan

Screenshot 2024-08-12 at 12 14 54 PM

threw a new error there to emulate the error. 

to trigger that line, replace an entire paragraph node, eg. pasting over the entire paragraph node 

Screenshot 2024-08-12 at 12 40 25 PM

### DIFF
--- a/packages/lexical/src/LexicalReconciler.ts
+++ b/packages/lexical/src/LexicalReconciler.ts
@@ -340,7 +340,18 @@ function reconcileElementTerminatingLineBreak(
       const element = dom.__lexicalLineBreak;
 
       if (element != null) {
-        dom.removeChild(element);
+        try {
+          dom.removeChild(element);
+        } catch (error) {
+          if (typeof error === 'object' && error != null) {
+            const msg = `${error.toString()} Parent: ${dom.tagName}, child: ${
+              element.tagName
+            }.`;
+            throw new Error(msg);
+          } else {
+            throw error;
+          }
+        }
       }
 
       // @ts-expect-error: internal field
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalReconciler.ts
```diff
@@ -501,7 +512,22 @@ function $reconcileChildren(
     } else {
       const lastDOM = getPrevElementByKeyOrThrow(prevFirstChildKey);
       const replacementDOM = $createNode(nextFrstChildKey, null, null);
-      dom.replaceChild(replacementDOM, lastDOM);
+      try {
+        dom.replaceChild(replacementDOM, lastDOM);
```
potatowagon, line 516: what could cause `The node to be replaced is not a child of this node` error here?

seems like the error suggest lastDOM is not a child of dom. did it get deleted before this line exec? and how?

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -340,7 +340,18 @@ function reconcileElementTerminatingLineBreak(
       const element = dom.__lexicalLineBreak;
 
       if (element != null) {
-        dom.removeChild(element);
+        try {
+          dom.removeChild(element);
```
potatowagon, line 344: what could cause `The node to be removed is not a child of this node` error here?

is dom.__lexicalLineBreak above deleted or returning outdated value? also where is dom.__lexicalLineBreak set?

zurfyx (approved):

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -501,7 +512,22 @@ function $reconcileChildren(
     } else {
       const lastDOM = getPrevElementByKeyOrThrow(prevFirstChildKey);
       const replacementDOM = $createNode(nextFrstChildKey, null, null);
-      dom.replaceChild(replacementDOM, lastDOM);
+      try {
+        dom.replaceChild(replacementDOM, lastDOM);
+      } catch (error) {
+        if (typeof error === 'object' && error != null) {
+          const msg = `${error.toString()} Parent: ${
+            dom.tagName
```
potatowagon, line 520: what are the best practices when logging errors in lexical? any methods to log the sanitised/privacy safe domNode/lexical node/editor state?

zurfyx: @potatowagon we'd use `invariant` and you can pass any params in, also keep in mind that internally logged errors have a not very generous max length

potatowagon:
funny the extended e2e tests signal didnt show up in the signal box but i can see they ran successfully (no failures) under the Checks (51) at the top of page

https://github.com/facebook/lexical/actions/runs/10352963152

PR #6517, opened 2024-08-13
[lexical-react] Fix: Fix React.startTransition on Webpack + React 17

### LINKED ISSUE #6516: Bug: Webpack + React 17 fails to compile on the usage of `React.startTransition` or `React["startTransition"]`
Lexical version: latest version

## Steps To Reproduce

1. some old projects will break with webpack + react 17.

turnerguo:
## Description

* Webpack + React 17 fails to compile on the usage of React.startTransition or React["startTransition"]

Closes #6516


etrepum (commented):
It probably makes sense to extract this into a shared module since it's used in multiple places and it would make it easier if it needs to be used again

etrepum (approved):

PR #6522, opened 2024-08-15
[lexical-list] Bug Fix: handle non-integer numbers in setIndent

### LINKED ISSUE #6521: Bug: browser lockup when converting paragraph node to list
Pasting a paragraph into Lexical can result in a non-integer `indent` value. If you then try to convert this paragraph to a list, the browser freezes due to an infinite loop.

Lexical version: 0.17.0 and latest on Playground as of 2024-08-15

## Steps To Reproduce

1. Open Moby Dick on Project Gutenberg: https://www.gutenberg.org/cache/epub/2701/pg2701-images.html
2. Scroll to "CHAPTER 1. Loomings.", copy the heading and first three paragraphs.
3. Open the Lexical Playground: https://playground.lexical.dev/
4. Paste the contents
5. Select one of the paragraphs and convert to a list

image

Here's an example editor JSON state (different tab) to show the indent value:

image

## The current behavior

The browser locks up and you have to kill the tab. In `ListItemNode.setIndent`, the node is indented/outdented until `currentIndent === indent`. This never happens because `indent` in a non-integer value.

image

## The expected behavior

The paragraph is converted to a list item. Ideally, `setIndent` gracefully handles fractional indent values as we already have Lexical documents where the paragraphs are stored with said values.

jrfitzsimmons:
## Description

Handles non-integer numbers in `ListItemNode.setIndent` to prevent infinite loop.

Closes #6521

### DIFF
--- a/packages/lexical-list/src/LexicalListItemNode.ts
+++ b/packages/lexical-list/src/LexicalListItemNode.ts
@@ -357,10 +357,9 @@ export class ListItemNode extends ElementNode {
   }
 
   setIndent(indent: number): this {
-    invariant(
-      typeof indent === 'number' && indent > -1,
-      'Invalid indent value.',
-    );
+    invariant(typeof indent === 'number', 'Invalid indent value.');
+    indent = Math.floor(indent)
+    invariant(indent >= 0, 'Indent value must be non-negative.');
     let currentIndent = this.getIndent();
     while (currentIndent !== indent) {
       if (currentIndent < indent) {


ivailop7 (approved):

etrepum (approved):

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -357,10 +357,9 @@ export class ListItemNode extends ElementNode {
   }
 
   setIndent(indent: number): this {
-    invariant(
-      typeof indent === 'number' && indent > -1,
-      'Invalid indent value.',
-    );
+    invariant(typeof indent === 'number', 'Invalid indent value.');
```
zurfyx, line 360: I wonder if we need to typecheck here when the type is already guaranteed to be a number by TS. While it is true that types are optional for users I feel like if I were to typecheck everywhere it would add a lot of noise

zurfyx (approved):

packages/lexical-list/src/LexicalListItemNode.ts
jrfitzsimmons: @acywatson had a similar comment on the PR that added this check (#4120)

etrepum: The Math.floor is going to guarantee it's a number but it could be NaN so we might still have a bad input bug?

james-atticus: That would still fail the `indent >= 0` check. Or are you saying remove the invariant above given the one below is will catch non-numbers. My vote would be leave it in as it's more explicit as to what's going on. I.e. not relying on Math.floor to produce NaN and then comparing that to 0. Happy to update if you prefer the latter though.

jrfitzsimmons: That would still fail the `indent >= 0` check. Or are you saying remove the invariant above given the one below will catch non-numbers. My vote would be leave it in as it's more explicit as to what's going on. I.e. not relying on Math.floor to produce NaN and then comparing that to 0. Happy to update if you prefer the latter though.

etrepum: Good point, I was just looking at it in isolation and not considering both invariants. I think it's fine as-is.

jrfitzsimmons: Cool. Is there anything I need to do to fix the extended tests failures? Can't see why this change would cause them to fail...

Issue #6402, opened 2024-07-15
Bug: Cut from a table doesn't remove values


circlingthesun:
Lexical version: v0.16.1

## Steps To Reproduce

1. Open playground
2. Insert a 5x5 table
3. Fill out a row with values
4. Select the row with aforementioned values
5. Right click and select cut or press cmd + x
6. Observe that the selected values are stil present

https://github.com/user-attachments/assets/deb58c76-ed2b-4189-9371-8645c7328d40

## The current behavior

After cutting the row the values are not removed

## The expected behavior

After cutting the cells should be blank

## Impact of fix

This affects any end user using tables. It's frustrating unexpected behaviour.

jrfitzsimmons:
+1 - from testing it seems like the content is added to the clipboard correctly (paste works fine) it's just the clearing of the cell contents that doesn't work.

PR #6528, opened 2024-08-19
[lexical-table] Fix a number of table Cut command scenarios

ivailop7:
Fixing a few Table Cut related issues:
1) When selected text + table and table is last node, Cut results in incorrect remaining paragraph text

https://github.com/user-attachments/assets/9a9c6239-0844-4b7a-accc-28bf8f10737f

2) When selected text + table and table is first node, Cut results in a table with single row and of single cell

https://github.com/user-attachments/assets/764cc4f1-e877-4f3c-a290-67dd78c0b284

3) When selection is inside a table, cut doesn't clear the cell contents after copy-ing (Resolves: #6402)

4) When the whole table is selected, cut should truly cut the table, so the user can paste elsewhere.


ivailop7:
@etrepum @zurfyx @potatowagon could I get a sign off here

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -371,6 +378,40 @@ export function applyTableHandlers(
...
+          );
+
+          if ($isTableSelection(selection)) {
+            $deleteCellHandler(event);
+            return true;
+          } else if ($isRangeSelection(selection)) {
+            editor.update(() => {
+              $deleteCellHandler(event);
+              selection.removeText();
+            });
```
etrepum, line 405: nit: editor.update is redundant in a command listener

```suggestion
            $deleteCellHandler(event);
            selection.removeText();
```

etrepum (dismissed):

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
ivailop7: updated

etrepum (approved):

PR #6530, opened 2024-08-19
[lexical-markdown][breaking change] Feature: multiline markdown transformers / mdx support

### LINKED ISSUE #2564: Bug: Markdown code node importing is hardcoded
Markdown functionality relies on passed transformers list, which allows passing built-in ones as well as custom nodes trasnfomers. But the code block is an exception - MarkdownImport always processes code blocks regardless of transformers list:

https://github.com/facebook/lexical/blob/2abc8244d5754860b96bfc6f875bdcbf23d4e390/packages/lexical-markdown/src/v2/MarkdownImport.ts#L54-L74

Which means even if we don't pass code blocks (or if we pass custom code node) the built-in version will be used.

AlessioGr:
### Motivation for this change

We would like to add mdx support to [Payload](https://github.com/payloadcms/payload/tree/beta), allowing mdx files to be easily import to and exported from lexical editors. Multiline markdown transformer support will allow JSX within mdx to be handled properly.

### Multiline markdown transformers

Currently, lexical only has one multiline markdown transformer for code nodes that is hardcoded. This comes with 2 issues:
-  users cannot extend lexical with their own multiline markdown transformers
- The code node markdown import transformer is executed (if there is code markdown (```) in the markdown string) even if code nodes are not enabled in the project, leading to errors

This PR adds a new `MultilineElementTransformer` type and changes the hardcoded code markdown transformer to be a `MultilineElementTransformer`.

I intentionally went for a separate type, contrary to what @fantactuka suggested [here](https://github.com/facebook/lexical/issues/2564#issuecomment-1176582653), but I'm happy to explore merging the types if it makes sense. Here are my reasons for making it a separate type:
- Within `createMarkdownImport` it's easier to determine if it's a multiline transformer - the algorithm for handling multiline transformers is very different
- Less bloated types. The `replace` function of the `MultilineElementTransformer` type drastically differs from the `replace` function of the `ElementTransformer` type: 
  - A multiline transformer does not need `isImport` as an argument in the `replace` function. It will only be run during imports, and not in the MarkdownShortcut plugin
  - There are 2 kinds of regex matches now (`openMatch` and `closeMatch`) and not just `match` - both should be available as arguments in `replace`
  - I added a new `linesInBetween` args to the `replace` arguments, instead of the `children` argument. Additionally, only the `rootNode` is passed as an argument. This is easier to understand / simpler and gives the developer full control over _which_ nodes to add to the root node and _how_, instead of lexical always creating a text node and a paragraph node. The new multiline code node is a good example for this - it does not need a `ParagraphNode` to be created
  - `linesInBetween` is the text matched between the open and close match, separated by lines. Basically, _anything in between the matches_. This is easy to understand, the developer now has easy access to anything in between the regex match, as well as the regex matches themselves

Regarding mdx (JSX) support: Now, with multiline markdown transformer support, we can easily utilize a general regex for matching any JSX tags, and then further parse that matched JSX node into the respective lexical node. An example for this can be found in the new tests here: https://github.com/facebook/lexical/blob/7d4a90435d47a26740da4ed38caa0895a4723de5/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts#L271

This also brings us to the second, smaller feature:

### Support terminating Element and MultilineElement transformers early, if the regex matched

If, during an import, the regex for one of your transformers matched, this will be the only transformer that will run. You cannot just decide that "hey, the regex matched, but I don't actually want to handle transforming this" inside the `replace` function.

This is possible in the `export` function of a transformer by returning null - that way, the next transformer will be called. This PR makes this possible for imports as well. Now, if you return `false` in the `replace` function, it will keep attempting to find a match, going to the next markdown transformer.

Use-case: We'd like to employ one, generic multiline markdown transformer for any JSX elements (mdx support). Then, I want to decide within the replace function if I'd like to handle this match or not, by reading the node type (the text between `< >`) - this is easier than writing unique regex for every single kind of html node I'd like to handle. That way, we'd just have one clean switch statement, and a great DX for allowing our users to register their own markdown handling for certain JSX elements

If there is no handler registered for this node type, I can now simply return false, thus signaling lexical that I don't want to handle this node and that lexical can keep looking for a more suitable transformer.

### JS Docs

This adds some JSDocs to the transformer types.

Closes #2564

## Test plan

See diffs of the LexicalMarkdown.test.ts file: https://github.com/facebook/lexical/pull/6530/files#diff-9a0c91bc82f2693a031ebe1165316e9726980b7f1d54de8681de1256e1d18303L233

Additionally, the tests for code nodes are still passing - this means the move from hardcoded code multiline transformer to actual code multiline transformer was successful

While working on this, I have also [encountered one test case that failed](https://github.com/facebook/lexical/pull/6530/files#diff-9a0c91bc82f2693a031ebe1165316e9726980b7f1d54de8681de1256e1d18303R216) - this fails in the current lexical version as well. As this is not within the scope of this PR, I did not fix it here - however I did include a TODO item, as well as a commented out test case


etrepum:
I haven't had a chance to look at any of this PR yet, but have you considered starting from a better parsing foundation and building a new lexical markdown API without any legacy constraints (e.g. leveraging remark)? The current one is a bit of a dead end with a handful of design flaws (although a lot less so with all of your work, under the assumption that the code looks good)

AlessioGr:
Hey @etrepum, I did consider something like remark & micromark, but those solutions would increase the bundle size by around 100kB. I'm trying to keep it as small possible. Additionally, something like micromark uses packages like `Acorn` for handling jsx. We've seen that package unleash hell when it's bundled into a Next.js app in the past, so I'd rather avoid that. This regex-based solution is very lean!

I also wouldn't know how well they would play together using the MarkdownShortcut plugin, which works quite well with regex. I've seen [this project](https://github.com/mdx-editor/editor/blob/main/src/importMarkdownToLexical.ts) use micromark for imports / exports, but it's still using the current solution regex-based solution for the MarkdownShortcut plugin.

I must say though, the regex-based multiline markdown handling from this PR works really well for us so far. I was half expecting us to run into some weird edge-cases that we wouldn't have if we used remark / parsed around some ast, but that's not the case. We're currently using a packaged version of @lexical/markdown with this PR included, and got a bi-directional mdx import / export working - with complex JSX and all that. It's pretty nice - no more editing our docs in plain mdx :p

If the JSX gets too complex to handle using regex, there's always the option to _**just**_  parse the children of the JSX element using micromark & work with the AST tree of that complex JSX element.  instead. But that's optional

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -182,6 +213,11 @@ describe('Markdown', () => {
       md: 'Hello ~~__*world*__~~!',
       skipExport: true,
     },
+    /*{
+      html: '<pre spellcheck="false"><span style="white-space: pre-wrap;">Inline Code</span></pre>',
+      md: '```Inline Code```',
+    },*/
+    // TODO: This test currently fails. Fix it
```
ivailop7, line 220: Still to be fixed?

ivailop7 (commented):
I'm on board on landing this one, as long as all tests pass and it's backwards-compatible. This PR could be a good opportunity to decouple the markdown package from the code package, which would be a huge win.

AlessioGr:
I have cleaned & simplified the code a bit, and fixed the todo issue with the single-line code block imports. Also do note that the multi-line code block transformer now also handles single-line code blocks.

I have completely disabled the normal code block transformer for imports as it was conflicting with the multiline transformer.  For example, for ```Inline Code``` it would take `Inline` as the language, because it was optimized to work with incomplete code blocks coming from the markdown shortcut plugin. And if I disabled the language detection for imports, it would still match something like ```ts - something only the multiline transformer should match as we need to detect the end of the code block for imports.
Now, import & shortcut code transformers are cleanly separated. The shortcut transformer can stay optimized for incomplete code blocks

Lmk what you think!

For a future PR, I think we should consider consolidating normal element transformers and multiline element transformers into one single transformer that can handle it all. Of course this would be a huge breaking change, so I don't wanna explore this in this PR.

I see some e2e's are now failing. I'll look into them tomorrow and fix whatever is not flaky

AlessioGr:
@ivailop7 I think this should be ready. The failing CI tests might be. E.g. here is [this failing CI test](https://github.com/facebook/lexical/actions/runs/10527769387/job/29171711665?pr=6530) running locally:

CleanShot 2024-08-23 at 11 09 13

`npm run test-e2e-chromium` also passes locally. Can you let me know if you think any _real_ failing e2e tests are left?

etrepum:
Some of these e2e tests appear to be failing consistently, e.g.

https://github.com/facebook/lexical/actions/runs/10527769387/job/29171708325?pr=6530
https://github.com/facebook/lexical/actions/runs/10527769387/job/29171711665?pr=6530
https://github.com/facebook/lexical/actions/runs/10527769387/job/29171715036?pr=6530
```
  1 failed
    [chromium] › packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs:472:3 › Hashtags › Should not break while importing and exporting multiple matches 
```

https://github.com/facebook/lexical/actions/runs/10527769387/job/29171705348?pr=6530
https://github.com/facebook/lexical/actions/runs/10527769387/job/29171704056?pr=6530
https://github.com/facebook/lexical/actions/runs/10527769387/job/29171712246?pr=6530
```
  1 failed
    [firefox] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:651:3 › Selection › Triple-clicking last cell in table should not select entire document 
```

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -48,14 +49,14 @@ export function createMarkdownExport(
       const child = children[i];
       const result = exportTopLevelElements(
         child,
-        byType.element,
+        [...byType.multilineElement, ...byType.element],
```
etrepum, line 52: You could float this out to the top-level of `createMarkdownExport`, add an additional parameter to `exportTopLevelArguments` (it's not a public api, no need to preserve its signature), or do the merge in `transformersByType` (unless there's some reason to keep them separate). No need to create a new Array each run.

packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -103,6 +107,87 @@ export function createMarkdownImport(
...
+ *
+ * @returns first element of the returned tuple is a boolean indicating if a multiline element was imported. The second element is the index of the last line that was processed.
+ */
+function $importMultiline(
+  lines: Array<string>,
+  startLineIndex: number,
+  multilineElementTransformers: Array<MultilineElementTransformer>,
+  rootNode: ElementNode,
+): [boolean, number] {
+  transformerLoop: for (const {
```
etrepum, line 120: I don't think these `transformerLoop` and `lineLoop` labels are necessary, it seems that all of the `continue` statements are for the current loop scope anyway, except for the last `continue transformerLoop` which could be a `break` since it's the tail of both loops. It makes it much harder to understand what the code is doing when you have to also keep track of the labels.

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -40,26 +40,75 @@ import {
...
+   * replace is called when markdown is imported or typed in the editor
+   */
   replace: (
     parentNode: ElementNode,
     children: Array<LexicalNode>,
     match: Array<string>,
+    /**
+     * Whether the match is from an import operation (e.g. through $convertFromMarkdownString) or not (e.g. through typing in the editor).
+     *
+     * @return return false to cancel the transform, even though the regex matched. Lexical will then search for the next transformer.
```
etrepum, line 69: I think this `@return` should go above since it's not related to the isImport parameter

packages/lexical-markdown/src/MarkdownExport.ts
AlessioGr: Fixed it, good thing you caught that. I went with moving them to the top of the function for simplicity. Keeping them separate it probably best given that they interfaces are very different

AlessioGr:
That specific test is broken in `main` too ("Should not break while importing and exporting multiple matches").

I have enabled slow typing for this test (this is the CI artifact on main):
[video.webm](https://github.com/user-attachments/assets/da983711-4f61-4ec6-aff0-8ac270f2c9bf)

No code node is created. Only happens in CI, works fine in local dev. Seems like in main, for some reason, the frequent pressing of the markdown button seems to unbreak it.

I'll try to figure out why the markdown transform is not triggered in CI

AlessioGr:
All tests are now passing! 🎉 


The discrepancies I had between local and CI e2e runs was due to CI running the test in plain-text mode. I was running them in rich-text mode locally - didn't notice the environment variable that was being set in CI.

The tests were basically failing because the new multiline code transformer, which is now used instead of the normal code transformer, does not handle "incomplete" markdown (e.g. "\`\`\`markdown hello" - no closing ```". In order to support incomplete markdown handling, I have added an `optional` property to `regExpEnd`, which can now be an object. If set to true, transformer will run even if no end regex has been matched. I have added new unit tests to test this.

This change now alleviates the need to have both a multiline and a normal ElementTransformer - the MultilineTransformer can handle it all. Thus, I have merged the `CODE` and `CODE_MULTILINE` transformers into just one `CODE` MultilineTransformer. MultilineTransformers can now also run during markdown shortcut transforms. I have made this change in https://github.com/facebook/lexical/pull/6530/commits/03ffd3db57e8af9f50deb08ad1d0d01ba279d66f

Please review the updated API. In the future, we could consider getting rid of the normal `ElementTransformer`, as the `MultilineElementTransformer` can handle it all, or merge them together in some other form - though I do not want to make any breaking changes in this PR.

Additionally, I'm considering making the arguments of the MultilineElementTransformer an single object argument, with the current arguments being object props. So,

`replace: ({ rootNode, children, startMatch, endMatch, linesInBetween })`

Instead of

`replace: (rootNode, children, startMatch, endMatch, linesInBetween)`


In my opinion, this be a cleaner API for the user. Better auto-suggestions, and we don't have to force them to type every single argument to get to something like `linesInBetween`. If we want to add arguments in the future, this will be especially bad. Additionally, we can improve the types by using union types to type `children` as null if `linesInBetween` is set, and `linesInBetween` as null if `children` is set - or even better, type it based on the `isImport` property.
However, the disadvantage is that this would be inconsistent with the `replace` function of the `ElementTransformer`. Let me know if we should do this change

potatowagon (approved):

potatowagon:
do u foresee any breaking change? code looks ok, i havnt spoted any major api changes, more like adding a multiline transformer as an option.

labelling it as breaking change for an internal memo to test further for any markdown regressions

AlessioGr:
> do u foresee any breaking change?

Not in this PR; though I foresee us merging MultilineElementTransformers and ElementTransformers together in the future, and that would be a breaking change

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -256,11 +326,72 @@ export const CODE: ElementTransformer = {
...
+      const textNode = $createTextNode(code);
+      codeBlockNode.append(textNode);
+      rootNode.append(codeBlockNode);
+    } else if (children) {
+      createBlockNode((match) => {
+        return $createCodeNode(match ? match[1] : undefined);
+      })(rootNode, children, startMatch, isImport);
+    }
+  },
+  type: 'multilineElement',
```
potatowagon, line 394: should this be multiline-element to follow the conventions of eg. `text-format`? any side effects if i raise a PR to rename this? @AlessioGr

PR #6535, opened 2024-08-21
Revert "Fix OverflowNode configuration"

### LINKED ISSUE #6534: Bug: Regression in test suite after merging #6027 (Fix OverflowNode configuration)
Test failures since #6027. It must've been in a state where the extended test suite wasn't running, so there were no plain text e2e tests run and the PR was green when it was merged.

For example:
```
  4 failed
    [webkit] › packages/lexical-playground/__tests__/e2e/CharacterLimit.spec.mjs:120:3 › CharacterLimit › UTF-16 › can type new lines inside overflow 
    [webkit] › packages/lexical-playground/__tests__/e2e/CharacterLimit.spec.mjs:175:3 › CharacterLimit › UTF-16 › can delete text in front and overflow is recomputed 
    [webkit] › packages/lexical-playground/__tests__/e2e/CharacterLimit.spec.mjs:120:3 › CharacterLimit › UTF-8 › can type new lines inside overflow 
    [webkit] › packages/lexical-playground/__tests__/e2e/CharacterLimit.spec.mjs:175:3 › CharacterLimit › UTF-8 › can delete text in front and overflow is recomputed 
```

Lexical version: nightly

## Steps To Reproduce

```
npm run test-e2e-plain-*
```

## The current behavior

Tests fail

## The expected behavior

Tests pass

## Impact of fix

Unblock PRs

etrepum:
Reverts facebook/lexical#6027

OverflowNode causes problems with LineBreakNode with this configuration change, causes the tests to fail in plain text mode (the insertion point always ends up *before* the trailing LineBreakNode) but the same strange behavior comes up if you press shift-return to insert a LineBreakNode into a ParagraphNode in rich text mode. I think we will need to revisit #6027 with additional tests.

Closes #6534

/cc @zurfyx @ivailop7

### DIFF
--- a/packages/lexical-overflow/src/index.ts
+++ b/packages/lexical-overflow/src/index.ts
@@ -69,15 +69,7 @@ export class OverflowNode extends ElementNode {
     return parent.insertNewAfter(selection, restoreSelection);
   }
 
-  canBeEmpty(): false {
-    return false;
-  }
-
-  isInline(): true {
-    return true;
-  }
-
-  excludeFromCopy(): true {
+  excludeFromCopy(): boolean {
     return true;
   }
 }


etrepum:
Pretty sure that this test is flaky so I'm going to restart that job:

```
  1 failed
    [chromium] › packages/lexical-playground/__tests__/e2e/Links.spec.mjs:1960:3 › Links › Can handle pressing Enter inside a Link containing multiple TextNodes 
```

etrepum:
The tests passed the second time around, it must be flaky.

potatowagon (approved):
hmm its not ideal that the e2e test signal is marked as skipped in the PR checks, with an overall passing signal. something seems to be up with the test workflows.

ivailop7 (approved):

PR #6538, opened 2024-08-21
[lexical-react] Fix multiple node selection deletion

ivailop7:
Selecting multiple nodes using Cmd + Shift + Click, which results in a NodeSelection with multiple nodes currently doesn't delete all of the selected nodes.

Before: 

https://github.com/user-attachments/assets/6b468486-494e-4067-b66f-b06a3f6b3625

After:

https://github.com/user-attachments/assets/7507b0c9-8f77-4427-b73d-25c099aff9da

Request for comment: The $onDelete method is super repetitive and identical for the exception of the 'isNodeOfType' check. Since most of these are in the playground, could do an util function, but an alternative solution is to modify this on the 'rich-text' plugin level deleteHandler and remove all of the $onDelete methods and overrides. Thoughts? 

@zurfyx @etrepum @potatowagon


zurfyx (approved):
> Request for comment: The $onDelete method is super repetitive and identical for the exception of the 'isNodeOfType' check. Since most of these are in the playground, could do an util function, but an alternative solution is to modify this on the 'rich-text' plugin level deleteHandler and remove all of the $onDelete methods and overrides. Thoughts?

Agree! IMO I'd make it into the rich/plaintext plugins as a command to mimic what we already do with RangeSelection.

`REMOVE_NODE_COMMAND` would be the equivalent of `DELETE_CHARACTER_COMMAND`. We implement `REMOVE_NODE_COMMAND` either in the selection module or rich/plain plugin and allow for plugins to override it if necessary.

Unfortunately, this non backward compatible change can accidentally break some products so we might need a version bump at the very least or rethink how we ship this. FWIW, when selected nodes > 1 this is always the desired behavior since you're not expected to interact with the node without resetting the selection.

ivailop7:
> > Request for comment: The $onDelete method is super repetitive and identical for the exception of the 'isNodeOfType' check. Since most of these are in the playground, could do an util function, but an alternative solution is to modify this on the 'rich-text' plugin level deleteHandler and remove all of the $onDelete methods and overrides. Thoughts?
> 
> Agree! IMO I'd make it into the rich/plaintext plugins as a command to mimic what we already do with RangeSelection.
> 
> `REMOVE_NODE_COMMAND` would be the equivalent of `DELETE_CHARACTER_COMMAND`. We implement `REMOVE_NODE_COMMAND` either in the selection module or rich/plain plugin and allow for plugins to override it if necessary.
> 
> Unfortunately, this non backward compatible change can accidentally break some products so we might need a version bump at the very least or rethink how we ship this. FWIW, when selected nodes > 1 this is always the desired behavior since you're not expected to interact with the node without resetting the selection.

OK, will merge this as is then and think on the REMOVE_NODE_COMMAND intro

PR #6539, opened 2024-08-21
[lexical-table] Stop selecting the whole table after pasting cells

### LINKED ISSUE #5656: Bug: Full table selected after cells copy/paste
I'm working with lexical tables and I noticed a change with the last version. When I select a bunch of cells and try to copy/paste it, the full table is selected after pasting. This doesn't happen if I explicitly select the area where I want to paste my cells.

Lexical version: 0.13.1

## Steps To Reproduce

https://www.loom.com/share/9960356e8d2b488ca9e95edc9357dc39?sid=70968f4c-2d27-4ab3-8f33-54e032a69128

Link to code example: Lexical Playground

## The current behavior

After copying and pasting cells without selecting a cell range, the full table is selected

## The expected behavior

After copying and pasting cells without selecting a cell range, only the pasted cells range should be selected

ivailop7:
Fixes #5656

Upon pasting a cells selection in a table when the target area is not a cells selection inside a table, it would select the whole table after the paste event. It worked correctly (not select the whole table) if the selected area to paste to was a cells selection.

Before:

https://github.com/user-attachments/assets/b6e75f2e-2ede-4da4-8ba4-2839c0b194c3

After:

https://github.com/user-attachments/assets/87d209f3-0feb-4833-a7b1-34193852d1f2

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -695,7 +695,6 @@ export function applyTableHandlers(
             newAnchorCellKey,
             newFocusCellKey,
           );
-          $setSelection(newTableSelection);
         }
         return true;
       },
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -66,10 +66,7 @@ import {$isTableCellNode} from './LexicalTableCellNode';
 import {$isTableNode} from './LexicalTableNode';
 import {TableDOMTable, TableObserver} from './LexicalTableObserver';
 import {$isTableRowNode} from './LexicalTableRowNode';
-import {
-  $createTableSelection,
-  $isTableSelection,
-} from './LexicalTableSelection';
+import {$isTableSelection} from './LexicalTableSelection';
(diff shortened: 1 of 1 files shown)


zurfyx (approved):

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -695,7 +695,6 @@ export function applyTableHandlers(
             newAnchorCellKey,
             newFocusCellKey,
           );
-          $setSelection(newTableSelection);
```
Shubhankerism, line 698: this entire `if` block can be removed

ivailop7: cleaning up

etrepum (approved):

moughxyz:
I was actually just investigating an issue with whole table selection as well :) https://github.com/facebook/lexical/pull/6542

I'm wondering if it should be ok for a PR like the one that was merged here to go through without tests? The problem is that the change made here is now indefensible against future changes. No one will know whether it's safe to modify code in this area, and no one will know whether their changes break something here.

I mention this in my PR:

> It's also my opinion that for large open source projects like this, tests should be the ultimate way of determining what stays and what goes. At some point there will no longer be anyone involved in the project who has all the historical context necessary to determine whether a change is harmful or beneficial, and its up to the tests to defend each change.

PR #6544, opened 2024-08-23
[lexical-table] Fix cut for pure text within cell

ivailop7:
Broke 'Cut' for pure text from within a table cell

Before: 

https://github.com/user-attachments/assets/ab143d1e-39d7-420c-a41c-a344a4aa3117

After:
Pastes correctly

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -396,8 +396,10 @@ export function applyTableHandlers(
             $deleteCellHandler(event);
             return true;
           } else if ($isRangeSelection(selection)) {
-            $deleteCellHandler(event);
-            selection.removeText();
+            editor.update(() => {
+              $deleteCellHandler(event);
+              selection.removeText();
+            });
             return true;
           }
         }


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -396,8 +396,10 @@ export function applyTableHandlers(
             $deleteCellHandler(event);
             return true;
           } else if ($isRangeSelection(selection)) {
-            $deleteCellHandler(event);
-            selection.removeText();
+            editor.update(() => {
```
ivailop7, line 399: turns out this was indeed needed as originally written, @etrepum

etrepum: I have a hard time believing that since the above `$isTableSelection` code doesn't use it and every command is processed in an `editor.update`

https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalUpdates.ts#L748-L754

etrepum: Maybe #6546 is the fix and it just seemed like the editor.update was related? I'm not sure what happened with that, maybe we the extended tests didn't run in full before #6542 was added to the merge queue

ivailop7: Unfortunately not, it was my commit, you can try the current playground and the preview for this PR. Maybe something deeper is happening, but this definitely makes a difference in behaviour.

etrepum: Ok so what's happening here is that `copyToClipboard` has an internal `editor.update` and the current semantics of `editor.update` are kinda broken.

```ts
editor.update(() => {
  console.log('A');
  editor.update(() => { console.log('B'); });
  console.log('C');
  editor.update(() => { console.log('D'); });
});
```

This will log 'A', 'C', 'B', 'D' although I suspect that most people would assume that it does not work like this.

etrepum: There's also the race condition because copyToClipboard is asynchronous and it gets the contents of the editor "later", it should really be refactored to get the clipboard data synchronously and then dispatch the copy event with that data rather than computing it from whatever the editor's state is when that event fires

ivailop7: Ok, let's merge this one to fix the current issue. A refactor can be done later.

etrepum: I have a better fix already implemented in #6550

PR #6545, opened 2024-08-23
[lexical-table] Support for vertical cell writing

ivailop7:
Resolves: #4437

https://github.com/user-attachments/assets/6fc1d1a8-2302-4a39-867b-c39c5c8fa54b

Mechanics:
Chaging the CSS attribute writing-mode to 'vertical-rl' or 'vertical-lr' makes the text direction vertical. Once the text is vertical, most people would center it, to make this work seamlessly, we need to convert the text alignment 'left' to 'bottom', 'center' to 'middle', right to 'top', etc. and apply on TableCell-level. I've updated the FORMAT_ELEMENT_COMMAND in a separate PR (#6582) so we don't have the TableCell coupled to the paragraph within.

if no writing-mode is defined - assume the default horizontal, if writing-mode is present validate only 'vertical-rl' and 'vertical-lr' and ignore all the other deprecated properties as per https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode#values


ivailop7:
@etrepum can you give this one a look, if you are OK with the direction.

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -489,6 +489,30 @@ function TableActionMenu({
     [editor],
   );
 
+  const toggleTextDirection = useCallback(() => {
+    editor.update(() => {
+      const selection = $getSelection();
+      if ($isRangeSelection(selection) || $isTableSelection(selection)) {
+        const [cell] = $getNodeTriplet(selection.anchor);
```
etrepum, line 496: The source for `$getNodeTriplet` seems a little suspect since it effectively uses `selection.anchor.getNode()` which could be possibly higher up in the table than inside of a cell? Seems kinda like a bad API to take a point or lexical node when we don't have an `$isLexicalNode` or `isPoint` guard for it to do that cleanly

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -489,6 +489,30 @@ function TableActionMenu({
...
+        const currentTextDirection = cell.getCellDirection();
+        const newTextDirection =
+          currentTextDirection === 'horizontal' ? 'vertical' : 'horizontal';
+        cell.setCellDirection(newTextDirection);
```
etrepum, line 500: maybe implement this as a `cell.toggleCellDirection()` which returns the new direction?

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -231,6 +255,35 @@ export class TableCellNode extends ElementNode {
...
+
+  getCellDirection(): 'horizontal' | 'vertical' {
+    if (this.getLatest().__writingMode) {
+      return 'vertical';
+    } else {
+      return 'horizontal';
+    }
+  }
+
+  setCellDirection(direction: 'horizontal' | 'vertical'): void {
```
etrepum, line 274: I think the general style is for node mutators to return this (or more accurately, the result of `this.getWritable()`) so you can do method chaining

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -231,6 +255,35 @@ export class TableCellNode extends ElementNode {
...
+    if (direction === 'vertical') {
+      const childTextDirection = this.getChildTextDirection();
+      if (childTextDirection === 'rtl') {
+        this.getWritable().__writingMode = 'vertical-rl';
+      } else {
+        this.getWritable().__writingMode = 'vertical-lr';
+      }
```
etrepum, line 281: This seems fragile since the text direction of the child could change and it wouldn't be reflected here?

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -164,6 +182,11 @@ export class TableCellNode extends ElementNode {
       element_.style.verticalAlign = 'top';
       element_.style.textAlign = 'start';
 
+      const writingMode = this.getCellDirection();
+      if (writingMode !== null) {
```
etrepum, line 186: This type is never null?

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -311,6 +367,10 @@ export function $convertTableCellNodeElement(
   if (backgroundColor !== '') {
     tableCellNode.__backgroundColor = backgroundColor;
   }
+  const writingMode = domNode_.style.writingMode;
+  if (writingMode) {
+    tableCellNode.__writingMode = writingMode;
+  }
```
etrepum, line 373: This seems fragile since there are writing modes other than `vertical-{rl,lr}` but other code in here assumes that if `__writingMode` is set then it is vertical (`getCellDirection`). Should have an explicit parse in here to only set writing modes that are understood (e.g. not `horizontal-tb` or any of the invalid, deprecated, experimental or global values that could be set here)

etrepum (commented):
Overall this seems like a reasonable approach but I am most concerned about the coupling between the cell node and its first child paragraph (unclear why we even really need to have that nesting, maybe something to do with shadow root assumptions?). In order to propagate the ltr/rtl we might want to have a node transform handle that (mark the parent dirty if the child changes its ltr/rtl). Might also make sense to handle the writingMode resolution entirely in createDOM/updateDOM (but I think the transform may still be required for the parent to be updated when the child changes its writing mode?)

ivailop7:
The quality of the PR wasn't bad, but no time to finish it. Someone else might use it as a starting point later on.

PR #6547, opened 2024-08-24
[lexical-table] feat: Add row striping

ivailop7:
Adding support for row striping, increases tables visual accessiblity.

https://github.com/user-attachments/assets/cbd92c66-0620-4e70-8346-bd8f74a3817b

There is a bug I'm struggling to pin down: once rowStriping is enabled via the Table Actions Plugin ONLY, the tableObserver is lost and any following changes - "Add Row Header", "Remove Column Header", "Toggle Row Striping", crash.


ivailop7:
@etrepum could you help me on this one, there something happening on a deeper level that I'm unable to pin down. If you run my branch and create a table -> enable row striping -> disable row striping (or add col header, etc.) it would crash, because it can no longer find a table observer. Not urgent, when you have some free time.

etrepum:
@ivailop7 I've added a commit that I think fixes the issue you're having, I did not do the rebase/conflict resolution

ivailop7:
> @ivailop7 I've added a commit that I think fixes the issue you're having, I did not do the rebase/conflict resolution

It does work, thanks Bob. A very interesting commit. I'll handle the remaining details.

etrepum:
I think the root cause was the incorrect type and management of the property that's added to the element and it's just a cast so there was no compiler checking. I added the abort signal because I also noticed that the listener cleanup there wasn't comprehensive when auditing the code. Probably would be better off moving most of that code out of effects for cross-platform and just simplicity reasons.

ivailop7:
> I think the root cause was the incorrect type and management of the property that's added to the element and it's just a cast so there was no compiler checking. I added the abort signal because I also noticed that the listener cleanup there wasn't comprehensive when auditing the code. Probably would be better off moving most of that code out of effects for cross-platform and just simplicity reasons.

When you say property, do you mean the TableNode rowStriping one?

Unfortunately, the issue seems to be still there. As in table selection gets broken after toggling row striping, no highlight shows up when doing a table selection or cells resizing works, until you do a table selection and then resizing cells works, so the event of enabling rowStriping is still the breaking point.

etrepum:
I meant the property attached to HTMLElement to store the TableObserver. Can you type up repro steps for what's happening now?

ivailop7:
> I meant the property attached to HTMLElement to store the TableObserver. Can you type up repro steps for what's happening now?

1. Create a table
2. Enable row striping
3. Try to resize the table (cursor doesn't change to two-way arrow) -> Can't
4. Try to do a table selection -> No highlight displayed, but selection is done. (The observer issue)
5. Now try to resize the table -> OK

ivailop7:
> > I meant the property attached to HTMLElement to store the TableObserver. Can you type up repro steps for what's happening now?
> 
> 1. Create a table
> 2. Enable row striping
> 3. Try to resize the table (cursor doesn't change to two-way arrow) -> Can't
> 4. Try to do a table selection -> No highlight displayed, but selection is done. (The observer issue)
> 5. Now try to resize the table -> OK

@etrepum let me know, if you still are struggling replicating. Will do a recording.

etrepum:
I've just been AFK all day, asked for repro instructions to save time for when I am back at a computer

ivailop7:
> I've just been AFK all day, asked for repro instructions to save time for when I am back at a computer

Sure, no rush. Appreciate the help here.

etrepum:
It looks like the problem is the updateDOM returning true, which creates a new HTMLElement and does not initialize a new TableObserver. Mutation listeners are kind of a bad abstraction for managing DOM because "update" could mean that an in-place DOM update occurred OR that the DOM was replaced (e.g. logically destroy then create would make more sense for this case, at least as far as DOM tracking goes). The simplest fix is to make the DOM update row striping in-place.

etrepum:
I've added a commit that does both fixes, the in-place update is more efficient and would've required less changes by itself. I thought it also made sense to handle updateDOM true case to demonstrate how to do it correctly and in case it's subclassed with an updateDOM that can return true.

ivailop7:
This is ready for review, can I get a review. Big shout out to Bob for helping on the Table Observer issues. @etrepum @zurfyx @potatowagon

etrepum (approved):
LGTM! but of course I had a hand in fixing some of the edge cases so it's probably worth another set of eyes

PR #6548, opened 2024-08-24
[lexical-table] Fix table selection paste as plain text

### LINKED ISSUE #6541: Bug: "text/plain" clipboard data of copied table cells is incorrect
When copying table cells from Lexical, the `text/plain` data is incorrect – it contains duplicated/nested versions of the content.

Copying this table:
image

Results in this text content when pasted in a text editor:
image

Lexical version: `v0.17.0+dev.esm` (current Lexical Playground version)

## Steps To Reproduce

1. Insert a table into the Lexical Playground
2. Select cells from the table, then copy the cells
3. Open a plain text editor and paste the content.
4. Expect just the original cell content, observe that it contains duplicated versions of the original content

## The current behavior

`text/plain` clipboard data of copied table cells contains duplicated content.  

## The expected behavior

`text/plain` clipboard data should be a simple representation of the original table cell content, with no duplication

## Impact of fix

Any users of tables will get correct content when pasting into plain text environments. The current issue does not seem to impact copy paste of the `text/html` or `application/x-lexical-editor` versions of the data on the clipboard.

ivailop7:
Fixes #6541

Before:
Double pasting incorrectly, since it was reading all nodes - rows and cells.

https://github.com/user-attachments/assets/8ddb613e-1a6b-4985-8052-ff51b9b20982

After:
Pasting correctly, since reading only cells.

https://github.com/user-attachments/assets/c1484903-388a-4f96-86f7-fb97322cd8cc

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelection.ts
+++ b/packages/lexical-table/src/LexicalTableSelection.ts
@@ -336,7 +336,7 @@ export class TableSelection implements BaseSelection {
     for (let i = 0; i < nodes.length; i++) {
       const node: TableCellNode = nodes[i];
       const row = node.__parent;
-      const nextRow = nodes[i + 1 < nodes.length ? i + 1 : 0].__parent;
+      const nextRow = (nodes[i + 1] || {}).__parent;
       textContent +=
         (node.getTextContent() || ' ') + (nextRow !== row ? '\n' : '\t');
     }
--- a/packages/lexical-table/src/LexicalTableSelection.ts
+++ b/packages/lexical-table/src/LexicalTableSelection.ts
@@ -331,10 +331,10 @@ export class TableSelection implements BaseSelection {
   }
 
   getTextContent(): string {
-    const nodes = this.getNodes();
+    const nodes = this.getNodes().filter((node) => $isTableCellNode(node));
     let textContent = '';
     for (let i = 0; i < nodes.length; i++) {
-      textContent += nodes[i].getTextContent();
+      textContent += nodes[i].getTextContent() + '\n';
     }
     return textContent;
   }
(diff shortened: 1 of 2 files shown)


etrepum (commented):
Isn't it expected for the content to be the rows delimited by newlines and the cells delimited by tabs?

ivailop7:
> Isn't it expected for the content to be the rows delimited by newlines and the cells delimited by tabs?

One of the more popular editors I tried does it this way - cell per new line, which I think makes more sense to manipulate further if one wanted the output as plain text. Another one I tried does it the way you decribed. I don't mind chaning the paste structure, I'm unsure on the right output layout.

etrepum:
The default behavior of the browser is to copy tables with tabs delimiting cells and newlines delimiting rows

ivailop7:
> The default behavior of the browser is to copy tables with tabs delimiting cells and newlines delimiting rows

updated,

https://github.com/user-attachments/assets/779b6056-fc94-4b37-bfa3-1edea71bff1c

I'm sure there's a more elegant way to concat correctly, but it does the job.

packages/lexical-table/src/LexicalTableSelection.ts
```diff
@@ -331,10 +331,14 @@ export class TableSelection implements BaseSelection {
...
-    const nodes = this.getNodes();
+    const nodes = this.getNodes().filter((node) => $isTableCellNode(node));
     let textContent = '';
     for (let i = 0; i < nodes.length; i++) {
-      textContent += nodes[i].getTextContent();
+      const node = nodes[i];
+      const row = node.__parent;
+      const nextRow = (nodes[i + 1] || {}).__parent;
+      textContent +=
+        (node.getTextContent() || ' ') + (nextRow !== row ? '\n' : '\t');
```
etrepum, line 341: I don't think the `|| ' '` matches browser expectations (when I tried it with a table the browser doesn't insert whitespace that isn't in the content, other than the delimiters) but the rest looks good to me. Would be nice to have some tests!

ivailop7: updated, added a test

etrepum (approved):
Looks good!

PR #6555, opened 2024-08-26
[lexical-playground] Bug Fix: fix comment timestamps

### LINKED ISSUE #6554: Bug: Playground comments have inconsistent timestamps in different window contexts
When adding collaborative comments, the timestamps are inconsistent in different window contexts.

Lexical version: 0.17.0

## Steps To Reproduce

1. Clone this repo and do relevant `npm i`.
2. In the playground package, run `npm run dev`.
3. Open http://localhost:3000/?isCollab=true in 2 different tabs.
4. Add some text and add a comment.
5. Wait a minute until the comment time stamp says "1 min. ago".
6. Reload only one browser tab.
7. Observe the comment time says "Just now" which is now different from the other tab.

Link to code example: use the playground in this repo (see reproduction steps above).

## The current behavior
Comment timestamps are inconsistent in different window contexts.

## The expected behavior
Comment timestamps are the same in different window contexts.

## Impact of fix
This isn't affected if you test with the **Split Screen** option as it's always the same window context. But for all other usual scenarios the inconsistent timestamps are disconcerting.

irahopkinson:
## Description

When adding collaborative comments in the playground, the comment timestamps are now consistent even in different window contexts.

Closes #6554

## Test plan

Follow the reproduction steps in issue #6554 and see that the comment timestamps are now the same.

### Before

Comment timestamps are inconsistent in different window contexts.

### After

Comment timestamps are the same in different window contexts.

## Note

There are other uses of `performance.now()` in this repo that might need to be reviewed.

### DIFF
--- a/packages/lexical-playground/src/commenting/index.ts
+++ b/packages/lexical-playground/src/commenting/index.ts
@@ -56,7 +56,10 @@ export function createComment(
     content,
     deleted: deleted === undefined ? false : deleted,
     id: id === undefined ? createUID() : id,
-    timeStamp: timeStamp === undefined ? performance.now() : timeStamp,
+    timeStamp:
+      timeStamp === undefined
+        ? performance.timeOrigin + performance.now()
+        : timeStamp,
     type: 'comment',
   };
 }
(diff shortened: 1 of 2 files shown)


etrepum (approved):
Looks good, thank you!

etrepum:
I don't think that any of the other `performance.now()` calls are distributed

PR #6557, opened 2024-08-26
[lexical-website] Feature: Traditional Chinese Localization

Edit-Mr:
## Description

Add Traditional Chinese Localization for docs

Here's an demo I deployed: <https://lexical-lexical-website-git-demo-editmrs-projects.vercel.app/zh-Hant/>


ivailop7:
Hi there, the deployment still fails. Also, I'm unsure how the maintenance is going to be done of this documentation going forward and keep it in sync with the english language one.

etrepum:
I agree with @ivailop7, this would be very difficult for the existing team to maintain. Maybe it would be better to maintain this externally, and when Lexical is more stable (>1.0 for sure) and has a larger team then official translations could be reconsidered, possibly following the [translations.react.dev](https://translations.react.dev/) model

childrentime:
I'm interested in adding Simplified Chinese documentation.

etrepum:
Closing as it's stale and translations can't be maintained by the current team at this stage of the project

PR #6558, opened 2024-08-26
Revert "[lexical-playground] Fix: in playground show component-menu w…

zurfyx:
Breaks internally, might need to revise the approach or our internal code...

…hen scroll (#6510)"

This reverts commit fd7a93d00359f391398adeef7aea6f783830cc8a.

## Description

Closes #

### Before

### After

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -494,7 +494,9 @@ export function useMenuAnchorRef(
     if (rootElement !== null && resolution !== null) {
       const {left, top, width, height} = resolution.getRect();
       const anchorHeight = anchorElementRef.current.offsetHeight; // use to position under anchor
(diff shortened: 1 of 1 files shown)


acywatson (approved):

etrepum:
Can you put together a repro that shows what goes wrong with this approach? It fixed all of the positioning issues that we could find, this revert is an OSS regression

etrepum:
Yeah, Floating UI is what I used in the past for doing this sort of thing and it worked well. I don't have a strong opinion on whether we should have a dependency handle floating UI or have an in-tree implementation without known issues but we should certainly not have an in-tree implementation that we know is incorrect for the cases that we demonstrate.

potatowagon:
> Can you put together a repro that shows what goes wrong with this approach? It fixed all of the positioning issues that we could find, this revert is an OSS regression

reintroducing the changes back in https://github.com/facebook/lexical/pull/6566 but with gating so we can investigate internally and fix intern

PR #6562, opened 2024-08-27
[lexical-code][breaking change] Bug Fix: explicitly import instead of `window.` to support code nodes in nodejs

### LINKED ISSUE #6563: Bug: Unable to create CodeNodes in a nodejs headless environment
Lexical version: 0.17.1

## Steps To Reproduce

1. Create a headless editor in a node.js environment
2. Create a code node
3. You'll see this error appear where says `window` is undefined

## The current behavior

Runtime error is thrown when trying to create a code node block in a server environment

## The expected behavior

Error is now thrown

nadine-nguyen:
## Description

Remove reference to `window` in `CodeNode` so the package can be used in node js. This is done by referencing Prism conditionally using `global.Prism` if it is in an environment where `window` is undefined

Closes #6563

### Before

`window is undefined` error appears when creating a code node in a headless editor in a nodejs environment 

### After

No error appears


ivailop7 (dismissed):

ivailop7:
There seems to be a build issue with Astro in the integration tests

scripts/__tests__/integration/fixtures/lexical-esm-astro-react › install & build succeeded

    ChildProcessError: Command failed: npm run build
    Named export 'languages' not found. The requested module 'prismjs' is a CommonJS module, which may not support all module.exports as named exports.
    CommonJS modules can always be imported via the default export, for example using:

    import pkg from 'prismjs';

PrismJS is kind of abandoned and doesn't supprot ESM, we should probably move to Shiki at this point.

ivailop7:
For this particular case, to unblock, since all the supported languages are listed in here: https://github.com/facebook/lexical/blob/main/packages/lexical-code/src/CodeHighlighterPrism.ts
You can export a const set with the listed languages here and check that particular set. Suboptimal, but should get things working.

nadine-nguyen:
@etrepum Thanks for the link!
@ivailop7 I've decided to switch between `window.Prism` vs `global.Prism`, depending on whether `window` is defined. The integration check is now passing with this change. Let me know your thoughts

etrepum (approved):
Looks good to me, but no idea if it will break Meta's build or not

ivailop7 (approved):

ivailop7:
@potatowagon could you check if this change would work for you internally?

potatowagon:
> @potatowagon could you check if this change would work for you internally?



> @potatowagon could you check if this change would work for you internally?

thanks for the headsup, labeled this as breaking change as a memo for further testing internally

PR #6566, opened 2024-08-27
[lexical-react] menu positioning: Unrevert PR6510 but with gating

potatowagon:
## Description
https://github.com/facebook/lexical/pull/6510 is correct but it broke implementations internally

hence it got reverted to mitigate the internal emergency: https://github.com/facebook/lexical/pull/6558

here, https://github.com/facebook/lexical/pull/6510 is reintroduced but gated behind the shouldIncludePageYOffset option to aid investigations internally and to fix internal to be compatible 

## Test plan

shouldIncludePageYOffset = true

https://github.com/user-attachments/assets/8721cc41-3486-4b0b-83c3-008b3ec4c173

---
shouldIncludePageYOffset = false

https://github.com/user-attachments/assets/db950846-7989-4489-9a33-36ef59cd0ceb

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -482,6 +482,7 @@ export function useMenuAnchorRef(
   setResolution: (r: MenuResolution | null) => void,
   className?: string,
   parent: HTMLElement = document.body,
+  shouldIncludePageYOffset: boolean = true,
 ): MutableRefObject<HTMLElement> {
   const [editor] = useLexicalComposerContext();
   const anchorElementRef = useRef<HTMLElement>(document.createElement('div'));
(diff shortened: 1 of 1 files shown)


zurfyx (dismissed):

packages/lexical-react/src/shared/LexicalMenu.ts
```diff
@@ -482,6 +482,7 @@ export function useMenuAnchorRef(
   setResolution: (r: MenuResolution | null) => void,
   className?: string,
   parent: HTMLElement = document.body,
+  shouldIncludePageYOffset: boolean = true,
```
zurfyx, line 485: Prob just fine for a temporary use case but otherwise I'd be keen a suffix __EXPERIMENTAL or __DEPRECATED so that it doesn't leak into product code in OSS which is harder to deprecate

potatowagon: suffixed with __EXPERIMENTAL

zurfyx (approved):

PR #6578, opened 2024-08-30
[lexical-react]: Fix incorrect addition of empty cells on table paste

Shubhankerism:
This PR fixes the issue with empty cells getting incorrectly added to a table upon paste.

## Before
When a table with rows with unequal sizes and merged cells is pasted, we add the empty cells in the wrong row.
For example, a table like this: 
image

table-transform-before

## After
Empty cells are now added to the correct row.

table-transform-after

### DIFF
--- a/packages/lexical-react/src/LexicalTablePlugin.ts
+++ b/packages/lexical-react/src/LexicalTablePlugin.ts
@@ -88,20 +88,18 @@ export function TablePlugin({
           return Math.max(curLength, row.length);
         }, 0);
         for (let i = 0; i < gridMap.length; ++i) {
-          const rowLength = gridMap[i].length;
+          const row = gridMap[i].filter((cell) => cell);
+          const rowLength = row.length;
           if (rowLength === maxRowLength) {
             continue;
           }
-          const lastCellMap = gridMap[i][rowLength - 1];
-          const lastRowCell = lastCellMap.cell;
+          const rowNode: TableRowNode | null = node.getChildAtIndex(i);
           for (let j = rowLength; j < maxRowLength; ++j) {
             // TODO: inherit header state from another header or body
             const newCell = $createTableCellNode(0);
             newCell.append($createParagraphNode());
-            if (lastRowCell !== null) {
-              lastRowCell.insertAfter(newCell);
-            } else {
-              $insertFirst(lastRowCell, newCell);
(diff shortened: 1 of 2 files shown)


ivailop7:
Fantastic! Well done! Please add a test for this copy-paste use case, there are quite a few to take inspiration from, so it should be easy to add one.

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -88,20 +88,18 @@ export function TablePlugin({
           return Math.max(curLength, row.length);
         }, 0);
         for (let i = 0; i < gridMap.length; ++i) {
-          const rowLength = gridMap[i].length;
+          const row = gridMap[i].filter((cell) => cell);
+          const rowLength = row.length;
```
etrepum, line 92: Since you're not using this array it would be more efficient to calculate it directly
```suggestion
          const rowLength = gridMap[i].reduce((acc, cell) => cell ? 1 + acc : acc, 0);
```

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -88,20 +88,18 @@ export function TablePlugin({
...
           if (rowLength === maxRowLength) {
             continue;
           }
-          const lastCellMap = gridMap[i][rowLength - 1];
-          const lastRowCell = lastCellMap.cell;
+          const rowNode: TableRowNode | null = node.getChildAtIndex(i);
```
etrepum, line 96: It would be more efficient to compute `node.getChildren()` outside of the loop and index it directly down here since each `getChildAtIndex` is a linked list traversal in a loop this ends up being geometric instead of linear.

Could also do the `if (rowNode)` here and continue or wrap it around the loop so you can skip the loop if it is merged.

etrepum (commented):
I'd love to see some tests with various merged cell conditions, I'm not familiar enough with this code to judge if it's a general fix or only handles the cases you ran into. The indexing seems maybe a bit suspicious (e.g. if there are merges in the middle of the table), but I didn't review `$computeTableMapSkipCellCheck`

ivailop7 (approved):

ivailop7:
@etrepum good with this one?

etrepum (approved):

PR #2638, opened 2022-07-13
$insertNodes: Selection-agnostic node insertion with Grid/Node selection support

zurfyx:
Node selection

https://user-images.githubusercontent.com/193447/178761488-d39987f7-2a90-4604-b4fc-dfbf57178f35.mov

Grid selection

https://user-images.githubusercontent.com/193447/178761618-316b267b-2105-48d5-9714-4a9268e0fc1b.mov

PR #6579, opened 2024-08-30
Bug Fix: Fix issue where selecting a cell then dragging outside of table would not select entire table

moughxyz:
This fixes a [regression](https://github.com/facebook/lexical/pull/6542#issuecomment-2318374970) caused by [this PR](https://github.com/facebook/lexical/pull/6542).

The PR above aimed to solve a destructive issue where triple-clicking in a table could inadvertently select the entire document. The solution caused a regression where dragging out from inside a table wouldn't select the entire table, even though the highlight made it seem like it was selected.

I spent a bit of time trying to fix both the regression and the original triple-click issue but unfortunately had to reach a compromise where I could fix the regression, and fix the triple-click issue, with the compromise that triple-clicking the last cell would end up selecting the first cell (rather than the whole document). That's fine enough for me, so long as the whole document isn't selected, but isn't exactly technically correct.

However I couldn't figure out why the first cell was being selected. The code looks right to me and inspecting the selection before it's made tells me that the last cell should be selected, but for some reason, it's being overriden somewhere to the first cell.

In any case I think this PR represents a suitable compromise and we can either merge it now if it makes sense, or wait to see if someone understands what a fix could look like that addresses both issues.

# Test plan

Tests were added and modified. I also added a utility function with the goal of making it easier to read and work with selection objects. Reading other people's selection objects, especially when reviewing the PR, is very hard to contextualize.

For example, this is hard to understand or conceptualize at first glance:

```js
const expectedSelection = {
  anchorOffset: 0,
  anchorPath: [1, 0, 0],
  focusOffset: 9,
  focusPath: [2, 0, 0],
};
```

You'd have to print out the DOM to really verify what's going on here as a reviewer.

I've created a helper function that allows for self-describing names:

```js
    const expectedSelection = createHumanReadableSelection({
      anchorOffset: {desc: 'beginning of cell', value: 0},
      anchorPath: [
        {desc: 'index of table in root', value: 1},
        {desc: 'first table row', value: 0},
        {desc: 'first cell', value: 0},
      ],
      focusOffset: {desc: 'full text length', value: 9},
      focusPath: [
        {desc: 'index of last paragraph', value: 2},
        {desc: 'index of first span', value: 0},
        {desc: 'index of text block', value: 0},
      ],
    });
```

Not required as part of this PR so I can remove it if it's not helpful.

## Before

https://github.com/user-attachments/assets/5c748a13-054e-4440-8669-75700e923d05

## After

https://github.com/user-attachments/assets/116d7ad8-7fbd-4d24-b54a-189cc1cf2454

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -709,7 +709,7 @@ test.describe.parallel('Selection', () => {
     browserName,
     legacyEvents,
   }) => {
-    test.skip(isPlainText);
+    test.skip(isPlainText || isCollab);
 
     await focusEditor(page);
     await insertTable(page, 1, 2);
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -714,15 +714,15 @@ test.describe.parallel('Selection', () => {
     await focusEditor(page);
     await insertTable(page, 1, 2);
     await moveToEditorEnd(page);
-    const lastCellText = 'Some text';
-    await page.keyboard.type(lastCellText);
+
+    const endParagraphText = 'Some text';
+    await page.keyboard.type(endParagraphText);
 
     const lastCell = page.locator(
       '.PlaygroundEditorTheme__tableCell:last-child',
     );
     await lastCell.click();
-    const cellText = 'Foo';
-    await page.keyboard.type(cellText);
+    await page.keyboard.type('Foo');
 
     const expectedSelection = createHumanReadableSelection({
       anchorOffset: {desc: 'beginning of cell', value: 0},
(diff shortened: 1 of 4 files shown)


moughxyz:
FYI this test is pretty flakey: https://github.com/facebook/lexical/blob/main/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs#L1307

Running it locally on macOS or Ubuntu 22.04 with `cross-env E2E_BROWSER=chromium E2E_EDITOR_MODE=rich-text-with-collab E2E_EVENTS_MODE=modern-events playwright test --project=\"chromium\"`, I can get it to fail 1/5 times or so.

In debug I can see that the left frame correctly inserts the image in the second column second row, but on the right frame, it inserts the image in the first column second row. The test flakes even without the changes in this PR and without the changes in https://github.com/facebook/lexical/pull/6542.

The test was introduced in https://github.com/facebook/lexical/pull/2638 by @zurfyx.

moughxyz:
@ivailop7 ready for review 🙏 If any suggestions please let me know.

ivailop7:
Will have a look next week.

etrepum (approved):
Looks like it works to me, and all of the test suites are passing

PR #6582, opened 2024-08-31
[lexical-table] Fix table cell format for selection within cell

ivailop7:
Currently, the TableCellNode tracks the format of the paragraph inside. This is applied correctly when selection is via table selection (range selection including a table, whole table selection, cells selection). 

There is a collection of cases if it's selection within the table cell, since the table cell is not selected the format doesn't propagate to the parent TableCellNode correctly and things get out of sync. This fixes those cases.

Before:

https://github.com/user-attachments/assets/bb1feffc-ab08-4232-9e86-3735eca43335

After:

Things are in sync

https://github.com/user-attachments/assets/fb459ff0-0898-4c81-a2e4-a001f796d86a

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -38,6 +38,7 @@ import {
   $getSelection,
   $isDecoratorNode,
   $isElementNode,
+  $isNodeSelection,
   $isRangeSelection,
   $isRootOrShadowRoot,
   $isTextNode,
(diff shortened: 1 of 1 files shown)


moughxyz:
Hi @ivailop7, I know it's not my place to say. But given you are able to reproduce and demonstrate this issue in the playground, it should also be possible to write an e2e test for this. Given there are many organizations that rely on this codebase, would you be willing to write one, or otherwise specify in your PR description why tests were not included for this change?

etrepum:
This kinda seems like something that should be handled by the selection class rather than a command? Since we do have a paragraph in every cell maybe it makes sense not to store this format there at all since it can and should already be stored on the paragraph (per #6583)

ivailop7:
> This kinda seems like something that should be handled by the selection class rather than a command? Since we do have a paragraph in every cell maybe it makes sense not to store this format there at all since it can and should already be stored on the paragraph (per #6583)

Yeah, that's probably a better approach and would make preserving the styling on empty cells more consistent. Need to redo it, but my holidays ended, so don't have as much time to do PRs. Will revisit in the coming weeks.

PR #6598, opened 2024-09-05
fix: delete horizontal node in table cell node

### LINKED ISSUE #6459: Bug: HorizontalRuleNode breaks TableCell in some case

keiseiTi:
## Description

fix #6459 

This commit solves the problem of continuing to execute subsequent logic when the first node in the ElementNode is decoratorNode

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1703,11 +1703,12 @@ export class RangeSelection implements BaseSelection {
       if (this.forwardDeletion(anchor, anchorNode, isBackward)) {
         return;
       }
-      if ($isElementNode(anchorNode)) {
-        const childNode = anchorNode.getChildAtIndex(0);
-        if ($isDecoratorNode(childNode) && anchor.offset === 0) {
-          return;
-        }
+      if (
+        anchor.offset === 0 &&
+        $isElementNode(anchorNode) &&
+        $isDecoratorNode(anchorNode.getFirstChild())
+      ) {
+        return;
       }
       // Handle the deletion around decorators.
       const focus = this.focus;
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1704,6 +1704,7 @@ export class RangeSelection implements BaseSelection {
         return;
       }
       if (
+        isBackward &&
         anchor.offset === 0 &&
         $isElementNode(anchorNode) &&
         $isDecoratorNode(anchorNode.getFirstChild())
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1703,7 +1703,12 @@ export class RangeSelection implements BaseSelection {
       if (this.forwardDeletion(anchor, anchorNode, isBackward)) {
         return;
       }
-
+      if ($isElementNode(anchorNode)) {
+        const childNode = anchorNode.getChildAtIndex(0);
+        if ($isDecoratorNode(childNode) && anchor.offset === 0) {
+          return;
+        }
+      }
```
etrepum, line 1711: I'd probably re-order this to short-circuit with fewer function calls, e.g.

```suggestion
      if (anchor.offset === 0 && $isElementNode(anchorNode) && $isDecoratorNode(anchorNode.getFirstChild())) {
        return;
      }
```

etrepum (commented):
It would be great to have a test that demonstrates that this does fix it

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1703,7 +1703,14 @@ export class RangeSelection implements BaseSelection {
       if (this.forwardDeletion(anchor, anchorNode, isBackward)) {
         return;
       }
-
+      if (
+        isBackward &&
+        anchor.offset === 0 &&
+        $isElementNode(anchorNode) &&
+        $isDecoratorNode(anchorNode.getFirstChild())
```
etrepum, line 1710: I think the best solution is probably `$isRootOrShadowRoot(anchorNode)`? It would be nice to add a test and maybe a comment around here to describe what we're trying to do here

zurfyx:
Thank you, can you expand on the problem? I believe there's an issue with selection and shadow root, backspace should never see beyond a shadow root. However, the horizontal line appears to be deletable as intended.

etrepum:
The problem is that the decorator logic doesn't check for a shadow root. When you delete the paragraph before the HR, put the selection there (offset 0 of the TableCellNode), delete again it ends up deleting the TableCellNode when the expected behavior would be a no-op

etrepum:
Are you planning to finish up this PR?

Issue #5853, opened 2024-04-09
Bug: There may be exceptions when merging tables


chaihan:
Lexical version:0.14.3

## Steps To Reproduce

1.Create a 5x5 table (not the experimental) in the Lexical Playground.
2.Merge two cells
3.After merging, merge the two cells below
4.Merge two merged cells again

Link to code example:

main.d4bb59e0.js:21 Uncaught Error: Internal Lexical error: invariant() is meant to be replaced at compile time. There is no runtime version. Error: updateEditor: selection has been lost because the previously selected nodes have been removed and selection wasn't moved to another node. Ensure selection changes after removing/replacing a selected node.
-->

## The current behavior
1.
![1](https://github.com/facebook/lexical/assets/70357130/79dbbf18-59e9-415e-9796-59c3c81ae87b)
2.
![2](https://github.com/facebook/lexical/assets/70357130/27a3e997-0b5d-4d90-a940-efe560653073)
3.
![3](https://github.com/facebook/lexical/assets/70357130/b55cd3cc-5918-454d-8dae-df210bded080)

## The expected behavior
Can merge normally

chaihan:
Will this issue be fixed later

ivailop7:
Duplicate of #4470

PR #6599, opened 2024-09-05
fix: merging cells onto already merged cells throws error

### LINKED ISSUE #4470: Bug: merging two merged cell --> Error #19: selection has been lost
When merging two merged cells , the following error happens: 

`updateEditor: selection has been lost because the previously selected nodes have been removed and selection wasn't moved to another node. Ensure selection changes after removing/replacing a selected node.`

Lexical version: latest on playground

## Steps To Reproduce

https://user-images.githubusercontent.com/19841175/236870912-cabf1288-148c-4c4b-b31e-25e3a3a8c2c5.mp4

rilrom:
## Description

Closes #4470, #5853.

This fix ensures that the merge cells button does not appear when attempting to merge already merged cells onto unmerged cells. This prevents an error from appearing which eventually causes the table to be unusable.

I do see potential for being able to merge already merged cells onto unmerged cells in certain situations, but that should be revisited separately, for now resolving the error is the priority.

## Test plan

1. Create a table.
2. Attempt to merge already merged cells onto unmerged cells.
3. The merge cells button should not appear in the table action menu.


etrepum:
I think the priority here should be test coverage, to confirm that this does actually fix the issue and to make it easier to properly fix the edge case that this fix is avoiding. If the problem is just that the selection is getting lost, I think that should be straightforward to fix properly.

rilrom:
> I think the priority here should be test coverage, to confirm that this does actually fix the issue and to make it easier to properly fix the edge case that this fix is avoiding. If the problem is just that the selection is getting lost, I think that should be straightforward to fix properly.

Understood, I will create a regression test for it.

etrepum:
After looking closer I think the correct fix here is #6607 but will take some time to write additional tests and probably fix another edge case or two while I'm in there

rilrom:
> After looking closer I think the correct fix here is #6607 but will take some time to write additional tests and probably fix another edge case or two while I'm in there

If you've managed to allow for merging already merged cells onto unmerged cells without throwing an error then I absolutely agree that's the better fix here. 

This has been a relatively long-standing issue so I'm glad I was able to get the discussion rolling.

I'll do some testing on your PR, thanks @etrepum.

PR #6609, opened 2024-09-07
[lexical-playground] Fix: Poll Option not clickable at some place after checked state

### LINKED ISSUE #6551: Bug: The Poll component doesn't get checked / unchecked each time on click
Lexical version: 0.14.3 or >0.14.3

## Steps To Reproduce

1. Insert the Poll component with the default two options.
2. Click on both options repeatedly to check/uncheck them on every click.
3. You will encounter the issue as shown here: https://d.pr/i/vHvKcU (used [lexical-playground](https://playground.lexical.dev/))

Link to code example:
    - [Lexical playground](https://github.com/facebook/lexical/tree/main/packages/lexical-playground).

## Current Behavior
The checkbox does not get checked/unchecked on each click.

## Expected Behavior
The checkbox should be checked/unchecked on each click.

## Impact of Fix
We are building an app where we are implementing the Lexical editor, so it is crucial for us that every component works smoothly.

iamvinayvk:
## Description

Poll option was not clickable after the checked state due to pseudo after element. Added `pointer-events:none` to let the events pass through the pseudo elements on checkbox change 

Closes #6551 

### Before

https://d.pr/i/vHvKcU

### After

https://github.com/user-attachments/assets/87bf261d-9bff-4f5a-a76f-5f48bfd1eae2

### DIFF
--- a/packages/lexical-playground/src/nodes/PollNode.css
+++ b/packages/lexical-playground/src/nodes/PollNode.css
@@ -105,6 +105,7 @@
   margin: 0;
   transform: rotate(45deg);
   border-width: 0 2px 2px 0;
+  pointer-events: none;
 }
 .PollNode__optionCheckbox {
   border: 0px;


etrepum (approved):

mdmuhtasimfuadfahim:
Thanks @etrepum. This solve the issue.

etrepum:
@iamvinayvk is the one to thank 🙌 , I just reviewed it!

PR #6612, opened 2024-09-09
Expose getStyleObjectFromCss in @lexical/selection

mofoshow:
### Problem
There is a need to use `getCSSFromStyleObject` together with `getStyleObjectFromCSS` in order to change a specific style that is applied to a node with `setStyle`. 

`getStyleObjectFromCSS` is available in the @lexical/selection package but not `getCSSFromStyleObject` even though the function exists in the utils file. 

This PR is simply exposing this function that already exists.

### DIFF
--- a/packages/lexical-selection/flow/LexicalSelection.js.flow
+++ b/packages/lexical-selection/flow/LexicalSelection.js.flow
@@ -20,7 +20,7 @@ declare export function $cloneWithProperties<T: LexicalNode>(node: T): T;
 declare export function getStyleObjectFromCSS(css: string): {
   [string]: string,
 };
-declare export funciton getCSSFromStyleObject(styles: Record<string, string>): string;
+declare export function getCSSFromStyleObject(styles: Record<string, string>): string;
 declare export function $patchStyleText(
   selection: BaseSelection,
   patch: {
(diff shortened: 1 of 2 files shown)


packages/lexical-selection/flow/LexicalSelection.js.flow
```diff
@@ -20,6 +20,7 @@ declare export function $cloneWithProperties<T: LexicalNode>(node: T): T;
 declare export function getStyleObjectFromCSS(css: string): {
   [string]: string,
 };
+declare export funciton getCSSFromStyleObject(styles: Record<string, string>): string;
```
etrepum, line 23:
```suggestion
declare export function getCSSFromStyleObject(styles: Record<string, string>): string;
```

etrepum: Don't think this will pass CI

packages/lexical-selection/flow/LexicalSelection.js.flow
```diff
@@ -20,6 +20,7 @@ declare export function $cloneWithProperties<T: LexicalNode>(node: T): T;
 declare export function getStyleObjectFromCSS(css: string): {
   [string]: string,
 };
+declare export function getCSSFromStyleObject(styles: Record<string, string>): string;
```
potatowagon, line 23: copied it over internally, no errors or formatting fixes needed. lgtm

packages/lexical-selection/flow/LexicalSelection.js.flow
mofoshow: gaaah my dyslexia coming out strong. Thank you for the fix

potatowagon (approved):

PR #6617, opened 2024-09-10
[lexical-markdown] Breaking Change: rename 'multilineElement' to 'multiline-element'

potatowagon:
## Description
to be consistent with other types like `text-match`, `text-format`

## Test plan

see automated tests

### DIFF
--- a/packages/lexical-markdown/src/utils.ts
+++ b/packages/lexical-markdown/src/utils.ts
@@ -432,7 +432,7 @@ export function transformersByType(transformers: Array<Transformer>): Readonly<{
 
   return {
     element: (byType.element || []) as Array<ElementTransformer>,
-    multilineElement: (byType.multilineElement ||
+    multilineElement: (byType['multiline-element'] ||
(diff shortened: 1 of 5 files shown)


packages/lexical-markdown/src/utils.ts
```diff
@@ -432,7 +432,7 @@ export function transformersByType(transformers: Array<Transformer>): Readonly<{
 
   return {
     element: (byType.element || []) as Array<ElementTransformer>,
-    multilineElement: (byType.multilineElement ||
+    multilineElement: (byType['multiline-element'] ||
```
potatowagon, line 435: im not sure of the historical reason why kebab case was originally chosen as the type name convention. 

im open to switching everything to lowerCamelCase if theres a compelling reason

etrepum (approved):
This looks fine to me, I don't see a compelling reason to use kebab convention other than backwards compatibility and consistency. We don't export these constants so I think we are kinda stuck with the existing value for text-format.

Issue #6624, opened 2024-09-12
Bug: table columns collapse when cells are merged


patrick-atticus:
This a bug in the playground, but is relevant to anyone trying to implement table resizing.

https://github.com/user-attachments/assets/7d0b24af-158a-4d71-8cee-27ad845c3f74

Merging cells collapses the column width, which is not restored when unmerging (unlike google docs).

Lexical version: v0.17.1

## Steps To Reproduce

1. create a single row table
2. resize the cells to have some custom widths
3. merge all the cells together
4. unmerge

## The current behavior

Table collapses to the size of a single cell. Unmerging expands to the minimum width

## The expected behavior

Table retains it's width but has a single cell. Unmerging should restore previous column widths

## Impact of fix
- Low severity
- Any user implementing tables would benefit from improvements
- Fix would contribute to the overall quality of tables in lexical

etrepum:
I don't think this is a bug, more like a feature request. Cells don't have a fixed width to them (until set by resize), it's all based on whatever your CSS is configured to do. When cells are merged, only the display metadata (style, background color, width) from the top-left cell is used and all of the other cells' metadata is discarded. Unmerge couldn't restore any of the metadata for any of the other cells because it's not anywhere in the document anymore.

I think at best what you could do, in your app, would be to check the widths before merge to see if anything custom is set. If so, calculate the display width of the merged area before merge and then set it on the result after merge. Unmerging the best you could do is distribute any custom width equally across the top row (and the content would all still be in the top-left cell).

etrepum:
The table code is different enough now that I'm going to close this. You can open a new feature request if you think that the current behavior (v0.20.2) should behave differently regarding column resizing

PR #6625, opened 2024-09-12
[lexical-table][lexical-playground] Feature: Add column widths to TableNode

### LINKED ISSUE #6623: Bug: table resizes wrong column when cell is merged
This a bug in the playground, but is relevant to anyone trying to implement table resizing.

When you resize a merged table cell, the column the cell starts in is resized, rather than the column where you're dragging.

Demo:

https://github.com/user-attachments/assets/1959cc03-f90e-4dbd-a0c9-33ca7c2311b9

Lexical version: v0.17.1

## Steps To Reproduce

1. create a column with >3 columns and > 2 rows
2. merge any cell in the first column with the cell to it's right
3. resize the right hand side of the merged cell

## The current behavior

The size of the first column is updated

## The expected behavior

The second column is expected to change size.

## Impact of fix
- Low severity
- Any user implementing tables would benefit from improvements
- Fix would contribute to the overall quality of tables in lexical

patrick-atticus:
## Description

This PR adds an optional `colWidths` property to TableNode, which renders a `<colgroup>` element within tables. 

Defining the column widths has 2 main advantages:
1. We can simplify table resizing logic, particularly around merged cell cases.
2. It enables better control of tables with the style `table-layout: fixed`

If using `table-layout: fixed`, you must set a table width. Lexical playground uses `max-content` which masks some issues. If you want to truncate text within the table, the best way is to change the table width to `fit-content` (and add css to truncate). 

However, without a `<col>` element, a table with layout `fixed` will use the first row to determine column widths for the rest of the table. If there are merged cells in the first row, this makes it impossible to correctly size (and resize) columns in the remaining rows.

Adding `colWidths` to TableNode is the cleanest way to enable the ability to truncate text in tables and simplify resizing.

### Changes
- Add optional `colWidth` to TableNode (an array of pixel width numbers)
- Update `createDOM` to render `<colgroup>` and `<col>` elements with widths, if `colWidths` is set.
- Update playground `TableCellResizerPlugin` to get and set `colWidths` for better resizing

Closes #6623 #6624 (resize bugs)

Now it is also possible to truncate text in cells. This was not possible before and is the main motivation for this PR

## Test plan

Since this changes the output of TableNode, it breaks many tests that expect specific HTML. I will update tests if maintainers are open to this approach.

### Demo
🔉 narrated
https://github.com/user-attachments/assets/aa7902f2-df6f-4975-b83c-2a28e7f086c5

### DIFF
--- a/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
@@ -43,7 +43,7 @@ type MousePosition = {
 type MouseDraggingDirection = 'right' | 'bottom';
 
 const MIN_ROW_HEIGHT = 33;
-const MIN_COLUMN_WIDTH = 75;
+const MIN_COLUMN_WIDTH = 92;
 
 function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
   const targetRef = useRef<HTMLElement | null>(null);
--- a/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
@@ -233,7 +233,7 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
     for (let row = 0; row < tableMap.length; row++) {
       for (let column = 0; column < tableMap[row].length; column++) {
         if (tableMap[row][column].cell === tableCellNode) {
-          return column + tableCellNode.getColSpan() - 1;
+          return column;
         }
       }
     }
--- a/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
@@ -270,9 +270,10 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
           if (width === undefined) {
             return;
           }
+          const newColWidths = [...colWidths];
           const newWidth = Math.max(width + widthChange, MIN_COLUMN_WIDTH);
-          colWidths[columnIndex] = newWidth;
-          tableNode.setColWidths(colWidths);
+          newColWidths[columnIndex] = newWidth;
+          tableNode.setColWidths(newColWidths);
         },
         {tag: 'skip-scroll-into-view'},
       );
--- a/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
@@ -5,12 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-import type {
-  TableCellNode,
-  TableDOMCell,
-  TableMapType,
-  TableMapValueType,
-} from '@lexical/table';
+import type {TableCellNode, TableDOMCell, TableMapType} from '@lexical/table';
 import type {LexicalEditor} from 'lexical';
 
 import './index.css';
@@ -24,6 +19,7 @@ import {
   $isTableCellNode,
   $isTableRowNode,
   getDOMCellFromTarget,
+  TableNode,
 } from '@lexical/table';
 import {calculateZoomLevel} from '@lexical/utils';
 import {$getNearestNodeFromDOMNode} from 'lexical';
@@ -47,7 +43,7 @@ type MousePosition = {
 type MouseDraggingDirection = 'right' | 'bottom';
 
 const MIN_ROW_HEIGHT = 33;
-const MIN_COLUMN_WIDTH = 50;
+const MIN_COLUMN_WIDTH = 75;
 
 function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
   const targetRef = useRef<HTMLElement | null>(null);
@@ -74,6 +70,20 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
   const isMouseDownOnEvent = (event: MouseEvent) => {
     return (event.buttons & 1) === 1;
   };
+
+  useEffect(() => {
+    return editor.registerNodeTransform(TableNode, (tableNode) => {
+      if (tableNode.getColWidths()) {
+        return tableNode;
+      }
+
+      const numColumns = tableNode.getColumnCount();
+      const columnWidth = MIN_COLUMN_WIDTH;
+
+      tableNode.setColWidths(Array(numColumns).fill(columnWidth));
+      return tableNode;
+    });
+  }, [editor]);
 
   useEffect(() => {
     const onMouseMove = (event: MouseEvent) => {
@@ -208,27 +218,6 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
     [activeCell, editor],
   );
 
-  const getCellNodeWidth = (
-    cell: TableCellNode,
-    activeEditor: LexicalEditor,
-  ): number | undefined => {
-    const width = cell.getWidth();
-    if (width !== undefined) {
-      return width;
-    }
-
-    const domCellNode = activeEditor.getElementByKey(cell.getKey());
-    if (domCellNode == null) {
-      return undefined;
-    }
-    const computedStyle = getComputedStyle(domCellNode);
-    return (
-      domCellNode.clientWidth -
-      parseFloat(computedStyle.paddingLeft) -
-      parseFloat(computedStyle.paddingRight)
-    );
-  };
-
   const getCellNodeHeight = (
     cell: TableCellNode,
     activeEditor: LexicalEditor,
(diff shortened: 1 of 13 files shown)


packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
```diff
@@ -242,7 +231,7 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
     for (let row = 0; row < tableMap.length; row++) {
       for (let column = 0; column < tableMap[row].length; column++) {
         if (tableMap[row][column].cell === tableCellNode) {
-          return column;
+          return column + tableCellNode.getColSpan() - 1;
```
etrepum, line 234: I think this fix would be nice to have outside of this PR

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -59,13 +82,24 @@ function setRowStriping(
...
 
+  getColWidths() {
+    const self = this.getLatest();
+    return self.__colWidths;
+  }
+
+  setColWidths(colWidths: number[]) {
+    const self = this.getWritable();
+    self.__colWidths = colWidths;
+  }
```
etrepum, line 99: By convention setters are chainable in Lexical (this isn't totally consistent yet)

```suggestion
  setColWidths(colWidths: number[]): this {
    const self = this.getWritable();
    self.__colWidths = colWidths;
    return self;
  }
```

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -83,19 +117,21 @@ export class TableNode extends ElementNode {
...
     tableNode.__rowStriping = serializedNode.rowStriping || false;
     return tableNode;
   }
 
-  constructor(key?: NodeKey) {
+  constructor(colWidths?: number[], key?: NodeKey) {
     super(key);
     this.__rowStriping = false;
+    this.__colWidths = colWidths;
   }
```
etrepum, line 129: For compatibility it would be better to leave this out of the constructor and put it in `afterCloneFrom`.

```suggestion
  constructor(key?: NodeKey) {
    super(key);
    this.__rowStriping = false;
  }
```

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -59,13 +82,24 @@ function setRowStriping(
...
+  }
+
+  setColWidths(colWidths: number[]) {
+    const self = this.getWritable();
+    self.__colWidths = colWidths;
+  }
+
   static clone(node: TableNode): TableNode {
-    return new TableNode(node.__key);
+    return new TableNode(node.__colWidths, node.__key);
```
etrepum, line 102:
```suggestion
    return new TableNode(node.__key);
```

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -83,19 +117,21 @@ export class TableNode extends ElementNode {
   }
 
   static importJSON(serializedNode: SerializedTableNode): TableNode {
-    const tableNode = $createTableNode();
+    const tableNode = $createTableNode(serializedNode.colWidths);
```
etrepum, line 120:
```suggestion
    const tableNode = $createTableNode();
    tableNode.__colWidths = serializedNode.colWidths;
```

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -306,8 +358,8 @@ export function $convertTableElement(
   return {node: tableNode};
 }
 
-export function $createTableNode(): TableNode {
-  return $applyNodeReplacement(new TableNode());
+export function $createTableNode(colWidths?: number[]): TableNode {
+  return $applyNodeReplacement(new TableNode(colWidths));
```
etrepum, line 362:
```suggestion
export function $createTableNode(): TableNode {
  return $applyNodeReplacement(new TableNode());
```

packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -471,6 +471,13 @@ export function $insertTableColumn__EXPERIMENTAL(insertAfter = true): void {
   if (firstInsertedCell !== null) {
     $moveSelectionToCell(firstInsertedCell);
   }
+  const colWidths = grid.getColWidths();
+  if (colWidths) {
+    const columnIndex = insertAfterColumn < 0 ? 0 : insertAfterColumn;
+    const newWidth = colWidths[columnIndex];
+    colWidths.splice(columnIndex, 0, newWidth);
```
etrepum, line 478: This is going to break history because you're mutating the array in-place, a new array needs to be created for this

packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -643,6 +650,11 @@ export function $deleteTableColumn__EXPERIMENTAL(): void {
     const {cell} = previousRow;
     $moveSelectionToCell(cell);
   }
+  const colWidths = grid.getColWidths();
+  if (colWidths) {
+    colWidths.splice(startColumn, selectedColumnCount);
```
etrepum, line 655: Another in-place mutation of the node state

etrepum (commented):
Overall this seems like a great approach. There are going to be some issues with sharing of `__colWidths` because node properties can not be shared across versions (unless they are effectively immutable). Removing the rest of the width support from TableCellNode would also make sense, and of course fixing the tests is going to be the big thing.

packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
patrick-atticus: Good call, this is really a separate issue https://github.com/facebook/lexical/pull/6630

patrick-atticus:
@etrepum thanks for the review!

> There are going to be some issues with sharing of __colWidths because node properties can not be shared across versions (unless they are effectively immutable).

Could you please expand on this? I'm not sure what this means

I'll take a look at the tests

patrick-atticus:
Thanks, I assume there's no other nodes that do this at the moment? Would it be enough to type the property as `readonly`, and `Object.freeze` the value in the setter? (see recent commit)

This worked well to flag an error in another place I had mutated it.

etrepum:
Using the readonly type is probably sufficient, we don't do a lot of runtime enforcement of these kinds of constraints that the type checker can enforce especially in production. Copying it everywhere even when not necessary is something to be avoided (e.g. copying in the setter, getter and clone would be most robust but do unnecessary work pretty often).

It's not done a lot in this codebase, I can't think of many properties that are a mutable type to begin with. It's normally primitive value types that don't have any sort of sharing issues.

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -85,6 +121,7 @@ export class TableNode extends ElementNode {
   static importJSON(serializedNode: SerializedTableNode): TableNode {
     const tableNode = $createTableNode();
     tableNode.__rowStriping = serializedNode.rowStriping || false;
+    tableNode.__colWidths = Object.freeze(serializedNode.colWidths);
```
etrepum, line 124:
```suggestion
    tableNode.__colWidths = serializedNode.colWidths;
```

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -59,17 +82,30 @@ function setRowStriping(
 export class TableNode extends ElementNode {
   /** @internal */
   __rowStriping: boolean;
+  __colWidths?: readonly number[];
 
   static getType(): string {
     return 'table';
   }
 
+  getColWidths() {
```
etrepum, line 91:
```suggestion
  getColWidths(): readonly number[] {
```

etrepum:
It looks like the tests are passing, when you think it's ready for review (this PR is still in draft) and then we'll review it again and merge when ready

patrick-atticus:
There's a lot of noise from fixing the tests, but I've tried to keep the commits clean.

I added one unit test that covers setting widths and inserting a column. The e2e tests have some resizing merged cells which verifies the colwidth output. Let me know if there's any other cases you'd like covered.

As far as I can tell the remaining failures are due to flakes.

etrepum:
I think the remaining failures are not just flaky, I gave the failed tests a couple additional runs. Some of it is HTML differences because of the added `colgroup`

```
[1]   2 failed
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Tables.spec.mjs:2277:3 › Tables › Select multiple merged cells (selection expands to a rectangle) 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Tables.spec.mjs:3378:3 › Tables › Can align text using Table selection 
```

patrick-atticus:
Ah I see now, thank you. Fixed.

As a side note, I wasn't able to run the collab e2e tests locally. With the server running `npm run start` and the collab tests  eg `npm run test-e2e-collab-chromium` the playground connection button state always starts as disconnected:

```
      -   unexpected value "Connect Collaborative Editing"


       at packages/lexical-playground/__tests__/utils/index.mjs:119

      117 |         await expect(
      118 |           frameLocator.locator('.action-button.connect'),
    > 119 |         ).toHaveAttribute('title', /Disconnect/);
```

etrepum:
Ah that's interesting, it looks like there's a dev mode collab bug probably due to react strict mode. I think because the effects run twice the provider is disconnecting the websocket, I wonder when that started. The ci versions appear to work (although those also trigger a build which can be a bit slower to start).

etrepum (approved):
Thank you for all your work on this!

patrick-atticus:
Thanks! Keen to make use of this

Issue #6583, opened 2024-09-01
Bug: Table formatting and styling not persisting for empty cells


ivailop7:
Lexical version: 0.17.1

Since the table cells are empty paragraphs without text nodes, applying formatting and styling to the cells doesn't persist when text is added later. Those changes should persist irrespective if cells are empty or not. A lot of people do table changes in advance before they start entering text, will also bring behaviour inline with other editors.

https://github.com/user-attachments/assets/ad020a2c-eb23-41f7-a57e-ef4b8294610a

etrepum:
I think this is a design flaw, not really a bug. The workaround that is used elsewhere is to store the format on the RangeSelection, which works but that of course is ephemeral, so if you move the collapsed cursor to some text that has some existing format then the new format goes away because it was on the selection and the selection changes its format to match the existing text. This is the same way google docs works (at least with paragraphs). I think you would need to come up with some other strategy for tables (google docs persists styles on table cells or if you're at the beginning of an empty paragraph).

ivailop7:
> It does look like there are textStyle and textFormat accessors on ParagraphNode that are currently only used by the reconciler to preserve the subtree's text style & format but that could probably be used directly by the Selection classes, at least when the ParagraphNode is empty.

Thanks, that's how I started solving it, wanted to make sure there is no better solution.

etrepum:
Not that I can see! If you were trying to preserve it on the table selection it would probably be for naught because when you focus for text entry it will have been thrown away to make a RangeSelection. I have to imagine that's what this accessor is for, really, to preserve the intended format of an empty node. Whatever you do here should probably also be applied to other kinds of selection because RangeSelection also has this problem (e.g. create empty paragraph, change style on paragraph without entering text, navigate somewhere else and then back, and it's gone; or select a few empty paragraphs and then any attempt to change the format with the toolbar doesn't work even though the bold/italic buttons change their state).

etrepum:
I believe that this was fixed in #6626, I can't repro

ivailop7:
Katsia fixed the formatting, but not styling. This one is still outstanding. Try selecting an empty table and change font size, and type in afterwards, styling doesn't persist.

ivailop7:
The formatting is now persisting, styling (colors, font-size) is still not getting applied. Renamed the issue.

PR #6626, opened 2024-09-12
[lexical-table]  Bug: Table formatting and styling not persisting for empty cells

KatsiarynaDzibrova:
## Description

*Apply format change to empty paragraphs. Before it was only applied to text nodes*

Closes #[6583](https://github.com/facebook/lexical/issues/6583)

### Before
Formatting is not applied to empty paragraphs
table-bug-before

### After
Formatting is applied to empty paragraphs
table-bug-after

### DIFF
--- a/packages/lexical-table/src/LexicalTableObserver.ts
+++ b/packages/lexical-table/src/LexicalTableObserver.ts
@@ -357,12 +357,7 @@ export class TableObserver {
       const anchor = formatSelection.anchor;
       const focus = formatSelection.focus;
 
-      const cellNodes: Array<TableCellNode> = [];
-      for (const node of selection.getNodes()) {
-        if ($isTableCellNode(node)) {
-          cellNodes.push(node);
-        }
-      }
+      const cellNodes = selection.getNodes().filter($isTableCellNode);
       const paragraph = cellNodes[0].getFirstChild();
       const alignFormatWith = $isParagraphNode(paragraph)
         ? paragraph.getFormatFlags(type, null)
--- a/packages/lexical-table/src/LexicalTableObserver.ts
+++ b/packages/lexical-table/src/LexicalTableObserver.ts
@@ -21,12 +21,13 @@ import {
   $getRoot,
   $getSelection,
   $isElementNode,
+  $isParagraphNode,
   $setSelection,
   SELECTION_CHANGE_COMMAND,
 } from 'lexical';
 import invariant from 'shared/invariant';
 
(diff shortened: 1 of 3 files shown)


ivailop7 (approved):
LGTM, @etrepum for a second pair of eyes

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1190,6 +1191,13 @@ export class RangeSelection implements BaseSelection {
       }
     }
 
+    selectedNodes.forEach((node) => {
+      if ($isParagraphNode(node)) {
+        const newFormat = node.getFormatFlags(formatType, null);
+        node.setTextFormat(newFormat);
+      }
+    });
```
etrepum, line 1199: We should implement alignment with the first cell of the selection. When the text/paragraphs have mixed formatting toggling bold should change the entire selection to have the same setting for bold.

https://github.com/user-attachments/assets/16687683-1fa8-4df9-957b-d4afee2c0723

packages/lexical-table/src/LexicalTableObserver.ts
```diff
@@ -357,7 +357,7 @@ export class TableObserver {
       const focus = formatSelection.focus;
 
       selection.getNodes().forEach((cellNode) => {
```
etrepum, line 359: I think this is where we would have to capture the format alignment

packages/lexical/src/LexicalSelection.ts
KatsiarynaDzibrova: Addressed in 25e38b6 and c995bf3

packages/lexical-table/src/LexicalTableObserver.ts
```diff
@@ -356,12 +357,21 @@ export class TableObserver {
...
-        if ($isTableCellNode(cellNode) && cellNode.getTextContentSize() !== 0) {
-          anchor.set(cellNode.getKey(), 0, 'element');
-          focus.set(cellNode.getKey(), cellNode.getChildrenSize(), 'element');
-          formatSelection.formatText(type);
+      const cellNodes: Array<TableCellNode> = [];
+      for (const node of selection.getNodes()) {
+        if ($isTableCellNode(node)) {
+          cellNodes.push(node);
         }
+      }
```
etrepum, line 365:
```suggestion
      const cellNodes = selection.getNodes().filter($isTableCellNode);
```

etrepum (approved):
Looks great, thank you!

etrepum (approved):

PR #6629, opened 2024-09-12
[lexical-markdown] Fix: normalize markdown in $convertFromMarkdownString to comply with CommonMark spec (2nd try)

GermanJablo:
## Description

See https://github.com/facebook/lexical/pull/6627

The only commits added since the first PR are:

- [fix single line code](https://github.com/facebook/lexical/pull/6629/commits/60fd030679206bf146eadae8af5b701c842eef24) - fixed an additional bug that I noticed on the fly, where code blocks that started and ended on the same line left the flag open, causing subsequent lines to not be normalized.
- [fix table](https://github.com/facebook/lexical/pull/6629/commits/57aa31f8c218f3dc1d8c9efd9015d393ffc05516) - fixed the problem with tables that @potatowagon mentioned

For both cases I added corresponding tests.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -160,7 +160,7 @@ const HEADING_REGEX = /^(#{1,6})\s/;
 const QUOTE_REGEX = /^>\s/;
 const CODE_START_REGEX = /^[ \t]*```(\w+)?/;
 const CODE_END_REGEX = /[ \t]*```$/;
-const CODE_SINGLE_LINE_REGEX = /^[ \t]*```(?:[^`]+|``?[^`])+```(?:[^`]|$)/;
+const CODE_SINGLE_LINE_REGEX = /^[ \t]*```[^`]+(?:(?:`{1,2}|`{4,})[^`]+)*```(?:[^`]|$)/;
 const TABLE_ROW_REG_EXP = /^(?:\|)(.+)(?:\|)\s?$/;
 const TABLE_ROW_DIVIDER_REG_EXP = /^(\| ?:?-*:? ?)+\|\s?$/;
 
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -127,7 +127,7 @@ export type MultilineElementTransformer = {
      */
     isImport: boolean,
   ) => boolean | void;
-  type: 'multilineElement';
+  type: 'multiline-element';
 };
 
 export type TextFormatTransformer = Readonly<{
@@ -399,7 +399,7 @@ export const CODE: MultilineElementTransformer = {
       })(rootNode, children, startMatch, isImport);
     }
   },
-  type: 'multilineElement',
+  type: 'multiline-element',
 };
 
 export const UNORDERED_LIST: ElementTransformer = {
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -160,7 +160,8 @@ const HEADING_REGEX = /^(#{1,6})\s/;
 const QUOTE_REGEX = /^>\s/;
 const CODE_START_REGEX = /^[ \t]*```(\w+)?/;
 const CODE_END_REGEX = /[ \t]*```$/;
-const CODE_SINGLE_LINE_REGEX = /^[ \t]*```[^`]+(?:(?:`{1,2}|`{4,})[^`]+)*```(?:[^`]|$)/;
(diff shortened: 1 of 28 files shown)


packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -153,6 +153,17 @@ export type TextMatchTransformer = Readonly<{
...
 }>;
 
+const ORDERED_LIST_REGEX = /^(\s*)(\d{1,})\.\s/;
+const UNORDERED_LIST_REGEX = /^(\s*)[-*+]\s/;
+const CHECK_LIST_REGEX = /^(\s*)(?:-\s)?\s?(\[(\s|x)?\])\s/i;
+const HEADING_REGEX = /^(#{1,6})\s/;
+const QUOTE_REGEX = /^>\s/;
+const CODE_START_REGEX = /^[ \t]*```(\w+)?/;
+const CODE_END_REGEX = /[ \t]*```$/;
+const CODE_SINGLE_LINE_REGEX = /^[ \t]*```(\w+)?[^`]*```/;
```
etrepum, line 163: I don't think this regex is correct, it doesn't handle embedded backticks and the `\w+` is not relevant to inline code. Maybe something like this would work?
```suggestion
const CODE_SINGLE_LINE_REGEX = /^[ \t]*```(?:[^`]+|``?[^`])+```(?:[^`]|$)/;
```

GermanJablo:
I can't reproduce the failing test. I've tried in plain text, firefox. Is the problem only on Linux?

GermanJablo:
Could you try to re-run the failed test?

I'm not sure if it's flaky. On the one hand, the test that fails is in a code block that parses to markdown. On the other hand, I can't reproduce it on my Mac or PC :/

etrepum:
Re-running failed jobs now

etrepum:
I can repro. Haven't looked at the code but there's a very bad performance issue in all browsers in that test case (hangs for >10 seconds on my laptop in chrome after the second click to the markdown button), but probably just bad enough that the firefox test fails.

```
npm run test-e2e-plain-firefox packages/lexical-playground/__tests__/e2e/Hashtags.spec.mjs:472:3
```


https://github.com/user-attachments/assets/4747cc67-c9f0-4128-8b01-dee8e6f11fcd

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -153,6 +153,17 @@ export type TextMatchTransformer = Readonly<{
...
 }>;
 
+const ORDERED_LIST_REGEX = /^(\s*)(\d{1,})\.\s/;
+const UNORDERED_LIST_REGEX = /^(\s*)[-*+]\s/;
+const CHECK_LIST_REGEX = /^(\s*)(?:-\s)?\s?(\[(\s|x)?\])\s/i;
+const HEADING_REGEX = /^(#{1,6})\s/;
+const QUOTE_REGEX = /^>\s/;
+const CODE_START_REGEX = /^[ \t]*```(\w+)?/;
+const CODE_END_REGEX = /[ \t]*```$/;
+const CODE_SINGLE_LINE_REGEX = /^[ \t]*```(?:[^`]+|``?[^`])+```(?:[^`]|$)/;
```
etrepum, line 163: The performance problem was that the regex did too much backtracking, this one does less and is closer to the commonmark spec

```suggestion
const CODE_SINGLE_LINE_REGEX = /^[ \t]*```[^`]+(?:(?:`{1,2}|`{4,})[^`]+)*```(?:[^`]|$)/;
```

etrepum (approved):

potatowagon:
https://github.com/user-attachments/assets/4b687c85-2cb9-4888-aca3-66e6a1e83afe

lgtm

packages/lexical-markdown/src/index.ts
```diff
@@ -82,11 +83,12 @@ function $convertFromMarkdownString(
   node?: ElementNode,
   shouldPreserveNewLines = false,
 ): void {
+  const sanitizedMarkdown = normalizeMarkdown(markdown);
```
potatowagon, line 86: can we add a condition here, to only normalizeMarkdown if shouldPreserveNewLines = false?

potatowagon, line 86:
```suggestion
  const sanitizedMarkdown = shouldPreserveNewLines ? markdown : normalizeMarkdown(markdown);
```

etrepum: I think we should add some test coverage for this to prevent such regressions in the future

potatowagon: added in d58c159c4a933f068bfa0be8792f878281e0d64a

potatowagon (approved):

PR #6630, opened 2024-09-12
[lexical-playground] Bug Fix: Fix table row/column index when resizing merged cells

### LINKED ISSUE #6623: Bug: table resizes wrong column when cell is merged

patrick-atticus:
## Description
When table cells are merged (either vertically or horizontally) and you try to resize them, it resizes the row or column of the top/left cell.

Update the index to account for colspan and rowspan so the correct column/row is resized.

Closes #6623

### Before

https://github.com/user-attachments/assets/75333ad5-c0fa-4800-b3c6-25d932d8b8df

### After

https://github.com/user-attachments/assets/473d6d33-f467-4e16-afbb-8bc947367c7d

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -1690,7 +1690,8 @@ test.describe.parallel('Tables', () => {
               <th
                 class="PlaygroundEditorTheme__tableCell PlaygroundEditorTheme__tableCellHeader"
                 colspan="2"
-                rowspan="2">
+                rowspan="2"
+                style="width: 217px">
                 <p class="PlaygroundEditorTheme__paragraph"><br /></p>
               </th>
               <th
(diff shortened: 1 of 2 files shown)


etrepum (approved):
Seems to work much better this way!

PR #6631, opened 2024-09-13
[lexical] Bug Fix: spliting AutoLinkNode

### LINKED ISSUE #6628: Bug: Line break inside link node appends text to the end of the current paragraph
When inserting a line break by pressing `Enter` while the cursor is inside a link the part of the link that is at the end will go to the end of the current paragraph and not to the start of the next line. Follow the steps to reproduce for an example or see the attached video.

Lexical version: 0.17.1

## Steps To Reproduce

1. Have the text `Some text. https://www.google.com Some other text.` in the editor https://playground.lexical.dev/
2. Set the cursor inside the link, for example between `c` and `o`: `https://www.google.c|om`
3. Press `ENTER`

https://github.com/user-attachments/assets/45573cee-8dbf-46f3-9442-9c631b9871f6

Link to code example:

## The current behavior
After following the steps to reproduce we have 3 paragraphs with the following content:
```
Some text. https://www.google.c
 Some other text.
om
```

## The expected behavior
It was expected to have only 2 paragraphs with the content:
```
Some text. https://www.google.c
om Some other text.
```

## Impact of fix
- Low severity
- Any user that uses link nodes may be affected

strdr4605:
## Description

Closes #6628

### Before

https://github.com/facebook/lexical/issues/6628

### After

https://github.com/user-attachments/assets/de32f92b-c246-41db-b82c-e12657414588

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
@@ -760,4 +760,63 @@ test.describe('Auto Links', () => {
       {ignoreClasses: true},
     );
   });
+
+  test('Can handle new lines in the middle of a URL for rich text', async ({
+    page,
+    isPlainText,
+  }) => {
+    test.skip(isPlainText);
+    await focusEditor(page);
+
+    await page.keyboard.type('Hello http://www.example.com test');
+    for (let i = 0; i < 8; i++) {
+      await page.keyboard.press('ArrowLeft');
+    }
+    await page.keyboard.press('Enter');
+
+    await assertHTML(
+      page,
+      html`
+        <p dir="ltr">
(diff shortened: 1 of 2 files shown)


potatowagon:
from the e2e test results, there seems to be a regression for copy paste

strdr4605:
> From the e2e test results, there seems to be a regression for copy paste

Hey @potatowagon, thanks for checking this PR. Before moving on try to fix the regression for copy-paste. I would like to know if the fix is good enough. I did the fix mostly to start a discussion around this issue and see that is the best solution.

strdr4605:
While trying to debug the failing tests I see a strange behavior of the editor in the test and reproducing it manually

image

Here my trying to do the same as the test

https://github.com/user-attachments/assets/d64296ad-9bbb-4ccd-9a96-5af96d08be87

The result is different. What I am missing?

PR #6634, opened 2024-09-14
[lexical-playground][ExcalidrawNode] Bug Fix: Preserve Excalidraw image dimensions after resizing

neysanfoo:
## Description
When resizing an Excalidraw image within the Lexical editor, the new dimensions are not preserved after certain editor actions. For example, try moving the cursor just to the left of the image and pressing Enter. This will cause the image to revert to its original size and a mismatch between the resize box and the image.

#### Updating the ExcalidrawNode Class:
1. Updated updateDOM method to ensure than when node is updated the dom reflects the new width and height
2. Update $createExcalidrawNode function to take in data, width, height

#### Modifying the ExcalidrawComponent:
1. Retrieved width and height from the node using the new getter methods.
2. Pass width and height to ExcalidrawImage

#### Updating the ExcalidrawImage Component:
1. Accepted width and height as props.
2. Apply the dimensions to the container's style

### Steps to Reproduce:
1. Insert an Excalidraw image into the editor.
5. Resize the image to make it larger.
6. Move the cursor to the left of the image.
7. Press Enter.

### Before
Observed Behavior: The image shrinks back to its original size, losing the resizing adjustments made by the user.

https://github.com/user-attachments/assets/c1011f32-e071-4083-9e5f-02fd3d4718b7

### After
Observed Behavior: The image retains its resized dimensions, preserving the user's adjustments even after the editor updates.

https://github.com/user-attachments/assets/6712543c-0b0b-4015-a0c5-4010162c5427

### DIFF
--- a/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
@@ -15,8 +15,6 @@ import {useLexicalNodeSelection} from '@lexical/react/useLexicalNodeSelection';
 import {mergeRegister} from '@lexical/utils';
 import {
   $getNodeByKey,
-  $getSelection,
-  $isNodeSelection,
   CLICK_COMMAND,
   COMMAND_PRIORITY_LOW,
   KEY_BACKSPACE_COMMAND,
@@ -48,13 +46,13 @@ export default function ExcalidrawComponent({
     useLexicalNodeSelection(nodeKey);
   const [isResizing, setIsResizing] = useState<boolean>(false);
 
-  const onDelete = useCallback(
+  const $onDelete = useCallback(
     (event: KeyboardEvent) => {
-      if (isSelected && $isNodeSelection($getSelection())) {
+      if (isSelected) {
         event.preventDefault();
         editor.update(() => {
           const node = $getNodeByKey(nodeKey);
-          if ($isExcalidrawNode(node)) {
+          if (node) {
             node.remove();
           }
         });
@@ -102,22 +100,22 @@ export default function ExcalidrawComponent({
       ),
       editor.registerCommand(
         KEY_DELETE_COMMAND,
-        onDelete,
+        $onDelete,
         COMMAND_PRIORITY_LOW,
       ),
       editor.registerCommand(
         KEY_BACKSPACE_COMMAND,
-        onDelete,
+        $onDelete,
         COMMAND_PRIORITY_LOW,
       ),
     );
-  }, [clearSelection, editor, isSelected, isResizing, onDelete, setSelected]);
+  }, [clearSelection, editor, isSelected, isResizing, $onDelete, setSelected]);
 
   const deleteNode = useCallback(() => {
     setModalOpen(false);
     return editor.update(() => {
       const node = $getNodeByKey(nodeKey);
-      if ($isExcalidrawNode(node)) {
+      if (node) {
         node.remove();
       }
     });
@@ -182,15 +180,15 @@ export default function ExcalidrawComponent({
     appState = {},
   } = useMemo(() => JSON.parse(data), [data]);
 
-  const [width, height] = useMemo(() => {
+  const [initialWidth, initialHeight] = useMemo(() => {
     let nodeWidth: 'inherit' | number = 'inherit';
     let nodeHeight: 'inherit' | number = 'inherit';
 
     editor.getEditorState().read(() => {
       const node = $getNodeByKey(nodeKey);
       if ($isExcalidrawNode(node)) {
-        nodeWidth = node.getWidth(); // Now exists
-        nodeHeight = node.getHeight(); // Now exists
+        nodeWidth = node.getWidth();
+        nodeHeight = node.getHeight();
       }
     });
 
@@ -223,8 +221,8 @@ export default function ExcalidrawComponent({
             elements={elements}
             files={files}
             appState={appState}
-            width={width}
-            height={height}
+            width={initialWidth}
+            height={initialHeight}
           />
           {isSelected && (
             <div
--- a/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
@@ -41,32 +41,30 @@ export default function ExcalidrawComponent({
   const [isModalOpen, setModalOpen] = useState<boolean>(
     data === '[]' && editor.isEditable(),
   );
-  const imageContainerRef = useRef<HTMLImageElement | null>(null);
+  const imageContainerRef = useRef<HTMLDivElement | null>(null);
   const buttonRef = useRef<HTMLButtonElement | null>(null);
   const captionButtonRef = useRef<HTMLButtonElement | null>(null);
   const [isSelected, setSelected, clearSelection] =
     useLexicalNodeSelection(nodeKey);
   const [isResizing, setIsResizing] = useState<boolean>(false);
 
-  const $onDelete = useCallback(
+  const onDelete = useCallback(
     (event: KeyboardEvent) => {
-      const deleteSelection = $getSelection();
-      if (isSelected && $isNodeSelection(deleteSelection)) {
+      if (isSelected && $isNodeSelection($getSelection())) {
         event.preventDefault();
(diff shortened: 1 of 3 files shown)


packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -41,32 +41,30 @@ export default function ExcalidrawComponent({
...
   const [isSelected, setSelected, clearSelection] =
     useLexicalNodeSelection(nodeKey);
   const [isResizing, setIsResizing] = useState<boolean>(false);
 
-  const $onDelete = useCallback(
+  const onDelete = useCallback(
     (event: KeyboardEvent) => {
-      const deleteSelection = $getSelection();
-      if (isSelected && $isNodeSelection(deleteSelection)) {
+      if (isSelected && $isNodeSelection($getSelection())) {
```
etrepum, line 53: I don't think this `$isNodeSelection($getSelection())` is doing anything useful anymore, now that the selection itself is not used

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -41,32 +41,30 @@ export default function ExcalidrawComponent({
...
         editor.update(() => {
-          deleteSelection.getNodes().forEach((node) => {
-            if ($isExcalidrawNode(node)) {
-              node.remove();
-            }
-          });
+          const node = $getNodeByKey(nodeKey);
+          if ($isExcalidrawNode(node)) {
+            node.remove();
+          }
```
etrepum, line 59: This type check is redundant, all nodes have `remove()` and node keys are only re-used for different versions of the same instance

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -41,32 +41,30 @@ export default function ExcalidrawComponent({
...
-  const imageContainerRef = useRef<HTMLImageElement | null>(null);
+  const imageContainerRef = useRef<HTMLDivElement | null>(null);
   const buttonRef = useRef<HTMLButtonElement | null>(null);
   const captionButtonRef = useRef<HTMLButtonElement | null>(null);
   const [isSelected, setSelected, clearSelection] =
     useLexicalNodeSelection(nodeKey);
   const [isResizing, setIsResizing] = useState<boolean>(false);
 
-  const $onDelete = useCallback(
+  const onDelete = useCallback(
```
etrepum, line 51: This is named `$onDelete` for a reason, since it calls dollar functions it must also be a dollar function. This rename should fail the lint
```suggestion
  const $onDelete = useCallback(
```

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -104,16 +102,16 @@ export default function ExcalidrawComponent({
       ),
       editor.registerCommand(
         KEY_DELETE_COMMAND,
-        $onDelete,
+        onDelete,
```
etrepum, line 105:
```suggestion
        $onDelete,
```

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -104,16 +102,16 @@ export default function ExcalidrawComponent({
...
         COMMAND_PRIORITY_LOW,
       ),
       editor.registerCommand(
         KEY_BACKSPACE_COMMAND,
-        $onDelete,
+        onDelete,
```
etrepum, line 110:
```suggestion
        $onDelete,
```

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -104,16 +102,16 @@ export default function ExcalidrawComponent({
...
         COMMAND_PRIORITY_LOW,
       ),
     );
-  }, [clearSelection, editor, isSelected, isResizing, $onDelete, setSelected]);
+  }, [clearSelection, editor, isSelected, isResizing, onDelete, setSelected]);
```
etrepum, line 114:
```suggestion
  }, [clearSelection, editor, isSelected, isResizing, $onDelete, setSelected]);
```

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -184,6 +182,21 @@ export default function ExcalidrawComponent({
...
+    editor.getEditorState().read(() => {
+      const node = $getNodeByKey(nodeKey);
+      if ($isExcalidrawNode(node)) {
+        nodeWidth = node.getWidth(); // Now exists
+        nodeHeight = node.getHeight(); // Now exists
+      }
+    });
+
+    return [nodeWidth, nodeHeight];
+  }, [editor, nodeKey]);
```
etrepum, line 198: This only runs once to get the initial height and width, because neither the `editor` or `nodeKey` ever change identity. If this is intentional behavior then maybe `initialHeight` and `initialWidth` would be better names.

packages/lexical-playground/src/nodes/ExcalidrawNode/index.tsx
```diff
@@ -29,13 +29,13 @@ const ExcalidrawComponent = React.lazy(() => import('./ExcalidrawComponent'));
...
-    width: Dimension;
-    height: Dimension;
+    width?: Dimension;
+    height?: Dimension;
   },
   SerializedLexicalNode
 >;
 
-function $convertExcalidrawElement(
+function convertExcalidrawElement(
```
etrepum, line 38: This shouldn't be renamed
```suggestion
function $convertExcalidrawElement(
```

packages/lexical-playground/src/nodes/ExcalidrawNode/index.tsx
```diff
@@ -48,10 +48,7 @@ function $convertExcalidrawElement(
     !widthStr || widthStr === 'inherit' ? 'inherit' : parseInt(widthStr, 10);
 
   if (excalidrawData) {
-    const node = $createExcalidrawNode();
-    node.__data = excalidrawData;
-    node.__height = height;
-    node.__width = width;
+    const node = new ExcalidrawNode(excalidrawData, width, height);
```
etrepum, line 51: The original was more correct, it's always preferable to use a constructor function so that `$applyNodeReplacement` can be used

```suggestion
    const node = $createExcalidrawNode(excalidrawData, width, height);
```

packages/lexical-playground/src/nodes/ExcalidrawNode/index.tsx
```diff
@@ -135,7 +139,7 @@ export class ExcalidrawNode extends DecoratorNode<JSX.Element> {
           return null;
         }
         return {
-          conversion: $convertExcalidrawElement,
+          conversion: convertExcalidrawElement,
```
etrepum, line 142:
```suggestion
          conversion: $convertExcalidrawElement,
```

packages/lexical-playground/src/nodes/ExcalidrawNode/index.tsx
```diff
@@ -169,11 +173,19 @@ export class ExcalidrawNode extends DecoratorNode<JSX.Element> {
     self.__data = data;
   }
 
+  getWidth(): Dimension {
+    return this.__width;
```
etrepum, line 177:
```suggestion
    return this.getLatest().__width;
```

packages/lexical-playground/src/nodes/ExcalidrawNode/index.tsx
```diff
@@ -169,11 +173,19 @@ export class ExcalidrawNode extends DecoratorNode<JSX.Element> {
...
+  }
+
   setWidth(width: Dimension): void {
     const self = this.getWritable();
     self.__width = width;
   }
 
+  getHeight(): Dimension {
+    return this.__height;
```
etrepum, line 186:
```suggestion
    return this.getLatest().__height;
```

etrepum (approved):
Looks good! Confirmed that resizing works as expected now at least in the situations I tried

PR #6641, opened 2024-09-18
Allow exporting a document fragment from the exportDOM function

skopz356:
Closes https://github.com/facebook/lexical/issues/5238

This PR allows a node's exportDOM function to return a [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) in addition to a HTMLElement or TextNode, which allows exporting a collection of elements with no wrapping element from that node.

See the linked issue for more details, but my basic use case is being able to remove a wrapping anchor tag from the HTML output if the href property is invalid, but keep any formatting (such as tags) within that anchor tag.

Based on: https://github.com/facebook/lexical/pull/5288

### DIFF
--- a/packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
+++ b/packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
@@ -6,9 +6,6 @@
  *
  */
 
-//@ts-ignore-next-line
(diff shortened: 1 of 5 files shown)


etrepum (approved):
Looks good and tests are passing!

PR #6642, opened 2024-09-18
[lexical-markdown] shouldMergeAdjacentLines as an option

potatowagon:
## Description

normalize markdown was introduced in https://github.com/facebook/lexical/pull/6629 . one of the significant behavior change this introduced is merging adjacent non empty lines, according to commonmark spec: https://spec.commonmark.org/0.24/#example-177

it is a welcomed change since that brings lexical's impl closer to commonmark. However to avoid breaking current callsites, im making the normalizedMarkdown an option default to true, so it is easier to rollback incase this new behavior causes a regression somewhere downstream

ultimately, if there is no known usecase to have shouldMergeAdjacentLines = false, will look to deprecate  shouldMergeAdjacentLines and have it always be true.

## Test plan

shouldMergeAdjacentLines = true

https://github.com/user-attachments/assets/5ec217e7-49a5-4119-b4c2-7466366fce47

---

shouldMergeAdjacentLines = false

https://github.com/user-attachments/assets/1c57abfa-db0c-4a8d-a29c-938764075b2d

added automated tests

### DIFF
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -65,7 +65,7 @@ describe('Markdown', () => {
     skipExport?: true;
     skipImport?: true;
     shouldPreserveNewLines?: true;
-    shouldNormalizeMarkdown?: true | false;
+    shouldMergeAdjacentLines?: true | false;
     customTransformers?: Transformer[];
   }>;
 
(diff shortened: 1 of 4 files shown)


packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -202,6 +203,17 @@ describe('Markdown', () => {
       html: '<p><i><em style="white-space: pre-wrap;">Hello </em></i><i><b><strong style="white-space: pre-wrap;">world</strong></b></i><i><em style="white-space: pre-wrap;">!</em></i></p>',
       md: '*Hello **world**!*',
     },
+    {
+      html: '<p><span style="white-space: pre-wrap;">helloworld</span></p>',
+      md: 'hello\nworld',
+      shouldNormalizeMarkdown: true,
+      skipExport: true,
```
potatowagon, line 210: skipping export because:

string -> md -> string
(string)
```
hello 
world
```
to (md)`helloworld`
to (string)`helloworld`

---

zurfyx (commented):
`shouldPreserveNewLines` is a specific flag while `shouldNormalizeMarkdown` is very high-level, I don't think they play well with each other and the code shows how we have to chain conditions to make it work.

I would consider targetting the specific parts of the MarkDown we don't want to normalize as opposed to the `normalize` function as a whole which is an internal implementation.

potatowagon:
> `shouldPreserveNewLines` is a specific flag while `shouldNormalizeMarkdown` is very high-level, I don't think they play well with each other and the code shows how we have to chain conditions to make it work.
> 
> I would consider targetting the specific parts of the MarkDown we don't want to normalize as opposed to the `normalize` function as a whole which is an internal implementation.

good point, i updated to shouldMergeAdjacentLines instead and placed its in a specific place in normalise markdown

zurfyx (approved):

PR #6644, opened 2024-09-19
[lexical-react][lexical-playground] Bug Fix: Workaround for yjs disconnect race in React StrictMode

### LINKED ISSUE #6640: Bug: Collab not connecting in dev mode
Lexical version: 0.17.1

## Steps To Reproduce

From https://github.com/facebook/lexical/pull/6625#issuecomment-2357260185

1. `npm run start`
2. `npm run test-e2e-chromium` (from another terminal)

## The current behavior

Fails because the websocket doesn't stay connected. I suspect this is a React strict mode related issue. CI doesn't test dev mode.

## The expected behavior

Websocket stays connected

## Impact of fix

I suspect that it would be pretty hard to work on collab right now with dev mode broken?

etrepum:
## Description

The PR in #6619 created a regression for local collab development by introducing a timing change in when disconnect was called. When reviewing the PR I didn't think it caused an issue because the deployed playground and CI tests worked fine, but when running the playground locally with vite dev mode it does cause an issue. 

https://github.com/facebook/lexical/pull/6619#discussion_r1754903692

Closes #6640

### Before

```
npm run start
npm run test-e2e-chromium # from another terminal
```

All tests fail.

Can also reproduce collab not working at http://localhost:3000/split/?isCollab=true

Note that this only happens when running locally with vite dev, it does not cause any problems in CI or in production.

### After

It all works as expected.

### DIFF
--- a/packages/lexical-react/src/shared/useYjsCollaboration.tsx
+++ b/packages/lexical-react/src/shared/useYjsCollaboration.tsx
@@ -155,9 +155,18 @@ export function useYjsCollaboration(
 
     return () => {
       if (isReloadingDoc.current === false) {
-        Promise.resolve(connectionPromise).then(() => {
+        if (connectionPromise) {
+          connectionPromise.then(disconnect);
+        } else {
+          // Workaround for race condition in StrictMode. It's possible there
+          // is a different race for the above case where connect returns a
+          // promise, but we don't have an example of that in-repo.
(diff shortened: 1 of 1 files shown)


potatowagon (approved):
thanks for the fix!

zurfyx:
Thank you Bob!

PR #6650, opened 2024-09-19
[lexical-table] Bug Fix: Fix crash in $deleteCellHandler

### LINKED ISSUE #6581: Bug: Crash when select all and delete with only table in the editor
Copy-pasting/ Inserting a table into the editor and deleting it by selecting all crashes the editor.

Lexical version: v0.17.1

## Steps To Reproduce
1. By copy/paste
    - Copy a table
    - Ctrl + A in the lexical editor
    - Paste
    - Ctrl + A
    - Press Del key

table-delete

2. By Inserting
    - Insert table
    - Remove everything before and after the table node
    - Ctrl + A
    - Press Del
    
table-delete-2

## The current behavior
Can see crashes

## The expected behavior
To not crash and delete the table

etrepum:
## Description

In certain situations `$deleteCellHandler` will remove a table without fixing up the selection and also return false, which turns into a crash in the editor's command listener because the selection's anchor or focus is no longer connected to the root.

TODO: tests

Closes #6581

### Before

1. Create a table
2. Remove the paragraph before the table
3. Select all from inside the table (anchor will be inside the table, focus will be outside the table)
4. Press delete
5. *Crash*

### After

No crash

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -336,22 +336,30 @@ export function applyTableHandlers(
     event: KeyboardEvent | ClipboardEvent | null,
   ): boolean => {
     const selection = $getSelection();
+    if (!($isTableSelection(selection) || $isRangeSelection(selection))) {
+      return false;
+    }
 
-    if (!$isSelectionInTable(selection, tableNode)) {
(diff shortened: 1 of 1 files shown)


potatowagon (approved):

ivailop7:
Thank you for your help on this one!

PR #6651, opened 2024-09-19
[lexical-markdown]: Refactor: allows omitting certain properties from TextMatchTransformers, adds jsdocs

AlessioGr:
In my project, I have a `TextMatchTransformer` that I use solely for  node => markdown string exports. For markdown string => node imports I'm using Multiline markdown transformers.

In order to accommodate this without having to pass fake stub properties to the `TextMatchTransformer`, this PR makes all properties optional.

Also added some much-needed JSDocs for this interface


etrepum (approved):
This seems reasonable to me, I did notice this last time I used the markdown implementation

etrepum:
The integration test failure is fixed by #6652

PR #6652, opened 2024-09-19
[lexical-code] Bug Fix: Annotate @lexical/code as having side-effects for Prism

etrepum:
## Description

The change in #6562 appears to have triggered an issue with astro's rollup configuration which will treeshake Prism out of @lexical/code, since it is only an import side-effect. Marking the package.json as having sideEffects is sufficient to fix it.

Note that #6650 currently has this commit included so the integration tests pass, but I made it a separate PR in case we want to review/merge separately.

### Before

Integration tests in #6650 were failing

### After

Integration tests pass

### DIFF
--- a/packages/lexical-code/package.json
+++ b/packages/lexical-code/package.json
@@ -25,7 +25,7 @@
     "@types/prismjs": "^1.26.0"
   },
   "module": "LexicalCode.mjs",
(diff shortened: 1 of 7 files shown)


potatowagon (approved):
thanks for fixing the failing integration tests!

PR #6656, opened 2024-09-21
[lexical-link] Bug Fix: Removing a link from descendants

levensta:
## Description

At the moment it is possible to delete LinkNode only when children are TextNode. But if there is another node inside the link above the text, deletion will not work.

With this PR I fix this behavior and show how it works using the MarkNode example (via adding a comment) in lexical-playground

### Before

https://github.com/user-attachments/assets/669a5ed9-926d-477c-9f7b-14c2a72d6577

### After

https://github.com/user-attachments/assets/6a56895c-2b08-4862-8787-ba0cef492d2d

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -18,7 +18,11 @@ import type {
   SerializedElementNode,
 } from 'lexical';
 
-import {addClassNamesToElement, isHTMLAnchorElement} from '@lexical/utils';
+import {
+  $findMatchingParent,
+  addClassNamesToElement,
+  isHTMLAnchorElement,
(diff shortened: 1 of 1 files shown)


etrepum (approved):
The logic looks like it's probably correct, but it would be nice to have a bit of test coverage!

There's probably some efficiency that could be gained by checking only the anchor and focus parents, otherwise the LinkNode itself would be present in the selection's nodes.

ivailop7:
+1 on a test covering this use case

potatowagon (approved):
looks good to merge! have noted the test case addition todo here: https://github.com/facebook/lexical/issues/6665

PR #6658, opened 2024-09-22
[lexical-table] Bug Fix: get table-cell background selection color from a class

### LINKED ISSUE #6655: Feature: Support Custom Color for Table Selection Highlight
## Description

I am working on a Lexical Editor for dark themes. It would be great to support custom colors for table selection highlights.

https://github.com/user-attachments/assets/86ef7a54-56cb-4dff-8177-03265b6c11b0

## Impact

This will help achieve a higher-quality dark theme.

hamza221:
## Description

Get table-cell background selection styles from a theme class `tableCellSelected` instead of directly updating the styles of the table cell nodes.

In order to upgrade to this version, you will need to add some CSS like this:

```css
.PlaygroundEditorTheme__tableCellSelected {
  caret-color: transparent;
}
.PlaygroundEditorTheme__tableCellSelected::after {
  position: absolute;
  left: 0;
  right: 0;
  bottom: 0;
  top: 0;
  background-color: rgb(172, 206, 247);
  opacity: 0.6;
  content: '';
  pointer-events: none;
}
```

The following unused theme properties were removed:

* tableCellEditing
* tableCellPrimarySelected
* tableCellSortedIndicator
* tableResizeRuler

Closes #6655

## After

https://github.com/user-attachments/assets/4086b6a8-b698-47fe-b854-084e223354e1

## Test plan

Existing tests were modified to no longer expect the styles to be added.


packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -167,7 +167,14 @@
   text-align: start;
 }
 .PlaygroundEditorTheme__tableCellSelected {
-  background-color: #c9dbf0;
+  background-color: rgb(172, 206, 247);
+}
+.PlaygroundEditorTheme__tableCellSelectedOverwrite {
```
etrepum, line 172: I suspect the unused table related classes are from the old GridSelection code that was removed some time ago, but the way that the CSS is managed makes it hard to know when things aren't used.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1313,12 +1312,9 @@ function $addHighlightToDOM(editor: LexicalEditor, cell: TableDOMCell): void {
...
   const backgroundColor = node.getBackgroundColor();
   if (backgroundColor === null) {
-    element.style.setProperty('background-color', `rgb(${BROWSER_BLUE_RGB})`);
+    element.classList.add('PlaygroundEditorTheme__tableCellSelected');
   } else {
-    element.style.setProperty(
-      'background-image',
-      `linear-gradient(to right, rgba(${BROWSER_BLUE_RGB},0.85), rgba(${BROWSER_BLUE_RGB},0.85))`,
-    );
+    element.classList.add('PlaygroundEditorTheme__tableCellSelectedOverwrite');
```
etrepum, line 1317: Playground classes don't belong in reusable code, the classes used should come from the editor's theme

hamza221:
- I left the old logic in, because I'm not about another way to set a default value (new to the code base) 
- I'll update the failing tests once the logic gets approved

ivailop7:
The CSS class is probably a leftover from the experimental table, which we deleted a year/year and a half ago. I think it's fine to reuse that CSS class, I did the same with reintroducing the hover add column/row buttons and reusing the existing CSS classes. I'm not in favour of adding an extra override CSS class, as long as you pick the CSS class name from the config.theme object use the existing CSS Class, also +1 for the test.

etrepum:
I agree that an override class is not necessary. Should be up to the user to make sure their configuration of classes and styles composes correctly.

matveeva-as:
Would be great to see this fix in the next release!

ivailop7:
@hamza221 are you interested in finishing this one?

hamza221:
I'll send a commit tomorrow

hamza221:
🤞

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -2570,27 +2590,27 @@ test.describe.parallel('Tables', () => {
...
-                style="background-color: rgb(172, 206, 247); caret-color: transparent">
+                style="caret-color: transparent">
                 <p class="PlaygroundEditorTheme__paragraph"><br /></p>
               </th>
               <th
-                class="PlaygroundEditorTheme__tableCell PlaygroundEditorTheme__tableCellHeader"
-                colspan="2"
-                style="background-color: rgb(172, 206, 247); caret-color: transparent">
+                class="PlaygroundEditorTheme__tableCell PlaygroundEditorTheme__tableCellHeader PlaygroundEditorTheme__tableCellSelected"
+                rowspan="2"
```
etrepum, line 2600: This changing from colspan to rowspan seems suspicious?
```suggestion
                colspan="2"
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1089,7 +1090,7 @@ export function $updateDOMForSelection(
 
     if (selectedCellNodes.has(lexicalNode)) {
       cell.highlighted = true;
-      $addHighlightToDOM(editor, cell);
+      $addHighlightToDOM(editor, cell, editor._config.theme);
```
etrepum, line 1093: I don't think this signature should change when the editor is an argument to this function already, would be easier to reference `editor._config.theme` from inside `$addHighlightToDOM`

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1304,21 +1305,22 @@ function selectTableCellNode(tableCell: TableCellNode, fromStart: boolean) {
 }
 
 const BROWSER_BLUE_RGB = '172,206,247';
-function $addHighlightToDOM(editor: LexicalEditor, cell: TableDOMCell): void {
+function $addHighlightToDOM(
+  editor: LexicalEditor,
+  cell: TableDOMCell,
+  editorThemeClasses: EditorThemeClasses,
+): void {
```
etrepum, line 1312:
```suggestion
function $addHighlightToDOM(editor: LexicalEditor, cell: TableDOMCell): void {
  const editorThemeClasses = editor._config.theme;
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1304,21 +1305,22 @@ function selectTableCellNode(tableCell: TableCellNode, fromStart: boolean) {
...
+  if (tableCellSelected) {
+    element.classList.add(tableCellSelected);
   } else {
-    element.style.setProperty(
-      'background-image',
-      `linear-gradient(to right, rgba(${BROWSER_BLUE_RGB},0.85), rgba(${BROWSER_BLUE_RGB},0.85))`,
-    );
+    element.style.setProperty('background-color', `rgb(${BROWSER_BLUE_RGB})`);
   }
   element.style.setProperty('caret-color', 'transparent');
```
etrepum, line 1325: I don't think there's any reason to use style directly if we are applying classes correctly

etrepum (commented):
I think this is getting very close! I believe there's one mistake in the test changes, and that it can be simplified further by not using element.style at all

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
hamza221: no it was ctrl c +ctrl v 😥  I'll revert it

etrepum:
Were you planning to finish this PR?

hamza221:
> ! I believe there's one mistake in the test changes, and that it can be simplified further by not using element.style at all

I'm okay with someone else taking over, I'm Afk until 2025

etrepum:
I'll take a look at finishing this after #6759

hamza221:
Thank you @etrepum

etrepum:
@ivailop7 might want to take a look at this one

ivailop7 (approved):
LGTM

PR #6660, opened 2024-09-24
[lexical-markdown] set shouldMergeAdjacentLines default to be false 

potatowagon:
## Description
For backwards compatibility, setting default shouldMergeAdjacentLines = false to not introduce breaking change from the previous npm release. 

Github Flavoured Markdown has the same behavior as shouldMergeAdjacentLines = false:

https://github.com/user-attachments/assets/c3dabbf7-d666-44dc-a586-2d03c0a01cec

## Test plan

npm run test-unit
npm run test-e2e-chromium

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -550,7 +550,7 @@ export const LINK: TextMatchTransformer = {
 
 export function normalizeMarkdown(
   input: string,
-  shouldMergeAdjacentLines = true,
+  shouldMergeAdjacentLines = false,
 ): string {
   const lines = input.split('\n');
   let inCodeBlock = false;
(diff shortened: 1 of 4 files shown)


potatowagon:
hi @GermanJablo please set `shouldMergeAdjacentLines = true` in your implementation to get the behavior in https://github.com/facebook/lexical/pull/6629

GermanJablo:
I wonder if it might be better to use a single prop like "GFM | Commonmark".

Right now, it's just one prop, but in the future, there could be many more for people targeting different specs. I don't think users will want to figure out the right combination of props to match their target spec.

For instance, when using MDX, you typically rely on plugins that handle different converters for each spec.

potatowagon:
> I wonder if it might be better to use a single prop like "GFM | Commonmark".
> 
> Right now, it's just one prop, but in the future, there could be many more for people targeting different specs. I don't think users will want to figure out the right combination of props to match their target spec.
> 
> For instance, when using MDX, you typically rely on plugins that handle different converters for each spec.

an enum for that would be nice, but likely would have to create a new api for that and mark the old one as deprecated, for backwards compatibilty. perhaps in a future PR

etrepum (approved):
LGTM, we should bias towards backwards compatibility here

PR #6664, opened 2024-09-24
Iterable $dfs

zurfyx:
`$dfs` is handy but not very flexible. The fact that it computes the entire array every time means that it's either all or nothing.

A common use case for a DFS is data format conversion, and data conversion formats that are modular will delegate the processing of a part of the tree to the responsible modules. An iterable version of `$dfs` makes it possible to stop and resume the iteration without paying the redundant cost.

Iterable is not just limited to this one use case, I'm confident that this small addition to the utils library is not a one-off.

This is how the application of this looks like in the wild (not a one-liner but I also don't think we want to overoptimize in one use case).

```
const initialDepth = $getDepth(node);
let iterator = $dfsIterator(initialNode);
let currentDFSNode;
while ((currentDFSNode = iterator.next().value) != null) {
  const currentNode = currentDFSNode.node;
  if (delegateSubtreeProcessing) {
    delegate(currentNode);

    const [nextNode, depthDiff] =
      $getNextSiblingOrParentSibling(currentNode);
    if (
      nextNode === null ||
      currentDFSNode.depth + depthDiff <= nodeDepth
    ) {
      break;
    }
    iterator = $dfsIterator(nextNode);
  }
}
```

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -171,8 +171,8 @@ export type DFSNode = Readonly<{
  * before backtracking and finding a new path. Consider solving a maze by hugging either wall, moving down a
  * branch until you hit a dead-end (leaf) and backtracking to find the nearest branching path and repeat.
  * It will then return all the nodes found in the search in an array of objects.
- * @param startNode - The node to start the search, if ommitted, it will start at the root node.
- * @param endNode - The node to end the search, if ommitted, it will find all descendants of the startingNode.
+ * @param startNode - The node to start the search, if omitted, it will start at the root node.
+ * @param endNode - The node to end the search, if omitted, it will find all descendants of the startingNode.
  * @returns An array of objects of all the nodes found by the search, including their depth into the tree.
  * \\{depth: number, node: LexicalNode\\} It will always return at least 1 node (the start node).
  */
@@ -198,8 +198,8 @@ const iteratorNotDone: <T>(value: T) => Readonly<{done: false; value: T}> = <T>(
 
 /**
  * $dfs iterator. Tree traversal is done on the fly as new values are requested with O(1) memory.
- * @param startNode - The node to start the search, if ommitted, it will start at the root node.
- * @param endNode - The node to end the search, if ommitted, it will find all descendants of the startingNode.
+ * @param startNode - The node to start the search, if omitted, it will start at the root node.
+ * @param endNode - The node to end the search, if omitted, it will find all descendants of the startingNode.
  * @returns An iterator, each yielded value is a DFSNode. It will always return at least 1 node (the start node).
  */
 export function $dfsIterator(
@@ -207,11 +207,10 @@ export function $dfsIterator(
   endNode?: LexicalNode,
 ): DFSIterator {
   const start = (startNode || $getRoot()).getLatest();
-  const end =
-    endNode ||
-    ($isElementNode(start) ? start.getLastDescendant() || start : start);
+  const startDepth = $getDepth(start);
+  const end = endNode;
   let node: null | LexicalNode = start;
-  let depth = $getDepth(start);
+  let depth = startDepth;
   let isFirstNext = true;
 
   const iterator: DFSIterator = {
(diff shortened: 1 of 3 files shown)


packages/lexical-utils/src/index.ts
```diff
@@ -166,56 +161,113 @@ export function mediaFileReader(
...
+    [Symbol.iterator](): DFSIterator {
+      return iterator;
+    },
+  };
+  return iterator;
+}
 
-  if (node !== null && node.is(end)) {
-    nodes.push({depth, node});
+export function $getNextSiblingOrParentSibling(
```
zurfyx, line 251: Intentionally exported, this 0 added cost function is a handy utility to skip a subtree or (unrelated) help find the next sibling that is not directly connected to the same parent.

packages/lexical-utils/src/index.ts
```diff
@@ -166,56 +161,113 @@ export function mediaFileReader(
...
+  value: undefined,
+};
+const iteratorNotDone: <T>(value: T) => Readonly<{done: false; value: T}> = <T>(
+  value: T,
+) => ({done: false, value});
+
+/**
+ * $dfs iterator. Tree traversal is done on the fly as new values are requested with O(1) memory.
+ * @param startNode - The node to start the search, if ommitted, it will start at the root node.
+ * @param endNode - The node to end the search, if ommitted, it will find all descendants of the startingNode.
```
etrepum, line 202:
```suggestion
 * @param endNode - The node to end the search, if omitted, it will find all descendants of the startingNode.
```

packages/lexical-utils/src/index.ts
```diff
@@ -166,56 +161,113 @@ export function mediaFileReader(
...
+ * @returns An iterator, each yielded value is a DFSNode. It will always return at least 1 node (the start node).
+ */
+export function $dfsIterator(
+  startNode?: LexicalNode,
+  endNode?: LexicalNode,
+): DFSIterator {
+  const start = (startNode || $getRoot()).getLatest();
+  const end =
```
etrepum, line 210: Perhaps this could be left optional and check depth < startDepth if undefined? Would save some traversing

packages/lexical-utils/src/index.ts
etrepum, line 251: I like exporting this, seems very useful for when you are sure you don't want to traverse all subtrees. It could probably use its own tests since it's exported though

packages/lexical-utils/src/index.ts
```diff
@@ -166,56 +161,113 @@ export function mediaFileReader(
...
+  value: T,
+) => ({done: false, value});
+
+/**
+ * $dfs iterator. Tree traversal is done on the fly as new values are requested with O(1) memory.
+ * @param startNode - The node to start the search, if ommitted, it will start at the root node.
+ * @param endNode - The node to end the search, if ommitted, it will find all descendants of the startingNode.
+ * @returns An iterator, each yielded value is a DFSNode. It will always return at least 1 node (the start node).
+ */
+export function $dfsIterator(
```
etrepum, line 205: I do wonder what kind of trouble users will get themselves into when they mutate the document and continue iterating

zurfyx: @etrepum fair observation, how much do we want to do to prevent misuse? I'm inclined to make it flexible and not just read-only. I believe we can control the flow by listening to editor updates when the iteration is started but I do foresee this traversal can be useful even in write mode to tweak some nodes as you go as long as you understand what you're doing...

etrepum: I don't think we should try very hard, we don't really have a way to do it even if we wanted to, but there ought to be some documentation about what you should or shouldn't do and what the semantics are. I'm writing this without another close look at the code but we should also be intentional about the next state in the iteration, e.g. we should probably know enough to find what the next key is going to be before we return the node from the iterator so that if the user replaces the given node the iterator doesn't halt, or if they insert something after that the new node doesn't end up in the iterator.

potatowagon (approved):

Issue #4283, opened 2023-04-07
Feature: Prevent an action from being added to the undo stack


Atrus619:
I'm working on implementing an image node that is built in multiple steps:
1. User uploads an image file
2. Image node is created, without a src attribute and suspends
3. API call goes to the backend to retrieve a pre-signed URL to upload the image as well as an s3 key where the image will be stored
4. Image component receives the s3 key and stores it for serialization (the src of the image is not serialized in order to compress it)
5. Image ends suspension
6. Later, when the saved editor is opened up again, the s3 key is used again to ask the backend for a pre-signed url to use as the src for the image (suspending while it waits async)

Because this is a multiple step process involving multiple changes to the editor, the action of adding an image to the editor cannot be undone in one `ctrl+z`. I would like to be able to specify in some way for a set of commands to not be added to the undo stack, or to be added to the undo stack as one batched change

I haven't been able to find any good examples online or in the documentation of how to use the `@lexical/history` library to do this, it seems this only really supports undoing, redoing, clearing the editor, and sharing history between components.

Atrus619:
this works great - thank you!

PR #6666, opened 2024-09-25
[lexical-playground] 3 Bug Fixes, 1 UX Improvement: All Regarding Excalidraw Node

neysanfoo:
#### Steps to Reproduce
1. Create some drawing (non-empty)
2. Save the drawing
3. Edit the drawing
4. Clear all the elements
5. Discard Changes

Current Behaviour:
- Drawing gets deleted due to bug in "discard" function that is deleting the node if scene is clear.

New Behaviour:
- Just close the modal. So the changes are actually discarded.

#### Steps to Reproduce
1. Create some drawing (non-empty)
2. Discard Changes

Current Behaviour:
- An excalidraw node is still inserted.

New Behaviour:
- Doesn't insert a node.

### Video of Bug 1 and 2 (Before Changes)
https://github.com/user-attachments/assets/c9379c0b-28ce-4386-9f8a-5827f33fe019

### Video of Bug 1 and 2 (After Changes)
https://github.com/user-attachments/assets/2c349e3a-14cb-4389-b82e-eb83e6ac679c

#### Steps to Reproduce
1. Create some drawing (non-empty)
2. Save the drawing
3. Undo x2
4. Redo

Current Behaviour:
- Firstly, it takes 2 undos to remove the node
- Modal opens up again on first redo

New Behaviour
- Do not add the intermediate steps, i.e. opening the modal, saving changes, or discarding changes, to the undo stack as per #4283 

### Video of UX Improvement 1 (Before Changes)

https://github.com/user-attachments/assets/be581e7c-ecac-40c7-a09a-34ed7601afd9

### Video of UX Improvement 1 (After Changes)

https://github.com/user-attachments/assets/21fe4cd8-2da0-4033-b938-40d3c2a669cc

### Bug 3
Lastly, removed a small bug DecoratorNode doesn't have an exportJSON method.

### Future Work
- Track the history of resizing, similar to how the image component works


packages/lexical-playground/src/nodes/ExcalidrawNode/index.tsx
```diff
@@ -84,7 +84,6 @@ export class ExcalidrawNode extends DecoratorNode<JSX.Element> {
 
   exportJSON(): SerializedExcalidrawNode {
     return {
-      ...super.exportJSON(),
```
etrepum, line 87: Nice catch!

etrepum (commented):
It's not really clear why the history-merge tag was added to these updates, can you explain? I think that might not be the desired behavior.

neysanfoo:
> It's not really clear why the history-merge tag was added to these updates, can you explain? I think that might not be the desired behavior.

So the history-merge tags that were added are to prevent having the modal in the undo stack (ref to videos of UX improvement). The rest of the code changes are to handle the bugs.

Consider the case: Open modal -> Draw -> Save. I guess there are 2 options for how I can imagine the UX working for undoing and redoing, the first is to count the entire process of opening, drawing and saving as a single action, which is what im proposing. The second option would be 1 undo opens the modal with the current data loaded and then another undo removes the node altogether, and similarly a single redo will open the modal with the data loaded and then another redo will put the image back.

However the current process right now is one undo will delete the image without opening the modal, the second undo will actually delete the node and then a redo will show an empty modal with no data loaded in it, which is a bit weird because we didnt see the modal open when undoing, and also its empty.

So to summarize the history-merge tags were added to the 3 functions to treat the actions of open modal -> draw -> save, open modal -> draw/dont draw -> discard, edit image -> change/no change -> save, edit image -> change/no change -> discard as single actions.

etrepum:
I am not particularly familiar with this code but I suspect that this might cause some problems with collab which is the only reason I can think of to have the node's state update while the modal is open. In the single user case I think a better strategy would be to only update the state that the decorator exposes to the editor when the modal closes?

neysanfoo:
I just tried excalidraw using collab. It doesn't seem to be even usable because modal opens on both screens. Would having the toolbar open the excalidraw modal, and then only creating the node on save make sense? I think that would solve both problems.

etrepum:
I think so, at least that's the approach I would take if I was building that node + plug-in

neysanfoo:
I have moved the code for opening the modal into the ui folder and decoupled the creation of the node from the opening/editing of the modal. The node is only created on saving the modal.

neysanfoo:
Any advice how to fix the failing tests?

ivailop7:
It was the flaky set of tests, all good now. @etrepum happy to merge?

etrepum (approved):
I tested this out and it the resizing of these nodes does not persist in history, although that is not a regression in this particular PR (maybe the last one though?)

neysanfoo:
> I tested this out and it the resizing of these nodes does not persist in history, although that is not a regression in this particular PR (maybe the last one though?)

I think this has never worked, I'm planning on fixing this next.

Issue #6662, opened 2024-09-24
Bug: useCharacterLimit autolink infinite loop bug


redstar08:
copy text and paste to Editor, type any link in overflow node, infinite loop
(eg.[https://x.com])

```
useCharacterLimit autolink infinite loop bug
type a link in overflow node
http://x.
````

<img width="566" alt="image" src="https://github.com/user-attachments/assets/8b830812-df2c-4062-84fd-3ea29967973f">

Lexical version: 0.17.1

## Steps To Reproduce

1. add CharacterLimitPlugin and AutoLinkPlugin or LinkPlugin
2. type a link in overflow node

Link to code example:
https://playground.lexical.dev/?isCharLimit=true

## The current behavior
infinite loop in useCharacterLimit -> $wrapOverflowedNodes

## The expected behavior
$wrapOverflowedNodes can deal autolink node or link node

## Impact of fix

Can I fix it by modify $wrapOverflowedNodes ?

see the comment below

redstar08:
## There is a way to fixed

import {$isAutoLinkNode, $isLinkNode} from '@lexical/link';

modify function $wrapOverflowedNodes 

Judge by the $isAutoLinkNode and $isLinkNode to avoid infinite loop

<img width="1586" alt="image" src="https://github.com/user-attachments/assets/cb1eb46d-5c17-4dfe-ba5d-e4cc1550b8b4">


### Expected effect after modification
<img width="1176" alt="image" src="https://github.com/user-attachments/assets/8e8c7106-b928-4061-b93f-4f708c28b4b5">

redstar08:
Please refer to PR #6668  for discussion

PR #6668, opened 2024-09-25
[lexical-react] Bug Fix: useCharacterLimit AutoLink infinite loop bug #6662

redstar08:
## Description

Bug Fix: useCharacterLimit AutoLink infinite loop bug #6662

### Steps to Reproduce

- add CharacterLimitPlugin and AutoLinkPlugin or LinkPlugin
- type a link in overflow node

Current Behaviour:
- when type a link in overflow node infinite loop in useCharacterLimit -> $wrapOverflowedNodes

New Behaviour:
- autolink is ok in overflow node

### DIFF
--- a/packages/lexical-react/src/shared/useCharacterLimit.ts
+++ b/packages/lexical-react/src/shared/useCharacterLimit.ts
@@ -8,6 +8,7 @@
 
 import type {LexicalEditor, LexicalNode} from 'lexical';
 
+import {$isAutoLinkNode, $isLinkNode} from '@lexical/link';
 import {
   $createOverflowNode,
   $isOverflowNode,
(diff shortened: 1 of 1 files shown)


etrepum (commented):
It does seem strange that this would be the best way to fix this bug. Ideally this wouldn't need to know anything about node types other than TextNode, ElementNode and DecoratorNode. Seems like a problem that is in the algorithm either here or in AutoLinkNode (e.g. if it's "grabbing" text from too far outside of its container). If we were to rewrite the character limit plugin in its own package without a react dependency it would be unexpected that it depends on the link package.

redstar08:
> It does seem strange that this would be the best way to fix this bug. Ideally this wouldn't need to know anything about node types other than TextNode, ElementNode and DecoratorNode. Seems like a problem that is in the algorithm either here or in AutoLinkNode (e.g. if it's "grabbing" text from too far outside of its container). If we were to rewrite the character limit plugin in its own package without a react dependency it would be unexpected that it depends on the link package.

The analysis is quite correct, it's seems to be the fastest, but not the best solution, and I try to reconsider from the algorithm in the $wrapOverflowedNodes function, and think a way how to avoid judging directly by AutoLink node.

Issue #6614, opened 2024-09-10
Bug: undo causes unexpected paragraph deletion in collaborative editor


mshafer:
When using `CollaborationPlugin` and two users are editing the same paragraph, an undo from one user can cause Lexical and the Yjs doc to get out of sync, leading to the whole paragraph getting cleared on page refresh.

Lexical version: 0.17.1

In this screen recording, observe the behaviour after User 1 (left) clicks the undo button: it unexpectedly deletes a character from User 2's word, then after a refresh of the page the whole paragraph is cleared:

https://github.com/user-attachments/assets/3f631b41-99a2-4cd3-8402-2a9334e06877

When we replicate the issue in our own application we have detection code that shows Lexical editor state and the Yjs document are getting out of sync immediately after the user presses undo. Lexical thinks the second paragraph has the text node with User 2's text, while the Yjs doc thinks the second paragraph node is empty. 

## Steps To Reproduce

1. Open two windows to https://playground.lexical.dev/?isCollab=true
2. On the left, put the cursor at the start of an empty second paragraph. (The issue still presents in the first paragraph, but this demonstrates it's not unique to the start of the document.)
3. On the left, type: `This is a`, `<backspace>`, `a test. `
4. On the right, type: `Word`
5. On the left, click the "Undo" button until all of `This is a test.` is removed. Sometimes this happens after one undo, sometimes multiple undo's.
6. Refresh the page

## The current behavior

The main unexpected behaviour is that after the undo operation, a refresh of the page causes the whole paragraph to be cleared, including the text from User 2.

In the specific case captured in the screen recording, it was also unexpected that User 1's undo operation deleted the `W` of `Word` from User 2.

## The expected behavior

User 1's undo operation should undo only the text they typed, and a refresh of the page after the undo should not lead to the paragraph content getting cleared. 

## Impact of fix

This bug can cause loss/corruption of data for all users of the Lexical collaboration plugin.

Note that although the data corruption behaviour is _usually_ the deleted paragraph as shown in the screen recording, in some cases we were able to reproduce the type of duplicated and garbled text seen in https://github.com/facebook/lexical/pull/6523 and https://github.com/facebook/lexical/pull/6374. We can't reliably reproduce this now, but it required both User 1 and User 2 performing edits on the affected paragraph right after pressing undo but **before** refreshing the page. We know at that point Lexical and Yjs are out of sync, so some combination of edits leads to further corruption.

mshafer:
[...]

PR #6670, opened 2024-09-26
[lexical-yjs][lexical-playground] Chore: Add test for collaboration undo bug

mike-atticus:
## Description

Adds a test for the issue https://github.com/facebook/lexical/issues/6614 where using undo in a collaborative session results in the Lexical editor content getting out of sync with the Yjs doc, which then leads to loss of document content. 

### Before

No tests to detect the bugs that lead to unexpected loss of editor content after using undo. 

### After

Playwright test added which uses the side-by-side collaborative playground editors to reproduce the bug after using undo.

Note that this test actually detects two separate bugs:

1. When the left-hand side collaborator clicks undo, the action removes a character of the text typed by the right-hand side collaborator, which shouldn't happen.
2. If you comment out the assertion that checks for bug 1, you'll see that the second paragraph is cleared (including RHS collaborator's text) after LHS collaborator refreshes their page.

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
@@ -230,4 +230,92 @@ test.describe('Collaboration', () => {
       focusPath: [1, 1, 0],
     });
   });
+
+  test('Undo with two collaborators editing same paragraph', async ({
+    isRichText,
+    page,
+    isCollab,
+    browserName,
+  }) => {
+    // test.skip(!isCollab || IS_MAC);
+    test.skip(!isCollab);
+
+    // Left collaborator types two paragraphs of text
+    await focusEditor(page);
+    await page.keyboard.type('Line 1');
+    await page.keyboard.press('Enter');
+    await sleep(1050); // default merge interval is 1000, add 50ms as overhead due to CI latency.
+    await page.keyboard.type('This is a test. ');
+
+    // Right collaborator types at the end of paragraph 2
+    await sleep(1050);
+    await page
+      .frameLocator('iframe[name="right"]')
+      .locator('[data-lexical-editor="true"]')
+      .focus();
+    await page.keyboard.press('ArrowDown'); // Move caret to end of paragraph 2
+    await page.keyboard.press('ArrowDown');
+    await page.keyboard.type('Word');
+
+    await assertHTML(
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">Line 1</span>
+        </p>
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">This is a test. Word</span>
+        </p>
+      `,
+    );
+
+    // Left collaborator undoes their second paragraph
+    await sleep(1050);
+    await page.frameLocator('iframe[name="left"]').getByLabel('Undo').click();
+
+    // Only left collaborator's text should have been undone
+    await assertHTML(
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">Line 1</span>
+        </p>
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">Word</span>
+        </p>
+      `,
+    );
+
+    // Left collaborator refreshes their page
+    await page.evaluate(() => {
+      document
+        .querySelector('iframe[name="left"]')
+        .contentDocument.location.reload();
+    });
+
+    // Page content should be the same as before the refresh
+    await assertHTML(
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">Line 1</span>
+        </p>
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">Word</span>
+        </p>
+      `,
+    );
+  });
 });


packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -230,4 +230,92 @@ test.describe('Collaboration', () => {
       focusPath: [1, 1, 0],
     });
   });
+
+  test('Undo with two collaborators editing same paragraph', async ({
```
potatowagon, line 234: this is failing on the ci :( 

the failure is reproducible across different envs so i dont think its a flaky failure

etrepum: That's the point of this PR, to demonstrate that bug

mike-atticus: Yea I created this test in case it helps others with reproducing the bug and confirming a fix, but it's not something that can be merged on its own. Let me know if I should switch this to a draft PR for now, if that's clearer for PR queue management.

potatowagon: i see, thanks for filing a detailed issue. if its not meant to be merged, converting to draft or adding a [do not merge] to title would work

mike-atticus: Title updated 👍

james-atticus:
@etrepum @ivailop7: @mike-atticus and I spent some more time on this and we think we have a fix...

**Our understanding of the issue**

When creating a new paragraph, it starts off with no children.

When a user types a character 'A', we add two things to the paragraph's `CollabElementNode` `_xmlText`: a `YMap` which represents the `TextNode` and the string 'A'. These are added to YJS's undo stack together.

The second user then types character 'B', which adds _only_ the string 'B' to `_xmlText`.

When the first user then clicks Undo, both the `YMap` and 'A' are deleted, however 'B' remains in `_xmlText`. @trueadm added handling for this case in [here](https://github.com/facebook/lexical/pull/1005/files#diff-db90e519357a15ba382479416977bd91e61ae416794f0da2f0c838b5bcdc29cbR188), cleaning up this dangling text when the document is reloaded. However, until that point, Lexical and YJS are out of sync.

**Proposed fix**

The same PR above added [some logic](https://github.com/facebook/lexical/pull/1005/files#diff-db90e519357a15ba382479416977bd91e61ae416794f0da2f0c838b5bcdc29cbR147-R156) for detecting when a text node was deleted leaving dangling text, and tried to merge said text into the previous text node. However, the checks were too strict: the `YMap` can be deleted when `delCount > 1`, as is the case with undo above.

The first change in this PR is fixing the check to properly handle cases when the `YMap` has been deleted from YJS (`if (offset === 0 && length === nodeSize) ...`). The other part is updating the dangling text handling:
* Get the dangling text with `spliceString`
* If there's a preceding text node, move the text to that node
* If there's no preceding text, remove the text from `_xmlText`; i.e.: the same thing as @trueadm's original fix but immediately

Note: this slightly differs from the test case that @mike-atticus originally submitted. At the time, we didn't really understand how Lexical and YJS synced with each other. Once we found @trueadm's PR, we realised the text node is fully deleted from YJS with the undo operation, so it was basically the same issue. Given this [TODO](https://github.com/facebook/lexical/pull/1005/files#diff-db90e519357a15ba382479416977bd91e61ae416794f0da2f0c838b5bcdc29cbR180-R187) about a "proper" fix, we opted to keep it simple and delete the dangling text.

**Next steps**

If Lexical you're are happy with this approach, we'll move the tests over to `Collaboration.test.ts` so that we can add checks on the YJS state and speed up the tests (no need for `sleep`).

etrepum:
The analysis and proposed fix makes sense to me, although I have never done an audit of the collab code so I could be missing something. I think the important things to consider are:

* Is the fix largely backwards compatible (at least such that coherent documents work the same way)? If not, is there another approach that would be more compatible?
* Does it add new edge cases?

The fix sounds simple enough that I think it would be backwards compatible and not cause new trouble, but it's hard to see without at least observing a full e2e collab test suite since there really is a lot going on in here.

Thank you for investigating further!

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -230,4 +230,163 @@ test.describe('Collaboration', () => {
...
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">Line 1</span>
+        </p>
+        <p class="PlaygroundEditorTheme__paragraph"><br /></p>
+      `,
+    );
```
james-atticus, line 292: Fails on `main` with:

```

    - Expected  - 1
    + Received  + 5

      <p
        class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
        dir="ltr">
        <span data-lexical-text="true">Line 1</span>
      </p>
    - <p class="PlaygroundEditorTheme__paragraph"><br /></p>
    + <p
    +   class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
    +   dir="ltr">
    +   <span data-lexical-text="true">ord</span>
    + </p>
```

The text is `ord` because Lexical skips a character, as it thinks there's still a `YMap` for the `TextNode`.

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -230,4 +230,163 @@ test.describe('Collaboration', () => {
...
+    await assertHTML(
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">normalBOLD</span>
+        </p>
+      `,
+    );
```
james-atticus, line 371: Fails on main with:

```

    - Expected  - 1
    + Received  + 4

      <p
        class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
        dir="ltr">
    -   <span data-lexical-text="true">normalBOLD</span>
    +   <span data-lexical-text="true">normal</span>
    +   <strong class="PlaygroundEditorTheme__textBold" data-lexical-text="true">
    +     OLD
    +   </strong>
```

Similar to the test above, a character is dropped because Lexical thinks there's still a `YMap` for the bolded `TextNode`.

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -230,4 +230,163 @@ test.describe('Collaboration', () => {
...
```
james-atticus, line 390: On main, this `assert` passes for the left editor (because the refresh got it back in sync) but not the right editor (which will only get back in sync when it refreshes).

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -230,4 +230,163 @@ test.describe('Collaboration', () => {
...
+      page,
+      html`
+        <p
+          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+          dir="ltr">
+          <span data-lexical-text="true">Line 1</span>
+        </p>
+        <p class="PlaygroundEditorTheme__paragraph"><br /></p>
+      `,
+    );
```
james-atticus, line 312: If you comment out the `assert` above, this passes on `main` because of [this handling](https://github.com/facebook/lexical/blob/4fbe7fec1a6f7c44bd828ec11460751f9cbcfe16/packages/lexical-yjs/src/CollabElementNode.ts#L208). The left editor's change is synced over to the right editor.

packages/lexical-yjs/src/CollabElementNode.ts
```diff
@@ -157,21 +157,25 @@ export class CollabElementNode {
...
-              length === nodeSize &&
-              // If the node has no keys, it's been deleted
-              Array.from(node._map.keys()).length === 0
-            ) {
-              // Merge the text node with previous.
-              prevCollabNode._text += node._text;
-              children.splice(nodeIndex, 1);
-            } else if (offset === 0 && delCount === nodeSize) {
-              // The entire thing needs removing
+            if (offset === 0 && length === nodeSize) {
```
james-atticus, line 160: Our observation is that `length < nodeSize` when you're just deleting some text from the node, and `length === nodeSize` if the `YMap` has been deleted. This matches up with the old code's first `if` branch (which also had `Array.from(node._map.keys()).length === 0` but we couldn't see why that was necessary). It also aligns with the second branch `delCount === nodeSize`, where `delCount = min(length, deletionSize)`.

james-atticus (commented):
> Is the fix largely backwards compatible?

I believe so, yes. We're only changing logic for processing `delete` deltas from YJS, which shouldn't impact loading existing documents (pretty sure YJS just gives you `insert` events on load). For existing documents that have dangling text at the start of a paragraph, that text will still be deleted from YJS on load by the logic [here](https://github.com/facebook/lexical/blob/4fbe7fec1a6f7c44bd828ec11460751f9cbcfe16/packages/lexical-yjs/src/CollabElementNode.ts#L208).

> Does it add new edge cases?

Not that we could think of. The change simplifies the overall deletion-handling logic; previously there were two branches that called `children.splice(nodeIndex, 1)`, each in very specific circumstances (either `delCount === 1` or `delCount === nodeSize)`. Now it's "if YMap is deleted, _always_ `splice` the CollabTextNode out", regardless of `delCount`, with some logic to either move dangling text to another TextNode or clean it up from YJS (`delCount` is used to determine what part of the text is left over).

We've pushed the code for the proposed fix as well as the two E2E tests that were failing before the change, covering the two paths for handling of dangling text. Are these what you meant by "full e2e collab test suite"? In case it's helpful for understanding how we got to these tests, I've added some comments inline highlighting where they're failing on `main`.

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -230,4 +230,163 @@ test.describe('Collaboration', () => {
...
   });
+
+  test('Remove dangling text from YJS when there is no preceding text node', async ({
+    isRichText,
+    page,
+    isCollab,
+    browserName,
+  }) => {
+    // test.skip(!isCollab || IS_MAC);
+    test.skip(!isCollab);
```
james-atticus, line 241: Ditto below

```suggestion
    test.skip(!isCollab || IS_MAC);
```

ivailop7:
Pretty confident this is backwards-compatible too. It doesn't modify the stored Ydoc and how to interpret it. I'm happy with the change and tests.

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
etrepum: Is there a reason this should be skipped on Mac? I am not seeing any platform specific code here, I think what this will do is make it harder for contributors to test locally on a Mac because some of the tests just won't run.

james-atticus: No particular reason, just pattern matching existing tests. Works fine on Mac so happy to leave as is.

etrepum (approved):
Given that all the test suites seem to be passing now I don't see any reason not to approve and merge this fix with additional tests. Thank you very much for doing all the hard work on this one!

PR #6675, opened 2024-09-26
[lexical] Bug Fix: Fix Table formatting when no TextNodes present

KatsiarynaDzibrova:
## Description
Don't rely on TextNode when applying formatting for empty ParagraphNode
The initial issue was fixed in https://github.com/facebook/lexical/pull/6626, but the scenario with no TextNodes present was missed in the final version of pr

## Test plan
2 tests added


ivailop7:
This is still not working for the styles - font size, font color, etc for the empty cells. Also when selecting the whole table and clicking on Bold, doesn't keep the toolbar button active, which if all cells have the applied said style should have active state.

KatsiarynaDzibrova:
Fixed toolbar buttons in https://github.com/facebook/lexical/pull/6678

Regarding text styles: ParagraphNode does not have properties to store them right now. We will need to add new properties in ParagraphNode and new create signals in Toolbar to process the action separately for table the same way it's done with formatting now. I'd rather do this in a separate pr as this is not a small change

ivailop7:
> Fixed toolbar buttons in #6678
> 
> Regarding text styles: ParagraphNode does not have properties to store them right now. We will need to add new properties in ParagraphNode and new create signals in Toolbar to process the action separately for table the same way it's done with formatting now. I'd rather do this in a separate pr as this is not a small change

Agreed. This PR is a great step forward.

ivailop7 (approved):

ivailop7:
@etrepum could I get a second pair of eyes, if you don't mind.

etrepum (approved):
LGTM, thank you!

PR #6678, opened 2024-09-27
[lexical-table][lexical-playground] Bug Fix: Make style buttons in toolbar respect table selection

KatsiarynaDzibrova:
## Description
Make style buttons in toolbar active for TableSelection

### Before
Styles indicators don't change
toolbar-bold-before

### After
Styles indicators change when selection includes styled text
toolbar-bold-after


ivailop7 (approved):
Great work!

PR #6679, opened 2024-09-27
Bug Fix: Shift+down selects an extra subsequent element  for Table selection

### LINKED ISSUE #6527: Bug: Shift+down selects an extra subsequent element
https://www.loom.com/share/78508267145b422081498fe08c461626?sid=13463229-4383-4ec6-80d2-741805a691c7

Lexical version: 0.17

## Steps To Reproduce

1. Insert a paragraph with some text.
2. Insert a 2x2 table.
3. Delete the preceding paragraph.
4. Insert a paragraph after the table, with some text.
5. Press "Shift+Down arrow" at the end of the first paragraph.

## The current behavior
Both the table and the second paragraph are selected.

## The expected behavior
Only the table should be selected.

## Impact of fix

- The bug doesn't happen with just a subsequent paragraph. If there is an image or a second table, they will be included in the new selection too. This could mess up deletion if we're trying to delete just the table. 
- The DOM selection is applied correctly when "shift+down"-ing into a table, with the next `focusNode` being the last cell. However, when the `selectionchange` event fires, the DOM selection now includes the next element after the table.

KatsiarynaDzibrova:
## Description
Only the table should be selected when shift+down. Also fixes scenario when selection is stuck when trying to unselect a table with arrow keys.

Closes #6527 for Tables

### Before
Paragraph after table is selected
table-selection-shiftdown-before

### After
Only table is selected
table-selection-shiftdown-after

### Before
Selection stuck
table-unselect-shiftdown-before

### After
Can unselect tables one by one
table-unselect-shiftdown-after


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1441,10 +1491,13 @@ function $handleArrowKey(
           }
           return false;
         } else {
-          const focusParentNode = $findMatchingParent(
+          let focusParentNode = $findMatchingParent(
             focusNode,
             (n) => $isElementNode(n) && !n.isInline(),
           );
+          if ($isTableCellNode(focusParentNode)) {
+            focusParentNode = getParentTable(focusParentNode);
```
ivailop7, line 1499: why not just $findMatchingParent(focusParentNode, $isTableNode), but needing that extra helper method, it will return the table node or null, just like your method.

ivailop7:
@etrepum, could you have a look at this one when you have a spare moment.

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -981,8 +981,8 @@ test.describe.parallel('Selection', () => {
     await assertSelection(page, {
       anchorOffset: 0,
       anchorPath: [0],
-      focusOffset: 0,
-      focusPath: [2],
+      focusOffset: 1,
+      focusPath: [1, 2, 1],
```
KatsiarynaDzibrova, line 985: Focus will now be inside the last cell of the table, not on the next element after table

etrepum (commented):
It still seems possible to get "stuck" if you remove any elements between the tables

For example, using [this document](https://lexical-playground-ehiwb4pzc-fbopensource.vercel.app/#doc=H4sIAAAAAAAAE-3bTWvCQBAG4P8y50USpRj3Wui5kEMP4mHd3ZrQbSZsJlqR_PfiF6iHYj1I0BcECWby8WzmZcnihrwrhWMuRjzpDUVm2X7bogwu-or09JaNmSJXRm-l5Ip01Yag6JPjtxHSRIrKyvlKSCeKZF170lSbaBbR1AUpWvrY7ApTReJ_5O1Qmew3c1mHbQl1_z-NmHnw1odwcZq5sV-LyG3lXjlwPB7Mcshrs9-l8Mb5o9RIUeTV4bdOPbFECglIQOIqiRuvI_Lq7DL-En0u3iEetINEAglIQOIqCcQwYhgtBwlIIIYfiRcxjJaDBCQQw4jhXkig5SABifvG8I3HubgXy-GjdFI0pKeToTr5zK4O-ucaTawIYvUHEpDAiiAm372QwEQLEpDAOxDEcI8eNEhAAhKIYcQwWg4SkOilBGIYMYyWgwQkHiKG770i2Nth2v258YymUxRMI7lZekc6HQ-zcTbKXkaTLFPUcBvttu49mPV-GE-rKRmk2SCh7hcvHDRMRTkAAA) with two tables and a paragraph, if you focus the last paragraph and then press shift up it works to select the first table but not the second

KatsiarynaDzibrova:
Addressed your comment in [0a47177](https://github.com/facebook/lexical/pull/6679/commits/0a471774e3a762b70d48b8fca5a7a3b43ac4a3db)

PR #6683, opened 2024-09-29
[lexical-yjs] Bug Fix: Add missing setLocalStateField method to ProviderAwareness type

collierrgbsitisfise:
## Description

The current type definition doesn't include the `setLocalStateField` method, which can lead to TypeScript errors when using this method and prevents proper code completion in IDEs. This change improves type safety and developer experience when working with the `lexical-yjs` package.

This PR adds the missing `setLocalStateField` method to the `ProviderAwareness` type in the `lexical-yjs` package([ref](https://github.com/yjs/y-protocols/blob/master/awareness.js#L141)). This method is present in the actual implementation of awareness providers but was not reflected in the TypeScript type definition.

### Before

--

### After

--

### DIFF
--- a/packages/lexical-react/src/__tests__/unit/LexicalCollaborationPlugin.test.tsx
+++ b/packages/lexical-react/src/__tests__/unit/LexicalCollaborationPlugin.test.tsx
@@ -53,6 +53,7 @@ describe(`LexicalCollaborationPlugin`, () => {
             off: () => {},
             on: () => {},
             setLocalState: () => {},
+            setLocalStateField: () => {},
           },
           connect: () => {},
           disconnect: () => {},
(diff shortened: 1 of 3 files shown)


etrepum (commented):
The tests don't typecheck with this change because it adds a new required property where one wasn't needed before. It's unclear what the utility of this backwards incompatible change would be since this method isn't called anywhere in @lexical/yjs

collierrgbsitisfise:
> The tests don't typecheck with this change because it adds a new required property where one wasn't needed before. It's unclear what the utility of this backwards incompatible change would be since this method isn't called anywhere in @lexical/yjs

[Updated tests as well according new awareness contract](https://github.com/facebook/lexical/pull/6683/commits/fe06417aba2fe22284d1747cfafb84e54244f0b3). Nothing functional  to test, so i replaced functional implementation with `// TODO` placeholder by analogy with [on/off methods](https://github.com/facebook/lexical/pull/6683/commits/fe06417aba2fe22284d1747cfafb84e54244f0b3#diff-c79ad391bc15c1972e508abdbed351bc4933e470907dde99715132498da52f06R98-R104)

ivailop7 (approved):
LGTM

etrepum (approved):

PR #6684, opened 2024-09-29
[lexical][lexical-list] Bug Fix: Retain indentation when list format is removed and reapplied

SeaBa55:
## Description

There is an issue in the lexical-list package where if the user removes the current applied list format, and then reapplies the same (or other) list format, indentation will not be retained. This results in a poor user experience with lists dropping indentation, for simply toggling the list format. 

After some debugging, I saw that there was an attempt to handle this behavior within the `$createListOrMerge` method of `formatList.ts`, however, the indentation was not being applied in the newly added list nodes.

I found that the indentation would be applied to the new list nodes, only if the following invocations are applied after the list is attached to root:
```
listItem.setFormat(node.getFormatType());
listItem.setIndent(node.getIndent());
```

This PR does not close any issue. I realize I should have created an issue prior, and if the team requires this I can do so. Just let me know. 

### Before

https://github.com/user-attachments/assets/42e96afa-22df-4541-bcd2-0e327416ca87

### After

https://github.com/user-attachments/assets/52b9ad0d-a646-4b92-9a16-c6b5ab750b11


ivailop7 (approved):
LGTM, can I get a second look from someone please

etrepum (approved):
It seems like it's probably fine but I would be a lot happier about merging if there was at least one test to show that it does fix the given issue

etrepum (approved):
I can't repro any issues with the tests under react-beta locally, so maybe it's some kind of CI cache issue. Changes look good otherwise!

PR #6686, opened 2024-09-30
[lexical-yjs] Enhance: Improve ProviderAwareness type in lexical-yjs package

collierrgbsitisfise:
## Description

Improving the type signature of the `getStates` method to support generic extensions of the `UserState` type. Addition to my prev [PR](https://github.com/facebook/lexical/pull/6683).

### Before

image
image

### After

image
image


collierrgbsitisfise:
> This doesn't really add any type safety because the generic type T is associated with the method and not the class, so you can call it multiple times with different types on the same instance.

As i understand there is no direct access to class(instance initialization) this logic is hidden by provider(`y-websocket` -> `y-protocols/awareness`). At least it gives possibility to extend **userState** instead of using **casts** or **ts-ignore**

As improvement mb, it make more sense also always marks extended fields as optional, so developers will be forced explicitly check it

```typescript
getStates: <T extends Record<string, unknown>>() => Map
  number,
  UserState & { [K in keyof T]?: T[K] }
>;
``` 

i believe it bring more value and type safety

etrepum:
Something like `UserState & Record<string, unknown>` would be safer, the generic in this way is only as safe as using an `as` cast but because it can be triggered by inference it's not obvious that you're doing an unsafe cast.

etrepum (commented):
I think there are three acceptable solutions to this:

* Add `Record<string, unknown>` or an index signature to UserState to allow arbitrary extra data
* Use interface instead of type to declare `UserState` so that it can be extended in user code with [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)
* Refactor all of the related types for yjs (Provider, ProviderAwareness, most of the functions that one of these as an argument) to make UserState a generic parameter that extends UserState (and defaults to UserState) but is carried around in the correct way for type safety

collierrgbsitisfise:
> I think there are three acceptable solutions to this:
> 
> * Add `Record<string, unknown>` or an index signature to UserState to allow arbitrary extra data
> * Use interface instead of type to declare `UserState` so that it can be extended in user code with [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)
> * Refactor all of the related types for yjs (Provider, ProviderAwareness, most of the functions that one of these as an argument) to make UserState a generic parameter that extends UserState (and defaults to UserState) but is carried around in the correct way for type safety

At the moment the first option seems to be the most optimal. Extended **UserState** in terms of next PR: https://github.com/facebook/lexical/pull/6724

etrepum:
I think we can close this in favor of https://github.com/facebook/lexical/pull/6724

PR #6688, opened 2024-10-01
[lexical] Bug Fix: removeText for token text nodes

### LINKED ISSUE #6687: Bug: removeText misbehaves for text nodes in token mode
Lexical version: 0.18.0

## Steps To Reproduce

1. Create a text node in `token` mode
2. Try to partially delete the text
3. It will mutate the node, even though, according to the docs, `token` nodes are immutable and will be deleted all at once.

Link to code example:

You can check it out on the current playground.

## The current behavior

When removing text, text nodes in `token` mode aren't deleted in full, but rather a new non-`token` node is created. This is the expected behaviour for `segmented` nodes, but not for `token` nodes.

## The expected behavior

When I remove text that is part of a `token` text node, the entire `token` text node is deleted.

## Impact of fix

The current version is not complying with the documented behaviour of `token` nodes, which can cause issues for everyone that is using TextNodes in `token` mode. Also, this is a regression introduced in #6456

adrianmxb:
## Description

This PR fixes an issue where text nodes in token mode were not being deleted properly when partially removed. According to the documentation, token nodes should be immutable and should be deleted as a whole. However, the current behaviour allowed partial deletion of the text within a token node, resulting in an unexpected mutation into a non-token node.

1. In `fixText` added another condition treating a `token` node similiar to an empty text node, effectively removing the node.
2. Consistent Return: `fixText` now always returns a node, eliminating the need for the || operator in assignments like firstNode = fixText(firstNode, del) || firstNode;. This simplifies the code and prevents potential undefined errors.

Closes #6687

### Before

CleanShot 2024-10-01 at 18 28 12

### After

CleanShot 2024-10-01 at 18 27 14

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1130,14 +1130,15 @@ export class RangeSelection implements BaseSelection {
     });
 
     const fixText = (node: TextNode, del: number) => {
-      if (node.getTextContent() === '') {
+      if (node.getTextContent() === '' || (del !== 0 && node.isToken())) {
         node.remove();
-      } else if (del !== 0 && $isTokenOrSegmented(node)) {
+      } else if (del !== 0 && node.isSegmented()) {
         const textNode = $createTextNode(node.getTextContent());
         textNode.setFormat(node.getFormat());
         textNode.setStyle(node.getStyle());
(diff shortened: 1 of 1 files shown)


adrianmxb:
Closed in favour of #6690

PR #6690, opened 2024-10-01
[lexical] Bug Fix: TextNode in token mode should not be split by removeText

### LINKED ISSUE #6687: Bug: removeText misbehaves for text nodes in token mode

etrepum:
## Description

The #6456 removeText rewrite introduced a regression in token mode behavior where it behaved too similarly to segmented nodes, it would be split and replaced with a plain TextNode instead of being deleted in full. This takes the simple approach of just expanding the selection outwards if the focus or anchor node contains any content from a token node.

Closes #6687

### Before

There were no tests to show that token TextNode should be deleted in full by removeText

### After

Now there are tests that cover removeText behavior for token TextNode

### DIFF
--- a/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
+++ b/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
@@ -344,6 +344,157 @@ describe('LexicalSelection tests', () => {
       });
     });
     describe('removeText', () => {
+      describe('with a leading TextNode and a trailing token TextNode', () => {
+        let leadingText: TextNode;
+        let trailingTokenText: TextNode;
+        let paragraph: ParagraphNode;
+        beforeEach(() => {
+          testEnv.editor.update(
+            () => {
+              leadingText = $createTextNode('leading text');
+              trailingTokenText =
+                $createTextNode('token text').setMode('token');
+              paragraph = $createParagraphNode().append(
+                leadingText,
+                trailingTokenText,
+              );
+              $getRoot().clear().append(paragraph);
+            },
+            {discrete: true},
+          );
+        });
+        test('remove all text', () => {
+          testEnv.editor.update(
+            () => {
+              const sel = $createRangeSelection();
+              sel.anchor.set(leadingText.getKey(), 0, 'text');
+              sel.focus.set(
+                trailingTokenText.getKey(),
+                trailingTokenText.getTextContentSize(),
+                'text',
+              );
+              $setSelection(sel);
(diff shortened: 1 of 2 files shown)


adrianmxb:
Appreciate the proposed changes and prefer this approach. Since I didn’t have the chance to follow up with tests earlier and you seem to have everything covered, I’ve gone ahead and closed my PR in favour of this one. That said, I do have a suggestion: it might be worth updating the `fixText` function to only create a new Text Node when in `segmented` mode. This would help prevent potential confusion later on and address the flawed code.

etrepum:
I don't think it's really worth the code churn to change `fixText`, the code now handles the token cases correctly because they all fall under `node.getTextContent() === ''` or `del === 0`. If I wrote this originally I might do it a bit differently but the logic down there doesn't seem wrong.

I do wonder if this makes sense at all since it seems that the logic to handle segmented deletion is elsewhere but if this better matches the previous behavior then it probably is the right thing to do.

adrianmxb:
Just to be clear, I only talk about replacing `$is`**Token**`OrSegmented(...)` through a `node.isSegmented()` call. I would argue that relying on implicit behaviours (like node.getTextContent() === '' or del === 0) to exclude token nodes isn't clear to readers.

Making the exclusion explicit avoids potential confusion.

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -344,6 +344,295 @@ describe('LexicalSelection tests', () => {
       });
     });
     describe('removeText', () => {
+      describe('with a leading TextNode and a trailing token TextNode', () => {
```
potatowagon, line 347: thanks for adding a test case

potatowagon (approved):

Issue #5062, opened 2023-09-28
Bug: --lexical-indent-base-value does not work upon initialization


immkoala:
I have a paragraph node with --lexical-indent-base-value set to 20px. 

However, during initialization, padding-inline-start is still set to 40px.

I debugged a bit and preliminarily suspect the following logic to be the issue:

```javascript
const indentationBaseValue =
    getComputedStyle(dom).getPropertyValue('--lexical-indent-base-value') ||
    DEFAULT_INDENT_VALUE;
```

When executing getComputedStyle(dom).getPropertyValue('--lexical-indent-base-value'), the dom node hasn't been inserted into the document yet, causing the retrieved value to be empty.

chschoeps:
Are there any updated regarding this issue? Is there any workaround?

etrepum:
Should be solved by now, e.g. #8536 #8466

Issue #4583, opened 2023-05-31
Bug: custom theme.indent value not respected when rendering a stored editor content


lysender:
When using the `TabIndentationPlugin` plugin, custom indent size from the theme css not respected when rendering back the editor content from storage (DB for example) to the editor. It is only respected during the time the content is being typed into the editor.

For example, the default indent is 40px. Tried to override it to 20px;

```
.PlaygroundEditorTheme__indent {
  --lexical-indent-base-value: 20px;
}
```

Lexical version: 0.11.1

![image](https://github.com/facebook/lexical/assets/114853/b540cce6-f98d-40f9-9abd-eab128292bb8)

## Steps To Reproduce

1. Setup an editor to have the `TabIndentationPlugin` plugin with 20px indention override on theme css
2. Start the editor with empty content
3. Enter a few lines of content with indention, 1 level of indention should be enough
4. Capture the editor state to json string, probably store it somewhere
5. Render the same editor but set the initial content to the value from above
6. Check the rendered content

Link to code example:

CodeSandbox: https://codesandbox.io/s/magical-david-o3y42f?file=/src/Editor.js

Demo repo: https://github.com/lysender/lexical-text-indent

## The current behavior

The rendered content still has 40px indention

## The expected behavior

Should have 20px indention

samirant15:
Seems like the issue is in the `.PlaygroundEditorTheme__indent` selector.

For me it worked by setting it at the root:
```css
:root {
  --lexical-indent-base-value: 20px;
}
```

vimutti77:
> Seems like the issue is in the `.PlaygroundEditorTheme__indent` selector.
> 
> For me it worked by setting it at the root:
> 
> ```css
> :root {
>   --lexical-indent-base-value: 20px;
> }
> ```

It doesn't work for me.

PR #6693, opened 2024-10-03
[lexical][lexical-list][lexical-rich-text]: Fix: Preserve indentation when serializing to and from HTML

### LINKED ISSUE #6082: Feature: paragraphNode should apply configured indent value from theme on exportDOM

GermanJablo:
## Description

This PR ensures that the indent property is preserved when importing and exporting HTML.

Closes #6082

Additionally, the following bugs are fixed:
- in `ParagraphNode.exportDOM` the correct default value is used. It was set to an old value.
- in `ParagraphNode.exportDOM` the indent is exported using the `padding-inline-start` property as used in the editor, instead of `text-indent`. Apparently, the person who set that did so because `padding-inline-start` has poor support in email clients. However, `text-indent` is a first-line indent, different from the lexical indent which encompasses the entire block. If someone needs to export for use in email, they should probably create their own converter.

### Considerations

This PR does not include serialization of custom indent values ​​defined in the editorConfig.theme. The problem is that the CSS variable `--lexical-indent-base-value` is used, but when importing html the variable is not accessible. It can be used in any context, even headless. This has already caused problems for other users:

https://github.com/facebook/lexical/issues/4583
https://github.com/facebook/lexical/issues/5062

As I see it, the solution to this would be that the indent spacing is not configured with CSS under the editorConfig theme property, but using javaScript. That would be a breaking change.

## Test plan

Tests are included for Paragraphs, Quotes and Headings.

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -155,7 +155,7 @@ export class ListNode extends ElementNode {
   }
 
   exportDOM(editor: LexicalEditor): DOMExportOutput {
-    const {element} = super.exportDOM(editor);
+    const element = this.createDOM(editor._config, editor);
     if (element && isHTMLElement(element)) {
       if (this.__start !== 1) {
         element.setAttribute('start', String(this.__start));
--- a/packages/lexical-playground/__tests__/unit/docSerialization.test.ts
+++ b/packages/lexical-playground/__tests__/unit/docSerialization.test.ts
@@ -15,7 +15,13 @@
  */
 
 import {serializedDocumentFromEditorState} from '@lexical/file';
-import {$createParagraphNode, $createTextNode, $getRoot} from 'lexical';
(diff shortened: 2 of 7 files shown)


GermanJablo:
I've opened a PR fixing the indent issue here: https://github.com/facebook/lexical/pull/6693

I'll do the same for text-align soon.

If you're interested, contributions are still welcome to make HTML serialization take into account custom indentation values ​​in `editorConfig.theme.indent`

GermanJablo:
I think the failed tests are flaky

etrepum (approved):

PR #6698, opened 2024-10-04
[lexical-playground] Bug Fix: match toolbar font size input with the rest of toolbar items in Read-Only mode

### LINKED ISSUE #6697: Bug: font size input doesn't seem disabled when playground is in Read-Only mode
https://github.com/user-attachments/assets/6bf4f972-f92e-474b-9e0f-2a0d98972e8f

Lexical version: v0.18.0

## Steps To Reproduce

1. Switch to Read-Only mode
2. Hover over the font size input

## The current behavior
The opacity doesn't match the rest of toolbar items and cursor is still `default`

## The expected behavior
The opacity should match the rest of toolbar items and cursor should be `not-allowed` just like the rest of the toolbar items.

## Impact of fix
The fix would make the font-size input match the rest of toolbar items.

bedre7:
## Description

Closes #6697 

### Before

https://github.com/user-attachments/assets/0c1cba0c-0b5d-475c-afbd-4d58aefe78cb

### After

https://github.com/user-attachments/assets/efd1bd53-da3c-4494-8f73-ab364f0001a8

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -1462,6 +1462,7 @@ button.toolbar-item {
   justify-content: space-between;
 }
 
+input.toolbar-item:disabled,
 button.toolbar-item:disabled {
   cursor: not-allowed;
 }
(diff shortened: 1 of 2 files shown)


vantage-ola:
this is actually neat...

PR #6703, opened 2024-10-04
[lexical] Bug Fix: Insertion into inline ElementNode should not crash

### LINKED ISSUE #6701: Bug: Pasting into custom inline ElementNode fails
Lexical version: 0.18.0

## Steps To Reproduce

1. Create a custom ElementNode, which has `isInline() { return true; }`
2. Paste content into the custom ElementNode
3. The paste fails with error message `Cannot use 'in' operator to search for '__language' in null`

And a video of me pressing "ctrl-v" to paste some text into the CustomElementNode: https://youtu.be/XJHWj9TZXIg

Link to code example: https://stackblitz.com/edit/facebook-lexical-urixtg?file=src%2FCustomElementNode.ts

## The current behavior
The paste fails with error message `Cannot use 'in' operator to search for '__language' in null`

## The expected behavior
The content is inserted into the custom ElementNode

## Impact of fix
You cannot paste into inline ElementNodes. At Lumen5, we are using inline ElementNodes to wrap groups of words together to demonstrate relationships between the words and other objects in the DOM. There are millions of users.

etrepum:
## Description

Closes #6701 

### Before

Constructing a document where the children of RootNode are inline ElementNode causes a crash on `selection.insertNodes(…)` (as in paste) because there is an assumption with no runtime invariant that all children of the root are block nodes.

### After

A runtime invariant was added to prevent that, and the incorrect type assertions were removed. A unit test is added to cover this particular case.

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -1732,7 +1732,7 @@ export function INTERNAL_$isBlock(
 export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
   node: LexicalNode,
   predicate: (ancestor: LexicalNode) => ancestor is NodeType,
-) {
+): NodeType | null {
   let parent = node;
   while (parent !== null && parent.getParent() !== null && !predicate(parent)) {
     parent = parent.getParentOrThrow();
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalUtils.ts
```diff
@@ -1732,7 +1732,7 @@ export function INTERNAL_$isBlock(
 export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
   node: LexicalNode,
   predicate: (ancestor: LexicalNode) => ancestor is NodeType,
-) {
+): NodeType | null {
```
etrepum, line 1735: This is just being a bit more explicit rather than relying on inference, usually it's best practice for exported types to be explicit.

ivailop7 (approved):

PR #6704, opened 2024-10-06
[lexical-playground] Bug Fix: Disable editing of Excalidraw Component in Read-Only Mode

neysanfoo:
## Description
Currently, the excalidraw component is still showing the edit modal button/double click to edit functionality when in read-only mode. Another problem is that this part of the code here is causing any read-only editor that uses this plugin to be editing mode.

```
useEffect(() => {
    if (isModalOpen) {
      editor.setEditable(false);
    } else {
      editor.setEditable(true);
    }
  }, [isModalOpen, editor]);
```

I have removed it all together, which means that the editor will not be in read-only mode when the modal is open, but that doesn't seem to cause any problems, and we don't put the editor in read-only mode for any of the other modal based plugins like equation/images.

### Changes in this PR
- Excalidraw modal editor cannot be opened in read-only mode
- Prevent selecting component when in read-only mode
- Clear selection when going from editing to read-only mode
- Event handlers and commands related to editing are not registered or executed in read-only mode
- Use the useLexicalEditor hook for consistency

### Before

https://github.com/user-attachments/assets/86eae3c6-72d6-48fa-bd22-857e2cf9418c

### After

https://github.com/user-attachments/assets/55ecc8eb-c258-4c3b-97e6-135227b7e10e


packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -40,8 +41,9 @@ export default function ExcalidrawComponent({
   height: 'inherit' | number;
 }): JSX.Element {
   const [editor] = useLexicalComposerContext();
+  const isEditable = useLexicalEditable();
   const [isModalOpen, setModalOpen] = useState<boolean>(
-    data === '[]' && editor.isEditable(),
+    data === '[]' && isEditable,
```
etrepum, line 46: This change doesn't do anything, this function is only called on the initial render to initialize the state, so the original code makes it slightly more obvious that it won't automatically track the current isEditable state.

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -66,16 +68,16 @@ export default function ExcalidrawComponent({
...
   useEffect(() => {
-    if (isModalOpen) {
-      editor.setEditable(false);
-    } else {
-      editor.setEditable(true);
+    if (!isEditable && isSelected) {
+      clearSelection();
     }
-  }, [isModalOpen, editor]);
+  }, [isEditable]);
```
etrepum, line 75: This is missing required dependencies, it would probably be better to just move it to the below effect that has all of the correct dependencies.

```ts
if (!isEditable) {
  if (isSelected) {
    clearSelection();
  }
  return;
}
```

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -198,20 +210,22 @@ export default function ExcalidrawComponent({
...
+      {isEditable && isModalOpen && (
+        <ExcalidrawModal
+          initialElements={elements}
+          initialFiles={files}
+          initialAppState={appState}
+          isShown={isModalOpen}
+          onDelete={deleteNode}
+          onClose={closeModal}
+          onSave={(els, aps, fls) => {
+            editor.setEditable(true);
```
etrepum, line 222: This probably shouldn't still be here, since it's not setting editable to false

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -175,8 +185,10 @@ export default function ExcalidrawComponent({
   };
 
   const openModal = useCallback(() => {
-    setModalOpen(true);
-  }, []);
+    if (isEditable) {
+      setModalOpen(true);
+    }
+  }, [isEditable]);
```
etrepum, line 191: This check is unnecessary because the button isn't rendered when it's not editable

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -130,7 +140,7 @@ export default function ExcalidrawComponent({
     aps: Partial<AppState>,
     fls: BinaryFiles,
   ) => {
-    if (!editor.isEditable()) {
+    if (!isEditable) {
       return;
     }
```
neysanfoo, line 145: Would this check be redundant since the modal (and thus onSave callback) only exists when isEditable is true? Thanks for all your feedback @etrepum.

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -176,7 +180,7 @@ export default function ExcalidrawComponent({
 
   const openModal = useCallback(() => {
     setModalOpen(true);
-  }, []);
+  }, [isEditable]);
```
etrepum, line 183:
```suggestion
  }, []);
```

etrepum (approved):

etrepum (approved):

PR #6706, opened 2024-10-06
[lexical-playground] Bug Fix: Disable table hover actions in read-only mode

neysanfoo:
## Description
Currently, the table hover actions allows users to add new rows and columns by hovering over the edges of the table, even when the editor is in read-only mode. Specifically the problems are: hover buttons are visible, interactions with the buttons are allowed, and the event handlers are active all while in read-only mode.

### Changes Introduced in this PR
- The hover buttons for adding rows and columns are hidden when the editor is in read-only mode and users cannot trigger hover actions in read-only mode.
- Disabled event listeners in read-only mode
- The `TableHoverActionsContainer` component returns null when the editor is not editable

### Before

https://github.com/user-attachments/assets/392cd48f-b507-4fc0-afc7-b61ad712cf20

### After

https://github.com/user-attachments/assets/f327a4a2-ccbe-4494-9b1f-5f7efff822d5

### DIFF
--- a/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
@@ -7,6 +7,7 @@
  */
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {useLexicalEditable} from '@lexical/react/useLexicalEditable';
 import {
   $getTableColumnIndexFromTableCellNode,
   $getTableRowIndexFromTableCellNode,
@@ -32,8 +33,9 @@ function TableHoverActionsContainer({
   anchorElem,
 }: {
   anchorElem: HTMLElement;
-}): JSX.Element {
+}): JSX.Element | null {
   const [editor] = useLexicalComposerContext();
+  const isEditable = useLexicalEditable();
   const [isShownRow, setShownRow] = useState<boolean>(false);
   const [isShownColumn, setShownColumn] = useState<boolean>(false);
   const [shouldListenMouseMove, setShouldListenMouseMove] =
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -246,8 +261,12 @@ export default function TableHoverActionsPlugin({
 }: {
   anchorElem?: HTMLElement;
 }): React.ReactPortal | null {
-  return createPortal(
-    <TableHoverActionsContainer anchorElem={anchorElem} />,
-    anchorElem,
-  );
+  const isEditable = useLexicalEditable();
+
+  return isEditable
```
rilrom, line 266: Is there a reason for all the additional `isEditable` checks? Wouldn't the check here suffice?

neysanfoo: Yeah this check is sufficient to prevent rendering the plugin when the editor is not editable, but the additional checks help with defensive programming. For example, when switching from editing to read-only mode, the main check will eventually unmount the component, but there is a brief period before the re-render completes, during this time, the internal checks ensure that not table modifications can occur. Basically preventing race conditions. For example, in a collobrative environment if one person sets the document as read-only, and at the same time another person trys to add a column before the rerendering has occurred.

rilrom: Good explanation, thanks!

etrepum: I don't think the defensive programming is useful in this case, the resolved state of the document over the network should override whatever the local settings are for any specific client.

neysanfoo: yeah i guess i dont really understand how collaborative editors work, I have removed the isEditable checks.

etrepum (approved):

PR #6710, opened 2024-10-06
feat: allow disable flush mutations

zhihengGet:
## Description
i would like to stop flush Mutation when rendering read-only html with lexcial, after render, i wnat to update dom with dom api not with lexical. i.e  user might update bg color, add  stuff to html...etc. 

basically, once rendered, ditch lexical. 

Closes #

### Before

### After

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -620,6 +620,8 @@ export class LexicalEditor {
   /** @internal */
   _editable: boolean;
   /** @internal */
+  _disableFlushMutations: boolean = false;
+  /** @internal */
   _blockCursorElement: null | HTMLDivElement;
 
   /** @internal */
--- a/packages/lexical/src/LexicalMutations.ts
+++ b/packages/lexical/src/LexicalMutations.ts
@@ -11,6 +11,7 @@ import type {LexicalEditor} from './LexicalEditor';
 import type {BaseSelection} from './LexicalSelection';
 
 import {IS_FIREFOX} from 'shared/environment';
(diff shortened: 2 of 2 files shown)


etrepum:
I don't think this is likely to be accepted, but you could clone the DOM tree that lexical renders, unmount lexical, and replace the previous contentEditable with the cloned content

zhihengGet:
> I don't think this is likely to be accepted, but you could clone the DOM tree that lexical renders, unmount lexical, and replace the previous contentEditable with the cloned content

hmm yeah, one problem is that some components are async i.e image, which means i need a way to wait for all components to be rendered then clone , i can use a setTimeout to repeatedly replace & replace but feel like a hacky workaround

and I still think if there is built-in way to 'eject' or detach lexical, that would be great. Thanks

etrepum:
If you set it as editable false and don't make updates to it then it won't ever flush mutations, you're free to do whatever you want to the DOM if you don't need to ever update it. Lexical only touches DOM for nodes that changed, and it only touches the selection when it's non-null and editable is true.

zhihengGet:
ahh, i solved! i can just do `editor.observer.disconnect() ` and `editor.setEditable=false` xd, thanks for th inspiration. 


hear what i try to accomplish when user click a button 

disable editable won't stop the mutation observer

```
  function updater() {
    const dom = document.body.getElementsByTagName('p');
    const [editor] = composerContext;
    editor.setEditable(false);
    for (let x of dom) {
      const d = document.createElement('span');
      d.innerText = 'aaaa';
      x.append(d).
    }
  }

```

PR #6713, opened 2024-10-07
[lexical][lexical-react][lexical-table] Fix: Add horizontal scroll for tables

### LINKED ISSUE #6480: Feature: Scroll for table
## Description

It would be great if the table had its own scroll bar. Right now, the scroll affects the entire document, which is not very convenient.
The video shows an example of how this is implemented in other projects.

https://github.com/user-attachments/assets/00519b81-a211-4d6f-bfe8-824a99980b9a

GermanJablo:
## Description

Closes #6480

Having tables overflow is a mistake, as their content can conflict with other elements of the interface:

image

Another incorrect behavior is the one that occurs in the Lexical Playground, where if the table is too large, the entire editor becomes larger. The horizontal scroll bar is not placed over the table, but over the entire content.

image

A super quick way to fix overflow is to add a max-width of 100% to the table.

The problem is, [as this user noted](https://github.com/facebook/lexical/issues/6480#issuecomment-2277817068), that this creates conflicts with the column width property.

This is what OneNote does and I've always hated it. Sometimes columns resize in a magical or hard-to-understand way. It's better to give the user the flexibility to adjust column widths to their liking and to have overflow if needed.

The correct solution to this is for tables to have an overflow scroll independent of the rest of the editor, similar to what Notion does:

image

The problem is that the `<table>` tag doesn't support overflow: scroll. A div is needed to wrap it.

This PR introduces a new `Scrollable` node that automatically wraps tables.

**BREAKING CHANGE**: `TablePlugin` now requires the editor to also register the `ScrollableNode` node. All tables are automatically wrapped in a `ScrollableNode`.

### Before

image

### After

image


packages/lexical-table/src/LexicalScrollableNode.ts
```diff
@@ -0,0 +1,125 @@
...
+} from 'lexical';
+
+import {$isTableNode} from './LexicalTableNode';
+
+/**
+ * This node allows the nodes it wraps to have a width greater than that of the editor,
+ * encapsulating only its content in a horizontal scroll. You'll want to use it to wrap
+ * nodes that have a "width" property and are NOT decoratorNodes. See TableNode as an example.
+ */
+export class ScrollableNode extends ElementNode {
```
GermanJablo, line 31: At first I wrapped `TableNode` and `ImageNode`, and so I put it in core. Then I realized that since it's a decoratorNode in `ImageNode` you can replicate the same behavior much easier without using this node. So I moved the node to `lexical-table`.
Wrapping another node with `ScrollableNode` should be easy if you want to do it in the future. At the moment `TableNode` is the only one that has a `width` property and it's not `DecoratorNode`.

packages/lexical-table/src/LexicalScrollableNode.ts
```diff
@@ -0,0 +1,125 @@
...
+    return 'scrollable';
+  }
+
+  static clone(node: ScrollableNode): ScrollableNode {
+    return new ScrollableNode(node.__key);
+  }
+
+  createDOM(_config: EditorConfig, _editor: LexicalEditor): HTMLElement {
+    const dom = document.createElement('div');
+    dom.classList.add('lexical-scrollable');
```
GermanJablo, line 42: Required for exportDOM and importDOM. I've included new tests to verify that it works correctly.

packages/lexical-table/src/LexicalScrollableNode.ts
etrepum: I think it would be tricky to use this class as-is with other element types because of the transform that requires the child to be a table

packages/lexical-table/src/LexicalScrollableNode.ts
```diff
@@ -0,0 +1,125 @@
...
+      if (firstChild && !$isTableNode(firstChild)) {
+        firstChild.remove();
+      }
+      if (
+        $isRangeSelection(selection) &&
+        selection.anchor.key === node.getKey()
+      ) {
+        firstChild.getFirstDescendant()!.selectStart();
+      }
+    };
```
etrepum, line 109: It's not entirely clear what this is defending against but it doesn't look like it's doing it correctly since the last branch can execute when firstChild is not a table (might not have a first descendant) and is not connected to the root. I would put the child check before the empty check, remove all children except the first that passes the table check, and then if there's no table node found then remove the node (and fix up the root if necessary).

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1389,18 +1396,47 @@ function $handleArrowKey(
...
+          : siblingNode0;
         if (!siblingNode || !$isTableNode(siblingNode)) {
           return false;
         }
         stopEvent(event);
-        siblingNode.selectEnd();
+        if (direction === 'backward') {
+          siblingNode.selectEnd();
+        } else {
+          const firstDescendant = siblingNode.getFirstDescendant()!;
```
etrepum, line 1413: it would be better to use a runtime invariant rather than a non-null assertion here, there isn't a runtime check that this is non-empty

etrepum (commented):
I don't have time for QA right now or to audit the tests but I took a quick look at the code. Seems like a useful improvement! The downside is that it is a bit backwards incompatible due to the additional node and requirements for table use. Would maybe be a bit nicer if it was done in a way that could be backwards compatible. @ivailop7 should have a look

packages/lexical-table/src/LexicalScrollableNode.ts
GermanJablo: Yes, that's true. At first I thought that it would only be necessary to copy the transform that the table has.
The problem is all the corrections I had to make to the selection in `LexicalTableSelectionHelpers.ts`

ivailop7:
Will give this a read this evening!

ivailop7:
Agreed, sorry, it's a big one. Ideally if this was an opt-in setting. Would have been a nobrainer for me, but it's a significant paradigm change. Want to understand the scope, a bit better if we can make it opt-inable.

GermanJablo:
I would love for it to be opt-in as well, but it would be necessary to duplicate a lot of tests and maintain the two paradigms that are indeed quite different.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1156,6 +1156,16 @@ export class RangeSelection implements BaseSelection {
       fixText(firstNode, del);
       return;
     }
+    if ($getRoot().getChildrenSize() > 1) {
```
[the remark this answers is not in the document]
GermanJablo: Without this, there were one or two e2e tests where the output ended with an extra empty paragraph. I think it was if the selection spanned from the beginning of the table to content outside the table.

GermanJablo:
> I believe we can have a toggle (default = false) on the scrollable node-level that if on, wouldn't apply the 'max-width of 100%' and we would be able to keep the existing behaviour if others rely on it. Thoughts?

The problem with the breaking change is in the introduction of a new node, as there can be logic that ignores its existence. Not with its max-width of 100%. Considering that the new node is going to be there for everyone, I can't think of any reason why anyone would want to give it a different behavior.

etrepum:
We discussed this PR in the weekly meeting today and it seems very promising and overall the right direction given the current constraints in the reconciler. However, the backwards compatibility issue makes it very tricky to accept as-is because it would likely cause some problems with existing deployments (particularly those using yjs, rolling out with multiple concurrent versions of lexical, etc.). It would be ideal if there was a way to preserve the possibility to deploy table nodes as-is without the ScrollableNode wrapper and then gracefully upgrade somehow to include the wrapper.

We think it will be possible to make the ScrollableNode opt-in by moving the transforms from static methods to separately exported transform functions that need to be registered explicitly with the editor (this could be abstracted with a register function that does a mergeRegister of all necessary transforms and a React plugin to make it easier to install). This would also allow future usage of ScrollableNode with other types of nodes as the logic would be in a function that can be replaced (e.g. by parameterizing the `$isTableNode` predicate).

It's acceptable that the tests will expect that this plugin is already registered so that work wouldn't have to be undone, but we'd probably want to add some minimal tests to show that some table functionality still works without the opt-in ScrollableNode.

Overall, if we can get this into an opt-in state to make the upgrade path smooth then we are onboard with merging this quickly!

Maybe something along these lines (with `$normalizePoint__EXPERIMENTAL` exported from LexicalNormalization like `$normalizeSelection` is?)

```ts
export interface ScrollableNodeConfig {
  scrollableChildNodes: readonly Klass<ElementNode>[];
  $isScrollableChild?: (node: LexicalNode | null) => boolean;
}

export function registerScrollableNodeTransform(
  editor: LexicalEditor,
  config: ScrollableNodeConfig,
): () => void {
  invariant(
    editor.hasNodes([ScrollableNode]),
    'TablePlugin: ScrollableNode not registered on editor',
  );
  const {
    scrollableChildNodes,
    $isScrollableChild = (node) =>
      node !== null &&
      scrollableChildNodes.some((klass) => node instanceof klass),
  } = config;
  return mergeRegister(
    editor.registerNodeTransform(ScrollableNode, (node) => {
      let onlyScrollableChild: LexicalNode | null = null;
      for (const child of node.getChildren()) {
        if (onlyScrollableChild === null && $isScrollableChild(child)) {
          onlyScrollableChild = child;
        } else {
          child.remove();
        }
      }
      if (onlyScrollableChild === null) {
        node.remove();
        const root = $getRoot();
        if (root.isEmpty()) {
          root.append($createParagraphNode());
        }
        return;
      }
      if (!selection) {
        return;
      }
      const nodeKey = node.getKey();
      for (const point of selection.getStartEndPoints() || []) {
        if (point.key === nodeKey) {
          $normalizePoint__EXPERIMENTAL(point);
        }
      }
    }),
    ...scrollableChildNodes.map((klass) =>
      editor.registerNodeTransform(klass, (node) => {
        if (!$isScrollableNode(parent)) {
          const scrollable = $createScrollableNode();
          node.insertBefore(scrollable);
          scrollable.append(node);
        }
      }),
    ),
  );
}
```

etrepum:
I added a commit here to show what I am proposing https://github.com/facebook/lexical/pull/6743 https://github.com/facebook/lexical/pull/6743/commits/42bf4f8315edcf06460d1ad45b8a64797f59a1ce

GermanJablo:
Thanks for the review @etrepum. Everything you say makes sense.

The biggest difficulty will be not in moving the transforms to a plugin, but in maintaining the 2 versions of `LexicalTableSelectionHelpers.ts`. I'll see if I can find a viable way to achieve this.

etrepum:
Maybe the best way to do that would be to instead override TableNode.selectNext, TableNode.selectPrevious, TableNode.insertBefore, etc. to check the parent and delegate if necessary?

I guess the insertBefore/insertAfter stuff could maybe be fixed in a more robust way by having the ScrollableNode transform move non-scrollable elements before or after itself instead of removing them

GermanJablo:
Hmm, you would also need to extend `getNextSibling`, `getPreviousSibling`, `getParent`, `getParentOrThrow`.

Also, that wouldn't solve every scenario. For example if a paragraph does `getNextSibling` and it's a scrollableNode?

etrepum:
I don't think that this code does any better at solving all of those scenarios, and I don't think that you would want the `get*` methods to lie to you. Updating the `select*` methods are just a form of preemptive normalization of the selection which is useful because necessary information is lost. You don't know whether the selection point should be left or right biased by the time that `selectNext` or `selectPrevious` returns.

GermanJablo:
I cherry picked your commit, and modified `LexicalTableSelectionHelpers` to modify its behavior depending on whether ScrollableNode is registered or not (the last commit so far).

The next step would be to retrieve the initial tests for cases where it is not registered. But I'm not sure what the best way to do that is.

I've made a PoC to make its registration configurable in the bottom left options menu in the playground. The commit is in another branch: https://github.com/facebook/lexical/commit/c7888966d53f564a4c1d92a130d371ff8c9fbae6

I see that with richtext/plaintext this + environment variables are used. But I don't know if I'm making it more complicated than necessary. In this case, it is only necessary to run a subset of tests with both configurations.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -76,6 +78,12 @@ import {$computeTableMap, $getNodeTriplet} from './LexicalTableUtils';
 
 const LEXICAL_ELEMENT_KEY = '__lexicalTableSelection';
 
+function $getScrollableOrTable(tableNode: TableNode) {
+  return $getEditor().hasNode(ScrollableNode)
+    ? (tableNode.getParentOrThrow() as ScrollableNode)
+    : tableNode;
+}
```
etrepum, line 85: It might be cleaner to put an $isScrollableNode check on the parent instead of using the editor config, that way you could toggle the behavior using only the plugin.

packages/lexical-table/src/LexicalScrollableNode.ts
```diff
@@ -0,0 +1,156 @@
...
+      let onlyScrollableChild: LexicalNode | null = null;
+      for (const child of node.getChildren()) {
+        if (onlyScrollableChild === null && $isScrollableChild(child)) {
+          onlyScrollableChild = child;
+        } else {
+          child.remove();
+        }
+      }
+      if (onlyScrollableChild === null) {
+        node.remove();
```
etrepum, line 123: based on the other discussions I think that these children should be moved out of the scrollable node either before or after the scrollable node depending on whether there's a child or not

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1790,8 +1835,8 @@ function $getExitingToNode(
   return anchorSibling && $isTableNode(anchorSibling)
     ? anchorSibling
     : direction === 'backward'
-    ? tableNode.getPreviousSibling()
-    : tableNode.getNextSibling();
+    ? tableNode.getParentOrThrow().getPreviousSibling()
+    : tableNode.getParentOrThrow().getNextSibling();
```
etrepum, line 1839:
```suggestion
    ? $getScrollableOrTable(tableNode).getPreviousSibling()
    : $getScrollableOrTable(tableNode).getNextSibling();
```

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -58,7 +58,7 @@ export function TablePlugin({
     if (!editor.hasNodes([TableNode, TableCellNode, TableRowNode])) {
       invariant(
         false,
-        'TablePlugin: TableNode, TableCellNode or TableRowNode not registered on editor',
+        'TablePlugin: TableNode, TableCellNode, TableRowNode or ScrollableNode not registered on editor',
```
etrepum, line 61:
```suggestion
        'TablePlugin: TableNode, TableCellNode or TableRowNode not registered on editor',
```

etrepum (commented):
This looks like a good approach to me. I'm not personally that concerned about having another toggle in the playground settings for separate e2e testing of scrollable and non-scrollable tables. I think a small suite of unit tests that build an editor without the scrollable plugin would be fine to demonstrate that things still work without it.

GermanJablo:
Sorry for the delay on this. I'm getting back to this PR and will try to finish it ASAP.

etrepum:
I did mock up an implementation based on this in https://github.com/facebook/lexical/pull/6759/commits/351dc2f9b72a447bfa173138c09865dd635bfacd (#6759) using a new proposed experimental getDOMSlot API which would allow the TableNode to control where its children are rendered. I didn't write any tests but it seemed to behave in the playground when I was testing it locally. Requires a setting to be enabled https://lexical-playground-irlbe8n5m-fbopensource.vercel.app/?tableHorizontalScroll=true

GermanJablo:
That looks super cool! I'm going to pause this PR until we see if `DOMSlot` works!

Issue #5170, opened 2023-10-26
Bug: `ListMaxIndentLevelPlugin` allows indent beyond `maxDepth` on paste


nooblyf:
The playground has a max depth of 7 for list indents but you can bypass it by pasting the indents as shown in the screen recording

https://github.com/facebook/lexical/assets/54323745/594036f3-5ffb-43ea-8a84-3bc72bd49ad7

Lexical version: `0.12.2`

## Steps To Reproduce

1. Go to https://playground.lexical.dev/ it has indent maxDepth of 7
2. Paste indented list having depth more than 7, you'll be able to paste it successfully

## The current behavior
maxDepth for `ListMaxIndentLevelPlugin` only stops directly indenting lists in the editor

## The expected behavior
Even when pasting indented content it should still adhere to maxDepth by the editor, it should remove those extra indents.

mdmuhtasimfuadfahim:
Hi @etrepum. I solved the issue. Please have a look [#6718](https://github.com/facebook/lexical/pull/6718)

PR #6718, opened 2024-10-09
[lexical-playground] Fix: `ListMaxIndentLevelPlugin` allows indent beyond maxDepth on paste 

mdmuhtasimfuadfahim:
## Description

The `ListMaxIndentLevelPlugin` restricts the maximum indent level for list items based on the defined `maxDepth`. However, when pasting content with a deeper indent, the plugin doesn't automatically enforce this restriction. This allows the pasted content to exceed the allowed indent level, resulting in a higher indent depth than the set `maxDepth`.

*Added logic to restrict paste behaviour for lists in `ListMaxIndentLevelPlugin`, ensuring that pasted content respects the defined `maxDepth` and prevents exceeding the allowed indent level.*

Closes #[6609](https://github.com/facebook/lexical/pull/6609)

### Before

[maxDepth issue](https://d.pr/v/IRF0rH)

### After

[solved maxDepth issue](https://d.pr/v/zw3uMQ)

Note: The solution restricts pasting beyond indent level 7, ensuring the list structure stays within the defined limit.

### DIFF
--- a/packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
+++ b/packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
@@ -16,6 +16,7 @@ import {
   $isRangeSelection,
   COMMAND_PRIORITY_CRITICAL,
   INDENT_CONTENT_COMMAND,
+  PASTE_COMMAND,
 } from 'lexical';
 import {useEffect} from 'react';
 
@@ -77,6 +78,14 @@ export default function ListMaxIndentLevelPlugin({
   useEffect(() => {
     return editor.registerCommand(
       INDENT_CONTENT_COMMAND,
+      () => $shouldPreventIndent(maxDepth),
+      COMMAND_PRIORITY_CRITICAL,
+    );
+  }, [editor, maxDepth]);
+
+  useEffect(() => {
+    return editor.registerCommand(
+      PASTE_COMMAND,
       () => $shouldPreventIndent(maxDepth),
       COMMAND_PRIORITY_CRITICAL,
     );


packages/lexical-playground/src/plugins/ListMaxIndentLevelPlugin/index.ts
```diff
@@ -81,5 +82,13 @@ export default function ListMaxIndentLevelPlugin({
       COMMAND_PRIORITY_CRITICAL,
     );
   }, [editor, maxDepth]);
+
+  useEffect(() => {
```
ivailop7, line 86: you can use mergeRegister to avoid a second useEffect, example usage in other part of the code: 
https://github.com/facebook/lexical/blob/d273333d26393c001cc4e2e1d8667f79ce78b86b/packages/lexical-react/src/LexicalAutoLinkPlugin.ts#L467-L509

ivailop7:
@mdmuhtasimfuadfahim don't forget to do CLA as well. Thanks!

mdmuhtasimfuadfahim:
> **[ivailop7](/ivailop7) ** c

Hi @ivailop7. I have updated the code. Please check and signed here `https://code.facebook.com/cla` with the same email I did the commits. I guess it will take 1 hour. Will it be okay for the Facebook CLA Check pass? If not. Please let me know what I should do for it. Thanks in advance.

mdmuhtasimfuadfahim:
Hi @ivailop7. Facebook CLA Check is done now. Please check the code.

ivailop7 (approved):

etrepum (commented):
I think what this does is prevent pasting at all when the indent is at the limit (even if the paste is just plain text), it doesn't prevent the problem from happening if the pasted content contains multiple levels of indent

mdmuhtasimfuadfahim:
> I think what this does is prevent pasting at all when the indent is at the limit (even if the paste is just plain text), it doesn't prevent the problem from happening if the pasted content contains multiple levels of indent

Hi @etrepum. If I don't prevent the pasting when the limit hits it generates the issue [#5170][(url)](https://github.com/facebook/lexical/issues/5170). Well let me check if I can allow the pasting for text but not to increase the indent level.

etrepum:
Yes, I understand that this prevents one situation where this issue happens, but I don't think it's a complete fix because pasted content can have embedded indentation. I also think it also prevents certain pastes that are expected to work. Possibly the only way to make this solution complete without unintended consequences would be through node transforms and not just commands.

etrepum:

PR #6719, opened 2024-10-10
[lexical] Bug Fix: lines were being deleted with `deleteLine`

taro-shono:
## Description

When I pressed ⌘ + Delete, the line was deleted. This is different from the behavior of other editors (such as textarea), so some users may find it to be a bug. Please see the video for more information about the behavior.

### Before

https://github.com/user-attachments/assets/c03a4503-4ab6-4639-afbc-3ec17d2089ce

### After

https://github.com/user-attachments/assets/34cd5c7b-5865-4d33-97fa-f9dd89679cc5


etrepum (commented):
It would be great if this change had a test, and it is necessary for all PRs to pass the formatting and type checks. This particular formatting issue (extra indent on all changed lines) can be fixed with `npm run prettier:fix` or by allowing husky to get set up properly which normally happens during `npm install`.

I think the logic might also not be correct, it seems to me that this extend behavior only happens in other editors (I tested VSCode and Chrome's textarea on github) when the selection is collapsed at offset 0.

taro-shono:
> I think the logic might also not be correct, it seems to me that this extend behavior only happens in other editors (I tested VSCode and Chrome's textarea on github) when the selection is collapsed at offset 0.

@etrepum What does that mean? Please explain in more detail.

etrepum:
A collapsed selection is where the focus and anchor are the same node and offset. It displays as a cursor, no text is highlighted.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1856,11 +1856,11 @@ export class RangeSelection implements BaseSelection {
...
-      // If selection is extended to cover text edge then extend it one character more
-      // to delete its parent element. Otherwise text content will be deleted but empty
-      // parent node will remain
-      const endPoint = isBackward ? this.focus : this.anchor;
-      if (endPoint.offset === 0) {
+      // If the selection starts at the beginning of a text node (offset 0),
+      // extend the selection by one character in the specified direction.
+      // This ensures that the parent element is deleted along with its content.
+      // Otherwise, only the text content will be deleted, leaving an empty parent node.
+      if (this.anchor.offset === 0) {
```
etrepum, line 1863: By collapsed I mean that this check should be more like this:

```suggestion
      if (this.isCollapsed() && this.anchor.offset === 0) {
```

taro-shono: @etrepum The isCollapsed check had already been made.

https://github.com/facebook/lexical/blob/787b3273023834d4aea44ccdee82405410a8ef8a/packages/lexical/src/LexicalSelection.ts#L1848

etrepum: You're right that a check was already made but the selection is mutated after the check is made with ` this.modify('extend', isBackward, 'lineboundary');` and an `extend` modification (if it does anything) will only change either the anchor or focus thus changing the selection so it's no longer collapsed

taro-shono: I see. I've also added a test for the logic of collapsed: [c26567c](https://github.com/facebook/lexical/pull/6719/commits/c26567c6e71a404902590f122703c43d041ee251)

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -270,6 +272,34 @@ test.describe.parallel('Selection', () => {
     );
   });
 
+  test('can delete line by collapse', async ({page, isPlainText}) => {
+    test.skip(isPlainText || !IS_MAC);
```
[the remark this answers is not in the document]
taro-shono: I think so too. But I don't know why this line: 
https://github.com/facebook/lexical/blob/c27f28194d485651e82537fa7c173840837f68ac/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs#L170
https://github.com/facebook/lexical/blob/c27f28194d485651e82537fa7c173840837f68ac/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs#L231

The above test is the same as the deleteLine test.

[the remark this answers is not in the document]
etrepum: They're not useful there either! All of them pass without it for me locally on my Mac. We can address that separately though

PR #6724, opened 2024-10-12
fix: preserve custom fields in Lexical-Yjs sync

collierrgbsitisfise:
When syncing Lexical selection to Yjs, custom fields in the local state were being dropped. This commit ensures all custom fields are preserved during the sync process, maintaining data integrity in collaborative editing scenarios.

Updated syncLexicalSelectionToYjs to include ...customFields Ensures backwards compatibility with existing implementations Improves reliability of custom data persistence in collaboration

prev [PR](https://github.com/facebook/lexical/pull/6683) where ProviderAwareness  type was extended with `setLocalStateField `

### DIFF
--- a/packages/lexical-yjs/src/SyncCursors.ts
+++ b/packages/lexical-yjs/src/SyncCursors.ts
@@ -502,6 +502,7 @@ export function syncLexicalSelectionToYjs(
     color,
     focusing,
     awarenessData,
+    ...customFields
   } = localState;
   let anchorPos = null;
   let focusPos = null;
@@ -531,6 +532,7 @@ export function syncLexicalSelectionToYjs(
       focusPos,
       focusing,
       name,
+      ...customFields,
     });
   }
 }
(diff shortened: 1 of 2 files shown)


packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -531,6 +532,7 @@ export function syncLexicalSelectionToYjs(
       focusPos,
       focusing,
       name,
+      ...customFields,
```
etrepum, line 535: It would be a bit more efficient to use `localState` instead, the `customFields` variable isn't needed:

```ts
    awareness.setLocalState({
      ...localState,
      anchorPos,
      awarenessData,
      color,
      focusPos,
      focusing,
      name,
    });
```

etrepum (dismissed):

packages/lexical-yjs/src/SyncCursors.ts
collierrgbsitisfise: @etrepum make senes, fixed!

etrepum (approved):

PR #6725, opened 2024-10-12
[lexical-playground] Table Hover Actions Layout Fixes

ivailop7:
- Fix up the anchoring logic for the relative add row/column buttons (Fixes for when the editor is an embedded widget)
- Fix the overlap of text entry when the buttons are displayed (videos below)

Before

https://github.com/user-attachments/assets/e98d44fd-39a0-4d8a-ad64-b7bbb2618842

After

https://github.com/user-attachments/assets/336bc756-ab66-449c-b60d-b919e1bb9176

### DIFF
--- a/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
@@ -39,8 +39,8 @@ function TableHoverActionsContainer({
   const [shouldListenMouseMove, setShouldListenMouseMove] =
     useState<boolean>(false);
   const [position, setPosition] = useState({});
-  const codeSetRef = useRef<Set<NodeKey>>(new Set());
-  const tableDOMNodeRef = useRef<HTMLElement | null>(null);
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -131,6 +132,13 @@ function TableHoverActionsContainer({
     250,
   );
 
+  // Hide the buttons on any table dimensions change to prevent last row cells
+  // overlap behind the 'Add Row' button when text entry changes cell height
+  const tableResizeObserver = new ResizeObserver((entries) => {
+    setShownRow(false);
+    setShownColumn(false);
+  });
```
etrepum, line 140: This violates the rules of react. Should be done with a useMemo, useState or useRef and used as a dependency in the effects (if not a ref).

ivailop7: yep, amended

etrepum (approved):

PR #6739, opened 2024-10-16
[lexical] Chore: Enable all selection tests on Mac

etrepum:
## Description

Some selection tests were disabled on Mac, possibly for some legacy reason (maybe before the meta/ctrl key press in tests was abstracted?), but they all work because we have keybindings for each of the actions they are performing.

### Before

Fewer selection tests ran on Mac

### After

All selection tests run on Mac


ivailop7 (approved):

etrepum:
The logic here was inverted, the tests are skipped when NOT on mac. I don't have an easy way to test that so I am abandoning this PR.

PR #6741, opened 2024-10-16
[lexical-table] Return inserted node from `$insertTableRow__EXPERIMENTAL` and `$insertTableColumn__EXPERIMENTAL`

amanharwara:
## Description

Currently, the `$insertTableRow__EXPERIMENTAL` function doesn't return anything, so if you want to insert a table row using it and do something with the inserted row, you'll have to do all the calculations again to find the inserted row.
With this PR, the function will return the inserted row node if it was successful and undefined if not. This removes the need for rewriting all the logic regarding spans just to find the inserted row.
Also returns the first inserted cell from `$insertTableColumn__EXPERIMENTAL`

### DIFF
--- a/packages/lexical-table/flow/LexicalTable.js.flow
+++ b/packages/lexical-table/flow/LexicalTable.js.flow
@@ -237,11 +237,11 @@ declare export function $deleteTableColumn(
 
 declare export function $insertTableRow__EXPERIMENTAL(
   insertAfter: boolean,
-): TableRowNode | void;
+): TableRowNode | null;
 
 declare export function $insertTableColumn__EXPERIMENTAL(
   insertAfter: boolean,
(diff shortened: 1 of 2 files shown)


ivailop7:
Would this actually work the way you intend to use it though? When you insert the row, the updated state needs to be commited, before using it, so you would need to do another editor.update(), at which point you could now get the just appended node key, etc after the committed state. Maybe @etrepum could confirm if I'm right or wrong here.

etrepum (approved):
I think it would be perfectly reasonable to do something with the result, e.g. populate some default content into the cells. The update commit concern is only relevant if you're trying to do something with the DOM after it's rendered.

ivailop7 (approved):

PR #6746, opened 2024-10-20
[lexical-table] [lexical-selection] Try to fix calling split on undefined

ivailop7:
Second highest error rate leading to editor crashes we are observing is the following error:
"Cannot read properties of undefined (reading "split")..."
Unfortunately, I don't know how to reproduce the error.

Have gone through the whole codebase and everywhere we call 'split' and have narrowed it down to 3 places (insertRawText method + the 2 in this PR), my gut feeling is that it's probably the table cell method, but the other 2 also could do with the defensive if statement.

EDIT: Will start only with the table plugin and CSS parsing edits

### DIFF
--- a/packages/lexical-selection/src/utils.ts
+++ b/packages/lexical-selection/src/utils.ts
@@ -176,6 +176,9 @@ export function createRectsFromDOMRange(
  */
 export function getStyleObjectFromRawCSS(css: string): Record<string, string> {
   const styleObject: Record<string, string> = {};
+  if (!css) {
+    return styleObject;
+  }
   const styles = css.split(';');
 
   for (const style of styles) {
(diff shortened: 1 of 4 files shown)


etrepum (commented):
I don't love adding defensive code that ignores input when the types are wrong, it's just hiding some incorrect cast somewhere else if these are anywhere near the actual problem. Maybe a better option would be to add a typeof string invariant to getStyleObjectFromRawCSS if that's a suspect and then you'd at least know that's where the failure is?

The textDecoration case seems awfully unlikely because that looks like a DOM node and those shouldn't have these kinds of surprises

potatowagon (approved):

ivailop7:
> I don't love adding defensive code that ignores input when the types are wrong, it's just hiding some incorrect cast somewhere else if these are anywhere near the actual problem. Maybe a better option would be to add a typeof string invariant to getStyleObjectFromRawCSS if that's a suspect and then you'd at least know that's where the failure is?
> 
> The textDecoration case seems awfully unlikely because that looks like a DOM node and those shouldn't have these kinds of surprises

Agreed, but I think for these 2 the changes make sense.

PR #6747, opened 2024-10-20
[lexical-selection] Fix __language in null

ivailop7:
5th highest error we are observing is: 'Cannot use 'in' operator to search for __language in null'
Not being able to reproduce, but confident the below should address this error in particular, since it seems like the only place in the code base we call '__language'

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1360,7 +1360,7 @@ export class RangeSelection implements BaseSelection {
 
     // CASE 1: insert inside a code block
     if ($isElementNode(firstBlock) && '__language' in firstBlock) {
-      if ('__language' in nodes[0]) {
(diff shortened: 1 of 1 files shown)


etrepum:
https://github.com/facebook/lexical/issues/6701 has a repro, this will certainly fix this exception but I think another one might happen because other invariants are being violated when this occurs https://github.com/facebook/lexical/issues/6701#issuecomment-2394768062

etrepum (approved):
Approving because it's an incremental improvement but I think this may just raise an exception elsewhere based on previous comment

ivailop7:
I'll rebase to latest nightly and see where things are at, before merging this one. Will keep open for a week/two.

ivailop7:
After rebasing to latest it's been 7 days without observing the error. This won't be needed it appears. Closing.

PR #6754, opened 2024-10-22
[lexical-playground] Feature: Add more keyboard shortcuts

### LINKED ISSUE #6734: Feature: Add keyboard shortcuts so the editor is much more accessible
More many things there are no keyboard shortcuts provided out of the box.
This is very bad escpacially for screenreader / only keyboard users, because it is very annoying and time consuming to go throu each menu just to make someting a headline or insert a list.

I would recommend to use the same keys as word where possible, many users are already familiar with them.

bedre7:
## Description

Closes #6734 

This PR adds more keyboard shortcuts to the Lexical Playground with a new plugin called `ShortcutsPlugin` to improve keyboard accessibility and user experience.

## Changes Made
- [X] Added `ShortcutsPlugin` to handle all keyboard shortcuts in one place and make adding more shortcuts in the future easier. (This doesn't include some existing shortcuts like, **bold** and **italic** which are implemented at the lower level.)
- [X] Created a new `ToolbarContext` to manage the toolbar state which is shared between the `ToolbarPlugin` and the `ShortcutsPlugin`
- [X] Moved common functions from `ToolbarPlugin.tsx` to a new file `utils.ts` to be shared between the `ToolbarPlugin` and the `ShortcutsPlugin`.
- [X] Moved existing Insert Link shortcut listener from the `ToolbarPlugin` to the `ShortcutsPlugin`.
- [X] Added title attribute and shortcut labels to the toolbar items to show the newly added keyboard shortcuts.
- [X] Added end-to-end tests for the new keyboard shortcuts.

## Choices Made

The issue was intended to add the keyboard shortcuts from MS Word, however, certain key combinations are reserved for browser usage and cannot be intercepted by client-side JavaScript, even if `event.preventDefault()` is called. (Ref: [Chromium Issue 41081444](https://bugs.chromium.org/p/chromium/issues/detail?id=41081444)). This is why we have chosen to use key bindings from well-known applications like Google Docs, Tiptap, and Slack, which are web-native applications. This choice ensures that users familiar with these applications will find the shortcuts intuitive and easy to use. Ultimately, the decision on the key bindings is up to the maintainers.

## Note
1. Some shortcuts might not work for all languages or keyboards.
2. The shortcuts have been added to the Playground only, so users can modify them as needed for their own applications.

## Test plan

End-to-end tests have been added for the new keyboard shortcuts.

## Acknowledgement
@etrepum for his guidance and letting me work on this feature.
@BennyAlex for creating this useful feature request.


etrepum (dismissed):
The code here looks great, very straightforward and easy to read through. I like the extensive tests and other refactoring here. I think this would provide a great example for someone who wanted to implement a full suite of keyboard shortcuts for their own editor.

packages/lexical-playground/src/plugins/ShortcutsPlugin/shortcuts.ts
```diff
@@ -0,0 +1,318 @@
...
+  // (Ctrl|⌘) + Alt + <key> shortcuts
+  NORMAL: IS_APPLE ? '⌘+Alt+0' : 'Ctrl+Alt+0',
+  HEADING1: IS_APPLE ? '⌘+Alt+1' : 'Ctrl+Alt+1',
+  HEADING2: IS_APPLE ? '⌘+Alt+2' : 'Ctrl+Alt+2',
+  HEADING3: IS_APPLE ? '⌘+Alt+3' : 'Ctrl+Alt+3',
+  BULLET_LIST: IS_APPLE ? '⌘+Alt+4' : 'Ctrl+Alt+4',
+  NUMBERED_LIST: IS_APPLE ? '⌘+Alt+5' : 'Ctrl+Alt+5',
+  CHECK_LIST: IS_APPLE ? '⌘+Alt+6' : 'Ctrl+Alt+6',
+  CODE_BLOCK: IS_APPLE ? '⌘+Alt+C' : 'Ctrl+Alt+C',
+  QUOTE: IS_APPLE ? '⌘+Alt+Q' : 'Ctrl+Alt+Q',
```
etrepum, line 23: These aren't labeled correctly for Mac, they should be Opt or ⌥ instead of Alt. There's no Alt key on Mac. You get Shift (⇧), Fn, Control (⌃), Option (⌥) and Command (⌘). They also unfortunately don't work as implemented because at least on a US keyboard layout the key changes due to option being pressed.

```
0 = º
1 = ¡
2 = ™
3 = £
```

Probably `KeyboardEvent.code` would be appropriate here (`'Digit0'`, `'Digit1'`, etc.)

bedre7: Thanks for the feedback! Oh my bad, I didn't realize that. Let's not merge this just yet in that case. I will make sure to fix them as you suggested. Thanks again for pointing this out.

etrepum: I would recommend refactoring it so the predicate functions just take a KeyboardEvent and you can destructure that for the checks, instead of destructuring it first and having to pass all of the parameters separately (this can also cause some hard to find bugs since many of them are the same type, so correctness is argument order dependent). This would also make it easier for a predicate like this one to check KeyboardEvent.code which is not currently passed to these functions.

bedre7: That makes sense, thanks!

etrepum (dismissed):
This looks and works great on Mac now, although there is a conflict that needs to be resolved before this can be merged

bedre7:
> This looks and works great on Mac now, although there is a conflict that needs to be resolved before this can be merged

The conflict has been resolved.

etrepum (approved):

ivailop7 (approved):
That's quite the refactor

PR #6755, opened 2024-10-22
Trust HTML for rich text clipboard

zurfyx:
Modern sites enforce trusted HTML. `DOMParse.parseFromString` will fail because the clipboard is not a trusted type. Given that we already post-process the HTML instead of blindly exposing it to the DOM I think it's safe to trust it here.


fantactuka (approved):
🚀

etrepum (approved):

PR #6587, opened 2024-09-03
[lexical] Feature:  customElements judgment when destroyNode

yaojiu19:
## Description
I define customElement, it include multiple elements，like:
```typescript
class FTable extends HTMLElement {
  sectionEle: HTMLDivElement
  tableEle: HTMLTableElement
  tableBodyEle: HTMLTableSectionElement
  constructor() {
    super()
    this.sectionEle = document.createElement('div')
    this.sectionEle.className = 'c-editor-doc-table-section'
    this.tableEle = document.createElement('table')
    this.tableEle.className = 'c-editor-doc-table'
    this.tableBodyEle = document.createElement('tbody')
    this.tableEle.append(this.tableBodyEle)
    this.sectionEle.append(this.tableEle)
  }
  connectedCallback() {
    this.append(this.sectionEle)
  }
  appendChild<T extends Node>(node: T): T {
    this.tableBodyEle.appendChild(node)
    return node
  }
  insertBefore<T extends Node>(node: T, child: Node | null): T {
    this.tableBodyEle.insertBefore(node, child === this.sectionEle ? this.tableBodyEle.childNodes[0] : child)
    return node
  }
  removeChild<T extends Node>(child: T): T {
    this.tableBodyEle.removeChild(child)
    return child
  }
  get children() {
    return this.tableBodyEle.children
  }
}

window.customElements.define('f-table', FTable)
```

create node, like:
```typescript
class TableNode extends ElementNode {
  ...
  createDOM(config: EditorConfig): HTMLElement {
    const element = document.createElement('f-table')
    return element
  }
  ...
}
```

now, cant destroyNode TableNode's children Element.
because TableNode's children Element's parentNode not equal to the customElement
```typescript
if (dom.parentNode === parentDOM) {
  parentDOM.removeChild(dom);
}
```

### Before

if (dom.parentNode === parentDOM) {
  parentDOM.removeChild(dom);
}

### After

modification  the customElement's children and add isCustomElementChild judgment
```typescript
if (dom.parentNode === parentDOM || isCustomElementChild(dom, parentDOM)) {
  parentDOM.removeChild(dom);
}
```

PR #6759, opened 2024-10-23
[lexical][lexical-table] Feature: Scrollable tables with experimental getDOMSlot API

### LINKED ISSUE #4487: Bug: Cannot place cursor at the end of the line (Safari)

### LINKED ISSUE #6480: Feature: Scroll for table

etrepum:
## Description

This refactors the reconciler's treatment of ElementNode to allow you to specify where children should be inserted, which should allow for nodes that have specific requirements like a wrapping DIV element for display or UI reasons. Currently exposed as an `@internal` API.

As a demonstration it implements the horizontal scroll for tables from #6713. The `TablePlugin` has a new `hasHorizontalScroll` property that defaults to `false` (for backwards compatibility considerations) but the playground has a new setting `tableHorizontalScroll` that defaults to true and can (only) be disabled with a query string setting.

<details>
<summary>Screen recording showing scrollable table</summary>

https://github.com/user-attachments/assets/a94320ca-4e90-487e-b0c1-1d2979ae538e
</details>

It also incorporates the img linebreak hack for Safari from https://github.com/facebook/lexical/pull/6282 since that part of the reconciler was modified anyway. I don't think it's perfect, I believe there may still be some arrow key navigation quirks in that specific scenario (end of line decorator), but it's a net improvement that can be refined further. I suspect that the full solution would include hiding or removing these img nodes during arrow key navigation.

<details>
<summary>Safari end-of-line decorator before</summary>

https://github.com/user-attachments/assets/7f1df365-5721-45d2-9d77-53c0bf417c45
</details>
<details>
<summary>Safari end-of-line decorator after imghack</summary>

https://github.com/user-attachments/assets/ea63c3a4-0059-445d-8c0f-f4d6d5febc76
</details>

Closes #6713
Closes #6282
Closes #6480
Closes #4487
Closes #6587 
Closes #5981

## 0.20 Upgrade Notes

* Scrollable tables are disabled by default in this initial release for compatibility reasons. You can add support for them with `<TablePlugin hasHorizontalScroll={true} />`.
* A new theme property `tableScrollableWrapper` was added, a warning will be issued if you are using scrollable tables without it. It should include `overflow-x: auto;`.

## New concepts

`DOMSlot` is an abstraction much like a DOM ParentNode with methods like `insertChild`, `replaceChild`, etc. Notably there are three components to a DOMSlot: `element`, `after`, and `before`. 
* `element` is the parentNode for any managed lexical children of this ElementNode. The default is the result of `createDOM`.
* `after` is a DOM node (or null) that all managed children will be after. Default is null.
* `before` is a DOM node (or null) that all managed children will be before (e.g. `this.element.insertChild(child, this.before)` which is equivalent to `this.element.appendNode(child)` in the null case). Default is null.

All internal interactions with managing children have been ported to use `elementNode.getDOMSlot(dom)`, so that the ElementNode has an opportunity to set up a non-default DOMSlot.

`setDOMUnmanaged` / `isDOMUnmanaged` - These functions are used to mark a DOM node as unmanaged, much like a decorator node, which means that the mutation observer shouldn't worry about it. You would use this for DOM nodes in a DOMSlot that do not have corresponding LexicalNodes.

`LexicalNode.reconcileObservedMutation` - some of the functionality that was hard-coded into the mutation observer and branched on node type has been moved to this method

## Test plan

* Unit tests were added for `<TablePlugin hasHorizontalScroll={false} />` and `<TablePlugin hasHorizontalScroll={true} />` to demonstrate that the tables are wrapped in a div.
* The e2e test suite currently sets `tableHorizontalScroll` set to false in several suites so fewer tests needed to be updated to account for the wrappers, but otherwise it is tested by default. The legacy behavior without a wrapper can be tested with the environment `E2E_TABLE_MODE=legacy`, but I opted not to add that full suite to our already long e2e tests since a lot of functionality is still tested in legacy mode outside of the Tables.spec.mjs suite.
* All existing e2e and unit tests pass


packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -79,6 +80,30 @@ function setRowStriping(
   }
 }
 
+const scrollableEditors = new WeakSet<LexicalEditor>();
```
etrepum, line 83: I don't love doing this but we also don't have another place to put this sort of state

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -322,17 +322,43 @@ test.describe('Collaboration', () => {
     // Left collaborator types two pieces of text in the same paragraph, but with different styling.
     await focusEditor(page);
     await page.keyboard.type('normal');
+    await assertHTML(
```
etrepum, line 325: Adding the assertions and the delay/sleeps here seems to sort out the flakiness of this test. It would be nice if we had a better way to wait on specific history events. I assume the arrow key delay fix has something to do with waiting a tick for the selection to get resolved correctly.

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/TablesHTMLCopyAndPaste.spec.mjs
```diff
@@ -17,7 +17,9 @@ import {
 } from '../../../utils/index.mjs';
 
 test.describe('HTML Tables CopyAndPaste', () => {
-  test.beforeEach(({isCollab, page}) => initialize({isCollab, page}));
+  test.beforeEach(({isCollab, page}) =>
+    initialize({isCollab, page, tableHorizontalScroll: false}),
```
etrepum, line 21: Use legacy behavior in these tests, no need to test the wrapper (it is not exported or imported)

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -53,7 +53,9 @@ import {
 } from '../utils/index.mjs';
 
 test.describe.parallel('Selection', () => {
-  test.beforeEach(({isCollab, page}) => initialize({isCollab, page}));
+  test.beforeEach(({isCollab, page}) =>
+    initialize({isCollab, page, tableHorizontalScroll: false}),
```
etrepum, line 57: Selection scenarios with the wrapper are tested in the tables suite, no need to refactor here

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -75,6 +77,28 @@ async function fillTablePartiallyWithText(page) {
...
+      : expectedHtml,
+    IS_TABLE_HORIZONTAL_SCROLL && expectedHtmlFrameRight !== undefined
+      ? wrapTableHtml(expectedHtmlFrameRight, options)
+      : expectedHtmlFrameRight,
+    options,
+    ...args,
+  );
+}
+
+const WRAPPER = IS_TABLE_HORIZONTAL_SCROLL ? [0] : [];
```
etrepum, line 100: We add this to all DOM anchor/focus paths at exactly the right place to account for the wrapper, if present

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -75,6 +77,28 @@ async function fillTablePartiallyWithText(page) {
   await page.keyboard.press('c');
 }
 
+async function assertHTML(
```
etrepum, line 80: Eliminate branching and noise in the tests by using the wrapper-less HTML and adding it to the expectations here if it is enabled

packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -34,6 +34,8 @@ export const IS_COLLAB =
 const IS_RICH_TEXT = process.env.E2E_EDITOR_MODE !== 'plain-text';
 const IS_PLAIN_TEXT = process.env.E2E_EDITOR_MODE === 'plain-text';
 export const LEGACY_EVENTS = process.env.E2E_EVENTS_MODE === 'legacy-events';
+export const IS_TABLE_HORIZONTAL_SCROLL =
```
etrepum, line 37: If `E2E_TABLE_MODE=legacy` then use `hasHorizontalScroll={false}` otherwise true.

packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -175,6 +195,16 @@ export async function clickSelectors(page, selectors) {
     await click(page, selectors[i]);
   }
 }
+
+function removeSafariLinebreakImgHack(actualHtml) {
```
etrepum, line 199: Like tables, we fix up the expectations rather than branching in the individual tests.

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -116,6 +116,14 @@
...
   min-width: 25px;
 }
+.PlaygroundEditorTheme__tableScrollableWrapper {
+  overflow-x: auto;
+  margin: 0px 25px 30px 0px;
+}
+.PlaygroundEditorTheme__tableScrollableWrapper > .PlaygroundEditorTheme__table {
+  /* Remove the table's margin and put it on the wrapper */
+  margin: 0;
+}
```
etrepum, line 126: The margin is only moved around in this way because we support both wrapper and wrapper-less with the same CSS. If/when we only support wrapper tables then we would remove the margin from the table directly and apply it only to the wrapper.

packages/lexical/src/LexicalEditor.ts
```diff
@@ -144,7 +144,9 @@ export type EditorThemeClasses = {
   tableCellSortedIndicator?: EditorThemeClassName;
   tableResizeRuler?: EditorThemeClassName;
   tableRow?: EditorThemeClassName;
+  tableScrollableWrapper?: EditorThemeClassName;
   tableSelected?: EditorThemeClassName;
+  tableSelection?: EditorThemeClassName;
```
etrepum, line 149: I just noticed that this one was missing from the type

packages/lexical/src/LexicalNode.ts
```diff
@@ -1160,6 +1169,16 @@ export class LexicalNode {
...
+
+  /**
+   * @internal
+   *
+   * When the reconciler detects that a node was mutated, this method
+   * may be called to restore the node to a known good state.
+   */
+  reconcileObservedMutation(dom: HTMLElement, editor: LexicalEditor): void {
+    this.markDirty();
+  }
```
etrepum, line 1181: This was moved here because TextNode and DecoratorNode had the same behavior but ElementNode didn't so we just move it to a method for specialization

GermanJablo:
> A new theme property tableScrollableWrapper was added, a warning will be issued if you are using scrollable tables without it. It should include overflow-x: auto;.

I would prefer to have the property handle adding `overflow-x: auto`. I understand your reasoning, it's CSS. But in this case I don't see any advantage in having to repeat the intention in the theme config.

etrepum:
The intention is that it's useful to have a class for other reasons, and it does't really make sense to set both a class and a style. We could unconditionally set the style, I don't have a very strong opinion about it, but there's plenty of other things that don't work if the theme isn't configured appropriately.

packages/lexical-playground/src/Settings.tsx
```diff
@@ -28,10 +28,11 @@ export default function Settings(): JSX.Element {
       isAutocomplete,
       showTreeView,
       showNestedEditorTreeView,
-      disableBeforeInput,
+      // disableBeforeInput,
```
ivailop7, line 31: maybe delete with the switch code below?

packages/lexical-playground/src/Settings.tsx
```diff
@@ -28,10 +28,11 @@ export default function Settings(): JSX.Element {
...
       showTableOfContents,
       shouldUseLexicalContextMenu,
       shouldPreserveNewLinesInMarkdown,
+      // tableHorizontalScroll,
```
ivailop7, line 35: uncomment?

packages/lexical-playground/src/Settings.tsx
```diff
@@ -167,6 +168,13 @@ export default function Settings(): JSX.Element {
             checked={shouldPreserveNewLinesInMarkdown}
             text="Preserve newlines in Markdown"
           />
+          {/* <Switch
```
ivailop7, line 171: I'm happy to have the switch in the playground 👍🏻

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -36,24 +38,50 @@ import {
...
 export function TablePlugin({
   hasCellMerge = true,
   hasCellBackgroundColor = true,
   hasTabHandler = true,
-}: {
-  hasCellMerge?: boolean;
-  hasCellBackgroundColor?: boolean;
-  hasTabHandler?: boolean;
-}): JSX.Element | null {
+  hasHorizontalScroll = false,
```
ivailop7, line 77: I'm not opposed to having it 'true' by default, since people can flip it back if needed.

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -142,6 +167,19 @@ export class TableNode extends ElementNode {
     };
   }
 
+  getDOMSlot(element: HTMLElement): ElementDOMSlot {
+    const tableElement =
+      (element.nodeName !== 'TABLE' && element.querySelector('table')) ||
```
ivailop7, line 172: I guess this condition is meant to target the wrapper element, since it feels a bit unintuitive, worth considering doing this via 'getting the parent' from the element, not saying it's better, just food for thought.

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -142,6 +167,19 @@ export class TableNode extends ElementNode {
...
+      element;
+    invariant(
+      tableElement.nodeName === 'TABLE',
+      'TableNode.getDOMSlot: createDOM() did not return a table',
+    );
+    return super
+      .getDOMSlot(tableElement)
+      .withAfter(tableElement.querySelector('colgroup'));
```
ivailop7, line 180: why colgroup here? I believe this is only for a table created from Lexical, but if we copy/paste an external table, they don't always have a colgroup (Excel I think was an example, or GDocs). Would this be a problem for copy-pasted tables?

ivailop7 (commented):
30% through, putting some comments will continue tomorrow

packages/lexical-playground/src/Settings.tsx
etrepum: I was mostly commenting things out for space, the UI is not very good with as many options as we have now so I was trying to limit the ones that shouldn't really get used. I don't think there's any reason for a user to toggle this one and it has to refresh the page to work anyway so it might as well be done by URL only

packages/lexical-playground/src/Settings.tsx
etrepum: I didn't really see a good reason why people would want to turn it on and off, we can do it in the URL with the tests, but I put it there just in case. No strong opinion here, but this UX is bad when there are a lot of options.

packages/lexical-react/src/LexicalTablePlugin.ts
etrepum: I thought it probably makes sense to have it be a completely seamless opt-in upgrade, since it does change the DOM structure and adds a semi-required class (there's a fallback style with a warning).

packages/lexical-table/src/LexicalTableNode.ts
etrepum: The element argument *is* the parent (`element` is the element `createDOM` returned). We need to check to see if it's already a table, and if not (`nodeName !== 'TABLE'`), then we need to find the table. It's written in this way to be very unopinionated about how a subclass might implement `createDOM`.

packages/lexical-table/src/LexicalTableNode.ts
etrepum: If `colgroup` is present then we want it to be the first child, all children slots are *after* it. If there is no colgroup, then this returns `null` (but that wouldn't happen unless a subclass intentionally does not create one for some reason). The `after` property is mostly pedantic because DOM management is largely done by inserting elements before the `before` property in the slot (which is null in this case, so elements are appended) but there are some reconciler situations where we want to know the index of the first child and/or know which parts of the DOM to ignore from the mutation observer in which case it does become relevant to mark this `after` slot.

Copy/paste is not relevant, this only called with the result of `createDOM`, it's not used in the implementation of `importDOM`.

packages/lexical-playground/src/Settings.tsx
ivailop7: Let make the switches smaller, I don't think the url query arg will be discoverable enough. Anyhow, your call.

etrepum: I left it out because I couldn't think of a good reason for someone to discover this toggle from the UX, but I'm happy to uncomment either way. I'll wait until the reviews are done before code churn on that nit.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1538,6 +1587,10 @@ function $handleArrowKey(
         }
       }
     }
+    if (direction === 'down' && $isScrollableTablesActive(editor)) {
+      // Enable Firefox workaround
```
ivailop7, line 1591: Firefox-specific bug? Do we know more about this one?

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -1882,14 +1941,29 @@ function $getTableEdgeCursorPosition(
     return undefined;
   }
 
-  const tableNodeParentDOM = editor.getElementByKey(tableNodeParent.getKey());
-  if (!tableNodeParentDOM) {
-    return undefined;
-  }
-
   // TODO: Add support for nested tables
```
ivailop7, line 1944: I think you can remove the TODO comment IMO, I don't think we'll support nested tables any time soon or if we should in a first place.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -441,14 +446,30 @@ export function $getNodeFromDOMNode(
...
+export function setNodeKeyOnDOMNode(
+  dom: Node,
+  editor: LexicalEditor,
+  key: NodeKey,
+) {
+  const prop = `__lexicalKey_${editor._key}`;
+  (dom as Node & Record<typeof prop, NodeKey | undefined>)[prop] = key;
+}
+
+export function getNodeKeyFromDOMNode(
```
ivailop7, line 465: excited about those utility functions!

ivailop7 (approved):
LGTM, for @zurfyx to give this a look too.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
etrepum: Yes I think I wrote some more about it elsewhere in this PR but in Firefox natively moving the down arrow will skip over the table when wrapped with a div so the last cell get focused instead of the first. Couldn't find a CSS arrangement that allows scrolling with keyboard down arrow navigation works in Firefox at the same time. We really should have better selection APIs because there isn't really a way to fix it before reconciliation in "user" code

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
etrepum: People write bugs about nested tables at least once a month. I don't think they're that far off of working, someone just has to care enough to write those tests and PR.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
etrepum: Now that I'm back at a computer here's a link to where the problem is described https://github.com/facebook/lexical/pull/6759/files#diff-9f98cefa6e4564feb2a48d20b6e4278bcf999351abeceb27122732af6cd82413R273-R281

matveeva-as:
Hi! Thanks a lot for this!! When are you planing to release a new version of Lexical with this changes?

mnlkrs:
Thanks a lot for the work :)

etrepum:
Releases typically happen monthly, but if you need to try it sooner you can always use a nightly release

zurfyx (commented):
Late review but that's awesome work! This decouples the strict Node to DOM Node map which has been an inconvenience in some cases.

### Terminology nits
While I like the flexibility of the DOMSlotAPI, `withAfter` still seems confusing to me. 
1. Ideally, we would've had something like the DOM where `insertBefore`/`insertAfter` would work very closely to the DOM API.
2. `withAfter` works the opposite way as I would anticipate `.withAfter(tableElement.querySelector('colgroup'))`. To me this an element that inserts another element before.
3. `slot.element` is fairly ambiguous when a wrapper is present. Why does `createDOM` have to return the wrapper whereas `domSlot` has control over the child?	

### Testing

Did we test DOMSlot with inlines? I believe that LexicalElementNode.test and the E2E tests cover the block case and the particularly domSlot API well but I think it's important to have a test that covers inlines with wrappers if it's not there already.

Similarly, validate how selection behaves when before and after elements are presents. The selection logic seems to cover resolution for discovery, but is this sufficient for adjusting anchor/focus points accordingly on all other events?

etrepum:
DOM doesn't have an `insertAfter` API, only `insertBefore(dom, before)`. The `after` and `before` properties (set by `withAfter` and `withBefore`) are the markers that would be used in those methods (if a native `insertAfter` existed). I don't think there is a non-confusing way to name those markers using the same terminology that DOM does, because the `before` sentinel is after the children and vice versa for `after`. If we name the properties relative to the children it will be opposite to how the DOM semantics are (we would be doing `insertBefore(dom, after)` in that case, which is more confusing IMO).

etrepum:
`createDOM` has to return the DOM that's inserted into the document, and if it didn't, we'd have to break compatibility by changing the LexicalNode interface. It was done this way so that no user code or types would have to change.

Naming of element is arbitrary, could be `elementThatChildrenAreInsertedInto` or something like that, I guess. If you have a good name then please suggest it. I think most names are going to be confusing in one way or another.

etrepum:
Inlines will be tested when we have a use case to add them. I wasn't going to add another week or two of effort to come up with and test a use case that we don't have yet for an internal experimental API. I am sure there are going to be selection quirks because of how we use platform native arrow key navigation but don't have a good way to fix them up post-event in user code yet. I am pretty certain that the lexical selection will be mapped to the closest lexical point to where the DOM selection is, but visually the caret might not end up where the user wants (especially if they don't provide the correct css and contentEditable settings for their accessory elements, if present)

matveeva-as:
Hi! I've found a bug but don't sure how it can be related to this PR, but 0.19.0 works good, 0.20.1-nightly.20241114.0 doesn't).

When we have two tables in the Editor and press arrowDown in the second table, focus goes to the first cell in the first table.

etrepum:
Can you report this as a new issue please?

etrepum:
Nevermind, should be fixed in #6839 and released with the next nightly

JeffreyQ:
Hi, I had some questions around `$getTableAndElementByKey`. 

I see that you implement it [here](
https://github.com/facebook/lexical/commit/506ef8946779083fc2e9801779b68cb92b517a19#diff-9f98cefa6e4564feb2a48d20b6e4278bcf999351abeceb27122732af6cd82413R63) in packages/lexical-table/src/LexicalTableObserver.ts. 

However, you are importing it [here](https://github.com/facebook/lexical/commit/506ef8946779083fc2e9801779b68cb92b517a19#diff-db8a6b898d340752a32e531d7b86950c444725c0e75172007e6ea1f05bceea2fR11-R23) from `@lexical/table` which is of version 0.20.0.

However, when I'm looking at `LexicalTableUtils.d.ts` file in version 0.20.0 on [npm](https://www.npmjs.com/package/@lexical/table?activeTab=code) I do not see a `$getTableAndElementByKey`. 
image

etrepum:
The entire lexical monorepo is versioned together. The playground is *not* using `@lexical/table` v0.20 from npm, it is built with the exact version that is in the same commit as the playground. The versions don't update in git until a new release is created.

JeffreyQ:
> The entire lexical monorepo is versioned together. The playground is _not_ using `@lexical/table` v0.20 from npm, it is built with the exact version that is in the same commit as the playground. The versions don't update in git until a new release is created.

Got it, thank you for clarifying!

PR #6760, opened 2024-10-25
[lexical-react] Feature: add attributes in LinkPlugin

taro-shono:
## Description

Add attributes when pasting a link into a text.

### Before

Can not add `target`

https://github.com/user-attachments/assets/175b85a0-8993-42b7-aa4b-7e0325536fe3

### After

Can add `target`

https://github.com/user-attachments/assets/c1a67866-8951-4f9d-acdf-c1750c4bd5eb

### DIFF
--- a/packages/lexical-react/src/LexicalLinkPlugin.ts
+++ b/packages/lexical-react/src/LexicalLinkPlugin.ts
@@ -6,7 +6,12 @@
  *
  */
 
-import {$toggleLink, LinkNode, TOGGLE_LINK_COMMAND} from '@lexical/link';
+import {
+  $toggleLink,
+  LinkAttributes,
+  LinkNode,
+  TOGGLE_LINK_COMMAND,
+} from '@lexical/link';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {mergeRegister, objectKlassEquals} from '@lexical/utils';
 import {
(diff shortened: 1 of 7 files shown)


potatowagon:
code lgtm, can we have some tests please to check that attributes are added to link on paste?

packages/lexical-react/src/LexicalLinkPlugin.ts
```diff
@@ -38,13 +44,18 @@ export function LinkPlugin({validateUrl}: Props): null {
...
             }
             return false;
           } else {
             const {url, target, rel, title} = payload;
-            $toggleLink(url, {rel, target, title});
+            $toggleLink(url, {
+              ...attributes,
+              rel,
+              target,
+              title,
```
potatowagon, line 57: this means that the link attribute contents of payload will overwrite the contents of attributes

potatowagon (dismissed):

taro-shono:
@potatowagon I have added a test, so please check it out.

potatowagon (approved):

potatowagon (approved):

potatowagon:
thanks for this!

aleksey-pro:
Thanks for plugin. But if i switch off attributes after link is created - it doesn't removes target="__blank" atribute from node.

PR #6761, opened 2024-10-25
[lexical-website] Chore: upgrade to Docusaurus v3.6 - Docusaurus Faster

slorber:
## Description

Upgrade Docusaurus to v3.6 (upcoming, WIP)

Enable the [Docusaurus Faster](https://github.com/facebook/docusaurus/issues/10556) options.

The site now bundles in 7 seconds instead of 50 seconds. 
(However the doc gen phase remains slow)

## Test plan

??? local or preview?

### Before

```bash
npm run build --prefix packages/lexical-website
```

### After

```bash
npm run build --prefix packages/lexical-website
```


package.json
```diff
@@ -144,6 +146,7 @@
     "concurrently": "^8.2.2",
     "confusing-browser-globals": "^1.0.10",
     "cross-env": "^7.0.3",
+    "docusaurus-plugin-internaldocs-fb": "1.18.4",
```
slorber, line 149: Temporary duplicated dep.

It looks like npm workspace hoisting has a weird behavior and doesn't hoist prerelease versions? 🤷‍♂️ This cause problem because some packages are hoisted and others not.

packages/lexical-website/package.json
```diff
@@ -14,13 +14,16 @@
     "write-heading-ids": "docusaurus write-heading-ids"
   },
   "dependencies": {
-    "@docusaurus/core": "^3.3.2",
-    "@docusaurus/preset-classic": "^3.3.2",
-    "@docusaurus/theme-search-algolia": "^3.3.2",
+    "@docusaurus/core": "3.5.2-canary-6121",
+    "@docusaurus/faster": "3.5.2-canary-6121",
+    "@docusaurus/babel": "3.5.2-canary-6121",
+    "@docusaurus/bundler": "3.5.2-canary-6121",
```
slorber, line 20: these 2 can be removed later: transitive deps

patches/docusaurus-plugin-internaldocs-fb+1.18.5.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/packages/lexical-website/node_modules/docusaurus-plugin-internaldocs-fb/index.js b/packages/lexical-website/node_modules/docusaurus-plugin-internaldocs-fb/index.js
```
slorber, line 1: Temporay patch, until the fb preset adds support for Rspack in an upcoming release

slorber (commented):
I'll let you know once it's safe to review/merge

For now, I mostly needed a v3 site to test the upcoming faster features. I don't have access to the facebook preset source code so the goal was to test what changes I should advise the preset devs to do to add proper support.

antonk52:
@slorber I've just released docusaurus-plugin-internaldocs-fb@1.19.0 that should work with rspack. the patch should no longer be needed

slorber:
Great, it's working 👍 

Just did the PR cleanup, ready to review!

This command works well locally but I can't access Vercel logs:

`npm run build --prefix packages/lexical-website`

etrepum (approved):
Looks great! Thank you

etrepum:
@slorber I did notice that docusaurus-plugin-internaldocs-fb has an implicit dependency on mercurial (`hg`) and it will spew stderr when using watch.

```
/bin/sh: hg: command not found
/bin/sh: hg: command not found
/bin/sh: hg: command not found
/bin/sh: hg: command not found
```

```
❯ rg "'hg\s" node_modules/
node_modules/docusaurus-plugin-internaldocs-fb/lib/tracking-file.js
92:    const p = child_process_1.default.exec('hg root');

node_modules/docusaurus-plugin-internaldocs-fb/index.js
115:            .execSync('hg config fbconduit.reponame')
126:        return child_process_1.default.execSync('hg root').toString().trim();

node_modules/docusaurus-plugin-internaldocs-fb/bin/build-fb.js
9:const hgRoot = execSync('hg root').toString().trim();

node_modules/docusaurus-plugin-internaldocs-fb/bin/start-fb.js
9:const hgRoot = execSync('hg root').toString().trim();

node_modules/docusaurus-plugin-internaldocs-fb/bin/auth.js
9:const hgRoot = execSync('hg root').toString().trim();

node_modules/docusaurus-plugin-internaldocs-fb/bin/fb.js
9:const hgRoot = execSync('hg root').toString().trim();
```

It seems to work just fine, could probably be "fixed" by redirecting that output to nowhere.

slorber:
@etrepum unfortunately I don't really know about this internal FB plugin, @antonk52 can tell you more about those `hg` calls.

Weren't they already there before?

etrepum:
The problem isn't so much that there are hg calls, it's that their exceptions are being ignored without capturing stderr

antonk52:
> unfortunately I don't really know about this internal FB plugin, @antonk52 can tell you more about those hg calls.
>
> Weren't they already there before?

These have been there for the past 2~2.5 years. I do agree that we should update it to only make these calls when inside a mercurial project. I can try to do this, but that would be no sooner than next week.

PR #6765, opened 2024-10-25
Bug Fix; Fix context menu pasting supporting lexical format to maintain style

### LINKED ISSUE #6720: Bug: use PASTE_COMMAND command lost some style

Lexical version: 0.18.0

## Steps To Reproduce

5. see the new paste text

## The current behavior
The new text is small
## The expected behavior
## Impact of fix
It can reproduce every time.

citruscai:
## Description
clipboard API doesnt support "application/x-lexical-editor" type which is causing pasting through context menu to lose orignal style of copied text
Trying to register application/x-lexical-format in the but $onPasteForRichText is only storing text/html and text/plain

Closes #6720  (attempts to)
 see video for more info

https://github.com/user-attachments/assets/bdba1daf-18e1-460b-8d8f-fde5264c8ecb

## Test plan
TBD
### After
TBD

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -136,7 +136,7 @@ export function $insertDataTransferForRichText(
 ): void {
   const lexicalString =
     dataTransfer.getData('application/x-lexical-editor') ||
-    dataTransfer.getData('application/x-lexical/editor');
+    dataTransfer.getData('web application/x-lexical-editor');
   if (lexicalString) {
     try {
       const payload = JSON.parse(lexicalString);
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -523,13 +523,12 @@ const clipboardDataFunctions = [
  */
 export function $getClipboardDataFromSelection(
   selection: BaseSelection | null = $getSelection(),
-  editor: LexicalEditor = $getEditor(),
 ): LexicalClipboardData {
   const clipboardData: LexicalClipboardData = {
     'text/plain': selection ? selection.getTextContent() : '',
-    'web application/x-lexical-editor': $getLexicalContent(editor, selection),
   };
   if (selection) {
+    const editor = $getEditor();
     for (const [mimeType, $editorFn] of clipboardDataFunctions) {
       const v = $editorFn(editor, selection);
       if (v !== null) {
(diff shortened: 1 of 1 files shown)


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -133,8 +134,9 @@ export function $insertDataTransferForRichText(
   selection: BaseSelection,
   editor: LexicalEditor,
 ): void {
-  const lexicalString = dataTransfer.getData('application/x-lexical-editor');
-
+  const lexicalString =
+    dataTransfer.getData('application/x-lexical-editor') ||
+    dataTransfer.getData('application/x-lexical/editor');
```
etrepum, line 139:
```suggestion
    dataTransfer.getData('web application/x-lexical-editor');
```

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -39,6 +39,7 @@ export interface LexicalClipboardData {
   'text/html'?: string | undefined;
   'application/x-lexical-editor'?: string | undefined;
   'text/plain': string;
+  'web application/x-lexical-editor'?: string | null;
```
etrepum, line 42:
```suggestion
  'web application/x-lexical-editor'?: string | undefined;
```

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -519,12 +523,13 @@ const clipboardDataFunctions = [
...
   selection: BaseSelection | null = $getSelection(),
+  editor: LexicalEditor = $getEditor(),
 ): LexicalClipboardData {
   const clipboardData: LexicalClipboardData = {
     'text/plain': selection ? selection.getTextContent() : '',
+    'web application/x-lexical-editor': $getLexicalContent(editor, selection),
   };
   if (selection) {
-    const editor = $getEditor();
     for (const [mimeType, $editorFn] of clipboardDataFunctions) {
```
etrepum, line 533: I don't think we need any of the changes in this function, since `clipboardDataFunctions` will handle the export of the new web custom type

citruscai: hi!! i committed your suggestions and ran it locally.  still having problem with paste losing style :(
image

ivailop7 (approved):

ivailop7:
Could we add a test please. Happy to merge quickly, once we have a test.

etrepum:
I think there may still be some issues here, perhaps only the async api can write web custom formats, or maybe it needs to be serialized in some other way to get picked up. The docs are not very clear on this. Possibly the execCommand hack could be done here to solve it a different way? I won't have a lot of time to investigate further until next week

citruscai:
update: think i found a solution and just tested it. need to clean my code up  and create tests and will push commits tomorrow thinkt im  also gonna make a new PR and close this one

PR #6772, opened 2024-10-27
[lexical] Feature: Add onUpdate function during update with $onUpdate

### LINKED ISSUE #6767: Feature: Register commands with update options
## Description
Allow the third argument to `editor.registerCommand()` to take an object that includes the priority as well as the `update` options. `registerCommand()` could then be called as follows:

```ts
editor.registerCommand(COMMAND_NAME, commandFunction, COMMAND_PRIORITY_LEVEL);
```

or 

```ts
editor.registerCommand(COMMAND_NAME, commandFunction, {
  discrete: true,
  onUpdate: // post-update function,
  priority: COMMAND_PRIORITY_LEVEL,
  skipTransforms: true
});
```

This would let us manage the command with more finesse, in situations where we need to use these options to narrow the scope of the command's changes.

For my current use case, I have an update listener that submits user input to the server. I also have a command that updates the editor with data from the server. While the command is operating, I don't want the listener to be active, or else it creates a feedback loop.

Currently I'm setting a flag in the command, and the update listener knows not to send to the server while that flag is set. I want to be able to unset the flag in an `onUpdate` callback function, but can't do so directly in the command. Right now I queue a microtask to do so, but that seems "unLexical."

landisdesign:
## Description

Adds `$onUpdate(fn)` to allow an in-progress update/command to append a function to the `_deferred` stack.

Closes #6767

### Before

This functionality did not exist. Calling `editor.update` with an `onUpdate` function was the only way to perform an action after an update.

### After

Performing an update such as:

```ts
editor.update(() => {
  // perform updates
  $onUpdate(() => {
    console.log('second onUpdate);
  });
}, {
  onUpdate: () => {
    console.log('first onUpdate');
  }
});
```

would emit `first onUpdate`, followed by `second onUpdate`, once the updates were flushed. This is essentially what the unit test does.

Successful unit test results


landisdesign:
Ugh started my code from an earlier version. Closing and reopening with correct starting point.

PR #6777, opened 2024-10-29
fix empty line issue

Shopiley:
## Description

Currently, empty lines on markdown disappear when the markdown button is toggled.
This pull request aim to ensure empty lines are maintained upon toggling markdown. 
I realised that setting a variable called `shouldPreserveNewLines`  to true across all instances it was used seemed to fix the issue.

Closes #

### Before

### After

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -30,7 +30,7 @@ import {isEmptyParagraph, transformersByType} from './utils';
  */
 export function createMarkdownExport(
   transformers: Array<Transformer>,
-  shouldPreserveNewLines: boolean = false,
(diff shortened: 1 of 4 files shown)


etrepum (commented):
The current default markdown behavior needs to be preserved for backwards compatibility reasons. The issue here is that it might be hard for users to discover this setting because there is a lack of documentation in the markdown package about this setting and the expected behavior. Unfortunately markdown is quite a complex specification mostly because there are many different ones (commonmark, github flavored markdown, mdx, other extensions etc.), many of them have conflicting requirements, and I think what has been accidentally created here is yet another "lexical markdown" spec which is not well defined but certain applications depend on exactly how it behaves now.

Shopiley:
Okay, so what I have to do is experiment with this setting and document the expected behaviour.

Issue #6768, opened 2024-10-27
Bug: When I type : in the editor, it becomes unresponsive.


Nagendra7509:
I'm trying to enter a date in the editor (ex: `26-10-2024 10:00:00`). However, after typing `26-10-2024 10:00` and pressing `:` the editor freezes and becomes unresponsive to any clicks.

https://github.com/user-attachments/assets/8b1593e4-c07a-4f3f-b270-a08c398bddf3

Lexical version: 0.18.0

I am using Lexical in Chrome browser (`Version 127.0.6533.88 (Official Build) (64-bit)`)

## NOTE: I've identified that this issue is caused by the `MarkdownShortcutPlugin`.

vantage-ola:
yes, i noticed that too..
even on the latest `v0.19.0` build

vantage-ola:
you working on it?

 @Nagendra7509

Nagendra7509:
I’m not actively working on it; I’ve only identified the plugin responsible for the issue. I haven’t explored the specific cause within the plugin.

PR #6778, opened 2024-10-30
#6768 fix to avoid infinite markdown shortcut matchers run

fantactuka:
Issue #6768 happens because of emoji markdown shortcut. It has conditional replacement which was not really supported for shortcuts:

https://github.com/facebook/lexical/blob/07ca5c83c2f512734976504ef9fae5d39b6072ae/packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts#L105-L108

1. Quick option is to disable it from list of shortcut transformers: 

https://github.com/facebook/lexical/blob/07ca5c83c2f512734976504ef9fae5d39b6072ae/packages/lexical-playground/src/plugins/MarkdownShortcutPlugin/index.tsx#L15

2. Another alternative would be to update emoji RegExp to match exact emoji names and instead of having `:([a-Z0-9]):` it'd be `:(smile|sad|happy|...|joy):`. The problem is that with current emoji-list that regex is 23kb long.

3. Enable conditional replacement (this PR). The reason it ran infinitely is because once we find match we call `textNode.splitText(matchOffset)` to extract text node that should be replaced:

https://github.com/facebook/lexical/blob/07ca5c83c2f512734976504ef9fae5d39b6072ae/packages/lexical-markdown/src/MarkdownShortcuts.ts#L174-L181

Even if we don't replace anything, it still causes full update cycle which triggers another transformers run. One thing is whether we should even call update cycle for unused result of `.splitText` (cc @zurfyx) given split nodes will reconcile back to a single text node. But on top of it, we don't really need to run any shortcuts logic if selection itself hasn't changed as we only suppose to do it when user types (i.e. selection moves).

Another (more minor) issue was in emoji regex that should ensure we only attempt to run matcher against end of node's text

### Test plan

Tests + sanity check

```
npm run test-e2e-chromium -- Markdown
✓ 52 passed (28.5s)
```

### DIFF
--- a/packages/lexical-markdown/src/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/MarkdownShortcuts.ts
@@ -480,10 +480,13 @@ export function registerMarkdownShortcuts(
       const selection = editorState.read($getSelection);
       const prevSelection = prevEditorState.read($getSelection);
 
+      // We expect selection to be a collapsed range and not match previous one (as we want
+      // to trigger transforms only as user types)
       if (
         !$isRangeSelection(prevSelection) ||
         !$isRangeSelection(selection) ||
-        !selection.isCollapsed()
+        !selection.isCollapsed() ||
+        selection.is(prevSelection)
       ) {
         return;
       }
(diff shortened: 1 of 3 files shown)


potatowagon (approved):
thanks for the quick fix!

etrepum (approved):

potatowagon:
closes: [6768](https://github.com/facebook/lexical/issues/6768)

PR #6780, opened 2024-10-30
Bug Fix:  ContextMenu Paste option not preserving style 

### LINKED ISSUE #6720: Bug: use PASTE_COMMAND command lost some style

citruscai:
## Description
When you paste a text with a different font size or custom style using the context menu, the pasted text loses it's style and is the default style.

Closes #6720 

### Before

https://github.com/user-attachments/assets/a2d22e83-ca36-4fff-a9e2-02b27dca678b

### After

https://github.com/user-attachments/assets/3428907e-bbaa-4dac-8349-3b5b687b8a02


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -1358,7 +1358,7 @@ function applyTextFormatFromStyle(
     if (shouldApply && !lexicalNode.hasFormat(shouldApply)) {
       lexicalNode.toggleFormat(shouldApply);
     }
-
+    lexicalNode.setStyle(style.cssText);
```
etrepum, line 1361: This is one workaround it but it breaks other things, it's not generally desirable or backwards compatible for *all* styles to be imported because the editor doesn't know how to work with most of the styles available in CSS and you would be stuck with them since you can't manage them from the editor. For example if you pasted some text with margins or padding or something like that you wouldn't really be able to get rid of it without deleting the text.

etrepum: In this particular case, the solution should be implemented such that it *only* works in the playground via some configuration and doesn't change the existing import behavior of TextNode

citruscai: so

> In this particular case, the solution should be implemented such that it _only_ works in the playground via some configuration and doesn't change the existing import behavior of TextNode

should i do something similiar like the above code? such as only supporting certain custom styles like font size and background color and not sure what else

etrepum: Yes, but in order to maintain backwards compatibility the code has to be done somewhere else. I replied to your discord with an untested sketch of some code that will probably get close, using the html import configuration option that will be documented via https://github.com/facebook/lexical/issues/6762

etrepum (dismissed):
This is looking pretty good, only a small suggestion left! Let's see if this affects any of the e2e tests that will run after approval

packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/ContextMenuCopyAndPaste.spec.mjs
```diff
@@ -52,4 +52,44 @@ test.describe('ContextMenuCopyAndPaste', () => {
       `,
     );
   });
+  test('Rich text Copy and Paste', async ({page, isRichText}) => {
+    test.skip(isRichText);
```
etrepum, line 56: I think this is backwards

```suggestion
    test.skip(!isRichText);
```

citruscai:
> This is looking pretty good, only a small suggestion left! Let's see if this affects any of the e2e tests that will run after approval

yipee!!

citruscai:
@etrepum  just fixed font size, the regex was returning null. now trying to fix background color and color

etrepum (approved):
Everything looks good, only one test failure this run and it's a flaky test that has passed before and passes locally for me. In the last commit I added some extra locking around clipboard related tests to hopefully make them more reliable.

I think this is good to go and will get merged shortly. Great work!

Issue #6740, opened 2024-10-16
Feature: Column Resizing Support in Layouts Plugin


harshmetkel24:
## Description
In one of our projects, we needed to support column resizing within the Layouts plugin. We were able to achieve this by adapting some of the logic already implemented for table resizing. Given that this is a fairly common requirement for projects dealing with dynamic layouts, it might be beneficial for other users as well.

Our specific use case involved a Reports module where we originally supported multiple widgets in a row. Previously, we used an external library to handle the resizing of these widgets. However, as we have now migrated fully to Lexical, we wanted to retain the same functionality using Lexical’s layout system. We used the columns layout to manage multiple widgets in a row, and by adding resizing functionality, we successfully mimicked the behavior we had with the external library.

### Benefits

	•	Consistency: Having built-in column resizing would create a seamless experience for projects that use the Layouts plugin.
	•	Usability: Many users may have similar requirements for flexible grid layouts where content is dynamically adjusted by users. Having this as a built-in feature would save the effort of manually implementing or integrating external libraries.
	•	Expandability: It provides more customization for layouts and makes the Layouts plugin more powerful in building complex, interactive content like reports, dashboards, and editors.

harshmetkel24:
Wanted to work on this

harshmetkel24:
I wanted to work on this!

ivailop7:
This would be an interesting improvement to the plugin. Feel free to give it a go!
If you are are going to use a library for the drag and resize side of things, my advice is to go for atlassian's https://github.com/atlassian/pragmatic-drag-and-drop
Which is really small and performant.

v-asad:
The PR seems to have been closed. Is there a PR or an issue relevant to this one that's in development?

PR #6787, opened 2024-11-01
[lexical-playground] Feature: columns resizing in layouts plugin

harshmetkel24:
[lexical-playground] Feature: support to add new columns side by side and column resize 

## Description
- Currently when adding `Columns Layout` promt asks to chose number of columns. But once added there is no way to add more columns or even resizing already added columns. If someone wants to change something they have to start fresh. Also there is no way to free resize columns, can only chose from predefined options.
- After this change users will be able to add more columns to end by clicking on `+` button. This behaviour is adapted from lexical-table only. Also user's can resize columns according to their need.

Closes #[6740](https://github.com/facebook/lexical/issues/6740)

### Before
image

No option to add new columns or resize existing ones

### After

https://github.com/user-attachments/assets/c8f2ed1f-1a8e-4ad3-831e-e0ba6e5318b1

Now users can add more columns and resize them as well according to their need

### DIFF
--- a/packages/lexical-playground/src/plugins/LayoutColumnResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/LayoutColumnResizer/index.tsx
@@ -1,0 +1,344 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import './index.css';
+
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {calculateZoomLevel} from '@lexical/utils';
+import {$getNearestNodeFromDOMNode, LexicalEditor} from 'lexical';
+import * as React from 'react';
+import {createPortal} from 'react-dom';
+
+import {
+  $findLayoutItemIndexGivenLayoutContainerNode,
+  $getLayoutContainerNodeIfLayoutItemNodeOrThrow,
+} from '../../nodes/LayoutContainerNode';
+import {$isLayoutItemNode} from '../../nodes/LayoutItemNode';
+
+export const MIN_LAYOUT_COLUMN_WIDTH = 100;
+
+interface Props {
+  editor: LexicalEditor;
+}
+
+type MousePosition = {
+  x: number;
+  y: number;
+};
+
+type MouseDraggingDirection = 'right' | 'bottom';
+
+const LayoutColumnResizer: React.FC<Props> = (props) => {
+  const {editor} = props;
+  const [mouseCurrentPos, updateMouseCurrentPos] =
+    React.useState<MousePosition | null>(null);
+
+  const [activeCell, updateActiveCell] = React.useState<HTMLElement | null>(
+    null,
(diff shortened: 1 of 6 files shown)


packages/lexical-playground/src/plugins/LayoutColumnResizer/index.tsx
```diff
@@ -0,0 +1,344 @@
...
+    };
+  }, [activeCell, draggingDirection, editor, isMouseDownOnEvent, resetState]);
+
+  const resizerStyles = getResizers();
+
+  return (
+    <div ref={resizerRef}>
+      {activeCell != null && !isMouseDown && (
+        <div
+          className="PlaygroundEditorTheme__resizer"
```
etrepum, line 321: This should ideally come from the theme rather than be hard-coded here, at least according to the coding conventions elsewhere in the playground

packages/lexical-playground/src/plugins/LayoutHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+    );
+
+    const isOutside = !(
+      layoutContainerNode ||
+      layoutItemNode ||
+      target.closest<HTMLElement>(
+        'button.PlaygroundEditorTheme__layoutAddColumns',
+      ) ||
+      target.closest<HTMLElement>('div.PlaygroundEditorTheme__resizer')
+    );
```
etrepum, line 60: These classes should come from the theme rather than be hard-coded here, at least according to the coding conventions elsewhere in the playground

packages/lexical-playground/src/plugins/LayoutHoverActionsPlugin/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+  }, [editor]);
+
+  if (!showColumnAction) {
+    return null;
+  }
+
+  return (
+    <button
+      className="PlaygroundEditorTheme__layoutAddColumns"
```
etrepum, line 235: This class should come from the theme rather than be hard-coded here, at least according to the coding conventions elsewhere in the playground

etrepum (approved):
Ideally there would be more usage of the theme rather than hard-coded classes, although perhaps a data attribute or some other identifier should be used for the cases where someone wants to use utility classes in their theme. It seems to work well enough in the cases that I tried

ivailop7:
I don't want to discourage the development of this plugin, but I think it might make more sense to have this one in a separate repo, vs inside the playground and we can list a number of other plugins people have built in the documentation.

potatowagon:
there seems to be a regression for this test: packages/lexical-playground/__tests__/e2e/Tables.spec.mjs:3743:3 › Tables › Paste and insert new lines after unmerging cells 
https://github.com/facebook/lexical/actions/runs/11691145148/job/32797500364?pr=6787

harshmetkel24:
> there seems to be a regression for this test: packages/lexical-playground/**tests**/e2e/Tables.spec.mjs:3743:3 › Tables › Paste and insert new lines after unmerging cells https://github.com/facebook/lexical/actions/runs/11691145148/job/32797500364?pr=6787

will look into this

PR #6788, opened 2024-11-03
Mention nodes shouldn't be spellcheck'd :)

sevki:
Makes names [no-spell-check](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck) which is nice for everyone not called John. 

image


ivailop7 (dismissed):

ivailop7:
Change is OK, but you would need to update the respective tests.

etrepum:
You could update the examples without being able to run them locally, they are not snapshot tests where tooling is used to update them. Should be fairly straightforward to find them with `class="mention"` .

Are you installing playwright with the given instructions (`npx install playwright`) and it still doesn't work? I don't think that the instructions really depend on too many globally installed things, but nixos is quite a bit different so not sure. https://lexical.dev/docs/testing#end-to-end-e2e-tests

packages/lexical-playground/src/nodes/MentionNode.ts
```diff
@@ -80,6 +80,7 @@ export class MentionNode extends TextNode {
     const dom = super.createDOM(config);
     dom.style.cssText = mentionStyle;
     dom.className = 'mention';
+    dom.spellcheck = false;
```
potatowagon, line 83: looks like a low risk change

potatowagon (approved):

PR #6791, opened 2024-11-04
[lexical-list] Bug Fix: Handle appending inline element nodes in ListNode.append

amanharwara:
Currently, `ListNode.append` will always convert an element node into a text node, even if the element node is inline and can be appended as a whole which can end up removing formatting and unwrapping things like links.

While this is rarely hit, since most of the code in the lexical codebase accounts for appending to a child list item instead of the list directly, I think it still makes sense for this to be correct.

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -200,8 +200,12 @@ export class ListNode extends ElementNode {
         if ($isListNode(currentNode)) {
           listItemNode.append(currentNode);
         } else if ($isElementNode(currentNode)) {
-          const textNode = $createTextNode(currentNode.getTextContent());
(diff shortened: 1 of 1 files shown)


etrepum (approved):
A test would be nice to demonstrate that this works as expected, but otherwise it seems like a positive change. I do wonder if this would be better done in a transform because there are other ways to violate this invariant (e.g. `listItemNode.replace(textNode)`).

Also makes me think that the implementation of `node.replace(otherNode)` should probably delegate to the containing ElementNode (like in the DOM API), right now it just swaps the pointers.

Shopiley:
Hi, I've added a test for this bug fix here: https://github.com/facebook/lexical/pull/6826

also wanted to point out that this PR fixes issue #6793 so we can close that.

PR #6810, opened 2024-11-08
[lexical-mark] Bug Fix: Stop MarkNode __ids array deep copy in clone

ebads67:
## Description

In this PR creation of a new `__ids` array is avoided in the clone function of the MarkNode, and the cloned MarkNode will have the same reference to` __ids` array of the original node. This has made an issue in the `syncPropertiesFromYjs` function where we [compare](https://github.com/facebook/lexical/blob/76f8569211780ee7e913176d99d23eddd1cd3b6c/packages/lexical-yjs/src/Utils.ts#L282C5-L282C35
) the old and new nodes' properties, and only make an update to ydoc if the properties are changed. In the case MarkNodes, because the clone function deep-copies the `__ids` array the property comparison detects a change, despite the arrays being identical.

The fix is keeping the same `__ids` array in the cloned node, and only creating new `__ids` array when the array is modified.

### DIFF
--- a/packages/lexical-mark/src/MarkNode.ts
+++ b/packages/lexical-mark/src/MarkNode.ts
@@ -39,7 +39,7 @@ export class MarkNode extends ElementNode {
   }
 
   static clone(node: MarkNode): MarkNode {
-    return new MarkNode(Array.from(node.__ids), node.__key);
+    return new MarkNode(node.__ids, node.__key);
   }
 
   static importDOM(): null {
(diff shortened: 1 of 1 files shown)


ivailop7 (dismissed):

ivailop7 (approved):
LGTM, @etrepum too have a look as well.

etrepum (approved):
LGTM, maybe a little more defensive than it needs to be, although this maintains type compatibility externally. The `$isMarkNode(self)` checks are useless if you wanted to simplify further, I can only imagine they are there from a time when `this.getLatest()` and `this.getWritable()` had the wrong types?

PR #6811, opened 2024-11-08
[lexical-playground] Fix: tabs do not show strikethrough/underline

### LINKED ISSUE #6808: Bug: Tabs do not show strikethrough
👋  Howdy maintainers, first of all big fan of the library! Thanks for all the hard work you put into this great software.

Lexical version: 0.19.0

## Steps To Reproduce

1. Go to lexical playground at https://playground.lexical.dev/
2. Type in `Hello	foo` which is hello press the Tab key then foo.
3. Select all the text
4. Format as strikethrough
5. unselect the text
This then shows only the typed text as struck through, the tab is blank.

Screenshot 2024-11-07 at 16 44 14

Link to code example:

## The current behavior

Tabs do not have a strikethrough.

## The expected behavior

Tabs should have strike through.

## Impact of fix
Would keep formatting consistent between lexical and other text editors.

vantage-ola:
## Description

Made some changes to `.PlaygroundEditorTheme__textStrikethrough` to show strikethroughs on tabs

Closes #6808 

### Before

Screenshot 2024-11-08 161836

After selecting the whole text, it shows the strikethrough is present
Screenshot 2024-11-08 162158

### After
The slight change in css makes the strikethrough visible 

Screenshot 2024-11-08 162320

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,9 +9,6 @@ on:
       - 'packages/lexical-website/**'
   pull_request:
     types: [opened, synchronize, reopened]
-    paths-ignore:
-      - 'examples/**'
-      - 'packages/lexical-website/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/packages/lexical-website/docs/react/plugins.md
+++ b/packages/lexical-website/docs/react/plugins.md
@@ -41,7 +41,7 @@ const initialConfig = {
 
 ### `LexicalPlainTextPlugin`
 
-React wrapper for `@lexical/plain-text` that adds major features for plain text editing, including typing, deletion and copy/pasting
+React wrapper for `@lexical/plain-text` that adds major features for plain text editing, including typing, deletion and copy/pasting.
 
 ```jsx
 <PlainTextPlugin
@@ -53,7 +53,7 @@ React wrapper for `@lexical/plain-text` that adds major features for plain text 
 
 ### `LexicalRichTextPlugin`
 
-React wrapper for `@lexical/rich-text` that adds major features for rich text editing, including typing, deletion, copy/pasting, indent/outdent and bold/italic/underline/strikethrough text formatting
+React wrapper for `@lexical/rich-text` that adds major features for rich text editing, including typing, deletion, copy/pasting, indent/outdent and bold/italic/underline/strikethrough text formatting.
 
 ```jsx
 <RichTextPlugin
@@ -65,7 +65,7 @@ React wrapper for `@lexical/rich-text` that adds major features for rich text ed
 
 ### `LexicalOnChangePlugin`
 
-Plugin that calls `onChange` whenever Lexical state is updated. Using `ignoreHistoryMergeTagChange` (`true` by default) and `ignoreSelectionChange` (`false` by default) can give more granular control over changes that are causing `onChange` call
+Plugin that calls `onChange` whenever Lexical state is updated. Using `ignoreHistoryMergeTagChange` (`true` by default) and `ignoreSelectionChange` (`false` by default) can give more granular control over changes that are causing `onChange` call.
 
 ```jsx
 <OnChangePlugin onChange={onChange} />
@@ -73,7 +73,7 @@ Plugin that calls `onChange` whenever Lexical state is updated. Using `ignoreHis
 
 ### `LexicalHistoryPlugin`
 
-React wrapper for `@lexical/history` that adds support for history stack management and `undo` / `redo` commands
+React wrapper for `@lexical/history` that adds support for history stack management and `undo` / `redo` commands.
 
 ```jsx
 <HistoryPlugin />
@@ -81,7 +81,7 @@ React wrapper for `@lexical/history` that adds support for history stack managem
 
 ### `LexicalLinkPlugin`
 
-React wrapper for `@lexical/link` that adds support for links, including `$toggleLink` command support that toggles link for selected text
+React wrapper for `@lexical/link` that adds support for links, including `$toggleLink` command support that toggles link for selected text.
 
 ```jsx
 <LinkPlugin />
@@ -107,7 +107,7 @@ React wrapper for `@lexical/list` that adds support for check lists. Note that i
 
 [![See API Documentation](/img/see-api-documentation.svg)](/docs/api/modules/lexical_react_LexicalTablePlugin)
 
-React wrapper for `@lexical/table` that adds support for tables
+React wrapper for `@lexical/table` that adds support for tables.
 
 ```jsx
 <TablePlugin />
@@ -153,7 +153,7 @@ const MATCHERS = [
 
 ### `LexicalClearEditorPlugin`
 
-Adds `clearEditor` command support to clear editor's content
+Adds `clearEditor` command support to clear editor's content.
 
 ```jsx
 <ClearEditorPlugin />
@@ -161,7 +161,7 @@ Adds `clearEditor` command support to clear editor's content
 
 ### `LexicalMarkdownShortcutPlugin`
 
-Adds markdown shortcut support: headings, lists, code blocks, quotes, links and inline styles (bold, italic, strikethrough)
+Adds markdown shortcut support: headings, lists, code blocks, quotes, links and inline styles (bold, italic, strikethrough).
 
 ```jsx
 <MarkdownShortcutPlugin />


packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -60,9 +60,22 @@
 .PlaygroundEditorTheme__textStrikethrough {
   text-decoration: line-through;
 }
-.PlaygroundEditorTheme__textUnderlineStrikethrough {
-  text-decoration: underline line-through;
```
etrepum, line 64: This deletion removes support for text that has both underline and strikethrough. I think you meant to delete the one above, although this would also need to be to remove the `line-through` and also target it with the rules below `::after`

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -60,9 +60,22 @@
...
+.PlaygroundEditorTheme__textStrikethrough {
+  position: relative;
+  text-decoration: none;
```
etrepum, line 65: this is redundant if the correct rule is deleted

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -60,9 +60,22 @@
...
+}
+
+.PlaygroundEditorTheme__textStrikethrough::after {
+  content: '';
+  position: absolute;
+  top: 50%;
+  left: 0;
+  width: 100%;
+  height: 2px;
+  background-color: black;
```
etrepum, line 75: this should be `currentColor` instead of black to match how text decoration works

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -60,9 +60,22 @@
...
+  text-decoration: none;
+}
+
+.PlaygroundEditorTheme__textStrikethrough::after {
+  content: '';
+  position: absolute;
+  top: 50%;
+  left: 0;
+  width: 100%;
+  height: 2px;
```
etrepum, line 74: In my testing it seemed like `0.1em` was a better height for this, with a `transform: translate(0,50%);` to move it slightly. That's probably font specific though. I tested in the dev tools by enabling the text-decoration and giving this a different color with opacity so I could see how they line up.

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
vantage-ola: very  valid. 👍

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
vantage-ola: would change that.. i noticed the different height, but i wasn't too sure about it.

vantage-ola:
i took another approach to fixing this.., i just jumped into it without thinking and was in an hurry to fix it... I had even forgotten both underline and strikethrough existed...

the reason I used `white-space: pre;`  is because normally HTML collapses multiple spaces into one, but` white-space: pre` treats them like` <pre>` tags

vantage-ola:
> i took another approach to fixing this.., i just jumped into it without thinking and was in an hurry to fix it... I had even forgotten both underline and strikethrough existed...
> 
> the reason I used `white-space: pre;` is because normally HTML collapses multiple spaces into one, but` white-space: pre` treats them like` \<pre\>` tags

no need for them actually. 😅 silly me

vantage-ola:
this should be good to go right?

etrepum (commented):
I don't think the pseudo-element approach works correctly when the text wraps 
Screenshot 2024-11-12 at 08 45 39

etrepum:
Text wrap means that the text is longer than one line, so it wraps to the next line. You can see it in the second paragraph which has strikethrough and underline, you can see not all of the text has strikethrough because the pseudo-element is not the same shape as the text it's supposed to draw a line through. The tab creates a new element so that's where the strikethrough resets and works until the next wrap.

bdgrichards:
Also when a single word wraps multiple lines this seems to break down, since the added line renders in the middle of the two/three/n lines:

https://github.com/user-attachments/assets/cd7cb7be-4045-472d-ae7e-24d2e904c3c1

vantage-ola:
@etrepum will give a better answer. but I think the issue fixes only the lexical playground. I dont know of the regular lexical editors :)

etrepum:
Since the fix will be presumably be primarily CSS based, it will be up to the user to configure their theme with the correct CSS to make it work. I think perhaps in this case the best solution might be to add a class to the TabNode so that it can be specifically targeted for a workaround rather than applying a workaround to all text nodes

etrepum:
Don't bother trying to replace the spaces, that will cause problems. I suggested a possible approach in my last comment, this workaround only needs to apply to tab nodes. If you add a class so they can be targeted you should be able to work around it for only those nodes and leave the other CSS alone so this fix doesn't break anything else.

vantage-ola:
I tried soemthing... but it breaks something in the tests... but i think the strikethrough/underlining or both works, atleast from what i tested...


```
Argument of type 'TextNode' is not assignable to parameter of type 'CodeHighlightNode | TabNode | LineBreakNode'.
      Property 'getClassName' is missing in type 'TextNode' but required in type 'TabNode'.

    265       let node = getFirstCodeNodeOfLine(firstSelectionNode);
                                                ~~~~~~~~~~~~~~~~~~

      packages/lexical/src/nodes/LexicalTabNode.ts:52:3
        52   getClassName(): string {
             ~~~~~~~~~~~~
        'getClassName' is declared here.
```

packages/lexical/src/nodes/LexicalTabNode.ts
```diff
@@ -47,6 +48,45 @@ export class TabNode extends TextNode {
...
+    prevNode: TabNode,
+    dom: HTMLElement,
+    config: EditorConfig,
+  ): boolean {
+    const updated = super.updateDOM(prevNode, dom, config);
+    if (updated) {
+      dom.className = this.getClassName();
+    }
+    return updated;
+  }
```
etrepum, line 88: This is far more work than it needs to be, the format classes are already assigned by the super implementation. You only need to add the class for TabNode in createDOM. This class should be defined in the theme, not hard-coded in this implementation.

etrepum (commented):
Not sure why that check failed, let's see if it happens again once the other issues are fixed. I can't review the functionality because it didn't deploy.

packages/lexical/src/nodes/LexicalTabNode.ts
vantage-ola: ok, thank you. will get to work on this. and make necessary adjustments.

vantage-ola:
so, only createDOM is needed in LexicalTabNode, but I noticed a number of tests are failing due to a new class name being introduced to  lexical. And while testing, I dont think the pseudo element approach will  ever work(but i may be wrong though)

for example:

```
 ● LexicalSelection tests › Paste text, move selection and delete word forward (#138)

    expect(received).toBe(expected) // Object.is equality

    Expected: "<div contenteditable=\"true\" style=\"user-select: text; white-space: pre-wrap; word-break: break-word;\" data-lexical-editor=\"true\"><p class=\"editor-paragraph\" dir=\"ltr\"><span data-lexical-text=\"true\">ABD</span><span data-lexical-text=\"true\">       </span><span data-lexical-text=\"true\">EFG</span></p></div>"
    Received: "<div contenteditable=\"true\" style=\"user-select: text; white-space: pre-wrap; word-break: break-word;\" data-lexical-editor=\"true\"><p class=\"editor-paragraph\" dir=\"ltr\"><span data-lexical-text=\"true\">ABD</span><span class=\"PlaygroundEditorTheme__tabNode\" data-lexical-text=\"true\">       </span><span data-lexical-text=\"true\">EFG</span></p></div>"
```

etrepum:
Yes, of course the tests that use tab node will have to change when a class is added.

vantage-ola:
@etrepum  I noticed something a while ago(totally unrelated) i dont know if its a bug..  sometimes when you press tab, it doesnt go to the next 4 spaces, it (folds the tab('4' spaces to '1' space), but when you do it again,  it goes 4 spaces forward..  Why is that? is that the right behaviour 
Screenshot 2024-11-22 232544

vantage-ola:
i have a suggestion @etrepum instead of adding the artificial strikethroughs/underlines or both for affected tab nodes... what of a visual tab indicator like i have here in this screenshot?

Screenshot 2024-11-22 235956

etrepum:
Regarding spacing with tabs - this is how they work. Tabs are not just wide spaces, they are variable width depending on how much space there is until the next horizontal tab stop.

https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size
https://en.wikipedia.org/wiki/Tab_key

As far as displaying them visually goes, that would be a different feature and is unrelated to this issue. Some people expect that the tab character would have text decoration in this particular scenario, as if it was part of the same span as the rest of the text, at least that's what something like Google Docs or Word (the app, not the web version) would do.

I'm not aware of an HTML based editor that can display tabs with text decoration, so I don't have an example to point you towards, but this also isn't an issue that I've researched myself beyond reviewing this PR.

packages/lexical/src/nodes/LexicalTabNode.ts
```diff
@@ -47,6 +48,12 @@ export class TabNode extends TextNode {
     return null;
   }
 
+  createDOM(config: EditorConfig): HTMLElement {
+    const dom = super.createDOM(config);
+    dom.classList.add('PlaygroundEditorTheme__tabNode');
```
etrepum, line 53: Since this is not a playground-only node, this class needs to come from the theme and not be hardcoded.

vantage-ola: ok, what do you mean by not hardcoded?

etrepum:
```ts
addClassNamesToElement(dom, config.theme.tab);
```

In addition you would need to modify the `EditorThemeClasses` type and add the property `tab: 'PlaygroundEditorTheme__tabNode'` to PlaygroundEditorTheme.ts

vantage-ola:
@etrepum can you please take a look at this pr, thanks 🙏

etrepum (commented):
The e2e test suite is failing, both in CI here and I can reproduce the failures locally.

```
  5 failed
    [chromium] › packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs:319:3 › CodeBlock › Can maintain indent when creating new lines 
    [chromium] › packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs:407:3 › CodeBlock › Can (un)indent multiple lines at once 
    [chromium] › packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs:891:3 › CodeBlock › When pressing CMD/Ctrl + Left, CMD/Ctrl + Right, the cursor should go to the start of the code 
    [chromium] › packages/lexical-playground/__tests__/e2e/Tab.spec.mjs:22:3 › Tab › can tab + IME ─
    [chromium] › packages/lexical-playground/__tests__/e2e/Tab.spec.mjs:92:3 › Tab › can tab inside code block #4399 
  21 skipped
  478 passed (6.4m)
```

You can run these suites locally by having `npm run start` in one terminal and then in another:

```
npm run test-e2e-chromium
```

To run a particular test or suite you can specify another argument, e.g.

```
npm run test-e2e-chromium packages/lexical-playground/__tests__/e2e/CodeBlock.spec.mjs:319:3
```

etrepum:
I think most of these tests failures probably just mean that you have to change the tests to expect the tabNode class, but they need to be fixed nonetheless.

vantage-ola:
ok, thank you... will take a look at them

vantage-ola:
i think its good to go now

packages/lexical/package.json
```diff
@@ -35,5 +35,8 @@
         "default": "./Lexical.js"
       }
     }
+  },
+  "dependencies": {
+    "@lexical/utils": "0.21.0"
```
etrepum, line 40: This change shouldn't be here, lexical shouldn't depend on @lexical/utils

```suggestion
```

packages/lexical/src/nodes/LexicalTabNode.ts
```diff
@@ -8,9 +8,11 @@
 
 import type {DOMConversionMap, NodeKey} from '../LexicalNode';
 
+import {addClassNamesToElement} from '@lexical/utils';
```
etrepum, line 11: Because this is in the core lexical package it shouldn't use that import from @lexical/utils because it would be a circular dependency. There is a similar private `getCachedClassNameArray` function in `'./LexicalUtils'` which can be imported and used, check out the implementation of ParagraphNode or TextNode to see how to use it.

etrepum (commented):
Other than the circular dependency I think this is good enough, there are still some slight difference in how the text is rendered which are clear especially when a background color is added but that's just a matter of tweaking the CSS accordingly. Since this CSS is in the playground and not part of the package itself I think that it's acceptable.

[Playground](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE-2UQUvcUBSF_4rcbqNMRqo0XRbanRTsru0iTt5oaMyTN0_bQfzvJWNBK4Ii0rl951vJDIzJOef77rWFrs8xHec2B2uuLcWYp7-Ls37oUhit-frgQxdy2w_WzCpbxnTeZmvqeWXnsQvW2Dh9NVhlq7wepi9O2sWP0xQvx253EYeYmp03h6Hbn9fvrbIcfmVr7Et7Mn1YX0w_2HxX2VVIqz6O1tQ31d1D56_30G_53jM3L_D4I18x59H0W4WgDFpYUKVBlbJKBGXQwoIqDaqUVSIogxYWVGlQpawSQRm0sKBKgypllQjKoIUFVRpUKatEUAYtLKjSoEpZJYIyaGFBlQZVyioRlEELC6o0qFJWiaAMWlhQpUGVskoEZdDCgioNqpRVIiiDFhZUaVClrBJBGbSwoEqDKmWVCMqghQVVGlQpq0RQBi0sqNKgSlklgjJoYUGVBlXKKhGUQQsLqjSoUlaJoAxaWFClQZWySgRl0MKCKg2qlFUiKIMWFlRp0LKzLuOYd5fteT-sm51PIabTvt0S0U-8ipdK_iEP_0kjQAIkz4KEVlAHdYAESLivqIM6ThoBEiDhvqIO6gAJkHBfnbVCJagDJEDCfUUd1AESD5UACfcVdVAHSICE-4o6jioBEiABEu4r6qAOkACJk0a4r6iDOkACJNxX1EEdR40ACZBwX1EHdYAESLivzlqhEtQBEiDhvqIO6gCJh0qAhPuKOqgDJEDCfUUdR5UACZAACfcVdVAHSIDESSPcV9RBHSABEu4r6qCOo0aABEi4r894jw_xMvUh7RyFnx70efx1PFWzLWD8NwM0QPMiaGgHpVAKaICGO-y3HapBKaABGu4wSjmsBmiABmi4wyiFUj6bARqg4Q6jFEoBjdNqgIY7jFIoBTRA47MZ7jBKoRTQAM32oaEdlEIpoAEa7rDfdqgGpYAGaLjDKOWwGqABGqDhDqMUSvlsBmiAhjuMUigFNE6rARruMEqhFNAAjc9muMMohVJAAzTbh4Z2UAqlgAZouMN-26EalAIaoOEOo5TDaoAGaICGO4xSKOWzGaABGu4wSqEU0DitBmi4wyiFUkADND6b4Q6jFEoBDdBsHxraQSmUAhqg4Q77bYdqUApogGa70HyvrOtTWOTNZxtysru3MqusH7sw5s3L_vknF21qT1N7cfbXf7p94sd7eabPx09msJe8QorxQY6byoZ2lY_bq9BZUx_u7x_U7-rZ2_nsoLJVvEyL6Xefh3Z9-xr3f22zvXm9N7Ob3zne9wsAQwEA)

Screenshot 2024-12-07 at 16 53 37

This PR

Screenshot 2024-12-07 at 16 53 28

packages/lexical/package.json
vantage-ola: ok. will delete that asap.

packages/lexical/src/nodes/LexicalTabNode.ts
vantage-ola: Ok, will do that.

package-lock.json
```diff
@@ -38927,7 +38927,10 @@
     },
     "packages/lexical": {
       "version": "0.21.0",
-      "license": "MIT"
+      "license": "MIT",
+      "dependencies": {
+        "@lexical/utils": "^0.21.0"
+      }
```
etrepum, line 38933: This file shouldn't have changed, maybe you need to run `npm install` to undo it? Or just revert it directly.

packages/lexical/src/nodes/LexicalTabNode.ts
etrepum: This code won't work, you still need to change it to not depend on @lexical/utils

etrepum (commented):
There are still some issues with the integrity check, maybe because package-lock.json is still changed unnecessarily, or maybe because there is still a dependency on @lexical/utils in LexicalTabNode.ts. Once you fix it, `npm run ci-check` should pass locally as well.

vantage-ola:
the strikethrough, underline or both  doesnt align when the font is different, i dont think there is a way the css can accomodate all of the fonts and they will be consistent across the playground editor(i may be wrong). 
Screenshot 2024-12-09 154408

package-lock.json
vantage-ola: npm install fixed it i think

etrepum (approved):

PR #6814, opened 2024-11-10
Fix: Aria attributes for ContentEditable are ignored

### LINKED ISSUE #6789: Bug: Aria attributes for ContentEditable are ignored
When adding aria-labelledby to the ContentEditable it is not visible in the dom.

<ContentEditable
            className={'ContentEditable__root'}
            aria-labelledby='TEST'
   />

vantage-ola:
## Description
Fix Aria attributes for ContentEditable element

Closes #6789

### Before

When adding aria-labelledby to the ContentEditable it is not visible in the dom.

```
<ContentEditable
className={'ContentEditable__root'}
aria-labelledby='TEST'
/>

```

### After

The aria-labelledby is visible in the the dom. you can check it out by inspecting element in browser DevTools to verify the ARIA attributes...

## TODO
Write a unit test to verify the attributes are properly rendered in the DOM

### DIFF
--- a/packages/lexical-react/src/shared/LexicalContentEditableElement.tsx
+++ b/packages/lexical-react/src/shared/LexicalContentEditableElement.tsx
@@ -88,7 +88,6 @@ function ContentEditableElementImpl(
 
   return (
     <div
-      {...rest}
       aria-activedescendant={isEditable ? ariaActiveDescendant : undefined}
       aria-autocomplete={isEditable ? ariaAutoComplete : 'none'}
       aria-controls={isEditable ? ariaControls : undefined}
@@ -118,6 +117,7 @@ function ContentEditableElementImpl(
       spellCheck={spellCheck}
       style={style}
       tabIndex={tabIndex}
+      {...rest}
     />
   );
 }


vantage-ola:
I dont know  if the tests are necessary... but while writing the tests,  If i wanted to write test cases where the isEditable is `false`, but I couldnt figure it out how to get it to work...

packages/lexical-react/src/__tests__/unit/LexicalContentEditableElement.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+    // Mock LexicalEditor
+    mockEditor = {
+      isEditable: jest.fn(() => true),
+      registerEditableListener: jest.fn((cb) => {
+        cb(true || false); // Simulate editable state
+        return jest.fn(); // Mock unregister function
+      }),
+      setRootElement: jest.fn(),
+    } as unknown as LexicalEditor;
```
etrepum, line 34: We don't use mocks to test the editor, you can look at other tests to see how we set up an editor for the tests

packages/lexical-react/src/__tests__/unit/LexicalContentEditableElement.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+    }
+    jest.restoreAllMocks(); // Restore mocks after each test
+  });
+
+  it('renders the correct ARIA attributes when editable', async () => {
+    await ReactTestUtils.act(async () => {
+      reactRoot.render(
+        <ContentEditableElement
+          editor={mockEditor}
+          ariaLabelledBy="test-label"
```
etrepum, line 53: I think these tests would all work as-is, without your changes, because they are using the renamed aria properties. We also need to test the cases described in the issue that you're trying to fix.

etrepum (approved):

PR #6817, opened 2024-11-12
[lexical-link] Test:  Removing link from node(children)

### LINKED ISSUE #6665: Add test for removing link from a node
removing link from a text node works, but not when the node has children (like mark node). This PR fixes it https://github.com/facebook/lexical/pull/6656/files. whats left is to add test

vantage-ola:
## Description
Test for removing link from a text node, when the node has children (like mark node)

Closes #6665

### DIFF
--- a/packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
+++ b/packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
@@ -13,8 +13,10 @@ import {
   LinkNode,
   SerializedLinkNode,
 } from '@lexical/link';
-import {MarkNode} from '@lexical/mark';
+import {$createMarkNode, $isMarkNode} from '@lexical/mark';
 import {
+  $createParagraphNode,
+  $createTextNode,
   $getRoot,
   $selectAll,
   ParagraphNode,
@@ -414,9 +416,9 @@ describe('LexicalLinkNode tests', () => {
     test('$toggleLink correctly removes link when textnode has children(like marknode)', async () => {
       const {editor} = testEnv;
       await editor.update(() => {
-        const paragraph = new ParagraphNode();
-        const precedingText = new TextNode('some '); // space after
-        const textNode = new TextNode('text');
+        const paragraph = $createParagraphNode();
+        const precedingText = $createTextNode('some '); // space after
+        const textNode = $createTextNode('text');
 
         paragraph.append(precedingText, textNode);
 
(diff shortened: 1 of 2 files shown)


packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -409,5 +410,81 @@ describe('LexicalLinkNode tests', () => {
...
+                type: 'text',
+              },
+            ],
+            ids: ['knetk'],
+            type: 'mark',
+          },
+        ],
+        type: 'paragraph',
+      });
+    });
```
etrepum, line 488: I think it would be better for these tests to inspect the document (e.g. with `editor.read`) rather than the JSON, since the test itself is not about JSON serialization

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -409,5 +410,81 @@ describe('LexicalLinkNode tests', () => {
...
+
+        paragraph.append(precedingText, textNode);
+
+        const linkNode = $createLinkNode('https://example.com/foo', {
+          rel: 'noreferrer',
+        });
+        textNode.insertAfter(linkNode);
+        linkNode.append(textNode);
+
+        const markNode = new MarkNode(['knetk']);
```
etrepum, line 429: always use `$create*` such as `$createTextNode`, `$createParagraphNode`, etc. to create nodes. The `new` operator should never be used directly except when implementing the clone method and the `$create` function.

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
vantage-ola: okay. fair enough. I would change that.

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
vantage-ola: ok. i had no idea. will work on that.

vantage-ola:
overall, apart from those suggested fixes. is the approach good to go?

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -409,5 +412,70 @@ describe('LexicalLinkNode tests', () => {
...
+        textNode.insertAfter(linkNode);
+        linkNode.append(textNode);
+
+        const markNode = $createMarkNode(['knetk']);
+        textNode.insertBefore(markNode);
+        markNode.append(textNode);
+        $getRoot().append(paragraph);
+      });
+
+      await editor.read(() => {
```
etrepum, line 437: No need to await for reads, nothing asynchronous is happening here

```suggestion
      editor.read(() => {
```

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -409,5 +412,70 @@ describe('LexicalLinkNode tests', () => {
...
+        }
+      });
+
+      await editor.update(() => {
+        $selectAll();
+        $toggleLink(null);
+      });
+
+      // Verify structure after link removal
+      await editor.read(() => {
```
etrepum, line 465:
```suggestion
      editor.read(() => {
```

etrepum (dismissed):

etrepum (approved):

Issue #5841, opened 2024-04-06
Bug: Composing Korean is not correctly handled on IOS


2wheeh:
Lexical version: 0.14.3 (the latest)

### on IOS:

Korean Keyboard inputs don't emit `CompositionEvents` (any of `compositionstart`, `compositionend`, `compositionupdate`) keeping `isComposing` as `false`, which means composition cycle for Korean is within just one key input. That’s same context as `keyCode` is not `229` on composing Korean on IOS. (Japanese, Chinese seem to act like in other OS)

Instead, they are handled in `onBeforeInput` with `deleteContentBackward` and `insertText` of `InputType`. When we type `ㄱ` => `가`, `ㄱ` is deleted and `가` is inserted using `beforeinput` events. While doing this, it clears composing, which is not accessible due to above reason, at least I don't know what API can help.

on BACKSPACE key input, its default behavior would be clearing composing and deleting the character. But it reaches `KEY_BACKSPACE_COMMAND` handlers (ref: [@lexical/rich-text](https://github.com/facebook/lexical/blob/4c30b5215f0269c069c6c01f5eea973163a77321/packages/lexical-rich-text/src/index.ts#L834), [@lexical/plain-text](https://github.com/facebook/lexical/blob/4c30b5215f0269c069c6c01f5eea973163a77321/packages/lexical-plain-text/src/index.ts#L269)) on `keyDown` where we prevent default and call deletion handler. Here we delete character but do not clear composing.

So this leads to push remnants when next key inputs come: 
- #3305

And since it's not handled by Lexical's composition event handlers (`onCompositionStart`, `onCompositionEndEmpl`, ...) there can be some other effects. 

i.e. it doesn't show the lying caret (?) for composing character and `LexicalEditor._compositionKey` is not utilized.

It's not easy to see all the effects since we don't have `e2e-ios` tests !

Link to code example: can reproduce in playground

## The current behavior

https://github.com/facebook/lexical/assets/40269597/e8883eb9-9ca3-479a-b970-376ca0ac8e74

## The expected behavior

Korean composition on IOS works well as much as with native browser contenteditable.

mayrang:
[...]

PR #6819, opened 2024-11-12
[lexical-rich-text][lexical-plain-text] workaround for Korean IME issue on iOS

2wheeh:
## Description

related to #5841, #3305.
based on #4814,

- adds follow-up commits to resolve side effect caused by blocking `isCollapsed` handling.
- adds same workaround to plain text plugin.

while we don't have `e2e-ios` tests, I checked the fix without `if (navigator.language === 'ko-KR' && IS_IOS)` passed tests on my local.

please include @Deadintegral as a co-author since it contains his work. 

Closes #4814

### Before

https://github.com/user-attachments/assets/c9483f3c-a8b9-4aa9-96b3-663d46d3d2e9

### After

https://github.com/user-attachments/assets/c294d341-bac4-4bed-af6a-15e094e33372

### DIFF
--- a/packages/lexical-plain-text/src/index.ts
+++ b/packages/lexical-plain-text/src/index.ts
@@ -274,6 +274,12 @@ export function registerPlainText(editor: LexicalEditor): () => void {
           return false;
         }
 
+        // Exception handling for iOS native behavior instead of Lexical's behavior when using Korean on iOS devices.
+        // more details - https://github.com/facebook/lexical/issues/5841
+        if (IS_IOS && navigator.language === 'ko-KR') {
+          return false;
+        }
+
         event.preventDefault();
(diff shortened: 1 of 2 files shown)


etrepum (dismissed):
This seems pretty low risk given the isolation to just iOS+Korean, I don't have any experience with IMEs so I can't really provide any guidance for what a proper fix would be and I am not even sure I could test it very easily (I couldn't get the korean IME to work in the iOS simulator, it never combines characters for me, and on my phone I don't think it would be very easy for me to have navigator.langauge set to korean)

etrepum (approved):

etrepum:
The only failing tests is a very flaky one (that is improved in #6759) so I'm going to merge

PR #6826, opened 2024-11-13
[lexical-link] Test: Appending inline element nodes to ListNode 

### LINKED ISSUE #6815: PRs with test-needed label could do with more test coverage
https://github.com/facebook/lexical/pulls?q=is%3Apr+label%3Atest-needed+is%3Aclosed

This is a list of PRs that were merged, but could do with more test coverage.

Shopiley:
## Description
Unit test for the recently updated ListNode.append() method, to verify that inline nodes (such as LinkNode) are wrapped in a ListItemNode without being converted to TextNode.

## Test Plan
**Test**: Adds a test case to append a LinkNode to a ListNode:

**Assertions**: 
- LinkNode is correctly wrapped as a ListItemNode in the ListNode.
- LinkNode retains its original type and is not converted to TextNode.
- The URL value in LinkNode remains unaltered.

Closes #6815

### DIFF
--- a/packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
+++ b/packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
@@ -277,23 +277,25 @@ describe('LexicalListNode tests', () => {
       editor.read(() => {
         const root = $getRoot();
 
-        const listNode = root.getFirstChild(); // returns type LexicalNode instead of ListNode
+        const listNode = root.getFirstChild();
         expect(listNode).not.toBeNull();
-        expect($isListNode(listNode)).toBe(true); // passes this assertion
+        expect($isListNode(listNode)).toBe(true);
 
-        const firstChild = listNode?.getFirstChild(); // Property 'getFirstChild' does not exist on type 'LexicalNode'
-        expect($isListItemNode(firstChild)).toBe(true);
-
-        if ($isListItemNode(firstChild)) {
-          const wrappedNode = firstChild?.getFirstChild();
-          expect(wrappedNode).not.toBeNull();
-          expect($isLinkNode(wrappedNode)).toBe(true);
-
-          expect((wrappedNode as LinkNode).getURL()).toBe(
-            'https://lexical.dev/',
-          );
-        } else {
+        if ($isListNode(listNode)) {
+          const firstChild = listNode.getFirstChild();
           expect($isListItemNode(firstChild)).toBe(true);
+
+          if ($isListItemNode(firstChild)) {
+            const wrappedNode = firstChild?.getFirstChild();
+            expect(wrappedNode).not.toBeNull();
+            expect($isLinkNode(wrappedNode)).toBe(true);
+
+            expect((wrappedNode as LinkNode).getURL()).toBe(
+              'https://lexical.dev/',
+            );
+          } else {
+            expect($isListItemNode(firstChild)).toBe(true);
+          }
         }
       });
     });
--- a/packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
+++ b/packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-import {$isLinkNode, LinkNode} from '@lexical/link';
+import {$createLinkNode, $isLinkNode, LinkNode} from '@lexical/link';
 import {$getRoot, ParagraphNode, TextNode} from 'lexical';
 import {initializeUnitTest} from 'lexical/src/__tests__/utils';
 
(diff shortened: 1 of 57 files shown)


packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
```diff
@@ -260,6 +261,31 @@ describe('LexicalListNode tests', () => {
       });
     });
 
+    test('ListNode.append() should wrap an InlineNode in a ListItemNode without converting it to TextNode', async () => {
+      const {editor} = testEnv;
+
+      await editor.update(() => {
```
etrepum, line 267: Generally speaking I think the right thing to do would be use update to create the nodes and insert them into the document, and then check to make sure that the document has the expected layout after the update (in a separate `editor.read(() => {…})` for example. This way we can be sure that the expected behavior persists after normalization, transforms and reconciliation.

packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
```diff
@@ -260,6 +261,31 @@ describe('LexicalListNode tests', () => {
...
+        if ($isListItemNode(firstChild)) {
+          const wrappedNode = firstChild?.getFirstChild();
+          expect($isLinkNode(wrappedNode)).toBe(true);
+
+          expect((wrappedNode as LinkNode).getURL()).toBe(
+            'https://lexical.dev/',
+          );
+        } else {
+          throw new Error('First child is not a ListItemNode');
+        }
```
etrepum, line 285: It would be better to fail with an expect, e.g. `expect($isListItemNode(firstChild)).toBe(true)`

packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
Shopiley: I tried using the update function to create the nodes. Then appended them to the root node of the document. In a separate read, I tried to check the layout by unpacking the nodes and verifying their types starting from the list node. But I realized that calling the `getFirstChild()` function on the root node returns a node of type LexicalNode instead of ListNode and somehow, it passes the `$isListNode()` check but errors out saying `Property 'getFirstChild' does not exist on type 'LexicalNode'` when I try to call `getFirstChild()` on the supposed ListNode.

`
test('ListNode.append() should wrap an InlineNode in a ListItemNode without converting it to TextNode', async () => {
      const {editor} = testEnv;

      await editor.update(() => {
        const listNode = new ListNode('bullet', 1);
        const linkNode = new LinkNode('https://lexical.dev/');

        listNode.append(linkNode);

        const root = $getRoot();
        root.append(listNode);
      });

      editor.read(() => {
        const root = $getRoot();

        const listNode = root.getFirstChild(); // returns type LexicalNode instead of ListNode
        expect(listNode).not.toBeNull();
        expect($isListNode(listNode)).toBe(true); // passes this assertion

        const firstChild = listNode?.getFirstChild(); // Property 'getFirstChild' does not exist on type 'LexicalNode'
        expect($isListItemNode(firstChild)).toBe(true);

        if ($isListItemNode(firstChild)) {
          const wrappedNode = firstChild?.getFirstChild();
          expect(wrappedNode).not.toBeNull();
          expect($isLinkNode(wrappedNode)).toBe(true);

          expect((wrappedNode as LinkNode).getURL()).toBe(
            'https://lexical.dev/',
          );
        } else {
          expect($isListItemNode(firstChild)).toBe(true);
        }
      });
    });
`

etrepum: This is normal, typescript can't use jest expect as an assertion. You can use an if statement, some other kind of assertion, or a cast, to ensure the type is checked.

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/41179

Shopiley: thanks! this was helpful. Made a new commit.

packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
```diff
@@ -260,6 +261,45 @@ describe('LexicalListNode tests', () => {
       });
     });
 
+    test('ListNode.append() should wrap an InlineNode in a ListItemNode without converting it to TextNode', async () => {
+      const {editor} = testEnv;
+
+      await editor.update(() => {
+        const listNode = new ListNode('bullet', 1);
+        const linkNode = new LinkNode('https://lexical.dev/');
```
etrepum, line 269: Nodes should never be created directly with the constructor except in the implementation of `clone` and the `$create*` function. Only functions such as `$createLinkNode` should be used elsewhere.

Shopiley: got it, done.

etrepum (approved):

PR #6827, opened 2024-11-13
[lexical-examples] Chore: Add DOMExportOutputMap type to the exportMap

### LINKED ISSUE #6825: [lexical-examples]: Export DOMExportOutputMap type from Lexical core for html configuration use case
Lexical version: V0.20.0

## Steps To Reproduce

While setting up a custom `html` export configuration in Lexical, I created an example for `DOMExportOutputMap`.
Currently, `DOMExportOutputMap` is not exported from the Lexical core, requiring the creation of a custom type to use this configuration in `LexicalComposer`.

## The current behavior
To use the `html` export configuration in `LexicalComposer`, users must manually create a custom type definition for `DOMExportOutputMap`. This is due to the type not being available for direct import from the Lexical core.
## The expected behavior
`DOMExportOutputMap` should be directly exportable from the Lexical core, enabling users to import it without needing to define custom types for common configurations.

## Impact of fix

Low severity
Promotes code reusability and aligns with the DRY (Don't Repeat Yourself) principle by reducing redundant type definitions.

Kingscliq:
## Description

To use the html export configuration in LexicalComposer, users must manually create a custom type definition for DOMExportOutputMap. This is due to the type not being available for direct import from the Lexical core.

Closes #6825 

## Test plan

DOMExportOutputMap should be directly exportable from the Lexical core, enabling users to import it without needing to define custom types for common configurations.

### DIFF
--- a/examples/react-rich/src/App.tsx
+++ b/examples/react-rich/src/App.tsx
@@ -16,6 +16,7 @@ import {
   $isTextNode,
   DOMConversionMap,
   DOMExportOutput,
+  DOMExportOutputMap,
   Klass,
   LexicalEditor,
   LexicalNode,
@@ -53,7 +54,7 @@ const removeStylesExportDOM = (
   return output;
 };
 
-const exportMap = new Map<
+const exportMap: DOMExportOutputMap = new Map<
   Klass<LexicalNode>,
   (editor: LexicalEditor, target: LexicalNode) => DOMExportOutput
 >([
--- a/examples/react-rich/src/App.tsx
+++ b/examples/react-rich/src/App.tsx
@@ -54,7 +54,10 @@ const removeStylesExportDOM = (
   return output;
 };
 
-const exportMap: DOMExportOutputMap = new Map([
+const exportMap: DOMExportOutputMap = new Map<
+  Klass<LexicalNode>,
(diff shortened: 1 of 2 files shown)


etrepum (approved):

examples/react-rich/src/App.tsx
```diff
@@ -53,7 +54,7 @@ const removeStylesExportDOM = (
   return output;
 };
 
-const exportMap = new Map<
+const exportMap: DOMExportOutputMap = new Map<
   Klass<LexicalNode>,
   (editor: LexicalEditor, target: LexicalNode) => DOMExportOutput
 >([
```
etrepum, line 60: This type of the value can be inferred from the explicit type of the variable now, so it can be even simpler than this:

```suggestion
const exportMap: DOMExportOutputMap = new Map([
```

etrepum (approved):

examples/react-rich/src/App.tsx
etrepum: Nevermind, it looks like the Klass<LexicalNode> is still required, so I'm reverting that suggestion.

etrepum (approved):

PR #6834, opened 2024-11-14
[lexical-table] Bug Fix: Resolve table selection issue when the mouse crosses over a portal

### LINKED ISSUE #6820: Bug: [lexical-table] Dragging over the table-cell-action-button interferes with selection
Lexical version: 0.20

## Steps To Reproduce

1. Create a table
2. Focus a cell so that the table-cell-action-button shows
3. Drag to create a selection but make sure the mouse moves over the table-cell-action-button

## The current behavior

https://github.com/user-attachments/assets/186d58f0-d870-4c51-af25-3f07f874f40f

## The expected behavior

Selection behaves as if the button was not there

## Impact of fix

Confusing behavior for anyone using the TableActionMenuPlugin (currently playground-only, but many people just copy this code).

etrepum:
## Description

When using table selection by mouse movement, if the mouse crosses over an element that is outside of the document, it ends up creating a native selection that messes up the intended table selection.

Unfortunately when researching this bug I accidentally learned how much of a tangled un-idiomatic mess this plugin is and I couldn't figure out how to fix it without a larger refactoring. The primary anti-patterns are basically nested and cascading updates.

For whatever historical reason, lexical uses the most error-prone design imaginable for nested updates, so they should be avoided at all cost (surprisingly, `dispatchCommand` avoids this issue by working in the manner that nested updates should have been implemented). TableObserver's methods are primarily used inside updates, but because sometimes they are called from mouse events they all have an explicit `editor.update` which is a mistake. I've renamed them all to `$methods` to be explicit on their dependencies and how to use them.

By doing this in a way that actually makes sense for Lexical, we will probably fix some bugs we don't know about yet, remove the cascading updates, and remove the setTimeout nonsense that was used to work around the nested and cascading updates.

This is not a complete refactor, just enough to get things working. I'm sure more logic could be cleaned up here, especially regarding nested tables.

Closes #6820
Closes #6861 (the code from that PR is also here)

### Before

https://github.com/user-attachments/assets/186d58f0-d870-4c51-af25-3f07f874f40f

### After

https://github.com/user-attachments/assets/4f175d4a-a710-4937-befb-4d17db769100

### DIFF
--- a/examples/react-rich/package-lock.json
+++ b/examples/react-rich/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "@lexical/react-rich-example",
-  "version": "0.15.0",
+  "version": "0.20.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@lexical/react-rich-example",
-      "version": "0.15.0",
+      "version": "0.20.0",
       "dependencies": {
-        "@lexical/react": "0.15.0",
-        "lexical": "0.15.0",
+        "@lexical/react": "0.20.0",
+        "lexical": "0.20.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0"
       },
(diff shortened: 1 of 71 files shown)


ivailop7 (approved):
Fantastic work! Really well done, Bob!

PR #6836, opened 2024-11-15
[lexical-playground] Bug Fix: Replace empty <img> tags with invisible <span> elements to prevent Safari issue

### LINKED ISSUE #6824: Bug: empty images around equations render as question marks on Safari
Lexical version: 0.20.0

## Steps To Reproduce

1. Open https://playground.lexical.dev/ in Safari
2. Add equation

## The current behavior

The equation shows up with question marks around it. The cause is the empty image tags that are added around the Katex equation, our current workaround is to omit them on Safari.

Screenshot 2024-11-13 at 16 38 48

snikhil24:
## Description

- Current Behavior: Equations display with question marks around them in Safari due to empty  tags used for spacing around the KaTeX equation
- Changes: This PR replaces the empty  tags with invisible <span> elements for spacing, which resolves the issue in Safari while maintaining functionality across other browsers.

Closes #6824

### Before

Screenshot 2024-11-14 at 11 54 18 PM

### After

image

### DIFF
--- a/packages/lexical-playground/src/ui/KatexRenderer.tsx
+++ b/packages/lexical-playground/src/ui/KatexRenderer.tsx
@@ -41,14 +41,14 @@ export default function KatexRenderer({
     // inner text from Katex. There didn't seem to be any other way of making this work,
     // without having a physical space.
     <>
-      <img src="#" alt="" />
+      <span style={{display: 'inline-block', height: '1px', width: '1px'}} />
       <span
         role="button"
         tabIndex={-1}
         onDoubleClick={onDoubleClick}
         ref={katexElementRef}
       />
-      <img src="#" alt="" />
+      <span style={{display: 'inline-block', height: '1px', width: '1px'}} />
     </>
   );
 }


etrepum (commented):
Did you test this on Android? It sounds like this workaround is specifically for an Android related purpose.

etrepum:
See also #4040 #1968

snikhil24:
I've tested the changes on Android using a Pixel device in two ways:

- On an Actual Mobile Device: I accessed the application through the browser on my Pixel phone and confirmed that the equations rendered correctly.

- Via Browser Simulation: I also used the 'Toggle Device Toolbar' feature in Chrome DevTools to simulate a Pixel device and verified the behavior.

Both tests showed the issue is resolved, and the equations render as expected with the invisible <span> elements in place of the empty  tags.

Since I’m new to open source contribution, I’d appreciate your input on whether this is the correct approach for testing Android behavior.

etrepum:
In this particular case the reason the images are there are for input reasons rather than rendering reasons. I think your testing approach is correct, but it sounds like you were testing the wrong thing (rendering instead of input). Probably the best thing to do here would be to set CSS and/or an src on the img tags so that they don't show placeholders on Safari, which is less likely to cause an input regression on Android.

etrepum:
This PR doesn't have a signed CLA and the review comments have not been addressed

PR #6837, opened 2024-11-15
[lexical-react] Fix(lexical-react): ContentEditable props type rename

ivaaaaann:
## Description

The name of the type seems too generic, so I suggest renaming it to:

Like the link below
https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalTablePlugin.ts

### DIFF
--- a/packages/lexical-react/src/LexicalContentEditable.tsx
+++ b/packages/lexical-react/src/LexicalContentEditable.tsx
@@ -29,6 +29,15 @@ export type ContentEditableProps = Omit<ElementProps, 'editor'> &
       }
   );
 
+  /**
+ * @deprecated This type has been renamed to `ContentEditableProps` to provide a clearer and more descriptive name.
+ * For backward compatibility, this type is still exported as `Props`, but it is recommended to migrate to using `ContentEditableProps` instead.
+ *
+ * @note This alias is maintained for compatibility purposes but may be removed in future versions.
+ * Please update your codebase to use `ContentEditableProps` to ensure long-term maintainability.
+ */ 
+export type Props = ContentEditableProps  
+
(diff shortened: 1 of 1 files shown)


etrepum:
I think this probably makes sense but it doesn't maintain backwards compatibility. I would recommend also exporting it as Props, maybe with a `@deprecated` docstring

ivaaaaann:
@etrepum Thanks for the nice comment, I didn't think about that part! I modified it according to your comment.

packages/lexical-react/src/LexicalContentEditable.tsx
```diff
@@ -29,10 +29,19 @@ export type Props = Omit<ElementProps, 'editor'> &
...
   );
 
+  /**
+ * @deprecated This type has been renamed to `ContentEditableProps` to provide a clearer and more descriptive name.
+ * For backward compatibility, this type is still exported as `Props`, but it is recommended to migrate to using `ContentEditableProps` instead.
+ *
+ * @note This alias is maintained for compatibility purposes but may be removed in future versions.
+ * Please update your codebase to use `ContentEditableProps` to ensure long-term maintainability.
+ */ 
+export type Props = ContentEditableProps
```
etrepum, line 39: I think the formatting here is causing CI to fail, `npm run prettier:fix` would resolve it

ivaaaaann: thank you i fixed [fix: prettier](https://github.com/facebook/lexical/pull/6837/commits/d832fc2a0028deebe9f43ae1f5418bb5b244beb7)

etrepum (approved):

Issue #6851, opened 2024-11-21
Bug: key backward cause ImageNode delete unexpected


emsurezhang:
<img width="1175" alt="截屏2024-11-21 21 01 37" src="https://github.com/user-attachments/assets/12f4ce72-684c-4be7-a68f-02f3d2b9387c">
In the playground, I insert an ImageNode and press enter to start a new ParagraphNode.
The cursor is at ParagraphNode which key is 4.
Now I press backward and expect only the ParagraphNode key 4 is deleted. But the result is the ImageNode is deleted too.
<img width="1129" alt="截屏2024-11-21 21 07 09" src="https://github.com/user-attachments/assets/94db724a-8ca7-4cf6-8753-245fc7e36cb1">

emsurezhang:
In LexicalSelection.ts removeText() function, I simply remove the node.remove line, it worked fine.

```
/**
   * Removes the text in the Selection, adjusting the EditorState accordingly.
   */
  removeText(): void {
    if (this.isCollapsed()) {
      return;
    }
    const {anchor, focus} = this;
    const selectedNodes = this.getNodes();
    const firstPoint = this.isBackward() ? focus : anchor;
    const lastPoint = this.isBackward() ? anchor : focus;
    let firstNode = firstPoint.getNode();
    let lastNode = lastPoint.getNode();
    const firstBlock = $getAncestor(firstNode, INTERNAL_$isBlock);
    const lastBlock = $getAncestor(lastNode, INTERNAL_$isBlock);
    // If a token is partially selected then move the selection to cover the whole selection
    if (
      $isTextNode(firstNode) &&
      firstNode.isToken() &&
      firstPoint.offset < firstNode.getTextContentSize()
    ) {
      firstPoint.offset = 0;
    }
    if (lastPoint.offset > 0 && $isTextNode(lastNode) && lastNode.isToken()) {
      lastPoint.offset = lastNode.getTextContentSize();
    }

    selectedNodes.forEach((node) => {
      if (
        !$hasAncestor(firstNode, node) &&
        !$hasAncestor(lastNode, node) &&
        node.getKey() !== firstNode.getKey() &&
        node.getKey() !== lastNode.getKey()
      ) {
        console.log("remove node", node.getKey())
        **// node.remove();**
      }
    });

    const fixText = (node: TextNode, del: number) => {
      if (node.getTextContent() === '') {
        node.remove();
      } else if (del !== 0 && $isTokenOrSegmented(node)) {
[...]

PR #6854, opened 2024-11-21
fix: bug-6851

emsurezhang:
https://github.com/facebook/lexical/issues/6851

update:
in LexicalSelection.ts file, change the native selection move mode from "extend" to "move", cause isCollapsed is true.
at line 1781:
this.modify('extend', isBackward, 'character'); => this.modify('move', isBackward, 'character');

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1778,7 +1778,7 @@ export class RangeSelection implements BaseSelection {
         possibleNode.selectStart();
         return;
       }
-      this.modify('extend', isBackward, 'character');
+      this.modify('move', isBackward, 'character');
 
       if (!this.isCollapsed()) {
         const focusNode = focus.type === 'text' ? focus.getNode() : null;


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1778,7 +1778,7 @@ export class RangeSelection implements BaseSelection {
         possibleNode.selectStart();
         return;
       }
-      this.modify('extend', isBackward, 'character');
+      this.modify('move', isBackward, 'character');
```
etrepum, line 1781: I don't think this change is correct but we can run the tests and see how many will fail. I haven't looked at this code closely recently but I believe the extend is intentional, to grow a collapsed selection by one character and then delete the contents of that selection.

etrepum: Here's a link to the unit test failure report: https://github.com/facebook/lexical/actions/runs/11956172962/job/33332485265?pr=6854

etrepum:
I'm going to close this as stale, since this change does not include tests and causes regressions elsewhere. Please feel free to re-open or create a new PR when you have something that works without regressions.

PR #6864, opened 2024-11-25
[lexical-utils] Bug Fix: Add feature detection to calculateZoomLevel

### LINKED ISSUE #6863: Refactor calculateZoomLevel to be an editor setting
Since Chrome 128 and https://chromestatus.com/feature/5198254868529152, the fix we did with calculateZoomLevel is no longer needed for later Chromium versions, however we now need a way to on/off the adjustment on demand for browser that have the flag On/Off or are on Chrome pre-128. Currently we have calculateZoomLevel at both lexical-core level and playground. We need a way to on/off it inside core, alongside playground.

etrepum:
## Description

Add feature detection to `calculateZoomLevel` in situations when we are not sure if the browser implements standardized CSS zoom. We still assume that Firefox does, but since only Chrome >= 128 implements it and Safari doesn't implement it yet we use feature detection (creating a div in the document but off-screen, measuring it with and without zoom, then removing it). It's measured twice in case there's a zoom on the body which could cancel out a static measurement.

Closes #6863

### Before

### After

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -660,7 +660,7 @@ function needsManualZoom(): boolean {
     document.body.appendChild(div);
     const noZoom = div.getBoundingClientRect();
     div.style.setProperty('zoom', '2');
-    NEEDS_MANUAL_ZOOM = div.getBoundingClientRect().width > noZoom.width;
+    NEEDS_MANUAL_ZOOM = div.getBoundingClientRect().width === noZoom.width;
     document.body.removeChild(div);
   }
   return NEEDS_MANUAL_ZOOM;
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -647,19 +647,38 @@ export function $insertFirst(parent: ElementNode, node: LexicalNode): void {
   }
 }
 
+let NEEDS_MANUAL_ZOOM = IS_FIREFOX || !CAN_USE_DOM ? false : undefined;
+function needsManualZoom(): boolean {
+  if (NEEDS_MANUAL_ZOOM === undefined) {
+    // If the browser implements standardized CSS zoom, then the client rect
+    // will be wider after zoom is applied
+    // https://chromestatus.com/feature/5198254868529152
+    // https://github.com/facebook/lexical/issues/6863
+    const div = document.createElement('div');
+    div.style.cssText =
+      'position: absolute; opacity: 0; width: 100px; left: -1000px;';
+    document.body.appendChild(div);
+    const noZoom = div.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


ivailop7:
This looks reasonable. Was originally thinking, since the method is used only in 3 places in 'lexical-react', if we have a way to get it out and have at playground level-only, but besides passing an extra argument to the 3 plugins (draggable block, checklist, context menu), dont have anything better. Chrome 128+ already has a pretty decent market share, so Ideally in 6-7 months we can remove the solution we come up with.

etrepum:
That's just the zoom CSS property and according to that table it's been supported in Chromium since 4 (2010). Running this code in Safari 18.1.1 (macOS) sets NEEDS_MANUAL_ZOOM to true where Firefox and Chromium set it to false.

```ts
    const div = document.createElement('div');
    div.style.cssText =
      'position: absolute; opacity: 0; width: 100px; left: -1000px;';
    document.body.appendChild(div);
    const noZoom = div.getBoundingClientRect();
    div.style.setProperty('zoom', '2');
    NEEDS_MANUAL_ZOOM = div.getBoundingClientRect().width === noZoom.width;
    document.body.removeChild(div);
```

etrepum:
Safari

Screenshot 2024-11-25 at 10 22 54

Chrome

Screenshot 2024-11-25 at 10 24 28

ivailop7:
> Safari
> 
> 
> 
> Screenshot 2024-11-25 at 10 22 54
> 
> 
> 
> Chrome
> 
> 
> 
> Screenshot 2024-11-25 at 10 24 28
> 
> 

Fair enough. I do remember trying it in WebKit a year+ ago and it worked. Thanks for checking.

etrepum:
We don't have any tests that cover this but I think it should be fine. I'm not impacted or interested enough to write e2e for this myself, but if anyone else wants to do that it would be appreciated.

potatowagon:
> We don't have any tests that cover this but I think it should be fine. I'm not impacted or interested enough to write e2e for this myself, but if anyone else wants to do that it would be appreciated.

have added tests-needed label

potatowagon (commented):
is this related to browser zoom? ie `ctrl` + `shift` + `+`/`-` to zoom in and out

ivailop7:
> is this related to browser zoom? ie `ctrl` + `shift` + `+`/`-` to zoom in and out

The browser-level zoom behaves correctly already. The problem this fixed is when you embed the editor in anything that uses CSS Zoom and the clickable regions would respect the browser-level zoom, but not the CSS zoom, despite resizing correctly, which is used for content reflow rather than transform, also the CSS Zoom didn't propagate to iframe tags prior to the Chromium-level fix.

ivailop7 (approved):
Appreciate the quick help on this!

PR #6869, opened 2024-11-26
Add Documentation for  RootNode's semantic and use case

### LINKED ISSUE #6404: Feature: Document RootNode semantics/special cases
## Description

There are some implicit assumptions about the RootNode that are not clearly spelled out in the documentation, for example:

* Can not be subclassed or used with node replacement (not extensible)
* Does not participate in mutation listeners (a root listener and/or update listener is sufficient for this)
* Can(?) be used with a node transform

Mostly the guiding principle here is that the RootNode is not really "part of the document", although in some cases it may "appear" to be (node transforms, serialization?)

## Impact

I think normally when people are trying to do these things with the RootNode they want document level metadata which may or may not be versioned along with the EditorState (e.g. support undo/redo). There just isn't any facility for that right now (maybe create some sort of shadow root under the root?). Documentation will at least help point them in the right direction (to come up with something on their own).

citruscai:
## Description
There are restrictions with RootNode that is not clearly laid out for the user. This PR expands more on that in documentation for Root Node. It should point user to write direction

Closes #6404

### DIFF
--- a/packages/lexical-website/docs/concepts/nodes.md
+++ b/packages/lexical-website/docs/concepts/nodes.md
@@ -25,6 +25,23 @@ There is only ever a single `RootNode` in an `EditorState` and it is always at t
 
 - To get the text content of the entire editor, you should use `rootNode.getTextContent()`.
 - To avoid selection issues, Lexical forbids insertion of text nodes directly into a `RootNode`.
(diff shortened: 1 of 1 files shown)


acywatson (approved):
This is a nice improvement. Thanks :)

PR #6872, opened 2024-11-26
Add SelectionAlwaysOnDisplay plugin

zurfyx:
This plugin already provides value in two different places internally, let's make it official

https://github.com/user-attachments/assets/5f61a947-6a95-4804-bddc-a2b77d353c1d


fantactuka (dismissed):

zurfyx:
I wonder if RetainSelection is a good/intuitive plugin name, given that we also have the (non-visible) selection concept

potatowagon (dismissed):

etrepum (approved):

etrepum:
The display isn't precisely the same as a normal selection (in chrome at least), but I think this provides a lot of value at least as a demonstration of the technique

https://github.com/user-attachments/assets/1e74ee77-7a07-4764-b37d-c0eeaa79d4d7

zurfyx:
> The display isn't precisely the same as a normal selection (in chrome at least), but I think this provides a lot of value at least as a demonstration of the technique
> 
>  retain-selection-rects.mov

I'd be happy to discuss this further in our next session. Ideally, this is an exact match of the actual browser representation, but my hunch is that browser provided rects is the closest the browser provides and the rest is on us which is why this is particularly hard problem.

A potential option to fix this problem is to always customize the selection render when this is enabled (https://github.com/facebook/lexical/commit/e934b9cfff64864e57ba873fd1bad031c4c17351). The trade-off is performance and the fact that we'd now trust selection to our built-in modules that are still experimental.

Overall, I feel like this is a small and reasonable trade-off but we can also offer the two variants and let the users decide. I can iterate on this on top, but then again, I'd love to hear your thoughts on this problem.

etrepum:
I'd probably look at the highlight API instead of trying to render it ourselves and probably provide some sort of straightforward way for decorators to get highlighted as well https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API

PR #6873, opened 2024-11-26
positionNodeOnRange fixes

zurfyx:
Some fixes/improvements over the existing functionality that ultimately powers RetainSelection (https://github.com/facebook/lexical/pull/6872).

1. Supports browser element selection
2. Adjustment to default styling
3. Fixes to positioning

Still works just like expected, general test on the other PR
Screenshot 2024-11-26 at 6 03 18 PM


packages/lexical-utils/src/positionNodeOnRange.ts
```diff
@@ -20,7 +20,27 @@ const mutationObserverConfig = {
   subtree: true,
 };
 
-export default function positionNodeOnRange(
+function prependDOMNode(parent: HTMLElement, node: HTMLElement) {
+  if (parent.hasChildNodes()) {
+    parent.insertBefore(node, parent.firstChild);
+  } else {
+    parent.append(node);
+  }
```
etrepum, line 28: null already means append, so you don't need to check it https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore#referencenode

```suggestion
  parent.insertBefore(node, parent.firstChild);
```

packages/lexical-utils/src/markSelection.ts
```diff
@@ -168,3 +160,14 @@ export default function markSelection(
     },
   );
 }
+
+function firstTextChild(node: Node): null | Text {
+  let current: Node | null = node;
+  while (current !== null) {
+    if (current instanceof Text) {
```
etrepum, line 167: I noticed that we have some code elsewhere that works with nodes from another document, it might make sense to check the nodeType instead of using `instanceof` (there's a lexical DOM_TEXT_TYPE constant for this) https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#node.text_node

zurfyx: Thanks for the callout! It turns out this same function is already implemented in LexicalUtils. I also revised another occurrence of `instanceof`

etrepum (approved):

PR #6874, opened 2024-11-27
Allow passing params to run-all

zurfyx:
This enables `npm run run-all -- Filename`


package.json
```diff
@@ -58,8 +58,8 @@
     "test-e2e-collab-ci-webkit": "npm run prepare-ci && cross-env E2E_PORT=4000 concurrently -k -s \"first\" \"npm run collab\" -P \"npm run test-e2e-collab-webkit -- {@}\"",
     "test-e2e-prod-ci-chromium": "npm run prepare-ci-prod && cross-env E2E_PORT=4000 npm run test-e2e-prod-chromium",
     "test-e2e-collab-prod-ci-chromium": " npm run prepare-ci-prod && cross-env E2E_PORT=4000 concurrently -k -s \"first\" \"npm run collab\" -P \"npm run test-e2e-collab-prod-chromium -- {@}\"",
-    "debug-run-all": "npm-run-all debug-test-e2e-*",
-    "run-all": "npm-run-all test-e2e-*",
+    "debug-run-all": "npm-run-all 'debug-test-e2e-* -- {1}' --",
+    "run-all": "npm-run-all 'test-e2e-* -- {1}' --",
```
etrepum, line 62: Given the names of the test suites, wouldn't this run both the CI and non-CI versions of these tests? Seems like some other refactoring should probably be done here

etrepum (approved):

package.json
zurfyx: @etrepum yes, you're right, definitely room for improvement

PR #6884, opened 2024-11-29
[lexical-table] Bug Fix: Fix scrollable table exportDOM

### LINKED ISSUE #6882: Scrollable tables do not copy content
Create any table, copy and paste on any other editor (so that it triggers HTML paste). It pastes blank

etrepum:
## Description

Fix TableNode.exportDOM regression when using scrollable tables

Closes #6882

## Test plan

I don't have time to do unit tests right now but this fixes it, if you wanted to release ASAP.

### Before

### After

### DIFF
--- a/packages/lexical-table/src/LexicalTableNode.ts
+++ b/packages/lexical-table/src/LexicalTableNode.ts
@@ -227,9 +227,12 @@ export class TableNode extends ElementNode {
   }
 
   exportDOM(editor: LexicalEditor): DOMExportOutput {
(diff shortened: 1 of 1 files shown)


zurfyx (approved):

etrepum:
The reason for this regression is because the exportDOM machinery (`$appendNodesToHTML`) doesn't currently have any support for getDOMSlot so the children ended up in the scrollable div instead of the table node since the result of createDOM was used directly as the element.

PR #6890, opened 2024-11-30
[lexical-website] Feature: Document the withKlass option for node replacement

### LINKED ISSUE #6398: Feature: Document the withKlass option for node replacement
## Description

When using node replacement, it's often useful to specify the `withKlass` option so that any behaviors that were intended for the original node type should work with the replaced node type (e.g. node transforms, mutation listeners, etc.).

## Impact

Anyone replacing built-in nodes with extended versions should probably be using this, but they wouldn't know that it's even an option unless they were looking at the types or the source code. Even then, it would be tricky to know what the semantics are unless you have a pretty thorough understanding of how the editor is constructed and how listeners work.

It may even be worth a lint and/or a warning when this is not specified.

Shopiley:
## Description
Currently, the withKlass option for node replacement is not documented, making it challenging for developers to use this functionality effectively.

This PR:
- Changes the concerned doc section header from **Node overrides** to **Node overrides / Node Replacements** as _node replacements_ is more in line with the terminology used in the code base.
- Includes an explanation of the arguments used for a node replacement type including the withKlass option.

Closes #6398 

### Before
No documentation available for withKlass.
### After
Updated documentation includes details for using withKlass in node replacement.
Verified documentation rendering.

### DIFF
--- a/packages/lexical-website/docs/concepts/node-replacement.md
+++ b/packages/lexical-website/docs/concepts/node-replacement.md
@@ -1,6 +1,6 @@
 
 
-# Node Overrides
+# Node Overrides / Node Replacements
 
 Some of the most commonly used Lexical Nodes are owned and maintained by the core library. For example, ParagraphNode, HeadingNode, QuoteNode, List(Item)Node etc - these are all provided by Lexical packages, which provides an easier out-of-the-box experience for some editor features, but makes it difficult to override their behavior. For instance, if you wanted to change the behavior of ListNode, you would typically extend the class and override the methods. However, how would you tell Lexical to use *your* ListNode subclass in the ListPlugin instead of using the core ListNode? That's where Node Overrides can help.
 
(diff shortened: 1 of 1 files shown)


packages/lexical-website/docs/concepts/node-replacement.md
```diff
@@ -1,6 +1,6 @@
 
 
-# Node Overrides
+# Node Overrides / Node Replacements
```
etrepum, line 3: Thank you for doing this, I often use the term node replacements because that's basically what they're called in the code and it's easy to forget that the documentation calls them something else!

packages/lexical-website/docs/concepts/node-replacement.md
```diff
@@ -22,6 +22,10 @@ const editorConfig = {
     ]
 }
 ```
+In the snippet above,
+- `replace`: Specifies the core node type to be replaced. 
+- `with`: Defines a transformation function to replace instances of the original node to the custom node.  
+- `withKlass`: This option ensures that behaviors associated with the original node type work seamlessly with the replacement. For instance, node transforms or listeners targeting ParagraphNode will also apply to CustomParagraphNode when withKlass is specified. Without this option, the custom node might not fully integrate with the editor's built-in features, leading to unexpected behavior.
```
etrepum, line 28:
```suggestion
- `withKlass`: This option ensures that behaviors associated with the original node type work seamlessly with the replacement. For instance, node transforms or mutation listeners targeting ParagraphNode will also apply to CustomParagraphNode when withKlass is specified. Without this option, the custom node might not fully integrate with the editor's built-in features, leading to unexpected behavior.
```

etrepum (approved):
This is great, it would be even better to include some API documentation for the `LexicalNodeReplacement` type as a follow-up by adding appropriately formatted comments to LexicalEditor.ts

etrepum (approved):

etrepum:
I'm going to merge this now, but re-open #6398 until we also have API documentation for this

packages/lexical-website/docs/concepts/node-replacement.md
Shopiley: yes! That's also why it took me forever to find it in the docs

Shopiley:
> This is great, it would be even better to include some API documentation for the `LexicalNodeReplacement` type as a follow-up by adding appropriately formatted comments to LexicalEditor.ts

okay, I could give a go at that as well

PR #6892, opened 2024-11-30
Fix selected table colors

zurfyx:
Make it more like the actual browser colors

| Before | After |
| --- | --- |
| Screenshot 2024-11-30 at 1 55 54 PM  | Screenshot 2024-11-30 at 1 55 31 PM  | 

We might also want to be backward compatible with the default selected theming previously built-in, I don't think we want to customize this often so it provides value even beyond backward compatibility (ref https://github.com/facebook/lexical/pull/6658) might follow up on this

### DIFF
--- a/examples/react-table/src/styles.css
+++ b/examples/react-table/src/styles.css
@@ -496,8 +496,8 @@ i.justify-align {
   right: 0;
   bottom: 0;
   top: 0;
-  background-color: rgb(172, 206, 247);
-  opacity: 0.6;
+  background-color: highlight;
+  mix-blend-mode: multiply;
   content: '';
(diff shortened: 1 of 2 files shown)


etrepum (approved):

etrepum:
The only backwards compatibility issue with #6658 is that previously some styles were applied directly to the DOM where now they are expected to come from CSS. There isn't really a way to make it backwards compatible without using `getComputedStyle` or something like that to see if the expected style properties did get set by the stylesheet. If you have a codebase that supports both versions of lexical you can update the CSS and it will work either way since the old direct style manipulation will just take precedence over the CSS.

PR #6894, opened 2024-12-02
[Breaking Change][lexical] Feature: New update tag: skip-dom-selection, $onUpdate now always called

### LINKED ISSUE #6514: Docs: Selection reconciliation & focus FAQs
It's very common for new users to stumble with focus issues with the editor. The way that Lexical reconciles its selection to the editor should be documented.

We might also want to consider ways to control this more precisely and to detect that it shouldn't change.

A pretty common response to focus related questions:

> This happens because the editor has a selection and when it updates the selection is reconciled to the DOM. The browser will always focus whatever is selected. You can either set the editor to not editable during this update, or set the selection to null

I think this issue is exacerbated by the fact that commands or registering plug-ins can cause the editor to be dirty and perform selection reconciliation even when it isn't obvious that it "should". Maybe a pair of update tags (or `$functions` that set editor state) to prevent or force selection change could help resolve this issue?

etrepum:
## Description

* Adds documentation about focus to the selection documentation.
* Adds a `'skip-dom-selection'` update tag. Both of these are relevant to a very common FAQ which is about how to avoid the editor taking over the selection with a programmatic update.
* Changes the type of `updateDOM` to be more correct, the `prevNode` argument should always be `this` (the same class, but a different instance). This fixes some subclassing issues.
* Changes the behavior of `onUpdate` and `$onUpdate` to always be called. Previously they would be skipped if the update containing them did not make any nodes dirty. It's often useful to use `$onUpdate` to attach some logic in a `dispatchCommand` that does not need to update nodes or the selection but needs to do something with the reconciled DOM or editor afterwards.
* registerNodeTransform now only marks relevant nodes as dirty, instead of all nodes

Closes #6514
Clsoes #4474

## Upgrade Notes

If you have code that expects `onUpdate` or `$onUpdate` to not be called if the update did not make any nodes or the selection dirty, then you will have to change your code accordingly.

Now when transforms are registered, only relevant nodes are marked dirty. Previously, all nodes in the editor were marked as dirty.

## Test plan

Unit tests were added for the new behavior of `onUpdate` / `$onUpdate` and  the `'skip-dom-selection'` tag.


packages/lexical-website/docs/concepts/selection.md
```diff
@@ -116,4 +116,54 @@ editor.update(() => {
...
+selection is reconciled to the DOM selection during reconciliation,
+and the browser's focus follows its DOM selection.
+
+If you want to make updates or dispatch commands to the editor without
+changing the selection, can use the `'skip-dom-selection'` update tag
+(added in v0.22.0):
+
+```js
+// Call this from an editor.update or command listener
+$addTag('skip-dom-selection');
```
ivailop7, line 135: shouldn't it be '$addUpdateTag' ?

ivailop7 (approved):

etrepum:
It looks like there's one failing flaky collab+firefox e2e test that now reliably fails in this PR. I'll investigate further before merge.

```
1 failed
[1]     [firefox] › packages/lexical-playground/__tests__/e2e/Autocomplete.spec.mjs:76:3 › Autocomplete › Can autocomplete in the same format as the original text 
```

zurfyx (dismissed):

etrepum:
So what's currently happening here is:

* Left editor changes the font size
* Left editor types a character or two
* Right editor sees the new text node with font size change, considers it a conflict for whatever reason, and creates a new paragraph
* Right editor's new paragraph ends up moving the browser's focus to the right editor, so the rest of the characters end up being typed into the Right editor's new paragraph

etrepum:
Autocomplete is a necessary condition for this, so it seems that interaction between autocomplete and collab is the real problem here

packages/lexical/src/LexicalEditor.ts
```diff
@@ -960,7 +970,10 @@ export class LexicalEditor {
       registeredNodes.push(registeredReplaceWithNode);
     }
 
-    markAllNodesAsDirty(this, klass.getType());
+    markNodesWithTypesAsDirty(
```
etrepum, line 973: I renamed this internal function because now it actually does what we need it to do

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -607,7 +607,8 @@ export function $commitPendingUpdates(
     editor._editable &&
     // domSelection will be null in headless
     domSelection !== null &&
-    (needsUpdate || pendingSelection === null || pendingSelection.dirty)
+    (needsUpdate || pendingSelection === null || pendingSelection.dirty) &&
+    !tags.has('skip-dom-selection')
```
etrepum, line 611: This is the logic that skips DOM selection with that tag

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -1005,6 +1006,7 @@ function $beginUpdate(
 
   const shouldUpdate =
     editor._dirtyType !== NO_DIRTY_NODES ||
+    editor._deferred.length > 0 ||
```
etrepum, line 1009: This is where we defer the update whenever there is an onUpdate, so you can attach the callback, do some mutations in another update issued that tick, and still get a callback

packages/lexical/src/LexicalUtils.ts
```diff
@@ -72,7 +72,6 @@ import {
   internalGetActiveEditorState,
   isCurrentlyReadOnlyMode,
   triggerCommandListeners,
-  updateEditor,
```
etrepum, line 75: I didn't see a reason to use this internal function here

packages/lexical/src/LexicalUtils.ts
```diff
@@ -498,22 +497,31 @@ export function getEditorStateTextContent(editorState: EditorState): string {
   return editorState.read(() => $getRoot().getTextContent());
 }
 
-export function markAllNodesAsDirty(editor: LexicalEditor, type: string): void {
-  // Mark all existing text nodes as dirty
-  updateEditor(
-    editor,
+export function markNodesWithTypesAsDirty(
```
etrepum, line 500: Previously whenever a transform was registered we marked all nodes in the editor as dirty, which was surprisingly the biggest factor in the collab+autocomplete+selection issue. Now it only marks relevant nodes as dirty, which in this case is none. We can use the same internal facility developed for registerMutationListener initialization to find the matching nodes by type in the reconciled editor state to determine which ones should be marked as dirty.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -498,22 +497,31 @@ export function getEditorStateTextContent(editorState: EditorState): string {
...
-      if (type === 'root') {
-        $getRoot().markDirty();
-        return;
-      }
-      const nodeMap = editorState._nodeMap;
-      for (const [, node] of nodeMap) {
-        node.markDirty();
+      for (const nodeMap of dirtyNodeMaps) {
+        for (const node of nodeMap.values()) {
+          node.markDirty();
```
etrepum, line 523: We don't need to check if the nodes are attached, worst case they just end up being re-inserted in the nodeMap and get GC'd on reconciliation.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1825,17 +1833,26 @@ export function getCachedTypeToNodeMap(
   );
   let typeToNodeMap = cachedNodeMaps.get(editorState);
   if (!typeToNodeMap) {
-    typeToNodeMap = new Map();
+    typeToNodeMap = computeTypeToNodeMap(editorState);
```
etrepum, line 1836: This is just extracting some logic to a function, nothing was changed here, I thought I might need to do it with a live EditorState but then realized that the frozen one will work even better since all nodes in the live editorstate are either in the frozen one or are already dirty.

packages/lexical/src/index.ts
```diff
@@ -29,6 +29,7 @@ export type {
   SerializedEditor,
   Spread,
   Transform,
+  UpdateListener,
```
etrepum, line 32: I thought I was going to need to use tags in the updateListener in the Autocomplete plugin and noticed this type was not exported

etrepum (commented):
This is passing all tests now, just needs a re-approval. Added some commentary on the new changes.

zurfyx (approved):

PR #6897, opened 2024-12-02
[lexical][lexical-rich-text][lexical-playground] Feature: Support capitalization format

bedre7:
## Description

This PR aims to add capitalization formats to text formatting options. Most text editors support capitalization formats. Currently, there is no way to change the capitalization of a written text without rewriting it entirely. The new capitalization formats added are lowercase, uppercase, and capitalize.

The behavior of capitalization format in other editors such as Google Docs, is as follows:
1. A piece of text can be selected and its capitalization can be changed
2. If a capitalization format was activated, it should revert to default when the Enter, Space, or Tab keys are pressed.

## Test plan

Unit tests and e2e tests have been added

### Before

There was no option to change the capitalization of existing text; users had to rewrite the entire text.

### After

A selection's capitalization can be changed, allowing users to easily switch between lowercase and uppercase formats without needing to rewrite the text.

https://github.com/user-attachments/assets/b2ccc356-3b61-46b9-8a3f-806abb6eb775

## Acknowledgement

* @Shopiley for her neat contribution.
* @etrepum for his suggestion and guidance.


etrepum (commented):
I think overall this looks great, but it would also make sense to support Title Case (text-transform: capitalize). Screenshot of Google Docs for reference:

Screenshot 2024-12-02 at 20 35 35

bedre7:
> I think overall this looks great, but it would also make sense to support Title Case (text-transform: capitalize). Screenshot of Google Docs for reference:
> 
> Screenshot 2024-12-02 at 20 35 35

Thanks for the feedback, that's a good point. I had implemented Title Case at first but then I realized there is an edge case that needs to be addressed. Unfortunately, text-transform: capitalize only capitalizes the first letter of each word, leaving the rest of the characters unchanged (for example: hELLO wORLD ends up being HELLO WORLD when it should have been Hello World). The only workaround I could find was converting the selection text to lowercase first, then the text-transform CSS property would work as expected. If this is acceptable, I'm happy to add it. What do you think?

Reference to the workaround I mentioned: https://stackoverflow.com/a/13793667/14920149

etrepum:
I think if we just call it capitalize it would be fine to use the css semantics. I don't think we should make it complicated.

etrepum (commented):
It looks like this is failing type checks because the theme wasn't updated, you can `npm run ci-check` locally to run the integrity tests.

etrepum (commented):
I think the toolbar changes also require e2e test changes, maybe there is some assertion based on the index of a specific toolbar item? Take a look at the failing check.

bedre7:
> I think the toolbar changes also require e2e test changes, maybe there is some assertion based on the index of a specific toolbar item? Take a look at the failing check.

Yes, it should be fixed once @Shopiley adds a test case for the 'capitalize' format.

bedre7:
Hi @etrepum, it's now ready for review.

packages/lexical/flow/Lexical.js.flow
```diff
@@ -585,7 +591,10 @@ export type TextFormatType =
   | 'highlight'
   | 'code'
   | 'subscript'
-  | 'superscript';
+  | 'superscript'
+  | 'lowercase'
+  | 'uppercase';
```
etrepum, line 596:
```suggestion
  | 'uppercase'
  | 'capitalize';
```

etrepum (approved):

etrepum (approved):
Looks great! The only thing I noticed was a missing 'capitalize' in the flow type, but I committed a fix for that. Will merge once I see what the test run looks like just to be sure.

PR #6902, opened 2024-12-03
[lexical][@lexical/selection] Feature: Unify $selectAll Implementations

### LINKED ISSUE #6893: Feature: Unify $selectAll implementations from lexical and @lexical/selection
## Description

`$selectAll` is exported from `@lexical/selection` and `lexical`. These functions do slightly different things (the one from `lexical` can create a new `RangeSelection` directly from root, the one from `@lexical/selection` always expands an existing `RangeSelection` and respects shadow roots), but from a user's perspective there's not usually a meaningful difference.

I think we should:

- Change the signature of `$selectAll` from `lexical` to `$selectAll(selection?: RangeSelection | null): RangeSelection` so it can either expand an existing RangeSelection (the `@lexical/selection` semantics) or create/update-and-set a RangeSelection (the `lexical` semantics). I think we should also return the selection (current behavior is void) because it's often useful to have a reference to the selection and it's annoying to separately call `$getSelection()` and assert-or-cast to ensure that it's the RangeSelection that you know it is.
- Remove the `$selectAll` implementation from `@lexical/selection` and replace it with a re-exported and deprecated reference to the one from the `lexical` package

## Impact

It will make it easier to use `$selectAll` in both cases, shrink the bundle size of `@lexical/selection`, and reduce the API surface area.

Monorepo usage:

- The `@lexical/selection` implementation is not used or tested in the monorepo.
- The `lexical` implementation is only used to implement `SELECT_ALL_COMMAND` and in a few unit tests.

elgh0ul:
### Current Behavior
- `$selectAll` is implemented in both the `lexical` and `@lexical/selection` packages, leading to redundancy and potential confusion.
- The `lexical` implementation creates a new `RangeSelection`, while the `@lexical/selection` version expands an existing selection.

### New Behavior
- Updated `$selectAll` in `lexical` to handle both cases:
  - If no `selection` is provided, it creates a new `RangeSelection`.
  - If a `selection` is provided, it expands it.
- Deprecated `$selectAll` in `@lexical/selection` by re-exporting the `lexical` version with a `@deprecated` annotation.

Closes #6893

### Before
- `$selectAll` in `@lexical/selection` expands an existing selection.
- `$selectAll` in `lexical` creates a new `RangeSelection`.

### After
- `$selectAll` in `lexical` handles both creating and expanding selections.


packages/lexical/src/LexicalUtils.ts
```diff
@@ -1084,10 +1084,49 @@ export function isSelectAll(
...
+      lastNode = lastNode.getParentOrThrow();
+    }
+
+    if (firstNode && lastNode) {
+      anchor.set(firstNode.getKey(), 0, firstType);
+      focus.set(lastNode.getKey(), lastOffset, lastType);
+    }
+
+    // Normalize and set the updated selection
+    $setSelection($normalizeSelection(selection));
```
etrepum, line 1122: This implementation shouldn't be calling `$setSelection`, if the user passes in a RangeSelection it's possible they do not want it to be also set (e.g. they are passing it to some other API that takes a selection, but do not want to affect the current editor state).

The `$normalizeSelection` is redundant here because that's what the rest of the code is doing. I'm sure most of it is just copied from the previous implementation, which likely predates `$normalizeSelection`. Since we do have `$normalizeSelection` we might as well simplify it all the way. Maybe something like this:

```suggestion
    const anchorNode = anchor.getNode();
    const topParent = anchorNode.getTopLevelElementOrThrow();
    const rootNode = topParent.getParentOrThrow();
    anchor.set(rootNode.getKey(), 0, 'element');
    focus.set(rootNode.getKey(), rootNode.getChildrenSize(), 'element');
    $normalizeSelection(selection);
```

elgh0ul:
Hey @etrepum,
Thanks for the feedback! I've incorporated these new changes and hopefully everything should be good now.

etrepum (approved):
Looks great, nice job!

PR #6906, opened 2024-12-04
[lexical-playground]: Fix empty layout item causes 100% CPU usage

basile-savouret:
[lexical-playground] Bug Fix : Prevent layout item from being empty which would leed into a 100% cpu usage

## Description
In playground if we write on 2 layout items and copy cut both of the item's content the right item will become empty. 
Then if we focus it and press "enter" to insert a new line the editor goes into an infinite loop and goes into 100% cpu usage without triggering any error.

To fix this I added the code I did in my company project. Which is a listener for any empty layout item node that will re-fill it with a paragraph node.

### Before

https://github.com/user-attachments/assets/5269f3b0-ceeb-4814-b5a4-07bdd27fc1dd

### After

https://github.com/user-attachments/assets/25831bb1-9382-48a2-bc82-a80830c68d75


etrepum (commented):
This would be better done with a node transform, mutation listeners happen after reconciliation so this implementation causes an update cascade

basile-savouret:
@etrepum i did modifications based on your comment, I hope this is what you had in mind.

etrepum (approved):
Looks great, thank you!

PR #6912, opened 2024-12-05
[lexical-list] Fix: Preserve original format after indenting list item 

### LINKED ISSUE #6476: Bug: Selection formatting/styling is lost when you indent a list
If you create a list item with some style, press enter to create a new item at the same level and then press tab you lose the formatting on the selection.

Lexical version: latest

## Steps To Reproduce

1. Create a list that has some styling, for example bold:

image

2. Press enter to create a new list item of the same level (bold styling remains on selection.

image

3. Press tab to indent the list item (bold styling disappears).

image

Link to code example:

Can easily be reproduced on the playground.

## The current behavior

Selection formatting is lost after indenting a list item.

## The expected behavior

Selection formatting is not lost after indenting a list item.

## Impact of fix

Currently an important bug on my app is caused by this.

citruscai:
## Description
This PR fixes the issue of format from previous list item being lost when you indent the new list item.
### Changes
- Make ListItemNode inherit from ParagraphNode to have properties for text format
- Set text format for new list item by getting the format from the selection

Closes #6476 

### Before

https://github.com/user-attachments/assets/564b3b3d-c4b2-431d-962c-5b33e7e0c7a7

### After

https://github.com/user-attachments/assets/63019777-992d-47f9-939e-f2c193301e51


etrepum:
Integrity tests are failing, looks like there are some typos in here

etrepum:
Still failing integrity checks

citruscai:
@etrepum only 3 e2e test failing and theyre related to links should i touch them?

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -81,20 +80,19 @@ export class ListItemNode extends ElementNode {
...
-  updateDOM(
-    prevNode: ListItemNode,
-    dom: HTMLElement,
-    config: EditorConfig,
-  ): boolean {
     const parent = this.getParent();
     if ($isListNode(parent) && parent.getListType() === 'check') {
-      updateListItemChecked(dom, this, prevNode, parent);
+      const listItemNode = $isListItemNode(prevNode) ? prevNode : null;
+      updateListItemChecked(dom, this, listItemNode, parent);
```
etrepum, line 91: This change is not necessary, the type is already `this` which is known to be a `ListItemNode` (or subclass)

```suggestion
      updateListItemChecked(dom, this, prevNode, parent);
```

packages/lexical-list/src/formatList.ts
```diff
@@ -478,12 +480,14 @@ export function $handleListInsertParagraph(): boolean {
   if (!$isRangeSelection(selection) || !selection.isCollapsed()) {
     return false;
   }
-  // Only run this code on empty list items
+  //Only run this code on empty list items
```
etrepum, line 483:
```suggestion
  // Only run this code on empty list items
```

packages/lexical-list/src/formatList.ts
```diff
@@ -524,9 +526,7 @@ export function $handleListInsertParagraph(): boolean {
       newList.append(sibling);
     });
   }
-
-  // Don't leave hanging nested empty lists
+  //Don't leave hanging nested empty lists
```
etrepum, line 529:
```suggestion
  // Don't leave hanging nested empty lists
```

packages/lexical-table/src/LexicalTableSelection.ts
```diff
@@ -220,7 +220,7 @@ export class TableSelection implements BaseSelection {
     cellNodes.forEach((cellNode: TableCellNode) => {
       const paragraph = cellNode.getFirstChild();
       if ($isParagraphNode(paragraph)) {
-        format |= paragraph.getTextFormat();
+        format |= paragraph.getTextFormat() || 0;
```
etrepum, line 223: We should not change the signature of this method

```suggestion
        format |= paragraph.getTextFormat();
```

packages/lexical/src/LexicalEvents.ts
```diff
@@ -354,8 +354,8 @@ function onSelectionChange(
               lastNode instanceof ParagraphNode &&
               lastNode.getChildrenSize() === 0
             ) {
-              selection.format = lastNode.getTextFormat();
-              selection.style = lastNode.getTextStyle();
+              selection.format = lastNode.getTextFormat() || 0;
+              selection.style = lastNode.getTextStyle() || '';
```
etrepum, line 358: These methods shouldn't change

```suggestion
              selection.format = lastNode.getTextFormat();
              selection.style = lastNode.getTextStyle();
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -38,8 +38,8 @@ import {$isTextNode, TextFormatType} from './LexicalTextNode';
 
 export type SerializedParagraphNode = Spread<
   {
-    textFormat: number;
-    textStyle: string;
+    textFormat?: number;
+    textStyle?: string;
```
etrepum, line 42: ParagraphNode shouldn't change in this way, it's not backwards compatible since these are existing and required, we would be generating JSON that the previous version may not be able to parse correctly.

```suggestion
    textFormat: number;
    textStyle: string;
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -48,8 +48,8 @@ export type SerializedParagraphNode = Spread<
 export class ParagraphNode extends ElementNode {
   ['constructor']!: KlassConstructor<typeof ParagraphNode>;
   /** @internal */
-  __textFormat: number;
-  __textStyle: string;
+  __textFormat?: number;
+  __textStyle?: string;
```
etrepum, line 52: This shouldn't change. There's no reason for these to be optional, they are always explicitly set in the constructor. There are a lot of other changes here that can be reverted.

```suggestion
  __textFormat: number;
  __textStyle: string;
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -63,7 +63,7 @@ export class ParagraphNode extends ElementNode {
 
   getTextFormat(): number {
     const self = this.getLatest();
-    return self.__textFormat;
+    return self.__textFormat || 0;
```
etrepum, line 66: Not needed

```suggestion
    return self.__textFormat;
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -74,7 +74,13 @@ export class ParagraphNode extends ElementNode {
...
   hasTextFormat(type: TextFormatType): boolean {
     const formatFlag = TEXT_TYPE_TO_FORMAT[type];
-    return (this.getTextFormat() & formatFlag) !== 0;
+    const textFormat = this.getTextFormat() || 0;
+
+    if (textFormat === 0) {
+      return false;
+    }
+
+    return (textFormat & formatFlag) !== 0;
```
etrepum, line 83: Not needed

```suggestion
return (this.getTextFormat() & formatFlag) !== 0;
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -85,12 +91,12 @@ export class ParagraphNode extends ElementNode {
   getFormatFlags(type: TextFormatType, alignWithFormat: null | number): number {
     const self = this.getLatest();
     const format = self.__textFormat;
-    return toggleTextFormatType(format, type, alignWithFormat);
+    return toggleTextFormatType(format || 0, type, alignWithFormat);
```
etrepum, line 94:
```suggestion
    return toggleTextFormatType(format, type, alignWithFormat);
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -85,12 +91,12 @@ export class ParagraphNode extends ElementNode {
...
   }
 
   getTextStyle(): string {
     const self = this.getLatest();
-    return self.__textStyle;
+    return self.__textStyle || '';
```
etrepum, line 99:
```suggestion
    return self.__textStyle;
```

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -190,7 +194,10 @@ export class ParagraphNode extends ElementNode {
     const direction = this.getDirection();
     newElement.setDirection(direction);
     newElement.setFormat(this.getFormatType());
-    newElement.setStyle(this.getTextStyle());
+    if (typeof this.getStyle() === 'string') {
+      newElement.setStyle(this.getStyle());
+    }
```
etrepum, line 199:
```suggestion
    newElement.setStyle(this.getStyle());
```

etrepum (commented):
This looks like it's getting close! I did see a lot of unnecessary changes here that should be reverted, but once that's done I think it might be good to go.

citruscai:
@etrepum now im getting this annoying error!!! grrr even tried pushing an empty commit and then i ran prettier locally and nothing happened 
image

etrepum:
That's a formatting issue, you need to run prettier:fix

packages/lexical-list/src/formatList.ts
```diff
@@ -242,10 +242,12 @@ export function removeList(editor: LexicalEditor): void {
...
           if ($isLeafNode(node)) {
-            const listItemNode = $getNearestNodeOfType(node, ListItemNode);
+            if ($isListItemNode(node)) {
+              const listItemNode = $getNearestNodeOfType(node, ListItemNode);
 
-            if (listItemNode != null) {
-              listNodes.add($getTopListNode(listItemNode));
+              if (listItemNode != null) {
+                listNodes.add($getTopListNode(listItemNode));
+              }
```
etrepum, line 250: I think this change is wrong. `$isLeafNode(node)` and `$isListItemNode(node)` are never true at the same time, so this block never does anything now. I guess there just isn't test coverage for what it's trying to do. Can you put this block back to the way it was and see if the tests still pass?

etrepum:
I'm currently trying to get a clean test run, please don't do any updates unless you have a reason to change the code manually (clicking the update button will run all of the tests again which is slowing this review down)

citruscai:
> I'm currently trying to get a clean test run, please don't do any updates unless you have a reason to change the code manually (clicking the update button will run all of the tests again which is slowing this review down)

no problem, sorry about that!

etrepum (approved):
Well done!

etrepum:
This change was reverted in #6929 due to some compatibility issues internally at Meta with the change in class hierarchy. The majority of this change is good, but in order to have better backwards compatibility the most pragmatic approach will be to create a new base class common to ParagraphNode and ListItemNode that has the common functionality so that `$isParagraphNode` and `$isListItemNode` can remain disjoint. Since this is a bit involved I'll take a first pass at this sometime this week.

PR #6917, opened 2024-12-06
[lexical-list] Bug Fix: Ensure new paragraph node retains selection styling when exiting list

aleksandr-lapushkin:
## Description
- Currently, when exiting a list by pressing enter multiple times, the selection style is lost
- This PR aims to ensure that the current style is retained

### Before
- Go to Playground
- Insert a new list
- Change text style, i.e. font, color or formatting
- Write a few bullet points
- Press enter twice to exit list
- Write more text - this will no longer have the correct styles applied

https://github.com/user-attachments/assets/0d88cf3a-8cff-437f-9ce5-1e6daf53d1bc

### After
- Go to Playground
- Insert a new list
- Change text style, i.e. font, color or formatting
- Write a few bullet points
- Press enter twice to exit list
- Write more text - this will be styled correctly

https://github.com/user-attachments/assets/9d6179c4-7046-438e-b4e2-a62983fb0b68


aleksandr-lapushkin:
Honestly not sure if there's a better way to do this, but seemed reasonable enough 🤷🏻

etrepum (commented):
It would be great if there was a unit and/or e2e test to show that this has the expected result

aleksandr-lapushkin:
> It would be great if there was a unit and/or e2e test to show that this has the expected result

Sure thing! Added an e2e test.

etrepum (approved):

PR #6919, opened 2024-12-08
[lexical-markdown] Refactor: Update regExpEnd optional type in MultilineElementTransformer

untilhamza:
[lexical-markdown] Refactor: Broaden regExpEnd optional type to boolean

## Description

**Current Behavior:**
In the [MultilineElementTransformer](cci:2://file:///Users/untilhamza/Developer/jenga-codelabs/lexical/packages/lexical-markdown/src/MarkdownTransformers.ts:76:0-206:48) interface, the `regExpEnd` property's `optional` field was unnecessarily restricted to only accept `true` as a value. This strict typing didn't provide any practical benefit and limited potential use cases where `false` might be a valid state.

**Changes:**
- Modified the type definition of `optional` in `regExpEnd` from `true` to `boolean`
- This change allows for more flexible usage of the optional flag
- Makes the type definition more intuitive and consistent with typical boolean flag patterns

Closes #[issue number]

### Before
The `regExpEnd` type was overly restrictive:
```typescript
regExpEnd?: RegExp | {
  optional?: true;  // Could only be true
  regExp: RegExp;
};
```

### After
The `regExpEnd` type now allows both true and false values:

```typescript
CopyInsert
regExpEnd?: RegExp | {
  optional?: boolean;  // Can be either true or false
  regExp: RegExp;
};
```

This change maintains backward compatibility while providing more flexibility in how the optional flag can be used.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -113,7 +113,7 @@ export type MultilineElementTransformer = {
          * Whether the end match is optional. If true, the end match is not required to match for the transformer to be triggered.
          * The entire text from regexpStart to the end of the document will then be matched.
          */
-        optional?: true;
+        optional?: boolean;
         regExp: RegExp;
       };
   /**


etrepum (commented):
The reason why you would use `optional?: true` is to make it clear that the default is false, and that there's no reason to specify it otherwise. I'm not sure this really has any benefit, there are other places in the code where an optional `true` is used as the type (such as `discrete?: true` and `skipTransforms?: true`), so if this change was desired I would expect that it would be done to the whole project. Does anyone else have opinions on this? @zurfyx @ivailop7 

Generally speaking it's best to use types to limit the number of possible runtime states that should to be considered. `{x?: boolean}` has 4 states and `{x?: true}` has 3 (one less each when `exactOptionalPropertyTypes` is used)

untilhamza:
> The reason why you would use `optional?: true` is to make it clear that the default is false, and that there's no reason to specify it otherwise. I'm not sure this really has any benefit, there are other places in the code where an optional `true` is used as the type (such as `discrete?: true` and `skipTransforms?: true`), so if this change was desired I would expect that it would be done to the whole project. Does anyone else have opinions on this? @zurfyx @ivailop7
> 
> Generally speaking it's best to use types to limit the number of possible runtime states that should to be considered. `{x?: boolean}` has 4 states and `{x?: true}` has 3 (one less each when `exactOptionalPropertyTypes` is used)

@etrepum , this is actually very insightful. However I did not know the default was false as I could not tell from just looking at the types. Is there a way to explicitly tell the user of the function that the default is false while keeping the type as `true`?

etrepum:
The type itself already says that either the setting doesn't mean anything at all, or the default is not true. What you could do to make it more explicit for people who don't understand that implication is to add more documentation to the type.

untilhamza:
I see what you mean. I will close this PR for now. Thanks for sharing the thinking behind it.

PR #6923, opened 2024-12-09
RFC: FloatingUI for LexicalMenu positioning

ivailop7:
To get rid of the popover menus positioning issues, floating-ui is the defacto standard for this problem, got it to work and works nicely. A few example to try:

Very long text now shows all the "/, @, :" menus correctly if the page is scrolled, also shows the menu correctly if typing on the last line in focus on the screen, it renders the menu upwards correctly. So this is technically the universal fix for 'popover' vs 'context menu'.

A few questions:
1) Are we happy to have floating-ui/react as a dependency for lexical-react?
2) Are we OK to remove the 'menuRenderFn' and unify the underer under LexicalMenu. So any plugin would just provide the menu items.

Clone and run locally to get a better idea of how it performs. Before I spend more time this, would like to gather some thoughts from the rest - @etrepum @zurfyx @fantactuka @potatowagon


etrepum (commented):
I'm 1000% for getting rid of Lexical's menu code and replacing it with something that works correctly without a large maintenance burden, floating-ui seems like the ideal choice. I also like floating-ui because it's not React dependent so when we start offering framework agnostic plugins it could work for that as well.

Our menu situation is a disaster, I don't think it works correctly for anyone but Meta (maybe?), and there aren't enough regression tests for anyone outside of Meta to fix it because any attempt to do so seems to end up getting reverted for reasons that are opaque to anyone else.

PR #6925, opened 2024-12-09
[lexical][lexical-list] Bug Fix: Style lost on list insertion from empty paragraph

aleksandr-lapushkin:
## Description
This PR attempts to fix a scenario where attempting to replace an empty Paragraph with `textStyle`, which also changes the current selection, would cause the style to be lost.
### Before
- Change font/color/size
- Write something
- Insert a new empty paragraph
- Insert a list
- Write some text
- Style is lost

https://github.com/user-attachments/assets/cebe40ed-940c-436f-94f5-a8882c0a1d02

### After
- Change font/color/size
- Write something
- Insert a new empty paragraph
- Insert a list
- Write some text
- Style is retained

https://github.com/user-attachments/assets/28fe0025-9e63-4b50-9323-2d6cdf2882a9

### DIFF
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -154,9 +154,14 @@ function $createListOrMerge(node: ElementNode, listType: ListType): ListNode {
     return node;
   }
 
+  const selection = $getSelection();
+  const selectionStyle = $isRangeSelection(selection) ? selection.style : '';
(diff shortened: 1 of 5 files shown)


etrepum:
I think you should wait on #6912 for this one, the implementation approaches are conflicting and it may fix the same issue.

etrepum:
AFAICT this is a solved problem, so this PR is no longer necessary

PR #6929, opened 2024-12-09
[lexical][@lexical/selection] Feature: add a generic `classes` property to all nodes for easier customization

GermanJablo:
Many places have discussed how node customization and extension could be made simpler (e.g. [[1](https://github.com/facebook/lexical/issues/3763)], [[2](https://github.com/facebook/lexical/pull/4254#issuecomment-1495101288)]).

We can divide the solutions into 2 groups:

1. make some methods not required when extending nodes by doing the work automatically. See my PRs about a [generic clone](https://github.com/facebook/lexical/pull/3920), and also [the one about export/importJSON](https://github.com/facebook/lexical/pull/3931). While they were closed for being stale they have been well received and I think there is still good will for a change in that direction. The challenge however is that the automatic methods could end up having worse performance. 
2. Minimize the number of use cases in which it is required to extend a node. While there have been some PRs that have helped in that direction (like the new [DOMSlot API](https://github.com/facebook/lexical/pull/6759)), in this PR I want to propose a new way to customize nodes with more flexibility and without the need to extend and replace nodes.

## Introducing Classes

Most of the time when users want to customize a node they just want to add a property to it, which ends up being reflected as a class in the dom element.

To satisfy that need we can introduced two new methods to all nodes: `getClasses` and `setClass`. 

```ts
export function CoolRedPlugin() {

  return (
    <button
      onClick={() => {
        editor.update(() => {
          $forEachSelectedTextNode((textNode) => {
            // setClass mutates the classes object where the key-value pairs follow the
            // format prefix-suffix for string values, or just prefix for boolean values.
            textNode.setClass('bg', 'red'); // adds the class bg-red
            // Perhaps you don't want to allow the same node to have
            // both text and background color defined at the same time...
            textNode.setClass('text', false); // ...so here you remove the class text-[color].
            textNode.setClass('cool', true); // adds the class cool (true values don't add a suffix)
          });
        });
      }}>
      Make text red and cool
    </button>
  );
}
```

<details>
<summary>EDITED: expand this to see the first API proposal</summary>

```ts
export function CoolRedPlugin() {

  return (
    <button
      onClick={() => {
        editor.update(() => {
          $forEachSelectedTextNode((textNode) => {
            // Allows mutation of the classes object where the key-value pairs follow the
            // format prefix-suffix for string values, or just prefix for true boolean values.
            textNode.mutateClasses((classes) => {
              classes.bg = 'red'; // adds the class bg-red
              // Perhaps you don't want to allow the same node to have
              // both text and background color defined at the same time...
              delete classes.text; // ...so here you remove the class text-[color].
              classes.cool = true; // adds the class cool (true values don't add a suffix)
            });
          });
        });
      }}>
      Make text red and cool
    </button>
  );
}
```

Under the hood the way I am implementing it is with a flat object whose values can be just strings or true.

```ts
type MutableClasses = {[classSuffix: string]: true | string};
type ReadOnlyClasses = {readonly [classSuffix: string]: true | string};

class LexicalNode {
  __classes?: ReadOnlyClasses;
  getClasses(): ReadOnlyClasses {
    return self.__classes || {};
  }
  mutateClasses(fn: (classes: MutableClasses) => void) {
    self.__classes = self.__classes || {};
    fn(self.__classes);
  }
```

</details>

## Considerations

1. why didn't I define it as an array of classes (strings), imitating the `Element.classList` API? 
    - For convenience and performance. A CSS property with multiple values is often mapped to a class. With an array, changing the background color for example would imply going through the whole array and eliminating the conflicting classes. This would be more verbose to program, slower, and more error-prone.
3. What about the styles property? 
    - This classes API is more flexible and easier to extend for the reasons I just gave. In addition, a class has the advantage that it can apply several styles at once, and its JSON serialization is smaller.
5. You can see I am not using `__classes` in the `createDOM`, `exportJSON`, `importJSON`, `clone` and `exportDOM` of each node, but in the places where those methods are invoked for 3 reasons:
    - it would require modifying more files.
    - if users extend a node they would need to repeat that process in several methods which degrades the value of this feature.
    - I am following a pattern that Lexical is already using with some properties. For example, `__indent` is applied in the reconciler and users do not have access to it.
6. These methods are not type-safe, (their purpose is precisely for users to use them as they wish). We could add a generic to `getClasses<T>` and `muteClasses<T>`.
7. To make this functionality easy to use in TextNodes, I created a new utility `$forEachSelectedTextNode` from `$patchStyleText`. There are other methods that could use it because they are very similar, such as `$RangeSelection.formatText`.
8. I'm making sure that the JSON serialization is as lightweight as possible (no export of forbidden values and no export if empty or undefined), something I've been wanting to do for a while for all the properties of all the nodes by the way :)
9. This PR does not introduce opinions about how the importDOM of the classes should be. If users want to paste HTML from external sources and maintain or transform their classes, they will probably want to extend their importDOM.

## Test plan

The PR contains tests.
Also, if you want to test it in the browser you can paste the `CoolRedPlugin` above in the playground editor.


packages/lexical-history/src/__tests__/unit/LexicalHistory.test.tsx
```diff
@@ -44,49 +44,52 @@ import {createRoot, Root} from 'react-dom/client';
 import * as ReactTestUtils from 'shared/react-test-utils';
 
 type SerializedCustomTextNode = Spread<
-  {type: ReturnType<typeof CustomTextNode.getType>; classes: string[]},
+  {type: ReturnType<typeof CustomTextNode.getType>; classList: string[]},
```
GermanJablo, line 47: The changes in this file are only a rename to avoid collisions and can be ignored.
I have also wondered what would happen if a user has extended a node by adding the classes property to it. Perhaps we could reserve a prefix and use something like `$classes` instead?

etrepum (commented):
I like the general idea here but I am not sure that this is the ideal API design (I do not have a concrete alternate proposal, this isn't something I have thought much about). Some initial thoughts:

- There's no consideration for more semantic meaning or how this interacts with themes
- It may not work very well for users who are using utility classes (e.g. tailwind)
- Namespace collisions (as you've identified in the example)
- Doesn't really distinguish between classes used for createDOM or exportDOM
- The implementation seems to copy more often than is necessary, given the access patterns __classes should be the readonly type and mutateClasses should have you work on a copy (and it should allow you to return a different object if you want)
- The prefixed variant seems useful for semantic reasons (e.g. `state: 'active'` and `state: 'inactive'` being mutually exclusive without having to toggle them separately) but is maybe a bit too prescriptive for exactly what the output class name looks like (although it is BEM compliant so it's not without precedent)
- It's likely that a lot of createDOM/updateDOM implementations won't be compatible with this up front particularly because LexicalNode doesn't have an implementation of it and this PR only considers TextNode
- No consideration for importDOM

GermanJablo:
Thank you for your thoughts!

(1), (2) and (6): I see no problem. Users can define the classes they want without any kind of prescription. They can choose `state-active` or just `active`. They can choose classes that match their utility framework or not.

(3): I think it would be good to reserve a prefix, since this problem actually applies every time Lexical adds a property or method. Or these occasions can also be marked as breaking changes...

(4) and (8): whether we should export the classes to the DOM is debatable. I don't think we should import them by default. It is left to the user to define the conversion, something [they can do without replacing the nodes](https://github.com/facebook/lexical/pull/4254).

(5) that would involve making a copy in each call to mutateClasses. I'm not sure I follow you, but if you have a suggestion on the API and could put it in the code comments or commit it would be great. 

(7): This PR works with all nodes. I made `forEachSelectedTextNode` because it is the most difficult use case, but you can modify the example plugin to add a class to a paragraph for example (`getRoot().getFirstChild().mutateClasses(...)`). It works fine. 

As a side note, I would like to add that I have reviewed all the nodes I have replaced and talked to other users who have heavily customized their nodes, and in all cases we have found that this API could have saved us the work.

etrepum:
The PR "works" for all nodes but it only has a DOM implementation for TextNode. If you add classes to any other node, they are not rendered, and it doesn't currently integrate with any theme properties. At this point, other than text nodes, it's just a property for storing some arbitrary metadata that has to be in a specific type for some reason.

Yes, I think reserving a prefix makes sense, especially in the class implementations but probably also in the serialization.

Regarding utility classes you would have something like `text-blue-600 hover:underline hover:cursor-pointer` that you are trying to apply for one semantic reason (e.g. this would style something like a link, so in normal css you might just have a `link` class). If you're applying it like that the meaning is more or less lost, and if you remove all of those classes then maybe you have removed something that maybe something else wanted there for a different semantic meaning. I'm pretty sure we already have issues with the way that the themes are currently implemented for people who are trying to use utility classes, because there's code that assumes that each theme class is disjoint, e.g. they assume that `add(propAClasses) -> add(propBClasses) -> remove(propAClasses) ~= add(propBClasses)` but if `intersect(propAClasses, propBClasses)` is non-empty then that assumption is false. If a more semantic approach is taken such that you're toggling a state and then all current states are used to generate the expected set of classes (possibly with a merge function like twMerge from tailwind-merge) then this problem goes away.

Probably what's really needed is some general way to store serializable metadata for a node, not unlike how DOM has attributes (with maybe special cases or utilities for dealing with things like classes, style, and dataset). Surely some people want to use something very much like this to add aria attributes or whatever instead of just classes. Maybe they want some sort of createDOM/updateDOM extensibility through editor configuration instead of subclassing (plus editor configuration).

I believe you that this API as-is would save some people some work, I just think that this kind of change that affects all nodes and all serialization warrants a higher level of scrutiny to make sure that we don't have API design regrets a few months down the road.

packages/lexical/src/LexicalNode.ts
```diff
@@ -185,6 +188,46 @@ export class LexicalNode {
...
+   *   delete currentClasses.active; // the node will no longer have the class `active`
+   * });
+   *
+   * @param fn A function that receives the current classes object and allows it to be mutated safely.
+   */
+  mutateClasses(fn: (classes: MutableClasses) => void) {
+    const self = this.getWritable();
+    self.__classes = self.__classes || {};
+    fn(self.__classes);
+  }
```
etrepum, line 230: This is what the sort of copy on write pattern I suggested would look like:

```suggestion
  mutateClasses(fn: (classes: MutableClasses) => void | ReadonlyClasses) {
    const self = this.getWritable();
    const classes = {...self.__classes};
    self.__classes = fn(classes) || classes;
  }
```

All other copies can go away since a readonly object doesn't need to be copied. The comparisons can also be short-circuited when the objects are identical which would be the case unless a mutation happened.

packages/lexical/src/LexicalNode.ts
```diff
@@ -170,6 +171,8 @@ export type DOMExportOutput = {
 };
 
 export type NodeKey = string;
+export type MutableClasses = {[classSuffix: string]: true | string};
+export type ReadOnlyClasses = {readonly [classSuffix: string]: true | string};
 
 export class LexicalNode {
```
etrepum, line 177: Adding `afterCloneFrom` support for `__classes` is probably also required if it's to be used anywhere but TextNode (and if so, it really shouldn't be here)

etrepum (commented):
Some notes on the mutation/copy API

packages/lexical/src/LexicalNode.ts
```diff
@@ -185,6 +188,55 @@ export class LexicalNode {
...
+   *   text: 'green', // node is rendered with class `text-green`,
+   *   active: true, // node is rendered with class `active`,
+   * }
+   * // Resulting classes: 'bg-red', 'text-green', and 'active'
+   *
+   * @returns The classes object.
+   */
+  getClasses(): ReadonlyClasses {
+    const self = this.getLatest();
+    return {...self.__classes};
```
etrepum, line 211: The previous `self.__classes || {}` was more performant, this always copies on read for no good reason. A no copy variant would use a variable initialized with `{}` to always return for empty classes, e.g. `self.__classes || EMPTY_CLASSES`

packages/lexical/src/LexicalNode.ts
```diff
@@ -185,6 +188,55 @@ export class LexicalNode {
...
+    const classes = {...self.__classes};
+    if (value === false) {
+      delete classes[key];
+    } else {
+      classes[key] = value;
+    }
+    if (Object.keys(classes).length > 0) {
+      self.__classes = classes;
+    }
+  }
```
etrepum, line 239: This pattern is N^2 if you are setting multiple classes because every time you add a class you iterate over everything. It also has a bug, you can't delete the last class, I would just unconditionally set it and determine whether or not to replace it with undefined on serialization. It's probably faster at runtime to always initialize `__classes` with an object because there's less branching. A mutateClasses and/or setClasses API would be better if making several mutations.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1841,6 +1844,9 @@ export function getCachedTypeToNodeMap(
 export function $cloneWithProperties<T extends LexicalNode>(latestNode: T): T {
   const constructor = latestNode.constructor;
   const mutableNode = constructor.clone(latestNode) as T;
+  if (latestNode.__classes) {
+    mutableNode.__classes = {...latestNode.__classes};
+  }
```
etrepum, line 1849: It would probably be better to do this in afterCloneFrom

GermanJablo:
Well, I see you reviewed my changes before I had time to explain them. That was fast! 😆

> The PR "works" for all nodes but it only has a DOM implementation for TextNode. If you add classes to any other node, they are not rendered, and it doesn't currently integrate with any theme properties. At this point, other than text nodes, it's just a property for storing some arbitrary metadata that has to be in a specific type for some reason.

This worked from the first commit of the PR. Maybe you didn't see that I was cloning somewhere else. I've now moved cloning to `afterCloneFrom`. Let me know if it still doesn't work for you with other nodes.

> Yes, I think reserving a prefix makes sense, especially in the class implementations but probably also in the serialization.

I have a local commit where I renamed to `__$classes`, `$getClasses`, `$setClass` and `$classes`. The problem is that objects starting with $ are wrapped in quotes. For example, for the test I introduced to work, I would have to replace:

`expect('$classes' in getSerializedParagraph(editor)).toBe(false);`

with:

`expect("'$classes'" in getSerializedParagraph(editor)).toBe(false);`

Which I don't like.
Maybe we can use a capital L as a prefix? Any other ideas?

> Surely some people want to use something very much like this to add aria attributes or whatever instead of just classes.

The nodes that Lexical provides do not need any aria attributes. If someone needs it, it is because they may be creating their own node anyway. 

> I believe you that this API as-is would save some people some work, I just think that this kind of change that affects all nodes and all serialization warrants a higher level of scrutiny to make sure that we don't have API design regrets a few months down the road.

I agree! And I see you've already seen the `setClasses` API I came up with afterwards. Arguably better than `mutateClasses`.

If I have left out anything other than address please let me know

packages/lexical/src/LexicalNode.ts
```diff
@@ -325,6 +323,7 @@ export class LexicalNode {
     this.__parent = prevNode.__parent;
     this.__next = prevNode.__next;
     this.__prev = prevNode.__prev;
+    this.__classes = {...prevNode.__classes};
```
etrepum: Readonly objects only need to be copied for mutation

```suggestion
    this.__classes = prevNode.__classes;
```

packages/lexical/src/LexicalNode.ts
```diff
@@ -273,6 +323,9 @@ export class LexicalNode {
     this.__parent = prevNode.__parent;
     this.__next = prevNode.__next;
     this.__prev = prevNode.__prev;
+    if (prevNode.__classes) {
+      this.__classes = {...prevNode.__classes};
+    }
```
etrepum, line 328:
```suggestion
    this.__classes = prevNode.__classes;
```

packages/lexical/src/LexicalNode.ts
```diff
@@ -185,6 +188,53 @@ export class LexicalNode {
...
+   *   text: 'green', // node is rendered with class `text-green`,
+   *   active: true, // node is rendered with class `active`,
+   * }
+   * // Resulting classes: 'bg-red', 'text-green', and 'active'
+   *
+   * @returns The classes object.
+   */
+  getClasses(): ReadonlyClasses {
+    const self = this.getLatest();
+    return self.__classes || {};
```
etrepum, line 211:
```suggestion
    return self.__classes;
```

packages/lexical/src/LexicalEditorState.ts
```diff
@@ -57,6 +57,9 @@ function exportNodeToJSON<SerializedNode extends SerializedLexicalNode>(
   node: LexicalNode,
 ): SerializedNode {
   const serializedNode = node.exportJSON();
+  if (node.__classes && Object.keys(node.__classes).length > 0) {
```
etrepum, line 60:
```suggestion
  if (Object.keys(node.__classes).length > 0) {
```

It does seem strange to force this here, before this change `node.exportJSON()` would always export the final result (other than children).

packages/lexical-website/sidebars.js
```diff
@@ -41,7 +41,7 @@ const sidebars = {
       items: [
         'concepts/editor-state',
         'concepts/nodes',
-        'concepts/node-replacement',
+        'concepts/node-customization',
```
etrepum, line 44: Is this rename worth breaking an existing URL?

packages/lexical/src/LexicalNode.ts
```diff
@@ -185,6 +188,53 @@ export class LexicalNode {
   __prev: null | NodeKey;
   /** @internal */
   __next: null | NodeKey;
+  /**
+   * Don't use this directly, use `this.getClasses()` and `this.mutateClasses()` instead
+   * @internal
+   */
+  __classes?: ReadonlyClasses;
```
etrepum, line 195:
```suggestion
  __classes: ReadonlyClasses;
```

This can be initialized to an empty object in the constructor, and then all other branching based on whether it's undefined or not can be eliminated.

packages/lexical/src/LexicalNode.ts
```diff
@@ -858,6 +911,15 @@ export class LexicalNode {
...
     const element = this.createDOM(editor._config, editor);
+    if (this.__classes) {
+      Object.entries(this.__classes).forEach(([classPrefix, classSufix]) => {
+        if (typeof classSufix === 'string') {
+          element.classList.add(`${classPrefix}-${classSufix}`);
+        } else if (typeof classSufix === 'boolean' && classSufix) {
+          element.classList.add(classPrefix);
+        }
+      });
+    }
```
etrepum, line 922: It also seems strange to me that this isn't just done in the base implementations for createDOM and updateDOM, there's also the question of where exactly this class should go in the presence of getDOMSlot and this might not be the right place.

packages/lexical/src/LexicalNormalization.ts
```diff
@@ -22,10 +22,16 @@ function $canSimpleTextNodesBeMerged(
   const node2Mode = node2.__mode;
   const node2Format = node2.__format;
   const node2Style = node2.__style;
+  const node1Classes = node1.__classes || {};
+  const node2Classes = node2.__classes || {};
```
etrepum, line 26:
```suggestion
  const node1Classes = node1.__classes;
  const node2Classes = node2.__classes;
  const node1ClassKeys = node1Classes !== node2Classes ? Object.keys(node1ClassKeys) : null;
```

packages/lexical/src/LexicalNormalization.ts
```diff
@@ -22,10 +22,16 @@ function $canSimpleTextNodesBeMerged(
...
   return (
     (node1Mode === null || node1Mode === node2Mode) &&
     (node1Format === null || node1Format === node2Format) &&
-    (node1Style === null || node1Style === node2Style)
+    (node1Style === null || node1Style === node2Style) &&
+    Object.keys(node1Classes).length === Object.keys(node2Classes).length &&
+    Object.keys(node1Classes).every(
+      (key) => node1Classes[key] === node2Classes[key],
+    )
```
etrepum, line 34: Checking the equivalence between to classes objects should really be in a function where it's easier to early exit and use a more performant for loop instead of `every`

```suggestion
    (node1ClassKeys === null ||
      (node1ClassKeys.length === Object.keys(node2Classes).length &&
        node1ClassKeys.every((key) => node1Classes[key] === node2Classes[key])))
```

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -632,8 +641,16 @@ function $reconcileNode(
...
+  const prevClasses = prevNode.__classes || {};
+  const classesChanged = !(
+    Object.keys(nextClasses).length === Object.keys(prevClasses).length &&
+    Object.keys(nextClasses).every(
+      (_key) => nextClasses[_key] === prevClasses[_key],
+    )
+  );
   // Update node. If it returns true, we need to unmount and re-create the node
-  if (nextNode.updateDOM(prevNode, dom, activeEditorConfig)) {
+  if (nextNode.updateDOM(prevNode, dom, activeEditorConfig) || classesChanged) {
```
etrepum, line 653: Do we really want to re-create the node every time classes change? That seems odd.

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -632,8 +641,16 @@ function $reconcileNode(
...
   }
 
+  const nextClasses = nextNode.__classes || {};
+  const prevClasses = prevNode.__classes || {};
+  const classesChanged = !(
+    Object.keys(nextClasses).length === Object.keys(prevClasses).length &&
+    Object.keys(nextClasses).every(
+      (_key) => nextClasses[_key] === prevClasses[_key],
+    )
+  );
```
etrepum, line 651: This should all be in an optimized function per previous comment in LexicalNormalizer

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -173,6 +173,15 @@ function $createNode(key: NodeKey, slot: ElementDOMSlot | null): HTMLElement {
     invariant(false, 'createNode: node does not exist in nodeMap');
   }
   const dom = node.createDOM(activeEditorConfig, activeEditor);
+  if (node.__classes) {
```
etrepum, line 176: The more I look at this the less I like that this adds DOM manipulation to every node that's outside of createDOM and doesn't give the author of the node class any control over what's happening

etrepum (commented):
I think overall I'm -0 on this proposal. I like the idea behind it, I think it would be great to have a bit more generic control over DOM nodes without subclassing (especially when mixing in some additional state). However, this approach seems like it's baked in at an awkward level of abstraction. Technically I think it should work if other maintainers agree that this is the approach we should adopt.

I would prefer to have something that is integrated with the normal lifecycle of createDOM/updateDOM/exportDOM. Yes, this would require more careful plumbing, and wouldn't necessarily "just work" for third party node classes that aren't calling the super methods to a base class in lexical or otherwise have explicit support for this, but I think that the trade-off for consistency and future-proofing would be worth it.

etrepum:
Regarding prefixes I think the simplest thing to do would be to document that the two-underscore prefix is reserved for use by Lexical, and implementing their own classes in the same style may result in namespace collisions. However, we also add methods occasionally without any sort of prefix, so I guess the real answer is that users just have to live with possible compatibility issues over time unless they do their own prefixing because that's just how OOP works.

There are plenty of use cases for aria roles, dataset, and other attributes (particularly id, maybe title or microdata related attributes) with the existing nodes. For aria we don't really have semantic elements so you may want to specify the role or region. For other attributes the most obvious one would be `id`, why shouldn't you be able to add an id attribute anywhere for deep linking purposes, especially when the editor is in readonly mode or for export? Makes just about as much sense as being able to add classes anywhere, and it requires storing state on the node so it's not something you can solve with pure html.export configuration. The dataset property has use cases similar to classes, for things where this metadata should be invisible. You could also technically use dataset (or any other arbitrary attribute scheme) to target elements with CSS, so if this code managed the dataset property then it could be used for the exact same purpose as this (but would arguably be a bit more useful, since you can store arbitrary key/value string pairs in dataset, so it can be used for things like colors or unique/external ids).

packages/lexical/src/LexicalReconciler.ts
GermanJablo: It is an opt-in feature. If you do not use setClass none will appear. 
The same happens with `__indent`.

packages/lexical/src/LexicalNormalization.ts
GermanJablo: `Array.every` exits early if it finds a counterexample

packages/lexical/src/LexicalReconciler.ts
GermanJablo: > `Array.every` exits early if it finds a counterexample

etrepum:
You're right that extending nodes is not composable, so maybe the right solution is to come up with an extension mechanism that is composable rather than hack in support to only mess with the class attribute in particular.

etrepum:
If we add an arbitrary bag of data that is guaranteed to be serialized/deserialized (at least with JSON) to the LexicalNode, then you could write a plugin that performs everything this proposal accomplishes with a similar interface, but also unlocks other possibilities.

* `$setNodeClass(node, className, value)` would update the data structure on the node, e.g. `node.setDataKey('classes', (prev) => ({...prev, [className]: value}))`
* The ClassPlugin registers an update listener to traverse dirty nodes and update the classList in their DOM accordingly (but you could also use a similar plugin to update any DOM for any node, or sync their changes with some external thing, based on other properties).

Missing pieces are:

* Arbitrary data property for each node that participates in the EditorState lifecycle (e.g. getWritable before updating, gets cloned, etc.)
* JSON serialization for the data property (would either need to be plumbed correctly, or hacked in like children are or like this proposal did for classes)

For full fidelity you'd probably want similar extension points for importDOM and exportDOM, something like middleware where you are able to get the result of what would happen without the plugin, and then have the opportunity to modify the result based on the context. This would probably be useful for other purposes.

This data property would also be supported by RootNode (just by virtue of where it would be implemented in the class hierarchy) so we would also finally have a way to store versioned document level data too (this pretty much covers any use case for isolated decorator nodes which don't really work).

GermanJablo:
I find the idea interesting, although it feels like a major abstraction for a feature that no one has asked for and that it could be implemented without breaking changes even if this PR had already been merged. 

As long as the JSON serialization ends up being the same, you could have the styles stored in the format that exists today, and the classes in that of this PR. 

And speaking of format, one of the things I like most about this PR is that classes follow the key-value format with the option to result in prefix-suffix, since manipulating an array of strings is often awkward. Maybe for things whose real abstraction is not a boolean but a set of options like `background-color` it could be used `data-key=value`, but I don't know.

You also mentioned a ClassPlugin. I would like to avoid that. However this is implemented, I would love for users to be able to set a class to the node and be done. 

> For full fidelity you'd probably want similar extension points for importDOM and exportDOM, something like middleware where you are able to get the result of what would happen without the plugin, and then have the opportunity to modify the result based on the context. This would probably be useful for other purposes.

Same thing I said just now. I would love exportDOM to work out of the box. One of the main points of all this is to reduce the complexity of customizing nodes. ImportDOM is trickier because the content can come from an external source like the clipboard, but for that there is already [this option](https://github.com/facebook/lexical/pull/4254).

zurfyx:
Thank you German, we appreciate your contributions as usual! We discussed this PR in the previous session.

Overall, there's some benefit in the newly introduced `setClass` API, and it does without any doubt make live easier in certain scenarios. However, we would be keen to have this `class` shortcut as a separate module rather than built-in into the core. Here's why:

**Cost**
This feature can be seen as an enhancement, a shortcut to CSS classes. Plugins can replicate this same behavior by accessing the rendered DOM element and applying the classes directly. It's inconvenient but the fact it's redundant and low number of times that this happens makes it hard to justify the budget size increase (even if rather small on the core). 

The recent DOM Slot API would instead fall into the category of low-use but a must-have to enable certain use cases.

Similarly, I'm personally not keen on the `update` extra props redundancy but this is unfortunately the product of legacy, since the `$` functions came later.

**Backward compatibility**
This PR introduces opinionated behavior that might not necessarily work well with existing products. The `classes` prop can potentially conflict and there is no workaround, and a few other items identified in your [discussion above](https://github.com/facebook/lexical/pull/6929#pullrequestreview-2490032787).

It is fair to say we have done many backward incompatible changes in the past year but storage is an area that must be backward compatible, or at least fully understand the impact that these can have on existing products.

---

We still think there is value in the class shortcut, do you think we should explore it as a utility function? We can also use the VC time this week to discuss it over if you prefer, I would love to hear more about your actual use case

GermanJablo:
Hi @zurfyx, thanks for the review!

> Plugins can replicate this same behavior by accessing the rendered DOM element and applying the classes directly.

Even if we patched the DOM outside of Lexical, the classes would have to be properties of the nodes. How are you visualizing the persistence of the classes if this is not the case?

> but the fact it's redundant and low number of times that this happens makes it hard to justify the budget size increase

What do you mean by redundant and budget size? I understand that Meta or other users may have lost interest if they have already achieved their goals in other ways. But I am trying to understand the disadvantages of introducing something like this. The cost in bundle size and runtime performance is negligible.

> The recent DOM Slot API would instead fall into the category of low-use but a must-have to enable certain use cases.

It helps that you talk about it as a "must-have". That PR started after my proposal for a scrollable-node. Yes, `DOMSlot` is more convenient, but it was possible to achieve its goals in another way without introducing a new API.

If that API was accepted, I think this one has much more merit to be so. As I said before, "I have reviewed all the nodes I have replaced and talked to other users who have heavily customized their nodes, and in all cases we have found that this API could have saved us the work.".

And to put into context the usefulness of this, it would be good to remember everything a user has to do today if he wants to make a colored TextNode:

<details>
  <summary>Expand to see example</summary>

```ts
export type SerializedColoredTextNode = Spread<
{
  color: string;
  type: 'colored';
  version: 1;
},
SerializedTextNode
>;

export class ColoredNode extends TextNode {
__color: string;

constructor(text: string, color?: string, key?: NodeKey): void {
  super(text, key);
  this.__color = color || "defaultColor";
}

static getType(): string {
  return 'colored';
}

setColor(color: string) {
  self.__color = color;
}

getColor(): string {
  return self.__color;
}

static clone(node: ColoredNode): ColoredNode {
  return new ColoredNode(node.__text, node.__color, node.__key);
}

  const element = super.createDOM(config);
  element.style.color = this.__color || "defaultColor";
  return element;
}

updateDOM(prevNode: ColoredNode, dom: HTMLElement, config: EditorConfig): boolean {
  const isUpdated = super.updateDOM(prevNode, dom, config);
  if (prevNode.__color !== this.__color) {
    dom.style.color = this.__color;
  }
  return isUpdated;
}

static importJSON(serializedNode: SerializedMentionNode): MentionNode {
  const node = new ColoredNode(serializedNode.text, node.serializedNode.color);
  node.setFormat(serializedNode.format);
  node.setDetail(serializedNode.detail);
  node.setMode(serializedNode.mode);
  node.setStyle(serializedNode.style);
  return node;
}

exportJSON(): SerializedHeadingNode {
  return {
    ...super.exportJSON(),
    color: this.getColor()
    tag: this.getTag(),
    type: 'colored',
    version: 1,
  };
}
}

export const nodes = [
  ColoredNode,
  {
    replace: TextNode,
    with: (node: TextNode) => new ColoredNode(node.__text),
    withKlass: ColoredNode,
  },
  // ...
]
```

</details>

I could list 100 common mistakes that aren't obvious and could go wrong when doing that. It doesn't seem like this should be the preferred way to customize a node. But still, this isn't my main reason for wanting to introduce a new API (more on that below).

> Backward compatibility

Are you seeing some incompatibility that I am not? How could this API break existing products?

> I would love to hear more about your actual use case

Improving DX is important, no doubt. But my main interest is what I mentioned above: "extending nodes is not composable. For example, in Payload the community can create their own plugins. If two of them extend the same node, it doesn't work".

This may not be crucial for a user like Meta who has full control over the editor's source code. But on platforms like ours this composability is crucial.

GermanJablo:
To be clear, I am super fine with introducing a more general or “holistic” API like the one you mention. Also, I think you have convinced me that `data-key=value` is more appropriate for key-value objects.  I can try to make a proof of concept. Do you have an idea of what the API would look like? Maybe something like this?

```ts
type Attributes = {
  classList: string[],
  id: string,
  data: Record<string, string>,
  aria: Record<string, string>,
}

class LexicalNode {
  attributes: Attributes
}
```

About the second paragraph, the problem is that it would break users who are already extending TextNode.

etrepum:
Regarding existing users who have already extended `TextNode` directly, I don't think that Lexical's reconciler should perpetually carry this burden to avoid a one-time refactor that would give them this extra functionality. Lexical is not 1.0. While backwards compatibility for a period of time is always a critical concern to support its major consumers, it's expected that the API is going to continue to evolve and projects will have to track it especially if they want new features.

If you really did want to simplify the `ColoredTextNode` implementation you pasted above all you would have to do is write two functions `$getTextColor` and `$setTextColor` that parse and manipulate the style. Something like this:

```ts
// getStyleObjectFromCSS and getCSSFromStyleObject are not currently exported but are in lexical-selection
function $getTextColor(node: TextNode): string {
  return getStyleObjectFromCSS(node.getStyle()).color || 'defaultColor';
}

function $setTextColor(node: TextNode, color: string): void {
  node.setStyle(getCSSFromStyleObject({...getStyleObjectFromCSS(node.getStyle()), color}));
}
```

I think it would be a bad idea to commit to a specific representation for managing attributes, because it prevents future optimization and refactoring. I think the API should be some combination of methods and/or $functions to read or manipulate them. IIRC in order to support today's collab this data structure would have to be both JSON serializable and read-only (or managed carefully with properties).

There are a couple things to consider here:

* Many plug-in use cases really just need per-node state that ends up in EditorState, and that storage doesn't necessarily need to be also serialized to the DOM
* A flat classList representation doesn't necessarily cover all use cases well, particularly since some people use utility classes. I think one of the roughest edges right now is that there isn't a more managed way to do this, if you scan the code for `removeClassNamesFromElement` and audit carefully you can find a lot of potential issues when there are overlapping sets of classes based on states. In your use case the hyphenated BEM style modifiers would also fall into this sort of bucket, which is why you were incentivized to come up with the opinionated approach that would let you determine the diffs and allow only one at a time by prefix.
* There are differing opinions on how things like classes should work for export, import, and managed editor DOM (e.g. #6968).

I don't have a concrete proposal for an API that covers everything well.

GermanJablo:
Yes, it works for styles, but that wasn't the point of the example. Any case where you need to add another common property (like a class) to a node requires all that ceremony.

> I think it would be a bad idea to commit to a specific representation for managing attributes, because it prevents future optimization and refactoring. I think the API should be some combination of methods and/or $functions to read or manipulate them.

That's a good point. Many DOM APIs make a lot of sense in specific contexts and could be easily replicated. The lowest level ones are `setAttribute` and `getAttribute`. Since they would be needed to provide something generic and the more specific ones can be derived from them, they seem like a good starting point.

My suggestion would be to provide `setAttribute` and `getAttribute` methods. JSON serialization would be a plain string object, just like you see in HTML, and Lexical should take care of that serialization as I do in this PR.

That would allow us to optimize or refactor some APIs later. For example, if we wanted to add `classList.toggle/add/remove/contains`, we could modify the `importJSON` and `exportJSON` of `attributes.class` to be a `Set` for performance reasons.

GermanJablo:
closed in favor of https://github.com/facebook/lexical/pull/7117

PR #6934, opened 2024-12-10
[lexical-playground] Refactor: editor styles should in PlaygroundEditorTheme.css

umaranis:
## Description

All editor styles are defined are defined in `PlaygroundEditorTheme.css` but two that are in `index.css`. This PR moves the two styles to their right place.


ivailop7 (commented):
You should include the animation as well.

umaranis:
@ivailop7 , shouldn't have missed that. Thanks for pointing out.

CursorBlink animation is included now.

etrepum (approved):

PR #6942, opened 2024-12-11
[lexical-yjs] Feature: Expose function to get anchor and focus nodes for given user awareness state

amanharwara:
This PR exposes a `getAnchorAndFocusCollabNodesForUserState` function extracted from the internal sync cursor functions that allows getting the anchor/focus nodes and offsets when provided a binding and a user state.

We want to implement a "scroll to user cursor" feature, but currently there is no way to get the anchor and focus nodes for a given user as a consumer of Lexical (except for the local, of course).

This also allows for potential custom implementations of collab cursors since the implementation provided in `@lexical/yjs` is not very customizable.

Screen recording confirming collab cursors work even after the change:

https://github.com/user-attachments/assets/ff7b186d-4b76-4231-b128-a5b6f044941a


amanharwara:
I can't reproduce the tests failing for both the commits locally. The tests failing locally when running `npm run test-e2e-collab-prod-chromium` are different from these, and unrelated to the change made here from what I can tell.

Anything else I need to do here?

packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -423,41 +448,30 @@ export function syncCursorPositions(
...
+              focusKey,
+              focusOffset,
+            );
+          } else {
+            const anchor = selection.anchor;
+            const focus = selection.focus;
+            anchor.key = anchorKey;
+            anchor.offset = anchorOffset;
+            focus.key = focusKey;
+            focus.offset = focusOffset;
```
etrepum, line 474: I understand that this is just copied from what was there before, but this should be using `$setPoint`  instead of setting the key and offset directly (setting of `type` is missing here). If the type of node changes this is incorrect and will cause weird bugs or crashes.

```suggestion
            $setPoint(selection.anchor, anchorKey, anchorOffset);
            $setPoint(selection.focus, focusKey, focusOffset);
```

packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -295,50 +295,75 @@ function updateCursor(
...
+    const selection = $getSelection();
+
+    if (!$isRangeSelection(selection)) {
+      return;
     }
+    const anchor = selection.anchor;
+    const focus = selection.focus;
+
+    $setPoint(anchor, anchorKey, anchorOffset);
+    $setPoint(focus, focusKey, focusOffset);
```
etrepum, line 366: No reason to have variables for these points, just adds more code

```suggestion
    $setPoint(selection.anchor, anchorKey, anchorOffset);
    $setPoint(selection.focus, focusKey, focusOffset);
```

packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -295,50 +295,75 @@ function updateCursor(
...
-  binding: Binding,
-  provider: Provider,
-): void {
-  const awareness = provider.awareness;
-  const localState = awareness.getLocalState();
+type AnyCollabNode =
+  | CollabDecoratorNode
+  | CollabElementNode
+  | CollabTextNode
+  | CollabLineBreakNode;
```
etrepum, line 302: It would make sense to also use this below in the definition of `getCollabNodeAndOffset`

etrepum (commented):
This seems like a good improvement but let's tighten it up a little bit before merge

packages/lexical-yjs/src/SyncCursors.ts
amanharwara: `selection.anchor` and `selection.focus` in the cursor selection aren't actual `Point`s and the `type` property isn't used anywhere where the cursor selection is used.

amanharwara: Refactoring to use actual `Point`s is of course possible, but I think it'd be doing unnecessary work for something that's not used. For e.g., to initially create the point you'd need to get the anchor and focus nodes from the editor state to be able to check whether the type should be `element` or `type`.

amanharwara: I tried refactoring to use Points [here](https://github.com/facebook/lexical/pull/6942/commits/d6f74b6923a2678faa59fbf633f4e6019a0f5f72), but have reverted that again since it seems to be making some extra changes like automatically deprecating and renaming syncCursorPositions because of rules-of-lexical lint rules.

etrepum: my fault, there's just not enough context the GitHub PR review UX for me to easily see that this kind of selection isn't a real selection, thanks for explaining!

etrepum (approved):

PR #6944, opened 2024-12-12
[lexical-list] Revert PR 6912

potatowagon:
Receiving reports about unexpected empty bullet lines appearing in the preview of the workplace editor

Screenshot 2024-12-11 at 10 00 33 PM

the sync diff containing this PR got reverted to mitigate the issue. Reverting this PR because:

- more discussion is needed about list item node inheriting from paragraph node (it is odd)
- not worth trading highly visible breaking changes for a rather edge case styling fix


etrepum:
Were you testing this with #6917 also applied? There's a fix in there that I believe is related to this issue

https://github.com/facebook/lexical/pull/6917/files#diff-ebf55b0878052c62d91299c29e05b3f3ef9026a05bc7881a7f0d7bb37ed2d0c9L514-R520

etrepum:
As far as rationale goes, the ListItemNode is basically supposed to support everything that a ParagraphNode does. Unlike a TableCellNode, it is not a shadow root that is expected to have children that are ParagraphNode, so to support all of the things that ParagraphNode supports then it should either be a ParagraphNode or separately implement all of the things that ParagraphNode implements, but the problem is that you can't just copy and paste that because the extensibility is not present in selections and the reconciler to make that work, there are $isParagraphNode special cases that can't be extended outside of the lexical package.

If we really don't want nodes like this to subclass ParagraphNode then we should move the features that ParagraphNode has into some new common base class, e.g. LexicalNode -> ElementNode -> *BlockNode* -> (ParagraphNode | ListItemNode) so that this sort of behavior can be inherited and baked into the lower levels without introducing dependencies from lexical to @lexical/list or elsewhere.

Sahejkm (approved):

potatowagon:
> Were you testing this with #6917 also applied? There's a fix in there that I believe is related to this issue
> 
> https://github.com/facebook/lexical/pull/6917/files#diff-ebf55b0878052c62d91299c29e05b3f3ef9026a05bc7881a7f0d7bb37ed2d0c9L514-R520

thanks for pointing out #6917. the sync diff didnt contain that fix. in that case ill put this revert on hold and test with that patch

potatowagon:
i tested with PR 6917 applied but it did not fix the issue sadly

Screenshot 2024-12-11 at 10 55 23 PM

ill proceed with this revert first

Sahejkm (approved):

potatowagon:
> If we really don't want nodes like this to subclass ParagraphNode then we should move the features that ParagraphNode has into some new common base class, e.g. LexicalNode -> ElementNode -> _BlockNode_ -> (ParagraphNode | ListItemNode) so that this sort of behavior can be inherited and baked into the lower levels without introducing dependencies from lexical to @lexical/list or elsewhere.

this might work better to prevent breaking changes from $isParagraphNode(...)

PR #6947, opened 2024-12-12
[@lexical/utils] Bug Fix: cursor positioning issue with CJK

### LINKED ISSUE #6946: Bug: wrong cursor position with CJK keyboard
`markSelection` utility function affects selection (cursor position) when I use CJK keyboards.

Lexical version: 0.21.0

## Steps To Reproduce

[Demo video](https://drive.google.com/file/d/1JRYiracJf1j2K0Z3uHfQyOLx1Zq6q80Z/view?usp=sharing)

1. Visit [demo](https://codesandbox.io/p/sandbox/confident-germain-9qjp55?file=%2Fsrc%2FEditor.js)
2. Type using CJK keyboard ([How to add CJK keyboard to MacOS?](https://www.koreanfluent.com/cross_cultural/korean_keyboard/korean-keyboard-mac-os-x.htm))

Link to code example:
https://codesandbox.io/p/sandbox/confident-germain-9qjp55?file=%2Fsrc%2FEditor.js

## The current behavior
Cursor jumps to beginning

## The expected behavior
Preserve cursor position where you type.

## Impact of fix
I use `markSelection` to highlight selection when the editor loses focus. It's very important feature for a better user experience. In fact, I believe it should be out of the box feature from Lexical.

mrdivyansh:
## Description
It fixes #6946. However, I am not sure why TextNode.updateDOM should cause any side-effect. 

Closes #6946

### DIFF
--- a/packages/lexical-utils/src/markSelection.ts
+++ b/packages/lexical-utils/src/markSelection.ts
@@ -9,7 +9,6 @@
 import {
   $getSelection,
   $isRangeSelection,
-  $isTextNode,
   type EditorState,
   ElementNode,
   getDOMTextNode,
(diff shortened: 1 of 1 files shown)


etrepum (commented):
I think you're right that this updateDOM call doesn't really make sense. What updateDOM does is return true if the node should be re-mounted, otherwise it returns false and does some in-place mutations (which is what is likely causing the issues here). If it would've returned true you can detect that because that DOM would be unmounted and createDOM would be called to create a new one so currentAnchorNodeDOM !== previousAnchorNodeDOM would be true. I think the proper fix would be to track those previous DOM variables in the closure outside of the compute function just like previousAnchorNode and previousFocusNode. This way the updateDOM call can be eliminated but the intent of that check can be preserved.

/cc @zurfyx

etrepum:
@mrdivyansh I created a draft PR #6961 with what I think this should look like, but I can't easily repro this bug so if you could take a look that would be great

PR #6948, opened 2024-12-12
[lexical-playground] [markdown] Fix: Block math equation is falsely converted to inline equation on MarkdownShortcutPlugin

### LINKED ISSUE #6936: Bug: Block math equation is falsely converted to inline equation on MarkdownShortcutPlugin

## Steps To Reproduce

2. Press the Markdown button

## The current behavior

## The expected behavior

## Potential Solutions

## Impact of fix

vantage-ola:
Closes #6936


vantage-ola:
https://github.com/facebook/lexical/issues/6936#issuecomment-2539436779   @etrepum

etrepum (commented):
I'm not sure if this is useful or not, but it certainly doesn't fix the bug reported in #6936

vantage-ola:
as of now, i haven't figured out the regex yet. the behaviour of the lexical playground as of now

Uploading Recording 2024-12-18 182216.mp4…

etrepum:
Are you planning to continue working on this, or should it be closed?

vantage-ola:
@etrepum I lost interest. Could not figure out the fix

etrepum:
No worries, thank you for the attempt!

PR #6966, opened 2024-12-14
[lexical-playground] Bug Fix: Allow scrolling if the table cell content overflows

### LINKED ISSUE #6931: Bug: Image becomes unresizable if embedded inside a table
Lexical 0.21

Trying to resize an image when inside a TableCell does not work:

https://github.com/user-attachments/assets/fc07f84a-fe1d-47f3-b30e-51b98d188be4

Parasaran-Python:
fix 6931: CSS changes in the playground to allow scrolling if the cell content overflows, and inheriting the same behavior for the nested `<p>` tag.

Bug Fix
## Description

Made changes to the playground CSS to allow scroll of table content, since there's a nested p tag, have added scroll:inherit to it

Closes #6931

## Test plan
Try scrolling of various contents inside playground table's cells
### Before

image

### After

https://github.com/user-attachments/assets/8605cc74-b300-4cda-bcc4-e2fa259c1d4f

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -15,6 +15,7 @@
 .PlaygroundEditorTheme__paragraph {
   margin: 0;
   position: relative;
+  overflow: inherit;
 }
 .PlaygroundEditorTheme__quote {
   margin: 0;
@@ -213,6 +214,7 @@
   padding: 6px 8px;
   position: relative;
   outline: none;
+  overflow: auto;
 }
 .PlaygroundEditorTheme__tableCellResizer {
   position: absolute;
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -15,12 +15,6 @@
 .PlaygroundEditorTheme__paragraph {
   margin: 0;
   position: relative;
-
-  /*
-    A firefox workaround to allow scrolling of overflowing table cell
-    ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1904159  
-  */
-  overflow: inherit;
 }
 .PlaygroundEditorTheme__quote {
   margin: 0;
@@ -220,6 +214,13 @@
   position: relative;
   outline: none;
   overflow: auto;
+}
+/*
+  A firefox workaround to allow scrolling of overflowing table cell
+  ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1904159
+*/
+.PlaygroundEditorTheme__tableCell > * {
+  overflow: inherit;
 }
 .PlaygroundEditorTheme__tableCellResizer {
   position: absolute;
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -15,6 +15,11 @@
 .PlaygroundEditorTheme__paragraph {
   margin: 0;
   position: relative;
+
+  /*
+    A firefox workaround to allow scrolling of overflowing table cell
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -15,6 +15,7 @@
 .PlaygroundEditorTheme__paragraph {
   margin: 0;
   position: relative;
+  overflow: inherit;
```
etrepum, line 18: In what situations would it be useful for this nested p tag to scroll separately?

etrepum (approved):
Overall I think the overflow is useful for table cells, but unsure why it was also applied to the nested p tag. Can you explain or demonstrate why that is useful?

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
Parasaran-Python: without this the image doesnt scroll, the p tag seems to be blocking the scroll of the image.

I could have very well applied the overflow prop to the p tag directly, but i felt it made more sense to apply it to td tag, and inherit to the child p tag

etrepum: Yes, the td tag needs it for sure. If remove this change with dev tools in a locally copy I can still scroll the cell horizontally. The only difference from here appears to be that the image gets clipped by the td padding on the right when it is enabled.

https://github.com/user-attachments/assets/b157583d-844b-4afe-a2ef-2b71a9d93ef0

Parasaran-Python: Works for me as well in chrome, didnt work for me in firefox, hence I had to add it.

Seems like this is a browser specific thing, I hope even you would face the same in a firefox instance.

Not sure how the behavior would be on safari!

etrepum: This does look like a firefox-specific issue, which is "fixed" but currently behind a feature flag that defaults to false https://bugzilla.mozilla.org/show_bug.cgi?id=1904159

etrepum: I would probably comment it out and put a note in there about Firefox compatibility since it looks a little weird in other browsers. I think if we set this now without any commentary this firefox workaround will just sit around long after Firefox supports table cell scrolling correctly by default.

Parasaran-Python: That's a great example of why cross-browser testing is so important! 😅

Regarding this PR, it might be safer to go ahead with this change, unless there's a risk of any potential breakage. What are your thoughts?

Parasaran-Python: Sure let me do that.

Parasaran-Python:
Comment has been added @etrepum  requesting you to re-review. 🙂

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
etrepum: Now that I've had a bit more time to think about this, let's move this to a separate CSS rule that specifically targets any direct descendant of the table cell. This both covers the intent of the change and is more correct (will work for other top-level blocks, not just ParagraphNode).

```css
/*
  A firefox workaround to allow scrolling of overflowing table cell
  ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1904159
*/
.PlaygroundEditorTheme__tableCell > * {
  overflow: inherit;
}
```

Parasaran-Python: Yes, make sense, will try this.

Parasaran-Python: This is done.

etrepum (approved):
Works great, thank you!

PR #6970, opened 2024-12-16
[Breaking Change][lexical] Feature: Add updateFromJSON and move more textFormat/textStyle to ElementNode

### LINKED ISSUE #5708: ElementNode format
There is a case that works well in three major editors: Word, GDoc and Pages that Lexical still doesn't do quite right. Preserving the format when there's no text.

While #5707 addresses part of the issue (retention of selection) it doesn't fix the retention while moving away.

There's a key ingredient that's missing here: persistance.

In HTML, browsers do this with:

```
<p>
  <b>
  </b>
</p>
```

In Lexical we never wanted to introduce this DOM structure because of redundancy, a11y (and the fact that we don't support nested HTML).

But we can do better, we can persist this format in the ElementNode. In the Lexical case, ElementNode would only store and use this in particular cases when the selection touches on the entirety of the Element. These are:

1. Collapsed selection on an empty block ElementNode
2. Entire block ElementNode selected (i.e. triple click)

### LINKED ISSUE #6476: Bug: Selection formatting/styling is lost when you indent a list

### LINKED ISSUE #6949: Feature: Preserve original format after indenting list item
Implement #6912 / #6944 but with a common base class for ParagraphNode and ListItemNode

See also:

_Originally posted by @etrepum in https://github.com/facebook/lexical/issues/6912#issuecomment-2539681890_

Closes: #6476

etrepum:
## Description

One of the biggest pain points and error-prone parts of extending lexical is the amount of boilerplate involved in subclassing nodes. In particular the worst of it is the `static importJSON` method that must be implemented. We can eventually make the boilerplate go away with some targeted refactoring.

The majority of this PR implements `updateFromJSON` which is the `importJSON` analog to the `afterCloneFrom` for `clone` from #6505. Basically the key concept here is to move all of the logic, other than the constructor itself (with as many defaults as possible), into an instance method, so that super can be called to inherit all of the base class logic.

More documentation has been added related to serialization best practices, including a recommendation *not* to use the `version` property as it does not compose well and is generally just ignored and/or redundant.

Before:

```ts
class BaseNode {
  static importJSON(serializedNode: SerializedBaseNode): BaseNode {
    const node = $createBaseNode();
    /* … a whole bunch of boilerplate that must be copied and pasted in subclasses */
    return node;
  }
}

class ExtendedNode extends BaseNode {
  static importJSON(serializedNode: SerializedExtendedNode): ExtendedNode {
    const node = $createExtendedNode();
    /* … a whole bunch of boilerplate that is copied from BaseClass */
    /* … some extra stuff specific to SerializedExtendedNode */
    return node;
  }
}
```

After:

```ts
class BaseNode {
  // We can eventually eliminate this boilerplate altogether, but makes sense in the short term
  static importJSON(serializedNode: SerializedBaseNode): BaseNode {
    return $createBaseNode().updateFromJSON(serializedNode);
  }

  updateFromJSON(serializedNode: Omit<SerializedBaseNode, 'type' | 'version'>): this {
    // for the true base cases of TextNode and ElementNode this will not have a super call
    const node = super.updateFromJSON(serializedNode);
    /* … the previous boilerplate, that can now be inherited */
    return node;
  }
}

class ExtendedNode extends BaseNode {
  // We can eventually eliminate this boilerplate altogether, but makes sense in the short term
  static importJSON(serializedNode: SerializedExtendedNode): ExtendedNode {
    return $createExtendedNode().updateFromJSON(serializedNode);
  }

  // this is now *optional* and can be implemented when extra setters are needed
  updateFromJSON(serializedNode: Omit<SerializedExtendedNode, 'type' | 'version'>): this {
    const node = super.updateFromJSON(serializedNode);
    /* … some extra stuff specific to SerializedExtendedNode */
    return node;
  }
}
```

The motivation for this was to move textFormat and textStyle to ElementNode, which to do properly would require also updating serialization everywhere, so it made sense to eliminate the boilerplate at the same time.

Note that the types here are not quite sound and are probably not possible to type in Flow, for the same class of reason that `updateDOM` and `afterCloneFrom` are also not sound. When a subclass implements this method with a more specific type, it's no longer sound to upcast it to the base class and call the method with the schema intended for the base class. I'm sure it would be possible to come up with a scheme that is sound, but I think it would require a large DX trade-off one way or another (such as using a new method name for each class like `updateFromTextNodeJSON`, moving it to a static method or exported function that must be called fully qualified, or forcing the `updateFromJSON` implementations to parse the JSON from something like `Record<unknown, unknown>` - which would be much safer but also very tedious with a runtime penalty). See #6998 for more information about this and some ideas for runtime checks.

Closes #6949
Closes #6476
Closes #5708 (although it does not implement a perfect solution to persist styles at arbitrary positions, it's better than status quo)

Might get closer to fixing #6583

Related: #3931

## Upgrade Notes

This change adds optional `textFormat` and `textStyle` properties to `SerializedElementNode`. If you have existing classes with those properties it could create a namespace clash that you will have to resolve one way or another.

TextNode and ElementNode subclasses should be updated to call the `updateFromJSON(serializedNode)` method from their `static importJSON` methods. If they don't, they won't support this new functionality, and will have to continue copy and pasting the super implementation of importJSON for correct behavior if the base class ever changes in the future.

You should consider dropping usage of the `version` field.

## Test plan

* Existing unit and e2e tests all pass (serialization tests and tests of the exact instance representation were modified)
* New e2e test for ListItemNode behavior
* Other tests and code were refactored to use this new facility, so it's well covered by everything else too

### Before

Styles were not preserved when creating a new ListItem

### After

Styles are preserved when creating a new ListItem (including e2e test)


zurfyx:
Thanks Bob, the `static` import has indeed been a problem for a very long time, we had this similar issue for clone where we hardcoded some properties just to make it easier for the common cases, I believe you fixed this one.

I like the solution you've presented and how it works with types!

I'm not so sure about the moving `textFormat` and `textStyle`, is this really what need? When is format relevant for an arbitrary element?

etrepum:
The textFormat and textStyle is only relevant to an element because we don't have another reliable way to preserve it when there isn't a TextNode present. Previously we were doing this only for ParagraphNode, but it probably makes sense for any element that can contain a TextNode (such as ListItemNode in this particular use case). It's not used directly on the element for any purpose, it's really just storage for the selection. Possibly a future direction might be to phase it out and insert some kind of format preservation node (a special empty TextNode that doesn't get normalized away or maybe even rendered, for example).

With regard to methods, the only real reason we need to add methods here is backwards compatibility. If we didn't have that constraint we could get rid of `static clone` and `static importFromJSON` and replace them with only `afterCloneFrom` and `updateFromJSON` which are both optional and only needed if you are adding state to the node. `exportJSON` could also be optional if we had the base classes use `type: this.getType()`. Implementing the property overrides for the collection of `afterCloneFrom`, `updateFromJSON`, `exportJSON` (and all of the property getters and setters) could be done with code gen at runtime (e.g. with something like a decorator) or compile time possibly with the assistance of a fourth method that just listed the properties that need to be carried around (but of course overriding could also customize this).

I think the next step here is to allow classes to drop `static clone`. The strategy here would be to deprecate the nodeKey argument and allow `$setNodeKey` to have some module global state for what the next `nodeKey` is that is set and change the implementation of `$cloneWithProperties` accordingly

```ts
  // before
  const mutableNode = constructor.clone(latestNode) as T;
  mutableNode.afterCloneFrom(latestNode);
```

```ts
  // after
  let mutableNode: T;
  if ('clone' in constructor) {
    // backwards compatibility
    mutableNode = constructor.clone() as T;
  } else {
    $setNextNodeKey(latestNode.getKey());
    // this does require all constructors to have a 0-arg form and a correct afterCloneFrom
    mutableNode = new constructor() as T;
    // if __DEV__ make sure the key was consumed by $setNodeKey(this, key);
  }
  mutableNode.afterCloneFrom(latestNode);
```

Instead of checking that these static methods exist we can instead have a lint that looks for extra properties on a node and if they exist then expect an `afterCloneFrom`, `updateFromJSON`, and `exportJSON` to be defined.

etrepum:
The exportJSON refactor is implemented in #6983

zurfyx (approved):

PR #6866, opened 2024-11-25
[lexical] Bug Fix: Missing variables and functions in Flow file

lytion:
Missing variables and functions in Flow file.

## Description
I found that some variables and function are missing in the lexical Flow file, here is a suggested fix.

PR #6977, opened 2024-12-18
[lexical] Bug Fix: Flow is missing some variables and functions

lytion:
## Description
Some variables and function are missing in flow file.

Here are classes affected:
- EditorThemeClasses
- BaseSelection
- NodeSelection
- RangeSelection
- DOMConversionFn

This is the following of #6866

### DIFF
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -362,7 +362,7 @@ export type DOMConversion = {
   conversion: DOMConversionFn,
   priority: 0 | 1 | 2 | 3 | 4,
 };
-export type DOMConversionFn = (element: Node) => DOMConversionOutput;
+export type DOMConversionFn = (element: Node) => DOMConversionOutput | null;
 export type DOMChildConversion = (
   lexicalNode: LexicalNode,
   parentLexicalNode: ?LexicalNode | null,
--- a/packages/lexical/flow/Lexical.js.flow
+++ b/packages/lexical/flow/Lexical.js.flow
@@ -262,6 +262,7 @@ export type EditorThemeClasses = {
     ulDepth?: Array<EditorThemeClassName>,
     ol?: EditorThemeClassName,
     olDepth?: Array<EditorThemeClassName>,
+    checklist?: EditorThemeClassName,
     listitem?: EditorThemeClassName,
     listitemChecked?: EditorThemeClassName,
     listitemUnchecked?: EditorThemeClassName,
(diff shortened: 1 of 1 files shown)


packages/lexical/flow/Lexical.js.flow
```diff
@@ -469,6 +473,8 @@ declare export class NodeSelection implements BaseSelection {
   dirty: boolean;
   constructor(objects: Set<NodeKey>): void;
   is(selection: null | BaseSelection): boolean;
+  isBackward(): boolean;
+  isCollapsed(): boolean;
```
etrepum, line 477: Is it useful to add this here, when it already declares that BaseSelection is implemented?

packages/lexical/flow/Lexical.js.flow
```diff
@@ -530,6 +538,8 @@ declare export class RangeSelection implements BaseSelection {
   insertNodes(nodes: Array<LexicalNode>): void;
   getCachedNodes(): null | Array<LexicalNode>;
   setCachedNodes(nodes: null | Array<LexicalNode>): void;
+  forwardDeletion(anchor: PointType, anchorNode: ElementNode | TextNode, isBackward: boolean): boolean;
+  getStartEndPoints(): null | [PointType, PointType];
```
etrepum, line 542: This also looks to be the same as BaseSelection

etrepum (approved):

packages/lexical/flow/Lexical.js.flow
lytion: Same as the other comment, if I remove those 2 declarations, `npm run ci-check` return this error

> Cannot implement BaseSelection [1] with RangeSelection because property getStartEndPoints is missing in
RangeSelection [2] but exists in BaseSelection [1]. [prop-missing]

packages/lexical/flow/Lexical.js.flow
lytion: If I remove those 2 declarations, `npm run ci-check` return this error

> Cannot implement BaseSelection [1] with NodeSelection because property isCollapsed is missing in NodeSelection [2] but exists in BaseSelection [1]. [prop-missing]

packages/lexical/flow/Lexical.js.flow
```diff
@@ -451,13 +452,16 @@ declare export function $isBlockElementNode(
 ): node is ElementNode;
 
 export interface BaseSelection {
-  clone(): BaseSelection;
   dirty: boolean;
+  clone(): BaseSelection;
```
potatowagon, line 456: curious about the use cases outside of meta for these added api

just a nit, rearranging clone to be after dirty would break the unenforced abc sorting for props

other then that lgtm

potatowagon (approved):

Husain5796:
### grand dragon 4d first prize 27/12/2024/2933

PR #6983, opened 2024-12-19
Refactor: exportJSON

GermanJablo:
exportJSON refactoring. Advantages:

- Less boilerplate on nodes. With this change many nodes no longer require `exportJSON`, and those that do no longer need to repeat "type" and "version".
- DRY. Any changes we make in the future can be done in a single node and propagated to subclasses. This will make reviewing PRs that modify JSON serialization easier, as far fewer files are modified.

Discussed internally with @etrepum

### DIFF
--- a/packages/lexical/src/__tests__/utils/index.tsx
+++ b/packages/lexical/src/__tests__/utils/index.tsx
@@ -182,13 +182,6 @@ export class TestElementNode extends ElementNode {
     return node;
   }
 
-  // exportJSON(): SerializedTestElementNode {
-  //   return {
-  //     ...super.exportJSON(),
-  //     type: 'test_block',
-  //   };
-  // }
-
   createDOM() {
     return document.createElement('div');
   }
@@ -216,13 +209,6 @@ export class TestTextNode extends TextNode {
   static importJSON(serializedNode: SerializedTestTextNode): TestTextNode {
     return new TestTextNode(serializedNode.text);
   }
-
-  exportJSON(): SerializedTestTextNode {
-    return {
-      ...super.exportJSON(),
-      type: 'test_text',
-    };
-  }
 }
 
 export type SerializedTestInlineElementNode = SerializedElementNode;
@@ -244,13 +230,6 @@ export class TestInlineElementNode extends ElementNode {
     node.setIndent(serializedNode.indent);
     node.setDirection(serializedNode.direction);
     return node;
-  }
-
-  exportJSON(): SerializedTestInlineElementNode {
-    return {
-      ...super.exportJSON(),
-      type: 'test_inline_block',
-    };
   }
 
   createDOM() {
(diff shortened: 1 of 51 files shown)


packages/lexical/src/__tests__/utils/index.tsx
```diff
@@ -182,13 +182,12 @@ export class TestElementNode extends ElementNode {
...
   }
 
-  exportJSON(): SerializedTestElementNode {
-    return {
-      ...super.exportJSON(),
-      type: 'test_block',
-      version: 1,
-    };
-  }
+  // exportJSON(): SerializedTestElementNode {
```
etrepum, line 185: I think all of the nodes defined in this file don't need a custom exportJSON implementation?

packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -792,8 +792,7 @@ export class ElementNode extends LexicalNode {
       direction: this.getDirection(),
       format: this.getFormatType(),
       indent: this.getIndent(),
-      type: 'element',
-      version: 1,
+      ...super.exportJSON(),
```
etrepum, line 795: This should go first in the JSON

packages/lexical/src/nodes/LexicalRootNode.ts
```diff
@@ -107,12 +107,11 @@ export class RootNode extends ElementNode {
 
   exportJSON(): SerializedRootNode {
```
etrepum, line 108: RootNode's exportJSON can be removed now, it's exactly the same as its base class now.

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -652,8 +652,7 @@ export class TextNode extends LexicalNode {
       mode: this.getMode(),
       style: this.getStyle(),
       text: this.getTextContent(),
-      type: 'text',
-      version: 1,
+      ...super.exportJSON(),
```
etrepum, line 655: Move this first

packages/lexical/src/nodes/LexicalElementNode.ts
GermanJablo: That's on purpose. The problem is that it would break the order of the properties and all tests that use stringified editorState would have to be changed.

packages/lexical/src/nodes/LexicalTextNode.ts
GermanJablo: Same as in ElementNode.

packages/lexical/src/nodes/LexicalElementNode.ts
etrepum: Maybe let's put a note in there so that this doesn't come up again, and that if we ever do change the serialized string representation then we can fix them up.

packages/lexical/src/nodes/LexicalRootNode.ts
GermanJablo: Good catch!

GermanJablo:
Comments resolved!

etrepum (approved):

PR #6994, opened 2024-12-24
[lexical-list] Bug Fix: Prevent error when calling formatList when selection is at root

amanharwara:
## Description

Currently, if the selection is at the root and you toggle one of the list options, it throws an error `Error: replace: cannot be called on root nodes`.

This PR fixes this by selecting the first child of the root if available, or creating a new paragraph, appending and selecting it before re-running the `insertList` function.

### Before

https://github.com/user-attachments/assets/af17623e-5240-4552-91f3-cb01222fe703

### After

https://github.com/user-attachments/assets/23e87498-03bd-45bc-a999-a61849854737

### DIFF
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -83,9 +83,7 @@ export function $insertList(listType: ListType): void {
           anchorNode.append(paragraph);
           nodes = paragraph.select().getNodes();
         }
-      }
-
-      if ($isSelectingEmptyListItem(anchorNode, nodes)) {
+      } else if ($isSelectingEmptyListItem(anchorNode, nodes)) {
         const list = $createListNode(listType);
 
         if ($isRootOrShadowRoot(anchorNodeParent)) {
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -9,11 +9,13 @@
 import {$getNearestNodeOfType} from '@lexical/utils';
 import {
   $createParagraphNode,
+  $getRoot,
   $getSelection,
   $isElementNode,
   $isLeafNode,
   $isRangeSelection,
   $isRootOrShadowRoot,
+  $setSelection,
   ElementNode,
   LexicalNode,
   NodeKey,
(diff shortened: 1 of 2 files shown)


etrepum (commented):
I suspect that this class of error would also exist for shadow roots, e.g. TableCellNode. It's also not a good idea to have this function recurse, as it will cause a nested update, and nested updates have weird semantics so this will likely be a source of a different very hard to understand bug.

What I would suggest here is using `$isRootOrShadowRoot` instead of checking anchor.key, and maybe to change the variables from const to let so that you can re-assign them after moving the selection around instead of recursing.

etrepum:
Alternatively if you really did want to recurse, you should extract the update function to a `$function` that is called by `insertList`, and that function could recurse. A unit or e2e test would also be great here to show that the problem is indeed fixed, and to prevent regressions in the future.

amanharwara:
Thanks for the suggestions. Have updated the code to use the `$isRootOrShadowRoot` guard, removed the recursion and added unit tests that cover the behavior.

packages/lexical-list/src/formatList.ts
```diff
@@ -77,6 +77,17 @@ export function insertList(editor: LexicalEditor, listType: ListType): void {
...
+        if ($isRootOrShadowRoot(anchorNode)) {
+          const firstChild = anchorNode.getFirstChild();
+          if (firstChild) {
+            nodes = firstChild.selectStart().getNodes();
+          } else {
+            const paragraph = $createParagraphNode();
+            anchorNode.append(paragraph);
+            nodes = paragraph.select().getNodes();
+          }
+        }
```
etrepum, line 89: It does seem strange that we're not also changing `anchorNode` and `anchorNodeParent` here, but maybe `$isSelectingEmptyListItem` is always false if this is the case? If that's so, I think it would make more sense to change that to be an `else if` to make it clear that these conditions are mutually exclusive.

amanharwara: I've changed it to an `else if ` since I don't think an empty list item should be a direct child of the root anyways.
By the way, do you think a new paragraph should be created if the selection is at the root regardless of whether the root is empty just so that nothing is unintentionally changed to a list?
I've added a test for the case where the selection is at the root which is not empty, but tbf I don't know how that would necessarily happen in an actual editor.

etrepum (approved):
Nice job!

etrepum (approved):

PR #6997, opened 2024-12-26
[lexical-react]perf: updating `useMemo` dependencies.

sanjaiyan-dev:
## Description

This PR refines `useMemo` dependency arrays to improve very minor performance.

Extremely sorry if I made any mistakes.

### DIFF
--- a/packages/lexical-react/src/shared/useYjsCollaboration.tsx
+++ b/packages/lexical-react/src/shared/useYjsCollaboration.tsx
@@ -201,7 +201,7 @@ export function useYjsCollaboration(
       <div ref={ref} />,
       (cursorsContainerRef && cursorsContainerRef.current) || document.body,
     );
-  }, [binding, cursorsContainerRef]);
+  }, [binding.cursorsContainer, cursorsContainerRef]);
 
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/shared/useYjsCollaboration.tsx
```diff
@@ -201,7 +201,7 @@ export function useYjsCollaboration(
       <div ref={ref} />,
       (cursorsContainerRef && cursorsContainerRef.current) || document.body,
     );
-  }, [binding, cursorsContainerRef]);
+  }, [binding.cursorsContainer, cursorsContainerRef]);
```
potatowagon, line 204: lets stick to binding as this will not pass the ci check

etrepum: Not only will this change not pass CI, the logic is also incorrect with this change. This is not a performance improvement, it's just wrong.

etrepum:
Closing as this is a one line change that is incorrect, and there is nothing that needs to be fixed in this location.

PR #7004, opened 2024-12-30
[lexical-markdown] Bug Fix: support link and inline code text formats

AlessioGr:
Fixes https://github.com/facebook/lexical/issues/5148. Additionally, it fixes an issue where formatted code blocks (e.g. 
```md
**`code`**
```
are not exported to markdown correctly after they have been imported.

This PR refactors the logic for applying text match and text format transformers to enable support for nested text formats within text match transformers, such as link nodes.

Unlike the previous attempt to fix it, this change does not include any node-specific logic. It fixes the root cause of the issue by ensuring that nested combinations of textmatch and textformat transformers are applied in optimal order.

**Before**

https://github.com/user-attachments/assets/8e2fef7f-7702-4b5e-89ef-48ac84fe24e3

**After**

https://github.com/user-attachments/assets/87af69a9-c8bd-40ac-9ca7-a0cb0e32d452

## The Problem

Previously, the import process for text transformers roughly followed this sequence:

ElementTransformers ($importBlocks) `=>` TextFormatTransformers `=if not found>` TextMatchTransformers

For link nodes containing formatted text, the process failed as follows:

1. Executed $importBlocks
2. Found code markdown within link => Run code text format transformer => create code textnode
3. The input text is split into 3 nodes: normal text, code text, and normal text. However, this fragmented structure prevents the link text match transformer from recognizing and creating a link node.

Initially, I attempted to solve this issue by adjusting the sequence to prioritize text match transformers:

ElementTransformers ($importBlocks) `=>` TextMatchTransformers `=>` TextFormatTransformers

While this resolved the issue with nested formats, it introduced a new problem in scenarios where links were wrapped by text format markdown, like this:

```md
Text **boldstart [text](https://lexical.dev) boldend** text
```

Now, the link is created first and we get `normal text, link, normal text`. However, the bold text transformer could no longer identify and apply formatting to the entire outer bold range.

## The Solution

Text format transformers already include logic to identify the outermost match, allowing them to handle scenarios like:

```md
One **two __three__ four**
```

In this case, the bold transformer runs first, followed by the italic transformer, ensuring proper formatting.

However, text match transformers currently lack similar logic. Consider this example:

```md
Text **boldstart [`text`](https://lexical.dev) boldend** text
```

The existing sequence processes it as:

1. Bold text
2. Code text
3. Link

However, to achieve correct results, the sequence should be:

1. Bold text
2. Link
3. Code text

To address this, the PR introduces logic for identifying the outermost match across **both** text match and text format transformers.

With this change, text match and text format transformers are treated as equals in priority, allowing their results to be compared directly. This ensures that the outermost match—whether from a text match or a text format transformer—is correctly identified and then applied, enabling seamless handling of nested text transformations.

### Implementation details

1. Find (not apply) outermost text match
2. Find (not apply) outermost text format
3. Determine if found text match or text format is the outermost match
4. Apply said text match **or** text format
5. Repeat this for the matched node, and the node before / afterwards if text was split, until no more text matches **and** text formats are found


AlessioGr:
Just squeezed in another fix for formatted inline code markdown, e.g.

```md
**`code`**
```

Didn't want to open a separate PR, as this one is a relatively large refactor and I wanted to make sure this fix is compatible

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -38,9 +38,20 @@ export function createMarkdownExport(
...
+    // <strong><code>code</code></strong> will be exported as `**Bold Code**`, as the code format will be applied first, and the bold format
+    // will be applied second and thus skipped entirely, as the code format will prevent any further formatting.
+    .sort((a, b) => {
+      if (a.format.includes('code') && !b.format.includes('code')) {
+        return 1;
+      } else if (!a.format.includes('code') && b.format.includes('code')) {
+        return -1;
+      } else {
+        return 0;
+      }
```
etrepum, line 53: fwiw you can also write this as `+a.format.includes('code') - +b.format.includes('code')` or `Number(a.format.includes('code')) - Number(b.format.includes('code'))` to be a little less JS quirk dependent but still only evaluate each condition once without having to use any extra variables

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -542,17 +544,14 @@ export const LINK: TextMatchTransformer = {
...
-      : `[${node.getTextContent()}](${node.getURL()})`;
-    const firstChild = node.getFirstChild();
-    // Add text styles only if link has single text node inside. If it's more
-    // then one we ignore it as markdown does not support nested styles for links
-    if (node.getChildrenSize() === 1 && $isTextNode(firstChild)) {
-      return exportFormat(firstChild, linkContent);
-    } else {
-      return linkContent;
-    }
+      ? `[${textContent}](${node.getURL()} "${title}")`
```
etrepum, line 551: Presumably this title should be escaped because there could be embedded `"` and/or `)`?

packages/lexical-markdown/src/importTextTransformers.ts
```diff
@@ -0,0 +1,150 @@
...
+      foundTextFormat.startIndex,
+      foundTextFormat.endIndex,
+      foundTextFormat.transformer,
+      foundTextFormat.match,
+    );
+
+    if (
+      result.nodeAfter &&
+      $isTextNode(result.nodeAfter) &&
+      !result.nodeAfter.hasFormat('code')
```
etrepum, line 69: With how often this expression is repeated it's probably worth making a function to cover `$isTextNode(node) && !node.hasFormat('code')`. Checking for non-null/undefined is redundant because `$isTextNode` already does that

etrepum (approved):
I didn't look too carefully at the logic in the while loop other than to confirm that it should still terminate, I think the unit test coverage is probably sufficient to show that it's not more wrong than the status quo

packages/lexical-markdown/src/MarkdownExport.ts
AlessioGr: good idea, I like that!

packages/lexical-markdown/src/importTextTransformers.ts
AlessioGr: done! I went for `canContainTransformableMarkdown` as that conveys the intent and makes the code where it's used more readable

AlessioGr: While doing this, found another area that I was able to optimize:

CleanShot 2025-01-20 at 23 06 22@2x

I think we can move the remaining logic up to the `$transform` call out of the `editor.update()` call as well - what do you think?

packages/lexical-markdown/src/MarkdownTransformers.ts
AlessioGr: Done - I escaped this with some simple regex.

Should we consider a library like `dompurify` to protect against XSS attacks here? While it's just markdown, XSS _could_ still be an issue if (and depending on _how_) this is rendered to HTML.

AlessioGr: I have reverted the escaping change. This broke a unit test and did not work when the markdown was re-imported to lexical.

The latter is a [separate issue](https://github.com/payloadcms/payload/issues/10289) I have experienced, where escaped markdown is not un-escaped properly when imported. E.g. `\*text\*` is imported as` \*text\*` instead of `*text*`

I think fixing this properly may not be trivial and is out of scope for this PR anyways, since this PR did not introduce this unescaped title being outputted

packages/lexical-markdown/src/importTextTransformers.ts
AlessioGr: Nevermind - this broke some tests, since `.hasFormat()` has to be called within the `editor.update()` callback

etrepum (approved):
I didn't do a very careful read of all the logic but this seems like a reasonable approach, has new tests, and passes existing tests. @potatowagon what do you think?

potatowagon:
> I didn't do a very careful read of all the logic but this seems like a reasonable approach, has new tests, and passes existing tests. @potatowagon what do you think?

lgtm! a very useful feature, thankyou

PR #7011, opened 2025-01-01
[lexical-playground] Fix table hover actions button position

ivailop7:
After introducing the getDOMSlot usage, the render coordinates for the table hover actions had to be adjusted.

Before:

https://github.com/user-attachments/assets/2621126e-1e7f-4950-b595-1a5f444b019e

After:

https://github.com/user-attachments/assets/ce0a85a8-9499-4368-9172-fdf688e835d2

### DIFF
--- a/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
@@ -115,6 +115,19 @@ function TableHoverActionsContainer({
           height: tableElemHeight,
         } = (tableDOMElement as HTMLTableElement).getBoundingClientRect();
 
+        // Adjust for using the scrollable table container
+        const parentElement = (tableDOMElement as HTMLTableElement)
+          .parentElement;
+        let tableHasScroll = false;
+        if (
+          parentElement &&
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
```diff
@@ -115,6 +115,19 @@ function TableHoverActionsContainer({
           height: tableElemHeight,
         } = (tableDOMElement as HTMLTableElement).getBoundingClientRect();
 
+        // Adjust for using the scrollable table container
+        const parentElement = (tableDOMElement as HTMLTableElement)
```
etrepum, line 119: A more robust check here would be to compare `tableDOMElement` to `parentElement = editor.getElementByKey(table.getKey())` - if they are not identical, then the latter is the outermost wrapping node. This is already computed above just not saved to its own variable.

etrepum (approved):
Not a complete fix because there's no scroll event handling (looking forward to #6923 so all of these things can be fixed). A visible button will remain in a fixed position while the table scrolls behind it. Also a bit of a hack because it doesn't use the theme, but it is playground-only code so that's a bit par for the course.

PR #7013, opened 2025-01-02
Merge TabIndentionPlugin and ListMaxIndentLevelPlugin plugins

fantactuka:
- Adds `INDENT_CONTENT_COMMAND` optional `maxIndent` param that handles both list and block indent
- Adds `TabIndentionPlugin` support for `maxIndent` 
- Removes `ListMaxIndentLevelPlugin` as redundant


ivailop7 (dismissed):

etrepum (commented):
Seems that this has at least one consistently failing test

```
  1) [chromium] › packages/lexical-playground/__tests__/e2e/ListMaxIndentLevel.spec.mjs:34:3 › Nested List › Can only indent until the max depth when list is empty 
```

fantactuka:
Moved max indent logic into intent plugin, should be easier to use vs explicitly passing max indent with every INDENT_CONTENT command

PR #7018, opened 2025-01-06
[lexical-react] Feature: Merge TabIndentionPlugin and ListMaxIndentLevelPlugin plugins

fantactuka:
Moved max indent logic to `TabIndentionPlugin` and updated it to work on non-list elements too


fantactuka:
```
Tables.spec.mjs:4078:3 › Tables › Paste and insert new lines after unmerging cells 
```

this one might be flaky, will re-run failures

fantactuka:
@etrepum we didn't expose `ListMaxIndentLevelPlugin` from playground and new `maxIndent` prop is optional, shouldn't be a breaking change or am I missing something

acywatson:
One possible issue - it appears we're now coupling these two functionalities. So you can't set a max indent without also enabled tab indents. Usually this probably won't matter, but I did want to point out the change.

etrepum:
I think there might be some subtle issue with collab here? I haven't been able to get a fully clean test run yet, the last two had this failure in common:

```
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:1092:3 › Selection › shift+arrowdown into a table, when the table is the only node, selects the whole table
```

Currently it's on attempt 13 https://github.com/facebook/lexical/actions/runs/12636094981/job/35283390649?pr=7018

etrepum:
I guess it's just unlucky, this time it had a different failure

```
[1]   1 failed
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Links.spec.mjs:1969:3 › Links › Can handle pressing Enter inside a Link containing multiple TextNodes 
[1]   11 flaky
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/ListsCopyAndPaste.spec.mjs:115:3 › Lists CopyAndPaste › Copy and paste of partial list items into the list 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/HorizontalRule.spec.mjs:31:3 › HorizontalRule › Can create a horizontal rule and move selection around it 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Links.spec.mjs:431:3 › Links › Can create a link with some text after, insert paragraph, then backspace, it should merge correctly 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Links.spec.mjs:2012:3 › Links › Can handle pressing Enter at the beginning of a Link 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/List.spec.mjs:77:3 › Nested List › Can create a list and partially copy some content out of it 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:520:3 › Selection › Can delete sibling elements forward 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:1046:3 › Selection › shift+arrowdown into a table, when the table is the last node, selects the whole table 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:1069:3 › Selection › shift+arrowup into a table, when the table is the first node, selects the whole table 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:1092:3 › Selection › shift+arrowdown into a table, when the table is the only node, selects the whole table 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Tables.spec.mjs:1662:3 › Tables › Table selection: can select multiple cells and insert an image 
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Toolbar.spec.mjs:42:3 › Toolbar › Insert image caption + table 
```

etrepum (approved):
The logic looks fine, per @acywatson's point I think if someone really wanted to do max indent without tab they could just resurrect the old plug-in or intercept KEY_TAB_COMMAND

fantactuka:
I think this could've been a separate plugin, but then I revisited popular editors and didn't see any of it using indent support without tab key (indent via UI controls only). I'll go ahead with merged version, and let's see if there's any feedback to have it as a standalone one

PR #7019, opened 2025-01-07
Fix: Empty images around equations render as question marks on Safari…

### LINKED ISSUE #6824: Bug: empty images around equations render as question marks on Safari

Codebreaker42:
[lexical-playground] Bug Fix: Remove empty images around equations causing question marks in Safari (#6824)

### Current Behavior
Equations rendered using the `KatexRenderer` component include empty `` tags as placeholders, which appear as question marks in Safari due to Safari’s handling of empty images.

### Changes Made
- Removed empty `` tags around the equation rendering in `KatexRenderer` component.
- Modified the comment to explain the new workaround.
- Verified that Android-specific composition issues with equations remain unaffected.

Closes #6824
Closes #6836

### Before
In Safari:
- Equations show up with question marks around them.
- Example: a+b=c

Rendered incorrectly due to `` placeholders.

*Screenshots/recordings showing the issue:*

!Before Screenshot {8D587512-54D7-41BB-838E-C6C9AFB9AF53}

### After
In Safari:
- Equations are rendered correctly without question marks.
- The issue with Android composition remains resolved even after removing the empty `` tags.

*Screenshots/recordings showing the fix:*

!After Screenshot   {AE51DBD3-074B-405F-8883-DB51D1F7E025}


packages/lexical-playground/src/ui/KatexRenderer.tsx
```diff
@@ -21,12 +22,15 @@ export default function KatexRenderer({
 }>): JSX.Element {
   const katexElementRef = useRef(null);
 
+  // Detect if the browser is Safari
+  const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
```
etrepum, line 26: This may cause some issues with SSR, so it might be better to do this with state and a layout effect or something like that, although I'm not sure if this component is compatible with SSR in the first place. Risk is fairly low since this is not a published component.

etrepum:
It looks like integrity checks are failing, you can run these locally with `npm run ci-check`. In this case I think it's a formatting issue that can be fixed with `npm run prettier:fix`

packages/lexical-playground/src/ui/KatexRenderer.tsx
Codebreaker42: Thank you for the feedback! You're absolutely right that directly using navigator.userAgent in the component could cause issues with SSR. I've updated the implementation to detect if the browser is Safari using a useEffect hook and state. This ensures the detection logic runs only on the client side, mitigating potential SSR issues.

Codebreaker42:
> It looks like integrity checks are failing, you can run these locally with `npm run ci-check`. In this case I think it's a formatting issue that can be fixed with `npm run prettier:fix`

Thank you for pointing that out! I'll run the integrity checks locally using npm run ci-check to identify the issue. If it's a formatting issue, I'll fix it using npm run prettier:fix as suggested.

I'll update the pull request once everything passes. Thanks again for the guidance!

etrepum (commented):
This version doesn't work for me with Safari 18.2 (`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.2 Safari/605.1.15`)
Screenshot 2025-01-06 at 22 23 09

etrepum:
This was tested in the preview app from the latest commit https://lexical-playground-9mxzhz6fh-fbopensource.vercel.app

PR #7022, opened 2025-01-07
[lexical] temporarily set editor to not writable when calling getCachedTypeToNodeMap

potatowagon:
## Description

Closes #

### Before

### After

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -909,9 +909,15 @@ export class LexicalEditor {
     klass: Klass<LexicalNode>,
   ): void {
     const prevEditorState = this._editorState;
+    const originalIsEditable = this.isEditable();
+    // Temporarily freeze the editor before calling getCachedTypeToNodeMap
+    if (originalIsEditable) {
+      this.setEditable(false);
+    }
     const nodeMap = getCachedTypeToNodeMap(prevEditorState).get(
       klass.getType(),
     );
+    this.setEditable(originalIsEditable);
     if (!nodeMap) {
       return;
     }
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -534,6 +534,7 @@ export function markNodesWithTypesAsDirty(
   // We only need to mark nodes dirty if they were in the previous state.
   // If they aren't, then they are by definition dirty already.
   const cachedMap = getCachedTypeToNodeMap(editor.getEditorState());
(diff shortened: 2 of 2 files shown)


etrepum:
This doesn’t seem like the right thing to do, and would cause extra events to fire, what are the preconditions for this problem?

potatowagon:
> This doesn’t seem like the right thing to do, and would cause extra events to fire, what are the preconditions for this problem?

im not sure what other side effects this would introduce, trying it out in a draft PR to experiment, so happy to close if this PR will cause nasty side effects. 

this is a followup to the breaking change in https://github.com/facebook/lexical/pull/6894. one trigger is a case of `editor.registerNodeTransform` being called from a useEffect in workplace. that cause workplace post composer to break. i fixed it on the intergration layer, but not sure where else could break as a result of getCachedTypeToNodeMap being called with a writable editor state

potatowagon:
>and would cause extra events to fire

this looks correct based on the unit tests failures

etrepum:
This should only be a problem if getEditorState returns a writable state, it’s never supposed to, because that is the state that was most recently reconciled. There is a workaround in there to handle an initial empty EditorState which is not frozen, but maybe there is also a case where your team is creating a non-empty EditorState that is not frozen but still returned from getEditorState somehow? I’m not currently at a computer to look into this but if you can come up with a repro I’m sure we can fix this cleanly

etrepum:
In this scenario what I would do, if I couldn't diagnose and fix the root cause of editor.getEditorState() returning a non-empty writable EditorState under some condition, is expose computeTypeToNodeMap and call that instead when this issue would trigger. You really don't want to populate this cache with a writable EditorState because there's no way to invalidate it if the EditorState changes (because it's not supposed to).

etrepum:
I think the root cause is probably `setRootElement(null)` which probably just shouldn't be doing this

```ts
        // If content editable is unmounted we'll reset editor state back to original
        // (or pending) editor state since there will be no reconciliation
        this._editorState = pendingEditorState;
        this._pendingEditorState = null;
        this._window = null;
```

ivailop7:
With Bob on this one, this doesn't seem like the right approach to solve this problem.

etrepum:
#7023 is a cleaner approach, but I can't be entirely sure this is exactly how they are getting the editor to have a writable `_editorState`

potatowagon:
> #7023 is a cleaner approach, but I can't be entirely sure this is exactly how they are getting the editor to have a writable `_editorState`

replied on discord to answer the how, closing this PR since its not the right approach

PR #7037, opened 2025-01-10
[Breaking Change][lexical][lexical-selection][lexical-list] Bug Fix: Fix infinite loop when splitting invalid ListItemNode

### LINKED ISSUE #7036: Bug: Weird list item state causes infinite loop on insertParagraph
Lexical version: 0.18.0 and above

## Steps To Reproduce

1. Create any list with an initial non-empty list item
2. Press Enter and then Tab to create a nested list item, add text to it
3. Press Enter again and in the empty nested list item, toggle off the current list (i.e. set block type to paragraph)
4. Start typing, which will cause the text to be inserted in the last non-nested list item next to the nested list
5. Press Enter again (basically trigger `selection.insertParagraph()` from that text node)

## The current behavior

Pressing Enter in that text node causes an infinite loop.

https://github.com/user-attachments/assets/a4e9b1ca-3b85-47f2-b028-1fa444c95c8c

On Firefox the page freezes indefinitely. On Chrome, it throws this error after a while of freezing:

image

## The expected behavior

It should correctly split the item (or create a new item if pressed Enter at the end)

## Impact of fix

This makes it effectively impossible to add something new to the document if the list item is at the end of the document. Even if it is not at the end of the document, this makes the document uneditable until you refresh the page.

etrepum:
## Breaking Changes

* `insertList` and `removeList` are now deprecated, use `$insertList` and `$removeList` instead.
* `INSERT_CHECK_LIST_COMMAND`, `INSERT_ORDERED_LIST_COMMAND`, `INSERT_UNORDERED_LIST_COMMAND` and `REMOVE_LIST_COMMAND` now update synchronously when dispatched from inside an update, rather than deferring a nested update (this is the expected behavior for commands)
* `editor.focus()` now happens synchronously when called from inside of an update, as if it was implemented with a command dispatch. Previously it would defer which is confusing behavior because the expected side-effect is to change the selection which you really do want to happen synchronously.

The breaking changes are the result of trying to unit test this functionality and finding out that editor.focus and list commands are asynchronous soup for no reason. These breaking changes are unlikely to affect any correctly written code, unless it is going very far out of the way to test or otherwise compensate for the previously maddening behavior.

## Description

There's a while loop in `$removeTextAndSplitBlock` that can reach a fixed point and never terminate if it reaches the root (which is not a block). This ensures that if it gets to a fixed point, the loop terminates.

Additionally this fixes some incorrect code in `$setBlocksType` which mutated a potentially cached `selection.getNodes()` and did not always update the selection when it should (this was part of the preconditions for #7036) so you'd end up with a selection that is not normalized correctly that allowed you to insert TextNode adjacent to ListNode where it shouldn't be.

As a small related cleanup, the `INTERNAL_$isBlock` function is now exported from the `lexical` package so that it can be used in the implementation of `@lexical/selection`, previously this function was copied which added unnecessary bundle size and maintenance burden to maintain two copies of this function. It's marked `@internal` so the doc tools shouldn't pick it up.

Closes #7036

### Before

Creating a ListItemNode that has a ListItem first child and a TextNode second child causes an infinite loop when the return key is pressed, see #7036

### After

A ParagraphNode is created and it is selected. A unit test covers this addition.

### DIFF
--- a/packages/lexical-list/README.md
+++ b/packages/lexical-list/README.md
@@ -9,12 +9,12 @@ The API of @lexical/list primarily consists of Lexical Nodes that encapsulate li
 
 ## Functions
 
-### insertList
+### $insertList
 
 As the name suggests, this inserts a list of the provided type according to an algorithm that tries to determine the best way to do that based on
-the current Selection. For instance, if some text is selected, insertList may try to move it into the first item in the list. See the API documentation for more detail.
+the current Selection. For instance, if some text is selected, $insertList may try to move it into the first item in the list. See the API documentation for more detail.
 
-### removeList
+### $removeList
 
 Attempts to remove lists inside the current selection based on a set of opinionated heuristics that implement conventional editor behaviors. For instance, it converts empty ListItemNodes into empty ParagraphNodes.
 
@@ -43,7 +43,7 @@ It's important to note that these commands don't have any functionality on their
 // MyListPlugin.ts
 
 editor.registerCommand(INSERT_UNORDERED_LIST_COMMAND, () => {
-    insertList(editor, 'bullet');
+    $insertList(editor, 'bullet');
     return true;
 }, COMMAND_PRIORITY_LOW);
 
--- a/packages/lexical-list/flow/LexicalList.js.flow
+++ b/packages/lexical-list/flow/LexicalList.js.flow
@@ -35,6 +35,10 @@ declare export function $isListNode(
   node: ?LexicalNode,
 ): node is ListNode;
 declare export function indentList(): void;
+declare export function $insertList(
+  listType: ListType,
+): void;
+/** @deprecated use {@link $insertList} from an update or command listener */
 declare export function insertList(
   editor: LexicalEditor,
   listType: ListType,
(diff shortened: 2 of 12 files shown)


ivailop7 (approved):

etrepum:
The test failure is a spurious flaky test failure, not worth running it any more times to try and get a clean run

CI
```
[1]   1 failed
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/ListsCopyAndPaste.spec.mjs:115:3 › Lists CopyAndPaste › Copy and paste of partial list items into the list 
```

Local
```
❯ npm run test-e2e-collab-chromium packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/ListsCopyAndPaste.spec.mjs:115:3

> @lexical/monorepo@0.23.1 test-e2e-collab-chromium
> cross-env E2E_BROWSER=chromium E2E_EDITOR_MODE=rich-text-with-collab playwright test --project="chromium" packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/ListsCopyAndPaste.spec.mjs:115:3


Running 1 test using 1 worker

  ✓  1 …dPaste.spec.mjs:115:3 › Lists CopyAndPaste › Copy and paste of partial list items into the list (2.5s)

  1 passed (3.5s)
```

PR #7038, opened 2025-01-10
[lexical] Fix whitespace stripping issue in textDOMNode conversion

Tranquiliz00:
## Description

The existing logic:
```textContent = textContent.replace(/\r/g, '').replace(/[ \t\n]+/g, ' ');```
will replace multiple whitespaces with a single whitespace, which will mess up the format of texts.

We have seen issues when people convert html to text node using lexical-html, especially for codes.

### Before

https://github.com/user-attachments/assets/88d71b94-9ca9-46c1-8340-0180c6948db7

### After

https://github.com/user-attachments/assets/defdd0c4-9912-4aeb-b323-2d32cccc1958

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -1197,7 +1197,10 @@ function $convertTextDOMNode(domNode: Node): DOMConversionOutput {
     }
     return {node: nodes};
   }
-  textContent = textContent.replace(/\r/g, '').replace(/[ \t\n]+/g, ' ');
+  textContent = textContent
+    .replace(/\r/g, '')
+    .replace(/[\t\n]+/g, ' ')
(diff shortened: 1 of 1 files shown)


Tranquiliz00:
It looks like stripping the whitespaces is the intended behavior based on existing tests?

etrepum:
Generally speaking, line breaks and tabs should be encoded as LineBreakNode and TabNode, not embedded in other TextNode. If you have some examples of this happening using existing functionality within the documented constraints then an issue with some examples would be a good place to start.

etrepum:
For example, see [insertRawText](https://lexical.dev/docs/api/classes/lexical.RangeSelection#insertrawtext) in https://github.com/facebook/lexical/blob/bd874a3edae21e4b89ba54be8c23c00ef83feb17/packages/lexical/src/LexicalSelection.ts#L708-L729

PR #7039, opened 2025-01-10
Clean up nested editor update

ivailop7:
Stumbed upon nested updates in the $onDelete methods while troubleshooting some selection issues. Not 100% sure if this is correct.


etrepum (approved):
This looks right to me, seems like a good idea to add a lint to discourage editor.update from inside a $function

PR #7041, opened 2025-01-11
[playground] Bug fix: prevent growing whitespaces in markdown table toggle

potatowagon:
### Before

https://github.com/user-attachments/assets/91e25c09-9c04-4cbf-8b17-652ce026c136

^ white spaces in table cell growing

### After

https://github.com/user-attachments/assets/841d33d2-ac91-4da6-b4e3-56fa781f7a80

^ num of white spaces remain constant


etrepum (approved):
I can see why this works but I think the actual bug is that on import `mapToTableCells` doesn't discard the required space before and after the `|` delimiter

Issue #6909, opened 2024-12-04
Bug: Cannot align tables


ivailop7:
Lexical 0.21

Applying alignment to a table doesn't take effect. Currently one cannot centre a table. I guess we are missing formatting properties on TableNode-level

etrepum:
In Firefox the CSS is such that the table is always full width, might make sense to tackle both at the same time if we are to support horizontal alignment for tables.

ivailop7:
Oh wow, I didn't even realize we even have different CSS for Firefox...

etrepum:
Well, the CSS is the same but Firefox does not render it the same way the other engines do. I think it regressed when the column width stuff was merged #6625 due to the `width: fit-content` instead of `width: max-content`.

Ulop:
Since there is no support at the library level, I solved the problem in my project with an additional `TableWrapper` node.

It wraps the table node in an additional `div`, with `display: flex; flex-direction: column`, and `align-items` according to `ElementNode::getFormatType`.

https://github.com/user-attachments/assets/4ac12563-5200-4aee-b918-759981ac280f



Works fine overall, but it would be nice to have an out-of-the-box solution.

PR #7044, opened 2025-01-12
[lexical-table] Support table alignment

ivailop7:
Fixes https://github.com/facebook/lexical/issues/6909

To differentiate between table alignment and cell text alignment, the table-level alignment logic only kicks in when the selection is the whole exact table. This does introduce the extra step if one wants to align all cell texts to do it in 2 steps, all columns except one and then the last column.

Before:

https://github.com/user-attachments/assets/d1851ec3-6844-4bb7-89f9-02bdf1000abf

After:

https://github.com/user-attachments/assets/30265b64-66c0-4b70-85d2-988f8eb7cb83

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -196,6 +196,13 @@
   /* Remove the table's margin and put it on the wrapper */
   margin: 0;
 }
+.PlaygroundEditorTheme__tableAlignmentCenter {
+  margin-left: auto !important;
+  margin-right: auto !important;
+}
+.PlaygroundEditorTheme__tableAlignmentRight {
+  margin-left: auto !important;
+}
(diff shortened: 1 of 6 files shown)


ivailop7:
Rewrote this with classNames vs using 'style', to leave the option for overridability to users, but unsure if ended up any better because I couldn't get the class waterfall to work nicely without !imporant to override the margins. In both cases (style vs classNames) it works correctly with both with and without tableWrapper

ivailop7:
Thanks @etrepum, implemented the PR comments

etrepum (commented):
Haven't looked closely at the changes but there are test failures now

ivailop7:
> Haven't looked closely at the changes but there are test failures now

Addressed

ivailop7:
@etrepum could we wrap this one up please.

etrepum (approved):
Looks good, and confirmed that it works as expected in the playground

PR #7045, opened 2025-01-12
[lexical] Bug Fix: Handle MutationObserver/input event re-ordering when using contentEditable inside of an iframe

### LINKED ISSUE #7028: Bug: Error: $validatePoint: anchor point.offset > node.getTextContentSize() in StencilJS
Lexical version:
"lexical": "0.23.0"
"@lexical/history": "0.23.0",
 "@lexical/html": "0.23.0",
 "@lexical/rich-text": "0.23.0",
 "@lexical/utils": "0.23.0",

Framework
Stencil "^4.23.0"

## Steps To Reproduce

1. Load Editor in an Iframe
2. Start typing more than one alphabet.
3. Error displayed in the browser console.

## Editor SetUp Class
`
import {
  createEditor,
} from 'lexical';

 initialConfig = {
  namespace: 'Lexical',
  // Register nodes specific for @lexical/rich-text
  nodes: [],
  onError: (error: Error) => {
    throw error;
  },
  // theme: {
  //   // Adding styling to Quote node, see lexical-styles.css
  //   // quote: 'lexical__quote ',  this is an example
  // },
};

private createEditor(editableContainer: HTMLElement): void {
    this.editor = createEditor(initialConfig);
    this.editor.setRootElement(editableContainer);

    mergeRegister(
      registerRichText(this.editor),
      registerHistory(this.editor, createEmptyHistoryState(), 300),
    );
  }

public async initialize(): Promise<void> {
    // Creates iframe element with srcDoc
    const iframeDocument = this.renderIframe();

    iframeDocument.onload = () => {
      const editableContainer = frameDocument.contentDocument?.querySelector('#lexical-editor') as HTMLElement;

      if (editableContainer) {
        this.createEditor(editableContainer);
      } else {
        // eslint-disable-next-line
        console.error('Editable container not found in iframe');
      }
    };
  }
`

etrepum:
## Description

Sometimes when an iframe is used as the contentEditable then the MutationObserver can fire after the input event, even if the mutation actually did happen. This causes problems that were undiagnosed prior to the addition of `$validatePoint` in `__DEV__` (part of #6759) because the Lexical selection would refer to DOM coordinates that were not yet valid in the Lexical document, because the DOM was mutated before getting the selection. When the mutation observer happens first this doesn't occur because either the mutation is applied and/or the selection code is skipped due to `getIsProcessingMutations()` (see `$internalCreateRangeSelection`). This is related to the optimization(?) in #794 that avoids explicitly flushing root mutations on input.

This PR:

- Adds a vanilla-js-iframe example as a testbed for checking that iframe events are handled correctly, with a `dev:monorepo` script so you can run it with the monorepo like the playground rather than with the npm installed lexical for easier development/debugging (maybe we add this to other examples?)
- Fixes up some incorrect $function naming
- Uses `updateEditorSync` instead of `updateEditor` in the places that I visited along the way, for in-order execution
- Adds a currently internal `event` option to `UpdateOptions` so that the reason for the update can be captured in a way that's accessible by the code in `$internalCreateRangeSelection` that checks to see if it's from a mutation or from some specific set of events before deciding whether to use the DOM selection or not. We don't want to use the DOM selection before input because it might refer to DOM coordinates that do not exist in the lexical document yet.
- Consistently use `getWindow` for `editor._window` in the core package (just a general clean-up while investigating)

Closes #7028

## Test plan

I'm not exactly sure how to reasonably add tests for this

### Before

Error when typing into the end of a TextNode in an iframe on Chrome or Firefox

### After

No more errors

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -75,7 +75,7 @@ import {
   $internalCreateRangeSelection,
   RangeSelection,
 } from './LexicalSelection';
-import {getActiveEditor, updateEditor} from './LexicalUpdates';
+import {getActiveEditor, updateEditorSync} from './LexicalUpdates';
 import {
   $flushMutations,
   $getNodeByKey,
@@ -201,7 +201,7 @@ function $shouldPreventDefaultAndInsertText(
   const focus = selection.focus;
   const anchorNode = anchor.getNode();
   const editor = getActiveEditor();
-  const domSelection = getDOMSelection(editor._window);
+  const domSelection = getDOMSelection(getWindow(editor));
   const domAnchorNode = domSelection !== null ? domSelection.anchorNode : null;
   const anchorKey = anchor.key;
   const backingAnchorElement = editor.getElementByKey(anchorKey);
@@ -290,7 +290,7 @@ function onSelectionChange(
       return;
     }
   }
-  updateEditor(editor, () => {
+  updateEditorSync(editor, () => {
     // Non-active editor don't need any extra logic for selection, it only needs update
     // to reconcile selection (set it to null) to ensure that only one editor has non-null selection.
     if (!isActive) {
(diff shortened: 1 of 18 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -290,7 +290,7 @@ function onSelectionChange(
       return;
     }
   }
-  updateEditor(editor, () => {
```
zurfyx, line 293: Do we foresee any performance impact of sync update events? I presume that this makes it possible for multiple reconciliations to happen almost simultaneously

zurfyx (approved):
That's a tricky one, thanks for looking into this! 😮

PR #7046, opened 2025-01-13
[lexical][lexical-utils][lexical-selection][lexical-table] Feature: NodeCaret abstraction for traversals and ranges

### LINKED ISSUE #7076: Regression: Font styles not applying to text in table cells
Lexical: 0.23.1 

After doing a git bisect, the following PR: https://github.com/facebook/lexical/pull/6981 introduced a regression, where upon selecting table cells and changing the font styles: color, background-color, etc are no longer applied.

Bug:

https://github.com/user-attachments/assets/ef737607-a3c2-4dc5-8f9d-04dae0a0da25

### LINKED ISSUE #7081: Bug: selection content not deleted when double click content and paste it in firefox.
Lexical version:
https://playground.lexical.dev
## Steps To Reproduce
Image

### **when copy a text like 123, double click to select content XXX and paste it, then ui show pasted content and selection content like 123XXX**

Image

Link to code example:

## The current behavior
current behavior is rich content show pasted content and selection content
## The expected behavior
expect ui only show pasted content, selection content should be deleted.
## Impact of fix
only occurs in firefox, chrome and edge is normally.

etrepum:
## Description

A fair amount of bugs and complexity in our code results from ambiguity and awkwardness around traversing the node tree, as well as issues around coordinates becoming stale after mutations. This is a proposal for a new set of abstractions that can be used in place of the typical `PointType` and `RangeSelection` to work through these sorts of traversal use cases with hopefully fewer bugs.

There are three major deficiencies that the design of `NodeCaret` hopes to address:

- PointType is not directional, and it's also not possible to infer a direction from a collapsed RangeSelection.
- Dealing with the borders of an ElementNode is full of quirks, especially when empty, frequently causing over-selection or under-selection or just having to frequently branch for empty and non-empty cases.
- Instability of PointType when doing anything "before" the point (e.g. inserting a node before a point "moves" the logical location of the point to the newly inserted node)

Docs:

* [Node Traversals with NodeCaret](https://lexical-git-fork-etrepum-lexical-caret-fbopensource.vercel.app/docs/concepts/traversals)

I can write up some flow types when/if we move forward with the proposal to reduce the maintenance burden if anything changes before acceptance.

## Cost

The current estimate looks like it's currently at `31.55 KB (+8.56% 🔺)` which is +~2.5kB for the production cjs build. There will be some follow-up PRs to reduce this further with additional refactoring to use this API more internally.

## Other changes

I didn't refactor everything with NodeCaret but chose a few places that would be straightforward and/or make it easier to fix a few bugs. More will come later which should reduce code size.

`lexical`
- Refactor `RangeSelection.removeText` with new algorithm based on CaretRange
- Clean up `TextNode.splitText` algorithm
- Fix over-selection in getNodes() (refactor based on NodeCaret coming later)

`@lexical/selection`
- Fixed `$forEachSelectedTextNode` to work with types other than RangeSelection
- Export `$patchStyle` for updating the style of an individual TextNode or collapsed RangeSelection
- Export new `$ensureForwardRangeSelection` function to flip anchor/focus when backwards
- `$setBlocksType`: Add new `$afterCreateElement` argument to allow user to specify how properties propagate to the new block (e.g. format and indent by default - fixes an indent related bug)
- Export new `$copyBlockFormatIndent` used as new `$setBlocksType` optional argument default

`@lexical/table`
- Export existing `$createTableSelectionFrom`
- Add more unit tests for TableSelection

`@lexical/utils`
- Refactor `$dfs` and related functions with NodeCaret implementations

Closes #7076
Closes #7081


zurfyx (commented):
Overall, I like the concept and the direction, thank you for putting this masterpiece together with such detailed documentation and test plan!

A couple high-level thoughts:

While most of the LexicalCaret methods are simple, this adds a major of piece of opinionated API that (will) touch on most of Lexical core methods, I'm curious how complicated it will for contributors to familiarize with it. Before that, selection and LexicalNode were mostly independent. However, it does make some of the core implementation easier to read when the verbose iteration logic is abstracted away, and your LexicalSelection fixes prove this.

Naming wise, @fantactuka proposes the use of cursor instead of caret. Previously, we introduced block "cursor" selection, I wonder why we decided to name this one caret instead.

I'll have it a second pass this week but I wonder what others think in the meantime.

etrepum:
> Overall, I like the concept and the direction, thank you for putting this masterpiece together with such detailed documentation and test plan!
> 
> A couple high-level thoughts:
> 
> While most of the LexicalCaret methods are simple, this adds a major of piece of opinionated API that (will) touch on most of Lexical core methods, I'm curious how complicated it will for contributors to familiarize with it. Before that, selection and LexicalNode were mostly independent. However, it does make some of the core implementation easier to read when the verbose iteration logic is abstracted away, and your LexicalSelection fixes prove this.

Selection and `LexicalNode` are still very much independent. `NodeCaret` has nothing to do with the selection specifically, it's an abstraction for finding a node adjacent to some origin node in some specific direction. It's just a building block that can be used to come up with things like selections (e.g. using two `PointCaret` you can make a `CaretRange` which is basically a more robust `RangeSelection`).

Even in situations like TableSelection, NodeSelection, etc. you still need to do document tree navigation and in those cases it will be very useful to have these abstractions at hand. In some cases RangeSelection could be used for that but it's awfully hard to work with them without setting the editor's selection which is counter-productive when you are working with another selection type.

> Naming wise, @fantactuka proposes the use of cursor instead of caret. Previously, we introduced block "cursor" selection, I wonder why we decided to name this one caret instead.

The words have very similar meaning, and I think it will be hard to avoid some confusion no matter which name is chosen. I settled on caret specifically because it is used more commonly for text, where cursor is the more general term. Cursor is used in other contexts in lexical already which is why I avoided that (blockCursor, css cursor for the mouse pointer, etc.).  See also:

- [Caret](https://developer.mozilla.org/en-US/docs/Glossary/Caret)
- [Caret navigation](https://en.wikipedia.org/wiki/Caret_navigation)

etrepum:
I think a more accurate way to think of it is that NodeCaret/PointCaret are a way to bring something like PointType (but more fit for purpose) to everywhere in Lexical in a way that *does not* depend on RangeSelection. Right now RangeSelection is *too intertwined* with LexicalNode and document navigation in general. For example there's probably been a hundred times where I want to programmatically create a selection that spans multiple nodes but the APIs to do that are just miserable because calling .select(...) on a node sets the EditorState's selection as a side-effect and always sets both points when you're really just trying to get an anchor or a focus. In order to get an anchor or focus from a node with NodeCaret you just create one with one function call.

zurfyx (approved):

etrepum:
Will merge this after v0.24.0 so there's some nightly releases before it goes live

etrepum:
Latest couple commits:
* Resolve conflict with main
* Added more docs, and copied and pasted the examples from the docs to find a small mistake or two
* Added flow types for the new stuff
* Refactored StepwiseIteratorConfig to use a hasNext guard instead of stop so the types didn't have to lie sometimes (this is one of the things that flow does better than typescript with implies)

Should be good to merge once re-approved

PR #7050, opened 2025-01-14
[lexical] Bug Fix: Normalize selection after applyDOMRange to account for Firefox differences

### LINKED ISSUE #6991: Bug: Selecting text in a link to/from the end of the link and deleting deletes the text that is not selected
Lexical version: 0.21.0

## Steps To Reproduce

1. Have a link node with a single text node, and another text node as the next sibling of the link node.
2. In the link, select text from the middle to the end or vice-versa. Either the anchor or the focus should have type `element`, offset `1` and key of the link node. This can be achieved by starting the selection from the middle then selecting the first character of the next text node and then back to the end of the link, or by starting the selection from the end of the link's text node.
3. Press backspace or delete to delete the characters

It doesn't seem possible to get the selection into this state if the link node has multiple child text nodes or if the link node does not have a text node sibling next to it.

Link to code example: https://playground.lexical.dev (as of 23rd dec)

## The current behavior

Pressing backspace or delete causes the text which is not selected to be deleted and the selection to be collapsed at the start of that text node.

https://github.com/user-attachments/assets/8d797746-8bf8-45a0-ac20-cfa7b8a29316

## The expected behavior

Only the text that was selected should be deleted.

## Impact of fix

This affects any user that tries to delete text at the end of a link and isn't careful about overselecting text initially. If the user starts selecting from the end of the link, then this happens even without overselecting from the sibling text node.

### LINKED ISSUE #7049: Bug: backspace in Firefox removed the MarkNode
Lexical version: Playground

## Steps To Reproduce

1. Go to playground in **Firefox**
2. Add a comment
3. Move the cursor, from the right, to the end of the node, using either arrows or backspace
4. Press backspace
5. Node is removed
## The current behavior

https://github.com/user-attachments/assets/f55ead76-8133-4b64-81f1-cf78477be806

## The expected behavior

To only remove a single character, like if you backspace from the middle of marknode

## Impact of fix

This seems to be a Firefox (tested on 134.0) issue only. I can't replicate it in Chrome

etrepum:
## Description

When we use native selection methods to move the cursor in Firefox it will sometimes resolve to an element point, rather than a text point, which violates some assumptions baked into removeText and other internals.

This adds a normalization step to RangeSelection.applyDOMRange (on all platforms) that can be manually toggled off if necessary. I don't think this is needed but I also haven't seen a full e2e test run yet.

Closes #6991
Closes #7049

### Before

Backspacing across the border of an ElementNode will cause the majority of that ElementNode to be deleted in Firefox

### After

Firefox behaves as other browsers do, and now we have an e2e test (with LinkNode) to demonstrate that the behavior is fixed

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Links.spec.mjs
@@ -524,6 +524,68 @@ test.describe.parallel('Links', () => {
     },
   );
 
+  test(
+    `Can backspace across a link and it deletes text, not the whole link`,
+    {
+      tag: '@flaky',
+    },
+    async ({page}) => {
+      await focusEditor(page);
+      await page.keyboard.type(' abc def ');
+      await moveLeft(page, 5);
+      await selectCharacters(page, 'left', 3);
+
+      // link
+      await click(page, '.link');
+      await click(page, '.link-confirm');
+
+      await assertHTML(
+        page,
+        html`
+          <p
+            class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+            dir="ltr">
+            <span data-lexical-text="true"></span>
+            <a
+              class="PlaygroundEditorTheme__link PlaygroundEditorTheme__ltr"
+              dir="ltr"
+              href="https://"
+              rel="noreferrer">
+              <span data-lexical-text="true">abc</span>
+            </a>
+            <span data-lexical-text="true">def</span>
+          </p>
+        `,
+      );
+
+      await moveRight(page, 4);
+
+      await page.keyboard.press('Backspace');
+      await page.keyboard.press('Backspace');
+      await page.keyboard.press('Backspace');
+      await page.keyboard.press('Backspace');
+
+      await assertHTML(
+        page,
+        html`
+          <p
+            class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
+            dir="ltr">
+            <span data-lexical-text="true"></span>
+            <a
+              class="PlaygroundEditorTheme__link PlaygroundEditorTheme__ltr"
+              dir="ltr"
+              href="https://"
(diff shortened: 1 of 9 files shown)


zurfyx (commented):
Thanks for fixing this! Do you know if this triggers an extra selection reconciliation? FWIW, this fix seems similar to triple click

etrepum:
> Thanks for fixing this! Do you know if this triggers an extra selection reconciliation? FWIW, this fix seems similar to triple click

I don't think so? The logic in there looks a bit wonky to me though, specifically:

* `RangeSelection.applyDOMSelection` unconditionally mutates the RangeSelection via `$setPointValues` but doesn't mark the selection as dirty (though it does set cachedNodes to null)
* `$normalizeSelection` uses `Point.set` to mutate its _selection and *will* mark it as dirty (as well as clearing the cache)

This means that when normalizing, the selection will always get marked as dirty, which would trigger `updateDOMSelection`, but that is probably already happening in most cases. Maybe it makes sense to add some guards to either Point.set or $normalizeSelection to only mark the selection as dirty if the point actually changed? That said, the contents of a selection can change even if the points don't, so cachedNodes with a writable EditorState still isn't really all that coherent unless it's managed very carefully.

There's also some compositionKey logic in Point.set which might be better placed somewhere else, but I'm not very familiar with how that hack works.

etrepum:
The remaining e2e test failure is flaky

Current run:
```
[1]   1 failed
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:520:3 › Selection › Can delete sibling elements forward 
```

A previous run:
```
   1 failed
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:1092:3 › Selection › shift+arrowdown into a table, when the table is the only node, selects the whole table 
```

zurfyx (approved):

benjick:
Thank you! 🙏

PR #7052, opened 2025-01-15
Revert [Breaking Change][lexical] Bug Fix: Commit updates on editorSetRootElement(null) (#7023)

potatowagon:
S482409: Editors on Meta's www were crashing with error: `Got unexpected null or undefined` on reloading the editor.

the root cause is a nullthrows at 

```
  const parentElementDOM = nullthrows(
    editor.getElementByKey(parentElementNode.getKey()),
  );
```
[MLCFloatingParagraphStylesPlugin.react.js](https://www.internalfb.com/code/www/[32c8b8ca986ab3fd52f78aabcc131eb033653175][blame]/html/js/shared/lexical_composer/plugins/MLCFloatingParagraphStylesPlugin.react.js?lines=61-63) in the www integration layer. More investigation needed to determine whats the best approach to handle null being returned to parentElementDOM.

PR7023 is suspected to cause the change, and carries more risk then expected. Reverting the sync diff with PR7023 mitigates the issue. sorry @etrepum to unblock the syncs, may we revert this PR first?

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -1101,15 +1101,11 @@ export class LexicalEditor {
           }
         }
       } else {
-        // When the content editable is unmounted we will still trigger a
-        // reconciliation so that any pending updates are flushed,
-        // to match the previous state change when
-        // `_editorState = pendingEditorState` was used, but by
-        // using a commit we preserve the readOnly invariant
-        // for editor.getEditorState().
+        // If content editable is unmounted we'll reset editor state back to original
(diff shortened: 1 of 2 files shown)


zurfyx (approved):

etrepum:
I think the real question here is why is an editor setting the root element to null and also expecting the associated DOM to be non-null? I don't have a strong opinion on whether the reconciliation here should be immediate or not but having setRootElement(nonNull) work significantly differently than setRootElement(null) is confusing behavior and should at least be documented and tested as the expected semantics if we are to leave this in place.

PR #7053, opened 2025-01-15
[lexical-yjs] Feature: Allow passing in custom `syncCursorPositions` function to collab hook

amanharwara:
## Description

We created a custom `syncCursorPositions` function in our app to allow for more advanced styling and behavior than what the implementation exported by `@lexical/yjs` allows (which requires some CSS trickery even just to customize basic styling)

However we were noticing that sometimes some of the cursors would use the styles from the Lexical impl and not ours, even though those styles didn't exist in our codebase. It turned out that the `syncYjsChangesToLexical` function calls the lexical implementation of `syncCursorPositions`, so depending on which implementation ran first, some cursors would use incorrect styles. Also, the lexical implementation removes cursors for user states which are not focusing, while ours does not. This would cause flickering of the cursors since the Lexical impl would remove them and then ours would add them back.

Currently we have used yarn patch to remove this call, but I think it makes sense to allow a custom function to be passed so that it becomes possible to have a custom implementation of collab cursors without the existing lexical implementation also running.

### DIFF
--- a/packages/lexical-react/src/shared/useYjsCollaboration.tsx
+++ b/packages/lexical-react/src/shared/useYjsCollaboration.tsx
@@ -6,7 +6,7 @@
  *
  */
 
-import type {Binding, Provider} from '@lexical/yjs';
+import type {Binding, Provider, SyncCursorPositionsFn} from '@lexical/yjs';
 import type {LexicalEditor} from 'lexical';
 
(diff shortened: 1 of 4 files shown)


etrepum (approved):
Seems reasonable and straightforward

PR #7055, opened 2025-01-15
fix: triple click around inline elements (links)

fantactuka:
Triple click on paragraph (or other block element) with link as a first child would select link instead of whole element's content

###  Before  
https://github.com/user-attachments/assets/1d1d03ff-eaf9-4cf0-8f5e-dbafa632ae67

### After
https://github.com/user-attachments/assets/e48c8cf5-9d03-4284-a112-fc03455989ff


etrepum (approved):
LGTM, confirmed the bug in the playground and confirmed the fix in the PR app

PR #7057, opened 2025-01-16
[lexical-table] Bug Fix: Prevent error if pasted table has empty row

amanharwara:
## Description

Currently, if a table was pasted that has an empty row, the `$tableTransform` would throw an error. This is because in `$computeTableMapSkipCellCheck`, an array for a row is only initialized inside the loop that goes through its children. This causes there to be a gap in the gridMap since an array for the empty row never gets initialized, which then causes `gridMap[i]` to be undefined and throw an error.

This PR fixes this by making sure an empty array is initialized at the index of the empty row, which in turn will be filled with the necessary amount of cells by `$tableTransform`

## Test plan

Added e2e test to cover this scenario

### Before

image

### After

image

### DIFF
--- a/packages/lexical-table/src/LexicalTableUtils.ts
+++ b/packages/lexical-table/src/LexicalTableUtils.ts
@@ -856,9 +856,7 @@ export function $computeTableMapSkipCellCheck(
       $isTableRowNode(row),
       'Expected TableNode children to be TableRowNode',
     );
-    if (row.getChildrenSize() === 0) {
-      tableMap[rowIdx] = [];
-    }
+    const startMapRow = getMapRow(rowIdx);
     for (
       let cell = row.getFirstChild(), colIdx = 0;
       cell != null;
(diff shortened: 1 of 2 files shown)


packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -856,6 +856,9 @@ export function $computeTableMapSkipCellCheck(
       $isTableRowNode(row),
       'Expected TableNode children to be TableRowNode',
     );
+    if (row.getChildrenSize() === 0) {
+      tableMap[rowIdx] = [];
+    }
```
etrepum, line 861: I don't think this is correct regarding rowspan, but you can initialize `startMapRow` here which will have the same effect and should be correct (at least with regard to that constraint). You'll have to remove the initialization below but I can't put a suggestion on code you didn't change in the review

```suggestion
    const startMapRow = getMapRow(rowIdx);
```

etrepum (approved):

PR #7058, opened 2025-01-16
docs: Change "here" link to more descriptive text

6TELOIV:
## Description
docs change

Links with text like "here" or "click here" are not as easily interpretable to screen reader users as links which describe the destination in the link text rather than before/after the link.

The existing text was not a failure of [SC 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) but this change should improve its accessibility.

Many other options of phrasing are possible (the beauty of communication and language!) to please modify if you have a phrase that better conveys the original's author intent.

## Test plan
n/a

### Before
You can see how it's used in the playground [here](https://github.com/facebook/lexical/blob/0775ab929e65723433626fa8c25900941e7f232f/packages/lexical-playground/src/Editor.tsx#L137)

### After
You can see how [TwitterPlugin is used in the playground](https://github.com/facebook/lexical/blob/0775ab929e65723433626fa8c25900941e7f232f/packages/lexical-playground/src/Editor.tsx#L137).


etrepum (approved):
Change looks good and it's more clear written this way by any measure. Thank you!

PR #7059, opened 2025-01-17
[lexical-list] Feature: Continuous list numbering

patrick-atticus:
## Description

Currently, each new numbered ListNode starts it's numbering at "1". The user can create a new list with a custom start value by typing the number, followed by a period, followed by space (eg. "12. "). If the user tries to create continuous lists this way, they will not stay in sync when new items are added/deleted from previous lists.

Continuous lists are a feature of word/google docs accessed via the context menu and are important in certain types of documents:
Screenshot 2025-01-14 at 3 22 13 PM

This PR adds this feature via the following changes:

- A new property `__continuePreviousNumbering` to ListNode
- ListNode method to update the numbering
- Listeners in ListPlugin to trigger updates

## Test plan

Included basic unit and playground tests, can add more if this PR is promising.

Demo:

https://github.com/user-attachments/assets/09eac18b-59e2-47e9-9d9e-3da2a77dd351

### DIFF
--- a/packages/lexical-list/src/index.ts
+++ b/packages/lexical-list/src/index.ts
@@ -10,8 +10,9 @@ import type {SerializedListItemNode} from './LexicalListItemNode';
 import type {ListType, SerializedListNode} from './LexicalListNode';
 import type {LexicalCommand, LexicalEditor} from 'lexical';
 
-import {mergeRegister} from '@lexical/utils';
+import {$dfs, mergeRegister} from '@lexical/utils';
 import {
+  $getRoot,
   COMMAND_PRIORITY_LOW,
   createCommand,
   INSERT_PARAGRAPH_COMMAND,
@@ -97,6 +98,18 @@ export function registerList(editor: LexicalEditor): () => void {
       },
       COMMAND_PRIORITY_LOW,
     ),
+    editor.registerNodeTransform(ListNode, () => {
+      let previousList: ListNode | null = null;
+      $dfs($getRoot()).forEach(({node}) => {
+        if (!$isListNode(node) || node.getListType() !== 'number') {
+          return;
+        }
+        if (previousList) {
+          node.updateStartFromPreviousList(previousList);
+        }
+        previousList = node;
+      });
+    }),
   );
   return removeListener;
 }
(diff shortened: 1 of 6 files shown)


packages/lexical-list/src/index.ts
```diff
@@ -97,10 +97,35 @@ export function registerList(editor: LexicalEditor): () => void {
       },
       COMMAND_PRIORITY_LOW,
     ),
+    editor.registerNodeTransform(ListNode, $updateListNumbering),
```
patrick-atticus, line 100: I've tried to use node transforms where possible, but ran in to a bug with this approach (see video). 

I know mutation listeners should not trigger additional updates, but one is needed to handle deleted list events. Would it be preferable to use it for all updates in this case?

https://github.com/user-attachments/assets/f1e42a52-3e32-4d28-84fb-63c9af26135f

etrepum (commented):
Ideally CSS counters could be leveraged to support this feature and let the browser do most of the work, this implementation approach seems problematic (looks like tests are failing too)

packages/lexical-list/src/index.ts
etrepum: Updating from an update listener is the same anti-pattern as updating from a mutation listener. The reason for an update should always have something to do with something external to the editor's state.

etrepum: Regarding deletion I don't think there's a good way to react to a deletion that doesn't cause an update cascade, unattached nodes are ignored from transforms, something else would have to be added to the editor API in order to support this, whether that's something like an update listener that runs before transforms or a registerTransform-like API that executes in the cases when nodes are no longer attached.

I think a solution that depends more on DOM manipulation and CSS counters can be achieved without something like this though, shouldn't really even need a transform to support it.

patrick-atticus:
Without a way to detect deletion, this approach seems like a dead end. 

A CSS approach is reasonable but wont easily work with the way we're generating PDFs in our project, so I'll have to shelf the feature for now.

etrepum:
If you need the feature put in a feature request. I think we'll be adding some more ways to tap into the update/reconciliation cycle, particularly before DOM reconciliation has happened, but feature requests can help guide the implementation constraints around it.

Issue #6370, opened 2024-07-04
Bug: In the Safari browser, during the compositing event process, the delete key exhibits unexpected behavior.


skyline0705:
Lexical version: 0.16.0

## Steps To Reproduce

1. Input Some Text, and then use a different language input, such as Chinese, trigger the composition event, and don't select word or press such as enter key
2. press backslash, when delete the first composition character, it will delete both the start composition character and the character before the first composition character

https://github.com/facebook/lexical/assets/1788280/59ab70e6-0075-411b-922e-5371f667143f

## The current behavior
it will delete the first composition character and the character before the first composition character
## The expected behavior
only delete the beginning composition character

## Impact of fix
I have try it on the version of v0.14.5, it works fine, and above v0.15.0, the mistake appears

yuluo-zy:
I'm also having the same problem, it's a general and serious problem, is there a temporary solution？

yuluo-zy:
Still available on the latest version 0.17

EruditionTu:
This is because on the Safari platform, when the composition input ends, the compositionend event is triggered before the keydown event, causing the isComposition check in onKeydown to be false, resulting in an extra character being deleted by the Backspace.Currently, Lexical has also made similar compatibility adjustments for Firefox, but these adjustments are specifically for the case where the compositionend event is triggered before the input event.https://github.com/facebook/lexical/blob/65ce66afeb2e6f67e40f6c0e2e43bdcf06f17320/packages/lexical/src/LexicalEvents.ts#L177C5-L177C31

etrepum:
This should've been closed by #7061 but a typo in the commit description prevented the automation from closing it

PR #7061, opened 2025-01-17
[lexical] Bug Fix: In the Safari browser, during the compositing event process, the delete key exhibits unexpected behavior.

EruditionTu:
## Description

Fix the issue on the Safari platform where Lexical triggers compositionend before keydown, resulting in an extra character being deleted.

Closes #https://github.com/facebook/lexical/issues/6370

### Before

1. Input Some Text, and then use a different language input, such as Chinese, trigger the composition event, and don't select word or press such as enter key
2. press backslash, when delete the first composition character, it will delete both the start composition character and the character before the first composition character

https://github.com/user-attachments/assets/ac9dd243-bd7f-4095-b749-d5a3df315c62

### After

1. only delete the beginning composition character

https://github.com/user-attachments/assets/154cd3fa-1656-4bfb-9fe8-091cd2ccd51f


etrepum (approved):
I can reproduce this bug in macOS Safari and confirm that this fixes it. I can not reproduce the bug in macOS Firefox or Chrome before this PR, and this PR did not cause any regressions on any of those platforms in my testing.

All of this state in here makes me think we should have a real FSM driving all of these input behaviors across several events, rather than just a jumble of module global variables.

Thank you for the clear video showing how to reproduce the bug, and the fix!

EruditionTu:
Yes, in the subsequent iterations of Lexical, more elegant compatibility handling for input events across various browser platforms is needed, rather than applying patches in this manner. Looking forward to a better Lexical.

Issue #7071, opened 2025-01-20
Bug: Pasting in block results in pasting the content twice in Firefox


GermanJablo:
Originally reported here: https://github.com/payloadcms/payload/issues/10634

Lexical version: 0.23.1

## Steps To Reproduce

1. open https://playground.lexical.dev/ in Firefox
2. Insert a poll component
3. When the selection is outside the poll, click on one of the inputs and paste something
4. The content will be pasted in duplicate inside and outside the poll

https://github.com/user-attachments/assets/e982ac21-9e7d-439e-95a3-75c6b78e0cbe

Apparently it's a selection issue. There is no generic logic that selects any decoratorNode when clicked, but this is done only on some specific decoratorNodes.

## The current behavior

The content will be pasted in duplicate inside and outside the poll

## The expected behavior

The content should be pasted only inside the poll

I have a fix on the way.

etrepum:
I think the best solution to this is probably something else. If it was merely a decorator node selection issue then the issue wouldn't be platform specific. It seems more like there's something wrong with the logic that determines whether to handle an event in a contentEditable=false or not, because also in chrome cmd-A is being captured while that input is focused but that probably shouldn't be the case. Firefox doesn't prevent cmd-A from going to the poll input. The review app in #7072 doesn't fix any of that.

GermanJablo:
I would describe it as a selection problem + something else (which makes it platform specific). 

The Lexical selection still being outside the decorator when the cursor is inside is a bug regardless of whether there is something wrong with the events on contentEditable.

This is just one of the possible side effects that can occur. The one described at the end of #6632 is another. Who knows how many more there may be.

etrepum:
There multiple scenarios where the lexical selection doesn't match the browser's selection by design, otherwise the SelectionAlwaysOnDisplay plug-in would have no purpose. I'm sure that some coherent solution exists but we should come up with something that solves all of the known quirks especially if we need to break compatibility to fix it.

[...]

Issue #7070, opened 2025-01-20
Bug:


Arturszott:
It's very hard to select two or more decorator nodes with mouse pointer in one line. 

## Steps To Reproduce

1. Add two decorator nodes
2. Try to select from right to left two of them.

## The current behavior

https://github.com/user-attachments/assets/45955e61-9086-49e4-9b81-2bc199058046

## The expected behavior

Decorator node is selected quicker and does not require to move the pointer to the end of it.

## Impact of fix
Decorator nodes are more comfortable to select.

Issue #6632, opened 2024-09-13
Bug: Selection does not update to null when selected inside a DecoratorNode


GermanJablo:
Lexical version: 0.17.1

Related to #6622, although this one is probably easier to fix as it doesn't conflict with elements like dialogue or modals.

## Steps To Reproduce

1. In https://playground.lexical.dev/  create a PollNode
2. select something outside the PollNode and then inside. 

## The current behavior

The selection is not updated

## The expected behavior

The selection is updated

## Impact of fix

This leads to undesired behavior. For example, cmd A + cmd X inside a decoratorNode removes all content from the editor without any visible selection. See https://github.com/payloadcms/payload/issues/8047

Issue #3748, opened 2023-01-18
Bug: Decorator block nodes selection selects all nodes or nothing


Piliuta:
Lexical version: 0.7.6

## Steps To Reproduce

1. Add a few decorator nodes (e.g. tweets) next to each other inside a text
2. Start selecting text before the nodes
3. Nodes do not become selected until the text after the nodes is selected (same for de-selection)

## The current behavior

https://user-images.githubusercontent.com/1575198/213175139-d1610d19-ad6e-4e8a-8ddc-2bd93d273bf2.mov

**Note:** This is not an issue with the horizontal rule nodes
https://user-images.githubusercontent.com/1575198/213175349-68ec35f0-7d92-4deb-8e12-cfb6a99d3efa.mov
P.s. on this video you can see a flickering issue, which is a separate one https://github.com/facebook/lexical/issues/3741

trueadm:
I think this is generally an issue with block nodes that contain content. Horizontal rules aren't affected as they don't contain any content. It's basically an annoying browser bug around `contenteditable=false`. There's little we can really do :/

PR #7072, opened 2025-01-20
[lexical-rich-text][lexical-playground][lexical-react]: Update selection when something is selected inside a DecoratorNode

GermanJablo:
Closes:
- https://github.com/facebook/lexical/issues/7071
- https://github.com/facebook/lexical/issues/6632

This PR:
- Update selection when something is selected inside a DecoratorNode.
- The logic is centralized in RichTextPlugin so that it does not have to be repeated in all nodes. This had caused some nodes to forget to implement this logic, or to be inconsistent:
  - Sometimes it seems that the intention was to deselect a decorator node if it was already selected with a second click.
  - Other times it was intended to deselect if shift + click was pressed.
  - Other times clicking on a selected decorator node leaves the selection unchanged. This is the default behavior chosen in this PR, but if anyone prefers either of the above two options please say so.
- I took the opportunity to improve a couple of details in Excalidraw. The option to open with double click was removed for reasons of simplicity and consistency, considering that the component already has a button at the top right to open the editor.

By having the behavior of the click command centralized, fixing other bugs later should be easier, such as:
- https://github.com/facebook/lexical/issues/7070
- https://github.com/facebook/lexical/issues/3748

### DIFF
--- a/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
@@ -16,9 +16,7 @@ import {useLexicalNodeSelection} from '@lexical/react/useLexicalNodeSelection';
 import {mergeRegister} from '@lexical/utils';
 import {
   $getNodeByKey,
-  CLICK_COMMAND,
   COMMAND_PRIORITY_LOW,
-  isDOMNode,
   KEY_BACKSPACE_COMMAND,
   KEY_DELETE_COMMAND,
 } from 'lexical';
@@ -75,35 +73,6 @@ export default function ExcalidrawComponent({
       return;
     }
     return mergeRegister(
-      editor.registerCommand(
-        CLICK_COMMAND,
-        (event: MouseEvent) => {
-          const buttonElem = buttonRef.current;
-          const eventTarget = event.target;
-
-          if (isResizing) {
-            return true;
-          }
-
-          if (
-            buttonElem !== null &&
-            isDOMNode(eventTarget) &&
-            buttonElem.contains(eventTarget)
-          ) {
-            if (!event.shiftKey) {
-              clearSelection();
-            }
-            setSelected(!isSelected);
-            if (event.detail > 1) {
-              setModalOpen(true);
-            }
-            return true;
-          }
-
-          return false;
-        },
-        COMMAND_PRIORITY_LOW,
-      ),
       editor.registerCommand(
         KEY_DELETE_COMMAND,
         $onDelete,
--- a/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
@@ -45,7 +45,6 @@ export default function ExcalidrawComponent({
     data === '[]' && editor.isEditable(),
   );
   const imageContainerRef = useRef<HTMLDivElement | null>(null);
-  const buttonRef = useRef<HTMLButtonElement | null>(null);
   const captionButtonRef = useRef<HTMLButtonElement | null>(null);
   const [isSelected, setSelected, clearSelection] =
     useLexicalNodeSelection(nodeKey);
@@ -190,9 +189,7 @@ export default function ExcalidrawComponent({
         />
       )}
       {elements.length > 0 && (
-        <button
-          ref={buttonRef}
-          className={`excalidraw-button ${isSelected ? 'selected' : ''}`}>
+        <div className={`excalidraw-button ${isSelected ? 'selected' : ''}`}>
           <ExcalidrawImage
             imageContainerRef={imageContainerRef}
             className="image"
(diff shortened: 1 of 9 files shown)


packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -234,9 +200,8 @@ export default function ExcalidrawComponent({
             height={height}
           />
           {isSelected && isEditable && (
```
GermanJablo, line 202: Even considering that double click was allowed, it's odd that the container was a button, and the button a div.
It was confusing to understand the click command I was deleting until I realized this (I thought buttonRef was the small button)

packages/lexical-playground/src/nodes/ExcalidrawNode/ExcalidrawComponent.tsx
```diff
@@ -234,9 +200,8 @@ export default function ExcalidrawComponent({
...
-            <div
+            <button
               className="image-edit-button"
```
GermanJablo, line 204: Even considering that double click was allowed, it's odd that the container was a button, and the button a div.
It was confusing to understand the click command I was deleting until I realized this (I thought buttonRef was the small button)

packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -300,11 +299,6 @@ export default function ImageComponent({
         },
         COMMAND_PRIORITY_LOW,
       ),
-      editor.registerCommand<MouseEvent>(
-        CLICK_COMMAND,
-        onClick,
```
GermanJablo, line 305: I'm not removing the onClick handler because it's used for the right button. See https://github.com/facebook/lexical/pull/5056

packages/lexical/src/LexicalEditor.ts
```diff
@@ -786,7 +786,7 @@ export class LexicalEditor {
    * deterministically in the order of registration.
    *
    * @param command - the command that will trigger the callback.
-   * @param listener - the function that will execute when the command is dispatched.
+   * @param listener - the function that will execute when the command is dispatched. Returns true to stop propagation, false to continue.
```
GermanJablo, line 789: Forgive me for this change that is irrelevant to the objective of the PR. But I always have to think about what true and false mean 😅

etrepum (commented):
It looks like there are failing tests

packages/lexical/src/LexicalEditor.ts
etrepum: This is a common source of confusion 👍

packages/lexical-rich-text/src/index.ts
```diff
@@ -565,13 +566,25 @@ export function registerRichText(editor: LexicalEditor): () => void {
...
+        if (!decorator) {
+          return false;
+        }
+        editor.update(() => {
+          const node = $getNearestNodeFromDOMNode(decorator);
+          if ($isDecoratorNode(node)) {
+            const selection = $createNodeSelection();
+            selection.add(node.getKey());
+            $setSelection(selection);
+          }
```
etrepum, line 585: This seems like a breaking change because now it works differently and only one node can be selected, where before you could select multiple nodes. This is also a bit tricky to override without having its own command.

GermanJablo: > This seems like a breaking change because now it works differently and only one node can be selected

For me this is more of a bug fix than a breaking change. Currently the selection was being set to null which is definitely worse and doesn't make much sense. Probably many people haven't noticed this because the most used nodes like ImageNode fix the behavior with their commands.

> where before you could select multiple nodes

From the frontend (let's say the playground), I know how to select multiple nodes with a rangeSelection, but I don't know of any way I could select multiple nodes with a NodeSelection, neither before nor after this PR. I think that it's a good thing, since I don't know in what scenario someone would want something like that.

> This is also a bit tricky to override without having its own command.

Same as above. I don't see this PR changing anything in that regard. Customizing the behavior used to require a command, now it does too.

etrepum: I get that this behavior makes more sense but it’s probably not backwards compatible, anyone upgrading to this PR is going to have to audit their code to be compliant. Without a backwards compatible upgrade path it might be tricky or at least take longer to get meta to accept it

etrepum: To select multiple nodes with a NodeSelection you click one and then shift-click the others. They don't even have to be adjacent. Try creating three divider nodes, click the first one, and then shift-click the third one. I don't think this makes a lot of sense to do for this particular kind of node, but it's how the editor works now and changing that would break compatibility.

The tricky thing about breaking compatibility is we don't know what we don't know. The unit and e2e test don't even come close to covering what people are doing in their own projects. Based on the scope of this affecting events for all decorators I think this would require at least some Meta folks looking closely at how all of their projects are using events and decorators. If we're going to just patch the problem I suspect there's a way we can do it that just brings Firefox behavior in alignment with the other platforms and doesn't change anything fundamental.

GermanJablo: I'm making a PR in our codebase to fix this on our end.
Feel free to close this PR if you feel the breaking change isn't worth it!
If, on the other hand, it is a change that you do want to incorporate, let me know and I'll take a look at the failing tests.

etrepum:

PR #7075, opened 2025-01-21
[lexical-react]Feature : CharacterCountPlugin

### LINKED ISSUE #7063: Feature: Request for Adding useCharacterCount Hook
## Description

I have implemented a new `useCharacterCount` hook inspired by the existing `useCharacterLimit` functionality. This hook helps track the number of characters in real-time within text input fields or rich text editors.

Although `useCharacterLimit` and `useCharacterCount` are not typically used together, an issue arises when both hooks are used simultaneously, causing the following error:

```
ToolbarContext.tsx:100 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
I am currently facing difficulties resolving this issue and would appreciate help fixing it so that both hooks can work together. Also, I would like to ask whether it would be appropriate to submit a pull request for this.
```
You can find the implementation of the `useCharacterCount` hook [here](https://github.com/h1un/lexical/commit/843a8693562a709f3c58dfa8cdb105b1adcdf424).

## Impact

Text Input Validation: Track the number of characters a user types in real-time and display an alert when the character count exceeds a set limit.
Rich Text Editing: Track character count in a rich text editor as users type, which can be useful for applications like document editing.
Form Validation: Dynamically validate form inputs based on character count, ensuring the form is only submitted if the character count is within the allowed range.

I request that the `useCharacterCount` hook be improved to work seamlessly alongside `useCharacterLimit`, resolving the issue when both hooks are used together. Also, I would like to know if it would be appropriate to submit a pull request for [this](https://github.com/h1un/lexical/commit/843a8693562a709f3c58dfa8cdb105b1adcdf424).

h1un:
## Description
I have implemented a new `useCharacterCount` hook inspired by the existing `useCharacterLimit` functionality. This hook helps track the number of characters in real-time within text input fields or rich text editors.

Closes #7063

### DIFF
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -7,6 +7,7 @@
  */
 
 import {AutoFocusPlugin} from '@lexical/react/LexicalAutoFocusPlugin';
+import {CharacterCountPlugin} from '@lexical/react/LexicalCharacterCountPlugin';
 import {CharacterLimitPlugin} from '@lexical/react/LexicalCharacterLimitPlugin';
 import {CheckListPlugin} from '@lexical/react/LexicalCheckListPlugin';
 import {ClearEditorPlugin} from '@lexical/react/LexicalClearEditorPlugin';
@@ -85,8 +86,10 @@ export default function Editor(): JSX.Element {
       isCollab,
       isAutocomplete,
       isMaxLength,
+      isCharCount,
       isCharLimit,
       hasLinkAttributes,
+      isCharCountUtf8,
       isCharLimitUtf8,
       isRichText,
       showTreeView,
@@ -256,6 +259,9 @@ export default function Editor(): JSX.Element {
             charset={isCharLimit ? 'UTF-16' : 'UTF-8'}
             maxLength={5}
           />
+        )}
+        {(isCharCount || isCharCountUtf8) && (
+          <CharacterCountPlugin charset={isCharCount ? 'UTF-16' : 'UTF-8'} />
         )}
         {isAutocomplete && <AutocompletePlugin />}
         <div>{showTableOfContents && <TableOfContentsPlugin />}</div>
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -734,6 +734,16 @@ i.page-break,
   color: red;
 }
 
+.characters-count {
+  color: #888;
+  font-size: 12px;
+  text-align: right;
+  display: block;
+  position: absolute;
+  left: 55px;
+  bottom: 5px;
+}
+
 .dropdown {
   z-index: 100;
   display: block;
--- a/packages/lexical-react/src/LexicalCharacterCountPlugin.tsx
+++ b/packages/lexical-react/src/LexicalCharacterCountPlugin.tsx
@@ -17,12 +17,15 @@ function utf8Length(text: string) {
 
 export function CharacterCountPlugin({
   charset = 'UTF-16',
+  render = DefaultRenderer,
 }: {
   charset?: 'UTF-16' | 'UTF-8';
+  render?: (characterCount: number) => JSX.Element;
 }): JSX.Element {
   const [editor] = useLexicalComposerContext();
-  const characterCount = useCharacterCount(editor, {
-    strlen: (text: string) => {
+
+  const strlen = React.useMemo(() => {
+    return (text: string) => {
       if (charset === 'UTF-8') {
         return utf8Length(text);
       } else if (charset === 'UTF-16') {
@@ -30,8 +33,14 @@ export function CharacterCountPlugin({
       } else {
         throw new Error('Unrecognized charset');
       }
-    },
-  });
+    };
+  }, [charset]);
 
+  const characterCount = useCharacterCount(editor, {strlen});
+
+  return render(characterCount);
+}
+
+function DefaultRenderer(characterCount: number) {
   return <span className="characters-count">{characterCount}</span>;
 }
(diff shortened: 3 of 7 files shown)


packages/lexical-react/src/shared/useCharacterCount.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  useEffect(() => {
+    // Check if OverflowNode is registered
+    if (!editor.hasNodes([OverflowNode])) {
+      invariant(
+        false,
+        'useCharacterCount: OverflowNode not registered on editor',
+      );
+    }
+  }, [editor]);
```
etrepum, line 33: Counting characters doesn't need OverflowNode

packages/lexical-react/src/LexicalCharacterCountPlugin.tsx
```diff
@@ -0,0 +1,37 @@
...
+    strlen: (text: string) => {
+      if (charset === 'UTF-8') {
+        return utf8Length(text);
+      } else if (charset === 'UTF-16') {
+        return text.length;
+      } else {
+        throw new Error('Unrecognized charset');
+      }
+    },
+  });
```
etrepum, line 34: This will cause the text content listener to be unregistered every time the editor changes because strlen changes causing the effect to re-fire every time. The options object or at least strlen should be computed with useMemo with this design.

packages/lexical-react/src/LexicalCharacterCountPlugin.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return <span className="characters-count">{characterCount}</span>;
```
etrepum, line 36: It would be better to return `<>{characterCount}</>` so the tag and class are entirely up to the user, lexical generally doesn't have any hard-coded class names.

h1un:
@etrepum 
I received your feedback and modified the code. Thank you.

etrepum:
I’ll take a look after you sign the CLA. This project can not accept PRs without a signed CLA.

packages/lexical-react/src/LexicalCharacterCountPlugin.tsx
etrepum: The only code in lexical that hard-codes class names is the playground, which is not published as a reusable module like this. Otherwise classes come from the editor theme (which is generally for nodes, not plug-ins). I think this should just be headless, where if you don't specify a render function it returns only the text.

packages/lexical-react/src/shared/useCharacterCount.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type OptionalProps = {
+  strlen?: (input: string) => number;
+};
+
+export function useCharacterCount(
+  editor: LexicalEditor,
+  optional: OptionalProps = Object.freeze({}),
+): number {
+  const {strlen = (input) => input.length} = optional;
```
etrepum, line 20: This has the same useMemo type problem, if this default is used a new function will be created each time the hook is called. If you extract this to a separate function it won't have that problem.

packages/lexical-react/src/shared/useCharacterCount.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+import type {LexicalEditor} from 'lexical';
+
+import {useEffect, useState} from 'react';
+
+type OptionalProps = {
```
etrepum, line 12: It would be easier to use and produce better API documentation if this was exported and had a more descriptive name like UseCharacterCountOptions

packages/lexical-react/src/shared/useCharacterCount.ts
```diff
@@ -0,0 +1,36 @@
...
+  strlen?: (input: string) => number;
+};
+
+export function useCharacterCount(
```
etrepum, line 16: This should have a `/** API documentation */`

packages/lexical-react/src/LexicalCharacterCountPlugin.tsx
```diff
@@ -0,0 +1,46 @@
...
+import * as React from 'react';
+
+import {useCharacterCount} from './shared/useCharacterCount';
+
+function utf8Length(text: string) {
+  const textEncoder = new TextEncoder();
+  return textEncoder.encode(text).length;
+}
+
+export function CharacterCountPlugin({
```
etrepum, line 18: This should have a `/** API documentation */`

packages/lexical-playground/src/Editor.tsx
```diff
@@ -257,6 +260,9 @@ export default function Editor(): JSX.Element {
             maxLength={5}
           />
         )}
+        {(isCharCount || isCharCountUtf8) && (
+          <CharacterCountPlugin charset={isCharCount ? 'UTF-16' : 'UTF-8'} />
```
etrepum, line 264: This is what it would look like when the output of the plugin is headless

```suggestion
          <span className="character-count">
            <CharacterCountPlugin charset={isCharCount ? 'UTF-16' : 'UTF-8'} />
          </span>
```

packages/lexical-playground/src/index.css
```diff
@@ -734,6 +734,16 @@ i.page-break,
   color: red;
 }
 
+.characters-count {
```
etrepum, line 737:
```suggestion
.character-count {
```

etrepum (commented):
Looks very close, mostly just needs API documentation

etrepum:
The feedback on this PR hasn't been addressed in a few months so it will be closed as stale. You may reopen it when/if you are able to finish it up.

PR #7077, opened 2025-01-22
[lexical-table] Feature: TableCellNode add verticalAlign attribute

liuweiGL:
## Description

TableCellNode adds the verticalAlign attribute so that users can control the vertical alignment of the table

### Before

### After

https://github.com/user-attachments/assets/92c0587f-793c-40af-ac39-2b8417a5c43d


ivailop7:
Thank you for this PR. I think it's almost there. In order to keep the node representation in the editorState compact and the HTML output smaller, I think the default for the verticalAlign should be 'undefined' since it's 'top' be default and it should only have a value if middle or bottom and nothing if 'top'. This way, you also won't have to modify the table tests and you can add a new table test specifically testing this new feature.

liuweiGL:
I have made the adjustments. Please review it for me again.

ivailop7 (dismissed):
LGTM, @etrepum to sign off as well.

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -188,6 +197,7 @@ export class TableCellNode extends ElementNode {
       colSpan: this.__colSpan,
       headerState: this.__headerState,
       rowSpan: this.__rowSpan,
+      verticalAlign: this.getVerticalAlign(),
```
etrepum, line 200: Something like this would prevent it from being serialized unless it is a non-default value:

```suggestion
      ...(isValidVerticalAlign(this.__verticalAlign) && { verticalAlign: this.__verticalAlign }),
```

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -49,6 +49,7 @@ export type SerializedTableCellNode = Spread<
     headerState: TableCellHeaderState;
     width?: number;
     backgroundColor?: null | string;
+    verticalAlign?: null | string;
```
etrepum, line 52: No reason to allow null to be serialized here 

```suggestion
    verticalAlign?: string;
```

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -65,6 +66,8 @@ export class TableCellNode extends ElementNode {
   __width?: number | undefined;
   /** @internal */
   __backgroundColor: null | string;
+  /** @internal */
+  __verticalAlign: null | string;
```
etrepum, line 70: Using undefined is a bit more natural here

```suggestion
  __verticalAlign?: undefined | string;
```

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -111,7 +115,8 @@ export class TableCellNode extends ElementNode {
       .setColSpan(serializedNode.colSpan || 1)
       .setRowSpan(serializedNode.rowSpan || 1)
       .setWidth(serializedNode.width || undefined)
-      .setBackgroundColor(serializedNode.backgroundColor || null);
+      .setBackgroundColor(serializedNode.backgroundColor || null)
+      .setVerticalAlign(serializedNode.verticalAlign || null);
```
etrepum, line 119:
```suggestion
      .setVerticalAlign(serializedNode.verticalAlign || undefined);
```

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -126,6 +131,7 @@ export class TableCellNode extends ElementNode {
     this.__headerState = headerState;
     this.__width = width;
     this.__backgroundColor = null;
+    this.__verticalAlign = null;
```
etrepum, line 134: No need to set when undefined is the default

```suggestion
```

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -249,6 +259,16 @@ export class TableCellNode extends ElementNode {
     return self;
   }
 
+  getVerticalAlign(): null | string {
```
etrepum, line 262:
```suggestion
  getVerticalAlign(): undefined | string {
```

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -249,6 +259,16 @@ export class TableCellNode extends ElementNode {
...
+    return this.getLatest().__verticalAlign;
+  }
+
+  setVerticalAlign(newVerticalAlign: null | string): this {
+    const self = this.getWritable();
+    self.__verticalAlign = newVerticalAlign;
+    return self;
+  }
```
etrepum, line 270:
```suggestion
  setVerticalAlign(newVerticalAlign: null | undefined | string): this {
    const self = this.getWritable();
    self.__verticalAlign = newVerticalAlign || undefined;
    return self;
  }
```

etrepum (commented):
I think we should tighten up this serialization just a little bit to show best practices. This way the unrelated tests won't have to change, and the JSON will be smaller.

etrepum (commented):
I think the only thing left here to do is revert the changes to the serialization tests where `"verticalAlign": null` was added. Code looks good otherwise!

liuweiGL:
@etrepum Everything is ok now

etrepum (approved):
Looks great, nice work!

PR #7086, opened 2025-01-23
[lexical-mark] Feature: include inline decorator nodes in marks

james-atticus:
## Description

`$wrapSelectionInMarkNode` currently wraps text nodes and inline element nodes, leaving other nodes unwrapped. This makes the function also include inline decorator nodes.

## Test plan

Marks only appear to be used by comments in the Playground. I couldn't see any existing unit tests for marks, or E2E tests for comments, so went with manual testing for this.

### Before

The mark does not include the equation node, though it is captured in the comment text:

image

### After

The mark includes the equation node:

image

### DIFF
--- a/packages/lexical-mark/__tests__/unit/LexicalMarkNode.test.ts
+++ b/packages/lexical-mark/__tests__/unit/LexicalMarkNode.test.ts
@@ -1,0 +1,185 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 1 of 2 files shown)


etrepum (commented):
It would be great to add some more tests around this, to prevent future regressions and to show that this works as intended

james-atticus:
@etrepum I'll add some tests on Tuesday

etrepum (approved):

PR #7097, opened 2025-01-25
[lexical-playground][table-cell-resizer] Bug Fix: column width resize issues

zruq:
The current `updateColumnWidth` implementation only resizes the current column, causing awkward behavior especially when combined with the `width: fit-content` table style (see "Before" video). This update adjusts the function to resize both the current and next column, and ensures the total table width only changes when resizing the last column, matching the behavior of other text editors.

**Before** 

https://github.com/user-attachments/assets/18dd319d-047d-4c46-aa98-00a48d188262

**After**

https://github.com/user-attachments/assets/30e51698-f624-41f1-b91f-336d50ad79ca


ivailop7:
Tests seem to be failing

packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -210,7 +210,6 @@
   overflow-y: scroll;
   overflow-x: scroll;
   table-layout: fixed;
-  width: fit-content;
```
etrepum, line 213: Removing this breaks scrollable tables

zruq: In Firefox it makes the table takes all the available width, which causes inconsistencies between the cell width in lexical and the rendered cell width.
Though using `max-content` instead it allows scrollable tables in both Firefox and Chrome.

zruq:
> I think the UX of this could probably use some work, it feels broken to use because when you create a table and try and resize any column but the last nothing happens. The current version of tables in lexical and other apps like google docs do not behave like this.

Tables in apps like google docs or microsoft word do behave like this. When columns are reduced to their minimum width, you can't resize them further (except for increasing the width of the last column). 
The UX problem is apparent because in lexical when a table is created all columns have minimum width, while in google docs the table takes full width of the page. So I think changing starting width for columns will fix this problem, but I'm not sure what's a good starting value

etrepum:
The whole table gets wider when you resize a column in google docs just after creating the table. I’m sure you’re right that at some point a maximum width is reached where the size has to be removed from an adjacent column but that’s not always the case and doesn’t really make sense when the tables are allowed to overflow and scroll like in lexical’s playground

zruq:
> The whole table gets wider when you resize a column in google docs
I was using pages format, that had the other behavior.
> doesn’t really make sense when the tables are allowed to overflow and scroll like in lexical’s playground
I see, thank you.

PR #7098, opened 2025-01-26
[lexical-yjs] Bug Fix: handle text node being split by Yjs redo

james-atticus:
## Description

Fixes a bug in `applyChildrenYjsDelta` which caused Lexical and Yjs to get out of sync. Once in this state, some changes would not sync to other clients; the changes would be lost on refresh.

This can be reproduced in the Playground with the following steps:

1. In the left editor, type "Hello World"
2. Select "World" and change the formatting to bold
3. In the right editor, type a character in the middle of "World", eg "Wor_ld"
4. In the left editor, click undo then redo
5. Any text added at the end in the left editor won't be synced to the right editor

When applying change deltas from Yjs, an `insertDelta` can be a string or a `sharedType`. If `currIndex` is pointing to some position within a text node and `insertDelta` is a string, then the new text is spliced in correctly. If `insertDelta` represents a shared type, then the newly created `collabNode` is inserted in the wrong position (_before_ said text node).

This PR updates the logic in `applyChildrenYjsDelta` to handle inserting a `sharedType` node in the middle of a text node.

## Test plan

The repro steps above are covered in an end-to-end test. I added a second E2E test to cover the case where the first `insertDelta` corresponds to an `ElementNode`, then the second delta is the `TextNode`.

### Before

https://github.com/user-attachments/assets/5e5d9186-3db2-4251-b118-35f57496b2e8

### After

https://github.com/user-attachments/assets/d9569c31-d1b2-4b2c-a766-f43debf48521

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
@@ -413,4 +413,256 @@ test.describe('Collaboration', () => {
       `,
     );
   });
+
+  test('Undo/redo where text node is split by formatting change', async ({
+    isRichText,
+    page,
+    isCollab,
+    browserName,
+  }) => {
+    test.skip(!isCollab);
+
+    // Left collaborator types two words, sets the second one to bold.
+    await focusEditor(page);
+    await page.keyboard.type('normal bold');
+
+    await sleep(1050);
+    await selectCharacters(page, 'left', 'bold'.length);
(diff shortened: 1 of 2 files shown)


ivailop7 (approved):

etrepum (approved):
I haven't done an audit of exactly how this works but it seems unlikely to be problematic given that it passes the existing e2e suite and the new test. Locally the change looks like it makes sense.

PR #7101, opened 2025-01-27
Nodes Lexical: 1 update to grammar - Update nodes.md

DocAdam:
## Description
1 update to grammar

'are that' should be switched to 'that are'


etrepum (commented):
I'm not quite sure that this is the best grammar fix since the validation referenced here is checking that an arbitrary node is a custom node (`$isCustomParagraphNode`), not some validation for nodes known to be the custom node type. It could be clearer either way, just not sure this change is in the right direction.

PR #7117, opened 2025-01-31
[lexical] Feature: add a generic state property to all nodes

GermanJablo:
This PR enables adding arbitrary state to nodes with automatic JSON serialization.

### 1. Easier node customization

The following compares adding a 'color' property to a text node using the current node replacement API versus the state API in this PR.

<details>
  <summary>With node replacement API</summary>

```ts
// IMPLEMENTATION
export type SerializedColoredTextNode = Spread<
{
  color: string;
  type: 'colored';
  version: 1;
},
SerializedTextNode
>;

export class ColoredNode extends TextNode {
__color: string;

constructor(text: string, color?: string, key?: NodeKey): void {
  super(text, key);
  this.__color = color || "defaultColor";
}

static getType(): string {
  return 'colored';
}

setColor(color: string) {
  self.__color = color;
}

getColor(): string {
  return self.__color;
}

static clone(node: ColoredNode): ColoredNode {
  return new ColoredNode(node.__text, node.__color, node.__key);
}

  const element = super.createDOM(config);
  element.style.color = this.__color || "defaultColor";
  return element;
}

updateDOM(prevNode: ColoredNode, dom: HTMLElement, config: EditorConfig): boolean {
  const isUpdated = super.updateDOM(prevNode, dom, config);
  if (prevNode.__color !== this.__color) {
    dom.style.color = this.__color;
  }
  return isUpdated;
}

static importJSON(serializedNode: SerializedMentionNode): MentionNode {
  const node = new ColoredNode(serializedNode.text, node.serializedNode.color);
  node.setFormat(serializedNode.format);
  node.setDetail(serializedNode.detail);
  node.setMode(serializedNode.mode);
  node.setStyle(serializedNode.style);
  return node;
}

exportJSON(): SerializedHeadingNode {
  return {
    ...super.exportJSON(),
    color: this.getColor()
    tag: this.getTag(),
    type: 'colored',
    version: 1,
  };
}
}

export const nodes = [
  ColoredNode,
  {
    replace: TextNode,
    with: (node: TextNode) => new ColoredNode(node.__text),
    withKlass: ColoredNode,
  },
  // ...
]

// USAGE
const textNode = $createTextNode();
textNode.setColor("blue");
const textColor = textNode.getColor() // "blue"

```

</details>

<details>
  <summary>With state API</summary>

```ts
// IMPLEMENTATION
const colorState = createState('color', {
  parse: (value: unknown) => (typeof value === 'string' ? value : undefined),
});

// USAGE
const textNode = $createTextNode();
$setState(textNode, colorState, 'blue');
const textColor = $getState(textNode, colorState); // "blue"

```

</details>

### 2. Composability

If two users replace the same node with the node replacement API, the system fails.
This is particularly relevant when you do not fully control the codebase. For example, when installing a Lexical plugin that requires a custom node.

### 3. Document metadata

Sometimes it can be useful to store some metadata about the document, such as its ID or creation date.
It can also sometimes be useful to store maps for faster lookups. For example, the comments plugin could have a map of comment IDs to node keys.
This API allows storing that information in RootNode.

___

Notes: 
- This PR is a new iteration of https://github.com/facebook/lexical/pull/6929 but less opinionated and with a more limited scope.
- This API does not come with opinions on how to render the state or serialize it to HTML. Therefore, the example comparison with node replacement API from section 1 is not entirely fair right now since it does not include create/updateDOM. However, the plan is to soon provide a plugin to handle this.
- The API design is the brainchild of @etrepum. Credit to him for also guiding the development.


.eslintrc.js
```diff
@@ -88,7 +88,10 @@ module.exports = {
         ],
         '@typescript-eslint/ban-ts-comment': OFF,
         '@typescript-eslint/no-this-alias': OFF,
-        '@typescript-eslint/no-unused-vars': [ERROR, {args: 'none'}],
+        '@typescript-eslint/no-unused-vars': [
+          ERROR,
+          {args: 'none', argsIgnorePattern: '^_', varsIgnorePattern: '^_'},
+        ],
```
GermanJablo, line 94: If I may, I would like to propose this form of opt-out which has legitimate uses. For example, in LexicalNode.test.ts when I define type variables just for the purpose of testing the types.
Also, while developing a WIP it's convenient to be able to mute warnings without having to write comments everywhere.

etrepum: I prefer this setting also, but maybe there’s some meta style guide reason for not having this? /cc @zurfyx

packages/lexical-website/docs/concepts/node-replacement.md
```diff
@@ -1,12 +1,49 @@
...
+
+// Or you can add all your state inside an object:
+type AwesomeLexical = {
+  color?: string;
+  bgColor?: string;
+  padding?: number
+}
+const awesomeLexical = createStateKey('awesome-lexical', /** your parse fn which returns AwesomeLexical type */)
+
+```
```
GermanJablo, line 38: I tried to see if collision checking could be done at runtime, but I came to the conclusion that it's not a good idea.
- Key registration would have to be global
- It would be difficult to differentiate it from re-executions of the same definition (e.g. useEffect)
- it might even be better to leave the conflict silent than crash the program. Two plugins using `bg-color` would probably do the same thing.

In any case, the recommendation here in the documentation would still be valid.

packages/lexical-website/docs/concepts/node-replacement.md
```diff
@@ -1,12 +1,49 @@
+# Node Customization
```
GermanJablo, line 1: I don't know how you handle redirects on your website, but I think it would be nice to rename the url to this title.

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
```diff
@@ -1497,3 +1506,125 @@ describe('LexicalNode tests', () => {
...
+          paragraph.setState(indentKey, 1);
+          const json2 = paragraph.exportJSON();
+          expect(json2.state).toStrictEqual({
+            indent: 1,
+          });
+          // TODO: How could exportJson know which stateKey maps to each key?
+          // set the default value explicitly
+          // paragraph.setState(indentKey, 0);
+          // const json3 = paragraph.exportJSON();
+          // expect(json3.state).not.toHaveProperty('indent');
```
GermanJablo, line 1587: As I said about collisions above, I think having a global registry would make things more complex. An alternative I can think of is using setState to remove the default value:
```ts
  setState<T extends StateKey>(k: T, v: T['value']) {
+   if (k.parse(undefined) === v) {
+     delete (self.__state as State)[k.key];
+   }
    (self.__state as State)[k.key] = v;
  }

```

etrepum: The map+object implementation that I described in the chat would resolve this since the keys of the map are state keys and not strings

packages/lexical-website/docs/concepts/node-replacement.md
etrepum: A map+object API would allow this kind of check on first use although to make this efficient you’d use Map<string, StateKey> and have the object be underlying storage (or have a separate object for parsed values)

packages/lexical-table/src/__tests__/unit/LexicalTableSelection.test.tsx
```diff
@@ -134,6 +134,7 @@ describe('table selection', () => {
       __parent: null,
       __prev: null,
       __size: 1,
+      __state: {},
```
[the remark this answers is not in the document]
etrepum: I think this will probably be necessary to implement for correct yjs behavior

packages/lexical-website/docs/concepts/node-replacement.md
```diff
@@ -1,12 +1,53 @@
...
+The advantages of using state over the replacement API are:
+1. Easier (less boilerplate)
+2. Composable (multiple plugins extending the same node causes failures)
+3. Allows metadata: useful for adding things to the RootNode.
+
+All you need to do is define keys with `createStateKey`, and then use it with the `setState` and `getState` methods.
+
+```ts
+// IMPLEMENTATION
+const color = createStateKey('color', { parse: (value) => value as string });
```
etrepum, line 16: We should demonstrate a correct parse function, this one is very incorrect because it silently includes undefined in the result but not its type even in the best case

packages/lexical-website/docs/concepts/node-replacement.md
```diff
@@ -1,12 +1,53 @@
...
+
+// USAGE
+const textNode = new TextNode();
```
etrepum, line 19: We should always demonstrate constructor function usage such as $createTextNode rather than directly constructing nodes

packages/lexical/src/LexicalNode.ts
```diff
@@ -184,6 +185,39 @@ export type DOMExportOutput = {
 
 export type NodeKey = string;
 
+type DeepImmutable<T> = T extends Map<infer K, infer V>
```
etrepum, line 188: With the current single object implementation all values must be JSON serializable so Map, Set, etc. can’t be supported (they don’t have a toJSON that would work)

packages/lexical/src/LexicalNode.ts
```diff
@@ -198,6 +232,19 @@ export class LexicalNode {
   __prev: null | NodeKey;
   /** @internal */
   __next: null | NodeKey;
+  /** @internal */
+  __state: DeepImmutable<State> = {};
+
+  getState<T extends StateKey>(k: T): T['value'] | undefined {
+    const self = this.getLatest();
+    // If the state is not set, return the default value
+    return self.__state[k.key] ?? k.parse(undefined);
```
etrepum, line 241: With a single object implementation this must always call parse because we don’t know if its directly from JSON or not

packages/lexical/src/LexicalNode.ts
```diff
@@ -198,6 +232,19 @@ export class LexicalNode {
...
+  }
+
+  setState<T extends StateKey>(k: T, v: T['value']) {
+    const self = this.getWritable();
+    (self.__state as State)[k.key] = v;
```
etrepum, line 246: Unless the state object is copied on every clone, or there’s some way to know if this should be frozen or not, then this must create a new object every time. Creating a new object on every set would probably also be the only way to support yjs without also tweaking enumerable (which would require explicit support for state in yjs, but at least it wouldn’t be guiding the implementation of this feature’s runtime state)

packages/lexical/src/LexicalNode.ts
```diff
@@ -934,6 +982,7 @@ export class LexicalNode {
   updateFromJSON(
     serializedNode: LexicalUpdateJSON<SerializedLexicalNode>,
   ): this {
+    this.__state = serializedNode.state || {};
```
etrepum, line 985: This logic also needs to go into afterCloneFrom, and probably worth adding tests to make sure that this feature doesn’t rewrite history (setting state shouldn’t affect previous reconciled versions of the node)

GermanJablo: Any idea how to test it? I tried comparing `_editorState` with `_pendingEditorState` without success

packages/lexical-table/src/__tests__/unit/LexicalTableSelection.test.tsx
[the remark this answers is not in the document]
GermanJablo: I already followed your suggestion to copy the state object in setState, so this suggestion is no longer necessary?
On the other hand, do you know what a unit test for yjs could look like?

packages/lexical/src/LexicalNode.ts
```diff
@@ -198,6 +232,19 @@ export class LexicalNode {
...
+  getState<T extends StateKey>(k: T): T['value'] {
+    const self = this.getLatest();
+    // If the state is not set, return the default value
+    return self.__state[k.key] ?? k.parse(undefined);
+  }
+
+  setState<T extends StateKey>(k: T, v: T['value']) {
+    const self = this.getWritable();
+    self.__state = {...self.__state, [k.key]: v};
+  }
```
etrepum, line 247: should return `this` for chaining like other mutator methods

```suggestion
  setState<T extends StateKey>(k: T, v: T['value']): this {
    const self = this.getWritable();
    self.__state = {...self.__state, [k.key]: v};
    return self;
  }
```

packages/lexical/src/LexicalNode.ts
```diff
@@ -198,6 +232,19 @@ export class LexicalNode {
   __prev: null | NodeKey;
   /** @internal */
   __next: null | NodeKey;
+  /** @internal */
+  __state: DeepImmutable<State> = {};
+
+  getState<T extends StateKey>(k: T): T['value'] {
+    const self = this.getLatest();
+    // If the state is not set, return the default value
+    return self.__state[k.key] ?? k.parse(undefined);
```
etrepum, line 241: This can be optimized later with a more complex representation of state, it's not correct as-is

```suggestion
    return k.parse(self.__state[k.key]);
```

packages/lexical-table/src/__tests__/unit/LexicalTableSelection.test.tsx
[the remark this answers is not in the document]
etrepum: I am not familiar with the intricacies of testing lexical-yjs off-hand, I just know its implementation is very problematic because it manipulates node properties directly rather than going through the constructor and well defined serializations. I'd find another node that uses a non-primitive property (something to do with comments or marks probably?) and see if there's any relevant testing. There might not be any good testing for this sort of thing currently.

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
```diff
@@ -1497,3 +1506,143 @@ describe('LexicalNode tests', () => {
...
+          });
+          root.setState(keyForString, 'hello');
+        };
+        expect(fn).toThrow();
+      });
+
+      test(`getState and setState`, async () => {
+        const keyForString = createStateKey('keyForString', {
+          parse: (value) => value as string,
+        });
```
etrepum, line 1535: This type is a lie because it will return undefined and call it a string. Not sure what the point of an intentionally wrong cast here is?

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
```diff
@@ -1497,3 +1506,143 @@ describe('LexicalNode tests', () => {
...
+
+          // TO-DO:
+          // It would be a bit nicer if createStateKey also gave a compilation error if parse tried to return null.
+          // It's not a big deal, because getState does give an error. The problem is that I can't activate it.
+          // The @expect-error does not work because strict mode is disabled in `tsconfig.test.json`.
+          const _keyForMaybeNull = createStateKey('keyForMaybeNull', {
+            parse: (value) => value as string | null,
+          });
+          // // @ts-expect-error - null is not a valid value
+          // const _maybeNullValue = root.getState(keyForMaybeNull);
```
etrepum, line 1561: I don't really understand your comment here, there's no error because you can explicitly cast unknown to anything that you like without any sort of warning or error in typescript. The only difference from `any` is that it requires an explicit "trust me bro". It's perfectly valid to have a function with this type signature, it's just the implementation that's wrong.

```suggestion
          const keyForMaybeNull = createStateKey('keyForMaybeNull', {
            parse: (value) => typeof value === 'string' ? value: null,
          });
          const maybeNullValue = root.getState(keyForMaybeNull);
          expect(maybeNullValue).toBe(null);
```

packages/lexical/src/LexicalNode.ts
```diff
@@ -184,6 +185,39 @@ export type DOMExportOutput = {
...
+  ? ReadonlyMap<DeepImmutable<K>, DeepImmutable<V>>
+  : T extends Set<infer S>
+  ? ReadonlySet<DeepImmutable<S>>
+  : T extends object
+  ? {
+      readonly [K in keyof T]: DeepImmutable<T[K]>;
+    }
+  : T;
+type State = {[Key in string]?: string | number | boolean | State};
+type StateValue = string | number | boolean | undefined | State;
```
etrepum, line 198: `null` should be a valid inhabitant of all of these types

packages/lexical/src/LexicalNode.ts
```diff
@@ -184,6 +185,39 @@ export type DOMExportOutput = {
...
+interface StateKey<
+  K extends string = string,
+  V extends StateValue = StateValue,
+> {
+  key: K;
+  // Here we are storing a default for convenience
+  value: DeepImmutable<V>;
+  parse: (value: unknown) => V;
```
etrepum, line 206:
```suggestion
  readonly key: K;
  // Here we are storing a default for convenience
  readonly value: DeepImmutable<V>;
  readonly parse: (value: unknown) => V;
```

packages/lexical/src/LexicalNode.ts
```diff
@@ -184,6 +185,39 @@ export type DOMExportOutput = {
...
+}
+interface StateKeyConfig<V extends StateValue = StateValue> {
+  parse: (value: unknown) => V;
```
etrepum, line 209:
```suggestion
  readonly parse: (value: unknown) => V;
```

packages/lexical/src/LexicalNode.ts
```diff
@@ -198,6 +232,19 @@ export class LexicalNode {
   __prev: null | NodeKey;
   /** @internal */
   __next: null | NodeKey;
+  /** @internal */
+  __state: DeepImmutable<State> = {};
```
etrepum, line 236: A more precise type would be something like this. `__state = 12` should not be valid.

```suggestion
  __state: { [k in string]?: DeepImmutable<State> } = {};
```

packages/lexical/src/LexicalNode.ts
etrepum:
```ts
let v0: RootNode;
let v1: RootNode;
const vk = createStateKey('vk', { parse: (v) => typeof v === 'number' ? v : null });
editor.update(
  () => {
    v0 = $getRoot();
    v0.setState(vk, 0);
    expect(v0.getState(vk)).toBe(0);
  },
  {discrete: true},
);
const state0 = editor.getEditorState();
editor.update(
  () => {
    v0.setState(vk, 1);
    // normally would use chaining for this but setState doesn't return this
    v1 = v0.getLatest();
    // This is testing getLatest()
    expect(v0.getState(vk)).toBe(1);
    expect(v1.getState(vk)).toBe(1);
    expect(v1.is(v0)).toBe(true);
  },
  {discrete: true},
);
const state1 = editor.getEditorState();
// Test that the correct version is returned and that they are independent
expect(state0.read(() => v0.getState(vk))).toBe(0);
expect(state1.read(() => v1.getState(vk))).toBe(1);
// Test that getLatest is used and not the __state property directly
expect(state0.read(() => v1.getState(vk))).toBe(0);
expect(state1.read(() => v0.getState(vk))).toBe(1);
```

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
GermanJablo: yeah, I wanted to test the type (line 1539). What you say I'm doing in 1544 to 1551. Here I could throw an error if value is not typeof string, but for the purposes of this test it doesn't matter.

packages/lexical/src/LexicalNode.ts
GermanJablo: I find it better to incentivize undefined for smaller serialization.

packages/lexical/src/LexicalNode.ts
GermanJablo: You are confusing the `State` and `StateValue` types. As it stands right now `__state = 12` is invalid

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
etrepum: Your intent to restrict the type doesn't really provide any benefit, given `parse: (value) => typeof value === 'string' ? value : null` then `null` will never be serialized if the implementation follows the guidelines that I proposed.

There are use cases where `null` is an explicit value that's different from the default (which could be undefined, or any other value), and your proposed type makes that use case impossible.

packages/lexical/src/LexicalNode.ts
etrepum: Not necessary. When null is the default then it won't be serialized. If null is not the default then it's an explicit choice to include null and there's no good reason to prevent that.

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
etrepum: Yeah but there's no good reason to lie here because we're not testing a runtime invariant. At least `(value as string) || ''` or something like that would make more sense. Sometimes people look at the tests to see how to use something and this is just lying for no good reason because undefined is always an inhabitant.

packages/lexical/src/__tests__/unit/LexicalNode.test.ts
etrepum: For more clarity here is a tsplayground demonstrating how I think this should all work https://tinyurl.com/rv8jutvt

<details>
<summary>code below for posterity</summary>

```ts
import { z } from 'zod';
import { Schema, Option } from 'effect';

export type DeepImmutable<T> = T extends Map<infer K, infer V>
  ? ReadonlyMap<DeepImmutable<K>, DeepImmutable<V>>
  : T extends Set<infer S>
  ? ReadonlySet<DeepImmutable<S>>
  : T extends object
  ? {
      readonly [K in keyof T]: DeepImmutable<T[K]>;
    }
  : T;
export type State = {[Key in string]?: StateValue};
// This could be relaxed in a future version, but for now we only
// support values that are directly JSON serializable
export type StateValue = string | number | boolean | null | undefined | State;
export interface StateKey<
  K extends string = string,
  V extends StateValue = StateValue,
> extends StateKeyConfig<V> {
  readonly key: K;
  // Here we are storing a default for convenience
  readonly value: DeepImmutable<V>;
}
export interface StateKeyConfig<V extends StateValue = StateValue> {
  readonly parse: (value?: unknown) => V;
}
export type StateKeyConfigValue<T extends StateKeyConfig> = ReturnType<T['parse']>;

export function createStateKey<K extends string, T extends StateKeyConfig<any>>(
  key: K,
  config: T,
): StateKey<K, StateKeyConfigValue<T>> {
  return {key, value: config.parse(undefined), parse: config.parse.bind(config)};
}

// It's best for us to demonstrate actual parsers, but we probably
// shouldn't ship a parse library since many users will already have
// something (zod, effect, whatever)
function parseString<Default extends StateValue>(defaultValue: Default): StateKeyConfig<string | Default> {
  return { parse: (v: unknown) => typeof v === 'string' ? v : defaultValue };
}
const skNull = createStateKey("test", parseString(null));
//    ^?
const skUndefined = createStateKey("test", parseString(undefined));
//    ^?
const skString = createStateKey("test", parseString(''));
//    ^?

const zkNull = createStateKey("test", z.string().nullable().catch(null));
//    ^?
const zkUndefined = createStateKey("test", z.string().optional().catch(undefined));
//    ^?
const zkString = createStateKey("test", z.string().catch(''));
//    ^?

function parseSchema<A extends StateValue, I, Default extends StateValue>(schema: Schema.Schema<A, I, never>, defaultValue: Default): StateKeyConfig<A | Default> {
  return { parse: (value) => Option.getOrElse(Schema.decodeUnknownOption(schema)(value), () => defaultValue) };
}

const ekNull = createStateKey("test", parseSchema(Schema.String, null));
//    ^?
const ekUndefined = createStateKey("test", parseSchema(Schema.String, undefined));
//    ^?
const ekString = createStateKey("test", parseSchema(Schema.String, ''));
//    ^?
```
</details>

packages/lexical/src/LexicalNode.ts
GermanJablo: I already consider null to be an error in the JS language, and although [I am not alone](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-null.md#why), I accept that it is debatable.
However, in this case I do think there's a good reason. Unlike other libraries like React, in Lexical the state is serialized.
If null is used as a non-default value, people are going to get a larger json, perhaps due to oversight or ignorance that undefined compresses better. I think it's a good idea to guide them with typescript here.

etrepum: I agree that having both undefined and null is not a particularly good idea, and would go even further to say that [having any pervasive null or undefined type is a mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/), but that's what JavaScript is and I don't think we should take a strong stance against it. lexical in general tends towards null vs. undefined so this design decision is inconsistent with how the rest of the library works.

Users would really have to go pretty far out of their way to have null as a non-default value. I think you are trying very hard to constrain the problem to force an optimization that would've happened without the constraint in any non-contrived scenario.

.eslintrc.js
GermanJablo: Marked as resolved to reduce noise in the PR. Feel free to reopen if there are objections.

packages/lexical-website/docs/concepts/node-replacement.md
GermanJablo: I'll note this down for later. Marked as resolved to remove PR noise

packages/lexical/src/LexicalNode.ts
```diff
@@ -882,9 +939,23 @@ export class LexicalNode {
...
    * */
   exportJSON(): SerializedLexicalNode {
+    const state: State = {};
+    Object.entries(this.__state).forEach(([key, value]) => {
+      const config = stateStore.get(key);
+      if (!config) {
+        throw new Error(
+          `There has been an attempt to export a state with the key "${key}", but it is not registered.`,
+        );
+      }
```
etrepum, line 949: I think it's better not to have an error here, particularly for versioning reasons. If you have two versions of the editor and the new version writes a document with a new key, it's better for the old version to leave it as-is rather than throw an error on save or discard it.

packages/lexical/src/LexicalNode.ts
```diff
@@ -184,6 +185,47 @@ export type DOMExportOutput = {
...
+  // Here we are storing a default for convenience
+  readonly value: DeepImmutable<V>;
+  readonly parse: (value: unknown) => V;
+}
+interface StateKeyConfig<V extends StateValue = StateValue> {
+  readonly parse: (value: unknown) => V;
+}
+type StateKeyConfigValue<T extends StateKeyConfig> = ReturnType<T['parse']>;
+
+const stateStore = new Map<string, StateKeyConfig>();
```
etrepum, line 213: We probably shouldn't have a global state store, you should be able to configure states to work differently for multiple editors on the page in the same way you can define multiple nodes with the same type. A `WeakMap<Editor, Map<string, StateKeyConfig>>` could work for this purpose, or it could be directly added to the editor. It would only be able to test on first use, but we can add state keys to editor/plugin/node config to detect conflicts early in a later version (e.g. with lexical builder)

packages/lexical/src/LexicalNode.ts
```diff
@@ -297,6 +312,12 @@ export class LexicalNode {
     this.__prev = null;
     this.__next = null;
     $setNodeKey(this, key);
+    Object.defineProperty(this, '_state', {
+      configurable: true,
+      enumerable: false,
+      value: new NodeState(),
+      writable: true,
+    });
```
GermanJablo, line 320: Something I thought about but haven't tested yet is whether it would work to use a private javascript property instead.

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -0,0 +1,131 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+/* eslint-disable dot-notation */
+
```
GermanJablo, line 9: The reason I suggest using bracket-notation here is to give these properties more privacy in a safer way than the `__` prefix.
In LexicalNode, I've defined `__state` because it's a pattern that's already all over the codebase and because the property is not enumerable.

see https://github.com/microsoft/TypeScript/issues/19335#issuecomment-338003928

etrepum (commented):
Overall I think this is a good approach. I'm currently working on some optimizations and basic support for lexical-yjs, after that I'll push the changes for you to take a look at.

As far as implementation details go, I've removed the DeepImmutable stuff. It's not something that's done elsewhere in lexical or for React.useState and it will make the types and certain use cases more awkward than they need to be. I've also removed the private bracket syntax stuff, it mostly just makes internal usage awkward and we don't try very hard to do implementation hiding elsewhere beyond marking things accordingly.

GermanJablo:
I don't mind the bracket syntax thing. I don't agree with DeepImmutable, though. I think it's a great improvement on DX without a noticeable disadvantage. I don't think React is the gold standard either. We didn't go with the tuple after all.

etrepum:
I'm pretty sure that the DeepImmutable stuff would get in the way of adding support for certain advanced use cases, like deeper integration with collab (e.g. supporting a YMap nested in the state), and it made some other things in the implementation awkward so it didn't seem worth the effort and extra syntax. The ideal way to provide that kind of immutability is to do it at the parse function, but that's something that the user should choose (e.g. zod or something like it).

I refactored the PollNode in the playground to use state instead of properties as a test of the collab support.

I did add some more convenience methods to the StateConfig, nodeGetter and nodeSetter, which reduces some of the boilerplate to add that functionality as node methods.

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -0,0 +1,365 @@
...
+  } else {
+    value = valueOrUpdater;
+  }
+  const writable = node.getWritable();
+  const state = $getWritableNodeState(writable);
+  $registerConfigToState(node, cfg, state);
+  state.knownState.set(cfg, value);
+  return writable;
+}
+
```
GermanJablo, line 215: For context, we had discussed internally 3 possible APIs:

```ts
// additional createState arguments omitted here for simplicity  

const paragraph = new ParagraphNode();  

// OPTION 1  
const indentKey = createState("indent");  
const indent = paragraph.getState(indentKey);  
paragraph.setIndent(indentKey, indent + 1);  

// OPTION 2  
const [getIndent, setIndent] = createState("indent");  
const indent = getIndent(paragraph);  
setIndent(paragraph, indent + 1);  

// OPTION 3  
const indentState = createState("indent");  
const indent = indentState.$get(paragraph);  
indentState.$set(paragraph, indent + 1);  

```

In this last commit, it seems you want to use all 3 at once, and add a 4th with `$getState` and `$setState`.

For simplicity I suggest we stick to just one.

If your preference is for option 3, I would:
- remove `$getState` and `$setState` (or not export them if you want to use them as an internal implementation).
- remove the `accessors` utility to get tuples. To do that I would go for option 2 instead, or use aliases: 
`const { $get: getIndent, $set: setIndent } = createState("indent")`.
- remove the inheritance pattern that appears in the tests, which is not exactly the same as option 1 but is quite similar

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -0,0 +1,365 @@
...
+   * Similarly, if your V is a function, then usage of {@link $setState}
+   * must use an updater function because your type will be indistinguishable
+   * from an updater function.
+   */
+  parse: (jsonValue: unknown) => V;
+  /**
+   * You may specify a function that converts V back to JSON.
+   * This is mandatory when V is not a JSON serializable type.
+   */
+  valueToJSON?: (parsed: V) => unknown;
```
GermanJablo, line 107: I understand the motivation for `valueToJson`, but I think that in order to have a much simpler and minimalist API it would be better to remove it. In terms of DX it is the same for the end user to have to write a toJSON as to have to write a fromJSON.
By omitting this option the user could use the fromJSON reviver instead when consuming the data.

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -0,0 +1,365 @@
...
+  /**
+   * Used to define the equality function so you can use an Array or Object
+   * as V and still omit default values from the exported JSON.
+   */
+  isEqual?: (a: V, b: V) => boolean;
```
GermanJablo, line 112: About `isEqual`: Similar to the comment about valueToJSON.
I don't think we should make things that complex. We can compare with `Object.is` like react does, and if people want to mutate an object or an array, it is their responsibility to clone it with a spread operator or whatever.

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -0,0 +1,365 @@
...
+ * Set the state defined by cfg on node. Like with `React.useState` you may
+ * directly specify the value or use an updater function that will be called
+ * with the previous value of the state on that node (which will be the
+ * `cfg.defaultValue` if not set).
+ *
+ * When an updater function is used, the node will only be marked dirty if
+ * `cfg.isEqual(prevValue, value)` is false.
+ *
+ * @param node The LexicalNode to set the state on
+ * @param cfg The configuration for this state
```
GermanJablo, line 188: I say we rename cfg to configState

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -0,0 +1,365 @@
...
+    this.parse = config.parse.bind(config);
+    this.valueToJSON = (config.valueToJSON || coerceToJSON).bind(config);
+    this.isEqual = (config.isEqual || Object.is).bind(config);
+    this.defaultValue = this.parse(undefined);
+    this.$get = (node) => $getState(node, this);
+    this.$set = (node, value) => $setState(node, this, value);
+  }
+  get accessors(): [$get: this['$get'], $set: this['$set']] {
+    return [this.$get, this.$set];
+  }
```
GermanJablo, line 44: about "get accessors". See comment I made on $setState

packages/lexical/src/index.ts
```diff
@@ -166,6 +166,16 @@ export type {
...
+export {
+  $getState,
+  $getWritableNodeState,
+  $setState,
+  createState,
+  type StateConfig,
+  type StateGetter,
+  type StateSetter,
+  type StateValueConfig,
+} from './LexicalNodeState';
```
GermanJablo, line 178: Why are we exporting anything other than `createState`?

packages/lexical-playground/src/nodes/PollNode.tsx
```diff
@@ -95,59 +119,56 @@ export class PollNode extends DecoratorNode<JSX.Element> {
...
-    return {
-      ...super.exportJSON(),
-      options: this.__options,
-      question: this.__question,
-    };
-  }
+  getQuestion = questionState.nodeGetter<this>();
+  setQuestion = questionState.nodeSetter<this>();
+  getOptions = optionsState.nodeGetter<this>();
+  setOptions = optionsState.nodeSetter<this>();
```
GermanJablo, line 125: same as I said about `LexcialNodeState.test.ts`

GermanJablo (commented):
Thanks, I see a lot of improvements. My biggest concerns though are:
- A very large API with many ways to achieve the same thing.
- Patterns for getters and setters: Object form, tuple form, inheritance, $getState / $setState.
- I'm not sure that allowing non-jsonifiable values ​​and introducing `valueToJson` will improve things.
- `isEqual` seems to complicate things unnecessarily. We can compare with `Object.is` like react does, and if people want to mutate an object or an array, it is their responsibility to clone it with a spread operator or whatever.
- exports that I see no reason for

I didn't check the yjs section. And the documentation may need to be updated depending on how the final API turns out.

packages/lexical/src/__tests__/unit/LexicalNodeState.test.ts
```diff
@@ -0,0 +1,277 @@
...
+    return 'state';
+  }
+  static clone(node: StateNode) {
+    return new StateNode(node.__key);
+  }
+  static importJSON(serializedNode: SerializedLexicalNode): TestNode {
+    return new StateNode().updateFromJSON(serializedNode);
+  }
+  getNumber = numberState.nodeGetter();
+  setNumber = numberState.nodeSetter();
```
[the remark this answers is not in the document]
etrepum: The lowest-level API is better to be something like:

```ts
$getState(node, cfg);
$setState(node, cfg, value);
```

Any methods we add to cfg or node are optional. If we want to simplify as much as possible I say we get rid of all the methods and just have these two functions as the interface.

I don't think the tuple makes any sense as the primitive, we're going to want a single object to reference the cfg, and you don't need to do anything with scope like react hooks are doing so it's only done at the module level.

Avoiding inheritance isn't the only use case for this proposal. I suspect that some people like the chaining API and this is a secondary benefit of state in that you can get rid of all your boilerplate for clone/json/etc. and just define the data type and you can generate the getter/setter if you want them. This also allows you to transition from directly implementing state to using NodeState without breaking any API like I did here (although some follow-up work will be required to let you flatten the JSON for these known keys).

packages/lexical/src/LexicalNode.ts
```diff
@@ -296,6 +301,12 @@ export class LexicalNode {
     this.__parent = null;
     this.__prev = null;
     this.__next = null;
+    Object.defineProperty(this, '__state', {
+      configurable: true,
+      enumerable: false,
+      value: undefined,
```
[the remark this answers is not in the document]
etrepum: The reason it's done this way is because state is shared across node versions until it changes, so you really want to start with a trivial state in the initializer and undefined was the best choice in this scenario because in the constructor the logic to figure out exactly which state to use would be more expensive than handling an empty value in the very few cases where this property is accessed directly. In previous versions, there wasn't state sharing across nodes and the NodeState didn't store a node reference.

NodeState is now initialized on the *first* setState or when updateFromJson is called with a *non-empty* state. If a node isn't using NodeState, then its __state property will never change from undefined.

Clones of that node (future versions) will re-use the previous NodeState until a writable state is requested so there is no penalty for updates that do not make updates to the NodeState.

packages/lexical/src/LexicalNodeState.ts
etrepum: This isn't practical when using data types that you don't want  to subclass or monkeypatch like Map or Set or when working with more exotic types like a Y.Map from yjs (which isn't fully supported just yet but could be with a few more lines of code in collab)

packages/lexical/src/LexicalNodeState.ts
etrepum: It's optional and for advanced usage. It is necessary if your data type is something like an array and you are not memoizing that default.

```ts
// This will always serialize
createState("array", {parse: (v): readonly string[] => Array.isArray(v) ? v : []});
// This will skip serialization only if the default was never changed or if cfg.defaultValue is always returned in your setters when it becomes empty
const emptyArray: readonly string[];
createState("array", {parse: (v): readonly string[] => Array.isArray(v) ? v : emptyArray});
// This will work as users probably expect it to
createState("array", {
  parse: (v): readonly string[] => Array.isArray(v) ? v : emptyArray,
  isEqual: (a, b) => a === b || a.length === b.length && JSON.stringify(a) === JSON.stringify(b),
});
```

packages/lexical/src/index.ts
etrepum: Collab needs access to the internals via `$getWritableNodeState` and it is marked `@internal`. `$getState` and `$setState` are the lowest level APIs that provide good encapsulation, I don't think we really need the methods on StateConfig.

The types are just obvious things you'd want to export otherwise you'd have to do awkward things like `ReturnType<typeof createState>` or whatever. Not exporting them doesn't help anything.

packages/lexical/src/LexicalNodeState.ts
GermanJablo: Yes, I guess it would make things easier. One idea then: we could leverage typescript to automatically warn the user that they need a `valueToJson`, if what they parse is not jsonifiable (doesn't extend State)

packages/lexical/src/__tests__/unit/LexicalNodeState.test.ts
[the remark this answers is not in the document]
GermanJablo: > If we want to simplify as much as possible I say we get rid of all the methods and just have these two functions as the interface

I'm fine with that. I don't have strong preferences for any of the (now) 4 options. I care more that it's just one.

packages/lexical/src/LexicalNodeState.ts
etrepum: TypeScript doesn't have any way for users to define warnings. It is something we could consider for the future in the linter, but that would be a pretty significant effort that I do not want to sign up for today.

It's probably possible to do some shenanigans with a very complex type function to provide an error for createState when V is not a primitive and an explicit encoder/isEqual are not provided, although I don't think having that page of type-level code (and the day or so that it would take to perfect it) is worth it right now. It would technically be a breaking change in the future but a pretty shallow one to fix. This is also not something I am volunteering to do, I don't think this additional safety for advanced use cases is high impact. Enabling advanced use cases is a lot more important to me than adding extra safety when doing those advanced things.

etrepum (commented):
With the latest pass I think we are in a pretty good place. More documentation/tests/examples would of course be useful (e.g. fine-grained testing of the minimal collab support) but I don't think those should necessarily be release blocking.

Some notable changes since the initial proposal:

* The public API is now function based:
  - `createState(key, stateValueConfig)`
  - `$getState(node, stateConfig)`
  - `$setState(node, stateConfig, valueOrUpdater)`
* React.useState style updater functions are supported
* StateValueConfig now has two optional properties for advanced use cases:
  - `parse: (jsonValue: unknown) => V` is the same, the required parsing function
  - `isEqual?: (a: T, b: T) => boolean` is a new optional function which is useful for advanced use cases such as using non-primitive types (e.g. use something other than Object.is)
  - `unparse?: (parsed: V) => unknown` is a new optional function which is useful for advanced use cases such as using non-primitive types (Array, Map, Set, Y.Map, etc.) or could even be used to prevent a value that is intended to be ephemeral from being serialized at all (useful for plugins like autocomplete, spell check, etc.)
* There are a number of optimizations to facilitate copy-on-write and parse-only-once semantics that don't really have any effect on the interface but should improve confidence about the cost of using this feature

Some minimal use-at-your-own-risk support is now built into collab, with the PollNode being rewritten to use state for its question and options. It does not currently support the old serialization outside of state, but providing a way for nodes to flatten specific keys is something worth exploring for when people do start migrating away from the boilerplate-heavy explicit properties.

In order to support all styles of usage, a small `makeStateWrapper` function is provided in @lexical/utils to provide the boilerplate for three styles of accessors.

```ts
const myStateWrapper = makeStateWrapper(createConfig(…));
// users may prefer this to the top-level functions
myStateWrapper.$get(node);
myStateWrapper.$set(node, valueOrUpdater);
// useState style
const [$getMyState, $setMyState] = myStateWrapper.accessors;
// method generators
class MyNode extends TextNode {
  // … normal boilerplate elided
  getMyState = myStateWrapper.makeGetterMethod<this>();
  setMyState = myStateWrapper.makeSetterMethod<this>();
}
```

The wrapper object is specifically useful for generating the methods because you can't both explicitly specify some (`this`) and infer other (`K`, `V`) generics in the same function call.

Some additional things to consider:
- Should we use `state` (JSON) and `__state` (LexicalNode) as the keys? We could easily use something even less likely to collide with user code and serializations such as `"@"`, `"_"`, `"$"` or even `""` with the added benefit of shaving off a few bytes each time it is serialized. For the `__state` property we could use a symbol instead. I don't have a strong preference about any of this.
- Flow types
- Check the formatting of the API docs
- Do we get rid of StateConfigWrapper, and/or eliminate some subset of it?

GermanJablo:
1. I honestly still don't understand the usefulness of isEqual. But if the rest of the team sees its usefulness, it doesn't matter :)
2. Now that we accept non-jsonifiable values ​​with `valueToJson`, I'm wondering if we shouldn't rename `parse` and `valueToJson` to (1) `exportJSON` and `importJSON` or (2) `toJSON` and `fromJSON`. I know that technically `parse` doesn't run, at least not yet, on exportJSON, but it's a technical subtlety and I think it would provide more clarity, as well as being more consistent with Lexical's nomenclature.
3. Similar to the above point, but I'm wondering if it's not a weird pattern to define the default value as the result of parse(undefined), and it would actually be more convenient to have one property for default, another for import, and another for export?
4. > Should we use state (JSON) and __state (LexicalNode) as the keys? We could easily use something even less likely to collide with user code and serializations such as "@", "_", "$" or even "" with the added benefit of shaving off a few bytes each time it is serialized... I don't have a strong preference about any of this.

    Same about the "no strong preference", but I actually think those are two good reasons. I like "_" or "$".
6. > Do we get rid of StateConfigWrapper, and/or eliminate some subset of it?

    One thing I liked about the option of providing ONLY the tuple or ONLY the object with `$get` and `$set`, was that it reduced the things to import. In this case for each state I have to import the `$getState`, `$setState` and the respective `stateConfig`. Anyway, I still maintain that I prefer to provide only one way to manipulate the state above all, so whatever API remains, I would get rid of `StateConfigWrapper`.
7. Among the pile of comments I can't find a specific conversation we had about `DeepImmutable`, maybe it was deleted. I'd like to leave it here as a reference, and see what others think. 
Originally I was wrapping the response of `$getState` in a generic [DeepImmutable](https://stackoverflow.com/a/78673120/10476393) to help people not unsafely mutate objects. @etrepum felt that it wasn't worth the trouble and could be a pain in the yjs binding. I think the latter is easily solved by having DeepImmutable wrap the response of `$getState`, but not the `__state` type. Anyway, I understand that it's not a conventional practice, but it seemed like a nice thing to have without a downside IMO. We could test how it works, and if we don't like it, remove it. The opposite is not true because it could be a breaking change.
7. NIT: There are still uses of the term `cfg` in some places in the tsdoc. I suggest renaming it to `stateConfig`.

etrepum:
1. At some point, you'll probably be happy it's there. It saves a lot of user code when the value is not a primitive type, otherwise the user has to be very careful with their default and mutators to make sure some memoized copy is provided whenever it is equivalent to the default.
2. `exportJSON` and `toJSON` are all methods with no arguments. `valueToJSON` is not a method and it has an argument. `toJSON` would also have weird behavior (an exception, possibly) if you managed to `JSON.stringify` the configuration object. I chose the name `parse` because that's what it does, it allows certain schema libraries like zod to directly be used as a configuration object, it's shorter than `importJSON`, and there are some scenarios where the transport isn't technically JSON (such as with yjs). I've renamed `valueToJSON` to `unparse` so it's very clear what it's supposed to do, without implying that it works some other way.
3. It doesn't really make sense to have a separate setting for a default, the parse function has to do something with undefined or invalid inputs anyway. Having the separate default just means having users write more code for no benefit.
4. I'll wait for input from others to see about whether we rename the keys. Having smaller serializations is nice, I'm not sure if anyone expects to get a subclass collision with `__state` or not.
5. There is still only one way in the core to do anything with state, the wrapper is in another package altogether for optional convenience so users can use the same things with a bit less code
6. I hid the conversations about DeepImmutable as outdated because it's not relevant to the current implementation, they're still there just collapsed to make this easier to read. You're underestimating the downsides on this for sure.
7. Changed the rest of the cfg references to stateConfig in the docs and implementation

etrepum:
I just thought of something else that we may want to add here, `$getState` may need an option to force reading the exact state from that node rather than the latest state for use in `updateDOM`. I think that’s the only situation where we intentionally pass old node references around without the EditorState that they came from. Maybe a better idea would be to have a `$getStateChanges(node, prevNode)` that returns a `Map<StateConfig, [value, prevValue]>` that is only populated with entries that differ.

etrepum:
After literally sleeping on it I went with the plan of adding an optional `version` argument to `$getState` where the user can choose `'latest'` (the default) or `'direct'`. There's also one more function `$getStateChange(node, prevNode, stateConfig)` that is a three line function to make it easier to do this correctly from `updateDOM`. Unit tests included, but since the PollNode example is a decorator it didn't make sense to exercise it there.

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -156,18 +159,55 @@ export function createState<K extends string, V>(
 }
 
 /**
- * The accessor for working with node state
+ * Given two versions of a node and a stateConfig, compare their state values
+ * using `$getState(nodeVersion, stateConfig, 'direct')`.
+ * If the values are equal according to `stateConfig.isEqual`, return `null`,
+ * otherwise return `[value, prevValue]`.
+ *
+ * This is useful for implementing updateDOM. Note that the `'direct'`
```
GermanJablo, line 167: Is it a feature intended for internal use, or do you see a use for it in user land?

etrepum: It's useful in userland for the same reasons you can read properties like __text directly in userland

etrepum:
There's one more edge case to think through, the automatic merge of TextNode doesn't consider state which will get in the way of using this for text. Will be a little tricky to do a generic comparison between states but not impossible

etrepum:
Some motivating cases for this:
* Don't really need MarkNode anymore, can just mark any node with a use-case-specific mark and use a plug-in to add an appropriate class or attribute to each marked node (probably using an update listener to find them and then manipulate DOM). Doesn't really make anything more complex because we already have to split MarkNodes across blocks.
* We could consider lifting style to state so we can scope concern specific properties like React does instead of working with the whole thing as a string (and get rid of the global memory leaking global cache of style string->object)
* Can add GUIDs to nodes for collab/persistence use cases
* Can store metadata on the root, which combined with marks could power something like comments or any other "invisible isolated decorator". Probably also eliminates some of the use cases for having a nested editor where the parent editor has some decorator to store metadata and maybe render some UI plus the "real" editor.
* Gets rid of a lot of need to subclass nodes
* When you do subclass nodes, you can skip all of the clone/json boilerplate if your additional properties are based on NodeState and the boilerplate will automatically elide default values from serialization
* With a little bit of extra work, non-primitive state values could directly support collab use cases (e.g. store a Y.Map directly) so you can control conflict resolution (NodeState itself already does this)
* Also, with some extra work, can provide full dev tools support since these values are completely introspectable either as direct values or via JSON interop (since we have a parse and unparse for any state)

Basically, other than backwards compat and *maybe* performance reasons for the lowest-level nodes, I don't think there is any motivating use case for adding extra __props to any node when you can use NodeState (which is what #7189 explores)

etrepum:
I arbitrarily picked "$" as the key to use but moved the choice to LexicalConstants (NODE_STATE_KEY) so if there's a preference for some other value once this gets reviewed then it is easy to change.

packages/lexical/src/LexicalNodeState.ts
```diff
@@ -0,0 +1,736 @@
...
+   * @internal
+   *
+   * Track the (versioned) node that this NodeState was created for, to
+   * facilitate copy-on-write for NodeState. When a LexicalNode is cloned,
+   * it will *reference* the NodeState from its prevNode. From the nextNode
+   * you can continue to read state without copying, but the first $setState
+   * will trigger a copy of the prevNode's NodeState with the node property
+   * updated.
+   */
+  readonly node: LexicalNode;
```
etrepum, line 347: This could be a WeakRef (someday…) or a WeakSet but the cost of that extra object is probably higher than the cost of referencing at most one old version of the node.

packages/lexical/src/LexicalNode.ts
```diff
@@ -296,6 +301,12 @@ export class LexicalNode {
     this.__parent = null;
     this.__prev = null;
     this.__next = null;
+    Object.defineProperty(this, '__state', {
+      configurable: true,
+      enumerable: false,
```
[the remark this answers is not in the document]
GermanJablo: Also so as not to break a handful of tests

etrepum (approved):
I'm going to put this in an approved state, but since I worked on this PR and it's a significant new API I will wait for feedback before merging.

Issue #5042, opened 2023-09-24
Bug: Unable to backspace inline images on Android


alexgleason:
Lexical version: https://playground.lexical.dev/

This seems related to https://github.com/facebook/lexical/issues/1965

## Steps To Reproduce

1. Add an inline image to the document on Android
2. Press backspace
3. The keyboard disappears

## Video

https://github.com/facebook/lexical/assets/3639540/d8ba07bd-5671-48b5-9f39-3f565b4c38d0

## Other info

In my application, I cannot backspace emojis on Android after implementing them in Lexical. I decided to see if maybe it was an issue in Lexical.

https://github.com/facebook/lexical/assets/3639540/f78ca3c5-2c5d-46af-843c-5cb1d03e4a32

My code is here: https://gitlab.com/soapbox-pub/soapbox/-/blob/main/src/features/compose/editor/nodes/emoji-node.tsx

alexgleason:
Adding `user-select: none` to my EmojiNode fixed it! https://github.com/facebook/lexical/pull/4040 

How incredibly delicate. I still don't get why Android backspace is so problematic.

I'm convinced now this is exactly the same problem as https://github.com/facebook/lexical/issues/1965. But #4040 is a workaround rather than a fix.

Jibbedi:
I have the same experience. The keyboard closes when hitting backspace in my decorator node which uses `user-select:none`. 

It also closes in the playground on inline images and equation nodes. 

I'm using the GBoard on a Samsung Device. 
Interestingly, if I switch to the stock Samsung Keyboard the nodes get deleted without closing the Keyboard. 

Maybe this helps narrowing the issue. 
Happy Holidays for all :-)

Vellaiyan-Marimuthu:
Has anyone fixed this issue?

Facundo1609:
Hey! I have the same problem when I try to delete an Image or an InlineImage in the playground example https://playground.lexical.dev/

I have this problem only if I use it on Android device, on iOS device it is working!

ThenTech:
Running into the same issue with a custom `DecoratorNode`, also only on mobile with GBoard. The backspace event does not seem to be triggered when pressing backspace when using the GBoard keybaord on Android. On desktop, everything works as expected.

Has anyone found a workaround for this? Or knows what options should be set/overridden when implementing `DecoratorNode` to let it behave consistently?

PR #7122, opened 2025-02-02
[lexical] Bug Fix: Fix Chrome on android deletion bugs

### LINKED ISSUE #4340: Bug: When deleting text with the keyboard of a device it's also removing text to the right of the pointer.

waynetee:
## Description

Fixes #4340, #5042 and #1965 

1. On Chrome on Android, deletions are usually handled by the browser instead of the `DELETE_CHARACTER_COMMAND`, however the browser is unable to deleted decorator nodes, this PR detects this scenario and dispatches `DELETE_CHARACTER_COMMAND` accordingly
2. When deleting text across paragraphs, Chrome on Android shifts the selection rightwards, this PR restores the correct selection

### Before
https://github.com/user-attachments/assets/b5842cfe-9fba-47a4-a063-6f13766b25f6

### After
https://github.com/user-attachments/assets/87ffbb2b-1dce-4b6a-ad72-a321b2ec41c7

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -606,7 +606,7 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
           const hasSelectedAllTextInNode =
             selection.anchor.offset === 0 &&
             selection.focus.offset === selectedNodeText.length;
-          const shouldLetBrowserHandleDelete =
+          let shouldLetBrowserHandleDelete =
             IS_ANDROID_CHROME &&
             isSelectionAnchorSameAsFocus &&
             !hasSelectedAllTextInNode &&
@@ -614,7 +614,9 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
           // Check if selection is collapsed and if the previous node is a decorator node
           // If so, the browser will not be able to handle the deletion
           if (shouldLetBrowserHandleDelete && selection.isCollapsed()) {
-            shouldLetBrowserHandleDelete = !$isDecoratorNode($getAdjacentNode(selection.anchor, true));
+            shouldLetBrowserHandleDelete = !$isDecoratorNode(
+              $getAdjacentNode(selection.anchor, true),
+            );
           }
           if (!shouldLetBrowserHandleDelete) {
             dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -613,14 +613,10 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
             selectedNodeCanInsertTextAfter;
           // Check if selection is collapsed and if the previous node is a decorator node
           // If so, the browser will not be able to handle the deletion
-          let browserAbleToDelete = true;
-          if (selection.isCollapsed()) {
-            const prevNode = $getAdjacentNode(selection.anchor, true);
-            if ($isDecoratorNode(prevNode)) {
-              browserAbleToDelete = false;
-            }
+          if (shouldLetBrowserHandleDelete && selection.isCollapsed()) {
+            shouldLetBrowserHandleDelete = !$isDecoratorNode($getAdjacentNode(selection.anchor, true));
           }
-          if (!(shouldLetBrowserHandleDelete && browserAbleToDelete)) {
+          if (!shouldLetBrowserHandleDelete) {
             dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
           }
         }
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -25,6 +25,7 @@ import {
   $getPreviousSelection,
   $getRoot,
   $getSelection,
+  $isDecoratorNode,
   $isElementNode,
   $isRangeSelection,
   $isRootNode,
@@ -79,6 +80,7 @@ import {getActiveEditor, updateEditorSync} from './LexicalUpdates';
 import {
   $findMatchingParent,
   $flushMutations,
+  $getAdjacentNode,
   $getNodeByKey,
   $isSelectionCapturedInDecorator,
   $isTokenOrSegmented,
@@ -609,7 +611,16 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
             isSelectionAnchorSameAsFocus &&
             !hasSelectedAllTextInNode &&
             selectedNodeCanInsertTextAfter;
-          if (!shouldLetBrowserHandleDelete) {
+          // Check if selection is collapsed and if the previous node is a decorator node
+          // If so, the browser will not be able to handle the deletion
+          let browserAbleToDelete = true;
+          if (selection.isCollapsed()) {
+            const prevNode = $getAdjacentNode(selection.anchor, true);
+            if ($isDecoratorNode(prevNode)) {
+              browserAbleToDelete = false;
+            }
+          }
+          if (!(shouldLetBrowserHandleDelete && browserAbleToDelete)) {
             dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
           }
         }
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalEvents.ts
```diff
@@ -609,7 +611,16 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
...
+          // Check if selection is collapsed and if the previous node is a decorator node
+          // If so, the browser will not be able to handle the deletion
+          let browserAbleToDelete = true;
+          if (selection.isCollapsed()) {
+            const prevNode = $getAdjacentNode(selection.anchor, true);
+            if ($isDecoratorNode(prevNode)) {
+              browserAbleToDelete = false;
+            }
+          }
+          if (!(shouldLetBrowserHandleDelete && browserAbleToDelete)) {
```
etrepum, line 623: It would make more sense to declare `shouldLetBrowserHandleDelete` with `let` so you can mutate it instead of having two variables here, or define a function that implements the logic that you could include when initially defining `shouldLetBrowserHandleDelete`.

e.g. 

```suggestion
          // Check if selection is collapsed and if the previous node is a decorator node
          // If so, the browser will not be able to handle the deletion
          if (shouldLetBrowserHandleDelete && selection.isCollapsed()) {
            shouldLetBrowserHandleDelete = !$isDecoratorNode($getAdjacentNode(selection.anchor, true));
          }
          if (!shouldLetBrowserHandleDelete) {
```

etrepum (approved):
This looks "fine" to me, I think the setTimeout is setting ourselves up for some future clean-up. I don't have an android chrome device to do any testing with so I am not volunteering to come up with anything more elegant. It would be nice if someone with access to this kind of device could verify before merge.

etrepum (approved):
Would be nice if someone with access to an Android + Chrome device could take a look at this before merge to confirm the bug & fix

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2905,6 +2906,18 @@ export function updateDOMSelection(
       nextFocusNode,
       nextFocusOffset,
     );
+    // When deleting text across multiple paragraphs, Chrome on Android shifts selection rightwards
+    // This is a workaround to restore the correct selection
+    if (IS_ANDROID_CHROME) {
+      setTimeout(() => {
```
potatowagon, line 2912: is the reason to wrap the workaround in setTimeout with no delay param to:

> remove the function from the execution queue and it will only be invoked after JavaScript has finished with the current execution queue

https://stackoverflow.com/questions/3580068/is-settimeout-with-no-delay-the-same-as-executing-the-function-instantly

potatowagon:
https://github.com/user-attachments/assets/c64fab3e-7d76-429a-b062-b82d0271ca8f

im unable to repro the issue on my android, could it be a keyboard setting? not sure if i need to enable maybe Right to Left language in my android chrome browser?

packages/lexical/src/LexicalSelection.ts
etrepum: The reason is that few people understand how lexical event handling works so they add hacks like this. It’s certainly not strictly necessary to use setTimeout here, but I don’t have access to android to trace exactly when/why this code needs to run. 

Usually the root cause for hacks like this is that an event comes in from the browser and then lexical handles that event without preventing default but also needs to do something after the platform handles the default. Lexical doesn’t have good/obvious primitives for queuing that up so people write code like this which “works” but has subtle problems of its own.

potatowagon:
testing from https://lexical-playground-git-fork-waynetee-fix-android-fbopensource.vercel.app/

https://github.com/user-attachments/assets/1fcafe4f-dcb0-4217-99f2-c98f21e8823e

i dont see anything breaking so this should be safe to merge

packages/lexical/src/LexicalSelection.ts
etrepum: The update cascade happens because lexical’s reconciliation happens at the next microtask which will be before the setTimeout callback runs, so a second reconciliation will happen because of this workaround. If the code was queued up in precisely the right way only one lexical reconciliation would occur.

potatowagon:
testing changes from https://lexical-playground-my5rjurxj-fbopensource.vercel.app/

https://github.com/user-attachments/assets/8815c00c-03c8-480b-ae62-7f058b0e2c12

i dont see any issues so should be safe to merge, tho i havnt yet been able to repro the bug

packages/lexical/src/LexicalSelection.ts
waynetee: @potatowagon Yep you are absolutely right.

Here's the current flow:
1. User presses backspace
2. `beforeinput` event is triggered, handling the deletion
3. Reconciliation happens
4. We call `domSelection.setBaseAndExtent()` , shifting the caret to the correct position
5. The browser emits `selectionchange`, incorrectly shifting the caret

Using `setTimeout` allows us to restore the selection after `selectionchange` has occurred.

@etrepum If you are against the use of `setTimeout` you are welcome to suggest a better fix. Perhaps you can try augmenting the `selectionchange` handler to detect this change and revert it there, but I personally don't think that would work better. If its performance you are worried about, I don't think it makes a big difference given that backspacing normally within the same line already requires two reconciliations per character, based on testing on my Pixel.

If you need access to an Android device, I suggest https://www.browserstack.com/

waynetee:
> im unable to repro the issue on my android, could it be a keyboard setting? not sure if i need to enable maybe Right to Left language in my android chrome browser?

Looks like you are using the Samsung keyboard, which doesn't seem to have this issue. Try using GBoard? It's the default browser on my Pixel device.

potatowagon:
> you are using the Samsung keyboard, which doesn't seem to have this issue. Try using GBoard? It's the default browser on my Pixel device.

thanks for this info. i have access to a google pixel too so let me try it there

potatowagon:
> > you are using the Samsung keyboard, which doesn't seem to have this issue. Try using GBoard? It's the default browser on my Pixel device.
> 
> thanks for this info. i have access to a google pixel too so let me try it there


https://github.com/user-attachments/assets/574931c5-3aa2-4108-a15a-a4c89bedb7f3

still unable to repro the bug on GBoard on a google pixel. Gboard About shown towards end of screen capture, has version info. could it be a particular keyboard setting? like language?

tested the changes in this PR on the google pixel and no issues either

waynetee:
The bug occurs when deleting across lines:

https://github.com/user-attachments/assets/e59d7e52-fd2a-4369-8665-712ef3ddb09f

etrepum (approved):
Using a "Real" Google Pixel device on a Sauce Labs trial I was able to reproduce the bug and confirm that this PR fixes it. The issue with `setTimeout` is less about performance and more about how complicated things get when these events are not happening in the order you expect them to which often leads to different bugs. I think this is fine for now, it clearly fixes an issue and probably doesn't yet cause other ones, but should be cleaned up in a more systematic way that doesn't require the setTimeout later. I don't have a specific suggestion for exactly what the best fix is without doing a much deeper dive into this part of the code while having a live android device accessible for debugging.

ThenTech:
I just tested this change on my current setup, where I have `DecoratorNode`s that render emojis (mentioned in #5042), but the issue still persists on mobile, specifically with the GBoard keyboard. When pressing backspace, the cursor jumps to the front of the node and then back behind the node, but the node itself does not get deleted.

etrepum:
Please file a new detailed issue with a full repro if you’d like anyone to look at this bug, comments on closed PRs will get lost.

packages/lexical/src/LexicalSelection.ts
etrepum: Turns out that this indeed did cause problems, I am guessing because it is over-writing some other changes to the selection that are expected to happen due to a different update or event. For now it will be reverted (#7218) because it causes worse problems in certain situations.

waynetee: You are right. Turns out that subsequent browser changes to selection are sometimes desired. Apologies to all those affected by the bugs introduced.

PR #7124, opened 2025-02-04
[workflow] disable build_script: build-prod in .github/workflows/size-limit.yml

potatowagon:
attempt to mitigate S487385

disabling build_script: build-prod will use the default build as per https://github.com/andresz1/size-limit-action/tree/v1/

incase this is insufficient, disabling the whole job https://github.com/facebook/lexical/pull/7125 would work to mitigate the sev


potatowagon:
no longer needed

PR #7125, opened 2025-02-04
[workflow] disable size-limit job

potatowagon:
mitigate S487385

risk of merging this: might not be able to catch size regressions on github. However on www, we can still catch size regressions on the sync diff triggered daily with unigraph signals

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -5,6 +5,8 @@ on:
       - main
 jobs:
   size-limit:
+    # remove below line to enable this job
+    if: false
     runs-on: ubuntu-latest
     env:
       CI_JOB_NUMBER: 1


etrepum (approved):
Removing the job makes more sense than making it do something useless. I don’t have any insight into what the sev is but if it’s privesc on GitHub actions there’s ways to fix that but hard for anyone outside of meta to do it because nobody else has the permissions

etrepum:
For more context the size-limit action was broken for months, it always reported no change because it was building main twice instead of the PR. Might as well remove it altogether if nobody at meta is interested in fixing it in a secure fashion

AaronDewes:
> Removing the job makes more sense than making it do something useless. I don’t have any insight into what the sev is but if it’s privesc on GitHub actions there’s ways to fix that but hard for anyone outside of meta to do it because nobody else has the permissions

Hi  :wave:

I'm the person who reported this issue originally.

A possible fix would be to just add:

```YAML
permissions:
  pull-requests: write
  contents: read
```

at the beginning of the workflow YAML file. No one at Meta would have to be involved for that with changing repo settings etc.

I don't want to get involved in your development workflow, so do whatever you prefer, but if you want to restore that action, that would be a very easy fix.

etrepum:
The size-limit action just isn’t safe to use with pull_request_target, the changes required would be more extensive than just that

AaronDewes:
Not really in my opinion.

Executing unknown code is not dangerous if it can't do anything except commenting on a PR.

PR #7127, opened 2025-02-04
[lexical-playground] Fix: Replace icon with Twitter with X

awjae:
## Description
Replace icon Twitter with X in playground embed menu 

### Before
image

### After
image


potatowagon (dismissed):
lgtm, thanks!

packages/lexical-playground/src/index.css
```diff
@@ -489,6 +489,10 @@ i.tweet {
   background-image: url(images/icons/tweet.svg);
 }
```
potatowagon, line 490: since this isnt use can look into cleaning it up, as well as the twitter svg. not merge blocking for this PR

awjae: I've added another commit based on your suggestions. Thanks! 😊

potatowagon (approved):

PR #7128, opened 2025-02-04
[workflow] Delete size limit report

potatowagon:
deleting .github/workflows/size-limit.yml to cleanup S487385 . it is only the reporter for size limit results, and has been broken for awhile

I kept the [size limit package](https://www.npmjs.com/package/size-limit) around since im guessing the size limit cli tool could help debug size regressions?  - lmk if this makes sense or should i uninstall size limit completely


etrepum (approved):
It would probably be better to eliminate it altogether. It doesn't really get looked at, it didn't even work for most of the time it was there (not being able to comment on anything from forks, only measuring dev bundles, and/or measuring main vs. main), and nobody has paid much attention to what it is actually measuring. When/if someone at Meta wants to invest in this project they can invest the effort to do it correctly, probably with a different tool that can be used in a secure fashion. The reason it was done this way was half to work around how the size-limit action works, and the other half was because people outside of Meta don't really have credentials to implement the secure portion of it (commenting based on the results of the measurement)

etrepum (approved):

PR #7047, opened 2025-01-14
[Breaking Change][*] Chore: Use terser for optimizing cjs prod build

etrepum:
## Breaking Change

This may cause some regression in the cjs build size or cause some other configuration related issue, particularly for www\. I suspect that most people in OSS are using esm builds at this point, but some extra scrutiny should be used at Meta when evaluating this PR.

## Description

The cjs prod build has used @ampproject/rollup-plugin-closure-compiler for some time, but that project is abandoned (last release Aug 2021) and causes hard to debug issues whenever you end up using syntax that is valid JS according to the rest of the toolchain but not supported by that abandoned project (e.g. property initializers in classes).

terser is already used for the lexical esm build which has been in production for some time now, it is currently the most popular tool for this purpose, and it is maintained on top of a current toolchain.

To investigate the difference between the two I audited our size-limit action since I knew there was an outstanding issue related to it (#6852). Turns out that issue is correct and the size-limit action hasn't done anything useful since we moved it to a pull_request_target trigger. Our size-limit action also only measured dev builds which is fast but not useful, so I changed it to prod. Unfortunately for (good) security reasons you can't really test these in a PR since the workflow code is guaranteed to come from main, so I had a few commits here testing the action with the pull_request trigger. We might still want to consider having a more abstract method for measuring size, e.g. publishing an artifact that a safer workflow trigger can measure.

Closes #6852

### Before

Get an acorn stack trace in prod builds that only includes the line number in some build artifact and not any real indication of where the error is (fortunately you can attach a debugger locally, but it's a pain) whenever modern JS syntax is used. Most recently I ran into this working on #7046 but this isn't the first time.

### After

Syntax support is consistent across all builds, dev and prod.

PR #7130, opened 2025-02-04
install babel plugin-transform-runtime

potatowagon:
T214354208 Getting regeneratorRuntime is not defined errors in lexical clipboard . These errors dont seem to cause any noticeble crashes or behavior when i tried to repro, tho i can repro the error in the logs.

Research:

About the error and possible fix:
[https://stackoverflow.com/questions/61755999/uncaught-referenceerror-regeneratorruntime-is-not-defined-in-react](https://l.facebook.com/l.php?u=https%3A%2F%2Fstackoverflow.com%2Fquestions%2F61755999%2Funcaught-referenceerror-regeneratorruntime-is-not-defined-in-react&h=AT3IyeCaEj6wIHvuEbY12AXfWvlTIQ5HG-wefFJYd0y3ZOV2mLwdgfGSdqutfdVQQupU2G4NRjlNE8v1h6PCePySDOzDPxeASddFQZPTAAZ8q7R8BDV7gfeHwsYH0f9I4F7fGKOo)

How async functions are related to the error:
[https://stackoverflow.com/questions/65378542/what-is-regenerator-runtime-npm-package-used-for](https://l.facebook.com/l.php?u=https%3A%2F%2Fstackoverflow.com%2Fquestions%2F65378542%2Fwhat-is-regenerator-runtime-npm-package-used-for&h=AT3IyeCaEj6wIHvuEbY12AXfWvlTIQ5HG-wefFJYd0y3ZOV2mLwdgfGSdqutfdVQQupU2G4NRjlNE8v1h6PCePySDOzDPxeASddFQZPTAAZ8q7R8BDV7gfeHwsYH0f9I4F7fGKOo)

fix steps: [https://babeljs.io/docs/babel-plugin-transform-runtime](https://l.facebook.com/l.php?u=https%3A%2F%2Fbabeljs.io%2Fdocs%2Fbabel-plugin-transform-runtime&h=AT3IyeCaEj6wIHvuEbY12AXfWvlTIQ5HG-wefFJYd0y3ZOV2mLwdgfGSdqutfdVQQupU2G4NRjlNE8v1h6PCePySDOzDPxeASddFQZPTAAZ8q7R8BDV7gfeHwsYH0f9I4F7fGKOo)

root cause:

- error coming from here: [https://github.com/facebook/lexical/blob/cleanup-size-limit/packages/lexical-clipboard/src/clipboard.ts#L408](https://l.facebook.com/l.php?u=https%3A%2F%2Fgithub.com%2Ffacebook%2Flexical%2Fblob%2Fcleanup-size-limit%2Fpackages%2Flexical-clipboard%2Fsrc%2Fclipboard.ts%23L408&h=AT3IyeCaEj6wIHvuEbY12AXfWvlTIQ5HG-wefFJYd0y3ZOV2mLwdgfGSdqutfdVQQupU2G4NRjlNE8v1h6PCePySDOzDPxeASddFQZPTAAZ8q7R8BDV7gfeHwsYH0f9I4F7fGKOo)

- thrown because of babel transpiler compatibility issues with export async function
could be introduced in [https://github.com/facebook/lexical/pull/7047/](https://l.facebook.com/l.php?u=https%3A%2F%2Fgithub.com%2Ffacebook%2Flexical%2Fpull%2F7047%2F&h=AT3IyeCaEj6wIHvuEbY12AXfWvlTIQ5HG

- unsure, may or may not be related to the build bundle change introduced in https://github.com/facebook/lexical/pull/7047/ because errors started to appear after syncing the most recent sync diff that contains PR7047 over to www

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "@babel/core": "^7.24.5",
     "@babel/eslint-parser": "^7.24.5",
     "@babel/plugin-transform-optional-catch-binding": "^7.24.1",
+    "@babel/plugin-transform-runtime": "^7.25.9",
     "@babel/preset-flow": "^7.24.1",
     "@babel/preset-react": "^7.24.1",
     "@babel/preset-typescript": "^7.24.1",
--- a/babel.config.js
+++ b/babel.config.js
@@ -14,6 +14,9 @@ module.exports = {
       require('./scripts/error-codes/transform-error-messages'),
       {noMinify: true},
     ],
+    // When this plugin is enabled, the useBuiltIns option in @babel/preset-env must not be set. Otherwise, this plugin may not able to completely sandbox the environment.
+    // https://babeljs.io/docs/babel-plugin-transform-runtime
+    ['@babel/plugin-transform-runtime'],
   ],
   presets: [
     [
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@babel/core": "^7.24.5",
         "@babel/eslint-parser": "^7.24.5",
         "@babel/plugin-transform-optional-catch-binding": "^7.24.1",
+        "@babel/plugin-transform-runtime": "^7.25.9",
         "@babel/preset-flow": "^7.24.1",
         "@babel/preset-react": "^7.24.1",
         "@babel/preset-typescript": "^7.24.1",


potatowagon:
draft because im waiting for npm install @babel/runtime to complete

package.json
```diff
@@ -112,6 +112,7 @@
     "@babel/core": "^7.24.5",
     "@babel/eslint-parser": "^7.24.5",
     "@babel/plugin-transform-optional-catch-binding": "^7.24.1",
+    "@babel/plugin-transform-runtime": "^7.25.9",
```
potatowagon, line 115: draft because im waiting for npm install @babel/runtime to complete

but its taking very long - aparently its a very large package and supposedly to go into prod dependencies, is this even a good idea? im afraid to bloat the install

potatowagon: install takes too long and hangs the editor each time it completes installing, im gonna try skipping installing @babel/runtime

given the issue only happens for prod builds, im moving  "@babel/plugin-transform-runtime" to prod dependency

im semi clueless about what im doing so happy to recieve any advice

etrepum:
This looks like a configuration issue outside of OSS lexical, whatever is consuming the LexicalClipboard.js is creating this problem. There's no reference to babel or regeneratorRuntime in any of the artifacts produced by OSS lexical. The referenced function in the output JS is the following (after prettier)

```ts
exports.copyToClipboard = async function (e, t, r) {
  if (null !== p) return !1;
  if (null !== t)
    return new Promise((n, o) => {
      e.update(() => {
        n(f(e, t, r));
      });
    });
  const i = e.getRootElement(),
    l = e._window || window,
    s = window.document,
    a = o.getDOMSelection(l);
  if (null === i || null === a) return !1;
  const c = s.createElement("span");
  (c.style.cssText = "position: fixed; top: -1000px;"),
    c.append(s.createTextNode("#")),
    i.append(c);
  const d = new Range();
  return (
    d.setStart(c, 0),
    d.setEnd(c, 1),
    a.removeAllRanges(),
    a.addRange(d),
    new Promise((t, i) => {
      const l = e.registerCommand(
        o.COPY_COMMAND,
        (o) => (
          n.objectKlassEquals(o, ClipboardEvent) &&
            (l(),
            null !== p && (window.clearTimeout(p), (p = null)),
            t(f(e, o, r))),
          !0
        ),
        o.COMMAND_PRIORITY_CRITICAL,
      );
      (p = window.setTimeout(() => {
        l(), (p = null), t(!1);
      }, 50)),
        s.execCommand("copy"),
        c.remove();
    })
  );
};
```

I'm not sure what exactly the closure compiler was doing to avoid triggering this configuration issue because it was very similar (from 0.23.1)

```ts
exports.copyToClipboard = async function (a, b, c) {
  if (null !== C) return !1;
  if (null !== b)
    return new Promise((h) => {
      a.update(() => {
        h(D(a, b, c));
      });
    });
  var d = a.getRootElement();
  let f = null == a._window ? window.document : a._window.document,
    k = p.getDOMSelection(a._window);
  if (null === d || null === k) return !1;
  let g = f.createElement("span");
  g.style.cssText = "position: fixed; top: -1000px;";
  g.append(f.createTextNode("#"));
  d.append(g);
  d = new Range();
  d.setStart(g, 0);
  d.setEnd(g, 1);
  k.removeAllRanges();
  k.addRange(d);
  return new Promise((h) => {
    let q = a.registerCommand(
      p.COPY_COMMAND,
      (l) => {
        n.objectKlassEquals(l, ClipboardEvent) &&
          (q(),
          null !== C && (window.clearTimeout(C), (C = null)),
          h(D(a, l, c)));
        return !0;
      },
      p.COMMAND_PRIORITY_CRITICAL,
    );
    C = window.setTimeout(() => {
      q();
      C = null;
      h(!1);
    }, 50);
    f.execCommand("copy");
    g.remove();
  });
};
```

potatowagon:
> Without any visibility into what the build toolchain looks like outside of OSS lexical I'm not sure I can provide any more insight other than to confirm that lexical itself isn't really the root cause of what's happening. Async functions have been supported in all browsers updated since ~2018 so there is no good reason for anyone to have a configuration in 2025 that would transpile them to generators with regeneratorRuntime.

yes u are right, i found out that haste does a validation of whether non es3 features are in use. the error could be a result of that. could explain why theres only the error but copy paste still works, since the browser supports es6. there is an annotation to turn of that non es3 vallidation: @noEnforceES3

ill be adding that in another PR

potatowagon:
closing in favor of - https://github.com/facebook/lexical/pull/7162

PR #7131, opened 2025-02-04
[ci] Chore: Switch to new SSH key

acywatson:
## Description

Rotates to a new SSH Key in Github Actions

## Test plan

Evaluate nightly release run. Or do a dry run, if possible?

### DIFF
--- a/.github/workflows/call-increment-version.yml
+++ b/.github/workflows/call-increment-version.yml
@@ -15,7 +15,7 @@ on:
         required: true
         type: string
     secrets:
-      SSH_KEY:
+      SSH_KEY_2:
         required: true
     outputs:
       version:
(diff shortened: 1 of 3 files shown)


zurfyx (approved):

etrepum (approved):

etrepum:
Should be able to test this by running the Create New Release Branch workflow, which only depends on this SSH_KEY_2 secret and not the NPM_TOKEN secret

etrepum:
Looks like it fails actually, see https://github.com/facebook/lexical/actions/runs/13142675139

You can run the workflow with this branch as a target

ivailop7 (dismissed):

zurfyx (approved):

potatowagon:
do we still need this?

PR #7134, opened 2025-02-04
[lexical-table] Support first column freeze

ivailop7:
Part one of freeze support in tables.
Have done the easier part, freezing a column

https://github.com/user-attachments/assets/dd1b3d80-ad21-44a6-a117-c8d1fb2cc47d

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -301,6 +301,7 @@ function getComment() {
     ' * @preserve-invariant-messages',
     ' * @preserve-whitespace',
     ' * @preventMunge',
+    ' * @noEnforceES3',
   );
   return lines.join('\n');
 }


packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -48,6 +48,7 @@ export type SerializedTableNode = Spread<
   {
     colWidths?: readonly number[];
     rowStriping?: boolean;
+    frozenFirstColumn?: boolean;
```
etrepum, line 51: For what it's worth, the Google Sheets API uses a frozenRowCount and frozenColumnCount. Freezing the first n rows or columns is probably not much harder than freezing one? Even if we only support 1 right now we could still future-proof the JSON by using a number.

https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/sheets

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -88,6 +89,20 @@ function setRowStriping(
   }
 }
 
+function setFirstColumnFreeze(
+  dom: HTMLElement,
+  config: EditorConfig,
+  forzenFirstColumn: boolean,
+): void {
+  if (forzenFirstColumn) {
```
etrepum, line 97:
```suggestion
  frozenFirstColumn: boolean,
): void {
  if (frozenFirstColumn) {
```

etrepum (commented):
Seems to work well! I think a number is a bit more future-proof than a boolean (the accessor could be a boolean for now, mostly just thinking about the JSON). I did notice a typo but otherwise this looks good. I only tested in Chrome though, Safari and Firefox might have CSS quirks with this sort of thing?

packages/lexical-table/src/LexicalTableNode.ts
ivailop7: This is a fantastic suggestion. I'll do it this way.

ivailop7:
@etrepum this one is good to go

etrepum (dismissed):
This looks like it's missing the `exportJSON` for frozenColumns (and it doesn't have any tests that exercise serialization). Could be fixed as a follow-up, can easily repro in the editor by copying and pasting a table that has a frozen column

ivailop7:
Oh, my oversight. I'll add the export JSON and add an extra test.

ivailop7:
@etrepum added the exportJSON part and state test. It should be good to go now.

ivailop7:
Can I get an approval please. @zurfyx @etrepum

PR #7135, opened 2025-02-04
[lexical][lexical-code] Refactor: Use NodeCaret to implement RangeSelection.getNodes()

etrepum:
## Description

Refactor RangeSelection.getNodes() and lexical-code getEndOfCodeInLine / getStartOfCodeInLine to use the new NodeCaret API internally.

Adds new APIs:
* $comparePointCaretNext (finally we can differentiate between a point being outside and inside an element at its border)
* $getCommonAncestor (determine if two nodes are the same, an ancestor/descendant of each other, or on separate branches. In the branch case it returns the common ancestor and the two siblings corresponding to the given points for efficient isBefore/isAfter)
* $getCommonAncestorResultBranchOrder (the computation to determine which sibling comes first isn't computed eagerly, you do that here)

Deprecations in lexical-code due to $ rule violations:
* getEndOfCodeInLine -> $getEndOfCodeInLine
* getStartOfCodeInLine -> $getEndOfCodeInLine

(commit history is long as it was stacked on #7046 and #7132 and not rebased)

Closes #6310 - it's not really possible to change the existing APIs to return coherent results, but now we have new APIs that can be used instead

## Test Plan

All existing unit and e2e tests pass. No expected behavior changes. Development of the getNodes() refactor was backed by an invariant that threw any time the previous and new implementations returned a different result, but that code has been removed after the whole suite was verified.


ivailop7 (approved):
LGTM. Approving to unblock, because I see it's been 3 weeks, ideally @fantactuka and @zurfyx  would be nice to give this a read too.

etrepum:
To be fair it wasn't marked for review until 5 days ago

zurfyx (approved):

PR #7152, opened 2025-02-08
[lexical] Chore: Add more tests for RangeSelection.getNodes()

etrepum:
## Description

As a follow-up to #7046 it makes sense to do more thorough unit testing so that the existing behavior of methods that are targeted for refactoring have a better specification to avoid compatibility issues. The goal is to refactor getNodes and related methods for compatibility.

## Test plan

This only adds more tests for the current behavior.


potatowagon (approved):
thankyou!

PR #7155, opened 2025-02-09
[Breaking Change][lexical][lexical-playground] Bug Fix: Improve character deletion around shadow roots and decorators

### LINKED ISSUE #5686: Bug: Unexpected behavior with transform selected lines to code block
If you select several lines and convert them into code block, the order of the nodes is disrupted, and extra nodes that were not selected end up in the code block.

Lexical version: 0.13.1

## Steps To Reproduce

1. Enter multiple lines
2. Select multiple lines between first and last lines
3. Try convert to code block

## The current behavior
The order of the nodes is disrupted, and extra nodes end up in the code block.

## The expected behavior
The order of the nodes is not violated, only the necessary nodes are included in the code block, the code block is inserted in the correct place

## Screenshots of unexpected behavior
Снимок экрана 2024-03-06 в 10 19 40
Снимок экрана 2024-03-06 в 10 19 47
Снимок экрана 2024-03-06 в 10 20 03

### LINKED ISSUE #6763: Bug: Pasting over content starting with an empty node appends instead of replace
Pasting over content that starts with an empty node leads to appending instead of replacement.

https://github.com/user-attachments/assets/9ed8bd65-6a1e-4312-979f-f5635a7691a8

Lexical version: 0.18.1+dev.esm

## Steps To Reproduce

1. go to playground. Press 'Enter'.
2. Type any text or insert any block in the 2nd line
3. Select all, then press paste

## The current behavior

Appends to the selected content instead of replacing it on paste

## The expected behavior

The text that is selected should replaced not append to.

## Impact of fix

Pasting content is pretty fundamental to a text editor. This behaviour can lead to weird behaviours for all users.

### LINKED ISSUE #7082: Bug: Block Type Change to Normal Fails for Selected Table Cells in Playground
Lexical version:  Latest (playground)

## Steps To Reproduce

1. Insert a TableNode
2. Select multiple cells
3. Change the block type to Heading, etc. (works correctly)
4. Change the block type to Normal (does not work)

https://github.com/user-attachments/assets/89fb0e78-fc00-427c-a66f-18faf1ca3103

## Link to code example

-  Playground

## The current behavior

-  Attempting to change the block type to Normal does not work.

## The expected behavior

-  The block type should change to Normal.

## Impact of fix

-  I don't think it has a major impact.

### LINKED ISSUE #7090: Bug: Backspace after a table causes unexpected deletion
Lexical version: v0.23.1

## Steps To Reproduce

In the playground:

1. Create a table (any size)
2. Move cursor to the paragraph after the table
3. Type some text
4. Move cursor to start of line
5. Press backspace
6. Observe that the line of text is forward deleted, except for the last character

https://github.com/user-attachments/assets/a8d3993e-6eb9-49b4-8fea-1e39b3284c6e

## The current behavior
A line of text is unexpectedly deleted in front of the cursor

## The expected behavior
Google Docs and Word simply no-op. Nothing is deleted and the cursor does not move.

## Impact of fix
Low, but unexpected deletions erode trust in the editor.

### LINKED ISSUE #7173: Bug: Pasting several paragraph blocks over a selection preceded by an empty paragraph doesn't overwrite the entire selection
Lexical version:

## Steps To Reproduce

1. Open Lexical playground
2. Delete all of the pre-filled text 
3. Press enter to leave an empty paragraph at the top
4. Type in some text on the next line
5. Elsewhere in the document, make two paragraph nodes and select/copy them to the clipboard -- you can delete them after, it doesn't matter
6. Highlight/select the top empty paragraph and the next paragraph (from steps 3/4)
7. Paste
8. Observe that the empty paragraph is inserted into and the rest of the selection is not deleted

Link to code example: This happens on the Lexical playground as is.

Can't seem to paste a video here, will try to add as a comment.

## The current behavior

The inserted text is just inserted in the position of the top-level paragraph, and all of the selected text remains in the editor.

## The expected behavior

The inserted text overwrites all of the selected text, leaving only what was pasted.

etrepum:
## Breaking Change

The heuristic used in `RangeSelection.deleteCharacter` to handle merging blocks and deleting decorators from a collapsed selection has been refactored for consistency. It now traverses the node tree directly instead of attempting to use browser selection APIs.

This generally has much more predictable results, aligned with how this method is expected to behave, but it is not "bug for bug" compatible with how it worked previously. Notably, it will no longer pierce shadow roots when merging blocks. For example, performing a forward deletion from the end of a block before a `CollapsibleContainerNode` would pierce the shadow root of its `CollapsibleTitleNode` and merge its first block child (e.g. the preceding paragraph and the title's paragraph would be merged). Since there was a test that expected this strange behavior, this functionality was moved to the `CollapsiblePlugin`'s `DELETE_CHARACTER_COMMAND` command listener.

## Description

Allowing the native platform to control deletion only really makes sense when the content to be deleted includes a TextNode, otherwise we are better off handling it ourselves. There's also some "interesting" edge cases around nested elements (with or without shadow roots) and decorators that need to be addressed by this sort of manual navigation.

Other fixes:
* `$setBlocksType` workaround for code nodes (playground only) #5686
* `$setBlocksType` support for selection types other than RangeSelection #7082
* Fixed the editor empty scenario when removing a TableNode that was RootNode's only child
* Added `selection.removeText()` to `RangeSelection.insertNodes(…)`

Closes #7082
Closes #7090
Closes #5686
Closes #7173
Closes #6763 
Closes #7184 

### Before

* Using character deletion from inside of a TextNode at the beginning of a block, when that block has a TableNode or block decorator previous sibling, would affect the TextNode (#7090)
* Removing a TableNode that was the only child of the RootNode would leave the RootNode with no children
* Various issues with $setBlocksType (see below)
* RangeSelection.insertNodes (used by paste) could leave existing content under certain conditions

### After

* All existing unit and e2e tests pass unchanged
* Removing a TableNode that was the only child of the RootNode now creates a ParagraphNode in its place
* Pressing delete at the block after a TableNode or DecoratorNode behaves predictably
* $setBlocksType works for table selections
* $setBlocksType works for normal
* $setBlocksType for code nodes in the playground works for partially selected blocks
* insertNodes removes the select text before adding anything


packages/lexical-playground/src/plugins/CollapsiblePlugin/index.ts
```diff
@@ -173,13 +176,67 @@ export default function CollapsiblePlugin(): null {
...
-          if (
-            !$isRangeSelection(selection) ||
-            !selection.isCollapsed() ||
-            selection.anchor.offset !== 0
-          ) {
+          if (!$isRangeSelection(selection) || !selection.isCollapsed()) {
+            return false;
+          }
+          if (!isBackward) {
+            // inherit strange forward delete behavior per e2e election test
```
ivailop7, line 185: If you want to change the behaviour that forward deletion should not delete the collapsible section, I'm with you.

etrepum: I don't have strong opinions about it, it's in a playground plug-in after all. I figured that breaking zero tests would maybe speed up review but it's been a week 🤷

packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
```diff
@@ -216,19 +218,54 @@ export const formatQuote = (editor: LexicalEditor, blockType: string) => {
...
+              // Handle case for both preceding and trailing content
+              sel.insertParagraph();
+            }
+          }
+        }
+        if (target) {
+          if ($isLineBreakNode(possibleLineBreak)) {
+            possibleLineBreak.remove();
           }
+          target.select().insertRawText(textContent);
```
zurfyx, line 267: How come this toolbar logic has now become so manual? Is this the final shape? It used to be pretty straightforward to replicate once it was backed by `insertNodes`

zurfyx (approved):

packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
etrepum: It's because insertNodes is buggy and I didn't want to also fix that in the scope of this PR

https://github.com/facebook/lexical/pull/7155/files#diff-c0aa95ba19e1cabb9a918e474930a02f54f63df8f1cd43f2eb96354b681e8071R240-R242

packages/lexical-playground/src/plugins/CollapsiblePlugin/index.ts
etrepum: I created a follow-up #7203 to address this separately

PR #7159, opened 2025-02-10
[lexical-website] Documentation Update: CaretType has sibling and child.

hastebrot:
## Description

Fixes small typo in "Node Traversals with NodeCaret".

The type is defined as `export type CaretType = 'sibling' | 'child';`.


etrepum (approved):
Good catch!

PR #7169, opened 2025-02-12
Update WWW build flag

zurfyx:
Turns out we were still transpiling `async` internally

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -294,6 +294,7 @@ function getComment() {
     0,
     ' *',
     ' * @fullSyntaxTransform',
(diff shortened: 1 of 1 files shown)


scripts/build.js
```diff
@@ -294,14 +294,14 @@ function getComment() {
     0,
     ' *',
     ' * @fullSyntaxTransform',
+    ' * @es6-async_DO_NOT_USE',
```
potatowagon, line 297: is there any documentation why this is do not use?

potatowagon (approved):

PR #7174, opened 2025-02-13
[fix] $dfs traversal from the middle of the tree leaves

fantactuka:
Running $dfs from non-first child node would reset caret and traverse from the first child:

```
    P1
T1  T2 T3

$dfs(T2) -> [T1, T2, T3]

while expected

[T2, T3]
```

@etrepum wondering if you recall why that $rewind() was added


zurfyx (approved):
Stamp to unblock but would definitely benefit from @etrepum's context here

etrepum (approved):
LGTM, I don't recall off the top of my head, looks like a mistake on my part. My best guess is that it was a workaround or something that didn't get changed when I refactored the iterator code a few times during the course of that PR. Either way we're much better off having a test that specifies the expected behavior!

PR #7175, opened 2025-02-13
[lexical] Bug Fix: Workaround for delete character with emoji grapheme customers that do not include non-BMP code points

### LINKED ISSUE #7163: Bug: ❤️ emoji cannot be deleted with a single keystroke
Lexical version: 0.24 (latest)

## Steps To Reproduce

1. Type ❤️ (https://emojipedia.org/red-heart#emoji)
2. Hit Backspace to  delete it

Link to code example:

https://playground.lexical.dev/#doc=H4sIAAAAAAAAE91TQU4DMQz8iuVzVLWFUsgDOCKkHhGHdONuI7LxKvG2RNW-gRdw4xG8hw_wBZQtqMAJIQ4Vp8QT2TPjUXZI1gnHhRgh1DuMzFLOau28jRRQ33wrLIlxHvVY4YpjY2S4NmwJNYaCeFSYJPsCoEKhe0GN8PL49Pr8UIDclqcBV7ihmBwH1JP-VqF1kSoZavQS8UBSRrlgKewJ34e0Jpo6mnb9ZdKe9PIgr5SLD0m9-iNLV7wFiRmEwZInIXACywxdcqGGpanuUmsqUmCChcwdbJ33UJP8pyWUXH_uJ3TeH7UdF5KQscCrIUwqSVbctCVenyFSwxuyxxDgL3iG3_1FbK_QmyQLU0zpyfzkYjo7n0_ms7OpwsRdrErftTe5jtwF-7kbx6Pp6WiM_RtvNUShRgQAAA

## The current behavior

The red heart is converted to a black heart (due to the two unicode characters)

Image

## The expected behavior

The entire grapheme should be removed

## Impact of fix

A bunch of these cases have already been solved by https://github.com/facebook/lexical/issues/17 but it seems there are a few emojis that are still not convered by the fix. 

I tested in Chrome, Safari and both behave the same

etrepum:
## Description

Lexical's deleteCharacter implementation basically uses the native platform to extend the selection by one character, and then removes the text in-between. In most situations this works well, but in some situations the concept of a character for selection and deletion are not quite the same – in several languages you want to be able to delete parts of a character that is visually composed, but it doesn't make sense to have to press the arrow key multiple times to navigate around that character.

The previous workaround looked specifically for surrogate pairs, which covers many cases where we want caret navigation and delete to work the same way, but the known case where this differs is when an emoji is a grapheme cluster built from only code points in the BMP. The reported example was '❤️' aka `String.fromCharCode(0x2764, 0xFE0F)` where neither of those code units are in the surrogate range.

Here's a google doc link with some example characters that may cause edge cases: https://docs.google.com/document/d/1tQCTPhXYFG2dYW4GN5Yln8yDSaHjM7XFxiBtye6clX0/edit?tab=t.0

Closes #7163

Related, but not addressed here: #5260

### Before

* Deleting ❤️ requires two backspace presses, with an intermediate state (that for whatever reason displays the same way on github but is a distinct code point: ❤)
* The HashtagPlugin would consider #️⃣ to be a hashtag because it is a # followed by a combining character which was for some reason allowed at the start

### After

* Deleting ❤️ requires one backspace press
* HashtagPlugin does not consider the #️⃣ key cap to be a hashtag


packages/lexical/src/LexicalSelection.ts
```diff
@@ -2018,19 +2082,63 @@ function $updateCaretSelectionForUnicodeCharacter(
...
+    const re = new RegExp('\\p{Emoji}', 'u');
+    const test = re.test.bind(re);
+    // check a few emoji
+    if (
+      // Emoji in the BMP (heart) with variation selector
+      test('\u2764\ufe0f') &&
+      // Emoji in the BMP (#) with variation selector
+      test('#\ufe0f\u20e3') &&
+      // Emoji outside the BMP (thumbs up) that is encoded with a surrogate pair
+      test('\ud83d\udc4d')
```
zurfyx, line 2131: How do we know these 3 are the only exceptions that do not fall under the surrogates range?

zurfyx (approved):
:O

packages/lexical/src/LexicalSelection.ts
etrepum: Using one of them should be sufficient, it's really just testing that the runtime parsed the regex at all. It's really just being a bit overly cautious to make sure that it matches our expectations if we're going to use it.

zurfyx (dismissed):

etrepum:
@zurfyx sorry for dismissing again, didn't realize you were on top of it already. just figured out how to get a bit more coverage in firefox

zurfyx (approved):

PR #7181, opened 2025-02-15
Position menu immediately

zurfyx:
Credit to Cindy Zhang.

The currently menu position logic can interfere with other (third-party) components that attempt to position the component nicely within the screen. Immediate position makes it play better with these.

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -491,7 +491,6 @@ export function useMenuAnchorRef(
   const anchorElementRef = useRef<HTMLElement | null>(
     CAN_USE_DOM ? document.createElement('div') : null,
   );
-  const hasPositioned = useRef(false);
   const positionMenu = useCallback(() => {
     if (anchorElementRef.current === null || parent === undefined) {
       return;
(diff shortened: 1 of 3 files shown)


potatowagon (dismissed):

packages/lexical-react/src/shared/LexicalMenu.ts
```diff
@@ -603,6 +596,15 @@ export function useMenuAnchorRef(
...
 
+  // Append the context for the menu immediately
+  const containerDiv = anchorElementRef.current;
+  if (containerDiv != null) {
+    containerDiv.style.position = 'absolute';
+    if (parent != null) {
+      parent.append(containerDiv);
+    }
+  }
+
```
potatowagon, line 607: causing many e2e tests to break, https://github.com/facebook/lexical/actions/runs/13342836119/job/37269678541?pr=7181

from the test failure videos & logs it looks like the test infra is unable to find the "Copy" option

>waiting for locator('#typeahead-menu [role="option"] :text("Copy")') to be visible

im unable to repro that issue manually. think need to update the test selector

potatowagon: tested manually the context menu seems fine


Uploading Screen Recording 2025-02-17 at 10.15.57 AM.mov…

so, likely a test infra issue

potatowagon: tests fixed in https://github.com/facebook/lexical/pull/7181/commits/8d522144c929b14269ec94e0b5362e01cdd715ce

for some reason #typeahead-menu selector isnt working for the test, could be because changes to parent container re-assignments, anyway changing the selector to be a closer ancestor worked. 

ran `npm run test-e2e-chromium` looks good locally

potatowagon (approved):

PR #7182, opened 2025-02-16
[lexical-playground] fix: hard coded theme classes for table hover actions

umaranis:
## Description

The [TableHoverActionsPlugin](packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx) has following class names hard coded instead of being referenced from theme.

- PlaygroundEditorTheme__tableAddRows
- PlaygroundEditorTheme__tableAddColumns
- PlaygroundEditorTheme__tableCell

This PR fixes the hard coded names.

## Test plan

There is no change in behaviour due to the fix. It only enables the Table Hover Action Plugin to keep working if the Playground theme is modified.

### DIFF
--- a/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableHoverActionsPlugin/index.tsx
@@ -24,7 +24,12 @@ import {
   TableRowNode,
 } from '@lexical/table';
 import {$findMatchingParent, mergeRegister} from '@lexical/utils';
-import {$getNearestNodeFromDOMNode, isHTMLElement, NodeKey} from 'lexical';
+import {
+  $getNearestNodeFromDOMNode,
+  EditorThemeClasses,
+  isHTMLElement,
+  NodeKey,
+} from 'lexical';
 import {useEffect, useMemo, useRef, useState} from 'react';
 import * as React from 'react';
 import {createPortal} from 'react-dom';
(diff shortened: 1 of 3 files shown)


etrepum (commented):
There are a couple problems with this once it's generalized:
- Having a missing theme will cause problems
- Defining a theme property with multiple classes (e.g. tailwind utility class style) will result in an incorrect selector, because only the first will have a preceding `.`

I think maybe we should have an invariant to check that there is a theme, and each of these keys exists on the theme, otherwise you will get strange behavior and/or runtime errors when you perform css queries on `td.undefined` or `td.relative text-xl`. Maybe a helper function like:

```ts
getThemeSelector(getTheme: () => EditorThemeClasses | null | undefined, name: keyof EditorThemeClasses): string {
  const className = getTheme()?.[name];
  invariant(typeof className === 'string', 'getThemeClass: required theme property %s not defined', name);
  return className.split(/\s+/g).map((cls) => `.${cls}`).join();
}
```

and then change the queries appropriately.

A more robust approach would be to refactor tables a bit to not need this, e.g. getMouseInfo can probably be refactored to go inside the editor.getEditorState().read() and use `$getNearestNodeFromDOMNode` directly to find if that node is inside of a table.

etrepum:
> Thanks @etrepum . I have updated the code based on your suggestion.
> 
> > A more robust approach would be to refactor tables a bit to not need this, e.g. getMouseInfo can probably be refactored to go inside the editor.getEditorState().read() and use $getNearestNodeFromDOMNode directly to find if that node is inside of a table.
> 
> This probably won't work because the add row and column buttons are outside the table.

It would work for tableDOMNode but not the button selectors. You don't really need selectors for those though, they could be probably tracked with refs from the plugins

etrepum (approved):

PR #7184, opened 2025-02-16
fix(6763): copy and paste with newline at the beginning

### LINKED ISSUE #6763: Bug: Pasting over content starting with an empty node appends instead of replace

ozanilkumar23:
[Lexical] : Fix duplicate content on copy and paste with newline at the beginning

Bug Fix

## Description

- When content with a newline at the start is copied and pasted again, it creates duplicate content, instead of replacing it by appending the text.
- This PR will fix the duplicate text issue and it will replace the whole text.

This PR checks if the first block is empty and it has a newline char then it will call this.insertParagraph(). 

Closes #6763

### Before

There is no change in behavior due to the fix.

### After

There is no change in behavior due to the fix.


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1430,7 +1430,16 @@ export class RangeSelection implements BaseSelection {
       $isElementNode(firstBlock) &&
       (!firstBlock.isEmpty() || firstBlock.canMergeWhenEmpty());
 
-    const shouldInsert = !$isElementNode(firstBlock) || !firstBlock.isEmpty();
+      let isFirstBlockEmpty = (firstBlock as ElementNode)?.isEmpty();
```
etrepum, line 1433: The `INTERNAL_$isBlock` guard will also return true for a top-level decorator nodes, so casting here is not safe

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1430,7 +1430,16 @@ export class RangeSelection implements BaseSelection {
...
+      // if first block is empty check its has newline chars 
+      if(isFirstBlockEmpty) {
+        const firstBlockParentContent = firstBlock?.getParent()?.getTextContent()?.split('\n') || [];
```
etrepum, line 1436: This is a very inefficient approach to solving this problem, as this could generally produce the textContent of the entire document just to look for a LineBreakNode. I also generally don't think it works because newlines are produced by any block ElementNode here so this would also be triggered by an empty paragraph block before the firstBlock in this selection.

etrepum (commented):
This does not pass the style checks (`npm run ci-check`; the prettier formatting and the lint against using the `?.` operator), does not have any new tests to confirm that it is solving the issue, and probably fails other tests.

I did already fix this particular issue more systematically in #7155, which is waiting for review

PR #7185, opened 2025-02-16
[lexical-react] Bug Fix: Make typeahead menu respect read-only mode

### LINKED ISSUE #7160: Bug: Typeahead menu does not respect read-only mode
Lexical version: [Assumed 0.24.0 from package here](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/package.json)

The common theme appears to be the `LexicalTypeaheadMenuPlugin` component

## Steps To Reproduce

Emoji example:

1. Open Lexical Playground example at https://playground.lexical.dev/
2. Ensure there's a colon with a space before it
3. Set playground to read-only mode
4. Focus cursor after colon
5. Choose an emoji from the list

Same for `@-mentions`.

Link to code example:

https://github.com/user-attachments/assets/7273d1a3-1e7d-4fd8-8ef3-916570c407ac

https://github.com/user-attachments/assets/c65a6128-56f5-495f-9ce2-baca5e7f9a82

## The current behavior

Emoji/mentions picker opens and allows insertion/modification of nodes

## The expected behavior

No picker should be present, or at least no emoji/mention should be insertable

## Impact of fix

Ensure true compliance of "read only" mode

kirandash:
## Description
Currently, the typeahead menu (emoji/mentions picker) does not respect the editor's read-only mode. When the editor is in read-only mode:
1. The typeahead menu still opens when typing triggers (: or @)
2. Users can still select and insert emojis/mentions
3. The menu stays open even if editor switches to read-only mode while menu is open

This PR fixes the issue by:
1. Adding a read-only mode check before showing the typeahead menu
2. Adding an editable state listener to close the menu when editor becomes read-only
3. Ensuring complete compliance with read-only mode behavior

Closes #7160

### Before

https://github.com/user-attachments/assets/61473734-e4db-486c-985e-f9480d9aa4c7

### After

https://github.com/user-attachments/assets/2eb0e34f-66b9-44c3-87b2-8e129e89defa

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -242,6 +242,12 @@ export function LexicalTypeaheadMenuPlugin<TOption extends MenuOption>({
   useEffect(() => {
     const updateListener = () => {
       editor.getEditorState().read(() => {
+        // Check if editor is in read-only mode
+        if (!editor.isEditable()) {
+          closeTypeahead();
+          return;
+        }
+
         const editorWindow = editor._window || window;
         const range = editorWindow.document.createRange();
         const selection = $getSelection();
@@ -296,6 +302,22 @@ export function LexicalTypeaheadMenuPlugin<TOption extends MenuOption>({
     closeTypeahead,
     openTypeahead,
   ]);
+
+  // Add effect to listen for editable state changes
+  useEffect(() => {
+    const editableListener = (isEditable: boolean) => {
+      if (!isEditable) {
+        closeTypeahead();
+      }
+    };
+
+    const removeEditableListener =
+      editor.registerEditableListener(editableListener);
+
+    return () => {
+      removeEditableListener();
+    };
+  }, [editor, closeTypeahead]);
 
   return resolution === null ||
     editor === null ||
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -297,6 +303,22 @@ export function LexicalTypeaheadMenuPlugin<TOption extends MenuOption>({
...
+        closeTypeahead();
+      }
+    };
+
+    const removeEditableListener =
+      editor.registerEditableListener(editableListener);
+
+    return () => {
+      removeEditableListener();
+    };
```
etrepum, line 319: Just a nit for future reference, but this could be simpler:

```suggestion
    return editor.registerEditableListener(editableListener);
```

etrepum (approved):
Looks good! Confirmed the fix

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
kirandash: @etrepum Thanks for the nit suggestion about simplifying the editable listener cleanup! I'll keep this in mind for future PRs. Sorry about the force push - I should have left the original approved version. Would you mind re-approving when you get a chance?

etrepum: Well if you're here, might as well make that simplification and I'll re-approve and add it to the merge queue once the CI test run finishes :)

kirandash: @etrepum I've pushed the simplified editable listener change using the implicit return approach. The change removes the intermediate variables and explicit return statement, making the code more concise while maintaining the same functionality. Looking forward to your re-review. Thanks  a lot for the feedback! 👍

etrepum (approved):
LGTM!

PR #6518, opened 2024-08-13
[lexical][lexical-devtools-core][lexical-playground] Bug Fix: fix TextNode importDom highlight formatting 

bfritscher:
## Description

Pull request #3583 by @moy2010 added a new textformat (which is helpful), but if we use HTML to serialize it gets lost because importDOM is not parsing the mark or if we past HTML with `<mark>`, but the editor is already capable to export `<mark>` and set IS_HIGHLIGHT.

- Added the missing mark entry
- Added a highlight formatting menu entry to playground for manual testing (including icon, and CSS theme)
- added the missing entry in devtools to show format: highlight
- added a test to check for mark on html past

Closes #6498

### After
image

PR #7186, opened 2025-02-16
[lexical][lexical-devtools-core][lexical-playground] Bug Fix: fix TextNode importDom highlight formatting

### LINKED ISSUE #6498: Bug: highlight TextFormatType is missing in importDOM losing state
Pull request https://github.com/facebook/lexical/pull/3583 by @moy2010 added a new textformat (which is helpful), but if we use HTML to serialize it gets lost because importDOM is not parsing the mark tag  in
https://github.com/facebook/lexical/blob/20e4ea1687bd846653a0cc2df8cbb610d8308307/packages/lexical/src/nodes/LexicalTextNode.ts#L552

Also the devtools do not show formatting for this new highlight feature because its mising in
https://github.com/facebook/lexical/blob/20e4ea1687bd846653a0cc2df8cbb610d8308307/packages/lexical-devtools-core/src/generateContent.ts#L58

Lexical version: 0.16.1

## Steps To Reproduce

1. exportDOM a TextNode with format highlighting
2. importDOM mark does not get parsed back to highligted text.,

kirandash:
## Description
This PR resolves merge conflicts in PR #6518 by @bfritscher which originally fixed this issue.

Changes made to resolve conflicts:
1. Integrated highlight feature with the newer centralized toolbar state management system
2. Added `isHighlight` to the `INITIAL_TOOLBAR_STATE` in ToolbarContext
3. Updated toolbar state handling to use `updateToolbarState` instead of individual state variables
4. Maintained all original highlight functionality while adapting to current codebase structure

Closes #6498
Closes #6518

### After

image


etrepum (approved):
Looks great, love that it has a test, confirmed the fix

PR #7187, opened 2025-02-16
[lexical-code][lexical-playground] Bug Fix: Fix selection boundaries in code block

### LINKED ISSUE #6311: Bug: in Playground typing in markdown source box, actually jumps and chars start appear outside this source box
Caret inside markdown source box:
image
Typing actually starts not in the source box:
image

kirandash:
## Description

Current behavior:
- When editing in markdown source mode, cursor selection can escape the code block boundaries
- After pressing arrow keys (up/down), typing appears outside the markdown source box
- Text input appears in the root editor instead of staying within the code block

Changes being added:
- Ensure cursor stays within code block when using arrow keys
- Verify text input remains contained within the code block
- Add test cases to verify and enforce proper selection boundaries
- Add assertions for correct node selection paths and offsets
- Test both start and end boundary conditions of the code block

Closes #6311

### Before

https://github.com/user-attachments/assets/b1d9337d-40b4-4883-b6d5-3757c498200b

### After

https://github.com/user-attachments/assets/33069c7f-681f-487b-ad45-a02198fe5509

Fixes #6311

### DIFF
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -880,22 +880,64 @@ export function registerCodeHighlighting(
     ),
     editor.registerCommand(
       KEY_ARROW_UP_COMMAND,
-      (payload): boolean => $handleShiftLines(KEY_ARROW_UP_COMMAND, payload),
+      (event) => {
+        const selection = $getSelection();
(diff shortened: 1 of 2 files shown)


etrepum (approved):
Looks great, confirmed that the playground still had the bug and this fixes it, and it seems the test would exercise that too. 👏

PR #7188, opened 2025-02-16
Replace substr with substring

ivailop7:
substr is deprecated, replacing with the current equivalent to silence the linters


packages/lexical-react/src/shared/LexicalMenu.ts
```diff
@@ -111,7 +111,7 @@ function getFullMatchOffset(
 ): number {
   let triggerOffset = offset;
   for (let i = triggerOffset; i <= entryText.length; i++) {
-    if (documentText.substr(-i) === entryText.substr(0, i)) {
+    if (documentText.substring(-i) === entryText.substring(0, i)) {
```
etrepum, line 114: This one is not equivalent, `substr(startIndex, length)` and `substring(startIndex, endIndex)` don't work the same for `substr(-i)` and `substring(-i)`. Probably should be something like `documentText.substring(documentText.length - i)` instead

etrepum (commented):
One of these things (is not like the others)

packages/lexical-react/src/shared/LexicalMenu.ts
ivailop7: Nice catch! Will fix this evening.

etrepum: just occurred to me that there's also the slice method, `documentText.slice(-i)` is shorter and does the same thing

etrepum: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice

ivailop7: The 'slice' shout out is a good one. updated!

etrepum (approved):
LGTM, could've been slice all around to save a few bytes, but this works just the same!

ivailop7:
> LGTM, could've been slice all around to save a few bytes, but this works just the same!

Should probably reuse the utils method in all other plugins in playground and should probably save more bytes, if minifaction doesn't pick those up already behind the scenes

PR #7192, opened 2025-02-17
[Breaking Change][lexical-table] Bug Fix: Prevent nested tables

### LINKED ISSUE #7154: Bug: Inner table cell selection also selects outer table cell
Lexical version: 0.24.0

## Steps To Reproduce

1. Create a table and insert another table within one of its cells.
2. Select cells in the inner table.

https://github.com/user-attachments/assets/912f549e-9eb9-401c-af16-fb9e3b99399c

## The current behavior

When selecting cells within the inner table, sometimes the outer table cells are also selected, causing unintended selections.

## The expected behavior

Only the cells within the inner table should be selected without affecting the outer table cells.

## Impact of fix

- It is disrupts the user’s ability to accurately select and manipulate data within nested tables.

kirandash:
## Description
Current behavior:
- Tables can be nested inside table cells through both insertion and paste operations
- This leads to unintended selection behavior where selecting cells in inner tables also selects outer table cells
- Nested tables are not a supported use case as confirmed by @ivailop7 

Changes being added:
- Prevent table insertion inside table cells via `INSERT_TABLE_COMMAND`
- Block table paste operations within cells via `SELECTION_INSERT_CLIPBOARD_NODES_COMMAND`
- Add comprehensive test coverage:
  - Unit tests to verify command handlers in `TablePlugin`
  - E2E tests in `Tables.spec.mjs` to verify user interactions

Closes #7154

## Breaking Changes

Nested tables have never worked correctly, but there was partial "support" for them in the code. While it may make sense to support them in the future, a safer default behavior is to prevent their use until the infrastructure is in place to support them.

### Before

https://github.com/user-attachments/assets/f5ee9e2c-e40f-46c9-80bc-3dea49492fd2

### After

https://github.com/user-attachments/assets/f66e6726-f6b4-42f2-aef0-c688a0f284f0

### DIFF
--- a/packages/lexical-table/src/LexicalTablePluginHelpers.ts
+++ b/packages/lexical-table/src/LexicalTablePluginHelpers.ts
@@ -37,6 +37,7 @@ import {$isTableNode, TableNode} from './LexicalTableNode';
 import {$getTableAndElementByKey, TableObserver} from './LexicalTableObserver';
 import {$isTableRowNode, TableRowNode} from './LexicalTableRowNode';
 import {
+  $findTableNode,
   applyTableHandlers,
   getTableElement,
   HTMLTableElementWithWithTableSelectionState,
@@ -58,18 +59,9 @@ function $insertTableCommandListener({
     return false;
   }
 
-  // Check if we're inside a table cell
-  let node = selection.anchor.getNode();
-  while (node !== null) {
-    if ($isTableCellNode(node)) {
-      // We're inside a table cell, don't allow nested tables
-      return false;
-    }
-    const parent = node.getParent();
-    if (parent === null) {
-      break;
-    }
-    node = parent;
+  // Prevent nested tables by checking if we're already inside a table
+  if ($findTableNode(selection.anchor.getNode())) {
+    return false;
   }
 
   const tableNode = $createTableNodeWithDimensions(
@@ -296,28 +288,9 @@ export function registerTablePlugin(editor: LexicalEditor): () => void {
         if (!$isRangeSelection(selection)) {
           return false;
         }
-
-        // Check if we're inside a table cell
-        let isInsideTableCell = false;
-        let currentNode = selection.anchor.getNode();
-        while (currentNode !== null) {
-          if ($isTableCellNode(currentNode)) {
-            isInsideTableCell = true;
-            break;
-          }
-          const parent = currentNode.getParent();
-          if (parent === null) {
-            break;
-          }
-          currentNode = parent;
-        }
-
-        // If inside table cell and clipboard contains a table, prevent the paste operation
-        if (isInsideTableCell && nodes.some($isTableNode)) {
-          return true;
-        }
-
-        return false;
+        const isInsideTableCell =
+          $findTableNode(selection.anchor.getNode()) !== null;
+        return isInsideTableCell && nodes.some($isTableNode);
       },
       COMMAND_PRIORITY_EDITOR,
     ),
--- a/packages/lexical-table/src/LexicalTablePluginHelpers.ts
+++ b/packages/lexical-table/src/LexicalTablePluginHelpers.ts
@@ -14,10 +14,13 @@ import {
 } from '@lexical/utils';
 import {
   $createParagraphNode,
+  $getSelection,
+  $isRangeSelection,
   $isTextNode,
   COMMAND_PRIORITY_EDITOR,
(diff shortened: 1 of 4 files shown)


packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -50,6 +53,25 @@ function $insertTableCommandListener({
...
+    if ($isTableCellNode(node)) {
+      // We're inside a table cell, don't allow nested tables
+      return false;
+    }
+    const parent = node.getParent();
+    if (parent === null) {
+      break;
+    }
+    node = parent;
+  }
```
etrepum, line 73:
```suggestion
  const node = $findTableNode(selection.anchor.getNode());
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -268,6 +290,37 @@ export function registerTablePlugin(editor: LexicalEditor): () => void {
...
+          if ($isTableCellNode(currentNode)) {
+            isInsideTableCell = true;
+            break;
+          }
+          const parent = currentNode.getParent();
+          if (parent === null) {
+            break;
+          }
+          currentNode = parent;
+        }
```
etrepum, line 313:
```suggestion
        const isInsideTableCell = $findTableNode(selection.anchor.getNode()) !== null;
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -268,6 +290,37 @@ export function registerTablePlugin(editor: LexicalEditor): () => void {
...
+
+        // If inside table cell and clipboard contains a table, prevent the paste operation
+        if (isInsideTableCell && nodes.some($isTableNode)) {
+          return true;
+        }
+
+        return false;
```
etrepum, line 320:
```suggestion
        return (isInsideTableCell && nodes.some($isTableNode));
```

etrepum (commented):
The code looks good but there are some simplifications we can apply before merge.

I think this is probably a net good change. It's small, isolated, has tests, and mostly prevents tables from being inserted in the UI without gutting the partial support for nested tables that exists elsewhere.

There are probably other situations where you could create tables programmatically, but rather than taking more extreme approaches (e.g. a transform) it probably makes sense to just chip away at nested table support. I think most of the work is handling DOM related issues with navigation (which probably shouldn't rely so much on DOM in this case).

Something that is probably going to come up is when a user pastes HTML with nested tables, the nested content is just going to disappear. We might want to accompany this change with some additional documentation somewhere that they should handle these cases in importDOM until nested tables are supposed.

packages/lexical-table/src/LexicalTablePluginHelpers.ts
kirandash: Thank you for the detailed review @etrepum! I've simplified the nested table checks using $findTableNode as suggested. Please let me know if you'd like to see any other simplifications.

kirandash:
> The code looks good but there are some simplifications we can apply before merge.
> 
> I think this is probably a net good change. It's small, isolated, has tests, and mostly prevents tables from being inserted in the UI without gutting the partial support for nested tables that exists elsewhere.
> 
> There are probably other situations where you could create tables programmatically, but rather than taking more extreme approaches (e.g. a transform) it probably makes sense to just chip away at nested table support. I think most of the work is handling DOM related issues with navigation (which probably shouldn't rely so much on DOM in this case).
> 
> Something that is probably going to come up is when a user pastes HTML with nested tables, the nested content is just going to disappear. We might want to accompany this change with some additional documentation somewhere that they should handle these cases in importDOM until nested tables are supposed.

Thanks for the detailed review @etrepum! I've updated the README to document the handling of nested tables, particularly focusing on the `importDOM` implementation as you suggested. The documentation now:

1. Clearly states that nested content will be removed by default during paste operations
2. Guides users on how to handle nested tables in their `importDOM` implementation
3. Provides high-level approaches for preserving nested table content

I haven't included specific code examples yet. Would you like me to add some practical examples of `importDOM` implementations to make it more concrete for users?

Let me know if you'd like any adjustments to the documentation or if there are other aspects we should cover.

etrepum (approved):
I think this level of documentation is totally sufficient, no need to go through the trouble of implementing that example code in this PR

PR #7193, opened 2025-02-17
Documentation Update: `ContentEditable` placeholder` in React Getting Started 

RobHannay:
## Description

When adding placeholders, I stumbled upon the `placeholder` prop within `ContentEditable`, the TODO, and subsequently #6171.

I had blindly copied the example from the docs, but that is out-of-date given this is the preferred prop now for a11y.

### Before

N/A

### After

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,12 @@ function Editor() {
   return (
     <LexicalComposer initialConfig={initialConfig}>
       <PlainTextPlugin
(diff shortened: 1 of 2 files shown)


etrepum (approved):
Looks good, also since #6814 all of the `...rest` attributes pass through directly so all aria-* properties are supported without renaming

PR #7194, opened 2025-02-17
[lexical-playground] Bug Fix: Optimize table cell resizer event listeners

### LINKED ISSUE #7099: Feature: Skip running TableCellResizer logic if no table is present in editor
Playing with react-scan, I see the TableCellResizer logic evaluates on every click inside the editor, irrespective if a table is present or not. If we have an efficient way to check if editorState has a tableNode, it might be worth adding a conditional in the plugin vs listening to all the event listeners at all times.

kirandash:
## Description

Current behavior:
- The `TableCellResizer` component's event handling logic runs regardless of whether a table exists in the editor
- This leads to unnecessary effect hook executions and potential performance overhead

Changes in this PR:
- Added a check to prevent event listener setup when no table exists
- The entire event handling effect is now skipped when there's no table in the editor
- This change improves performance by avoiding unnecessary effect hook executions

Closes #7099

### Before

https://github.com/user-attachments/assets/e8608f95-a4cf-467d-85ba-f0884ad20bb4

### After

https://github.com/user-attachments/assets/f34832bf-8b64-4987-afd9-af7538fd6aea

### DIFF
--- a/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
@@ -6,7 +6,7 @@
  *
  */
 import type {TableCellNode, TableDOMCell, TableMapType} from '@lexical/table';
-import type {LexicalEditor} from 'lexical';
+import type {LexicalEditor, NodeKey} from 'lexical';
 import type {JSX} from 'react';
 
 import './index.css';
@@ -18,19 +18,13 @@ import {
   $getTableNodeFromLexicalNodeOrThrow,
   $getTableRowIndexFromTableCellNode,
   $isTableCellNode,
-  $isTableNode,
   $isTableRowNode,
   getDOMCellFromTarget,
   getTableElement,
   TableNode,
 } from '@lexical/table';
 import {calculateZoomLevel} from '@lexical/utils';
-import {
-  $getNearestNodeFromDOMNode,
-  $getRoot,
-  isHTMLElement,
-  LexicalNode,
-} from 'lexical';
+import {$getNearestNodeFromDOMNode, isHTMLElement} from 'lexical';
 import * as React from 'react';
 import {
   MouseEventHandler,
@@ -58,6 +52,7 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
   const resizerRef = useRef<HTMLDivElement | null>(null);
   const tableRectRef = useRef<ClientRect | null>(null);
   const [hasTable, setHasTable] = useState(false);
+  const tableKeysRef = useRef<Set<NodeKey>>(new Set());
 
   const mouseStartPosRef = useRef<MousePosition | null>(null);
   const [mouseCurrentPos, updateMouseCurrentPos] =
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
```diff
@@ -74,20 +81,43 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
...
-      }
+    // Check for table existence using editorState
+    const unregisterUpdate = editor.registerUpdateListener(() => {
+      editor.getEditorState().read(() => {
+        const hasTableNow = $getRoot()
+          .getChildren()
+          .some((node: LexicalNode) => $isTableNode(node));
+        setHasTable(hasTableNow);
+      });
+    });
```
etrepum, line 92: This isn't really comprehensive because root isn't the only place a table can go, could be in some other shadow root. A mutation listener would work without having to traverse the whole document to find table nodes. Something like this:

```ts
const tableKeys = new Set<NodeKey>();
const unreg = editor.registerMutationListener(TableNode, (nodeMutations) => {
  for (const [nodeKey, mutation] of nodeMutations) {
    if (mutation === 'destroyed') {
      tableKeys.delete(nodeKey);
    } else {
      tableKeys.add(nodeKey);
    }
  }
  setHasTable(tableKeys.size > 0);
});
```

kirandash: Thanks for the detailed feedback @etrepum! I've implemented your suggestion using `registerMutationListener` and pushed the changes.

Let me know if you'd like me to make any further adjustments to the implementation.

packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
```diff
@@ -74,20 +76,48 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
...
-    });
+        tableNode.setColWidths(Array(numColumns).fill(columnWidth));
+        return tableNode;
+      },
+    );
+
+    return () => {
+      unregisterMutation();
+      unregisterTransform();
+    };
```
etrepum, line 113: The normal pattern for this is to use

```suggestion
    return mergeRegsiter(
      unregisterMutation,
      unregisterTransform,
    );
```

Although typically the calls are also inline, which reduces code size a bit further and makes sure they're all actually called

```
    return mergeRegsiter(
      editor.registerMutationListener(…),
      editor.registerNodeTransform(…),
    );
```

packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
```diff
@@ -51,6 +51,8 @@ function TableCellResizer({editor}: {editor: LexicalEditor}): JSX.Element {
   const targetRef = useRef<HTMLElement | null>(null);
   const resizerRef = useRef<HTMLDivElement | null>(null);
   const tableRectRef = useRef<ClientRect | null>(null);
+  const [hasTable, setHasTable] = useState(false);
+  const tableKeysRef = useRef<Set<NodeKey>>(new Set());
```
etrepum, line 55: This should actually be inside of the effect, just as a regular variable in the closure. `registerMutationListener` is going to fully initialize it, but the unregisterMutationListener won't send a callback that clears it… so having it out here makes it inconsistent if the editor changes while the plugin is unmounted for whatever reason.

packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
kirandash: Thanks! I've updated the code to use `mergeRegister` for combining the cleanup functions and also made the calls inline. Will keep it in mind for future PRs.

packages/lexical-playground/src/plugins/TableCellResizer/index.tsx
kirandash: Thanks for explaining the consequence of keeping `tableKeys` outside the effect! I had initially implemented your suggested mutation listener code but missed the importance of placing `tableKeys` inside the effect closure. Now I understand that keeping it outside could lead to stale state if the editor changes while unmounted. I've moved it inside the effect as you suggested.

etrepum (approved):
Another nice one! Feel free to reach out on the [discord](https://discord.gg/KmG4wQnnD9) if you have any questions, I'm pretty familiar with most of lexical at this point

kirandash:
> Another nice one! Feel free to reach out on the [discord](https://discord.gg/KmG4wQnnD9) if you have any questions, I'm pretty familiar with most of lexical at this point

Thank you for all your help and guidancel! I'm learning a lot about Lexical and better coding practices through your thorough reviews. Really appreciate you taking the time to explain the reasoning behind suggestions. Will definitely reach out on Discord for any questions!

PR #7196, opened 2025-02-18
[ci] Chore: Upgrade astro to fix mysterious integration test failure

etrepum:
## Description

astro integration test strangely started failing. Locally it seems that the server would just hang on requests instead of serving them. Looks like it might've been the --no-package-lock usage, which was intended to make sure we were testing a recent version of the frameworks under test, but it's not really worth the hassle on our end to have that setting.

### Before

Integration tests would fail for no good reason (e.g. https://github.com/facebook/lexical/actions/runs/13383012921/job/37374781282)

### After

Integration tests pass again

### DIFF
--- a/scripts/__tests__/integration/fixtures/lexical-esm-astro-react/package-lock.json
+++ b/scripts/__tests__/integration/fixtures/lexical-esm-astro-react/package-lock.json
@@ -1,21 +1,21 @@
 {
   "name": "lexical-esm-astro-react",
-  "version": "0.22.0",
+  "version": "0.24.0",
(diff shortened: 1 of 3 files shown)


etrepum:
I guess it must just be some kind of poisoned cache in the other #7117 PR but I don't really have sufficient privileges to investigate further (or just try and reset the cache). I've disabled the test in that PR, we're not currently getting a lot of value out of it but it would be nice to have examples with other frameworks at some point.

ivailop7 (approved):

PR #7202, opened 2025-02-18
[lexical] Bug Fix: catch setTimeout wrapped setBaseAndExtent call

StormVanDerPol:
## Description

https://github.com/facebook/lexical/pull/7122 added a `setBaseAndExtent` call on `LexicalSelection.ts` line `2935` that as far as I understand is intended to be caught and ignored if an error is thrown, but since it's wrapped inside of a `setTimeout` callback it's still uncaught. 

Would be nice if it was handled the same way as the `setBaseAndExtent` call which isn't in the `setTimeout` callback (line `2925`). if that's the intention at least. 

lmk if it makes sense, thx.

### Before
potential Uncaught `Failed to execute 'setBaseAndExtent'` may happen

### After
potential Uncaught `Failed to execute 'setBaseAndExtent'` don't happen, as I assume was the intention

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -2932,12 +2932,18 @@ export function updateDOMSelection(
     // This is a workaround to restore the correct selection
     if (IS_ANDROID_CHROME) {
       setTimeout(() => {
-        domSelection.setBaseAndExtent(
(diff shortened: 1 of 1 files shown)


potatowagon (approved):
yes its meant to be ignored. i was thinking of adding a check for index size error conditions but this works too

PR #7204, opened 2025-02-18
[Breaking Change][lexical-playground]: Refactor: Remove special case for collapsible forward deletion

### LINKED ISSUE #7203: Bug: Forward deletion of CollapsibleContainerNode inconsistent with backwards deletion
Lexical version: v0.24.0

## Steps To Reproduce

[example doc](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE9VUTUsDMRD9L-8cy7YVdfMDPAs9ioe4GdvANFlmZ4ul7NWz-BP9JZKtpR9C0aqgp2QezJv3XpisQD5okok6JdgVJCXNZzUL7IUi7O1B4UldYNjC4CHJ3Gl_nSdPsIgZYRg0uuQMwEDpUWHx-vSSi2Wd4R4zWJA0IUXYYXdn4INQpX0dW-Ytf2YJ0VNcz3rnqJ24qbh6tke0nne9VZbLyUZNZ45Z-xGfGpTp807BKvgVqyfMqRKzq5twz3S28bGj_Gh4p-ZVpahZxv9PbOvke9oPOV2IJAc-Up1jV2npw6ucvqDPf2BBvz6m_7L2pHYG7BqduAV52OHluLwqx6OiLC5Kgya1UuW-G3bLqaQ2-t1uFIPR-aBA9wZSRDjNGwUAAA)

See e2e selection test 'Can delete forward a Collapsible' that encodes this behavior.

1. Position the caret after the right arrow (there needs to be some inline content in the paragraph above, otherwise the empty paragraph gets deleted)
2. Press option-delete for forward delete

When #7155 was implemented to make decorators and elements with shadow roots behave correctly, a workaround was implemented to preserve the behavior encoded by this test.

https://github.com/facebook/lexical/pull/7155#discussion_r1959325465_

## The current behavior

The node's content gets spliced into the document

## The expected behavior

The collapsible gets deleted

## Impact of fix

Consistency goes up, code size goes down

etrepum:
## Description

Now that we have more robust handling of deleteCharacter around decorators and shadow roots, we can get rid of the expectation that CollapsibleContainerNode does something weird. AFAICT this behavior was a result of the implementation and not something that was specifically intended?

Follow-up to #7155 
Related #3082 #4257 https://github.com/facebook/lexical/blame/f5e6329c3773a3e968eeb06a50ae002ce8b1318f/packages/lexical-playground/src/plugins/CollapsiblePlugin/index.ts#L169

Closes #7203

### Before

Delete forward before a CollapsibleNode would pierce the shadow root and unroll its content at the caret

### After

Delete forward before a CollapsibleNode will do nothing, like other shadow roots (e.g. table)

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -380,7 +380,7 @@ test.describe.parallel('Selection', () => {
     );
   });
 
-  test('Can delete forward a Collapsible', async ({page, isPlainText}) => {
+  test(`Can't delete forward a Collapsible`, async ({page, isPlainText}) => {
     test.skip(isPlainText);
     if (!IS_MAC) {
       // Do Windows/Linux have equivalent shortcuts?
@@ -389,6 +389,7 @@ test.describe.parallel('Selection', () => {
     await focusEditor(page);
     await page.keyboard.type('abc');
     await insertCollapsible(page);
+    await page.keyboard.type('title');
     await moveToEditorBeginning(page);
     await moveRight(page, 3);
     await deleteForward(page);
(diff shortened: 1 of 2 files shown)


ivailop7 (dismissed):

etrepum:
Looks like there's probably a webkit-specific issue to handle, will take a look today

ivailop7 (approved):

PR #7205, opened 2025-02-18
[lexical-table] Feature: Support TableNode.__style in createDOM and updateDOM

### LINKED ISSUE #7201: Feature: TableNode styling
TableNode ignores LexicalNode.__style prop completely.
it could be a good idea to take it into account.

currently I found myself doing this:
```
TableNode.importDOM = function ()
{
	return {
		...tableImporters,
		table: () => ({
			conversion: function (domNode)
			{
				const node = $createTableNode();
				node.setStyle(domNode.style.cssText);
				return {node};
			},
			priority: 0
		})
	};
};

const org_TableNode_createDOM = TableNode.prototype.createDOM;
TableNode.prototype.createDOM = function ()
{
	const dom = org_TableNode_createDOM.apply(this, arguments);
	if (this.getStyle())
		dom.style.cssText = this.getStyle();
	return dom;
};

//updates active editor's table style
const org_TableNode_updateDOM = TableNode.prototype.updateDOM;
TableNode.prototype.updateDOM = function (prevNode, dom)
{
	dom.style.cssText = this.getStyle();
	return org_TableNode_updateDOM.apply(this, arguments);
};
```

etrepum:
## Description

TableNode ignored its __style attribute, this PR adds support for it

Closes #7201

### Before

`tableNode.setStyle(…)` has no effect on the DOM

### After

`tableNode.setStyle(…)` is reconciled to the DOM, and this is unit tested.

### DIFF
--- a/packages/lexical-table/src/LexicalTableNode.ts
+++ b/packages/lexical-table/src/LexicalTableNode.ts
@@ -264,6 +264,9 @@ export class TableNode extends ElementNode {
 
   createDOM(config: EditorConfig, editor?: LexicalEditor): HTMLElement {
     const tableElement = document.createElement('table');
+    if (this.__style) {
+      tableElement.style.cssText = this.__style;
+    }
     const colGroup = document.createElement('colgroup');
     tableElement.appendChild(colGroup);
     updateColgroup(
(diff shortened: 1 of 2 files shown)


ivailop7 (approved):

allenhehe:
I might be missing something, but shouldn't `importDOM` also handle `__style` transfer to ensure existing tables function correctly?

I noticed that TableNode's `importDOM` doesn't seem to assign CSS styles, as seen here: 
- https://github.com/facebook/lexical/blob/0777324d026642a94305b4ca873cc77b7a68dd62/packages/lexical-table/src/LexicalTableNode.ts#L210
- https://github.com/facebook/lexical/blob/0777324d026642a94305b4ca873cc77b7a68dd62/packages/lexical-table/src/LexicalTableNode.ts#L611

Furthermore, when copying colored text from external sites into the Lexical Playground, the colors are lost. According to the [documentation](https://lexical.dev/docs/concepts/serialization#handling-extended-html-styling), overriding importDOM is required to retain such styling.

Is this behavior intentional for both tables and text nodes, or is it a known limitation?

PR #7214, opened 2025-02-20
[lexical-playground] [markdown] fix: block equations falsely converted to inline in markdown conversion

### LINKED ISSUE #6936: Bug: Block math equation is falsely converted to inline equation on MarkdownShortcutPlugin

nmzabith:
## Description

This pull request includes changes to improve the handling of inline and block equations in the `EquationNode` and `MarkdownTransformers` classes. The most important changes include adding a method to determine if an equation is inline and modifying the regex patterns to better handle both inline and block equations.

Improvements to equation handling:

* [`packages/lexical-playground/src/nodes/EquationNode.tsx`](diffhunk://#diff-920e946ba05ca18c451cc214819702b0d9b24ea4d9648e55b5d79406c4659835R148-R151): Added `isInline` method to the `EquationNode` class to determine if an equation is inline.

* [`packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts`](diffhunk://#diff-73191b3f45b7609f2b2ea7229779b6dcb62706ca175504a5157eec6bd76d6039R120-R141): Modified the `EQUATION` transformer to use the `isInline` method and updated the regex patterns to better handle both inline and block equations.## Description

Closes #6936 

## Test plan

1. add the below markdown content through MarkdownShortcutPlugin
2. Press the Markdown button
3. Press the Markdown button (Block equations wont converted to inline)

Other way to test

2. Press the Markdown button
3. Press the Markdown button (Block equations wont converted to inline)

```
# KaTeX Equations in Markdown

## Inline Equations

You can write inline equations like this: $E = mc^2$ or $a^2 + b^2 = c^2$.

A more complex inline equation example:

\( \sum_{n=1}^{\infty} \frac{1}{n^2} = \frac{\pi^2}{6} \)

## Block Equations

Block equations are useful for displaying more complex mathematical expressions. You can use them by wrapping in `$$`.

### Example 1: Quadratic Formula

$$ x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a} $$

### Example 2: Fourier Transform

$$ F(k) = \int_{-\infty}^{\infty} f(x) e^{-2\pi i k x} \,dx $$

### Example 3: Einstein Field Equations

$$ R_{\mu\nu} - \frac{1}{2} R g_{\mu\nu} + \Lambda g_{\mu\nu} = \frac{8\pi G}{c^4} T_{\mu\nu} $$

### Example 4: Probability Distribution (Gaussian Function)

$$ P(x) = \frac{1}{\sigma \sqrt{2\pi}} e^{-\frac{(x - \mu)^2}{2\sigma^2}} $$

### Example 5: Summation and Product Notation

$$ \sum_{i=1}^{n} i = \frac{n(n+1)}{2} $$

$$ \prod_{k=1}^{n} k = n! $$

## Conclusion

KaTeX makes it easy to include both simple and complex equations in Markdown. Whether using inline math \( x^2 + y^2 = z^2 \) or block equations, it's a powerful tool for documentation and note-taking.

```
### Before

image

### After

image


etrepum (commented):
It would be great to add some tests to this PR to confirm that it works as expected. There's a `deleteCharacter merges children from adjacent blocks even if the previous leaf is an inline decorator` regression test that uses the equation node, you could use that as a starting point. There are also some unit tests for LexicalMarkdown.

C0casio45:
After some test on my side, the code for equation is not working consistently, and I came up with this modification


```js
replace: (textNode, match) => {
    const fullMatch = match[0];
    const input = match.input;
    const blockEq = match[1] === fullMatch ? match[2] : match[1];
    const inlineEq = match[1] === fullMatch ? match[3] : match[2];

    // If the equation has $$ it's a block equation
    const isBlock = input.startsWith("$$") && input.endsWith("$$");
    const equation = isBlock ? blockEq : inlineEq;

    console.log("EQUATION", equation);

    if (!equation || (input.startsWith("$$") && !input.endsWith("$$"))) {
      return;
    }

    const equationNode = $createEquationNode(equation.trim(), !isBlock);
    textNode.replace(equationNode);
  }
```

etrepum:

PR #7217, opened 2025-02-20
[lexical-yjs] Bug Fix: Normalize multiple adjacent merge conflicts in one block

james-atticus:
## Description

If you have multiple adjacent text nodes in a single block, and they're all merged (normalised) in a single update, and that update is either tagged `collaboration` or `historic`, then Lexical and Yjs get out of sync. This is caused by collab nodes being removed from their parent too early in `$handleNormalizationMergeConflicts`, which leads to offset calculations being wrong.

## Test plan

I don't know how the editor gets into this state through normal user behaviour, but we've seen it in production multiple times now. Most recently today, where a user had a document with lots of repeated content, similar to https://github.com/facebook/lexical/issues/6614#issuecomment-2424855275 (we're on 0.24.0 though).

I'm still working on a repro which could be added to E2E tests, but for now I've resorted to manually getting the Yjs doc into this weird state in a lower-level test.

### Before

```
  ● Collaboration › Should handle multiple text nodes being normalized due to merge conflict

    Expected: "<p><span data-lexical-text=\"true\">123</span></p>"
    Received: "<p><span data-lexical-text=\"true\">13</span></p>"

      465 |     });
      466 |
    > 467 |     expect(client1.getHTML()).toEqual(
          |                               ^
      468 |       '<p><span data-lexical-text="true">123</span></p>',
      469 |     );
      470 |     expect(client1.getHTML()).toEqual(client2.getHTML());

      at Object.<anonymous> (packages/lexical-react/src/__tests__/unit/Collaboration.test.ts:467:31)
```

### After

Tests pass.

### DIFF
--- a/packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
+++ b/packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
@@ -14,6 +14,7 @@ import {
   TextNode,
   UNDO_COMMAND,
 } from 'lexical';
+import * as Y from 'yjs';
 
 import {Client, createTestConnection, waitForReact} from './utils';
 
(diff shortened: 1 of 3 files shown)


packages/lexical-yjs/src/SyncEditorStates.ts
```diff
@@ -175,22 +176,25 @@ function $handleNormalizationMergeConflicts(
 
         const parent = collabNode._parent;
         collabNode._normalized = true;
-
         parent._xmlText.delete(offset, 1);
 
-        collabNodeMap.delete(nodeKey);
-        const parentChildren = parent._children;
-        const index = parentChildren.indexOf(collabNode);
-        parentChildren.splice(index, 1);
```
james-atticus, line 184: We're removing the collab node's `YMap` from the parent XML, but its _text_ is still there. When calculating the offset for subsequent nodes (i.e. the next loop iteration), we still need to count that text.

If you remove the collab text node from its parent, then in the next loop iteration, `collabNode.getOffset()` above will return the wrong value, and `parent._xmlText.delete(offset, 1);` ends up deleting part of the text instead of the `YMap`.

packages/lexical-yjs/src/SyncEditorStates.ts
```diff
@@ -175,22 +176,25 @@ function $handleNormalizationMergeConflicts(
...
+    const nodeKey = collabNode.getKey();
+    collabNodeMap.delete(nodeKey);
+    const parentChildren = collabNode._parent._children;
+    const index = parentChildren.indexOf(collabNode);
+    parentChildren.splice(index, 1);
+  }
+
   for (let i = 0; i < mergedNodes.length; i++) {
     const [collabNode, text] = mergedNodes[i];
-    if (collabNode instanceof CollabTextNode && typeof text === 'string') {
```
james-atticus, line 191: Drive-by: commit history says this check was added to make `tsc` happy, but I thought it was more clear to just have a type for `mergedNodes` defined above.

packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
```diff
@@ -409,4 +410,70 @@ describe('Collaboration', () => {
...
+        const paragraph = root.getFirstChild<ParagraphNode>()!;
+        paragraph.append($createTextNode('1'));
+      });
+    });
+
+    expect(client1.getHTML()).toEqual(
+      '<p><span data-lexical-text="true">1</span></p>',
+    );
+    expect(client1.getDocJSON()).toEqual({
+      root: '[object Object]1',
```
etrepum, line 439: This sort of test seems really weird to me, but I can see you're just copying that technique from elsewhere. `toMatchObject` might work better for what this is trying to do?

etrepum (approved):
LGTM! Nice catch, looks like a tricky one to figure out. Great job adding some test coverage for the problem as well.

packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
james-atticus: Yeah it's not great. I think it stems from the use of YMap embeds in YXmlText. YXmlText's `toString` just [converts](https://github.com/yjs/yjs/blob/2fba694/src/types/YXmlText.js#L94) these to `[object Object]`. `toMatchObject` wouldn't change this so I'm inclined to leave as is - it would be unexpected for there to be keys other than `root` on the doc JSON.

PR #7226, opened 2025-02-21
[lexical] Chore: Added missing `isInline` function to TextNode

mrdivyansh:
## Description
LexicalNode expects the extender Node to implement `isInline`. However, it is missing from `TextNode` implementation.

https://github.com/facebook/lexical/blob/054fb46f13261e0e2115995bb0a71426c04fdf3f/packages/lexical/src/LexicalNode.ts#L317-L323

### Before

- Invoke `isInline` function on a `TextNode` instance.
- `Invariant` will throw an error -- `LexicalNode: Node TextNode does not implement .isInline().`

### After

- Invoke `isInline` function on a `TextNode` instance.
- The function should return `true`.


mrdivyansh:
Not sure if this PR requires any unit tests.

etrepum:
This might make sense, but there should already be guards to prevent this from happening on nodes other than ElementNode and DecoratorNode. Did you run into a situation where this exception occurs? If so, that’s where the unit test should be. I think otherwise it might make more sense to move this to the ElementNode and DecoratorNode interfaces specifically because those are the only two types of nodes where it may return something meaningful.

packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -473,6 +473,13 @@ export class TextNode extends LexicalNode {
     return true;
   }
 
+  /**
+   * @returns true if the text node is inline, false otherwise.
+   */
+  isInline(): boolean {
```
etrepum, line 479: If this does make sense to implement on TextNode, it would make sense to also narrow the implementation to return only true, because it's not permitted for any subclass to return false

```suggestion
  isInline(): true {
```

etrepum:
Also if we do decide to implement this, there's a fourth node that extends LexicalNode that should implement isInline: LineBreakNode. That would cover all of the expected cases, since the value for LexicalNode isn't exported there's no reasonable way to add another class to the hierarchy.

PR #7227, opened 2025-02-21
[lexical-link] Bug Fix: Remove anchors with empty href during DOM import

mrdivyansh:
## Description
Currently, `LinkNode` allows importing DOM with an empty `href`. However, IMO, it doesn't make sense during HTML paste. 

### Before

https://github.com/user-attachments/assets/a2ce4f75-0a7d-451a-8f7e-2ad6791cfcb1

### After

https://github.com/user-attachments/assets/b5adb8bc-6931-4df1-9c4e-b63ae31af911

### DIFF
--- a/packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
+++ b/packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
@@ -6,6 +6,7 @@
  *
  */
 
+import {$insertDataTransferForRichText} from '@lexical/clipboard';
 import {
   $createLinkNode,
   $isLinkNode,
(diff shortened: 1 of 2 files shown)


etrepum (commented):
I'm not sure this is the best place to encode this normalization, it's easy enough to do this in editor html.import configuration or with a LinkNode transform if it's something that's appropriate for your editor. It looks like this was an intentional decision to allow empty links rather than to normalize them away https://github.com/facebook/lexical/issues/3194

mrdivyansh:
@etrepum , I don't see a valid use case for an imported link without an href. I'll address this on my end. Feel free to close the PR. (let me know if you want me to close the PR)

PR #7228, opened 2025-02-21
[@lexical/list] Feature: export `ListNodeTagType`

DaniGuardiola:
## Description

Export `ListNodeTagType` since it's helpful to consumers. Otherwise, they need to do something like the following to obtain the type:

```ts
import type { ListNodeTagType } from 'node_modules/@lexical/list/LexicalListNode'
```

## Test plan

n/a

### DIFF
--- a/packages/lexical-list/src/index.ts
+++ b/packages/lexical-list/src/index.ts
@@ -7,7 +7,11 @@
  */
 
 import type {SerializedListItemNode} from './LexicalListItemNode';
-import type {ListType, SerializedListNode} from './LexicalListNode';
+import type {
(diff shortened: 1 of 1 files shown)


etrepum (approved):
FWIW a much better general solution to acquiring types that should've been exported is to make an alias to a value of that type. Since TS is structural this is basically indistinguishable from the real thing unless there are generics involved.

stuff like this will work in a pinch:

```ts
type ListNodeTagType = SerializedListNode['tag'];
```

```ts
type ListNodeTagType = ReturnType<ListNode['getTag']>;
```

PR #7229, opened 2025-02-21
[lexical-playground] Bug Fix Table Action Menu dropdown positioning

umaranis:
## Description

Table Action Menu is not positioned properly when the window is vertically scrolled.

### Before

https://github.com/user-attachments/assets/be36b015-03b1-47f3-b6ab-db2c8bc382b6

### After

https://github.com/user-attachments/assets/b816f51e-c2e1-4b2f-941e-1e017379639f

I have tested the following scenarios and they work fine after the fix:
- when there is enough space below the Menu Button to display the Menu
- when there is not enough space below the Menu Button but enough space above
- when there is not enough space below and above the Menu Button
- tested the above three scenarios when page is scrolled and not scrolled


etrepum (approved):
Seems like an improvement. There are still issues if scrolling when this is open, but I would rather see that sort of thing fixed systematically e.g. by integrating floating-ui #6923

PR #7237, opened 2025-02-23
[lexical-react] fix: ensure attributes are set immediately on menu

sodenn:
## Description
With the changes introduced in Lexical 0.25.0 (notably #7181 and #7164), the menu now renders immediately. However, this has led to two issues:

**1. Missing attributes on initial render**
  The menu container `<div>` is rendered before its attributes are assigned. As a result:
  - Styles defined via `anchorClassName` are not applied immediately, causing visibility issues. The attributes only appear after the user makes further inputs. In my case, the menu is covered by a dialog because the `z-index` from my custom class is not applied initially.
  - Other missing attributes, such as `role` and `aria-label`, make automated testing more difficult.

**2. Persistent DOM Element (not addressed in this PR)**
The menu container `<div>` remains in the DOM even when the menu is closed. This can lead to duplicate IDs, potentially causing accessibility and testing issues.

**Fix in This PR**
This PR ensures that all attributes are correctly assigned as soon as the menu is rendered.

### Before
before

### After
after1
after2


etrepum (dismissed):

etrepum:
Perhaps a path for a short term fix might be to only assign the id to the active menu, so there's only one typeahead-menu in the document at a time. That probably wouldn't require changing the tests and other logic that depends on the typeahead-menu id

etrepum (dismissed):
This looks good to me and seems to work well with manual testing in the playground, nice job! Will merge assuming the extended test suite is clear.

packages/lexical-playground/__tests__/e2e/Mentions.spec.mjs
```diff
@@ -1019,4 +1020,15 @@ test.describe('Mentions', () => {
       focusPath: [0, 1, 0],
     });
   });
+
+  test(`Sets correct attributes on typeahead menu container`, async ({
+    page,
+  }) => {
```
etrepum, line 1026:
```suggestion
  test(`Sets correct attributes on typeahead menu container`, async ({
    page,
    isCollab,
  }) => {
    test.skip(isCollab);
```

etrepum (commented):
Looks like there are some test failures, I don't think it's meaningful to perform this specific test on collab so we should just skip it there

etrepum (approved):

PR #7239, opened 2025-02-24
Fix selection shifting when deleting paragraphs on android

### LINKED ISSUE #4340: Bug: When deleting text with the keyboard of a device it's also removing text to the right of the pointer.

waynetee:
## Description
Fixes #4340 

This is a second attempt to fix the abovementioned bug, following the unsuccessful attempt in #7122 (apologies for the issues introduced then)

In this PR, selection is only restored during deletion and when several other criteria have been met, this should minimise the chances of any unintended side effects.

### Before
https://github.com/user-attachments/assets/6f555449-8112-400b-8d21-ed35b249f428

### After
https://github.com/user-attachments/assets/7617e10e-faf6-4d25-b1ab-e73c5f8a0dad

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -622,6 +622,41 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
           }
           if (!shouldLetBrowserHandleDelete) {
             dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
+
+            // When deleting across paragraphs, Chrome on Android incorrectly shifts the selection rightwards
+            // We detect this and restore the selection accordingly
+            const selectionAfterDelete = $getSelection();
+            if (
+              IS_ANDROID_CHROME &&
+              $isRangeSelection(selectionAfterDelete) &&
+              selectionAfterDelete.isCollapsed()
+            ) {
+              setTimeout(() =>
+                editor.update(() => {
+                  // Verify that selection has shifted rightwards
+                  const newSelection = $getSelection();
+                  if (
+                    !$isRangeSelection(newSelection) ||
+                    !newSelection.isCollapsed()
+                  ) {
+                    return;
+                  }
+                  const newAnchor = newSelection.anchor;
+                  const newAnchorSibling = newAnchor
+                    .getNode()
+                    .getPreviousSibling();
+                  const prevAnchor = selectionAfterDelete.anchor;
+                  if (
+                    (newAnchor.key === prevAnchor.key &&
+                      newAnchor.offset === prevAnchor.offset + 1) || // Rightward shift in same node
+                    (newAnchor.offset === 1 &&
+                      prevAnchor.getNode().is(newAnchorSibling)) // Rightward shift into sibling node
+                  ) {
+                    $setSelection(selectionAfterDelete.clone()); // Restore selection if shifted
+                  }
+                }),
+              );
+            }
           }
         }
         return;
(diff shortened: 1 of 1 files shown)


etrepum (commented):
I think the ideal approach here is to come up with a solution that does not depend on setTimeout. This kind of solution causes a lot of trouble down the line with debugging, possibly creating extra history entries, etc. even when it does work.

waynetee:
@etrepum I agree with you that in a ideal world we would not need to use to use `setTimeout`. I did spend a significant amount of time investigating this issue and as far as I can tell originates from the browser, making a clean fix difficult.

I see a few ways forward:
1. We merge this PR
2. Rather than using `setTimeout` in the `beforeinput` handler, we apply the fix in the `selectionchange` handler. This is workable, but requires a way to sync it to the deletion event. We would likely end up with a fix not much cleaner than this and might end up having to use `setTimeout` anyways.
3. We solve the problem in the browser itself, this would be ideal, but it doesn't seem like any progress has been made on the issue since it was filed early last year https://issues.chromium.org/issues/331320419
4. You or someone else volunteers to investigate this issue to try to come up with a better fix. I would be happy to assist in this case.

I personally think this is a significant bug, it would be good if we can resolve it in a timely manner.

etrepum (commented):
As far as I can tell with a Google Pixel 5 on Sauce Labs this build still has the bug

waynetee:
Turns out the fix was skipped over by an earlier optimisation in certain cases. It should be fixed now, and I have tested it on both Pixel and Samsung, and with the SwiftKey keyboard too.

Thanks for helping to test this thoroughly.

etrepum (approved):
This works with the Pixel and Samsung devices I tried on browser stack with the google, samsung and microsoft keyboards. Does anyone else want to give this a test before merge? /cc @zurfyx @potatowagon

packages/lexical/src/LexicalEvents.ts
```diff
@@ -622,6 +647,18 @@ function onBeforeInput(event: InputEvent, editor: LexicalEditor): void {
           }
           if (!shouldLetBrowserHandleDelete) {
             dispatchCommand(editor, DELETE_CHARACTER_COMMAND, true);
+            // When deleting across paragraphs, Chrome on Android incorrectly shifts the selection rightwards
+            // We save the correct selection to restore later during handling of selectionchange event
+            const selectionAfterDelete = $getSelection();
+            if (
+              IS_ANDROID_CHROME &&
```
zurfyx, line 654: Is it only specific to Chrome?

zurfyx (approved):

packages/lexical/src/LexicalEvents.ts
waynetee: Chromium browsers, based on my testing

PR #7244, opened 2025-02-25
[Update] Upgrade docusaurus version

lilshady:
## Description
This pull request is to update the  docusaurus version
currently, inside lexical-website, the docusaurus/plugin-client-redirects has a different version(3.7.0) so the `docusaurus build` would fail.

### Before

when run `docusaurus build`

 Error: Invalid name=docusaurus-plugin-client-redirects version number=3.7.0.
  All official [@docusaurus](https://www.internalfb.com/intern/profile/docusaurus)/* packages should have the exact same version as [@docusaurus](https://www.internalfb.com/intern/profile/docusaurus)/core (number=3.6.0).

### After

able to run `docusaurus build`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -175,281 +175,147 @@
       }
     },
     "node_modules/@algolia/autocomplete-core": {
-      "version": "1.9.3",
-      "resolved": "https://registry.npmjs.org/@algolia/autocomplete-core/-/autocomplete-core-1.9.3.tgz",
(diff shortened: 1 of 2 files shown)


zurfyx (approved):
Thank you, I can't remember why we used exact versions in the past. But this seems reasonable

potatowagon:
for tesing, u can view the generated lexical website at the vercel preview link commented by vercel bot

i checked it looks fine

PR #7252, opened 2025-02-26
[fix] Add missing flow types for node event plugin

fantactuka:


packages/lexical-react/flow/LexicalNodeEventPlugin.js.flow
```diff
@@ -7,6 +7,14 @@
  * @flow strict
  */
 
-/**
- * LexicalNodeEventPlugin
- */
+import type {LexicalEditor, LexicalNode, NodeKey} from 'Lexical';
```
etrepum, line 10:
```suggestion
import type {LexicalEditor, LexicalNode, NodeKey} from 'lexical';
```

etrepum (approved):

PR #7254, opened 2025-02-26
[@lexical/code] Feature: export `SerializedCodeHighlightNode`

DaniGuardiola:
## Description

Export `SerializedCodeHighlightNode` since it's helpful to consumers. 

## Test plan

n/a

### DIFF
--- a/packages/lexical-code/src/CodeHighlightNode.ts
+++ b/packages/lexical-code/src/CodeHighlightNode.ts
@@ -38,7 +38,7 @@ import {$createCodeNode} from './CodeNode';
 
 export const DEFAULT_CODE_LANGUAGE = 'javascript';
 
-type SerializedCodeHighlightNode = Spread<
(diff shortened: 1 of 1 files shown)


etrepum (commented):
This isn't sufficient to actually use it outside of this package, you'd also have to export from the entrypoint (index.ts).

In the meantime you should be able to use `ReturnType<CodeHighlightNode['exportJSON']>` to refer to this value. It might make more sense to use a generic type in your application like `type JSONForNode<T extends LexicalNode> = ReturnType<T['exportJSON']>` instead of trying to make sure every declared serialization type is exported.

ivailop7 (approved):

ivailop7 (changes requested):
as per Bob's comments

etrepum:
Closed due to staleness, review feedback not addressed

PR #7255, opened 2025-02-27
[lexical][lexical-mark] Bug Fix: Identify <mark> as inline element

### LINKED ISSUE #4641: Bug: $generateNodesFromDOM() removes spaces around <Mark> Node
I have created a custom Mark node simular to original, that can be serialized/deserialized and stored in data base. After updating from 0.9.1 to 0.11.1 a bug appeared that seems to remove spaces around Mark node. 

Lexical version: 0.11.1

## Steps To Reproduce
1. Convert Node to MarkNode
2. Store changes in database as HTML
3. Next time Lexical updates editor from HTML -> Lexical using $generateNodesFromDOM() spaces around Mark node will be removed.

## The current behavior
image

## The expected behavior
image

PeterDekkers:
## Description
The [&lt;mark&gt;](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-mark-element) element was not correctly identified as an inline element.

Closes #4641

## Test plan

All pre-existing unit tests pass. A simple `<mark>`-specific test was added to the `lexical-mark` unit tests.

### Before

```typescript
const dom = new DOMParser().parseFromString(
    `<html><body>Foo <mark>Bar</mark> !</body></html>`,
    'text/html',
);
```
Output:
```html
<span data-lexical-text="true">Foo</span><mark data-lexical-text="true"><span>Bar</span></mark><span data-lexical-text="true">!</span>
```

### After

```typescript
const dom = new DOMParser().parseFromString(
    `<html><body>Foo <mark>Bar</mark> !</body></html>`,
    'text/html',
);
```
Output:
```html
<span data-lexical-text="true">Foo </span><mark data-lexical-text="true"><span>Bar</span></mark><span data-lexical-text="true"> !</span>
```

### DIFF
--- a/packages/lexical-mark/__tests__/unit/LexicalMarkNode.test.ts
+++ b/packages/lexical-mark/__tests__/unit/LexicalMarkNode.test.ts
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
+import {$generateNodesFromDOM} from '@lexical/html';
 import {$wrapSelectionInMarkNode, MarkNode} from '@lexical/mark';
 import {
   $createParagraphNode,
@@ -12,6 +13,7 @@ import {
   $createTextNode,
   $getRoot,
   ParagraphNode,
+  TextNode,
 } from 'lexical';
 import {
   $createTestDecoratorNode,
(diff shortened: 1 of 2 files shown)


etrepum (approved):
Code looks good and the test looks like it should confirm that it's working correctly. Will merge assuming the e2e test suite also looks good!

PR #7256, opened 2025-02-27
[lexical] Bug Fix: Point.isBefore could return incorrect result due to normalization

### LINKED ISSUE #3181: Bug: `Point`'s `isBefore` calculation can be incorrect
Reported internally by @olegbl:

> As far as I can tell, the calculation inside `Point.isBefore` is not quite right when comparing an Element Point to a Text Point:
> ```js
> if ($isElementNode(aNode)) {
>   const aNodeDescendant = aNode.getDescendantByIndex(aOffset);
>   aNode = aNodeDescendant != null ? aNodeDescendant : aNode;
> }
> ```
> 
> changes the value of aNode but does not change aOffset.
> As a result, when it gets used down below in
> ```js
> if (aNode === bNode) {
>   return aOffset < bOffset;
> }
> ```
> The aOffset is being applied twice and leads to an incorrect result.
> e.g.:
> ```js
> // nodes
> 1 (element node)
>   2 (text node)
>   3 (element node)
>     4 (text node)
>     5 (text node)
>                   <-- point1 is here
>     6 (text node) <-- point2 is inside here
>   7 (text node)
> 
> // point1
> key = 3
> offset = 2
> 
> // point2
> key = 6
> offset = 1
> 
> // test:
> point1.isBefore(point2) // this should be true
> 
> // aNode = node 3
> // aOffset = 2
> // bNode = node 6
> // bOffset = 1
> 
> if ($isElementNode(aNode)) {
>   const aNodeDescendant = aNode.getDescendantByIndex(aOffset);
>   aNode = aNodeDescendant != null ? aNodeDescendant : aNode;
> }
> 
> // aNode = node 6
> 
> if (aNode === bNode) { // true
>   return aOffset < bOffset; // 2 < 1 = false (this is wrong)
> }
> ```

etrepum:
## Description

Fixes a long-standing bug in Point.isBefore by using the NodeCaret implementation of `$comparePointCaretNext`

Closes #3181

### Before

If two points that are not normalized are compared, and isBefore's incorrect normalization causes the compared nodes to match (without fixing up the offsets) then the result can be wrong because the offsets refer to a different node entirely.

### After

The normalization is now correctly performed with carets, and there's a unit test to confirm that the known broken situation is now fixed.

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -98,10 +98,18 @@ export function $setBlocksType<T extends ElementNode>(
     prevNode.replace(element, true);
     if (newSelection) {
       if (key === newSelection.anchor.key) {
-        newSelection.anchor.key = element.getKey();
+        newSelection.anchor.set(
+          element.getKey(),
+          newSelection.anchor.offset,
+          newSelection.anchor.type,
+        );
       }
       if (key === newSelection.focus.key) {
-        newSelection.focus.key = element.getKey();
+        newSelection.focus.set(
+          element.getKey(),
+          newSelection.focus.offset,
+          newSelection.focus.type,
+        );
       }
     }
   }
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -17,6 +17,7 @@ import invariant from 'shared/invariant';
 import {
   $caretFromPoint,
   $caretRangeFromSelection,
+  $comparePointCaretNext,
   $createLineBreakNode,
   $createParagraphNode,
   $createTextNode,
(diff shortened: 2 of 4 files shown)


packages/lexical-selection/src/range-selection.ts
```diff
@@ -98,10 +98,18 @@ export function $setBlocksType<T extends ElementNode>(
     prevNode.replace(element, true);
     if (newSelection) {
       if (key === newSelection.anchor.key) {
-        newSelection.anchor.key = element.getKey();
+        newSelection.anchor.set(
```
etrepum, line 101: This function was the only remaining place I could find, outside of the tests, where we were manually setting properties on the Point. This is probably something we should disallow in the future (e.g. make it readonly in the PointType interface) so that caches and such can be updated correctly. I can see us eventually wanting to store some more information on the point such as a cached NodeCaret or a direction or something like that which we may clear when it's changed.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -151,23 +152,12 @@ export class Point {
...
-      aNode = aNodeDescendant != null ? aNodeDescendant : aNode;
-    }
-    if ($isElementNode(bNode)) {
-      const bNodeDescendant = bNode.getDescendantByIndex<ElementNode>(bOffset);
-      bNode = bNodeDescendant != null ? bNodeDescendant : bNode;
-    }
-    if (aNode === bNode) {
-      return aOffset < bOffset;
+    if (this.key === b.key) {
+      return this.offset < b.offset;
```
etrepum, line 156: This is left in only as an optimization, could be removed for code size

packages/lexical/src/LexicalSelection.ts
```diff
@@ -151,23 +152,12 @@ export class Point {
...
     }
-    return aNode.isBefore(bNode);
+    const aCaret = $normalizeCaret($caretFromPoint(this, 'next'));
+    const bCaret = $normalizeCaret($caretFromPoint(b, 'next'));
+    return $comparePointCaretNext(aCaret, bCaret) < 0;
```
etrepum, line 160: Normalization is done here for backwards compatibility. PointCaret's ordering before normalization is more precise - two points can refer to the same "visual" location but be "logically" distinct. An 'element' point with the offset of a TextNode is before the a 'text' point at offset 0 to the same TextNode (because you would descend into the TextNode after encountering it as a child of the element), but these are indistinguishable visually and to the descendant-based algorithm initially used by Point.

There are cases where the Point class can still distinguish between "logical" points that are "visually" the same. For example given a textNode that has a text node sibling, the 'text' point at textNode with offset textNode.getTextContentSize() and the 'text' point at textNode.getNextSibling() with offset 0 are visually indistinguishable, but the first point isBefore the second point. This is true both before and after this fix, PointCaret normalization does not bias TextPointCaret.

ivailop7 (approved):

Issue #6998, opened 2024-12-26
Feature: Add runtime invariants for methods with unsound types: updateDOM, updateFromJSON, afterCloneFrom


etrepum:
## Description

Follow-up to #6970. There's basically a variance violation in these methods, they are only correct when `this` is not up-casted to a base class. There isn't really a straightforward typing rule we can use to prove this that has reasonable DX, but we can add a simple invariant to prove that there isn't a casting violation.

```ts
class ExtendedTextNode extends TextNode {
  afterCloneFrom(prevNode: this): void {
    super.afterCloneFrom(prevNode)
      .setNewProperty(this.getNewProperty());
  }
}

// This will throw an error about a missing `getNewProperty` method at runtime,
// but passes type checks due to the up-cast to TextNode
const extendedTextNode: TextNode = $createExtendedTextNode();
extendedTextNode.afterCloneFrom($createTextNode());
```

We can add an invariant to check this, it still won't get caught by the type checker, but will provide for better runtime errors:

```ts
class ExtendedTextNode extends TextNode {
  afterCloneFrom(prevNode: this): void {
    invariant(prevNode instanceof this.constructor, 'prevNode must be an instance of ExtendedTextNode');
    super.afterCloneFrom(prevNode)
      .setNewProperty(this.getNewProperty());
  }
}
```

## Impact

This will improve error messages for incorrect usage of the library in places where the type checker can't help.

zurfyx:
Since the introduction of `$applyNodeReplacement` I've always wanted to see it evolve into a sort of Node builder. This function solves the actual problem (node replacement) but 1) it's optional, there's no enforcement enforcement of any kind 2) it's not extensible.

A Node creation mechanism that is controlled by us allows us to do extend the creation logic seamlessly, runtime checks and optimizations.

I like that your proposal touches on this idea.

---

```
static registerNode(reg: NodeRegistration<TextNode>): void {
```

Why is this instance pre-created? I would have expected the user to create a `NodeRegistration` instance where the Node would be part of it, and then Lexical would take either a `Node` or a `NodeRegistration` class (for backward compatibility reasons). But I can also see how your solution integrates more seamlessly with the current model and potentially work best with the extended Nodes.

etrepum:
It doesn't really matter whether the NodeRegistration is created by the user (no static method requirement with that, but it is a separate identifier to export and use) or by the editor. If it's created by the user then it has to be a sort of inert abstract configuration object that editors will evaluate whenever they are constructed.

[...]

Issue #5572, opened 2024-01-31
Feature: support for a static `register` method


GermanJablo:
At https://github.com/facebook/lexical/pull/4030 @acywatson incorporated an experimental static method called `transform`. In his words:
> The fix I'm proposing here is to introduce a method to nodes that allows them to declare a transform on themselves, rather than registering it through the register API. This allows us to keep this logic encapsulated at the node level, rather than moving it to the plugin, and it should otherwise function the same.

Which has led me to think... Why not do the same with the other registersX functions?

A `register` method could be used that works just like the `mergeRegister` function in React.

This could eliminate the need for some plugins tightly tied to a particular node, such as `ImagePlugin` or `TablePlugin`. It would also not be necessary to check within the plugin if said node is registered or throw an error if not.

related: https://github.com/facebook/lexical/discussions/2739

etrepum:
With $config and extensions all of these use cases should be addressed

PR #7258, opened 2025-02-27
[WIP] Feature: $config protocol + NodeState registration/flattening

etrepum:
## Description

Based on the future needs of NodeState (#7117) - specifically allowing nodes to statically declare their required state - this RFC refactors LexicalNode and the associated internals to remove the need for any static methods, vastly reducing the boilerplate in order to subclass nodes. This is a squashed/rebased version of the original #7189 PR that explored this feature.

The short story is that this sort of thing will let us get better type-level information about nodes. We can't have `node.getType()` ever give a more specific type than `string` but we could write a `$getType(node)` that does infer to the exact type if we want. Same thing for `node.exportJSON()` and similar methods.

It also facilitates scrapping all of the node subclass boilerplate, except for this one method, so long as the class has a compatible signature (trivial constructor)

<details>
<summary>Example of extension and inference capability from the tests</summary>

```ts
const numberState = createState('numberState', {
  parse: (v) => (typeof v === 'number' ? v : 0),
});
const boolState = createState('boolState', {parse: Boolean});
class StateNode extends TestNode {
  $config() {
    return this.config('state', {
      extends: TestNode,
      stateConfigs: [{flat: true, stateConfig: numberState}, boolState],
    });
  }
  getNumber() {
    return $getState(this, numberState);
  }
  setNumber(valueOrUpdater: StateValueOrUpdater<typeof numberState>): this {
    return $setState(this, numberState, valueOrUpdater);
  }
}

const extraState = createState('extra', {parse: String});
class ExtraStateNode extends StateNode {
  $config() {
    return this.config('extra-state', {
      extends: StateNode,
      stateConfigs: [{flat: true, stateConfig: extraState}],
    });
  }
}

type _TestExtraStateNodeExportJSON = Expect<
  Equal<
    ExtraStateNodeExportJSON,
    {
      state?:
        | (Record<string, unknown> & {
            boolState?: boolean | undefined;
          })
        | undefined;
      version: number;
      type: 'extra-state';
      numberState?: number | undefined;
      extra?: string | undefined;
    }
  >
>;
```
</details>

## How it works

The proposed protocol for declaring nodes would scale down to something like this:

```ts
class CustomTextNode extends TextNode {
  $config() {
    return this.config('custom-text', {extends: TextNode});
  }
}
```

The two method interface is useful for TypeScript reasons, basically. The outer `getStaticNodeConfig` (could come up with a shorter and/or more memorable name) is the API and the `this.configureNode(…)` is a helper method that has the right generics to make it easy to return something with the right shape and type. The types are very delicate, if you don't have all of the annotations just right then the type will simplify into something without information that can be extracted. Basically it is just returning an object like this, but in a way where TypeScript doesn't collapse the type into something simpler (kind of like using `as const` or `satisfies` in the right place):

```ts
{'custom-text': {type: 'custom-text', extends: TextNode}}
```

On the LexicalEditor side, during createEditor, `CustomTextNode.prototype.$config()` is called and can grab the type and any other useful metadata (e.g. a defined transform, registered state, anything else we want to put in there). For compatibility reasons it will also monkeypatch `getType`, `importJSON`, and `clone` static methods onto the node class if they are not already there. It adds a module global variable to inject cloned node keys so it doesn't need to break any compatibility with multi-argument constructors, it only requires that they have a 0-argument form.

I've also put a `$create` method in there which reduces the boilerplate and increases the efficiency there, assuming that we'd be willing to deprecate `with` in the node overrides and go directly to `withKlass` so that we don't have to construct any intermediate garbage when that's configured.

I think this covers most of the breaking-ish changes I would want to make to the lowest level stuff, in exchange for better DX and no loss of performance (probably better, even if just code size improves).

## Future direction

I think this is where we fix the rest of the #6998 things here with some combination of middleware-style functions or automatic super calls (like `$afterCloneFrom`) instead of methods so that we can control things better without variance violations.

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -27,8 +27,8 @@ import {
   NodeKey,
   SerializedElementNode,
   Spread,
+  StaticNodeConfigRecord,
 } from 'lexical';
-import invariant from 'shared/invariant';
 import normalizeClassNames from 'shared/normalizeClassNames';
 
 import {$createListItemNode, $isListItemNode, ListItemNode} from '.';
@@ -59,6 +59,19 @@ export class ListNode extends ElementNode {
   __start: number;
   /** @internal */
   __listType: ListType;
+
+  /** @internal */
+  $config(): StaticNodeConfigRecord<
+    'list',
+    {$transform: (node: ListNode) => void}
+  > {
+    return this.config('list', {
+      $transform: (node: ListNode): void => {
+        mergeNextSiblingListIfSameType(node);
+        updateChildrenListItemValue(node);
+      },
+    });
+  }
 
   static getType(): string {
     return 'list';
@@ -127,14 +140,6 @@ export class ListNode extends ElementNode {
     $setListThemeClassNames(dom, config.theme, this);
 
     return false;
-  }
-
-  static transform(): (node: LexicalNode) => void {
-    return (node: LexicalNode) => {
-      invariant($isListNode(node), 'node is not a ListNode');
-      mergeNextSiblingListIfSameType(node);
-      updateChildrenListItemValue(node);
-    };
   }
 
   static importDOM(): DOMConversionMap | null {
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -23,8 +23,16 @@ import {
   $isTextNode,
   type DecoratorNode,
   ElementNode,
+  NODE_STATE_KEY,
 } from '.';
-import {$updateStateFromJSON, type NodeState} from './LexicalNodeState';
+import {PROTOTYPE_CONFIG_METHOD} from './LexicalConstants';
+import {
+  $updateStateFromJSON,
+  type NodeState,
+  type NodeStateJSON,
+  type Prettify,
+  type RequiredNodeStateConfig,
+} from './LexicalNodeState';
 import {
   $getSelection,
   $isNodeSelection,
@@ -48,6 +56,7 @@ import {
   $setNodeKey,
   $setSelection,
   errorOnInsertTextNodeOnRoot,
+  getRegisteredNode,
   internalMarkNodeAsDirty,
   removeFromParent,
 } from './LexicalUtils';
@@ -62,8 +71,148 @@ export type SerializedLexicalNode = {
   type: string;
   /** A numeric version for this schema, defaulting to 1, but not generally recommended for use */
   version: number;
+  /**
+   * Any state persisted with the NodeState API that is not
+   * configured for flat storage
+   */
   [NODE_STATE_KEY]?: Record<string, unknown>;
 };
+
+/**
+ * EXPERIMENTAL
+ * The configuration of a node returned by LexicalNode.$config()
+ *
+ * @example
+ * ```ts
+ * class CustomText extends TextNode {
+ *   $config() {
+ *     return this.config('custom-text', {extends: TextNode}};
+ *   }
+ * }
+ * ```
+ */
+export interface StaticNodeConfigValue<
+  T extends LexicalNode,
+  Type extends string,
+> {
+  /**
+   * The exact type of T.getType(), e.g. 'text' - the method itself must
+   * have a more generic 'string' type to be compatible wtih subclassing.
+   */
+  readonly type?: Type;
+  /**
+   * An alternative to the internal static transform() method
+   * that provides better DX
+   */
+  readonly $transform?: (node: T) => void;
+  /**
+   * EXPERIMENTAL
+   *
+   * An array of RequiredNodeStateConfig to initialize your node with
+   * its state requirements. This may be used to configure serialization of
+   * that state.
+   *
+   * This function will be called (at most) once per editor initialization,
+   * directly on your node's prototype. It must not depend on any state
+   * initialized in the constructor.
+   *
+   * @example
+   * ```ts
+   * const flatState = createState("flat", {parse: parseNumber});
+   * const nestedState = createState("nested", {parse: parseNumber});
+   * class MyNode extends TextNode {
+   *   $config() {
+   *     return this.config(
+   *       'my-node',
+   *       {
+   *         extends: TextNode,
+   *         stateConfigs: [
+   *           { stateConfig: flatState, flat: true},
+   *           nestedState,
+   *         ]
+   *       },
+   *     );
+   *   }
+   * }
+   * ```
+   */
+  readonly stateConfigs?: readonly RequiredNodeStateConfig[];
+  /**
+   * If specified, this must be the exact superclass of the node. It is not
+   * checked at compile time and it is provided automatically at runtime.
+   *
+   * You would want to specify this when you are extending a node that
+   * has non-trivial configuration in its $config such
+   * as required state. If you do not specify this, the inferred
+   * types for your node class might be missing some of that.
+   */
+  readonly extends?: Klass<LexicalNode>;
+}
+
+/**
+ * This is the type of LexicalNode.$config() that can be
+ * overridden by subclasses.
+ */
+export type BaseStaticNodeConfig = {
+  readonly [K in string]?: StaticNodeConfigValue<LexicalNode, string>;
+};
+
+/**
+ * Used to extract the node and type from a StaticNodeConfigRecord
+ */
+export type StaticNodeConfig<
+  T extends LexicalNode,
+  Type extends string,
+> = BaseStaticNodeConfig & {
+  readonly [K in Type]?: StaticNodeConfigValue<T, Type>;
+};
+
+/**
+ * Any StaticNodeConfigValue (for generics and collections)
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type AnyStaticNodeConfigValue = StaticNodeConfigValue<any, any>;
+
+/**
+ * @internal
+ *
+ * This is the more specific type than BaseStaticNodeConfig that a subclass
+ * should return from $config()
+ */
+export type StaticNodeConfigRecord<
+  Type extends string,
+  Config extends AnyStaticNodeConfigValue,
+> = BaseStaticNodeConfig & {
+  readonly [K in Type]?: Config;
+};
+
+/**
+ * Extract the type from a node based on its $config
+ *
+ * @example
+ * ```ts
+ * type TextNodeType = GetStaticNodeType<TextNode>;
+ *      // ? 'text'
+ * ```
+ */
+export type GetStaticNodeType<T extends LexicalNode> = ReturnType<
+  T[typeof PROTOTYPE_CONFIG_METHOD]
+> extends StaticNodeConfig<T, infer Type>
+  ? Type
+  : string;
+
+/**
+ * The most precise type we can infer for the JSON that will
+ * be produced by T.exportJSON().
+ *
+ * Do not use this for the return type of T.exportJSON()! It must be
+ * a more generic type to be compatible with subclassing.
+ */
+export type LexicalExportJSON<T extends LexicalNode> = Prettify<
+  Omit<ReturnType<T['exportJSON']>, 'type'> & {
+    type: GetStaticNodeType<T>;
+  } & NodeStateJSON<T>
+>;
 
 /**
  * Omit the children, type, and version properties from the given SerializedLexicalNode definition.
@@ -236,6 +385,40 @@ export class LexicalNode {
       'LexicalNode: Node %s does not implement .clone().',
       this.name,
     );
+  }
+
+  /**
+   * Override this to implement the new static node configuration protocol,
+   * this method is called directly on the prototype and must not depend
+   * on anything initialized in the constructor. Generally it should be
+   * a trivial implementation.
+   *
+   * @example
+   * ```ts
+   * class MyNode extends TextNode {
+   *   $config() {
+   *     return this.config('my-node', {extends: TextNode});
+   *   }
+   * }
+   * ```
+   */
+  $config(): BaseStaticNodeConfig {
+    return {};
+  }
+
+  /**
+   * This is a convenience method for $config that
+   * aids in type inference. See {@link LexicalNode.$config}
+   * for example usage.
+   */
+  config<Type extends string, Config extends StaticNodeConfigValue<this, Type>>(
+    type: Type,
+    config: Config,
+  ): StaticNodeConfigRecord<Type, Config> {
+    const parentKlass =
+      config.extends || Object.getPrototypeOf(this.constructor);
+    Object.assign(config, {extends: parentKlass, type});
+    return {[type]: config} as StaticNodeConfigRecord<Type, Config>;
   }
 
   /**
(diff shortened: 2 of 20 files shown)


zurfyx (commented):
I quite like this proposal this API ticks the inheritance and type-safe customizability boxes, thank you Bob as usual!

I'm not a fan of the repeated `extends` but The major downside is the fact that this emphasizes JSON structures as opposed to inheritance. This is a shift from our typical paradigm that relies on class based structure. While arguably this `$config` function is inheritable, the definition of the all properties within this `$config` happens declarative as part of the initialization, what would have typically been additional properties on the Node.

To put the code in perspective, this is the API proposed in this PR:
```
return this.config('listitem', {
      $transform: (node: ListItemNode): void => { ... },
      stateConfigs: [
         {flat: true, stateConfig: numberState},
         boolState],
      ],
});
```

For the sake of `$create` I don't think this API is needed but I can see how this help, and I would definitely be keen to move toward this direction to double down on the idea behind `$applyNodeReplacement` that is currently very specific to that job in particular.

packages/lexical/src/LexicalNode.ts
```diff
@@ -67,9 +75,149 @@ export type SerializedLexicalNode = {
...
+  /**
+   * The exact type of T.getType(), e.g. 'text' - the method itself must
+   * have a more generic 'string' type to be compatible wtih subclassing.
+   */
+  readonly type?: Type;
+  /**
+   * An alternative to the internal static transform() method
+   * that provides better DX
+   */
+  readonly $transform?: (node: T) => void;
```
GermanJablo, line 111: Maybe this is a good opportunity to do this: https://github.com/facebook/lexical/issues/5572
I don't see why transforms could be registered in the node definition but not any other type of listener or command.

packages/lexical/src/LexicalNode.ts
```diff
@@ -243,6 +391,40 @@ export class LexicalNode {
...
+   * This is a convenience method for $config that
+   * aids in type inference. See {@link LexicalNode.$config}
+   * for example usage.
+   */
+  config<Type extends string, Config extends StaticNodeConfigValue<this, Type>>(
+    type: Type,
+    config: Config,
+  ): StaticNodeConfigRecord<Type, Config> {
+    const parentKlass =
+      config.extends || Object.getPrototypeOf(this.constructor);
```
GermanJablo, line 423: Why is the superclass repeated in 2 places? Couldn't it be lookup'd in the prototype chain?

GermanJablo:
So, if I understand correctly, "flat" would be to be able to use this API on existing properties in a backwards compatible way, right?

packages/lexical/src/LexicalNode.ts
etrepum: Having transform here is mostly because some nodes won’t work correctly at all without their transform, and that there’s already a static method for it. I would prefer to revisit that after we add extensions, since they also cover that functionality.

nodes should really be minimal and only include functionality relevant to the lexical document, IMO they shouldn’t even have DOM code in them, so would be better to not add more to them today. SSR, headless, React Native, Reacf-only rendering, etc. use cases would be easier if nodes didn’t have extra stuff.

etrepum:
Yes “flat” moves a property out of the $ key to the root of the node’s JSON, allowing for backwards/forwards compatibility of the serialization

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -78,8 +88,15 @@ export class ListNode extends ElementNode {
     this.__start = start;
   }
 
+  afterCloneFrom(prevNode: this): void {
+    super.afterCloneFrom(prevNode);
+    this.__listType = prevNode.__listType;
+    this.__tag = prevNode.__tag;
+    this.__start = prevNode.__start;
+  }
```
zurfyx, line 96: I understand that this is to fulfil the constructor on the automated clone mechanism. What is your suggestion to get rid of this? Is it possible to move these to state props?

packages/lexical-website/docs/concepts/nodes.mdx
```diff
@@ -343,3 +342,467 @@ export function $isVideoNode(
...
+
+export class ColoredNode extends TextNode {
+  // highlight-start
+  $config() {
+    return this.config('colored', {
+      extends: TextNode,
+      // This defines the serialization of the color NodeState as
+      // a flat property of the SerializedLexicalNode JSON instead of
+      // nesting it in the '$' property
+      stateConfigs: [{flat: true, stateConfig: colorState}],
```
zurfyx, line 522: Can we describe when a flat property is preferred vs `$` properties alongside with their representation covered in the tests?
```
[NODE_STATE_KEY]: {boolState: true},
  numberState: 2,
  type: 'state',
```

packages/lexical-website/docs/concepts/nodes.mdx
```diff
@@ -343,3 +342,467 @@ export function $isVideoNode(
...
+  decorate(): ReactNode {
+    // highlight-next-line
+    return <VideoPlayer videoID={$getState(this, idState)} />;
+  }
+}
+
+export function $createVideoNode(id: string): VideoNode {
+  // highlight-next-line
+  return $setState($create(VideoNode), idState, id);
+}
```
zurfyx, line 720: I wonder if we should promote paramless properties as well for creation functions, but I haven't thought too deeply about this

zurfyx (commented):
Thanks Bob! Do you think we can get the Flow types in before merging?

packages/lexical-list/src/LexicalListNode.ts
etrepum: These could absolutely be NodeState to eliminate the boilerplate, but since this class (and those properties) is part of the public lexical-list API and is used with lexical-yjs there are additional considerations for how that migration from properties-to-state would work

packages/lexical-website/docs/concepts/nodes.mdx
etrepum: The motivation here is to eliminate the boilerplate if you're migrating from an existing serialization that uses flat properties (e.g. in a property-to-state migration). There's also the aesthetic reason why someone would rather have all of the statically known properties be top-level in the serialization.

packages/lexical-website/docs/concepts/nodes.mdx
etrepum: There isn't really a reason to write a boilerplate parameterless creation function at all since you have `$create(NodeClass)` which is even the same number of characters as `$createNodeClass()`. Ensuring that all classes have a parameterless constructor and ideally avoiding direct use of the `new` operator (so that the `$applyNodeReplacement` equivalent can happen before any construction happens) are the key things

etrepum:
@zurfyx I will write the flow types when/if there's general buy-in for this API, ideally that step would be the last iteration. Are you happy enough with how it looks for that work to start?

zurfyx (approved):

lilshady:
Hey @etrepum @zurfyx  I have a question regarding this change. 
[Here](https://github.com/facebook/lexical/blob/960adf903067747cb3d1ef7389e32bb5f34f9072/packages/lexical/src/LexicalNode.ts#L412) we have an invariant that all nodes have to implement `getType`. 
Along with this PR, do we have to change this invariant?  e.g. provide a default implementation for it?

etrepum:
The data returned by $config is used to implement a boilerplate static getType for backwards compatibility when the node is registered with an editor.

lilshady:
> The data returned by $config is used to implement a boilerplate static getType for backwards compatibility when the node is registered with an editor.

Yes, but I see in the `getStaticNodeConfig`, we also call the [getType](https://github.com/facebook/lexical/blob/960adf903067747cb3d1ef7389e32bb5f34f9072/packages/lexical/src/LexicalUtils.ts#L2028)? 
Could it be that `getType` is being invoked before the injection process? 
Correct me if I am wrong: My understanding is that we're using `getType` to retrieve the associated nodeConfig(but it seems that getType has been removed). Then injecting the getType implementation with () => ownNodeType.

Context is that I do see some tests are failing at meta internally, the stack trace is like:
```
 at formatDevErrorMessage (html/shared/lexical/Lexical.dev.js:32:7)
      at Function.getType (html/shared/lexical/Lexical.dev.js:3223:1)
      at getStaticNodeConfig (html/shared/lexical/Lexical.dev.js:12202:57)
      at _loop2 (html/shared/lexical/Lexical.dev.js:10008:1)
      at Object.createEditor (html/shared/lexical/Lexical.dev.js:10056:59)
```

etrepum:
getType is only called there when the node class directly implements it (the hasOwn check) and it is not one of the three abstract base classes (DecoratorNode, ElementNode, and LexicalNode). I can't really debug this without seeing the full context, possibly the haste compiler at FB is doing something non-standard when defining classes in that environment? This code is assuming that classes work according to the ES specification.

```ts
class Base { static foo() { return 'base'; } }
class Child extends Base { }
```

```
> Child.foo()
'base'
> Object.hasOwn(Child, 'foo')
false
> Object.hasOwn(Base, 'foo')
true
```

lilshady:
> getType is only called there when the node class directly implements it (the hasOwn check) and it is not one of the three abstract base classes (DecoratorNode, ElementNode, and LexicalNode). I can't really debug this without seeing the full context, possibly the haste compiler at FB is doing something non-standard when defining classes in that environment? This code is assuming that classes work according to the ES specification.
> 
> ```ts
> class Base { static foo() { return 'base'; } }
> class Child extends Base { }
> ```
> 
> ```
> > Child.foo()
> 'base'
> > Object.hasOwn(Child, 'foo')
> false
> > Object.hasOwn(Base, 'foo')
> true
> ```

Thanks! Let me debug it further.

lilshady:
> getType is only called there when the node class directly implements it (the hasOwn check) and it is not one of the three abstract base classes (DecoratorNode, ElementNode, and LexicalNode). I can't really debug this without seeing the full context, possibly the haste compiler at FB is doing something non-standard when defining classes in that environment? This code is assuming that classes work according to the ES specification.
> 
> ```ts
> class Base { static foo() { return 'base'; } }
> class Child extends Base { }
> ```
> 
> ```
> > Child.foo()
> 'base'
> > Object.hasOwn(Child, 'foo')
> false
> > Object.hasOwn(Base, 'foo')
> true
> ```

Yes,  the root cause is  that `hasOwn` returns `True` instead of `False` in this case. Still figuring it out how we can address this issue at meta.

etrepum:
Hard to say what the right workaround would be without knowing exactly how far Meta's implementation strays from the standard but if it respects the prototype chain correctly then something like this would likely work:

```ts
export function hasOwn(o: object, k: string): boolean {
  return (
    Object.prototype.hasOwnProperty.call(o, k) &&
    // workaround for meta
    o[k as keyof typeof o] !== Object.getPrototypeOf(o)[k]
  );
}
```

etrepum:
Of course this is only a valid workaround in the limited context for which this is used, so maybe it should be renamed to `hasOwnMethodOverride` or something like that so that it's clear that it's not a general ponyfill for Object.hasOwn

lilshady:
> Hard to say what the right workaround would be without knowing exactly how far Meta's implementation strays from the standard but if it respects the prototype chain correctly then something like this would likely work:
> 
> ```ts
> export function hasOwn(o: object, k: string): boolean {
>   return (
>     Object.prototype.hasOwnProperty.call(o, k) &&
>     // workaround for meta
>     o[k as keyof typeof o] !== Object.getPrototypeOf(o)[k]
>   );
> }
> ```

Thank you for providing this workaround. However, after further debugging, I've found that it doesn't correctly respect the prototype chain. Specifically,` Object.getPrototypeOf(o)[k]` returns undefined at meta(I guess some internal optimizations involved here ). I'm considering an alternative approach.
> ```ts
> export function hasOwnMethodOverride(o: object, k: string): boolean {
>   return (
>     Object.prototype.hasOwnProperty.call(o, k) &&
>     // workaround for meta
>    && o[k as keyof typeof o] !== LexicalNode[k]
>   );
> }
> ```

And another issue is 

```ts
 if (__DEV__) {
        invariant(
          klass.length === 0,
          '%s (type %s) must implement a static clone method since its constructor has %s required arguments (expecting 0). Use an explicit default in the first argument of your constructor(prop: T=X, nodeKey?: NodeKey).',
          klass.name,
          ownNodeType,
          String(klass.length),
        );
      }
```
At meta, it  doesn't ignore default values and optional arguments, resulting in incorrect counts (e.g., ListNode returns 3 instead of 0). I'm considering downgrading this to a warning, although it's not ideal as it deviates from the `fail fast` principle.
 If you have any suggestions or alternative approaches, please let me know. Thank you! 
cc @zurfyx

etrepum:
Your implementation of hasOwnMethod only works for static methods. The implementation uses hasOwn for both instance and static methods. Would need a separate workaround for instance methods.

I don't know if there's a good solution for the clone warning. If I couldn't fix meta's compilers/optimizations to work according to spec I would just turn the warning off for meta and do that check at lint time rather than runtime. Should be straightforward to implement a lint to check that all constructor arguments are optional or have defaults. The warning is useful for open source because not everyone uses the lint and it saves us from getting bug reports about user error.

lilshady:
How about we have two separate methods, (For instance method, I believe meta behaves the same):
```ts
function hasOwnStaticMethod(o, k) {
  return (Object.prototype.hasOwnProperty.call(o, k) && o[k as keyof typeof o] !== LexicalNode[k]);
}

function hasOwnInstanceMethod(o, k) {
  return Object.prototype.hasOwnProperty.call(o, k);
}
```
only for decorate, we will use `hasOwnInstanceMethod`, for other places we will use `hasOwnStaticMethod`?

etrepum:
That should work fine

lilshady:
Regarding the clone warning, I debugged for a while and I don't think I can change the behavior at Meta. 
Since it's an invariant in the dev environment, it will cause the entire page to break instead of displaying a warning or alert that can be turned off. 
I know changing it to console.warn instead of an invariant seems a bad idea, but can't figure out another solution at this moment.

etrepum:
Why not add a variable like `__META_ONLY__` that you can use to isolate workarounds for the bugs in Meta's tool chain?

zurfyx:
@etrepum for reference, the root of the problem appears to be our use of `babel/helpers/inheritLoose` (which is applied automatically).

This function is, as the name suggests, is intentionally loose and bubbles up properties that otherwise would be discovered via the prototype chain.

i.e.

```
class Base { static foo() { return 'base'; } }
class Child extends Base { }
```

```
Base=function(){function Base(){}Base.
foo=function foo(){
return'base';
};return Base;}();var Child=function(_Base){babelHelpers.inheritsLoose(Child,_Base);function Child(){return _Base.apply(this,arguments)||this;}return Child;}(Base);
```

A flag can work, at the end of the day we already have our own build script. I wonder if there's an alternative besides having to fork and maintain two versions of the logic.

etrepum:
The inheritance chain issue is sufficiently solved by checking value equality with the LexicalNode or LexicalNode.prototype directly. It won't handle warning in all of the edge cases where you're subclassing a subclass that isn't compliant with the newer protocols but that's not really worse than today.

The clone invariant is something that can probably only be handled with a variable injected by the build. There's some chance that it can be detected at runtime by including a sample of a constructor with default args and measuring its length but that will only work when lexical and the code using lexical are compiled in the same way.

etrepum:
I've created a PR #7659 that may address these issues

vchirikov:
this api looks like hooks in react and do more confusion vs clean class concept 😕. Also by the docs $config uses NodeState API which `This API is experimental` according to the docs.

Will you support old `Legacy static methods and properties` way in the future or it will be completely removed in next versions?

etrepum:
The static methods will continue to be supported for the foreseeable future.

`$config` does not depend on NodeState, it provides extra functionality for NodeState that is not possible with the static methods.

PR #7264, opened 2025-02-28
[@lexical/react]  Bug Fix: anchor element not cleanup when component unmount

yuanyan:
## Description

anchor element not cleanup when component unmount

image


etrepum (approved):
Confirmed that this bug exists in the playground (can be reproduced easily by toggling the context menu setting on and off) and that this PR fixes the issue. Nice catch! Will merge after the e2e test suite finishes running, assuming that it's still passing.

etrepum (approved):

PR #7271, opened 2025-03-02
[lexical] Bug Fix: invalid import from self

AlessioGr:
Updating to 0.26.0 causes our project to crash executing a standalone node script that imports from `lexical` (this cannot be reproduced when importing lexical from a bundler).

The reason for that is this import that made it into the lexical build:

CleanShot 2025-03-01 at 21 30 26@2x

This is caused by the following newly introduced import statement in `packages/lexical/src/LexicalNode.ts`:

`import {Klass, KlassConstructor, NODE_STATE_KEY} from 'lexical';`

Imports from self like that are not valid and cause issues. I have verified that removing it fixes our issue.

Notably, such self-import statements can easily slip in (especially with VS Code’s auto-import suggestions). To help guard against this, we’ve created a custom ESLint rule ([no-imports-from-self](https://github.com/payloadcms/payload/blob/main/packages/eslint-plugin/customRules/no-imports-from-self.js)) that warns developers if they accidentally import from the same package. I think it would be a good idea to introduce something like that in this monorepo - feel free to copy this rule or use our `@payloadcms/eslint-plugin` package if interested.

## To Reproduce

1. Create a script.js:

```ts
import { TabNode } from 'lexical'
console.log(TabNode)
```

2. Run it using `node src/script.js`
3. The following error will be thrown:

```
Warning: Detected unsettled top-level await at
file:///Users/alessio/Documents/GitHub/payload20/node_modules/.pnpm/lexical@0.26.0/node_modules/lexical/Lexical.node.mjs:9
const mod = await (process.env.NODE_ENV !== 'production' ? import('./Lexical.dev.mjs') : import('./Lexical.prod.mjs'));
```

### DIFF
--- a/packages/lexical/src/LexicalCommands.ts
+++ b/packages/lexical/src/LexicalCommands.ts
@@ -6,13 +6,11 @@
  *
  */
 
-import type {
-  BaseSelection,
-  ElementFormatType,
-  LexicalCommand,
-  LexicalNode,
-  TextFormatType,
-} from 'lexical';
+import type {LexicalCommand} from './LexicalEditor';
+import type {LexicalNode} from './LexicalNode';
+import type {BaseSelection} from './LexicalSelection';
+import type {ElementFormatType} from './nodes/LexicalElementNode';
+import type {TextFormatType} from './nodes/LexicalTextNode';
 
 export type PasteCommandType = ClipboardEvent | InputEvent | KeyboardEvent;
 
--- a/packages/lexical/src/LexicalNode.ts
+++ b/packages/lexical/src/LexicalNode.ts
@@ -7,10 +7,14 @@
  */
 
 /* eslint-disable no-constant-condition */
-import type {EditorConfig, LexicalEditor} from './LexicalEditor';
+import type {
+  EditorConfig,
+  Klass,
+  KlassConstructor,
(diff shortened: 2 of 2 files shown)


etrepum (approved):
Not sure how you would run into problems with this unless you were not using the build artifacts from lexical? Avoiding this problem in the future is only feasible if you can also provide a unit test of some kind that can detect it.

etrepum:
Never mind, I can see how this causes node problems now. Please go ahead and submit a PR that adds the appropriate lint rules or tests to ensure this doesn’t reoccur.

AlessioGr:
Will do, thank you for the speedy reply! And yea, adding a unit test for this will be tricky, as this only happens when importing from the built, installed version of lexical.

A unit test would need to:

1. Build the lexical package
2. Run `npm pack`
3. Move the tgz file into some test directory
4. modify the package.json to install it
5. Run `npm install`
6. Then it can execute the script

Given that this should happen in _any_ node script that imports `lexical` outside a bundler, would it be possible to get an early `0.26.1` release?

etrepum:
We already have some tests that do that sort of thing (e.g. the integration tests), so it’s not something too crazy.

Yes, there will be a patch release next week.

Issue #5381, opened 2023-12-13
Bug: prismjs is included as a dependency of @lexical/markdown


markulrich:
It appears that prismjs gets included as a (very large) dependency of @lexical/markdown, swelling the bundle size.

## Steps To Reproduce

See https://bundlephobia.com/package/@lexical/markdown@0.12.5 and that prismjs swells the bundle size

## The current behavior

Prismjs is included in the bundle

## The expected behavior

Prismjs is not included, and the bundle size is <100kb

markulrich:
Also see note from @acywatson at https://github.com/facebook/lexical/discussions/3778#discussioncomment-6694006. Thank you!!

vadimkantorov:
Btw, just curious, what is the expected `ls node_modules | wc -l` after `npm install` of a freshly-cloned https://github.com/facebook/lexical ?

For me this initial `npm install` led to more than 1000...

etrepum:
I think this is really just bundlephobia telling you what the worst case is, if you were to use every single thing with no tree-shaking. Can you demonstrate that this actually happens in practice, with a real bundler in an application that is not using code nodes?

vadimkantorov:
No, but I am also not a web developer and now knowledgeable of proper use of bundlers etc. So the mere number of fetched modules scared me and I was wondering what are some typical bundle size and if I'm doing anything wrong with this 1000 of modules...

etrepum:
I was commenting on the original post, not your off-topic reply.

AlessioGr:
Including Prism in @lexical/markdown caused a messed up issue for us where Node would randomly exit the process if @lexical/markdown was imported dynamically (or any dynamically imported module in the chain imported from @lexical/markdown). Getting rid of the `@lexical/code` import fixed it for us: https://github.com/payloadcms/payload/pull/9382.

Wondering if anyone else has experienced this issue

etrepum:
I'm sure that could happen depending on how your bundler omits modules and what kind of exception handling your code has, but the anecdote isn't really of much use without a repro. Of course the best thing to do would be to refactor things accordingly to make prism explicitly optional, the next best thing would be to demonstrate a working bundler configuration that excludes the majority of prism (e.g. the languages).

AlessioGr:
[...]

PR #7272, opened 2025-03-02
chore: add no-imports-from-self eslint rule to monorepo

AlessioGr:
This adds a new `no-imports-from-self` eslint rule to this monorepo, which will prevent regression of https://github.com/facebook/lexical/pull/7271 - cc @etrepum

CleanShot 2025-03-01 at 23 17 16@2x

The rule is taken from [here](https://github.com/payloadcms/payload/blob/main/packages/eslint-plugin/customRules/no-imports-from-self.js). We cannot install `@payloadcms/eslint-plugin` directly, as that's an esm-only package which would require us to move to the flat eslint config.

I added the rule to scripts/eslint, where we can also add more custom monorepo-only eslint rules in the future. Could move it to the monorepo root, but I didn't want to bloat it with yet another folder.

## Todo

Runnin npm lint shows that there are 110 violations of this rule - mostly in tests:

CleanShot 2025-03-01 at 23 25 51@2x

I will fix those imports shortly

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -161,6 +161,7 @@ module.exports = {
     'no-only-tests',
     'lexical',
     '@lexical',
+    'lexical-monorepo',
   ],
 
   // Stop ESLint from looking for a configuration file in parent folders
@@ -169,7 +170,6 @@ module.exports = {
   // and then enable some React specific ones.
   rules: {
     'accessor-pairs': OFF,
-
     'consistent-return': OFF,
     curly: [ERROR, 'all'],
     'dot-location': [ERROR, 'property'],
@@ -199,6 +199,8 @@ module.exports = {
     'jsx-quotes': [ERROR, 'prefer-double'],
 
     'keyword-spacing': [ERROR, {after: true, before: true}],
+
+    'lexical-monorepo/no-imports-from-self': ERROR,
 
     // Enforced by Prettier
     // TODO: Prettier doesn't handle long strings or long comments. Not a big
--- a/scripts/eslint/no-imports-from-self.js
+++ b/scripts/eslint/no-imports-from-self.js
@@ -1,0 +1,92 @@
+/**
+ * From: https://github.com/payloadcms/payload/blob/main/packages/eslint-plugin/customRules/no-imports-from-self.js
+ *
+ * MIT License
+ *
+ * Copyright (c) 2018-2025 Payload CMS, Inc. <info@payloadcms.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * 'Software'), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+'use strict';
+
+const fs = require('fs');
+const path = require('node:path');
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  create(context) {
+    let packageName = null;
+
+    return {
+      ImportDeclaration(node) {
+        const importPath = node.source.value;
+        const pkgName = getPackageName(context, packageName);
+        if (pkgName && importPath.startsWith(pkgName)) {
+          context.report({
+            message: `Package "${pkgName}" should not import from itself. Use relative instead.`,
+            node,
+          });
+        }
+      },
+    };
+  },
+
+  meta: {
+    docs: {
+      category: 'Best Practices',
+      description: 'Disallow a package from importing from itself',
+      recommended: true,
+    },
+    schema: [],
+  },
+};
+
+/**
+ * @param {import('eslint').Rule.RuleContext} context
+ * @param {string|undefined} packageName
+ */
+function getPackageName(context, packageName) {
+  if (packageName) {
+    return packageName;
+  }
+
+  const fileName = context.getFilename();
+  const pkg = findNearestPackageJson(path.dirname(fileName));
+  if (pkg) {
+    return pkg.name;
+  }
+}
+
+/**
+ * @param {string} startDir
+ */
+function findNearestPackageJson(startDir) {
+  let currentDir = startDir;
+  while (currentDir !== path.dirname(currentDir)) {
+    // Root directory check
+    const pkgPath = path.join(currentDir, 'package.json');
+    if (fs.existsSync(pkgPath)) {
+      const pkgContent = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+      return pkgContent;
+    }
+    currentDir = path.dirname(currentDir);
+  }
+  return null;
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,6 +58,7 @@
         "eslint-plugin-jest": "^28.5.0",
         "eslint-plugin-jsx-a11y": "^6.8.0",
         "eslint-plugin-lexical": "file:./eslint-plugin",
+        "eslint-plugin-lexical-monorepo": "file:./scripts/eslint",
         "eslint-plugin-no-function-declare-after-return": "^1.1.0",
         "eslint-plugin-no-only-tests": "^3.1.0",
         "eslint-plugin-react": "^7.34.1",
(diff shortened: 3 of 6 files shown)


etrepum (commented):
I think this needs to be modified a bit to work for this repo.

Tests and private packages (shared, lexical-devtools, lexical-playground) should be excluded from this rule, they aren't a build artifact and shouldn't be subject to these rules.

Type-only imports (`import type`) shouldn't matter as they are erased and the .d.ts files that are produced are ignored by the runtime (and typescript doesn't have this class of issue).

etrepum:
lexical-react is also correct in importing '@lexical/react/X' from '@lexical/react/Y', if these were relative imports then code will be duplicated across the multiple module entrypoints which is not really desirable. The rule should probably be about modules and not packages (although this is currently the only package that has multiple entrypoints so it could be special cased in a fairly straightforward way).

AlessioGr:
> lexical-react is also correct in importing '@lexical/react/X' from '@lexical/react/Y', if these were relative imports then code will be duplicated across the multiple module entrypoints which is not really desirable

Yep, you're right. I tested it, and node seems to handle the @lexical/react/X and @lexical/react/Y self-imports just fine. It appears the problematic self-import in the previous issue might have only caused problems because it was a recursive reference, which isn’t the case for @lexical/react/X, since those bundles are separate.

Interestingly, the same issue occurs with @lexical/code. A few weeks back, I [commented about @lexical/markdown breaking if it imports @lexical/code](https://github.com/facebook/lexical/issues/5381#issuecomment-2566745990), and I initially thought it was due to Prism. But it turns out it was another self-reference within @lexical/code. This brings a lot of clarity! What a nightmare to debug if it's unclear that the issue stems from self-imports hah

I’m going to open a separate PR to remove the remaining invalid self-imports first, so we can hopefully include those fixes in next week’s release.

AlessioGr:
Nevermind that, just saw that the `@lexical/code` self-import is type-only. The `@lexical/code` issue occurs because it is importing from `lexical`, which had that self-import issue. Must have been something unrelated then (although the error message, "Warning: Detected unsettled top-level await at", was the same)

.eslintrc.js
```diff
@@ -161,6 +161,7 @@ module.exports = {
     'no-only-tests',
     'lexical',
```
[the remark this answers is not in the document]
AlessioGr: Thanks for clarifying! Done

scripts/eslint/no-imports-from-self.js
```diff
@@ -0,0 +1,92 @@
+/**
```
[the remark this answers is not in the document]
AlessioGr: Could you confirm that this can't be accepted if we include both the Payload MIT and Meta MIT license headers? I updated it to mention Payload under `Portions Copyright`, which I believe should satisfy the MIT requirements given that we share the same license.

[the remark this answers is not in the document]
etrepum: I don’t work at Meta, am not a lawyer, and haven’t read the CLA recently so I can’t answer that. The PR will probably sit on hold for a while until someone at Meta can figure out whether it’s acceptable or not if it doesn’t follow the standard license header template (even if compliant with MIT). Basically if payload can assign copyright to Meta for this file it will look like everything else and can be approved more quickly.

eslint-plugin/src/rules/no-imports-from-self.js
```diff
@@ -0,0 +1,83 @@
...
+ * From: https://github.com/payloadcms/payload/blob/main/packages/eslint-plugin/customRules/no-imports-from-self.js
+ *
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * Portions Copyright (c) Payload CMS, Inc. <info@payloadcms.com>
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+// eslint-disable-next-line header/header
```
etrepum, line 11: I think this will probably sit in review for at least a few extra days, possibly weeks, unless the copyright assignment and header is consistent with the rest of the project. The CLA does state:

> 2. Grant of Copyright License. Subject to the terms and conditions of this Agreement, You hereby grant to Meta and to recipients of software distributed by Meta a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable copyright license to reproduce, prepare derivative works of, publicly display, publicly perform, sublicense, and distribute Your Contributions and such derivative works.

The attribution for the origin and Payload CMS could be elsewhere in the code, but I think having it here is going to cause some delays because it doesn't look like a normal copyright assignment per the CLA.

If you can't do this then I can look at implementing something equivalent to this from scratch when I have time which wouldn't be subject to copyright issues.

eslint-plugin/src/rules/no-imports-from-self.js
```diff
@@ -0,0 +1,83 @@
...
+
+    return {
+      ImportDeclaration(node) {
+        // Skip type-only imports
+        if (node.importKind === 'type' || node.importKind === 'typeof') {
+          return;
+        }
+
+        const importPath = node.source.value;
+        const pkgName = getPackageName(context, packageName);
```
etrepum, line 30: Since we're reading the package.json anyway it probably makes sense to use that to determine if it's in a private package or not rather than explicitly listing them all in configuration. The tests of course still have to be excluded specifically.

eslint-plugin/src/rules/no-imports-from-self.js
```diff
@@ -0,0 +1,83 @@
...
+// eslint-disable-next-line header/header
+'use strict';
+
+const fs = require('fs');
+const path = require('node:path');
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  create(context) {
+    let packageName = null;
```
etrepum, line 20: This cache doesn't appear to be used, which means it's traversing the filesystem to find and read package.json on every import

.eslintrc.js
```diff
@@ -114,6 +114,18 @@ module.exports = {
         'lexical/no-optional-chaining': OFF,
       },
     },
+    {
+      files: [
+        'packages/lexical-playground/**',
+        'packages/lexical-devtools/**',
+        'packages/lexical-react/**',
```
etrepum, line 121: lexical-react could have similar problems, there are still entrypoints with fork modules, it's really just the rule's implementation that's wrong because the actual problem derives from modules that import themselves.

eslint-plugin/src/rules/no-imports-from-self.js
AlessioGr: > The attribution for the origin and Payload CMS could be elsewhere in the code

Where do you suggest we add the attribution?

.eslintrc.js
AlessioGr: I have addressed this by changing

`if (importPath.startsWith(packageName)) {`

to

`if (importPath === packageJson.name) {`


This won't 100% cover all cases, e.g. `packages/lexical-react/src/LexicalMarkdownShortcutPlugin.tsx` could still incorrectly import from itself (`'@lexical/react/LexicalMarkdownShortcutPlugin'`) - but this is very unlikely to happen. 

Alternatively, we'd have to then parse `package.jsons#exports` which is not worth the additional overhead IMO. What do you think?

eslint-plugin/src/rules/no-imports-from-self.js
etrepum: I think just below the template header is fine. I don't think it's meaningful to try and add "portions copyright" or anything like that to the template header itself, because the CLA requires this grant of copyright license, but you could point towards the original implementation.

.eslintrc.js
etrepum: It's up to you, we can always add it later. There is already some code to do that sort of thing in scripts/shared/ that's used for other reasons. Probably makes sense to directly work with PackagesMetadata since packagesManager doesn't currently have a mechanism to clear the cache as it's not currently used in any sort of language-server-like capacity.

eslint-plugin/src/rules/no-imports-from-self.js
AlessioGr: I moved it below the license header - let me know if it's fine now

etrepum: I'll see how quickly I can get confirmation on that, it looks ok to me personally but since it still says copyright with multiple entities someone at Meta should have a look.

etrepum (approved):
Code looks good, waiting for clarification from Meta on the copyright comments

.eslintrc.js
AlessioGr: Neat utility - I have added support for parsing the package.json#exports if it imports from a self-subpath export. Let me know if you like it - it's a bit more overhead, on the other hand the package metadata is now cached on the module scope

eslint-plugin/src/rules/no-imports-from-self.js
etrepum: Based on feedback so far, if we can get this down to one line just attributing to where the original came from then it should be good to go. I think the primary concern here is the precedent, if contributions started carrying multi-line attributions it would get unwieldy pretty quickly.

AlessioGr: Omitting the rest of the license might be fine given that lexical is already MIT-licensed - but I think including the copyright notice (the `(c) Payload CMS, Inc` part) would be a requirement in order to stay compliant with the MIT license:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

This won't be a problem from our side, so if you don't mind it, I'm happy to omit it and just link to the original code

eslint-plugin/src/rules/no-imports-from-self.js
```diff
@@ -0,0 +1,123 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+/**
+ * From: https://github.com/payloadcms/payload/blob/main/packages/eslint-plugin/customRules/no-imports-from-self.js
+ * Copyright (c) Payload CMS, Inc. <info@payloadcms.com>
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
```
zurfyx, line 14: Let's just keep it to a liner as we've previously done with other references in this repo that add historical value. I like the fact that we're easily able to trace back to the source (which may be updated over time), given that using the package directly is not viable. Thank you!

```suggestion
// From: https://github.com/payloadcms/payload/blob/main/packages/eslint-plugin/customRules/no-imports-from-self.js
```

zurfyx (approved):

zurfyx:
LGTM, thank you Alessio!

PR #7279, opened 2025-03-03
feat: add HR theme config for selected state

RobHannay:
## Description

Closes #

### Before

### After


packages/lexical-react/src/LexicalHorizontalRuleNode.tsx
```diff
@@ -99,7 +99,7 @@ function HorizontalRuleComponent({nodeKey}: {nodeKey: NodeKey}) {
 
   useEffect(() => {
     const hrElem = editor.getElementByKey(nodeKey);
-    const isSelectedClassName = 'selected';
+    const isSelectedClassName = editor._config.theme.hrSelected ?? 'selected';
```
RobHannay, line 102: Fallback to avoid a breaking change

packages/lexical/src/LexicalEditor.ts
```diff
@@ -116,6 +116,7 @@ export type EditorThemeClasses = {
     h6?: EditorThemeClassName;
   };
   hr?: EditorThemeClassName;
+  hrSelected?: EditorThemeClassName;
```
RobHannay, line 119: Naming follows similar convention to the `table` class names below.

Nesting `hr` classes object would be a breaking change

etrepum (approved):
Looks good!

ivailop7:
We could have just used ':active' CSS pseudoclass on the current hr class instead of adding more to the theme object.

RobHannay:
fwiw, this kind of stuff could be solved with custom data attributes too, like `[data-selected]`. This is what React Aria Components does and I'm a big fan. However, that would also be a bigger discussion imo whether that's a pattern that would be preferred across all (React) plugins

etrepum:
I'm also a fan of the data-* approach for this kind of state (whether using react or not), that approach will fix some coherency issues with how theme classes are currently applied and removed when you're using utility style classes. Using a separate namespace from classes for ephemeral state fixes the problem entirely without a larger refactoring of how themes work and is still totally compatible with something like tailwind.

PR #7285, opened 2025-03-04
[lexical] Feature: Add listStyleType support to ListNode for customizable list styles

Tiscs:
## Description

Add listStyleType support to ListNode for customizable list styles.

### Before

By default:

image

### After

When set to `'circle'`

image

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -41,6 +41,7 @@ import {$getListDepth, $wrapInListItem} from './utils';
 export type SerializedListNode = Spread<
   {
     listType: ListType;
+    listStyleType?: ListStyleType;
     start: number;
     tag: ListNodeTagType;
   },
@@ -48,6 +49,27 @@ export type SerializedListNode = Spread<
 >;
 
 export type ListType = 'number' | 'bullet' | 'check';
+
+export type ListStyleType =
+  | undefined // default
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | (string & {}) // string means custom style
+  | 'none'
+  | 'disc'
+  | 'circle'
+  | 'square'
+  | 'decimal'
+  | 'decimal-leading-zero'
+  | 'lower-roman'
+  | 'upper-roman'
+  | 'lower-greek'
+  | 'lower-alpha'
+  | 'lower-latin'
+  | 'upper-alpha'
+  | 'upper-latin'
+  | 'armenian'
+  | 'georgian'
+  | 'hebrew';
 
 export type ListNodeTagType = 'ul' | 'ol';
 
@@ -59,6 +81,8 @@ export class ListNode extends ElementNode {
   __start: number;
   /** @internal */
   __listType: ListType;
+  /** @internal */
+  __listStyleType: ListStyleType;
 
   static getType(): string {
     return 'list';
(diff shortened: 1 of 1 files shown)


packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -49,6 +50,26 @@ export type SerializedListNode = Spread<
 
 export type ListType = 'number' | 'bullet' | 'check';
 
+export type ListStyleType =
+  | null // null means default
```
etrepum, line 54:
```suggestion
  | undefined // default
```

`undefined` is probably an easier default to manage in this case

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -49,6 +50,26 @@ export type SerializedListNode = Spread<
...
+  | string // string means custom style
```
etrepum, line 55: The only utility of having a union of string and some specific strings would be for autocomplete, but with `string` here by itself will default the autocomplete. There is a workaround though: https://www.charpeni.com/blog/open-ended-unions-and-autocomplete-with-typescript#solution
 
```suggestion
  | (string & {}) // string means custom style
```

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -70,11 +93,17 @@ export class ListNode extends ElementNode {
...
+    listType: ListType = 'number',
+    start: number = 1,
+    key?: NodeKey,
+    listStyleType?: ListStyleType,
+  ) {
     super(key);
     const _listType = TAG_TO_LIST_TYPE[listType] || listType;
     this.__listType = _listType;
     this.__tag = _listType === 'number' ? 'ol' : 'ul';
+    this.__listStyleType = listStyleType ?? null;
```
etrepum, line 106:
```suggestion
    this.__listStyleType = listStyleType;
```

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -114,6 +153,7 @@ export class ListNode extends ElementNode {
     }
     // @ts-expect-error Internal field.
     dom.__lexicalListType = this.__listType;
+    dom.style.setProperty('list-style-type', this.__listStyleType);
```
etrepum, line 156: This also needs to be done in `updateDOM` since this value can change

Tiscs:
> Is there a reason why you decided to add this property instead of using `listNode.setStyle('list-style-type: …')` and implementing the `__style` property in createDOM/updateDOM/exportDOM? It's already there, it's just that the DOM methods aren't currently using it (not for a good reason).

The rationale for adding the `__listStyleType` property is to provide a more specific and type-safe way to control list styling, rather than reusing a generic `__style` property.

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -49,6 +50,26 @@ export type SerializedListNode = Spread<
 
 export type ListType = 'number' | 'bullet' | 'check';
 
+export type ListStyleType =
+  | undefined // default
+  | (string & object) // string means custom style
```
etrepum, line 55:
```suggestion
  // eslint-disable-next-line @typescript-eslint/ban-types
  | (string & {}) // string means custom style
```

object and {} don't mean the same thing, using object makes it not accept arbitrary strings because `string & object = never`

[Playground Link](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAGlC8UDOwBOBLAdgcygMigHsAjAKwgGNgBuAKAHp6pmA9AfltEigE0Fk0WXAQDeAXzqNmUdrVpA)

etrepum (commented):
The general consensus of the contributor team is that this PR is well written (although it doesn't yet have tests or any support in the playground) but probably doesn't make sense to accept at this time because it will conflict with use of setStyle since that currently overwrites all styles on the node (if/when that's implemented by this node's createDOM/updateDOM).

Treating styles in a higher-level way (e.g. as a record instead of a string) is something that's likely to come up in the next few months so something like this could be revisited then, but using that facility rather than an extra prop on the class.

In the meantime it would be best to implement this sort of thing as a ListNode subclass and use a node override to use your subclass instead of the default ListNode.

PR #7289, opened 2025-03-04
[lexical-playground] Reuse guid method

ivailop7:
We had the same method a couple of times across the codebase. A small clean up.

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -65,6 +65,7 @@ export {default as positionNodeOnRange} from './positionNodeOnRange';
 export {default as selectionAlwaysOnDisplay} from './selectionAlwaysOnDisplay';
 export {
   $splitNode,
+  createUID,
   isBlockDomNode,
   isHTMLAnchorElement,
   isHTMLElement,
--- a/packages/lexical/src/index.ts
+++ b/packages/lexical/src/index.ts
@@ -239,6 +239,7 @@ export {
   $setCompositionKey,
   $setSelection,
   $splitNode,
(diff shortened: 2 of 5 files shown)


packages/lexical-utils/src/index.ts
```diff
@@ -65,6 +65,7 @@ export {default as positionNodeOnRange} from './positionNodeOnRange';
 export {default as selectionAlwaysOnDisplay} from './selectionAlwaysOnDisplay';
 export {
   $splitNode,
+  createUID,
```
etrepum, line 68: I don't think we need to export this from both lexical and @lexical/utils? It's already in lexical so maybe just import it from there rather than include its implementation twice?

packages/lexical/src/index.ts
```diff
@@ -239,6 +239,7 @@ export {
   $setCompositionKey,
   $setSelection,
   $splitNode,
+  createUID,
```
etrepum, line 242: We should add some API docs to this function if we're going to export it. I wouldn't say it's a good UID implementation, just a quick hack and only has a key space of <12M with up to 5 letters (there's no guarantee that there are *any* characters because it's not a good algorithm, if the representation happened to be all digits it could even return an empty string). For things that don't have to support super old browsers it might make sense to just use [randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) when possible (in secure contexts) - otherwise I would recommend that we fix the implementation so it's at least guaranteed to produce a 5 letter string, or adopt something like the [nanoid](https://github.com/ai/nanoid) algorithm.

PR #7297, opened 2025-03-05
[lexical-table] Add table cell selection handler for touch devices

ibastawisi:
## Description
[commit #fcb7666](https://github.com/facebook/lexical/commit/fcb7666) introduced a regression for table cell selection on touch devices, previously clicking a cell while cursor is inside another cell would select the cells in between, clicking again collapses the selection.

* [`packages/lexical-table/src/LexicalTableSelectionHelpers.ts`](diffhunk://#diff-4efac82ecef2990d8b44ca49b63c50e2767e9496650884a89ba7111c57152e15R309-R341): Implemented a function to handle click events, specifically for touch input, and update the table selection accordingly.

### DIFF
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -223,10 +223,8 @@ export default function Editor(): JSX.Element {
             <CollapsiblePlugin />
             <PageBreakPlugin />
             <LayoutPlugin />
-            {floatingAnchorElem && !isSmallWidthViewport && (
+            {floatingAnchorElem && (
               <>
-                <DraggableBlockPlugin anchorElem={floatingAnchorElem} />
-                <CodeActionMenuPlugin anchorElem={floatingAnchorElem} />
                 <FloatingLinkEditorPlugin
                   anchorElem={floatingAnchorElem}
                   isLinkEditMode={isLinkEditMode}
@@ -236,6 +234,12 @@ export default function Editor(): JSX.Element {
                   anchorElem={floatingAnchorElem}
                   cellMerge={true}
                 />
+              </>
+            )}
+            {floatingAnchorElem && !isSmallWidthViewport && (
+              <>
+                <DraggableBlockPlugin anchorElem={floatingAnchorElem} />
+                <CodeActionMenuPlugin anchorElem={floatingAnchorElem} />
                 <TableHoverActionsPlugin anchorElem={floatingAnchorElem} />
                 <FloatingTextFormatToolbarPlugin
                   anchorElem={floatingAnchorElem}
--- a/packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
@@ -907,7 +907,7 @@ function TableCellActionMenuContainer({
 
   useEffect(() => {
     // We call the $moveMenu callback every time the selection changes,
-    // once up front, and once after each mouseUp
+    // once up front, and once after each pointerUp
     let timeoutId: ReturnType<typeof setTimeout> | undefined = undefined;
     const callback = () => {
       timeoutId = undefined;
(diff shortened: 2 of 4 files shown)


etrepum (commented):
I am not able to make this work on Safari on iOS 18.3.1 with either the simulator or a real device. It did work in browerstack with a Samsung Galaxy S23 v13.0

etrepum:
Listening to more specific events rather than expecting some particular behavior from a click event sounds like the right approach to me, but I don't have a lot of recent experience dealing with touch device workarounds.

etrepum:
I think this particular solution only works well when the table doesn't require scrolling, it would be great if we could have the touchend be ignored if it didn't also scroll the table

754B5941-6F37-400E-8153-B7245719B394_4_5005_c

ibastawisi:
> I think this particular solution only works well when the table doesn't require scrolling, it would be great if we could have the touchend be ignored if it didn't also scroll the table
> 
>  ScreenRecording_03-05-2025.17-22-52_1.mp4

I see, the click event is best to prevent accidental touches, I guess there's no way to achieve the same behavior with just the touchend.
so..., how about reverting to click command and detecting touch with a global flag set on touchstart (and maybe cleared on touch end)? seems a little to much but would be bulletproof, I guess.

ibastawisi:
> This doesn't work on iOS Safari for me (either simulator or real device), the table selection doesn't happen at all, it behaves the same way that it does without this patch. It does work in browerstack with a Samsung Galaxy S23 v13.0 (Chrome)

Yeah for some reason browserstack is slow with vite dev server on my machine, I even tried a cloudflared tunnel and it was still very slow so I was optimistic andskipped local testing this time..
But the good news is that detecting pointerType works fine on safari with this method, I tested the resizer pr and it works there!
So I guess the problem is with selection, I will try to debug it properly tomorrow.. maybe I just need to stop the click event.. let's wait and see 🙏

etrepum:
Great, let me know if you get stuck, I have easy access to iOS and could probably help sometime later this week if needed

ibastawisi:
I found that safari updates the selection before triggering the click event, which makes click command unusable for this scenario.
Can I just change the mouse events with pointer events? if so a simple check added to Firefox workaround would achieve the desired behavior!
image

etrepum:
Edit: looks like at least one e2e test is failing, I will take a look once it's all passing :) 

```
 1 failed
    [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:1041:3 › Selection › Triple-clicking last cell in table should not select entire document 
```

ibastawisi:
> Edit: looks like at least one e2e test is failing, I will take a look once it's all passing :)
> 
> ```
>  1 failed
>     [chromium] › packages/lexical-playground/__tests__/e2e/Selection.spec.mjs:1041:3 › Selection › Triple-clicking last cell in table should not select entire document 
> ```

It seems that you can't track click count with `pointerdown` event!
https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail

reverting this one to a `mousedown` should pass the tests

etrepum:
Looks like the resize handler is failing now, I'll take a look at the changes to that one because I think it was good and maybe after merging that PR this one will be in better shape

ibastawisi:
> Looks like the resize handler is failing now, I'll take a look at the changes to that one because I think it was good and maybe after merging that PR this one will be in better shape

all the failed tests have Firefox in common, gonna debug it locally.

etrepum (approved):
Confirmed expected behavior on macOS Chrome, Safari, Firefox with clicking various table cells, shift-clicking to create selections, and click+drag to create selections.

Also confirmed on iOS Simulator, iOS hardware and Samsung Galaxy S23 on Browserstack. I do still have trouble selecting the whole table if there is scrolling, because as soon as you put your finger down to scroll it selects so you lose the anchor by the time the desired focus is visible. I don't think that this limitation should block the feature, but maybe we should try and do a follow-up to polish that aspect

ibastawisi:
> Confirmed expected behavior on macOS Chrome, Safari, Firefox with clicking various table cells, shift-clicking to create selections, and click+drag to create selections.
> 
> Also confirmed on iOS Simulator, iOS hardware and Samsung Galaxy S23 on Browserstack. I do still have trouble selecting the whole table if there is scrolling, because as soon as you put your finger down to scroll it selects so you lose the anchor by the time the desired focus is visible. I don't think that this limitation should block the feature, but maybe we should try and do a follow-up to polish that aspect

yeah so tricky!
maybe we'd better handle it in a selection change command, I'll think about it.

etrepum:
Let's go ahead and merge this now, because I think this behavior is an improvement even if not perfect, and the polish can be explored in a follow-up PR (hopefully soon). Nice work!

PR #7309, opened 2025-03-08
[lexical-table] Bug Fix: Fix unintended touch table cell selection when scrolling

### LINKED ISSUE #7308: Bug: Selecting table cells on touch devices happens unintentionally when scrolling
Follow-up to #7297

Selecting a range of table cells is done by tapping one cell (the anchor) to create a collapsed RangeSelection, and then tapping another cell (the focus) to create the TableSelection. This works well when the table does not require scrolling, but initiating a scroll is also considered a tap in this case and creates an unintended selection.

We should explore a different sequence of events to create the TableSelection from a focus on mobile. Possibly we would stash the intended focus cell and scroll information on pointerdown and create the TableSelection on pointerup if scrolling did not occur, or react to a selection change, or some other sequence of events.

ibastawisi:
This pull request adds a new property to track the type of pointer used and delegates the creation of a table selection until selection change occurs.

This approach would fix #7308.

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -247,6 +247,7 @@ export function applyTableHandlers(
   };
 
   const onPointerDown = (event: PointerEvent) => {
+    tableObserver.pointerType = event.pointerType;
     if (event.button !== 0 || !isDOMNode(event.target) || !editorWindow) {
       return;
     }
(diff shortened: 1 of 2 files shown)


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -260,9 +261,8 @@ export function applyTableHandlers(
         // if we go ahead and make the table selection now it will work
         // Handle case when tapping on a cell with touch device
```
ibastawisi, line 262: forgot to remove that line!

etrepum (approved):
LGTM!

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -999,6 +998,36 @@ export function applyTableHandlers(
...
+                  ),
+                );
+                tableObserver.$setFocusCellForSelection(
+                  $getObserverCellFromCellNodeOrThrow(
+                    tableObserver,
+                    focusCellNode,
+                  ),
+                  true,
+                );
+                tableObserver.pointerType = null;
```
ibastawisi, line 1028: Maybe we should reset the flag regardless of anchor and focus being different, this would nly trigger table selection on consecutive tabs on different cells without selection changes between them..

The downside is inconsistent behavior and maybe the flag state gets lost if multiple discrete selection changes happen following a single tab.

PR #7316, opened 2025-03-12
[lexical-table] Bug Fix: Add fallback selection to InsertTableCommand

patrick-atticus:
## Description
If you try to insert a table asynchronously when the editor is not focused, the table is not inserted (because there is no selection)

This can happen if you focus an input field (like row count in the playground), and then dispatch INSERT_TABLE_COMMAND async.

Screenshot 2025-03-12 at 3 01 13 pm

Tables are not inserted async in the playground so it's not a problem, but it can easily be reproduced by wrapping the insert command in a setTimeout in packages/lexical-playground/src/plugins/TablePlugin.tsx

This PR adds a fallback to `$getPreviousSelection()` in the table insert command handler. Same as https://github.com/facebook/lexical/pull/4570

Note that that actual insertion uses `$insertNodeToNearestRoot`, which already uses the same fallback. This change simply prevents the command from exiting early.


etrepum (commented):
Adding a test would be a good idea, since as you said the playground doesn’t do this there’s no confirmation that this fixes anything

patrick-atticus:
Unfortunately I could not reproduce this in the headless unit test environment.

I can't do an e2e test without changing playground behaviour. I updated the playground so now the insert is async and the button renders as disabled, which is nice when you insert a huge table and it takes a moment.

Now the table tests (eg `Can a table be inserted from the toolbar`) will fail without the updated `$insertTableCommandListener`.

I'll understand if you don't want these playground changes.

packages/lexical-playground/src/plugins/TablePlugin.tsx
```diff
@@ -101,12 +101,16 @@ export function InsertTableDialog({
...
+    // Large tables can take a moment. Defer insert to render button disabled
+    setTimeout(() => {
+      activeEditor.dispatchCommand(INSERT_TABLE_COMMAND, {
+        columns,
+        rows,
+      });
 
-    onClose();
+      onClose();
+    }, 0);
```
etrepum, line 113: Let's roll this change back, I don't see a good reason to demonstrate this anti-pattern in playground code that people will be copying. Table creation and rendering isn't async there's no real utility to disabling anything since the UI isn't going to be responsive while any of that is happening.

I don't see why you couldn't do this in a unit test with `$setSelection(null); editor.dispatchCommand(INSERT_TABLE_COMMAND, …);` given a previous update that had a selection.

patrick-atticus:
Earlier I was missing discreet updates to get the test sequencing correct.

I've also added a test for basic table insertion. The other tests checked that nested tables was NOT being inserted, and since the commands weren't registered, no tables could be inserted.

etrepum (approved):

PR #7317, opened 2025-03-12
[WIP][lexical][lexical-selection] Manually handle selection around TextNode boundaries

### LINKED ISSUE #7301: Feature: Proper Handling of Deletions Across Adjacent Unmergeable Text Spans in Grid Containers
## Description

When deleting across adjacent unmergeable text spans inside a `display: inline-grid` (or presumably `display: grid`) container, Lexical does not properly extend the selection into the next span, causing deletion to get stuck.

Previously, it was possible to work around this issue by manually extending the selection when deleting across boundaries (see [`span-deletion.ts`](https://stackblitz.com/edit/lexical-adjacent-span-deletion?file=src%2Fspan-deletion.ts)), but this workaround no longer functions in Lexical 0.27.0+ due to a change in selection syncing ([related change](https://github.com/facebook/lexical/pull/7270/files#diff-78b86cfd1ec931911a488d41ac3b779722121b8657aa59062b78856fa63cae12R1573-R1578)).

The expected behavior is for `RangeSelection.modify` to correctly move from the edge of one text node to the next before applying the native selection change, similar to `$modifySelectionAroundDecoratorsAndBlocks`.

### Steps to Reproduce

A minimal reproduction can be found here: [StackBlitz](https://stackblitz.com/edit/lexical-adjacent-span-deletion?file=src%2FApp.tsx).

1. Position the caret between "A" and "B".
2. Press <kbd>Del</kbd> to delete "B".
3. Press <kbd>Del</kbd> again.

#### Expected Results
- Both "B" and "C" should be deleted sequentially.

#### Actual Results

- Deletion gets stuck either at "B" or at "C" (it depends on if the caret initially lies at the start of the "B" span or the end of the "A" span).

This only occurs when using `display: inline-grid`. Disabling `inline-grid` allows deletion to function normally.

### Use cases

- Any applications requiring adjacent text nodes inside a container like `display: grid` (ruby text editing in our case)

### Alternative Solutions Considered

- Manually adjusting the selection before deletion (which worked in 0.26 but no longer works in 0.27 due to the updated selection syncing behavior).
- Finding an alternative layout approach that avoids using `grid` etc. (In our specific case, browser support for ruby layout has come a long way since we started but still only Firefox supports the necessary HTML tags to render mono ruby without interleaving the transcription nodes.)

## Impact

- This affects developers working with custom nodes that put multiple adjacent (unmergeable) text nodes inside certain containers (at least `display: inline-grid` but possibly others like `display: flex` too).
- The issue prevents deletion across text spans, breaking expected editor behavior.
- Fixing this would improve selection handling and deletion reliability in specialized text-editing scenarios.

(As discussed with @etrepum on Discord)

etrepum:
## Description

See #7301 - when alternative display is used for TextNode elements (e.g. not inline span) the behavior of the DOM's Selection.modify doesn't really work anymore because it jumps from one container to the next instead of across to the next logical character. By catching these scenarios and moving the DOM selection into the next element in advance this handles all of the (known) scenarios with arrows and deletions.

The arrow keys need to be handled unconditionally at TextNode boundaries because the selection is often out of sync with exactly what the Lexical points are.

Other issues likely still exist when moving by word or line, but character is the important granularity for most situations.

For future consideration we might want to move RangeSelection.modify into a command, or parts of it into a command, so that people can more easily work around these sorts of issues.

Closes #7301

## Test plan

TBD - The plan will probably be to add an e2e suite outside of the playground with the project sitting in `scripts/__tests__/integration/fixtures/lexical-adjacent-span-deletion` which can be used to demonstrate that this bug is fixed.

### Before

Run with release version of lexical, observe that arrow keys and delete doesn't work consistently (the provided workarounds `useNavigateBase` and `useSpanDeletion` are disabled)

```
cd scripts/__tests__/integration/fixtures/lexical-adjacent-span-deletion
npm install
npm run dev
```

### After

Run with the monorepo version of lexical, observe that the arrow keys and delete work correctly
```
cd scripts/__tests__/integration/fixtures/lexical-adjacent-span-deletion
npm install
npm run monorepo:dev
```

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -27,12 +27,12 @@ import {
   $isChildCaret,
   $isDecoratorNode,
   $isElementNode,
-  $isExtendableTextPointCaret,
   $isLeafNode,
   $isRangeSelection,
   $isRootNode,
   $isRootOrShadowRoot,
   $isTextNode,
+  $isTextPointCaret,
   $setSelection,
   INTERNAL_$isBlock,
 } from 'lexical';
@@ -430,8 +430,19 @@ export function $shouldOverrideDefaultCharacterSelection(
     selection.focus,
     isBackward ? 'previous' : 'next',
   );
-  if ($isExtendableTextPointCaret(focusCaret)) {
-    return false;
+
+  if ($isTextPointCaret(focusCaret)) {
+    // https://github.com/facebook/lexical/issues/7301
+    // Always handle the event manually at the boundaries of a TextNode
+    // because the DOM selection may not be normalized to the same Text
+    // child as the point. It could be on another node entirely that
+    // happens to be zero distance away.
+    // Conversely, *never* handle the event if it is internal to the
+    // TextNode.
+    return (
+      focusCaret.offset === 0 ||
+      focusCaret.offset === focusCaret.origin.getTextContentSize()
+    );
   }
   for (const nextCaret of $extendCaretToRange(focusCaret)) {
     if ($isChildCaret(nextCaret)) {
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1554,30 +1554,59 @@ export class RangeSelection implements BaseSelection {
     ) {
       removeDOMBlockCursorElement(blockCursorElement, editor, rootElement);
     }
-    if (this.dirty) {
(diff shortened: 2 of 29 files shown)


etrepum:
Ignore the failing unit tests for now, I don't think that suite has any real value as-is (it's testing a mock implementation that doesn't really behave like any real browser) and needs to be converted to an e2e suite.

PR #7318, opened 2025-03-12
[Breaking Change][lexical-table] Bug Fix: Table cell line breaks behave differently from the intended HTML behavior.

### LINKED ISSUE #7261: Bug: When there is only an inline DecoratorNode as a child of a ParagraphNode, a <br> is added at the end.
When there is only an inline DecoratorNode as a child of a ParagraphNode, an unnecessary `<br>` tag is added at the end. This issue persists even in the latest version of the Lexical package.

There is a mismatch case related to isLineBreakNode in the logic of $convertTableCellNodeElement.
```ts
    forChild: (lexicalNode, parentLexicalNode) => {
      if ($isTableCellNode(parentLexicalNode) && !$isElementNode(lexicalNode)) {
        const paragraphNode = $createParagraphNode();
        if (
          $isLineBreakNode(lexicalNode) &&
          lexicalNode.getTextContent() === '\n'
        ) {
          return null;
        }
        if ($isTextNode(lexicalNode)) {
          if (hasBoldFontWeight) {
            lexicalNode.toggleFormat('bold');
          }
          if (hasLinethroughTextDecoration) {
            lexicalNode.toggleFormat('strikethrough');
          }
          if (hasItalicFontStyle) {
            lexicalNode.toggleFormat('italic');
          }
          if (hasUnderlineTextDecoration) {
            lexicalNode.toggleFormat('underline');
          }
        }
        paragraphNode.append(lexicalNode);
        return paragraphNode;
      }

      return lexicalNode;
    }
```

Lexical version: 0.25.0, 0.24.0

## Steps To Reproduce

1. Create a ParagraphNode.
2. Add only an inline DecoratorNode as a child of the ParagraphNode.
3. Check the result to see that an unnecessary <br> tag is added at the end.

Link to code example: https://playground.lexical.dev/ - inline Equation

Image

## The current behavior

When there is only an inline DecoratorNode as a child of a ParagraphNode, an unnecessary <br> tag is added at the end.

## The expected behavior

When there is only an inline DecoratorNode as a child of a ParagraphNode, no <br> tag should be added.

## Impact of fix

This bug can cause unnecessary line breaks in the user interface, affecting user experience.

dineug:
## Breaking Change

The importDOM implementation for TableCellNode has been corrected to behave as other Lexical roots do with regard to converting `<br>` tags to `LineBreakNode` and wrapping runs of top-level inline nodes (including inline decorators) with `ParagraphNode`. If you have any specific workarounds accounting for the previous behavior, you can remove them. The one difference from other import paths is that a solitary `<br>` (e.g. `<td><br></td>`) is converted to an empty `ParagraphNode` rather than a `ParagraphNode` containing a `LineBreakNode`.

## Description
Table cell line breaks behave differently from the intended HTML behavior.

Modify the logic to align with the intended HTML line breaks.

Closes #7261

## Test plan

Add TC to `TableCellNode.importDOM`

### Before

| case | html | editor |
| ---- | ---- | ------ |
| inline decorator | \<td><br> \<test-decorator>\</test-decorator><br> 1<br>\</td> | \<td><br> \<p><br> \<test-decorator>\</test-decorator><br> \<br /><br> \</p><br> \<p><br> \<span>1\</span><br> \</p><br>\</td> |
| br | \<td><br> 1<br> \<br /><br> \<br /><br> \<br /><br> \<br /><br> 2<br>\</td> | \<td><br> \<p><br> \<span>1\</span><br> \</p><br> \<p><br> \<span>2\</span><br> \</p><br>\</td> |

### After

| case | html | editor |
| ---- | ---- | ------ |
| inline decorator | \<td><br> \<test-decorator>\</test-decorator><br> 1<br>\</td> | \<td><br> \<p><br> \<test-decorator>\</test-decorator><br> \<span>1\</span><br> \</p><br>\</td> |
| br | \<td><br> 1<br> \<br /><br> \<br /><br> \<br /><br> \<br /><br> 2<br>\</td> | \<td><br> \<p><br> \<span>1\</span><br> \<br /><br> \<br /><br> \<br /><br> \<br /><br> \<span>2\</span><br> \</p><br>\</td> |

### DIFF
--- a/packages/lexical-table/src/LexicalTableCellNode.ts
+++ b/packages/lexical-table/src/LexicalTableCellNode.ts
@@ -15,6 +15,7 @@ import type {
   LexicalNode,
   LexicalUpdateJSON,
   NodeKey,
+  ParagraphNode,
   SerializedElementNode,
   Spread,
 } from 'lexical';
@@ -23,7 +24,7 @@ import {addClassNamesToElement} from '@lexical/utils';
 import {
   $applyNodeReplacement,
   $createParagraphNode,
-  $isElementNode,
+  $isInlineElementOrDecoratorNode,
   $isLineBreakNode,
   $isTextNode,
   ElementNode,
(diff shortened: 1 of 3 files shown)


etrepum (commented):
It's probably not expected that TableCell import behaves differently than elsewhere, I would expect

input

```html
<td>
  1
  <br />
  <br />
  <br />
  <br />
  2
</td>
```

output

```html
<td>
  <p>
    <span>1</span>
    <br />
    <br />
    <br />
    <br />
    <span>2</span>
  </p>
</td>
```

I don't think each `<br />` should create a new block, that's not typically how lexical works.

For example, if you paste this text/html into the root of a lexical editor:

```html
1<br><br><br>2
```

Then you will get a document with this structure:

```
 root
  └ (47) paragraph 
    ├ (48) text "1"
    ├ (49) linebreak 
    ├ (50) linebreak 
    ├ (51) linebreak 
>   └ (52) text "2"
```

The same thing applies when you paste that same html into a table cell.

I don't really understand what you're trying to "fix" here (at least in the br case, I have not investigated the decorator case)

dineug:
I have made the changes you mentioned! @etrepum

etrepum:
It doesn't look like test are passing with these changes

packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -367,8 +368,24 @@ export function $convertTableCellNodeElement(
...
+        }
+      };
+
       for (const child of childLexicalNodes) {
-        if (!$isElementNode(child)) {
+        if (
+          $isInlineElementOrDecoratorNode(child) ||
+          $isTextNode(child) ||
+          $isLineBreakNode(child)
+        ) {
```
dineug, line 388: I have refined the conditions to be more accurate.

etrepum (approved):
Looks good, nice work!

PR #7323, opened 2025-03-12
[Breaking Change][lexical][lexical-playground] Feature: Add a default delete handler for NodeSelection

### LINKED ISSUE #7319: Bug: when deleting consecutive DecoratorNodes, sometimes two are deleted at once
If a DecoratorNode is selected and I press Backspace, if there are two or more DecoratorNodes in a row, two of them are deleted at once.

Lexical version: 0.27.0

## Steps To Reproduce

1. Go to the Lexical Playground
2. Insert a Horizontal Rule
3. Press left arrow so that it's selected
4. Copy it
5. Press right arrow to deselect the hr
6. Paste two times (for a total of 3 hrs)
7. This should result in this state:
```
 root
  ├ (95) paragraph 
  ├ (96) horizontalrule 
  ├ (98) horizontalrule 
  └ (100) horizontalrule 
```
8. Now select the bottommost horizontal rule
  - As an aside, pressing left arrow from the bottom of the editor seems to select the _center_ hr at first
9. Press backspace
10. Only one should be deleted, but instead two are deleted, leaving the topmost one

This also occurs with pressing the Delete key when selecting either the top or middle hr; the next two hrs are deleted.

## The current behavior

Two hrs are deleted at a time in the case described above

## The expected behavior

Only one hr is deleted

## Impact of fix

I pointed to the playground and used hrs to illustrate the example, but we're using custom DecoratorNodes in our application (that use the same `$onDelete` logic as [LexicalHorizontalRuleNode](https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalHorizontalRuleNode.tsx)) and this leads to surprising behavior for users when multiple are deleted at once. In the case of just two DecoratorNodes back to back, there is a workaround by selecting just one and either pressing Delete or Backspace to force a single node at the edge to be deleted, but we wouldn't expect users to be aware of how to do that.

etrepum:
## Breaking Change

If you have any `$onDelete` handlers copied from the playground for `KEY_DELETE_COMMAND` and `KEY_BACKSPACE_COMMAND`, you can delete them all now. If you leave them in, it will have similar bugs as prior to this PR.

The default `KEY_DELETE_COMMAND` and `KEY_BACKSPACE_COMMAND` handlers now preventDefault and call `DELETE_CHARACTER_COMMAND` for NodeSelection (in addition to the existing RangeSelection behavior).

The `DELETE_CHARACTER_COMMAND` handler now handles NodeSelection by calling the new `NodeSelection.deleteNodes()` method which places a new RangeSelection where the first selected node was (if any, and if it was the current selection) and then removes all of the selected nodes.

## Description

NodeSelection has no default handler for when the delete or backspace keys are pressed. This results in having every individual decorator node plugin register the same handler to incorrectly deal with this situation.

By centralizing the handler to a default location we can fix up the selection after a delete and remove a bunch of boilerplate code (while still having the option to override this behavior in the same way).

Closes #7319

## Test plan

New e2e regression tests added to catch this specific behavior

### DIFF
--- a/packages/lexical-playground/__tests__/regression/7319-delete-character-backward-nodeselection.spec.mjs
+++ b/packages/lexical-playground/__tests__/regression/7319-delete-character-backward-nodeselection.spec.mjs
@@ -1,0 +1,101 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {deleteBackward, moveLeft} from '../keyboardShortcuts/index.mjs';
+import {
+  assertHTML,
+  focusEditor,
+  html,
+  initialize,
+  pasteFromClipboard,
+  test,
+} from '../utils/index.mjs';
+
+test.describe('Regression tests for #7319', () => {
+  test.beforeEach(({isPlainText, isCollab, page}) =>
+    initialize({isCollab, isPlainText, page}),
+  );
+
+  test(`deleteCharacter after hr with RangeSelection`, async ({
+    page,
+    isCollab,
+    isPlainText,
+  }) => {
+    test.skip(isCollab || isPlainText);
+    await focusEditor(page);
+    await page.keyboard.press('Enter');
+    const hrCount = 3;
+    for (let i = 0; i < hrCount; i++) {
+      await pasteFromClipboard(page, {
+        'text/html': html`
+          <hr />
+        `,
+      });
+    }
+    async function assertHR(count) {
+      const expectedHtml = html`
+        <p><br /></p>
+        ${Array.from(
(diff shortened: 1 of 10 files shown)


thorn0:
Love that you provide tips for code copied from the playground. Please include such tips in release notes.

etrepum:
Breaking Changes notes go in the release notes

thorn0:
Sorry for going off-topic, my request isn't just about this PR. I recently updated Lexical in a project where the editor was based on the playground, and I had to go through half a year's worth of changes in the playground code between 0.16.1 and 0.27. I'm concerned I might have missed something important.

etrepum:
You probably did, the release notes didn’t start including this sort of information until a few months ago. You’ll probably want to audit all of the CSS/theme properties especially around selections and tables.

ivailop7 (approved):

PR #7338, opened 2025-03-15
[lexical-playground][lexical-react] Feature: Push Draggable Element to Parent

sescandell:
## Description
Allows passing a function to receive the DOM element identified by the DraggableBlockPlugin. 

An example application of this is the addition of a "+" button that allows the creation of a block above the identified one. This feature is particularly useful in the playground, as it allows adding a new block at the top of the page without having to drag and drop, for example.

### Before

image

### After

image

### DIFF
--- a/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
@@ -64,7 +64,7 @@ export default function DraggableBlockPlugin({
         <div ref={targetLineRef} className="draggable-block-target-line" />
       }
       isOnMenu={isOnMenu}
-      onTargetElementChanged={setDraggableElement}
+      onElementChanged={setDraggableElement}
     />
   );
 }
--- a/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
@@ -44,7 +44,7 @@ export default function DraggableBlockPlugin({
       }
 
       const pNode = $createParagraphNode();
-      if (e.metaKey || e.ctrlKey) {
+      if (e.altKey || e.ctrlKey) {
         node.insertBefore(pNode);
       } else {
         node.insertAfter(pNode);
--- a/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
@@ -9,8 +9,10 @@ import type {JSX} from 'react';
 
 import './index.css';
 
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {DraggableBlockPlugin_EXPERIMENTAL} from '@lexical/react/LexicalDraggableBlockPlugin';
-import {useRef} from 'react';
+import {$createParagraphNode, $getNearestNodeFromDOMNode} from 'lexical';
+import {useRef, useState} from 'react';
 
 const DRAGGABLE_BLOCK_MENU_CLASSNAME = 'draggable-block-menu';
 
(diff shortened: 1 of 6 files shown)


etrepum (commented):
This implementation looks good but I think it might make sense to mirror Notion's UI and have the default behavior for this button be insert after rather than insert before.

Screenshot 2025-03-15 at 21 50 24

ivailop7:
+1 on Bob's comment for "insert after" I'd even go as far and say it even makes sense to render the component picker menu for the user to select any widget, not just a paragraph.

sescandell:
> This implementation looks good but I think it might make sense to mirror Notion's UI and have the default behavior for this button be insert after rather than insert before.
> 

I just pushed an update based on your comment. I replicated Notion's behavior: by default, the new node is inserted after the current one. If you press the Cmd (or Ctrl) key while clicking, it inserts the new node before the current one.



> +1 on Bob's comment for "insert after" I'd even go as far and say it even makes sense to render the component picker menu for the user to select any widget, not just a paragraph.

I completely agree! However, there isn't a ready-to-use "component picker" at the moment. The existing one is tightly coupled with the typeahead plugin, so extracting it or duplicating some of its code would be necessary.

I'm not sure how to easily create a "menu" within the playground next to my button. Maybe we could tackle this in a separate PR?

etrepum:
I agree that it makes sense to handle the dialog in a separate PR. It is basically the same interface as the `/` typeahead if following notion's UX. The block paragraph is created eagerly anyway, so the menu would just be a state change that happens after this behavior.

Any reason why you chose cmd-click instead of opt-click like Notion?

sescandell:
> Any reason why you chose cmd-click instead of opt-click like Notion?

I just used the same as in FloatingLinkEditorPlugin: https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx#L369 
I can change it. What is "Option-Click"?

etrepum:
Option is the Mac equivalent of the Alt key https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/altKey

ivailop7:
Can I pitch shift+click on Windows, which always made more sense to me for such actions.

sescandell:
Notion is using the AltKey on windows... which seems... unnatural for me. But I'm not a big user of notion.

I changed for the option command. Let me know if you want the AltKey for Windows

packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -366,7 +366,7 @@ function useFloatingLinkEditorToolbar(
           if ($isRangeSelection(selection)) {
             const node = getSelectedNode(selection);
             const linkNode = $findMatchingParent(node, $isLinkNode);
-            if ($isLinkNode(linkNode) && (payload.metaKey || payload.ctrlKey)) {
+            if ($isLinkNode(linkNode) && (payload.altKey || payload.ctrlKey)) {
```
etrepum, line 369: It looks like this is the wrong event handler, this one is to open links (meta is correct for that)
```suggestion
            if ($isLinkNode(linkNode) && (payload.metaKey || payload.ctrlKey)) {
```

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.tsx
```diff
@@ -23,8 +25,33 @@ export default function DraggableBlockPlugin({
...
+    }
+
+    editor.update(() => {
+      const node = $getNearestNodeFromDOMNode(draggableElement);
+      if (!node) {
+        return;
+      }
+
+      const pNode = $createParagraphNode();
+      if (e.metaKey || e.ctrlKey) {
```
etrepum, line 47: This is the place that should be altKey
```suggestion
      if (e.altKey || e.ctrlKey) {
```

sescandell:
@etrepum Sorry... careless mistake. Thanks for the correction!

etrepum (approved):
LGTM! Nice job

PR #7340, opened 2025-03-15
[lexical-utils] Feature: Add type predicate to objectKlassEquals

2wheeh:
## Description

- adds type predicate to `objectKlassEquals`
- removes unnecessary type assertions following `objectKlassEquals` calls

it is backward compatible since the return type is still boolean but just with type predicate.


etrepum (approved):
LGTM!

PR #7341, opened 2025-03-15
[Breaking Change][lexical] Bug Fix: Fix bug in transformer loop that would cause nodes not to get reconciled

### LINKED ISSUE #7333: Bug: the next list was deleted when delete paragraph
Lexical version: 0.72.2

## Steps To Reproduce

https://github.com/user-attachments/assets/23270dbb-4a64-49fa-abfd-d51783c8bc2a

etrepum:
## Breaking Change

This PR moves the incorrectly overridden `append` methods for RootNode and ListNode and moves it to the ElementNode's primitive `splice` method. For RootNode this means that the exception you'd get for appending a leaf node to the root will be thrown in all situations when that node is inserted, rather than just append. For ListNode this means that the automatic ListItemNode wrapping applies in all situations when children are inserted into the node, not just append.

## Description

The `$applyAllTransforms` loop has a fast path to ignore ElementNode that were not intentionally marked as dirty (ancestors of parents).

```ts
      if (nodeKey !== 'root' && !intentionallyMarkedAsDirty) {
        continue;
      }
```

This is a sensible thing to do, but the way it was implemented *discarded* the fact that this node was not intentionally marked as dirty, which is information that the reconciler needs. The reconciler starts a traversal from the root and only reconciles dirty leaves and elements (regardless of whether intentional or not). If this information goes missing, that subtree does not get reconciled, and you can end up with situations where nodes exist in the NodeMap but were not rendered to the DOM if they were moved in a particular way (such as in #7333) *and* a transform exists.

While I was researching this issue I cleaned up the append overrides for RootNode and ListItemNode and added some documentation about how reconciliation and transforms interact.

Closes #7333

## Test plan

See #7333 and added unit test

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -193,31 +193,26 @@ export class ListNode extends ElementNode {
     return false;
   }
 
-  append(...nodesToAppend: LexicalNode[]): this {
-    for (let i = 0; i < nodesToAppend.length; i++) {
-      const currentNode = nodesToAppend[i];
-
-      if ($isListItemNode(currentNode)) {
-        super.append(currentNode);
-      } else {
-        const listItemNode = $createListItemNode();
-
-        if ($isListNode(currentNode)) {
-          listItemNode.append(currentNode);
-        } else if ($isElementNode(currentNode)) {
-          if (currentNode.isInline()) {
-            listItemNode.append(currentNode);
-          } else {
-            const textNode = $createTextNode(currentNode.getTextContent());
-            listItemNode.append(textNode);
-          }
-        } else {
-          listItemNode.append(currentNode);
+  splice(
+    start: number,
+    deleteCount: number,
+    nodesToInsert: LexicalNode[],
+  ): this {
+    let listItemNodesToInsert = nodesToInsert;
+    for (let i = 0; i < nodesToInsert.length; i++) {
+      const node = nodesToInsert[i];
+      if (!$isListItemNode(node)) {
+        if (listItemNodesToInsert === nodesToInsert) {
+          listItemNodesToInsert = [...nodesToInsert];
         }
-        super.append(listItemNode);
+        listItemNodesToInsert[i] = $createListItemNode().append(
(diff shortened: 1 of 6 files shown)


packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -193,31 +193,26 @@ export class ListNode extends ElementNode {
...
-        } else if ($isElementNode(currentNode)) {
-          if (currentNode.isInline()) {
-            listItemNode.append(currentNode);
-          } else {
-            const textNode = $createTextNode(currentNode.getTextContent());
-            listItemNode.append(textNode);
-          }
-        } else {
-          listItemNode.append(currentNode);
+  splice(
```
etrepum, line 196: This is unrelated cleanup, but overriding append should never be done since splice is the primitive method

packages/lexical/src/LexicalUpdates.ts
```diff
@@ -308,6 +308,7 @@ function $applyAllTransforms(
     for (const currentUntransformedDirtyElement of untransformedDirtyElements) {
       const nodeKey = currentUntransformedDirtyElement[0];
       const intentionallyMarkedAsDirty = currentUntransformedDirtyElement[1];
+      dirtyElements.set(nodeKey, intentionallyMarkedAsDirty);
```
etrepum, line 311: Moving this single line is the actual fix

packages/lexical/src/nodes/LexicalRootNode.ts
```diff
@@ -82,18 +82,18 @@ export class RootNode extends ElementNode {
...
-  append(...nodesToAppend: LexicalNode[]): this {
-    for (let i = 0; i < nodesToAppend.length; i++) {
-      const node = nodesToAppend[i];
-      if (!$isElementNode(node) && !$isDecoratorNode(node)) {
-        invariant(
-          false,
-          'rootNode.append: Only element or decorator nodes can be appended to the root node',
-        );
-      }
+  splice(
```
etrepum, line 85: Another unrelated cleanup, this was the other case where append was incorrectly overridden

ivailop7 (approved):

PR #7342, opened 2025-03-16
[Breaking Change][lexical][lexical-utils]: Bug Fix: Handle canBeEmpty in $splitNodes

### LINKED ISSUE #6849: Bug: If you add a table on a list it adds it above the list and results in invalid exported HTML
If you add an unordered list and then on the list add a table it will add it above the list item and the exported HTML will contain an empty unordered list tag which is invalid HTML.

Lexical version: 0.17.1

## Steps To Reproduce

1. Add unordered list
2. Click on the top list element
3. Add a table
4. Observe the table is above the list element and if you call the `$generateHtmlFromNodes` function it will contain an empty `<ul>` tag

Link to code example:

I noticed this in my own app using the `INSERT_TABLE_COMMAND` command but was able to reproduce it in the Lexical playground so I thought it was easier if I just demoed it using that. 

https://github.com/user-attachments/assets/4858287e-071a-484e-803b-05117e99fc88

## The current behavior
Table gets added above the list item and exported HTML contains an empty `<ul>` tag

## The expected behavior
Table gets added below the list item and exported HTML does not contain a empty `<ul>` tag

## Impact of fix
- This bug seems somewhat but not catastrophically bad. To get around it I need to strip out these empty list elements after exporting the editor content
- This happens consistently for me, both in my own app and in the Lexical playground. I've just mentioned unordered lists for simplicity but it also applies to ordered lists
- This seems like a bit of an edge case but as it is general behaviour I think it could benefit lots of users

etrepum:
## Breaking Change

Previously `$insertNodeToNearestRoot` could create empty ElementNode when splitting the current node, now it respects `canBeEmpty()` and will not split in those cases.

## Description

This change makes `$insertNodeToNearestRoot` to respect the `canBeEmpty()` method of the nodes it is splitting and provides infrastructure to better control splitting (`SplitAtPointCaretNextOptions`, `$insertNodeToNearestRootAtCaret`, `$splitAtPointCaretNext`).

New lexical APIs:
  * `$splitAtPointCaretNext` - a lower-level API to split at the given caret exactly once (not recursive) with full control over how that happens
  * `$getCollapsedCaretRange(pointCaret)` - convenience for `$getCaretRange(pointCaret, pointCaret)` to reduce redundant code

New @lexical/utils API:
  * `$insertNodeToNearestRootAtCaret` - lower-level API that does `$insertNodeToNearestRoot` recursive split and then insert without the quirks and selection changes (those happen in the `$insertNodeToNearestRoot` function).

I'm not very fond of the manner in which nodes are split (the empty splits are a bit awkward especially when not a ParagraphNode), but this doesn't break backwards compatibility in any case that we test and this function is only used in the playground and the implementation of commands that can be easily overridden (e.g. `INSERT_HORIZONTAL_RULE_COMMAND`). Now that the lower-level functions are exposed, people can control how this works for better UX.

Closes #6849

## Test plan

All existing e2e tests pass, new ListNode split test

### Before

Splitting an empty ListItemNode creates a broken list:

```html
<ul class="PlaygroundEditorTheme__ul"></ul>
<hr>
<ul class="PlaygroundEditorTheme__ul">
  <li value="1" class="PlaygroundEditorTheme__listItem"></li>
</ul>
```

### After

Splitting an empty ListItemNode creates a complete second list:

```html
<ul class="PlaygroundEditorTheme__ul">
  <li value="1" class="PlaygroundEditorTheme__listItem"></li>
</ul>
<hr>
<ul class="PlaygroundEditorTheme__ul">
  <li value="1" class="PlaygroundEditorTheme__listItem"></li>
</ul>
```

### DIFF
--- a/packages/lexical/src/__tests__/utils/index.tsx
+++ b/packages/lexical/src/__tests__/utils/index.tsx
@@ -26,6 +26,7 @@ import {
 import {HeadingNode, QuoteNode} from '@lexical/rich-text';
 import {TableCellNode, TableNode, TableRowNode} from '@lexical/table';
 import {expect} from '@playwright/test';
+import * as prettierSync from '@prettier/sync';
 import {
   $isRangeSelection,
   createEditor,
@@ -45,7 +46,6 @@ import {
   TextNode,
 } from 'lexical';
 import path from 'path';
-import * as prettier from 'prettier';
 import * as React from 'react';
 import {createRef} from 'react';
 import {createRoot} from 'react-dom/client';
@@ -58,7 +58,7 @@ import {
 } from '../../LexicalEditor';
 import {resetRandomKey} from '../../LexicalUtils';
 
-const prettierConfig = prettier.resolveConfig.sync(
+const prettierConfig = prettierSync.resolveConfig(
   path.resolve(__dirname, '../../../../.prettierrc'),
 );
 
@@ -784,7 +784,7 @@ export function expectHtmlToBeEqual(actual: string, expected: string): void {
 }
 
 export function prettifyHtml(s: string): string {
-  return prettier.format(s.replace(/\n/g, ''), {
+  return prettierSync.format(s.replace(/\n/g, ''), {
     ...prettierConfig,
     parser: 'html',
   });
--- a/packages/lexical/src/__tests__/utils/index.tsx
+++ b/packages/lexical/src/__tests__/utils/index.tsx
@@ -26,7 +26,6 @@ import {
 import {HeadingNode, QuoteNode} from '@lexical/rich-text';
 import {TableCellNode, TableNode, TableRowNode} from '@lexical/table';
 import {expect} from '@playwright/test';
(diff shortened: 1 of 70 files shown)


packages/lexical/src/__tests__/utils/index.tsx
```diff
@@ -335,9 +337,13 @@ export function $createTestExcludeFromCopyElementNode(): TestExcludeFromCopyElem
...
 }
 
-export type SerializedTestDecoratorNode = SerializedLexicalNode;
+export type SerializedTestDecoratorNode = Spread<
+  SerializedLexicalNode,
+  {block?: boolean}
+>;
 
 export class TestDecoratorNode extends DecoratorNode<JSX.Element> {
+  __block: boolean = false;
```
etrepum, line 346: DecoratorNode are inline by default but some of these tests were inserting it directly into the root. I added this to make that reasonable.

Issue #2715, opened 2022-07-27
Bug: Can't backslash escape markdown characters


cduff:
Lexical version: 0.3.8

## Steps To Reproduce

Markdown usually supports backslash escaping of characters.

https://spec.commonmark.org/0.30/#backslash-escapes

For example, `\*not italic\*` should not display as italic as the asterisk characters are backslash escaped.

`$convertFromMarkdownString` does not appear to handle backslash escaping correctly.

Link to code example:

https://codesandbox.io/s/lexical-md-escape-issue-lyfh9y?file=/src/Editor.js

## The current behavior

The escaping is ignored, resulting in italic text:

![image](https://user-images.githubusercontent.com/7855146/181142385-a017cf5c-de61-4c15-9945-822f83b6309a.png)

## The expected behavior

The expected output would be the text `*not italic*` with no italics.

digitalgopnik:
:+1:

digitalgopnik:
![escaped_not_working](https://user-images.githubusercontent.com/2630316/183086146-4fa52209-17af-4008-8f4d-b6c7cdfe356b.gif)

digitalgopnik:
I think we need the some regex-gods here, to extend the ```TRANSFORMER```-regexes? :slightly_smiling_face:

digitalgopnik:
Maybe a solution?

I am using the following regex for matching italic-markdown (underscore) -> (asterisk is analog, just replace ```\_``` with ```\*```)
`
const UNESCAPE_ITALIC_UNDERSCORE_IMPORT_REGEX =
  /([\_])(?<!(?:\1|\w).)(?![_*\s])(.*?[^_*\s])(?=\1)([\_])(?!\w|\3)/;
const UNESCAPE_ITALIC_UNDERSCORE_REGEX =
  /([\_])(?<!(?:\1|\w).)(?![_*\s])(.*?[^_*\s])(?=\1)([\_])(?!\w|\3)/;
`

My transformer for italic_underscore looks like this:
`
export const UNESCAPE_ITALIC_UNDERSCORE: TextMatchTransformer = {
  export: () => null,
  importRegExp: UNESCAPE_ITALIC_UNDERSCORE_IMPORT_REGEX,
  regExp: UNESCAPE_ITALIC_UNDERSCORE_REGEX,
  replace: (textNode, _) => {
    const notEscapedUnderscoreRegex = /(?<![\\]{1})[\_]{1}/g;
[...]

PR #7353, opened 2025-03-19
[Breaking Change][lexical-markdown] Feature: Support escaping markdown characters

AlessioGr:
Fixes https://github.com/facebook/lexical/issues/2715

This PR adds full support for escaping markdown characters during import and export. For example, `\*` now correctly becomes a literal `*` in the editor, and when exporting back to markdown, those special characters are automatically escaped (e.g. `*` becomes `\*`), preventing accidental formatting on re-import.

Safari [now has much better support for negative lookbehind regex](https://caniuse.com/?search=Negative%20Lookbehind), which is why I removed the conditions specific to it.

### Markdown Imports

Previously, any escaped characters (e.g. `\*`) were imported as literal `\*`. Now the backslash is stripped, and the escaped character (`*`) is imported instead.

### Markdown Exports

Previously, all markdown characters that were part of a text node (e.g. `**`) were exported unescaped. Now they are exported escaped (e.g. `\*\*`), preventing unwanted formatting if the text is re-imported.

### Example:

- **Old Behavior**
	- Text node: `Some **text**`
	- Exported to Markdown as: `Some **text**` (bold if re-imported)
	- Text node If re-imported: Some **text**
- **New Behavior**
	- Text node: `Some **text**`
	- Exported to Markdown as: `Some \*\*text\*\*` (literal asterisks if re-imported)
	- Text node If re-imported: `Some **text**`

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -207,6 +207,12 @@ function exportTextFormat(
   // bring the whitespace back. So our returned string looks like this: "   **foo**   "
   const frozenString = textContent.trim();
   let output = frozenString;
+
+  if (!node.hasFormat('code')) {
+    // Escape any markdown characters in the text content
+    output = output.replace(/([*_`~\\])/g, '\\$1');
+  }
+
   // the opening tags to be added to the result
   let openingTags = '';
   // the closing tags to be added to the result
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -297,10 +297,6 @@ function createTextFormatTransformersIndex(
   const openTagsRegExp: string[] = [];
   const escapeRegExp = `(?<![\\\\])`;
 
-  // Sort `textTransformers` so that longer tags come before shorter ones
-  // (prevents `*` from "stealing" text that should match `**`).
-  textTransformers.sort((a, b) => b.tag.length - a.tag.length);
-
   for (const transformer of textTransformers) {
     const {tag} = transformer;
     transformersByTag[tag] = transformer;
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -300,7 +300,6 @@ function createTextFormatTransformersIndex(
   for (const transformer of textTransformers) {
     const {tag} = transformer;
     transformersByTag[tag] = transformer;
-
     const tagRegExp = tag.replace(/(\*|\^|\+)/g, '\\$1');
     openTagsRegExp.push(tagRegExp);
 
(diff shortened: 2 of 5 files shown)


AlessioGr:
Regarding the failing "Should not break while importing and exporting multiple matches" Hashtag test:

This is how it worked before:

https://github.com/user-attachments/assets/1daf55c0-fa6b-4a01-848d-022dd95cdf5b

This is how it works now:

https://github.com/user-attachments/assets/076b800a-c22b-47f3-9df5-df3bd4cefad5

The test passed in richt-text mode and only fails in plain-text mode. I think that the new behavior is actually more correct - it explicitly escapes the ``` block.  \`\`\`markdown was written out in plain text.

If the text node containing \`\`\`markdown is converted to Markdown, then \\\`\\\`\\\`markdown is the correct behavior, as it ensures that the text  \`\`\`markdown is preserved correctly. It converted a text node to markdown, not an actual code block node.

I would suggest disabling this particular text for plain text mode. The intended behavior seems to be for \`\`\`markdown to create a markdown block while typing, which is not the case in plain text mode

etrepum:
I don't think it really makes sense to have markdown enabled for plain text, so disabling those tests probably makes sense. Either way this is a breaking change and the consequences should be clearly spelled out so they can be included in the release notes

AlessioGr:
> I don't think it really makes sense to have markdown enabled for plain text, so disabling those tests probably makes sense.

Agreed, I have disabled that test. Test suite should be passing now.

> Either way this is a breaking change and the consequences should be clearly spelled out so they can be included in the release notes

I have adjusted my PR description and marked it as breaking change. Technically it could also be considered a bug fix instead of a breaking change. Many projects, e.g. TypeScript, do not consider breaking incorrect behavior as breaking change. Let me know if you prefer it this way

etrepum:
People rely on broken behavior in Lexical all the time, better to document it than not. I think the biggest breaking change here is the input rather than the output (although the correct output of this version isn't parsable by older versions). That said, Lexical's markdown implementation has always done a bad job at being compliant with markdown/commonmark specs.

etrepum (approved):
This seems like a positive improvement to me! However, I know that Meta has a conservative backwards compatibility policy and I don't know if negative lookbehind is included, and if this would affect their use of the markdown package. Will have to wait for some feedback on their end before merge.

potatowagon (approved):
thanks for this fix

potatowagon:
> People rely on broken behavior in Lexical all the time, better to document it than not. I think the biggest breaking change here is the input rather than the output (although the correct output of this version isn't parsable by older versions). That said, Lexical's markdown implementation has always done a bad job at being compliant with markdown/commonmark specs.

theres a usecase to fix this in meta too. in meta we have been relying a large part on deprecated draft js markdown convertion. we have recently started shifting to lexical markdown and this was one of the bugs we have to fix. so it is timely tohave this PR merged

linsanyi:
However, when the exported data is processed by the LLM, the format has now changed to end_code => end\\_code, which may cause the LLM to fail to correctly understand the meaning of this prompt

PR #7359, opened 2025-03-21
[lexical-react][lexical-playground] Feature: Makes an inline node more easily linkable

sescandell:
## Description

Currently if you want to put a link around an image, you need to be clever and "select" around the image. It works in a way, but is not easy to use neither well presented.
Thanks to these changes, we globally allow to put a link on an inline node such as the ImageNode from the playground
### Before
_Nothing to show_

### After

image

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -29,6 +29,7 @@ import {
   $applyNodeReplacement,
   $getSelection,
   $isElementNode,
+  $isNodeSelection,
   $isRangeSelection,
   $normalizeSelection__EXPERIMENTAL,
   $setSelection,
@@ -562,9 +563,13 @@ export function $toggleLink(
   const rel = attributes.rel === undefined ? 'noreferrer' : attributes.rel;
   const selection = $getSelection();
 
-  if (!$isRangeSelection(selection)) {
+  if (
+    !$isRangeSelection(selection) &&
+    (!$isNodeSelection(selection) || !selection.getNodes()[0].isInline())
+  ) {
     return;
   }
+
   const nodes = selection.extract();
 
   if (url === null) {
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -28,7 +28,9 @@ import {
 import {
   $applyNodeReplacement,
   $getSelection,
+  $isDecoratorNode,
   $isElementNode,
+  $isInlineElementOrDecoratorNode,
   $isNodeSelection,
   $isRangeSelection,
   $normalizeSelection__EXPERIMENTAL,
@@ -502,6 +504,21 @@ function $getPointNode(point: Point, offset: number): LexicalNode | null {
   return null;
 }
 
+function $extractNodes(selection: BaseSelection | null) {
+  if (!selection) {
+    return [];
+  }
+
+  if ($isRangeSelection(selection)) {
+    return selection.extract();
+  }
+
+  if ($isNodeSelection(selection)) {
+    return selection.extract().filter($isInlineElementOrDecoratorNode);
+  }
+  return [];
+}
+
 /**
  * Preserve the logical start/end of a RangeSelection in situations where
  * the point is an element that may be reparented in the callback.
@@ -514,6 +531,7 @@ function $withSelectedNodes<T>($fn: () => T): T {
   if (!$isRangeSelection(initialSelection)) {
     return $fn();
   }
+
   const normalized = $normalizeSelection__EXPERIMENTAL(initialSelection);
   const isBackwards = normalized.isBackward();
   const anchorNode = $getPointNode(normalized.anchor, isBackwards ? -1 : 0);
@@ -563,14 +581,11 @@ export function $toggleLink(
   const rel = attributes.rel === undefined ? 'noreferrer' : attributes.rel;
   const selection = $getSelection();
 
-  if (
-    !$isRangeSelection(selection) &&
-    (!$isNodeSelection(selection) || !selection.getNodes()[0].isInline())
-  ) {
+  const nodes = $extractNodes(selection);
+
+  if (nodes.length === 0) {
     return;
   }
-
-  const nodes = selection.extract();
 
   if (url === null) {
     // Remove LinkNodes
@@ -632,6 +647,12 @@ export function $toggleLink(
         updateLinkNode(parentLinkNode);
         continue;
       }
+
+      // Pas sûr....
+      if ($isDecoratorNode(node) && !node.isInline()) {
+        continue;
+      }
+
       if ($isElementNode(node)) {
         if (!node.isInline()) {
           // Ignore block nodes, if there are any children we will see them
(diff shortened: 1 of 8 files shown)


packages/lexical-link/src/index.ts
```diff
@@ -562,9 +563,13 @@ export function $toggleLink(
   const rel = attributes.rel === undefined ? 'noreferrer' : attributes.rel;
   const selection = $getSelection();
 
-  if (!$isRangeSelection(selection)) {
+  if (
+    !$isRangeSelection(selection) &&
+    (!$isNodeSelection(selection) || !selection.getNodes()[0].isInline())
```
etrepum, line 568: This should really check all of the nodes since they're all processed later, e.g. `!selection.getNodes().some((node) => !node.isInline())`. Since you're using this sort of pattern in multiple places probably a utility function in lexical-selection makes sense.

packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -70,8 +71,13 @@ function FloatingLinkEditor({
 
   const $updateLinkEditor = useCallback(() => {
     const selection = $getSelection();
-    if ($isRangeSelection(selection)) {
-      const node = getSelectedNode(selection);
+    const node = $isRangeSelection(selection)
+      ? getSelectedNode(selection)
+      : $isNodeSelection(selection) && selection.getNodes()[0]?.isInline()
+      ? selection.getNodes()[0]
```
etrepum, line 77: incomplete check

packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -105,10 +111,17 @@ function FloatingLinkEditor({
       const domRect: DOMRect | undefined =
         nativeSelection.focusNode?.parentElement?.getBoundingClientRect();
       if (domRect) {
-        domRect.y += 40;
         setFloatingElemPositionForLinkEditor(domRect, editorElem, anchorElem);
       }
       setLastSelection(selection);
+    } else if ($isNodeSelection(selection)) {
+      const domRect = editor
+        .getElementByKey(selection.getNodes()[0].getKey())
```
etrepum, line 119: also incomplete check

packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -311,37 +324,50 @@ function useFloatingLinkEditorToolbar(
...
-        }
+      if (!selection) {
+        setIsLink(false);
+        return;
+      }
+
+      const focusNode = $isRangeSelection(selection)
+        ? getSelectedNode(selection)
+        : $isNodeSelection(selection) && selection.getNodes()[0]?.isInline()
+        ? selection.getNodes()[0]
```
etrepum, line 335: incomplete check

etrepum (commented):
Seems like a great start, it would be nice to clean up the functionality so that it's more correct in the general case. I would recommend adding some function(s) to @lexical/selection to facilitate the shared logic between these different modules so that this code is simpler, and that it's easier to do these things elsewhere too

packages/lexical-link/src/index.ts
sescandell: Does it make sense to say that if `getNodes()` contains more than one element, we should not allow creating a link? Even if all the elements are "Inline", it doesn't really make sense to say "transform this into a link" since they could be completely separate.

Am I missing something?

etrepum: In the RangeSelection case it will create multiple LinkNode if your selection crosses block boundaries, so having it work differently in the NodeSelection case doesn't really make sense either.

sescandell: OK, I didn’t notice that.

I’ll look into how to update everything.

Thanks! 👍

packages/lexical-playground/src/utils/getSelectedNode.ts
```diff
@@ -6,22 +6,61 @@
...
+
     return $isAtNodeEnd(anchor) ? anchorNode : focusNode;
   }
+
+  if ($isNodeSelection(selection)) {
+    const nodes = selection
+      .getNodes()
+      .filter(
+        (n) => $isTextNode(n) || $isElementNode(n) || $isDecoratorNode<T>(n),
+      );
```
etrepum, line 49: What's the purpose of this? To avoid LineBreakNode? This filter is why the code fails typecheck because the guard isn't implied with this version of TypeScript.

packages/lexical-playground/src/utils/getSelectedNode.ts
```diff
@@ -6,22 +6,61 @@
...
+
+    if (nodes.length === 0) {
+      return null;
+    }
+
+    if (nodes.length === 1) {
```
etrepum, line 55: There's no reason to special-case this, and isBackward isn't implemented for NodeSelections so it all ends up being nodes[0] either way

packages/lexical-playground/src/utils/getSelectedNode.ts
sescandell: Actually, this update wasn’t intended to be reviewed yet — I’m still working on it. Sorry, I shouldn’t have pushed it.

To answer your question: no, I can’t limit it to $isTextNode and $isElementNode because, for the primary objective of this PR, this wouldn’t work for ImageComponent (which is a DecoratorNode).

The goal is not to restrict it to specific blocks. In the initial implementation, getSelectedNode was designed to return either a TextNode or an ElementNode, and I wanted to stay as close as possible to that. However, in a NodeSelection, it can actually return more. For formatting purposes (e.g., links, formatting, toolbar), DecoratorNode should be allowed to support components like ImageComponent.

That said, after looking at your comment (and the code), I realize that, aside from LineBreakNode, there aren’t any other relevant types of elements.

I’m still working on this. I’ll mention you once I think it’s ready for review.

Thanks!

sescandell: Hi @etrepum,

I’ve removed the filter and updated the return type of getSelectedNode.
In previous usages, all cases either already checked the type of the node before using it or didn’t require the check at all.

Let me know your thoughts.

sescandell: @etrepum, do you need anything else from me for this PR?

I noticed that another very similar PR has been merged, but this one also handles the toolbar link button update, prevents adding a link around a "non-inline node", and improves the floating link editor positioning for non-textual nodes.

Would you like me to update this PR to align with the latest changes on master, or are these enhancements not of interest? Just asking to avoid unnecessary work.

Thanks!

etrepum: This one hasn't been reviewed and merged because the tests are not passing. If you merge with master and get everything passing then I'd be happy to review the additional features this PR provides

etrepum:

PR #7366, opened 2025-03-23
[lexical-link][lexical-playground] Feature: Add support for image links via NodeSelection

### LINKED ISSUE #4645: Feature: Link over an image
Allow link on an image. So currently you can insert image in the editor but you can not link the image to some link which is a common practice in HTML.

kirandash:
### Current Behavior
Currently, the `$toggleLink` function in `@lexical/link` package only supports `RangeSelection`, making it impossible to add links to images or other nodes that use `NodeSelection`. The `FloatingLinkEditorPlugin` also doesn't show up when clicking on images, preventing users from adding or editing links on images.

### Changes in this PR
This PR adds support for linking images in Lexical by:
1. Extending `$toggleLink` to handle both `RangeSelection` and `NodeSelection`
2. Updating FloatingLinkEditorPlugin to:
   - Show link toolbar when clicking images
   - Handle link editing for image nodes
   - Position toolbar correctly for image selections
3. Adding comprehensive test coverage for:
   - Adding/editing/removing links on single images

Closes #4645

### Before

https://github.com/user-attachments/assets/3ff74605-82ca-4573-8d83-9c5d25e814bd

### After

https://github.com/user-attachments/assets/53a768fe-48fe-4d2f-bf6a-61e98ac69732

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -576,47 +576,43 @@ export function $toggleLink(
       return;
     }
 
-    const node = nodes[0];
-    if (url === null) {
-      // If we're removing the link and the node is inside a link, move it out
-      const linkParent = $findMatchingParent(
-        node,
-        (parent): parent is LinkNode =>
-          !$isAutoLinkNode(parent) && $isLinkNode(parent),
-      );
-      if (linkParent) {
-        linkParent.insertBefore(node);
-        if (linkParent.getChildren().length === 0) {
-          linkParent.remove();
+    // Handle all selected nodes
+    nodes.forEach((node) => {
+      if (url === null) {
+        // Remove link
+        const linkParent = $findMatchingParent(
+          node,
+          (parent): parent is LinkNode =>
+            !$isAutoLinkNode(parent) && $isLinkNode(parent),
+        );
+        if (linkParent) {
+          linkParent.insertBefore(node);
+          if (linkParent.getChildren().length === 0) {
+            linkParent.remove();
+          }
+        }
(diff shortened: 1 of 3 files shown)


etrepum:
A NodeSelection can have more than one node selected (you can shift-click to add nodes to the existing selection), it might be surprising if only the first selected node gets wrapped in a LinkNode

kirandash:
> A NodeSelection can have more than one node selected (you can shift-click to add nodes to the existing selection), it might be surprising if only the first selected node gets wrapped in a LinkNode

Thanks for the feedback! I've updated the `$toggleLink` function to handle all nodes in a `NodeSelection` and added test coverage to verify this behavior.

Please let me know if you'd like me to make any additional improvements.

packages/lexical-link/src/index.ts
```diff
@@ -562,9 +563,60 @@ export function $toggleLink(
...
+      if (url === null) {
+        // Remove link
+        const linkParent = $findMatchingParent(
+          node,
+          (parent): parent is LinkNode =>
+            !$isAutoLinkNode(parent) && $isLinkNode(parent),
+        );
+        if (linkParent) {
+          linkParent.insertBefore(node);
+          if (linkParent.getChildren().length === 0) {
```
etrepum, line 590: There's a faster path to see if a node is empty or not, calling getChildren requires fetching the latest version of each child from the editor state whereas isEmpty doesn't involve any other nodes

```suggestion
          if (linkParent.isEmpty()) {
```

etrepum (approved):

Issue #7383, opened 2025-03-26
Bug: [lexical-playground] clear-formatting doesn't clear text-left/right/center alignment when no formatting to text


harshmetkel24:
Lexical version: all

## Steps To Reproduce

1. Add any text to playground
2. right align it, without any other formatting like bold, font-size, etc
3. now try to do clear formatting

Link to code example:
I've verified this fix
<img width="1365" alt="Image" src="https://github.com/user-attachments/assets/c20fcabc-6b32-4238-b06c-1daf54204483" />

## The current behavior
pressing `cmd+\` do left align text until bold or any other formatting is applied to text.

https://github.com/user-attachments/assets/61607411-d177-4f3c-9caf-2db195b99f43

## The expected behavior
pressing `cmd+\` clears right align and text is left aligned

https://github.com/user-attachments/assets/aef76a8b-ff17-4c7b-a098-92e2d8340b00

## Impact of fix
clear formatting will be improved and cover wider range

harshmetkel24:
I've worked out ths solution and tested as well. Can you guys assign this to me

kirandash:
> I've worked out ths solution and tested as well. Can you guys assign this to me

Thanks for your interest! Just to note - I'm not a core maintainer, but from what I've seen in the Lexical community, you can go ahead and submit a PR with your solution when it's ready. No formal assignment needed. The maintainers will review it when submitted.

harshmetkel24:
> > I've worked out ths solution and tested as well. Can you guys assign this to me
> 
> Thanks for your interest! Just to note - I'm not a core maintainer, but from what I've seen in the Lexical community, you can go ahead and submit a PR with your solution when it's ready. No formal assignment needed. The maintainers will review it when submitted.

Thanks for the reply will do that 🙌🏾

PR #7384, opened 2025-03-26
[lexical-playground] Feature: clear blockelement formatting along with textNode, even when #7383

harshmetkel24:
## Steps To Reproduce

1. Add any text to playground
2. right align it, without any other formatting like bold, font-size, etc
3. now try to do clear formatting

## The current behavior
pressing `cmd+\` do left align text until bold or any other formatting is applied to text.

https://github.com/user-attachments/assets/61607411-d177-4f3c-9caf-2db195b99f43

## The expected behavior
pressing `cmd+\` clears right align and text is left aligned

https://github.com/user-attachments/assets/aef76a8b-ff17-4c7b-a098-92e2d8340b00

## Impact of fix
clear formatting will be improved and cover wider range

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
@@ -276,7 +276,11 @@ export const clearFormatting = (editor: LexicalEditor) => {
           }
           if (textNode.__format !== 0) {
             textNode.setFormat(0);
-            $getNearestBlockElementAncestorOrThrow(textNode).setFormat('');
+          }
+          const nearestBlockElement =
+            $getNearestBlockElementAncestorOrThrow(textNode);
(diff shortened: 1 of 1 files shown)


etrepum (commented):
For this kind of change it would be best if there was a test to show that this works as expected, which will help ensure that it continues to keep working that way

harshmetkel24:
> For this kind of change it would be best if there was a test to show that this works as expected, which will help ensure that it continues to keep working that way

Okay can you suggest some existing related test's file. Actually this is first time I am contributing

harshmetkel24:
@etrepum I've added tests 
1. case which was already working, having BIU formatting along with right/center align clears formatting.
2. case when only right/center align is there clear formatting has no effect

etrepum (approved):
Looks good, nice work!

harshmetkel24:
> Looks good, nice work!

Thanks 🙌🏾

PR #7388, opened 2025-03-26
[lexical-playground] Bug Fix: Use natural dimensions for inherited image size

### LINKED ISSUE #7376: Bug: Inserting Image via url
Inserting image via url inside the [playground dev](https://playground.lexical.dev/) ..for the very first time ..do not show the image.

Any change further (typing or adding another image).. renders correctly.

Steps to reproduce:
- Open playground dev . Add image via url.

Attaching screenshot.

Image

Here the DOM tree clearly shows there is an image but its not rendering in the editor.

kirandash:
### Current Behavior
Currently, when an image is inserted with 'inherit' dimensions in the Lexical playground, it defaults to arbitrary 200x200 dimensions without considering the image's natural size. This causes visual inconsistency and incorrect image rendering on initial load.

### Changes in this PR
1. Using the image's natural dimensions (naturalWidth/naturalHeight) when width/height is set to 'inherit'
2. Maintaining proper aspect ratio during scaling
3. Adding proper dimension detection on both initial mount and image load
4. Keeping the existing maxWidth and maxHeight (500px) constraints for editor usability
5. Using fallback dimensions (200x200) only while the image is loading

Closes #7376

### Before

image

### After

image

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Images.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Images.spec.mjs
@@ -418,7 +418,7 @@ test.describe('Images', () => {
                 alt="lexical logo"
                 draggable="false"
                 src="https://lexical.dev/img/logo.svg"
-                style="height: inherit; max-width: 500px; width: inherit;" />
+                style="height: 112px; max-width: 500px; width: 500px" />
             </div>
           </span>
           <span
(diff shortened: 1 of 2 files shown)


etrepum (approved):
Looks great! I notice that the tests depend on a particular URL on the website which could make things fail unexpectedly in the future but it doesn't seem to be the only test that's doing it so we can worry about that another time

PR #7390, opened 2025-03-27
[RFC][lexical-markdown] Replace whitespace with &#160 when the string consists only of whitespace.

### LINKED ISSUE #6526: Bug: Styling format should be removed from the whitespace
Hello ,
There an issue that I found out after I copy paste from Microsoft word.
If you apply a styling to a whitespace, it results with asterisk **** or ~~~~ in the markdown.

Lexical version: 0.17.0

## The current behavior
Screen Shot 2024-08-19 at 9 31 41 PM
Screen Shot 2024-08-19 at 9 31 52 PM

## The expected behavior
Screen Shot 2024-08-19 at 9 32 07 PM

## Impact of fix

I am new to lexical, and I am having a hard time to fix the issue for a company. Is there a solution such as Markdown Transformer to add it to the array of transformers ?

```
export const PLAYGROUND_TRANSFORMERS: Array<Transformer> = [
  HR,
  ...ELEMENT_TRANSFORMERS,
  ...TEXT_FORMAT_TRANSFORMERS,
  ...TEXT_MATCH_TRANSFORMERS
];

```

lilshady:
## Description
Currently, `exportTextFormat` in `MardownExports` trim the input first to handle the case looking like `**   foo   **`
An edge case here is if the string consists only of whitespaces, the trimmed string will be empty.
Based on the discussion [here](https://github.com/facebook/lexical/issues/6526#issuecomment-2725146825), one possible solution is replacing the whitespace with `&#160;`, which is also compatible with CommonMark.

Closes #6526 

### Before

https://github.com/user-attachments/assets/8a8882e7-db70-4401-9461-31db07fc5b88

### After

https://github.com/user-attachments/assets/772fe8fa-634c-40ce-a217-3a8db2eca2e9

------------------
Test other whitespaces

https://github.com/user-attachments/assets/b0497b32-c613-42b7-a288-b4b026468fad

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -25,6 +25,7 @@ import {
 } from './MarkdownTransformers';
 import {isEmptyParagraph, transformersByType} from './utils';
 
+const WHITESPACE_CHAR = '&#160;';
 /**
  * Renders string from markdown. The selection is moved to the start after the operation.
  */
@@ -205,14 +206,19 @@ function exportTextFormat(
   // Where it would be invalid markdown to generate: "**   foo   **"
   // We instead want to trim the whitespace out, apply formatting, and then
   // bring the whitespace back. So our returned string looks like this: "   **foo**   "
+  // for whitespace only string case like: "  ", we replace the whitespace with &#160;
   const frozenString = textContent.trim();
   let output = frozenString;
+  const isWhitespaceOnly = frozenString.length === 0;
+
+  if (isWhitespaceOnly) {
+    output = textContent.replace(/\s/g, WHITESPACE_CHAR);
+  }
 
   if (!node.hasFormat('code')) {
     // Escape any markdown characters in the text content
     output = output.replace(/([*_`~\\])/g, '\\$1');
   }
-
   // the opening tags to be added to the result
   let openingTags = '';
   // the closing tags to be added to the result
@@ -286,6 +292,9 @@ function exportTextFormat(
   }
 
   output = openingTags + output + closingTagsAfter;
+  if (isWhitespaceOnly) {
+    return closingTagsBefore + output;
+  }
   // Replace trimmed version of textContent ensuring surrounding whitespace is not modified
   return closingTagsBefore + textContent.replace(frozenString, () => output);
 }
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -209,9 +209,10 @@ function exportTextFormat(
   // for whitespace only string case like: "  ", we replace the whitespace with &#160;
   const frozenString = textContent.trim();
   let output = frozenString;
-  const isWhitespaceOnly = frozenString.length === 0;
+  const isFormattedWhitespace =
+    frozenString.length === 0 && node.getFormat() !== 0;
 
-  if (isWhitespaceOnly) {
+  if (isFormattedWhitespace) {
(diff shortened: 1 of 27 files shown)


packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -205,14 +206,20 @@ function exportTextFormat(
...
   // We instead want to trim the whitespace out, apply formatting, and then
   // bring the whitespace back. So our returned string looks like this: "   **foo**   "
+  // for whitespace only string case like: "  ", we replace the whitespace with &#160;
   const frozenString = textContent.trim();
   let output = frozenString;
+  const isFormattedWhitespace =
+    frozenString.length === 0 && node.getFormat() !== 0;
+
+  if (isFormattedWhitespace) {
+    output = textContent.replace(/\s/g, WHITESPACE_CHAR);
```
etrepum, line 216: This is a little strange because it replaces all whitespace with nbsp, is there no way to use an encoding that round-trips losslessly?

lilshady: if we want to convert the whitespace to its corresponding entity code, one way I think of is something like

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -205,14 +205,23 @@ function exportTextFormat(
...
+  // for whitespace only string case like: "  ", we replace the whitespace with &#160;
   const frozenString = textContent.trim();
   let output = frozenString;
+  const isFormattedWhitespace =
+    frozenString.length === 0 && node.getFormat() !== 0;
+
+  if (isFormattedWhitespace) {
+    // Convert whitespaces to the code entity
+    output = [...textContent]
+      .map((char) => '&#' + char.charCodeAt(0) + ';')
```
etrepum, line 217: Let's use fromCodePoint even though it doesn't make any difference for these specific whitespace characters we'll be encoding. HTML entities are based on code points and not UTF-16 code units. Might as well correctly support decoding any entity with this syntax.

```suggestion
      .map((char) => '&#' + char.codePointAt(0) + ';')
```

packages/lexical-markdown/src/importTextTransformers.ts
```diff
@@ -133,6 +133,10 @@ export function importTextTransformers(
 
   // Handle escape characters
   const textContent = textNode.getTextContent();
-  const escapedText = textContent.replace(/\\([*_`~])/g, '$1');
+  const escapedText = textContent
+    .replace(/\\([*_`~])/g, '$1')
+    .replace(/&#(\d+);/g, (_, codePoint) => {
+      return String.fromCharCode(codePoint);
```
etrepum, line 139:
```suggestion
      return String.fromCodePoint(codePoint);
```

etrepum (commented):
An alternative encoding we could use is to instead apply this escape to any leading and trailing whitespace instead of doing a trim at all, which might be preferable since it would better preserve the original document. What do you think?

packages/lexical-markdown/src/importTextTransformers.ts
lilshady: Thanks! TIL.

lilshady:
> An alternative encoding we could use is to instead apply this escape to any leading and trailing whitespace instead of doing a trim at all, which might be preferable since it would better preserve the original document. What do you think?

Yes, that also makes sense to me, let me make this change.

lilshady:
3123

> An alternative encoding we could use is to instead apply this escape to any leading and trailing whitespace instead of doing a trim at all, which might be preferable since it would better preserve the original document. What do you think?

accidentally messed up this PR, have created a new one: #7400

PR #7393, opened 2025-03-27
[lexical-rich-text] Bug Fix: Fix right and up arrow key navigation with decorator nodes

### LINKED ISSUE #7392: Bug: right click blocks cursor after nodeSelction
Initially facing the issue I thought this is because of some `KEY_ARROW_RIGHT_COMMAND`. But I tried locally starting playground to every `registerCommand`  was not able to find any culprit. If someone have any idea around possible causes for this please help me with that

Lexical version: all

## Steps To Reproduce

1. Add any decorator node (my guess this issue is related to decorator node, might be only related to nodeSelection as well) for demo purpose I am using `ExcaliDraw`
2. Now select excalidraw by clicking over it
3. Next press right arrow key you should observe nothing happens
4. but as can be seen in video if we click left key twice cursor moves to side of decorator node and now clicking right arrow key will bring cursor correctly to right of decorator node as expected.

## The current behavior

https://github.com/user-attachments/assets/e5a04258-8438-4334-9b57-db6c1b49e9b5

## The expected behavior
Cursor should not lost instead we should be able to move cursor to right of decorator node.

## Impact of fix

Keyboard accessibility will improve if this is fixed

kirandash:
### Current behavior:
- When clicking on an Excalidraw decorator, the button becomes active and becomes the target of keyboard events
- The right arrow key handler checks if the target is inside a decorator (`!$isTargetWithinDecorator`) and ignores the event
- This check was originally added in PR #2895 for all arrow keys to fix keyboard controls on internal decorators
- Later, PR #2929 removed this check from left arrow key, but it remained for right arrow key
- This creates an inconsistent behavior where right arrow doesn't work initially, but works after pressing left arrow

### Changes in this PR:
- Remove the `!$isTargetWithinDecorator` check from `KEY_ARROW_RIGHT_COMMAND` to match the behavior of `KEY_ARROW_LEFT_COMMAND`
- This allows consistent bi-directional arrow key navigation when decorator nodes are selected
- The change follows the same pattern as PR #2929 which successfully removed this check from left arrow key

Closes #7392

### Before
1. Add an Excalidraw node to the editor
2. Click to select the Excalidraw node
3. Press right arrow key - nothing happens
4. Press left arrow key - selection moves left
5. Press right arrow key - now it works

https://github.com/user-attachments/assets/75640105-e8c5-429f-91be-6cd5baafbb0a

### After
1. Add an Excalidraw node to the editor
2. Click to select the Excalidraw node
3. Press right arrow key - selection moves right
4. Press left arrow key - selection moves left
5. Both arrow keys work consistently in either direction

https://github.com/user-attachments/assets/82eaa525-184f-4d45-b854-489cbdd3ea83

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -747,10 +747,7 @@ export function registerRichText(editor: LexicalEditor): () => void {
       KEY_ARROW_UP_COMMAND,
       (event) => {
         const selection = $getSelection();
-        if (
-          $isNodeSelection(selection) &&
-          !$isTargetWithinDecorator(event.target as HTMLElement)
-        ) {
+        if ($isNodeSelection(selection)) {
           // If selection is on a node, let's try and move selection
           // back to being a range selection.
           const nodes = selection.getNodes();
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -839,10 +839,7 @@ export function registerRichText(editor: LexicalEditor): () => void {
       KEY_ARROW_RIGHT_COMMAND,
       (event) => {
         const selection = $getSelection();
-        if (
-          $isNodeSelection(selection) &&
-          !$isTargetWithinDecorator(event.target as HTMLElement)
-        ) {
+        if ($isNodeSelection(selection)) {
           // If selection is on a node, let's try and move selection
           // back to being a range selection.
           const nodes = selection.getNodes();


etrepum:
IIRC the up arrow also had this when I briefly looked earlier

kirandash:
> IIRC the up arrow also had this when I briefly looked earlier

Thanks for catching that! I've updated the PR to remove the decorator check from both the up and right arrow commands to ensure consistent navigation behavior across all directions.

etrepum (approved):
Looks great, will merge assuming extended tests are in good shape!

PR #7395, opened 2025-03-27
[lexical-markdown] Bug Fix: Prevent Markdown shortcuts from applying to code-formatted text

### LINKED ISSUE #7349: Bug: Markdown shortcuts should not apply to code nodes
Lexical version: 0.28.0

Type the following text into the playground:

```
`void*` and `int*`
```

## The current behavior

Produces: <code>void</code><i> and <code>int</code></i>

Notice that the `*` inside `void*` was matched to the `*` inside `int*`.

## The expected behavior

Expected to produce: <code>void*</code> and <code>int*</code>

## Impact of fix

This would impact anyone using markdown shortcuts for typing inline code where you would not expect your code formatted text to be modified as you type.

takuyakanbr:
## Description

Prevent text-format Markdown shortcuts from matching with text in code-formatted text nodes.

For example, this ensures that typing the following:
```
`void*` and `int*`
```
results in "<code>void*</code> and <code>int*</code>" instead of "<code>void</code><i> and <code>int</code></i>".

Closes #7349

## Test plan

Run E2E tests: `npm run test-e2e-chromium Markdown.spec`

### Before

https://github.com/user-attachments/assets/51f3483c-2fa8-419f-affb-2d1e00b2ccd4

### After

https://github.com/user-attachments/assets/134489a7-d770-48e5-b771-69351e9ddf4c

### DIFF
--- a/packages/lexical-markdown/src/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/MarkdownShortcuts.ts
@@ -254,6 +254,9 @@ function $runTextFormatTransformers(
       }
 
       if ($isTextNode(sibling)) {
+        if (sibling.hasFormat('code')) {
+          continue;
+        }
         const siblingTextContent = sibling.getTextContent();
         openNode = sibling;
         openTagStartIndex = getOpenTagStartIndex(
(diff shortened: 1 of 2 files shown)


etrepum:
There definitely is some testing for markdown shortcuts in the e2e suite, there just isn't a single file that focuses specifically on testing them. You'll see stuff like:

```js
    await page.keyboard.type('> Testing for quote node');
```

or 

```js
    await page.keyboard.type('[link](https://lexical.dev)');
```

potatowagon (approved):

etrepum (approved):
Will merge assuming other extended tests pass, nice job!

PR #7397, opened 2025-03-28
[lexical] Chore: Change $getTextNodeOffset invariant to warn in prod (error in __DEV__)

etrepum:
## Description

Change `$getTextNodeOffset` boundary checking to clamp with minified warning in prod and error in `__DEV__`. Adds a new `devInvariant` function with associated transform to provide this error-in-dev-warning-in-prod functionality that re-uses the same error map and transform-error-messages babel plugin.

## Test plan

We don't really have a facility to easily test behavior that only happens in `!__DEV__` but the transform itself is tested

### DIFF
--- a/packages/lexical/src/caret/LexicalCaret.ts
+++ b/packages/lexical/src/caret/LexicalCaret.ts
@@ -902,14 +902,21 @@ export function $getTextNodeOffset(
   offset: number | CaretDirection,
 ): number {
   const size = origin.getTextContentSize();
-  const numericOffset =
+  let numericOffset =
     offset === 'next' ? size : offset === 'previous' ? 0 : offset;
-  invariant(
-    numericOffset >= 0 && numericOffset <= size,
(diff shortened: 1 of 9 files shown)


packages/lexical/src/caret/LexicalCaret.ts
```diff
@@ -902,14 +903,19 @@ export function $getTextNodeOffset(
...
+  if (numericOffset < 0 || numericOffset > size) {
+    devInvariant(
+      false,
+      '$getTextNodeOffset: invalid offset %s for size %s at key %s',
+      String(offset),
+      String(size),
+      origin.getKey(),
+    );
+    // Clamp invalid offsets in prod
+    numericOffset = numericOffset < 0 ? 0 : size;
```
etrepum, line 917: This branch has no tests, review carefully

lilshady: This looks good to me!

lilshady (approved):

PR #7399, opened 2025-03-28
 [lexical-playground] Feature: opening component picker on insert click

harshmetkel24:
Opening component picker menu to next line on click to insert button in dragblock plugin. 
**Inspiration: Notion**

### Before
Clicking on insert ("+") button only add a new line below or above the current line.

https://github.com/user-attachments/assets/cc59c497-5e34-4a1b-9053-f79a20f7cae9

### After
Now user can open up component picker menu below and above current line clicking on "+" button. 

https://github.com/user-attachments/assets/32a9b1d0-b246-40ea-a1ba-16535a8663bf

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -228,19 +228,19 @@ pre::-webkit-scrollbar-thumb {
 }
 
 #options-button::after {
-  background-image: url(images/icons/gear.svg);
+  background-image: url(images/icons/gear.svg) !important;
 }
 
 #test-recorder-button::after {
-  background-image: url(images/icons/journal-code.svg);
+  background-image: url(images/icons/journal-code.svg) !important;
 }
 
 #paste-log-button::after {
-  background-image: url(images/icons/clipboard.svg);
+  background-image: url(images/icons/clipboard.svg) !important;
 }
 
 #docs-button::after {
-  background-image: url(images/icons/file-earmark-text.svg);
+  background-image: url(images/icons/file-earmark-text.svg) !important;
 }
 
 #test-recorder-button-snapshot {
@@ -248,15 +248,15 @@ pre::-webkit-scrollbar-thumb {
 }
 
 #test-recorder-button-snapshot::after {
-  background-image: url(images/icons/camera.svg);
+  background-image: url(images/icons/camera.svg) !important;
 }
 
 #test-recorder-button-copy::after {
-  background-image: url(images/icons/clipboard.svg);
+  background-image: url(images/icons/clipboard.svg) !important;
 }
 
 #test-recorder-button-download::after {
-  background-image: url(images/icons/download.svg);
+  background-image: url(images/icons/download.svg) !important;
 }
 
 .typeahead-popover {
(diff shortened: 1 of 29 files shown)


etrepum (commented):
It would be nice if it didn't also insert the `/` but it is a simple implementation this way. Anyone else have thoughts on this? @ivailop7?

ivailop7:
> It would be nice if it didn't also insert the `/` but it is a simple implementation this way. Anyone else have thoughts on this? @ivailop7?

I agree with @etrepum, would be nice if this component was positioned relative to the "+" and append the node vs append a paragraph and type "/" for the typeahead to trigger. Essentially the menu gets coupled to people needing to support the "/" typeahead behaviour for this + menu to work.

harshmetkel24:
After reviewing the current implementation of `ComponentPickerPlugin`, it appears that integrating relative positioning for the menu might be challenging. The`LexicalTypeAheadMenuPlugin` is designed to function entirely based on the / trigger logic. Given this, the available options for positioning the menu are limited to what is provided by the existing component menu component.

This is my initial analysis upon reviewing the implementation. However, I will explore potential solutions to determine the best approach.

harshmetkel24:
@etrepum @ivailop7 
I've updated code to not append '/' and open a menu absolutely positioned instead

etrepum:
Looks like it's still not passing eslint, you can run those checks locally with `npm run ci-check`. The autofixer should be able to handle most of those fixes

harshmetkel24:
https://github.com/Sparticuz/chromium/issues/63#issuecomment-1456569524

Yeah I tried that but I keep getting this error
image

etrepum:
It looks like that happens if you have an M1 mac without rosetta installed and the app is x86_64 only (which it looks like flow is). `softwareupdate --install-rosetta` should fix the root cause.

You could also run the components of the CI check without flow.

```
    "ci-check": "npm-run-all --parallel tsc tsc-extension flow prettier lint",
```

`npm run prettier`, `npm run lint`, `npm run tsc`, etc.

packages/lexical-playground/src/index.css
```diff
@@ -228,35 +228,35 @@ pre::-webkit-scrollbar-thumb {
 }
 
 #options-button::after {
-  background-image: url(images/icons/gear.svg);
+  background-image: url(images/icons/gear.svg) !important;
```
ivailop7, line 231: why do we need these !important overrides ?

packages/lexical-playground/src/index.css
```diff
@@ -355,6 +355,37 @@ pre::-webkit-scrollbar-thumb {
   width: 200px;
 }
 
+.floating-component-picker-menu {
```
ivailop7, line 358: I think having the same styling as the regular component picker and just overriding the position or nessessarity properties makes for smaller code base.

ivailop7: you have the styles in index.css and inside the draggableblockplugin folder too. I think this styling could be removed from the index.css and have it properly encapsulated with the plugin.

ivailop7:
I'm also seeing an issue if for node selected is meant to show a dialog - image, table, columns layount, it just appends a paragraph right now, rather than show the correct dialog.

packages/lexical-playground/src/index.css
harshmetkel24: I needed to do that becuase icons are displayed with className .icon. In dragblockPlugin we use .icon class to show dragger icon. But yes instead of adding `!imprtant` I can change class name to resolve className collision

packages/lexical-playground/src/index.css
harshmetkel24: Yess sure will do that

harshmetkel24:
> I'm also seeing an issue if for node selected is meant to show a dialog - image, table, columns layount, it just appends a paragraph right now, rather than show the correct dialog.

oh yes, will check that as well

PR #7400, opened 2025-03-29
[RFC][lexical-markdown] Replace whitespace with code point when the string has leading and trailing whitespaces

### LINKED ISSUE #6526: Bug: Styling format should be removed from the whitespace

lilshady:
## Description
Currently, `exportTextFormat` in `MardownExports` trim the input first to handle the case looking like `**   foo   **`
An edge case here is if the string consists only of whitespaces, the trimmed string will be empty.
Based on the discussion [here](https://github.com/facebook/lexical/issues/6526#issuecomment-2725146825), one possible solution is replacing the whitespace with its corresponding code points,  which is also compatible with CommonMark.

Also, modify some test cases that have leading and trailing whitespaces.
Closes #6526 

### Before

https://github.com/user-attachments/assets/8a8882e7-db70-4401-9461-31db07fc5b88

### After

https://github.com/user-attachments/assets/772fe8fa-634c-40ce-a217-3a8db2eca2e9

------------------
Test other whitespaces

https://github.com/user-attachments/assets/b0497b32-c613-42b7-a288-b4b026468fad

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -203,10 +203,13 @@ function exportTextFormat(
 ): string {
   // This function handles the case of a string looking like this: "   foo   "
   // Where it would be invalid markdown to generate: "**   foo   **"
-  // We instead want to trim the whitespace out, apply formatting, and then
-  // bring the whitespace back. So our returned string looks like this: "   **foo**   "
-  const frozenString = textContent.trim();
-  let output = frozenString;
+  // If the node has no format, we use the original text.
+  // Otherwise, we escape leading and trailing whitespaces to their corresponding code points,
+  // ensuring the returned string maintains its original formatting, e.g., "**&#32;&#32;&#32;foo&#32;&#32;&#32;**".
+  let output =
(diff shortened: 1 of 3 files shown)


etrepum (approved):
Looks like it works well, nice job!

PR #7401, opened 2025-03-29
[Bug Fix] Clone a node and use  `$setSelection` instead of assigning dirty to true directly. 

lilshady:
## Description
Currently, we mark the selection dirty and force the selection back to it.
We freeze the selection object in `LexicalUpdates::$commitPendingUpdates` when in `DEV` mode for performance reasons
```javascript
 if (!pendingEditorState._readOnly) {
    pendingEditorState._readOnly = true;
    if (__DEV__) {
      handleDEVOnlyPendingUpdateGuarantees(pendingEditorState);
      if ($isRangeSelection(pendingSelection)) {
        Object.freeze(pendingSelection.anchor);
        Object.freeze(pendingSelection.focus);
      }
      Object.freeze(pendingSelection);
    }
  }
```
In some cases, we have observed this error:
```
Uncaught TypeError: Cannot assign to read only property 'dirty' of object '#<RangeSelection>'
```
To prevent this, we can clone a node and use  `$setSelection` instead of assigning dirty to true directly. 

## Test plan
existing tests.

### Before
Have seen the error above in console when in dev mode
### After
The error disappears.

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -34,6 +34,7 @@ import {
 import {
   $addUpdateTag,
   $onUpdate,
+  $setSelection,
   createUID,
   dispatchCommand,
   getCachedClassNameArray,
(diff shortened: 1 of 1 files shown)


packages/lexical/src/LexicalEditor.ts
```diff
@@ -1302,12 +1302,18 @@ export class LexicalEditor {
       // This ensures that iOS does not trigger caps lock upon focus
       rootElement.setAttribute('autocapitalize', 'off');
       updateEditorSync(this, () => {
-        const selection = $getSelection();
+        let selection = $getSelection();
```
etrepum, line 1305: This isn't re-used anywhere with the suggestion below

```suggestion
        const selection = $getSelection();
```

packages/lexical/src/LexicalEditor.ts
```diff
@@ -1302,12 +1302,18 @@ export class LexicalEditor {
...
           // Marking the selection dirty will force the selection back to it
-          selection.dirty = true;
+          // In DEV mode, we freeze the selection in $commitPendingUpdates
+          // so clone a node here to prevent an error.
+          if (__DEV__ && Object.isFrozen(selection)) {
+            selection = selection.clone();
+          }
+          if (!selection.dirty) {
+            selection.dirty = true;
+          }
```
etrepum, line 1316: This is just going to mask a bug in the code if it works at all, I think the correct solution would look more like this:

```suggestion
          if (!selection.dirty) {
            $setSelection(selection.clone());
          }
```

lilshady: Thanks! Checked the implementation of `$setSelection` and it makes more sense here.
TIL again

PR #7403, opened 2025-03-30
[lexical-devtools-core] Bug Fix: Update debug view to show KEY_ESCAPE_COMMAND immediately

### LINKED ISSUE #7379: Bug: `KEY_ESCAPE_COMMAND` is not logged in the playground debugger until it receives focus again
Lexical version: 0.29.0+dev.esm

## Steps To Reproduce

1. Open the Playground
2. Click/focus the editor
3. Press Escape
4. Observe that KEY_ESCAPE_COMMAND isn't logged in the debug view
5. Click outside the editor and back in it
6. KEY_ESCAPE_COMMAND now appears in the debug view before other commands, e.g. CLICK_COMMAND from the interaction in (5)

## The current behavior
KEY_ESCAPE_COMMAND is not logged in the debugger until it receives focus again

## The expected behavior
The command is logged immediately

## Impact of fix

Low, I think. It seems that the timing of KEY_ESCAPE_COMMAND listeners is working as expected, because if I attach a command handler to KEY_ESCAPE_COMMAND and log inside of it, it logs at the expected time (i.e. immediately after escape is pressed). I had originally thought there was a bug with the Escape key handler, but it seems to just be the debug view.

kirandash:
### Current behavior:
- The debug view in TreeView component only updates when the editor state changes
- Commands that don't modify state (like KEY_ESCAPE_COMMAND) are not displayed until the editor regains focus
- This creates confusion for developers trying to debug command interactions

### Changes in this PR:
- Added a new `commandsLog` prop to TreeView to track commands independently of state changes
- Modified the update logic to trigger view refresh when either editor state or commands log changes
- Preserved existing time-travel functionality by only recording state changes in timeline

Closes #7379

### Before

1. Open the Playground
2. Click/focus the editor
3. Press Escape
4. KEY_ESCAPE_COMMAND is not logged in the debug view
5. Click outside the editor and back in
6. Only then does KEY_ESCAPE_COMMAND appear in the debug view

### After

1. Open the Playground
2. Click/focus the editor
3. Press Escape
4. KEY_ESCAPE_COMMAND is immediately logged in the debug view
5. No need to click outside and back in to see the command

https://github.com/user-attachments/assets/8d8a8b9a-65a4-42b0-95ec-e6796bc842e6

### DIFF
--- a/packages/lexical-devtools-core/src/TreeView.tsx
+++ b/packages/lexical-devtools-core/src/TreeView.tsx
@@ -95,11 +95,15 @@ export const TreeView = forwardRef<
       lastCommandsLogRef.current !== commandsLog;
 
     if (shouldUpdate) {
+      // Check if it's a real editor state change
+      const isEditorStateChange = lastEditorStateRef.current !== editorState;
+
       lastEditorStateRef.current = editorState;
       lastCommandsLogRef.current = commandsLog;
       generateTree(showExportDOM);
 
-      if (!timeTravelEnabled && lastEditorStateRef.current !== editorState) {
+      // Only record in time travel if there was an actual editor state change
+      if (!timeTravelEnabled && isEditorStateChange) {
         setTimeStampedEditorStates((currentEditorStates) => [
           ...currentEditorStates,
           [Date.now(), editorState],
--- a/packages/lexical-devtools-core/src/TreeView.tsx
+++ b/packages/lexical-devtools-core/src/TreeView.tsx
@@ -27,6 +27,7 @@ export const TreeView = forwardRef<
     generateContent: (exportDOM: boolean) => Promise<string>;
     setEditorState: (state: EditorState, options?: EditorSetOptions) => void;
     setEditorReadOnly: (isReadonly: boolean) => void;
+    commandsLog?: string[];
   }
 >(function TreeViewWrapped(
   {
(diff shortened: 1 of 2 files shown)


packages/lexical-devtools-core/src/TreeView.tsx
```diff
@@ -85,12 +89,17 @@ export const TreeView = forwardRef<
...
+      lastEditorStateRef.current !== editorState ||
+      lastCommandsLogRef.current !== commandsLog;
+
+    if (shouldUpdate) {
       lastEditorStateRef.current = editorState;
+      lastCommandsLogRef.current = commandsLog;
       generateTree(showExportDOM);
 
-      if (!timeTravelEnabled) {
+      if (!timeTravelEnabled && lastEditorStateRef.current !== editorState) {
```
takuyakanbr, line 102: Wouldn't the second condition always be `false` due to L98?

kirandash: Good catch! You're right. this condition would always be `false` since we set `lastEditorStateRef.current = editorState` on L98. 

I've moved the editor state change detection before the reference update in the latest commit [`50d2d84` (#7403)](https://github.com/facebook/lexical/pull/7403/commits/50d2d84cf9490d0e46072684f6ce943221d0d47f), so we can properly distinguish between state changes and command-only updates.

Thanks for the review!

takuyakanbr (approved):

PR #7404, opened 2025-03-30
[lexical-react] Feature: Add option to disable first item auto-selection in menus

### LINKED ISSUE #7364: Feature: Do not select first item on render of LexicalMenu / LexicalTypeaheadMenuPlugin
## Description

The current logic does not support the "do not select any elements on render". It will be nice to have an option do nothing `on enter` for `LexicalTypeaheadMenuPlugin` at least.

https://github.com/facebook/lexical/blob/55cc35fcae11e61b0037f548d1c39ca79f28f4cf/packages/lexical-react/src/shared/LexicalMenu.ts#L336-L342

Current workaround - creating an empty(noop) + hidden first option :(

kirandash:
### Current behavior:
- LexicalMenu and LexicalTypeaheadMenuPlugin automatically select the first menu item when rendered
- Users have to work around this by creating a hidden no-op first option
- No built-in way to disable the auto-selection behavior

### Changes in this PR:
- Added new `autoSelectFirstItem` prop (defaults to true for backward compatibility)
- When set to false, no menu item is selected on initial render
- Improved keyboard navigation to handle unselected state:
  - Arrow down selects first item
  - Arrow up selects last item
  - Once an item is selected, maintains existing cycling behavior
- Added TypeScript types and prop interfaces for better type safety

Closes #7364

### Before

When opening any menu (e.g., component picker with "/"), the first item is always automatically selected

### After

When `autoSelectFirstItem={false}`:
- Menu opens with no item selected
- Arrow keys work as expected to select first/last items
- All other menu functionality remains unchanged

https://github.com/user-attachments/assets/6a212a76-c9cb-4733-a950-c9d6bdd10e00

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -199,6 +199,7 @@ export type TypeaheadMenuPluginProps<TOption extends MenuOption> = {
   anchorClassName?: string;
   commandPriority?: CommandListenerPriority;
   parent?: HTMLElement;
+  autoSelectFirstItem?: boolean;
 };
 
 export function LexicalTypeaheadMenuPlugin<TOption extends MenuOption>({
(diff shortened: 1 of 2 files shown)


takuyakanbr (commented):
Just a naming nit: I feel it's more common to call this kind of functionality "preselection" instead of "auto-selection". But otherwise LGTM

kirandash:
> Just a naming nit: I feel it's more common to call this kind of functionality "preselection" instead of "auto-selection". But otherwise LGTM

Thanks for the naming suggestion! I've updated the prop name from `autoSelectFirstItem` to `preselectFirstItem` to use more common terminology. The functionality remains the same.

takuyakanbr (approved):

Issue #7140, opened 2025-02-06
Bug: selection points to missing nodes after calling $deleteTableRow__EXPERIMENTAL() on all table rows


mike-atticus:
Calling `$deleteTableRow__EXPERIMENTAL()` when all table rows are selected produces an editor state (processed by update listeners) that has an invalid selection. The selection points to nodes that have been removed from the editor, so if you call `$getSelection()?.getNodes()` you will get an error `Point.getNode: node not found`.

**Note**: in the recording below the rendered HTML table doesn't show the selection, but the tree view below the editor shows the table cells are getting selected.

https://github.com/user-attachments/assets/30f574d1-1684-4672-a006-c814342307fc

Lexical version: 0.23.1

## Steps To Reproduce

1. Have an editor state with a `table` and at least one other sibling node before it (e.g. `paragraph`).
2. Have an update listener that calls `$getSelection?.getNodes()` on every update. 
3. Select all rows of the table, then take an action that runs `$deleteTableRow__EXPERIMENTAL()`.
4. Observe the error `Point.getNode: node not found` is thrown.

Link to code example: https://stackblitz.com/edit/facebook-lexical-akmwfitx?file=src%2FApp.tsx

## The current behavior

The error `Point.getNode: node not found` is thrown because the selection is pointing to nodes that have been removed from the document.

## The expected behavior

The selection should be fixed up (pointing to valid nodes) before update listeners are called with new updates. 

## Impact of fix

The most severe impact of this bug exists when you also use `CollaborationPlugin`, as this error causes Lexical and the Yjs doc to get out of sync. The table deletion doesn't propagate to the Yjs doc, so until the page is refreshed there is risk that user edits are corrupting the document in this unexpected state, and upon refresh users will see the table re-appear unexpectedly.

As a workaround, the error can be caught and handled to avoid this problem.

etrepum:
I haven't fixed the problem but here are the relevant changes to the theme so you can see the selection properly

https://stackblitz.com/edit/facebook-lexical-akmwfitx?file=src%2Fstyles.css,src%2FExampleTheme.ts,index.html,src%2Fmain.tsx

mike-atticus:
Thanks! Have updated the link in the original description.

PR #7408, opened 2025-03-30
[lexical-table] Feature: Improve logic for pasting table into table

### LINKED ISSUE #7262: Feature: Add missing rows and columns when pasting a table.
## Description
Automatically add missing rows and columns when pasting a table.

### Current Behavior

https://github.com/user-attachments/assets/a693d376-4efe-4607-86da-f92fed4745f4

### Expected Behavior

https://github.com/user-attachments/assets/ad8b3273-203d-4828-9d2a-f63f4b776940

## Impact
- Enhances user experience through more intuitive table operations.

### LINKED ISSUE #7263: Feature: When pasting a table, it does not work if there are merged cells in the target table.
## Description
When pasting a table, it does not work if there are merged cells in the target table.

### Current Behavior

https://github.com/user-attachments/assets/24dfeff4-5b77-42e1-8857-33971bfbff5f

### Expected Behavior

https://github.com/user-attachments/assets/f3df4552-ee8b-4cc3-8288-b08936b5c36a

## Impact
- Enhances user experience through more intuitive table operations.

takuyakanbr:
## Description

Improve how we handle pasting table into table:
- Properly account for merged cells in both tables. Presence of merged cells currently causes very buggy behavior when pasting (including the possibility of creating nested tables).
- If the current selection is a range selection, allow the table being modified to grow if necessary to contain the table being pasted. This is in line with GDocs behavior.
- If the current selection is a table selection, constrain the modifications to be within the selection boundary. This is in line with GDocs behavior.

Closes #7262
Closes #7263

## Test plan

Added E2E tests for table pasting: `npm run test-e2e-chromium Tables.spec`

### Before

https://github.com/user-attachments/assets/d2f03651-8149-4697-b1f5-b7a222d8e650

### After

https://github.com/user-attachments/assets/174211b4-c454-44e6-9e8b-c58b80fb39f5

## Additional notes

The new logic for pasting tables can be summarized as:
1. Unmerge all cells in affected area of the table being modified.
2. Expand the table being modified so that it can contain the table being pasted (if range selection).
3. Merge cells in the table being modified to match the table being pasted.
4. Set content in all affected cells to match the table being pasted.

To support the new logic, I've added a few additional table utils (in `lexical-table` package) that works on TableCellNode(s)—insert row, insert column, merge cells, unmerge cell. The logic for merging cells is taken from the `TableActionMenuPlugin` in `lexical-playground` package.

Some potential followups:
- I can't seem to be able to trigger pasting using cmd+V when there is a table selection (on Chrome). You can still use the browser context menu, but this should probably be fixed.
- Unlike GDocs, I'm not currently making multiple copies of the table being pasted when the selection boundary is larger than the size of the table being pasted. If there is interest in this, it could be a potential follow-up item.

### DIFF
--- a/packages/lexical-table/src/LexicalTableUtils.ts
+++ b/packages/lexical-table/src/LexicalTableUtils.ts
@@ -14,6 +14,7 @@ import {
   $createParagraphNode,
   $createTextNode,
   $getSelection,
+  $isParagraphNode,
   $isRangeSelection,
   LexicalNode,
 } from 'lexical';
@@ -262,21 +263,45 @@ export function $insertTableRow__EXPERIMENTAL(
   const focus = selection.focus.getNode();
   const [anchorCell] = $getNodeTriplet(anchor);
   const [focusCell, , grid] = $getNodeTriplet(focus);
-  const [gridMap, focusCellMap, anchorCellMap] = $computeTableMap(
+  const [, focusCellMap, anchorCellMap] = $computeTableMap(
     grid,
     focusCell,
     anchorCell,
   );
-  const columnCount = gridMap[0].length;
   const {startRow: anchorStartRow} = anchorCellMap;
   const {startRow: focusStartRow} = focusCellMap;
+  if (insertAfter) {
+    return $insertTableRowAtNode(
+      anchorStartRow + anchorCell.__rowSpan >
+        focusStartRow + focusCell.__rowSpan
+        ? anchorCell
+        : focusCell,
+      true,
+    );
+  } else {
+    return $insertTableRowAtNode(
+      focusStartRow < anchorStartRow ? focusCell : anchorCell,
+      false,
+    );
+  }
+}
+
+/**
+ * Inserts a table row before or after the given cell node,
+ * taking into account any spans. If successful, returns the
+ * inserted table row node.
+ */
+export function $insertTableRowAtNode(
+  cellNode: TableCellNode,
+  insertAfter = true,
+): TableRowNode | null {
+  const [, , grid] = $getNodeTriplet(cellNode);
+  const [gridMap, cellMap] = $computeTableMap(grid, cellNode, cellNode);
+  const columnCount = gridMap[0].length;
+  const {startRow: cellStartRow} = cellMap;
   let insertedRow: TableRowNode | null = null;
   if (insertAfter) {
-    const insertAfterEndRow =
-      Math.max(
-        focusStartRow + focusCell.__rowSpan,
-        anchorStartRow + anchorCell.__rowSpan,
-      ) - 1;
+    const insertAfterEndRow = cellStartRow + cellNode.__rowSpan - 1;
     const insertAfterEndRowMap = gridMap[insertAfterEndRow];
     const newRow = $createTableRowNode();
     for (let i = 0; i < columnCount; i++) {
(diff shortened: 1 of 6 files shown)


packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -725,14 +778,132 @@ function $insertFirst(parent: ElementNode, node: LexicalNode): void {
   }
 }
 
+export function $mergeCells(cellNodes: TableCellNode[]): TableCellNode | null {
```
takuyakanbr, line 781: (This function, along with the helper function following it, are extracted out from the `TableActionMenuPlugin` in the `lexical-playground` package, with no change in logic.)

packages/lexical-table/src/LexicalTableUtils.ts
```diff
@@ -262,21 +263,45 @@ export function $insertTableRow__EXPERIMENTAL(
...
+    );
+  }
+}
+
+/**
+ * Inserts a table row before or after the given cell node,
+ * taking into account any spans. If successful, returns the
+ * inserted table row node.
+ */
+export function $insertTableRowAtNode__EXPERIMENTAL(
```
takuyakanbr, line 294: (The "experimental" tag for insert row/column follows the existing functions that they are being extracted from. Open qns: Those function were added 2 years ago, so I'm wondering if we could promote them to non-experimental and get rid of the legacy versions.)

etrepum:
While you're in this part of the code there's another issue with one of these table functions that is probably a simple fix: https://github.com/facebook/lexical/issues/7140

etrepum:
Regarding the paste with cmd+V I think the problem is that the browser needs to have an active selection in order to handle paste. I think the real solution to this problem will be to create a selection in the browser that is promptly removed or otherwise specifically ignored by all of the code that tries to reconcile a change in the browser's selection to a new lexical RangeSelection (e.g. which happens while starting every update unless certain conditions are met, plus the selectionchange listener). This is probably also relevant to NodeSelection or any other custom selection. What may be needed is another hack like BlockCursorElement but for holding the browser's selection while a non-RangeSelection is in use.

packages/lexical-table/src/LexicalTableUtils.ts
etrepum: In the case for new functions I don't think they should have these experimental names, if a function is experimental it would be better to just document it appropriately (e.g. with `@internal` if we don't want it to show up in docs). I think this naming convention was a failed experiment and not a convention worth following for new code.

etrepum (commented):
I didn't do a careful audit of the code or tests but I did some manual testing and it seems to work much better now. Will approve once the functions are renamed without the EXPERIMENTAL postfix.

packages/lexical-table/src/LexicalTableUtils.ts
takuyakanbr: I've renamed the new functions. For the existing "experimental" insert row/column functions, there are legacy exported functions using the same name (without the `__EXPERIMENTAL` tag) but with different APIs 😅

etrepum: hah! well I guess we need to come up with new names then, would be hard to do a proper deprecation if the call signature changes. Doesn't have to be in the scope of this PR

etrepum (approved):

PR #7409, opened 2025-03-30
FloatingUI for LexicalMenu positioning

ivailop7:
Since the Core team agreed that it's OK for the @lexical/react package to depend on [floating UI](https://floating-ui.com/) to fix the difficult problem of popovers positioning. This PR begins this transition. This PR does this for the typeahead-based menus.

This PR also introduces a breaking change, because it removes the custom `menuRenderFn` prop in LexicalMenu so one does not need to pass it in each typeahead widget (Mention, Forward slash, Emoji Picker, AutoEmbed), but reuses the same style and configuration for all them. This ensures consistency and reduces the overall codebase size and removes a bit of the duplicated code with respect to the playground code. 

There are other places where FloatingUI can be used like 'floating format toolbar', 'add comment popover', 'context menu', 'drag handle positioning', 'table hover action buttons', 'dropdown (toolbar and table action items)', 'the new + icon menu' those can come later on.


ivailop7:
Currently has a bug with the Paste event from the custom context menu, where it appens to the start of the editor, vs where the selection was done. I guess it's a focus-related thing. Yet to pin down.

ivailop7:
I have found a much cleaner approach to do this. I'll do the ContextMenu by itself not using the LexicalMenu component at all. So this would allow to clean up the LexicalMenu and get rid of the pointer vs caret coordinates derivation code in the useMenuAnchor hook and make easier to reason about menu positioning.

ivailop7:
Closed in favour of https://github.com/facebook/lexical/pull/7623

PR #7412, opened 2025-04-01
[lexical] Bug Fix: Fix forward line deletion when using control+K

takuyakanbr:
Fix forward line deletion when using control+K on Mac.

On the Mac, control+K [performs forward line deletion](https://support.apple.com/en-sg/102650#:~:text=Control%2DK%3A%20delete%20the%20text%20between%20the%20insertion%20point%20and%20the%20end%20of%20the%20line%20or%20paragraph.), which is the same function that Lexical already supports / handles via cmd+fn+delete. We're not currently explicitly handling the control+K key combination in Lexical, resulting in unexpected behavior when it is performed.

(Fixes an internally-reported bug: T193999537.)

## Test plan

Added a test to `Selection.mjs.spec`:
`npm run test-e2e-chromium Selection`

### Before

https://github.com/user-attachments/assets/72fad707-5bcc-4241-b59d-b5bbe5b29a23

### After

https://github.com/user-attachments/assets/825df656-1906-4fca-866a-1a0d1c1f494a

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -911,8 +911,15 @@ export function isDeleteLineBackward(key: string, metaKey: boolean): boolean {
   return IS_APPLE && metaKey && isBackspace(key);
 }
 
-export function isDeleteLineForward(key: string, metaKey: boolean): boolean {
-  return IS_APPLE && metaKey && isDelete(key);
+export function isDeleteLineForward(
+  key: string,
+  metaKey: boolean,
+  ctrlKey: boolean,
+): boolean {
+  return (
+    IS_APPLE &&
+    ((metaKey && isDelete(key)) || (ctrlKey && key.toLowerCase() === 'k'))
+  );
 }
 
 export function isDeleteBackward(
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalUtils.ts
```diff
@@ -911,8 +911,15 @@ export function isDeleteLineBackward(key: string, metaKey: boolean): boolean {
...
-export function isDeleteLineForward(key: string, metaKey: boolean): boolean {
-  return IS_APPLE && metaKey && isDelete(key);
+export function isDeleteLineForward(
+  key: string,
+  metaKey: boolean,
+  ctrlKey: boolean,
+): boolean {
+  return (
+    IS_APPLE &&
+    ((metaKey && isDelete(key)) || (ctrlKey && key.toLowerCase() === 'k'))
```
etrepum, line 921: This is just general feedback but these keybindings all match too broadly and we should probably refactor all of the keypress handling in LexicalEvents and LexicalUtils, e.g. cmd-ctrl-k or shift-ctrl-k works the same as ctrl-k because we don't specifically check for the absence of other modifiers.

etrepum (approved):
Current test failures look like the result of a transient npm outage but I would recommend re-running the checks and confirming that they do all pass before merge

etrepum:
Just restarted the failed checks

PR #7420, opened 2025-04-03
[lexical-list] Feature: Enforce strict list indentation

dineug:
## Description

We are adding a strict indent feature for lists.
This feature helps in creating an accurate Markdown converter.
Additionally, it can be optionally enabled as needed.

### Before

https://github.com/user-attachments/assets/75fd4f1f-7b1c-49e3-96f2-fc7748b0c567

### After

https://github.com/user-attachments/assets/e1235286-b37a-4ca8-bd35-ee08439a576d

### DIFF
--- a/packages/lexical-list/src/index.ts
+++ b/packages/lexical-list/src/index.ts
@@ -14,7 +14,7 @@ import type {
 } from './LexicalListNode';
 import type {LexicalCommand, LexicalEditor} from 'lexical';
 
-import {mergeRegister} from '@lexical/utils';
+import {$findMatchingParent, mergeRegister} from '@lexical/utils';
 import {
   $getSelection,
   $isRangeSelection,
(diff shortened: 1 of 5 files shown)


etrepum (approved):
This looks reasonable to me, although I think the transform algorithm could probably be optimized if a bit more state was maintained particularly in the recursive case

dineug:
I have incorporated the code review! @etrepum

packages/lexical-list/src/index.ts
```diff
@@ -145,6 +145,89 @@ export function registerList(editor: LexicalEditor): () => void {
...
+          $formatListIndentStrict(child);
+
+          const firstChild = child.getFirstChild();
+          if ($isListNode(firstChild)) {
+            queue.push(firstChild);
+          }
+        }
+      }
+    }
+  };
```
etrepum, line 206: This isn't really what I meant by preserving information in the recursive case. For example, `startingListItemNode` can be computed here with much less redundant work in each `$formatListIndentStrict`. I also don't think it needs to go very deep because I think modifying a list's indentation can only break the indent invariant for itself or its list children so the recursion can stop if no modifications are made (or really even if they are, since the transform is going to run again anyway, but preemptive modifications would do fewer rounds of transforms).

etrepum (approved):
As before I think this looks fine for now, it's up to you if you'd like to optimize the algorithm

dineug:
Thanks for the detailed explanation! @etrepum 
I think it should be okay to merge now.

PR #7429, opened 2025-04-04
[lexical-list] Feature: export registerCheckList

### LINKED ISSUE #7428: Feature: make autolink and checklist plugins framework agnostic
## Description
Most plugins are implemented as framework agnostic lexical packages with a thin react specific wrapper in lexical-react. That is not the case with autolink and checklist, why is that? Are there any specific reason for this, like moving away from being framework agnostic, or the plugins are work in progress not ready for general use?

I don't use react, but found it really easy to integrate and extend lexical. However with autolink and checklist I had to copy the source code from the react package and make some small changes (removing react imports and exporting as a register*** method).

I would propose that these implementations are moved out of the react package and into core plugins, with thin react wrappers left (as for most other core plugins).

This would benefit users in all other frameworks...

## Impact
Autolink and Checklist usable from frameworks other than react.

atsjo:
## Description

Move implementation of CheckList from @lexical/react to @lexical/list, and export registerCheckList function. The original plugin in @lexical/react now calls registerCheckList. Also changed from using insertList to $insertList, since the first is deprecated

Partially Closes #7428 

## Test plan
no change in logic, only moved current implementation to make it available without react


packages/lexical-react/src/LexicalCheckListPlugin.tsx
```diff
@@ -6,301 +6,16 @@
...
-      : sibling.getFirstChild();
-
-    if (!$isListNode(firstChild)) {
-      return sibling;
-    }
-
-    sibling = backward ? firstChild.getLastChild() : firstChild.getFirstChild();
-  }
-
+    return mergeRegister(registerCheckList(editor));
```
etrepum, line 18: mergeRegister is redundant if there's just one
```suggestion
    return registerCheckList(editor);
```

packages/lexical-list/src/checkList.ts
```diff
@@ -0,0 +1,300 @@
...
+      return sibling;
+    }
+
+    sibling = backward ? firstChild.getLastChild() : firstChild.getFirstChild();
+  }
+
+  return null;
+}
+
+function handleArrownUpOrDown(
```
etrepum, line 268:
```suggestion
function handleArrowUpOrDown(
```

I get that this is just copied and pasted but there's a typo in this name

packages/lexical-list/src/checkList.ts
```diff
@@ -0,0 +1,300 @@
...
+      () => {
+        $insertList('check');
+        return true;
+      },
+      COMMAND_PRIORITY_LOW,
+    ),
+    editor.registerCommand<KeyboardEvent>(
+      KEY_ARROW_DOWN_COMMAND,
+      (event) => {
+        return handleArrownUpOrDown(event, editor, false);
```
etrepum, line 54:
```suggestion
        return handleArrowUpOrDown(event, editor, false);
```

packages/lexical-list/src/checkList.ts
```diff
@@ -0,0 +1,300 @@
...
+      },
+      COMMAND_PRIORITY_LOW,
+    ),
+    editor.registerCommand<KeyboardEvent>(
+      KEY_ARROW_UP_COMMAND,
+      (event) => {
+        return handleArrownUpOrDown(event, editor, true);
```
etrepum, line 61:
```suggestion
        return handleArrowUpOrDown(event, editor, true);
```

etrepum (dismissed):
Looks good! I think since this already cleans up some deprecations it would make sense to also fix a few internal typos

kirandash:
> Looks good! I think since this already cleans up some deprecations it would make sense to also fix a few internal typos

I've created a PR at https://github.com/facebook/lexical/pull/7437 that fixes this and other typos across the codebase, including the `handleArrownUpOrDown` typo you mentioned.

PR #7431, opened 2025-04-05
[lexical-playground] Bug Fix: Immediate broken image display on load failure

### LINKED ISSUE #7387: Bug: image-broken.svg fails to load in Lexical editor when image loading fails
Lexical version:0.29.0

## Steps To Reproduce

1. Try to load an invalid or inaccessible image URL in the Lexical editor
2. When the image fails to load, the fallback image-broken.svg should display
3. However, the image-broken.svg itself fails to load because it's trying to load from the wrong path
4. It displays after making some changes to the editor ex: typing something

## The current behavior
- When the image fails to load, the fallback image-broken.svg does not display

https://github.com/user-attachments/assets/60acfca2-9647-4f3a-90cc-90475654a4bc

## The expected behavior
- When an image fails to load, the fallback image-broken.svg should display correctly and instantly

## Impact of fix
- Happens every time image url fails to load

kirandash:
### Current behavior:
- When an image fails to load in the Lexical editor, the fallback image-broken.svg does not display immediately
- The fallback image only appears after making changes to the editor (like typing)
- The image-broken.svg fails to load because it's trying to load from the wrong path

### Changes in this PR:
- Improved error handling in `useSuspenseImage` hook to properly handle image loading failures
- Modified LazyImage component to maintain proper error state and show fallback immediately
- Ensured failed images aren't cached to allow retry attempts

Closes #7387

### Before
- Try to load an invalid image URL in the editor
- Observe that the broken image fallback doesn't appear immediately
- Need to make changes to the editor to see the fallback

https://github.com/user-attachments/assets/74df4cdf-acb9-4aad-baf9-9b370035ecdc

### After

- Load an invalid image URL in the editor
- Broken image fallback appears immediately
- No need to make additional changes to see the fallback
- Verify that retrying with a valid URL works

https://github.com/user-attachments/assets/c19a74e1-8526-4b16-968e-beb33c5716a7

### DIFF
--- a/packages/lexical-playground/src/nodes/ImageComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ImageComponent.tsx
@@ -73,7 +73,8 @@ function useSuspenseImage(src: string) {
         resolve(null);
       };
       img.onerror = () => {
-        imageCache.add(src);
+        // Don't cache failed images to allow retry
+        resolve(null);
       };
     });
   }
@@ -102,11 +103,11 @@ function LazyImage({
   width: 'inherit' | number;
   onError: () => void;
 }): JSX.Element {
-  useSuspenseImage(src);
   const [dimensions, setDimensions] = useState<{
     width: number;
     height: number;
   } | null>(null);
+  const [hasError, setHasError] = useState(false);
   const isSVGImage = isSVG(src);
 
   // Set initial dimensions for SVG images
@@ -119,6 +120,16 @@ function LazyImage({
       });
     }
   }, [imageRef, isSVGImage]);
+
+  try {
+    useSuspenseImage(src);
+  } catch (error) {
+    // Let the component handle the error through onError
+  }
+
+  if (hasError) {
+    return <BrokenImage />;
+  }
 
   // Calculate final dimensions with proper scaling
   const calculateDimensions = () => {
@@ -168,7 +179,10 @@ function LazyImage({
       alt={altText}
       ref={imageRef}
       style={imageStyle}
-      onError={onError}
+      onError={() => {
+        setHasError(true);
+        onError();
+      }}
       draggable="false"
       onLoad={(e) => {
         if (isSVGImage) {
@@ -193,6 +207,7 @@ function BrokenImage(): JSX.Element {
         width: 200,
       }}
       draggable="false"
+      alt="Broken image"
     />
   );
 }
--- a/packages/lexical-playground/src/nodes/ImageComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ImageComponent.tsx
@@ -124,8 +124,14 @@ function LazyImage({
   }, [imageRef, isSVGImage]);
 
   const hasError = useSuspenseImage(src);
+
+  useEffect(() => {
+    if (hasError) {
+      onError();
+    }
+  }, [hasError, onError]);
+
   if (hasError) {
-    onError();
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -73,7 +73,8 @@ function useSuspenseImage(src: string) {
         resolve(null);
       };
       img.onerror = () => {
-        imageCache.add(src);
+        // Don't cache failed images to allow retry
+        resolve(null);
```
takuyakanbr, line 77: I feel like just adding this line alone would suffice, without all the other changes associated with error handling / caching. Even if the URL is added to the cache, a refetch can still be triggered when the `` element (L176) gets rendered.

packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -120,6 +121,16 @@ function LazyImage({
     }
   }, [imageRef, isSVGImage]);
 
+  try {
+    useSuspenseImage(src);
+  } catch (error) {
+    // Let the component handle the error through onError
+  }
```
takuyakanbr, line 128: Wrapping `useSuspenseImage` with try-catch basically negates the purpose of throwing a Promise in the function and the use of Suspense in the parent. It negates the value of having that function itself, actually, since the `` element will be immediately rendered and that by itself would cause the image to be loaded and error handling to be triggered.

etrepum: This is correct, but also `useSuspenseImage` isn't really a good example of how to implement suspense because it doesn't return a stable promise ("in practice" I think that won't matter with how this is used, but it might be better to demonstrate best practices). Probably a "good" implementation would be more like this (untested):

```ts
const imageCache = new Map<string, Promise<bool> | bool>();
// returns true if there was an error, false otherwise
function useSuspenseImage(src: string): boolean {
  let cached = imageCache.get(src);
  if (typeof cached === 'boolean') {
    return cached;
  } else if (!cached) {
    cached = new Promise<boolean>((resolve) => {
      const img = new Image();
      img.src = src;
      img.onload = () => resolve(false);
      img.onerror = () => resolve(true);
    }).then((hasError) => imageCache.set(src, hasError));
    imageCache.set(src, cached);
  }
  throw cached;
}
```

Could be simplified a bit further in React 19 with `use` since you wouldn't need to track the promise's state directly (could use a `Map<string, Promise<bool>>`).

kirandash: Thanks for the detailed feedback @takuyakanbr and @etrepum! I understand the concerns about proper Suspense implementation. I've updated the code to remove the try-catch block and follow the suggested pattern.

Let me know if you'd like me to adjust this approach further.

packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -120,6 +123,12 @@ function LazyImage({
     }
   }, [imageRef, isSVGImage]);
 
+  const hasError = useSuspenseImage(src);
+  if (hasError) {
+    onError();
```
takuyakanbr, line 128: This is going to modify component state during render, which React doesn't like (you should see a console warning). We should move this into a `useEffect`.

etrepum (approved):
LGTM!

PR #7438, opened 2025-04-07
[lexical] Feature: Add text-transform styles to exported HTML

### LINKED ISSUE #7328: Feature: lowercase, uppercase an capitalize doesn't export as style
Currently all those options only add classes chosen with theme, meaning that whenever I use the exported html in another place that doesn't have the class, the text will not be transformed.

Lexical version: 0.27.0

## Steps To Reproduce

1. Add text, set it to either one of capitalize, uppercase or lowercase. 
2. In "export dom" you can see the span class changes to the one in your theme, but there is no style.

## The current behavior
Applying capitalization, uppercase and lowercase only adds new class to the text span, however, it doesn't add any style to the span, meaning it will not work when class is not present.

## The expected behavior
Applying those options should edit the style with appropriate text-transform to make the html portable without the need of custom classes.

## Impact of fix
Everyone who exports to HTML and wants to use those options in their toolbar will benefit from it.

kirandash:
### Current behavior:
When using text formatting options like lowercase, uppercase, and capitalize, Lexical only adds CSS classes to the exported HTML without the necessary inline styles. This means the formatting doesn't work when the content is viewed outside of Lexical's CSS context.

### Changes in this PR:
- This PR adds inline `text-transform` styles to the HTML elements during export via `TextNode.exportDOM()`. Now when users export content with these text transformations, the formatted text will maintain its appearance even when viewed without the Lexical CSS classes.
- Unit tests have been added to verify that the exported HTML contains the correct text-transform styles for all capitalization formats (lowercase, uppercase, and capitalize).

Closes #7328

### Before
When applying uppercase, lowercase, or capitalize to text and exporting the HTML, the elements only had CSS classes but no inline styles. The exported HTML looked like:
```html
<span class="PlaygroundEditorTheme__textUppercase">TEXT</span>
```
This formatting was lost when viewed outside of Lexical's styling context.

### After
Now when exporting text with capitalization formats, the exported HTML includes the appropriate inline style:
```html
<span class="PlaygroundEditorTheme__textUppercase" style="text-transform: uppercase;">TEXT</span>
```
This ensures the text transformation is preserved even without Lexical's CSS classes.

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -645,6 +645,16 @@ export class TextNode extends LexicalNode {
       'Expected TextNode createDOM to always return a HTMLElement',
     );
     element.style.whiteSpace = 'pre-wrap';
+
+    // Add text-transform styles for capitalization formats
+    if (this.hasFormat('lowercase')) {
+      element.style.textTransform = 'lowercase';
+    } else if (this.hasFormat('uppercase')) {
+      element.style.textTransform = 'uppercase';
+    } else if (this.hasFormat('capitalize')) {
+      element.style.textTransform = 'capitalize';
+    }
+
     // This is the only way to properly add support for most clients,
     // even if it's semantically incorrect to have to resort to using
     // <b>, <u>, <s>, <i> elements.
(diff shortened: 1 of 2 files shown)


ivailop7:
Can we get @etrepum and @zurfyx to comment on this change as well please.

etrepum (approved):
This looks right, what it may be missing is importDOM support for the other direction

PR #7439, opened 2025-04-07
[lexical-selection] Bug Fix: Correct caret movement in vertical-rl writing mode

### LINKED ISSUE #7363: Bug: Incorrect Caret Movement in Vertical Text Mode (writing-mode: vertical-rl)
Lexical version:
v0.28.0+dev.esm (using Playground)

## Steps To Reproduce

1. Enter [playground](https://playground.lexical.dev/)
2. Open the browser DevTools and add `writing-mode: vertical-rl;` to `.ContentEditable__root`.
3. In the body of the text, press EnterKey several times to create a few blank lines. (`<p><br><p>`)
4. Place a caret on the blank lines and press the left or right arrow key.
5. The caret will move in the opposite direction.

## The current behavior

Press the right arrow key to move left and the left arrow key to move right.
This incorrect behavior only occurs in blank lines (`<p><br><p>`). When the paragraph contains text, the arrow keys function correctly.

It does not occur in a standard `contenteditable` element without Lexical.

## The expected behavior

When the right arrow key is pressed, the caret should move to the right.
When the left arrow key is pressed, the caret should move to the left.

## Impact of fix

Since `writing-mode: vertical-rl;` is used in traditional Japanese typography, this fix will significantly improve the experience for Japanese users in particular.

kirandash:
### Current behavior:
In vertical-rl writing mode, when navigating between empty paragraphs, arrow keys move in the opposite direction from expected and also blocks the navigation to previous nodes. Right arrow moves left and left arrow moves right and is blocked. This only happens in empty paragraphs (`<p><br></p>`).

### Changes in this PR:
This PR fixes the incorrect caret movement in vertical-rl writing mode by:
1. Modifying `$moveCharacter` to correctly handle vertical writing mode orientation
2. Adding support for detecting vertical writing mode in the editor
3. Ensuring correct behavior for empty paragraphs by inverting arrow key directions
4. Passing the editor instance to arrow key handlers in both rich-text and plain-text plugins

Closes #7363

### Before

Arrow keys behave incorrectly in vertical-rl writing mode:
- Right arrow key moves left (instead of right), keep pressing left but arrow does not navigate to previous paragraphs
- Left arrow key moves right (instead of left), keep pressing right but arrow does not navigate to previous paragraphs
This incorrect behavior only occurs in empty paragraphs.

https://github.com/user-attachments/assets/31a009a4-1968-453c-a72c-ce6f0a336fa1

### After

Arrow keys now behave correctly in vertical-rl writing mode:
- Right arrow key moves right
- Left arrow key moves left

https://github.com/user-attachments/assets/9644d0f8-3aab-4bee-b9d5-8b9437c71f0d

### Testing Note
I attempted to add E2E tests to verify this fix, but encountered challenges in getting them to pass. This is likely due to the complexity of simulating and asserting cursor movement behavior in vertical writing mode.

Instead, I've included a video demonstration showing the working fix in the UI.

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -9,7 +9,6 @@
 import type {
   BaseSelection,
   ElementNode,
-  LexicalEditor,
   LexicalNode,
   NodeKey,
   Point,
@@ -22,6 +21,7 @@ import {
   $caretFromPoint,
   $createRangeSelection,
   $extendCaretToRange,
+  $getEditor,
   $getPreviousSelection,
   $getSelection,
   $hasAncestor,
@@ -418,18 +418,37 @@ export function $wrapNodesImpl(
 }
 
 /**
+ * Tests if the selection's parent element has vertical writing mode.
+ * @param selection - The selection whose parent to test.
+ * @returns true if the selection's parent has vertical writing mode (writing-mode: vertical-rl), false otherwise.
+ */
+export function $isEditorVerticalOrientation(
+  selection: RangeSelection,
+): boolean {
+  const anchorNode = selection.anchor.getNode();
+  const parent = $isRootNode(anchorNode)
+    ? anchorNode
+    : anchorNode.getParentOrThrow();
+  const editor = $getEditor();
+  const domElement = editor.getElementByKey(parent.getKey());
+  if (domElement === null) {
+    return false;
+  }
+  const computedStyle = window.getComputedStyle(domElement);
+  return computedStyle.writingMode === 'vertical-rl';
+}
+
+/**
  * Determines if the default character selection should be overridden. Used with DecoratorNodes
  * @param selection - The selection whose default character selection may need to be overridden.
  * @param isBackward - Is the selection backwards (the focus comes before the anchor)?
- * @param editor - The editor instance to check for vertical writing mode.
  * @returns true if it should be overridden, false if not.
  */
 export function $shouldOverrideDefaultCharacterSelection(
   selection: RangeSelection,
   isBackward: boolean,
-  editor?: LexicalEditor,
 ): boolean {
-  const isVertical = editor ? $isEditorVerticalOrientation(editor) : false;
+  const isVertical = $isEditorVerticalOrientation(selection);
 
   // In vertical writing mode, we adjust the direction for correct caret movement
   const adjustedIsBackward = isVertical ? !isBackward : isBackward;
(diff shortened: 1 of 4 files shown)


etrepum (commented):
Let's refactor it to not pass the editor explicitly, you can use `$getEditor()` in the one place it's needed.

Similarly I think that the vertical writing mode should be based on where the selection is (similar to how $isParentElementRTL works).

Probably this vertical writing mode should be part of the lexical model rather than only in DOM but it would be a reasonable fix to do it this way in the meantime

kirandash:
> Let's refactor it to not pass the editor explicitly, you can use `$getEditor()` in the one place it's needed.
> 
> Similarly I think that the vertical writing mode should be based on where the selection is (similar to how $isParentElementRTL works).
> 
> Probably this vertical writing mode should be part of the lexical model rather than only in DOM but it would be a reasonable fix to do it this way in the meantime

Thanks for the review! I've pushed a commit [f9dbfbd](https://github.com/facebook/lexical/pull/7439/commits/f9dbfbd156b7e1d8fe2a05060826ae4cd8416ed7) that implements your suggestions:

1. Refactored to remove the explicit `editor` parameter and use `$getEditor()` internally
2. Updated the vertical writing mode check to be selection-based, similar to `$isParentElementRTL`

Please let me know if you'd like me to make any other changes to this PR!

packages/lexical-selection/src/range-selection.ts
```diff
@@ -416,6 +417,27 @@ export function $wrapNodesImpl(
...
+  const anchorNode = selection.anchor.getNode();
+  const parent = $isRootNode(anchorNode)
+    ? anchorNode
+    : anchorNode.getParentOrThrow();
+  const editor = $getEditor();
+  const domElement = editor.getElementByKey(parent.getKey());
+  if (domElement === null) {
+    return false;
+  }
+  const computedStyle = window.getComputedStyle(domElement);
```
etrepum, line 437: When iframes are in use, the editor's window may not match the global window

```suggestion
  const computedStyle = getWindow(editor).getComputedStyle(domElement);
```

Could also use `domElement.ownerDocument.defaultView` instead of `getWindow(editor)`, should be the same thing at least under the current constraints of how lexical DOM is allowed to work

etrepum (commented):
Haven't done any manual testing yet but there's one small issue with the use of window

packages/lexical-selection/src/range-selection.ts
kirandash: Thanks for the suggestion! I've pushed a [commit](https://github.com/facebook/lexical/pull/7439/commits/147dcf3fe3377b75e0b9f9b151398a3daf63da5f) using `domElement.ownerDocument.defaultView`.

etrepum (approved):
I suspect that this will be revisited but it does seem like an improvement. I don't have any experience with how the vertical-rl writing mode is supposed to work

PR #7441, opened 2025-04-07
[lexical][lexical-website] Feature: Document and export common update tags

### LINKED ISSUE #5992: Feature: document and export common/known update tags
If we export the update tags that are used internally it'll be easier to document what they're for and people will be less likely to make a typo which is hard to debug since it's used as an extensible `Set<string>` and there is no way a dev mode would be able to detect 'unregistered' use since there is no registration process like commands.

Some examples in the monorepo that I'm aware of:

* historic
* history-push
* history-merge
* paste
* collaboration
* skip-collab
* skip-scroll-into-view

It might also be a good time to think about how and whether updates with different tags should really be merged, since you could have two updates in the same tick with very different semantics. It might be the case that for correct behavior they should be reconciled separately at least if they were not part of the same top-level update. It seems reasonable that cascading transforms and any other sort of explicit nested update should inherit tags but if the updates were completely separate it might not make sense. I think this sort of bug would be very hard to catch and reproduce, and tags are probably used rarely enough that it wouldn't be a performance impact to reconcile more than once in a tick (iff updates with different tag sets are to be combined) to avoid this kind of incoherence.

kirandash:
### Current behavior:
- Update tags are used internally but not exported or documented
- No validation for typos in tag names
- No comprehensive documentation about tag behavior and lifecycle
- No test coverage for update tag functionality

### Changes being added:
- Create `LexicalUpdateTags.ts` to export common update tags as constants
- Add tag validation to `$addUpdateTag` to catch typos
- Add comprehensive documentation in `updates.md` about:
  - Tag usage and examples
  - Tag lifecycle (clearing after updates)
  - Common update tags and their purposes
  - Tag validation
- Add test suite covering all update tag functionality

Closes #5992

### Before
- No validation for incorrect tag names
- No clear documentation about tag behavior
- Tags were internal implementation details
- No dedicated test file for update tags

### After

Added comprehensive test suite in `LexicalUpdateTags.test.ts` that verifies:
- All exported tags are in `KNOWN_UPDATE_TAGS`
- `$addUpdateTag` and `$hasUpdateTag` functionality
- Multiple tag support
- Tag validation warnings
- Tag lifecycle (clearing after updates)

All tests are passing and validate the documented behavior.


packages/lexical-website/docs/concepts/updates.md
```diff
@@ -0,0 +1,86 @@
+# Updates
+
+Updates in Lexical are the way you can mutate the editor state. All updates are run asynchronously and batched by default. This means that if you have multiple updates back-to-back, they'll be batched together and run in a single update. This is done for performance reasons, as it means we can avoid unnecessary re-renders and DOM updates.
```
etrepum, line 3: This isn't accurate, updates are synchronous (except in nested update scenarios which should be deprecated) but the reconciliation (DOM updates) is batched for performance

packages/lexical-website/docs/concepts/updates.md
```diff
@@ -0,0 +1,86 @@
...
+
+You can check if a tag is present using `$hasUpdateTag()`:
+```js
+editor.update(() => {
+  $addUpdateTag('my-tag');
+  console.log($hasUpdateTag('my-tag')); // true
+});
+```
+
+Note: Update tags are cleared after each update. If you need to check for tags, make sure to do it within the same update callback where they were added.
```
etrepum, line 42: Typically update tags are checked in update and mutation listeners with the `tags` and `updateTags` properties in their respective payloads (as demonstrated below), it's less common to read update tags from inside the same update batch

packages/lexical-website/docs/concepts/updates.md
```diff
@@ -0,0 +1,86 @@
...
+- `skip-dom-selection`: Prevents updating the DOM selection (useful for updates that shouldn't affect focus)
+
+### Tag Validation
+
+To help catch typos and ensure you're using known tags, you can enable tag validation by passing `true` as the second argument to `$addUpdateTag`:
+
+```js
+editor.update(() => {
+  // Will warn if "unknown-tag" is not a known tag
+  $addUpdateTag('unknown-tag', true);
```
etrepum, line 64: I think a better approach here would be to recommend that people use the exported constants rather than string literals to avoid typos (e.g. `HISTORIC_TAG` instead of `'historic'`).

packages/lexical/src/LexicalUpdateTags.ts
```diff
@@ -0,0 +1,67 @@
...
+export const KNOWN_UPDATE_TAGS = new Set([
+  HISTORIC_TAG,
+  HISTORY_PUSH_TAG,
+  HISTORY_MERGE_TAG,
+  PASTE_TAG,
+  COLLABORATION_TAG,
+  SKIP_COLLAB_TAG,
+  SKIP_SCROLL_INTO_VIEW_TAG,
+  SKIP_DOM_SELECTION_TAG,
+]);
```
etrepum, line 67: I don't think we need this

packages/lexical/src/index.ts
```diff
@@ -296,3 +296,16 @@ export type {
...
+export {
+  COLLABORATION_TAG,
+  HISTORIC_TAG,
+  HISTORY_MERGE_TAG,
+  HISTORY_PUSH_TAG,
+  KNOWN_UPDATE_TAGS,
+  PASTE_TAG,
+  SKIP_COLLAB_TAG,
+  SKIP_DOM_SELECTION_TAG,
+  SKIP_SCROLL_INTO_VIEW_TAG,
```
etrepum, line 310: This is the useful part for preventing typos

kirandash:
Thanks for the thorough review @etrepum! I apologize for the confusion in the documentation. I had misunderstood how updates worked, incorrectly assuming they were asynchronous. Thanks for the clarification! 

I've addressed all your feedback:
1. [Fixed the documentation about updates being synchronous, with only DOM reconciliation being batched](https://github.com/facebook/lexical/issues/5992)
2. [Improved documentation for update tag usage patterns](https://github.com/facebook/lexical/pull/7441/commits/4c63ee88ca4b3af188b332ef281ed5d251d6043f)
3. [Use exported constants for update tag examples](https://github.com/facebook/lexical/pull/7441/commits/fb34ab4c4e154a21af36acc4687f43f3432b0a4f)
4. [Removed the KNOWN_UPDATE_TAGS Set and runtime validation as suggested](https://github.com/facebook/lexical/pull/7441/commits/5a722ea58a8ad90ff51348785e64bbdb16cdea08)

Please take another look when you have a chance. Let me know if anything needs further adjustment!

packages/lexical-website/docs/concepts/updates.md
```diff
@@ -0,0 +1,132 @@
+# Updates
```
etrepum, line 1: This new file needs to be referenced in packages/lexical-website/sidebars.js so you can find it in the documentation. I think this is also why the preview isn't rendered with a sidebar

new page:
https://lexical-git-fork-bgwebagency-feat-5992-docu-7750e0-fbopensource.vercel.app/docs/concepts/updates

existing page in that section:
https://lexical-git-fork-bgwebagency-feat-5992-docu-7750e0-fbopensource.vercel.app/docs/concepts/nodes

etrepum (commented):
I think this looks great other than the missing sidebar reference. Perhaps we should create a follow-up issue to have the existing code use these new constants to match the recommendation from the new docs?

kirandash:
> I think this looks great other than the missing sidebar reference. Perhaps we should create a follow-up issue to have the existing code use these new constants to match the recommendation from the new docs?

Thanks for the review! I've fixed the missing sidebar reference and created a follow-up issue to migrate the existing code to use the new constants at #7446. 

Let me know if there's anything else you'd like me to change or improve in this PR!

etrepum (approved):
Looks great!

PR #7442, opened 2025-04-07
[lexical][lexical-website] Feature: Improve createCommand developer experience

### LINKED ISSUE #6021: Feature: Dev experience for createCommand
It would be great if we could come up with some ways to enable better introspection and debuggability in prod builds for `createCommand`. The transform that `createCommand` does right now is irreversible and the choice of prod vs. dev build infects user code because the type information is thrown away in the constructor.

```typescript
export function createCommand<T>(type?: string): LexicalCommand<T> {
  return __DEV__ ? {type} : {};
}
```

I suggest we change the implementation to an unconditional `return {type};`

```typescript
export function createCommand<T>(type?: string): LexicalCommand<T> {
  return {type};
}
```

This leaves the decision of whether to provide a type or not to the caller. Might increase the bundle size by probably about 1kb if we do nothing else?

### Compression options

If we want to sacrifice DX in the same way that we do now, but in a way that only affects library code, we can have a transform like `transform-error-messages` simply remove the argument in prod builds.

If a particularly optimization-focused caller really wants to conditionally throw away the type, they can already do that pretty easily with their own wrapper, a ternary at each call site, or a transform (as above).

Another compression option here would be to have something like transform-error-messages that replaces longer strings with shorter ones, e.g. ':1' and we would expect that any command that has a type starting with ':' should be found in an out-of-bundle-table (that the dev tools would ship with). If we were byte-conscious we could use a more dense encoding than decimals.

A slight modification to the above would be to change the type to `type?: string | number` which would let us have an out-of-bundle table and serial id transform, to save a few uncompressed bytes for each command (since numbers don't have 2 characters of syntax overhead).

RE #2942
/cc @StyleT @acywatson @thegreatercurve

kirandash:
### Current behavior:
Currently, createCommand has two issues:
1. It uses a `__DEV__` conditional that makes the transform irreversible:
```ts
export function createCommand<T>(type?: string): LexicalCommand<T> {
  return __DEV__ ? {type} : {};
}
```
2. The type information is thrown away in production builds, making debugging difficult and affecting user code.

### Changes being added:
This PR improves the developer experience for createCommand by:
1. Removing the `__DEV__` conditional to make transforms reversible
2. Using a transform table approach that:
   - Maps built-in commands to short numeric IDs in production
   - Preserves descriptive string types for custom commands
   - Maintains debuggability while optimizing bundle size
3. Adding CommandTypes enum for better type safety
4. Adding comprehensive test coverage
5. Updating documentation to clarify built-in vs custom command usage

The bundle size impact (approximately 1kb) is minimized through the transform table approach, while providing much better developer experience.

Closes #6021

### Before

```ts
// In development
createCommand('CUSTOM_COMMAND') // returns {type: 'CUSTOM_COMMAND'}
// In production
createCommand('CUSTOM_COMMAND') // returns {} - all type info lost
```

### After

```ts
// Built-in commands - optimized in production
createCommand('SELECTION_CHANGE_COMMAND') // returns {type: 1}
// Custom commands - preserved in both dev and prod
createCommand('CUSTOM_COMMAND') // returns {type: 'CUSTOM_COMMAND'}
```

Added test suite covering:
- Built-in command transformation
- Custom command handling
- Type safety
- Case sensitivity
- Undefined type handling


etrepum (commented):
The point of having a numeric lookup table would be to only include it in the dev tools and not the production build at all (e.g. the transform for `invariant` has the codes.json table that is only on the website and not shipped in production builds). The strategy used in this PR makes things more complicated and achieves the opposite effect (larger instead of smaller bundle, since now the bundle contains the string `'BLUR_COMMAND'`, the key for the string `BLUR` and the number `44`).

I think a simple approach would be best, let's undo most of this PR and remove the `__DEV__` check so that `type` is unconditionally included. I think the bundle size difference will be negligible and not worth creating a lookup table, the command names are short and there's only about a screen full of them.

kirandash:
> The point of having a numeric lookup table would be to only include it in the dev tools and not the production build at all (e.g. the transform for `invariant` has the codes.json table that is only on the website and not shipped in production builds). The strategy used in this PR makes things more complicated and achieves the opposite effect (larger instead of smaller bundle, since now the bundle contains the string `'BLUR_COMMAND'`, the key for the string `BLUR` and the number `44`).
> 
> I think a simple approach would be best, let's undo most of this PR and remove the `__DEV__` check so that `type` is unconditionally included. I think the bundle size difference will be negligible and not worth creating a lookup table, the command names are short and there's only about a screen full of them.

Thank you for the detailed feedback! I understand that my approach with the transform table was counterproductive.

I've simplified the implementation to just remove the `__DEV__` check and unconditionally include the type.

Please let me know if you'd like me to make any further adjustments.

Thanks again for helping me understand the better approach here! 🙏

etrepum (approved):
I updated the PR description to match the current implementation and did measurements of the lexical package built with build-prod before and after

```
$ cat packages/lexical/dist/Lexical.js | wc -c 
  125839
$ cat packages/lexical/dist/Lexical.js | wc -c 
  126857
```

```
$ cat packages/lexical/dist/Lexical.js | gzip | wc -c 
   39847
$ cat packages/lexical/dist/Lexical.js | gzip | wc -c 
   40224
```

PR #7443, opened 2025-04-07
[lexical] Bug Fix exact modifier matching for built-in keyboard shortcuts

### LINKED ISSUE #7434: Bug: Keybindings match too broadly
Lexical version: all

## Steps to reproduce:

1. Focus the playground
2. Issue a keyboard shortcut with modifiers that overlap with but do not match exactly the intended shortcut for an existing listener (e.g. Shift+Cmd+K when the existing listener is intended for Cmd+K).

## The current behavior

Lexical matches Shift+Cmd+K as Cmd+K because it does not check *all* modifiers when doing keyboard shortcut matching

## The expected behavior

Built-in keyboard shortcuts should exactly match the intended keyboard shortcut, and not handle it if other modifiers are also pressed

## Impact of fix

The built-in command listeners for keyboard presses can unintentionally preventDefault and do something else when native behavior is intended

_Originally posted by @etrepum in https://github.com/facebook/lexical/pull/7412#discussion_r2023097692_

10tacion:
## Description
Built-in keyboard shortcuts like `Cmd+B` could previously be triggered even when extra modifiers such as `Shift` were held (e.g., `Shift+Cmd+B`). This was due to Lexical not checking all modifier keys when matching keyboard shortcuts.

To address this, a new helper function `isExactShortcutMatch` was implemented and applied to the relevant key handling logic in the playground. This ensures that shortcuts only trigger when the modifier keys match exactly.

Although the original issue (#7434) used `Cmd+K` and `Shift+Cmd+K` as examples, those already behave correctly in the playground. To verify the fix, `Cmd+B` and `Shift+Cmd+B` were used instead, since `Cmd+B` triggers bold formatting.

If this change looks good, I’d appreciate any guidance on how best to add a corresponding test case for this behavior. Would it make sense to place it in one of the existing e2e test suites (e.g., Selection or Bold), or should a new test be added specifically for exact shortcut matching?

Closes #7434

### Before

- `Shift+Cmd+B` would incorrectly trigger bold formatting (same as `Cmd+B`)
- Modifier keys were not strictly checked

https://github.com/user-attachments/assets/8ac03f10-83cc-42e7-beb7-eb4e24fc4eba

### After

- `Cmd+B` correctly toggles bold
- `Shift+Cmd+B` does nothing, as expected
- Verified manually in the playground editor

https://github.com/user-attachments/assets/9e562ba7-7b92-4899-a3c3-daf135b4821f


packages/lexical/src/LexicalUtils.ts
```diff
@@ -834,270 +834,305 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
...
-  shiftKey: boolean,
-  metaKey: boolean,
-  ctrlKey: boolean,
+  event: Pick<
+    KeyboardEvent,
+    'key' | 'metaKey' | 'ctrlKey' | 'shiftKey' | 'altKey'
+  >,
 ): boolean {
-  if (shiftKey) {
+  if (event.shiftKey) {
```
etrepum, line 1054: This if statement can be deleted since the match is exact now

etrepum (commented):
This is a great start, some of the refactoring looks like it broke a few of the IS_APPLE conditions

packages/lexical/src/LexicalUtils.ts
```diff
@@ -834,270 +834,305 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
...
-export function isOpenLineBreak(key: string, ctrlKey: boolean): boolean {
+export function isOpenLineBreak(
+  event: Pick<
+    KeyboardEvent,
+    'key' | 'metaKey' | 'ctrlKey' | 'shiftKey' | 'altKey'
+  >,
+): boolean {
   // 79 = KeyO
-  return IS_APPLE && ctrlKey && key.toLowerCase() === 'o';
+  return isExactShortcutMatch(event, 'o', {metaKey: IS_APPLE});
```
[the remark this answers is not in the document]
10tacion: Thanks for the clarification!

#### I noticed the original code checks for Ctrl + O on macOS only:
```
return IS_APPLE && ctrlKey && key.toLowerCase() === 'o';

```
#### Your version uses both Ctrl and Cmd:
```
return IS_APPLE && isExactShortcutMatch(event, 'o', {ctrlKey: true, metaKey: true});
```
Just wanted to double-check—was the intended behavior to trigger this shortcut specifically with Cmd + Ctrl + O on macOS? Or was it originally meant to respond to just Ctrl + O?

10tacion:
Thanks for the review! I’ve incorporated your suggestions into the latest changes.

packages/lexical/src/LexicalUtils.ts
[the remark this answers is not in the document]
etrepum: The original Ctrl+O only on Apple was correct. The metaKey mistake came from looking at your change.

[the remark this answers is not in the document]
10tacion: Got it — I've applied the change to match the original behavior.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -834,270 +834,144 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
...
+  return isExactShortcutMatch(event, 'Enter', {shiftKey: true});
 }
 
 // Inserts a new line after the selection
 
-export function isOpenLineBreak(key: string, ctrlKey: boolean): boolean {
+export function isOpenLineBreak(event: KeyboardEventModifiers): boolean {
   // 79 = KeyO
-  return IS_APPLE && ctrlKey && key.toLowerCase() === 'o';
+  return IS_APPLE && isExactShortcutMatch(event, 'o', {metaKey: true});
```
etrepum, line 886:
```suggestion
  return IS_APPLE && isExactShortcutMatch(event, 'o', {ctrlKey: true});
```

packages/lexical/src/LexicalUtils.ts
```diff
@@ -834,270 +834,144 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
...
-export function isItalic(
-  key: string,
-  altKey: boolean,
-  metaKey: boolean,
-  ctrlKey: boolean,
-): boolean {
-  return (
-    key.toLowerCase() === 'i' && !altKey && controlOrMeta(metaKey, ctrlKey)
-  );
+const CONTROL_OR_META = {ctrlKey: !IS_APPLE, metaKey: IS_APPLE};
```
etrepum, line 856:
```suggestion
const CONTROL_OR_META = {ctrlKey: !IS_APPLE, metaKey: IS_APPLE};
const CONTROL_OR_ALT = {ctrlKey: !IS_APPLE, altKey: IS_APPLE};
```

packages/lexical/src/LexicalUtils.ts
```diff
@@ -834,270 +834,144 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
...
 }
 
-export function isDeleteWordBackward(
-  key: string,
-  altKey: boolean,
-  ctrlKey: boolean,
-): boolean {
-  return isBackspace(key) && (IS_APPLE ? altKey : ctrlKey);
+export function isDeleteWordBackward(event: KeyboardEventModifiers): boolean {
+  return isExactShortcutMatch(event, 'Backspace', CONTROL_OR_META);
```
etrepum, line 890:
```suggestion
  return isExactShortcutMatch(event, 'Backspace', CONTROL_OR_ALT);
```

packages/lexical/src/LexicalUtils.ts
```diff
@@ -834,270 +834,144 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
...
 }
 
-export function isDeleteWordForward(
-  key: string,
-  altKey: boolean,
-  ctrlKey: boolean,
-): boolean {
-  return isDelete(key) && (IS_APPLE ? altKey : ctrlKey);
+export function isDeleteWordForward(event: KeyboardEventModifiers): boolean {
+  return isExactShortcutMatch(event, 'Delete', CONTROL_OR_META);
```
etrepum, line 894:
```suggestion
  return isExactShortcutMatch(event, 'Delete', CONTROL_OR_ALT);
```

etrepum (commented):
Looks like there's a couple more mistakes in here, might be the reason why the tests are failing. I'll go ahead and apply these to see

etrepum:
It looks like tests are still failing, so there must still be some behavior change. Without further investigation I can't be sure if the problem is the test issuing keys that were matched incorrectly before which requires fixing of the tests, or if there's still some mistakes to catch in this PR.

10tacion:
Thanks for pointing this out — I’ll take a closer look to figure out what’s causing the issue. Appreciate you highlighting the potential areas to investigate.

10tacion:
I'm currently looking into the exact shortcut matching behavior, and I had a quick question for clarification:

Are all built-in shortcuts expected to match modifiers strictly (e.g., Cmd+B should not trigger with Shift+Cmd+B), or are there some cases where additional modifiers are intentionally allowed?

etrepum:
I don't think there are any situations where the additional modifiers are allowed in the keyboard handlers, although there may be functions that are used in multiple places that in some situations are supposed to only match a specific key ignoring modifiers (e.g. `isDelete`). Perhaps a good solution there is to have constants like `const DELETE_KEY = 'Delete'` and use that instead of a function when that is the intent

10tacion:
I noticed that after selecting table cells using Table selection, keyboard shortcuts like Cmd+A, Cut, Paste, Align, Insert Image, etc. no longer work. The shortcuts are ignored completely.

This started happening after I introduced stricter key matching logic (e.g. isExactShortcutMatch), but I’m not sure if it's directly related.

Do you have any idea what might be causing this?

10tacion:
I've analyzed the root cause of why our tests are failing after applying the isExactShortcutMatch function. 
Before the change, keyboard events were processed independently by extracting individual properties (key, ctrlKey, etc.) and passing them separately to handler functions. This allowed text input in table cells to work correctly, including sequences like typing "bb".
After implementing isExactShortcutMatch, we're now passing the entire event object to handler functions, which changes how events are interpreted. The key issue is that when typing characters in sequence (like "bb"), the second "b" might be incorrectly interpreted as a shortcut because the function doesn't properly distinguish between regular text input and shortcuts in table contexts.

etrepum:
I don’t think that analysis is correct, event objects don’t change while an event is being processed

10tacion:
I realized the issue wasn’t just about strict matching

- Strictly match a key + exact modifier combination (e.g. Cmd+Z)
- Loosely match a key while excluding some modifiers (e.g. Tab without Ctrl — Shift+Tab should still be allowed)

To reflect that, I added isKeyWithExcludedModifiers. For example:

```javascript
export function isTab(event: KeyboardEventModifiers): boolean {
  return isKeyWithExcludedModifiers(event, 'Tab', {
    ctrlKey: true,
    metaKey: true,
  });
}
```
This replaces previous isExactShortcutMatch usage where strict matching wasn’t necessary.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -835,270 +835,176 @@ export function $shouldInsertTextAfterOrBeforeTextNode(
...
+  KeyboardEvent,
+  'key' | 'metaKey' | 'ctrlKey' | 'shiftKey' | 'altKey'
+>;
 
-export function isBold(
-  key: string,
-  altKey: boolean,
-  metaKey: boolean,
-  ctrlKey: boolean,
+export function isKeyWithExcludedModifiers(
```
etrepum, line 843: I think a better way to do this would be to change the exact match function around a bit to allow for an 'any' state, for example:

```ts
export type KeyboardEventModifiers = Pick<KeyboardEvent, "key" | "altKey" | "ctrlKey" | "shiftKey" | "metaKey">;
export type KeyboardEventModifierMask = {
  [K in Exclude<keyof KeyboardEventModifiers, "key">]?: boolean | undefined | 'any';
};
function matchModifier(
  event: KeyboardEventModifiers,
  mask: KeyboardEventModifierMask,
  prop: keyof KeyboardEventModifierMask,
): boolean {
  const expected = mask[prop] || false;
  return expected === 'any' || expected === event[prop];
}
export function isExactShortcutMatch(
  event: KeyboardEventModifiers,
  expectedKey: string,
  mask: KeyboardEventModifierMask,
): boolean {
  return (
    matchModifier(event, mask, 'altKey') &&
    matchModifier(event, mask, 'ctrlKey') &&
    matchModifier(event, mask, 'shiftKey') &&
    matchModifier(event, mask, 'metaKey') &&
    event.key.toLowerCase() === expectedKey.toLowerCase()
  );
}
```

This way instead of implicitly building the set of ignored modifiers you explicitly choose which modifiers to ignore. Less error-prone this way.

In the future I think what we would want to do is have these matching functions return data so that we can use that to build a data structure to do the matching in constant time.

10tacion: Thanks for the helpful insight — especially the idea of returning match data for constant-time lookups. That’s a great direction I hadn’t considered.

It also made me revisit the matching logic. I thought treating undefined in the modifier mask as 'any' would help avoid potential issues, since interpreting it as false could accidentally reject valid key inputs when a modifier is simply omitted. Based on that, I updated the logic and removed isKeyWithExcludedModifiers, consolidating everything into isExactShortcutMatch for simplicity.

Let me know if you have any suggestions or improvements!

etrepum: Really the only values that should be specified explicitly are true or ‘any’. false and undefined are there mostly to make the types a bit cleaner to work with. I don’t think that requiring every modifier to be explicitly listed does anything but to increase code size and make them a bit less convenient to write (which encourages copy and paste and leads to mistakes)

10tacion: Thank you for the detailed explanation — it helped a lot in understanding the intended usage. I've reflected that in the latest changes.

10tacion:
I really appreciate all the detailed feedback and support throughout this PR. I've applied the suggested changes and believe everything is now in order. Please let me know if any further adjustments are needed!

etrepum:
Tentatively I think this is in a good state, but the `'any'` workarounds should get a more careful audit that I haven't had time for (to verify all use sites of those commands). It will be another few days before merge, but no feedback to address at the moment.

etrepum (approved):
Thanks for your patience and making this nice improvement!

I made some minor changes to keep the implementation consistent and make this code more generally useful outside of the core. Mostly what I did was write comments for the commands related to event processing here so that there's API documentation for what they are supposed to mean, especially in cases where some or all modifiers are ignored.

takuyakanbr:
Hmm just realized this should probably be considered a breaking change. Internally we have quite a bit of plugins relying on the non-exact matching for custom handling of the Enter key (on event `KEY_ENTER_COMMAND`)

etrepum:
I don’t think it was intentional to make the KEY_[X]_COMMAND behave differently. The code was a mess, some of them did check for modifiers but this PR may have added a few more constraints

10tacion:
I’m aware that the change might affect existing behaviors more broadly, and I’m happy to help if any adjustments are needed.

PR #7447, opened 2025-04-09
[lexical-table][lexical-playground] Bug Fix: Handle backspace deletion of tables with merged cells

### LINKED ISSUE #7444: Bug: When there are merged cells in a table, the backspace key does not delete the table when the entire table is selected.
When there are merged cells in a table, the backspace key does not delete the table when the entire table is selected.
It works normally when there are no merged cells in the table.

Lexical version: 0.30.0

## The current behavior

https://github.com/user-attachments/assets/c48d76fa-187f-4c6b-947f-22cf94699f94

## The expected behavior

https://github.com/user-attachments/assets/fbfd5946-d195-4822-bcb5-1fee127e3520

## Impact of fix
- It provides consistent usability regardless of merged cells.

kirandash:
### Current behavior:
Currently, when a table containing merged cells is fully selected and the backspace key is pressed, the table does not get deleted. This behavior is inconsistent with tables that don't have merged cells, where backspace deletion works as expected.

### Changes in this PR:
This PR fixes the table deletion behavior to work consistently regardless of whether the table has merged cells or not. When a table is fully selected and backspace is pressed, the table will be properly deleted and replaced with an empty paragraph, maintaining consistent behavior across all table types.

Changes made:
1. Updated `TableObserver.$clearText` to properly handle merged cells during table deletion
2. Added e2e test to verify table deletion behavior with merged cells

Closes #7444

### Before

When a table with merged cells is fully selected and backspace is pressed, the table remains in place and is not deleted.

https://github.com/user-attachments/assets/5a3ad33e-2532-4b93-bb1d-f3d64d2c4444

### After

When a table with merged cells is fully selected and backspace is pressed, the table is properly deleted and replaced with an empty paragraph, matching the behavior of regular tables.

https://github.com/user-attachments/assets/dbb170b9-c1bd-4107-9f2b-5d015c901963

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -6840,6 +6840,51 @@ test.describe.parallel('Tables', () => {
       `,
     );
   });
+
+  test('Can delete table when fully selected with merged cells', async ({
+    page,
+    isPlainText,
+    isCollab,
+  }) => {
+    test.skip(isPlainText);
+    await initialize({isCollab, page});
+
+    await focusEditor(page);
+
+    // Insert a 3x3 table
+    await insertTable(page, 3, 3);
+
+    // Merge some cells to create a complex merged cell structure
+    await selectCellsFromTableCords(
+      page,
+      {x: 0, y: 0},
+      {x: 1, y: 1},
+      true,
+      false,
+    );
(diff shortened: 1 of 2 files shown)


etrepum (dismissed):
Looks like the right approach to me, will merge assuming tests are all passing!

etrepum (approved):

PR #7453, opened 2025-04-10
[lexical-yjs][lexical] Refactor: Simplify removeFromParent internal operations

### LINKED ISSUE #7452: Refactor: Simplify internal operations in removeFromParent function
## Description
The `removeFromParent` function is currently duplicated in `packages/lexical-yjs/src/Utils.ts` and `packages/lexical/src/LexicalUtils.ts` with identical implementations. They both have a TODO comment noting that the function "duplicates a bunch of operations" internally. The function performs multiple `getWritable()` calls and has nested conditional logic that can be simplified while maintaining the same functionality.

### Areas for Improvement
1. Multiple `getWritable()` calls on related nodes
2. Separate conditional blocks for related pointer updates
3. Nested null checks that could be flattened
4. Repeated sibling key access

### Proposed Solution
1. Batch related pointer updates together
2. Get writable nodes upfront to avoid repeated calls
3. Simplify conditional logic for better readability
4. Maintain the same functionality with cleaner code structure

### Core Maintainers
- Clearer code flow and intent
- Easier to maintain and debug
- Better testability with simplified logic

### Performance Benefits
- Reduced number of `getWritable()` calls

### Usage Context
The function is used in core node removal operations:
- Text editing (backspace/delete)
- List manipulation
- Node replacement
- Selection changes

The improvement would make the code more maintainable while keeping the same robust functionality that handles parent-child relationship cleanup

kirandash:
### Current behavior:
- The removeFromParent function in both packages makes multiple separate getWritable() calls
- Has nested conditional logic for sibling pointer updates
- Contains repeated null checks and sibling key access operations
- Has a TODO comment noting operation duplication

### Changes being added:
- Batch all getWritable() calls at the start of the function
- Consolidate related pointer updates into cleaner blocks
- Simplify null checks and conditional logic
- Maintain identical functionality with improved code structure
- Remove the TODO comment as the operations are now optimized

Closes #7452

## Test plan

All existing tests should continue to pass as the functionality remains unchanged.

### DIFF
--- a/packages/lexical-yjs/src/Utils.ts
+++ b/packages/lexical-yjs/src/Utils.ts
@@ -566,47 +566,38 @@ export function removeFromParent(node: LexicalNode): void {
     const writableParent = oldParent.getWritable();
     const prevSibling = node.getPreviousSibling();
     const nextSibling = node.getNextSibling();
-    // TODO: this function duplicates a bunch of operations, can be simplified.
+
+    // Get writable siblings once
+    const writablePrevSibling =
+      prevSibling !== null ? prevSibling.getWritable() : null;
+    const writableNextSibling =
+      nextSibling !== null ? nextSibling.getWritable() : null;
+
+    // Update parent's first/last pointers
     if (prevSibling === null) {
-      if (nextSibling !== null) {
-        const writableNextSibling = nextSibling.getWritable();
-        writableParent.__first = nextSibling.__key;
-        writableNextSibling.__prev = null;
-      } else {
-        writableParent.__first = null;
-      }
-    } else {
-      const writablePrevSibling = prevSibling.getWritable();
-      if (nextSibling !== null) {
-        const writableNextSibling = nextSibling.getWritable();
-        writableNextSibling.__prev = writablePrevSibling.__key;
-        writablePrevSibling.__next = writableNextSibling.__key;
-      } else {
-        writablePrevSibling.__next = null;
-      }
-      writableNode.__prev = null;
+      writableParent.__first = nextSibling !== null ? nextSibling.__key : null;
     }
     if (nextSibling === null) {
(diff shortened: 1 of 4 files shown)


packages/lexical-yjs/src/Utils.ts
```diff
@@ -566,47 +566,38 @@ export function removeFromParent(node: LexicalNode): void {
     const writableParent = oldParent.getWritable();
     const prevSibling = node.getPreviousSibling();
     const nextSibling = node.getNextSibling();
-    // TODO: this function duplicates a bunch of operations, can be simplified.
+
```
etrepum, line 569: Does this function differ at all from the one in LexicalUtils? if not, perhaps we should instead export the version from LexicalUtils from the lexical package (likely marked as `@internal` to discourage anyone from using it) and remove this one altogether.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -345,47 +345,38 @@ export function removeFromParent(node: LexicalNode): void {
     const writableParent = oldParent.getWritable();
     const prevSibling = node.getPreviousSibling();
     const nextSibling = node.getNextSibling();
-    // TODO: this function duplicates a bunch of operations, can be simplified.
+
+    // Get writable siblings once
+    const writablePrevSibling =
+      prevSibling !== null ? prevSibling.getWritable() : null;
+    const writableNextSibling =
+      nextSibling !== null ? nextSibling.getWritable() : null;
```
etrepum, line 353: you could reduce code size a little more by storing the nextSiblingKey and prevSiblingKey in variables since these values are used for both updating the parent's `__first` and `__last` as well as updating the sibling `__next` and __prev`

etrepum (approved):
Looks good!

PR #7457, opened 2025-04-10
[lexical] Chore: Update internalMarkNodeAsDirty TODO comment

### LINKED ISSUE #7456: Refactor: Split internalMarkNodeAsDirty into Dedicated Element/Leaf Functions
## Description
The current `internalMarkNodeAsDirty` function handles both element and leaf nodes in a single function, with branching logic. As noted in the TODO, this should be split into two dedicated functions for better separation of concerns.

### Current Implementation
```typescript
export function internalMarkNodeAsDirty(node: LexicalNode): void {
  errorOnInfiniteTransforms();
  const latest = node.getLatest();
  const parent = latest.__parent;
  const editorState = getActiveEditorState();
  const nodeMap = editorState._nodeMap;
  const dirtyElements = editor._dirtyElements;
  if (parent !== null) {
    internalMarkParentElementsAsDirty(parent, nodeMap, dirtyElements);
  }
  const key = latest.__key;
  editor._dirtyType = HAS_DIRTY_NODES;
  if ($isElementNode(node)) {
    dirtyElements.set(key, true);
  } else {
    editor._dirtyLeaves.add(key);
  }
}
```

### Proposed Solution

Split into two dedicated internal functions:

```typescript
function internalMarkElementNodeAsDirty(node: ElementNode): void {
  errorOnInfiniteTransforms();
  const latest = node.getLatest();
  // Element-specific logic
}

function internalMarkLeafNodeAsDirty(node: TextNode | LineBreakNode | DecoratorNode<unknown>): void {
  errorOnInfiniteTransforms();
  const latest = node.getLatest();
  // Leaf-specific logic
}

export function internalMarkNodeAsDirty(node: LexicalNode): void {
  if ($isElementNode(node)) {
    internalMarkElementNodeAsDirty(node);
  } else if ($isLeafNode(node)) {
    internalMarkLeafNodeAsDirty(node);
  } else {
    invariant(false, 'Unknown node type'); or internalMarkLeafAsDirty(node as DecoratorNode<unknown>);
  }
}
```

### Benefits
1. Better separation of concerns
2. More maintainable code
3. Type safety improvements
4. Clearer code paths for each node type
5. Easier to extend or modify behavior for specific node types

### Who Would Benefit
1. **Core Maintainers**
   - Clearer code organization
   - Easier to maintain and modify node-specific logic

2. **Plugin Developers**
   - Better understanding of how different node types are handled
   - Clearer extension points for custom node types

## Implementation Notes
- No breaking changes to public API
- Internal refactoring only
- Should maintain existing behavior

kirandash:
### Current behavior:
- `internalMarkNodeAsDirty` has a TODO comment suggesting to split it into separate element/leaf functions
- After analysis and review feedback, the current single-function implementation is actually preferred

### Changes being made:
1. Remove the outdated TODO comment as the current implementation is better because:
   - Avoids unnecessary code duplication
   - Maintains better performance
   - Keeps code size minimal
   - Improves maintainability

The change is minimal but important - removing an outdated TODO comment that would lead developers down a suboptimal path. The current implementation is more maintainable and efficient than splitting it would be.

Closes #7456

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -407,7 +407,6 @@ export function internalMarkNodeAsDirty(node: LexicalNode): void {
   if ($isElementNode(node)) {
     dirtyElements.set(key, true);
   } else {
-    // TODO split internally MarkNodeAsDirty into two dedicated Element/leave functions
     editor._dirtyLeaves.add(key);
   }
 }


packages/lexical/src/LexicalUtils.ts
```diff
@@ -404,11 +420,20 @@ export function internalMarkNodeAsDirty(node: LexicalNode): void {
...
-    dirtyElements.set(key, true);
+    internalMarkElementAsDirty(node);
+  } else if ($isLeafNode(node)) {
+    internalMarkLeafAsDirty(node);
   } else {
-    // TODO split internally MarkNodeAsDirty into two dedicated Element/leave functions
-    editor._dirtyLeaves.add(key);
+    // For custom nodes that don't extend the standard leaf types,
+    // we treat them as decorator nodes for dirty marking purposes
+    internalMarkLeafAsDirty(node as DecoratorNode<unknown>);
```
etrepum, line 436: It can be assumed that anything that isn't an element is a leaf node, it's not allowed to subclass LexicalNode directly (the value is not even exported, so it is tricky to even try)

etrepum (commented):
I think the real improvement here would be to delete the comment and leave the code the same. Taking one function and making it into three functions, where two of them have a lot of duplicated code, seems like a net negative on most fronts to me (performance, code size, maintainability)

kirandash:
> I think the real improvement here would be to delete the comment and leave the code the same. Taking one function and making it into three functions, where two of them have a lot of duplicated code, seems like a net negative on most fronts to me (performance, code size, maintainability)

Thanks @etrepum for the feedback! I actually had the same concerns about code duplication while implementing the split. But since I saw the TODO, I thought there might have been a specific reason for wanting to split these functions that I was missing. 

I've now updated the PR to simply remove the TODO comment.

etrepum (approved):

PR #7477, opened 2025-04-16
[lexical-react] Feature: make check for entity boundary configurable in LexicalTypeaheadMenuPlugin

asevich:
### What is the current behavior that you are modifying?
LexicalTypeaheadMenuPlugin can only be triggered if there is no TextNode before trigger, i.e. whitespace or a paragraph and there is no way to configure this behaviour, i.e.
- @Bob@B - won't trigger menu

### What are the behavior or changes that are being added by this PR?
Make this behaviour configurable by ignoring element boundary check


asevich:
Does this PR break some collab tests or are they just flaky and needs a rerun?

etrepum:
I think #7482 should solve the flaky test situation that is blocking this one

etrepum (approved):
Seems straightforward, but would take a fair bit of infrastructure to add e2e testing for this. Seems fine to allow this override for now.

PR #7479, opened 2025-04-17
[lexical] Chore: Revert Enter command to use inexact matching

takuyakanbr:
## Description

#7443 modified built-in keyboard shortcuts (including Enter and Shift+Enter) to use exact modifier matching. Unfortunately internally we have a decent amount of plugins relying on the old inexact matching to handle the Enter key (with various modifiers), across both public and internal products. This PR reverts the triggering of `KEY_ENTER_COMMAND` to the state before #7443, to give us time to migrate our plugins to use `KEY_DOWN_COMMAND` instead (if we want to move ahead with exact-matching for the Enter command).

## Test plan

Existing tests

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -926,11 +926,20 @@ export function isUnderline(event: KeyboardEventModifiers): boolean {
 }
 
 export function isParagraph(event: KeyboardEventModifiers): boolean {
-  return isExactShortcutMatch(event, 'Enter', {});
+  return isExactShortcutMatch(event, 'Enter', {
+    altKey: 'any',
+    ctrlKey: 'any',
+    metaKey: 'any',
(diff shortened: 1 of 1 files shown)


lilshady (approved):

lilshady:
cc: @etrepum @10tacion

etrepum:
I think these semantics are fine but you should change the docs on the command to reflect how KEY_ENTER_COMMAND is matched

etrepum:
I think in general the KEY_X_COMMANDs are confusing because you would expect them to either ignore all modifiers or only fire when no modifiers are down but most of them do not behave in either way.

The way it probably should have worked is to have a KEY_DOWN_COMMAND handle dispatching of KEY_X_COMMANDs, ignoring any modifiers and short-circuit if any of them return true before handling the editor priority shortcuts.

PR #7480, opened 2025-04-17
[lexical-playground] Chore: Skip flaky autolinks test in collab

etrepum:
## Description

Skips this test in collab as it seems flaky in that environment. Primarily this test seems to be testing the regex so it doesn't seem particularly valuable to verify the collab behavior

```
1 failed
[1]     [chromium] › packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs:491:3 › Auto Links › Can convert URLs into links 
```

## Test plan

Removes a test from running in collab


etrepum:
Superceded by #7482

PR #7486, opened 2025-04-21
[lexical-react][lexical-yjs] Feature: Add cursor filters

simonecervini:
Added a new optional `cursorFilter` parameter to the `syncCursorPositions` function and exposed it as a `LexicalCollaborationPlugin`’s prop.

The function is defined as `(clientID: number, awareness: UserState) => boolean` and is useful for hiding cursors without impacting awareness data.

## Motivation

Let’s consider an editor with 3 active presences:
- User A
- User B
- User B, in another tab (or side-by-side window)

So we have 3 different client IDs, and each user can see up to two cursors. It’s not necessarily a problem that User B can see the cursor state from the other tab/window, but it can cause issues when you start to consider refreshes and failures of the provider to push the new awareness state when the tab is closed. In the worst case, you have to wait for the timeout to see the old cursors get “garbage collected,” and during this time, a new cursor can appear with each refresh (which will still be automatically deleted after, say, 30 seconds).

In the attached video, I show this scenario with a custom provider that doesn’t include the listener recommended by Yjs
```ts
window.addEventListener('beforeunload', () => {
  awarenessProtocol.removeAwarenessStates(
    this.awareness, [doc.clientID], 'window unload'
  )
})
```

We can easily solve this problem by hiding, only for User B, the cursor from the other tab. Hence the `cursorFilter` addition.

https://github.com/user-attachments/assets/1ba0aa93-dec0-49f3-94d7-35589427f8fc

### DIFF
--- a/packages/lexical-yjs/src/SyncCursors.ts
+++ b/packages/lexical-yjs/src/SyncCursors.ts
@@ -410,14 +410,18 @@ function getCollabNodeAndOffset(
   return [null, 0];
 }
 
+export type CursorFilter = (clientID: number, awareness: UserState) => boolean;
+
 export type SyncCursorPositionsFn = (
   binding: Binding,
   provider: Provider,
+  cursorFilter?: CursorFilter,
 ) => void;
 
 export function syncCursorPositions(
   binding: Binding,
   provider: Provider,
+  cursorFilter?: CursorFilter,
 ): void {
   const awarenessStates = Array.from(provider.awareness.getStates());
   const localClientID = binding.clientID;
@@ -429,8 +433,9 @@ export function syncCursorPositions(
   for (let i = 0; i < awarenessStates.length; i++) {
     const awarenessState = awarenessStates[i];
     const [clientID, awareness] = awarenessState;
+    const filtered = cursorFilter ? cursorFilter(clientID, awareness) : true;
 
-    if (clientID !== localClientID) {
+    if (clientID !== localClientID && filtered) {
       visitedClientIDs.add(clientID);
       const {name, color, focusing} = awareness;
       let selection = null;
--- a/packages/lexical-react/src/LexicalCollaborationPlugin.tsx
+++ b/packages/lexical-react/src/LexicalCollaborationPlugin.tsx
@@ -17,11 +17,14 @@ import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {
   Binding,
   createBinding,
+  type CursorFilter,
   ExcludedProperties,
   Provider,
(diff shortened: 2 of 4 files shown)


etrepum (commented):
I think a better refactoring of this would add syncCursorPositions as an argument to the plugin for parity with useYjsCollaboration and have the caller be responsible for wrapping it with extra options or functionality if they want to.

For filtering the positions it might be nice to have an options argument to syncCursorPositions to get the awareness states that defaults to `(_binding, provider) => provider.awareness.getStates()` which would let you introduce filtering or other transformations as needed.

simonecervini:
@etrepum we could probably make `getAwarenessStates` parameterless, since the `binding` and `provider` objects are likely redundant when providing a custom sync function. Consider the following example:
```ts
const myCustomSyncFunction: SyncCursorPositionsFn = (binding, provider) => {
  syncCursorPositions(binding, provider, {
    // _provider here is unnecessary — we can just use provider
    getAwarenessStates: (_, _provider) => {
      const filteredStates = _provider.awareness.getStates();
      for (const clientId of userClientIds) {
        filteredStates.delete(clientId);
      }
      return filteredStates;
    },
  });
};
```

etrepum:
Having parameters allows you to define it at the module level rather than as a closure, makes reuse easier.

packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -413,13 +413,23 @@ function getCollabNodeAndOffset(
...
+  ) => Map<number, UserState>;
+};
+
 export function syncCursorPositions(
   binding: Binding,
   provider: Provider,
+  options?: SyncCursorPositionsOptions,
 ): void {
-  const awarenessStates = Array.from(provider.awareness.getStates());
+  const {getAwarenessStates = provider.awareness.getStates} = options ?? {};
```
etrepum, line 431: There are a lot of test failures, maybe because of this line? `const fn = foo.bar; fn()` is not equivalent to `foo.bar()` in javascript when classes are involved due to the way methods and `this` work. This would have to be either `provider.awareness.getStates.bind(provider.awareness)` or something equivalent like `() => provider.awareness.getStates()` or defining a module level function that handles it.

```ts
function getAwarenessStatesDefault(_binding: Binding, provider: Provider): Map<number, UserState> {
  return provider.awareness.getStates();
}
```

simonecervini: Yeah, I’d go with a module-level function then, I think it's clearer personally

etrepum (approved):

PR #7494, opened 2025-04-26
[lexical][lexical-utils] Refactor: deprecate $getAncestor in favour of $findMatchingParent

### LINKED ISSUE #5311: Duplicate function `$findMatchingParent` and `$getAncestor`
Maybe it's not that important... but it seems that these two functions do exactly the same thing.

The only difference is that `getAncestor` is a type guard as well.

What should be done? Options

1. Modify the return type of `$findMatchingParent` so that it is also a type guard, and make one function call the other, but leave both to avoid a breaking change.
2. Delete one of the two functions. In case of removing `getAncestor` make `$findMatchingParent` a type guard.

Javv4d:
[lexical][lexical-utils] Refactor: deprecate `$getAncestor` in favour of `$findMatchingParent`

## Description

Issue #5311 notes that **`$getAncestor` and `$findMatchingParent` duplicate logic**.  
This PR:

1. **Re-exports ` $getAncestor` as a deprecated alias** (typedoc deprecation tag).
2. **Extends `$findMatchingParent`’s overload** so it can be used as the original type-guard.
3. **Replaces ~300 internal call-sites** across the monorepo (`rg \$getAncestor | wc -l` before/after).
4. **Adds a dedicated unit test** (`packages/lexical-utils/src/__tests__/unit/findMatchingParent.test.ts`)
   that exercises both the boolean-predicate and type-predicate overloads.

No external behaviour or public API surface changes (other than the deprecation warning).

Closes #5311

### Automated

| Package            | Command                  | Result |
|--------------------|--------------------------|--------|
| root               | `npm run unit-test` (Jest)       | **1 failed → 1 failed** (baseline unchanged) |
| lexical-utils only | `npm run test -t findMatchingParent` | **new test passes** |

### Manual

* Built playground (`yarn start`), verified editor still functions (node insertion, selection, list indentation etc.).
* Ran `yarn lint` + `yarn prettier --check` (clean).

## Notes for reviewers

*Kept the alias export so downstream packages that import `$getAncestor` don’t break; a follow-up PR can remove it in the next semver major.*

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -662,7 +662,7 @@ export function $toggleLink(
     const firstNode = nodes[0];
     // if the first node is a LinkNode or if its
     // parent is a LinkNode, we update the URL, target and rel.
-    const linkNode = $getAncestor(firstNode, $isLinkNode);
+    const linkNode = $findMatchingParent(firstNode, $isLinkNode);
     if (linkNode !== null) {
       return updateLinkNode(linkNode);
     }
@@ -674,7 +674,7 @@ export function $toggleLink(
       if (!node.isAttached()) {
         continue;
       }
-      const parentLinkNode = $getAncestor(node, $isLinkNode);
+      const parentLinkNode = $findMatchingParent(node, $isLinkNode);
       if (parentLinkNode) {
         updateLinkNode(parentLinkNode);
         continue;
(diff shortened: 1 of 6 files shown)


etrepum (commented):
The `// #5311` comments don't really provide any value to future readers of the code, let's get rid of them all.

Not sure what the playground build failure is, I don't have access to see that, but I suspect it's because some of the new imports are coming from `@lexical/utils` instead of `lexical`. The separate implementation of `$findMatchingParent` in `@lexical/utils` can be replaced with an alias to the version in `lexical` and deprecated as well.

The CLA will have to be accepted before this PR is considered for merge

etrepum:
Looks like the suspicion is likely correct, the [unit tests fail with the following error](https://github.com/facebook/lexical/actions/runs/14676431983/job/41209823787?pr=7494):
```
Error: packages/lexical-selection/src/range-selection.ts has an undeclared dependency in its import of @lexical/utils.
Add the following to the dependencies in packages/lexical-selection/package.json: "@lexical/utils": "0.30.0"
```

etrepum:
This PR can not be considered until the CLA is accepted. Please re-open or create a new PR if you intend to accept the CLA and fix the issues with it.

PR #7498, opened 2025-04-29
[lexical-table] Bug Fix: Add missing DOM import for TableNode frozen rows and columns

Iaotle:
## Description
Currently the table node does not set frozen columns and frozen rows on conversion. Since a replacement `TableNode` has no access to the `domNode` in `$convertTableElement`, only row striping is correctly set on the node.


etrepum:
Adding tests for this behavior would avoid future regressions

Iaotle:
I don't really have the time to add tests for this. Debugging this was already quite time consuming unfortunately

etrepum (approved):

etrepum:
Will merge after #7502

PR #7499, opened 2025-04-29
[lexical-markdown] Bug Fix: Link Transformer URL Protocol Handling

### LINKED ISSUE #7497: Bug: Link Transformer doesn't handle URLs without HTTP/HTTPS protocols correctly
# Link Transformer doesn't handle URLs without protocols correctly

When creating links without protocols (like `[youtube](www.youtube.com)`) using Link transformer, they're treated as relative paths instead of external URLs. This causes links to navigate to incorrect destinations (e.g., https://playground.lexical.dev/www.youtube.com instead of https://www.youtube.com).

**Lexical version**: 0.3.0

## Steps To Reproduce
1. Create a new link in the Lexical playground using link transformer
2. Enter a URL without protocol (e.g., `[youtube](www.youtube.com)`)
3. Click on the link in the editor
4. Observe that the link opens `https://playground.lexical.dev/www.youtube.com` instead of `https://www.youtube.com`

## Link to code example
The issue can be reproduced in the Lexical playground: https://playground.lexical.dev/

The problematic code appears to be in:
https://github.com/facebook/lexical/blob/1b08c62441a1fd40b84a0d9cea2f43c94aeca0cc/packages/lexical-markdown/src/MarkdownTransformers.ts#L540

## The current behavior
When creating a link without a protocol (e.g., `www.youtube.com`), the link is treated as a relative path rather than an external URL.

For example, creating a link with text "YouTube" pointing to `www.youtube.com` results in a link that navigates to:
`https://playground.lexical.dev/www.youtube.com` (or whatever the base URL is) instead of `https://www.youtube.com`

This happens because the link transformer isn't properly detecting and handling URLs that don't include a protocol.

## The expected behavior
The link transformer should recognize common URL patterns that lack URL protocols (like "www.example.com") and automatically prepend "https://". For specialized content like emails or phone numbers, it should add the appropriate protocol prefixes ("mailto:" or "tel:") when creating the link.

So when a user creates a link to `www.youtube.com`, it should be treated as `https://www.youtube.com` and navigate to the correct external site.

## Impact of fix
- **Severity**: Medium - This is a usability issue that affects a common use case (linking to external sites)
- This fix would benefit all users of Lexical who create links, as it's a common pattern to type URLs without the protocol
- Specifically, it would improve the experience for content creators who frequently link to external resources
- The current behavior is counterintuitive and could cause confusion when links don't work as expected

I have already implemented a fix locally and would be happy to submit a PR once this issue is approved.

AashishRichhariya:
## Description

Currently, when creating links without protocols (like `[youtube](www.youtube.com)`) using the Link transformer, they're treated as relative paths instead of external URLs. This causes links to navigate to incorrect destinations (e.g., https://playground.lexical.dev/www.youtube.com instead of https://www.youtube.com).

This PR:
1. Adds URL handling utility functions to detect and format different types of URLs
2. Updates the LINK transformer to use these utilities
3. Adds tests to ensure the behavior works correctly

The fix improves the user experience for content creators by:
- Adding https:// to common URLs without protocols
- Adding mailto: to email addresses
- Adding tel: to phone numbers
- Preserving relative URLs

Closes #7497

### Before
When clicking a link created as `[YouTube](www.youtube.com)`, the browser navigates to a relative URL like `https://playground.lexical.dev/www.youtube.com` instead of `https://www.youtube.com`.

https://github.com/user-attachments/assets/f2361acd-901c-4d53-b32f-57731f29ce73

### After
When clicking a link created as `[YouTube](www.youtube.com)`, the browser correctly navigates to `https://www.youtube.com`.

https://github.com/user-attachments/assets/054ab3ac-2c47-4689-ab0e-8ddac6d26554

### DIFF
--- a/packages/lexical-markdown/src/utils.ts
+++ b/packages/lexical-markdown/src/utils.ts
@@ -460,3 +460,43 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
       MARKDOWN_EMPTY_LINE_REG_EXP.test(firstChild.getTextContent()))
   );
 }
+
+// URL validation patterns
+export const EMAIL_REGEX =
+  /(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))/;
+
+export const PHONE_NUMBER_REGEX = /^\+?[0-9\s()-]{5,}$/;
+
+/**
+ * Formats a URL string by adding appropriate protocol if missing
+ *
+ * @param url - URL to format
+ * @returns Formatted URL with appropriate protocol
+ */
+export function formatUrl(url: string): string {
+  // Check if URL already has a protocol
+  if (url.match(/^[a-z]+:/i)) {
+    // URL already has a protocol, leave it as is
+    return url;
+  }
+  // Check if it's a relative path
+  else if (url.startsWith('/')) {
+    // Relative path, leave it as is
+    return url;
+  }
+  // Check for email address
+  else if (EMAIL_REGEX.test(url)) {
+    return `mailto:${url}`;
+  }
+  // Check for phone number
+  else if (PHONE_NUMBER_REGEX.test(url)) {
+    return `tel:${url}`;
+  }
+  // For everything else that looks like a URL (starting with www. or containing dots)
+  else if (/^www\.|(\.[a-z]{2,})/i.test(url)) {
+    return `https://${url}`;
+  }
+
+  // If nothing matches, return the original URL
+  return url;
+}
--- a/packages/lexical-markdown/src/utils.ts
+++ b/packages/lexical-markdown/src/utils.ts
@@ -476,18 +476,13 @@ export function formatUrl(url: string): string {
     return url;
   }
   // Check if it's a relative path (starting with '/', '.', or '#')
-  else if (url.startsWith('/') || url.startsWith('.') || url.startsWith('#')) {
+  else if (url.match(/^[/#.]/)) {
     // Relative path, leave it as is
     return url;
   }
 
-  // Check for email address - ensure it has exactly one '@' with non-empty username and domain portions
-  else if (
-    url.includes('@') &&
-    url.split('@').length === 2 &&
-    url.split('@')[0].length > 0 &&
-    url.split('@')[1].length > 0
-  ) {
+  // Check for email address
+  else if (url.includes('@')) {
     return `mailto:${url}`;
   }
 
@@ -495,11 +490,7 @@ export function formatUrl(url: string): string {
   else if (PHONE_NUMBER_REGEX.test(url)) {
     return `tel:${url}`;
   }
-  // For everything else that looks like a URL (starting with www. or containing dots)
-  else if (/^www\.|(\.[a-z]{2,})/i.test(url)) {
-    return `https://${url}`;
-  }
 
-  // If nothing matches, return the original URL
-  return url;
+  // For everything else, return with https:// prefix
+  return `https://${url}`;
 }
--- a/packages/lexical-markdown/src/utils.ts
+++ b/packages/lexical-markdown/src/utils.ts
@@ -481,8 +481,13 @@ export function formatUrl(url: string): string {
     return url;
   }
 
-  // Check for email address
-  else if (url.includes('@')) {
+  // Check for email address - ensure it has exactly one '@' with non-empty username and domain portions
+  else if (
+    url.includes('@') &&
+    url.split('@').length === 2 &&
+    url.split('@')[0].length > 0 &&
+    url.split('@')[1].length > 0
+  ) {
     return `mailto:${url}`;
   }
 
--- a/packages/lexical-markdown/src/utils.ts
+++ b/packages/lexical-markdown/src/utils.ts
@@ -461,10 +461,6 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
   );
 }
 
-// URL validation patterns
-export const EMAIL_REGEX =
-  /(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))/;
-
 export const PHONE_NUMBER_REGEX = /^\+?[0-9\s()-]{5,}$/;
 
 /**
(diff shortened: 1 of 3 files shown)


packages/lexical-markdown/src/utils.ts
```diff
@@ -460,3 +460,43 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
...
+ * @returns Formatted URL with appropriate protocol
+ */
+export function formatUrl(url: string): string {
+  // Check if URL already has a protocol
+  if (url.match(/^[a-z]+:/i)) {
+    // URL already has a protocol, leave it as is
+    return url;
+  }
+  // Check if it's a relative path
+  else if (url.startsWith('/')) {
```
etrepum, line 483: Starting with a `'.'` or `'#'` is also a relative URL

packages/lexical-markdown/src/utils.ts
```diff
@@ -460,3 +460,43 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
       MARKDOWN_EMPTY_LINE_REG_EXP.test(firstChild.getTextContent()))
   );
 }
+
+// URL validation patterns
+export const EMAIL_REGEX =
+  /(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))/;
```
etrepum, line 466: I don't think there's any reason to try this hard when you finding an `@` is probably just as good or better as a heuristic for guessing the mailto protocol. Seems weird to explicitly only support ipv4 and ascii domain names in 2025

packages/lexical-markdown/src/utils.ts
```diff
@@ -460,3 +460,43 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
...
+
+/**
+ * Formats a URL string by adding appropriate protocol if missing
+ *
+ * @param url - URL to format
+ * @returns Formatted URL with appropriate protocol
+ */
+export function formatUrl(url: string): string {
+  // Check if URL already has a protocol
+  if (url.match(/^[a-z]+:/i)) {
```
etrepum, line 478: This doesn't quite match the URI spec https://datatracker.ietf.org/doc/html/rfc3986#section-3.1 (although a browser is not often going to do something useful with something like `s3://` or `ms-excel:` they are still valid)

```suggestion
  if (url.match(/^[a-z][a-z0-9+.-]*:/i)) {
```

etrepum (commented):
Looks like a good start but I think the heuristics could be refined a bit further for spec compliance

AashishRichhariya:
@etrepum Thanks for the feedback. I've incorporated the changes in latest commit.

packages/lexical-markdown/src/utils.ts
```diff
@@ -460,3 +460,46 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
...
+ * @returns Formatted URL with appropriate protocol
+ */
+export function formatUrl(url: string): string {
+  // Check if URL already has a protocol
+  if (url.match(/^[a-z][a-z0-9+.-]*:/i)) {
+    // URL already has a protocol, leave it as is
+    return url;
+  }
+  // Check if it's a relative path (starting with '/', '.', or '#')
+  else if (url.startsWith('/') || url.startsWith('.') || url.startsWith('#')) {
```
etrepum, line 479:
```suggestion
  else if (url.match(/^[/#.]/)) {
```

packages/lexical-markdown/src/utils.ts
```diff
@@ -460,3 +460,46 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
...
+
+  // Check for email address - ensure it has exactly one '@' with non-empty username and domain portions
+  else if (
+    url.includes('@') &&
+    url.split('@').length === 2 &&
+    url.split('@')[0].length > 0 &&
+    url.split('@')[1].length > 0
+  ) {
+    return `mailto:${url}`;
+  }
```
etrepum, line 492: It's inefficient to do it this way since the split is computed three times (and the first split is also equivalent to the includes). There are also esoteric cases where you *could* have more than one @ in the local part of an address if the local part is quoted. I think it's better to leave it as simply `url.includes('@')` or maybe something like `url.match(/^(?:".+"|[^@]+)@[^@]+/)` and assume that if someone is writing a link they will specify a protocol if it's not supposed to be treated as an email address. It's not like we try and validate URIs that do have a scheme nor is there really something we can do if we consider the parse to be a warning or failure.

```suggestion
  else if (url.includes('@')) {
    return `mailto:${url}`;
  }
```

See also:
* https://en.wikipedia.org/wiki/Email_address#Examples
* https://datatracker.ietf.org/doc/html/rfc2822#section-3.4.1
* https://datatracker.ietf.org/doc/html/rfc2822#section-3.2.5

packages/lexical-markdown/src/utils.ts
```diff
@@ -460,3 +460,46 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
...
+
+  // Check for phone number
+  else if (PHONE_NUMBER_REGEX.test(url)) {
+    return `tel:${url}`;
+  }
+  // For everything else that looks like a URL (starting with www. or containing dots)
+  else if (/^www\.|(\.[a-z]{2,})/i.test(url)) {
```
etrepum, line 499: 0-9 and hyphens (not at the start or end) are also valid domain name characters (and then there are many many more possible code points with IDN). Maybe the best fallback would be to use ```return `https://${url}`;``` instead of `return url;` instead of using any heuristic here. I'm having a hard time thinking of a reasonable use case for a link href that doesn't match any of the above and isn't a https URI.

etrepum (commented):
I think this is still trying a bit too hard to make sure the heuristic is correct, at the expense of not allowing some valid things to match. Very close though!

packages/lexical-markdown/src/utils.ts
AashishRichhariya: Interesting, I didn't know we can have a quoted string in email. This does make this split invalid. 

`url.includes('@')` For checks as simple as these, it will also pass for these types of strings: `'@abcd'`, `'abcd@'` and `'@'`. 

Would this be an accepted behaviour?

packages/lexical-markdown/src/utils.ts
AashishRichhariya: This is much better. I will use this change.

packages/lexical-markdown/src/utils.ts
AashishRichhariya: Sounds good. I will remove the `else if` condifition, and return `https://${url}` for all fallback cases.

packages/lexical-markdown/src/utils.ts
etrepum: Yes, I think so. Most of those would be invalid as an https url too since @ cant be in the domain. This is just transforming the url into what is most likely to be useful, it’s not a strict validator.

AashishRichhariya:
I've made the suggested changes @etrepum

packages/lexical-markdown/src/utils.ts
etrepum: There are going to be false positives no matter which heuristic you use, but the user can always fix it by either specifying a scheme or starting the relative URL with one of the three symbols `[./#]` (`@` and `./@`as relative URLs point to the same resource). There are no relative links that are unrepresentable, and it doesn’t clutter the implementation with heuristics that are large and complicated (which is the only way to do it correctly) or have false negatives (other than relative URLs without that prefix).

I think the simple heuristic that I proposed will be easiest for users to understand and will have the expected behavior more often.

AashishRichhariya: Sounds good. I've made these changes in latest commit.

etrepum (approved):
Looks great!

packages/lexical-markdown/src/utils.ts
```diff
@@ -460,3 +460,37 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
...
+
+export const PHONE_NUMBER_REGEX = /^\+?[0-9\s()-]{5,}$/;
+
+/**
+ * Formats a URL string by adding appropriate protocol if missing
+ *
+ * @param url - URL to format
+ * @returns Formatted URL with appropriate protocol
+ */
+export function formatUrl(url: string): string {
```
potatowagon, line 472: this should be optional. enforcing this is breaking usecases that dynamically build urls

a middle ground is to have this behind a flag defaulted to false

aashishr-meelance: Can you help me understand those usecases? Perhaps with an example?

etrepum: I think we can probably sidestep adding an option and cover both use cases by moving this code to LinkNode. It would render and export the href using the formatter and preserve what was typed in the lexical document (and markdown export)

PR #7504, opened 2025-05-01
[lexical-list] Bug Fix: Inherit marker styles when indenting list items

### LINKED ISSUE #7491: Bug: missing --listitem-marker-font-size on first indent
Lexical version: 0.3.0

## Steps To Reproduce

Repro steps:
- create 3 paragraphs
- select all
- convert to unordered list (any type of list)
- Indent the second list item
- now the created PlaygroundEditorTheme__nestedListItem list item does NOT have `--listitem-marker-font-size` property

## Repro steps visually
Image

## The current behavior
`--listitem-marker-font-size` is missing from the nested list item

## The expected behavior
`--listitem-marker-font-size` should be on the nested list item

## Impact of fix
List styling which relies on this --listitem-marker-font-size

It does still look good in the Playground, as it is adding a fixed 32px left margin to list items
But in my case, I rely on the `--listitem-marker-font-size` as I set a dynamic left margin based on that css variable

kirandash:
### Current Behavior:
Currently, when a list item with marker styles (like `--listitem-marker-font-size`) is indented, the parent list item that gets created does not inherit these styles. This causes inconsistent marker styling when indenting list items.

### Changes in this PR:
1. Modified `$handleIndent` in `formatList.ts` to copy the text style from the list item being indented to the newly created parent list item
2. Added unit tests to verify the marker style inheritance behavior

Closes #7491

### Before

lexical-bug-7491

When a list item with custom marker styles is indented, the parent list item loses those styles:
```html
<li><!-- Parent loses the marker style -->
  <ul>
    <li style="--listitem-marker-font-size: 60px;">Child item</li>
  </ul>
</li>
```

### After

https://github.com/user-attachments/assets/2ce40bd5-4eb4-4807-a5be-252a0688d958

The parent list item now inherits the marker styles when a child is indented:
```html
<li style="--listitem-marker-font-size: 60px;">
  <ul>
    <li style="--listitem-marker-font-size: 60px;">Child item</li>
  </ul>
</li>
```

### DIFF
--- a/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
+++ b/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
@@ -25,6 +25,7 @@ import {
   ListItemNode,
   ListNode,
 } from '../..';
+import {$handleIndent} from '../../formatList';
 
 const editorConfig = Object.freeze({
   namespace: '',
@@ -1512,6 +1513,33 @@ describe('LexicalListItemNode tests', () => {
             </div>
           `,
         );
+      });
+    });
+
+    test('ListItemNode marker style inheritance on indent', async () => {
+      const {editor} = testEnv;
+
+      await editor.update(() => {
+        const root = $getRoot();
+        const listNode = $createListNode('bullet');
+        const listItem1 = $createListItemNode();
+        const listItem2 = $createListItemNode();
+
+        // Set marker style on listItem2
+        listItem2.__textStyle = 'font-size: 19px;';
+
+        listNode.append(listItem1, listItem2);
+        root.append(listNode);
+
+        // Indent listItem2
+        $handleIndent(listItem2);
+
+        // Get the parent list item after indent
+        const parentListItem = listItem2.getParentOrThrow().getParentOrThrow();
+        expect($isListItemNode(parentListItem)).toBe(true);
+
+        // Check if marker style was inherited
+        expect(parentListItem.__textStyle).toBe('font-size: 19px;');
       });
     });
   });
--- a/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
+++ b/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
@@ -1526,7 +1526,7 @@ describe('LexicalListItemNode tests', () => {
         const listItem2 = $createListItemNode();
 
         // Set marker style on listItem2
-        listItem2.__textStyle = 'font-size: 19px;';
+        listItem2.setTextStyle('font-size: 19px;');
 
         listNode.append(listItem1, listItem2);
         root.append(listNode);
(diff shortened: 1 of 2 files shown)


packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
```diff
@@ -1514,5 +1515,32 @@ describe('LexicalListItemNode tests', () => {
...
+      const {editor} = testEnv;
+
+      await editor.update(() => {
+        const root = $getRoot();
+        const listNode = $createListNode('bullet');
+        const listItem1 = $createListItemNode();
+        const listItem2 = $createListItemNode();
+
+        // Set marker style on listItem2
+        listItem2.__textStyle = 'font-size: 19px;';
```
etrepum, line 1529:
```suggestion
        listItem2.setTextStyle('font-size: 19px;');
```

packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
```diff
@@ -1514,5 +1515,32 @@ describe('LexicalListItemNode tests', () => {
...
+
+        // Indent listItem2
+        $handleIndent(listItem2);
+
+        // Get the parent list item after indent
+        const parentListItem = listItem2.getParentOrThrow().getParentOrThrow();
+        expect($isListItemNode(parentListItem)).toBe(true);
+
+        // Check if marker style was inherited
+        expect(parentListItem.__textStyle).toBe('font-size: 19px;');
```
etrepum, line 1542:
```suggestion
        expect(parentListItem.getTextStyle()).toBe('font-size: 19px;');
```

etrepum (commented):
This should be using the public accessors, there's no need to drop down to the implementation of the property here

packages/lexical-list/src/formatList.ts
```diff
@@ -400,6 +400,8 @@ export function $handleIndent(listItemNode: ListItemNode): void {
       const newListItem = $createListItemNode()
         .setTextFormat(parent.getTextFormat())
         .setTextStyle(parent.getTextStyle());
+      // Copy the text style from the list item being indented to preserve marker styles
+      newListItem.setTextStyle(listItemNode.getTextStyle());
```
etrepum, line 404: Looking more closely I think the fix can be even simpler, the text style is just coming from the wrong node. No need to set it twice. It does make me think that both format and style should probably come from the same node though

```suggestion
        .setTextFormat(listItemNode.getTextFormat())
        .setTextStyle(listItemNode.getTextStyle());
```

kirandash: Thanks for catching that @etrepum! I've [simplified the code](https://github.com/facebook/lexical/pull/7504/commits/d3f20dc10d70fa186a1afafeafb8bb96a561fe86) to:
1. Get both format and style from the listItemNode being indented
2. Remove the redundant style setting
3. Make the format and style handling consistent

Thanks for the thorough review. Please let me know if this looks better!

etrepum:
I don't know why github is failing to start CI, it doesn't appear that restarting the jobs has any effect, and I don't have permissions on this fork to push another commit. The last commit passed, and both test-unit and ci-check pass locally so I'm just going to merge it as it's a very isolated and straightforward fix.

etrepum:
The merge queue seems to be blocked on those jobs to start as well, so I think this might need an empty commit or something to trigger CI so it can be merged.

etrepum (approved):

etrepum (approved):

etrepum:
I'm going to try closing and re-opening the PR to see if that resets the stuck CI

PR #7435, opened 2025-04-06
[lexical-playground] Bug Fix: Fix typeahead menu positioning during scroll

kirandash:
### Current behavior:
- The typeahead menus (slash menu, emoji picker, mentions, etc.) disappear when scrolling down in the editor
- This occurs because the menu uses `position: fixed` without proper positioning values
- The menu stays in its initial position relative to the viewport instead of moving with the cursor

### Changes being added:
- Changed typeahead popover positioning from `fixed` to `absolute`
- Added positioning context container with `position: relative`
- Updated all typeahead menu implementations (ComponentPicker, EmojiPicker, Mentions, AutoEmbed, ContextMenu) to use the new container

Closes #7430

### Before
The slash menu and other typeahead menus disappear when scrolling down:

https://github.com/user-attachments/assets/2a6acc69-38db-4a84-bc74-ef80887a1b4a

### After
The menus stay properly positioned relative to their trigger points during scrolling:

https://github.com/user-attachments/assets/25dfd844-4e9f-418c-92b3-f0e7885ef512

Tested the following scenarios:
1. ComponentPickerPlugin
   - Type "/" at various scroll positions
   - Menu stays visible and properly positioned
2. Emoji picker
   - Type ":" at various scroll positions
   - Menu stays visible and follows cursor
3. Mentions plugin
   - Type "@" at various scroll positions
   - Menu stays visible and follows cursor
4. AutoEmbed plugin
   - Enter URL at various scroll positions
   - Menu stays visible
5. Context menu
   - Right-click on text at various scroll positions
   - Menu appears at correct position

PR #7506, opened 2025-05-03
[lexical-playground] Bug Fix: LexicalTypeaheadMenuPlugin Positioning When Scrolled

### LINKED ISSUE #7430: Bug: [lexical-react] [lexical-playground] slash menu disappears scrolling down
Without scrolling down:
Image

When you scroll down:
Image

Lexical version: LTS

## Steps To Reproduce

Go onto the playground, hit enter a few dozen times and scroll down, then type out a `/`, and the `<LexicalTypeaheadMenuPlugin />` isn't visible.

Link to code example:

## The current behavior
You cannot see the slash command menu when scrolling down.

## The expected behavior
You should see the slash command menu when scrolling down.

### LINKED ISSUE #7459: Bug: LexicalTypeaheadMenuPlugin incorrect position when scrolled
Hello, i'm struggling finding a solution for this behavior, where the anchor position is calculated but is incorrect if there is a scroll

(i cant upload video here, here is a screenshot, i provided the repro steps)
Image

Lexical version: latest

## Steps To Reproduce

1. Go to public lexical playground
2. Add enough lines so a scroll appear
3. Scroll a little
4. Open a LexicalTypeaheadMenuPlugin shortcut (ie: type /)
5. The menu is not in the correct position

Link to code example:
Not applicable, reproductible on lexical-playground
I tested adding the 'parent' props, it change a bit the behavior but is not fixing

## The current behavior
The menu is far from the text where the cursor is

## The expected behavior
The menu should be at the same position that if there is no scroll

## Impact of fix
How severe is this bug? 
On little screens, it is very impacting as the need to scroll is very usual, but on long text, it is problematic even on big screens

Who would benefit from this fix?
In theory, any project that need a typeahead (eg: for mentioning users / things) and that do this in scrollable text would benefit to have this bug fixed

AashishRichhariya:
## Description
Currently, the LexicalTypeaheadMenuPlugin displays menus in incorrect positions when the editor content is scrolled. This creates a misalignment between where the user is typing and where the menu appears, with the discrepancy increasing as the user scrolls more.

This PR:
1. Changes the CSS positioning strategy for the typeahead popover from `fixed` to `relative`
2. Maintains all other styling properties to ensure visual consistency

The fix improves user experience by ensuring that the typeahead menu always appears at the cursor's position regardless of scroll state, making the editor more usable on:
* Screens of all sizes, especially smaller ones where scrolling is inevitable
* Documents with longer content that require scrolling
* Any implementation using the typeahead menu for mentions, commands, or other features

Closes #7459
Closes #7430

### Before
When scrolling down in the editor and triggering the typeahead menu (by typing `/`), the menu appears far away from the cursor position.

https://github.com/user-attachments/assets/16f2dad3-2ab2-4977-8a1e-bca46c705221

### After
When scrolling down in the editor and triggering the typeahead menu, the menu appears directly at the cursor position regardless of scroll state.

https://github.com/user-attachments/assets/7608b32d-c019-4841-a1bf-2247da9848b4


kirandash:
I've also been working on this issue in PR #7435, which takes a similar approach of changing the positioning from `fixed` to `relative`. However, I closed my PR after @ivailop7 mentioned they're working on a more general solution in #7409.

This PR looks like it might fix the immediate issue. However, it might be worth waiting for changes from @ivailop7.

etrepum:
I think the trade-off here is that some of these fixes may conflict with whatever @ivailop7 is working on. However, if the fix is really just one line of CSS I think the benefits of merging prior to the general fix would outweigh the costs of resolving a one line merge conflict.

AashishRichhariya:
@kirandash @etrepum @ivailop7 I think since this is a very simple 1 line css fix, we should merge this change for immediate fix while @ivailop7 is working on much more comprehensive solution, which is understandablt taking time. It will improve the user experience in short term, while the general fix is on the way. 

Curious for you thoughts.

etrepum (approved):

Issue #7514, opened 2025-05-08
Bug: Deleting the last character of a link inside brackets reduces indentation level


ArchitW:
Lexical version: 0.31.0

## Steps To Reproduce

1. Create a nested indentation via bullets
2. Write text inside brackets: "[link]" on the nested line
3. Select the text, create a link from it to any URL
4. Place the cursor on the last letter of the text, before the right bracket
5. Hit delete

Link to code example:

## The current behavior
Instead of deleting the character, reduce the indentation
## The expected behavior
Indentation should not get affected

PR #7515, opened 2025-05-08
[lexical-playground] [lexical-richtext] BugFix Deleting Last Character of Link Inside Brackets Reduces Indentation Level

ArchitW:
## Description

This PR addresses an edge case where deleting a character at the end of a link text incorrectly removes the indent. To resolve this issue, we've added a check to verify if the node preceding the cursor is a link. If it is, we skip the OUTDENT_CONTENT_COMMAND, ensuring that the indent remains intact.

Closes https://github.com/facebook/lexical/issues/7514

### Before

https://github.com/user-attachments/assets/442f0bc9-77f2-4db0-a78e-94482d36a6a7

### After

https://github.com/user-attachments/assets/4891d477-f2ff-4c7c-8439-2a0cf22671b2

### DIFF
--- a/packages/lexical-rich-text/src/index.ts
+++ b/packages/lexical-rich-text/src/index.ts
@@ -876,6 +876,10 @@ export function registerRichText(editor: LexicalEditor): () => void {
             anchor.offset === 0 &&
             !$isRootNode(anchorNode)
           ) {
+            const prevNode = anchorNode.getPreviousSibling();
+            if (prevNode && prevNode.__type === 'link') {
+              return false;
+            }
             const element = $getNearestBlockElementAncestorOrThrow(anchorNode);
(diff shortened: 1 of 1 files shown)


ivailop7:
I think the issue might be related to a command from the markdown plugin, because this only happens if the link is inside square brackets and has indentation. I don't think the fix for this lies in the rich-text package.

etrepum:
The brackets are a red herring here, based on some experimentation I think it happens any time you press backspace at the beginning of a TextNode that is in a ListItemNode. You don't need a LinkNode at all, e.g. it could be `(decorator)(text)` and if you press backspace at the beginning of the text it will dedent instead of deleting the decorator. Either way, this is the wrong fix, but not for the reasons that @ivailop7 suspected.

PR #7518, opened 2025-05-09
[lexical-playground] Bug Fix: Improve keyboard accessibility for toolbar formatting commands

### LINKED ISSUE #6006: better support for keyboard accessibility
In general using "Enter" on a toolbar that makes changes to lexical is problematic.

Lets say I have a button that toggles "bold" and it's triggered using the "Enter" key on the mouse-down event which calls dispatchCommand(FORMAT_TEXT_COMMAND...) - this causes lexical to get focus and then handle the Enter key which can replace the selected text with a new line.

adding the e.preventDefault to the event solves this problem but the focus still moves from the toolbar to the editor.

I'm not sure how to handle this...

simon0x1800:
## Description

Current behavior:
- When using Enter key on toolbar buttons (e.g., bold), it triggers FORMAT_TEXT_COMMAND
- This causes the editor to get focus and creates an unwanted new line
- Even with e.preventDefault(), focus still moves from toolbar to editor
- This creates poor keyboard accessibility experience

Changes in this PR:
- Added SKIP_DOM_SELECTION_TAG to prevent editor from stealing focus when dispatching commands
- Wrapped formatting commands in editor.update() with SKIP_DOM_SELECTION_TAG
- Applied to all formatting buttons (bold, italic, underline, strikethrough, code)
- Maintained proper keyboard event handling (preventDefault and stopPropagation)

This solution aligns with Lexical's documentation which states that SKIP_DOM_SELECTION_TAG prevents the editor from stealing focus when dispatching commands.

Closes #6006

### Before
When using keyboard navigation (Tab + Enter) on toolbar formatting buttons:
- Button triggers formatting but loses focus
- Editor gains focus
- Unwanted newline is created
- Poor keyboard accessibility experience

### After
When using keyboard navigation (Tab + Enter) on toolbar formatting buttons:
- Button triggers formatting and maintains focus
- Editor does not steal focus
- No unwanted newline is created
- Proper keyboard accessibility maintained
- Screen reader users can navigate and use toolbar buttons effectively


packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -6,7 +6,12 @@
  *
  */
 
-import type {JSX} from 'react';
+/** @jsxRuntime classic */
```
etrepum, line 9: I think you added some extra changes in here that don’t belong in this PR

etrepum:
Closing as stale as this PR includes unrelated changes, does not build correctly, and no updates have happened in weeks. If you intend to fix it, you may re-open.

PR #7520, opened 2025-05-09
[lexical-react][lexical-playground] Bug Fix: Make typeahead punctuation configurable; allow underscores and dashes in emoji queries 

### LINKED ISSUE #7519: Bug:  If you set query including underscore(_) or dash(-) when using `LexicalTypeaheadMenuPlugin`, onQueryChange doesn't pass query param.
Lexical version: 0.30.0

## Steps To Reproduce

1. enter text start with colon to open menu
2. add _ to match pre-defined emoji prefix in [lexical playground](https://playground.lexical.dev/)

Link to code example: lexical guide code 👉 https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/EmojiPickerPlugin/index.tsx

## The current behavior

https://github.com/user-attachments/assets/25c5e9c2-65ee-4050-b822-c0b205905afd

## The expected behavior

query is passed as param and show menu including 'man_...'

kirandash:
### Current behavior:
The typeahead menu does not recognize queries containing underscores (`_`) or dashes (`-`). For example, typing `:man_beard` or `:t-rex` does not trigger the typeahead menu as expected.

### Change introduced:  
This PR makes the punctuation set for typeahead queries configurable via an optional parameter in `useBasicTypeaheadTriggerMatch`.

- The default punctuation set is restored for backwards compatibility.
- The emoji plugin now passes a custom punctuation set, allowing underscores and dashes in emoji queries.

This enables the typeahead menu to correctly match and display suggestions for queries like `:man_beard` and `:t-rex` in the emoji picker, without affecting other typeahead usages.

Closes #7519

### Before

https://github.com/user-attachments/assets/69979fde-8946-4aa0-b529-91000429490e

### After

image

image

### DIFF
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -35,7 +35,7 @@ import {startTransition} from 'shared/reactPatches';
 import {LexicalMenu, MenuOption, useMenuAnchorRef} from './shared/LexicalMenu';
 
 export const PUNCTUATION =
-  '\\.,\\+\\*\\?\\$\\@\\|#{}\\(\\)\\^\\-\\[\\]\\\\/!%\'"~=<>_:;';
+  '\\.,\\+\\*\\?\\$\\@\\|#{}\\(\\)\\^\\[\\]\\\\/!%\'"~=<>:;';
 
 function getTextUpToAnchor(selection: RangeSelection): string | null {
   const anchor = selection.anchor;
--- a/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
@@ -35,7 +35,7 @@ import {startTransition} from 'shared/reactPatches';
 import {LexicalMenu, MenuOption, useMenuAnchorRef} from './shared/LexicalMenu';
 
 export const PUNCTUATION =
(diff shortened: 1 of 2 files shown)


packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -35,7 +35,7 @@ import {startTransition} from 'shared/reactPatches';
 import {LexicalMenu, MenuOption, useMenuAnchorRef} from './shared/LexicalMenu';
 
 export const PUNCTUATION =
-  '\\.,\\+\\*\\?\\$\\@\\|#{}\\(\\)\\^\\-\\[\\]\\\\/!%\'"~=<>_:;';
+  '\\.,\\+\\*\\?\\$\\@\\|#{}\\(\\)\\^\\[\\]\\\\/!%\'"~=<>:;';
```
etrepum, line 38: I wonder if it's worth breaking backwards compatibility for useBasicTypeaheadTriggerMatch in this way, maybe instead the punctuation could be optionally overridden and the emoji plugin could provide its own set

kirandash: Thank you for the suggestion!
I've [updated the implementation](https://github.com/facebook/lexical/pull/7520/commits/72893765d76c27c3b3335f626e5db5e2905ded7b) so that the punctuation set is now configurable via an optional parameter to `useBasicTypeaheadTriggerMatch`. 
The original `PUNCTUATION` constant is restored for backwards compatibility.
The emoji plugin now passes its own punctuation set to allow underscores and dashes.

Let me know if you have any further feedback!

etrepum (approved):

PR #7545, opened 2025-05-18
[lexical][lexical-playground] Bug Fix: Support Apple Pencil

### LINKED ISSUE #7542: Bug:
Lexical version: v0.31.1

## Steps To Reproduce

1. Go to the playground on Safari using iPad with iPadOs 18.5 and an Apple Pencil
2. Tap on the screen somewhere in the editor with the Apple Pencil to move the selection
3. Using the Apple Pencil open the Insert dropdown, select Image, then insert the Sample image
4. Notice that no sample image was inserted
5. Tap with your finger on the somewhere in the document and perform the same steps
6. Notice that a Sample image was still not inserted
7. Refresh the page and perform the sample image insert using your finger
8. Notice that the image inserts fine if you avoid touching the editor with your Apple Pencil

## The current behavior

https://github.com/user-attachments/assets/26917363-3332-4db4-a1b7-906cd8c40ff4

## The expected behavior

I expect to be able to insert images using an Apple Pencil

## Impact of fix

This bug prevents anyone from interacting with Lexical using an Apple Pencil

circlingthesun:
## Description

Currently Apple Pencil inputs effectively breaks the editor because pen inputs do not update the selection state. This PR ensures pen input events update the selection state in the same way as touch input events.

Apple pencil inputs also create keyboard events when making scribbles. The key property for these events are undefined and broke the shortcuts plugin. This PR also fixes that.

Closes #7542

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -514,8 +514,8 @@ function onClick(event: PointerEvent, editor: LexicalEditor): void {
             }
           }
         }
-      } else if (event.pointerType === 'touch') {
-        // This is used to update the selection on touch devices when the user clicks on text after a
+      } else if (event.pointerType === 'touch' || event.pointerType === 'pen') {
+        // This is used to update the selection on touch devices (including Apple Pencil) when the user clicks on text after a
         // node selection. See isSelectionChangeFromMouseDown for the inverse
         const domAnchorNode = domSelection.anchorNode;
         // If the user is attempting to click selection back onto text, then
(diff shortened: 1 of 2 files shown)


etrepum (dismissed):
I can't test this but I think it looks good, assuming the extended test suite still passes!

etrepum (dismissed):

packages/lexical/src/LexicalEvents.ts
```diff
@@ -542,7 +542,7 @@ function onPointerDown(event: PointerEvent, editor: LexicalEditor) {
   // TODO implement text drag & drop
   const target = event.target;
   const pointerType = event.pointerType;
-  if (isDOMNode(target) && pointerType !== 'touch' && event.button === 0) {
+  if (isDOMNode(target) && pointerType !== 'touch' && pointerType !== 'pen' && event.button === 0) {
```
etrepum, line 545: This is needed to pass the formatting checks

```suggestion
  if (
    isDOMNode(target) &&
    pointerType !== 'touch' &&
    pointerType !== 'pen' &&
    event.button === 0
  ) {
```

ivailop7:
> I can't test this but I think it looks good, assuming the extended test suite still passes!

I can test this, will confirm in the coming days.

ivailop7 (approved):
Verified it works with an Apple Pencil on an iPad.

PR #7558, opened 2025-05-21
[lexical-playground][lexical-list] Bug Fix: Made checklist icon fully scalable, clickable, and properly spaced at large font sizes

### LINKED ISSUE #7554: Bug:
Lexical version: 0.31.2

## Steps To Reproduce

1. Add a checklist
2. increase the font size

## The current behavior:
On increasing the font size the size of the checkbox remains unchanged.
Image

## The expected behavior:
On increasing the font size the size of the checkbox also matches the font

Image

## Impact of fix
This fixes the overall checklist UX

swrsaini:
**Current behavior:**

- The checklist icon (checkbox) does not scale properly with large font sizes.
- The clickable area of the checkbox does not cover the entire icon at larger sizes, making it hard to toggle.
- The checkmark inside the icon remains small and is not visually balanced.
- The margin between the icon and the checklist item text is inconsistent at larger sizes.

**Changes introduced by this PR:**

- Ensured the checklist icon and its checkmark scale proportionally with the font size of the checklist item.
- Made the entire icon area fully clickable, regardless of size.
- Improved the visual alignment and thickness of the checkmark for better visibility.
- Fixed the margin between the icon and the item text for consistent spacing at all sizes.

Closes #7554 

**Before**

- Checklist icon and checkmark did not scale with large font sizes.
- Only part of the icon was clickable at large sizes.
- Margin between icon and text was inconsistent.
Screenshot 2025-05-22 040524

**After**

- Checklist icon and checkmark scale smoothly with font size.
- The entire icon is clickable, even at large sizes.
- Margin between icon and text is consistent and visually balanced.

https://github.com/user-attachments/assets/31c26900-4971-4b57-995a-dda7ed795685


etrepum (commented):
The CSS changes here make the text caret invisible when there is no text in the list item (at least in Chrome 136.0.7103.114 macOS). Probably the flex? If you turn off flex the cursor is there but in the wrong place, because without a span element there's no margin to move it away from the checkbox.

swrsaini:
Made some changes, i have added a <span> with zero-width space for caret in checklist items to fix that issue

etrepum:
Tests are failing, and I don't think it's appropriate or necessary to change the markup to make this work correctly.

swrsaini:
I removed the span tag that was causing DOM manipulation and replaced it with a zero-width space to handle the caret position. All test cases have passed successfully.

etrepum (commented):
Still not seeing a cursor in Chrome or Safari until text is entered.
The cursor is not aligned correctly in Firefox until text is entered.

Screenshot from firefox:
Screenshot 2025-05-26 at 20 31 29

swrsaini:
@etrepum  I’ve added a single space character as a text node to ensure the caret stays properly aligned and visible across all browsers. I’ve tested it on Chrome and Firefox, and it looks good. Could you please double-check on your end?

swrsaini:
I don’t think this can be fixed without changing the markup. The main issue is that browsers especially Firefox don’t reliably render the caret in empty contenteditable elements when you're using only CSS tricks like pseudo-elements, flexbox, or zero-width spaces. If the element looks filled but is still empty in the DOM, the caret often disappears or ends up in the wrong place. Is there any other solution i can go with?

etrepum:
It works without the changes in this PR, I think the problem is specifically the use of flex

swrsaini:
Yes flex was causing this issue.

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -95,42 +95,54 @@ export class ListItemNode extends ElementNode {
...
+    // Set custom property for font-size from textStyle if present
+    const textStyleObj = getStyleObjectFromCSS(this.__textStyle);
+    if (textStyleObj['font-size']) {
+      dom.style.setProperty(
+        '--listitem-marker-font-size',
+        textStyleObj['font-size'],
+      );
+    } else {
+      dom.style.removeProperty('--listitem-marker-font-size');
+    }
```
etrepum, line 127:
```suggestion
```

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -95,42 +95,54 @@ export class ListItemNode extends ElementNode {
...
     if (prevStyle !== nextStyle) {
       if (nextStyle === '') {
         dom.removeAttribute('style');
       } else {
-        dom.style.cssText = nextStyle;
+        dom.style.cssText += nextStyle;
```
etrepum, line 132:
```suggestion
        dom.style.cssText = nextStyle;
```

etrepum (commented):
It doesn't look like this code was removed, at least not in what's on github right now

PR #7564, opened 2025-05-23
[lexical-markdown] Bug fix: Prevent transform from removing nodes if the replace function returns false

### LINKED ISSUE #7563: Bug: Markdown Shortcuts Transformer wrongly removes the matched node before the replace function runs
When using the markdown shortcuts plugin with custom transformers, if a transformer's `replace` function returns `false` to cancel the transformation, the leading node of the matching node is still removed from the document. This happens because the `leadingNode.remove()` is called before checking the return value of the `replace` function in this line https://github.com/facebook/lexical/blob/b61a1ea1c26449b4c7ce0e5e6667ee0d753043aa/packages/lexical-markdown/src/MarkdownShortcuts.ts#L73

This seems unintended since it doesn't respect the documentation of the replace function: `@return return false to cancel the transform, even though the regex matched. Lexical will then search for the next transformer.`

Fixing this would enable writing transformers that do not replace the content even if they match, under certain conditions. For example preventing turning headers into numbered lists if you type `1. Header`, like Notion and other popular editor do.

Lexical version: v0.31.2

## Steps To Reproduce

1. Create a markdown transformer that returns `false` from its `replace` function
2. Type a markdown shortcut that would trigger the transformer
3. The document is modified despite the transformation being cancelled

## The current behaviour

The leading node gets remove even if the replace is cancelled.

## The expected behavior

Returning `false` from the replace function should cancel the transform and not remove the nodes.

## Impact of fix

This issue prevents applying markdown transformers conditionally. For example, if you don't want to transform headings into numbered lists which is a common use case. 

Let me know what you think. Happy to put up a PR.

rodrigomotion:
## Description
Currently if you return `false` from a Markdown transformer replace function, the expectation is that the transform will be cancelled, however the matched nodes are still being modified (removed) because the leading node gets removed before checking the return value of the replace function. This prevents implementing some functionality, such as not turning headers into numbered lists, and could cause other issues since when the next transform runs it won't have the leading node anymore.

I tried to keep it simple and just moved the `remove` function so that it runs only if the replace function does not return false. All tests pass.

Closes #7563

## Test plan

Ran all the unit and e2e tests and they all pass.

### Before and after video

https://share.cleanshot.com/JYWVlb4d

### DIFF
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -223,7 +223,7 @@ const CODE_TAG_COUNTER_EXAMPLE: MultilineElementTransformer = {
   type: 'multiline-element',
 };
 
-export const CENCELED_HEADING_REPLACE_EXAMPLE: ElementTransformer = {
+export const CANCELED_HEADING_REPLACE_EXAMPLE: ElementTransformer = {
   dependencies: [HeadingNode],
   export: () => {
     return null;
@@ -768,7 +768,7 @@ describe('Markdown', () => {
       ],
     });
 
-    registerMarkdownShortcuts(editor, [CENCELED_HEADING_REPLACE_EXAMPLE]);
+    registerMarkdownShortcuts(editor, [CANCELED_HEADING_REPLACE_EXAMPLE]);
 
     editor.update(
       () => {


rodrigomotion:
Seems like one of the test suites is timing out but it's unrelated to this change?

etrepum:
Likely, I will restart it. Do you plan to write some additional tests or make any other changes before review? It is still marked WIP.

rodrigomotion:
> Likely, I will restart it. Do you plan to write some additional tests or make any other changes before review? It is still marked WIP.

I was trying to write a test, but I'm not sure how to set this up without changing some of the transformers, so I won't unless someone thinks it's required. If the current tests pass I'll leave it as is.

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -213,6 +223,18 @@ const CODE_TAG_COUNTER_EXAMPLE: MultilineElementTransformer = {
   type: 'multiline-element',
 };
 
+export const CENCELED_HEADING_REPLACE_EXAMPLE: ElementTransformer = {
```
etrepum, line 226:
```suggestion
export const CANCELED_HEADING_REPLACE_EXAMPLE: ElementTransformer = {
```

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -734,6 +756,99 @@ describe('Markdown', () => {
...
+        HeadingNode,
+        ListNode,
+        ListItemNode,
+        QuoteNode,
+        CodeNode,
+        LinkNode,
+      ],
+    });
+
+    registerMarkdownShortcuts(editor, [CENCELED_HEADING_REPLACE_EXAMPLE]);
```
etrepum, line 771:
```suggestion
    registerMarkdownShortcuts(editor, [CANCELED_HEADING_REPLACE_EXAMPLE]);
```

etrepum (approved):
Tentatively looks good, other than this typo in the tests. Will wait to see the result of the full test run.

darikprescott:
Shut this shit down.    I'm deleting it all

Yahoo Mail: Search, Organize, Conquer 
 
  On Sun, May 25, 2025 at 11:59 AM, ***@***.***> wrote:   



Merged #7564 into main.

PR #7568, opened 2025-05-26
[lexical-selection][lexical-playground] Bug Fix: unbulleting an image doesn't work #5698

### LINKED ISSUE #5698: Bug: Bulleting an image when image is selected crashes

kamilpitula:
## Description

Bug Fix: unbulleting an image doesn't work

Closes #5698

### Steps to reproduce
1. Add an image
2. Select the image
3. Change formatting to bullet list
4. Change formatting back to normal

### Before

The image is still an element of a bullet list, also notice that when you change formatting when the image is selected the dropdown is not updated.

### After

The image is unbulleted correctly, formatting dropdown is updated accordingly.

https://github.com/user-attachments/assets/6e987e46-b940-4598-b30d-d9a48c004de6

### DIFF
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -16,7 +16,7 @@ import type {
   TextNode,
 } from 'lexical';
 
-import { TableSelection } from '@lexical/table';
+import {TableSelection} from '@lexical/table';
 import {
   $caretFromPoint,
   $createRangeSelection,
@@ -39,7 +39,7 @@ import {
 } from 'lexical';
 import invariant from 'shared/invariant';
 
-import { getStyleObjectFromCSS } from './utils';
+import {getStyleObjectFromCSS} from './utils';
 
 export function $copyBlockFormatIndent(
   srcNode: ElementNode,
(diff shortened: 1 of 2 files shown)


etrepum (commented):
This PR is not formatted with the same rules as the monorepo, `npm run prettier:fix` should address that

packages/lexical-selection/src/range-selection.ts
```diff
@@ -94,6 +94,11 @@ export function $setBlocksType<T extends ElementNode>(
   for (const node of selection.getNodes()) {
     if ($isElementNode(node) && INTERNAL_$isBlock(node)) {
       blockMap.set(node.getKey(), node);
+    } else {
```
etrepum, line 97:
```suggestion
    } else if (anchorAndFocus === null) {
```

This shouldn't be necessary for RangeSelection due to the anchorAndFocus code above (all containing blocks are either fully inside the RangeSelection or ancestors of the anchor or focus)

etrepum (approved):

PR #7573, opened 2025-05-27
[lexical-code] Feature: Improve Tab/Indent/Outdent to match VSCode

### LINKED ISSUE #7561: Review of CodeBlock Indent/Outdent/Tab features
After https://github.com/facebook/lexical/issues/7541, it was realized that there are some corner cases of  Indent/Outdent/Tab that are not correctly handled in Lexical.

it was decided to try and mimic the VSCode features as people probably expect the same behavior from a web based code editor.

This issue will be used to document the way VSCode handles Indent/Outdent/Tab features and compare it to the current status of Lexical. It will then be the base for adding new tests and modify the behavior to match these tests.

jeromew:
## Description

This is a Feature/Bugfix PR that gives the CodeBlock Tab/Indent/Outdent mechanism the same behavior as VSCode.

It was observed that the Lexical Tab/Indent/Outdent strategy had corner case peculiarities that can be pain points depending on the habits of the user.

After #7541 fixed some bugs, it was noted that Lexical was still not on par with VSCode's behavior and that it would make sense to match this behavior as it can nowadays be considered as "expected behavior". Issue #7561 was created to document the observed behaviors in VSCode.

The PR adds a test suite in `packages/lexical-code/src/__tests__/unit/LexicalCodeNodeTabs.test.ts` that makes it easy to declare new scenarios. It adds 37 unit tests in the Tab/Indent/Outdent domain.

The code modifications in this PR passes all 37 tests while the current HEAD fails on 15 tests.

Closes #7561

passes
`npm run test-unit` and `npm run test-e2e-chromium`

### DIFF
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -22,13 +22,18 @@ import {
   $createPoint,
   $createTabNode,
   $createTextNode,
+  $getCaretRange,
   $getNodeByKey,
   $getSelection,
+  $getSiblingCaret,
+  $getTextPointCaret,
   $insertNodes,
   $isLineBreakNode,
   $isRangeSelection,
   $isTabNode,
   $isTextNode,
+  $normalizeCaret,
+  $setSelectionFromCaretRange,
   COMMAND_PRIORITY_LOW,
   INDENT_CONTENT_COMMAND,
   INSERT_TAB_COMMAND,
@@ -648,9 +653,12 @@ function $handleMultilineIndent(type: LexicalCommand<void>): boolean {
     const tabNode = $createTabNode();
     const lineBreakNode = $createLineBreakNode();
     selection.insertNodes([tabNode, lineBreakNode]);
-    const parent = tabNode.getParentOrThrow();
-    const index = tabNode.getIndexWithinParent();
-    parent.select(index, index + 2);
+    $setSelectionFromCaretRange(
+      $getCaretRange(
+        $getTextPointCaret(tabNode, 'next', 0),
+        $normalizeCaret($getSiblingCaret(lineBreakNode, 'next')),
+      ),
+    );
     return true;
   }
 
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -23,6 +23,7 @@ import {
   $createTabNode,
   $createTextNode,
   $getCaretRange,
+  $getCaretRangeInDirection,
   $getNodeByKey,
   $getSelection,
   $getSiblingCaret,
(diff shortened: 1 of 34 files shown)


packages/lexical-code/src/CodeHighlighter.ts
```diff
@@ -593,28 +629,60 @@ function $handleMultilineIndent(type: LexicalCommand<void>): boolean {
...
+    codeLinesLength === 0 &&
+    type === INDENT_CONTENT_COMMAND &&
+    selection.getTextContent() === '\n'
+  ) {
+    const tabNode = $createTabNode();
+    const lineBreakNode = $createLineBreakNode();
+    selection.insertNodes([tabNode, lineBreakNode]);
+    const parent = tabNode.getParentOrThrow();
+    const index = tabNode.getIndexWithinParent();
+    parent.select(index, index + 2);
```
etrepum, line 653: I think this can create a selection that's a bit different than what is naturally created in the editor, generally if the boundary of a selection is at a TextNode it will always be a text point. The select method of ElementNode doesn't do this normalization. The NodeCaret API provides a bit of a better abstraction for this, e.g.

```ts
$setSelectionFromCaretRange(
  $getCaretRange(
    $getTextPointCaret(tabNode, 'next', 0),
    $normalizeCaret($getSiblingCaret(lineBreakNode, 'next')),
  ),
);
```

etrepum (approved):

packages/lexical-code/src/CodeHighlighter.ts
jeromew: I'll also try to see if I can re-use the original LineBreakNode that is already present. I guess selection.getNodes() would reply with the only LineBreakNode that is in the selection in this case

jeromew:
@etrepum note that I'll need to change this last commit. I realized it does not maintain the forwardness of the selection.

do you think I should modify/duplicate the 37 tests to make sure that they all work correctly when selection is done in a forward manner and that the forwardness is maintained ?

packages/lexical-code/src/__tests__/unit/LexicalCodeNodeTabs.test.ts
```diff
@@ -0,0 +1,222 @@
...
+        ['>|-line1>li|ne2', '>|line1>li|ne2', 'outdent'],
+        ['>li|ne1>li|ne2', '>-li|ne1>-li|ne2', 'tab'],
+        ['>li|ne1>li|ne2', '>-li|ne1>-li|ne2', 'indent'],
+        ['>-li|ne1>>-li|ne2', '>--li|ne1>>--li|ne2', 'tab'],
+        ['>-li|ne1>>-li|ne2', '>--li|ne1>>--li|ne2', 'indent'],
+        ['>-li|ne1>>-li|ne2', '>li|ne1>>li|ne2', 'outdent'],
+        ['>|-line1>->-line2|', '>|--line1>-->--line2|', 'tab'],
+      ];
+      suite.forEach((scenario) => {
+        test(`testing ${scenario[2]}: ${scenario[0]} => ${scenario[1]} `, async () => {
```
etrepum, line 82: This is probably how I would implement the directionality

```suggestion
      suite.forEach((scenario) => {
        ['forwards', 'backwards'].forEach((direction) => {
          test(`testing ${scenario[2]}: ${scenario[0]} => ${scenario[1]} ${direction}`, async () => {
```

packages/lexical-code/src/__tests__/unit/LexicalCodeNodeTabs.test.ts
```diff
@@ -0,0 +1,222 @@
...
+                );
+              } else {
+                selection.focus.set(
+                  codeNode.getKey(),
+                  parentIndex + 1,
+                  'element',
+                );
+              }
+            }
+            $setSelection(selection);
```
etrepum, line 186:
```suggestion
            $setSelectionFromCaretRange(
              $getCaretRangeInDirection(
                $caretRangeFromSelection(selection),
                direction === 'forwards' ? 'next' : 'previous',
              ),
            );
```

This could also be simplified if PointCaret was used for anchor/focus instead of manipulating a RangeSelection.

jeromew:
@etrepum I added the code to have the tests also run in backward selection + fixed the 2 tests that failed in backward mode (they were the 2 tests related to the addition of [tabNode, lineBreakNode] that were initially selected only in forward mode)

I did not see a link between the failing tests (collab-mac & flaky) and the PR.

etrepum (approved):

PR #7579, opened 2025-05-29
[lexical] Fix: ctrl+a doesn't work well in plain text mode on Firefox

### LINKED ISSUE #7578: Bug: Ctrl+A doesn't work well in PlainTextPlugin on Firefox
Lexical version: 0.31.2

## Steps To Reproduce

1. Clone lexical repository.
2. Move `examples/react-plain-text`.
3. Run npm install & npm run dev.
4. Visit dev site using Firefox.
5. Enter any text which ends with LineBreak (e.g. `aaa\naaa\n`)
6. Enter `Ctrl+A`.

## The current behavior

Selection contains only first line text.

```
 root
  └ (2) paragraph 
>   ├ (3) text "aaa"
    ├ (4) linebreak 
    ├ (5) text "aaa"
    └ (6) linebreak 

 selection: range  
  ├ anchor { key: 3, offset: 0, type: text }
  └ focus { key: 2, offset: 1, type: element }
```

## The expected behavior

Selection contains all elements.

```
 root
  └ (2) paragraph 
>   ├ (3) text "aaa"
>   ├ (4) linebreak 
>   ├ (5) text "aaa"
>   └ (6) linebreak 

 selection: range  
  ├ anchor { key: 3, offset: 0, type: text }
  └ focus { key: 2, offset: 4, type: element }
```

## Additions

It works well on Chrome.

wataru-chocola:
## Description

Enable override of selectall introduced by #4671 for Firefox.

Closes #7578 

### Before

before

### After

after


etrepum (commented):
I don't think skipping the tests is the best approach, especially because it means that the tests don't cover what you're trying to fix here

wataru-chocola:
> I don't think skipping the tests is the best approach, especially because it means that the tests don't cover what you're trying to fix here

You are right.
I fixed test cases again.

PR #7581, opened 2025-05-29
Fix lexical offset invalid selection

### LINKED ISSUE #7580: Bug: @lexical/offset makes invalid selection if offset points to inline node
Lexical version: 0.31.2

## Steps To Reproduce

1. Clone repository.
3. Visit dev site.
4. Input `aaaaa\naa`.

Link to code example:

## The current behavior

Image

```
client:912 [vite] connected.
    at Lexical.dev.mjs:8347:9
overrideMethod @ hook.js:608
```

## The expected behavior

No error occurs.

wataru-chocola:
## Description

Fixed anchor / focus offset of the selection created by OffsetView.createSelectionFromOffsets.

Closes #7580


packages/lexical-offset/src/index.ts
```diff
@@ -158,20 +158,18 @@ export class OffsetView {
...
       }
     } else if (startOffsetNode.type === 'inline') {
-      startKey = startNode.getParentOrThrow().getKey();
-      startOffset =
-        end > startOffsetNode.start
-          ? startOffsetNode.end
-          : startOffsetNode.start;
+      const parent = startNode.getParentOrThrow();
+      startKey = parent.getKey();
+      startOffset = parent.getChildrenKeys().indexOf(startNode.getKey());
```
etrepum, line 163:
```suggestion
      startOffset = startNode.getIndexWithinParent();
```

packages/lexical-offset/src/index.ts
```diff
@@ -158,20 +158,18 @@ export class OffsetView {
...
     if (endOffsetNode.type === 'text') {
       endOffset = end - endOffsetNode.start;
       endType = 'text';
     } else if (endOffsetNode.type === 'inline') {
-      endKey = endNode.getParentOrThrow().getKey();
-      endOffset =
-        end > endOffsetNode.start ? endOffsetNode.end : endOffsetNode.start;
+      const parent = endNode.getParentOrThrow();
+      endKey = parent.getKey();
+      endOffset = parent.getChildrenKeys().indexOf(endNode.getKey());
```
etrepum, line 172:
```suggestion
      endOffset = endNode.getIndexWithinParent();
```

etrepum:
Can someone else take a close look at this one? I don't see any usage of `@lexical/offset` except in collab and I'm not sure what the existing test coverage from there looks like if there was really anything /cc @james-atticus @amanharwara

james-atticus:
With respect to collab specifically, only `$createChildrenArray` is used which is a simple function that's not affected by this change. It's not doing anything specific to offsets and could probably live in core `lexical`.

Looking at the [original PR](https://github.com/facebook/lexical/pull/664/files#diff-2ef5aa39ab8ecb21e707aadf59e4966c46db0eaa8db6c59f944b81a766e1bfe8R189) though, I'm pretty sure these start/end values are intended to be absolute offsets rather than offsets relative to the parent -- `state.offset` (where start/end come from) is only ever increased by 1 or the length of the text. This is pretty similar to ProseMirror's [Indexing](https://prosemirror.net/docs/guide/), except that it doesn't count entering/leaving a block, which makes sense given that offsets [were previously used](https://github.com/facebook/lexical/pull/876/files) for selection restoration in collab.

etrepum:
After having a closer look myself, it looks like @james-atticus is right and that this probably isn't the correct fix. I think if we are going to fix anything in this module we should have a more comprehensive test suite that shows that it handles the diffing and offset clamping as expected in more than just this one case, and include the behavior when nodes are nested (e.g. an inline element node is not covered by this test case)

wataru-chocola:
Thank you for reviewing, but I don't quite understand comments.

I don't think this PR changed anything related to absolute offsets.
OffsetView.createSelectionFromOffsets converts absolute start / end offsets to selection.
This PR fixed offset calculation of **selection**, which is inherently relative within anchor / focus node.

Do I misunderstand something?

etrepum:
Possibly not, I did not give it a very thorough audit because I would have to carefully read all of the code in context to make sure it is correct. There just aren’t enough tests, and absolutely no usage in the playground or examples, to give me any confidence that this module does what it’s supposed to do before or after this change.

james-atticus:
My mistake, @wataru-chocola is correct. I was assuming that the original behaviour (returning absolute offset) was intentional, but that's incorrect. For the `text` case:

```
    if (startNode.type === 'text') {
      startOffset = start - startNode.start;
      startType = 'text';
```

`start` and `startNode.start` are both absolute indexes, so `startOffset` ends up being relative to the text node. This PR is doing the same for `inline` nodes.

The part I'm still unsure on though is whether it's correct to drop the `end > startOffsetNode.start` check. Going back to the original PR, the check looks to be there to handle some behaviour change based on `isCollapsed`. Haven't got my head around exactly what it's meant to do though. As @etrepum says, hard to know what the original intent was without tests or playground usage.

PR #7589, opened 2025-05-31
[lexical-react] Bug Fix: page freezes when typing a link in an overflow area

### LINKED ISSUE #6662: Bug: useCharacterLimit autolink infinite loop bug

### LINKED ISSUE #7474: Bug: Page freezes when typing a link in overflow area with Character Limit plugin enabled
When the Character Limit plugin is active in the playground, typing a link after exceeding the character limit causes the entire page to freeze and become unresponsive.

Image
Image
Image

kamilpitula:
## Description
Fix for an infinite loop bug caused by auto link plugin and overflow that were wrapping each other indefinitely.
Added an additional check within $wrapOverflowedNodes function that skips leaf nodes that already have an overflow parent.

Closes #7474
Closes #6662 
Closes #6668

### Before

When typing in an autolink with a character limit on, page freezes

### After

Page doesn't freeze anymore

https://github.com/user-attachments/assets/07db956b-d9a2-46ab-b2b8-e000f09d64b7

### DIFF
--- a/packages/lexical-react/src/shared/useCharacterLimit.ts
+++ b/packages/lexical-react/src/shared/useCharacterLimit.ts
@@ -173,6 +173,9 @@ function $wrapOverflowedNodes(offset: number): void {
 
   for (let i = 0; i < dfsNodesLength; i += 1) {
     const {node} = dfsNodes[i];
+    const hasOverflowParent = !!node
+      .getParents()
+      .find((p) => p instanceof OverflowNode);
 
     if ($isOverflowNode(node)) {
       const previousLength = accumulatedLength;
@@ -215,7 +218,7 @@ function $wrapOverflowedNodes(offset: number): void {
           $unwrapNode(node);
         }
       }
-    } else if ($isLeafNode(node)) {
+    } else if ($isLeafNode(node) && !hasOverflowParent) {
       const previousAccumulatedLength = accumulatedLength;
       accumulatedLength += node.getTextContentSize();
 
--- a/packages/lexical-react/src/shared/useCharacterLimit.ts
+++ b/packages/lexical-react/src/shared/useCharacterLimit.ts
@@ -14,7 +14,12 @@ import {
   OverflowNode,
 } from '@lexical/overflow';
 import {$rootTextContent} from '@lexical/text';
-import {$dfs, $unwrapNode, mergeRegister} from '@lexical/utils';
+import {
+  $dfs,
+  $findMatchingParent,
+  $unwrapNode,
+  mergeRegister,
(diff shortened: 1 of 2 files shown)


packages/lexical-react/src/shared/useCharacterLimit.ts
```diff
@@ -173,6 +173,9 @@ function $wrapOverflowedNodes(offset: number): void {
 
   for (let i = 0; i < dfsNodesLength; i += 1) {
     const {node} = dfsNodes[i];
+    const hasOverflowParent = !!node
+      .getParents()
+      .find((p) => p instanceof OverflowNode);
```
etrepum, line 178: Probably makes sense to compute the full condition here since $isLeafNode is cheap and traversing the parents is not so cheap.

```suggestion
    const needsOverflowParent = $isLeafNode(node) && !$findMatchingParent(node, $isOverflowNode);
```

packages/lexical-react/src/shared/useCharacterLimit.ts
```diff
@@ -215,7 +218,7 @@ function $wrapOverflowedNodes(offset: number): void {
           $unwrapNode(node);
         }
       }
-    } else if ($isLeafNode(node)) {
+    } else if ($isLeafNode(node) && !hasOverflowParent) {
```
etrepum, line 221:
```suggestion
    } else if (needsOverflowParent) {
```

etrepum (commented):
Looks good, I like the simplicity of this solution and the addition of a test! I think using the $isOverflowNode predicate would be more idiomatic, instanceof is rarely used directly in lexical code

packages/lexical-react/src/shared/useCharacterLimit.ts
kamilpitula: Good point! I changed the condition as you suggested.

etrepum (approved):

PR #7600, opened 2025-06-03
[lexical] Bug Fix: synced the export html output with import DOM expectations and browser DOM

jvithlani:
## Description

**Context:**
Currently the Lexical reconciler adds an extra `br` tag in the DOM to ensure the br tag added by us is not ignored by the parent block element.
So if there are n consecutive LB nodes in a block node, there will be n+1 `br` elements in the DOM. This hold true only if the last element of the block node is LB node.
This is handled by `reconcileElementTerminatingLineBreak` helper present in `packages/lexical/src/LexicalReconciler.ts`

In order to handle this, the importDOM method of the lineBreak node, ignores the last br node present in the imported HTML to avoid adding extra line breaks.

**Issue:**
The export DOM method does not export the extra br tag, which is anticipated to be present in the output by the import method, hence if we have 3 LB nodes added, there will be 4 br elements in the DOM. but the exported output will only have 3 LB nodes.
If we try to parse the same HTML output in our editor, we will trim away the last LB node, leading to only 2 line breaks in the given output which leads to a discrepancy.

I logged the HTML output in our plaground, attaching screenshots below
Before:
Here there are 2 nodes inserted in the video, 3 BR elements in the DOM but only 2 br elements are exported

https://github.com/user-attachments/assets/67b8fa78-4c3d-4192-8a52-e3e8cb897a48

After:
Here there are 2 nodes inserted in the video, 3 BR elements in the DOM but 3 br elements are exported

Screenshot 2025-06-03 at 4 44 00 PM

https://github.com/user-attachments/assets/5977dac1-d93e-4ffd-ba60-274b0cd23cce

Closes #

### Before

### After

### DIFF
--- a/packages/lexical/src/nodes/LexicalLineBreakNode.ts
+++ b/packages/lexical/src/nodes/LexicalLineBreakNode.ts
@@ -6,10 +6,11 @@
  *
  */
 
-import type {KlassConstructor} from '../LexicalEditor';
+import type {KlassConstructor, LexicalEditor} from '../LexicalEditor';
 import type {
   DOMConversionMap,
   DOMConversionOutput,
+  DOMExportOutput,
   NodeKey,
   SerializedLexicalNode,
 } from '../LexicalNode';
@@ -48,6 +49,18 @@ export class LineBreakNode extends LexicalNode {
 
   updateDOM(): false {
     return false;
+  }
+
+  override exportDOM(editor: LexicalEditor): DOMExportOutput {
+    const {element} = super.exportDOM(editor);
+    if (this.getNextSibling() !== null) {
+      return {element};
+    }
+    const documentFragment = document.createDocumentFragment();
+    const extraLineBreakElement = this.createDOM();
+    documentFragment.appendChild(element as HTMLElement);
+    documentFragment.appendChild(extraLineBreakElement);
+    return {element: documentFragment};
   }
 
   isInline(): true {


etrepum (commented):
I'm not sure this is the right way around, IIRC the extra `<br>` is there for input reasons and not rendering reasons so it would probably make more sense to remove this quirk importDOM rather than add a quirk to exportDOM. We could add a `data-lexical-linebreak="true"` attribute to these managed line breaks in the reconciler to make them easy to ignore if their rendered output is piped back in to importDOM.

Either way if we are changing behavior here we should make sure to add test coverage for it.

jvithlani:
@etrepum Sure, will add the tests once we finalise the approach
About removing the quirk from importDOM i visited the PR in which this was added https://github.com/facebook/lexical/pull/6395 it was done to make it consistent with how HTML reacts to the br tags. This is also helpful to main the WYSIWYG editor, so i felt this is the right thing to do, export the HTML which will render corectly even if rendered outside lexical editor, let me know your thoughts im open to both.

jvithlani:
@potatowagon bumping on this

jvithlani:
@etrepum what can we do here next?

etrepum:
We can wait and see if @potatowagon responds or proceed with the suggestion I had to add an attribute to managed line breaks for this purpose and add test coverage for the new behavior

potatowagon:
> add a data-lexical-linebreak="true" attribute to these managed line breaks in the reconciler

im not too clear what this approach would look like, is it a heavy lift?

etrepum:
That was my suggestion, so that it can be determined whether the linebreak is part of the document or a rendering artifact that should be excluded on import. The PR as it stands tries to export the extra linebreak, I guess as a workaround for what #6395 does

etrepum:
It should not be a heavy lift to set an attribute on managed linebreaks. The "hardest" part will be to fix the tests that are sensitive to that change in the rendered HTML, but that should be very mechanical.

jvithlani:
If we just wanna ignore the imported extra BR, we can rather remove that from the importDOM method instead of exporting and only ignoring that bit. As far as i understood this importDOM is written to handle even non-lexical generated HTML, so rather than just ignoring the br tag with our custom data-attribute we can take away that importing logic

jvithlani:
@etrepum thoughts?

etrepum:
I'm waiting to hear @potatowagon's thoughts because this PR is relevant to their #6395

jvithlani:
@potatowagon bumping on this, its open for a long time

jvithlani:
@etrepum are we good to merge this? this has been open for a while now

etrepum:
I don't think it should be merged as-is, for the same reasons as the comments I made previously https://github.com/facebook/lexical/pull/7600#pullrequestreview-2893090447

PR #7607, opened 2025-06-06
[lexical-react] Chore: Remove confusing return value

eliottvincent:
## Description
- What is the current behavior that you are modifying?
In the AutoLinkPlugin, specifically in the `TOGGLE_LINK_COMMAND` handler, a value is returned from a `forEach` loop, which has no effect, is confusing and doesn't make much sense.
- What are the behavior or changes that are being added by this PR?
The return value was removed.

### DIFF
--- a/packages/lexical-react/src/LexicalAutoLinkPlugin.ts
+++ b/packages/lexical-react/src/LexicalAutoLinkPlugin.ts
@@ -492,7 +492,8 @@ function useAutoLink(
             return false;
           }
           const nodes = selection.extract();
-          nodes.forEach((node) => {
+          for (let i = 0; i < nodes.length; i++) {
+            const node = nodes[i];
             const parent = node.getParent();
 
             if ($isAutoLinkNode(parent)) {
@@ -501,7 +502,7 @@ function useAutoLink(
               parent.markDirty();
               return true;
             }
-          });
+          }
           return false;
         },
         COMMAND_PRIORITY_LOW,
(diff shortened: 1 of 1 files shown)


etrepum (commented):
I think the correct fix may be to remove the `return true;` - what is the actual behavior that you are trying to fix with this PR?

eliottvincent:
> I think the correct fix may be to remove the `return true;` - what is the actual behavior that you are trying to fix with this PR?

Hey @etrepum , we are inside a command handler, so the return value (`true` / `false`) actually has a meaning. As of right now, whatever happens in the `TOGGLE_LINK_COMMAND` handler of AutoLinkPlugin, the return value will always be `false`, even if the AutoLink node was actually toggled. Meaning that any other command handler for `TOGGLE_LINK_COMMAND` will trigger afterwards.

etrepum:
I understand how command handlers work, but I think that the most appropriate fix for this listener is probably to remove the return statement altogether. It looks like what you're trying to do is fix the code to make sense (you are correct, there *is* something about this code that doesn't make sense), but not considering what the effect of this listener is supposed to be.

What this command is doing is toggling all link nodes in a selection, there is no clear reason why you would only want it to toggle the first AutoLinkNode and then stop. Returning true would prevent further AutoLinkNodes from being toggled, and also avoid calling any other TOGGLE_LINK_COMMAND listener that was registered at a lower priority or registered after at the same priority.

etrepum:
Either way, if this does fix a behavior issue, then we should describe what issue is being fixed and add an e2e test to show that it continues to work as expected.

eliottvincent:
You're right, I totally missed the fact that this command could be called on a selection with more than one matching node.
I just removed the `return true;` which is the correct way of fixing things and avoiding confusion.

Thanks for your feedback!

etrepum (approved):

etrepum:
Can you fix up the description of this PR to match the current implementation? Looks good to merge otherwise!

eliottvincent:
Done!

PR #7616, opened 2025-06-11
[lexical-yjs][lexical-react] Refactor: split out useYjsCollaborationInternal with shared v1/v2 logic

james-atticus:
## Description
Introduces a new `useYjsCollaborationV2__EXPERIMENTAL` and splits out shared logic from the original implementation into `useYjsCollaborationInternal` (open to naming suggestions).

The plan is to have entirely new `syncLexicalUpdateToYjs` and `syncYjsChangesToLexical` implementations for v2. Tree shaking should mean that consumers who only use one version won't have a noticeable increase in bundle size.

## Test plan

Straight refactor, should be no changes to any existing tests. V2 binding implementation not included in this PR, so no tests for that yet.


packages/lexical-react/src/shared/useYjsCollaboration.tsx
```diff
@@ -161,6 +122,119 @@ export function useYjsCollaboration(
...
+    color,
+    binding,
+    setDoc,
+    cursorsContainerRef,
+    awarenessData,
+    handleBootstrap,
+  );
+}
+
+export function useYjsCollaborationV2__EXPERIMENTAL(
```
james-atticus, line 172: I've removed `shouldBootstrap` from v2 because it's a big foot-gun in v1. The consistent recommendation in Discord is to set it to false, so figured may as well use this opportunity to remove it entirely.

`syncCursorPositionsFn` is tied specifically to `Binding`. I haven't looked into cursor syncing for the rewrite yet, when I get to that, I'll try and make the function generic.

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -38,7 +37,7 @@ test.describe('Collaboration', () => {
     isCollab,
     browserName,
   }) => {
-    test.skip(!isCollab || IS_MAC);
+    test.skip(!isCollab);
```
james-atticus, line 40: Drive-by: this test works fine on mac

packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
```diff
@@ -32,443 +32,441 @@ describe('Collaboration', () => {
...
-
-        const paragraph = root.getFirstChild<ParagraphNode>();
-
-        const text = $createTextNode('Hello world');
-
-        paragraph!.append(text);
+    if (useCollabV2) {
+      // Manually bootstrap editor state.
+      await waitForReact(() => {
+        client1.update(() => $getRoot().append($createParagraphNode()));
```
james-atticus: Needed because v2 doesn't support `shouldBootstrap`

packages/lexical-yjs/src/Bindings.ts
```diff
@@ -46,10 +46,42 @@ export type Binding = BaseBinding & {
   root: CollabElementNode;
 };
 
+export type LexicalMapping = Map<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  YAbstractType<any>,
+  // Either a node if type is YXmlElement or an Array of text nodes if YXmlText
+  LexicalNode | Array<TextNode>
+>;
```
james-atticus: Equivalent of the [type in y-prosemirror](https://github.com/yjs/y-prosemirror/blob/v1.3.5/src/plugins/sync-plugin.js#L51-L54).

packages/lexical-yjs/src/SyncEditorStates.ts
```diff
@@ -304,3 +309,117 @@ export function syncLexicalUpdateToYjs(
...
+  const dirtyElements = new Set<NodeKey>();
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const collectDirty = (_value: any, type: YAbstractType<any>) => {
+    if (binding.mapping.has(type)) {
+      const node = binding.mapping.get(type)!;
+      if (!(node instanceof Array)) {
+        dirtyElements.add(node.getKey());
+      }
+    }
+  };
```
james-atticus: In `y-prosemirror`, for any YType that's been modified, the corresponding entry is [deleted from the mapping](https://github.com/yjs/y-prosemirror/blob/v1.3.5/src/plugins/sync-plugin.js#L607-L617). This then causes the sync code to recreate the code in [createNodeIfNotExists](https://github.com/yjs/y-prosemirror/blob/v1.3.5/src/plugins/sync-plugin.js#L688C7-L688C28). In Lexical however, we don't want to do this - we want to update the existing node if possible (i.e. keep the same nodeKey).

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,650 @@
...
+    ? !item.deleted
+    : snapshot.sv.has(item.id.client) &&
+      /** @type {number} */
+      snapshot.sv.get(item.id.client)! > item.id.clock &&
+      !Y.isDeleted(snapshot.ds, item.id);
+
+/**
+ * @return Returns node if node could be created. Otherwise it deletes the yjs type and returns null
+ */
+export const $createOrUpdateNodeFromYElement = (
```
james-atticus, line 82: Per above, this is basically the same as [createNodeFromYElement](https://github.com/yjs/y-prosemirror/blob/v1.3.5/src/plugins/sync-plugin.js#L724), except that it takes `dirtyElements` and updates existing nodes.

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,650 @@
...
+  return nodes;
+};
+
+const createTypeFromTextNodes = (
+  nodes: TextNode[],
+  meta: BindingV2,
+): Y.XmlText => {
+  const type = new Y.XmlText();
+  const delta = nodes.map((node) => ({
+    attributes: {__properties: propertiesToAttributes(node, meta)},
```
james-atticus: y-prosemirror has marks (`marksToAttributes(node.marks, meta)`) for text nodes and attributes (`node.attrs`) for element nodes. Lexical does have this same delineation - just got the one `propertiesToAttributes` helper function.

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,650 @@
...
+        break;
+      }
+    }
+  }
+  // find number of matching elements from right
+  for (; right + left + 1 < minCnt; right++) {
+    const rightY = yChildren[yChildCnt - right - 1];
+    const rightP = pChildren[pChildCnt - right - 1];
+    if (mappedIdentity(meta.mapping.get(rightY), rightP)) {
+      if (rightP instanceof ElementNode && dirtyElements.has(rightP.getKey())) {
```
james-atticus: This is where `dirtyElements` is used to update YFragments without removing them from the mapping.

packages/lexical-playground/src/appSettings.ts
```diff
@@ -33,6 +33,7 @@ export const DEFAULT_SETTINGS = {
   tableCellBackgroundColor: true,
   tableCellMerge: true,
   tableHorizontalScroll: true,
+  useCollabV2: false,
```
james-atticus, line 38: Adding as a setting but not exposing it through the settings menu yet, given how early days this still is.

packages/lexical-yjs/src/Bindings.ts
```diff
@@ -63,16 +102,26 @@ export function createBinding(
...
+  excludedProperties?: ExcludedProperties,
+): BindingV2 {
+  invariant(
+    doc !== undefined && doc !== null,
+    'createBinding: doc is null or undefined',
+  );
+  return {
+    ...createBaseBinding(editor, id, doc, docMap, excludedProperties),
+    mapping: new Map(),
+    root: doc.get('root-v2', XmlElement) as XmlElement,
```
etrepum: I like that this has a different default to avoid bad behavior when different versions of collab are used with the same doc, although maybe it should be parameterized further in some way to also solve the use case from #6483?

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,647 @@
...
+    return node;
+  } catch (e) {
+    // an error occured while creating the node. This is probably a result of a concurrent action.
+    // TODO(collab-v2): also delete the mapped node from editor state.
+    el.doc!.transact((transaction) => {
+      el._item!.delete(transaction);
+    }, meta);
+    meta.mapping.delete(el);
+    return null;
+  }
```
etrepum: Would be nice to narrow this down to specific exceptions we'd expect to see and/or to a smaller portion of that code. A whole lot is happening in that block and this would make it quite hard to debug.

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,647 @@
...
+      nodes.pop()!.remove();
+    }
+  } catch (e) {
+    // an error occured while creating the node. This is probably a result of a concurrent action.
+    // TODO(collab-v2): also delete the mapped text nodes from editor state.
+    text.doc!.transact((transaction) => {
+      text._item!.delete(transaction);
+    });
+    return null;
+  }
```
etrepum: Same as above, I think this will make debugging very difficult

etrepum (commented):
I haven't been able to do invest the time to do a very thorough review, but the things that caught my eye were mostly just the sections where this is handling exceptions indiscriminately

packages/lexical-yjs/src/Bindings.ts
james-atticus: Can do.

packages/lexical-yjs/src/SyncV2.ts
james-atticus: I'd been meaning to come back to this. Given it's in the functions for going from Yjs->Lexical, I'd rather just throw all errors and let the Lexical error boundary handle it. Just deleting data from Yjs seems needlessly dangerous. (It's only like this now because that's how y-prosemirror works.)

etrepum:
Might be worth thinking through how future migration use cases can be handled (e.g. #7780)

packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
```diff
@@ -160,12 +160,21 @@ test.describe('Collaboration', () => {
...
+        focusPath: [0, 0, 0],
+      });
+    } else {
+      await assertSelection(page, {
+        anchorOffset: 1,
+        anchorPath: [0],
+        focusOffset: 1,
+        focusPath: [0],
+      });
+    }
```
james-atticus, line 177: Same overall outcome. This was the driver for #7794.

packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs
```diff
@@ -1101,7 +1108,7 @@ const TYPED_MARKDOWN_HTML = html`
   <p class="PlaygroundEditorTheme__paragraph" dir="auto">
     <span data-lexical-text="true">This is</span>
     <strong
-      class="PlaygroundEditorTheme__textBold PlaygroundEditorTheme__textStrikethrough PlaygroundEditorTheme__textItalic"
+      class="PlaygroundEditorTheme__textBold PlaygroundEditorTheme__textItalic PlaygroundEditorTheme__textStrikethrough"
```
james-atticus, line 1111: I was seeing class names be applied in a different order between left and right clients. Didn't get to the bottom of why, by changing the order of the markdown modifiers above then the test here seemed to fix it.

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -6700,6 +6688,8 @@ test.describe.parallel('Tables', () => {
       false,
     );
 
+    await sleep(1050);
```
james-atticus, line 6693: To ensure that undo on v2 doesn't delete the whole table.

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -1447,7 +1448,7 @@ test.describe.parallel('Tables', () => {
   });
 
   test(
-    'Table selection: can select multiple cells and insert an image',
```
james-atticus, line 1451: There's nothing special about images here from what I can tell, so switched it to another decorator node so that the test can still run on v2.

packages/lexical-react/src/LexicalCollaborationPlugin.tsx
```diff
@@ -208,3 +181,177 @@ function YjsCollaborationCursors({
 
   return cursors;
 }
+
+type CollaborationPluginV2Props = {
+  id: string;
+  providerFactory: ProviderFactory;
+  shouldBootstrap: boolean;
```
james-atticus: I was hoping to avoid `shouldBootstrap` but it was a nightmare getting existing tests to pass without it. One option would be to do something like:

```suggestion
  shouldBootstrap?: false;
```

Then when it's used in the playground with boolean or in unit tests, add a `ts-expect-error` comment explaining why it's overridden for those specific cases (but reinforcing that it shouldn't be overridden elsewhere). At least that way you _really_ have to opt-in.

Either that, or call it `__shouldBootstrapUnsafeInternalUseOnly`.

packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -234,6 +234,7 @@ export class TableNode extends ElementNode {
     this.__rowStriping = false;
     this.__frozenColumnCount = 0;
     this.__frozenRowCount = 0;
+    this.__colWidths = undefined;
```
james-atticus, line 237: All properties need to have default values for `initializeNodeProperties` to work. This also impacts current collab code FWIW, it's just that the playground inserts tables with default colWidths set.

packages/lexical-yjs/flow/LexicalYjs.js.flow
```diff
@@ -106,7 +106,7 @@ export type Binding = {
   docMap: Map<string, Doc>,
   editor: LexicalEditor,
   id: string,
-  nodeProperties: Map<string, Array<string>>,
+  nodeProperties: Map<string, {[property: string]: mixed}>,
```
james-atticus, line 109: Not familiar with Flow - is this correct?

packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -293,19 +350,20 @@ function updateCursor(
...
 }
-
 type AnyCollabNode =
   | CollabDecoratorNode
   | CollabElementNode
   | CollabTextNode
   | CollabLineBreakNode;
 
+/**
+ * @deprecated Use `$getAnchorAndFocusForUserState` instead.
```
james-atticus, line 359: Don't think we should be exposing collab node types outside of this package.

packages/lexical-react/src/LexicalCollaborationPlugin.tsx
```diff
@@ -208,3 +199,72 @@ function YjsCollaborationCursors({
...
+  excludedProperties,
+  awarenessData,
+}: CollaborationPluginV2Props): JSX.Element {
+  const collabContext = useCollaborationContext(username, cursorColor);
+  const {yjsDocMap, name, color} = collabContext;
+
+  const [editor] = useLexicalComposerContext();
+  useCollabActive(collabContext, editor);
+
+  const doc = useMemo(() => docProp || new Doc(), [docProp]);
```
james-atticus: Still not entirely happy with this. If the ID changes, presumably you're connecting to a different document, so you should be throwing out the doc and creating a new one. Might have to go with a doc factory prop instead.

james-atticus: That said, if you're changing the ID without throwing away the whole editor/editor state, then I can't see things working well anyway.

packages/lexical-code-shiki/src/CodeHighlighterShiki.ts
```diff
@@ -777,7 +777,7 @@ export function registerCodeHighlighting(
       editor.registerMutationListener(
         CodeNode,
         (mutations) => {
-          editor.update(() => {
+          editor.read(() => {
```
etrepum: editor.read can trigger a flush if there is a pending update (there shouldn't be, but sometimes people do that sort of thing in mutations). AFAICT `updateCodeGutter` doesn't perform any mutations or need `$getEditor()` so the options argument or update isn't needed.

```suggestion
          editor.getEditorState().read(() => {
```

packages/lexical-code/src/CodeHighlighterPrism.ts
```diff
@@ -771,7 +771,7 @@ export function registerCodeHighlighting(
       editor.registerMutationListener(
         CodeNode,
         (mutations) => {
-          editor.update(() => {
+          editor.read(() => {
```
etrepum:
```suggestion
          editor.getEditorState().read(() => {
```

packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -753,6 +758,22 @@ export async function insertDateTime(page) {
   await sleep(500);
 }
 
+export function getExpectedDateTimeHtml() {
```
etrepum: I don't think there's anything particularly easy we can do about this but this is a bit fragile if you happen to be running tests around midnight in whatever local TZ you're using.

Unrelated to the scope of this PR, `date.toString()` is probably not a great representation for dates, would be better to use something like ISO 8601.

```
> new Intl.DateTimeFormat('sv-SE', {dateStyle: 'short'}).format(new Date(2025, 8, 18))
'2025-09-18'
> new Date(2025, 8, 18).toString()
'Tue Sep 18 2025 00:00:00 GMT-0800 (Pacific Standard Time)'
```

packages/lexical-react/src/LexicalCollaborationPlugin.tsx
etrepum: I think the unsafe name probably makes sense, but what we really need to do is provide a full example that doesn't bootstrap so people have a good template to start with

packages/lexical-react/src/LexicalCollaborationPlugin.tsx
```diff
@@ -208,3 +198,70 @@ function YjsCollaborationCursors({
...
+const useCollabActive = (
+  collabContext: CollaborationContextType,
+  editor: LexicalEditor,
+) => {
+  useEffect(() => {
+    collabContext.isCollabActive = true;
+
+    return () => {
+      // Resetting flag only when unmount top level editor collab plugin. Nested
+      // editors (e.g. image caption) should unmount without affecting it
```
etrepum, line 259: A better way to do this would be to track the editors, e.g. `Set<LexicalEditor>` and then you can check set membership to see if it's active for some editor and set size to see if it's active for any editor.

The context object could have a `get isCollabActive() { return this.collabEditors.size > 0; }` or whatever if we had a reason to preserve API compatibility for that context object

packages/lexical-react/src/shared/useYjsCollaboration.tsx
```diff
@@ -116,54 +102,229 @@ export function useYjsCollaboration(
...
+    const removeListener = editor.registerUpdateListener(
+      ({
+        prevEditorState,
+        editorState,
+        dirtyLeaves,
+        dirtyElements,
+        normalizedNodes,
+        tags,
+      }) => {
+        if (tags.has(SKIP_COLLAB_TAG) === false) {
```
etrepum:
```suggestion
        if (!tags.has(SKIP_COLLAB_TAG)) {
```

packages/lexical-yjs/flow/LexicalYjs.js.flow
etrepum: Looks right to me (but I also haven't seriously used flow since ~2018 other than lexical maintenance), could also be `Record<string, mixed>` and `Record<string, unknown>` on the other side. AFAIK there is no practical difference other than not having to give an arbitrary name to the keys.

packages/lexical-yjs/src/Bindings.ts
```diff
@@ -12,37 +12,72 @@ import type {CollabLineBreakNode} from './CollabLineBreakNode';
...
+import {Doc, XmlElement, XmlText} from 'yjs';
 
 import {Provider} from '.';
 import {$createCollabElementNode} from './CollabElementNode';
+import {CollabV2Mapping} from './CollabV2Mapping';
+import {initializeNodeProperties} from './Utils';
 
 export type ClientID = number;
-export type Binding = {
+export type BaseBinding = {
```
etrepum: FWIW it's better in typescript to use interfaces instead of types and intersection for this kind of thing. I think most of the `type X = {}` in lexical vs. `interface X {}` is because of flow habits (which has exact types by default, something that's hard to do with typescript).

packages/lexical-yjs/src/SyncCursors.ts
```diff
@@ -335,8 +392,93 @@ export function getAnchorAndFocusCollabNodesForUserState(
...
+      }
+    }
+    return {
+      anchorKey: anchorNode !== null ? anchorNode.getKey() : null,
+      anchorOffset,
+      focusKey: focusNode !== null ? focusNode.getKey() : null,
+      focusOffset,
+    };
+  } else {
+    invariant(false, 'getAnchorAndFocusForUserState: unknown binding type');
```
etrepum: we could eliminate the need for this kind of runtime exhaustiveness check if we used something like `export type AnyBinding = Binding | BindingV2;` used instead of `BaseBinding` as an argument to functions like this.

If a new type were ever added to the union then the lack of this sort of `else` branch would let the compiler tell you which code needs to be refactored to match it.

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,856 @@
...
+        el,
+        new Snapshot(prevSnapshot.ds, snapshot.sv),
+      ).forEach($createChildren);
+    }
+
+    $spliceChildren(node, children);
+  }
+
+  const attrs = el.getAttributes(snapshot);
+  // TODO(collab-v2): support for ychange
```
etrepum, line 160: Is this something that has to be addressed in the binding before release? I'm not that familiar with yjs internals

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,856 @@
...
+        ? computeYChange('added', el._item!.id)
+        : {type: 'added'};
+    }
+  }
+  */
+  const properties: Record<string, unknown> = {
+    ...getDefaultNodeProperties(node, binding),
+  };
+  const state: Record<string, unknown> = {};
+  for (const k in attrs) {
```
etrepum, line 178: Is it not possible to store the node state in its own Y.Map or equivalent? The goal here was to avoid a lot of redundant serialization. If we can sync only what changes we don't need to encode and decode the whole state record every time.

packages/lexical-yjs/src/SyncV2.ts
```diff
@@ -0,0 +1,856 @@
...
+        );
+        if (ns !== null) {
+          ns.forEach((textchild) => {
+            if (textchild !== null) {
+              children.push(textchild);
+            }
+          });
+        }
+      } else {
+        invariant(false, 'XmlHook is not supported');
```
etrepum, line 143: What is XmlHook? I didn't see that in the yjs api

packages/lexical-playground/__tests__/utils/index.mjs
james-atticus: The other option would be to use a different inline decorator node like `Equation`

packages/lexical-react/src/LexicalCollaborationPlugin.tsx
james-atticus: #7830 for the docs, renamed the prop in this PR.

packages/lexical-react/src/LexicalCollaborationPlugin.tsx
james-atticus: The implementation in this PR is the same as it was before, I've just moved the code. I can make the change if you prefer, I was just trying to avoid modifying logic when it wasn't required.

packages/lexical-yjs/src/Bindings.ts
james-atticus: Updated

packages/lexical-yjs/src/SyncCursors.ts
james-atticus: Good call!

packages/lexical-yjs/src/SyncV2.ts
james-atticus: `ychange` is Yjs' name for [metadata](https://github.com/yjs/yjs/blob/v13/src/types/YText.js#L1048-L1061) that's added when you call `toDelta` on `Y.Text`. y-prosemirror uses the same name for [element attributes](https://github.com/yjs/y-prosemirror/blob/master/src/plugins/sync-plugin.js#L791-L799). Support is purely additive, doesn't need to be added before release.

I'm planning to use it in a future PR to add support for diffing two different Yjs snapshots in the editor view. WIP code is [here](https://github.com/james-atticus/lexical/compare/yjs-v2-boilerplate...yjs-v2-snapshot-diff?expand=1), screenshot:

image

packages/lexical-yjs/src/SyncV2.ts
james-atticus: https://github.com/yjs/yjs/blob/v13/src/types/YXmlHook.js

The typing for this function comes from https://github.com/yjs/yjs/blob/v13/src/types/YXmlFragment.js#L371-L378

packages/lexical-yjs/src/SyncV2.ts
james-atticus: I can take a closer look at this. Pretty sure it'll work for elements, not so confident with `YText` attributes.

james-atticus: Turns out XmlElement attributes are stored in a YMap, so we get it free. I've updated the code below to only sync changed state keys.

No luck with XmlText though, doesn't seem compatible with the approach of syncing via `to/applyDelta` which is using `ContentFormat` to apply properties/state to ranges. Yjs does have some optimisations under the hood to not send unchanged values though.

etrepum (approved):
I think this is in a good state, it certainly seems to be an improvement over v1. Since it's experimental and not a default I think it's fairly low risk?

It's a shame that NodeState has to be serialized to/from string for Text nodes but if that becomes a bottleneck I'm sure that's something we could deal with later one way or another.

PR #7620, opened 2025-06-13
[lexical-react][lexical-table] Refactor: Optimize table observers

vect-spasquereau:
Hi Lexical team! 👋

This PR introduces an internal refactor of how `TableObserver`s are managed in the editor, aiming to improve performance when working with documents that include many tables.

## Problem
Currently, each table registers its own `TableObserver`, with several DOM event listeners (e.g. pointerdown, pointermove, pointerup) and more than 10 command registered. This works fine at small scale, but I noticed performance issues when the editor contains many tables. Especially, as each table registers redundant listeners, is leads to latency on pointer events (up to several hundreds of ms per `pointerdown` event 😕, according to document complexity)

Example (before):
Clicking in the document triggers multiple `pointerdownCallback` executions across all registered tables:
image

## What this PR does
- Centralizes event listeners, reducing the total number of active DOM listeners
- Centralizes all command listeners into a shared handler

Example (with these changes):
Same document, same action — now only one `pointerdown` is triggered:
image

## Compatibility
Same behavior for table selection, formatting, cut/paste, etc. All previous commands and handlers are preserved as-is.

I can make this an opt-in or behind a flag if you'd prefer — happy to adjust 🙌

Thanks for your awesome work on Lexical! 💛


packages/lexical-react/src/shared/useMap.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+import {useRef} from 'react';
+
+export function useMap<K, V>(initial?: Iterable<[K, V]>) {
```
etrepum: I'm not sure these are well-designed hooks. They will construct a new Map on every iteration when needed or not, and every call results in a new return object with new methods for each property. 

In either case the result seems indistinguishable from something like this (other than making the methods "bound" to the instance):

```ts
export function useMap<K, V>(initial?: Iterable<[K, V]>) {
  return useMemo(() => new Map(initial));
}
```

packages/lexical-table/src/index.ts
```diff
@@ -32,7 +32,6 @@ export {$getTableAndElementByKey, TableObserver} from './LexicalTableObserver';
 export {
   registerTableCellUnmergeTransform,
   registerTablePlugin,
-  registerTableSelectionObserver,
```
etrepum, line 35: This is a breaking API change, ideally there would be a way to maintain compatibility

packages/lexical-table/src/index.ts
```diff
@@ -53,12 +52,29 @@ export {
 } from './LexicalTableSelection';
 export type {HTMLTableElementWithWithTableSelectionState} from './LexicalTableSelectionHelpers';
 export {
+  $addHighlightStyleToTable,
```
etrepum: All of the extra exports here add support burden because now they're public API that shouldn't break, if there isn't an immediate use for them (e.g. in lexical-playground or lexical-react) I wouldn't export any of it.

etrepum (commented):
Before a full review this will have to pass all of the ci checks, it looks like the formatting doesn't match the prettier rules (`npm run prettier:fix` can resolve this) for packages/lexical-react/package.json and also unit tests are failing. https://github.com/facebook/lexical/actions/runs/15639431987/job/44062922469?pr=7620

I only did a short scan of some of the code and left a few comments. Ultimately I think this will be hard to accept without full backwards compatibility (certainly the tests but ideally the API as well).

The table observer code might even be due for a more substantial architectural refactor than this consolidation. It's not particularly well done code in the first place. The last pass I did over it was just fixing bugs in-place rather than coming up with a better solution altogether from first principles.

vect-spasquereau:
> Before a full review this will have to pass all of the ci checks, it looks like the formatting doesn't match the prettier rules (`npm run prettier:fix` can resolve this) for packages/lexical-react/package.json and also unit tests are failing. https://github.com/facebook/lexical/actions/runs/15639431987/job/44062922469?pr=7620
> 
> I only did a short scan of some of the code and left a few comments. Ultimately I think this will be hard to accept without full backwards compatibility (certainly the tests but ideally the API as well).
> 
> The table observer code might even be due for a more substantial architectural refactor than this consolidation. It's not particularly well done code in the first place. The last pass I did over it was just fixing bugs in-place rather than coming up with a better solution altogether from first principles.

Hello @etrepum👋 Thanks a lot for the feedback — totally makes sense.

You're right, I’ve been a bit hasty changing the API exposure too much in this version, and I agree that this PR introduces more surface than it probably should at this stage.

I’m currently working on a more compliant version in that matter. I can update this PR or create a new one if you prefer. Feel free to close this, if you prefer to keep things clean — no problem at all 🙌
Thanks again!

etrepum:
It's fine to leave this PR open or create a new one when you're ready, up to you. I will mark this as draft in the meantime

vect-spasquereau:
Hello @etrepum 👋

This is ready for review again! ✅

CI and tests should now be green.

You might notice that in `LexicalTablePlugin`, I had to check for `editor._window` before registering the observers.
Without this check, `editor._window` ends up `undefined` during tests — which is a bit unexpected, since it should be initialized via the test setup 🤷‍♂️

etrepum:
I don't have time to do a code review or any debugging right now but editor._window is initialized when the root element is attached which is done by the content editable. If the registerOptimizedTableSelectionObserver code used a root listener to register the events then it would probably work better.

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -66,10 +72,27 @@ export function TablePlugin({
...
+        hasTabHandler,
+      );
+    }
+
+    return () => {
+      if (unregisterObservers) {
+        unregisterObservers();
+      }
+    };
+  }, [editor, editor._window, hasTabHandler, useOptimizedMode]);
```
etrepum: using editor._window as a dependency here doesn't do anything, because there's nothing that would trigger a render when it changes. You would need to subscribe to its changes one way or another, e.g.

```ts
function useEditorWindow(editor): null | Window {
  const callbacks = useMemo(() => ({
    getServerSnapshot: () => null,
    getSnapshot: () => getEditorWindow(editor),
    subscribe: (onStoreChange: () => void) => editor.registerRootListener(onStoreChange),
  }), [editor]);
  return useSyncExternalStore(
    callbacks.subscribe,
    callbacks.getSnapshot,
    callbacks.getServerSnapshot,
  );
}
```

However, implementing it correctly is better done by having the register function use a root listener instead of trying to observe it here (this also makes it less react-dependent).

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -294,6 +342,1118 @@ export function registerTableSelectionObserver(
...
+  dom: HTMLTableElement;
+  node: TableNode;
+};
+
+export function registerOptimizedTableSelectionObserver(
+  editor: LexicalEditor,
+  hasTabHandler: boolean = true,
+): () => void {
+  const rootElement = editor.getRootElement();
+  const editorWindow = getEditorWindow(editor);
```
etrepum: Correct usage requires that you use a root listener instead of observing this once

vect-spasquereau:
Closing this PR, not up to date, still issues with selection change command, I'm working on this as a background task

PR #7626, opened 2025-06-16
[lexical-headless-test] Enhance navigator environment check with object type and structure validation

0xobedient:
Lexical version: 0.32.1

[lexical-headless] 
[env-utils] Feature: improve navigator detection logic for Node environments

Choose from the following PR Types:
Chore

## Description
lexical-headless must guarantee testing in a Node environment, not in jsDom.
To ensure this, the following comment is placed at the top of the file to explicitly run tests in the Node environment:
```javascript
 /**
    * @jest-environment node
 */
```
While testing 'should be headless environment', we verified that both window and document are undefined as expected.
However, navigator was found to be assigned an empty object and identified as an Object, which is unexpected.

To address this, we added logic to check whether navigator is a valid one created by the test environment (i.e., not a fake or placeholder object), ensuring a more accurate environment validation.

### Before
Test Failing because ``typeof navigator``is returning "object". 
image

### After
All Test succeeded.
image

### DIFF
--- a/packages/shared/src/isEmptyNavigator.ts
+++ b/packages/shared/src/isEmptyNavigator.ts
@@ -1,0 +1,15 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+export default function isEmptyNavigator() {
+  return (
+    typeof navigator === 'object' &&
+    Object.prototype.toString.call(navigator) === '[object Object]' &&
+    Object.keys(navigator).length === 0
+  );
+}
--- a/packages/lexical-devtools-core/src/generateContent.ts
+++ b/packages/lexical-devtools-core/src/generateContent.ts
@@ -35,7 +35,7 @@ import {LexicalCommandLog} from './useLexicalCommandsLog';
 export type CustomPrintNodeFn = (
   node: LexicalNode,
   obfuscateText?: boolean,
-) => string;
+) => string | undefined;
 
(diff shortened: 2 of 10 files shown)


packages/shared/src/isEmptyNavigator.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+export default function isEmptyNavigator() {
+  return (
+    typeof navigator === 'object' &&
+    Object.prototype.toString.call(navigator) === '[object Object]' &&
```
lilshady: I feel `Object.prototype.toString.call()` check is redundant since we're already checking the type?

0xobedient: Thanks for your comment.

In a jsDom environment, a properly defined navigator object returns `[object Navigator]` as shown in the referenced image. If `Object.prototype.toString.call(navigator) === '[object Object]'`, it means that the navigator is not properly defined and is instead a manually assigned empty object. This suggests that the environment is likely a node environment.

Brower:
image

jsDom(test without flag`@jest-environment node`):
image

0xobedient:
> How is this a net improvement, if the tests don't run in the configuration where this may have failed? It changes one test, which must have already passed, and seems to make no changes to any runtime code.
> 
> 
> 
> If this code is only relevant to tests, it should probably be colocated with the tests rather than placed in shared. We don't have other test utilities in the shared package.

You are right — this change does not introduce any modifications to the runtime code. As you suggested, I will move `isEmptyNavigator` into the test file. Thanks for review.

PR #7631, opened 2025-06-18
RFC [lexical] Fix: propagate directioned text up the tree in reconciler

james-atticus:
## Issue

Depending on the order of operations and what nodes are marked dirty, LexicalReconciler behaves differently: sometimes you end up with `dir="rtl/ltr"` (`p`, `a`) all the way up the DOM and sometimes it's just on child nodes of the tree (just `a`). For example in "Auto Links" tests, [Does not create redundant auto-link](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs#L225) has the direction set on `p` but [Handles multiple autolinks in a row](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs#L300) does not.

In some cases, eg "ListItemNode RTL behavior" tests, `dir` ends up on the root `contenteditable` node. Looking at [list items handle mixed LTR and RTL content](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts#L1451), the root and `ul` are `rtl` but the first `li` is `ltr`. This seems to be a last-write-wins situation: in the playground, if you have a table with RTL text in the first cell, then table becomes RTL; if you then add LTR text in another cell, the table gets switched back to LTR.

The TL;DR is that you end up with different behaviour based on whether the reconciler uses `subTreeDirectionedTextContent` or falls back to `activeEditorDirection`.

## Proposed Fix

Currently, `__lexicalTextContent` is attached to each DOM node up the tree. This is built up in [$createChildren](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical/src/LexicalReconciler.ts#L290) and [$reconcileNode](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical/src/LexicalReconciler.ts#L226). I.e. for the root node, `__lexicalTextContent` is the concatenated content of all children.

The same is not done for `__lexicalDirTextContent` - the subtree text is not appended in [$createChildrenWithDirection](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical/src/LexicalReconciler.ts#L259) nor [$reconcileChildrenWithDirection](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical/src/LexicalReconciler.ts#L439). This PR changes the behaviour to do the same as `__lexicalTextContent`, i.e. propagate up.

By always propagating the directioned text content up, the last-write-wins behaviour is removed, and instead the result is that a given node's direction will be the set based on the first directioned text of its descendants.

A side effect of this change is that `hasEmptyDirectionedText` will no longer be true in most cases, so `dir="ltr"` is going to be set on almost all DOM nodes - [this check](https://github.com/facebook/lexical/blob/v0.32.1/packages/lexical/src/LexicalReconciler.ts#L392) will no longer evaluate to true.

Wanted to get input on this change before I go and update all the test cases, as this will impact almost every file.

### DIFF
--- a/packages/lexical/src/LexicalReconciler.ts
+++ b/packages/lexical/src/LexicalReconciler.ts
@@ -256,7 +256,8 @@ function $createChildrenWithDirection(
   subTreeDirectionedTextContent = '';
   $createChildren(children, element, 0, endIndex, element.getDOMSlot(dom));
   reconcileBlockDirection(element, dom);
-  subTreeDirectionedTextContent = previousSubTreeDirectionedTextContent;
+  subTreeDirectionedTextContent =
+    previousSubTreeDirectionedTextContent + subTreeDirectionedTextContent;
 }
 
 function $createChildren(
@@ -357,7 +358,7 @@ function reconcileBlockDirection(
 ): void {
   const previousSubTreeDirectionTextContent: string =
     dom.__lexicalDirTextContent || '';
-  const previousDirection: string = dom.__lexicalDir || '';
+  const previousDirection: string | null = dom.dir || null;
 
   if (
     previousSubTreeDirectionTextContent !== subTreeDirectionedTextContent ||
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalReconciler.ts
```diff
@@ -357,7 +358,7 @@ function reconcileBlockDirection(
 ): void {
   const previousSubTreeDirectionTextContent: string =
     dom.__lexicalDirTextContent || '';
-  const previousDirection: string = dom.__lexicalDir || '';
+  const previousDirection: string | null = dom.dir || null;
```
james-atticus, line 361: This change isn't directly related to the PR, fixed it while I was in the area.

If an empty node was marked dirty, `dom.__lexicalDir` would be set to `ltr`, but because of the `hasEmptyDirectionedTextContent` check, `dom.dir` would not be set. If the node was then updated with directioned text, then `previousDirection === direction`, so the update logic below would be skipped at `dom.dir` would not be set. The unit test below was added to cover this case.

etrepum:
Related #2610

james-atticus:
Closing in favour of #7727

PR #7636, opened 2025-06-21
[lexical-playground] Bug Fix: Allow deleting empty column layouts via backspace 

### LINKED ISSUE #7615: Bug: Columns Layout delete
I'm not sure if it's a bug, but Column Layouts cannot be deleted using (only) the keyboard.
The only way to delete a layout is by selecting it and pressing delete.
Tested on the playground.

Lakshmanshankar:
## Description
Previously, column layouts could not be deleted using only the keyboard — even if they were empty.  
This PR enables backspace to delete an empty column layout **only when**:
- Cursor is at offset 0 in the first column
- All columns are empty

Closes #7615 

### Before

- Pressing backspace at the start of an empty column layout did nothing.

### After
[Recording.webm](https://github.com/user-attachments/assets/3e99ba4f-daa7-428e-94db-d3605f91167b)

### DIFF
--- a/packages/lexical-playground/src/plugins/LayoutPlugin/LayoutPlugin.tsx
+++ b/packages/lexical-playground/src/plugins/LayoutPlugin/LayoutPlugin.tsx
@@ -26,6 +26,7 @@ import {
   KEY_ARROW_LEFT_COMMAND,
   KEY_ARROW_RIGHT_COMMAND,
   KEY_ARROW_UP_COMMAND,
+  KEY_BACKSPACE_COMMAND,
 } from 'lexical';
 import {useEffect} from 'react';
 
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/LayoutPlugin/LayoutPlugin.tsx
```diff
@@ -97,6 +98,57 @@ export function LayoutPlugin(): null {
...
+
+        if (
+          !$isLayoutContainerNode(container) ||
+          !$isLayoutItemNode(currentLayoutItem)
+        ) {
+          return false;
+        }
+
+        const firstLayoutItemKey = container.getFirstChild()?.__key;
+        if (firstLayoutItemKey !== currentLayoutItem?.__key) {
```
etrepum:
```suggestion
        const firstLayoutItemKey = container.getFirstChild()?.getKey();
        if (firstLayoutItemKey !== currentLayoutItem?.getKey()) {
```

packages/lexical-playground/src/plugins/LayoutPlugin/LayoutPlugin.tsx
```diff
@@ -145,6 +197,11 @@ export function LayoutPlugin(): null {
         () => $onEscape(true),
         COMMAND_PRIORITY_LOW,
       ),
+      editor.registerCommand(
+        KEY_BACKSPACE_COMMAND,
+        () => $deleteEmptyLayoutOnBackspace(),
+        COMMAND_PRIORITY_LOW,
+      ),
```
etrepum: I think this can be more easily implemented by having LayoutItemNode implement collapseAtStart with a similar implementation, e.g.

```ts
function $isEmptyLayoutItemNode(node: LexicalNode): boolean {
  return $isLayoutItemNode(node) && !node.getFirstDescendant();
}
export class LayoutItemNode extends ElementNode {
  // …
  collapseAtStart(): boolean {
    const parent = this.getParentNodeOrThrow();
    if (
      this.is(parent.getFirstChild()) &&
      parent.getChildren().every($isEmptyLayoutItemNode)
    ) {
      this.parent.replace($createParagraphNode()).select();
      return true;
    }
    return false;
  }
}
```

etrepum (commented):
The big thing missing here is a test that shows that this works, but I think it can be cleaned up a bit more by using collapseAtStart instead of handling backspace specifically

Lakshmanshankar:
Hi [etrepum](https://github.com/etrepum),

I've pushed the changes to use `collapseAtStart()` as suggested, and removed the manual Backspace handling logic. I’ve also verified the behavior in the playground, and it works as expected.

Just wanted to confirm — when you mentioned "The big thing missing here is a test that shows that this works," were you referring to adding a unit test under something like "/lexical-playground/unit/LayoutItem.test.tsx"?

Happy to add that if you'd like. Thanks again for the helpful feedback!

etrepum:
Yes, a unit or e2e test is missing

ivailop7 (approved):

PR #7637, opened 2025-06-21
[lexical-table] Bug Fix: Ignore SELECTION_INSERT_CLIPBOARD_NODES_COMMAND dispatched from child editors

### LINKED ISSUE #7635: Bug: Editor freezes when pasting text copied from websites into a nested editor state
The editor consistently freezes when pasting text copied from websites into a nested editor state.

Lexical version: Observed in v0.30.0 and v0.32.1

## Steps To Reproduce

2. Delete everything and insert sample image
3. Add caption
4. Type something into the caption (e.g. `foo bar`)
5. Delete what you typed into the caption
6. Copy text from somewhere in github.com (e.g. some random text from a GitHub issue)
7. Paste it into the image caption in Lexical Playground (note: image caption is a nested editor state)
8. The editor is now frozen

Demo video:

https://github.com/user-attachments/assets/3f168bd2-15bc-4667-8b9d-1bcb370ac94d

## The current behavior

Pasting text from websites into the editor causes it to freeze.

## The expected behavior

Pasting text from websites doesn't cause the editor to freeze.

## Impact of fix

This is a very severe bug. We use nested editor states in several custom Lexical plugins. Our users commonly paste content into them, which causes the editor to freeze and become unusable.

etrepum:
## Description

Minimal fix for `SELECTION_INSERT_CLIPBOARD_NODES_COMMAND` command listeners to prevent parent editor from breaking when content is pasted into a nested editor.

See https://github.com/facebook/lexical/issues/7635#issuecomment-2993812367 for full discussion of the root cause and more complete ways to avoid this type of problem.

Separately, as a follow-up, it would make sense to also address the issue in lexical's error handling that doesn't manage to get the editor into a working state after this error is encountered.

Closes #7635

## Test plan

Added an e2e test to demonstrate that the parent editor is still responsive to edits after editing and clearing an image caption then pasting into the caption. Confirmed that this test fails if the bug fix commit is not applied.


johnnyoshika:
Thank you so much for this PR and attention to this matter!

ivailop7 (approved):

johnnyoshika:
Thank you for this. Any idea when a new release will be published?

etrepum:
Nightly releases are published every weekday. Other releases are generally once a month, will be soon.

PR #7638, opened 2025-06-22
[lexical-playground] [lexical-list] add ability to change ordered list start number

### LINKED ISSUE #7621: Feature: Add ability to reset/change ordered list numbering
## Description
At the moment, it is possible to start a list at an arbitrary number, but it only happens through starting a new list with the beginning number. For example, if you want a list that starts at 5, then outside of any node, start typing <kbd>5</kbd><kbd>.</kbd><kbd>space</kbd> and the text will get turned into an ordered list item starting at 5.

It would be great if there was a simple way through the included UI to reset the numbering from a specific number.

achaljhawar:
## Description

This PR introduces a feature allowing users to modify the starting number of an existing ordered list directly through the toolbar in the playground. 

Closes #7621
 
## Test plan

- Create an ordered list.
- Select an item in the ordered list.
- You should see a new number input field appear in the toolbar, next to the block type dropdown, showing the current start number .
- Change the number in that input field (e.g., to 9).
- Verify that the list in the editor updates its numbering accordingly (e.g., starts from 9, 10, 11...).

### Before

Currently, Lexical allows starting an ordered list with an arbitrary number only by creating a new list with that specific starting number. There is no UI mechanism to change the start number of an *existing* ordered list.

### After

https://github.com/user-attachments/assets/f2f8165e-2bbb-4101-b15c-6caa217b7677

### DIFF
--- a/packages/lexical-playground/src/index.css
+++ b/packages/lexical-playground/src/index.css
@@ -48,7 +48,7 @@ header h1 {
 .editor-shell {
   margin: 20px auto;
   border-radius: 2px;
-  max-width: 1140px;
+  max-width: 1100px;
   color: #000;
   position: relative;
   line-height: 1.7;
@@ -1491,7 +1491,6 @@ button.action-button:disabled {
 
 .toolbar {
   display: flex;
-  justify-content: center;
   margin-bottom: 1px;
   background: #fff;
   padding: 4px;
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
@@ -15,7 +15,7 @@ import {
   getLanguageFriendlyName,
 } from '@lexical/code';
 import {$isLinkNode, TOGGLE_LINK_COMMAND} from '@lexical/link';
-import {$isListNode, ListNode, UPDATE_LIST_START_COMMAND} from '@lexical/list';
+import {$isListNode, ListNode} from '@lexical/list';
 import {INSERT_EMBED_COMMAND} from '@lexical/react/LexicalAutoEmbedPlugin';
 import {INSERT_HORIZONTAL_RULE_COMMAND} from '@lexical/react/LexicalHorizontalRuleNode';
 import {$isHeadingNode} from '@lexical/rich-text';
@@ -581,21 +581,10 @@ export default function ToolbarPlugin({
             : element.getListType();
 
           updateToolbarState('blockType', type);
-          if (type === 'number') {
-            const start = parentList
-              ? parentList.getStart()
-              : element.getStart();
-            updateToolbarState('listStartNumber', start);
-          } else {
-            updateToolbarState('listStartNumber', null);
-          }
         } else {
-          updateToolbarState('listStartNumber', null);
           $handleHeadingNode(element);
           $handleCodeNode(element);
         }
-      } else {
-        updateToolbarState('listStartNumber', null);
       }
 
       // Handle buttons
@@ -830,30 +819,6 @@ export default function ToolbarPlugin({
               rootType={toolbarState.rootType}
               editor={activeEditor}
             />
-            {toolbarState.blockType === 'number' &&
-              toolbarState.listStartNumber !== null && (
-                <input
-                  type="number"
-                  value={toolbarState.listStartNumber}
-                  disabled={!isEditable}
-                  onChange={(e) => {
-                    const newStart = parseInt(e.target.value, 10);
-                    if (
-                      !isNaN(newStart) &&
-                      newStart >= 0 &&
-                      selectedElementKey
-                    ) {
-                      activeEditor.dispatchCommand(UPDATE_LIST_START_COMMAND, {
-                        listNodeKey: selectedElementKey,
-                        newStart,
-                      });
-                    }
-                  }}
-                  className="toolbar-item list-start-input"
-                  aria-label="List start number"
-                  style={{marginLeft: '5px', marginRight: '5px', width: '50px'}}
-                />
-              )}
             <Divider />
           </>
         )}


packages/lexical-list/src/formatList.ts
```diff
@@ -548,3 +550,25 @@ export function $handleListInsertParagraph(): boolean {
...
+  newStart: number,
+): void {
+  editor.update(() => {
+    const listNode = $getNodeByKey(listNodeKey);
+    if ($isListNode(listNode) && listNode.getListType() === 'number') {
+      listNode.setStart(newStart);
+      updateChildrenListItemValue(listNode);
+    }
+  });
+}
```
etrepum: Using editor here is unnecessary and actively error-prone (nesting editor.update calls has very unexpected behavior). Other functions with this sort of API have already been deprecated. The command listener already has an implicit editor.update and can directly call dollar functions.

```suggestion
/**
 * Handles the command to update the start number of an ordered list.
 * It retrieves the ListNode by its key, sets the new start number,
 * and updates the values of its child ListItemNodes.
 * @param editor - The LexicalEditor instance.
 * @param listNodeKey - The key of the ListNode to update.
 * @param newStart - The new start number for the list.
 */
export function $handleUpdateListStart(
  listNodeKey: NodeKey,
  newStart: number,
): void {
  const listNode = $getNodeByKey(listNodeKey);
  if ($isListNode(listNode) && listNode.getListType() === 'number') {
    listNode.setStart(newStart);
    updateChildrenListItemValue(listNode);
  }
}
```

packages/lexical-list/src/index.ts
```diff
@@ -77,6 +79,15 @@ export function registerList(editor: LexicalEditor): () => void {
       },
       COMMAND_PRIORITY_LOW,
     ),
+    editor.registerCommand(
+      UPDATE_LIST_START_COMMAND,
+      (payload) => {
+        const {listNodeKey, newStart} = payload;
+        $handleUpdateListStart(editor, listNodeKey, newStart);
```
etrepum: It probably makes more sense to inline the implementation of `$handleUpdateListStart` here since it's short, not used anywhere else, and isn't particularly useful as a public API.

achaljhawar: I have done the inline implementation now.

packages/lexical/src/LexicalCommands.ts
```diff
@@ -203,6 +203,11 @@ export const CLEAR_EDITOR_COMMAND: LexicalCommand<void> = createCommand(
 export const CLEAR_HISTORY_COMMAND: LexicalCommand<void> = createCommand(
   'CLEAR_HISTORY_COMMAND',
 );
+
+export const UPDATE_LIST_START_COMMAND: LexicalCommand<{
+  listNodeKey: NodeKey;
+  newStart: number;
+}> = createCommand('UPDATE_LIST_START_COMMAND');
```
etrepum: This command isn't relevant to the lexical core, it should be defined in and exported from lexical-list with the other list commands

etrepum:
A unit or e2e test to show that this works would also be nice

packages/lexical/src/LexicalCommands.ts
achaljhawar: I have moved the command to lexical-list now

packages/lexical/src/LexicalCommands.ts
```diff
@@ -203,6 +203,7 @@ export const CLEAR_EDITOR_COMMAND: LexicalCommand<void> = createCommand(
 export const CLEAR_HISTORY_COMMAND: LexicalCommand<void> = createCommand(
   'CLEAR_HISTORY_COMMAND',
 );
+
```
etrepum:
```suggestion
```

etrepum (approved):
Looks good! Will merge once the test suite has finished running successfully

etrepum:
I just had a closer look and perhaps we might want to reconsider this UI since it causes internal scrolling in the toolbar, even with a full width browser on a laptop screen. The code and implementation is otherwise clean, it's really just a UI concern.

https://github.com/user-attachments/assets/eb3f7fbd-8d4b-490c-b091-57b8f763cec8

achaljhawar:
> I just had a closer look and perhaps we might want to reconsider this UI since it causes internal scrolling in the toolbar, even with a full width browser on a laptop screen. The code and implementation is otherwise clean, it's really just a UI concern.
> 
>  numbered-list-toolbar.mov

yeah I initially tried to fix this problem by decreasing the size of buttons on the toolbar when a ordered list is clicked so that it could accommodate a new field, but the scroll was still persistent.

achaljhawar:
@etrepum Is this okay if I increase the size of the playground by 40px and center align the toolbar the scroll doesn't come 

https://github.com/user-attachments/assets/e7ec9f84-b9c9-4454-aad8-ffee5d626f93

should I commit this?

etrepum:
@ivailop7 do you have any opinions on this UI?

ivailop7:
> @ivailop7 do you have any opinions on this UI?

Not a fan of the UI change. I'd keep only the changes to the plugin, but skip the toolbar one. If people want to make use of the new functionality, it should be documented how to use 'UPDATE_LIST_START_COMMAND' and skip the UI part altogether.

achaljhawar:
> @etrepum Is this okay if I increase the size of the playground by 40px and center align the toolbar the scroll doesn't come
> 
>  Lexical.Playground.3.mp4 
> should I commit this?

@ivailop7 what do you think about this?

ivailop7:
> > @etrepum Is this okay if I increase the size of the playground by 40px and center align the toolbar the scroll doesn't come
> 
> > 
> 
> >  Lexical.Playground.3.mp4 
> 
> > should I commit this?
> 
> 
> 
> @ivailop7 what do you think about this?

Let's just scope the changes in this PR to the command for now and skip the toolbar and UI changes for now altogether.

achaljhawar:
@ivailop7 I have removed all the ui changes as you asked

ivailop7 (approved):

lachieh:
@ivailop7 would you mind if I made another issue to surface the command in the UI? I am imagining something like the table feature where right-clicking on the list item/s brings up a menu with "change list numbering" which then opens a modal with the number input. If that seems out of scope, I'll leave it alone.

PR #7641, opened 2025-06-24
[lexical] Bug Fix: copy NodeState when splitting text nodes

randal-atticus:
## Description
Fixes node state not being copied to the new sibling nodes when splitting text nodes.

## Test plan
Unit test created for this behaviour (and segmented text nodes)

### Before

Formatting the middle of a TextNode with state applied to it would drop the styling from the split parts.

https://github.com/user-attachments/assets/ad9a01ee-603e-42f3-85f9-38dddf4720ee

### After

(Difficult to re-record the above situation, but the test should demonstrate the desired outcome)

### DIFF
--- a/packages/lexical/src/nodes/LexicalTextNode.ts
+++ b/packages/lexical/src/nodes/LexicalTextNode.ts
@@ -983,6 +983,7 @@ export class TextNode extends LexicalNode {
     const format = self.getFormat();
     const style = self.getStyle();
     const detail = self.__detail;
+    const state = self.__state;
     let hasReplacedSelf = false;
 
     // Prepare to handle selection
@@ -1007,6 +1008,7 @@ export class TextNode extends LexicalNode {
       writableNode.__format = format;
       writableNode.__style = style;
       writableNode.__detail = detail;
+      writableNode.__state = state;
       hasReplacedSelf = true;
     } else {
       // For the first part, update the existing node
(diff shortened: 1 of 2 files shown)


packages/lexical/src/nodes/LexicalTextNode.ts
```diff
@@ -983,6 +983,7 @@ export class TextNode extends LexicalNode {
     const format = self.getFormat();
     const style = self.getStyle();
     const detail = self.__detail;
+    const state = self.__state;
```
etrepum: This is going to cause an aliasing issue when `state && state.node === self`, needs to be done a bit differently. Maybe a utility function like this:

```ts
function $cloneNodeState<T extends LexicalNode>(from: T, to: T): undefined | NodeState<T> {
  const state = from.__state;
  // Make a copy if the NodeState is writable to avoid aliasing
  return state && state.node === from
    ? state.getWritable(to)
    : state;
}
```

With usage like this:

```ts
writableNode.__state = $cloneNodeState(self, writableNode);
```

packages/lexical/src/nodes/__tests__/unit/LexicalTextNode.test.tsx
```diff
@@ -631,6 +634,40 @@ describe('LexicalTextNode tests', () => {
...
+
+    test('copies state to all nodes', async () => {
+      await update(() => {
+        const textNode = $createTextNode('hello world');
+        const state = createState('state', {
+          parse: (v) => v,
+          unparse: (v) => v,
+        });
+        $setState(textNode, state, 'foo');
+        const splits = textNode.splitText(3, 5);
```
etrepum, line 646: To observe the aliasing issue you could do `$setState(textNode, state, 'bar')` after the split and observe that they all changed (because `splits[0] === textNode` and `textNode.__state.node === textNode` here)

etrepum (commented):
Once the aliasing issue is addressed this looks like a good idea

PR #7642, opened 2025-06-24
[lexical-devtools-core] Chore: add NodeState to TreeView

randal-atticus:
## Description
Adds NodeState to TreeView, aiding debugging

### Before

Screenshot 2025-06-24 at 12 03 42 pm

### After

Screenshot 2025-06-24 at 12 03 15 pm

### DIFF
--- a/packages/lexical-devtools-core/src/generateContent.ts
+++ b/packages/lexical-devtools-core/src/generateContent.ts
@@ -315,6 +315,7 @@ function printAllTextNodeProperties(node: TextNode) {
     printFormatProperties(node),
     printDetailProperties(node),
     printModeProperties(node),
+    printStateProperties(node),
   ]
     .filter(Boolean)
     .join(', ');
@@ -325,6 +326,7 @@ function printAllLinkNodeProperties(node: LinkNode) {
     printTargetProperties(node),
     printRelProperties(node),
     printTitleProperties(node),
+    printStateProperties(node),
   ]
     .filter(Boolean)
     .join(', ');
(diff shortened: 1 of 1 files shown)


packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -396,6 +398,21 @@ function printTitleProperties(node: LinkNode) {
...
+function printStateProperties(node: LexicalNode) {
+  if (!node.__state) {
+    return false;
+  }
+  const states = [];
+  for (const [stateType, value] of node.__state.knownState.entries()) {
+    states.push(`[${stateType.key}: ${value}]`);
+  }
+  let str = states.join(',');
+  if (str != null) {
```
etrepum: The null check isn't going to do anything

```suggestion
  if (str !== '') {
```

packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -396,6 +398,21 @@ function printTitleProperties(node: LinkNode) {
...
 }
 
+function printStateProperties(node: LexicalNode) {
+  if (!node.__state) {
+    return false;
+  }
+  const states = [];
+  for (const [stateType, value] of node.__state.knownState.entries()) {
+    states.push(`[${stateType.key}: ${value}]`);
+  }
```
etrepum: probably makes sense to discard defaults here, and use unparse of the configs or possibly just work with the state.toJSON() serialization without being too tied to the internals. Could use a WeakMap to cache the result string for a given NodeState if performance is a concern.

packages/lexical-devtools-core/src/generateContent.ts
```diff
@@ -396,6 +398,25 @@ function printTitleProperties(node: LinkNode) {
...
+  if (!node.__state) {
+    return false;
+  }
+  const states = [];
+  for (const [stateType, value] of node.__state.knownState.entries()) {
+    if (stateType.isEqual(value, stateType.defaultValue)) {
+      continue;
+    }
+    const textValue = stateType.unparse(value);
+    states.push(`[${stateType.key}: ${textValue}]`);
```
etrepum: textValue isn't necessarily a string, could be anything JSON serializable (including an array or object).
```suggestion
    const textValue = JSON.stringify(stateType.unparse(value));
    states.push(`[${stateType.key}: ${textValue}]`);
```

PR #7647, opened 2025-06-25
[lexical-react] Bug Fix: aria-prohibited-attr violation on LexicalContentEditableElement

### LINKED ISSUE #7500: Bug: `role="textbox"` should be preserved in read-only mode
When a Lexical editor is **read-only**, its root `<div contenteditable>` **drops** `role="textbox"`.  
Playwright’s `locator.fill()` (and any API depending on the same autowait logic) rejects the element and throws:

```
Element is not an <input>, <textarea>, <select> or [contenteditable]
and does not have a role allowing [aria-readonly]
```

Every editor instance starts in this transient read-only state for a split second during initialisation, so tests frequently stumble over the error even though the editor soon becomes editable.

Lexical version: **0.30.0**

## Steps To Reproduce

1. clone https://github.com/thorn0/lexical-repro
2. `npm install`
3. `npm run test`

## The current behavior

* `role="textbox"` is **removed** while `aria-readonly="true"` is present.  
* Playwright (and assistive tech relying on ARIA roles) no longer recognises the element as a textbox, so autowait in `locator.fill()` aborts with the error shown above.

## The expected behavior

`role="textbox"` should remain on the element regardless of `aria-readonly`.  
With both attributes (`role="textbox" aria-readonly="true"`), Playwright can interact or wait safely, and assistive technologies keep the correct semantic.

## Impact of fix
* **Severity :** breaks E2E suites that use Playwright on apps built with Lexical, makes tests flaky.
* **Beneficiaries :** Any developers/components/tests that rely on Playwright autowait, plus users of assistive technology (WCAG recommends keeping the role and toggling only `aria-readonly`).

so99ynoodles:
## Description

- removing `role` in read-only editor leads to an a11y violation `aria-prohibited-attr`

Closes #7500

## Test plan
- add `jest-axe` for a11y check

### Before
`npm run test-unit packages/lexical-react/src/__tests__/unit/LexicalContentEditableElement.test.tsx`

image

### After
`npm run test-unit packages/lexical-react/src/__tests__/unit/LexicalContentEditableElement.test.tsx`

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11567,6 +11567,27 @@
         "pretty-format": "^29.0.0"
       }
     },
+    "node_modules/@types/jest-axe": {
+      "version": "3.5.9",
+      "resolved": "https://registry.npmjs.org/@types/jest-axe/-/jest-axe-3.5.9.tgz",
+      "integrity": "sha512-z98CzR0yVDalCEuhGXXO4/zN4HHuSebAukXDjTLJyjEAgoUf1H1i+sr7SUB/mz8CRS/03/XChsx0dcLjHkndoQ==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/jest": "*",
+        "axe-core": "^3.5.5"
+      }
+    },
+    "node_modules/@types/jest-axe/node_modules/axe-core": {
+      "version": "3.5.6",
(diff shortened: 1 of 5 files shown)


etrepum:
I think this looks fine but maybe @zurfyx should take a look

PR #7654, opened 2025-06-27
[lexical] Bug Fix: allow same mutation listener fn to be registered to multiple nodes

james-atticus:
## Description

`registerMutationListener` currently maps from the listener function to the corresponding node class. If you then call `registerMutationListener` with a different class, it overwrites the first. This also means that the cleanup function from the first class could end up removing the listener for the second class.

`registerNodeTransform` on the other hand already allows you to register the same listener function to multiple classes. This change brings the same behaviour to `registerMutationListener`.

## Test plan

Added unit test, fails before change, passes after.

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -244,7 +244,7 @@ export type Transform<T extends LexicalNode> = (node: T) => void;
 
 export type ErrorHandler = (error: Error) => void;
 
-export type MutationListeners = Map<MutationListener, Klass<LexicalNode>>;
+export type MutationListeners = Map<MutationListener, Set<Klass<LexicalNode>>>;
 
 export type MutatedNodes = Map<Klass<LexicalNode>, Map<NodeKey, NodeMutation>>;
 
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalEditor.ts
```diff
@@ -948,7 +948,10 @@ export class LexicalEditor {
       this.getRegisteredNode(klass),
     ).klass;
     const mutations = this._listeners.mutation;
-    mutations.set(listener, klassToMutate);
+    if (!mutations.has(listener)) {
```
etrepum: Just a small nit but if the same sort of klassSet variable approach is used here there could be a single get instead of has and then get. We don't need to check separately because undefined is not a valid value for the entry to have.

etrepum (approved):
Seems like the right thing to do, honestly surprised it worked this way. I guess most use cases were defining this function with a closure or just not registering multiple classes.

packages/lexical/src/LexicalEditor.ts
james-atticus: Something like this?

```
    let klassSet = mutations.get(listener);
    if (klassSet === undefined) {
      klassSet = new Set();
      mutations.set(listener, klassSet);
    }
    klassSet.add(klassToMutate);
```

I'm happy either way. I went with the current approach to avoid `let`.

etrepum: Yeah, I think that using let is a lesser sin than having more runtime code (and a non-null type assertion). There's plenty of mutable variables in this codebase

james-atticus: Done.

ivailop7 (approved):

PR #7655, opened 2025-06-27
[lexical-list] Bug Fix: Make tap target for checklists bigger so touch inputs register

### LINKED ISSUE #7651: Bug: Unable to check checkboxes on iPhone
Lexical version: v0.32.1

## Steps To Reproduce

1. Open lexical playground on an iPhone running Safari and iOS 18.5 (This has also been reported on Android, but I don't have a device on hand to reproduce)
2. Insert a checklist
3. Tap any of the checkboxes

https://github.com/user-attachments/assets/87ff26a7-47fc-4708-b3bf-a429f931f832

## The current behavior
I can't check a checkbox on my iPhone

## The expected behavior
I would like to be able to check checkboxes

## Impact of fix
This bug makes checklists unusable on iPhones and likely most other mobile devices

circlingthesun:
## Description

This makes the area in which touch events are registered bigger

Closes #7651

### DIFF
--- a/packages/lexical-list/src/checkList.ts
+++ b/packages/lexical-list/src/checkList.ts
@@ -156,10 +156,7 @@ export function registerCheckList(editor: LexicalEditor) {
   );
 }
 
-function handleCheckItemEvent(
-  event: PointerEvent | MouseEvent | TouchEvent,
-  callback: () => void,
-) {
+function handleCheckItemEvent(event: PointerEvent, callback: () => void) {
   const target = event.target;
 
   if (!isHTMLElement(target)) {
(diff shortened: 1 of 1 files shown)


ivailop7:
Does this respect CSS zoom after ignoring the calculateZoom scaling?

circlingthesun:
@ivailop7 it does now. Sorry, I didn't fully understand what was going on there. Applying zoom: 2 to the body worked in Chrome and Firefox, but I see now that it caused problems in Safari.

etrepum (approved):
Seems to work well!

I did notice that there's a bug with LexicalMenu along the way but that's unrelated to this PR: if you use the component picker to create the checklist then it will leave an empty div on top of the second checkbox preventing clicks or touches.

PR #7656, opened 2025-06-27
[lexical-table] Bug Fix: Fix table selection for touch devices

### LINKED ISSUE #7650: Bug: Table cell selection triggers unexpectedly on touch inputs
Lexical version: v0.32.1

## Steps To Reproduce

1. Open lexical playground on any mobile device or mobile device emulator on Chrome
2. Insert a table
3. Click in any cell
4. Click in any other cell
5. The cells between them are now selected

https://github.com/user-attachments/assets/22a03083-d5e4-4c95-a6d5-d0994c86796f

Link to code example:

## The current behavior
A cell selection is triggered every time a the cursor is moved from one cell to another

## The expected behavior
Being able to move my cursor without selecting table cells.

## Impact of fix
It makes the table component usable on mobile

circlingthesun:
## Description

Currently tapping on cell in a table on mobile while the selection is in another cell creates a cell selection, this change ensures that it only happens when dragging.

Closes #7650

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -301,7 +301,11 @@ export function applyTableHandlers(
             );
           }
         } else {
-          tableObserver.$setAnchorCellForSelection(targetCell);
+          // Only set anchor cell for selection if this is not a simple touch tap
+          // Touch taps should not initiate table selection mode
+          if (event.pointerType !== 'touch') {
+            tableObserver.$setAnchorCellForSelection(targetCell);
+          }
         }
       });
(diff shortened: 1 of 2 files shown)


etrepum:
I think this is probably closer to the ideal behavior but scrolling removes the selection and with this scheme it's not possible to select more than a screen full of the table since it doesn't scroll internally with the selection

https://github.com/user-attachments/assets/892c0457-c20f-4958-a86b-0cd7cefcfb15

circlingthesun:
@etrepum this behaviour also happens on the current playground version:

https://github.com/user-attachments/assets/d1bb5129-97e9-4022-8318-4f5579e494bc

PR #7657, opened 2025-06-28
[lexical-clipboard] Inherit style when typing after pasting rich text

### LINKED ISSUE #7490: Bug: Selection changes when typing after pasting
When you paste some content from outside, say some text with some background color and different font size, and i just continue typing from there the selection should ideally contain those values and the new added text should look the same, but thats not the case

Lexical version: v0.30.0

## Steps To Reproduce

1. copy some content from outside, paste it
2. immediately type something

Link to code example:

## The current behavior

https://github.com/user-attachments/assets/06de5d5c-5607-45b1-8600-d80f22102f5c

## The expected behavior

https://github.com/user-attachments/assets/dc268f5c-a454-4471-8c1a-925601ccc5cd

## Impact of fix
This will just improve the overall UX for copy pasting

achaljhawar:
## Description

When pasting rich text content, the editor selection's active format and style are now updated to match the style of the last character of the pasted content.

This ensures that continuing to type after a paste operation will correctly inherit the formatting (e.g., bold, italics, background-color, font-size) from the pasted material.

Closes #7490 

### Before

https://github.com/user-attachments/assets/826cfda6-7994-49dd-8d82-0d9f658a0ffa

/automated-tests*

### After

https://github.com/user-attachments/assets/2d9f4731-90de-4ad3-99fd-73189779ff29


etrepum (commented):
It probably makes sense to move this code into its own function (called by `$insertGeneratedNodes`) and write some tests for it.

The NodeCaret APIs could be used to iterate through the nodes in the 'previous' direction until a TextNode or block element is found

achaljhawar:
@etrepum how exactly should I use the the node caret api here?

etrepum:
Probably something like this (untested):

```ts
// usage: nodeToInspect = $getClosestTextNodeInBlock($caretFromPoint(selection.anchor, 'previous'));
function $getClosestTextNodeInBlock<D extends CaretDirection>(anchorCaret: PointCaret<D>): null | TextNode {
  if ($isTextPointCaret(anchorCaret)) {
    return anchorCaret.origin;
  }
  const range = $getCaretRange(anchorCaret, $getChildCaret($getRoot(), 'next').getFlipped());
  for (const caret of range) {
    if ($isTextNode(caret.origin)) {
      return caret.origin;
    } else if ($isBlockElementNode(caret.origin)) {
      break;
    }
  }
  return null;
}
```

achaljhawar:
@etrepum I have implemented nodecaret apis update the logic as you said.

etrepum (commented):
The code all looks good and I confirmed that it works as expected in the playground, the only thing missing here is a unit or e2e test to show that it works (and to prevent future regressions)

achaljhawar:
> The code all looks good and I confirmed that it works as expected in the playground, the only thing missing here is a unit or e2e test to show that it works (and to prevent future regressions)

where exactly should I write the tests for this ? (the file path)

packages/lexical/src/__tests__/unit/HTMLCopyAndPaste.test.ts
```diff
@@ -107,6 +113,16 @@ describe('HTMLCopyAndPaste tests', () => {
...
             $insertDataTransferForRichText(dataTransfer, selection, editor);
+            if (testCase.plainTextInsert) {
+              editor.update(() => {
+                const newSelection = $getSelection();
+                invariant(
+                  $isRangeSelection(newSelection),
+                  'isRangeSelection(newSelection) for plainTextInsert',
+                );
+                newSelection.insertText(testCase.plainTextInsert);
+              });
```
etrepum:
```suggestion
              const newSelection = $getSelection();
              invariant(
                $isRangeSelection(newSelection),
                'isRangeSelection(newSelection) for plainTextInsert',
              );
              newSelection.insertText(testCase.plainTextInsert);
```

etrepum (approved):
Looks good! Nice work

PR #7659, opened 2025-06-30
[lexical] Bug Fix: Workarounds in $config protocol for loose inheritance

etrepum:
## Description

Workarounds for #7258 combined with loose inheritance (not compliant with ES6)

## Test plan

Can't really test this in-repo very easily since lexical is built as ES6

### DIFF
--- a/packages/lexical/src/LexicalEditor.ts
+++ b/packages/lexical/src/LexicalEditor.ts
@@ -44,11 +44,12 @@ import {
   getDefaultView,
   getDOMSelection,
   getStaticNodeConfig,
-  hasOwn,
+  hasOwnExportDOM,
+  hasOwnStaticMethod,
   markNodesWithTypesAsDirty,
 } from './LexicalUtils';
 import {ArtificialNode__DO_NOT_USE} from './nodes/ArtificialNode';
-import {$isDecoratorNode} from './nodes/LexicalDecoratorNode';
+import {$isDecoratorNode, DecoratorNode} from './nodes/LexicalDecoratorNode';
 import {LineBreakNode} from './nodes/LexicalLineBreakNode';
 import {ParagraphNode} from './nodes/LexicalParagraphNode';
 import {RootNode} from './nodes/LexicalRootNode';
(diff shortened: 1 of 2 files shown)


packages/lexical/src/LexicalEditor.ts
```diff
@@ -567,24 +569,27 @@ export function createEditor(editorConfig?: CreateEditorArgs): LexicalEditor {
...
-            if (!hasOwn(klass, method)) {
+          (['getType', 'clone'] as const).forEach((method) => {
+            if (!hasOwnStaticMethod(klass, method)) {
               console.warn(`${name} must implement static "${method}" method`);
             }
           });
-          if (!hasOwn(klass, 'importDOM') && hasOwn(klass, 'exportDOM')) {
+          if (
+            !hasOwnStaticMethod(klass, 'importDOM') &&
+            hasOwnExportDOM(klass)
```
etrepum, line 579: This warning wasn't even working before because it was checking the class rather than the prototype 🤦

packages/lexical/src/LexicalEditor.ts
```diff
@@ -567,24 +569,27 @@ export function createEditor(editorConfig?: CreateEditorArgs): LexicalEditor {
...
+          ) {
             console.warn(
               `${name} should implement "importDOM" if using a custom "exportDOM" method to ensure HTML serialization (important for copy & paste) works as expected`,
             );
           }
           if ($isDecoratorNode(proto)) {
-            if (!hasOwn(proto, 'decorate')) {
+            if (proto.decorate === DecoratorNode.prototype.decorate) {
```
etrepum, line 586: I think this one can be weak since it might make sense in some scenarios to subclass a DecoratorNode but not override its decorate method

zurfyx (approved):
Thank you for the quick follow ups as usual!

packages/lexical/src/LexicalUtils.ts
```diff
@@ -2040,11 +2057,14 @@ export function getStaticNodeConfig(klass: Klass<LexicalNode>): {
...
+    if (!hasOwnStaticMethod(klass, 'getType')) {
       klass.getType = () => ownNodeType;
     }
-    if (!hasOwn(klass, 'clone')) {
-      if (__DEV__) {
+    if (!hasOwnStaticMethod(klass, 'clone')) {
+      // TextNode.length > 0 will only be true if the compiler output
+      // is not ES6 compliant, in which case we can not provide this
+      // warning
+      if (__DEV__ && TextNode.length === 0) {
```
lilshady, line 2067: this is so smart ❤️

PR #7662, opened 2025-07-01
[lexical-code][lexical-code-shiki][lexical-markdown][lexical-playground][lexical-devtools] Feature: Experimental Shiki support for code highlighting

jeromew:
This PR creates a new `lexical-code-shiki` package that adds support for the [Shiki](https://shiki.style/) highlighting engine.

By default, [lexical-playground] still uses the Prism highlighting engine in [lexical-code]. New setting options in th playground make it possible to enable Shiki highlighting.

[lexical-code-shiki] implements dynamic support for all languages and themes supported by Shiki.

Prism support can be considered backward compatible except for one minor aspect enabling CodeBlocks to have dynamic loading of languages (see Breaking Changes below). This affects a small number of tests in [lexical-code] and [lexical-markdown].

The support is considered experimental in the sense that more work is expected on it before shiki support should be considered backwards compatible. Some things are still subject to change such as:
 - Build time preloading of syntax and theme modules
 - Restructuring shiki highlighted code blocks to be line-oriented (and having a gutter implementation for it)
 - Extracting prism to its own optional package (breaking change)

## Breaking Changes

CodeNow now applies the `data-highlight-language` DOM attribute after highlighting, rather than immediately. The `data-language` attribute behavior is unchanged. This post-highlighting setting also applies to `getIsSyntaxHighlightSupported()` and its underlying `__isSyntaxHighlightSupported` property. Previously this would have been set eagerly because all known Prism languages were known and loaded at build time, now it represents whether the syntax highlighting has occurred or not.

## Tests

All tests related to Prism have been left unmodified except those mentioned related to the Breaking changes.
Tests have been added to Shiki to control that all the Tab/Indent/Outdent cases pass.
Manual tests have been applied, but other Shiki tests that imply the tokenization of the code have been postponed until the restructuring of shiki highlighted code blocks to be line-oriented, as all the expected html snippets will need updating. These Shiki tests will be written by taking all the Prism tests as a starting point.
The Prism tests themselves will be kept as-is.


packages/lexical-code/src/CodeNode.ts
```diff
@@ -338,24 +381,42 @@ export class CodeNode extends ElementNode {
...
 
 export function $createCodeNode(
   language?: string | null | undefined,
+  theme?: string | null | undefined,
 ): CodeNode {
-  return $applyNodeReplacement(new CodeNode(language));
+  const node = new CodeNode();
+  node.setLanguage(language);
+  node.setTheme(theme);
+  return $applyNodeReplacement(node);
```
etrepum: This should generally go the other way around

```suggestion
  return $applyNodeReplacement(new CodeNode())
    .setLanguage(language)
    .setTheme(theme);
```

There's also the new `$create` API (will be in the next release any day now, but is in main) where you can do `$create(CodeNode)` instead of `$applyNodeReplacement(new CodeNode())`

packages/lexical-code/src/CodeNode.ts
```diff
@@ -34,44 +34,24 @@ import {
...
   $createCodeHighlightNode,
-  $getFirstCodeNodeOfLine,
   $isCodeHighlightNode,
 } from './CodeHighlightNode';
+import {$getFirstCodeNodeOfLine} from './FlatStructureUtils';
 
 export type SerializedCodeNode = Spread<
   {
     language: string | null | undefined;
+    theme: string | null | undefined;
```
etrepum: It doesn't make a runtime difference but as far as the types are concerned it is more correct to mark this as an optional property (language should've been optional as well)

```suggestion
    theme?: string | null | undefined;
```

packages/lexical-code/src/CodeHighlighterShiki.ts
```diff
@@ -49,164 +49,53 @@ import {
...
   defaultLanguage: DEFAULT_CODE_LANGUAGE,
-  tokenize(code: string, language?: string): (string | Token)[] {
-    return Prism.tokenize(
-      code,
-      Prism.languages[language || ''] || Prism.languages[this.defaultLanguage],
-    );
+  defaultTheme: DEFAULT_CODE_THEME,
+  tokenize(code: string, language?: string): Array<string | Token> {
+    // Not Implemented for Shiki
+    return [];
```
etrepum: Since prism and shiki are using separate Tokenizer interface definitions I don't think this needs to be there?

packages/lexical-code/src/CodeHighlighterShiki.ts
```diff
@@ -265,17 +154,44 @@ function codeNodeTransform(
...
+  if (isCodeLanguageLoaded(language)) {
+    if (!node.getIsSyntaxHighlightSupported()) {
+      node.setIsSyntaxHighlightSupported(true);
+    }
+  } else {
+    if (node.getIsSyntaxHighlightSupported()) {
+      node.setIsSyntaxHighlightSupported(false);
+    }
+    loadCodeLanguage(language, editor, nodeKey);
     return;
```
etrepum: returning here will cause a waterfall because it forces the code language to be loaded before the theme is loaded, they could be loaded concurrently

packages/lexical-code/src/FacadeShiki.ts
```diff
@@ -0,0 +1,217 @@
...
+  createHighlighterCoreSync,
+  getTokenStyleObject,
+  isSpecialLang,
+  isSpecialTheme,
+  stringifyTokenStyle,
+} from '@shikijs/core';
+import {createJavaScriptRegexEngine} from '@shikijs/engine-javascript';
+// pre-load necessary language modules: default language, markdown for markdown import/export
+import lang_javascript from '@shikijs/langs/javascript';
+import lang_markdown from '@shikijs/langs/markdown';
```
etrepum: Ideally we'd figure out how to make this something that the user can choose, although that does get a bit complicated as far as bundlers and npm dependencies go. I think we'd have to have a peer dependency on shikijs and the user could pass in the shiki instance. If not, we'd provide some default like this does now.

packages/lexical-code/src/FacadeShiki.ts
```diff
@@ -0,0 +1,217 @@
...
+
+function getDiffedLanguage(language: string) {
+  const DIFF_LANGUAGE_REGEX = /^diff-([\w-]+)/i;
+  const diffLanguageMatch = DIFF_LANGUAGE_REGEX.exec(language);
+  return diffLanguageMatch ? diffLanguageMatch[1] : null;
+}
+
+export function isCodeLanguageLoaded(language: string) {
+  const diffedLanguage = getDiffedLanguage(language);
+  const langId = diffedLanguage ? diffedLanguage : language;
```
etrepum:
```suggestion
  const langId = diffedLanguage || language;
```

packages/lexical-code/src/FacadeShiki.ts
```diff
@@ -0,0 +1,217 @@
...
+
+      // here we know that the language is loaded
+      // make sure the code is highlighed with the correct language
+      if (editor && codeNodeKey) {
+        editor.update(() => {
+          const codeNode = $getNodeByKey(codeNodeKey);
+          if ($isCodeNode(codeNode)) {
+            codeNode.setLanguage(language);
+            codeNode.setIsSyntaxHighlightSupported(true);
+          }
```
etrepum: I think these properties should've already been set

```suggestion
          if ($isCodeNode(codeNode) && codeNode.getLanguage() === language && !codeNode.getIsSyntaxHighlightSupported()) {
            codeNode.setIsSyntaxHighlightSupported(true);
          }
```

packages/lexical-code/src/FacadeShiki.ts
```diff
@@ -0,0 +1,217 @@
...
+  const langId = diffedLanguage ? diffedLanguage : language;
+  if (!isCodeLanguageLoaded(langId)) {
+    const languageInfo = bundledLanguagesInfo.find(
+      (desc) =>
+        desc.id === langId || (desc.aliases && desc.aliases.includes(langId)),
+    );
+    if (languageInfo) {
+      // in case we arrive here concurrently (not yet loaded language is loaded twice)
+      // shiki's synchronous checks make sure to load it only once
+      await shiki.loadLanguage(languageInfo.import());
```
etrepum: using an await will force this to get deferred to the next microtask and cause some extra rendering, there should really be something that will use promises only when loading is in flight. similar to how promises work with react suspense.

packages/lexical-code/src/FacadeShiki.ts
```diff
@@ -0,0 +1,217 @@
...
+) {
+  if (!isCodeThemeLoaded(theme)) {
+    const themeInfo = bundledThemesInfo.find((info) => info.id === theme);
+    if (themeInfo) {
+      await shiki.loadTheme(themeInfo.import());
+      if (editor && codeNodeKey) {
+        editor.update(() => {
+          const codeNode = $getNodeByKey(codeNodeKey);
+          if ($isCodeNode(codeNode)) {
+            codeNode.setTheme(theme);
```
etrepum: There should be some way to do this only once when the theme transitions from a not loaded to a loaded state to avoid running the transforms more times than necessary

```suggestion
            codeNode.markDirty();
```

jeromew:
@etrepum thanks for the comments I will look into them. For now I am blocked on the build process in scripts/build.js that seems to not like my dynamic imports. I will try to find time to look into it but am not familiar with this script and the extent to which I can change it.

i think the problem arises when the build script tries to bundle https://github.com/shikijs/shiki/blob/main/packages/shiki/src/langs-bundle-full.ts and https://github.com/shikijs/shiki/blob/main/packages/shiki/src/themes.ts which have all the dynamic imports for all shiki languages

etrepum:
Just do it without the async imports for now and that can be handled separately.

packages/lexical-code/src/CodeNode.ts
jeromew: ok i went for chaining on the $create API

packages/lexical-code/src/FacadeShiki.ts
jeromew: I am not sure I get the full extent of your comment. I removed the await and replaced it with 
`shiki.loadLanguage(languageInfo.import()).then(..)`
or do you mean `languageInfo.import().then((module)=>shiki.loadLanguage(module); ..)` ?

What I found about react suspense is 

> React catches that promise, pauses rendering, and shows the fallback UI you provided to <Suspense> . When the promise resolves, React tries rendering again. This is possible because in JavaScript, you can synchronously stop a function by throwing. React leverages this to "pause" rendering until the data is ready.

can you clarify what you mean by "there should really be something that will use promises only when loading is in flight. similar to how promises work with react suspense." ?

currently, CodeNodeTransform 
 - fire-and-forget the loadAsset(language or theme) when they're not loaded + bails out if either needed loading. After loading, codeNode is marked as dirty
 
I could also not bailout after observing that loading is needed and render the code with lang "plain" + theme "none" which are always loaded

packages/lexical-code/src/FacadeShiki.ts
jeromew: yes, `codeNode.markDirty();` is sufficient indeed. I am not sure if your comment refers to an additional optimization that could be done to lower the number of transform applications between the time when we receive the setTheme & the time we know the theme has been loaded.

I am not sure I am fully aware of what is the overhead difference between `codeNode.setTheme(theme);` and `codeNode.markDirty();`.

scripts/build.js
```diff
@@ -40,10 +41,33 @@ const modulePackageMappings = Object.fromEntries(
...
+function getShikiAssets(assetType) {
+  return glob
+    .sync(
+      path.resolve(
+        path.dirname(__dirname),
+        'node_modules/@shikijs/' + assetType + '/dist/*.mjs',
+      ),
+    )
+    .map((p) => path.basename(p, '.mjs'));
+}
```
etrepum: Based on the extended test suite failures it looks like there might be an issue with a windows path in one of the build script changes. Probably related to glob and path.resolve. It likely needs `{windowsPathsNoEscape: true}` but possibly also conversion of the `\` back to `/`. You can check other uses of glob in the repo for examples

packages/lexical-code-shiki/src/FlatStructureUtils.ts
etrepum, line 1: This file looks almost exactly the same as the one in `@lexical/code`, so I think we should just have that one

packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts
```diff
@@ -944,15 +945,15 @@ describe('LexicalCodeNode tests', () => {
...
+          '<code spellcheck="false" data-language="javascript" dir="ltr"><span data-lexical-text="true">const lexical = "awesome"</span></code>',
         );
         registerRichText(editor);
         registerTabIndentation(editor);
         registerCodeHighlighting(editor);
         await Promise.resolve(undefined);
         // after transforms
         expect(testEnv.innerHTML).toBe(
-          '<code spellcheck="false" data-language="javascript" data-highlight-language="javascript" dir="ltr" data-gutter="1"><span data-lexical-text="true">const</span><span data-lexical-text="true"> lexical </span><span data-lexical-text="true">=</span><span data-lexical-text="true"> </span><span data-lexical-text="true">"awesome"</span></code>',
+          '<code spellcheck="false" data-language="javascript" dir="ltr" data-gutter="1" data-highlight-language="javascript"><span data-lexical-text="true">const</span><span data-lexical-text="true"> lexical </span><span data-lexical-text="true">=</span><span data-lexical-text="true"> </span><span data-lexical-text="true">"awesome"</span></code>',
```
etrepum: changing this test to use `expectHtmlToBeEqual` from `lexical/src/__tests__/utils` would make it more resilient to this semantically irrelevant change of attribute order

packages/lexical-playground/src/Editor.tsx
```diff
@@ -38,7 +38,8 @@ import AutocompletePlugin from './plugins/AutocompletePlugin';
 import AutoEmbedPlugin from './plugins/AutoEmbedPlugin';
 import AutoLinkPlugin from './plugins/AutoLinkPlugin';
 import CodeActionMenuPlugin from './plugins/CodeActionMenuPlugin';
-import CodeHighlightPlugin from './plugins/CodeHighlightPlugin';
+import {default as CodeHighlightPrismPlugin} from './plugins/CodeHighlightPlugin';
+import {default as CodeHighlightShikiPlugin} from './plugins/CodeHighlightShikiPlugin';
```
etrepum:
```suggestion
import CodeHighlightPrismPlugin from './plugins/CodeHighlightPlugin';
import CodeHighlightShikiPlugin from './plugins/CodeHighlightShikiPlugin';
```

This could follow the convention of the other playground plugins (not a big fan of default exports but consistency makes sense). Optionally CodeHighlightPlugin could be renamed and refactored to match how it's now used. The playground itself is not a published API so no need for backwards compatibility.

etrepum (commented):
Overall I think this approach makes sense, I will probably do another pass this weekend to look closer for potential issues/races with the lazy loading. Very solid work, I think people are really going to like having this option!

packages/lexical-code-shiki/src/FlatStructureUtils.ts
jeromew: I think it will depend on what we decide regarding the Lexical Node layout for shiki. If we do not keep the Flat layout, and go with something more line oriented then this file will change name and will have heavy modifications adapting to the new layout structure traversal.
BTW: you are right currently it is exactly the same file.

packages/lexical-playground/src/Editor.tsx
jeromew: ok I renamed CodeHighlightPlugin to CodeHighlightPrismPlugin and followed the plugin import convention there.

maybe I misunderstood what you meant by "refactored to match how it's now used" ; do you mean creating a lexical-code-prism module ?

packages/lexical-code-shiki/src/FlatStructureUtils.ts
jeromew: This file was coming from my first re-organization where I thought it could be useful in the future to have traversal helpers for different kind of Lexical Node Layouts/Hierarchies (I see the CodeBlocks as a sort of Island of structure) in order to be able to implement Code Highlighing plugins more easily : Choose a Lexical Node Layout, Implement a Facade spitting out this layout, and bind it to the corresponding Highlighter. But that is maybe a pipe dream and more of a constraint than anything else.

I thought FlatStructureUtils.ts could grow to implement generic helpers associated with the Array<LineBreakNode|TabNode|CodeHighlightNode> structure or in a distant future any rather Flat structure.

This "FlatStructureUtils.ts" file is not very big and could as well be integrated inside the CodeHighlighter.js.

Before my re-org it was split between CodeHighlighter.js and CodeHighlightNode.js ; tell me what you prefer.

BTW the only real difference between Shiki & Prism, apart from the Facade, is 
 - the instanciation of the tokenizer
 - the handling of theme attr for shiki
 - the equality test for the diff algorithm because with shiki attr.style must be taken into account

jeromew: it seems github swallowed my message.
For now, I removed FlatStructureUtils.ts from lexical-code-shiki and now use the one from lexical-code. I realized that the 4 functions inside are exported by lexical-code anyway for Prism backward compatibility so we may as well use them for immediate readability until the Lexical Node CodeBlock Layout is maybe changed to a Line based approach.

packages/lexical-playground/src/Editor.tsx
etrepum: The only refactoring I meant was also renaming the function that was exported to match the module name and it looks like you did that

packages/lexical-code-shiki/src/FlatStructureUtils.ts
etrepum: The functions to navigate the structure of CodeNode (whether flat or a line oriented approach) I think are fine to all have in `@lexical/code`, the things we really want to isolate in `@lexical/code-shiki` are the direct dependencies on shikijs. When we move prism out of the `@lexical/code` core I think those utilities should stay put.

etrepum (dismissed):
I think this is in a good state! I do like the idea of a line-based approach for shiki, which will certainly help with other UI affordances like making the gutters correct, but I think that can be handled as a follow-up. We might want to do some additional doc changes to note that this package is experimental and we expect to change the internals for that purpose.

Any thoughts on adding this `@lexical/code-shiki` engine? /cc @ivailop7 @zurfyx 

The big highlights for me are:
* Shiki is a maintained dependency
* It leverages a standard TextMate grammar format for its syntax (same as vscode)
* Lazy loading of language and theme files without too many hacks and without risk of breaking prism

Some follow-up items to consider:
* Extracting prism to its own optional package (breaking change)
* Build time preloading of syntax and theme modules
* Restructuring shiki highlighted code blocks to be line-oriented (and having a gutter implementation for it)

etrepum:
One thing that would be helpful while this gets some feedback from other maintainers would be to reformat the PR title/description to match the [pull request template](https://raw.githubusercontent.com/facebook/lexical/refs/heads/main/.github/pull_request_template.md)

It would also be good to note the (minor) breaking change prominently, typically with something like this:

```
## Breaking Changes

The `data-highlight-language` DOM attribute is now applied to CodeNode after highlighting, rather than immediately. The `data-language` attribute behavior is unchanged.
```

fantactuka:
> Some follow-up items to consider

One other thing we've been missing is just pure `<code>` element that has no dependencies on highlighting. E.g., with code as part of rich text package, and highlighting being a separate package that handles parsing/tokenization using 3p libraries.

etrepum:
> One other thing we've been missing is just pure `<code>` element that has no dependencies on highlighting

This PR is a big step in that direction, it decoupled CodeNode from Prism and any highlighting dependencies (which is why the timing of setting the highlight language attribute changed). The remaining step would primarily be to break compatibility by moving the Prism code to its own package.

etrepum:
You can test out the playground with shiki enabled here:
https://lexical-playground-git-fork-jeromew-code-shiki-fbopensource.vercel.app/?isCodeShiki=true

You can see it without any highlighting plugin enabled here:
https://lexical-playground-git-fork-jeromew-code-shiki-fbopensource.vercel.app/?isCodeHighlighted=false

vadimkantorov:
> This PR is a big step in that direction, it decoupled CodeNode from Prism and any highlighting dependencies

Awesome :)

Like so, lexical-markdown can import triple-backticks and still render it maybe even without dependencies on Prism/Shiki (if highlighting is not needed, just some display of block of code) or even on lexical-code

etrepum:
CodeNode will remain in lexical-code, eventually prism support will be moved to an optional package.

PR #7666, opened 2025-07-03
[lexical-link] Chore: Fix test names

noamzaks:
Will look later for other possible places with this sort of issue


packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -62,7 +62,7 @@ describe('LexicalLinkNode tests', () => {
       expect(() => $createLinkNode('')).toThrow();
     });
 
-    test('LineBreakNode.clone()', async () => {
+    test('LineNode.clone()', async () => {
```
```suggestion
    test('LinkNode.clone()', async () => {
```

noamzaks:
Thanks for the quick review.

I'm quite busy migrating my app to Lexical (a smoother process than expected!) so it might be best to merge now and hopefully I can look later to try and find similar cases.

Issue #6728, opened 2024-10-14
Feature: Add Maximum Change Limit and Optimization for Lexical History Plugin


dineug:
## Description

After reading [this article](https://emergence-engineering.com/blog/lexical-prosemirror-comparison), I traced where memory usage increases and found that it differs depending on whether the Lexical history plugin is enabled in the production build environment.

It would be helpful if the Lexical history plugin had an option to set a maximum limit on change history. Additionally, optimizing how changes are stored could improve performance.

### Use HistoryPlugin
![Lexical-JSHeapUsedSize](https://github.com/user-attachments/assets/bd4a69cd-be48-4fbf-ba1e-96b1e0d9ba6f)

### None HistoryPlugin
![Lexical-JSHeapUsedSize](https://github.com/user-attachments/assets/fb2d836b-51bd-4076-8588-78e2591189ad)

## Impact

This feature would enhance the user experience when editing large documents by providing smoother performance.

etrepum:
It is possible to manage the history size without any additions, since the data structures are public. Something like this (untested):

```ts
function pruneHistoryListener({redoStack, undoStack}: HistoryState, maxHistory: number): UpdateListener {
  return () => {
    redoStack.length = Math.max(0, Math.min(maxHistory, redoStack.length));
    undoStack.length = Math.max(0, Math.min(maxHistory - redoStack.length, undoStack.length));
  };
}

function LimitedHistoryPlugin(props: {
  delay?: number = 1000;
  maxHistory?: number = 50;
  externalHistoryState?: HistoryState;
}) {
  const historyState = useMemo(() => externalHistoryState || createEmptyHistoryState(), [externalHistoryState]);
  const [editor] = useLexicalComposerContext();
  useEffect(() => {
    return mergeRegister(
      registerHistory(editor, historyState, delay),
      editor.registerUpdateListener(pruneHistoryListener(historyState, Math.max(0, maxHistory))),
    );
  }, [editor, historyState, delay, maxHistory]);
  return null;
}
```

etrepum:
Of course if you were looking for something more precise about memory usage you'd probably want to consider the number of nodes in history (weight each history entry by the size of its nodeMap) rather than the number of history entries. Creating a more efficient data structure than simply referencing all EditorStates is an even better option, but with a lot more complexity.

dineug:
I tested it based on maxHistory 100, and it was very effective!
Thank you! @etrepum 

[...]

PR #7674, opened 2025-07-06
[WIP][lexical] Feature: GenMap - A generational copy-on-write NodeMap implementation

etrepum:
## Description

Replace the naïve Map implementation of NodeMap with a version optimized for structural sharing.

GenMap has two "modes":

* Readonly Proxy - when `_prev` is set to some previous GenMap it proxies all read operations to that object. This is used to avoid any unnecessary work when there are no changes to state.
* Mutable - On first mutation it sets `_prev` to undefined and directly performs read and write operations

In Mutable mode there are two Maps:

* `_old` - A read-only snapshot of some previous state (equivalent to an old Map based NodeMap), generated by the last collection
* `_nursery` - The mutable working set of nodes (or tombstones) that have changed from `_old`. Tombstones are used to mark deletions of nodes that are still present in `_old` but will be removed at the next collection

When transitioning to Mutable mode one of two things will happen:

* A collection - this generates the next `_old` by creating a Map from every entry from `_prev`

```ts
this._old = new Map(prev.entries());
this._nursery = new Map();
```

* A copy - the `_old` from `_prev` is referenced (no copy) and `_nursery` from `_prev` is copied 

```ts
this._old = prev._old;
this._nursery = new Map(prev._nursery);
```

This is an optimization because the cost of copying a Map is linear in both memory and time. In large documents, most nodes are not updated on each update cycle, so we are only paying the cost to copy the smaller nursery. Occasionally we pay the full cost to do a collection, but this is amortized since we don't do it often.

Related:
* #6728
* https://emergence-engineering.com/blog/lexical-prosemirror-comparison
* [discord discussion](https://discord.com/channels/953974421008293909/964928369793843230/1303854088596881510)

## Test plan

All existing suites pass

## TODO

- [ ] Benchmarks
- [ ] More specific tests

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -8,7 +8,6 @@
 
 import {
   $caretFromPoint,
-  $cloneWithProperties,
   $createParagraphNode,
   $getAdjacentChildCaret,
   $getCaretInDirection,
@@ -32,6 +31,7 @@ import {
   $setState,
   $splitAtPointCaretNext,
   type CaretDirection,
+  cloneEditorState,
   type EditorState,
   ElementNode,
   type Klass,
@@ -535,15 +535,15 @@ export function $restoreEditorState(
   editorState: EditorState,
 ): void {
   const FULL_RECONCILE = 2;
-  const nodeMap = new Map();
   const activeEditorState = editor._pendingEditorState;
 
-  for (const [key, node] of editorState._nodeMap) {
-    nodeMap.set(key, $cloneWithProperties(node));
-  }
-
   if (activeEditorState) {
-    activeEditorState._nodeMap = nodeMap;
+    activeEditorState._nodeMap = cloneEditorState(editorState)._nodeMap;
+    editor._cloneNotNeeded.clear();
+    editor._dirtyLeaves = new Set();
+    editor._dirtyElements.clear();
+  } else {
+    editor._editorState = editorState;
   }
 
   editor._dirtyType = FULL_RECONCILE;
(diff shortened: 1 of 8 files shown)


packages/lexical-utils/src/index.ts
```diff
@@ -535,15 +535,15 @@ export function $restoreEditorState(
   editorState: EditorState,
 ): void {
   const FULL_RECONCILE = 2;
-  const nodeMap = new Map();
   const activeEditorState = editor._pendingEditorState;
 
-  for (const [key, node] of editorState._nodeMap) {
-    nodeMap.set(key, $cloneWithProperties(node));
-  }
```
etrepum: I couldn't figure out why it was done this way, there aren't really any tests that exercise this properly. I could only guess that this was a bodge that would be better fixed by updating `_cloneNotNeeded` accordingly?

packages/lexical-utils/src/index.ts
```diff
@@ -535,15 +535,15 @@ export function $restoreEditorState(
...
-
   if (activeEditorState) {
-    activeEditorState._nodeMap = nodeMap;
+    activeEditorState._nodeMap = cloneEditorState(editorState)._nodeMap;
+    editor._cloneNotNeeded.clear();
+    editor._dirtyLeaves = new Set();
+    editor._dirtyElements.clear();
+  } else {
+    editor._editorState = editorState;
```
etrepum: The code didn't do anything at all if there was no active state, this seemed like the right thing to do? Not really sure without good tests either way.

Issue #7675, opened 2025-07-06
Feature: Introduce a linear Position API for Lexical Nodes


salihudickson:
## Description

I'd like to propose that each node track its start position and offset, as this will help make range selections more accurate.

## Motivation
tldr; Selecting an entire line with a triple click causes the selection to sometimes bleed into the next node. If we knew the absolute start and end positions of each node then we can more accurately normalize selections and eliminate this problem.

I was taking a look at an issue #7592, which points out that using a triple click in a code block causes all of the content in the code block to be selected instead of just the line that was triple clicked. After poking around a bit, turns out this behavior is caused by 2 PRs ([#4512](https://github.com/facebook/lexical/pull/4512) and [#7055](https://github.com/facebook/lexical/pull/7055) that were added to address the fact that triple clicking on a specific line can sometimes cause the selection to bleed into the next line. This does not show up visually, you only notice when you try to add some text formats, like Headings or bullet lists, and It's most prevalent on google chrome and other chromium based browsers.

But the PRs referenced above do not actually solve this problem, plus they have a side effect (as referenced above). After taking a closer look, I realized that the problem is caused by the fact that when a selection is made on the text editor, Lexical does some normalizing to try to remove nodes that seem like they have been selected but when you run some checks it's clear that they have not been selected. But it has to be conservative with these checks to ensure nodes that have actually been selected are not removed, so It assumes that if the selection selects any part of a node and that node has some content, then the entire node has been selected. And this is not always the case, for example when you triple click on a paragraph node, it sometimes selects everything in that node up until the start of the next node, and with the current behavior, lexical includes the whole next node. So I believe we need a way to allow the node selection to be a bit more certain when normalizing selections. 

My solution to this would be to have each node track its absolute start position (starting at 0 from the start of the root node) and its offset, we can then provide a separate normalized selection function that uses these to try to ensure that selections do not include nodes that were not intended. We can then allow users choose which function they want to use depending on how conservative they want the normalization to be.

## Alternative solutions
I considered walking through the Lexical Node tree to determine the position of each node when needed, but for obvious reasons this will become very inefficient when more than just a few hundred nodes are on the document.

#### a video of the triple click bug i am referring to above

https://github.com/user-attachments/assets/79b5767b-428e-443f-bba4-7bfac5bc50df

#### a video demonstrating how the bug has been resolved by properly normalizing the selection with the approach I described above.

https://github.com/user-attachments/assets/89371099-69fd-43bd-8e1a-14286143f227

[...]

PR #7677, opened 2025-07-08
[lexical-code][lexical][lexical-playground][lexical-list][lexical-rich-text] Feature: adjust text selection by node position before applying block-level formats

SalihuDickson:
### Current Behavior

When you triple-click to select an entire line in the editor, the start of the next line is sometimes unintentionally included in the selection. This happens due to the browser’s default selection behavior and is not visibly noticeable to the user. However, it becomes apparent when applying block-level formats — since these styles affect the entire line, the formatting gets applied to both the intended line and part of the next one. This issue is most common in Chromium-based browsers.
How This PR Fixes It.
https://github.com/user-attachments/assets/2b59e59c-155b-4f79-8896-d7e88ce9cb12

### How this PR addresses this
This PR addresses the issue by normalizing the selection before applying block-level formats. It calculates the absolute numeric position of the selection’s end, then filters out any nodes whose start position is greater than or equal to that endpoint. This ensures that only the intended line receives the formatting.
https://github.com/user-attachments/assets/832a1350-622b-4418-be18-29fbceb67455

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -2701,6 +2701,95 @@ function $validatePoint(
   }
 }
 
+function $traverseNodes(
+  callback: (
+    node: LexicalNode,
+    absoluteOffset: number,
+  ) => {found: boolean; newOffset: number},
+) {
+  const root = $getRoot();
+  let absoluteOffset = 0;
+
+  const traverse = (node: LexicalNode): boolean => {
+    const result = callback(node, absoluteOffset);
+
+    if (result.found) {
+      absoluteOffset = result.newOffset;
+      return true; // Found - stop traversal
+    }
+
+    if ($isTextNode(node)) {
+      absoluteOffset += node.getTextContent().length;
+    } else if ($isElementNode(node)) {
+      return node.getChildren().some((child) => traverse(child));
+    }
+
+    return false;
+  };
+
+  root.getChildren().some((child) => traverse(child));
+  return absoluteOffset;
+}
+
+function $calculateSelectionPositions() {
+  const selection = $getSelection();
+
+  if (!$isRangeSelection(selection)) {
+    return {from: 0, to: 0};
+  }
+
+  const anchorPos = $calculateAbsolutePosition(selection.anchor);
+  const focusPos = $calculateAbsolutePosition(selection.focus);
+
+  return {
+    from: Math.min(anchorPos, focusPos),
+    to: Math.max(anchorPos, focusPos),
+  };
+}
+
+function $calculateAbsolutePosition(target: PointType | LexicalNode): number {
+  const targetKey =
+    target instanceof Object && 'key' in target ? target.key : target.getKey();
+  const targetOffset =
+    target instanceof Object && 'offset' in target ? target.offset : 0;
+
+  return $traverseNodes((node, currentOffset) => {
+    if (node.getKey() === targetKey) {
+      return {
+        found: true,
+        newOffset: currentOffset + targetOffset,
+      };
+    }
+    return {found: false, newOffset: currentOffset};
+  });
+}
+
+export function $normalizeSelectionByPosition(): null | BaseSelection {
+  const selection = getActiveEditorState()._selection;
+
+  if ($isRangeSelection(selection)) {
+    const {to} = $calculateSelectionPositions();
+    const allNodes = selection.getNodes();
+
+    // Filter nodes to make sure the all nodes are within the selected range
+    const filteredNodes = allNodes.filter((node) => {
+      const nodeStart = $calculateAbsolutePosition(node);
+      return nodeStart < to - 1 && $isTextNode(node);
+    });
+
+    if (filteredNodes.length > 0) {
+      const firstNode = filteredNodes[0];
+      const lastNode = filteredNodes[filteredNodes.length - 1];
+      const lastNodeEndOffset = lastNode.getTextContent().length;
+
+      selection.anchor.set(firstNode.getKey(), 0, 'text');
+      selection.focus.set(lastNode.getKey(), lastNodeEndOffset, 'text');
+    }
+  }
+
+  return selection;
+}
+
 export function $getSelection(): null | BaseSelection {
   const editorState = getActiveEditorState();
   return editorState._selection;
--- a/packages/lexical/src/index.ts
+++ b/packages/lexical/src/index.ts
@@ -228,6 +228,7 @@ export {
   $isBlockElementNode,
   $isNodeSelection,
   $isRangeSelection,
+  $normalizeSelectionByPosition as $normalizeSelectionByPosition__EXPERIMENTAL,
 } from './LexicalSelection';
 export {$parseSerializedNode, isCurrentlyReadOnlyMode} from './LexicalUpdates';
 export {
--- a/packages/lexical-code/src/CodeHighlighter.ts
+++ b/packages/lexical-code/src/CodeHighlighter.ts
@@ -34,6 +34,7 @@ import {
   $isTabNode,
   $isTextNode,
   $normalizeCaret,
+  $normalizeSelectionByPosition__EXPERIMENTAL as $normalizeSelectionByPosition,
   $setSelectionFromCaretRange,
   COMMAND_PRIORITY_LOW,
   INDENT_CONTENT_COMMAND,
@@ -646,7 +647,7 @@ function $handleTab(shiftKey: boolean): null | LexicalCommand<void> {
 }
 
 function $handleMultilineIndent(type: LexicalCommand<void>): boolean {
-  const selection = $getSelection();
+  const selection = $normalizeSelectionByPosition();
   if (!$isRangeSelection(selection) || !$isSelectionInCode(selection)) {
     return false;
   }
--- a/packages/lexical-list/src/formatList.ts
+++ b/packages/lexical-list/src/formatList.ts
@@ -16,6 +16,7 @@ import {
   $isRangeSelection,
   $isRootOrShadowRoot,
   $normalizeCaret,
+  $normalizeSelectionByPosition__EXPERIMENTAL as $normalizeSelectionByPosition,
   $setPointFromCaret,
   ElementNode,
   LexicalNode,
@@ -63,7 +64,7 @@ function $isSelectingEmptyListItem(
  * @param listType - The type of list, "number" | "bullet" | "check".
  */
 export function $insertList(listType: ListType): void {
-  const selection = $getSelection();
+  const selection = $normalizeSelectionByPosition();
 
   if (selection !== null) {
     let nodes = selection.getNodes();
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
@@ -27,6 +27,7 @@ import {
   $getSelection,
   $isRangeSelection,
   $isTextNode,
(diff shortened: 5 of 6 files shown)


salihudickson:
@etrepum, this is the PR i spoke about in #7675. Please check it out.

etrepum:
A simpler and more efficient approach here would be to have a function to normalize a RangeSelection such that the points are moved closer to each other so long as it can be done in a "zero distance" manner. It's not necessary to know the absolute positions, only the relative positions, and even then you don't have to know how much is between them only which direction they are from each other.

Either way, the most important part of this work would be to write a few tests to show that it works as expected in all of the scenarios it's supposed to cover. The actual change itself is less important than the tests that define the behavior and prevent it from regressing in the future.

etrepum:
IIRC the normalization that currently happens already has a right-bias so mostly it's the focus point that would need to move (assuming the selection is already normalized to be forwards such that the focus is not before the anchor).

salihudickson:
@etrepum, yeah that makes sense. Thanks for the suggestion.

And I'll be sure to add a bunch of test cases.

etrepum:
The tests are still failing, you can run unit tests locally with `npm run test-unit`

salihudickson:
> The tests are still failing, you can run unit tests locally with `npm run test-unit`

😅 yeah I've found the failing tests and I'm working on a workaround

salihudickson:
Hey @etrepum, can you please take a look at this? 

I've fixed the logic that was breaking the tests, and I've added unit and e2e regression tests.

etrepum:
It looks like there are conflicts with main that have to be resolved before the tests can be approved to run

salihudickson:
> It looks like there are conflicts with main that have to be resolved before the tests can be approved to run

Just did a rebase, the tests should be able to run now.

salihudickson:
hey @etrepum, please can you authorize the integration tests, they seem to be failing on some platforms and I'm trying to debug in CI

etrepum (commented):
Using module mocks will make these tests fragile because it simulates an environment that doesn't conform to normal expectations. It would be much better to write tests that are able to set up the initial conditions appropriately so that mocks are not necessary.

salihudickson:
> Using module mocks will make these tests fragile because it simulates an environment that doesn't conform to normal expectations. It would be much better to write tests that are able to set up the initial conditions appropriately so that mocks are not necessary.

I used the mocks in the unit tests to try to isolate some logic in the main function. But I also have tests that check the results  for the same unit without mocks. But I can take out the tests that use mocks if you believe it's not ideal. I could separate the logic it into it's own function and test it that way.

etrepum:
I think the mock tests written in this way are a liability and we shouldn't use them in this way, if the functionality is sufficiently tested with normal unit tests I would just get rid of the mocks. If you're trying to test smaller units that aren't exported from the lexical package then you can export the functions from the internal modules that define them and import the internal modules directly from the tests, there are several unit test examples of this (e.g. `import {LexicalNode} from '../../LexicalNode'`)

salihudickson:
Sure @etrepum, thanks for the suggestions.

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -812,6 +818,607 @@ describe('LexicalSelection tests', () => {
...
+          $getRoot().append(paragraph);
+
+          const startTime = performance.now();
+          const result = $getCharacterOffsetBetweenNodes(nodes[0], nodes[99]);
+          const endTime = performance.now();
+
+          expect(result).toBeGreaterThan(0);
+          expect(endTime - startTime).toBeLessThan(100); // Should complete in reasonable time
+        });
+      });
```
etrepum: Benchmarks are not a good fit for unit tests, it would be much better to have tests that cover more complicated structures such as inline ElementNode (e.g. LinkNode), DecoratorNode, and possibly also define what the behavior should be if the two points cross a shadow root boundary (e.g. different TableCellNode)

packages/lexical/src/index.ts
```diff
@@ -220,14 +220,17 @@ export {
...
   $getCharacterOffsets,
   $getPreviousSelection,
   $getSelection,
+  $getSelectionLength,
   $getTextContent,
   $insertNodes,
   $isBlockElementNode,
   $isNodeSelection,
   $isRangeSelection,
+  $normalizeSelectionByPosition as $normalizeSelectionByPosition__EXPERIMENTAL,
```
etrepum: I don't think it's useful to have the `__EXPERIMENTAL` naming scheme, this usually just results in the name being around for much longer than it should for backwards compatibility reasons. The best way to mark functions as experimental is in the API documentation (generated from the comment before the function)

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
 
+export function $getCharacterOffsetBetweenNodes(
+  startNode: LexicalNode,
+  targetNode: LexicalNode,
+): number | null {
+  const parentNode = $getCommonAncestor(startNode, targetNode);
+  if (
+    !parentNode ||
+    !parentNode.commonAncestor ||
+    !('getChildren' in parentNode.commonAncestor)
```
etrepum:
```suggestion
    !$isElementNode(parentNode.commonAncestor)
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
+  ) {
+    return targetNode.getKey() === startNode.getKey() ? 0 : null;
```
etrepum:
```suggestion
    return targetNode.is(startNode) ? 0 : null;
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
+  }
+
+  let offset = 0;
+  let hasStarted = false;
+
+  const traverse = (node: LexicalNode): boolean => {
+    if (node.getKey() === targetNode.getKey()) {
+      return true;
+    } else if (!hasStarted && node.getKey() === startNode.getKey()) {
```
etrepum:
```suggestion
    if (node.is(targetNode)) {
      return true;
    } else if (!hasStarted && node.is(startNode)) {
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
+      hasStarted = true;
+    }
+
+    if (hasStarted && $isTextNode(node)) {
+      offset += node.getTextContent().length;
```
etrepum:
```suggestion
      offset += node.getTextContentSize();
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
+    } else if ($isElementNode(node)) {
+      return node.getChildren().some((child) => traverse(child));
+    }
+
+    return false;
+  };
+
+  // Start traversal from common ancestor
+  const found = (parentNode.commonAncestor as ElementNode)
```
etrepum:
```suggestion
  const found = parentNode.commonAncestor
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
+
+  const {anchor, focus} = selection;
+
+  // Use the higher offset for same node, otherwise use offset from last node
+  const nodeOffset =
+    anchor.key === focus.key
+      ? Math.max(anchor.offset, focus.offset)
+      : (anchor.key === lastNode.getKey() ? anchor : focus).offset;
+
+  return offset + nodeOffset;
```
etrepum: this seems a bit suspect, it's not necessarily the case that both points are text

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
+
+  return found ? offset : null;
+}
+
+export function $getSelectionLength() {
+  const selection = $getSelection();
+
+  if (!$isRangeSelection(selection)) {
+    return 0;
+  }
```
etrepum: It would make more sense for this to only accept the types of selection that it can produce a meaningful answer for

```suggestion
export function $getSelectionLength(selection: RangeSelection): number {
```

packages/lexical/src/LexicalSelection.ts
```diff
@@ -2711,6 +2712,131 @@ function $validatePoint(
...
+}
+
+export function $normalizeSelectionByPosition(): null | BaseSelection {
+  const selection = $getSelection();
+  const selectionLength = $getSelectionLength();
+  if (!selection || !selectionLength) {
+    return selection;
+  }
+
+  const allNodes = selection.getNodes() ?? [];
```
etrepum: getNodes always returns a LexicalNode[]

```suggestion
  const allNodes = selection.getNodes();
```

salihudickson:
Hey, @etrepum, thanks for the in-depth review. I'll go ahead and check out the NodeCaret API, and make the other necessary changes.

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
salihudickson: Sure, but the normalization doesn't really interact with most nodes, just the boundaries between them, that's why I didn't think it would be necessary to test for them. I did write a test for decorator nodes though. I'll just go ahead and remove the bench mark test. I really just added it cause the first iteration of my solution was pretty inefficient and was running a nested for loop.

packages/lexical/src/LexicalSelection.ts
salihudickson: The offset is 0 when the point it's in is not a text node.

salihudickson:
> Probably the best place to perform this normalization would be after a selection change due to a triple click is detected, rather than find every mutation that might be called after a triple click.

The issue only applies to block level formats, so those are the only mutations I'm applying it to. Normalizing after triple click won't fix the issue because the selection is retrieved with `getSelection()` before the formatting is applied and that contains the selection that spills into the next line

etrepum:
The problem *is* that the selection spills into the next line on triple click, when the expectation is that it would not.

salihudickson:
> The problem _is_ that the selection spills into the next line on triple click, when the expectation is that it would not.

That's true but that happens whenever you run the `$getSelection` function. You always get the selection that spills into the next line. So even if the selection is normalized on triple click, whenever the user applies so block format, the function that applies the format runs `$getSelection` and retrieves the spilled selection. This infact is why the other attempts at a solution have been ineffective, and just created side effects.

etrepum:
Here's the sequence of events:

- Triple click event happens, ignored by lexical
- Browser changes the selection to the whole element plus an overlap into the leading edge of the next element
- Lexical changes its selection based on the selectionchange to match the browser's selection
- Later on the user changes the block using this selection with overflow

If the lexical selection is corrected, then the problem goes away. Your fix does it in the last step and only works around the problem in the specific cases that are addressed in this PR. If the selection is fixed earlier, it fixes the problem more comprehensively

salihudickson:
@etrepum, I was under the impression that the selection was converted from a DOM selection whenever the `$getSelection` function is called. In light of your explanation, I have been able to rework the solution to only be applied when a triple click is detected.

salihudickson:
hey @etrepum, could you take another look at this PR?

etrepum:
It looks like the tests are not passing, I will take a look after the tests pass. When updating this PR with main you may have to make some additional changes because the HTML output has changed after #7727 (the ltr/rtl classes are gone and the dir attribute may be set differently or not at all)

PR #7680, opened 2025-07-09
fix(hashtag-plugin): exclude emoji keycap (from hashtag regex match)

### LINKED ISSUE #7678: Bug: Hashtag regex is matching hashtag emoji
Lexical version: 0.33.0

## Steps To Reproduce
Hard to reproduce without getting access to the Hootsuite codebase, which is a private GitHub.
1.
2.

Link to code example:

## The current behavior

Image

## The expected behavior

Image

Sahil-Sevda:
- **Current behavior**: The hashtag regex matches emoji keycap `#️⃣` (`#\uFE0F\u20E3`), causing rendering issues and incorrect hashtag detection in Chrome and Safari.
- **Change introduced**: Added a negative lookahead in the `getHashtagRegexString()` function to exclude emoji keycap sequences from matching.

Closes #7678 

### Before

- Typing `#️⃣` in Chrome/Safari breaks rendering and incorrectly triggers the hashtag plugin.

### After

- Typing `#️⃣` no longer triggers hashtag parsing.
- Regular hashtags (e.g., `#test123`) still work as expected.

### DIFF
--- a/packages/lexical-react/src/LexicalHashtagPlugin.ts
+++ b/packages/lexical-react/src/LexicalHashtagPlugin.ts
@@ -229,7 +229,9 @@ function getHashtagRegexString(): string {
   const hashtagAlpha = '[' + alpha + ']';
   const hashtagAlphanumeric = '[' + alphanumeric + ']';
   const hashtagBoundary = '^|$|[^&/' + alphanumeric + ']';
-  const hashCharList = '[' + hashChars + ']';
+  // const hashCharList = '[' + hashChars + ']';
+   // Prevent matching emoji hashtag: '#️⃣' == '#\uFE0F\u20E3'
+  const hashCharList = '(?!' + '[' + hashChars + ']' + '\\uFE0F\\u20E3)' + '[' + hashChars + ']';
 
   // A hashtag contains characters, numbers and underscores,
   // but not all numbers.


etrepum (commented):
Adding a test to show that this works would be great

Sahil-Sevda:
Adding two videos for reference:

- **Before Fix:** Demonstrates how the hashtag plugin was behaving previously.
- **After Fix:** Shows the corrected behavior with the updated hashtag regex logic.

These videos help validate the change visually.
[after.webm](https://github.com/user-attachments/assets/8a1ee721-bf71-40d5-8162-bf17dc3ed47e)

https://github.com/user-attachments/assets/f06e1d5c-d08f-4040-8cff-a77461b6009b

etrepum:
Tests are more important than videos, they show that it works now but more importantly they will detect if that behavior ever changes due to some other PR

etrepum:
As far as I can tell, this unit test passes on main without any change to the module other than to export `getHashtagRegexString`, so it's not clear if this fixes the problem or not

Sahil-Sevda:
Thanks for the review! Just to clarify:

- The root issue here is that the regex incorrectly matches emoji hashtags like `#️⃣`, which causes Lexical’s hashtag plugin to activate and leads to rendering issues in Chrome and Safari.
- This isn't easily reproducible with unit tests alone because the rendering bug is browser-specific and doesn't manifest in a jsdom environment.

My fix adds a negative lookahead in the `getHashtagRegexString()` to prevent emoji keycap sequences (`#️⃣` == `#\uFE0F\u20E3`) from being matched.

Although the test passes on `main`, the bug is reproducible in the browser (as reported here #7678  and this change prevents that faulty matching.

Let me know if you'd prefer a browser test instead — but given the complexity, I believe a regex-based unit test with a documented context should suffice.

etrepum:
I want a test that will fail on main and pass with this change. This test passes on main (after only adding the export), so it doesn’t really demonstrate any improvement. An e2e test that can show the problem in a browser would be ideal, but a unit test that shows that this change changes behavior in a positive way would be sufficient. Right now we have a test that doesn’t show any change between this code and main, so it’s not clear that the code change has any effect.

maahd-shahzad-hs:
Hi just wondering if there is an update on this? Did you decide on your testing strategy? I am the external dev who created the original bug report. Thanks for working on this btw! Let me know if I can help in any way or if you have any questions. But looking at the PR, it seems like you are on the right track. Doing a negative lookahead should solve this problem.

Sahil-Sevda:
Thanks for following up! I’m finalizing the testing strategy and will have an update soon.

Sahil-Sevda:
@maahd-shahzad-hs Could you please share the exact reproduction steps you used for this issue? I’ve tried to reproduce it with Jest and E2E tests in both Chrome and Safari, but haven’t had consistent results. Your input would be really helpful—thanks!

maahd-shahzad-hs:
Hi @Sahil-Sevda. Sorry for the delay in getting back to you. We were trying to figure out the best way for you to reproduce this without needing to access our private repos. We were trying to create a JS Fiddle and noticed something. I am not able to reproduce the issue in the [lexical playground here](https://playground.lexical.dev). 

Did you try reproducing the issue before fixing it. If yes, were you manually able to reproduce it on your end?

maahd-shahzad-hs:
I tested locally with our stack. Seems like your latest update has fixed the issue. I created the issue on July 8 and version 0.33.1 was released on July 10 according to npmjs.com. Our company is shutting down next week to relax and unwind so I won't be able to respond. We are all on PTO. But will update you when I get back. On a first look it seems like it's working now. So no action may be needed from you at this point. 

https://github.com/user-attachments/assets/7f2810a7-e238-4b19-8544-be07ffb25796
https://www.npmjs.com/package/lexical

maahd-shahzad-hs:
Ok I can confirm that V0.33.1 has resolved the issue on our end. So you probably don't need to merge this PR. Unless you see another reason to.

takuyakanbr:
Closing this as https://github.com/facebook/lexical/issues/7678 has already been closed

PR #7686, opened 2025-07-10
[lexical-selection] Bug Fix: Correct backward inversion for RTL

### LINKED ISSUE #7685: Bug: Selection movement incorrect for RTL entering decorator node
https://github.com/user-attachments/assets/7143b0f1-ec1d-4e3e-b5da-ca2e453a1c4e

Lexical version: v0.33.0

## Steps To Reproduce

1. Write two RTL paragraphs.
2. Add a divider in between both paragraphs.
3. Try to move selection from the end of the first paragraph forward (with the left arrow key) or from the start of the second paragraph backward (with the right arrow key).

## The current behavior

Some "weird" location is selected, and only after pressing the left key the divider is focused.

## The expected behavior

The divider should be focused.

## Impact of fix
RTL users.

noamzaks:
## Description
This updates the `$shouldOverrideDefaultCharacterSelection` to the correct "backward" value when inside an RTL element.

Closes #7685.

### Before

https://github.com/user-attachments/assets/7143b0f1-ec1d-4e3e-b5da-ca2e453a1c4e

### After

https://github.com/user-attachments/assets/90ba1799-9ab7-4329-95bb-a33cbe874cf3

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -1066,7 +1066,12 @@ test.describe.parallel('Selection', () => {
     );
   });
 
-  test('Select previous with RTL (DecoratorNode) #7685', async ({page}) => {
+  test('Select previous with RTL (DecoratorNode) #7685', async ({
+    page,
+    isPlainText,
+    isCollab,
+  }) => {
+    test.skip(isPlainText || isCollab);
     await page.keyboard.type('קצת מלל');
     await insertHorizontalRule(page);
     await page.keyboard.type('עוד');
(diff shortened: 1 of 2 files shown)


noamzaks:
As can be seen in the "After" video, exiting the decorator nodes should probably be inverted in some cases as well. However this is more tricky, as it's probably possible that before+after aren't the same direction, in which case it's unknown what should occur.

In any case, I think when before+after are both RTL it's expected that a right arrow from the decorator node will move to the _previous_ node instead of the next one and the left arrow will move to the _next_ node, where I think changes can be made to the lexical-rich-text/src/index.ts file in the appropriate command registrations.

I will probably open an issue (and hopefully PR) for this after #7685 is resolved.

etrepum (commented):
This looks like a reasonable approach, although it would be fantastic to add an e2e test to show that it does what it's supposed to. Should be fairly straightforward since this is reproducible without adding anything to the playground, there are other tests in `packages/lexical-playground/__tests__/e2e/Selection.spec.mjs` (but nothing that tests rtl behavior yet)

noamzaks:
I added a test case for this (and verified it fails before adding the change).
I had an issue using `moveToLineEnd` or `moveToLineBeginning` for some reason.

etrepum:
It’s likely that move line to end/beginning has the same class of issue with rtl that needs a different workaround. The e2e test helpers tend to use input rather than some back door to force a specific state. It does look like these tests are failing, maybe only on some specific browser/platform combinations.

noamzaks:
Looks like plain text and collab should be disabled for this test.

etrepum (approved):
Nice work, thank you!

noamzaks:
I think this is still broken when its in the end of the paragraph for instance. Investigating now.

PR #7693, opened 2025-07-15
[lexical-playground]Bug Fix: Image links lose link state when dragged

### LINKED ISSUE #7692: Bug: Non-text links lose link state when dragged
Lexical version: https://playground.lexical.dev/

## Steps To Reproduce

1. Insert an image

2. Add a link to the image

3. Drag the image to another location (click and hold the image itself, not the draggable tab on the left)

4. Observe link is now lost

## The current behavior
Image loses its link

## The expected behavior
Link is retained. This is similar to the behaviour in google docs.

## Impact of fix
Editors who do not implement the draggable tab are still able to move link-images without losing the link.

lilshady:
## Description
I guess this is kind of an edge case: When the image is wrapped within a Link Node, toggle the link upon drop

Closes #7692

### Before

https://github.com/user-attachments/assets/44abe44b-3ebf-4cfe-ae8c-d127623c13d0

### After

https://github.com/user-attachments/assets/b7913055-5454-4a3d-bb3a-711bd0477cd6

### DIFF
--- a/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
@@ -8,8 +8,18 @@
 
 import type {JSX} from 'react';
 
+import {
+  $isAutoLinkNode,
+  $isLinkNode,
+  LinkNode,
+  TOGGLE_LINK_COMMAND,
+} from '@lexical/link';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
-import {$wrapNodeInElement, mergeRegister} from '@lexical/utils';
+import {
+  $findMatchingParent,
+  $wrapNodeInElement,
+  mergeRegister,
+} from '@lexical/utils';
 import {
   $createParagraphNode,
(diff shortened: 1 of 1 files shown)


takuyakanbr (approved):
Would be good to add a test case in `Images.spec.mjs`

packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
```diff
@@ -324,6 +339,9 @@ function $onDrop(event: DragEvent, editor: LexicalEditor): boolean {
     }
     $setSelection(rangeSelection);
     editor.dispatchCommand(INSERT_IMAGE_COMMAND, data);
+    if (existingLink) {
+      editor.dispatchCommand(TOGGLE_LINK_COMMAND, existingLink.getURL());
```
etrepum, line 343: might be worth preserving the other link attributes too, `TOGGLE_LINK_COMMAND` can take an object with url, target, rel, title (although there's currently no utility function to extract a `({url: string} & LinkAttributes)` object from a LinkNode.

PR #7696, opened 2025-07-16
feat(list): add disableTakeFocusOnClick option to CheckListPlugin (#7689)

### LINKED ISSUE #7689: Feature: Add an option to disable focus on click in CheckListPlugin
## Description

In `checkList.ts`, `handleClick` explicitly focuses the node. This is a problem on mobile, as it makes it impossible to check or uncheck list items without causing the keyboard to automatically animate in.

### Use cases 
- Toggling checkboxes without focusing the editor on mobile.

### Alternative solutions
- We rolled out our own plugin, but it has broken on us with lexical updates in the last update.

## Impact
Adding a `disableTakeFocusOnClick?=false` parameter to plugin initialization would have no negative impact on existing users but would better support checklists on mobile.

mrbuddhu:
## Summary

Adds a `disableTakeFocusOnClick` option to the CheckListPlugin, allowing users to disable focusing the editor when a checklist item is clicked. This is especially useful on mobile, where focusing brings up the keyboard unnecessarily.

Fixes #7689

## Test Plan

- Pass `disableTakeFocusOnClick={true}` to the CheckListPlugin.
- On mobile, clicking a checklist item toggles it without focusing or opening the keyboard.
- Default behavior (focus on click) is unchanged for existing users.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -23,6 +23,7 @@ component_syntax=true
 module.name_mapper='^lexical$' -> '<PROJECT_ROOT>/packages/lexical/flow/Lexical.js.flow'
 module.name_mapper='^@lexical/clipboard$' -> '<PROJECT_ROOT>/packages/lexical-clipboard/flow/LexicalClipboard.js.flow'
 module.name_mapper='^@lexical/code$' -> '<PROJECT_ROOT>/packages/lexical-code/flow/LexicalCode.js.flow'
+module.name_mapper='^@lexical/code-shiki$' -> '<PROJECT_ROOT>/packages/lexical-code-shiki/flow/LexicalCodeShiki.js.flow'
 module.name_mapper='^@lexical/devtools-core$' -> '<PROJECT_ROOT>/packages/lexical-devtools-core/flow/LexicalDevtoolsCore.js.flow'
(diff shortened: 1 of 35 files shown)


mrbuddhu:
Signed CLA

ivailop7:
Not stealing the cursor focus upon clicking a checkbox should be the default behaviour IMO. 

Playing with the PR preview, it only works if I tick the checkbox that is on the same line as the caret, but not if I check/uncheck box 1, when the caret is on line 3, it still steals the focus.

packages/lexical-list/src/index.ts
```diff
@@ -278,3 +278,8 @@ export function insertList(editor: LexicalEditor, listType: ListType): void {
 export function removeList(editor: LexicalEditor): void {
   editor.update(() => $removeList());
 }
+
+export function registerCheckList(
+  editor: LexicalEditor,
+  options?: { disableTakeFocusOnClick?: boolean }
+): () => void;
```
etrepum, line 285: This shouldn't be here

```suggestion
```

etrepum (commented):
I think it would be more useful if it just worked in the expected way, especially on mobile. It's hard to review this without any live environment or tests that demonstrates the usage, it just adds a toggle but you can't turn it on without editing the code and running it locally.

For all I know there could be a separate issue related to [focus](https://lexical.dev/docs/concepts/selection#focus) if there happens to be an existing selection when this node is mutated but I can't quickly verify if that's the case or not.

PR #7698, opened 2025-07-17
fix(Image):Image Links loose link state when dragged

Jynx2004:
I guess this is kind of an edge case: When the image is wrapped within a Link Node,  add a parent link node to the image and then update the editor.

Closes https://github.com/facebook/lexical/issues/7692

### DIFF
--- a/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
@@ -8,11 +8,13 @@
 
 import type {JSX} from 'react';
 
+import {$createLinkNode, LinkNode} from '@lexical/link';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {$wrapNodeInElement, mergeRegister} from '@lexical/utils';
 import {
   $createParagraphNode,
   $createRangeSelection,
+  $getNodeByKey,
   $getSelection,
   $insertNodes,
   $isNodeSelection,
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
```diff
@@ -306,25 +309,67 @@ function $onDragover(event: DragEvent): boolean {
...
 function $onDrop(event: DragEvent, editor: LexicalEditor): boolean {
+  // Get the currently selected image node
   const node = $getImageNodeInSelection();
   if (!node) {
-    return false;
+    return false; // No image node selected, exit early
+  }
+
+  // Get the parent node's key
+  const parent_key = String(node.__parent);
```
etrepum, line 319: There's a `getParent` method.

packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
```diff
@@ -306,25 +309,67 @@ function $onDragover(event: DragEvent): boolean {
...
+  let link = '';
+  if (node) {
+    // Check if the parent node is a link
+    const nodelink = $getNodeByKey(parent_key);
+    if (nodelink?.__type === 'link') {
```
etrepum, line 324: There's an `$isLinkNode` function

etrepum (commented):
I don't think that all of these extra comments are helpful. There's already a shorter and approved PR that should solve this issue #7693

packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
Jynx2004: Hi @etrepum,

Thanks for pointing out the getParent method

etrepum:
This issue this addresses was fixed by #7693

Issue #7691, opened 2025-07-14
Bug: Code Block formatting applies to unintended adjacent lines


romancavry:
Lexical version: ^0.32.1, 0.33.1

## Steps To Reproduce

1. In rich-text editor, enter three lines of the text:
aaa
bbb
ccc

2. Select only the middle line (bbb)

3. Apply code block formatting to the selection

4. Observe that the resulting code block unexpectedly includes ccc, instead of just wrapping the selected line

Link to code example: https://playground.lexical.dev/

**Worth noting**: applying a code block to the last line works as expected

https://github.com/user-attachments/assets/94aec4ed-47fd-461e-b8ea-5714382f4b19

shunnobir:
I couldn't reproduce the bug. I have tried numerous times. It produced the correct output.

etrepum:
it's not well described in the issue but you can tell from the screenshot that it is using line breaks and not paragraphs, to repro do this for step 1:

* Go to an empty playground https://playground.lexical.dev/?emptyEditor=true
* Type the following: "aaa" shift+return "bbb" shift+return "ccc"

shunnobir:
Thanks for the clarification. Didn't notice the usage of line breaks.

PR #7699, opened 2025-07-17
fix(Code Block):Code Block formatting applies to unintended adjacent lines

Jynx2004:
This pull request updates the formatCode function in formatCode.ts to improve the handling of code block insertion for non-collapsed selections. The changes modify the else condition to achieve the following:

**New Behavior:** When a non-collapsed selection is detected, the function now:
Creates a new code node.
Inserts the code node as a sibling immediately below the selected node (using anchorNode.insertAfter(codeNode)).
Moves the selected text content into the new code node.
Removes the original selected content.

**Rationale:** This ensures the code block is placed directly after the selected element in the editor's node structure, maintaining a cleaner and more intuitive document hierarchy compared to the previous approach.

**Impact:** The change improves the user experience by correctly positioning code blocks relative to the selection, avoiding nested or misplaced nodes.

Closes https://github.com/facebook/lexical/issues/7691

### DIFF
--- a/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
@@ -8,13 +8,11 @@
 
 import type {JSX} from 'react';
 
-import {$createLinkNode, LinkNode} from '@lexical/link';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {$wrapNodeInElement, mergeRegister} from '@lexical/utils';
 import {
   $createParagraphNode,
   $createRangeSelection,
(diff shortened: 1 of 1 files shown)


etrepum (commented):
This PR needs tests to demonstrate that it works. You should also test that it works when multiple lines are selected.

I can confirm that it doesn't work correctly when two lines separated by a linebreak are selected and then converted to a code block.

Jynx2004:
> This PR needs tests to demonstrate that it works. You should also test that it works when multiple lines are selected.
> 
> I can confirm that it doesn't work correctly when two lines separated by a linebreak are selected and then converted to a code block.

Sure will do the testing considering all the cases

PR #7703, opened 2025-07-18
Fix(Page Break) : Children of root nodes must be elements or decorators" invariant isn't enforced

Jynx2004:
**Summary:**
Updated the CONTROLLED_TEXT_INSERTION_COMMAND handler in tableObserver.js to improve paragraph insertion behavior within table cells when the last child is a page-break node.

**Changes Made:**
Modified the logic in the CONTROLLED_TEXT_INSERTION_COMMAND handler to insert a new paragraph node containing the text payload directly after the last child of a TableCellNode when:
The last child is a page-break node.
No edge position is defined (edgePosition === undefined).
Replaced the existing $insertParagraphAtTableEdge function call with direct node manipulation using the insertAfter method to append the new paragraph immediately after the page-break node.
Added selectEnd to the new paragraph node to position the cursor at the end of the inserted paragraph, enhancing user experience.

**Impact:**
Ensures precise placement of new content within table cells following a page-break node.
Improves editor usability by correctly positioning the cursor after insertion.

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -691,10 +691,23 @@ export function applyTableHandlers(
               selection,
               tableNode,
             );
+            const lastChild = tableCellNode.getLastChild();
+            const isLastChildPageBreak =
+              lastChild && lastChild.getType() === 'page-break'; // Adjust 'page-break' to the actual node type for page breaks
             if (edgePosition) {
               $insertParagraphAtTableEdge(edgePosition, tableNode, [
                 $createTextNode(payload),
               ]);
+              return true;
+            } else if (isLastChildPageBreak && edgePosition === undefined) {
+              // Create a new paragraph node with the payload text
+              const newParagraph = $createParagraphNode();
+              newParagraph.append($createTextNode(payload));
+              // Insert the new paragraph after the last child (page-break)
+              lastChild.insertAfter(newParagraph);
+              // Optionally, move the selection to the new paragraph
+              newParagraph.selectEnd();
+
               return true;
             }
           }
--- a/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ImagesPlugin/index.tsx
@@ -8,13 +8,11 @@
 
 import type {JSX} from 'react';
 
-import {$createLinkNode, LinkNode} from '@lexical/link';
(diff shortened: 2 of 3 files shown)


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -691,10 +691,23 @@ export function applyTableHandlers(
               selection,
               tableNode,
             );
+            const lastChild = tableCellNode.getLastChild();
+            const isLastChildPageBreak =
+              lastChild && lastChild.getType() === 'page-break'; // Adjust 'page-break' to the actual node type for page breaks
```
etrepum, line 696: page-break is currently a playground-only node that is not distributed with lexical so it doesn't really make sense to encode specific behavior for it in the `@lexical/table` package. This method of type checking is also not compatible with subclasses which would have a different type (this is important for node replacement).

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -691,10 +691,23 @@ export function applyTableHandlers(
...
               ]);
+              return true;
+            } else if (isLastChildPageBreak && edgePosition === undefined) {
+              // Create a new paragraph node with the payload text
+              const newParagraph = $createParagraphNode();
+              newParagraph.append($createTextNode(payload));
+              // Insert the new paragraph after the last child (page-break)
+              lastChild.insertAfter(newParagraph);
+              // Optionally, move the selection to the new paragraph
+              newParagraph.selectEnd();
```
etrepum, line 709: optional isn't a very accurate description since this always happens and there's no way to control it

etrepum (commented):
This is probably a reasonable idea but it doesn't really make sense to change `@lexical/table` to support nodes that are only in the playground. There are two approaches that would make more sense to merge:

1. Figure out a more general heuristic to use, it's likely that the issue you're trying to solve happens any time the last node of the cell is a DecoratorNode that is not inline, for example (`$isDecoratorNode(node) && !node.isInline()`). I don't think page break nodes are generally very special, so it doesn't make sense to solve one specific manifestation of the problem when you can solve the entire class of problem.
2. Implement workarounds for playground nodes in the playground itself (not sure what the best solution for this would be, would require auditing more of the relevant code, and it's likely much harder than the above approach).

It's also important to implement at least one test that demonstrates that this addition actually does what it intends to, and the test will prevent that behavior from regressing as the code around it changes.

PR #7705, opened 2025-07-19
[lexical-react] Clean up old mouseup events from context menu

ivailop7:
Floating UI handles the click away, removing the old mouseup event listener and timeout from the new context menu implementation.


ivailop7:
@etrepum could I get an approval here

PR #7707, opened 2025-07-20
[lexical-playground] DateTime Plugin

ivailop7:
Add DateTime Plugin with typeahead completions for 'Date', 'Today', 'Tomorrow', 'Yesterday'.

The popover rendering positioning respects the widget positioning and where it is relative to the edges.
If a time is added to widget, it defaults to the user's browser local time zone. In the future someone else can extend it further to support timezone selection using 'tzdb'.

https://github.com/user-attachments/assets/96405cc6-a45b-422d-bb45-8adc341a523a

- Doesn't support text formatting


ivailop7:
@etrepum Would you know why marking the decorator node as inline does not actually work? Something is off with the selection bit.

etrepum:
I don't think it works inline because of the css, divs are `display: block` by default. The node itself is inline from lexical's perspective, it's more of a rendering concern.

To make it work with formatting you could make it not a decorator, it could be a TextNode subclass, although we don't currently have a selection/navigation mode by default that would treat a `'token'` node as one unit as far as keyboard and pointer selection is concerned (I don't know how meta is using token mode, but maybe we could just fix that, or add a method or plugin to change that behavior for some or all token TextNode).

ivailop7:
> I don't think it works inline because of the css, divs are `display: block` by default. The node itself is inline from lexical's perspective, it's more of a rendering concern.
> 
> To make it work with formatting you could make it not a decorator, it could be a TextNode subclass, although we don't currently have a selection/navigation mode by default that would treat a `'token'` node as one unit as far as keyboard and pointer selection is concerned (I don't know how meta is using token mode, but maybe we could just fix that, or add a method or plugin to change that behavior for some or all token TextNode).

I started with Element node initially, but the decorator just allowed for nice isolation of the component code with all the event listeners. I do agree, the node is closer to a mention or a hashtag with just a 'onclick' behaviour. I'll look into the CSS. Maybe we can get @zurfyx or @fantactuka to help with the 'token' part?

vadimkantorov:
It also appears there exist an HTML element for semantic representation of date/time: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time

Might be better to use `<time>` for the HTML representation?

IMO would also be great to somehow separate out the mature Playground plugins into a separate package (maybe `lexical-extra-plugins`?), so that they can be re-used/updated simpler than not forgetting to re-copy-pasting

ivailop7:
> It also appears there exist an HTML element for semantic representation of date/time: [developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time)
> 
> Might be better to use `<time>` for the HTML representation?
> 
> IMO would also be great to somehow separate out the mature Playground plugins into a separate package (maybe `lexical-extra-plugins`?), so that they can be re-used/updated simpler than not forgetting to re-copy-pasting

Didn't know about that one! It's a good shout! Will have a look.

ivailop7:
@etrepum this is ready to go! Added a e2e test, made the node inline with proper selection highlight.

Kept the importDOM/exportDOM, since the 'time' html tag doesn't seem to be widespreadly used.

vadimkantorov:
For time html tag - I guess it's available in browsers, but not handled in any special way. Maybe exists currently mainly for semantic annotation

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -0,0 +1,153 @@
...
+): DOMConversionOutput | null {
+  const dateTimeValue = domNode.getAttribute('data-lexical-datetime');
+  if (dateTimeValue) {
+    const node = $createDateTimeNode(new Date(Date.parse(dateTimeValue)));
+    return {node};
+  }
+  return null;
+}
+
+export class DateTimeNode extends DecoratorNode<JSX.Element> {
```
etrepum: You could get rid of a lot of the boilerplate using NodeState (and more easily fix the serialization bugs) and optionally also $config to elide the rest of it. I'm also not sure how collab would deal with a Date here, but if it was NodeState it would go through a well-defined representation (by default JSON.stringify will serialize a Date to an ISO 8601 formatted string).

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -0,0 +1,153 @@
...
+    return new DateTimeNode(node.__dateTime, node.__key);
+  }
+
+  constructor(dateTime: Date | undefined, key?: NodeKey) {
+    super(key);
+    this.__dateTime = dateTime;
+  }
+
+  static importJSON(serializedNode: SerializedDateTimeNode): DateTimeNode {
+    return $createDateTimeNode(serializedNode.dateTime).updateFromJSON(
```
etrepum: dateTime is never actually a Date after JSON.parse, it will be a string.

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -0,0 +1,153 @@
...
+  SerializedLexicalNode,
+  Spread,
+} from 'lexical';
+import * as React from 'react';
+
+const DateTimeComponent = React.lazy(() => import('./DateTimeComponent'));
+
+export type SerializedDateTimeNode = Spread<
+  {
+    dateTime?: Date;
```
etrepum: JSON doesn't round-trip Date, so this doesn't work

etrepum (commented):
I didn't spend any time looking at the UI but I can see that the DateTimeNode implementation has some bugs with serialization and possibly collab that should be straightforward to fix by switching from an explicit property to NodeState.

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -41,109 +60,77 @@ function $convertDateTimeElement(
...
-
-  static getType(): string {
-    return 'datetime';
-  }
-
-  static clone(node: DateTimeNode): DateTimeNode {
-    return new DateTimeNode(node.__dateTime, node.__key);
-  }
+const dateTimeState = createState('dateTime', {
+  parse: (v) => (v instanceof Date ? v : new Date()),
```
etrepum:
```suggestion
  parse: (v) => (v instanceof Date ? v : new Date(v)),
```

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
etrepum: If this export is kept then it should be fixed since it won't actually serialize to Date

```suggestion
    dateTime?: string;
```

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
etrepum: Generally speaking this will never be parsed from a Date, parse is only really used when coming from JSON (when using the setter it is not called), but it would be fine to support it I guess.

packages/lexical-playground/src/plugins/DateTimePlugin/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+} from 'lexical';
+import {useEffect} from 'react';
+
+import {
+  $createDateTimeNode,
+  DateTimeNode,
+} from '../../nodes/DateTimeNode/DateTimeNode';
+
+type CommandPayload = {
+  dateTime: Date | undefined;
```
etrepum:
```suggestion
  dateTime: Date;
```

etrepum (commented):
Almost there, still doesn't quite deserialize correctly (never parses from string, which it will be after JSON) and there's a type error

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const dateTimeValue = domNode.getAttribute('data-lexical-datetime');
+  if (dateTimeValue) {
+    const node = $createDateTimeNode(new Date(Date.parse(dateTimeValue)));
+    return {node};
+  }
+  return null;
+}
+
+const dateTimeState = createState('dateTime', {
+  parse: (v) => new Date(v as string),
```
etrepum, line 64:
```suggestion
  parse: (v) => new Date(v as string),
  unparse: (v) => v.toISOString(),
```

I think this might need an explicit encoding for collab support

vadimkantorov:
Just for information, here is the Clipboard HTML from Google Docs' `@Today` element:
```html
<meta charset="utf-8">
<b style="font-weight:normal;" id="docs-internal-guid-8ead19bd-7fff-859b-a6a8-bd93d5d97918">
<p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;">
<span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;" data-rich-links="{&quot;dat_df&quot;:{&quot;fres_frt&quot;:1,&quot;dfie_ts&quot;:{&quot;tv&quot;:{&quot;tv_s&quot;:1754913600,&quot;tv_n&quot;:0}},&quot;dfie_l&quot;:&quot;en&quot;,&quot;dfie_p&quot;:{&quot;fres_frt&quot;:0,&quot;tres_tv&quot;:&quot;MMM d, y&quot;},&quot;dfie_dt&quot;:&quot;Aug 11, 2025&quot;,&quot;dfie_pt&quot;:3,&quot;dfie_tpt&quot;:0,&quot;dfie_tzi&quot;:&quot;&quot;},&quot;type&quot;:&quot;date&quot;}">Aug 11, 2025</span>
</p>
</b>
<br class="Apple-interchange-newline">
```

Will now Lexical be able to paste this HTML as DateTime Node?

Ha-ha, I'm still sad, no-one actually uses `<time>` for representing datetime semantics :)

ivailop7:
> Just for information, here is the Clipboard HTML from Google Docs' `@Today` element:
> 
> ```html
> 
> <meta charset="utf-8">
> 
> <b style="font-weight:normal;" id="docs-internal-guid-8ead19bd-7fff-859b-a6a8-bd93d5d97918">
> 
> <p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;">
> 
> <span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;" data-rich-links="{&quot;dat_df&quot;:{&quot;fres_frt&quot;:1,&quot;dfie_ts&quot;:{&quot;tv&quot;:{&quot;tv_s&quot;:1754913600,&quot;tv_n&quot;:0}},&quot;dfie_l&quot;:&quot;en&quot;,&quot;dfie_p&quot;:{&quot;fres_frt&quot;:0,&quot;tres_tv&quot;:&quot;MMM d, y&quot;},&quot;dfie_dt&quot;:&quot;Aug 11, 2025&quot;,&quot;dfie_pt&quot;:3,&quot;dfie_tpt&quot;:0,&quot;dfie_tzi&quot;:&quot;&quot;},&quot;type&quot;:&quot;date&quot;}">Aug 11, 2025</span>
> 
> </p>
> 
> </b>
> 
> <br class="Apple-interchange-newline">
> 
> ```
> 
> 
> 
> Will now Lexical be able to paste this HTML as DateTime Node?
> 
> 
> 
> Ha-ha, I'm still sad, no-one actually uses `<time>` for representing datetime semantics :)

I'm currently on holidays. If you raise the PR, I'd be more than happy to review and merge 😃.

PR #7715, opened 2025-07-22
[lexical] Bug Fix: call $validatePoint on resolvedFocusPoint

vchirikov:
## Description

Looks like the original author want to call `$validatePoint` on `resolvedAnchorPoint` AND `resolvedFocusPoint`, but accidentally call it only on `resolvedAnchorPoint` twice.


ivailop7 (approved):

etrepum:
It looks like this has probably found a bug given the amount of tests that are failing

PR #7719, opened 2025-07-22
[lexical-playground] Bug Fix: include font sizes in pt as well in parseAllowedFontSize

### LINKED ISSUE #7718: Bug: when content from google docs or microsoft word, font-size is lost
Lexical version: latest

## Steps To Reproduce

1. Copy content from Google docs or Ms word with some font-size applied
2. Paste into lexical, you will see font size reset to default (15 in case of playground)

Link to code example:

## The current behavior

https://github.com/user-attachments/assets/2edfb61d-fff4-4f72-b078-362c6966d630

## The expected behavior

https://github.com/user-attachments/assets/8a26aacf-72d8-435d-83dc-b9addac5a042

## Impact of fix

Copy paste will improve from non-lexical editor to lexical editor

harshmetkel24:
## Description
Consider font sizes in pt as well along with px in `parseAllowedFontSize` as most of editors like Google docs and MS word export font-sizes in pt when copied content

Closes #7718
Closes #7726

## Test plan
tests added for custom font-size in HTMLCopyAndPaste.test.ts
### Before

https://github.com/user-attachments/assets/3d5ed1a4-5780-44b9-92a3-25a735b37728

### 

After

https://github.com/user-attachments/assets/65c2d61c-d1dc-4805-9460-e73b15542bb1

### DIFF
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
@@ -23,13 +23,21 @@ import {
 } from './utils';
 
 export function parseAllowedFontSize(input: string): string {
-  const match = input.match(/^(\d+(?:\.\d+)?)px$/);
-  if (match) {
-    const n = Number(match[1]);
-    if (n >= MIN_ALLOWED_FONT_SIZE && n <= MAX_ALLOWED_FONT_SIZE) {
-      return input;
-    }
+  const pxMatch = input.match(/^(\d+(?:\.\d+)?)px$/);
+  const ptMatch = input.match(/^(\d+(?:\.\d+)?)pt$/);
+
+  let pxValue: number = pxMatch ? Number(pxMatch[1]) : undefined;
+
+  if (ptMatch) {
+    pxValue = Number(ptMatch[1]) * (4 / 3);
+  } else {
+    return '';
   }
+
+  if (pxValue >= MIN_ALLOWED_FONT_SIZE && pxValue <= MAX_ALLOWED_FONT_SIZE) {
+    return input;
+  }
+
   return '';
 }
 
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
@@ -26,7 +26,7 @@ export function parseAllowedFontSize(input: string): string {
   const pxMatch = input.match(/^(\d+(?:\.\d+)?)px$/);
   const ptMatch = input.match(/^(\d+(?:\.\d+)?)pt$/);
 
-  let pxValue: number = pxMatch ? Number(pxMatch[1]) : undefined;
+  let pxValue: number | undefined = pxMatch ? Number(pxMatch[1]) : undefined;
 
   if (ptMatch) {
     pxValue = Number(ptMatch[1]) * (4 / 3);
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
@@ -30,7 +30,7 @@ export function parseAllowedFontSize(input: string): string {
   const pxValue =
     Number(pxOrptMatch[1]) * (pxOrptMatch[2] === 'pt' ? 4 / 3 : 1);
   if (pxValue >= MIN_ALLOWED_FONT_SIZE && pxValue <= MAX_ALLOWED_FONT_SIZE) {
-    return input;
+    return pxValue + 'px';
   }
 
   return '';
--- a/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
+++ b/packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
@@ -30,11 +30,13 @@ export function parseAllowedFontSize(input: string): string {
 
   if (ptMatch) {
     pxValue = Number(ptMatch[1]) * (4 / 3);
-  } else {
-    return '';
   }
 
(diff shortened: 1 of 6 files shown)


harshmetkel24:
I’ve resolved the lint error, but the test is still failing. After investigating, I realized my fix only addresses the issue in the playground, which doesn’t require updating unit tests in core.

I’d appreciate your guidance on how to proceed. Here are the options I’m considering:
1. Remove the test I added: This avoids misleading others but might allow regressions or encourage unsupported workarounds.
2. Fix it in lexicalTextNode: This would involve applying styles during importDOM, as shown in the first screenshot. However, this causes unintended font-size styles to propagate broadly, resulting in more test failures (see second screenshot).

Let me know what you think is the best direction.
Thanks!

image
image

etrepum:
You can test the behavior of the playground with e2e tests. I do not think the default behavior of TextNode should be changed

harshmetkel24:
oh looks like some other tests related to copy paste are failing now 😢

etrepum:
Yes it looks like other tests are expecting the font-size in pt to be ignored on paste, so those expectations would have to change.

harshmetkel24:
yess fixed

packages/lexical-playground/src/plugins/ToolbarPlugin/fontSize.tsx
```diff
@@ -23,13 +23,21 @@ import {
...
+  if (ptMatch) {
+    pxValue = Number(ptMatch[1]) * (4 / 3);
+  } else {
+    return '';
   }
+
+  if (pxValue >= MIN_ALLOWED_FONT_SIZE && pxValue <= MAX_ALLOWED_FONT_SIZE) {
+    return input;
+  }
+
```
etrepum: I think this is wrong because either pxMatch or ptMatch will be truthy, but this will return '' whenever ptMatch is not  truthy. I'm surprised we don't have any test coverage for that, so I think we should add a test that makes sure that the px case is also tested.

```suggestion
  const pxptMatch = input.match(/^(\d+(?:\.\d+)?)(px|pt)$/);
  if (!pxptMatch) {
    return '';
  }

  const pxValue = Number(pxptMatch[1]) * (pxptMatch[2] === 'pt' ? 4 / 3 : 1);
  if (pxValue >= MIN_ALLOWED_FONT_SIZE && pxValue <= MAX_ALLOWED_FONT_SIZE) {
    return input;
  }

```

etrepum (commented):
When inspecting the playground for this I noticed that the toolbar displays the raw number associated with the font-size, the number parsing should be used for this functionality as well.

```ts
          <FontSize
            selectionFontSize={toolbarState.fontSize.slice(0, -2)}
            editor={activeEditor}
            disabled={!isEditable}
```

Here's a [playground link](https://lexical-playground-git-fork-harshmetkel24-7718-fbopensource.vercel.app/#doc=H4sIAAAAAAAAE92UMWvDMBSE_4q5qQU12FbcgrJnLngsHVRLSQSKZJ6fQ9Pg_17kBNK0i1taKJmkO6S79y3vAGscR6pZs4U6gGLkdDYb5w3ZAPX0SRjL2nmoXGAVaat5vG6jsVAIyfEQ6Hjvk7GKge8692ZVVhQtL5roI6mM1i83ZVWJLBdZfruAANtXhkJ6lNS-Tb9HU2BnqXMxQBXDs4BxZBseNTwTznMAAi4YG44znUJaTXpNut1cJB0bl2eCJOvT1BjEr1GXU6jLa6OWU6jltVHPp1DPr426mkJd_Ufqy9bQe_8npT-AG_fw14r6e2t1GAS87rjWO2ugiodKymJeyfsyLwW62FOT0h693q8p9sF8rEQ-k3JWYHgHfgG4VCUGAAA) for this PR with some content that uses pt font sizes

harshmetkel24:
> When inspecting the playground for this I noticed that the toolbar displays the raw number associated with the font-size, the number parsing should be used for this functionality as well.
> 
> ```ts
>           <FontSize
>             selectionFontSize={toolbarState.fontSize.slice(0, -2)}
>             editor={activeEditor}
>             disabled={!isEditable}
> ```
> 
> Here's a [playground link](https://lexical-playground-git-fork-harshmetkel24-7718-fbopensource.vercel.app/#doc=H4sIAAAAAAAAE92UMWvDMBSE_4q5qQU12FbcgrJnLngsHVRLSQSKZJ6fQ9Pg_17kBNK0i1taKJmkO6S79y3vAGscR6pZs4U6gGLkdDYb5w3ZAPX0SRjL2nmoXGAVaat5vG6jsVAIyfEQ6Hjvk7GKge8692ZVVhQtL5roI6mM1i83ZVWJLBdZfruAANtXhkJ6lNS-Tb9HU2BnqXMxQBXDs4BxZBseNTwTznMAAi4YG44znUJaTXpNut1cJB0bl2eCJOvT1BjEr1GXU6jLa6OWU6jltVHPp1DPr426mkJd_Ufqy9bQe_8npT-AG_fw14r6e2t1GAS87rjWO2ugiodKymJeyfsyLwW62FOT0h693q8p9sF8rEQ-k3JWYHgHfgG4VCUGAAA) for this PR with some content that uses pt font sizes

yes

etrepum:
It looks like some tests are failing because the font-size is a floating point and the precision is different depending on the browser

harshmetkel24:
oh okay I got that. I am thinking to keep font-size in style in pt only. Just round off pt value to nearest px value in toolbar
This should work right?

etrepum:
Yes handling it in a different way in the toolbar will likely work, the parsing should be split into two functions so the toolbar code can just get the px value as a number

harshmetkel24:
Yes agree, I made changes for that

etrepum:
The formatting check is failing `npm run prettier:fix` will typically resolve that

Issue #7606, opened 2025-06-05
Bug: Fixing cursor position after equation


meadow-tulips:
https://github.com/user-attachments/assets/e0e42270-3687-4b36-875c-7f2fc87f490d

Please take a look at the recording. 

Here I am trying to get my cursor behind the equation using mouse clicks.
But cursor stays at the start.

Jynx2004:
After trying all the possible things , one that could probably work is creating a custom inline paragraph by assigning display to be inline flex along with width to be auto . Will provide a merge request soon

PR #7720, opened 2025-07-23
fix(Equation):Fixing cursor position after equation

Jynx2004:
1. Added **_InlineParagraphNode_**, a new Lexical node extending ElementNode to support inline paragraph rendering. 

2. It inherits all properties from ParagraphNode and applies width: auto, minWidth: 1px, and display: inline-flex styles for inline-flex layout

3. Closes : #7606

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -691,23 +691,10 @@ export function applyTableHandlers(
               selection,
               tableNode,
             );
-            const lastChild = tableCellNode.getLastChild();
-            const isLastChildPageBreak =
-              lastChild && lastChild.getType() === 'page-break'; // Adjust 'page-break' to the actual node type for page breaks
             if (edgePosition) {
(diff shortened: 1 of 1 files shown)


etrepum (commented):
I'm certain that there is a simpler and robust solution to this, navigating the cursor with the arrow keys works fine, the only issue is creating the same selection with a click.

I think this approach will cause different problems, will only help with this one specific DecoratorNode when it's likely to be an issue for other inline DecoratorNodes, adds a lot of unnecessary complexity a a semantically meaningless element to the document, has no tests, etc. I would not recommend going further with this approach.

Jynx2004:
> I'm certain that there is a simpler and robust solution to this, navigating the cursor with the arrow keys works fine, the only issue is creating the same selection with a click.
> 
> I think this approach will cause different problems, will only help with this one specific DecoratorNode when it's likely to be an issue for other inline DecoratorNodes, adds a lot of unnecessary complexity a a semantically meaningless element to the document, has no tests, etc. I would not recommend going further with this approach.

Hi @etrepum , will try to find a better solution for this

Issue #7709, opened 2025-07-21
Bug: triggering an update from an mutation/update listener can lead to Yjs document corruption


james-atticus:
Lexical version: 0.33.1

## Steps To Reproduce

1. Register a mutation/update listener which triggers another editor update.
2. Register a mutation listener which calls `editor.read()`
3. Do an action which triggers the first mutation/update listener.

Example failing test: https://github.com/facebook/lexical/commit/635cb9665

## The current behavior

The `read()` from the second mutation listener calls `$commitPendingUpdates` which triggers update listeners. However, update listeners from the first update may not have not yet been called. This can lead to corruption of the Yjs document.

<img width="486" height="358" alt="Image" src="https://github.com/user-attachments/assets/9159758b-2483-42f6-a3fd-cc61318fc956" />

## The expected behavior

Update listeners are called in-order and Yjs is not corrupted.

## Impact of fix

For Yjs specifically, these updates being called out of order can lead to data loss or complete failure to load the document (`Expected shared type to include type attribute`).

Acknowledge that triggering an update from an update listener is an anti-pattern, but reporting this anyway given the impact.

etrepum:
Probably makes sense to at least add a stern warning about the anti-pattern when it’s detected. EditorState.read should be used in these sorts of cases, Editor.read is primarily useful for DOM event handlers and tests

james-atticus:
In our particular case, the `editor.read` was in a completely separate plugin that was doing DOM manipulation so couldn't use `editorState.read`.

Agree a warning would be good, but for Yjs in particular, if there's a way to detect it from the plugin and throw an error, that'd be even better.

Alternatively, could we defer updates from mutation/update listeners until the initial commit is finished, the same way that's done for `onUpdate` with `editor._deferred`?

etrepum:
Having some updates be synchronous and other updates deferred with the same code is madness. We should have an explicit way to do a deferred update, but it would likely still fail in the same way here due to the editor.read. 

You can use EditorState.read anywhere you have a reference to the editor with getEditorState. The only caveat is that it will only reflect what has been reconciled.

james-atticus:
Fair call. I guess we could make `lexical-yjs` track the last editor state it saw, and if the next `prevEditorState` that comes in isn't the same, it could queue the update(s) itself. Eg: have a `Map` from prevEditorState` to `editorState`.

Yeah we try and use editor state where possible, but some functions like `$getNearestNodeFromDOMNode` don't let you do that.

[...]

PR #7724, opened 2025-07-24
Fix trigger reconcilation on nested read

zurfyx:
**Problem:**

A combination of nested updates + reads can cause a missed reconcilation, consequently leading to further errors such as https://lexical.dev/docs/error?code=75&v=6 or mutating (should've been frozen) nodes.

For example:
1. `dispatchCommand(KEY_DOWN)`
2. (dispatch command triggers an update by command design)
3. User triggers a function such as submit message that contains a combination of `editor.read` and `editor.update`
4. `editor.read` flushes by design
5. `editor.update` has no `pendingEditorState` as it has been flushed. Hence, reconcilation never happens.

**Solution**

We can either recreate the EditorState or ignore the flush sync behavior on read. Because we already support `flushSync` flag in nested updates this PR does the first. 

**Note for self**

We can't just reuse the discrete functionality within update as in 

```
read<T>(callbackFn: () => T): T {
    const pendingEditorState = this._pendingEditorState;
    if (pendingEditorState !== null) {
      this.update(() => {}, {discrete: true});
    }
    return this.getEditorState().read(callbackFn, {editor: this});
  }
```

This will trigger a (parent) update and dismiss the concept of queued updates.

### DIFF
--- a/packages/lexical/src/LexicalUpdates.ts
+++ b/packages/lexical/src/LexicalUpdates.ts
@@ -840,7 +840,7 @@ function triggerDeferredUpdateCallbacks(
   }
 }
 
-function processNestedUpdates(
+function $processNestedUpdates(
   editor: LexicalEditor,
   initialSkipTransforms?: boolean,
 ): boolean {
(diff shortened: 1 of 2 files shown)


etrepum:
Related #7709

etrepum (approved):

PR #7726, opened 2025-07-24
Fix(Format Error): when content from google docs or microsoft word, font-size is lost

### LINKED ISSUE #7718: Bug: when content from google docs or microsoft word, font-size is lost

Jynx2004:
Merge Request: Fix Font-Size Loss for Text Nodes from Google Docs/Microsoft Word (#7718)

**Description**

This change addresses issue #7718, where font-size styles are lost when importing content from Google Docs or Microsoft Word into the Lexical editor. Specifically, text nodes (#text) were not inheriting styles from their parent elements, resulting in missing formatting such as font-size.

**Changes**

Added a new condition in the createNodesFromDOM function to handle text nodes (node.nodeName === '#text').

Checks if the text node has a non-empty/non-null value (node.nodeValue) and if its parent node has a style attribute.

If conditions are met, applies the parent's style attribute value (e.g., font-size:11pt;) to the first TextNode in transformOutput.node, ensuring styles like font-size are preserved.

Uses getAttribute('style') to safely retrieve the parent's inline styles, defaulting to an empty string if no styles exist.

**Impact**

Fixes font-size loss for text nodes when importing content from Google Docs or Microsoft Word.

Preserves inline styles (e.g., font-size, color) applied to parent elements, ensuring accurate rendering in the Lexical editor.

Maintains existing behavior for non-text nodes and nodes without parent styles.

No breaking changes; the change is isolated to text nodes with styled parents.

**Testing**

Verified that text nodes from Google Docs/Microsoft Word imports now retain parent styles (e.g., font-size:11pt;).

Confirmed no regressions for non-text nodes or nodes without parent styles.

Tested with sample documents containing styled text to ensure styles are correctly applied to TextNode.__style.

**Closes #7718**


etrepum:
There is already an in-progress PR #7719 that solves this in a more appropriate way, with tests. It's not desired behavior for the default TextNode implementation to copy *all* styles. The expected behavior is to only handle specific styles that are part of the TextNode model.

Jynx2004:
> There is already an in-progress PR #7719 that solves this in a more appropriate way, with tests. It's not desired behavior for the default TextNode implementation to copy _all_ styles. The expected behavior is to only handle specific styles that are part of the TextNode model.

Can i just take the font size attribute from the complete style of the Parent Node and apply it ?
Using a regular expression i can pick the font-size attribute.

etrepum:
No, there's also the matter of keeping the toolbar in sync and writing tests. The approach in PR https://github.com/facebook/lexical/pull/7719 is preferable and almost done. I wouldn't invest any more time in this PR since it will just get closed once that PR is merged.

Jynx2004:
> No, there's also the matter of keeping the toolbar in sync and writing tests. The approach in PR #7719 is preferable and almost done. I wouldn't invest any more time in this PR since it will just get closed once that PR is merged.

Okay

PR #7727, opened 2025-07-25
[lexical] Feature: allow ElementNode direction to be overridden

### LINKED ISSUE #4930: Bug:  change node direction does not work!
i want to change node direction manual, but setDirection of node instance doesnt work! is there any way change dir of a node? because i have something like GPS: دارد and editor recognize it as ltr but i want to change it to rtl manually but i cant and it doesnt work

    "lexical": "^0.11.1",
 "@lexical/react": "^0.11.1",

### LINKED ISSUE #4998: Bug: serialized `direction` in headless editor is always `null`
Lexical version: 0.12.2

## Steps To Reproduce

1. Instantiate an headless editor
2. Add a `ParagraphNode` and a non-empty `TextNode`
3. Log the serizialized editor state

Link to code example: https://codesandbox.io/s/lexical-headless-direction-2c6cg4

Screenshot 2023-09-14 at 11 45 25

Screenshot 2023-09-14 at 11 45 46

## The current behavior

In both the serialized `ParagraphNode` and `RootNode` the direction is `null`, instead of `ltr`.

By taking a quick look at the codebase, it seems that the `direction` is set by the DOM reconciler. It is inferred from a regex on the text content. The headless editor is never reconciled, therefore the `direction` is always `null`, I think it could either: 

- be similarly inferred from the text content
- never be serialized, since it is inferred at runtime anyways

## The expected behavior

The direction should be inferred to be `ltr`.

Our use case is to dynamically create serialized nodes (think on the server). The problem we are facing is that the serialized nodes differ before/after reconciliation, because of the `direction` property. To solve this we are manually calling `setDirection` on the nodes, but that is causing other issues depending on where the serialized node is used.

### LINKED ISSUE #7632: Bug: Calling setDirection() has no visible effect in DOM like setFormat()
### Title: `setDirection()` does not reflect direction changes in DOM like `setFormat()`

---

**Description:**

Hi,  
I'm working on a customized version of `lexical-playground` to use it as a base for a component library. I tried to add a `TextDirectionPlugin` to handle keyboard shortcuts (`Ctrl+Shift+Alt+R/L`) for setting text direction (`rtl` / `ltr`), similar to how `FORMAT_ELEMENT_COMMAND` handles text alignment.

I followed the same structure as Lexical uses for `FORMAT_ELEMENT_COMMAND`, but created my own `DIRECT_ELEMENT_COMMAND` and used `element.setDirection(direction)` instead of `element.setFormat(...)`.

It seems the reconciler detect direction only based on `RTL_REGEX` regular expression and not consider the __dir property of node.

---

**Lexical Version**: T 

---

### 🔁 Steps To Reproduce

1. Create a direction plugin as shown here:  
   https://github.com/Pourianof/lexical-playground/blob/master/src/plugins/TextDirectionPlugin.tsx

2. Add the plugin to the editor:  
   https://github.com/Pourianof/lexical-playground/blob/master/src/Editor.tsx#L203

---

### ❌ Current Behavior

Calling `element.setDirection('rtl')` or `'ltr'` correctly updates the editor state,  
but **no visual change happens in the DOM** — no `dir` attribute is applied to the block's DOM node.  
As a result, text direction is not reflected in the layout or rendering.

---

### ✅ Expected Behavior

Calling `element.setDirection(...)` on a block-level element (e.g., ParagraphNode, HeadingNode) should update the DOM — applying `dir="rtl"` or `dir="ltr"` — similar to how `setFormat(...)` updates `text-align`.

---

### 💡 Possible Cause / Suggestion

It seems the reconciler detect direction only based on RTL_REGEX regular expression and not consider the __dir property of node.

If this is the expected behavior by design, it would be helpful to mention it in the documentation — so developers know they must override `createDOM()` themselves to apply `dir`.

Otherwise, it would be ideal if Lexical core handled applying `dir` attributes automatically when `setDirection()` is called, just like it does with `setFormat()` and `text-align`.

---

### 🌍 Impact

This is especially useful for users of **right-to-left languages** (e.g., Persian, Arabic, Hebrew) where mixed content (e.g., English inside Arabic text) is common, and direction control is essential.

Supporting `setDirection()` would allow users to toggle direction dynamically without having to write custom nodes or styles.

---

Thanks for all the amazing work on Lexical 🙏 Looking forward to your feedback!

james-atticus:
## Description

Removes the custom direction logic in `LexicalReconciler` in favour of delegating the responsibility of detecting text direction to the browser, as described in https://github.com/facebook/lexical/issues/2610#issuecomment-2986760988. 

Closes #4930
Closes #7632
Closes #4998

#### `ltr` and `rtl` theme CSS classes removed

`ltr` and `rtl` theme CSS classes are no longer added to DOM elements. If you have custom CSS to set `text-align` (or any other properties) for bidirectional text, this can be removed. You can target elements that have customized their `dir` by targeting the attr in CSS, for example:

```css
[dir=rtl] { text-align: right; }
```

#### `$isParentElementRTL` requires an active editor

`$isParentElementRTL` now relies on accessing the DOM to determine element direction, meaning there must be an active editor when calling the function. If you have code that calls `$isParentElementRTL` inside `editorState.read()`, you will need to pass in an `editor` argument. For example:

```ts
const isRTL = editorState.read(
  () => {
    return $isRangeSelection(selection) && $isParentElementRTL(selection);
  },
  {editor}
);
```

#### `__dir` property is synced to/from Yjs (again)

Prior to #7330, the text direction of nodes (as determined by the reconciler) was synced to Yjs, even though this value was ignored on other clients. This property is now used for any manually-set direction attribute on `ElementNode`.

If you have documents that were created prior to v0.30.0, the previously-reconciled direction will now be read and applied to nodes in the editor. This may lead to unexpected behaviour if, for example, your document contains a paragraph that contained RTL text but content changed to LTR in >= v0.30.0.

## Test plan

Starting off with just the base change and associated unit tests to get feedback before going and updating all the unit/E2E tests that this breaks.

### DIFF
--- a/examples/react-rich/src/App.tsx
+++ b/examples/react-rich/src/App.tsx
@@ -43,13 +43,10 @@ const removeStylesExportDOM = (
     // in i, b, and strong tags.
     for (const el of [
       output.element,
-      ...output.element.querySelectorAll('[style],[class],[dir="auto"]'),
+      ...output.element.querySelectorAll('[style],[class]'),
     ]) {
       el.removeAttribute('class');
       el.removeAttribute('style');
-      if (el.getAttribute('dir') === 'ltr') {
-        el.removeAttribute('dir');
-      }
     }
   }
   return output;
--- a/packages/lexical/src/LexicalReconciler.ts
+++ b/packages/lexical/src/LexicalReconciler.ts
@@ -589,13 +589,17 @@ function $reconcileNode(
         // Root node direction changing from set to unset (or vice versa)
         // changes how children's direction is calculated.
         $isRootNode(nextNode) &&
-        !nextNode.__dir !== !prevNode.__dir &&
         // Can skip if all children already reconciled.
         !treatAllNodesAsDirty
       ) {
-        for (const child of nextNode.getChildren<ElementNode>()) {
-          const childDom = getElementByKeyOrThrow(activeEditor, child.getKey());
-          $setElementDirection(childDom, child);
+        for (const child of nextNode.getChildren()) {
+          if ($isElementNode(child)) {
+            const childDom = getElementByKeyOrThrow(
+              activeEditor,
+              child.getKey(),
+            );
+            $setElementDirection(childDom, child);
+          }
         }
       }
     }


packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -763,7 +763,7 @@ export default function ToolbarPlugin({
   }, [editor, $updateToolbar, setActiveEditor]);
 
   useEffect(() => {
-    activeEditor.getEditorState().read(() => {
+    activeEditor.read(() => {
```
etrepum: any reason why this was done?

james-atticus: Without it, you get:

```
Uncaught Error: Unable to find an active editor. This method can only be used synchronously during the callback of editor.update() or editor.read(). Detected on the page: 1 compatible editor(s) with version 0.33.1+git
    at formatDevErrorMessage (formatDevErrorMessage.ts:12:9)
    at getActiveEditor (LexicalUpdates.ts:119:7)
    at $getEditor (LexicalUtils.ts:1885:10)
    at $getComputedStyle (range-selection.ts:442:18)
    at $isParentElementRTL (range-selection.ts:516:25)
    at index.tsx:631:35
    at index.tsx:767:7
    at readEditorState (LexicalUpdates.ts:477:12)
    at EditorState.read (LexicalEditorState.ts:116:12)
    at index.tsx:766:35
```

The alternative would be `activeEditor.getEditorState().read(() => {}, {editor: activeEditor})` similar to below but that felt overly verbose when `activeEditor.read(...)` works.

etrepum: The downside is the side-effect of flushing any pending state which isn't necessarily desirable unless you're sure that there isn't any (most dom events) or that it's useful (tests)

Pourianof:
As someone who speaks in RTL language, I can confirm that the layout of the lists in this mode is much more acceptable because in the old mode, the left and right alignment of the items, especially on relatively wide screens, made the items get lost and difficult to read.
As for tables, I also feel that the new method makes for a more orderly and visually readable layout, which makes it easier for the mind to predict and read the position of the text. 
But what if we don't want to adjust the direction based on the previous paragraph? i saw no option for changing the defaults, the direction options just change the single cell direction

james-atticus:
@Pourianof thanks for the feedback.

> But what if we don't want to adjust the direction based on the previous paragraph?

Can you clarify what you mean by this? You should be able to set the direction on the root node to change the direction for all paragraphs, or on each paragraph if you want to control their direction individually.

Pourianof:
> Can you clarify what you mean by this? You should be able to set the direction on the root node to change the direction for all paragraphs, or on each paragraph if you want to control their direction individually.

@james-atticus 
Maybe my mistake. You were not wat to implement the direction adjustment buttons, instead you wanted to implement the default behavior and fix the direction setting bugs. Did the problem I mentioned in #7632  also fixed which calling `element.setDirection(direction)` on element has not the effect as `element.setFormat` ?

james-atticus:
@Pourianof Yes, calling `setDirection` will now work. The tests have some examples: https://github.com/facebook/lexical/pull/7727/files#diff-c09925ebf2a87808d4e448cbad5c0ed0667966c9d5aea6024a92ec123a15af37R63-R116

etrepum:
If you want to test setting the direction on the root node you can open up a devtools javascript console and perform an update. It's a bit awkward without easy access to the lexical module's exports but this will work:

```
(() => { let editor = document.querySelector('[contenteditable=true]').__lexicalEditor; let rootNode = editor.getEditorState()._nodeMap.get('root'); editor.update(() => { rootNode.setDirection('rtl') }) })()
```

For setting a specific paragraph you'd want to traverse it from the root node, e.g. `rootNode.getChildAtIndex(0).setDirection('rtl')`

packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
```diff
@@ -1383,139 +1383,6 @@ describe('LexicalListItemNode tests', () => {
       });
     });
 
-    describe('ListItemNode RTL behavior', () => {
```
james-atticus, line 1386: No longer have custom behaviour for this.

packages/lexical-playground/__tests__/e2e/Autocomplete.spec.mjs
```diff
@@ -35,9 +35,7 @@ test.describe('Autocomplete', () => {
     await assertHTML(
       page,
       html`
-        <p
-          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
```
james-atticus, line 39: TODO: remove `PlaygroundEditorTheme__ltr` and `PlaygroundEditorTheme__rtl` from example styles

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/HTMLCopyAndPaste.spec.mjs
```diff
@@ -56,22 +56,20 @@ test.describe('HTML CopyAndPaste', () => {
...
 
     await pasteFromClipboard(page, clipboard);
 
     await assertHTML(
       page,
       html`
-        <p
-          class="PlaygroundEditorTheme__paragraph PlaygroundEditorTheme__ltr"
-          dir="ltr">
+        <p class="PlaygroundEditorTheme__paragraph" dir="auto">
```
james-atticus, line 67: The clipboard content above has `dir="ltr"` however we don't read that in `importDOM`. We do include `dir` in `exportDOM` - that was a little easier because it's just in `LexicalElementNode`, whereas import would mean including it in a whole lot of places and this PR is already pretty big.

Given we don't currently read `dir` in `importDOM`, can we treat adding support as a separate change from this one?

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/ListsHTMLCopyAndPaste.spec.mjs
```diff
@@ -32,7 +32,16 @@ test.describe('HTML Lists CopyAndPaste', () => {
 
     await assertHTML(
       page,
-      '<ul class="PlaygroundEditorTheme__ul"><li value="1" class="PlaygroundEditorTheme__listItem PlaygroundEditorTheme__ltr" dir="ltr"><span data-lexical-text="true">Hello</span></li><li value="2" class="PlaygroundEditorTheme__listItem PlaygroundEditorTheme__ltr" dir="ltr"><span data-lexical-text="true">world!</span></li></ul>',
+      html`
```
james-atticus, line 35: Took the opportunity to switch to tagged templates for readability. The actual change is pulling dir up to the top list element and removing it from list items.

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -765,71 +765,39 @@ test.describe.parallel('Tables', () => {
...
               </td>
               <td>
-                <p dir="ltr"><span data-lexical-text="true">f</span></p>
+                <p><span data-lexical-text="true">f</span></p>
               </td>
             </tr>
           </table>
-          <p><br /></p>
-        `,
-        html`
```
james-atticus, line 800: No longer need separate left/right HTML for collab

packages/lexical-playground/__tests__/regression/7163-graphemes.spec.mjs
```diff
@@ -214,17 +193,6 @@ test.describe('Grapheme deleteCharacter', () => {
           page,
           selectionFromOffset(codeUnits, graphemePath),
         );
-        if (dir !== 'rtl') {
-          // It's unclear why Firefox navigates differently in these specific,
```
james-atticus, line 218: Doesn't any more!

packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -58,12 +58,12 @@ export function wrapTableHtml(expected, {ignoreClasses = false} = {}) {
   return html`
     ${expected
       .replace(
-        /<table/g,
-        `<div${
+        /<table([^>]*)(dir="\w+")([^>]*)>/g,
+        `<div $2${
```
james-atticus, line 63: It's not the nicest regex, but couldn't see a better way to move the `dir` attribute from tables to their parent scrollable wrappers.

packages/lexical-table/src/__tests__/unit/LexicalTableNode.test.tsx
```diff
@@ -110,6 +113,17 @@ describe('LexicalTableNode tests', () => {
           hasHorizontalScroll ? wrapTableHtml(expected) : expected,
         );
       }
+
+      function expectReconciledTableHtmlToBeEqual(
```
james-atticus, line 117: Some of the tests check inside `editor.update`, i.e. before the reconciler adds the `dir` attribute. Having a separate util functions `expect(Reconciled)TableHtmlToBeEqual` felt the cleanest option.

packages/lexical/src/__tests__/unit/LexicalSerialization.test.ts
```diff
@@ -110,7 +110,7 @@ describe('LexicalSerialization tests', () => {
...
..."paragraph","version":1,"textFormat":0,"textStyle":""}],"direction":null,"format":"","indent":0,"type":"tablecell","version":1,"backgroundColor":null,"colSpan":1,"headerState":0,"rowSpan":1},{"children":[{"children":[],"direction":null,"format":"","indent":0,"type":"paragraph","version":1,"textFormat":0,"textStyle":""}],"direction":null,"format":"","indent":0,"type":"tablecell","version":1,"backgroundColor":null,"colSpan":1,"headerState":0,"rowSpan":1},{"children":[{"children":[],"direction":null,"format":"","indent":0,"type":"paragraph","version":1,"textFormat":0,"textStyle":""}],"direction":null,"format":"","indent":0,"type":"tablecell","version":1,"backgroundColor":null,"colSpan":1,"headerState":0,"rowSpan":1},{"children":[{"children":[],"direction":null,"format":"","indent":0,"type":"paragraph","version":1,"textFormat":0,"textStyle":""}],"direction":null,"format":"","indent":0,"type":"tablecell","version":1,"backgroundColor":null,"colSpan":1,"headerState":0,"rowSpan":1}],"direction":null,"format":"","indent":0,"type":"tablerow","version":1}],"direction":null,"format":"","indent":0,"type":"table","version":1}],"direction":null,"format":"","indent":0,"type":"root","version":1}}`;
```
james-atticus, line 113: `"direction":"ltr"` -> `"direction":null`

packages/lexical/src/__tests__/unit/LexicalEditor.test.tsx
```diff
@@ -1530,17 +1530,19 @@ describe('LexicalEditor tests', () => {
...
-        $getRoot().append(paragraph);
+        $getRoot().clear().append(paragraph);
       });
       const stringifiedEditorState = JSON.stringify(editor.getEditorState());
       const parsedEditorStateFromObject = editor.parseEditorState(
         JSON.parse(stringifiedEditorState),
       );
       parsedEditorStateFromObject.read(() => {
         const root = $getRoot();
+        expect(root.getFirstChild<ParagraphNode>()!.getDirection()).toBe('rtl');
```
james-atticus, line 1545: Checking that direction is now exported to/import from JSON correctly. Let me know if there's a better place to put this test.

packages/lexical-react/src/__tests__/unit/Collaboration.test.ts
```diff
@@ -300,41 +299,30 @@ describe('Collaboration', () => {
...
     expect(client1.getHTML()).toEqual(
-      '<p dir="rtl"><span data-lexical-text="true">فرعي</span></p><p><br><br></p>',
-    );
-
-    // Check that the second paragraph still has RTL direction on client 2, as __dir is not synced.
-    expect(client2.getHTML()).toEqual(
-      '<p dir="rtl"><span data-lexical-text="true">فرعي</span></p><p dir="rtl"><br><br></p>',
+      '<p dir="rtl"><span data-lexical-text="true">hello</span></p>',
     );
+    expect(client2.getHTML()).toEqual(client1.getHTML());
```
james-atticus, line 325: Direction will now be the same on two different clients syncing via Yjs

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -665,6 +582,23 @@ function $reconcileNode(
...
       editorTextContent += DOUBLE_LINE_BREAK;
     }
+
+    if (treatAllNodesAsDirty || nextNode.__dir !== prevNode.__dir) {
+      $setElementDirection(dom, nextNode);
+      if (
+        // Root node direction changing from set to unset (or vice versa)
+        // changes how children's direction is calculated.
+        $isRootNode(nextNode) &&
+        !nextNode.__dir !== !prevNode.__dir &&
```
james-atticus:
```suggestion
```

examples/react-rich/src/App.tsx
```diff
@@ -43,7 +43,7 @@ const removeStylesExportDOM = (
     // in i, b, and strong tags.
     for (const el of [
       output.element,
-      ...output.element.querySelectorAll('[style],[class],[dir="ltr"]'),
+      ...output.element.querySelectorAll('[style],[class],[dir="auto"]'),
```
etrepum: I think this particular case should be to remove the selector since dir="auto" already doesn't seem to be exported from exportDOM by default, but the below code should also change to remove the code that handles pruning the dir attribute

```suggestion
      ...output.element.querySelectorAll('[style],[class]'),
```

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/HTMLCopyAndPaste.spec.mjs
etrepum: I think that's probably fine, especially since I think `dir="auto"` may handle most of the UX expectations if the text is not ltr. It's worth at least a follow-up issue to note that `dir` is not yet imported whether we implement it in the same release as this PR or not.

packages/lexical-playground/__tests__/utils/index.mjs
etrepum: it's a test, doesn't need to be elegant!

packages/lexical-yjs/src/Utils.ts
```diff
@@ -51,7 +51,6 @@ const elementExcludedProperties = new Set<string>([
   '__first',
   '__last',
   '__size',
-  '__dir',
```
etrepum, line 54: This is certainly the right thing to do, but for the breaking changes notes it's probably worth thinking through if there are edge cases for what happens when legacy computed-direction clients interact with the same documents as new manual-direction clients. I think it's probably fine since the legacy clients will ignore this attribute and will not have set it previously in a document that they created, but I haven't put a lot of thought into it.

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -665,6 +582,23 @@ function $reconcileNode(
...
+        // Can skip if all children already reconciled.
+        !treatAllNodesAsDirty
+      ) {
+        for (const child of nextNode.getChildren<ElementNode>()) {
+          const childDom = getElementByKeyOrThrow(activeEditor, child.getKey());
+          $setElementDirection(childDom, child);
+        }
```
etrepum: RootNode can have DecoratorNode children, so this cast is not correct

```suggestion
        for (const child of nextNode.getChildren()) {
          if ($isElementNode(child)) {
            const childDom = getElementByKeyOrThrow(activeEditor, child.getKey());
            $setElementDirection(childDom, child);
          }
        }
```

packages/lexical/src/__tests__/unit/LexicalReconciler.test.ts
```diff
@@ -30,24 +31,133 @@ describe('LexicalReconciler', () => {
...
+      const directions = editor.read(() => {
+        return $getRoot()
+          .getChildren<ParagraphNode>()
+          .map((child) => $getReconciledDirection(child));
       });
-      expect(para3Dir).toEqual('ltr');
+      expect(directions).toEqual(['auto', 'auto', 'auto']);
+    });
+
+    test('Should not set direction of root node children if root node has direction', async () => {
```
etrepum, line 42: Just marking this spot in case anyone else is looking for tests that were added in this PR to confirm the new behavior

packages/lexical/src/__tests__/unit/LexicalSerialization.test.ts
etrepum: Will be nice when we clean up the JSON representation to not include so many redundant defaults!

packages/lexical-playground/__tests__/e2e/Autocomplete.spec.mjs
james-atticus: Done

packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/HTMLCopyAndPaste.spec.mjs
james-atticus: #7765

packages/lexical/src/LexicalReconciler.ts
james-atticus: Good catch, fixed.

packages/lexical-yjs/src/Utils.ts
james-atticus: Clients that have `__dir` excluded will be fine as `isExcludedProperty` is used for both syncing from and to Yjs.

What might cause problems though is any documents created before #7330 merged as they will have direction set from the existing reconciler logic.

One option is to just include a warning in the release notes. Reviving the `__dir` attribute from old Ydocs should have pretty minimal impact. It'd be the same as anyone who's currently serialising their documents to JSON as they will have scattered `"__dir": "ltr"` too. Anyone who wants to switch to the new behaviour would have to add a node transform or something to clear `__dir` from all nodes.

The alternative would be to change the name of the property to `__direction` or something. That feels like a bigger breaking change though.

etrepum (approved):
Looks great!

etrepum:
From my perspective the only thing left to do (ideally before merge) is to tighten up the PR description to clearly spell out the breaking changes so it will be easy to paste them into the release notes

etrepum:
Everything looks great here, just giving the rest of the team an opportunity to take a look before merge

packages/lexical-playground/src/plugins/ToolbarPlugin/index.tsx
```diff
@@ -764,9 +764,12 @@ export default function ToolbarPlugin({
...
 
   useEffect(() => {
-    activeEditor.getEditorState().read(() => {
-      $updateToolbar();
-    });
+    activeEditor.getEditorState().read(
+      () => {
+        $updateToolbar();
+      },
+      {editor: activeEditor},
```
zurfyx, line 771: When is this line required? What would break without it?

etrepum: Per the breaking changes notes, this is needed to compute the direction from DOM

PR #7731, opened 2025-07-29
[lexical-website] Documentation Update: Add editor context gotcha documentation for editorState.read()

mrdivyansh:
## Description

**Current behavior:** 
The documentation does not explain that `editor.getEditorState().read()` lacks editor context by default, causing functions like `$getEditor()` to throw "Unable to find an active editor" errors. This is a common gotcha that developers discover through trial and error.

**Changes being added:**
- Enhanced the existing tip sections in `intro.md` and `README.md` with clear warnings about editor context differences
- Added a dedicated FAQ entry explaining why `$getEditor()` fails in `editorState.read()` and how to fix it
- Added warning callout in `concepts/editor-state.md` right after the `editorState.read()` example
- Provided multiple solutions with clear ❌/✅ examples showing what works and what doesn't
- Added guidance on when to use `editor.read()` vs `editor.getEditorState().read()`

### Before

*Developers encountering this issue would:*
- Try `editor.getEditorState().read(() => $getEditor())` 
- Get cryptic "Unable to find an active editor" error
- Have to dig through tests or source code to understand the `{ editor }` option
- No clear guidance on when to use which read method

### After

*With the updated documentation:*
- Clear warnings in multiple locations where developers will see them
- Prominent ⚠️ gotcha sections with visual ❌/✅ indicators
- Dedicated FAQ entry for this specific error
- Multiple solution examples showing both the explicit editor option and using `editor.read()` instead
- Guidance on when to use each approach

https://github.com/user-attachments/assets/586cbf76-f82f-4e28-a705-31be0efd19e3

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,27 @@ pending state, where node transforms and DOM reconciliation may not have run yet
 `editor.getEditorState().read()` will use the latest reconciled `EditorState` (after any node transforms,
 DOM reconciliation, etc. have already run), any pending `editor.update` mutations will not yet be visible.
 
+**⚠️ Editor Context Gotcha:** `editor.read()` provides editor context (allowing functions like `$getEditor()` to work),
+while `editor.getEditorState().read()` does **not** provide editor context by default. If you need editor context
+with `editorState.read()`, you must explicitly provide it:
+
+```js
+// ❌ This will throw "Unable to find an active editor"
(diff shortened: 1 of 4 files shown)


etrepum:
Why not improve the error message and API docs? People will almost certainly find those a lot more easily.

PR #7732, opened 2025-07-30
[lexical-code][lexical-playground] Chore: Added type exports for serialized nodes

davidbraginsky:
## Description

I have noticed that all serialized nodes have their types exported from their respective packages with the exception of SerializedCodeHighlightNode and SerializedCollapsibleContainerNode.

### DIFF
--- a/packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleContainerNode.ts
+++ b/packages/lexical-playground/src/plugins/CollapsiblePlugin/CollapsibleContainerNode.ts
@@ -27,7 +27,7 @@ import {
 
 import {setDomHiddenUntilFound} from './CollapsibleUtils';
 
-type SerializedCollapsibleContainerNode = Spread<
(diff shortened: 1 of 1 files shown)


packages/lexical-code/src/CodeHighlightNode.ts
```diff
@@ -24,7 +24,7 @@ import {$applyNodeReplacement, ElementNode, TextNode} from 'lexical';
 
 import {$createCodeNode} from './CodeNode';
 
-type SerializedCodeHighlightNode = Spread<
+export type SerializedCodeHighlightNode = Spread<
```
etrepum, line 27: This isn't sufficient to use the type because it's not exported from the package's entrypoint (index.ts)

etrepum (commented):
I don't think there's all that much use for this in general, unless you want it to show up in the API documentation. There are other ways to get to that type such as using `LexicalExportJSON`

Issue #7736, opened 2025-08-04
Bug: Selection doesnt retain when text is selected from right to left


jvithlani:
In lexical playground when the toggle for retain selection is turned on, and the text is selected from right to left and clicked on the toolbar the selection doesnt retain

Lexical version:

## Steps To Reproduce

1. Type 'hello world'
2. Copy world right to left
3. Click on toolbar, the selection is gone

Video:

https://github.com/user-attachments/assets/7bc54d5f-ea84-4a7d-ac53-fc72fbad84c6

## The current behavior
Selection doesnt retain

## The expected behavior
Selection retains

## Impact of fix
Its minor, but better UX

etrepum:
Probably a simple fix to markSelection for whoever is interested in resolving this

jvithlani:
@etrepum done!

PR #7737, opened 2025-08-04
[lexical-utils] fix: Backward selection was not being retained

jvithlani:
## Description

**Problem**
The selectionAlwaysOnDisplay utility wasn't working correctly for right-to-left text selections (when users drag from right to left). The visual selection mark would not appear properly when the editor lost focus after making a right-to-left selection.

**Root Cause**
The issue was in the rangeFromPoints function in markSelection.ts. The function was only checking focusNode.isBefore(anchorNode) to determine selection direction, which works for selections across different DOM nodes but fails for selections within the same text node.

For same-node selections:
Left-to-right: anchor.offset < focus.offset (e.g., 5 → 10)
Right-to-left: anchor.offset > focus.offset (e.g., 10 → 5)
The original logic didn't account for this offset-based direction detection.

**Solution**
Enhanced the selection direction detection logic to handle both cases:

Closes https://github.com/facebook/lexical/issues/7736

### Before

https://github.com/user-attachments/assets/7bc54d5f-ea84-4a7d-ac53-fc72fbad84c6

### After

https://github.com/user-attachments/assets/9c82f76c-3cc5-4678-80c9-028bc17dd0b5

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/SelectionAlwaysOnDisplay.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/SelectionAlwaysOnDisplay.spec.mjs
@@ -6,7 +6,7 @@
  *
  */
 
-import {selectAll} from '../keyboardShortcuts/index.mjs';
+import {selectAll, selectCharacters} from '../keyboardShortcuts/index.mjs';
 import {
   evaluate,
   expect,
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/__tests__/e2e/SelectionAlwaysOnDisplay.spec.mjs
```diff
@@ -58,5 +58,52 @@ test.describe('SelectionAlwaysOnDisplay', () => {
     await expect(widthDifference).toBeLessThanOrEqual(5);
     await expect(heightDifference).toBeLessThanOrEqual(5);
   });
+
+  test(`retain selection works with reverse selection`, async ({
```
jvithlani, line 62: confirmed this fails on the main branch

etrepum (commented):
A simpler fix would be to use the isBackwards method of RangeSelection, e.g.

```ts
function $getOrderedSelectionPoints(selection: RangeSelection): [PointType, PointType] {
  // there's no reason for this to be nullable when called from RangeSelection, that could be fixed
  const points = selection.getStartEndPoints()!;
  return selection.isBackward() ? points.reverse() : points;
}
```

With this method to retrieve the anchor and focus, you could remove the broken direction code from `rangeFromPoints` altogether.

jvithlani:
something like this @etrepum ?

PR #7743, opened 2025-08-07
[lexical-examples]  Bug Fix: missing dependency @lexical/utils at examples/react-rich-collab

debuggingfuture:
### Expected vs Actual
- `npm i && npm run dev` run out of the box
- however below error is thrown

```
11:55:27 PM [vite] Internal server error: Failed to resolve import "@lexical/utils" from "src/plugins/ToolbarPlugin.tsx". Does the file exist?
```

### Root Cause & Changes
- `package-lock.json` is outdated, adding `@lexical/utils` in `package.json` and `npm i`  resolves the issue 
- seems other packages also has outdated `package-lock.json`. Shall we bump all of them & run `npm i` at build process everytime?

### After

- tested `npm i && npm run dev` able to start

### DIFF
--- a/examples/react-rich-collab/package-lock.json
+++ b/examples/react-rich-collab/package-lock.json
@@ -1,16 +1,17 @@
 {
   "name": "@lexical/react-rich-collab-example",
-  "version": "0.29.0",
+  "version": "0.33.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@lexical/react-rich-collab-example",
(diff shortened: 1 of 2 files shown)


etrepum:
The examples are not part of the build toolchain so these sorts of things can fall through the cracks

etrepum:
There's also a bit of a bootstrapping issue with the package-lock files for the examples because they can't be updated (in any conventional way) until after the new version is published to npm

etrepum (approved):

PR #7749, opened 2025-08-10
Stabilize selected text drag and drop functionality

asnmi:
## Description
*Stabilize selected text drag and drop functionality*

##before

https://github.com/user-attachments/assets/a37673b4-0ef3-4a4b-a4ec-2180f9d88b0f

##after

https://github.com/user-attachments/assets/dd4f28da-0332-4dee-aa48-160bbd53e5a7


etrepum:
While there certainly are edge cases in the existing implementation I don't think this fix is correct

https://github.com/user-attachments/assets/d55172b1-2de6-463f-9b4c-b68fdbda7e0a

PR #7753, opened 2025-08-12
[lexical-offset] Bug Fix: inline element selection, match user input observed behaviour with text nodes

### LINKED ISSUE #7580: Bug: @lexical/offset makes invalid selection if offset points to inline node

felzsoca:
## Description
inline element selection (multiple)
prefer text node end over next node start (TODO should detail out separately for collapsed selection?)

Closes #7580 (TODO double-check and create bugs for the other issues...)

### Before

### After


etrepum:
The most important thing to do when adding fixes here would be to add tests that show what the fixes are supposed to do, confirm that they work, and ensure that they continue to work as the rest of the code changes.

felzsoca:
Hi @etrepum,

thanks for checking the PR! You are right, the PR is not there yet, hence I created just as a draft one.
I started to add unit tests locally, but as there was none before for this package, it takes a little longer. I also add some nice helpers via which it is easier to arrange the tests.
I plan to have multiple commits with different test cases, just would like to have an 'acceptable' first one as I already had an idea for arranging nodes. Will get back to it tomorrow.

PR #7757, opened 2025-08-13
[lexical-react] Bug Fix: Add $config $transform inheritance to LexicalNestedComposer

### LINKED ISSUE #7756: Bug: Nested composers not running $config transforms for nodes inherited from parent
Nested composers seem to not be running node transforms defined via the new $config API, for nodes that are inherited from the parent editor. A specific manifestation of this issue is that list item indices are not being auto-incremented as the user types out a list (see screenshot below).

Image

I'm guessing this could be because we're not pulling in the transforms defined via the $config API in 
[LexicalNestedComposer::getTransformSetFromKlass](https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalNestedComposer.tsx#L32-L37).

## More context about our setup

We have a custom "collapsible" node that uses `LexicalNestedComposer` in its implementation. The nested composer inherits nodes from the parent editor as no nodes are specified when the child editor is created. The code fragment below should help to illustrate the setup.

```
export class CollapsibleNode extends DecoratorNode<React.Node> {
  __content: LexicalEditor;

  constructor(
    content?: LexicalEditor,
    key?: NodeKey,
  ) {
    super(key);
    if (content) {
      this.__content = content;
    } else {
      this.__content = createEditor();
    }
  }

  decorate(): React.MixedElement {
    return (
      <LexicalNestedComposer
        initialEditor={this.__content}
        initialTheme={[...]}>
        <RichTextPlugin [...] />
        [...]
      </LexicalNestedComposer>
    );
  }

  [...]
}
```

etrepum:
## Description

When the editor used in a LexicalNestedComposer does not specify a nodes configuration it inherits them from the parent, and also creates the associated transforms. This behavior was not previously implemented for `$config` `$transform`.
Closes #7756

## Test plan

New unit test to confirm expected behavior


takuyakanbr (approved):
Thanks for the quick fix!

PR #7766, opened 2025-08-17
[lexical-website] Chore: Update docusaurus-plugin-typedoc

### LINKED ISSUE #6072: Feature: [lexical-website] Upgrade docusaurus-plugin-typedoc and typedoc-plugin-markdown
It seems that a lot of improvements have happened to docusaurus-plugin-typedoc, seems to fix some issues that I've noticed with how the docs are generated and organized, but it's a backwards incompatible change.

We should:

* Upgrade these packages
* Update the configuration accordingly (sidebars in particular, but we may also be able to get rid of some of our other plugins that exist just to work around previous bugs)
* Update all of the static links to the documentation in the monorepo (the URL structure changes, but we also only have these references in the README files so it's not a big task)
* Consider adding redirect routes in lexical-website so that old links redirect to their new location rather than 404

etrepum:
## Description

Updates the API documentation build to use the latest docusaurus-plugin-typedoc and typedoc-plugin-markdown.

Was a lot harder than I expected to track down all of the changes we'd need to make this work similarly to the current version.

The biggest change is that now all of the classes and interfaces are documented in the module they are exported from, there is no longer a separate file for each class and interface or a global index of all classes and all interfaces. In order to get the URLs to be the same I had to write some extra plugins.

Before: https://lexical.dev/docs/api/
After: https://lexical-git-fork-etrepum-update-docusaurus-fbopensource.vercel.app/docs/api/

Other relevant changes:
* Updated CI node-version to current LTS node versions (now 22.19.0 and 20.19.4; previously 20.11.0 and 18.18.0)
* Switched prettier config from js to json (avoiding cjs here since we are not type: module in package.json yet)
* Updated prettier from 2.8.8 to 3.6.2, which causes churn elsewhere (primarily in https://github.com/facebook/lexical/pull/7766/commits/a1f21cd3e7e5ddf3392ac21848154f33d8e223fe) and required some other changes to build infrastructure since it has an async API now. Previously we had two versions of prettier (one for build & one for app usage) which was a bit problematic.
* Updated typescript to 5.9.2

Closes #6072

## Test plan

There aren't tests for the documentation, but the build process does throw errors if there are broken links.


packages/lexical-utils/src/index.ts
```diff
@@ -872,35 +873,6 @@ export function $unwrapNode(node: ElementNode): void {
...
 
-/**
- * Returns the Node sibling when this exists, otherwise the closest parent sibling. For example
- * R -> P -> T1, T2
- *   -> P2
- * returns T2 for node T1, P2 for node T2, and null for node P2.
- * @param node LexicalNode.
- * @returns An array (tuple) containing the found Lexical node and the depth difference, or null, if this node doesn't exist.
- */
-export function $getAdjacentSiblingOrParentSiblingCaret<
```
etrepum, line 883: This was already defined separately in lexical so re-exporting for compatibility

packages/lexical-website/src/plugins/lexical-typedoc-plugin-legacy-router/index.mjs
```diff
@@ -0,0 +1,54 @@
...
+            ),
+        )
+      ) {
+        comment.summary.pop();
+      }
+    },
+  );
+}
+
+class LegacyRouter extends ModuleRouter {
```
etrepum, line 44: The paths all changed in this version of typedoc so this is a custom router that uses a URL scheme that's backwards compatible with ours

packages/lexical-website/src/plugins/lexical-typedoc-plugin-legacy-router/index.mjs
```diff
@@ -0,0 +1,54 @@
...
+ * @typedef {import('typedoc').Reflection} Reflection
+ * @typedef {import('typedoc').RouterTarget} RouterTarget
+ */
+
+/** @param {Application} app */
+export function load(app) {
+  app.renderer.defineRouter('legacy', LegacyRouter);
+  app.converter.on(
+    Converter.EVENT_CREATE_DECLARATION,
+    function removeLicenseComment(ctx, reflection) {
```
etrepum, line 25: This avoids attaching the license to the first definition in the module if it doesn't have its own comment

packages/lexical-website/plugins/webpack-buffer/index.js
```diff
@@ -16,6 +16,18 @@ module.exports = async function (context, options) {
           new currentBundler.instance.ProvidePlugin({
             Buffer: ['buffer', 'Buffer'],
           }),
+          function disableExpensiveBundlerOptimizationPlugin() {
+            return {
+              configureWebpack(_config, _isServer) {
+                return {
+                  optimization: {
+                    concatenateModules: false,
```
etrepum, line 24: This is a performance improvement for the build, but may produce more js (~3%) for the clients to download https://github.com/facebook/docusaurus/discussions/11199

etrepum:
@zurfyx I think something in the admin side of the repo needs to be flipped so this can land, it appears to be looking for statuses of specific checks with the old matrix node-versions

PR #7770, opened 2025-08-22
[Chore] Update flow suppressions to use error code

SamChou19815:
## Description

Flow will ban codeless suppressions soon. This PR adds error codes to a few codeless suppressions in lexical.

## Test plan

flow

### DIFF
--- a/packages/lexical-link/flow/LexicalLink.js.flow
+++ b/packages/lexical-link/flow/LexicalLink.js.flow
@@ -72,7 +72,7 @@ export type SerializedAutoLinkNode = {
 };
 declare export class AutoLinkNode extends LinkNode {
   static getType(): string;
-  // $FlowFixMe clone method inheritance
+  // $FlowFixMe[incompatible-type] clone method inheritance
   static clone(node: AutoLinkNode): AutoLinkNode;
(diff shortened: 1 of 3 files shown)


packages/lexical-link/flow/LexicalLink.js.flow
```diff
@@ -72,7 +72,7 @@ export type SerializedAutoLinkNode = {
 };
 declare export class AutoLinkNode extends LinkNode {
   static getType(): string;
-  // $FlowFixMe clone method inheritance
+  // $FlowFixMe[incompatible-type] clone method inheritance
```
etrepum:
```
Cannot extend `LinkNode` [1] with `AutoLinkNode` because `AutoLinkNode` [2] is incompatible with `LinkNode` [3] in the
first parameter of property `clone`. [incompatible-extend]
```

SamChou19815: `incompatible-type` is the error code on latest version of flow with `experimental.error_code_migration=new` (and will be the default in the next version), which is what's used internally.

Revert the change for now, since the version used in the repo is too old. I will let the maintainer deal with this.

PR #7772, opened 2025-08-23
[lexical] Bug Fix: Correct exit direction from decorator nodes in RTL

### LINKED ISSUE #7771: Bug: Selection movement incorrect for RTL exiting decorator node
https://github.com/user-attachments/assets/015b166a-9dc2-46b1-a599-c8f1cff2c33e

Lexical version: v0.34.0

## Steps To Reproduce

1. Write a right-to-left paragraph.
2. Insert some decorator node, e.g. a date node.
3. Try to move the cursor from the start to the end or vice versa.

## The current behavior

When the decorator node is selected and the right/left arrow key is pressed, the selection is always moved before/after.

## The expected behavior

If the paragraph is RTL and the right arrow key is pressed, the selection should move before instead of after, and the same being said for the left arrow key.

## Impact of fix
RTL users.

### LINKED ISSUE #7775: Bug: Selection movement incorrect for RTL going back from last node in paragraph
https://github.com/user-attachments/assets/6adbaad2-8c01-47af-9d66-02f2a32b9117

Lexical version: v0.34.0

## Steps To Reproduce

1. Write a right-to-left paragraph.
2. Insert some decorator node, e.g. a date node, at the end of the paragraph.
3. Try to move right from the end of the paragraph - the date node focus is skipped

## The current behavior

The selection skips the focus node, in fact, going in the opposite direction focuses it (in our case, at the end of the RTL paragraph, pressing the left arrow).

## The expected behavior

The focus should be on the date time element.

## Impact of fix
RTL users.

noamzaks:
## Description
This updates the right/left key arrow handlers to correct whether the previous/next node should be selected, for RTL scenarios.

Closes #7771.
Closes #7775.

## Test plan

Added a test for this.

### Before

https://github.com/user-attachments/assets/015b166a-9dc2-46b1-a599-c8f1cff2c33e

### After

https://github.com/user-attachments/assets/543d3d0a-468d-40b8-bc54-f0af870c74e9

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -36,6 +36,7 @@ import {
   html,
   initialize,
   insertCollapsible,
+  insertDateTime,
   insertHorizontalRule,
   insertImageCaption,
   insertSampleImage,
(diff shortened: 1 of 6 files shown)


noamzaks:
Switching to draft because tests are not working yet

noamzaks:
I set the test to skip on Firefox, but there's another bug (that's unmodified by the changes) in Firefox where selection _can't_ be after the DateTime component if it's at the end of an RTL paragraph.

Opened an issue for this in #7783.

Issue #7777, opened 2025-08-26
`LexicalNodeContextMenuPlugin` Blocking Right Click on whole screen


htmujahid:
This is causing blocking behavior in whole document in https://shadcn-editor.vercel.app

https://github.com/facebook/lexical/blob/9028e4151c52d3957bf2df456655a9ee147ca5cb/packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx#L251

Kepron:
same here

etrepum:
Should be a pretty straightforward fix, please feel free to submit a PR.

etrepum:
Fixed by #7778

PR #7778, opened 2025-08-26
[lexical-react] Fix the context menu event listener element

ivailop7:
Fix for #7777

### DIFF
--- a/packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx
@@ -248,12 +248,14 @@ const NodeContextMenuPlugin = forwardRef<
       setIsOpen(true);
     }
 
-    const editorElement = editor.getRootElement();
-    if (editorElement) {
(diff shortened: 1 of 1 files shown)


packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx
```diff
@@ -248,10 +248,12 @@ const NodeContextMenuPlugin = forwardRef<
...
-    document.addEventListener('contextmenu', onContextMenu);
-    return () => {
-      document.removeEventListener('contextmenu', onContextMenu);
-    };
+    const editorElement = editor.getRootElement();
+    if (editorElement) {
+      editorElement.addEventListener('contextmenu', onContextMenu);
+      return () =>
+        editorElement.removeEventListener('contextmenu', onContextMenu);
+    }
```
etrepum: this isn't coherent if the root element ever changes for whatever reason, this should use a RootListener one way or another (either outside with state or inside here)

etrepum (approved):

PR #7786, opened 2025-08-31
[lexical-playground] DatetimeNode DOM updates

ivailop7:
Fixes for importDOM & createDOM methods. Supports copy-pasted content from Google Docs that includes a date widget.


packages/lexical-playground/src/nodes/DateTimeNode/DateTimeComponent.tsx
```diff
@@ -47,8 +47,21 @@ export default function DateTimeComponent({
...
+  const [timeValue, setTimeValue] = useState(() => {
+    const hours = dateTime?.getHours();
+    const minutes = dateTime?.getMinutes();
+    if (hours !== 0 || minutes !== 0) {
+      return `${hours?.toString().padStart(2, '0')}:${minutes
+        ?.toString()
+        .padStart(2, '0')}`;
+    }
+    return '00:00';
+  });
```
etrepum: I think these are maybe not correct for `dateTime === undefined`? `includeTime` will be true if it is `undefined` and `timeValue` will be `undefined:undefined`

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -57,6 +57,14 @@ function $convertDateTimeElement(
     const node = $createDateTimeNode(new Date(Date.parse(dateTimeValue)));
     return {node};
   }
+  const gDocsDateTimePayload = domNode.getAttribute('data-rich-links');
+  if (gDocsDateTimePayload) {
+    const parsed = JSON.parse(gDocsDateTimePayload);
+    const node = $createDateTimeNode(
+      new Date(Date.parse(parsed?.dat_df?.dfie_dt || '')),
+    );
```
etrepum: might be worth checking if Date.parse is NaN because those invalid date objects are weird (all of the accessors return NaN) and will also be !== 0

etrepum (commented):
I didn't audit the code thoroughly to see where the dateTime can be undefined or how invalid input can occur but it's not very defensive against that sort of thing

ivailop7:
@etrepum addressed

Issue #7788, opened 2025-09-01
Bug: Lexical still focused after clicking outside in Firefox


taweber-acn:
Lexical version: v0.34.0

## Steps To Reproduce

1. Go to Lexical Playground in Firefox (142.0.1)
2. Clear Editor
3. Paste following content
**foo**
bar
4. Move Cursor to start of content and click
5. Move Mouse over second content row
6. Move Mouse to right until cursor is outside of the editor
7. Click

## The current behavior
Lexical is still focused

## The expected behavior
Lexical loose focus by clicking outside

RohithMacharla11:
👋 Hi, I’ve worked on a fix for this issue. The problem was that Firefox wasn’t properly triggering the expected blur event when clicking outside the editor. I’ve added a check to ensure the editor correctly dispatches BLUR_COMMAND in this case.

I’ve opened a PR here: [[#7798 ](https://github.com/facebook/lexical/pull/7798)] — would appreciate it if you could take a look. 🚀

taweber-acn:
Hi,

i tested under https://lexical-playground-git-fork-rohithmacharla1-28ba1c-fbopensource.vercel.app and unfortunately I can still reproduce the problem there.

PR #7798, opened 2025-09-05
fix(editor): ensure blur event triggers correctly in Firefox

RohithMacharla11:
This PR fixes an issue where the Lexical editor in Firefox stays focused even after clicking outside.

**Steps to Reproduce:**

Open Lexical Playground in Firefox.

Type some text.

Move the cursor inside, then click outside the editor.

The editor remains focused.

**Fix Implemented:**

Added a mousedown listener for Firefox to detect outside clicks.

Manually dispatches BLUR_COMMAND to align behavior with Chrome/Edge.

Before:
Editor stays focused in Firefox when clicking outside.

After:
Editor correctly loses focus in Firefox when clicking outside.

https://github.com/user-attachments/assets/061caa56-fe29-4c5f-8629-adb97aaa5797


etrepum (commented):
This PR contains unrelated changes that should be removed and according to the author of the issue it doesn't fix the problem https://github.com/facebook/lexical/issues/7788#issuecomment-3257452402

PR #7800, opened 2025-09-08
RFC [lexical-list] Fix: updating list type to/from checked should update child DOM elements

james-atticus:
## Description

`ListItemNode` reconciliation depends on the properties of the parent node. Specifically, attributes are added/removed depending on whether the list type is `checked`. If the type changes to/from `checked` but the children are not manually marked as dirty, then they are not updated.

One option is a `ListNode` node transform which marks the children as dirty, however that doesn't work for collab which sets node properties directly.

Another option, which I've implemented in this PR to demonstrate, is to add a function to `LexicalNode` which allows a node to indicate to the reconciler that it depends on its parent in some way. The reconciler can check this when iterating `ElementNode` children and reconcile such nodes.

## Test plan

Added an example failing test case which passes with the reconciler change.


packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -126,15 +126,16 @@ export class ListItemNode extends ElementNode {
     return element;
   }
 
+  domDependsOnParent(): true {
```
james-atticus: Not at all attached to this name, just the first thing that came to mind.

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -126,15 +126,16 @@ export class ListItemNode extends ElementNode {
...
+    return true;
+  }
+
   updateListItemDOM(
     prevNode: ListItemNode | null,
     dom: HTMLLIElement,
     config: EditorConfig,
   ) {
-    const parent = this.getParent();
-    if ($isListNode(parent) && parent.getListType() === 'check') {
```
james-atticus, line 135: Drive-by: if the parent becomes not-`check`, then we should be removing the check attributes.

etrepum:
I don't think it's more elegant but this dependency could also be expressed with a MutationListener or updateDOM change to ListNode that manually calls updateDOM on its existing children when checked changes. I think we might have some other ways to express this kind of thing once extensions land, so maybe we should avoid adding new code to the reconciler in the meantime?

james-atticus:
Forgot `updateDOM` could be used for this. It already had a `prevNode.__tag !== this.__tag` check so I've added a check for `__listType` too.

packages/lexical-list/src/__tests__/unit/LexicalListNode.test.ts
```diff
@@ -301,6 +302,32 @@ describe('LexicalListNode tests', () => {
       });
     });
 
+    test('Should update list children when switching from checklist to bullet', async () => {
+      const {editor} = testEnv;
+
+      await waitForReact(() => {
```
etrepum, line 308: fwiw `{discrete: true}` or even just `await` should be sufficient for the updateDOM to have already occurred (since the deferred update was scheduled earlier on the microtask queue than the await), `waitForReact` is probably only useful for decorators

PR #7804, opened 2025-09-08
[lexical][lexical-playground] Feature: Improve focus management in Toolbar and adds SKIP_SELECTION_FOCUS_TAG

KaiPrince:
## Description

In this PR, I am addressing the issue of focus being moved to the editor when a toolbar button is activated with the keyboard, as reported in #6006.

I have made the following behavioural changes:

1. Focus outlines are visible on the toolbar buttons
2. When activating a toolbar button (either with space or enter), the focus remains on the button
3. When activating a toolbar dropdown item, the focus returns to the dropdown opener button

This PR partially addresses #6006.

This PR does not address the behaviour of modals such as the insert image button, or the trap created by the tab intent plugin.

### SKIP_SELECTION_FOCUS_TAG

Normally when an editor.update reconciles its selection to the DOM, the focus is moved to the rootElement if it does not already have focus. In order to ensure that the DOM selection changes but the focus remains on the button, a new `SKIP_SELECTION_FOCUS_TAG` has been added which is more fine-grained than `SKIP_DOM_SELECTION` which avoids changing the DOM selection or focus.

### Before

Toolbar button:

Before | After
--------|-------
image | image

Dropdown button:

Before | After
--------|-------
image | image

### After

Toolbar button:

Before | After
--------|-------
image | image

Dropdown button:

Before | After
--------|-------
image| image


packages/lexical-playground/src/index.css
```diff
@@ -1511,6 +1511,10 @@ button.action-button:disabled {
   overflow-y: hidden; /* disable vertical scroll*/
 }
 
+.toolbar .toolbar-item:focus {
+  outline: 2px solid rgb(60, 132, 244);
```
KaiPrince, line 1442: I couldn't find a standardized focus colour, so I copied this colour from https://github.com/facebook/lexical/blob/9571a9e5bd420aeab4359110203ba89a12a49399/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css#L585

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -73,7 +75,7 @@ function DropDownItems({
   onClose,
 }: {
   children: React.ReactNode;
-  dropDownRef: React.Ref<HTMLDivElement>;
+  dropDownRef: React.RefObject<HTMLDivElement>;
```
KaiPrince: Constrained the accepted type here so that the below `.current` would type-match, and so I won't be fired.

image

image

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -88,12 +90,14 @@ function DropDownItems({
   );
 
   const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    const key = event.key;
+    if (key === 'Escape') {
+      onClose();
+    }
```
KaiPrince, line 99: Moved this above the `(!items)` check, because some dropdowns, e.g. the colour picker don't have any buttons registered, but should still be closable.

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -132,8 +136,12 @@ function DropDownItems({
 
     if (highlightedItem && highlightedItem.current) {
       highlightedItem.current.focus();
+    } else if (!items) {
+      if (dropDownRef && dropDownRef.current) {
+        focusNearestDescendant(dropDownRef.current);
```
KaiPrince: For colour pickers, etc. we want to focus the first field.

packages/lexical-playground/src/utils/focusUtils.ts
KaiPrince, line 1: Created some utility functions for later reuse. Alternatively, could have just added a dependency to a library like tabbable

etrepum (commented):
It looks like this is failing a lot of tests

KaiPrince:
Hmm, yes.. I'll look into it. I'm betting it's just the tests themselves that need to be updated.

KaiPrince:
Nope I take that back it's not the tests it's me.

packages/lexical-playground/src/ui/DropDown.tsx
etrepum: That's not what that means, those are just internal types that you shouldn't use as an API, it's fine for react internals to use those types. The change is correct for other reasons, since the code here expects exactly a RefObject (and not the callback style of ref)

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -132,8 +136,12 @@ function DropDownItems({
 
     if (highlightedItem && highlightedItem.current) {
       highlightedItem.current.focus();
+    } else if (!items) {
+      if (dropDownRef && dropDownRef.current) {
```
etrepum: I understand you're just copying code from elsewhere but these ref objects are always defined (at least when using typescript), no reason to check
```suggestion
      if (dropDownRef.current) {
```

packages/lexical-playground/src/utils/focusUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+      childNode as HTMLElement,
+    );
+
+    if (focusableChild) {
+      return focusableChild;
+    }
+  }
+
+  return null;
+};
```
etrepum: pretty sure that this whole thing can be done with a call to querySelector?

packages/lexical-playground/src/utils/focusUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+): HTMLElement | null => {
+  const el = findFirstFocusableDescendant(startElement);
+  el?.focus();
+  return el;
+};
+
+export const isKeyboardInput = (
+  event: MouseEvent | PointerEvent | React.MouseEvent,
+): boolean => {
+  if (event instanceof PointerEvent) {
```
etrepum: generally speaking it's best to avoid `instanceof` checks directly on DOM related classes because it will be false if that event was generated from a different iframe or shadow root

packages/lexical/src/LexicalUpdateTags.ts
```diff
@@ -52,6 +52,8 @@ export const SKIP_SCROLL_INTO_VIEW_TAG = 'skip-scroll-into-view';
  */
 export const SKIP_DOM_SELECTION_TAG = 'skip-dom-selection';
 
+export const SKIP_SELECTION_FOCUS_TAG = 'skip-selection-focus';
```
etrepum, line 59: This should have a jsdoc comment so it gets API docs

etrepum:
Overall the usability of this seems nice, I just think a few things could be cleaned up!

I did notice that the 'Click to add below' button can get focus (after the editor) even when it's not visible, e.g. with the editor focused and a wide enough browser if you press escape and then tab it will look like nothing is focused but if you check the activeElement you'll see it's that button (but its container has the style `opacity: 0; transform: translate(-10000px, -10000px);` instead of something that would prevent focus/input).

I don't spend a lot of time thinking about keyboard UX specifically but it might be nice to have some way to quickly send focus back to the editor, not sure if there's a standard key sequence in other apps for that. The first thing I thought of would be that escape could do that (if you're on some other part of the editor's UI, like the toolbar), since right now with the tab plugin you're already using escape to move focus out of the editor.

packages/lexical-playground/src/utils/focusUtils.ts
KaiPrince: Yeah probably. I've always seen it done this way so I assumed it was for performance reasons. I'll swap it with a query selector

KaiPrince:
The "click to add below" button is something I plan on addressing in a future PR while I'm also working adding escaping to the tab indentation plugin. I have a separate dev branch where I address it by swapping the inline styles applied to the element with class names so that I can just add `:not(:focused)`.

As far as quickly moving focus back to the editor, I think that would best be addressed by switching the toolbar's interaction model to use arrow keys instead of tab. That way the tab key would move you past the whole toolbar and back to the editor, rather than having to tab through each control. See [Toolbar Pattern | ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/) for an example.

packages/lexical-playground/src/utils/focusUtils.ts
KaiPrince: Huh, TIL

packages/lexical-playground/src/ui/ColorPicker.tsx
```diff
@@ -96,24 +112,16 @@ export default function ColorPicker({
...
+    emitOnChange(newColor.hex);
   };
 
-  useEffect(() => {
-    // Check if the dropdown is actually active
-    if (innerDivRef.current !== null && onChange) {
-      onChange(selfColor.hex, skipAddingToHistoryStack);
-      setInputColor(selfColor.hex);
-    }
-  }, [selfColor, onChange]);
```
KaiPrince, line 107: I removed this because it seems to go against React guidelines on useEffect: https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers. I've replaced it with individual calls in each event handler.

As a side effect, this also fixed the odd behaviour that selecting text and then opening the color picker splits the text node even if you don't change the text color.

packages/lexical-playground/src/ui/DropDown.tsx
KaiPrince: I was joking about being fired of course, but even based on what you said, this project probably shouldn't be relying on react internals.

packages/lexical-playground/src/ui/ColorPicker.tsx
```diff
@@ -96,24 +112,16 @@ export default function ColorPicker({
...
+    const newColor = transformColor('hex', basicColor);
 
-  useEffect(() => {
-    if (color === undefined) {
-      return;
-    }
-    const newColor = transformColor('hex', color);
     setSelfColor(newColor);
     setInputColor(newColor.hex);
-  }, [color]);
```
KaiPrince, line 116: Similarly, this seems to go against React's recommendations: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

packages/lexical-playground/src/ui/DropDown.tsx
etrepum: It's literally not. Ref is the more general type that accepts both types of refs (callback and object), plus some internal stuff that you don't need to think about. That sort of naming means that you shouldn't write it yourself, it's fine if it gets pulled in by using the public API.

etrepum:
npm run ci-check is failing:
```
Error: packages/lexical-playground/src/ui/DropDown.tsx(281,13): error TS2322: Type 'RefObject<HTMLDivElement | null>' is not assignable to type 'RefObject<HTMLDivElement>'.
  Type 'HTMLDivElement | null' is not assignable to type 'HTMLDivElement'.
    Type 'null' is not assignable to type 'HTMLDivElement'.
ERROR: "tsc" exited with 2.
```

https://github.com/facebook/lexical/actions/runs/18002165768/job/51264652086?pr=7804

KaiPrince:
Huh. I'm not seeing that issue locally. Let me investigate

packages/lexical-playground/src/ui/DropDown.tsx
KaiPrince: Ah okay gotcha 👍

packages/lexical-playground/src/ui/DropDown.tsx
```diff
@@ -253,7 +276,10 @@ export default function DropDown({
 
       {showDropDown &&
         createPortal(
-          <DropDownItems dropDownRef={dropDownRef} onClose={handleClose}>
+          <DropDownItems
+            dropDownRef={dropDownRef as React.RefObject<HTMLDivElement>}
```
etrepum: You shouldn't need to add a cast, casts are unsafe. The correct place to fix this are the types themselves. This is a direct consequence of the Ref vs. RefObject thing you keep trying to fix but was actually correct before and now it isn't.

```
    type Ref<T> = RefCallback<T> | RefObject<T | null> | null;
```

The definition of `RefCallback` contains the internal implementation details that you keep running away from, but using `Ref` and `RefCallback` as-is are perfectly correct and is not something that needs to be avoided or "fixed".

KaiPrince: Hmm yes I'm not pleased with using a cast here either. Based on observation the casted type seemed to be correct, but now that you pasted the type definition, it seems like I have the wrong version of react installed because I get a different type def:

image

Could you link me the source for the definition you pasted?

Oh and just to clarify I'm not trying to run away from RefCallback, I'm just trying to best utilize the type system by narrowing the acceptable types to what we actually would expect. Having a stricter interface keeps the implementation simple since we don't have to add behaviour to handle unexpected inputs. e.g. `DropDownItems` is an internal component of `Dropdown`, which we know only uses RefObject and never RefCallback, so by being more strict on the input type we don't need to add a check for handling RefCallback, leading to simpler code.

KaiPrince: I'll try resetting my `node_modules` and running `npm ci`. In the meantime, adjusting the prop type to match the version you pasted fixes it in CI: https://github.com/KaiPrince/lexical/pull/2/commits/3cd1c6c9d9815096ae9c2db317a1378b849d21f6

If you still feel strongly that I should revert it back to just `Ref`, let me know.

packages/lexical/src/LexicalUpdateTags.ts
```diff
@@ -52,6 +52,12 @@ export const SKIP_SCROLL_INTO_VIEW_TAG = 'skip-scroll-into-view';
  */
 export const SKIP_DOM_SELECTION_TAG = 'skip-dom-selection';
 
+/**
+ * Indicated that after changing the selection, the editor should not focus itself
+ * This tag is ignored if SKIP_DOM_SELECTION_TAG is used
```
etrepum:
```suggestion
 * Indicates that after changing the selection, the editor should not focus itself
 * This tag is ignored if {@link SKIP_DOM_SELECTION_TAG} is used
```

PR #7812, opened 2025-09-10
[lexical-markdown] Fix: errors in shouldMergeAdjacentLines

GermanJablo:
WIP

## Description

Closes #

### Before

### After

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -198,17 +198,20 @@ export type TextMatchTransformer = Readonly<{
   type: 'text-match';
 }>;
 
-const ORDERED_LIST_REGEX = /^(\s*)(\d{1,})\.\s/;
+const EMPTY_OR_WHITESPACE_ONLY = /^[\t ]*$/;
+const ORDERED_LIST_REGEX = /^(\s*)(\d+)\.\s/;
 const UNORDERED_LIST_REGEX = /^(\s*)[-*+]\s/;
 const CHECK_LIST_REGEX = /^(\s*)(?:-\s)?\s?(\[(\s|x)?\])\s/i;
 const HEADING_REGEX = /^(#{1,6})\s/;
 const QUOTE_REGEX = /^>\s/;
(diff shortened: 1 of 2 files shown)


packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -212,6 +212,9 @@ const CODE_SINGLE_LINE_REGEX =
   /^[ \t]*```[^`]+(?:(?:`{1,2}|`{4,})[^`]+)*```(?:[^`]|$)/;
 const TABLE_ROW_REG_EXP = /^(?:\|)(.+)(?:\|)\s?$/;
 const TABLE_ROW_DIVIDER_REG_EXP = /^(\| ?:?-*:? ?)+\|\s?$/;
+const TAG_START_REGEX = /^<[a-z_][\w-]*(?:\s[^<>]*)?\/?>/i;
+const TAG_END_REGEX = /^<\/[a-z_][\w-]*\s*>/i;
+const ENDS_WITH = (regex: RegExp) => new RegExp(`${regex.source}$`);
```
etrepum: There's a subtle bug in here because `regex.source` doesn't carry the flags, so it loses case insensitivity. You probably mean this:

```suggestion
const ENDS_WITH = (regex: RegExp) => new RegExp(`(?:${regex.source})$`, regex.flags);
```

I also added the grouping to prevent a future bug because if the source was something like `/foo|bar/` this would transform it to `/foo|bar$/` which is not the same (only "bar" has to be at the end).

It's a real bug in this implementation, the test cases in this PR just don't cover it.

```
> const TAG_START_REGEX = /^<[a-z_][\w-]*(?:\s[^<>]*)?\/?>/i;
> const ENDS_WITH_TAG_START_REGEX = new RegExp(`${TAG_START_REGEX.source}$`);
> ENDS_WITH_TAG_START_REGEX.test('<foo>')
true
> ENDS_WITH_TAG_START_REGEX.test('<Foo>')
false
> TAG_START_REGEX.test('<Foo>')
true
```

GermanJablo: You are right, thank you!

PR #7814, opened 2025-09-13
[lexical][lexical-selection][lexical-utils] Refactor: Consolidate ancestor lookup via findMatchingParent

### LINKED ISSUE #5311: Duplicate function `$findMatchingParent` and `$getAncestor`

cnaples79:
## Description

- Current behavior: `$findMatchingParent` and `$getAncestor` logic lived in multiple places (`@lexical/utils`, `@lexical/selection`), leading to duplicated implementations and layering violations. This duplication also surfaced as duplicate symbol/implementation errors in CI when both versions were in play.

- Changes in this PR:
  - Move a single, typed implementation of `$findMatchingParent` into core (`packages/lexical/src/LexicalUtils.ts`) with overloads (type‑predicate support).
  - Export `$findMatchingParent` from core (`packages/lexical/src/index.ts`).
  - `@lexical/utils`: re‑export `$findMatchingParent` from `lexical` and remove the duplicate implementation.
  - `@lexical/selection`: provide a local exported `$getAncestor` wrapper that delegates to core `$findMatchingParent` (import from `lexical`), avoiding a utils dependency and removing local duplication.
  - Remove stray PR‑only comment in `@lexical/link` (no functional change).

No runtime behavior changes are intended; this is a consolidation/refactor to respect package layering and eliminate duplicate code.

Closes #5311

### Before

- Multiple implementations of ancestor lookup existed across packages.
- CI surfaced TypeScript/esbuild errors like duplicate symbol/implementation for `$findMatchingParent` when both versions compiled.

### After

- A single source of truth for ancestor lookup in core, with utils re‑export and selection delegating wrapper.
- Packages compile against the unified helper; behavior remains the same and is covered by existing unit/e2e tests (e.g., selection, links, lists, tables).

*No visual changes.*

### DIFF
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -645,7 +645,7 @@ export function $toggleLink(
     const firstNode = nodes[0];
     // if the first node is a LinkNode or if its
     // parent is a LinkNode, we update the URL, target and rel.
-    const linkNode = $getAncestor(firstNode, $isLinkNode);
+    const linkNode = $findMatchingParent(firstNode, $isLinkNode);
     if (linkNode !== null) {
       return updateLinkNode(linkNode);
     }
@@ -657,7 +657,7 @@ export function $toggleLink(
       if (!node.isAttached()) {
         continue;
       }
-      const parentLinkNode = $getAncestor(node, $isLinkNode);
+      const parentLinkNode = $findMatchingParent(node, $isLinkNode);
       if (parentLinkNode) {
         updateLinkNode(parentLinkNode);
         continue;
@@ -701,16 +701,7 @@ export function $toggleLink(
 /** @deprecated renamed to {@link $toggleLink} by @lexical/eslint-plugin rules-of-lexical */
 export const toggleLink = $toggleLink;
 
-function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
-  node: LexicalNode,
-  predicate: (ancestor: LexicalNode) => ancestor is NodeType,
-) {
-  let parent = node;
-  while (parent !== null && parent.getParent() !== null && !predicate(parent)) {
-    parent = parent.getParentOrThrow();
-  }
-  return predicate(parent) ? parent : null;
-}
+// Removed duplicate $getAncestor in favor of shared $findMatchingParent from @lexical/utils
 
 const PHONE_NUMBER_REGEX = /^\+?[0-9\s()-]{5,}$/;
 
--- a/packages/lexical-link/src/index.ts
+++ b/packages/lexical-link/src/index.ts
@@ -701,8 +701,6 @@ export function $toggleLink(
 /** @deprecated renamed to {@link $toggleLink} by @lexical/eslint-plugin rules-of-lexical */
 export const toggleLink = $toggleLink;
 
-// Removed duplicate $getAncestor in favor of shared $findMatchingParent from @lexical/utils
-
 const PHONE_NUMBER_REGEX = /^\+?[0-9\s()-]{5,}$/;
 
 /**
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -44,16 +44,6 @@ import {
   getStyleObjectFromCSS,
 } from './utils';
 
-// Local helper to preserve existing callers. Delegates to the shared
-// implementation available from core without introducing a dependency on
-// @lexical/utils (per package layering guidance).
-function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
-  node: LexicalNode,
-  predicate: (ancestor: LexicalNode) => ancestor is NodeType,
-) {
-  return $findMatchingParent(node, predicate) as NodeType | null;
-}
-
 export function $copyBlockFormatIndent(
   srcNode: ElementNode,
   destNode: ElementNode,
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -21,6 +21,7 @@ import {
   $caretFromPoint,
   $createRangeSelection,
   $extendCaretToRange,
+  $findMatchingParent,
   $getPreviousSelection,
   $getSelection,
   $hasAncestor,
@@ -36,7 +37,6 @@ import {
   INTERNAL_$isBlock,
 } from 'lexical';
 import invariant from 'shared/invariant';
-import { $findMatchingParent } from '@lexical/utils';
 
 import {
   $getComputedStyleForElement,
--- a/packages/lexical-selection/src/range-selection.ts
+++ b/packages/lexical-selection/src/range-selection.ts
@@ -44,6 +44,16 @@ import {
   getStyleObjectFromCSS,
 } from './utils';
 
+// Local helper to preserve existing callers. Delegates to the shared
+// implementation available from core without introducing a dependency on
+// @lexical/utils (per package layering guidance).
+function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
+  node: LexicalNode,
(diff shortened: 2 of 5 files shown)


packages/lexical-selection/src/range-selection.ts
```diff
@@ -36,6 +36,7 @@ import {
   INTERNAL_$isBlock,
 } from 'lexical';
 import invariant from 'shared/invariant';
+import { $findMatchingParent } from '@lexical/utils';
```
etrepum: @lexical/selection can't depend on @lexical/utils

etrepum:
The follow-up work, or equivalent, is necessary for this to work at all because `@lexical/utils` depends on `@lexical/selection` so you can't have `@lexical/selection` depend on `@lexical/utils`.

cnaples79:
@etrepum I see! Thanks for the feedback, I'll update the fix and PR.

cnaples79:
Thanks for the pointer on package layering, @etrepum. I’ve updated the PR to avoid the circular dependency by moving  into core (), exporting it via , and re-exporting from .\n\nChanges: \n- Add  to  with the same typed overloads and docs.\n- Export from .\n- Update  to import from  (not ).\n- Re-export  from  and remove the duplicate local implementation.\n\nThis keeps  free of a dependency on , removes the duplication, and aligns with the guidance from #5311. Happy to iterate further if you'd prefer a different placement or naming.

cnaples79:
Pushed a follow-up to fix the Vercel failure: removed the duplicate  const and added proper overloads to the existing function in  (single implementation). This should unblock the playground build. If there’s a preferred location or naming for the overloads/docs, happy to tweak.

cnaples79:
Found and fixed the Vercel build failure! 🎉

**Issue**: There were duplicate function declarations causing TypeScript compilation errors:
- Overload declarations on lines 981-988 ✅
- Duplicate `export function` declaration on line 1789 ❌

**Fix**: Removed the `export` keyword from line 1789. The function is already properly exported via the overload declarations above.

The TypeScript pattern should be:
1. Export overload declarations 
2. Single implementation function (without export keyword)

This should resolve the build failure. The playground build was already working, so this was specifically affecting the main lexical package build.

packages/lexical-link/src/index.ts
```diff
@@ -701,16 +701,7 @@ export function $toggleLink(
...
-  node: LexicalNode,
-  predicate: (ancestor: LexicalNode) => ancestor is NodeType,
-) {
-  let parent = node;
-  while (parent !== null && parent.getParent() !== null && !predicate(parent)) {
-    parent = parent.getParentOrThrow();
-  }
-  return predicate(parent) ? parent : null;
-}
+// Removed duplicate $getAncestor in favor of shared $findMatchingParent from @lexical/utils
```
etrepum: No need for this comment, only really makes sense in the context of the PR, not the remaining code
```suggestion
```

packages/lexical-selection/src/range-selection.ts
```diff
@@ -646,9 +647,6 @@ export function $getAncestor<NodeType extends LexicalNode = LexicalNode>(
   node: LexicalNode,
   predicate: (ancestor: LexicalNode) => ancestor is NodeType,
 ) {
-  let parent = node;
-  while (parent !== null && parent.getParent() !== null && !predicate(parent)) {
-    parent = parent.getParentOrThrow();
-  }
-  return predicate(parent) ? parent : null;
+  // Delegate to shared implementation to avoid duplication.
+  return $findMatchingParent(node, predicate) as NodeType | null;
```
etrepum: This doesn't remove `$getAncestor` or fix the type of `$findMatchingParent` per https://github.com/facebook/lexical/issues/5311#issuecomment-2512076499

packages/lexical-utils/src/index.ts
```diff
@@ -415,31 +415,7 @@ export type DOMNodeToLexicalConversionMap = Record<
...
-    if (findFn(curr)) {
-      return curr;
-    }
-
-    curr = curr.getParent();
-  }
-
-  return null;
-};
+export { $findMatchingParent } from 'lexical';
```
etrepum: This doesn't work, it's used in this module so you need to import it then separately export it.

Are you running this code to test it? Pretty sure `npm run ci-check` would've caught this pretty quickly

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1764,15 +1786,15 @@ export function $splitNode(
...
+function $findMatchingParent(
   startingNode: LexicalNode,
   findFn: (node: LexicalNode) => boolean,
 ): LexicalNode | null {
   let curr: ElementNode | LexicalNode | null = startingNode;
 
   while (curr !== $getRoot() && curr != null) {
     if (findFn(curr)) {
-      return curr;
+      return curr as LexicalNode; // T is inferred via overload when appropriate
```
etrepum: This cast seems redundant, and all casts are potentially unsafe so you should avoid them unless absolutely necessary

cnaples79:
I've updated the PR and addressed your feedback @etrepum!

packages/lexical/src/LexicalUtils.ts
```diff
@@ -2141,3 +2124,37 @@ export function $create<T extends LexicalNode>(klass: Klass<T>): T {
...
+    startingNode: LexicalNode,
+    findFn: (node: LexicalNode) => boolean,
+  ): LexicalNode | null;
+} = (
+  startingNode: LexicalNode,
+  findFn: (node: LexicalNode) => boolean,
+): LexicalNode | null => {
+  let curr: ElementNode | LexicalNode | null = startingNode;
+
+  while (curr !== $getRoot() && curr != null) {
```
etrepum: I know that this is just code moved from elsewhere but this is more correct because getLatest isn't called on startingNode first.

```suggestion
  while (curr != null && !$isRootNode(curr)) {
```

packages/lexical/src/LexicalUtils.ts
```diff
@@ -2141,3 +2124,37 @@ export function $create<T extends LexicalNode>(klass: Klass<T>): T {
...
   return new registeredNode.klass() as T;
 }
+
+/**
+ * Starts with a node and moves up the tree (toward the root node) to find a matching node based on
+ * the search parameters of the findFn. (Consider JavaScripts' .find() function where a testing function must be
+ * passed as an argument. eg. if( (node) => node.__type === 'div') ) return true; otherwise return false
+ * @param startingNode - The node where the search starts.
+ * @param findFn - A testing function that returns true if the current node satisfies the testing parameters.
+ * @returns A parent node that matches the findFn parameters, or null if one wasn't found.
```
etrepum: This just makes the description more precise, its treatment of the startingNode and RootNode are not covered by the existing documentation.

```suggestion
 * @returns `startingNode` or one of its ancestors that matches the `findFn` predicate and is not the `RootNode`, or `null` if no match was found.
```

etrepum (approved):
Made some suggestions for improvements to the existing code. Audited our existing usage of `$getAncestor` (not publicly exported, and does not depend on the implementation differences with `$findMatchingParent`, so it's fine to remove).

etrepum:
This looks good, other than the PR title and description. Can you change these to match the repository's pull request template? It's important for how we put together the release notes.

https://raw.githubusercontent.com/facebook/lexical/refs/heads/main/.github/pull_request_template.md

cnaples79:
@etrepum done! Please let me know if any other changes to the PR need to be made! And thank you again for your feedback.

PR #7816, opened 2025-09-15
[lexical-yjs][lexical-playground] Breaking change: remove clientID from collab context

james-atticus:
## Description

Removes `clientID` from collaboration context. This ID was unreliable as it could be updated by nested collab editors. Consumers should retrieve the client ID by looking up the correct document in the context's `yjsDocMap`, or switch to using a different ID.

Also updated `PollComponent` to use name rather than client ID, so two clients with the same name will update each other's vote, which arguably is more correct.

### DIFF
--- a/packages/lexical-playground/src/nodes/PollComponent.tsx
+++ b/packages/lexical-playground/src/nodes/PollComponent.tsx
@@ -24,7 +24,6 @@ import {
   COMMAND_PRIORITY_LOW,
   NodeKey,
 } from 'lexical';
-import * as React from 'react';
 import {useEffect, useMemo, useRef, useState} from 'react';
 
 import Button from '../ui/Button';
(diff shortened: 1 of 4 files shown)


etrepum (commented):
It seems like there should be some sort of context where you get access to the binding (and if it's changed), but I don't see one. I wonder if there is some sort of need to get the clientID?

I looked a little closer and was horrified to see that CollaborationContext is just a global variable in a React Context disguise, so none of the changes to it are observable and it's not safe for multiple collaboration instances. You can tell because CollaborationContext.Provider is never used!

james-atticus:
@etrepum I don't think we can do that, at least with this context provider. Nested editors use the same collab context (because they need the same yjsDocMap) but create their own binding.

If someone needs the clientID, they can get it from `yjsDocMap.get(id).clientID`. Well, in theory... because of [#2637](https://github.com/facebook/lexical/issues/2637), [#6271](https://github.com/facebook/lexical/issues/6271) and strict mode, the doc gets removed from the map in dev. I'll fix that in v2 but I don't see a good way to fix it in v1 without a big breaking change.

etrepum:
I think this seems reasonable but we should probably get some feedback from a major consumer of collab. @ivailop7?

ivailop7 (approved):
We don't rely on it. LGTM.

PR #7818, opened 2025-09-15
[lexical-react] Breaking change: enforce use of collab context provider

james-atticus:
## Description

Flagged by @etrepum in #7816, `useCollaborationContext` didn't enforce that a provider was used. Indeed, neither the playground nor any of the examples used a provider. This could lead to significant issues with collab if two editors with the same ID were used on the page.

---

### `useCollaborationContext` now requires that a context provider is used

Users of the `<CollaborationPlugin>` or `useCollaborationContext` must add a wrapping `<LexicalCollaboration>` component. For example:

```typescript
  <LexicalCollaboration>
    <RichTextPlugin ... />
    <CollaborationPlugin id="lexical-editor" ... />
  <LexicalCollaboration>
</LexicalComposer>
```

The collaboration context is shared for nested editors, so the provider must only be used on the top-level editor.


packages/lexical-react/src/__tests__/unit/LexicalCollaborationPlugin.test.tsx
```diff
@@ -65,21 +66,25 @@ describe(`LexicalCollaborationPlugin`, () => {
...
-            shouldBootstrap={false}
-          />
-          <RichTextPlugin
-            contentEditable={<ContentEditable className="editor-input" />}
-            placeholder={
-              <div className="editor-placeholder">Enter some rich text...</div>
-            }
-            ErrorBoundary={LexicalErrorBoundary}
-          />
+          <LexicalCollaboration>
```
etrepum: I think this might really need to go around LexicalComposer because if any DecoratorNode needs the context it might not find it from here

tsconfig.json
```diff
@@ -62,7 +62,7 @@
         "./packages/lexical-react/src/LexicalClickableLinkPlugin.tsx"
       ],
       "@lexical/react/LexicalCollaborationContext": [
-        "./packages/lexical-react/src/LexicalCollaborationContext.ts"
+        "packages/lexical-react/src/LexicalCollaborationContext.tsx"
```
etrepum, line 65: I think this is your doc failure

etrepum: (or some other tsconfig file is not appropriately changed)

packages/lexical-react/src/LexicalCollaborationContext.tsx
```diff
@@ -38,19 +39,37 @@ const entries = [
...
 
 export function useCollaborationContext(
   username?: string,
   color?: string,
 ): CollaborationContextType {
   const collabContext = useContext(CollaborationContext);
+  invariant(
+    collabContext != null,
+    'useCollaborationContext: no context provider found',
+  );
```
etrepum: We could possibly make this transition a little smoother by transitionally having a global context default that we use when the context is null and print a warning. There's `devInvariant` which compiles to an error in dev and warning in prod that would maybe be good for this. The warning or error could also have a reference to this PR or some brief instructions to wrap your composer with `<LexicalCollaboration>`

etrepum (approved):
This looks right to me but we could make the transition a bit easier if we choose to

etrepum (dismissed):

etrepum:
Will approve again once the conflict from merging #7816 is resolved

etrepum (approved):

PR #7790, opened 2025-09-01
Add Lexical Shadow DOM Support

aleksandr-konovalov:
# Lexical Shadow DOM Support - Complete Implementation

Complete Shadow DOM support for Lexical editor implemented through a systematic 4-phase development approach.

## 🎯 Implementation Overview

This document describes the complete Shadow DOM support implementation in Lexical, developed through 4 systematic phases:

1. **[Phase 1: Core Infrastructure](#phase-1-core-infrastructure)** - Fundamental Shadow DOM APIs and ecosystem integration
2. **[Phase 2: Browser Compatibility](#phase-2-browser-compatibility)** - Cross-browser API standardization  
3. **[Phase 3: Deletion Commands](#phase-3-deletion-commands)** - Character and line deletion support
4. **[Phase 4: Word Operations](#phase-4-word-operations)** - Advanced word-level deletion

## 📊 Complete Feature Matrix

| Feature | Standard DOM | Shadow DOM | Implementation Status |
|---------|--------------|------------|----------------------|
| **Text Input** | ✅ | ✅ | Phase 1 - Complete |
| **Text Selection** | ✅ | ✅ | Phase 1 - Complete |
| **Character Deletion** | ✅ | ✅ | Phase 3 - Complete |
| **Word Deletion** | ✅ | ✅ | Phase 4 - Complete |
| **Line Deletion** | ✅ | ✅ | Phase 3 - Complete |
| **Copy/Paste** | ✅ | ✅ | Phase 1 - Complete |
| **Formatting** | ✅ | ✅ | Phase 1 - Complete |
| **All Plugins** | ✅ | ✅ | Phase 1 - Complete |

## 🌐 Browser Support Matrix

| Browser | getComposedRanges | Support Level |
|---------|-------------------|---------------|
| **Chrome 125+** | ✅ Primary | Full Support |
| **Firefox 132+** | ✅ Primary | Full Support |
| **Safari 17+** | ✅ Primary | Full Support |
| **Edge 125+** | ✅ Primary | Full Support |
| **Older Browsers** | ❌ | Graceful Degradation |

> **Note**: The experimental `shadowRoot.getSelection()` API was intentionally not used due to limited browser support and unstable behavior.

### Testing in Playground

```bash
npm run dev
# Open http://localhost:3000
# Click ⚙️ Settings → Toggle "Shadow DOM"
# Test all features: typing, deletion, formatting
```

### Zero-Config Integration

```typescript
import { createEditor } from 'lexical';

[...]

PR #7822, opened 2025-09-17
[lexical-clipboard] Fix: copy correct selection when editor in different window/document

james-atticus:
## Description

When the editor is mounted in a different window, for example when using [react-new-window](https://www.npmjs.com/package/react-new-window), then `copyToClipboard` copies the selection from the main window.

## Test plan

Tried to get a repro working with unit tests but couldn't get selection working with JSDOM.

### Before

Copy using context menu gets the text from the parent window's selection:

https://github.com/user-attachments/assets/5eaeed0c-2910-4dd9-8315-c2aae9e4665f

### After

https://github.com/user-attachments/assets/d30c01e7-3b11-4c26-9995-82a9bf1b3cd1

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -468,7 +468,7 @@ export async function copyToClipboard(
 
   const rootElement = editor.getRootElement();
   const editorWindow = editor._window || window;
-  const windowDocument = window.document;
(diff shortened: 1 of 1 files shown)


ivailop7 (approved):

etrepum:
#7790 probably fixes some more of these issues

etrepum:
(or at least something like #7790, that someone is interested in actually finishing)

PR #7830, opened 2025-09-19
[lexical-website] Chore: add example for how to bootstrap collab ydoc on the server

james-atticus:
## Description

`shouldBootstrap` is not recommended for use, but there's currently no examples of how to bootstrap a document on the server. This change adds a basic example to the website docs.

Split out from https://github.com/facebook/lexical/pull/7616/files#r2360141159


etrepum (approved):
This is definitely an improvement but I think the real win would be a runnable examples/* project that has a full stack project with a websocket server that uses this bootstrap function (or perhaps refactor one of the existing collab examples accordingly)

etrepum (approved):

Issue #4960, opened 2023-09-02
Feature: Server-Side Rendering (SSR) Support for Lexical Editor


agarwalamn:
### Use Case:

We are currently working on a project similar to Reddit, where our users utilize the Lexical Editor to create posts and comments. These posts are then displayed on their timelines using a read-only editor. This is a fundamental use case where we believe Server-Side Rendering (SSR) support is essential. SSR would greatly benefit us by allowing us to render these posts on the server, leading to improvements in user experience and SEO.

### Current Issue:
Our observation is that the Lexical Editor's rendering is currently limited exclusively to the client-side due to its reliance on client-side rendering mechanisms. Consequently, posts are only rendered on the client side, which can have some drawbacks.

### Challenges:
One of the challenges we've encountered is the time it takes for the editor to process data on the client side. This processing delay leads to noticeable UI jumps after content is loaded, which is not an ideal user experience.

### Proposed Solutions:
We propose exploring two potential solutions to address these challenges:

#### Server-Side Rendering with JSDOM:
We suggest investigating the feasibility of enabling SSR for the Lexical Editor by utilizing libraries like JSDOM. This approach would empower us to render the editor within a Node.js/server environment, thus improving initial load times and overall performance.

#### Render-Only Editor Mode:
Another avenue to explore is the creation of a "render-only" mode for the Lexical Editor. This mode would be optimized for server-side rendering, offering a simplified and resource-efficient way to render the lexical tree.

### How This Benefits the Project:
Implementing these solutions would not only enhance the user experience but also make our platform more SEO-friendly. It would address the limitations of client-side rendering, reduce processing time, and eliminate UI jumps, resulting in a smoother and more efficient application.

acywatson:
> Our observation is that the Lexical Editor's rendering is currently limited exclusively to the client-side due to its reliance on client-side rendering mechanisms

Can you be more specific about what you've observed here?

acywatson:
> One of the challenges we've encountered is the time it takes for the editor to process data on the client side.

Can you give an example here? What exactly are you trying to do that leads to the undesirable behavior?

agarwalamn:
Hi @acywatson, Adding details below

#### Project Goal:

The primary objective is to create a community platform similar to Reddit, where users can submit new posts and engage in discussions through comments.
We utilize a lexical editor for both creating and rendering posts.

#### Implementation:

- Our tech stack includes `Next.js`.
[...]

PR #7833, opened 2025-09-19
[lexical-headless] Feature: @lexical/headless/dom module for easier SSR

### LINKED ISSUE #3616: Can't be able to use `generateHtmlFromNodes` correctly.

etrepum:
## Description

This adds a new entrypoint to `@lexical/headless` at `@lexical/headless/dom` that uses conditional exports so that importing it from node brings in a happy-dom implementation of `withDOM` and importing it from the browser brings in a trivial `withDOM = (f) => f(window)` implementation. This is using the same sort of trick that react-dom uses to have separate implementations for react-server and everything else.

RE #4960 - Not closing this one because it could use examples and maybe some framework support. Could even implement hydration either with JSON (like react with some json blob on the page) or HTML (if the user is careful to have an exportDOM/importDOM that round-trips). A react-server conditional export somewhere in @lexical/react would possibly make sense for that use case.

Closes #3616

## Test plan

Adds unit tests. Manually tested that adding withDOM to the playground brings in the trivial browser version.

Follow-up work would be to add examples and more usage, but those are a lot easier to write once there is a version published to npm (at least a nightly).

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -116,7 +116,8 @@ const thirdPartyExternals = [
   'react-dom',
   'yjs',
   'y-websocket',
-  ...(isWWW ? [] : ['react-error-boundary', '@floating-ui/react', 'happy-dom']),
+  'happy-dom',
+  ...(isWWW ? [] : ['react-error-boundary', '@floating-ui/react']),
 ];
 const thirdPartyExternalsRegExp = new RegExp(
   `^(${thirdPartyExternals.join('|')})(\\/|$)`,


etrepum:
I will wait to merge this until after #7706 to avoid approval and merge conflict churn on that larger PR

etrepum:
@ivailop7 this needs a re-approval after fixing the merge conflict

ivailop7:
> @ivailop7 this needs a re-approval after fixing the merge conflict

Done!

PR #7835, opened 2025-09-19
[lexical-clipboard] Bug Fix: Update Lexical Clipboard with Empty Selection

jkjk822:
## Description
`$copyToClipboardEvent` updates a `ClipboardEvent` based on the current selection - however returning null for an absent or empty selection will prevent the Lexical content from getting cleared for an empty selection. This means the 'text/plain' and 'text/html' contents will update to empty strings, but the 'application/x-lexical-editor' will still have whatever stale value was present previously.

An alternative fix to the one committed already is to delete the content when null, rather than skipping, i.e.
```
const v = $editorFn(editor, selection);
if (v !== null) {
  clipboardData[mimeType] = v;
} else {
  delete clipboardData[mimeType]
}
```
This would preserve the `null` return for `$getLexicalContent`, though it doesn't look like anything in the repo depends on it.

## Test plan

No affected tests - though I could probably add one if desired.

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -588,7 +588,11 @@ export function setLexicalClipboardDataTransfer(
   clipboardData: DataTransfer,
   data: LexicalClipboardData,
 ) {
-  clipboardData.clearData();
+  for (const [k] of clipboardDataFunctions) {
+    if (data[k] === undefined) {
+      clipboardData.setData(k, '');
+    }
+  }
   for (const k in data) {
     const v = data[k as keyof LexicalClipboardData];
     if (v !== undefined) {
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -588,6 +588,7 @@ export function setLexicalClipboardDataTransfer(
   clipboardData: DataTransfer,
   data: LexicalClipboardData,
 ) {
+  clipboardData.clearData();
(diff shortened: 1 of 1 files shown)


etrepum (commented):
I don't think changing the semantics of `$getLexicalContent` is the correct fix. This is part of the public API and it's not clear why it should change.

The fix should probably be closer to the problem, probably in `setLexicalClipboardDataTransfer`. Maybe just a call to `clipboardData.clearData()` before the updates.

jkjk822:
Ah, good idea. That is much simpler and handles other stale data cases outside Lexical as well. I confirmed it still fixes the issue in my repo as well.

If this looks good I can try to write a quick test.

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -588,6 +588,7 @@ export function setLexicalClipboardDataTransfer(
   clipboardData: DataTransfer,
   data: LexicalClipboardData,
 ) {
+  clipboardData.clearData();
```
etrepum: It looks like that method is unfortunately not available in that context, I guess the next best thing is set all of the defaults.

```suggestion
  for (const [k] of clipboardDataFunctions) {
    if (data[k] === undefined) {
      clipboardData.setData(k, '');
    }
  }
```

jkjk822: Looks good to me!

jkjk822:
Looks good, thanks for cleaning up and merging.

PR #7836, opened 2025-09-19
[lexical] Bug Fix: Fix Bugs with `RangeSelection#extract`

jkjk822:
## Description
Fix two bugs:

- Using stale `lastIndex` - this is no longer accurate if `selectedNodes.shift()` happens above.
- `selectedNodes` is a view into `_cachedNodes` underneath (so changes are reflected) - except during `splitText`, `_cachedNodes` is reset to `null`, so later changes will not be reflected.

See commits for deeper explanation of each bug.

## Test plan

No affected tests - though I could probably add one if desired.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1534,7 +1534,7 @@ export class RangeSelection implements BaseSelection {
         selectedNodes.pop();
       } else if (endOffset !== lastNodeTextLength) {
         [lastNode] = lastNode.splitText(endOffset);
-        selectedNodes[lastIndex] = lastNode;
+        selectedNodes[selectedNodes.length - 1] = lastNode;
       }
     }
     return selectedNodes;
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -1523,6 +1523,7 @@ export class RangeSelection implements BaseSelection {
         selectedNodes.shift();
       } else if (startOffset !== 0) {
         [, firstNode] = firstNode.splitText(startOffset);
+        this._cachedNodes = selectedNodes;
         selectedNodes[0] = firstNode;
       }
     }
@@ -1534,6 +1535,7 @@ export class RangeSelection implements BaseSelection {
         selectedNodes.pop();
       } else if (endOffset !== lastNodeTextLength) {
         [lastNode] = lastNode.splitText(endOffset);
+        this._cachedNodes = selectedNodes;
         selectedNodes[selectedNodes.length - 1] = lastNode;
       }
     }


etrepum:
Please do add tests if you're fixing bugs. Without a test there's nothing stopping a future regression.

jkjk822:
Ah, will do.

etrepum:
For what it's worth I think this probably has other bugs related to splitting and re-using the cached selection. I think a more reliable strategy would be to do any slicing of text nodes first, then get the extracted nodes.

The caret APIs are better at dealing with this sort of thing, but nobody has refactored extract to use them.

jkjk822:
Makes sense. Regardless this fix is useful for avoiding a custom patch when I next update my repo from upstream.

Adding a test should also ensure a future refactor doesn't regress and break me again.

etrepum:
I think manually updating cachedNodes is not a good fix, setting it to null when the selection changes is what's supposed to happen, it will just be recomputed the next time getNodes is called. It's never really supposed to be updated in-place, the logic to maintain that sort of cache is very error-prone and hard to maintain. If this repo had good readonly hygiene the return type of getNodes would be a readonly array so you wouldn't be able to mutate it without copying first since it is a shared cache.

It would make sense for `extract()` to either do the splitting first so it can return getNodes directly (since it will have the split nodes) or it should make a copy of the array at the start or at least before it starts mutating it.

The `lastIndex` thing seems like a good fix if the bug is observing lastNode twice at the end of the array.

etrepum:
It's not expected behavior for `extract()` to guarantee that `getNodes()` gets the same output, they use subtly different algorithms. The only expected side-effect is the TextNode splitting, it shouldn't otherwise be moving the anchor and focus points around, so only the changes directly caused by splitting of TextNodes should be observable from a subsequent call to `getNodes()`.

`_cachedNodes` will be cleared by splitText when the selection is the editor's selection. The cache is completely incoherent if you are mutating a selection that isn't the editor's current selection, because the mutators only concern themselves with updating the current selection not every selection object that might exist.

Both of these methods are kinda bad, which is a big reason why the caret APIs exist. Any fixes to them should be really precise, and I don't think this strategy (trying to force things into the cache that won't be computed the same way when the cache is cleared) is the right one.

jkjk822:
> The only expected side-effect is the TextNode splitting, it shouldn't otherwise be moving the anchor and focus points around, so only the changes directly caused by splitting of TextNodes should be observable from a subsequent call to getNodes().

Unfortunately sometimes the changes aren't observable regardless, which is part of the issue I was hoping to solve. For example, if I remove the `_cachedNodes` changes and run my tests, you can see the selection will be stale (_ denotes selection):

```
pa_ragrap_h text
# selection.extract() -> new TextNode("ragrap")
# selection.getNodes() -> stale TextNode("paragraph text") that is no longer in document
```

> It would make sense for `extract()` to either do the splitting first so it can return getNodes directly (since it will have the split nodes) or it should make a copy of the array at the start or at least before it starts mutating it.

So I'm not sure your suggestion would work since `getNodes()`'s value is stale. Making `selectedNodes` a copy of `_cachedNodes` rather than a reference does seem like the other reasonable alternative to updating `_cachedNodes`. I can make that update, though it doesn't solve the staleness. I guess I could update the anchor/focus here, or make `splitText` actually update it on the selection. Curious to hear what you'd recommend.

-------

Overall, I'm just hoping to get the nicest behavior with minimal changes. Right now my repo will fail with a stale `getNodes()` since that is the method used in `LexicalNode#isSelected()`.

etrepum:
I don't think you quite follow what I'm saying here. The `_cachedNodes` *should* be cleared. It isn't when the selection isn't the editor's current selection, because splitText doesn't know which selection it was called from. That should be fixed, with a single `this.setCachedNodes(null)` in this extract method (this can even go immediately after the getNodes call). The wrong solution is to mutate the result of getNodes and/or try and manually set `_cachedNodes` to anything but `null` except when computed by `getNodes`.

etrepum:
It would also make sense to make a copy of the result of `getNodes()` since it's mutated in-place and that would change previously observed values. Unlikely to happen, but might as well avoid another potential bug.

etrepum (approved):

jkjk822:
Nice, happy to see the tests passing with this style rather than modifying `_cachedNodes`. Thanks for finishing up and merging.

PR #7844, opened 2025-09-22
[lexical-react] Bug Fix: Add visibility margin to `isTriggerVisibleInNearestScrollContainer` to prevent popover from wrongfully closing

### LINKED ISSUE #7837: Bug: LexicalTypeaheadMenuPlugin closes when trying to scroll inside the dropdown list
In the LexicalTypeaheadMenuPlugin dropdown, when I'm rendering a list of about 50 users, and the dropdown overflows with a scrollbar, trying to go down the user list by using the arrow keys or scrolling with the mouse closes the dropdown. 

When using the arrow keys, the items in view are iterated, but closes the moment I'm about to scroll. Same thing happens when I try to scroll with my mouse by hovering over it and using the scroll wheel, or by clicking and dragging the scrollbar (the dropdown closes during the drag, not the click). What's more interesting is that the onClose event does not fire when it closes like this.

Lexical version: 0.35.0

## Steps To Reproduce

1. Click on the editor and type in @a, it should prop a dropdown which is overflowing with a scrollbar 
2. Try scrolling the dropdown with your mouse, or arrow keys

Link to code example: https://codesandbox.io/p/sandbox/qrz5n5

## The current behavior
Dropdown closes instead of scrolling

## The expected behavior
Dropdown should scroll and not close

## Impact of fix
The bug is quite severe; it is blocking me from shipping a core feature.

dwrth:
## Description

Whenever the editors total height is so small that the LexicalMenu's popover is outside the bounds of it's parent element, the `isTriggerVisibleInNearestScrollContainer` results in `false` whenever the `useEffect` inside `useDynamicPositioning` is run. This happens on scrolling and also navigating all the way to the bottom via the arrow keys.

This PR proposes the addition of a margin to the visibility check in order for the popover trigger to still be considered "visible" even if it's slightly out of bounds.

Closes #7837

### Before

https://github.com/user-attachments/assets/98c5efe0-783c-44e7-886a-2362ea27e9bd

### After

https://github.com/user-attachments/assets/48f8ffff-ad33-44e2-ae4f-23e3be36b911

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -194,7 +194,12 @@ function isTriggerVisibleInNearestScrollContainer(
 ): boolean {
   const tRect = targetElement.getBoundingClientRect();
   const cRect = containerElement.getBoundingClientRect();
-  return tRect.top > cRect.top && tRect.top < cRect.bottom;
+
+  const VISIBILITY_MARGIN_PX = 6;
+  return (
+    tRect.top >= cRect.top - VISIBILITY_MARGIN_PX &&
+    tRect.top <= cRect.bottom + VISIBILITY_MARGIN_PX
+  );
 }
 
 // Reposition the menu on scroll, window resize, and element resize.


etrepum (approved):
I like how simple this workaround is!

PR #7845, opened 2025-09-22
[lexical-markdown][lexical-playground] Bug Fix: Keep indent when switching with markdown

### LINKED ISSUE #7820: Bug: Indent lost when switching to Markdown
Lexical version: 0.35.0 (latest version)

## Steps To Reproduce

1. Go to Lexical playground https://playground.lexical.dev
2. Type something in the editor
3. Click on indent button
4. Click on Markdown button to switch into Markdown code block (which I call Markdown view)
5. Notice indent is missing
6. Click again on Markdown button
7. Indent has been lost

## The current behavior
Indent is lost when switching from default view to Markdown.

## The expected behavior
Switching to Markdown view should not alter text. Indent should be kept when switching between Markdown and default view.

## Impact of fix
I would say it's an important issue because there's data loss.

## Observation
This issue doesn't happen when adding an indent while in Markdown view, the reason is a tab node is created and this one is kept when switching view mode.
By looking at the code I see that's because `$handleMultilineIndent` from `CodeHighlightPrism.ts` is called.
I think we should be able to fix this issue by reproducing the same behavior as this function or at least creating a tab node when calling `INDENT_CONTENT_COMMAND`.
I can do a PR but I didn't found how commands are working yet, if anyone can explain I'll be glad to do it.

lytion:
## Description
Reported issue: #7820
To summarize, if you use indent button and then switch to markdown, indent will be lost.

I decided to fix this issue by adding a transformer by I think it could also be fixed by handling it directly in `createMarkdownExport` and `createMarkdownImport`

This transformer simply add '\t' when exporting to markdown and remove those '\t' and set indent when importing from markdown. This only happen if '\t' is found at the beginning of the string and if the parent is a paragraph to avoid applying it to lists.

Closes #7820

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -35,6 +35,8 @@ import {
 import {
   $createLineBreakNode,
   $createTextNode,
+  $isParagraphNode,
+  $isTextNode,
   ElementNode,
   Klass,
   LexicalNode,
@@ -209,6 +211,7 @@ const CODE_SINGLE_LINE_REGEX =
   /^[ \t]*```[^`]+(?:(?:`{1,2}|`{4,})[^`]+)*```(?:[^`]|$)/;
 const TABLE_ROW_REG_EXP = /^(?:\|)(.+)(?:\|)\s?$/;
 const TABLE_ROW_DIVIDER_REG_EXP = /^(\| ?:?-*:? ?)+\|\s?$/;
+const INDENT_REGEX = /^\t+/;
 
 const createBlockNode = (
   createNode: (match: Array<string>) => ElementNode,
@@ -592,6 +595,37 @@ export const LINK: TextMatchTransformer = {
     return linkTextNode;
   },
   trigger: ')',
+  type: 'text-match',
+};
+
+export const INDENT: TextMatchTransformer = {
+  dependencies: [],
+  export: (node, exportChildren, exportFormat) => {
+    const parentNode = node.getParent();
+    const textContent = node.getTextContent();
+    if (
+      !$isParagraphNode(parentNode) ||
+      !$isTextNode(node) ||
+      parentNode.getFirstChild() !== node
+    ) {
+      return null;
+    }
+    const indent = parentNode.getIndent();
+    const textWithFormat = exportFormat(node, textContent);
+    return textWithFormat ? '\t'.repeat(indent) + textWithFormat : null;
+  },
+  importRegExp: INDENT_REGEX,
+  regExp: INDENT_REGEX,
+  replace: (textNode, match) => {
+    const [indents] = match;
+    const parentNode = textNode.getParent();
+    if (!parentNode || !$isParagraphNode(parentNode)) {
+      return;
+    }
+    parentNode.setIndent(indents.length);
+    textNode.setTextContent(textNode.getTextContent().replace(/^\t+/, ''));
+    return textNode;
+  },
   type: 'text-match',
 };
 
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -669,6 +669,22 @@ describe('Markdown', () => {
       html: '<p><span style="white-space: pre-wrap;">[h</span><a href="https://lexical.dev"><span style="white-space: pre-wrap;">ello</span></a><a href="https://lexical.dev"><span style="white-space: pre-wrap;">world</span></a></p>',
       md: '[h[ello](https://lexical.dev)[world](https://lexical.dev)',
     },
+    {
+      html: '<p style="padding-inline-start: 40px;"><span style="white-space: pre-wrap;">Hello Word</span></p>',
+      md: '\tHello Word',
+    },
+    {
+      html: '<p style="padding-inline-start: 80px;"><span style="white-space: pre-wrap;">Hello Word</span></p>',
+      md: '\t\tHello Word',
+    },
+    {
+      html: '<p><span style="white-space: pre-wrap;">Hello\t Word</span></p>',
+      md: 'Hello\t Word',
+    },
+    {
+      html: '<p style="padding-inline-start: 40px;"><i><em style="white-space: pre-wrap;">Hello</em></i><span style="white-space: pre-wrap;"> Word</span></p>',
+      md: '\t*Hello* Word',
+    },
   ];
 
   const HIGHLIGHT_TEXT_MATCH_IMPORT: TextMatchTransformer = {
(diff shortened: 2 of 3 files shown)


etrepum:
Failing lint https://github.com/facebook/lexical/actions/runs/17919434734/job/50950181819?pr=7845

```
/home/runner/work/lexical/lexical/packages/lexical-markdown/src/index.ts
Error:   31:3  error  'INDENT' is defined but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars
```

etrepum:
Now the prettier check is failing https://github.com/facebook/lexical/actions/runs/17919849090/job/50951631139

You can run the integrity checks locally with `npm run ci-check`, and if the husky git commit hooks are installed then it will catch these formatting issues locally as well. You can fix the prettier issue with `npm run prettier:fix`

fantactuka:
Given it's not a standard spec transformation, it might be better to keep it as a separate exported transformer instead of including it into default transformers set. And then add it explicitly into [playground's transformers list](https://github.com/facebook/lexical/blob/63958a2f715aea80287ac1e94f8264f44a824674/packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts#L310)

lytion:
Is there some way to run tests in a linux environment or should I test it in a vm ?
Edit: Looks like updating the branch fix the e2e test

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -622,6 +625,37 @@ export const LINK: TextMatchTransformer = {
...
 
+export const INDENT: TextMatchTransformer = {
+  dependencies: [],
+  export: (node, exportChildren, exportFormat) => {
+    const parentNode = node.getParent();
+    const textContent = node.getTextContent();
+    if (
+      !$isParagraphNode(parentNode) ||
+      !$isTextNode(node) ||
+      parentNode.getFirstChild() !== node
```
etrepum: It's generally better to use is or key comparison to avoid some potential issues if the instance isn't the latest one from the EditorState (very unlikely in this case)

```suggestion
      !node.is(parentNode.getFirstChild())
```

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -622,6 +625,37 @@ export const LINK: TextMatchTransformer = {
...
+  importRegExp: INDENT_REGEX,
+  regExp: INDENT_REGEX,
+  replace: (textNode, match) => {
+    const [indents] = match;
+    const parentNode = textNode.getParent();
+    if (!parentNode || !$isParagraphNode(parentNode)) {
+      return;
+    }
+    parentNode.setIndent(indents.length);
+    textNode.setTextContent(textNode.getTextContent().replace(/^\t+/, ''));
```
etrepum, line 738: This seems like maybe there's a bug elsewhere because tabs are usually supposed to use TabNode

etrepum: I noticed that this PR was updated with main but this comment hasn't been addressed

lytion: Sorry I missed the comment.
I think you're right, when clicking on "left indent" button in normal mode, there's no `INDENT_CONTENT_COMMAND` which I believe should happen and therefore should add a TabNode.

lytion: I want to make sure we're on the same page.
From what I see:
Markdown use \t for indent and tab key.
CodeHighlight convert \t into TabNode when converting to markdown.
Nodes use __indent and TabNode.
Here the transformer set the correct number of indent and remove all \t found from the start of the line.
If I understand correctly, what you're talking about is to convert all \t into TabNode even if CodeHighlight is disabled.
I can create a TAB transformer that does that, but maybe that's a separate issue

etrepum: Generally speaking there should never be '\t' or a '\n' in a TextNode, anything that creates a TextNode with those characters in the contents probably has a bug. The solution isn't a transformer, it's to fix the code that's creating those TextNodes. I don't think there is currently a helper to do this other than RangeSelection.insertRawText but one could be added to make this a little easier.

lytion: This is a quick and dirty refactor but is this the direction you are thinking about ?
```
diff --git a/packages/lexical-markdown/src/MarkdownImport.ts b/packages/lexical-markdown/src/MarkdownImport.ts
index d64037d91..851707205 100644
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts

+// Copy of LexicalSelection insertRawText function
+function insertRawText(text: string): LexicalNode[] {
+  const parts = text.split(/(\r?\n|\t)/);
+  const nodes = [];
+  const length = parts.length;
+  for (let i = 0; i < length; i++) {
+  const part = parts[i];
+  if (part === '\n' || part === '\r\n') {
+    nodes.push($createLineBreakNode());
+  } else if (part === '\t') {
+    nodes.push($createTabNode());
+  } else {
+    nodes.push($createTextNode(part));
+  }
+}
+  return nodes;
+}
+
 function $importBlocks(
   lineText: string,
   rootNode: ElementNode,
@@ -227,27 +248,39 @@ function $importBlocks(
   textMatchTransformers: Array<TextMatchTransformer>,
   shouldPreserveNewLines: boolean,
 ) {
-  const textNode = $createTextNode(lineText);
+  const nodes = insertRawText(lineText);
   const elementNode = $createParagraphNode();
-  elementNode.append(textNode);
+  elementNode.append(...nodes);
   rootNode.append(elementNode);
 
   for (const {regExp, replace} of elementTransformers) {
-    const match = lineText.match(regExp);
-
-    if (match) {
-      textNode.setTextContent(lineText.slice(match[0].length));
-      if (replace(elementNode, [textNode], match, true) !== false) {
+    for (const node of nodes) {
+      if (!$isTextNode(node)) {
         break;
       }
+      const match = node.getTextContent().match(regExp);
+
+      if (match) {
+        node.setTextContent(lineText.slice(match[0].length));
+        if (replace(elementNode, [node], match, true) !== false) {
+          break;
+        }
+      }
     }
   }
 
-  importTextTransformers(
-    textNode,
-    textFormatTransformersIndex,
-    textMatchTransformers,
-  );
+  for (const node of nodes) {
+    if (!$isTextNode(node)) {
+      break;
+    }
+    // TODO: Do this for TabNode too (maybe Line
+    importTextTransformers(
+      node,
+      textFormatTransformersIndex,
+      textMatchTransformers,
+    );
+  }
+
```

etrepum: Hard to say without a thorough audit of that code but probably something like that. It might be the case that it’s better to fix it up afterwards in the markdown case depending on how that code works.

lytion: I made a new PR to address this: #8211

lytion:
https://github.com/facebook/lexical/pull/8211 has been merged. Closing this PR.

PR #7854, opened 2025-09-24
[lexical] Bug Fix: account for Apple WebView in `setManagedLineBreak`

elpnt:
## Description

Currently, `setManagedLineBreak` only accounts for `IS_IOS` and `IS_SAFARI` when applying the WebKit-specific hack for line breaks.  
However, in Apple’s WebView environments (such as Tauri with wry on macOS), `IS_APPLE_WEBKIT` is `true` but not considered, which can lead to inconsistent line break rendering and cursor behavior.

This PR updates the logic so that the WebKit-specific hack also applies when `IS_APPLE_WEBKIT` is true. This ensures consistent behavior across iOS Safari, standalone Safari, and embedded Apple WebView environments.

### Before

https://github.com/user-attachments/assets/84ebddcb-8217-40b3-8e93-e5f8a7d81a9f

- On Apple WebView (Tauri on macOS), line breaks inserted by decorator nodes were not managed correctly.  
- Cursor movement and rendering could become inconsistent.

### After

https://github.com/user-attachments/assets/b455c6a9-9ec2-4a60-981c-02f6415d129c

- In Apple WebView, the `decorator` line break type is now handled the same way as in Safari.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -31,6 +31,7 @@ module.name_mapper='^@lexical/extension$' -> '<PROJECT_ROOT>/packages/lexical-ex
 module.name_mapper='^@lexical/file$' -> '<PROJECT_ROOT>/packages/lexical-file/flow/LexicalFile.js.flow'
 module.name_mapper='^@lexical/hashtag$' -> '<PROJECT_ROOT>/packages/lexical-hashtag/flow/LexicalHashtag.js.flow'
 module.name_mapper='^@lexical/headless$' -> '<PROJECT_ROOT>/packages/lexical-headless/flow/LexicalHeadless.js.flow'
+module.name_mapper='^@lexical/headless/dom$' -> '<PROJECT_ROOT>/packages/lexical-headless/flow/dom.js.flow'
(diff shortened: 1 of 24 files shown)


ivailop7 (dismissed):

etrepum (approved):

etrepum:
I think this is probably good, but there is some chance that our e2e tests on mac+chromium are not catching some regression where we expect Chrome to also be IS_APPLE_WEBKIT on macOS. We might want to note that as a breaking change. Here's an example Chrome userAgent:
`'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/140.0.0.0 Safari/537.36'`

elpnt:
Are there any cases where Chrome on macOS is expected to be `IS_APPLE_WEBKIT`?
I think this isn’t a breaking change, since the previous `IS_APPLE_WEBKIT` already accounted for `!IS_CHROME`.

etrepum (approved):
It looks like you're right, I just missed that at first glance because it's on another line. I think this should be good to go then

PR #7875, opened 2025-09-25
Preventing empty string selection

### LINKED ISSUE #7862: Bug:
Lexical version: v0.33.1 (but should also exist in latest 0.35)

## Steps To Reproduce

1. Put cursor into a text field but don't select anything
2. Trigger a copy (i.e. cmd+c on mac)

## The current behavior

An empty string is copied into the clipboard

## The expected behavior

Clipboard is not updated. Should not be able to copy "nothing".

Offending code is likely here: https://github.com/facebook/lexical/blob/a278e96ff7d271ba7a3d85c9e6483121205d00c5/packages/lexical-clipboard/src/clipboard.ts#L535

It only checks if selection is null, but it should also check if the RangeSelection is "empty"

Image

niikkhilsharma:
## Description

- The current behavior allows copying an empty string to the clipboard when the cursor is placed inside the editor when no text is selected.
- This PR modifies the clipboard copy logic to also check if the selection is collapsed, preventing the clipboard from being overwritten with an empty string.
- The condition now ensures that if the selection is null or collapsed, the copy action is skipped.

Closes #7862

### Before
- Copying with no selection would copy an empty string to the clipboard.

https://github.com/user-attachments/assets/90577645-9f2f-40fc-94d0-3d71f44199f5

### After
- Copying with no selection does not update the clipboard.
- Verified manually to ensure clipboard remains unchanged when their is no text to copy.

https://github.com/user-attachments/assets/727568e7-8a04-4e39-8cc1-dff9b5ec1062


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -519,7 +519,7 @@ function $copyToClipboardEvent(
 ): boolean {
   if (data === undefined) {
     const domSelection = getDOMSelection(editor._window);
-    if (!domSelection) {
+    if (!domSelection || domSelection.isCollapsed) {
```
etrepum: This is always true, you forgot to call the method. I'm surprised that eslint isn't complaining about this.

```suggestion
    if (!domSelection || domSelection.isCollapsed()) {
```

etrepum (commented):
It would be best to also add tests to show that this change fixes what you expect it to fix, otherwise there's nothing to stop it from changing behavior again in the future.

packages/lexical-clipboard/src/clipboard.ts
niikkhilsharma: I don't have knowledge about writing tests. But surely I'll learn and write those. It would be really helpful if you can guide me a little here.

etrepum: https://lexical.dev/docs/api/#running-tests (note that the e2e tests assume you are running the playground in another tab)

I'd suggest taking a look at the other clipboard related tests to see how they work. Tests for this would probably belong in this file: https://github.com/facebook/lexical/blob/main/packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs

niikkhilsharma: This is not a function it's a boolean value on domSelection
`const domSelection = getDOMSelection(editor._window);`

image

etrepum: Ok well the selection used for determining what is copied should be based on the lexical selection, which might not be exactly represented by the DOM selection, so that's probably why it doesn't work.

niikkhilsharma: if you put something like
`if (!domSelection || domSelection.isCollapsed) {
      return false;
    }`
    you're able to check weather the node is empty or not and so, I think it works fine as in the video also above it is working fine.

etrepum: It might work in the video but it doesn't work in the general case because there are failing tests. Some types of selection don't use the DOM selection such as with tables, but copy should still work.

PR #7878, opened 2025-09-26
[lexical]Refactor: npm audit fix

lytion:
## Description
I saw there was some vulnerabilities so I run some `npm audit fix`.
unit tests seem to pass so I'm pushing this to see if the CI find something.
There's still two issues:
### 2 high (same package: cross-spawn):
[child-process-promise](https://www.npmjs.com/package/child-process-promise/v/2.2.0): it's using cross-spawn@4.0.2 which as [ReDoS vulnerability](https://github.com/advisories/GHSA-3xgq-45jj-v275)
This package seem to be mostly use when publishing a new release and from what I read ReDoS seem to be a problem if it uses regex so I think it should be fine.

[npm-run-all](https://www.npmjs.com/package/npm-run-all/v/1.7.0): Same as above, shouldn't be a big issue as this run npm commands.

### 1 critical:
eslint-plugin-lexical: It is flagged as malware infected which is weird as `package.json` implement it from a file: `"eslint-plugin-lexical": "file:./eslint-plugin"`
I don't know how we could fix that one.

### Before

`npm audit` returns 48 vulnerabilities (4 low, 38 moderate, 4 high, 2 critical)

### After

`npm audit` returns 18 vulnerabilities (15 moderate, 2 high, 1 critical)


lytion:
Hmm, don't know what happen in vercel, maybe I can revert the package causing the issue

etrepum:
None of those highlighted vulnerabilities are problematic as used in this repo. The critical one is a false positive, obviously a local file in the repo isn't a problem. I think perhaps you did this change before updating to main instead of after which might be causing the build issue.

lytion:
>I think perhaps you did this change before updating to main instead of after which might be causing the build issue.

I don't think so, a deployment failed after I pushed merge conflict [6ecd029](https://github.com/facebook/lexical/pull/7878/commits/6ecd0292f768b0eeb9cdc992410da5f99d761293)

etrepum:
I don't have access to vercel to see the build error but you can probably reproduce it with something like `npm ci && npm run -w packages/lexical-website build`. If the build fails we can't accept a PR, so unless it can be fixed the only thing that can be done is to close this.

etrepum:
I'm going to close this because it breaks the build, there isn't really anything that can be done with this PR since it's just opaque changes to package-lock.json. If you manage to figure out how to do this without breaking the website build feel free to re-open or create a new PR. None of the audit issues are problematic for users of lexical as described, because cross-spawn is only used during the build process with static arguments that don't trigger a ReDoS.

The eslint-plugin-lexical issue is certainly a false positive because it's not even coming from npm and doesn't have any dependencies. It was also recently moved and renamed under the `@lexical/` scope to `@lexical/eslint-plugin-internal` so probably doesn't come up in the audit report anymore and wouldn't be susceptible to an untrusted party squatting on the name if you were to somehow fetch it from npm instead of the local directory.

PR #7880, opened 2025-09-27
[lexical-clipboard] Preventing copying empty string

### LINKED ISSUE #7862: Bug:

niikkhilsharma:
## Description

- The current behavior allows copying an empty string to the clipboard when the cursor is placed inside the editor when no text is selected.
- This PR modifies the clipboard copy logic to also check if the selection is collapsed, preventing the clipboard from being overwritten with an empty string.
- The condition now ensures that if the selection is null or collapsed, the copy action is skipped.

Closes #7862

### Before
- Copying with no selection would copy an empty string to the clipboard.

https://github.com/user-attachments/assets/90577645-9f2f-40fc-94d0-3d71f44199f5

### After
- Copying with no selection does not update the clipboard.
- Verified manually to ensure clipboard remains unchanged when their is no text to copy.

https://github.com/user-attachments/assets/727568e7-8a04-4e39-8cc1-dff9b5ec1062


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -519,6 +519,12 @@ function $copyToClipboardEvent(
 ): boolean {
   if (data === undefined) {
     const domSelection = getDOMSelection(editor._window);
+    const editorSelection = $getSelection();
+
+    if (editorSelection && editorSelection.isCollapsed()) {
+      return false;
+    }
+
```
etrepum: If you look further down there's already a call to $getSelection, this condition should be added there instead

niikkhilsharma: Sure

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -537,7 +537,7 @@ function $copyToClipboardEvent(
     ) {
       return false;
     }
-    const selection = $getSelection();
+
     if (selection === null) {
       return false;
     }
```
etrepum: With a small change to the above statement this becomes redundant.

```suggestion
```

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -519,9 +519,9 @@ function $copyToClipboardEvent(
 ): boolean {
   if (data === undefined) {
     const domSelection = getDOMSelection(editor._window);
-    const editorSelection = $getSelection();
+    const selection = $getSelection();
 
-    if (editorSelection && editorSelection.isCollapsed()) {
+    if (selection && selection.isCollapsed()) {
```
etrepum:
```suggestion
    if (!selection || selection.isCollapsed()) {
```

etrepum (dismissed):
This looks good assuming the existing tests pass, but the one thing it's missing is a test to show that it works since no existing test expects this new behavior.

niikkhilsharma:
Yes, i'm working on writing tests for this new behaviour

niikkhilsharma:
@etrepum 
`
  test('Copy and paste empty string', async ({isRichText, page}) => {
    await focusEditor(page);

    await assertHTML(
      page,
      html`
        <p class="PlaygroundEditorTheme__paragraph" dir="auto"><br /></p>
      `,
    );

    // Set clipboard to some initial content to verify it doesn't get overwritten
    await withExclusiveClipboardAccess(async () => {
      await page.evaluate(() =>
        navigator.clipboard.writeText('initial content'),
      );
    });

    await withExclusiveClipboardAccess(async () => {
      await copyToClipboard(page);
    });

    await pasteFromClipboard(page, {'text/plain': 'test content'});

    // Check editor content after paste, making sure it only contains the pasted content
    if (isRichText) {
      await assertHTML(
        page,
        html`
          <p class="PlaygroundEditorTheme__paragraph" dir="auto">
            <span data-lexical-text="true">test content</span>
          </p>
        `,
      );
    } else {
      await assertHTML(
        page,
        html`
          <p class="PlaygroundEditorTheme__paragraph" dir="auto">
            <span data-lexical-text="true">test content</span>
          </p>
        `,
      );
    }
  });`
  
  i was writing this, but this required write permission, should I add the write permission here in the respective file.

etrepum:
You can select something on the page and copy it which wouldn't require write permission. This works in more browsers so it's a better test.

niikkhilsharma:
We want to implement no-copying behavior only when selecting an empty string. 
Selecting something won't work. I need a way to test that my clipboard content has not been over-written after performing copy on an empty editor

etrepum:
Yes. Select something, copy it. Now the clipboard has something. Then collapse the selection and copy again. Confirm that the clipboard still has the original content.

etrepum:
Another sequence would be:
* Select something
* Cut (which copies it to the clipboard, removes it, and collapses the selection)
* Confirm that the content was removed
* Copy (the selection is collapsed so this should do nothing)
* Paste
* Confirm that the content has been added back

niikkhilsharma:
Sure, Thank you

niikkhilsharma:
@etrepum brother can you please guide me where is the issue here.

niikkhilsharma:
@etrepum can you please see why these tests are failing

etrepum:
The tests are failing in webkit and firefox because you wrote them in a way that only works in chromium. It's failing at `await context.grantPermissions(['clipboard-read', 'clipboard-write']);` you can look at the test failures, by clicking on the failed check in the github PR UI e.g. https://github.com/facebook/lexical/actions/runs/18090320132/job/51469283931?pr=7880

Can be reproduced locally as well, here's how you would run that test directly (with `npm run start` in another tab):

```
npm run test-e2e-firefox packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs:967
```

Current output:
```
   Error: browserContext.grantPermissions: Unknown permission: clipboard-read

      970 |     context,
      971 |   }) => {
    > 972 |     await context.grantPermissions(['clipboard-read', 'clipboard-write']);
          |                   ^
      973 |     await focusEditor(page);
      974 |
      975 |     await page.keyboard.type('Hello world');
        at /Users/bob/src/lexical/packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs:972:19

    attachment #1: video (video/webm) ──────────────────────────────────────────────────────────────
    test-results/e2e-CopyAndPaste-lexical-C-aa88b-lection-preserves-clipboard-firefox-retry1/video.webm
    ────────────────────────────────────────────────────────────────────────────────────────────────

  1 failed
    [firefox] › packages/lexical-playground/__tests__/e2e/CopyAndPaste/lexical/CopyAndPaste.spec.mjs:967:3 › CopyAndPaste › Cut then copy empty selection preserves clipboard 
```

I've just pushed a commit that fixes the test to work correctly without using any chrome specific code

niikkhilsharma:
Yes, it's working now
Thank you

etrepum (approved):

PR #7892, opened 2025-09-30
[lexical-list][lexical-markdown] Feature: Keep list marker for markdown round-trip

lytion:
## Description
I'm submitting a proposal so list markers are not loss anymore when switching from/to markdown.
This is a continuation in the idea that switching from/to markdown should not modify typed text.

These changes add a state to NodeList that save list marker between markdown import/export.

## Test plan
In lexical playground
In view mode:
1. Type `* hello`
2. Switch to markdown

### Before
-> markdown is `- hello`

### After
-> markdown is `* hello`, it kept marker
Switching again and again keep marker.

## Unit tests
Added unit tests in `LexicalMarkdown.test.ts` that verify a marker is kept between import/export
Added e2e tests in `Markdown.spec.mjs` that verify markers are kept when typing in lexical view and when copy/pasting.


etrepum:
I don't think this really belongs in ListItemNode, lexical is not really a markdown native editor so it doesn't make sense to always try and store markdown related things even when you are not using it.

One thing you could do is to write custom markdown transformers that use NodeState to attach this metadata to the nodes it creates so that they can be used on markdown serialization (and that data would end up in the JSON for those nodes as well). https://lexical.dev/docs/concepts/node-state

The issues you had writing tests probably had something to do with how the editors are configured in the specific places you were trying to write tests. There is no check list transformer in the markdown package, but there is one in the playground, so you'd have to copy that into the tests to use it. A primary reason for that is that such a transformer would fail if the check list plugin is not installed. Extensions will solve this once someone implements a markdown extension, but that's why it only ships with default transformers that are likely to be enabled in a basic rich text editor.

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -210,6 +213,10 @@ const CODE_SINGLE_LINE_REGEX =
 const TABLE_ROW_REG_EXP = /^(?:\|)(.+)(?:\|)\s?$/;
 const TABLE_ROW_DIVIDER_REG_EXP = /^(\| ?:?-*:? ?)+\|\s?$/;
 
+export const listMarkerState = createState('listMarker', {
```
etrepum: might be worth coming up with a more globally unique name for this, since it does only apply to markdown and it's not allowed for two different states on the same node to have the same name

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -251,6 +258,10 @@ const listReplace = (listType: ListType): ElementTransformer['replace'] => {
     const listItem = $createListItemNode(
       listType === 'check' ? match[3] === 'x' : undefined,
     );
+    if (listType === 'bullet' || listType === 'check') {
+      const marker = match[0].trim()[0];
+      $setState(listItem, listMarkerState, marker);
```
etrepum: This state should really be on the list rather than the list item since it's not really permitted to mix different markers in the same unordered list. It's also a lot more compact as far as serialization goes.

if you put it on the ListItemNode what will happen is that you deserialize from markdown, then edit in lexical, you will end up with your custom markers and then any new items will be the default marker. You'll also run into the same sort of issues when copy and paste because the list items might have been created with different markers than their siblings.

Would be a good idea to have some test coverage that checks that the lists have consistent markers after such operations

packages/lexical-markdown/src/MarkdownTransformers.ts
lytion: What about `mdListMarkerState` ?

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -258,11 +258,14 @@ const listReplace = (listType: ListType): ElementTransformer['replace'] => {
...
-      $setState(listItem, listMarkerState, marker);
-    }
+    const listMarker =
+      listType === 'bullet' || listType === 'check'
+        ? match[0].trim()[0]
+        : undefined;
     if ($isListNode(nextNode) && nextNode.getListType() === listType) {
+      if (listMarker) {
+        $setState(nextNode, listMarkerState, listMarker);
+      }
```
lytion, line 268: I'm not fully satisfied of this bunch of if condition but haven't found a better way to do it.

Except maybe putting `if ($isListNode(nextNode) && nextNode.getListType() === listType)` and the other if/else in variable to handle everything at the end. I didn't do that as it doesn't really simplify the code imo

packages/lexical-markdown/src/MarkdownTransformers.ts
etrepum: mdListMarker seems reasonable, "state" is redundant

```suggestion
export const listMarkerState = createState('mdListMarker', {
```

etrepum (approved):

etrepum (approved):

etrepum:
I think this is good to merge but could you fix up the PR description to match the current state of the PR (e.g. there are tests now)? This will help when putting together the release notes.

lytion:
Done, I refactored the Unit Test part by:
>Added unit tests in LexicalMarkdown.test.ts that verify a marker is kept between import/export
Added e2e tests in Markdown.spec.mjs that verify markers are kept when typing in lexical view and when copy/pasting.

And added a small description of what the PR does

lytion:
It seems I'm missing something to merge it, I don't have the merge button.
I'm sorry I only started using GitHub PR system to work on Lexical, I'm not familiar with it. I can't find what's preventing me from merging

etrepum:
The merge button is only there for maintainers

PR #7898, opened 2025-10-01
[lexical-react] Bug Fix: Prevent out-of-bounds selection in LexicalMenu when options shrink

### LINKED ISSUE #7897: Bug: LexicalMenu reads option.ref when options[newSelectedIndex] is undefined
Lexical version: v0.36.2+dev.esm (Playground)

## Steps To Reproduce

1. Open the Lexical Playground
2. Enable mentions (@) and type the first letter to open the suggestion list.
3. Move the cursor to the last entry (Arrow Down).
4. Keep typing quickly so the list refreshes and shrinks before `selectedIndex` is reset. You can try with 'pooja'. 
6. Quickly press Arrow Down again

Or for a better success rate at reproducing it, have Hammerspoon doing it:
```lua
hs.hotkey.bind({"ctrl", "alt"}, "p", function()
    hs.eventtap.keyStrokes("@")

    hs.eventtap.keyStrokes("p")

    hs.eventtap.keyStroke({}, "down")
    hs.eventtap.keyStroke({}, "down")
    hs.eventtap.keyStroke({}, "down")
    hs.eventtap.keyStroke({}, "down")

    hs.timer.usleep(1000000)

    hs.eventtap.keyStrokes("ooj")

    hs.eventtap.keyStroke({}, "down")
end)

```

## The current behavior
Crash in the browser console
Image

If I'm not mistaken, this is happening here: https://github.com/facebook/lexical/blob/8cbd7260914acbbdbc34c2e35cdda40023a63783/packages/lexical-react/src/shared/LexicalMenu.ts#L385

## The expected behavior
The highlighted index should reset (or short-circuit) when the list shrinks so we never dereference a missing option.

## Impact of fix

Prevents a crash under fast typing in typeahead/mentions flows used by any app relying on LexicalTypeaheadMenuPlugin

robertoalvarezalonso:
## Description

When the typeahead menu’s options list shrinks (e.g., user types quickly and the suggestions refresh) before selectedIndex is reset, the code may try to read option.ref from an out-of-range index, causing a runtime error and breaking keyboard navigation. The affected logic lives in `packages/lexical-react/src/shared/LexicalMenu.ts`

**What this PR changes**

- Clamp selectedIndex whenever options change so it always stays within [0, options.length - 1]. If options is empty, clear the selection.
- Add a defensive guard before accessing a selected option (avoid reading ref from undefined).

Closes #7897 

### Before

https://github.com/user-attachments/assets/997b3279-1fca-4a42-871d-64ca80378747

### After

https://github.com/user-attachments/assets/d8fd9d4d-907b-4920-a444-e7f85e0a3508

### DIFF
--- a/packages/lexical-react/src/shared/LexicalMenu.ts
+++ b/packages/lexical-react/src/shared/LexicalMenu.ts
@@ -377,12 +377,12 @@ export function LexicalMenu<TOption extends MenuOption>({
             const newSelectedIndex =
               selectedIndex === null
                 ? 0
-                : selectedIndex !== options.length - 1
+                : selectedIndex < options.length - 1
                   ? selectedIndex + 1
                   : 0;
             updateSelectedIndex(newSelectedIndex);
             const option = options[newSelectedIndex];
-            if (option.ref != null && option.ref.current) {
+            if (option && option.ref != null && option.ref.current) {
               editor.dispatchCommand(
                 SCROLL_TYPEAHEAD_OPTION_INTO_VIEW_COMMAND,
                 {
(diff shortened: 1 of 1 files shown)


ivailop7 (approved):

etrepum:
Thank you! The PR just merged in #7899 should solve this, with a slightly more robust clamp.

PR #7900, opened 2025-10-03
[lexical-playground] Feature: Caption has been added to exportDOM and importDOM

### LINKED ISSUE #5025: image caption conversion
https://github.com/facebook/lexical/blob/ad18d11b3768877ee8fa2926debb895617b97d27/packages/lexical-playground/src/nodes/ImageNode.tsx#L45

I carefully reviewed the Playground project's image-related code,  if  import an  image DOM element  with a caption 'This is an image description,  how to display the caption correctly in editor?

Kepron:
## Description
These changes will allow `exportDOM` and `importDOM` to use the caption feature. Thanks to @etrepum on the Discord channel for guiding me on this.

_Also, I’m still new to GitHub and in the learning stage. If there’s any mistake in the PR, please be kind with me._

Closes #5025

### Before

- Users could only add images without caption on html export.

### After

- Now users can use caption for html export.

### DIFF
--- a/packages/lexical-playground/src/nodes/ImageNode.tsx
+++ b/packages/lexical-playground/src/nodes/ImageNode.tsx
@@ -65,19 +65,19 @@ function isGoogleDocCheckboxImg(img: HTMLImageElement): boolean {
 
 function $convertImageElement(domNode: Node): null | DOMConversionOutput {
   const img = domNode as HTMLImageElement;
-  if (img.src.startsWith("file:///") || isGoogleDocCheckboxImg(img)) {
+  if (img.src.startsWith('file:///') || isGoogleDocCheckboxImg(img)) {
     return null;
   }
-  const { alt: altText, src, width, height } = img;
+  const {alt: altText, src, width, height} = img;
 
   const parentElement = img.parentElement;
-  const isInsideFigure = parentElement && parentElement?.tagName === "FIGURE";
+  const isInsideFigure = parentElement && parentElement?.tagName === 'FIGURE';
 
   if (isInsideFigure) {
-    let captionText = "";
-    const figcaption = parentElement.querySelector("figcaption");
+    let captionText = '';
+    const figcaption = parentElement.querySelector('figcaption');
     if (figcaption) {
-      captionText = figcaption.textContent || figcaption.innerText || "";
+      captionText = figcaption.textContent || figcaption.innerText || '';
       figcaption.remove(); // This is to prevent Lexical from adding an extra span after inserting an image because of the figcaption.
     }
     const captionEditor = createEditor({
@@ -93,16 +93,16 @@ function $convertImageElement(domNode: Node): null | DOMConversionOutput {
     }
     const node = $createImageNode({
       altText,
+      caption: captionEditor,
       height,
+      showCaption: true,
       src,
       width,
-      showCaption: true,
-      caption: captionEditor,
     });
-    return { node };
+    return {node};
   }
-  const node = $createImageNode({ altText, height, src, width });
-  return { node };
+  const node = $createImageNode({altText, height, src, width});
+  return {node};
 }
 
 export type SerializedImageNode = Spread<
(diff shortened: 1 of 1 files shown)


Kepron:
I discovered that every time an image with a caption is imported, an extra `p` element with a `br` child is added above and below it. I’m turning the PR into a draft until I fix this bug.

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -320,13 +320,11 @@ function $appendNodesToJSON(
...
-    clone =
-      $isTextNode(clone) && selection !== null
-        ? $sliceSelectedTextNodeContent(selection, clone)
-        : clone;
-    target = clone;
+  if (selection !== null && $isTextNode(target)) {
+    target = $sliceSelectedTextNodeContent(
+      selection,
+      $cloneWithProperties(target),
+    );
```
etrepum, line 327: This is just a performance improvement, it was doing some unnecessary work by cloning every node when there is a selection instead of just nodes it intends to modify

packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -454,7 +484,7 @@ export default function ImageComponent({
         {showCaption && (
           <div className="image-caption-container">
             <LexicalNestedComposer initialEditor={caption}>
-              <AutoFocusPlugin />
+              <DisableCaptionOnBlur setShowCaption={setShowCaption} />
```
etrepum, line 487: This is a UI improvement, if you remove the caption content and the caption editor loses focus then it will revert to the same state as it had before clicking the add caption button

packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -454,7 +484,7 @@ export default function ImageComponent({
         {showCaption && (
           <div className="image-caption-container">
             <LexicalNestedComposer initialEditor={caption}>
-              <AutoFocusPlugin />
```
etrepum, line 457: It didn't make sense to autofocus in all cases (e.g. when an image with caption was pasted), so I removed this plugin

packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -390,11 +413,18 @@ export default function ImageComponent({
...
       if ($isImageNode(node)) {
-        node.setShowCaption(true);
+        node.setShowCaption(show);
+        if (show) {
+          node.__caption.update(() => {
+            if (!$getSelection()) {
+              $getRoot().selectEnd();
+            }
+          });
+        }
```
etrepum, line 427: This is the equivalent to the AutofocusPlugin so when you click the add caption button it focuses the caption editor

packages/lexical-playground/src/nodes/ImageNode.tsx
```diff
@@ -62,14 +75,28 @@ function isGoogleDocCheckboxImg(img: HTMLImageElement): boolean {
 
 function $convertImageElement(domNode: Node): null | DOMConversionOutput {
   const img = domNode as HTMLImageElement;
-  if (img.src.startsWith('file:///') || isGoogleDocCheckboxImg(img)) {
+  const src = img.getAttribute('src');
```
etrepum, line 78: using getAttribute prevents it from adding the origin when the original src was a relative path

packages/lexical-playground/src/nodes/ImageNode.tsx
```diff
@@ -137,17 +164,87 @@ export class ImageNode extends DecoratorNode<JSX.Element> {
...
+    }
+
+    return {element: imgElement};
   }
 
   static importDOM(): DOMConversionMap | null {
     return {
-      img: (node: Node) => ({
+      figcaption: () => ({
+        conversion: () => ({node: null}),
```
etrepum, line 214: figcaption is imported directly from the figure importer, we ignore it here to save some work

packages/lexical/src/LexicalNode.ts
```diff
@@ -546,7 +546,6 @@ export class LexicalNode {
 
     if (__DEV__) {
       if (this.__type !== 'root') {
-        errorOnReadOnly();
```
etrepum, line 549: This is a bug fix where `$cloneWithProperties` would throw when it shouldn't. The invariant this is checking is already checked appropriately above in `$setNodeKey`

packages/lexical-playground/src/nodes/ImageNode.tsx
```diff
@@ -137,17 +164,87 @@ export class ImageNode extends DecoratorNode<JSX.Element> {
...
+        priority: 0,
+      }),
+      figure: () => ({
+        conversion: (node) => {
+          return {
+            after: (childNodes) => {
```
etrepum, line 220: We expect to parse an img node from the children, and if so then we find the associated caption and generate the caption editor's content from its DOM

etrepum (commented):
I added an e2e test, made it import/export html instead of just plain text, and fixed some bugs here

PR #7906, opened 2025-10-06
[lexical-playground] Bug Fix: Export ParagraphNode as div role="paragraph" when it contains block DOM

etrepum:
## Description

Follow-up to #7900 which added `<figure>` export to `` nodes with caption.

The HTML spec does not allow the p tag to contain block-level elements ([mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/p)). figure and figcaption are block-level elements (as are some other things rendered by decorators that may end up inline). There's no way for us to say that the `<figure>` node is actually not used inline in this case, nor is it possible to really attach any other sort of HTML caption to an `img` node (the alt or title could be used, but is [not advised by mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#the_title_attribute))

> Paragraphs are [block-level elements](https://developer.mozilla.org/en-US/docs/Glossary/Block-level_content), and notably will automatically close if another block-level element is parsed before the closing `</p>` tag.

Using an element without Tag Omission such as the div element ([mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/div)) in these cases will allow us to serialize the figure node inside of a logical paragraph

### Before

The symptom of this is that when you use a DOMParser that conforms to the HTML5 spec is that you will split `<p>` tags at every `<figure>` or other block-level element.

### After

When a block-level child of a `<p>` tag is detected in the playground's html export override for ParagraphNode, it will replace that `<p>` tag with `<div role="paragraph">` which avoids the issue.


packages/lexical-playground/__tests__/unit/docSerialization.test.ts
```diff
@@ -1,11 +1,3 @@
-/**
```
etrepum, line 1: This environment is not needed anymore now that we are using vitest

packages/lexical-playground/src/buildHTMLConfig.tsx
```diff
@@ -0,0 +1,129 @@
...
+          }
+          return output;
+        },
+      };
+    };
+  }
+
+  return importMap;
+}
+function buildExportMap(): DOMExportOutputMap {
```
etrepum, line 86: This is the new functionality to override paragraph export

packages/lexical-playground/src/App.tsx
```diff
@@ -202,7 +133,7 @@ function App(): JSX.Element {
           : emptyEditor
             ? undefined
             : $prepopulatedRichText,
-        html: {import: buildImportMap()},
+        html: buildHTMLConfig(),
```
etrepum, line 136: all of the import/export code was extracted to another file because some imports in App.tsx are not importable from node (excalidraw stuff)

PR #7914, opened 2025-10-12
[lexical] Feature: Create empty paragraph exiting edge nodes

### LINKED ISSUE #7912: Feature: create empty paragraph before/after code block if it's the first/last node
## Description

Image

## Impact
All users.

noamzaks:
## Description

Closes #7912.


noamzaks:
Waiting to see if this looks like a good direction before continuing, if anyone has the time to take a look

etrepum:
I think this would be better implemented as an opt-in extension rather than at the lowest level. There are other situations where it’s preferable that extra paragraphs shouldn’t be created (when the editor has a more fixed layout, like the playground in markdown editing mode)

PR #7919, opened 2025-10-13
[lexical-markdown] Bug Fix: Prevent Markdown links with empty string link text from being automatically removed

adambolcsfoldi:
## Description

Currently any markdown link where the _link text_ is an empty string, e.g. `[](https://lexical.dev)`, will be automatically removed on transformation. According to the [CommonMark Spec](https://spec.commonmark.org/0.31.2/#example-487), it's a valid link.
With this pull request the removal of the link's text node is prevented by using the _Unmergeable_ flag.

Not sure what the desired behaviour is here. Admittedly, a link without a corresponding text isn't terribly useful unless handled by the UI layer in some way. 
An option would be to not parse it as a valid link as is currently done when the _link destination_ is empty string (`[some link title]()`), even despite being compliant with the [CommonMark Spec](https://spec.commonmark.org/0.31.2/#example-487).

### Before

https://github.com/user-attachments/assets/9c9d6943-cfe1-4bcd-9988-d0353b4aa01e

### After

https://github.com/user-attachments/assets/9c68ca12-9282-413b-9172-a308b8794fe1

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -604,6 +604,7 @@ export const LINK: TextMatchTransformer = {
     }
     const linkTextNode = $createTextNode(parsedLinkText);
     linkTextNode.setFormat(textNode.getFormat());
+    linkTextNode.toggleUnmergeable();
     linkNode.append(linkTextNode);
     textNode.replace(linkNode);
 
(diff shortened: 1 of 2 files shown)


etrepum:
I don't think that strict compliance is useful here, an empty unmergeable TextNode is going to cause other problems (you can't really interact with it). A zwsp might be better if you really wanted to parse such a construct. There are plenty of other instances where better conformance to the commonmark spec would be useful but I don't think this is one of them.

adambolcsfoldi:
> I don't think that strict compliance is useful here, an empty unmergeable TextNode is going to cause other problems (you can't really interact with it). A zwsp might be better if you really wanted to parse such a construct. There are plenty of other instances where better conformance to the commonmark spec would be useful but I don't think this is one of them.

I can definitely get behind that. 
However, I do think the current behaviour where the user input just vanishes feels quite broken. Should it then instead not be parsed as a markdown link at all, unless the length of the link text > 0? 
So, `[](https://lexical.dev)` will not be parsed as a link, the same way `[some text]()` isn't now, but `[a](https://lexical.dev)` would be and result in [a](https://lexical.dev).

etrepum:
That's probably slightly better than dropping it altogether, but I don't have a strong preference about that. lexical isn't specifically a markdown editor so lossless transformation isn't really the expectation. It's certainly something that could be built but not very easily with the architecture of the current `@lexical/markdown` package (but there's nothing special about that package, someone could publish or contribute a better one).

adambolcsfoldi:
Closed in favour of a more useful solution here: https://github.com/facebook/lexical/pull/7923

PR #7921, opened 2025-10-13
[lexical-code] Bug Fix: Respect RTL when moving to line start/end in code blocks

### LINKED ISSUE #7913: Bug: command+arrow moves to opposite side in RTL in code blocks
Lexical version: v0.37.0+dev.esm

## Steps To Reproduce

1. Write an RTL line, e.g. "איזה כיף" inside a code block
2. Press command+right or command+left

## The current behavior
The caret moves to the opposite end of the paragraph.

## The expected behavior
The caret moves to the correct end of the paragraph.

## Impact of fix
Small impact for RTL users.

ashmod:
## Description

  - Current behavior
      - In code blocks, Command(CTRL)+Left/Right (`MOVE_TO_START`/`END`) on RTL lines moves the caret to the opposite visual end of the line.
  - New behavior
      - Detect the line’s direction and flip `MOVE_TO_START`/`END` semantics for RTL so the caret moves to the correct visual ends.
      - LTR behavior remains unchanged. Non-code content is unaffected.
  - Implementation
      - Add `$getCodeLineDirection`(anchor) in @lexical/code:
          - Scans the code line for the first strong LTR/RTL character across CodeHighlightNode/TabNode segments.
          - Falls back to the parent element’s direction; returns null if indeterminate.
      - Use `$getCodeLineDirection` in both highlighters (Prism, Shiki) to determine whether to invert `MOVE_TO_START/END`

  Closes #7913

  ## Test plan

  ### Before

  - Steps to reproduce
      - Insert a code block; type an RTL string (e.g., "تجربة").
      - Place the caret within the line.
      - Press Command(CTRL)+Left (MOVE_TO_START) or Command(CTRL)+Right (`MOVE_TO_END`).
      - Caret jumps to the opposite visual end of the line.
  - Automated tests
      - No RTL-specific unit coverage in code blocks for `MOVE_TO_START`/`END`.

  ### After

  - Steps to verify
      - Repeat the steps above.
      - `MOVE_TO_START` moves caret to visual left; `MOVE_TO_END` moves caret to visual right on RTL lines.
      - LTR lines behave as before.
  - Automated tests
      - Added RTL test cases to `LexicalCodeNode.test.ts`

### DIFF
--- a/packages/lexical-code/src/FlatStructureUtils.ts
+++ b/packages/lexical-code/src/FlatStructureUtils.ts
@@ -74,12 +74,9 @@ export function $getCodeLineDirection(
 
   while (node !== null) {
     if ($isCodeHighlightNode(node)) {
-      const text = node.getTextContent();
-      for (let i = 0; i < text.length; i++) {
-        const direction = getTextDirection(text[i]);
-        if (direction !== null) {
-          return direction;
-        }
+      const direction = getTextDirection(node.getTextContent());
+      if (direction !== null) {
+        return direction;
       }
     }
 


packages/lexical-code/src/index.ts
```diff
@@ -53,3 +55,5 @@ export const getLastCodeNodeOfLine = $getLastCodeNodeOfLine;
 export const getEndOfCodeInLine = $getEndOfCodeInLine;
 /** @deprecated renamed to {@link $getStartOfCodeInLine} by @lexical/eslint-plugin rules-of-lexical */
 export const getStartOfCodeInLine = $getStartOfCodeInLine;
+/** @deprecated renamed to {@link $getCodeLineDirection} by @lexical/eslint-plugin rules-of-lexical */
+export const getCodeLineDirection = $getCodeLineDirection;
```
etrepum:
```suggestion
```

It doesn't really make sense to add deprecated APIs that didn't exist before 😆

packages/lexical-code/src/FlatStructureUtils.ts
```diff
@@ -16,11 +16,39 @@ import type {
...
+    const character = text[i];
+    if (RTL_TEXT_REGEX.test(character)) {
+      return 'rtl';
+    }
+    if (LTR_TEXT_REGEX.test(character)) {
+      return 'ltr';
+    }
+  }
+  return null;
+}
```
etrepum: All of this could probably be eliminated by exporting and re-using the currently internal utility function `getTextDirection` from the lexical package

ashmod: updated in [12665e9](https://github.com/facebook/lexical/pull/7921/commits/12665e9e008a1d75456b53524670b770378fb5dc)

packages/lexical-code/src/FlatStructureUtils.ts
```diff
@@ -51,6 +57,50 @@ export function $getLastCodeNodeOfLine(
...
+
+  while (node !== null) {
+    if ($isCodeHighlightNode(node)) {
+      const text = node.getTextContent();
+      for (let i = 0; i < text.length; i++) {
+        const direction = getTextDirection(text[i]);
+        if (direction !== null) {
+          return direction;
+        }
+      }
```
etrepum: It's very inefficient (and less correct) to do this one code unit at a time

```suggestion
      const direction = getTextDirection(node.getTextContent());
      if (direction !== null) {
        return direction;
      }
```

ashmod: Thanks, I've applied this. I initially thought `getTextDirection` would need to be called per unit since it looks up the first RTL character, which I assumed might break cases where a user types in LTR and then switches to RTL. But it seems to be handled correctly.

noamzaks:
This looks pretty good to me, although I don't have the best knowledge in BiDi specifications it looks familiar.

Thank you! :)

PR #7941, opened 2025-10-22
[lexical] Bug Fix: retain selection during updates on unfocused editor

fantactuka:
## Description

There's an issue when editor loses selection state when it's blurred (DOM selection is elsewhere), and editor update is triggered without user event (so window.event is null): 

https://stackblitz.com/edit/lexical-selection-retention?file=index.html

E2e test is replicating the same thing: adds content and selection state to the editor, focus into font-size input, triggers programmatic update. Selection state is lost

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -1667,8 +1667,9 @@ test.describe.parallel('Selection', () => {
   test('programatic update on blurred editor does not kill selection', async ({
     page,
     isPlainText,
+    isCollab,
   }) => {
-    test.skip(isPlainText);
+    test.skip(isPlainText || isCollab);
     await focusEditor(page);
     await page.keyboard.type('Hello');
     await page.locator('input.font-size-input').focus();


fantactuka:
Found it when editor's text color was updated with debounce (so there's no associated window.event), and caret is within color picker input. `skip-dom-selection` tag does not do much, since selection is being set to null way before - when creating initial selection for pending editor state:

https://github.com/fantactuka/lexical/blob/9e4c5bdfe61264f9e5d594704fa4ed1c87b7cc1f/packages/lexical/src/LexicalUpdates.ts#L945-L949

packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
```diff
@@ -1663,4 +1663,52 @@ test.describe.parallel('Selection', () => {
       focusPath: [1, 1, 0],
     });
   });
+
+  test('programatic update on blurred editor does not kill selection', async ({
+    page,
+    isPlainText,
+  }) => {
+    test.skip(isPlainText);
```
etrepum: I think this also needs to skip collab

PR #7942, opened 2025-10-22
[lexical-clipboard] Bug Fix: Log exceptions in clipboard paste handler

### LINKED ISSUE #7940: Bug: Exception is swallowed when pasting into editor
Lexical version: 0.37.0

## Steps To Reproduce

1. Use Editor's initialConfig: { html: { import: { span: (el) => { throw new Error("Unintentional error."); } } } }, i.e. have custom import handling for an HTML element and do an unintentional error there.
2. Copy any HTML containing <span> element.
3. When pasting, the exception is swallowed - it's not shown anywhere, so developer has no clue why the import is behaving weirdly (it looks like it falls back to something like non-HTML plain-text paste).

Link to code example:

[https://stackblitz.com/edit/facebook-lexical-7nhfkbzs?file=src%2FApp.tsx](https://stackblitz.com/edit/facebook-lexical-7nhfkbzs?file=src%2FApp.tsx)

## The current behavior
Exception is swallowed. Likely due to:
[https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts#L148](https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts#L148)
[https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts#L168](https://github.com/facebook/lexical/blob/main/packages/lexical-clipboard/src/clipboard.ts#L168)

## The expected behavior
Exception is logged into console.

## Impact of fix
Developers will thank you for not making it harder for them ;-).

niikkhilsharma:
This PR fixes issue #7940 where exceptions thrown during HTML paste operations were silently swallowed, making it impossible for developers to debug import handlers.

Current behavior: When an error occurs in custom HTML import handlers (e.g., html.import.span), the exception is caught in clipboard.ts but never displayed or logged. Developers have no visibility into why paste operations fail or behave unexpectedly.

Changes made: Modified the error handling in @lexical/clipboard to properly log exceptions during HTML paste operations. The exceptions are now surfaced to the console with full stack traces, allowing developers to identify and fix issues in their custom import handlers.

The fix ensures that while paste operations remain stable in production, developers get the diagnostic information they need during development and debugging.

Closes #7940

### After
image

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -1663,4 +1663,53 @@ test.describe.parallel('Selection', () => {
       focusPath: [1, 1, 0],
     });
   });
+
+  test('programatic update on blurred editor does not kill selection', async ({
+    page,
+    isPlainText,
+    isCollab,
+  }) => {
+    test.skip(isPlainText || isCollab);
+    await focusEditor(page);
+    await page.keyboard.type('Hello');
+    await page.locator('input.font-size-input').focus();
+
+    // It is important that his update is not called via UI event (e.g., as onClick handler)
+    // as internal code relies on window.event to track those
+    await page.evaluate(() => {
+      const editor = document.querySelector(
+        'div[contenteditable="true"]',
+      ).__lexicalEditor;
+
+      if (editor._editorState._selection == null) {
+        throw new Error('Expected selection to be no null');
+      }
+
+      return new Promise((resolve) => {
+        editor.update(
+          () => {
+            for (const node of editor._editorState._nodeMap) {
+              if (node.type === 'text') {
+                node.toggleFormat('bold');
+              }
+            }
+          },
+          {
+            onUpdate: resolve,
+            tag: 'skip-dom-selection',
+          },
+        );
+      });
+    });
+
+    await page.evaluate(() => {
+      const editor = document.querySelector(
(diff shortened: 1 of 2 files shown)


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -144,8 +144,9 @@ export function $insertDataTransferForRichText(
         const nodes = $generateNodesFromSerializedNodes(payload.nodes);
         return $insertGeneratedNodes(editor, nodes, selection);
       }
-    } catch {
-      // Fail silently.
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log(error);
```
etrepum:
```suggestion
      console.error(error);
```

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -164,8 +165,9 @@ export function $insertDataTransferForRichText(
       );
       const nodes = $generateNodesFromDOM(editor, dom);
       return $insertGeneratedNodes(editor, nodes, selection);
-    } catch {
-      // Fail silently.
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log(error);
```
etrepum:
```suggestion
      console.error(error);
```

etrepum (commented):
The error handling elsewhere uses console.error rather than console.log

PR #7944, opened 2025-10-23
[lexical-link] Feature: Enable Selective Removal Within Linked Text

### LINKED ISSUE #7943: Bug Fix: Enable Selective Removal Within Linked Text
When attempting to remove part of a linked word, lexical will remove the link from the entire text rather than just the selected portion. This forces the user to reapply the link to the intended text.

Lexical version: 0.32.1

## Steps To Reproduce

1. Write some text and select it.
2. Apply a link to the text.
3. Select a portion of that.
4. Try to unlink a portion of the text.
5. The entire link text will become unlinked and not the selected portion. 

## The current behavior

When selecting a portion of the text to unlink, it unlinks the entire text.

## The expected behavior

If I'm selecting a portion of the linked text, like some extra space for example, I expect only that portion of the text to be unlinked. 

## Impact of fix
This happens anytime I want to unlink text

normtronics:
## Description

This PR allows for specific highlighted text to be be unlinked. This fixes when attempting to remove part of a linked word, lexical will remove the link from the entire text rather than just the selected portion. This forces the user to reapply the link to the intended text.

Closes #7943 

## Test plan
1. Write some text and select it.
2. Apply a link to the text.
3. Select a portion of that.
4. Try to unlink a portion of the text.
5. The selected text should be unlinked, not the entire text.

### Before
Trying to unlink a portion of the text would cause the entire text to become unlinked.

### After
When selecting a portion of the text to unlink, only that portion will become unlinked.

### DIFF
--- a/packages/lexical-clipboard/src/clipboard.ts
+++ b/packages/lexical-clipboard/src/clipboard.ts
@@ -144,8 +144,9 @@ export function $insertDataTransferForRichText(
         const nodes = $generateNodesFromSerializedNodes(payload.nodes);
         return $insertGeneratedNodes(editor, nodes, selection);
       }
-    } catch {
-      // Fail silently.
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
     }
   }
 
@@ -164,8 +165,9 @@ export function $insertDataTransferForRichText(
       );
       const nodes = $generateNodesFromDOM(editor, dom);
       return $insertGeneratedNodes(editor, nodes, selection);
-    } catch {
-      // Fail silently.
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
     }
   }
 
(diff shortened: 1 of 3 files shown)


normtronics:
@etrepum After switching to Lexical for some clients, they noticed an issue with unlinking portions of text. It became cumbersome when they accidentally linked spaces at the end of a selection because when they tried to unlink just the space, it would remove the link from the entire text instead.

What should have been a one-step process turned into two.

efjacobson:
> Is there some specific editor(s) that this change is inspired by? It seems that most other editors we have looked at don't really have this behavior.

@etrepum thanks for the quick merge. you can see this behavior in [ProseMirror](https://prosemirror.net/), [Quill](https://quilljs.com/playground/react), [CKEditor](https://ckeditor.com/ckeditor-5/demo/feature-rich/), and the (3rd party, i know) [DraftJS link/anchor plugin](https://www.draft-js-plugins.com/plugin/anchor). the DraftJs plugin being the reason our clients are used to having that functionality

PR #7954, opened 2025-10-28
Fix: Unicode URL regex for Arabic and multilingual domain/link support (#7883)

### LINKED ISSUE #7883: Bug: Arabic links bug
Image

Image

it should be a valid link

BiswajitHemram:
## Description

Fixes an issue where **Arabic and other non-Latin URLs** (e.g., Arabic, Cyrillic, Chinese, accented domains) were **not recognized as valid links** by the `AutoLinkPlugin`.

After this change, **Unicode URLs** will be correctly detected and linked, improving multilingual support for the Lexical Playground and other apps using this plugin.

Closes #7883

###  Steps to reproduce
1. Go to [https://lexical.dev/](https://lexical.dev/)
2. Open the [Playground](https://playground.lexical.dev/)
3. Paste one of the following URLs into the editor:

   * `https://qabilah.com/posts/عربي`
   * `http://пример.рф/новости`
   * `https://例子.中国/测试`
4. The links should now be **automatically detected** and **clickable**.

### Before

before-fix

### After

after-fix


packages/lexical-playground/src/plugins/AutoLinkPlugin/index.tsx
```diff
@@ -15,7 +15,7 @@ import {
 import * as React from 'react';
 
 const URL_REGEX =
-  /((https?:\/\/(www\.)?)|(www\.))[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)(?<![-.+():%])/;
+  /^((https?:\/\/(www\.)?)|(www\.))[\p{L}\p{N}\p{M}_-]{1,63}(\.[\p{L}\p{N}\p{M}_-]{1,63})+(:\d+)?(\/[\p{L}\p{N}\p{M}\p{S}\p{P}@:%_+.~#?&//=-]*)?$/u;
```
fantactuka, line 19: `^...$` will only work if url is the only thing on a line, if there's anything else it won't work

It breaks query params `https://lexical.dev?error=123`

It's more lose than original one in terms of special chars, e.g., allowing `https://foo.bar_` as valid URL

Could you address those, please

fantactuka: You can also run AutoLink e2e tests locally to verify solution. Could also add more to cover the fix:

```bash
npm run test-e2e-chromium -- packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
```

etrepum: Adding new tests to show that it works as intended would be appreciated as well

BiswajitHemram: Thanks for the clarification, @fantactuka! @etrepum

I’ve understood your points:
* `^...$` anchors should be removed so the regex can detect URLs within a line.
* Query parameters like `?error=123` should be fully supported.
* Underscore (_) is not valid in domain names, so I’ll remove it from the domain-matching part.

I’ll update the regex accordingly and add these cases to make sure the code doesn’t break in future tests.

etrepum: Looks like the requested changes have been completed, blocked on @fantactuka to review since changes were requested

BiswajitHemram: @etrepum @fantactuka, could you please let me know what the exact problem is?

etrepum:
It looks like the issue with this is Firefox-specific, you can reproduce locally by running those e2e tests.

In one terminal start the app with `npm run start`, and in another run the e2e tests, here's one of the failing tests:

```
npm run test-e2e-firefox packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs:757:3 
```

It looks like it's dropping a character. To isolate I would just open up Firefox and test the regex out and see if it's happening at that level or something higher-level like text input.

etrepum:
It looks like the problem is `page.keyboard.type` on Firefox with unicode characters. If you change these to `page.keyboard.insertText` then the tests should pass.

PR #7964, opened 2025-10-31
[lexical] Feature: add commands for beforeinput, input, compositionstart/end events

james-atticus:
## Description

In order to any sort of track changes/suggestion mode feature, consumers need to be able to intercept/override events that mutate editor state. This is already the case for many operations, but not for input and composition events.

This PR adds commands for `beforeinput`, `input`, `compositionstart` and `compositionend`. The existing functions for these events have been repurposed into command handlers which are registered when creating a new editor instance.

## Test plan

One test was relying on `editor._commands` being empty by default; updated that test to be slightly more targeted.

### DIFF
--- a/packages/lexical/src/LexicalEvents.ts
+++ b/packages/lexical/src/LexicalEvents.ts
@@ -1048,7 +1048,7 @@ function onCompositionStart(
   event: CompositionEvent,
   editor: LexicalEditor,
 ): void {
-  dispatchCommand(editor, COMPOSITION_START_COMMAND, event as CompositionEvent);
+  dispatchCommand(editor, COMPOSITION_START_COMMAND, event);
 }
 
 function $handleCompositionStart(event: CompositionEvent): boolean {


etrepum (commented):
I think this is probably right and I'll give it a closer look later but one thing to think about here is that with nested editors there's this weird "feature" where commands from one editor can bubble to the parent editor if no handler (at any priority) returns true. A thorough review to make sure that the way these events are handled in nested editor situations in the same way with the old method and this new command based method are equivalent.

packages/lexical/src/LexicalCommands.ts
```diff
@@ -37,6 +37,15 @@ export const SELECTION_INSERT_CLIPBOARD_NODES_COMMAND: LexicalCommand<{
...
   createCommand('CLICK_COMMAND');
+export const BEFORE_INPUT_COMMAND: LexicalCommand<InputEvent> = createCommand(
+  'BEFORE_INPUT_COMMAND',
+);
+export const INPUT_COMMAND: LexicalCommand<InputEvent> =
+  createCommand('INPUT_COMMAND');
+export const COMPOSITION_START_COMMAND: LexicalCommand<CompositionEvent> =
+  createCommand('COMPOSITION_START_COMMAND');
+export const COMPOSITION_END_COMMAND: LexicalCommand<CompositionEvent> =
+  createCommand('COMPOSITION_END_COMMAND');
```
james-atticus, line 48: TODO: export these from `index.ts`

etrepum: I'll wait for these changes before reviewing

packages/lexical/src/LexicalEvents.ts
```diff
@@ -905,138 +941,156 @@ function onInput(event: InputEvent, editor: LexicalEditor): void {
...
+  return true;
 }
 
 function onCompositionStart(
   event: CompositionEvent,
   editor: LexicalEditor,
 ): void {
-  updateEditorSync(editor, () => {
-    const selection = $getSelection();
+  dispatchCommand(editor, COMPOSITION_START_COMMAND, event as CompositionEvent);
```
etrepum:
```suggestion
  dispatchCommand(editor, COMPOSITION_START_COMMAND, event);
```

nit: redundant cast

etrepum (dismissed):
This looks good to me, since all of these new command listeners are unconditionally returning true and the manner in which the DOM events are being listened to doesn't change there doesn't seem to be any risk of different behavior for editor propagation

code-alexander:
This broke KEY_DOWN_COMMAND for me. Was able to resolve it by changing the priority to COMMAND_PRIORITY_HIGH instead of COMMAND_PRIORITY_EDITOR.

Sharing in case it's helpful.

mahendrimd:
thanks @code-alexander , for me `COMMAND_PRIORITY_NORMAL` is enough
I think this is somehow related to the PR on [LexicalEditor.ts#665](https://github.com/james-atticus/lexical/blob/a7f597d46ab9eb90b0e3cbfebe1e250cb67cc635/packages/lexical/src/LexicalEditor.ts#L665), where keydown and beforeinput in `registerDefaultCommandHandlers` use `COMMAND_PRIORITY_EDITOR`

PR #7969, opened 2025-11-03
Feature: Update Lexical link types for Flow

thegreatercurve:
## Description

We're updating the the Flow `dom` types for event listener options to be more consistent with the spec: https://dom.spec.whatwg.org/#eventtarget. 

This diff updates the relevant Flow types in Lexical to now be consistent with the new globals.

## Test plan

```
npm run build-www
```


etrepum (dismissed):

takuyakanbr:
The typings for the `dom` library need to be updated in `flow-typed`. That's where we're pulling the `dom` types from (as of #7785), since `flow-bin` is no longer bundling them.

takuyakanbr:
Changes have been made in #7973

PR #7971, opened 2025-11-03
[lexical] Feature: Only serialize ElementNode textFormat and textStyle when necessary

### LINKED ISSUE #7968: Editor State from Headless Lexical and React Lexical slightly differs
Hi!

There is a lot of context in the issue referenced below, but to be short: it seems that extracting the Editor State from a headless or a React Lexical differs slightly (as shown in the context below).

This causes a few problems when trying to use the library with form libraries and to calculate dirty states. One of the fix is to manually manipulate the Editor State to remove the flaky properties, but it's risky as I'm touching Lexical internal properties with no guarantees that the system won't break in the future.

Is it possible to make sure that for the same content, the 2 Lexical editors can output the same Editor States?

---

Hi guys

We are facing a gnarly problem with our implementation of Lexical and React Hook Form.

## Introduction
We have had normal inputs with html inside, and we are trying to implement Lexical to replace them. In our DB I store the html string, then the frontend fetches this html string, uses a **headless Lexical** to convert the html to a Lexical Editor State, and then use that calculated Editor State to initialize our Form, and by consequence Lexical as well. This is of course an adapter logic just for the migration; as soon as the user saves the Lexical again, we'll store directly the json stringified Editor State and not the html string as we used to have before.

This is the `htmlToEditorStateHeadless` function by the way, using the same nodes and config as our React Lexical editor:
```ts
/**
 * Convert an HTML string into a stringified Lexical EditorState using a headless editor.
 * Returns an empty string if the produced state has no text content.
 */
export function htmlToEditorStateHeadless(html: string, options: Options = {}): string {
  const content = html.trim?.();
  if (!content) return '';

  const { nodes = NODES, namespace = NAMESPACE } = options;

  try {
    const editor = createHeadlessEditor({ namespace, onError, nodes });

    editor.update(
      () => {
        const dom = parser.parseFromString(content, 'text/html');
        const nodesFromDom = $generateNodesFromDOM(editor, dom);
        $getRoot().clear();
        $getRoot().select();
        $insertNodes(nodesFromDom);
      },
      {
        discrete: true,
      },
    );

    const editorState = editor.getEditorState();

    let isEmpty = false;
    editorState.read(() => {
      isEmpty = $isRootTextContentEmpty(false, true);
    });
    if (isEmpty) return '';

    return JSON.stringify(editorState);
  } catch (err) {
    onError(err);
    return '';
  }
}
```

## Problem
The problem we are facing is that, as soon as the user touches the Lexical editor, the RHF goes `dirty`. I checked the diff between the initialization Lexical state and the new one after touching it, and it seems that the difference appears here

image

There is `textFormat: 3` that appears twice in the modified version. **This only happens on html strings that have some kind of rich texts inside, like bold, italic, etc.**.

## Improvement attempt
First I was wondering why just the mere clicking on the Lexical Editor would trigger a change, then I found out that the `OnChange` plugin has a property called `ignoreSelectionChange`. This fixed the fact that the change would happen on merely clicking in the editor but it didn't fix the underlying problem.

In fact, now if I add a letter to the Editor and delete it right after, the underlying Editor State results modified and, by consequence, the form is dirty. In this case, I'd expect the state to be equal to initialization, which is not.

---

Any suggestions on what could cause the problem, or what could we do better to avoid this problem or achieve our migration? Thank you 🙏

_Originally posted by @buondevid in https://github.com/facebook/lexical/discussions/7965_

etrepum:
## Breaking Change

Previously the derived properties `textFormat` and `textStyle` would always be serialized to JSON if not set to the default values. These should only be useful when the ElementNode does not currently have any TextNode children, to preserve formatting choices, and will be recomputed when reconciled. Now they are only reconciled when they are determined to be useful (in an ElementNode with no direct TextNode children that is not a root or shadow root).

## Description

Avoid serializing derived `textFormat` and `textStyle` whenever their values would be useless (in a root or shadow root that should not have TextNode children) or recomputed on first reconcile (when the ElementNode has a TextNode child).

Also fixes a bug where in some cases the ElementNode's format and style could be coming from two different TextNodes. It should really only come from the first TextNode.

Closes #7968

## Future Considerations

We should eventually start changing the JSON in backwards incompatible ways, notably to remove default and useless values (e.g. the version attribute). They make the data less compact and provide no value other than to be compatible with very old versions of lexical.

## Test plan

New unit tests and all existing tests should pass (nothing exercised these edge cases)

### DIFF
--- a/packages/lexical-code/src/CodeHighlighterPrism.ts
+++ b/packages/lexical-code/src/CodeHighlighterPrism.ts
@@ -149,6 +149,7 @@ function codeNodeTransform(
   tokenizer: Tokenizer,
 ) {
   const nodeKey = node.getKey();
+  const cacheKey = editor.getKey() + '/' + nodeKey;
 
   // When new code block inserted it might not have language selected
   if (node.getLanguage() === undefined) {
@@ -168,11 +169,11 @@ function codeNodeTransform(
     return;
   }
 
-  if (nodesCurrentlyHighlighting.has(nodeKey)) {
+  if (nodesCurrentlyHighlighting.has(cacheKey)) {
     return;
   }
 
-  nodesCurrentlyHighlighting.add(nodeKey);
+  nodesCurrentlyHighlighting.add(cacheKey);
 
   // Using nested update call to pass `skipTransforms` since we don't want
   // each individual CodeHighlightNode to be transformed again as it's already
@@ -211,7 +212,7 @@ function codeNodeTransform(
     },
     {
       onUpdate: () => {
-        nodesCurrentlyHighlighting.delete(nodeKey);
+        nodesCurrentlyHighlighting.delete(cacheKey);
       },
       skipTransforms: true,
     },


ivailop7 (approved):

buondevid:
Thank you for this @etrepum!

PR #7972, opened 2025-11-04
[lexical-markdown] Option: $convertToMarkdownString with shouldPreserveWhitespaces

baptistejamin:
[lexical-markdown] Option: $convertToMarkdownString with shouldPreserveWhitespaces

## Description

By default, $convertToMarkdownString replaces whitespaces in somecases with HTML chars: `**&#32;&#32;&#32;foo&#32;&#32;&#32;**`, while it's not a big deal when exporting from Lexical and then rendering into Lexical, it breaks on other Markdown renderers.

### DIFF
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -1047,6 +1047,78 @@ describe('Markdown', () => {
   });
 });
 
+describe('whitespace preservation in export', () => {
+  it('exports leading/trailing whitespaces as HTML code points by default', () => {
+    const editor = createHeadlessEditor({
+      nodes: [
+        HeadingNode,
+        ListNode,
+        ListItemNode,
+        QuoteNode,
+        CodeNode,
+        LinkNode,
+      ],
+    });
+
+    const html =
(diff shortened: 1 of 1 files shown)


etrepum:
This fails the formatting check (should be able to fix with `npm run prettier:fix`) and is a breaking change because the default is not the current behavior

baptistejamin:
Thank you for the report.

It has been fixed.

etrepum (commented):
The only thing missing here are tests to show that it does what it's supposed to

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -1047,6 +1047,78 @@ describe('Markdown', () => {
...
+      {discrete: true},
+    );
+
+    const exported = editor
+      .getEditorState()
+      .read(() =>
+        $convertToMarkdownString([...TRANSFORMERS], undefined, false, true),
+      );
+
+    expect(exported).toBe('**   Hello   **');
```
etrepum, line 1118: Looking closer at what this is supposed to do now that there are tests I don't think this is useful, because this is not something most markdown parsers will accept.

For example here's what github does with it:

**   Hello   **

The commonmark reference implementation [also doesn't parse it as formatting](https://spec.commonmark.org/dingus/?text=**%20%20%20Hello%20%20%20**)

PR #7978, opened 2025-11-08
[lexical] Bug fix: $dfsCaretIterator should be able to stop at its last descendant

nigelgutzmann:
## Description
When you have an element node and you want to iterate through all it's children, you might do something like:

```
$dfsIterator(elementNode, elementNode.getLastDescendant());
```

You would expect it to end at the last descendant inside of the elementNode. However, currently it will iterate through all the descendants, and then just continue right on through to the end of the entire editor's tree.

I checked the condition where endCaret is null even though the user did pass in an endNode. In this case, we should get the adjacent parent as the end caret.
Closes #

## Test plan
Unit test included.

### Before
It would return all nodes to the end of the editor's tree

### After
It will properly stop at the node that is passed in as the endNode

### DIFF
--- a/packages/lexical-utils/src/index.ts
+++ b/packages/lexical-utils/src/index.ts
@@ -198,8 +198,10 @@ export interface DFSNode {
  * before backtracking and finding a new path. Consider solving a maze by hugging either wall, moving down a
  * branch until you hit a dead-end (leaf) and backtracking to find the nearest branching path and repeat.
  * It will then return all the nodes found in the search in an array of objects.
+ * Preorder traversal is used, meaning that nodes are listed in the order of when they are FIRST encountered.
  * @param startNode - The node to start the search (inclusive), if omitted, it will start at the root node.
- * @param endNode - The node to end the search (inclusive), if omitted, it will find all descendants of the startingNode.
+ * @param endNode - The node to end the search (inclusive), if omitted, it will find all descendants of the startingNode. If endNode
+ * is an ElementNode, it will stop before visiting any of its children.
  * @returns An array of objects of all the nodes found by the search, including their depth into the tree.
  * \\{depth: number, node: LexicalNode\\} It will always return at least 1 node (the start node).
  */
@@ -237,8 +239,10 @@ export function $reverseDfs(
 
 /**
  * $dfs iterator (left to right). Tree traversal is done on the fly as new values are requested with O(1) memory.
+ * Preorder traversal is used, meaning that nodes are iterated over in the order of when they are FIRST encountered.
  * @param startNode - The node to start the search (inclusive), if omitted, it will start at the root node.
  * @param endNode - The node to end the search (inclusive), if omitted, it will find all descendants of the startingNode.
+ * If endNode is an ElementNode, the iterator will end as soon as it reaches the endNode (no children will be visited).
  * @returns An iterator, each yielded value is a DFSNode. It will always return at least 1 node (the start node).
  */
 export function $dfsIterator(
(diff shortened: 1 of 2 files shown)


packages/lexical-utils/src/index.ts
```diff
@@ -269,11 +269,18 @@ function $dfsCaretIterator<D extends CaretDirection>(
...
+    endCaret = $getAdjacentChildCaret(
+      $getChildCaretOrSelf($getSiblingCaret(endNode, direction)),
+    );
+    if (endCaret === null) {
+      const next = $getAdjacentSiblingOrParentSiblingCaret(
+        $getSiblingCaret(endNode, direction),
+      );
+      endCaret = next && next[0];
+    }
+  }
```
etrepum: This looks like a copy pasted version of `$getEndCaret(endNode, direction)`


```suggestion
  const endCaret = endNode
    ? $getAdjacentChildCaret(
        $getChildCaretOrSelf($getSiblingCaret(endNode, direction)),
      ) || $getEndCaret(endNode, direction)
    : $getEndCaret(start, direction);
```

etrepum:
While there is clearly some sort of bug here I think if we're going to try and fix anything with these functions it should be a more comprehensive fix to the tests and API docs to fully specify what the behavior actually should be. The caret APIs were added because functions like this one (and many of the functions that work with selections) are so ambiguous about how the edges and ElementNodes should be handled. These functions aren't even used at all in lexical itself (outside of a few tests) except for a single `$dfs()` call in `useCharacterLimit` to iterate the entire document with no specified start or end nodes.

From the API docs it's not really even clear whether endNode should be inclusive or not nor does it specify what happens if it's an ElementNode with children.

nigelgutzmann:
@etrepum , I made the changes you requested - simplified the code and added to the jsdocs for the relevant functions. To maintain backwards compatibility, both startNode and endNode should be inclusive.

However, just to be clear, the problem being solved here is not whether the startNode and endNode are inclusive or not, but that in this situation (where endNode has no next sibling), the caller supplies an endNode, the iterator hits the end node, and then just keeps going and going all the way to the end of the ENTIRE editor tree. The API is pretty clear that if the caller supplies an endNode, the iterator should stop at the endNode - no ambiguity there.

nigelgutzmann:
I don't really understand your point about ElementNodes being ambiguous. To me, it seems like the rules of Depth-First-Search are pretty clear when it comes to a node with children. Can you explain your point a bit more?

etrepum:
My point is about an ElementNode being used as the endNode. ElementNode are basically visited twice, once on enter and once on exit. You could stop at or after either depending on the expected semantics of the function, because without carets to distinguish between them it’s very ambiguous. Depending on the definition there may or may not be a way to specify “include this ElementNode but not any of its children”.

etrepum:
To be clear there’s definitely a problem with the implementation of endNode before this PR, but it’s also not really clear what this function was really supposed to do with endNode. There were not any docs, tests, or usage to specify what the edge case handling should do.

packages/lexical-utils/src/index.ts
```diff
@@ -198,8 +198,8 @@ export interface DFSNode {
  * before backtracking and finding a new path. Consider solving a maze by hugging either wall, moving down a
  * branch until you hit a dead-end (leaf) and backtracking to find the nearest branching path and repeat.
  * It will then return all the nodes found in the search in an array of objects.
- * @param startNode - The node to start the search, if omitted, it will start at the root node.
- * @param endNode - The node to end the search, if omitted, it will find all descendants of the startingNode.
+ * @param startNode - The node to start the search (inclusive), if omitted, it will start at the root node.
+ * @param endNode - The node to end the search (inclusive), if omitted, it will find all descendants of the startingNode.
```
etrepum: This does not specify (in docs or tests) whether endNode is also inclusive of all descendants of endNode

nigelgutzmann:
Got it! Thanks for the explanation. I can make those changes! 👍

etrepum (approved):
This looks great, now the function clearly specifies how it's supposed to behave so it's finally clear what is a bug and what isn't

PR #7979, opened 2025-11-08
[lexical-markdown] Bug fix: skip format of the empty string on export 

### LINKED ISSUE #7575: Bug: Trailing whitespace being formatted produces invalid Common Mark Markdown
Lexical version: 0.31.2

## Steps To Reproduce

2. Mark in the sentence `Make "sure to check "out the various[...]` (only mark the Text in double quotes and the trailing whitespace is important) and format it bold by pressing "B" Symbol
3. The trailing whitespace get's converted to HTML representation `&#32;` and behind the result will become `Make **sure to check&#32;**out the various[...]`
4. That is invalid common mark (and also gfm) and would other editors to print the surrounding "**" before and after.
5. Please compare to this playground https://spec.commonmark.org/dingus/?text=test%20**test**%20test%5C%0Atest%20**test%20**%20test%5C%0Atest%20**test%20**test%5C%0Atest%20**test%26%2332%3B**test%5C%0Atest%20**test%26%2332%3B**%20test

Link to code example:

## The current behavior
HTML space `&#32;` is added, but space to following word is removed - which breaks Common Mark Syntax

## The expected behavior
Any valid Common Mark dealing the trailing whitespace, see Step 5 in reproduction, to get a idea, what it could be (maybe examples are incomplete and there is much more)

## Impact of fix

Happens always if within a text some words are marked to be formatted and makes it impossible to process the resulting Common Mark properly with other tools.

reinseth:
Leading and trailing whitespaces inside formatted text are already moved outside the formatted region to ensure correct markdown. But if the formatted text only contain whitespace, then we end up with formatting for the empty string, e.g. `**** ` for a bold whitespace. So we need to skip the formatting in these cases.

The previous attempt at fixing this in #7400 used html entities to escape leading/trailing whitespaces inside formatted texts, but this breaks commonmark.

Example that breaks when using html entities:
https://spec.commonmark.org/dingus/?text=bold%20space%20with%20adjacent%20text%3A%20**%26%2332%3B**adjacent%0A%0A

Reverts #7400
Closes #7575

## Test plan

Automated test in `LexicalMarkdown.test.ts`

### Before

Skjermbilde 2025-11-08 kl  23 46 19

Skjermbilde 2025-11-08 kl  23 46 30

### After

Skjermbilde 2025-11-08 kl  23 48 21

Skjermbilde 2025-11-08 kl  23 48 48

### DIFF
--- a/packages/lexical-markdown/src/importTextTransformers.ts
+++ b/packages/lexical-markdown/src/importTextTransformers.ts
@@ -135,6 +135,10 @@ export function importTextTransformers(
 
   // Handle escape characters
   const textContent = textNode.getTextContent();
-  const escapedText = textContent.replace(/\\([*_`~\\])/g, '$1');
+  const escapedText = textContent
+    .replace(/\\([*_`~\\])/g, '$1')
+    .replace(/&#(\d+);/g, (_, codePoint) => {
+      return String.fromCodePoint(codePoint);
+    });
   textNode.setTextContent(escapedText);
 }


reinseth:
Just to be clear: this PR fixes the original issue reported in #6526. It has no tests for the issues reported in #7575 since the code that caused them has been reverted.

etrepum:
I think this may need a more targeted approach than a revert, the escaped entities do work in many cases, for example, here are the removed tests and how github renders them (all as expected)

```
**Hello&#32;~~world~~**!

**~~Hello&#32;*world*~~**~~!~~

*Hello&#32;**world**!*

Text **boldstart&#32;[text](https://lexical.dev)&#32;boldend** text

Text **boldstart&#32;[`text`](https://lexical.dev)&#32;boldend** text

It ***~~works&#32;[with links](https://lexical.io)~~*** too

It ***~~works&#32;[with links](https://lexical.io)&#32;too~~***!

**&#160;**
```

**Hello&#32;~~world~~**!

**~~Hello&#32;*world*~~**~~!~~

*Hello&#32;**world**!*

Text **boldstart&#32;[text](https://lexical.dev)&#32;boldend** text

Text **boldstart&#32;[`text`](https://lexical.dev)&#32;boldend** text

It ***~~works&#32;[with links](https://lexical.io)~~*** too

It ***~~works&#32;[with links](https://lexical.io)&#32;too~~***!

**&#160;**

reinseth:
That's unrelated to the original bug report in #6526 (_Styling should be removed from the whitespace_). You should consider doing that in a separate change to prevent mixing concerns.

reinseth:
As a sidenote:
When creating examples like this, you should also add variants with adjacent text:

```
**&#160;**
**&#160;**adjacent
```

**&#160;**
**&#160;**adjacent

As you can see, the last one breaks. Escaping leading/trailing/single whitespace/newline as an html entity is a hack that only partially work.

etrepum:
I didn’t create the examples, they are from the test suite that’s being reverted in this PR. That’s why I’m suggesting that this should be a more targeted fix to specifically test and address the scenarios that don’t work.

reinseth:
I don't follow you. #7400 only adds a single test case (the newline case) and then augments other cases with verification that whitespaces are escaped on export. And we don't want the whitespaces escaped.

https://github.com/facebook/lexical/pull/7400/files#diff-9a0c91bc82f2693a031ebe1165316e9726980b7f1d54de8681de1256e1d18303

reinseth:
@etrepum: checking in. How would you like to proceed with this? I didn't quite follow what changes from the previous PR that you would like to retain. Maybe those aspects are properties of the markdown plugin that should be expressed in a different PR / feature? (considering that the original bug was that the markdown export produced `****` on bolded whitespace.)

aleksandrsjevdokimovswv:
As a Lexical user/consumer on a large-scale platform, I wanted to share a production perspective.

Ever since the change in https://github.com/facebook/lexical/pull/7400, we had to revert to v0.28.0 of Lexical and are seriously considering patching the HTML-entity changes back to their original state if we upgrade again. This was due to restricted portability and rendering issues in Markdown workflows, which affected most of our Markdown to HTML parsers in our platform.

This PR restores the expected behaviour, aligns with CommonMark, and feels like a much safer and more robust solution in the long run.

Even with the reverted changes, we currently have to regex out `****` cases to prevent incorrect parsing in Parsedown (MD → HTML).

etrepum:
The blocker here is that because it's changing tests that aren't problematic (listed above in comments) it needs a more thorough review that I simply have not had time for, and won't have time to do for at least another week. If it was a targeted fix that only affected serialization for incorrect markdown then it would be a quicker review. If another maintainer is interested they are welcome to participate here.

reinseth:
Ok, good to know. You could review commit by commit, where the [first commit](https://github.com/facebook/lexical/pull/7979/commits/d663537859239c136d46e089e84a40b6f82e366f) is the revert, and the [second commit](https://github.com/facebook/lexical/pull/7979/commits/21d87d6ebbc0dbb14e7229f6a7a7b42c081ceedd) is the actual change, which is a small and targeted change.

etrepum:
It doesn’t really work like that, since the commit you’re reverting is released all of the changes have to be considered

aleksandrsjevdokimovswv:
Hi, I'd like to see if this PR is going anywhere or it's in some sort of a "limbo" state?

reinseth:
@etrepum: checking back in. If you don't have the time for this, could you check with some of the other maintainers if they're interested in taking over this?

etrepum:
I can take another look at this one if the conflicts are resolved

reinseth:
@etrepum: I've rebased my branch onto `main` and resolved all conflicts caused by 51a83a75f168ffd1cd211ef15c5b144f3cb96242 and f1e4f66014377b1f2595aec2b0ee17f5b7ef4dfc.

etrepum:
Unclear what the root cause is at first glance but it looks like one of the markdown e2e tests is failing in a strange way (at least after the normalization that assertHTML does) https://github.com/facebook/lexical/actions/runs/22301277010/job/64587995265?pr=7979#step:11:285

There shouldn't really ever be an empty span produced by a TextNode since those are normalized away by the reconciler

```
1) [chromium] › packages/lexical-playground/__tests__/e2e/Markdown.spec.mjs:815:3 › Markdown › can convert markdown text into rich text 

    Error: innerHTML of contenteditable in page did not match

    - Expected  - 0
    + Received  + 1

    @@ -49,10 +49,11 @@
        <strong
          class="PlaygroundEditorTheme__textBold PlaygroundEditorTheme__textItalic PlaygroundEditorTheme__textStrikethrough"
          data-lexical-text="true">
          works
        </strong>
    +   <span data-lexical-text="true"></span>
        <a class="PlaygroundEditorTheme__link" href="https://lexical.io/">
          <strong
            class="PlaygroundEditorTheme__textBold PlaygroundEditorTheme__textItalic PlaygroundEditorTheme__textStrikethrough"
            data-lexical-text="true">
            with links
```

reinseth:
I'll investigate why this happens. I'll report back if I need clarification of any of the existing functionality.

reinseth:
Did some testing myself, and the e2e-tests seems to be flaky. Got the same results on the first run, but all passed on the second run.

I'll merge the branch shortly 👍

reinseth:
I've tracked down source of the issue: 51a83a75f168ffd1cd211ef15c5b144f3cb96242
I cannot say anything as to _why_ it happens. The change in the noted commit is pretty extensive. So any help here will be appreciated.

etrepum:
I’m not seeing the failure in other PRs, so I don’t think that commit is causing the problem on its own. Not sure what the problem is here without really diving in

reinseth:
No, the commit doesn't introduce any problems in itself, but the introduced functionality is somehow in conflict with the changes in this PR.

packages/lexical-markdown/src/importTextTransformers.ts
```diff
@@ -135,10 +135,6 @@ export function importTextTransformers(
 
   // Handle escape characters
   const textContent = textNode.getTextContent();
-  const escapedText = textContent
-    .replace(/\\([*_`~\\])/g, '$1')
-    .replace(/&#(\d+);/g, (_, codePoint) => {
-      return String.fromCodePoint(codePoint);
-    });
+  const escapedText = textContent.replace(/\\([*_`~\\])/g, '$1');
```
etrepum: Not the cause of the current failure but this change may make it parse previously generated markdown incorrectly

reinseth: That's a good point. I'll restore this.

reinseth:
I've reproduced the case from the e2e test in the unit test. I'll dig into this deeper and see if I can find a solution.

reinseth:
After more investigation, the behaviour observed by the e2e-test is correct. The e2e test does the following

- pastes the markdown including `It ~~___works [with links](https://lexical.io)___~~ too` in a markdown code block
- converts to html
- converts back to markdown
- converts back to html againg

What happens is that since we no longer includes leading/trailing whitespace inside formatted text on export, and the fact that an unrelated change moves formatting of links _inside_ the link text, the exported markdown becomes `It ***~~works~~*** [***~~with links~~***](https://lexical.io) too`. When converting back to html we correctly get an empty span between `works` and the link.

Here are screenshots from the sequence:

Pasted the markdown inside a markdown block:
Skjermbilde 2026-02-26 kl  09 48 57

Converted to html:
Skjermbilde 2026-02-26 kl  09 49 05

Converted back to markdown:
Skjermbilde 2026-02-26 kl  09 49 15

Converted back to html again:
Skjermbilde 2026-02-26 kl  09 49 21

etrepum:
Pretty sure that #8170 has enough overlap with this to resolve the same issue, that PR has more tests

reinseth:
Yes, #8170 fixes the issue that this PR also fixes. Great!
It's a shame, though, that it took 4 months and many hours on my part for this.

etrepum:
Thank you for your efforts, sometimes these things happen. The work you put in to investigating this problem and proposing a fix was certainly helpful especially on the review side. The sticking point here was that compatibility with lexical's output had to be considered since #7400 had made it into a release.

PR #7981, opened 2025-11-10
Chore: Update dependencies to resolve vulnerabilities

sherylkong18:
## Description

Update the versions of several packages to fix vulnerabilities

vite -> 7.1.11
`npm audit fix` for root, extension-react-table, extension-sveltekit-ssr-hydration
`npm i astro@5.14.3` for scripts/tests/integration/fixtures/lexical-esm-astro-react

Closes #

### Before

### After


etrepum:
It looks like one of the upgrades here is consistently causing test failures, hard to say which because there are several. I have noticed that updating playwright does break many of the tests due to some timing changes, so maybe something like that is happening and the tests will need to be fixed in order to upgrade.

PR #7983, opened 2025-11-11
[lexical-table] Feature: add config for opting in to nested tables

james-atticus:
## Description

#7192 made changes to try and prevent nested tables. They weren't removed from the document tree however, and could still be inserted when not using commands. @zurfyx also advised that nested tables are wanted at Meta, even if they're not fully supported (we also want to be able to use them).

This PR adds a config option to the plugin/extension allowing developers to opt in to allowing nested tables.

Note: there are no fixes/improvements for nested tables included (eg: #7154 is still an issue). The config option is marked `@experimental` to convey that nested tables are not properly supported.

### DIFF
--- a/packages/lexical-react/src/LexicalTablePlugin.ts
+++ b/packages/lexical-react/src/LexicalTablePlugin.ts
@@ -8,10 +8,10 @@
 
 import type {JSX} from 'react';
 
+import {signal} from '@lexical/extension';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {
   $isScrollableTablesActive,
-  registerPreventNestedTablesHandlers,
   registerTableCellUnmergeTransform,
   registerTablePlugin,
   registerTableSelectionObserver,
@@ -19,7 +19,7 @@ import {
   TableCellNode,
   TableNode,
 } from '@lexical/table';
-import {useEffect} from 'react';
+import {useEffect, useMemo} from 'react';
 
 export interface TablePluginProps {
   /**
@@ -72,7 +72,15 @@ export function TablePlugin({
     }
   }, [editor, hasHorizontalScroll]);
 
-  useEffect(() => registerTablePlugin(editor), [editor]);
+  const hasNestedTablesSignal = useMemo(() => signal(false), []);
+  if (hasNestedTablesSignal.peek() !== hasNestedTables) {
+    hasNestedTablesSignal.value = hasNestedTables;
+  }
+
+  useEffect(
+    () => registerTablePlugin(editor, {hasNestedTables: hasNestedTablesSignal}),
+    [editor, hasNestedTablesSignal],
+  );
 
   useEffect(
     () => registerTableSelectionObserver(editor, hasTabHandler),
@@ -97,13 +105,6 @@ export function TablePlugin({
       }
     });
   }, [editor, hasCellBackgroundColor, hasCellMerge]);
-
-  // Prevent nested tables being inserted when the feature isn't enabled
-  useEffect(() => {
-    if (!hasNestedTables) {
-      return registerPreventNestedTablesHandlers(editor);
-    }
-  }, [editor, hasNestedTables]);
 
   return null;
 }
--- a/packages/lexical-table/src/LexicalTablePluginHelpers.ts
+++ b/packages/lexical-table/src/LexicalTablePluginHelpers.ts
@@ -6,6 +6,7 @@
  *
  */
 
+import {Signal, signal} from '@lexical/extension';
 import {
   $findMatchingParent,
   $insertFirst,
@@ -23,7 +24,6 @@ import {
   $isTextNode,
   CLICK_COMMAND,
   COMMAND_PRIORITY_EDITOR,
-  COMMAND_PRIORITY_LOW,
   ElementNode,
   isDOMNode,
   LexicalEditor,
@@ -57,13 +57,17 @@ import {
   $getNodeTriplet,
 } from './LexicalTableUtils';
 
-function $insertTableCommandListener({
-  rows,
-  columns,
-  includeHeaders,
-}: InsertTableCommandPayload): boolean {
+function $insertTable(
+  {rows, columns, includeHeaders}: InsertTableCommandPayload,
+  hasNestedTables: boolean,
+): boolean {
   const selection = $getSelection() || $getPreviousSelection();
   if (!selection || !$isRangeSelection(selection)) {
+    return false;
+  }
+
+  // Prevent nested tables by checking if we're already inside a table
+  if (!hasNestedTables && $findTableNode(selection.anchor.getNode())) {
     return false;
   }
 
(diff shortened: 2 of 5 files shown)


packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -99,33 +124,37 @@ describe('Table', () => {
...
+    extension.output.hasNestedTables.value = true;
 
-    // Try to paste a table inside the cell
-    await editor.update(() => {
-      const tableNode = $createTableNode();
-      const selection = $getSelection();
-      if (selection === null) {
-        throw new Error('Expected valid selection');
-      }
-      editor.dispatchCommand(SELECTION_INSERT_CLIPBOARD_NODES_COMMAND, {
```
james-atticus: This wasn't actually testing anything before: verified by removing the command handler and the test still passed. We need to instead use `$insertGeneratedNodes` which only progresses if all the handlers for `SELECTION_INSERT_CLIPBOARD_NODES_COMMAND` return false.

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -99,33 +124,37 @@ describe('Table', () => {
...
-      const row = $createTableRowNode();
-      const cell = $createTableCellNode();
-      row.append(cell);
-      table.append(row);
-      root.append(table);
-      cell.select();
-    });
+  it('Allows nested tables when hasNestedTables is true', async () => {
+    const extension = getExtensionDependencyFromEditor(editor, TableExtension);
+    extension.output.hasNestedTables.value = true;
```
james-atticus, line 123: @etrepum I couldn't find any examples in the codebase where extension config was being updated after-the-fact. Let me know if this isn't the right way to go about it.

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -140,7 +169,115 @@ describe('Table', () => {
...
+          const cell = $createTableCellNode();
+          const paragraph = $createParagraphNode();
+          cell.append(paragraph);
+          row.append(cell);
+          table.append(row);
+          root.append(table);
+          paragraph.select();
+        },
+        {discrete: true},
+      );
```
james-atticus, line 180: Switched from async to `discrete: true`, and added a paragraph inside the cell (otherwise `$insertGeneratedNodes` throws). Test is otherwise the same as it was before.

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -59,28 +78,34 @@ describe('Table', () => {
...
-      const root = $getRoot();
-      const table = $createTableNode();
-      const row = $createTableRowNode();
-      const cell = $createTableCellNode();
-      row.append(cell);
-      table.append(row);
-      root.append(table);
-      cell.select();
-    });
+  it('Prevents nested tables by default', async () => {
```
james-atticus, line 81: Github is struggling a bit with this diff. Recommend reviewing commits individually and with whitespace hidden.

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -228,6 +224,37 @@ export function registerTableCellUnmergeTransform(
   });
 }
 
+export function registerPreventNestedTablesHandlers(
```
james-atticus: This might be better named as just `registerPreventNestedTables`. Open to suggestions.

packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
```diff
@@ -268,9 +268,62 @@ test.describe.parallel('Tables', () => {
...
-    // Note: Tests for nested table navigation ("Can exit the first/last cell of a nested table into the parent table cell")
-    // have been removed since nested tables are no longer supported.
-    // See: https://github.com/facebook/lexical/issues/7154
+    test(`Can exit the first cell of a nested table into the parent table cell`, async ({
+      page,
+      isPlainText,
+      isCollab,
+    }) => {
+      test.skip(isPlainText);
+      await initialize({hasNestedTables: true, isCollab, page});
```
james-atticus, line 277: Same as the tests removed in #7192 but with `hasNestedTables: true` added.

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
etrepum: When output properties are signals they're intended to be mutable after the fact, doing it this way is fine. It's a little awkward because this would normally just be configured up front, but it is a sort of partial test that the signal is wired up correctly.

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -90,5 +98,12 @@ export function TablePlugin({
     });
   }, [editor, hasCellBackgroundColor, hasCellMerge]);
 
+  // Prevent nested tables being inserted when the feature isn't enabled
+  useEffect(() => {
+    if (!hasNestedTables) {
+      return registerPreventNestedTablesHandlers(editor);
+    }
+  }, [editor, hasNestedTables]);
+
```
etrepum: This might be a good opportunity to just stop adding features to the react plugin and only support it in the extension

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -58,4 +77,207 @@ describe('Table', () => {
...
+        });
+      },
+      {discrete: true},
+    );
+
+    // Verify no nested table was created
+    editor.getEditorState().read(() => {
+      const root = $getRoot();
+      const table = root.getFirstChild();
+      if (!$isTableNode(table)) {
```
etrepum: not that we are using vitest you can use its `assert` to do this sort of thing in a more concise way

packages/lexical-table/src/index.ts
```diff
@@ -31,6 +31,7 @@ export {
 export type {TableDOMCell} from './LexicalTableObserver';
 export {$getTableAndElementByKey, TableObserver} from './LexicalTableObserver';
 export {
+  registerPreventNestedTablesHandlers,
```
etrepum: If we only support this feature in the extension we could not export this function at all and then its name wouldn't matter

packages/lexical-table/src/LexicalTablePluginHelpers.ts
etrepum: For the INSERT_TABLE_COMMAND handler this could also be implemented by passing the signal through to the `insertTableCommandListener` so that it can `peek()` the current value, which would prevent the need for registering a separate command listener at a higher priority to check this condition

packages/lexical-react/src/LexicalTablePlugin.ts
james-atticus: I took a stab at this but I think it'd require a large refactor of the playground App/Editor components. We'd have to replace TablePlugin with TableExtension, and I wasn't sure how to properly update the extension's config based on the values from `useSettings`.

I've switched to using signals for `registerTablePlugin` though, which avoids having to add a new `registerPreventNestedTablesHandlers` function.

etrepum: Basically the same way you did in the test should work, from a useEffect to control the timing. Not a big deal either way

etrepum (approved):

PR #7984, opened 2025-11-13
[lexical-react] [lexical-playground] Remove old ContextMenu, consolidate LexicalMenu render

ivailop7:
## Breaking Changes

Removes deprecated `LexicalContextMenu` API from `@lexical/react` for the `LexicalNodeContextMenu` API introduced in v0.32.0 (#7509)

## Before

```typescript
 <LexicalContextMenuPlugin
      options={options}
      onSelectOption={onSelectOption}
      onWillOpen={onWillOpen}
      menuRenderFn={menuRenderFn}
 />
```

## Now

```typescript
<NodeContextMenuPlugin
      className="PlaygroundEditorTheme__contextMenu"
      itemClassName="PlaygroundEditorTheme__contextMenuItem"
      separatorClassName="PlaygroundEditorTheme__contextMenuSeparator"
      items={items}
/>
```

A complete usage example is available [here](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx)

## Description

* Remove deprecated LexicalContextMenu, LexicalNodeContextMenu should be used. The example in the playground has been using it for quite some time now.
* Consolidate rendering of all inline popovers using 'menuRenderFn' and remove all associated instances of the custom rendering to use the same styles as the typeahead one.
* Remove unused PUNCTUATION variable

### DIFF
--- a/packages/lexical-playground/src/plugins/MentionsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/MentionsPlugin/index.tsx
@@ -563,7 +563,6 @@ function getPossibleQueryMatch(text: string): MenuTextMatch | null {
 class MentionTypeaheadOption extends MenuOption {
   name: string;
   picture: JSX.Element;
-  title: JSX.Element;
 
   constructor(name: string, picture: JSX.Element) {
     super(name);
--- a/packages/lexical-react/src/shared/LexicalMenu.tsx
+++ b/packages/lexical-react/src/shared/LexicalMenu.tsx
@@ -45,7 +45,7 @@ export class MenuOption {
   key: string;
   ref?: RefObject<HTMLElement | null>;
   icon?: JSX.Element;
-  title?: string | null;
+  title?: JSX.Element | string | null;
 
   constructor(key: string) {
     this.key = key;


etrepum:
I fixed up the description but it might be a good idea to write up a quick example of what migration looks like in case anyone is depending on the deprecated API. We also don't have API docs for the old or new APIs.

packages/lexical-react/flow/LexicalAutoEmbedPlugin.js.flow
```diff
@@ -43,7 +42,6 @@ type LexicalAutoEmbedPluginProps<TEmbedConfig> = {
     embedFn: () => void,
     dismissFn: () => void,
   ) => Array<AutoEmbedOption>,
-  menuRenderFn: MenuRenderFn<AutoEmbedOption>,
```
zurfyx, line 46: This is a breaking change (React package), was this intentional? Do we still have an option to customize the menu now? @ivailop7

ivailop7: it's using the typeahead menu styles for all LexicalMenu-based menus, excluding the ContextMenu-one those are definable in the theme object.

PR #7995, opened 2025-11-17
[lexical-react]: Add LexicalMenu styling to theme

ivailop7:
Since removing the menuRenderFn from the LexicalMenu component to allow for refactoring of the rendering and positioning, removed some styling flexibility. This PR adds the following theme classes for the LexicalMenu in the react package:

```typescript
  menu?: {
    container?: EditorThemeClassName;
    item?: EditorThemeClassName;
    itemSelected?: EditorThemeClassName;
    itemText?: EditorThemeClassName;
  };
```
this should provide enough flexibility to style completely the LexicalMenu using pure CSS.

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -777,7 +777,7 @@ test.describe.parallel('Tables', () => {
 
       await waitForSelector(
         page,
-        `div[class="typeahead-popover mentions-menu"] ul li:first-child.selected`,
+        `div[class="typeahead-popover"] ul li:first-child.PlaygroundEditorTheme__menuItemSelected`,
       );
 
       await moveDown(page, 1);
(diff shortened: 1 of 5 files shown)


ivailop7:
> Thanks for looking at this. Adding theming is good but doesn't quite meet the need where you have different styling for menus within the same editor.
> 
> As an example, we have a mentions typeahead menu and a text shortcuts typeahead menu with different styling. Currently we do this with `menuRenderFn` that constructs a container with a different class name on each. Eg:
> 
> ```css
> .MentionsPlugin_typeahead-menu {
>     width: 320px;
>     max-height: 300px;
>     overflow-y: auto;
> }
> 
> .MentionsPlugin_typeahead-menu .item {
>     color: var(--text-dark);
>     font-size: var(--font-medium);
>     font-weight: bold;
> }
> ```
> 
> I've pushed another [commit here](https://github.com/facebook/lexical/commit/8177474d5c776ca7decf4612b57722f649e4d938) on top of your change which adds a prop that would allow us to do this (doesn't look like I can open an PR into your fork).

thanks, gave you edit to my fork, you should be able to push to the branch.

etrepum:
Are we abandoning this?

ivailop7:
Seems so, for now. Closing.

PR #7997, opened 2025-11-18
Revert "[lexical-react] [lexical-playground] Remove old ContextMenu, consolidate LexicalMenu render"

ivailop7:
Reverts facebook/lexical#7984


james-atticus:
@zurfyx we'd be fine with #7995; we're not doing anything with our menus that can't be replicated with CSS for the container/div/ul/li elements.

PR #8001, opened 2025-11-21
[lexical-react] Annotate @deprecated to menuRenderFn  with NodeContext…

thatmichael85:
# What 
Annotate menuRenderFn with @deprecated and point to the link in the playground after https://github.com/facebook/lexical/pull/7997

# Why
To alert users that this prop is deprecated and to migrate to LexicalNodeContextMenu apis. 

The annotation at packages/lexical-react/flow/LexicalContextMenuPlugin.js.flow isn't too helpful

e.g
image

### DIFF
--- a/packages/lexical-extension/src/TabIndentationExtension.ts
+++ b/packages/lexical-extension/src/TabIndentationExtension.ts
@@ -6,9 +6,17 @@
  *
  */
 
-import type {LexicalCommand, LexicalEditor, RangeSelection} from 'lexical';
+import type {
+  ElementNode,
+  LexicalCommand,
+  LexicalEditor,
+  RangeSelection,
+} from 'lexical';
 
-import {$getNearestBlockElementAncestorOrThrow} from '@lexical/utils';
+import {
+  $getNearestBlockElementAncestorOrThrow,
+  $handleIndentAndOutdent,
+} from '@lexical/utils';
 import {
   $createRangeSelection,
   $getSelection,
@@ -59,9 +67,18 @@ function $indentOverTab(selection: RangeSelection): boolean {
   return false;
 }
 
+export type CanIndentPredicate = (node: ElementNode) => boolean;
+
+function $defaultCanIndent(node: ElementNode) {
+  return node.canBeEmpty();
+}
+
 export function registerTabIndentation(
   editor: LexicalEditor,
   maxIndent?: number | ReadonlySignal<null | number>,
+  $canIndent:
+    | CanIndentPredicate
+    | ReadonlySignal<CanIndentPredicate> = $defaultCanIndent,
 ) {
   return mergeRegister(
     editor.registerCommand<KeyboardEvent>(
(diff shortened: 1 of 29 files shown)


packages/lexical-react/src/LexicalContextMenuPlugin.tsx
```diff
@@ -45,6 +45,10 @@ export type LexicalContextMenuPluginProps<TOption extends MenuOption> = {
   onClose?: () => void;
   onWillOpen?: (event: MouseEvent) => void;
   onOpen?: (resolution: MenuResolution) => void;
+  /**
+   * @deprecated Use LexicalNodeContextMenuPlugin instead. Here is an example for using NodeContextMenuPlugin:
+   * https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/ContextMenuPlugin/index.tsx
+   */
   menuRenderFn: ContextMenuRenderFn<TOption>;
```
zurfyx, line 52: Besides marking it as deprecated, is it possible to restore the new API that @ivailop7 introduced so that we only get to this as a workaround for surfaces that are time-consuming to migrate? I see there's still a few playground callsites that are using this property already.

packages/lexical-react/flow/LexicalAutoEmbedPlugin.js.flow
```diff
@@ -43,6 +43,10 @@ type LexicalAutoEmbedPluginProps<TEmbedConfig> = {
     embedFn: () => void,
     dismissFn: () => void,
   ) => Array<AutoEmbedOption>,
+ /**
+ * @deprecated Use NodeContextMenuPlugin instead. Here is an example for using NodeContextMenuPlugin:
```
ivailop7: This is not correct. NodeContextMenuPlugin is not a replacement for AutoEmbedPlugin

packages/lexical-react/flow/LexicalNodeMenuPlugin.js.flow
```diff
@@ -55,6 +55,10 @@ type NodeMenuPluginProps<TOption> = {
   nodeKey: NodeKey | null,
   onClose?: () => void,
   onOpen?: (resolution: MenuResolution) => void,
+/**
+ * @deprecated Use NodeContextMenuPlugin instead. Here is an example for using NodeContextMenuPlugin:
```
ivailop7: This is also not correct. NodeContextMenuPlugin is a replacement only for ContextMenuPlugin, nothing else.

packages/lexical-react/flow/LexicalTypeaheadMenuPlugin.js.flow
```diff
@@ -66,6 +66,10 @@ export type TypeaheadMenuPluginProps<TOption> = {
     matchingString: string,
   ) => void,
   options: Array<TOption>,
+/**
```
ivailop7: Remove this one as well please.

packages/lexical-react/src/LexicalAutoEmbedPlugin.tsx
```diff
@@ -84,6 +84,10 @@ type LexicalAutoEmbedPluginProps<TEmbedConfig extends EmbedConfig> = {
     embedFn: () => void,
     dismissFn: () => void,
   ) => Array<AutoEmbedOption>;
+  /**
```
ivailop7: Incorrect as well.

packages/lexical-react/src/LexicalNodeMenuPlugin.tsx
```diff
@@ -34,6 +34,10 @@ export type NodeMenuPluginProps<TOption extends MenuOption> = {
   nodeKey: NodeKey | null;
   onClose?: () => void;
   onOpen?: (resolution: MenuResolution) => void;
+  /**
```
ivailop7: Not true

packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
```diff
@@ -202,6 +202,10 @@ export type TypeaheadMenuPluginProps<TOption extends MenuOption> = {
     matchingString: string,
   ) => void;
   options: Array<TOption>;
+  /**
```
ivailop7: Please remove this as well.

ivailop7 (commented):
A lot of incorrect deprecation comments in this PR. 
Putting this here to clarify:
LexicalNodeContextMenu is a replacement ONLY for LexicalContextMenu. The Playground example has been using the new one for 5-6 months+. So the LexicalContextMenu should be fully deletable in its current state.

No other deprecation of menuRenderFn in AutoEmbed, NodeMenu or Typeahead is happening, since they need to have their interface changed, to use the one I did in the now reverted PR, but maybe keep menuRenderFn alongside it for now, as an escape hatch.

PR #8003, opened 2025-11-23
[lexical-selection]: Fix applying styles for empty table cells

ivailop7:
This an attempt to fix: https://github.com/facebook/lexical/issues/6583

I'm not 100% sure if this is the right approach, but it works!

Before:

https://github.com/user-attachments/assets/fa32351b-b13a-4e81-8a07-ccb73b1fe8a9

After:

https://github.com/user-attachments/assets/9462aa18-ab76-4f38-a9a8-d5b067c82fb6


etrepum (approved):

etrepum:
This seems like a reasonable approach to me!

PR #8030, opened 2025-12-16
Bug Fix : fix cursor looping and incorrect paragraph insertion after tables

### LINKED ISSUE #7999: Bug: Odd selection behavior with table as last element in document
Lexical version: 0.38.2

## Steps To Reproduce

1. Insert a table and delete anything after it.
2. Click to the right of the table.
3. Press right arrow key repeatedly.
 

## The current behavior

When you click to the right of the table the cursor briefly shows next to the table before moving into the last table cell.

The cursor goes in a loop consisting of 
1. The last cell of the table.
2. A block cursor beneath the table.
3. The right hand side of the table.

Additionally, pressing enter when the cursor is on the right hand side of the table inserts a newline above the table rather than beneath it.

## The expected behavior

I'd expect the cursor to go beneath the table when pressing the right arrow from the last cell. And then to stay there rather than cycle back up.

## Impact of fix

Mostly polish, but it can be hard to figure out how to add content after the table. Had a user tell me it couldn't be done.

Jynx2004:
## Description

This PR fixes an issue where cursor navigation becomes inconsistent when a table is the final block in the document.

Fixes #7999

## Problem

When a table is the last node and all content after it is deleted:

- Clicking to the right of the table briefly places the cursor there, then moves it back into the last table cell
- Repeatedly pressing the right arrow key causes the cursor to loop between:
  - The last table cell
  - A block cursor beneath the table
  - The right-hand side of the table
- Pressing **Enter** while the cursor appears on the right-hand side of the table inserts a new paragraph **above** the table instead of below it

This makes it difficult for users to understand how to add content after a table.

## Cause

When there is no text in the direction of navigation, Lexical attempts to explore adjacent nodes. If the previous node is a table, this logic re-enters table selection and causes the cursor to cycle back into the table.

## Fix

Short-circuit navigation/deletion logic when:

- There is no text in the navigation direction, and
- The immediately preceding node is a table

This prevents Lexical from re-entering table selection logic and allows the cursor to correctly move and remain below the table.

## Result

- Cursor moves beneath the table when navigating right from the last cell
- Cursor no longer cycles back into the table
- Pressing **Enter** inserts a paragraph **after** the table
- Matches expected behavior of common rich-text editors

## Impact

Primarily a polish improvement, but significantly improves usability and prevents user confusion when editing content after tables.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,162 @@
+# Lexical Agent Guide
+
+This file provides detailed guidance for AI agents and automated tools working with the Lexical codebase.
+
+## Build, Test, and Development Commands
+
+### Building
+- `npm run build` - Build all packages in development mode
+- `npm run build-prod` - Clean and build all packages in production mode
+- `npm run build-release` - Build production release with error codes
+- `npm run build-types` - Build TypeScript type definitions and validate them
+
+### Testing
+- `npm run test-unit` - Run all unit tests (Vitest)
+- `npm run test-unit-watch` - Run unit tests in watch mode
+- `npm run test-e2e-chromium` - Run E2E tests in Chromium (requires dev server running)
+- `npm run test-e2e-firefox` - Run E2E tests in Firefox
+- `npm run test-e2e-webkit` - Run E2E tests in WebKit
+- `npm run debug-test-e2e-chromium` - Run E2E tests in debug mode (headed)
+- `npm run debug-test-unit` - Debug unit tests with inspector
+
+For E2E testing workflow:
+1. Start the dev server: `npm run start` (or `npm run dev` if you don't need collab)
+2. In another terminal: `npm run test-e2e-chromium`
+
+### Development Servers
+- `npm run start` - Start playground dev server + collab server (http://localhost:3000)
+- `npm run dev` - Start only the playground dev server (no collab)
+- `npm run start:website` - Start Docusaurus website (http://localhost:3001)
+- `npm run collab` - Start collab server on localhost:1234
+
+### Code Quality
+- `npm run lint` - Run ESLint on all files
+- `npm run prettier` - Check code formatting
+- `npm run prettier:fix` - Auto-fix formatting issues
+- `npm run flow` - Run Flow type checker
+- `npm run tsc` - Run TypeScript compiler
+- `npm run ci-check` - Run all checks (TypeScript, Flow, Prettier, ESLint)
+
+## High-Level Architecture
+
+### Core Concepts
+
+Lexical is built around several key architectural concepts that work together:
+
+**Editor Instance** - Created via `createEditor()`, wires everything together. Manages the EditorState, registers listeners/commands/transforms, and handles DOM reconciliation.
+
+**EditorState** - Immutable data model representing the editor content. Contains:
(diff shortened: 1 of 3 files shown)


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1747,12 +1747,17 @@ export class RangeSelection implements BaseSelection {
       const direction = isBackward ? 'previous' : 'next';
       const initialCaret = $caretFromPoint(anchor, direction);
       const initialRange = $extendCaretToRange(initialCaret);
+      const elementBefore =
+        anchorNode.__prev && $getNodeByKey(anchorNode.__prev);
```
etrepum: This is better written as `anchorNode.getPreviousSibling()`

packages/lexical/src/LexicalSelection.ts
```diff
@@ -1747,12 +1747,17 @@ export class RangeSelection implements BaseSelection {
...
       if (
         initialRange
           .getTextSlices()
           .every((slice) => slice === null || slice.distance === 0)
       ) {
         // There's no text in the direction of the deletion so we can explore our options
+        if (elementBefore && elementBefore.__type === 'table') {
+          return;
+        }
```
etrepum: This is not an acceptable solution because LexicalSelection shouldn't know about tables and the type of a table could be any string (in situations where a node override is used to replace TableNode with a subclass)

etrepum (commented):
This isn't a complete workaround and it does not add any tests to show that it works as expected

packages/lexical/src/LexicalSelection.ts
Jynx2004: How can we identify the paragraph node which is just after the table in hierarchy other than the above method ?
If we recognize it then we can prevent deletion of it which resolves the bug.

etrepum: Usually this would be done by implementing the functionality in one of the table plugins.

packages/lexical/src/LexicalSelection.ts
Jynx2004: In my latest merge request i have used  `anchorNode.getPreviousSibling()` instead of `anchorNode.__prev  and $getNodeByKey(anchorNode.__prev)`

packages/lexical/src/LexicalSelection.ts
Jynx2004: Will work on improving `elementBefore && elementBefore.__type === 'table'`

Jynx2004: The function `$TableNode` checks whether a given node is an instance of class TableNode but we cannot use it in LexicalSelection .

Jynx2004: > Usually this would be done by implementing the functionality in one of the table plugins.

Issue lies in the paragraphNode which is inserted below the table when the table is created . When we delete the paragraph then the cursor gets stuck in a loop when we click the right arrow. So we need to prevent the deletion of that node so that the bug gets resolved.

etrepum: Right, this workaround should not be in this file at all.

PR #8031, opened 2025-12-16
Agent documentation

zurfyx:
We may want to start intentionally incorporating LLMs into the repo for the sake of development velocity. This can be experimental for now but it has played out very well internally so I'm optimistic.

The idea is to be docs first and Claude / others having an easy time identifying the relevant docs. To make this more effective we may also want to be more strict when it comes to documentation for new features, and complex developer pieces in the core, just like we enforce a test per PR now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,14 +98,14 @@ npm run test-e2e-chromium
 npm run ci-check
 ```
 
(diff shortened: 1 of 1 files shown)


etrepum (approved):
I wonder if it’s better or worse that our rest API docs are only in docusaurus. Serving agent-appropriate docs seems like functionality it should have

PR #8035, opened 2025-12-17
npm -> pnpm

zurfyx:
We have made the decision to migrate to PNPM. We believe Lexical can benefit from its security and performance.

This PR encapsulates the minimum viables changes to get there, which boil down to:
* Replacing `npm` for `pnpm` and adjusting the syntax
* Declaring _phantom_ dependencies
* Small tweaks to the CI (including an upgrade to Node 22 LTS)

Test plan:
- [x] Compare build side-by-side
- [x] Compare (Meta) WWW build side-by-side
- [x] Meta's PNPM internal deployment (D89383966)
- [x] CI passes
- [ ] NPM publish (let's merge this PR first)
- [x] Updating CONTRIBUTING.md
- [ ] Create new package (will be followed up separately after `docs/`)

### DIFF
--- a/scripts/override-react.js
+++ b/scripts/override-react.js
@@ -48,7 +48,7 @@ async function main() {
   console.log(cmd);
   await exec(cmd);
   await exec(
-    `git checkout package-lock.json package.json packages/*/package.json`,
+    `git checkout pnpm-lock.yaml package.json packages/*/package.json`,
   );
 }
 


.pnpmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
etrepum: I don't think this is a config option at all in pnpm

package.json
```diff
@@ -8,33 +8,33 @@
...
     "clean": "node scripts/clean.js",
     "extract-codes": "node scripts/build.js --codes",
     "flow": "node ./scripts/check-flow-types.js",
     "flow-typed-install": "flow-typed install -s -o --ignoreDeps dev peer bundled",
     "tsc": "tsc",
-    "tsc-extension": "npm run compile -w @lexical/devtools",
+    "tsc-extension": "pnpm --filter @lexical/devtools run compile",
     "tsc-watch": "tsc -w",
-    "collab": "cross-env HOST=localhost PORT=1234 npx y-websocket-server",
+    "collab": "cross-env HOST=localhost PORT=1234 npx y-websocket",
```
thatmichael85: oh is this expected? 
from y-websocket-server to y-websocket?

.github/workflows/call-core-tests.yml
```diff
@@ -3,12 +3,14 @@ name: Lexical Core Tests
...
 
+permissions: {}
+
 jobs:
   integrity:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [22.19.0]
+        node-version: [22.x]
```
etrepum, line 13: Is there a strong reason to keep 22.x instead of upgrading to the active lts 24.x?

examples/vanilla-js-plugin/package.json
```diff
@@ -10,16 +10,25 @@
...
+    "@lexical/history": "file:../../npm/lexical-history-0.39.0.tgz",
+    "@lexical/html": "file:../../npm/lexical-html-0.39.0.tgz",
+    "@lexical/list": "file:../../npm/lexical-list-0.39.0.tgz",
+    "@lexical/rich-text": "file:../../npm/lexical-rich-text-0.39.0.tgz",
+    "@lexical/selection": "file:../../npm/lexical-selection-0.39.0.tgz",
+    "@lexical/table": "file:../../npm/lexical-table-0.39.0.tgz",
+    "@lexical/utils": "file:../../npm/lexical-utils-0.39.0.tgz",
     "emoji-datasource-facebook": "15.1.2",
-    "lexical": "0.39.0"
+    "lexical": "file:../../npm/lexical-0.39.0.tgz"
```
etrepum: this looks like an accident

examples/extension-sveltekit-ssr-hydration/package.json
```diff
@@ -52,5 +52,8 @@
     "vite-plugin-devtools-json": "^1.0.0",
     "vitest": "^3.2.3",
     "vitest-browser-svelte": "^0.1.0"
+  },
+  "optionalDependencies": {
+    "@rollup/rollup-linux-x64-gnu": "4.6.1"
```
etrepum: I wonder if we might want to add other platforms to this or to include @rollup/wasm-node or something

etrepum (commented):
Haven't had a chance to look through all of it or try it locally yet but this is what I noticed with a quick look

.github/workflows/call-core-tests.yml
zurfyx: For some reason, 24.x and 25.x hang when running Playwright, which I why I intentionally put an `engine` restriction of `< 23.0`. I was thinking to look at it separately,  maybe upgrading the latest Playwright version will do but there's already a lot going on in this PR

scripts/__tests__/integration/utils.js
```diff
@@ -153,7 +149,9 @@ function describeExample(packageJsonPath, bodyFun = undefined) {
       test(
         'tests pass',
         async () => {
-          await withCwd(exampleDir, () => expectSuccessfulExec('npm run test'));
+          await withCwd(exampleDir, () =>
+            expectSuccessfulExec('pnpm run test'),
```
etrepum: This is a bit inconsistent, right now everything else with examples and integration tests is run with npm and uses package-lock.json. I think it happens to work because the node_modules is set up properly but it is a bit strange

scripts/override-react.js
```diff
@@ -44,11 +44,11 @@ async function main() {
       fs.writeJsonSync(fn, json, {spaces: 2});
     }
   });
-  const cmd = `npm i ${packages.map((pkg) => `${pkg}@${version}`).join(' ')}`;
+  const cmd = `pnpm add -w ${packages.map((pkg) => `${pkg}@${version}`).join(' ')}`;
   console.log(cmd);
   await exec(cmd);
   await exec(
-    `git checkout package-lock.json package.json packages/*/package.json`,
+    `git checkout pnpm-lock.yaml package.json packages/*/package.json`,
```
etrepum: This might not be working correctly because the examples are still using npm and package-lock.json?

CONTRIBUTING.md
```diff
@@ -7,8 +7,16 @@ possible.
 
 We actively welcome your pull requests.
 
+This project uses [pnpm](https://pnpm.io/) as its package manager. If you don't have it installed:
+
+```bash
+npm install -g pnpm
+# or if you have Node.js 16.13+
+corepack enable
```
etrepum: I think these instructions might be a bit out of date, the latest node versions have moved corepack out of the distribution so it might be something more like

npx corepack@latest enable

package.json
```diff
@@ -8,33 +8,33 @@
     "packages/*"
   ],
   "engines": {
-    "npm": ">=8.2.3",
-    "yarn": ">=1"
+    "node": ">=18.0.0 <23.0.0",
```
etrepum: Is there a reason to explicitly prevent newer node versions from being used? We're not doing anything that's forwards incompatible

etrepum:
I confirmed that installing and running the unit tests & e2e tests works as expected.

After a closer look I think it's probably worth auditing each `pnpm run` that uses `--` since those might not be correct anymore. Those exist because `npm run arg --foo` has `npm` parse the argument `--foo` but `pnpm run arg --foo` passes the argument to the script like `npm run arg -- --foo` which is what's wanted in 99.99% of the cases (and how npx behaves). It also gets really messy in npm when scripts invoke other scripts so you have multiple layers of `--`.

The solution is probably to just get rid of all the `--`.

scripts/__tests__/integration/utils.js
zurfyx: @etrepum This was actually intentional but I'm happy to do either way. My thought process is, documentation and examples are user-oriented, and folks are the most familiar with NPM. For developers in our repo, PNPM is preferable option. Would you agree with this / or do you think that these examples are more test oriented as opposed to user facing?

etrepum: I think either is fine but it should be consistent, if we’re installing example with npm we should be testing the examples with npm (and making sure to handle the correct lock files). I don’t think it’s a big deal either way, the examples should still work for users with any package manager, the only difference should be whether a lockfile is used or not.

zurfyx: Ah yes, you're completely right, sorry I misread the code pointer

thedjpetersen:
Neat - curious to learn more about the advantages of PNPM

etrepum:
The primary motivation for this PR is better security defaults. There are also some developer ergonomic improvements such as better script argument passing with `pnpm run` vs. `npm run` (no need for `--` escaping).

PR #8036, opened 2025-12-18
[lexical-link][Bug Fix] avoid auto-linking urls in code snippets

### LINKED ISSUE #8028: Bug: AutoLink produces links inside non-highlighted code-blocks
The AutoLinkPlugin (registerAutoLink underneath) does not take code blocks into account, and inserts AutoLinks in code blocks, even though this is not usually desired. This only happens if a highlighter plugin is not used. Can be easily reproduced in the Lexical playground.

Lexical version: 

## Steps To Reproduce

1. In Lexical playground, turn off code highlighting
2.Add a code block and type a link in the code block

Link to code example: https://playground.lexical.dev/?isCodeHighlighted=false

## The current behavior
Links are highlighted in code blocks

## The expected behavior
AutoLink should ignore links in code blocks

## Impact of fix

My current use case expects code blocks to look like code instead of having interactive links. For example, most links included are not meant to be opened by the user, such as in code examples of API calls. 
If the current behaviour is desired in some cases, an option could be added to disable highlighting in certain contexts such as code blocks.

xnt:
## Description

The AutoLinkPlugin (registerAutoLink underneath) does not take code blocks into account, and inserts AutoLinks in code blocks, even though this is not usually desired. This only happens if a highlighter plugin is not used. Can be easily reproduced in the Lexical playground.

This fix validates that the current block is not a code block before trying to insert auto-links. Also, adds a test with that check.

Closes #8028

### Before

image

### After

image

image

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
@@ -18,7 +18,6 @@ import {
 import {
   assertHTML,
   click,
-  expect,
   focusEditor,
   html,
   initialize,


etrepum (commented):
This introduces a dependency on @lexical/code from @lexical/link which breaks the build and is not desired. One way or another this issue should be addressed in a way that doesn't introduce this kind of dependency (note also that checking `getType()` is also not an acceptable solution because the type of a CodeNode can be any string if it's subclassed).

A cleaner possibility might be to add configuration to AutoLinkConfig to allow you to specify some predicates that control which nodes shouldn't be considered. The extensions could have a peer dependency to set this up automatically for CodeNode.

etrepum:
#8123 should be a sufficient fix for this issue

PR #8042, opened 2025-12-19
[lexical-playground]: Draggable block handle gliding effect

ivailop7:
Add a gliding effect to the draggable block handle and a fade-in and fade-out when the handle is shown/hidden.
To remove the 'fly-in' effect due to the current design where the handle is positioned by default at -10000,-10000, made a minor change to make the handle not visible vs positioned off screen when the mouse cursor is not in the textarea. I'm not sure why this was approach taken in a first place. I'm not aware of any accessibility or performance benefits to have it off screen vs hidden.

https://github.com/user-attachments/assets/190894ef-70bb-45bc-86d6-f7b24f824242


etrepum:
Not sure what is causing the deployment failures, could be something transient? Not sure how to trigger another build without another commit

ivailop7:
Not sure either, thought we blocked Vercel builds, but I see other PRs are fine. Will wait on the big pnpm PR to go through and then merge these 2.

zurfyx:
> Not sure either, thought we blocked Vercel builds, but I see other PRs are fine. Will wait on the big pnpm PR to go through and then merge these 2.

Thank you, pnpm PR is landed. We did block Vercel builds for some time, but at this point in time everything is fully functional.

etrepum (approved):
I'm not quite sure what's going on here, maybe some layout race condition, but the handle now gets stuck in offset positions after a drop (until it's moved to another element)

https://github.com/user-attachments/assets/89e80b55-58dd-48cd-9f19-ade62bbafe99

ivailop7:
Noted, that's weird. Will have a look in the coming days.

ivailop7:
Fixed the onDrop issue, it was also happening for the very first row when entering the textarea with the cursor. All good now. Merging.

packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.css
```diff
@@ -6,9 +6,12 @@
   position: absolute;
   left: 0;
   top: 0;
-  will-change: transform;
+  will-change: transform, opacity;
   display: flex;
   gap: 2px;
+  transition:
+    transform 140ms ease-in-out,
+    opacity 160ms ease-in-out;
```
alx-ppv, line 14: There's a transition for opacity, but the opacity is never 0. The code that sets the opacity to 0 is removed in LexicalDraggableBlockPlugin.tsx, if I understand correctly. Even in the playground I don't see the animation.

PR #8043, opened 2025-12-19
[lexical-playground]: Color table resize handle

ivailop7:
Highlight the table border under the mouse cursor to indicate better resizability of the table column. Having this only for the columns, not for the rows.

https://github.com/user-attachments/assets/a552096a-afea-4872-88ff-a4335535b058


etrepum:
This also had the deployment failures

PR #8049, opened 2025-12-23
[lexical-list] Bug Fix: fix selection issue from list transform on linebreak

### LINKED ISSUE #8046: Bug: Selection issue from list transform on multi-line paragraph
Lexical version: 0.39.0
## Steps To Reproduce

1. In a blank editor, enter a line of text, then insert a `linebreak` with `Cmd + Shift`
2. Transform the paragraph into a list (ordered or unordered)
3. Previously, the selection was on the `linebreak`, after transforming it into a list, the selection is on the list node itself. We also receive this console warning: `IndexSizeError: Failed to execute 'setBaseAndExtent' on 'Selection': There is no child at offset 2.`

Link to code example:
This can be reproduced on the Lexical playground. It doesn't make sense to create a codesandbox for this, as it needs to be reproduced step by step. Here's a screen recording of how to reproduce it.

https://github.com/user-attachments/assets/4bbd31ef-dd1a-47d8-af11-e63413430103

## The current behavior
After transforming a paragraph ending in a `linebreak` node into a list, we get the following console warning: `IndexSizeError: Failed to execute 'setBaseAndExtent' on 'Selection': There is no child at offset 2.`

Additionally, our selection is set to the list node, as opposed to the `linebreak` node where we left our selection state before the transformation.
## The expected behavior
Expected behavior is that we don't get the console warning, and our selection state is preserved, on the `linebreak` node.
## Impact of fix
- This bug affects a few live customers of an app using Lexical.
- If focus is preserved on list transformation, this issue is more noticeable. If the editor is refocused after the transformation, the cursor jumps from the position of the `linebreak` node, up to the beginning of the list.

dizsmek:
## Description

**Current behavior:**
When a paragraph ends with a `LineBreakNode` and the cursor is positioned at the end, 
Lexical creates an element-anchored selection with an offset pointing after the last child. 
When this paragraph is converted to a list, Lexical's selection reconciliation automatically 
maps the selection from the paragraph to the newly created ListNode, preserving the offset. 
However, the ListNode only has 1 child (the ListItemNode), so any offset greater than 1 
becomes invalid on the ListNode. 

This causes:
1. IndexSizeError in the console: "Failed to execute 'setBaseAndExtent' on 'Selection': 
   There is no child at offset X"
2. Incorrect cursor position - the selection state becomes corrupted

**New behavior:**
After a paragraph is replaced with a list structure, the fix checks if the selection 
was automatically mapped to the ListNode. If so, it redirects the selection to anchor 
to the ListItemNode instead, while preserving the original offset.

This works because the ListItemNode contains the same children as the original paragraph, 
making the offset valid on the ListItemNode even though it would be invalid on the ListNode.

Closes #8046 

### Before

https://github.com/user-attachments/assets/424ab626-29ca-4226-91da-dbc252961050

### After

https://github.com/user-attachments/assets/16d6b92c-446b-4357-8f4e-07f8bf5c9004

### DIFF
--- a/.github/workflows/call-npm-publish.yml
+++ b/.github/workflows/call-npm-publish.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: pnpm/action-setup@v4
         with:
           version: 9
-      - run: pnpm install --frozen-lockfile
+      - run: pnpm install
       - run: pnpm run prepare-release
       - run: node ./scripts/npm/release.js --non-interactive $DRY_RUN_ARG $IGNORE_PREVIOUSLY_PUBLISHED_ARG --channel='${{ inputs.channel }}'
         env:
--- a/.pnpmrc
+++ b/.pnpmrc
@@ -2,5 +2,3 @@
 # This is needed because test files use ESM (.mjs) which has issues
 # with pnpm's default symlinked structure
 node-linker=hoisted
-
-
--- a/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.css
+++ b/packages/lexical-playground/src/plugins/DraggableBlockPlugin/index.css
@@ -6,9 +6,12 @@
   position: absolute;
   left: 0;
   top: 0;
-  will-change: transform;
+  will-change: transform, opacity;
   display: flex;
   gap: 2px;
+  transition:
+    transform 140ms ease-in-out,
+    opacity 160ms ease-in-out;
 }
 
 .draggable-block-menu .icon {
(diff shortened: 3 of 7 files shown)


etrepum (commented):
Typically the getKey accessor is used instead of __key. It would also be a good idea to add a test to this PR to show that it fixes the issue.

dizsmek:
@etrepum added a test and replaced `__key`.

etrepum (approved):

PR #8052, opened 2025-12-25
[lexical-playground]: Draggable handle and dropdown CSS zoom fix

ivailop7:
Fix handle positioning when CSS zoom is applied. 

Before:

https://github.com/user-attachments/assets/2fcaf053-d1ef-4113-9685-8c7003e1601e

After: 

https://github.com/user-attachments/assets/2e0ac3c6-8252-4386-9874-62b572525680

Also fixed a similar issue for the toolbar dropdown menus

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 ## Features
 
-- **Framework Agnostic Core** - Works with any UI framework, with official [React bindings](https://lexical.dev/docs/react/getting-started)
+- **Framework Agnostic Core** - Works with any UI framework, with official [React bindings](https://lexical.dev/docs/getting-started/react)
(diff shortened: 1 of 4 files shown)


packages/lexical-utils/flow/LexicalUtils.js.flow
```diff
@@ -132,7 +132,7 @@ declare export function $splitNode(
   offset: number,
 ): [ElementNode | null, ElementNode];
 
-declare export function calculateZoomLevel(element: Element | null): number;
+declare export function calculateZoomLevel(element: Element | null, useManualZoom: boolean = false): number;
```
etrepum:
```suggestion
declare export function calculateZoomLevel(element: Element | null, useManualZoom?: boolean): number;
```

PR #8057, opened 2025-12-29
[lexical-playground]: New Table Hover Actions Plugin

ivailop7:
Paving the way for a better tables user experience, by replacing the current TableHoverActionsPlugin. Keeping the old one in the codebase for now, despite not being used, will add an issue to delete eventually. 

The new one reuses the existing FloatingUI dependency and allows to insert a row or a column at any position. This removes a number of issues with oversized tables and vertical scrolling.

The buttons only appear when the mouse cursor is in the top row or first column.

https://github.com/user-attachments/assets/00aa144b-b24f-4cec-9ece-6d21309eb61f


etrepum (approved):
This is a much nicer UX!

PR #8060, opened 2025-12-30
[lexical-playground]: Column Sort for Basic Table

ivailop7:
Add basic sorting to table columns. 

Behaviour:
On mouse hover over the top row, it shows a sort button that shows a menu for Ascending or Descending sorting of the column. Upon selection it sorts the rows in the table.It excludes the header row from the sort if such is present.
The sorting is no-op when the table contains any merged cells.

https://github.com/user-attachments/assets/8ef1750a-a20b-457a-bdf4-120a67c35809

### DIFF
--- a/packages/lexical-playground/src/plugins/TableHoverActionsV2Plugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableHoverActionsV2Plugin/index.tsx
@@ -23,15 +23,15 @@ import {
   $insertTableColumnAtSelection,
   $insertTableRowAtSelection,
   $isTableCellNode,
+  $isTableNode,
   $isTableRowNode,
-  TableNode,
-  type TableRowNode,
+  type TableNode,
 } from '@lexical/table';
 import {
   $getChildCaret,
   $getNearestNodeFromDOMNode,
   $getSiblingCaret,
-  EditorThemeClasses,
+  type EditorThemeClasses,
   isHTMLElement,
 } from 'lexical';
 import {useEffect, useRef, useState} from 'react';
@@ -54,17 +54,17 @@ const LEFT_BUTTON_OVERHANG = SIDE_INDICATOR_SIZE_PX / 2;
 function $isSimpleTable(table: TableNode): boolean {
   const rows = table.getChildren();
   let columns: null | number = null;
-  for (const row of rows as TableRowNode[]) {
+  for (const row of rows) {
     if (!$isTableRowNode(row)) {
       return false;
     }
     if (columns === null) {
-      columns = (row as TableRowNode).getChildrenSize();
+      columns = row.getChildrenSize();
     }
-    if ((row as TableRowNode).getChildrenSize() !== columns) {
+    if (row.getChildrenSize() !== columns) {
       return false;
     }
-    const cells = (row as TableRowNode).getChildren();
+    const cells = row.getChildren();
     for (const cell of cells) {
       if (
         !$isTableCellNode(cell) ||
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/TableHoverActionsV2Plugin/index.tsx
```diff
@@ -294,9 +306,66 @@ function TableHoverActionsV2({
...
+        : null;
+      const bodyRows: Array<TableRowNode> = rows
+        .slice(1)
+        .filter((row): row is TableRowNode => $isTableRowNode(row));
+
+      if (!headerRow || bodyRows.length === 0) {
+        return;
+      }
+
+      bodyRows.forEach((row) => row.remove());
```
etrepum: You shouldn't need to do this, moving them around in the document should do any necessary fix-ups

packages/lexical-playground/src/plugins/TableHoverActionsV2Plugin/index.tsx
```diff
@@ -294,9 +306,66 @@ function TableHoverActionsV2({
...
+        const bCell = b.getChildAtIndex(colIndex);
+        const aText = aCell ? aCell.getTextContent() : '';
+        const bText = bCell ? bCell.getTextContent() : '';
+        const result = aText.localeCompare(bText, undefined, {numeric: true});
+        return direction === 'asc' ? result : -result;
+      });
+
+      for (const bodyRow of bodyRows) {
+        headerRow.insertAfter(bodyRow);
+      }
```
etrepum: I think this is inserting them backwards, something like this should work to insert them in the same order as the array:

```ts
$getSiblingCaret(headerRow, 'next').splice(0, bodyRows);
```

packages/lexical-playground/src/plugins/TableHoverActionsV2Plugin/index.tsx
```diff
@@ -294,9 +306,66 @@ function TableHoverActionsV2({
...
+        return;
+      }
+
+      const colIndex = cellNode.getIndexWithinParent();
+      const rows = tableNode.getChildren();
+
+      if (rows.length <= 1) {
+        return;
+      }
+
```
etrepum: This could be a good place to do the merge cell check to just bail if there are any merges until the logic works correctly, something like:

```ts
function $isSimpleTable(table: TableNode): boolean {
  const rows = table.getChildren();
  let columns: undefined | number;
  for (const row of rows) {
    if ($isTableRowNode(row)) {
      return false;
    }
    if (columns === undefined) {
      columns = row.getChildrenSize();
    }
    if (row.getChildrenSize() !== columns) {
      return false;
    }
    const cells = row.getChildren();
    for (const cell of cells) {
      if (!$isTableCellNode(cell) || cell.getRowSpan() !== 1 || cell.getColSpan() !== 1) {
        return false;
      }
    }
  }
  return (columns || 0) > 0;
}
```

packages/lexical-playground/src/plugins/TableHoverActionsV2Plugin/index.tsx
```diff
@@ -294,9 +306,66 @@ function TableHoverActionsV2({
...
+      const headerRow = $isTableRowNode(rows[0])
+        ? (rows[0] as TableRowNode)
```
etrepum: I don't think this cast is necessary?

```suggestion
        ? rows[0]
```

ivailop7:
@etrepum thanks for the pointers! Addressed the comments and made it no-op when there are merged cells. I think is in a mergable state now.

PR #8062, opened 2026-01-04
[lexical-react] Bug Fix: Clear remote cursor immediately on collaborator refresh

### LINKED ISSUE #8061: Bug: The cursor lags during collaborative editing.
## Steps To Reproduce

1. Use Chrome's split view feature to create two tabs, and open https://playground.lexical.dev/?isCollab=true in both tabs.
2. Refresh one of these tabs. The cursor will remain in the same position and will only disappear after a while (about 10 seconds). This issue occurs randomly with relatively high frequency.

## The current behavior

Image

## The expected behavior
This issue no longer exists.

aldoprogrammer:
### Descriptions
When a collaborator tab is refreshed or closed, their remote cursor
remains visible for several seconds due to delayed awareness cleanup.

This change clears the Yjs awareness state on page unload so the remote
cursor is removed immediately instead of waiting for a timeout.

Closes #8061

## Before
image
see that after refresh one tab, the other tab still see the previous position of cursor from the tab that we refrehed

## After
image
no delay cursor dissapear after one tab is refreshed.

### DIFF
--- a/packages/lexical-react/src/__tests__/unit/utils.tsx
+++ b/packages/lexical-react/src/__tests__/unit/utils.tsx
@@ -99,7 +99,7 @@ export class Client implements Provider {
     getStates: () => Map<number, UserState>;
     off(): void;
     on(): void;
-    setLocalState: (state: UserState) => void;
+    setLocalState: (state: UserState | null) => void;
     setLocalStateField: (field: string, value: unknown) => void;
   };
 
@@ -121,7 +121,7 @@ export class Client implements Provider {
         // TODO
       },
 
-      setLocalState: (state) => {
+      setLocalState: (state: UserState | null) => {
         this._awarenessState = state;
       },
       setLocalStateField: (field: string, value: unknown) => {


packages/lexical-react/src/shared/useYjsCollaboration.tsx
```diff
@@ -418,6 +418,37 @@ function useProvider(
...
+      // causing them to remove the cursor immediately instead of waiting for timeout
+      // Note: Yjs awareness.setLocalState accepts null, but the Lexical type definition
+      // doesn't reflect this. The runtime behavior is correct.
+      try {
+        // @ts-expect-error - Yjs awareness.setLocalState accepts null to signal disconnection,
+        // but the Lexical type definition only allows UserState
+        provider.awareness.setLocalState(null);
+      } catch (_e) {
+        // Ignore errors during cleanup (e.g., if provider is already disconnected)
+      }
```
etrepum, line 434: Any reason not to fix this bug?

aldoprogrammer: Fixed, updated the type definitions in both TypeScript (packages/lexical-yjs/src/index.ts) and Flow (packages/lexical-yjs/flow/LexicalYjs.js.flow) to allow null for setLocalState, and removed the @ts-expect-error suppression. The code is now properly typed.

etrepum: I think you need to push that commit

aldoprogrammer: done

etrepum (approved):

aldoprogrammer:
Just fixed it, it should pass now

PR #8063, opened 2026-01-04
[lexical-playground]: Rearrange table columns

ivailop7:
Add a drag handle button to the table hover action buttons. 
Leverages the atlaskit npm packages. 
Also fixes a number of FloatingUI issues.
Only works if the table does not contain any merged cells.

https://github.com/user-attachments/assets/6d84368f-98b4-49d5-8027-62d71a360507

### DIFF
--- a/packages/lexical-extension/src/TabIndentationExtension.ts
+++ b/packages/lexical-extension/src/TabIndentationExtension.ts
@@ -6,9 +6,17 @@
  *
  */
 
-import type {LexicalCommand, LexicalEditor, RangeSelection} from 'lexical';
+import type {
+  ElementNode,
+  LexicalCommand,
+  LexicalEditor,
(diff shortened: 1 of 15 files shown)


etrepum:
Looks like this one needs new test expectations

ivailop7:
@etrepum this one is ready for rereview

package.json
```diff
@@ -209,6 +209,8 @@
     "y-websocket": "^1.5.4"
   },
   "dependencies": {
+    "@atlaskit/drag-and-drop-indicator": "^0.11.1",
+    "@atlaskit/pragmatic-drag-and-drop": "^1.7.7",
```
etrepum: I think these should go into packages/lexical-playground/package.json

ivailop7: yep, fixing this.

PR #8065, opened 2026-01-05
[lexical][@lexical/react] Bug Fix: Fix cursor disappearing in Firefox when dragging blocks

### LINKED ISSUE #8064: Bug: Cursor disappears when dragging or interacting with the left-side section (hand icon gesture)
### Description
When interacting with the section on the left side (shows hand icon gesture), the text cursor becomes invisible. Typing still works, but the cursor is not visible until after some time. Clicking in the editor does not make it visible immediately. 

### Steps to Reproduce
1. Open a Lexical editor instance.
2. Grab or drag the left-side section (hand gesture icon area).
3. Observe that the text cursor disappears.
4. Try typing: text appears but cursor remains invisible.
5. Wait for some time: cursor reappears.

### Expected Behavior
The cursor should remain visible and follow the typing location regardless of dragging or interacting with the left section.

### Actual Behavior
Cursor disappears temporarily after interacting with the left section, even though typing still works. This happened on Firefox only, Chrome and Edge Works Fine

### Demo
[Watch Here](https://youtu.be/s7j1M9L5ssg)

### Environment
- Lexical version: 0.39.0
- Browser: Firefox 146.0.1. 
- OS: Windows
- Node version: 20.19.5 

### Additional Context
- This behavior may be related to cursor state or selection handling when the left-hand section is manipulated.

### Impact of fix
This bug causes confusion for users because the text cursor is invisible after interacting with the left-hand section. Users can still type, but they cannot see where their input will appear, leading to potential mistakes. Fixing this will improve the editing experience and reduce user frustration, especially for Firefox users. All users editing in this scenario will benefit.

aldoprogrammer:
## Description

**Problem:**
In Firefox (version 146.0.1+), the text cursor disappears when interacting with the draggable block handle (hand icon). The cursor becomes invisible when starting a drag operation, even though typing still works. This issue does not occur in Chrome or Edge.

**Root Cause:**
Firefox fires a `blur` event on the editor root element when clicking the drag handle, which happens *before* the `dragstart` event. This causes focus loss and makes the cursor invisible.

**Solution:**
1. **Blur handler fix** (`LexicalEvents.ts`): Detect when blur is caused by clicking the drag handle (by checking if `relatedTarget` has class `draggable-block-menu`). When detected, immediately restore focus synchronously and prevent the `BLUR_COMMAND` from being dispatched.

2. **Drag start fallback** (`LexicalDraggableBlockPlugin.tsx`): Add synchronous focus restoration in `onDragStart` as a fallback to ensure cursor visibility during drag operations.

**Changes:**
- Modified `packages/lexical/src/LexicalEvents.ts` - Added Firefox-specific blur handler logic
- Modified `packages/lexical-react/src/LexicalDraggableBlockPlugin.tsx` - Added focus restoration in `onDragStart` and `onDragEnd`

### Before
We can't see the pointer cursor poisitoin, after we make a drag action on left side, please check the demo ini #8064 to see it.

### After
Now the cursor position remains visible after we did drag action on left side. 

### Demo
[Watch here](https://youtu.be/3aPlQozLgz4)

**Testing:**
- ✅ Tested in Firefox 146.0.1 - cursor remains visible during drag
- ✅ Tested in Chrome - no regressions
- ✅ Tested in Edge - no regressions
- ✅ Normal typing behavior unchanged in all browsers
- ✅ Unit tests pass (LexicalSelection tests)
- ✅ Linting passes

Closes #8064

### DIFF
--- a/packages/lexical-react/src/__tests__/unit/utils.tsx
+++ b/packages/lexical-react/src/__tests__/unit/utils.tsx
@@ -99,7 +99,7 @@ export class Client implements Provider {
     getStates: () => Map<number, UserState>;
     off(): void;
     on(): void;
-    setLocalState: (state: UserState) => void;
+    setLocalState: (state: UserState | null) => void;
     setLocalStateField: (field: string, value: unknown) => void;
   };
 
@@ -121,7 +121,7 @@ export class Client implements Provider {
         // TODO
       },
 
-      setLocalState: (state) => {
+      setLocalState: (state: UserState | null) => {
         this._awarenessState = state;
       },
       setLocalStateField: (field: string, value: unknown) => {


aldoprogrammer:
Thanks for the feedback! I've refactored the fix to keep the workaround contained within the plugin as requested.

**Changes made:**
- Removed the hardcoded `draggable-block-menu` class check from `LexicalEvents.ts`
- Moved the blur prevention logic to `LexicalDraggableBlockPlugin.tsx` using the existing `isOnMenu` function (which is already provided to the plugin)
- Added a blur event listener via `registerRootListener` that detects when blur is caused by clicking the drag handle
- Added a `BLUR_COMMAND` handler as a fallback to intercept the command if needed

The plugin now handles its own blur prevention without any hardcoded class names in the core, maintaining proper separation of concerns. The fix still works correctly in Firefox - the cursor remains visible when dragging blocks.

etrepum:
Right now this PR also includes the changes from #8062 so I think it will fail the type check for the same reason. I'll take another look at this one once that PR is fixed and merged.

packages/lexical/src/LexicalSelection.ts
```diff
@@ -3095,6 +3096,50 @@ export function updateDOMSelection(
...
+      !rootElement.contains(document.activeElement))
+  ) {
+    // Restore focus immediately to ensure cursor visibility
+    rootElement.focus({preventScroll: true});
+    // Immediately re-apply selection after focus to ensure Firefox shows cursor
+    // Firefox sometimes loses the selection when focus is restored, so we need to
+    // explicitly set it again right after focus
+    try {
+      // Use requestAnimationFrame to ensure this happens in the next frame after focus
+      requestAnimationFrame(() => {
```
etrepum: This seems like it could be problematic since there could be another update which changes the selection before the rAF callback

packages/lexical-react/src/LexicalDraggableBlockPlugin.tsx
```diff
@@ -409,6 +416,15 @@ function useDraggableBlockMenu(
       }
       setDraggableBlockElem(null);
 
+      // Firefox-specific fix: Use editor.focus() after drop to properly restore
+      // both focus and selection. This ensures cursor visibility immediately.
+      if (IS_FIREFOX) {
+        // Use a microtask to ensure this happens after the drop update completes
+        Promise.resolve().then(() => {
+          editor.focus();
+        });
```
etrepum: Did you try using $onUpdate first?

```suggestion
        $onUpdate(() => {
          editor.focus();
        });
```

packages/lexical/src/LexicalSelection.ts
aldoprogrammer: Yeah, just noticed it. Using requestAnimationFrame here could race with another update that modifies the selection. I’ll remove this and handle it synchronously within the update cycle instead.

packages/lexical-react/src/LexicalDraggableBlockPlugin.tsx
aldoprogrammer: Not yet, good call. I’ll try using $onUpdate first and update the PR.

aldoprogrammer:
I’ve updated the PR. Please let me know if I missed anything.

etrepum (approved):

PR #8066, opened 2026-01-05
[lexical-playground] Add button shows the Component Picker

ivailop7:
The + icon next to the draggable block handle now shows the ComponentPicker menu vs adding a paragraph below.

https://github.com/user-attachments/assets/c1844530-ccee-40e5-bdff-b23d067d61c4


ivailop7:
@etrepum @zurfyx could I get this one signed off?

PR #8070, opened 2026-01-11
[lexical][@lexical/link] Bug Fix: Fix infinite transform loop in AutoLinkPlugin

### LINKED ISSUE #7646: Bug: Pattern makes AutoLinkPlugin throw error:  One or more transforms are endlessly triggering additional transforms
I'm using lexical to display read-only content. I utilize `AutoLinkPlugin` to convert the pattern `/#\d{4,}/` to a link (for example, `#1234`). I noticed that, starting from lexical version 0.16.1, when there's a period plus non-space after that pattern, `AutoLinkPlugin` throws the error `One or more transforms are endlessly triggering additional transforms`.

Examples:

1. Content like `Check this out! #1234.Another thing...` makes `AutoLinkPlugin` throw the error (because there's a period + non-space after `#1234`)
2. With content like `Check this out! #1234. Another thing...` the AutoLinkPlugin works fine (there's no period + non-space after `#1234`)

Lexical version: 0.32.1

## Steps To Reproduce

1. Clone the repo
2. Install the dependencies (`npm install`)
3. Run the app (`npm run start`)
4. Check the browser console. The error should be logged there.

Image

Link to code example:

https://github.com/daedmunoz/lexical-autolink-issue

Everything is in file `src/App.tsx`. MATCHERS array with the pattern is below.
```typescript
const MATCHERS = [
  createLinkMatcherWithRegExp(/#\d{4,}/, (text: string) => `/#/segment1/segment2/${text.replace("#", "")}`),
];

```

## The current behavior
The content `#1234` is not a link, and there's an error `One or more transforms are endlessly triggering additional transforms` logged in the console.
## The expected behavior
The content `#1234` should be a link, and there should be no `One or more transforms are endlessly triggering additional transforms` error in the console.

## Impact of fix
The impact of this fix can be high since it is currenlty affecting some clients of the company I work for. In fact, that error is breaking the whole webapp when users visit the page when we display content with those patterns.

The temporal solution I found is to use a previous version of the plugin `AutoLinkPlugin` (v0.16.0) but it would be very nice that you guys fix this issue so that content like `#1234.Someting else` work in the last version of your amazing library.

aldoprogrammer:
## Description

Fixes infinite transform loop in `AutoLinkPlugin` when initializing content with patterns like `#1234.Another` where the matcher matches `#1234` but the text continues with a period (`.`). The period after the match causes boundary validation to fail, which triggers link unwrapping, which re-triggers the transform, creating an infinite loop.

## Current Behavior

When initializing editor content with text like `#1234.Another` where `#1234` matches a link matcher but is followed by a period, the AutoLink transform enters an infinite loop:
- Transform matches `#1234` and creates an `AutoLinkNode`
- Boundary validation fails because `.Another` follows the match
- Link is unwrapped, triggering another transform
- Process repeats indefinitely
- Error: "One or more transforms are endlessly triggering additional transforms"

## Changes in this PR

Made the `registerAutoLink` transform idempotent by adding guards to prevent re-processing nodes that are already part of an `AutoLinkNode`:

1. **`$handleLinkCreation`**: Added early return if any input node is already linked, and skip creating a new link if matching nodes are already part of an `AutoLinkNode`.

2. **`handleBadNeighbors`**: Added early return if `textNode` is already part of an `AutoLinkNode`, and added validation checks before appending to siblings to ensure nodes are still siblings and the combined text would still form a valid link.

3. **`registerAutoLink` transform**: Added check to skip processing if `textNode` is already part of an `AutoLinkNode`.

These changes ensure the transform only processes nodes that actually need link creation/editing, preventing infinite loops while maintaining correct link behavior.

## Testing

- Added unit test `registerAutoLink does not cause infinite transform loop with #1234.Another` that reproduces the issue and verifies the fix
- All existing tests pass (68/68)

Closes #7646

### DIFF
--- a/packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
+++ b/packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
@@ -22,7 +22,7 @@ import {
   TextNode,
 } from 'lexical/src';
 import {initializeUnitTest} from 'lexical/src/__tests__/utils';
-import {describe, expect, test} from 'vitest';
+import {assert, describe, expect, test} from 'vitest';
 
 const editorConfig = Object.freeze({
   namespace: '',
@@ -65,29 +65,27 @@ describe('LexicalAutoLinkExtension tests', () => {
       // Verify content is correct and that #1234 was converted to an AutoLinkNode
       editor.read(() => {
         const root = $getRoot();
-        const paragraph = root.getFirstChild();
-
-        // Verify paragraph exists (not empty root) and is an ElementNode
-        expect(paragraph).not.toBeNull();
-        expect($isParagraphNode(paragraph)).toBe(true);
-
-        const paragraphNode = paragraph as ParagraphNode;
+        const paragraphNode = root.getFirstChild();
+        assert(
+          $isParagraphNode(paragraphNode),
+          'first root child must be a ParagraphNode',
+        );
         expect(paragraphNode.getTextContent()).toBe('#1234.Another');
 
         // Verify that #1234 was converted to an AutoLinkNode
-        const firstChild = paragraphNode.getFirstChild();
-        expect(firstChild).not.toBeNull();
-        expect($isAutoLinkNode(firstChild)).toBe(true);
+        const autoLinkNode = paragraphNode.getFirstChild();
+        assert(
+          $isAutoLinkNode(autoLinkNode),
+          'first child must be an AutoLinkNode',
+        );
 
         // The AutoLinkNode should contain "#1234" only (the matched portion)
-        const autoLinkNode = firstChild as AutoLinkNode;
         expect(autoLinkNode.getTextContent()).toBe('#1234');
 
         // Verify that ".Another" is separate text after the link (unmatched portion)
         const nextSibling = autoLinkNode.getNextSibling();
-        expect(nextSibling).not.toBeNull();
-        expect($isTextNode(nextSibling)).toBe(true);
-        expect((nextSibling as TextNode).getTextContent()).toBe('.Another');
+        assert($isTextNode(nextSibling), 'next sibling must be a TextNode');
+        expect(nextSibling.getTextContent()).toBe('.Another');
       });
 
       unregister();


etrepum:
It looks like this PR is failing e2e tests

aldoprogrammer:
Ok, just fixed it

packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      await editor.update(() => {
+        const root = $getRoot();
+        const paragraph = $createParagraphNode();
+        const textNode = $createTextNode('#1234.Another');
+        paragraph.append(textNode);
+        root.append(paragraph);
+      });
+
+      // Wait for transforms to settle
+      await new Promise((resolve) => setTimeout(resolve, 100));
```
etrepum: No need for this, just slows the tests down. Transforms generally should not cause cascading updates, it will all be resolved in one microtask before the await is resolved.

```suggestion
```

packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        });
+      } catch (error: unknown) {
+        if (
+          error instanceof Error &&
+          error.message.includes('infinite transform')
+        ) {
+          transformErrorThrown = true;
+        }
+        throw error;
+      }
```
etrepum: The transform error would not be thrown here, it would happen in the update cycle above

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -476,16 +534,20 @@ export function registerAutoLink(
...
-          textNode.isSimpleText() &&
-          (startsWithSeparator(textNode.getTextContent()) ||
-            !$isAutoLinkNode(previous))
-        ) {
-          const textNodesToMatch = getTextNodesToMatch(textNode);
-          $handleLinkCreation(textNodesToMatch, matchers, onChange);
-        }
+        // Only process if textNode is not already part of an AutoLinkNode
+        const currentParent = textNode.getParent();
+        if (!$isAutoLinkNode(currentParent)) {
```
etrepum: currentParent === parent and we already know it's not an AutoLinkNode because it's not a LinkNode, I don't think any of the changes in this function are necessary.

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -393,29 +416,64 @@ function handleBadNeighbors(
...
     !nextSibling.getIsUnlinked() &&
     !endsWithSeparator(text)
   ) {
-    replaceWithChildren(nextSibling);
-    handleLinkEdit(nextSibling, matchers, onChange);
-    onChange(null, nextSibling.getURL());
+    // Check if the nextSibling is still a sibling (hasn't been moved) to prevent loops
+    const currentNextSibling = textNode.getNextSibling();
+    if (
+      currentNextSibling === nextSibling &&
```
etrepum: Going from not dirty to dirty will change the identity of currentNextSibling even if it hasn't materially changed


```suggestion
    // Check if the nextSibling is still a sibling (hasn't been moved) to prevent loops
    if (
      nextSibling.is(textNode.getNextSibling()) &&
```

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -393,29 +416,64 @@ function handleBadNeighbors(
...
+    return;
   }
 
+  // Handle case: textNode added AFTER a link, making link invalid
+  // Check if previousSibling is a link and adding this textNode makes it invalid
+  if ($isAutoLinkNode(previousSibling) && !previousSibling.getIsUnlinked()) {
+    // Check if the textNode is still a sibling (hasn't been moved) to prevent loops
+    const currentPreviousSibling = textNode.getPreviousSibling();
+    if (
+      currentPreviousSibling === previousSibling &&
```
etrepum:
```suggestion
    if (
      previousSibling.is(textNode.getPreviousSibling()) &&
```

packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      // Verify content is correct
+      const state = editor.getEditorState();
+      state.read(() => {
+        const root = $getRoot();
+        const paragraph = root.getFirstChild();
+        if (paragraph) {
+          const textContent = paragraph.getTextContent();
+          expect(textContent).toBe('#1234.Another');
+        }
```
etrepum: This test would also pass if the root was empty, which is a bit weird.

I would expect this test to also confirm whether or not there is supposed to be an AutoLinkNode here and what portion of the text it contains

packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        const root = $getRoot();
+        const paragraph = $createParagraphNode();
+        const textNode = $createTextNode('#1234.Another');
+        paragraph.append(textNode);
+        root.append(paragraph);
+      });
+
+      // Verify content is correct
+      const state = editor.getEditorState();
+      state.read(() => {
```
etrepum:
```suggestion
      editor.read(() => {
```

aldoprogrammer:
Plesae do check again, just push another one.

packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      // Verify content is correct and that #1234 was converted to an AutoLinkNode
+      editor.read(() => {
+        const root = $getRoot();
+        const paragraph = root.getFirstChild();
+
+        // Verify paragraph exists (not empty root) and is an ElementNode
+        expect(paragraph).not.toBeNull();
+        expect($isParagraphNode(paragraph)).toBe(true);
+
+        const paragraphNode = paragraph as ParagraphNode;
```
etrepum: vitest has an [assert](https://vitest.dev/api/assert) function that lets you write expectations that will also refine the types so you don't need to do any casting

```suggestion
        const paragraphNode = root.getFirstChild();
        assert($isParagraphNode(paragraphNode), 'first root child must be a ParagraphNode');
```

etrepum (commented):
This looks good but let's clean up the tests to use `assert`, should reduce a lot of the code

aldoprogrammer:
oke cool, now another one pushed.

etrepum (approved):

PR #8071, opened 2026-01-12
[lexical-react] Chore: Expose onReposition prop on SelectionAlwaysOnDisplay

takuyakanbr:
Expose `onReposition` prop on `LexicalSelectionAlwaysOnDisplay` plugin and `selectionAlwaysOnDisplay` util.

I'd like to be able to customize how the persisted selections are rendered in an internal use case (custom background color based on active theme).

## Test plan

`pnpm run tsc`
`pnpm run flow`

### DIFF
--- a/packages/lexical-react/flow/LexicalSelectionAlwaysOnDisplay.js.flow
+++ b/packages/lexical-react/flow/LexicalSelectionAlwaysOnDisplay.js.flow
@@ -10,6 +10,6 @@
 /**
  * LexicalSelectionAlwaysOnDisplay
  */
-declare export function SelectionAlwaysOnDisplay(
+declare export function SelectionAlwaysOnDisplay({
   onReposition?: (node: Array<HTMLElement>) => void,
-): null;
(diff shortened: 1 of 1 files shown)


packages/lexical-react/flow/LexicalSelectionAlwaysOnDisplay.js.flow
```diff
@@ -10,4 +10,6 @@
 /**
  * LexicalSelectionAlwaysOnDisplay
  */
-declare export function SelectionAlwaysOnDisplay(): null;
+declare export function SelectionAlwaysOnDisplay(
+  onReposition?: (node: Array<HTMLElement>) => void,
+): null;
```
etrepum: I think this type is wrong because it's a Props object in typescript

etrepum:
Approving because the incorrect flow type probably doesn't affect anyone outside of Meta and we don't dismiss reviews on update so you can fix it and merge when ready

PR #8072, opened 2026-01-12
[lexical-playground] Feature: Create paragraphs around code

### LINKED ISSUE #7912: Feature: create empty paragraph before/after code block if it's the first/last node

jvithlani:
## Description
This PR ensures that a paragraph node always exists before and after a code block. Currently, the absence of these nodes prevents the up and down arrow keys from functioning correctly. This change guarantees a "buffer" node (defaulting to a paragraph) to ensure consistent keyboard navigation.

Closes #7912

### Before

https://github.com/user-attachments/assets/5c1c05e3-911d-4892-8bea-8844791ef2f6

### After

https://github.com/user-attachments/assets/305c72c2-293b-4bfa-baa6-17c96251d809

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CodeActionMenu.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CodeActionMenu.spec.mjs
@@ -195,6 +195,7 @@ test.describe('CodeActionMenu', () => {
     await assertHTML(
       page,
       `
+        <p class="PlaygroundEditorTheme__paragraph" dir="auto"><br /></p>
         <code
           class="PlaygroundEditorTheme__code"
           dir="auto"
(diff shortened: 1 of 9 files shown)


jvithlani:
https://github.com/facebook/lexical/pull/7914#issuecomment-3394527724

Just added a new plugin, @etrepum let me know if this doesnt belong to playground folder

etrepum:
It's fine to have it in the playground, but this implementation breaks the markdown view functionality which expects the editor to contain only a CodeNode. This plugin would have to be aware of that editor mode and disabled accordingly.

etrepum:
You can run the e2e tests locally, have the playground running in one terminal with `pnpm run start` and run the tests in another with `pnpm run test-e2e-chromium` https://github.com/facebook/lexical/blob/main/CONTRIBUTING.md#pull-requests

etrepum:
This PR has been closed because it is stale, if you intend to finish it please re-open or create a new PR

PR #8076, opened 2026-01-14
[lexical][@lexical/table] Fix Ctrl+A to select all cells in table with merged cells #8074

### LINKED ISSUE #8074: Bug: Ctrl+A fails to select all cells in table with merged cells when table is the only editor content
When cells are merged in a table, pressing `Ctrl+A` does not select all cells within the table. Instead, only a subset of cells are selected, breaking the expected selection behavior for table content.

Lexical version: 0.39.0

## Steps To Reproduce

1. Insert a table into the editor.
2. Remove all other nodes from the editor, leaving only the table.
3. Merge two cells in the last column.
4. Place the cursor inside any cell of the table.
5. Press `Ctrl+A`.

Link to code example:
https://playground.lexical.dev
Video:
https://github.com/user-attachments/assets/1b57c791-df24-4244-ad27-c75c576cdc59

## The current behavior
`Ctrl+A` fails to select all cells when the table contains merged cells and is the only element in the editor.

## The expected behavior
`Ctrl+A` should select all cells within the table.

aldoprogrammer:
## Description
- Fix Ctrl+A so it selects all table cells (including merged cells) when a table is the only editor content.
- Add table SELECT_ALL handler that only intercepts when the table is the sole root child.
- Add regression tests: unit (SELECT_ALL scenarios) and E2E for merged-cell table-as-only-content.

## Test plan
- pnpm run test-unit -- lexical-table
- pnpm run test-e2e-chromium -- packages/lexical-playground/__tests__/e2e/Tables.spec.mjs

## Before
`Ctrl+A `fails to select all cells when the table contains merged cells and is the only element in the editor. see from issue demo #8074 

## After
Success to select all cells when the table contains merged cells and is the only element in the editor. Demo [here](https://youtu.be/c_cg4cSGMS4)

Closes #8074

### DIFF
--- a/packages/lexical-table/src/LexicalTablePluginHelpers.ts
+++ b/packages/lexical-table/src/LexicalTablePluginHelpers.ts
@@ -26,7 +26,7 @@ import {
   $setSelection,
   CLICK_COMMAND,
   COMMAND_PRIORITY_EDITOR,
-  COMMAND_PRIORITY_HIGH,
+  COMMAND_PRIORITY_LOW,
   ElementNode,
   isDOMNode,
   LexicalEditor,
@@ -204,25 +204,7 @@ function $tableSelectAllCommand(): boolean {
   // This is required to reproduce the bug: table must be the only content, no empty paragraphs
   // This prevents breaking other tests that expect RangeSelection when there's content outside table
   const root = $getRoot();
-  const rootChildren = root.getChildren();
-
-  // Find the table's index in root children
-  let tableIndex = -1;
-  for (let i = 0; i < rootChildren.length; i++) {
-    if (rootChildren[i].is(tableNode)) {
-      tableIndex = i;
-      break;
-    }
-  }
-
-  if (tableIndex === -1) {
-    return false; // Table not found in root children (shouldn't happen)
-  }
-
-  // STRICT: Table must be the ONLY child - no other nodes before or after
-  // This matches the bug reproduction scenario where table is the only content
-  if (rootChildren.length !== 1) {
-    // There are other nodes (even empty paragraphs) - don't intercept
+  if (!root.is(tableNode.getParent()) || root.getChildrenSize() !== 1) {
     return false;
   }
 
@@ -432,7 +414,7 @@ export function registerTablePlugin(
     editor.registerCommand(
       SELECT_ALL_COMMAND,
       $tableSelectAllCommand,
-      COMMAND_PRIORITY_HIGH,
+      COMMAND_PRIORITY_LOW,
     ),
     editor.registerCommand(
       CLICK_COMMAND,
--- a/packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
+++ b/packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
@@ -34,6 +34,7 @@ import {
   $isRangeSelection,
   defineExtension,
   LexicalEditor,
+  NodeKey,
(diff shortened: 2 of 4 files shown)


packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -182,6 +187,78 @@ function $tableClickCommand(event: MouseEvent): boolean {
...
+  if (tableIndex === -1) {
+    return false; // Table not found in root children (shouldn't happen)
+  }
+
+  // STRICT: Table must be the ONLY child - no other nodes before or after
+  // This matches the bug reproduction scenario where table is the only content
+  if (rootChildren.length !== 1) {
+    // There are other nodes (even empty paragraphs) - don't intercept
+    return false;
+  }
```
etrepum, line 209: This is a lot of work, there are much simpler ways to find out if the table is the only child. The one with the least amount of code is probably something like this:

```ts
if (!root.is(tableNode.getParent()) || root.getChildrenSize() !== 1) {
  return false;
}
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -352,6 +429,11 @@ export function registerTablePlugin(
       },
       COMMAND_PRIORITY_EDITOR,
     ),
+    editor.registerCommand(
+      SELECT_ALL_COMMAND,
+      $tableSelectAllCommand,
+      COMMAND_PRIORITY_HIGH,
```
etrepum: This should be the lowest priority that implements the desired behavior which is probably this:

```suggestion
      COMMAND_PRIORITY_LOW,
```

packages/lexical-table/src/__tests__/unit/LexicalTableNode.test.tsx
```diff
@@ -763,7 +763,7 @@ describe('LexicalTableNode tests', () => {
                 </table>
               `,
             );
-          });
+          }, 15000);
```
etrepum, line 768: What's this for?

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -324,4 +331,271 @@ describe('TableExtension', () => {
...
+        // Verify selection spans from first to last cell
+        expect(anchorCell.getKey()).toBe(firstCellMap.cell.getKey());
+        expect(focusCell.getKey()).toBe(lastCellMap.cell.getKey());
+
+        // Verify all unique cells are in the selection (merged cells should appear once)
+        const selectedNodes = selection.getNodes().filter($isTableCellNode);
+        const uniqueCellKeys = new Set(
+          selectedNodes.map((node) => node.getKey()),
+        );
+        const totalUniqueCells = new Set<string>();
```
etrepum: Generally the NodeKey type alias is used any time keys are being used, even though it is currently indistinguishable from string.

```suggestion
        const totalUniqueCells = new Set<NodeKey>();
```

packages/lexical-table/src/__tests__/unit/LexicalTableNode.test.tsx
aldoprogrammer: This timeout increase is for a pre-existing test` ("Copy table with caption from an external source")` that was occasionally timing out.

etrepum (approved):

PR #8078, opened 2026-01-15
[lexical-link] Bug Fix: Toggle links with nested children

patrick-atticus:
## Description

When a link has nested children, the TOGGLE_LINK_COMMAND does nothing (it should remove the link). This PR adds handling for such cases.

Here's an example of the problematic lexical state:
```
 root
  └ (47) paragraph 
    └ (48) link "https://lexical.dev/"
      └ (49) heading 
>       └ (50) text "Lexical"
```
This can be reproduced by copying the title of a google search result and pasting it into the editor. 

## Changes

The link removal logic only checks the immediate parent of selected nodes, so it cant find the LinkNode if the parent is a HeadingNode.

1. Changed link detection in RangeSelection to use $findMatchingParent() instead of getParent(), allowing it to traverse up and find ancestor LinkNodes.
2. Updated $splitLinkAtSelection() to identify children that contain extracted nodes, not just children that are extracted nodes.

## Test plan

Captured the behaviour in a unit test, which now passes

### DIFF
--- a/packages/lexical-link/src/LexicalLinkNode.ts
+++ b/packages/lexical-link/src/LexicalLinkNode.ts
@@ -549,9 +549,16 @@ function $splitLinkAtSelection(
   );
 
   const allChildren = parentLink.getChildren();
-  const extractedChildren = allChildren.filter((child) =>
-    extractedKeys.has(child.getKey()),
-  );
+  // Check if a child is an extracted node OR contains an extracted node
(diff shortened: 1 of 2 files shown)


patrick-atticus:
I have actually worked on the normalization already - did you want me to include it in this PR or a separate change?

etrepum:
That's not necessary, having normalization be separate from this makes more sense. I was just leaving some comments while I was waiting to make sure the extended test suite still passed

PR #8082, opened 2026-01-17
[lexical][@lexical/table] Prevent nested table pastes while preserving table-selection pasting

### LINKED ISSUE #8077: Bug: Paste tables in tables even if hasNestedTables: false
This is a screen recording from the Lexical playground:

https://github.com/user-attachments/assets/efee7e2c-f387-41b1-8b15-59986490bc41

Lexical version: latest on playground

## Steps To Reproduce

1. Select a table and copy it to clipboard
2. Select multiple cells of the table
3. Right click and paste

## The current behavior
The copied content from the clipboard gets inserted, including a table, within the table.

## The expected behavior
The table should be stripped out when pasting into a table.

## Impact of fix
This could prevent some abuse of content blocks.

aldoprogrammer:
### Summary

Prevent nested tables when pasting into table cells.

### Testing
Original issue demo: #8077
After having conflict on this pr, pull latest and merged in main: [issue still persist](https://youtu.be/na9r6hgEotU)
Demo fix: [video](https://youtu.be/s5rMfbHZEfg)

Closes #8077

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -116,25 +116,6 @@ const isPointerDownOnEvent = (event: PointerEvent) => {
   return (event.buttons & 1) === 1;
 };
 
-type TableSelectionLike = BaseSelection & {
-  anchor: RangeSelection['anchor'];
-  focus: RangeSelection['focus'];
-  tableKey: NodeKey;
-};
-
-function isTableSelectionLike(
-  selection: BaseSelection,
-): selection is TableSelectionLike {
-  return (
-    $isTableSelection(selection) ||
-    (typeof selection === 'object' &&
-      selection !== null &&
-      'tableKey' in selection &&
-      'anchor' in selection &&
-      'focus' in selection)
-  );
-}
-
 export function isHTMLTableElement(el: unknown): el is HTMLTableElement {
   return isHTMLElement(el) && el.nodeName === 'TABLE';
 }
@@ -183,9 +164,9 @@ export function $insertTableNodesFromClipboard(
   nodes: Array<LexicalNode>,
   selection: BaseSelection,
 ): boolean {
-  const isTableSelectionLikeSelection = isTableSelectionLike(selection);
+  const isTableSelection = $isTableSelection(selection);
   const isRangeSelection = $isRangeSelection(selection);
-  const anchorAndFocus = isTableSelectionLikeSelection
+  const anchorAndFocus = isTableSelection
     ? [selection.anchor, selection.focus]
     : isRangeSelection
       ? selection.getStartEndPoints()
@@ -199,20 +180,18 @@ export function $insertTableNodesFromClipboard(
       $findMatchingParent(selection.focus.getNode(), (n) =>
         $isTableCellNode(n),
       ) !== null) ||
-    isTableSelectionLikeSelection;
-  const templateGrid = nodes.find(
-    (node): node is TableNode =>
-      $isTableNode(node) || node.getType() === 'table',
+    isTableSelection;
+  const templateGrid = nodes.find((node): node is TableNode =>
+    $isTableNode(node),
   );
   const hasOnlyTableNodes =
-    nodes.length > 0 &&
-    nodes.every((node) => $isTableNode(node) || node.getType() === 'table');
+    nodes.length > 0 && nodes.every((node) => $isTableNode(node));
 
   if (
     !isSelectionInsideOfGrid ||
     anchorAndFocus === null ||
     templateGrid === undefined ||
-    (!isTableSelectionLikeSelection && !hasOnlyTableNodes)
+    (!isTableSelection && !hasOnlyTableNodes)
   ) {
     return false;
   }
@@ -252,7 +231,7 @@ export function $insertTableNodesFromClipboard(
   let affectedRowCount = templateGridMap.length;
   let affectedColCount = affectedRowCount > 0 ? templateGridMap[0].length : 0;
 
-  if (isTableSelectionLikeSelection) {
+  if (isTableSelection) {
     // If we have a table selection, we'll only modify the cells within
     // the selection boundary.
     const selectionBoundary = $computeTableCellRectBoundary(
@@ -363,7 +342,7 @@ export function $insertTableNodesFromClipboard(
     }
   }
 
-  if (isTableSelectionLikeSelection && didPerformMergeOperations) {
+  if (isTableSelection && didPerformMergeOperations) {
     // reset the table selection in case the anchor or focus cell was
     // removed via merge operations
     const [finalGridMap] = $computeTableMapSkipCellCheck(


randal-atticus:
@aldoprogrammer Apologies, I might have caused the conflicts in this PR due to #8088, which I did preparation for another change relating to nested tables. That PR doesn't change any behaviour but it restructured `SELECTION_INSERT_CLIPBOARD_NODES_COMMAND` in a similar way. 

For the avoidance of doubt, when `hasNestedTables: true`, a mixed-node clipboard should still insert a table into a cell (rather than pasting over the table) - that's consistent with how other editors seem to handle nested tables. Is that behaviour preserved here?

aldoprogrammer:
Please do check again

aldoprogrammer:
> @aldoprogrammer Apologies, I might have caused the conflicts in this PR due to #8088, which I did preparation for another change relating to nested tables. That PR doesn't change any behaviour but it restructured `SELECTION_INSERT_CLIPBOARD_NODES_COMMAND` in a similar way.
> 
> For the avoidance of doubt, when `hasNestedTables: true`, a mixed-node clipboard should still insert a table into a cell (rather than pasting over the table) - that's consistent with how other editors seem to handle nested tables. Is that behaviour preserved here?

Yes @randal-atticus , the behavior is preserved. When hasNestedTables: true, mixed-node clipboards return false early, allowing the default paste handler to insert the table into the cell as a nested table rather than pasting over it.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -109,6 +116,25 @@ const isPointerDownOnEvent = (event: PointerEvent) => {
...
+  return (
+    $isTableSelection(selection) ||
+    (typeof selection === 'object' &&
+      selection !== null &&
+      'tableKey' in selection &&
+      'anchor' in selection &&
+      'focus' in selection)
+  );
+}
+
```
etrepum: In what situation does `$isTableSelection` not work for this use case?

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -153,6 +179,208 @@ export function $findParentTableCellNodeInTable(
...
+        $isTableCellNode(n),
+      ) !== null) ||
+    isTableSelectionLikeSelection;
+  const templateGrid = nodes.find(
+    (node): node is TableNode =>
+      $isTableNode(node) || node.getType() === 'table',
+  );
+  const hasOnlyTableNodes =
+    nodes.length > 0 &&
+    nodes.every((node) => $isTableNode(node) || node.getType() === 'table');
```
etrepum: getType should never be used for this use case, what situation are you trying to handle where `$isTableNode` doesn't work?

aldoprogrammer:
Thanks for the feedback! Removed the unnecessary `isTableSelectionLike` function (using `$isTableSelection` directly) and replaced `getType() === 'table'` with `$isTableNode()`. Fixed in the latest commit.  
@etrepum

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -153,6 +160,206 @@ export function $findParentTableCellNodeInTable(
...
+      ) !== null &&
+      $findMatchingParent(selection.focus.getNode(), (n) =>
+        $isTableCellNode(n),
+      ) !== null) ||
+    isTableSelection;
+  const templateGrid = nodes.find((node): node is TableNode =>
+    $isTableNode(node),
+  );
+  const hasOnlyTableNodes =
+    nodes.length > 0 && nodes.every((node) => $isTableNode(node));
```
etrepum, line 188: All of these guards can be used as-is, you don't have to wrap them in another function
```suggestion
  const isSelectionInsideOfGrid =
    (isRangeSelection &&
      anchorAndFocus !== null &&
      $findMatchingParent(selection.anchor.getNode(), $isTableCellNode) !== null &&
      $findMatchingParent(selection.focus.getNode(), $isTableCellNode) !== null) ||
    isTableSelection;
  const templateGrid = nodes.find($isTableNode);
  const hasOnlyTableNodes =
    nodes.length > 0 && nodes.every($isTableNode);
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -153,6 +160,206 @@ export function $findParentTableCellNodeInTable(
...
+    (!isTableSelection && !hasOnlyTableNodes)
+  ) {
+    return false;
+  }
+
+  const [anchor, focus] = anchorAndFocus;
+  const [anchorCellNode, anchorRowNode, gridNode] = $getNodeTriplet(anchor);
+  const focusCellNode = $findMatchingParent(focus.getNode(), (n) =>
+    $isTableCellNode(n),
+  );
```
etrepum, line 203:
```suggestion
  const focusCellNode = $findMatchingParent(focus.getNode(), $isTableCellNode);
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -153,6 +160,206 @@ export function $findParentTableCellNodeInTable(
...
+        continue; // cell is not a merged cell
+      }
+      $unmergeCellNode(cellMap.cell);
+      unmergedKeys.add(cellMap.cell.getKey());
+      didPerformMergeOperations = true;
+    }
+  }
+
+  let [interimGridMap] = $computeTableMapSkipCellCheck(
+    gridNode.getWritable(),
```
etrepum, line 275: The getWritable here is redundant, all of the accessor methods will already do this as needed. The identity of gridNode or direct access to its properties are not used here

```suggestion
    gridNode,
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -153,6 +160,206 @@ export function $findParentTableCellNodeInTable(
...
+    $insertTableRowAtNode(cellMap.cell);
+  }
+  const colsToInsert = affectedColCount - initialColCount + startCol;
+  for (let i = 0; i < colsToInsert; i++) {
+    const cellMap = interimGridMap[0][initialColCount - 1];
+    $insertTableColumnAtNode(cellMap.cell, true, false);
+  }
+
+  [interimGridMap] = $computeTableMapSkipCellCheck(
+    gridNode.getWritable(),
```
etrepum, line 293:
```suggestion
    gridNode,
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -153,6 +160,206 @@ export function $findParentTableCellNodeInTable(
...
+        if ($isTextNode(child)) {
+          const paragraphNode = $createParagraphNode();
+          paragraphNode.append(child);
+          cell.append(child);
+        } else {
+          cell.append(child);
+        }
+      });
+      originalChildren.forEach((n) => n.remove());
+    }
```
etrepum, line 342: This logic is wrong, for example the paragraphNode code does absolutely nothing because the child is moved out of the paragraphNode into the cell so the paragraphNode just becomes garbage. It's probably better to just assume that the template table is already normalized (or the destination will be normalized after the fact) than to try and do some ad-hoc normalization here.


```suggestion
      cell.splice(0, cell.getChildrenSize(), templateCell.getChildren());
```

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -153,6 +160,206 @@ export function $findParentTableCellNodeInTable(
...
+  }
+
+  if (isTableSelection && didPerformMergeOperations) {
+    // reset the table selection in case the anchor or focus cell was
+    // removed via merge operations
+    const [finalGridMap] = $computeTableMapSkipCellCheck(
+      gridNode.getWritable(),
```
etrepum, line 349:
```suggestion
      gridNode,
```

etrepum:

PR #8087, opened 2026-01-19
Fix(Playground): Prevent code block line wrapping to keep line number…

### LINKED ISSUE #4361: Bug: code block lines
Code block does not show line numbers all the way until the end

image

Sa-Te:
## Description
Fixed an issue where line numbers in the Code Block gutter would desynchronize or disappear when code lines wrapped due to window resizing.

## Fix
Added `white-space: pre;` to `.PlaygroundEditorTheme__code`. This forces the code block to scroll horizontally instead of wrapping, ensuring that 1 visual line always equals 1 logical line number. This matches the behavior of standard code editors (VS Code, GitHub).

## Test Plan
1. Open Playground.
2. Insert a Code Block.
3. Paste a large snippet (200+ lines) with long lines of text.
4. Resize the window to be narrow.
5. Observed: Horizontal scrollbar appears, and line numbers remain perfectly aligned with the text.

Fixes #4361

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -183,7 +183,7 @@
   content: attr(data-gutter);
   position: sticky;
   background-color: #eee;
-  left: 0px;
+  left: 0;
   float: left;
   top: 0;
   z-index: 3;


Sa-Te:
@etrepum Thanks for the feedback! I've updated the implementation to match the requested behavior:

-  Switched the gutter to position: sticky and float: left. It now stays pinned to the left edge while the code scrolls horizontally underneath it.
- Alignment: Used negative margins on the gutter (margin-top: -8px) to counteract the container padding, ensuring the line numbers align perfectly with the text baselines.
- Layout: Removed the container's left padding to let the sticky gutter handle the spacing naturally.

Validated that vertical scrolling still behaves natively (page scroll) and horizontal scrolling is contained within the block.

PR #8088, opened 2026-01-20
[lexical-table] Chore: Fix test for nested table pasting

randal-atticus:
## Description

When `hasNestedTables: true`, and you paste a table into another table, it flattens it. I'm working on adding a way to allow a table to be pasted into another table, and realised the test assumes we can already do that.

Before changing anything I want to make sure the test obeys the current behaviour.

### Before
https://github.com/user-attachments/assets/bdef9260-1c89-47c4-8892-1fda2b124fc1

### After

Behaviour unchanged:
https://github.com/user-attachments/assets/14e47f23-801a-4e7e-b5c1-4aad40507d0d

### DIFF
--- a/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
+++ b/packages/lexical-table/src/LexicalTableSelectionHelpers.ts
@@ -190,10 +190,19 @@ export function applyTableHandlers(
     detachTableObserverFromTableElement(tableElement, tableObserver),
   );
 
-  const createPointerHandlers = () => {
+  const createPointerHandlers = (startingCell: TableDOMCell | null) => {
     if (tableObserver.isSelecting) {
       return;
     }
+    tableObserver.isSelecting = true;
+
+    // Set anchor immediately if starting cell provided (handles direct drag without click)
+    if (startingCell !== null && tableObserver.anchorCell === null) {
+      editor.update(() => {
+        tableObserver.$setAnchorCellForSelection(startingCell);
+      });
+    }
+
     const onPointerUp = () => {
       tableObserver.isSelecting = false;
       editorWindow.removeEventListener('pointerup', onPointerUp);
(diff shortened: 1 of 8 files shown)


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -750,209 +743,6 @@ export function applyTableHandlers(
     ),
   );
 
-  tableObserver.listenersToRemove.add(
```
randal-atticus, line 771: I've hoisted this entire command handler up to LexicalTablePluginHelper, the only change is where `editor != dispatchEditor` is checked.

I had to move this up because _none_ of these command handlers were being registered in tests.
Separately, I was wondering if you think all of the command handlers should be lifted out `applyTableHandlers`, because:

- These handlers are not registered in tests (unless [wrapped in a certain way](https://github.com/facebook/lexical/blob/dbe2fcc66c8357c7cd816be91db9c27d7cadb405/packages/lexical-table/src/__tests__/unit/LexicalTableNode.test.tsx#L1846)) 
- This function is a mix of both DOM event handlers and lexical command handlers, perhaps should be registered separately
- The command handlers are registered somewhat redundantly for each table (e.g. if you have 100 tables in the doc, there will be 100 `CUT_COMMAND` handlers, 100 `FORMAT_TEXT_COMMAND` handlers etc).

etrepum: Commands being registered per-table doesn't really make a lot of sense.  I guess whoever did that originally was thinking that it would make sense to only register the commands when at least one table is in the document, and just didn't optimize for >1 table? These are the kinds of problems that extensions are much better at.

randal-atticus:
Sorry, fixing tests...

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -397,7 +406,14 @@ export function registerTablePlugin(
     ),
     editor.registerCommand(
       SELECTION_INSERT_CLIPBOARD_NODES_COMMAND,
-      ({nodes, selection}, dispatchEditor) => {
+      (selectionPayload, dispatchEditor) => {
```
randal-atticus, line 409: this is structured a bit oddly but is effectively mirroring the previous implementation 1:1 (because the old `SELECTION_INSERT_CLIPBOARD_NODES_COMMAND` had `COMMAND_PRIORITY_HIGH` and fell through to this handler afterwards).

etrepum (approved):

PR #8092, opened 2026-01-25
[lexical-list] Feature: Preserve ordered list numbering when split by blocks or paragraphs

### LINKED ISSUE #8037: Bug: 有序列表中间插入code block，序号被打断了
Image

Lexical version:

## Steps To Reproduce

1.
2.

Link to code example:

Sa-Te:
## Summary

Fixes #8037

Currently, when an Ordered List is interrupted by a Block Node (like a Code Block) or split via the Enter key (creating a Paragraph), the subsequent list resets its numbering to `1`.

This PR fixes the numbering continuity by:
1.  **Updating Logic (`LexicalListItemNode` & `formatList`):** Changed the list splitting logic to calculate the `newStart` index based on the splitting item's `indexWithinParent` rather than relying on fragile sibling counts.
2.  **Updating Renderer (`LexicalListNode`):** Added a check in `updateDOM` to ensure changes to the `start` attribute are actually applied to the DOM.

## Test Plan

**repro steps:**
1. Open Playground.
2. Create an ordered list:
   1. Item A
   2. Item B
   3. Item C
3. Place cursor at the end of "Item B" and press Enter (creating a new list item).
4. Insert a Code Block (or press Enter again to create a Paragraph).

**Expected Behavior (Fixed):**
The list below the block should continue numbering at **3**.

**Actual Behavior (Before Fix):**
The list below the block resets to **1**.

## Implementation Details

- **LexicalListNode.ts:** Added `prevNode.__start !== this.__start` check in `updateDOM`.
- **LexicalListItemNode.ts:** Updated `replace` and `insertAfter` to calculate `newStart` using `getIndexWithinParent()`.
- **formatList.ts:** Updated `$handleListInsertParagraph` to calculate `newStart` using `getIndexWithinParent()` when splitting a list with a paragraph.

### DIFF
--- a/packages/lexical-list/src/index.ts
+++ b/packages/lexical-list/src/index.ts
@@ -76,9 +76,9 @@ export const REMOVE_LIST_COMMAND: LexicalCommand<void> = createCommand(
   'REMOVE_LIST_COMMAND',
 );
 
-export type RegisterListOptions = {
+export interface RegisterListOptions {
   restoreNumbering?: boolean;
-};
+}
 
 export function registerList(
   editor: LexicalEditor,
@@ -298,6 +298,7 @@ export interface ListConfig {
    * When `false` (default), indentation is more flexible.
    */
   hasStrictIndent: boolean;
+  shouldPreserveNumbering: boolean;
 }
 
 /**
@@ -308,13 +309,20 @@ export const ListExtension = defineExtension({
   build(editor, config, state) {
     return namedSignals(config);
   },
-  config: safeCast<ListConfig>({hasStrictIndent: false}),
+  config: safeCast<ListConfig>({
+    hasStrictIndent: false,
+    shouldPreserveNumbering: false,
+  }),
   name: '@lexical/list/List',
   nodes: () => [ListNode, ListItemNode],
   register(editor, config, state) {
     const stores = state.getOutput();
     return mergeRegister(
-      registerList(editor),
+      effect(() => {
+        return registerList(editor, {
+          restoreNumbering: stores.shouldPreserveNumbering.value,
+        });
+      }),
       effect(() =>
         stores.hasStrictIndent.value
           ? registerListStrictIndentTransform(editor)
--- a/packages/lexical-list/src/utils.ts
+++ b/packages/lexical-list/src/utils.ts
@@ -208,7 +208,5 @@ export function $getNewListStart(
   list: ListNode,
   listItem: ListItemNode,
 ): number {
-  const children = list.getChildren();
-  const index = children.indexOf(listItem);
-  return list.getStart() + index;
+  return list.getStart() + listItem.getIndexWithinParent();
 }
--- a/packages/lexical-list/src/__tests__/unit/ListExtension.test.ts
+++ b/packages/lexical-list/src/__tests__/unit/ListExtension.test.ts
@@ -9,12 +9,14 @@ import {buildEditorFromExtensions, defineExtension} from '@lexical/extension';
 import {
   $createListItemNode,
   $createListNode,
+  $isListItemNode,
   $isListNode,
   CheckListExtension,
   ListExtension,
 } from '@lexical/list';
 import {RichTextExtension} from '@lexical/rich-text';
-import {$createTextNode, $getRoot} from 'lexical';
+import {$createTextNode, $getRoot, INSERT_PARAGRAPH_COMMAND} from 'lexical';
+import {invariant} from 'lexical/src/__tests__/utils';
 import {describe, expect, it} from 'vitest';
 
 // TODO: write more tests here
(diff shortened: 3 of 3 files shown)


packages/lexical-list/src/formatList.ts
```diff
@@ -570,7 +570,15 @@ export function $handleListInsertParagraph(): boolean {
...
   if (nextSiblings.length > 0) {
-    const newList = $createListNode(parent.getListType());
+    const listType = parent.getListType();
+    let newStart = 1;
+
+    if (listType === 'number') {
+      newStart = parent.getStart() + listItem.getIndexWithinParent();
+    }
+
+    const newList = $createListNode(parent.getListType(), newStart);
```
etrepum: Since this code is basically repeated 3 times maybe it should be consolidated into a function

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -227,14 +227,25 @@ export class ListItemNode extends ElementNode {
...
-      const newList = $createListNode(list.getListType());
+      // Split the list and calculate the new start index
       let nextSibling = this.getNextSibling();
+      const nextSiblings = [];
       while (nextSibling) {
-        const nodeToAppend = nextSibling;
+        nextSiblings.push(nextSibling);
         nextSibling = nextSibling.getNextSibling();
-        newList.append(nodeToAppend);
       }
```
etrepum, line 237: This is a long way of writing

```ts
const nextSiblings = this.getNextSiblings();
```

Sa-Te: Thanks for the guidance! Its a real honor to learn from a mentor like you.

I've refactored the code to use getNextSiblings() and consolidated the start index calculation into a $getNewListStart helper in utils.ts to reduce duplication. I also added a regression test in LexicalListItemNode.test.ts to verify numbering continuity when splitting lists.

levensta:
Hello, @Sa-Te. If this becomes the default behavior, how can I reset the counter for the split list section? I'd also like to be able to configure this as an option. I think that for some users this change will come as a surprise and will require additional redesign of those services that are integrated with lexical-based editors

Sa-Te:
> Hello, @Sa-Te. If this becomes the default behavior, how can I reset the counter for the split list section? I'd also like to be able to configure this as an option. I think that for some users this change will come as a surprise and will require additional redesign of those services that are integrated with lexical-based editors

Cheers for the feedback, mate!
The main idea here was to sort out the current behaviour, which feels a bit wonky compared to standard kit like Google Docs or Word. Most users get a bit miffed when their list resets to 1 automatically just because they popped a paragraph in the middle.

As for resetting the counter manually, that would likely be handled by the UI (like a 'Restart Numbering' context menu) manipulating the 'start' attribute, rather than relying on the default split behaviour.

I’m happy to leave it to the maintainers to decide if this needs to be behind a configuration flag, but personally, I reckon bringing it in line with standard rich-text expectations is the way forward.

levensta:
> The main idea here was to sort out the current behaviour, which feels a bit wonky compared to standard kit like Google Docs or Word. Most users get a bit miffed when their list resets to 1 automatically just because they popped a paragraph in the middle.
> 
> As for resetting the counter manually, that would likely be handled by the UI (like a 'Restart Numbering' context menu) manipulating the 'start' attribute, rather than relying on the default split behaviour.
> 
> I’m happy to leave it to the maintainers to decide if this needs to be behind a configuration flag, but personally, I reckon bringing it in line with standard rich-text expectations is the way forward.

Google Docs and Word really don't reset the counter, and I understand the motivation to bring the behavior in line with popular editors. But this is only one of the ways a list can be split. For example, Notion, just like lexical in its current state, splits list into separate nodes when you try to insert a paragraph between items. 

I'm not opposed to your change, but if you made it possible to reset the new default behavior, it would make the feature truly excellent

etrepum (commented):
What do you think about adding an option to the list extension that defaults to the previous behavior, but can be configured to have this new behavior? I think that would solve the compatibility issue with people that want to preserve the existing behavior

Sa-Te:
> What do you think about adding an option to the list extension that defaults to the previous behavior, but can be configured to have this new behavior? I think that would solve the compatibility issue with people that want to preserve the existing behavior

Thanks @etrepum and @levensta for the feedback! I've refactored the approach to be fully opt-in and backward compatible, moving the logic from the Node level to the Command level.

Changes in this update:
Reverted LexicalListItemNode: The low-level insertAfter / replace methods now behave exactly as before (resetting numbering to 1 on split). This ensures no breaking changes for existing integrations.

Configurable Logic: Updated $handleListInsertParagraph (and registerList) to accept a shouldPreserveNumbering option.
false (Default): Resets new list to 1.
true (Opt-in): Calculates the correct continuity (e.g., 1, 2, [split], 3).

React Plugin: Updated <ListPlugin /> to accept a shouldPreserveNumbering prop.

Playground: Enabled this feature by default in the Playground so the "smart" behavior is visible there.

Tests: Added unit tests covering both scenarios:
    Verifying that default usage resets numbering (Legacy).
    Verifying that { shouldPreserveNumbering: true } correctly maintains continuity.
    
Note: I have disabled the feature in the Playground Editor.tsx by default to ensure existing E2E tests pass. The logic is verified via Unit Tests.

packages/lexical-list/src/index.ts
```diff
@@ -76,7 +76,14 @@ export const REMOVE_LIST_COMMAND: LexicalCommand<void> = createCommand(
   'REMOVE_LIST_COMMAND',
 );
 
-export function registerList(editor: LexicalEditor): () => void {
+export type RegisterListOptions = {
+  restoreNumbering?: boolean;
+};
```
etrepum:
```suggestion
export interface RegisterListOptions {
  restoreNumbering?: boolean;
}
```

packages/lexical-list/src/utils.ts
```diff
@@ -200,3 +200,15 @@ export function $wrapInListItem(node: LexicalNode): ListItemNode {
...
+
+/**
+ * Calculates the start value for a new list created by splitting an existing list.
+ */
+export function $getNewListStart(
+  list: ListNode,
+  listItem: ListItemNode,
+): number {
+  const children = list.getChildren();
+  const index = children.indexOf(listItem);
```
etrepum:
```suggestion
  const index = listItem.getIndexWithinParent();
```

etrepum (commented):
This is missing support in `ListExtension`, the configuration should also be in its `ListConfig`.

PR #8093, opened 2026-01-25
[lexical-markdown] Bug Fix: Replace regex-based format matching with …

### LINKED ISSUE #8073: Bug: Markdown emphasis parser fails on overlapping delimiters (*text**more***) - CommonMark compliance issue
(Note: this bug description generated by Claude.ai)

# Bug: Markdown emphasis parser fails on overlapping delimiters (CommonMark compliance issue)

## Description

Lexical's `$convertFromMarkdownString` incorrectly parses markdown emphasis when single and double asterisk delimiters overlap in certain patterns. This violates the CommonMark specification for emphasis delimiter matching.

## Environment

- **Lexical version**: 0.39.0
- **Package**: `@lexical/markdown`
- **Browser**: All browsers (parser issue)

## Bug Type

Parser incorrectly handles CommonMark emphasis delimiter rules

## Expected Behavior (per CommonMark spec)

According to the [CommonMark specification](https://spec.commonmark.org/0.31.2/#emphasis-and-strong-emphasis), the markdown string:

```markdown
ggggg*jhhhh**dsjdhjdj***
```

Should parse as:

```html
<p>ggggg<em>jhhhh<strong>dsjdhjdj</strong></em></p>
```

**Breakdown:**
- First `*` opens italic
- `**` opens bold inside the italic
- First two asterisks of `***` close the bold
- Last `*` of `***` closes the italic

## Actual Behavior (Lexical v0.39.0)

Lexical parses it as:

```html
<p>ggggg*jhhhh<strong>dsjdhjdj</strong>*</p>
```

The first and last asterisks are treated as literal characters instead of emphasis delimiters.

## Test Cases

Here's a comprehensive test showing which patterns work and which fail:

```typescript
import { $convertFromMarkdownString } from "@lexical/markdown";
import { $getRoot } from "lexical";

// ✅ WORKS: Simple cases
$convertFromMarkdownString("*italic*", TRANSFORMERS);
// Result: <em>italic</em> ✓

$convertFromMarkdownString("**bold**", TRANSFORMERS);
// Result: <strong>bold</strong> ✓

$convertFromMarkdownString("***bold and italic***", TRANSFORMERS);
// Result: <em><strong>bold and italic</strong></em> ✓

// ✅ WORKS: Properly balanced nesting
$convertFromMarkdownString("**bold with *italic* inside**", TRANSFORMERS);
// Result: <strong>bold with <em>italic</em> inside</strong> ✓

$convertFromMarkdownString("*italic with **bold** inside*", TRANSFORMERS);
// Result: <em>italic with <strong>bold</strong> inside</em> ✓

// ❌ FAILS: Overlapping delimiters (opener before opener, closer after closer)
$convertFromMarkdownString("ggggg*jhhhh**dsjdhjdj***", TRANSFORMERS);
// Expected: ggggg<em>jhhhh<strong>dsjdhjdj</strong></em>
// Actual: ggggg*jhhhh<strong>dsjdhjdj</strong>* ✗
```

## Root Cause Analysis

After investigating the `@lexical/markdown` source code, the issue appears to be in the `findOutermostMatch` function in `LexicalMarkdown.dev.js` (lines 308-340).

### The Algorithm Flaw

The current algorithm:
1. Finds all opening tags using `openTagsRegExp` in a **left-to-right scan**
2. For each opening tag (in order of appearance), tries to match its full pattern
3. Returns the **first successful match**

This is a greedy left-to-right approach that doesn't respect CommonMark's emphasis delimiter matching rules.

### Why It Fails on `*jhhhh**dsjdhjdj***`

1. Scanner finds `*` at position 5, `**` at position 11
2. Tries to match `*...*` for position 5 → **NO MATCH** (finds `*jhhhh**` which has `**` at the end)
3. Tries to match `**...**` for position 11 → **MATCH!** (`**dsjdhjdj**`)
4. Applies bold formatting, leaving `*jhhhh` and `*` as literal characters

### CommonMark's Algorithm

Per the CommonMark spec, emphasis delimiter matching should:
1. Identify delimiter runs and apply **flanking rules**
2. Process emphasis from **innermost to outermost**
3. Match delimiters based on **precedence** (*** before ** before *)
4. Handle overlapping by respecting the **delimiter stack**

The CommonMark algorithm is significantly more complex than Lexical's current implementation. See [CommonMark spec section 6.2](https://spec.commonmark.org/0.31.2/#emphasis-and-strong-emphasis) for the full algorithm.

### Minimal Test Case

```typescript
import { createHeadlessEditor } from '@lexical/headless';
import { $convertFromMarkdownString, TRANSFORMERS } from '@lexical/markdown';
import { $getRoot } from 'lexical';

const editor = createHeadlessEditor({
  nodes: [],
  onError: (error) => console.error(error),
});

editor.update(() => {
  $convertFromMarkdownString('ggggg*jhhhh**dsjdhjdj***', TRANSFORMERS);
  const root = $getRoot();
  console.log(root.getTextContent()); // Shows: ggggg*jhhhhdsjdhjdj*
});
```

### Automated Test

```typescript
describe('@lexical/markdown emphasis parsing', () => {
  test('should handle overlapping emphasis delimiters per CommonMark spec', () => {
    const editor = createHeadlessEditor({
      nodes: [],
      onError: (error) => console.error(error),
    });

    editor.update(() => {
      $convertFromMarkdownString('ggggg*jhhhh**dsjdhjdj***', TRANSFORMERS);
      const root = $getRoot();
      const html = root.getFirstChild()?.exportDOM().element?.innerHTML;

      // Should have nested em and strong tags
      expect(html).toContain('<em>');
      expect(html).toContain('<strong>');
      expect(html).toMatch(/ggggg<em>jhhhh<strong>dsjdhjdj<\/strong><\/em>/);

      // Should NOT have literal asterisks
      expect(html).not.toContain('*');
    });
  });
});
```

## Related Issues

- #2632 - Similar issue with `**Bold *Italic***` (fixed in v0.22.0 via PR #5758)
- The fix in #5758 addressed one pattern but didn't fully implement CommonMark's delimiter matching algorithm

## Proposed Solution

Implement CommonMark's full emphasis delimiter algorithm in `findOutermostMatch` or replace it with a proper delimiter stack-based approach. Key changes needed:

1. **Apply flanking rules** to determine which delimiter runs can open/close
2. **Process by precedence**: *** before ** before *
3. **Track delimiter stack** to handle overlapping cases
4. **Match from innermost to outermost** instead of left-to-right

Alternatively, consider using a battle-tested CommonMark parser like `commonmark.js` or `markdown-it` for the import path, then convert to Lexical nodes.

## Impact

- **Severity**: Medium - affects CommonMark compliance
- **Frequency**: Low - specific pattern, but occurs in real-world markdown (especially LLM-generated content)
- **Workaround**: Users can avoid this pattern and use properly balanced emphasis like `**bold with *italic* inside**`

## Additional Context

This issue was discovered while working with markdown content in a rich text editor. The pattern appears occasionally in user-generated content and more frequently in LLM-generated markdown.

CommonMark compliance is important for interoperability with other markdown tools and ensuring user expectations are met when importing markdown content.

kimseongyu:
### Current Behavior
Currently, Markdown texts are converted into Lexical nodes using regex. In issue #8073, this approach fails to correctly identify nested or complex matching pairs.

For example, with the pattern `*text**text***`, the opening tag `*` should find its corresponding closing tag. But since regex searches for an independently existing closing `*`, it cannot match correctly.

### Changes in This PR
To resolve these inconsistencies, I have implemented the [CommonMark Delimiter Algorithm](https://spec.commonmark.org/0.31.2/#emphasis-and-strong-emphasis) for processing emphasis and strong emphasis.

The algorithm proceeds as follows:
1. Scan text to build a delimiter stack with canOpen/canClose properties
2. Process delimiters to find matching pairs using flanking rules and the rule of 3
3. Return the outermost matched emphasis

Additionally, this PR fixes the issue where formats inside code spans were incorrectly processed. However, there are still some remaining issues. inline elements other than code spans (e.g., links, raw HTML) are handled by text match transformers, making them difficult to address with the current implementation. A new conversion approach may be needed to fully resolve these cases.

As a result, all tests pass. One previously incorrect test was fixed, and 5 new test cases were added.

Closes #8073

### Before

https://github.com/user-attachments/assets/e3e9a3ac-19d8-4afb-b15d-67dc5fea7f0c

### After

https://github.com/user-attachments/assets/ca7da7f3-9411-4dae-9faf-17983e164f8e

### DIFF
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -736,9 +736,9 @@ describe('Markdown', () => {
       mdAfterExport: '\\*foo *bar baz*',
     },
     {
-      html: '<p><span style="white-space: pre-wrap;">*a </span><code spellcheck="false" style="white-space: pre-wrap;"><span>*</span></code><span style="white-space: pre-wrap;"> b </span><code spellcheck="false" style="white-space: pre-wrap;"><span>x</span></code><span style="white-space: pre-wrap;">*</span></p>',
+      html: '<p><i><em style="white-space: pre-wrap;">a </em></i><i><code spellcheck="false" style="white-space: pre-wrap;"><em>*</em></code></i><i><em style="white-space: pre-wrap;"> b </em></i><i><code spellcheck="false" style="white-space: pre-wrap;"><em>x</em></code></i></p>',
       md: '*a `*` b `x`*',
-      mdAfterExport: '\\*a `*` b `x`\\*',
+      mdAfterExport: '*a&#32;`*`&#32;b&#32;`x`*',
     },
     {
       html: '<p><span style="white-space: pre-wrap;">_foo_bar</span></p>',
--- a/packages/lexical-markdown/src/importTextFormatTransformer.ts
+++ b/packages/lexical-markdown/src/importTextFormatTransformer.ts
@@ -38,21 +38,38 @@ export function findOutermostTextFormatTransformer(
   // (e.g., links, raw HTML) may need similar treatment in the future.
   const codeRegex = textFormatTransformersIndex.fullMatchRegExpByTag['`'];
   const codeTransformer = textFormatTransformersIndex.transformersByTag['`'];
+
+  const excludeRanges: Array<{start: number; end: number}> = [];
   let codeMatch = null;
   if (codeRegex && codeTransformer) {
-    codeRegex.lastIndex = 0;
-    const codeRegexMatch = codeRegex.exec(textContent);
-    codeMatch = codeRegexMatch
-      ? {
-          content: codeRegexMatch[2],
-          endIndex: codeRegexMatch.index + codeRegexMatch[0].length,
-          startIndex: codeRegexMatch.index,
+    const globalRegex = new RegExp(codeRegex.source, 'g');
+    const matches = Array.from(textContent.matchAll(globalRegex));
+
+    for (const match of matches) {
+      const startIndex = match.index!;
+      const endIndex = startIndex + match[0].length;
+
+      if (!codeMatch) {
+        codeMatch = {
+          content: match[2],
+          endIndex,
+          startIndex,
           tag: '`',
-        }
-      : null;
+        };
+      }
+
+      excludeRanges.push({
+        end: endIndex,
+        start: startIndex,
+      });
+    }
   }
 
-  const delimiters = scanDelimiters(textContent, textFormatTransformersIndex);
+  const delimiters = scanDelimiters(
+    textContent,
+    textFormatTransformersIndex,
+    excludeRanges,
+  );
   const emphasisMatch =
     delimiters.length > 0
       ? processEmphasis(textContent, delimiters, textFormatTransformersIndex)
@@ -105,6 +122,7 @@ export function findOutermostTextFormatTransformer(
 function scanDelimiters(
   text: string,
   transformersIndex: TextFormatTransformersIndex,
+  excludeRanges: Array<{start: number; end: number}> = [],
 ): Delimiter[] {
   const delimiters: Delimiter[] = [];
   const delimiterChars = new Set(
@@ -113,11 +131,25 @@ function scanDelimiters(
       .map((tag) => tag[0]),
   );
 
+  const isEscaped = (index: number): boolean => {
+    let count = 0;
+    for (let i = index - 1; i >= 0 && text[i] === '\\'; i--) {
+      count++;
+    }
+    return count % 2 === 1;
+  };
+
+  const isInExcludedRange = (index: number): boolean => {
+    return excludeRanges.some(
+      (range) => index >= range.start && index < range.end,
+    );
+  };
+
   let i = 0;
   while (i < text.length) {
     const char = text[i];
 
-    if (!delimiterChars.has(char) || isEscaped(text, i)) {
+    if (!delimiterChars.has(char) || isEscaped(i) || isInExcludedRange(i)) {
       i++;
       continue;
     }
@@ -302,14 +334,6 @@ function isFlanking(
     WHITESPACE.test(secondary) ||
     PUNCTUATION.test(secondary)
   );
-}
-
-function isEscaped(text: string, index: number): boolean {
-  let count = 0;
-  for (let i = index - 1; i >= 0 && text[i] === '\\'; i--) {
-    count++;
-  }
-  return count % 2 === 1;
 }
 
 export function importTextFormatTransformer(


packages/lexical-markdown/src/importTextFormatTransformer.ts
```diff
@@ -22,69 +32,280 @@ export function findOutermostTextFormatTransformer(
...
-  const match = findOutermostMatch(textContent, textFormatTransformersIndex);
 
-  if (!match) {
-    return null;
+  // Find code span first. Emphasis delimiters inside inline elements (e.g., code spans)
+  // should not be processed. Currently only code spans are handled; other inline elements
+  // (e.g., links, raw HTML) may need similar treatment in the future.
+  const codeRegex = textFormatTransformersIndex.fullMatchRegExpByTag['`'];
+  const codeTransformer = textFormatTransformersIndex.transformersByTag['`'];
+  const codeRegexMatch = codeRegex.exec(textContent);
```
etrepum: It might make sense to explicitly set `codeRegex.lastIndex = 0` in case that regexp ever gets a `g` flag.

packages/lexical-markdown/src/importTextFormatTransformer.ts
```diff
@@ -22,69 +32,280 @@ export function findOutermostTextFormatTransformer(
...
+    return null;
+  }
 
-  const transformer: TextFormatTransformer =
-    textFormatTransformersIndex.transformersByTag[match[1]];
+  const regexMatch = [
+    textContent.slice(resultMatch.startIndex, resultMatch.endIndex),
+    resultMatch.tag,
+    resultMatch.content,
+  ] as RegExpMatchArray;
```
etrepum: Using a cast here is not as type-safe, all of the extra properties are optional so this should work just fine

```suggestion
  const regexMatch: RegExpMatchArray = [
    textContent.slice(resultMatch.startIndex, resultMatch.endIndex),
    resultMatch.tag,
    resultMatch.content,
  ];
```

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -707,15 +707,44 @@ describe('Markdown', () => {
...
+    },
+    {
+      html: '<p><span style="white-space: pre-wrap;">*foo </span><i><em style="white-space: pre-wrap;">bar baz</em></i></p>',
+      md: '*foo *bar baz*',
+      mdAfterExport: '\\*foo *bar baz*',
+    },
+    {
+      html: '<p><span style="white-space: pre-wrap;">*a </span><code spellcheck="false" style="white-space: pre-wrap;"><span>*</span></code><span style="white-space: pre-wrap;">*</span></p>',
+      md: '*a `*`*',
+      mdAfterExport: '\\*a `*`\\*',
```
AlessioGr: I think a test for multiple code formats would be useful, e.g.

```md
*a `*` b `x`*
```

AlessioGr:
I think this is a good change!

kimseongyu:
Thank you! I've addressed the feedback in the latest commit. I also fixed the case with ````*a `*` b `x`*````.

Previously, the delimiter stack included delimiters inside code spans, causing incorrect matching. Now, I exclude code span ranges from delimiter scanning to ensure correct behavior.

[playground](https://spec.commonmark.org/dingus/?text=*a%20%60*%60%20b%20%60x%60*)

PR #8094, opened 2026-01-26
Fix(table): Support 'scope' attribute in HTML import.

Sa-Te:
Currently, the table cell importer assumes all `<th>` elements are Row Headers. This ignores the scope attribute if present (e.g., `<th scope="col">`), leading to incorrect header states for semantic tables.

Fix: Updated $convertTableCellNodeElement to check for scope="col" and scope="row". It defaults to Row Header if scope is missing, preserving existing fallback behavior while enabling high-fidelity imports for semantic HTML.

Tests: Added unit tests in LexicalTableCellNode.test.ts to verify scope="col", scope="row", and missing scope scenarios.

### DIFF
--- a/packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
+++ b/packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
@@ -9,14 +9,16 @@
 import {$generateHtmlFromNodes, $generateNodesFromDOM} from '@lexical/html';
 import {
   $createTableCellNode,
+  $isTableCellNode,
   TableCellHeaderStates,
   TableCellNode,
 } from '@lexical/table';
-import {$getRoot} from 'lexical';
+import {$getRoot, DOMConversionOutput} from 'lexical';
 import {
   expectHtmlToBeEqual,
   html,
   initializeUnitTest,
+  invariant,
 } from 'lexical/src/__tests__/utils';
 import {describe, expect, test} from 'vitest';
 
@@ -235,6 +237,21 @@ describe('LexicalTableCellNode tests', () => {
       return specs.conversion(element);
     };
 
+    const expectTableCellNode = (result: DOMConversionOutput | null) => {
+      const node = result?.node;
+
+      if (Array.isArray(node)) {
+        throw new Error('Expected a single node, but got an array');
+      }
+
+      invariant(
+        $isTableCellNode(node),
+        'Expected result.node to be a TableCellNode',
+      );
+
+      return node;
+    };
+
     test('DOM Conversion: <th> with scope="col" becomes COLUMN header', async () => {
       const {editor} = testEnv;
 
@@ -244,11 +261,9 @@ describe('LexicalTableCellNode tests', () => {
 
         const result = convertHTMLTag(th);
 
-        const node = result!.node as TableCellNode;
+        const node = expectTableCellNode(result);
 
-        expect(
-          (node as TableCellNode & {__headerState: number}).__headerState,
-        ).toBe(TableCellHeaderStates.COLUMN);
+        expect(node.getHeaderStyles()).toBe(TableCellHeaderStates.COLUMN);
       });
     });
 
@@ -261,11 +276,9 @@ describe('LexicalTableCellNode tests', () => {
 
         const result = convertHTMLTag(th);
 
-        const node = result!.node as TableCellNode;
+        const node = expectTableCellNode(result);
 
-        expect(
-          (node as TableCellNode & {__headerState: number}).__headerState,
-        ).toBe(TableCellHeaderStates.ROW);
+        expect(node.getHeaderStyles()).toBe(TableCellHeaderStates.ROW);
       });
     });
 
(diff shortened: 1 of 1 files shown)


packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
```diff
@@ -16,6 +16,11 @@ import {
 } from 'lexical/src/__tests__/utils';
 import {describe, expect, test} from 'vitest';
 
+import {
+  $convertTableCellNodeElement,
+  TableCellNode,
+} from '../../LexicalTableCellNode';
+
```
etrepum: It would be better to test the public interface here rather than import these implementation details

Sa-Te: Good point. I've refactored the tests to use TableCellNode.importDOM() instead of importing the internal helper function directly. Checks passed!

packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
```diff
@@ -213,5 +217,72 @@ describe('LexicalTableCellNode tests', () => {
...
+      await editor.update(() => {
+        const th = document.createElement('th');
+        th.setAttribute('scope', 'col');
+
+        const result = convertHTMLTag(th);
+
+        const node = result!.node as TableCellNode;
+
+        expect(
+          (node as TableCellNode & {__headerState: number}).__headerState,
```
etrepum: I don't know why you're doing this cast?
```suggestion
          node.getHeaderState(),
```

packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
```diff
@@ -213,5 +217,72 @@ describe('LexicalTableCellNode tests', () => {
...
+      await editor.update(() => {
+        const th = document.createElement('th');
+        th.setAttribute('scope', 'row');
+
+        const result = convertHTMLTag(th);
+
+        const node = result!.node as TableCellNode;
+
+        expect(
+          (node as TableCellNode & {__headerState: number}).__headerState,
```
etrepum:
```suggestion
          node.getHeaderState(),
```

packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
```diff
@@ -213,5 +217,72 @@ describe('LexicalTableCellNode tests', () => {
...
+
+      await editor.update(() => {
+        const th = document.createElement('th');
+        // No scope attribute set
+        const result = convertHTMLTag(th);
+
+        const node = result!.node;
+
+        expect(
+          (node as TableCellNode & {__headerState: number}).__headerState,
```
etrepum:
```suggestion
          node.getHeaderState(),
```

packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
```diff
@@ -213,5 +217,72 @@ describe('LexicalTableCellNode tests', () => {
...
+    test('DOM Conversion: <th> with scope="col" becomes COLUMN header', async () => {
+      const {editor} = testEnv;
+
+      await editor.update(() => {
+        const th = document.createElement('th');
+        th.setAttribute('scope', 'col');
+
+        const result = convertHTMLTag(th);
+
+        const node = result!.node as TableCellNode;
```
etrepum:
```suggestion
        const node = result!.node;
        assert($isTableCellNode(node), 'Expected TableCellNode');
```

packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
```diff
@@ -213,5 +217,72 @@ describe('LexicalTableCellNode tests', () => {
...
+    test('DOM Conversion: <th> with scope="row" becomes ROW header', async () => {
+      const {editor} = testEnv;
+
+      await editor.update(() => {
+        const th = document.createElement('th');
+        th.setAttribute('scope', 'row');
+
+        const result = convertHTMLTag(th);
+
+        const node = result!.node as TableCellNode;
```
etrepum:
```suggestion
        const node = result!.node;
        assert($isTableCellNode(node), 'Expected TableCellNode');
```

packages/lexical-table/src/__tests__/unit/LexicalTableCellNode.test.ts
```diff
@@ -213,5 +217,72 @@ describe('LexicalTableCellNode tests', () => {
...
+
+    test('DOM Conversion: <th> without scope defaults to ROW header', async () => {
+      const {editor} = testEnv;
+
+      await editor.update(() => {
+        const th = document.createElement('th');
+        // No scope attribute set
+        const result = convertHTMLTag(th);
+
+        const node = result!.node;
```
etrepum:
```suggestion
        const node = result!.node;
        assert($isTableCellNode(node), 'Expected TableCellNode');
```

etrepum (approved):

PR #8097, opened 2026-01-27
[lexical-table][lexical-playground] Feature: Implement "fit nested tables" for nested table pasting

randal-atticus:
## Description
Adds a `hasFitNestedTables` flag to `TablePlugin` that resizes nested tables when pasted into table cells. This is optional, because if you have `hasHorizontalScroll` (scrollable tables) or scrollable table cells (as in Playground), you don't really need this. However, we _do_ need this for a use case that is designed to be closer to MS Word's tables, and I imagine this could be useful in core.

https://github.com/user-attachments/assets/dc1d62c4-db29-4a0e-8ae0-1fc161227047

### Before

Tables are pasted at full width. This _looks_ okay with `hasHorizontalScroll`, but not all Lexical users will be using this flag.

https://github.com/user-attachments/assets/aeaf2183-0c2e-4ebd-aefe-c26b75a15564

### After

The pasted table is resized to fit inside the parent cell, accounting for "insets" (padding + border). This is done recursively.

Basic case:

https://github.com/user-attachments/assets/dc1d62c4-db29-4a0e-8ae0-1fc161227047

Deeply nested tables are also resized:

https://github.com/user-attachments/assets/d0feb304-4db1-497b-baf3-974f8480ef44

Pasting into a merged cell:

https://github.com/user-attachments/assets/f2472c5e-4406-431b-b9c8-9e22beb4b9d0

### DIFF
--- a/packages/lexical-table/src/LexicalTablePluginHelpers.ts
+++ b/packages/lexical-table/src/LexicalTablePluginHelpers.ts
@@ -767,9 +767,7 @@ function $calculateCellInsets(cell: TableCellNode) {
   );
 }
 
-function $getTotalTableWidth(table: TableNode) {
-  const colWidths = table.getColWidths();
-  invariant(!!colWidths, 'Tables without colWidths are not supported');
+function $getTotalTableWidth(colWidths: readonly number[]) {
   return colWidths.reduce((curWidth, width) => curWidth + width, 0);
 }
 
@@ -785,21 +783,19 @@ function $resizeTableToFitCell(
   parentCellWidth: number,
   borderBoxInsets: number,
 ) {
-  if (node.getColWidths() === undefined) {
+  const oldColWidths = node.getColWidths();
+  if (!oldColWidths) {
     return node;
   }
 
   const usableWidth = parentCellWidth - borderBoxInsets;
-  const tableWidth = $getTotalTableWidth(node);
+  const tableWidth = $getTotalTableWidth(oldColWidths);
   if (tableWidth <= usableWidth) {
     return node;
   }
 
   const proportionalWidth = usableWidth / tableWidth;
-  const oldColWidths = node.getColWidths();
-  if (oldColWidths) {
-    node.setColWidths(oldColWidths.map((width) => width * proportionalWidth));
-  }
+  node.setColWidths(oldColWidths.map((width) => width * proportionalWidth));
 
   const rowChildren = node.getChildren().filter($isTableRowNode);
   for (const rowChild of rowChildren) {


packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -633,3 +671,157 @@ function $tableSelectionInsertClipboardNodesCommand(
...
+  if (colWidths) {
+    return colWidths[columnIndex];
+  }
+  return cell.getWidth();
+}
+
+/**
+ * Returns horizontal insets of the given cell (padding + border).
+ *
+ * TODO: merged cells?
```
randal-atticus: Note to self: Probably don't need to handle this case, because you are only ever pasting into the unmerged cell, and either way, merged cells have the same insets as normal cells.

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -633,3 +671,157 @@ function $tableSelectionInsertClipboardNodesCommand(
...
+    return colWidths.reduce((curWidth, width) => curWidth + width, 0);
+  }
+  const tableRow = table.getFirstChild();
+
+  invariant(
+    $isTableRowNode(tableRow),
+    'Expected first child of a Table to be a TableRowNode',
+  );
+
+  // TODO merged cells?
```
randal-atticus: Need to consider if we handle the case where `colWidths` is not set (i.e. older tables) - As far as I know, we prefer to have colWidths for all tables. If that is the case, perhaps we just don't do the auto-fit for tables without colWidths?

packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -58,17 +58,31 @@ function wrapAndSlowDown(method, delay) {
   };
 }
 
-export function wrapTableHtml(expected, {ignoreClasses = false} = {}) {
+export function wrapTableHtml(
```
randal-atticus, line 61: For whatever reason, `"dir="auto"` wasn't being applied on the nested table when it was pasted. Instead of debugging that, I just added `ignoreDir`. Then I tried to make it a bit more explicit about what it was doing (moving "dir" up to the div, and optionally adding the scrollable wrapper class).

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -462,12 +476,37 @@ function $tableSelectionInsertClipboardNodesCommand(
...
+  // When pasting multiple nodes (including tables) into a cell, update the table to fit.
+  if (isRangeSelection && hasNestedTables.peek()) {
+    return $insertTableNodesIntoCells(
+      nodes,
+      selection,
+      hasFitNestedTables.peek(),
+    );
+  }
+
+  // If we reached this point, there's a table in the selection and nested tables are not allowed - reject the paste.
```
randal-atticus, line 498: I'm aware this behaviour is different from #8082. In that PR, tables are stripped from the payload. The non-nested-table case isn't a focus of this specific PR so I didn't try to replicate that - but there will be a merge conflict.

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -215,6 +218,58 @@ describe('TableExtension', () => {
...
+      // Verify a nested table was created
+      editor.getEditorState().read(() => {
+        const root = $getRoot();
+        const table = root.getFirstChild();
+        assert($isTableNode(table), 'Expected table node');
+        const row = table.getFirstChild();
+        assert($isElementNode(row), 'Expected row node');
+        const cell = row.getFirstChild();
+        assert($isElementNode(cell), 'Expected cell node');
+        const cellChildren = cell.getChildren();
```
etrepum:
```suggestion
        assert($isTableRowNode(row), 'Expected row node');
        const cell = row.getFirstChild();
        assert($isTableCellNode(cell), 'Expected cell node');
        const cellChildren = cell.getChildren();
```

packages/lexical-table/src/__tests__/unit/LexicalTableExtension.test.ts
```diff
@@ -215,6 +218,58 @@ describe('TableExtension', () => {
...
+          root.append(table);
+          paragraph.select();
+        },
+        {discrete: true},
+      );
+
+      // Try to paste a table inside the cell
+      editor.update(
+        () => {
+          const tableNode = $createTableNode();
```
etrepum: Maybe this should create some sort of non-degenerate table? We don't really have support for zero cell tables, this could easily stop working if there was ever a normalization for that invariant

packages/lexical-react/src/LexicalTablePlugin.ts
```diff
@@ -108,3 +117,11 @@ export function TablePlugin({
...
   return null;
 }
+
+function usePropSignal<T>(value: T): Signal<T> {
+  const configSignal = useMemo(() => signal(value), [value]);
+  if (configSignal.peek() !== value) {
+    configSignal.value = value;
+  }
+  return configSignal;
+}
```
etrepum, line 127: `useMemo(() => signal(value), [value])` will return a new signal and thus re-register the plugin whenever the value changes.

```suggestion
function usePropSignal<T>(value: T): Signal<T> {
  const [configSignal] = useState(() => signal(value));
  if (configSignal.peek() !== value) {
    configSignal.value = value;
  }
  return configSignal;
}
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -447,9 +451,19 @@ function $tableSelectionInsertClipboardNodesCommand(
...
     typeof SELECTION_INSERT_CLIPBOARD_NODES_COMMAND
   >,
+  hasNestedTables: Signal<boolean>,
+  hasFitNestedTables: Signal<boolean>,
 ) {
   const {nodes, selection} = selectionPayload;
-  const anchorAndFocus = selection.getStartEndPoints();
+
+  const hasTables = nodes.some(
+    (n) => $isTableNode(n) || $dfs(n).some((d) => $isTableNode(d.node)),
```
etrepum, line 460: $dfs here does materialize an array, probably not a big deal but could be a bottleneck for a big paste

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -633,3 +671,150 @@ function $tableSelectionInsertClipboardNodesCommand(
...
+  const paddingLeft =
+    window.getComputedStyle(cellDOM).getPropertyValue('padding-left') || '0px';
+  const paddingRight =
+    window.getComputedStyle(cellDOM).getPropertyValue('padding-right') || '0px';
+  const borderLeftWidth =
+    window.getComputedStyle(cellDOM).getPropertyValue('border-left-width') ||
+    '0px';
+  const borderRightWidth =
+    window.getComputedStyle(cellDOM).getPropertyValue('padding-right-width') ||
+    '0px';
```
etrepum: this should probably only compute the style for cellDOM once into a variable

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -633,3 +671,150 @@ function $tableSelectionInsertClipboardNodesCommand(
...
+  const tableWidth = $getTotalTableWidth(node);
+  if (tableWidth <= usableWidth) {
+    return node;
+  }
+
+  const proportionalWidth = usableWidth / tableWidth;
+  const oldColWidths = node.getColWidths();
+  if (oldColWidths) {
+    node.setColWidths(oldColWidths.map((width) => width * proportionalWidth));
+  }
```
etrepum: This is calling getColWidths three times when once would be sufficient if the `oldWidthCols` variable was set first and used later, e.g. the `$getTotalTableWidths` function could take the `readonly number[]` since you have already narrowed it to be not-undefined
```suggestion
  const oldColWidths = node.getColWidths();
  if (!oldColWidths) {
    return node;
  }

  const usableWidth = parentCellWidth - borderBoxInsets;
  const tableWidth = $getTotalTableWidth(oldColWidths);
  if (tableWidth <= usableWidth) {
    return node;
  }

  const proportionalWidth = usableWidth / tableWidth;
  node.setColWidths(oldColWidths.map((width) => width * proportionalWidth));
```

(this suggestion doesn't include the refactor of `$getTotalTableWidth`)

packages/lexical-table/src/LexicalTablePluginHelpers.ts
randal-atticus: Hmm. Is there a more efficient way to figure out if the payload contains a table at all?

randal-atticus: Seems like we always need to do some kind of recursive traversal.

When `hasFitNestedTables: true`, we have to do a recursive traversal (well... I just realised if you have a TableNode within another ElementNode it won't be reached by the loop in `$resizeTableToFitCell` - will fix)

When `hasNestedTables: false`, we block the paste. But per #8082 it sounds like we want to instead be stripping all deeply nested tables (thereby requiring a recursive traversal anyway). 

The only time we don't need to recursively find/operate on tables is `hasNestedTables: true, hasFitNestedTables: false`.

etrepum: The performance issue isn’t that it’s traversing, it’s all of the garbage it generates to do that traversal (Arrays that are built and destroyed immediately)

etrepum (approved):

PR #8099, opened 2026-01-28
[lexical] Bug Fix: Refactor RootNode.__cachedText computation for coherency

### LINKED ISSUE #8096: Bug: $getRoot().getTextContent() new line inconsistencies
Lexical version: 0.38.2

## Steps To Reproduce

1. Load previously stored serialized state with multiple paragraphs.
2. Notice (console) how there are 4x \n between paragraphs.
3. Now add a space at the end.
4. Notice how there are now 2x \n between paragraphs.
5. Select end and press Enter 2x and type anything.
6. Notice there are 4x \n.

Link to code example: https://codesandbox.io/p/devbox/priceless-meadow-kmlmj2

## The current behavior

Inconsistent number of new lines in between paragraphs in $getRoot().getTextContent() output.

## The expected behavior

Consistent number of new lines in between paragraphs in $getRoot().getTextContent() output.

## Impact of fix

I am using `$getRoot().getTextContent()` to get text diff between initial and final state. The inconsistencies report fake addition/removals of new lines.

etrepum:
## Description

The way that the reconciler computes `RootNode.__cachedText` was inconsistent with the way that `ElementNode.getTextContent()` works. There were some cache coherency issues with the way ElementNode cached text was handled.

This refactors the logic such that:
* the text is only accumulated in one variable (previously the `subTreeTextContent` and `editorTextContent` were computed independently for no obvious reason)
* the `DOUBLE_LINE_BREAK` condition is handled when reconciling children of a node, rather than trying to append it to the cache of the ElementNode itself
* nodes are only reconciled once (some edge cases could end up processing a node multiple times since there wasn't a continue in the loop when a prevNode is destroyed)

Closes #8096

## Test plan

All existing tests pass, and new unit tests pass.

### DIFF
--- a/packages/lexical/src/LexicalReconciler.ts
+++ b/packages/lexical/src/LexicalReconciler.ts
@@ -14,9 +14,8 @@ import type {
   RegisteredNodes,
 } from './LexicalEditor';
 import type {LexicalPrivateDOM, NodeKey, NodeMap} from './LexicalNode';
-import type {ElementDOMSlot} from './nodes/LexicalElementNode';
+import type {ElementDOMSlot, ElementNode} from './nodes/LexicalElementNode';
 
-import devInvariant from 'shared/devInvariant';
 import invariant from 'shared/invariant';
 
 import {
@@ -25,7 +24,6 @@ import {
   $isLineBreakNode,
   $isRootNode,
   $isTextNode,
-  ElementNode,
 } from '.';
 import {
   DOUBLE_LINE_BREAK,
@@ -210,9 +208,11 @@ function $createNode(key: NodeKey, slot: ElementDOMSlot | null): HTMLElement {
     if (indent !== 0) {
       setElementIndent(dom, indent);
     }
-    const endIndex = childrenSize - 1;
-    const children = createChildrenArray(node, activeNextNodeMap);
-    $createChildren(children, node, 0, endIndex, node.getDOMSlot(dom));
+    if (childrenSize !== 0) {
+      const endIndex = childrenSize - 1;
+      const children = createChildrenArray(node, activeNextNodeMap);
+      $createChildren(children, node, 0, endIndex, node.getDOMSlot(dom));
+    }
 
     const format = node.__format;
 
@@ -389,7 +389,6 @@ function $reconcileChildren(
   const nextChildrenSize = nextElement.__size;
   subTreeTextContent = '';
   const dom: HTMLElement & LexicalPrivateDOM = slot.element;
-  dom.__lexicalTextContent = undefined;
 
   if (prevChildrenSize === 1 && nextChildrenSize === 1) {
     const prevFirstChildKey: NodeKey = prevElement.__first!;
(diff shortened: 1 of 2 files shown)


packages/lexical/src/__tests__/unit/LexicalNode.test.ts
```diff
@@ -769,27 +769,30 @@ describe('LexicalNode tests', () => {
...
+            expect(barTextNode.getCommonAncestor(bazTextNode)).toBe(null);
+            barParagraphNode.append(barTextNode);
+            bazParagraphNode.append(bazTextNode);
+            expect(barTextNode.getCommonAncestor(bazTextNode)).toBe(null);
+            expect(bazTextNode.getCommonAncestor(bazTextNode)).toBe(
+              bazParagraphNode,
+            );
+            rootNode.append(barParagraphNode, bazParagraphNode);
+          },
+          {discrete: true},
```
etrepum, line 794: This just makes the stack trace nicer when the invariant fails, doesn't change the nature of the test

packages/lexical/src/nodes/__tests__/unit/LexicalRootNode.test.ts
```diff
@@ -38,6 +38,9 @@ describe('LexicalRootNode tests', () => {
 
         expect(root.__cachedText).toBe(text);
 
+        expect(root.__cachedText).toBe(
+          ElementNode.prototype.getTextContent.call(root),
+        );
```
etrepum, line 43: This is a cleaner way to do the rootCopy thing, assuming RootNode doesn't do any meaningful changes to its implementation of getTextContent

packages/lexical/src/LexicalNode.ts
```diff
@@ -243,7 +243,6 @@ export type LexicalUpdateJSON<T extends SerializedLexicalNode> = Omit<
 export interface LexicalPrivateDOM {
   __lexicalTextContent?: string | undefined | null;
   __lexicalLineBreak?: HTMLBRElement | HTMLImageElement | undefined | null;
-  __lexicalDirTextContent?: string | undefined | null;
```
etrepum, line 246: This was unused

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -49,7 +50,6 @@ type IntentionallyMarkedAsDirtyElement = boolean;
 let subTreeTextContent = '';
 let subTreeTextFormat: number | null = null;
 let subTreeTextStyle: string | null = null;
-let editorTextContent = '';
```
etrepum, line 52: This is entirely redundant, subTreeTextContent does the same thing

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -266,7 +258,7 @@ function $createNode(key: NodeKey, slot: ElementDOMSlot | null): HTMLElement {
 
 function $createChildren(
   children: Array<NodeKey>,
-  element: ElementNode & LexicalPrivateDOM,
```
etrepum, line 269: This was a mistake, LexicalPrivateDOM is for DOM and not Lexical nodes

packages/lexical/src/LexicalReconciler.ts
```diff
@@ -555,7 +549,12 @@ function $reconcileNode(
     return replacementDOM;
   }
 
-  if ($isElementNode(prevNode) && $isElementNode(nextNode)) {
+  if ($isElementNode(prevNode)) {
+    invariant(
+      $isElementNode(nextNode),
+      'Node with key %s changed from ElementNode to !ElementNode',
+      key,
+    );
```
etrepum, line 556: Since the key must refer to the same exact class of element an invariant makes more sense

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1312,12 +1312,6 @@ export function dispatchCommand<TCommand extends LexicalCommand<unknown>>(
   return triggerCommandListeners(editor, command, payload);
 }
 
-export function $textContentRequiresDoubleLinebreakAtEnd(
```
etrepum, line 1315: All use cases of this function were inlined because the isLastChild condition is known at all call sites

PR #8102, opened 2026-01-29
[lexical] Bug Fix: Preserve paragraph alignment when pasting

### LINKED ISSUE #8101: Bug: Text alignment is lost on copy-paste
**Lexical version:** latest (tested on https://playground.lexical.dev)

## Steps To Reproduce
1. Create multiple paragraphs.
2. Set the **first paragraph** alignment to **center** or **right**.
3. Copy the first paragraph.
4. Move the cursor to another paragraph that has **default left alignment**.
5. Paste the copied content.

[Demo](https://youtu.be/f83vxQGogIs)

## Link to code example

## The current behavior
The pasted text is reset to **left-aligned**, even though the copied text was center- or right-aligned.

## The expected behavior
The pasted text should **preserve its original alignment** (center or right).

## Impact of fix
This affects basic formatting workflows. Users expect copy-paste to retain alignment, especially in document-style editors.

aldoprogrammer:
## Description
Fixes paste alignment loss when copying from non-left aligned paragraphs. Covers both full paragraph and text-only selection; pasted content retains original alignment.Tests added for HTML/JSON paste and copy/paste scenarios.

Closes #8101

### Before
Can't persist the alignment copied text
### After
[Demo](https://youtu.be/yv7wbE5zmMI)


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -281,6 +283,15 @@ export interface BaseSerializedNode {
...
 }
 
+function isSerializedParagraphNode(
+  node: BaseSerializedNode,
+): node is SerializedElementNode {
+  return (
+    node.type === 'paragraph' &&
+    Array.isArray((node as {children?: unknown}).children)
+  );
+}
```
etrepum: This isn't a very extensible way to do this test, since this won't work for any ParagraphNode subclass. We also don't really have any other code that works with the exported JSON in any way other than to deserialize it.

etrepum (commented):
This needs a closer look from someone to come up with a better suggestion, but I don't think this is the right approach to solve this problem

packages/lexical-clipboard/src/clipboard.ts
aldoprogrammer: I’ve removed the hardcoded paragraph JSON assumptions and now derive the wrapper from the actual paragraph node’s `exportJSON()`, only replacing children. This preserves paragraph subclasses and avoids manual JSON handling beyond the children swap

packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -123,6 +123,15 @@ export class ParagraphNode extends ElementNode {
...
   }
 
+  extractWithChild(
+    child: LexicalNode,
+    selection: BaseSelection | null,
+    destination: 'clone' | 'html',
+  ): boolean {
+    const formatType = this.getFormatType();
+    return formatType !== '' && formatType !== 'left' && formatType !== 'start';
+  }
```
etrepum, line 133: I just did some final checks here and the problem with this approach is that it doesn't verify that the entire paragraph is selected when preserving the alignment. It's unexpected behavior to copy part of a paragraph and then on paste discover that it creates a new paragraph instead of just copying that text.

etrepum:

PR #8105, opened 2026-01-31
feat: add focus event option for the checklist extension

### LINKED ISSUE #7689: Feature: Add an option to disable focus on click in CheckListPlugin

rayterion:
## Description

This PR adds an optional configuration parameter, `disableTakeFocusOnClick`, to `CheckListExtension`. This allows users to enable or disable editor focus when clicking on checklist items, as described in the related issue:

- Closes #7689

The changes are implemented in a way that does not break existing usages of `CheckListExtension` or the `CheckListPlugin` React component. The new parameter is optional and defaults to the current behavior.

Note: In the file `LexicalCheckListPlugin.js.flow`, I added what I believe is the correct change based on patterns observed in similar files. However, I did not deeply verify the purpose of all files in that directory, so this part should be reviewed.

## Test plan

I tested the behavior by running pnpm run start and verifying how checklist items behave when clicked, both with and without the `disableTakeFocusOnClick` option enabled.
Steps:
1. Open `packages/lexical-playground/src/appSettings.ts` and set `shouldDisableFocusOnClickChecklist` to either false or true (default is false)
2. run `pnpm run start`
3. create a checklist in the editor (shortcut Ctrl+Shift+9 on pc).
4. Verify the focus behavior

I have not yet explicitly tested whether the mobile keyboard correctly dismisses when the option is set to true. However, in theory, this should work as expected, since the keyboard should only appear when the editor receives focus.

## Additional Context

I noticed there is an existing open PR with the same goal as this one, but it was not completed by its author. For reference:

- #7696

### DIFF
--- a/packages/lexical-list/src/checkList.ts
+++ b/packages/lexical-list/src/checkList.ts
@@ -304,6 +304,8 @@ function handleClick(
             if (disableFocusOnClick) {
               $addUpdateTag(SKIP_SELECTION_FOCUS_TAG);
               $addUpdateTag(SKIP_DOM_SELECTION_TAG);
+            } else {
+              domNode.focus();
             }
             node.toggleChecked();
           }
@@ -323,12 +325,13 @@ function handleSelectDefaults(
   event: MouseEvent | TouchEvent,
   disableTakeFocusOnClick: boolean,
 ) {
-  if (disableTakeFocusOnClick) {
-    handleCheckItemEvent(event, () => {
-      event.preventDefault();
+  handleCheckItemEvent(event, () => {
+    // Prevents caret moving when clicking on check mark.
+    event.preventDefault();
+    if (disableTakeFocusOnClick) {
       event.stopPropagation();
-    });
-  }
+    }
+  });
 }
 
 function getActiveCheckListItem(): HTMLElement | null {


packages/lexical-list/src/checkList.ts
```diff
@@ -38,7 +40,26 @@ export const INSERT_CHECK_LIST_COMMAND: LexicalCommand<void> = createCommand(
   'INSERT_CHECK_LIST_COMMAND',
 );
 
-export function registerCheckList(editor: LexicalEditor) {
+let disableTakeFocusOnClick = false;
```
etrepum: Using a module global is not the right way to implement this, there can be multiple editors with different settings. The setting should be stored with the extension's output or config.

packages/lexical-react/flow/LexicalCheckListPlugin.js.flow
```diff
@@ -7,4 +7,4 @@
  * @flow strict
  */
 
-declare export function CheckListPlugin(): null;
+declare export function CheckListPlugin({disableTakeFocusOnClick: boolean}): null;
```
etrepum: This type isn't the same as the typescript one because it isn't optional with all optional properties. I don't even think the syntax will pass because the argument isn't given a name

packages/lexical-list/src/index.ts
```diff
@@ -314,6 +314,10 @@ export const ListExtension = defineExtension({
   },
 });
 
+export type CheckListConfig = {
+  disableTakeFocusOnClick?: boolean;
+};
```
etrepum: config properties shouldn't be optional

```suggestion
export interface CheckListConfig {
  disableTakeFocusOnClick: boolean;
}
```

packages/lexical-list/src/checkList.ts
```diff
@@ -191,45 +270,71 @@ function handleCheckItemEvent(event: PointerEvent, callback: () => void) {
...
+ * Prevents default focus switch behavior
+ *
+ * @param event might be of type PointerEvent, MouseEvent, or TouchEvent, hence the generic Event type
+ *
+ */
+function handleSelectDefaults(
+  event: Event,
+  disableTakeFocusOnClick: boolean = false,
+) {
+  if (disableTakeFocusOnClick === true) {
```
etrepum: Using `=== true` for a boolean is just extra code for no reason

```suggestion
  if (disableTakeFocusOnClick) {
```

packages/lexical-list/src/checkList.ts
```diff
@@ -191,45 +270,71 @@ function handleCheckItemEvent(event: PointerEvent, callback: () => void) {
...
-  });
+/**
+ * Prevents default focus switch behavior
+ *
+ * @param event might be of type PointerEvent, MouseEvent, or TouchEvent, hence the generic Event type
+ *
+ */
+function handleSelectDefaults(
+  event: Event,
+  disableTakeFocusOnClick: boolean = false,
```
etrepum: No need to make this optional with a default, it's always explicitly passed.

```suggestion
  disableTakeFocusOnClick: boolean,
```

packages/lexical-list/src/checkList.ts
```diff
@@ -191,45 +270,71 @@ function handleCheckItemEvent(event: PointerEvent, callback: () => void) {
...
+      : clientXInPixels > rect.left - clickAreaPadding &&
+        clientXInPixels < rect.left + beforeWidthInPixels + clickAreaPadding
   ) {
     callback();
   }
 }
 
-function handleClick(event: Event) {
-  handleCheckItemEvent(event as PointerEvent, () => {
+function handleClick(event: Event, disableFocusOnClick: boolean = false) {
```
etrepum:
```suggestion
function handleClick(event: Event, disableFocusOnClick: boolean) {
```

packages/lexical-list/src/checkList.ts
```diff
@@ -191,45 +270,71 @@ function handleCheckItemEvent(event: PointerEvent, callback: () => void) {
...
+
+            if ($isListItemNode(node)) {
+              if (disableFocusOnClick) {
+                domNode.focus();
+              }
+              node.toggleChecked();
+            }
+          },
+          {tag: tags},
+        );
```
etrepum:
```suggestion
        editor.update(
          () => {
            const node = $getNearestNodeFromDOMNode(domNode);

            if ($isListItemNode(node)) {
              if (disableFocusOnClick) {
                $addUpdateTag(SKIP_SELECTION_FOCUS_TAG);
                $addUpdateTag(SKIP_DOM_SELECTION_TAG);
                domNode.focus();
              }
              node.toggleChecked();
            }
          },
        );
```

packages/lexical-list/src/checkList.ts
```diff
@@ -142,21 +166,59 @@ export function registerCheckList(editor: LexicalEditor) {
...
+            capture: true,
+          },
+        );
       }
     }),
   );
 }
 
-function handleCheckItemEvent(event: PointerEvent, callback: () => void) {
+function handleCheckItemEvent(event: Event, callback: () => void) {
```
etrepum: It's better to use a more specific type here so narrowing will work without casts

```suggestion
function handleCheckItemEvent(event: MouseEvent | TouchEvent, callback: () => void) {
```

packages/lexical-list/src/checkList.ts
```diff
@@ -179,10 +241,27 @@ function handleCheckItemEvent(event: PointerEvent, callback: () => void) {
...
+  let clientX: number | null = null;
+  let pointerType: string | null = null;
+
+  if ('clientX' in event) {
+    clientX = (event as MouseEvent).clientX;
+  } else if ('touches' in event) {
+    const touches = (event as TouchEvent).touches;
+    if (touches.length > 0) {
+      clientX = touches[0].clientX;
+      pointerType = 'touch';
```
etrepum: These casts should be unnecessary if the function's type is more specific

```suggestion
    clientX = event.clientX;
  } else if ('touches' in event) {
    const touches = event.touches;
    if (touches.length > 0) {
      clientX = touches[0].clientX;
      pointerType = 'touch';
```

packages/lexical-list/src/checkList.ts
```diff
@@ -38,7 +40,29 @@ export const INSERT_CHECK_LIST_COMMAND: LexicalCommand<void> = createCommand(
...
+export function registerCheckList(
+  editor: LexicalEditor,
+  options?: {disableTakeFocusOnClick?: boolean},
+) {
+  let disableTakeFocusOnClick = false;
+  if (options) {
+    if (options.disableTakeFocusOnClick === true) {
+      disableTakeFocusOnClick = true;
+    }
+  }
```
etrepum:
```suggestion
  const disableTakeFocusOnClick = options && options.disableTakeFocusOnClick || false;
```

etrepum:
This feature still doesn't have any tests, when you add some I will take another look

rayterion:
> This feature still doesn't have any tests, when you add some I will take another look

I've added the test file. I took a little longer because i was trying to understand the standards for testing in Lexical, but i think i got it right.

packages/lexical-list/src/checkList.ts
```diff
@@ -304,7 +304,6 @@ function handleClick(
             if (disableFocusOnClick) {
               $addUpdateTag(SKIP_SELECTION_FOCUS_TAG);
               $addUpdateTag(SKIP_DOM_SELECTION_TAG);
-              domNode.focus();
```
rayterion: should not move focus at all here

packages/lexical-playground/__tests__/e2e/List.spec.mjs
```diff
@@ -75,6 +75,92 @@ test.beforeEach(({isPlainText}) => {
...
+    // Click on the checkbox marker area (left side of the item)
+    const box = await checklistItem.boundingBox();
+    await page.mouse.click(box.x + 10, box.y + box.height / 2, {
+      button: 'left',
+    });
+
+    // The item toggles and focus moves to the list item when enabled.
+    await expect(checklistItem).toHaveAttribute('aria-checked', 'true');
+    const isEditorFocused = await page.evaluate(() => {
+      return document.activeElement.getAttribute('role') === 'textbox';
```
rayterion: The editor's attr is "textbox". is there a better way to check if the editor is focused?

rayterion:
E2E tests added. Tackling those conflicts now.

packages/lexical-list/src/__tests__/unit/checkList.test.ts
rayterion, line 1: I feel like the E2E tests are doing a much better job of explaining the logic. Do we still need these unit tests?

etrepum: No, the unit tests are doing a lot of mocking and are probably not useful

packages/lexical-playground/__tests__/e2e/List.spec.mjs
etrepum: Can’t you compare the activeElement with the editor.getRootElement()?

etrepum:
Looks like maybe there's an issue with firefox here since those e2e tests are failing

rayterion:
> Looks like maybe there's an issue with firefox here since those e2e tests are failing

I'll probably start fixing it monday.

Once i finish fixing it, should i mention you somehow or maybe requesting a review...?

etrepum:
I should get a notification when new commits are pushed

PR #8104, opened 2026-01-31
v0.40.0

etrepum:
v0.40.0 is a monthly release primarily focusing on bug fixes and infrastructure such as the move from npm to pnpm.

Some very notable changes include:
- Some major fixes and refactoring to markdown parsing #8093 #8085
- Several of the most common utilities were moved from @lexical/utils to lexical #8106
- A cache coherency bug in `RootNode.getTextContent()` was fixed #8099

## Highlights

Monorepo:
- 🧹  #8035 Chore: Transition from npm to pnpm (also #8045 #8048 #8050 #8051 #8054 #8106) 

Link:
- ✅ #8070 Fix: Fix infinite transform loop in AutoLinkPlugin
- ✅ #8078 Fix: Toggle links with nested children

List:
- ✅ #8049 Fix: fix selection issue from list transform on linebreak
- ✅ #8068 Fix: Treat whitespace-only list items as empty when pressing Enter

Markdown:
- ✅ #8085 Fix: Fix incorrect format tag placement at link boundaries
- ✅ #8093 Fix: Replace regex-based format matching

Core:
- ✅ #8069 Fix: format removed on multi selection after replace
- ✅ #8099 Fix: Refactor RootNode.__cachedText computation for coherency

Table:
- ✅ #8076 Fix: Fix Ctrl+A to select all cells in table with merged cells
- ✅ #8081 Fix: Fix inconsistent multi-cell selection in 2x2 tables
- 🧹  #8088 Chore: Fix test for nested table pasting

Utils:
- 🧹 #8106 Chore: Move functions

React:
- ✅ #8062 Fix: Clear remote cursor immediately on collaborator refresh
- ✅ #8065 Fix: Fix cursor disappearing in Firefox when dragging blocks
- ✅ #8071 Chore: Expose onReposition prop on SelectionAlwaysOnDisplay

Playground:
- 🆕  #8043 Feature: Color table resize handle
- 🆕  #8042 Feature: Draggable block handle gliding effect
- ✅ #8052 Fix: Draggable handle and dropdown CSS zoom fix 
- 🆕 #8057 Feature: New Table Hover Actions Plugin
- 🆕 #8060 Feature: Column Sort for Basic Table
- 🆕 #8066 Feature: Add button shows the component picker
- ✅ #8087 Fix: Prevent code block line wrapping to keep line number

## What's Changed
* v0.39.0 by @etrepum in https://github.com/facebook/lexical/pull/8021
[...]

PR #8106, opened 2026-02-01
[lexical][lexical-utils] Chore: Use workspace:* for monorepo versions and fix some cyclic dependencies

etrepum:
## Description

This should fix the failures that are blocking the v0.40.0 release (#8104)

This fixes updateVersion to use workspace:* versions consistently. The normalization of these to exact versions should happen on publish by pnpm.

This moves a few very commonly used functions from @lexical/utils to lexical:
* mergeRegister
* addClassNames
* removeClassNames
* normalizeClassNames (internal)

It also removes some cyclic dependencies from @lexical/utils (it had depended on @lexical/table and @lexical/list which in turn depend on it, but this dependency is only used in tests which we don't need to declare in package.json)

## Test plan

Unit tests should still pass, also after running the increment version workflow

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -297,29 +297,29 @@ importers:
   packages/lexical-clipboard:
     dependencies:
       '@lexical/html':
-        specifier: 0.39.0
-        version: 0.39.0
+        specifier: workspace:*
+        version: link:../lexical-html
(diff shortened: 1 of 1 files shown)


ivailop7:
Approving to unblock, for when the failures get fixed.

PR #8107, opened 2026-02-02
[lexical-playground] Chore: Add (disabled) E2E tests for nested table drag-selection

randal-atticus:
## Description
I was working on a fix for #7154, that is, improving the selection behaviour for nested cells.

I gave up on the fix because the current selection handler in `LexicalTableSelectionHelpers.ts` is registered for _each_ table. As it stands, a selection in a nested cell is also considered a selection inside the parent cell. Both tables end up firing $setSelection and triggering each others `SELECTION_CHANGED_COMMAND`, and they end up fighting for control. This leads to flickering and the wrong element being selected, see:

https://github.com/user-attachments/assets/da3fba80-5ba2-4782-b15f-ca281d840adf

**I would like to refactor table selection to use a monolithic handler which is only registered and fired once.**

I propose merging these test to demonstrate the bug and then refactoring (with a fix for these tests) in another PR.

### Before

Visual demonstration of the E2E tests (if the `.fixme` is removed). Apologies for the poor quality, had to downscale to fit Github limits.

https://github.com/user-attachments/assets/d1568aec-d20a-4c15-8f77-4e94402ce156

### After

No changes to logic.

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -7574,165 +7574,171 @@ test.describe.parallel('Tables', () => {
     });
   });
 
-  test('Drag-select inside nested table to outside table (backwards) selects nested table only', async ({
-    page,
-    isPlainText,
-    isCollab,
-  }) => {
-    test.skip(isPlainText);
-    await initialize({hasNestedTables: true, isCollab, page});
+  test.describe.fixme(
+    'Drag-select nested table tests',
+    'These tests are all erroneously selecting the parent cell',
+    () => {
+      test('Drag-select out of nested table (backwards) does not select parent cell', async ({
+        page,
+        isPlainText,
+        isCollab,
+      }) => {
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/__tests__/utils/index.mjs
```diff
@@ -829,7 +829,7 @@ export async function dragMouse(
   if (mouseDown) {
     await page.mouse.down();
   }
-  await page.mouse.move(toX, toY, slow ? 10 : 1);
```
randal-atticus, line 874: I think this argument was wrong? The dragging behaviour was only testable once I used the `steps` parameter instead of passing the number in. https://playwright.dev/docs/api/class-mouse#mouse-move

I changed the function to take the number of steps rather than a boolean, and reduced the "slow" value from 10 to 5 to speed up some tests (at `10` steps, the test takes an additional 200ms)

etrepum:
Test still appear to be failing here

randal-atticus:
Will just work on the updated selection handler rather than adding tests for the (broken) behaviour

PR #8113, opened 2026-02-03
[lexical] Chore: Stabilize playground collab WebKit E2E test waits

### LINKED ISSUE #8112: Flaky WebKit collab E2E tests in playground (sleep(50) too short)
Hi! I'm seeing intermittent failures in the Lexical playground collab E2E suite on WebKit that usually pass on re-run. This looks like timing/synchronization rather than a real regression.

I noticed this while upgrading svelte-lexical (a Svelte port of the playground) through the 0.33 to 0.40 chain. The webkit collab job failed sporadically and passed on re-run.

In packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs there are a couple of places that use very short fixed sleeps, e.g. await sleep(50) around operations like Undo / delete + cross-frame actions.

Examples:
- Test: Remove dangling text from YJS when there is no preceding text node (there is a sleep(50) right before clicking Undo)
- Test: $handleNormalizationMergeConflicts handles nodes that have been reparented (there is a sleep(50) between delete in left frame and delete in right frame)

Hypothesis: 50ms is sometimes not enough for Yjs merge / editor state propagation on CI WebKit, so the next action/assertion happens too early.

Suggested fix: Replace the fixed sleep(50) with an explicit wait for the expected state before proceeding. For example, wait via expect(async () => { ...assertHTML/assertSelection... }).toPass(...) or wait for a UI/editor condition that implies the merge/normalization has completed. If an explicit condition is hard to define, increase the sleep conservatively, but I think explicit waits would be more stable long-term.

If you want, I can open a PR with a minimal change (only replacing those sleeps with explicit waits), keeping the rest of the test logic intact.

screenfluent:
Fixes #8112.

This replaces a couple of short fixed sleeps in packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs with explicit waits:

- Wait for the Undo button to be enabled before clicking (removes sleep(50)).
- Wait for both editors to converge after the first delete before performing the second cross-frame delete (removes sleep(50)).

No behavior assertions changed; this only makes the existing test steps more deterministic.

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Collaboration.spec.mjs
@@ -255,7 +255,9 @@ test.describe('Collaboration', () => {
     );
 
     // Left collaborator undoes their text in the second paragraph.
-    const undoButton = page.frameLocator('iframe[name="left"]').getByLabel('Undo');
+    const undoButton = page
+      .frameLocator('iframe[name="left"]')
+      .getByLabel('Undo');
     await expect(undoButton).toBeEnabled();
     await undoButton.click();
 


etrepum:
This PR can't be considered without acceptance of the CLA, and it seems to be failing the prettier ci-check. `pnpm run prettier:fix` should address that issue.

screenfluent:
> This PR can't be considered without acceptance of the CLA, and it seems to be failing the prettier ci-check. `pnpm run prettier:fix` should address that issue.

Hey @etrepum thanks! CLA is accepted and the Prettier check is fixed now, so the PR should be ready for review.

etrepum (approved):

PR #8114, opened 2026-02-04
[lexical-extension] Feature: Implement DecoratorTextExtension applying format to DecoratorTextNode

levensta:
## Description

Added a new extension `DecoratorTextExtension` which provides a new `DecoratorTextNode` class and the ability to apply text formatting to it. `DecoratorTextNode` — a new node extended from `DecoratorNode`. The implementation of the node is simple and is actually the same `DecoratorNode`, but with formatting methods exactly the same as those of `TextNode`.

This feature can be useful for decorator nodes that need to be rendered in the middle of text and behave like text when the user wants to convert a part in a some format. Support for specific formats and how they affect the visual component of the decorator is left up to the user. As an example, I made how this might work for DateTimeNode in lexical-playground

### Before

https://github.com/user-attachments/assets/ae85ebc4-de20-4dea-90b2-82048bbcb53c

### After

https://github.com/user-attachments/assets/44a54dd8-bd0b-4731-b51b-c870e6d32cf4

### DIFF
--- a/packages/lexical-playground/src/nodes/DateTimeNode/DateTimeComponent.tsx
+++ b/packages/lexical-playground/src/nodes/DateTimeNode/DateTimeComponent.tsx
@@ -42,13 +42,13 @@ import {DayPicker} from 'react-day-picker';
 
 import {$isDateTimeNode, type DateTimeNode} from './DateTimeNode';
 
-const FORMAT_CLASSES = {
-  [IS_BOLD]: 'bold',
-  [IS_HIGHLIGHT]: 'highlight',
-  [IS_ITALIC]: 'italic',
-  [IS_STRIKETHROUGH]: 'strikethrough',
-  [IS_UNDERLINE]: 'underline',
-};
+const FORMAT_CLASSES = [
+  [IS_BOLD, 'bold'],
+  [IS_HIGHLIGHT, 'highlight'],
+  [IS_ITALIC, 'italic'],
+  [IS_STRIKETHROUGH, 'strikethrough'],
+  [IS_UNDERLINE, 'underline'],
+] as const;
 
 const userTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
 
@@ -200,8 +200,8 @@ export default function DateTimeComponent({
   };
 
   const classNames = ['dateTimePill'];
-  for (const [flag, className] of Object.entries(FORMAT_CLASSES)) {
-    if (format & Number(flag)) {
+  for (const [flag, className] of FORMAT_CLASSES) {
+    if (format & flag) {
       classNames.push(className);
     }
   }


packages/lexical-react/flow/LexicalDecoratorTextNode.js.flow
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+/**
+ * LexicalDecoratorTextNode
+ */
```
levensta: This was generated by the `update-flowconfig` command, but I didn't understand whether I needed to add the types manually. The `build-types` command returned an error. Perhaps it would be worth adding a brief explanation to CONTRIBUTING.md of how to correctly add new exportable things to packages 🤔

etrepum: `pnpm run update-flowconfig` only creates the files, it doesn't try and write the flow types for you.

`pnpm run build-types` is unrelated to flow, it only runs typescript. it does look like there's a problem with it worth fixing or filing an issue for, it tries to type check the build products in a way that doesn't make sense, so `pnpm build-types` will fail if you have run `pnpm run build` without `pnpm run clean` or equivalent.

levensta: >  it does look like there's a problem with it worth fixing or filing an issue for

I attempted to run the command on node v24.11.1. After switching to v22.22.0, `pnpm run build-types` executed successfully

levensta:
> I haven't taken a very close look but this does seem like a lot of code for this feature, without any tests, and it appears to be failing existing tests. I suspect that it could be simpler.

I simplified the implementation a little and fixed fundamental errors that caused the tests to fail

packages/lexical-playground/__tests__/e2e/TextFormatting.spec.mjs
```diff
@@ -1224,10 +1224,12 @@ test.describe.parallel('TextFormatting', () => {
       await toggleBold(page);
       await assertHTML(
         page,
+        // After formatting the text, the selection will be reset from the decorator node,
+        // so it will retain its previous format when toggleBold is triggered again
```
levensta, line 1228: I guess this is related to the TextPointType update in selection.formatText https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalSelection.ts#L1319-L1325

https://github.com/user-attachments/assets/f6c850b4-f5c6-4cd2-9aaf-ff15ccc588e6

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -46,39 +49,86 @@ export type SerializedDateTimeNode = Spread<
...
+    };
   }
   const gDocsDateTimePayload = domNode.getAttribute('data-rich-links');
   if (!gDocsDateTimePayload) {
     return null;
   }
   const parsed = JSON.parse(gDocsDateTimePayload);
-  const parsedDate = Date.parse(parsed?.dat_df?.dfie_dt || '');
+  const parsedDate =
+    parsed?.dat_df?.dfie_ts.tv.tv_s * 1000 ||
```
etrepum: This parser will throw if the structure is not correct

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -111,11 +161,31 @@ export class DateTimeNode extends DecoratorNode<JSX.Element> {
...
+    }
+    if (this.hasFormat('italic')) {
+      textDom = wrapElementWith(textDom, 'i');
+    }
+    if (this.hasFormat('strikethrough')) {
+      textDom = wrapElementWith(textDom, 's');
+    }
+    if (this.hasFormat('underline')) {
+      textDom = wrapElementWith(textDom, 'u');
+    }
```
etrepum: This seems like functionality that should be provided by the DecoratorTextNode module

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -46,39 +49,86 @@ export type SerializedDateTimeNode = Spread<
...
+  }
+  if (hasLinethroughTextDecoration && !lexicalNode.hasFormat('strikethrough')) {
+    lexicalNode.toggleFormat('strikethrough');
+  }
+  if (hasItalicFontStyle && !lexicalNode.hasFormat('italic')) {
+    lexicalNode.toggleFormat('italic');
+  }
+
+  return lexicalNode;
+}
```
etrepum: This could be provided by DecoratorTextNode

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
levensta: I deliberately left `DecoratorTextNode` without a specific DOM import/export implementation so that it could serve as a tool for setting the format of the node itself, while the specific implementation may be in the extended node if necessary. Similarly, `DecoratorBlockNode` does not necessarily need to have a specific HTML serialization implementation.

In `DateTimeNode`, I included an example of how this might look. It does not support all formats; for example, subscript and superscript on a date would look strange, so they are not displayed in the render or exportDOM, but on another user custom node it might be okay, so the choice is up to the user. 

The format tag itself is inside the date tag, which is the ideal place in this case, since the child is guaranteed to be text. In more complex cases, the format may be needed outside the tag, in which case the `importDOM` implementation will be completely different, so I would leave `DecoratorTextNode` as is and serialization in `DateTimeNode` for inspiration

ivailop7:
It would be nice if the font size and font color attributes worked as well.

levensta:
> It would be nice if the font size and font color attributes worked as well.

Playground sets styles through the [`$patchStyleText` function](https://github.com/facebook/lexical/blob/main/packages/lexical-selection/src/lexical-node.ts#L307-L356), which works with `TextNode` or `ElementNode`. A primitive solution would be to make this function work with the new `DecoratorTextNode` type, but I think it's wrong to pull the dependency from the lexical-react package into the lower-level lexical-selection

I think the right way is to add style management to the playground via a separate plugin and the `PATCH_TEXT_STYLE_COMMAND` command, as done in [one of the examples](https://github.com/facebook/lexical/blob/main/examples/node-state-style/src/styleState.ts#L458-L467). Then override the command handler in the `TextWithFormattedContents` component, which I implemented in this PR, and apply styles to `DecoratorTextNode`.

Anyway, I would suggest doing it in a separate PR, since supporting styles for non-text nodes is a non-trivial feature, especially if you need to come up with a mechanism to limit the set of possible styles for a specific node/class.

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
etrepum: I'm not saying that this functionality should be the exportDOM/importDOM methods, just that it should be in some common module. Could be a function. Most implementations will use the same code, no reason to copy and paste it into every implementation. In the cases where it makes sense to parameterize it there could be a data structure argument for supported formats

levensta: I moved these functions to the DecoratorTextNode file and added support for all formats with the ability to override the tag map

packages/lexical-react/src/LexicalTextWithFormattedContents.tsx
```diff
@@ -0,0 +1,93 @@
...
+type Props = Readonly<{
+  children: ReactNode;
+  nodeKey: NodeKey;
+  className: Readonly<{
+    base: string;
+    focus: string;
+  }>;
+}>;
+
+export function TextWithFormattedContents({
```
etrepum: This doesn't look like the best separation of concerns, it looks like it would register a command for each one of these nodes in the document rather than consolidating it to a single plugin. I know there is some old code that does this already but it's not a good way to do things. HorizontalRuleExtension is an example of something recent that consolidates the command listeners.

packages/lexical-react/src/LexicalDecoratorTextNode.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  updateDOM(): false {
+    return false;
+  }
+}
+
+export function $createDecoratorTextNode() {
+  return $applyNodeReplacement(new DecoratorTextNode());
+}
+
```
etrepum: No need for this since it's not useful to create them directly

packages/lexical-react/src/LexicalDecoratorTextNode.ts
```diff
@@ -0,0 +1,211 @@
...
+} from 'lexical';
+
+export type SerializedDecoratorTextNode = Spread<
+  {
+    format: number;
+  },
+  SerializedLexicalNode
+>;
+
+export class DecoratorTextNode extends DecoratorNode<JSX.Element> {
```
etrepum: This can be simplified by using $config and NodeState https://lexical.dev/docs/concepts/nodes#creating-custom-nodes-with-config-and-nodestate

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -111,11 +136,17 @@ export class DateTimeNode extends DecoratorNode<JSX.Element> {
...
+    let textDom: HTMLElement | Text = document.createTextNode(
+      getDateTimeText(this.getDateTime()),
+    );
     element.setAttribute(
       'data-lexical-datetime',
       this.getDateTime()?.toString() || '',
     );
+
+    textDom = applyFormatToDom(this, textDom, tagToFormat);
+    element.appendChild(textDom);
```
etrepum: No real need to re-assign textDOM here

```suggestion
    element.appendChild(applyFormatToDom(this, textDom, tagToFormat));
```

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeNode.tsx
```diff
@@ -111,11 +136,17 @@ export class DateTimeNode extends DecoratorNode<JSX.Element> {
 
   exportDOM(): DOMExportOutput {
     const element = document.createElement('span');
-    element.textContent = getDateTimeText(this.getDateTime());
+    let textDom: HTMLElement | Text = document.createTextNode(
```
etrepum:
```suggestion
    const textDom = document.createTextNode(
```

packages/lexical-react/src/LexicalTextWithFormattedContents.tsx
levensta: Thanks for the tip. I rewrote it from a component to an extension and removed the className setting because with the current implementation it is redundant and much easier to set all the necessary classes inside the component itself

packages/lexical-extension/src/DecoratorTextExtension.ts
```diff
@@ -0,0 +1,235 @@
...
+            } else {
+              const decoratorText = $findMatchingParent(
+                node,
+                $isDecoratorTextNode,
+              );
+              if (decoratorText !== null) {
+                decoratorText.toggleFormat(formatType);
+              }
+            }
+          }
```
etrepum: `$isDecoratorTextNode` will never return true for an `ElementNode` so there's no need to check parents

```suggestion
        if ($isNodeSelection(selection) || $isRangeSelection(selection)) {
          for (const node of selection.getNodes()) {
            if ($isDecoratorTextNode(node)) {
              node.toggleFormat(formatType);
            }
          }
        }
```

packages/lexical-playground/src/nodes/DateTimeNode/DateTimeComponent.tsx
```diff
@@ -181,11 +199,18 @@ export default function DateTimeComponent({
     });
   };
 
+  const classNames = [
+    'dateTimePill',
+    ...Object.entries(formatClassMap)
+      .filter(([flag]) => format & Number(flag))
+      .map(([, className]) => className),
```
etrepum: These can array returning methods be fused into a flatMap, although it would probably make more sense to have formatClassMap just be an array of pairs in the first place rather than converting it to one on demand every time.

```suggestion
      .flatMap(([flag, className]) => format & Number(flag) ? [className] : [])
```

or

```ts
// This would be top-level in the module
const FORMAT_CLASSES = [
  [IS_BOLD, 'bold'],
  [IS_HIGHLIGHT, 'highlight'],
  [IS_ITALIC, 'italic'],
  [IS_STRIKETHROUGH, 'strikethrough'],
  [IS_UNDERLINE, 'underline'],
] as const;

// …
const classNames = ['dateTimePill'];
for (const [flag, className] of FORMAT_CLASSES) {
  if (format & flag) {
    classNames.push(className);
  }
}
```

levensta: > although it would probably make more sense to have formatClassMap just be an array of pairs in the first place rather than converting it to one on demand every time

Agreed, done

PR #8115, opened 2026-02-05
[lexical] Feature: Support legacy 'align' attribute in ParagraphNode importDOM

Sa-Te:
## Summary

This PR fixes an issue where pasting legacy HTML content (e.g., from older editors or Word) containing the `align` attribute on `<p>` tags would lose its alignment information.

Previously, `ParagraphNode.importDOM` only checked `element.style.textAlign`. It now falls back to the `align` attribute if no CSS alignment is present.

## Details

- Modified `$convertParagraphElement` in `LexicalParagraphNode.ts` to check `element.getAttribute('align')` as a fallback.
- Ensures modern CSS `text-align` still takes priority over the legacy attribute.
- Added comprehensive unit tests covering:
- Legacy `<p align="right">`
- Modern `<p style="text-align: center">`
- Conflict resolution (CSS priority)

## Test Plan

- [x] Added new unit tests in `LexicalParagraphNode.test.ts`
- [x] Verified `pnpm run test-unit packages/lexical/src/nodes/__tests__/unit/LexicalParagraphNode.test.ts` passes.

### DIFF
--- a/packages/lexical/src/nodes/LexicalParagraphNode.ts
+++ b/packages/lexical/src/nodes/LexicalParagraphNode.ts
@@ -24,6 +24,7 @@ import type {
   SerializedElementNode,
 } from './LexicalElementNode';
 
+import {ELEMENT_TYPE_TO_FORMAT} from '../LexicalConstants';
 import {
   $applyNodeReplacement,
   getCachedClassNameArray,
@@ -176,7 +177,9 @@ function $convertParagraphElement(element: HTMLElement): DOMConversionOutput {
   if (node.getFormatType() === '') {
     const align = element.getAttribute('align');
     if (align) {
-      node.setFormat(align as ElementFormatType);
+      if (align && align in ELEMENT_TYPE_TO_FORMAT) {
+        node.setFormat(align as ElementFormatType);
+      }
     }
   }
   return {node};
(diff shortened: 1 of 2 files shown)


packages/lexical/src/nodes/LexicalParagraphNode.ts
```diff
@@ -170,6 +170,15 @@ function $convertParagraphElement(element: HTMLElement): DOMConversionOutput {
...
     setNodeIndentFromDOM(element, node);
   }
+
+  // Check legacy 'align' attribute
+  // Only use this if no format was set by CSS
+  if (node.getFormatType() === '') {
+    const align = element.getAttribute('align');
+    if (align) {
+      node.setFormat(align as ElementFormatType);
+    }
```
etrepum, line 183:
```suggestion
    if (align && align in ELEMENT_TYPE_TO_FORMAT) {
      node.setFormat(align);
    }
```

`setFormat` doesn't handle unvalidated inputs correctly, will set `__format` to `undefined` instead of `0`. This should either parse with `align in ELEMENT_TYPE_TO_FORMAT` and/or it should include an improvement to `setFormat` to handle invalid inputs more gracefully by using `0` in that situation.

```ts
  setFormat(type: ElementFormatType): this {
    self.__format = type !== '' ? ELEMENT_TYPE_TO_FORMAT[type] : 0;
    return this;
  }
```

etrepum (approved):

PR #8116, opened 2026-02-05
[lexical-markdown] Bug Fix: Fix nested fenced code blocks parsing and export

### LINKED ISSUE #8109: Bug: Nested fenced code blocks (4 backticks) are corrupted
Lexical version: indirect dependency (version not exposed by Sveltia CMS)

## Steps To Reproduce

1. Create or edit a markdown document in the Sveltia CMS editor.
2. Switch to markdown view.
3. Add nested fenced code blocks using 4 backticks to wrap content that contains a 3-backtick code block, e.g.:

`````markdown
````markdown
## Development Setup

After cloning, run:

```bash
pip install pre-commit
pre-commit install
```
````
`````

4. Save the markdown content.
5. Close the editor.
6. Reopen the same markdown document in the editor.
7. Observe the malformed markdown output.

Link to code example: https://playground.lexical.dev/#doc=H4sIAAAAAAAACtWVwW7CMAyGX6Uy14DKhtYtt0l7gEnsNk1qaEwbkTpV6rAh1HefUjYGnBCMiZ0S_7F_f744a0Bt2PkpK0aQa_DOcTyLyljtkUC-HgQaWRkLMhUwd75W3F9rpxEkUFQsCGh5ZaMAAhg_GCTkeR6DVRPlwmkcVqasrCkrBgFL9K1xBHIsYKu_bJKbQAUHxfG9E2cQ1MovtHunEzn6ZKuoDKrEnuTLxRrCmUe12DM4B3UwSJ5widY1NRInU-TQnEM9s65YHId8sZke54w-KawjQ6VIfCD530fK83ym2uqPxjiVsjFNYqhlZW3SeBwWrq4NXzv0FvSb_cqBf2-_XRLxOMbuTYA2HoueSVKw9qdVNDSkkTZtd9wO5tyuSgnbzXuCc_8l7dF10bvlqVqiBjnOsjSdZHc3k_txJqB1wRex7tmqVeldIL1bDeno9mGUQvcJoRgyfvsGAAA

Click the **Convert from Markdown** button

## The current behavior

When reopening a document that contains nested fenced code blocks (4 backticks wrapping a 3-backtick fenced block), the editor corrupts the structure: it misinterprets and breaks the nesting boundaries, leading to malformed markdown where code blocks are opened/closed incorrectly.

## The expected behavior

The nested fenced code block structure should be preserved exactly as it was entered, with the outer 4-backtick fence correctly containing the inner 3-backtick code block, following CommonMark rules for nested fences.

## Impact of fix

Severity: High for users documenting examples of code within markdown (e.g., docs, tutorials).

Frequency: Happens whenever nested code fences are used and the document is reopened.

Who would benefit: Content authors using advanced markdown features; anyone relying on accurate editing of code examples.

Metrics (if known): Affects all markdown fields in the CMS editor where nested code blocks are used.

### PR DESCRIPTION
**Pull request by abdulalim110**
## Description

**Current behavior:**

Nested fenced code blocks (e.g., using 4 backticks to wrap 3 backticks) are corrupted during the Markdown round-trip. The parser prematurely closes the code block at the first encounter of a 3-backtick sequence, causing the remaining content and the original closing fence to leak into the editor as plain text.

**Changes added by this PR:**

1. **State Preservation:** Introduced `codeFenceState` using `createState` to store the specific fence length (3, 4, 5+ backticks) used during import, ensuring the same length is used during export.
    
2. **Dynamic Fence Detection:** Implemented `handleImportAfterStartMatch` with a dynamic Regex to ensure a code block only closes when it finds a fence with a length matching the opening fence.
    
3. **Edge Case Fixes:**
    
    - Improved single-line code block detection to prevent content (like in ` ```Single line``` `) from being incorrectly parsed as a language attribute.
        
    - Added EOF (End of File) handling for unclosed code blocks to ensure they are correctly captured as code nodes instead of standard paragraphs.
        
    - Refined the `CODE` transformer Regex to properly separate the fence and language capture groups.
        

Closes #8109

### Before

Markdown content with nested fences would be rendered incorrectly in the Lexical editor. For example, a 4-backtick block would "break" at the inner 3-backtick block, leading to malformed output and extra escaped backticks ( `\```` `) on export.

### After

**Automated Tests:**

Added new unit tests in `LexicalMarkdown.test.ts`:

- `can round-trip nested fenced code blocks (4 backticks wrapping 3 backticks)`
    
- `can round-trip deeply nested fenced code blocks (5 backticks wrapping 4 backticks)`
    
- Verified that single-line blocks and incomplete/unclosed tags still pass and render correctly as per CommonMark expectations.
    

**Manual Verification:**

Tested in the playground by switching between Markdown and Rich Text views with nested code examples. The structure and backtick count are now perfectly preserved.

### DIFF
--- a/packages/lexical-list/src/LexicalListNode.ts
+++ b/packages/lexical-list/src/LexicalListNode.ts
@@ -146,6 +146,10 @@ export class ListNode extends ElementNode {
 
     $setListThemeClassNames(dom, config.theme, this);
 
+    if (prevNode.__start !== this.__start) {
+      dom.setAttribute('start', String(this.__start));
+    }
+
     return false;
   }
 
--- a/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
+++ b/packages/lexical-list/src/__tests__/unit/LexicalListItemNode.test.ts
@@ -9,6 +9,7 @@
 import {
   $createParagraphNode,
   $createRangeSelection,
+  $createTextNode,
   $getRoot,
   TextNode,
 } from 'lexical';
@@ -26,7 +27,7 @@ import {
   ListItemNode,
   ListNode,
 } from '../..';
-import {$handleIndent} from '../../formatList';
+import {$handleIndent, $handleListInsertParagraph} from '../../formatList';
 
 const editorConfig = Object.freeze({
   namespace: '',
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: packages/lexical-markdown/src/MarkdownTransformers.ts [hunk 1]
```diff
@@ -221,6 +221,15 @@ export const listMarkerState = createState('mdListMarker', {
   parse: (v) => (typeof v === 'string' && /^[-*+]$/.test(v) ? v : '-'),
 });
 
+export const codeFenceState = createState<string, string>('mdCodeFence', {
+  parse: (val) => {
+    if (typeof val === 'string' && val.startsWith('`')) {
+      return val;
+    }
```
etrepum: Might as well guarantee that it's a valid fence

```suggestion
    if (typeof val === 'string' && /^`{3,}$/.test(val) {
      return val;
    }
```

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -402,24 +411,95 @@ export const QUOTE: ElementTransformer = {
 
 export const CODE: MultilineElementTransformer = {
   dependencies: [CodeNode],
+
   export: (node: LexicalNode) => {
     if (!$isCodeNode(node)) {
       return null;
     }
     const textContent = node.getTextContent();
+    const fence = $getState(node, codeFenceState) || '```';
```
```suggestion
    const fence = $getState(node, codeFenceState);
```

It would be a good idea to also verify that textContent does not contain the fence and to compute a longer fence if necessary.

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -402,24 +411,95 @@ export const QUOTE: ElementTransformer = {
...
+      const endMatch = afterFence.match(singleLineEndRegex);
+      const content = afterFence.slice(0, afterFence.lastIndexOf(endMatch![0]));
+
+      const fakeStartMatch = [...startMatch];
+      fakeStartMatch[2] = '';
+
+      CODE.replace(
+        rootNode,
+        null,
+        fakeStartMatch as RegExpMatchArray,
```
```suggestion
      const fakeStartMatch: RegExpMatchArray = [...startMatch];
      fakeStartMatch[2] = '';

      CODE.replace(
        rootNode,
        null,
        fakeStartMatch,
```

etrepum (commented):
Overall this looks like a good direction but I think it can handle a few more edge cases

packages/lexical-markdown/src/MarkdownTransformers.ts
abdulalim110: Thanks for the review 

I applied the changes, but with two small adjustments:

1. I kept the `|| '```'` fallback because `$getState` might return `null` (e.g. for newly created nodes), and I need `fence` to be a guaranteed string for the `indexOf` check to be safe.
2. I used `let` instead of `const` so I can dynamically compute a longer fence length if the content contains backticks, as per your suggestion.

etrepum: $getState does not return null.

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -402,24 +411,103 @@ export const QUOTE: ElementTransformer = {
 
 export const CODE: MultilineElementTransformer = {
   dependencies: [CodeNode],
+
   export: (node: LexicalNode) => {
     if (!$isCodeNode(node)) {
       return null;
     }
     const textContent = node.getTextContent();
+    let fence = $getState(node, codeFenceState) || '```';
```
etrepum: The return value of `$getState(node, state)` will always return a value in the domain of `state.parse(…)`. The default value is `state.parse(undefined)` which in this case is `'```'`

```suggestion
    let fence = $getState(node, codeFenceState);
```

packages/lexical-markdown/src/MarkdownTransformers.ts
abdulalim110: ah, thanks for the clarification 

I missed that internal detail about $getState.

I've removed `|| '```'` and pushed the update

abdulalim110:
@etrepum I've added the test case (`computes fence dynamically...`) as requested.

this PR now addresses all feedback (redundant fallback removed + new test).

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -985,6 +985,106 @@ describe('Markdown', () => {
...
+    });
+
+    editor.update(
+      () => {
+        // Create a CodeNode without setting fence state (uses default ```)
+        const codeBlockNode = $createCodeNode('markdown');
+        // Content contains ``` which conflicts with default fence
+        const textNode = $createTextNode('```js\nconsole.log("hello");\n```');
+        codeBlockNode.append(textNode);
+        $getRoot().append(codeBlockNode);
```
etrepum: This wouldn't be the representation, new lines are `$createLineBreakNode()`, they are never embedded in a TextNode. RangeSelection.insertRawText can do this parsing.

```suggestion
        $getRoot()
          .append(codeBlockNode)
          .selectEnd()
          .insertRawText('```js\nconsole.log("hello");\n```');
```

PR #8118, opened 2026-02-06
[lexical] Bug Fix: Remove empty parent node in nested list

### LINKED ISSUE #8117: Bug: Nested list not remove when delete selection
Lexical version:
 0.40.0
## Steps To Reproduce

1. Insert a list node with a nested list.
2. Select the entire nested list and part of the parent list.
3. Press Backspace to delete the selected content.
The nested list is not fully removed; instead, it shifts forward.
Image

Link to code example:

## The current behavior
The nested list is not removed
## The expected behavior
The nested list is emoved

Senasiko:
## Description

This PR fixes an issue where removing a range that contains a nested list leaves behind an extra list item.
Closes #8117

## Test plan

image

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -206,8 +206,8 @@ const UNORDERED_LIST_REGEX = /^(\s*)[-*+]\s/;
 const CHECK_LIST_REGEX = /^(\s*)(?:[-*+]\s)?\s?(\[(\s|x)?\])\s/i;
 const HEADING_REGEX = /^(#{1,6})\s/;
 const QUOTE_REGEX = /^>\s/;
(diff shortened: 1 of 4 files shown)


packages/lexical/src/caret/LexicalCaretUtils.ts
```diff
@@ -344,7 +344,18 @@ export function $removeTextFromCaretRange<D extends CaretDirection>(
...
+    focusBlock.remove(true);
+    while (parent) {
+      if (parent.isEmpty()) {
+        const element = parent;
+        parent = parent.getParent();
+        element.remove(true);
+      } else {
+        break;
+      }
+    }
```
etrepum: Could be a simpler loop here
```suggestion
    while (parent && parent.isEmpty()) {
      const element = parent;
      parent = parent.getParent();
      element.remove(true);
    }
```

PR #8121, opened 2026-02-07
[lexical] Bug Fix: Do not move anchor when inserting composition start char

### LINKED ISSUE #8119: Bug: composition breaks when replacing selection starting with formatted text
Lexical version: 0.40.0

## Steps To Reproduce

1. Type "hello world" in the editor
2. Apply a format only to the first word
3. Select the entire string
4. Start typing a composition character(e.g. Korean '가' by typing 'ㄱ' then 'ㅏ')

Link to code example: https://playground.lexical.dev/#doc=H4sIAAAAAAAAE52QQUvEQAyF_4q-87BMpd3F-QGehR7Fw9CJ24HsTEnT1bL0v8t0lXVFRDyF90K-l-QEClGztOqV4E6QnLXUro8chBLc0zcRSH1kOGvwkuXgFa4yOORAcEjFYRiMOnMxYKD0pnDoiTkXOQ-lsboGR5Ix5gRXLeZHtv0L--Y1C4fb3-jPBiEKdbrqNDFfMgoppkDpnPfBGLz4vfihvwKdMx8ulxfZfm70j5j141erLgbsR239kQJctdvZuqmbZnu_vTMY8yRdmXtkP-8lTyl8nYbd1HZjsbwDTqVV89oBAAA

## The current behavior

https://github.com/user-attachments/assets/18085240-66ac-46a1-a0fc-e2576d78407b

The first composition input disappears, characters don't combine correctly.

## The expected behavior

The entire selection should be replaced by the first stroke of the composition, and the state should remain active to complete the character

kimseongyu:
### Current Behavior

Lexical inserts a `COMPOSITION_START_CHAR` to ensure the browser correctly handles IME composition. However, when this character is inserted, the selection anchor moves forward and first composition stroke is in a selected state. Windows Chrome manages to continue the composition session, But macOS Chrome drops the first stroke and composing is failed.

### Changes in This PR

- Anchor position is continued when the `COMPOSITION_START_CHAR` is inserted at the beginning of a composition session.
- After composition, Clean up the `COMPOSITION_START_CHAR` in the text and reconcile the anchor and focus positions appropriately.

Closes #8119

### Before

https://github.com/user-attachments/assets/b584c35e-149d-4feb-bd5e-5d3886c59592

### After

https://github.com/user-attachments/assets/859e3add-3639-46eb-8242-91e106dbc795

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -28,13 +28,7 @@ import type {
 import type {RootNode} from './nodes/LexicalRootNode';
 
 import {CAN_USE_DOM} from 'shared/canUseDOM';
-import {
-  IS_APPLE,
-  IS_APPLE_WEBKIT,
-  IS_FIREFOX,
-  IS_IOS,
-  IS_SAFARI,
-} from 'shared/environment';
+import {IS_APPLE, IS_APPLE_WEBKIT, IS_IOS, IS_SAFARI} from 'shared/environment';
 import invariant from 'shared/invariant';
 
 import {
@@ -776,9 +770,15 @@ export function $updateTextNodeFromDOMContent(
     const isComposing = node.isComposing();
     let normalizedTextContent = textContent;
 
-    if (IS_FIREFOX || compositionEnd) {
-      const charsToStrip = [COMPOSITION_START_CHAR, COMPOSITION_SUFFIX];
-      for (const char of charsToStrip) {
+    if (isComposing || compositionEnd) {
+      if (textContent.endsWith(COMPOSITION_SUFFIX)) {
+        normalizedTextContent = textContent.slice(
+          0,
+          -COMPOSITION_SUFFIX.length,
+        );
+      }
+      if (compositionEnd) {
+        const char = COMPOSITION_START_CHAR;
         let index;
         while ((index = normalizedTextContent.indexOf(char)) !== -1) {
           normalizedTextContent =


etrepum:
Looks like this is failing e2e tests

kimseongyu:
Thank you!

Before I finalize the fix, I have a question regarding a behavior I observed in the E2E test(Can type Hiragana via IME part way through a mention).

In this test case, when the Hiragana character 'す' is entered, a COMPOSITION_SUFFIX (an invisible character) is inserted. I noticed that when I try to move the cursor from 'す' back to the 'e' in "Luke", I have to press the arrow key twice to pass the invisible character.

Is this behavior intentional or unintended side effect? If it is a side effect, I am going to update the E2E test accordingly.

- [E2E test link](https://github.com/facebook/lexical/blob/70c1d237052b29ce06207e8b5aad092369c225a3/packages/lexical-playground/__tests__/e2e/Composition.spec.mjs#L744)
- [playground](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE52QQU7DMBBFr1L9tVU5UDWSD8CKBVKWiIUVD60V144mTmlURWqy5QDA9chFkFNQKUtWM380f_7THEHGxsBF1JGgjuAQYqrl1jrD5KEe_whDUVsHJQWeA-90nNtdMAQFnyYOAk3sXBpAINIhQuG-rWjxeXqdho9peFtM4zgN79MwFlX3ol1FnFa7Oplmh8CeuLHBQ2X9k4CxTGWctW-du6SnDOsN-TPJ941as96wrrdXh840dxfuJIsf1n_EzP-6Qu0FnG5iofdkoLI8l-vbbCXXWX4j0ISWy-R7cLrbcGi9-e2GXK7kUqL_AsaPXTqYAQAA)

etrepum:
It’s only intentional if that’s how IME normally works in native text fields.

kimseongyu:
Lexical [inserts](https://github.com/facebook/lexical/blob/70c1d237052b29ce06207e8b5aad092369c225a3/packages/lexical/src/LexicalEvents.ts#L1054) COMPOSITION_SUFFIX to prevent input errors during composition. However, it is currently only [removed](https://github.com/facebook/lexical/blob/70c1d237052b29ce06207e8b5aad092369c225a3/packages/lexical/src/LexicalUtils.ts#L755) from the end of the text, leaving intermediate ones behind. These suffixes cause unexpected behavior. I will update the related e2e tests to ensure all instances are properly cleaned up.

kimseongyu:
The test is passed. But there is incorrectly behavior in the firefox when input the composition text. So, I am working on fix it.

packages/lexical/src/LexicalEvents.ts
```diff
@@ -1005,7 +1007,16 @@ function $handleInput(event: InputEvent): boolean {
...
+      if (editor.isComposing()) {
+        editor.update(
+          () => {
+            dispatchCommand(editor, CONTROLLED_TEXT_INSERTION_COMMAND, data);
+          },
+          {tag: HISTORY_MERGE_TAG},
+        );
+      } else {
+        dispatchCommand(editor, CONTROLLED_TEXT_INSERTION_COMMAND, data);
+      }
```
etrepum, line 1014: nesting editor.update is almost never what you want, it will run this code asynchronously (after the current update). Does this not work?

```suggestion
      if (editor.isComposing()) {
        $addUpdateTag(HISTORY_MERGE_TAG);
      }
      dispatchCommand(editor, CONTROLLED_TEXT_INSERTION_COMMAND, data);
```

etrepum:
At least with Chrome on macOS it looks like the cursor is in the wrong place when doing composition with this PR (compared to the current playground and Apple's TextEdit input behavior) - this PR is on the left

https://github.com/user-attachments/assets/9365ba9b-19b1-4bcd-8811-3931fef819b8

kimseongyu:
Thanks for the catch! I'm already on it. The bug manifests as a simple position error on macOS, but on Windows, it causes false input position when formatting is applied during composition. I'll include the fix for this in my next push with your previous review.

etrepum:
You should be able to cherry-pick this commit which may include changes that I didn't mention in review comments https://github.com/facebook/lexical/pull/8133/changes/8d35f04712c4749e053609810323d5cf9916b160

kimseongyu:
@etrepum  I resolved all issue in this PR and the tests are all passes!


https://github.com/user-attachments/assets/4fe514c1-fd3b-4476-b964-006b18689e2d

etrepum (approved):
Code looks good, thank you! The lexical team doesn't currently have any active contributors with much IME expertise so this is very helpful

PR #8122, opened 2026-02-08
[lexical-extension][lexical-rich-text][lexical-react] Feature: Support configuration of indentable nodes

levensta:
## Breaking Change

`TabIndentationExtension` and  `TabIndentationPlugin` now enforce max indentation level on a per-block basis. Previously, any block at max indentation in the selection would prevent all indentation of the selection.

## Description

I added a new option to the `TabIndentation` plugin and extension. The essence of the feature is to provide an interface for a limited selection of classes that can be indented by pressing Tab.

Although it is possible to override the `canIndent` method for classes that do not need indents, this generates a lot of boilerplate code. For example, in a real-world case, you may need to be able to set nested lists using Tab, but not set indents for all other elements. To do this, you would need to extend each class used from lexical to override only one method, `canIndent` With the new option in the plugin, this task is simplified

It is now also possible to set indents for each node within a selection independently of each other. If one of the nodes has the maximum indent value, the indent will be added to the others. Previously, Tab did not work for any nodes if there was at least one maximum value among the nodes.

### Before

All elements can be indented. If one of the selected nodes has the maximum indentation, no indentation will be set for the other elements.

https://github.com/user-attachments/assets/1eb366b7-1e6e-4e42-aea4-0f89f31e5352

### After

All elements can be indented. Nodes with maximum indentation are ignored, while the rest elements are indented

https://github.com/user-attachments/assets/05d436ea-33be-4c03-a8c5-d4c0791558f5

If indents are only needed for specific nodes. For example, for lists

```jsx
<TabIndentationPlugin maxIndent={7} $canIndent={$isListItemNode} />
```

https://github.com/user-attachments/assets/aadc06ae-029e-4b1f-948c-3e64d061b23d

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Autocomplete.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Autocomplete.spec.mjs
@@ -28,49 +28,53 @@ test.describe('Autocomplete', () => {
   test.beforeEach(({isCollab, page}) =>
     initialize({isAutocomplete: true, isCollab, page}),
   );
-  test('Can autocomplete a word', async ({page, isPlainText}) => {
-    await focusEditor(page);
-    await page.keyboard.type('Sort by alpha');
-    await sleep(500);
-    await assertHTML(
-      page,
-      html`
-        <p class="PlaygroundEditorTheme__paragraph" dir="auto">
-          <span data-lexical-text="true">Sort by alpha</span>
-          <span
-            class="PlaygroundEditorTheme__autocomplete"
-            style="font-size: 15px"
-            data-lexical-text="true">
-            betical (TAB)
-          </span>
-        </p>
-      `,
-      html`
-        <p class="PlaygroundEditorTheme__paragraph" dir="auto">
-          <span data-lexical-text="true">Sort by alpha</span>
-          <span
-            class="PlaygroundEditorTheme__autocomplete"
-            style="font-size: 15px; display: none"
-            data-lexical-text="true">
-            betical (TAB)
-          </span>
-        </p>
-      `,
-    );
-    await page.keyboard.press('Tab');
-    await page.keyboard.type(' order:');
-    await assertHTML(
-      page,
-      html`
-        <p class="PlaygroundEditorTheme__paragraph" dir="auto">
-          <span data-lexical-text="true">Sort by alpha</span>
-          <span style="font-size: 15px" data-lexical-text="true">
-            betical order:
-          </span>
-        </p>
-      `,
-    );
-  });
+  test(
+    'Can autocomplete a word',
+    {tag: '@flaky'},
+    async ({page, isPlainText}) => {
+      await focusEditor(page);
+      await page.keyboard.type('Sort by alpha');
+      await sleep(500);
+      await assertHTML(
+        page,
+        html`
+          <p class="PlaygroundEditorTheme__paragraph" dir="auto">
+            <span data-lexical-text="true">Sort by alpha</span>
+            <span
+              class="PlaygroundEditorTheme__autocomplete"
+              style="font-size: 15px"
+              data-lexical-text="true">
+              betical (TAB)
+            </span>
+          </p>
+        `,
+        html`
+          <p class="PlaygroundEditorTheme__paragraph" dir="auto">
+            <span data-lexical-text="true">Sort by alpha</span>
+            <span
+              class="PlaygroundEditorTheme__autocomplete"
+              style="font-size: 15px; display: none"
+              data-lexical-text="true">
+              betical (TAB)
+            </span>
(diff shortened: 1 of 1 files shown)


levensta:
> I would suggest an even simpler approach where something like `$canIndent: (node: ElementNode) => boolean` is the configuration with a default implementation of `(node) => node.canIndent()` and then the user can make any possible override of behavior from there.

Good idea. I did it and also extended the current e2e test for indentation, since now the maximum indentation level can be set by selected nodes independently of each other. I think it looks expected and natural, watch the demo video in the updated description

packages/lexical-extension/src/TabIndentationExtension.ts
```diff
@@ -59,9 +65,14 @@ function $indentOverTab(selection: RangeSelection): boolean {
...
 }
 
+export type CanIndentPredicate = (node: ElementNode) => boolean;
+
 export function registerTabIndentation(
   editor: LexicalEditor,
   maxIndent?: number | ReadonlySignal<null | number>,
+  $canIndent: CanIndentPredicate | ReadonlySignal<CanIndentPredicate> = (
+    node,
+  ) => node.canIndent(),
```
etrepum: Better to define this default function at the top-level with a name since it's used twice (here and in the config)

packages/lexical-react/src/LexicalTabIndentationPlugin.tsx
```diff
@@ -17,11 +20,21 @@ export {registerTabIndentation};
  * recommend using this plugin as it could negatively affect accessibility for keyboard
  * users, causing focus to become trapped within the editor.
  */
-export function TabIndentationPlugin({maxIndent}: {maxIndent?: number}): null {
+export function TabIndentationPlugin({
+  maxIndent,
+  $canIndent = (node) => node.canIndent(),
```
etrepum:
```suggestion
  $canIndent,
```

This default comes from elsewhere, no need to define it again

packages/lexical-rich-text/src/index.ts
```diff
@@ -514,7 +514,7 @@ export function eventFiles(
   return [hasFiles, Array.from(dataTransfer.files), hasContent];
 }
 
-function $handleIndentAndOutdent(
+export function $handleIndentAndOutdent(
```
etrepum: Since this is public API now it should probably be given some docs

levensta: done

etrepum:
I think the integration tests are failing because this introduced a circular dependency from extension -> rich-text -> dragon -> extension

levensta:
> I think the integration tests are failing because this introduced a circular dependency from extension -> rich-text -> dragon -> extension

thanks, fixed

levensta:
> Can you update the e2e test to preserve the existing tests? It's difficult to see if this is backwards compatible or not because it looks like the tests have also changed in-place.

The test had to be changed anyway because it checked the old behavior of the plugin and failed after the changes. Previously, it was expected that none of the elements would set indentation if at least one of them reached the maximum value


Maybe this workaround will help you 😅
Try copying the string from assertHTML and pasting it as `text/html` into the preview environment.

- old test https://github.com/facebook/lexical/blob/main/packages/lexical-playground/__tests__/e2e/Indentation.spec.mjs#L704
- new test https://github.com/levensta/lexical/blob/nodes-for-tabindentation-plugin/packages/lexical-playground/__tests__/e2e/Indentation.spec.mjs#L779-L929

<details><summary>console code</summary>

```js
async function copyHtmlToClipboard(html) {
  const blob = new Blob([html], { type: "text/html" });
  const data = new ClipboardItem({ "text/html": blob });

  await navigator.clipboard.write([data]);
  console.log("HTML сopied as text/html");
}


// paste html here
const html = `

`;

// focus on the document during the timer
setTimeout(() => copyHtmlToClipboard(html), 1500); 

```

</details

etrepum:
My point here is that we should verify that if you don’t add any new settings then the behavior shouldn’t change.

levensta:
> My point here is that we should verify that if you don’t add any new settings then the behavior shouldn’t change.

That's the point. Without changing the settings, the test fails with an error because it checked different behavior.

https://github.com/facebook/lexical/actions/runs/21797687397/job/62887841955#step:11:293

I can revert to the previous behavior and add a check for the existence of the maximum indent value among the selected elements so that the test does not fail, but I think from a UX perspective it is better to give the user the ability to indent all remaining nodes than not to give it at all.

etrepum:
I think this test might need to be looked at:
```
[chromium] › packages/lexical-playground/__tests__/e2e/Autocomplete.spec.mjs:31:3 › Autocomplete › Can autocomplete a word 
```

Issue #8126, opened 2026-02-10
Bug: Incorrect behavior when pressing Enter on a link


trinhvinhtruong96:
Lexical version:
v0.40.0
## Steps To Reproduce

1. Go to https://playground.lexical.dev/ 
2. Click clear and type "https://playground.lexical.dev playground"
3. Move the caret to the middle of the link and hit enter (or shift-enter)

see: 

https://github.com/user-attachments/assets/79c51ca5-4852-4b1b-a7a1-85cf57d3842b

Link to code example: 

## The current behavior
The text after the caret moves to a new line, but it doesn't keep the original order.

## The expected behavior
The text after the caret moves to a new line, and it keep the original order.

PR #8127, opened 2026-02-10
[lexical-playground][lexical-link] Bug Fix: Prevent AutoLinkNode from creating extra paragraphs

### LINKED ISSUE #5882: Bug: Additional paragraph inserted when pressing "Enter" within AutoLinkNode
When pressing "Enter" and the cursor is within an AutoLinkNode, the link is split into two paragraphs, but there is also an empty paragraph is inserted in between. This is also reproducible in the Playground.

Lexical version: 0.14.3

## Steps To Reproduce

1. Visit playground and delete demo text
2. Type "http://example.com"
3. Press left arrow key to move cursor one character to the left
4. Press "Enter"

https://github.com/facebook/lexical/assets/5648131/04f1b334-e41d-4f46-94d2-2e01cbbdd52a

## The current behavior

There are 3 paragraphs, because an additional empty `ParagraphBlock` is inserted.

## The expected behavior

There should be 2 paragraphs, one with text content `http://example.co`, and one with text content `m`.

polyrainbow:
## Description

Bug fix to prevent AutoLinkNode from creating an extra paragraph, when pressing Enter inside an AutoLinkNode.

Root Cause: AutoLinkNode.insertNewAfter() delegated to its parent (ParagraphNode) to create a new block, which caused a premature paragraph split during $splitNodeAtPoint. Then insertParagraph created another paragraph for the actual Enter key handling — resulting in 3 paragraphs instead of 2.

Closes #5882

### Before

https://github.com/user-attachments/assets/80dff92a-0ff3-4e9d-bdff-245bed0bc758

### After

https://github.com/user-attachments/assets/674b91f9-d20b-48fe-b0fc-2a3b12ee1da8

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/ComponentPicker.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/ComponentPicker.spec.mjs
@@ -1,0 +1,123 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {
+  assertHTML,
+  click,
+  focusEditor,
+  html,
+  initialize,
+  test,
+  waitForSelector,
+} from '../utils/index.mjs';
+
+test.describe('ComponentPicker', () => {
+  test('Can insert a heading using the component picker slash command', async ({
+    page,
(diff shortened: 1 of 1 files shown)


trinhvinhtruong96:
This should also fix issue #8126. Thank you @polyrainbow

etrepum (approved):

Issue #8129, opened 2026-02-10
Bug: markdown link transformer wraps the entire line into a link


levensta:
Lexical version: 0.40.0

## Steps To Reproduce

1. The string must contain a text node in the form of a markdown-like link. Insert this
   ```
   [text](https://x.com) — this is markdown link. Try to write ')'. 
   ```
2. Manually write the markdown link on the same line: `[text](https://x.com)`

Link to code example:

https://playground.lexical.dev/#doc=H4sIAAAAAAAAE51QS2rDMBC9iphNGhCOYpqG6gBdF9xdyEJY01hElsxonMQEQw_RE_YkRU5LmtJVVzPvwbzPnAGt40gVG0bQZ6AYOc-6cd4SBtCbX8AiG-dBKwmvkVrD09pGi6AhZMaDhMSDzwRIYDwxaNjkub1rmLukF4tTUcd2Lj7e3gU3LgmXRGtob-MxCO_CvhAvNAiO4kiOUczms0L8JZENhi5bTT4SDkjJxQB6OW4lWEdY84RD7_01c07mgsVwyf-l0RkyOzJdcyN06fB0bZth9d3wHzbTl2-ijhK8SVyZA1rQy_VarcvVQ_moypWEFHuq892zN8OOYh_sz2tQxb0qFIyfULPppM4BAAA

## The current behavior

Аfter the ')' character is triggered, the entire line will wrap in LinkNode. In the text part of the link, square brackets will be cut off at the edges.

https://github.com/user-attachments/assets/42a0be86-ba82-4342-984e-8935b868a1d4

## The expected behavior

Only the current link being entered is transformed

<img width="1181" height="706" alt="Image" src="https://github.com/user-attachments/assets/ae16e59e-c735-4cb0-a81b-c753ddf11dba" />

## Impact of fix

In theory, this may prevent you from writing  content for an article on the use of markdown

PR #8130, opened 2026-02-10
[lexical-markdown] Fix: Prevent Link Transformer from consuming preceding text

Sa-Te:
The Markdown link transformer was using the regex match index relative to the entire text content, rather than relative to the specific TextNode being transformed. This caused text preceding the link (e.g., Start [link](url)) to be incorrectly sliced out if the match index was non-zero.

Fix:
- Updated replace in LINK transformer to calculate the match index relative to the current textNode.
- Added logic to splitText only if the local index > 0.

Test Plan:
- Added unit test MarkdownTransformers.test.ts covering the scenario where text precedes a markdown link.
- Verified manually in Playground.

### DIFF
--- a/packages/lexical-markdown/src/__tests__/unit/MarkdownTransformers.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/MarkdownTransformers.test.ts
@@ -16,11 +16,12 @@ describe('MarkdownLinkBug', () => {
   test('Expect text BEFORE a markdown link to be preserved', () => {
     const editor = createHeadlessEditor({
       nodes: [LinkNode],
-      onError(error) {},
+      onError(error) {
+        throw error;
+      },
     });
 
     editor.update(() => {
-      // creating a scenario: "Start [test](url)"
       const root = $getRoot();
       const paragraph = $createParagraphNode();
       const textNode = $createTextNode('Start [test](url)');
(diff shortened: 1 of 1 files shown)


packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -681,6 +681,18 @@ export const LINK: TextMatchTransformer = {
     /(?:\[(.+?)\])(?:\((?:([^()\s]+)(?:\s"((?:[^"]*\\")*[^"]*)"\s*)?)\))$/,
   replace: (textNode, match) => {
     const [, linkText, linkUrl, linkTitle] = match;
+    const matchText = match[0];
+
+    const textContent = textNode.getTextContent();
+    const localMatchIndex = textContent.indexOf(matchText);
+    if (localMatchIndex === -1) {
+      return;
+    }
```
etrepum, line 690: In what scenario would this be any different from this code?

```suggestion
    const localMatchIndex = match.index || 0;
```

Sa-Te: I tried the optimization, but it caused regressions in the LexicalMarkdown import tests. It seems match.index can become stale if other transformers modify the node content during the pass. Reverting to indexOf fixes the regression.

etrepum: It seems that the importer already splits the text based on the match before calling the replace function so localMatchIndex is always 0

Sa-Te:
> This test passes even if the MarkdownTransformers changes are reverted, so it's not clear that it's fixing anything

@etrepum Thanks for the review!

I investigated your comment about localMatchIndex always being 0. While that is true for simple cases, I found that relying on match.index (or assuming 0) actually causes 11 regressions in the lexical-markdown unit test suite.

The Issue:
In cases of nested formatting (e.g., **Bold** [Link](url)), a preceding transformer (like Bold) modifies the text node structure before the Link transformer runs. This causes the original match.index to become stale relative to the new text node state.

Evidence:
When I revert to match.index, tests like can import "**text [link](https://lexical.dev)**" fail because the index is no longer accurate.

The Fix:
Switching to textNode.getTextContent().indexOf(match[0]) calculates the correct index on the current node state. I verified this locally, and it results in a 100% pass rate (283 tests), fixing the regressions.

etrepum:
I don't see any test failures when reverting your changes, even in the new test, so if this fixes anything it needs additional tests that will fail without your changes.

Relying on match.index indeed does not work (without other refactoring) because the importer splits the TextNode before calling the replace function so the string is already split, at least in the import cases.

Sa-Te:
> I don't see any test failures when reverting your changes, even in the new test, so if this fixes anything it needs additional tests that will fail without your changes.
> 
> Relying on match.index indeed does not work (without other refactoring) because the importer splits the TextNode before calling the replace function so the string is already split, at least in the import cases.

@etrepum Thanks for the feedback! I've added a new unit test (LINK.replace handles stale match indices) that isolates the specific race condition where match.index becomes stale.

The Issue:
When Markdown formatting is nested (e.g., **Bold** [Link](url)), the Bold transformer runs first and modifies the text node structure. This invalidates the original match.index (which was calculated on the full string) by the time the Link transformer runs.

Without this fix (using match.index): The new test fails, and I also see regressions in 11 existing tests (specifically regarding nested formatting).

With this fix (using indexOf): The new test passes.

This confirms that searching the current text node is necessary to handle these nested cases correctly.

etrepum:
I think these unit tests are a bit too artificial to be useful, they really should be called by the transformer infrastructure (shortcuts/importer) to show that they are working as expected in the environments that they are used from. I don't think these tests are a good simulation of how that works.

etrepum:
It does not look like this code fixes #8129 which is the only related issue that I'm aware of, so either this doesn't fix anything, or it fixes an issue that I haven't seen a reproduction for.

PR #8131, opened 2026-02-10
[lexical-table] Fix: Preserve Background Color when pasting table rows inside table.

Sa-Te:
The Issue:
When pasting rows or cells into an existing table, the TableCellNode creation logic in $insertTableIntoGrid was copying the cell's children (content) but failing to copy the cell's own style properties (specifically backgroundColor). This resulted in the background color being lost upon paste.

The Fix:

Updated $insertTableIntoGrid in LexicalTablePlugin.ts (or LexicalTableSelectionHelpers.ts).

Added logic to check templateCell.getBackgroundColor() and apply it to the target cell before inserting children.

Test Plan:

Manual Verification: Verified in Playground.

- Created a table.

- Applied a background color to a row.

- Copied the row.

- Pasted it into a new row.

Result: Background color is now preserved (previously lost).

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/TablesHTMLCopyAndPaste.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/CopyAndPaste/html/TablesHTMLCopyAndPaste.spec.mjs
@@ -342,14 +342,14 @@ test.describe('HTML Tables CopyAndPaste', () => {
           </colgroup>
           <tr>
             <th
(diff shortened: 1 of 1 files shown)


pnpm-lock.yaml
```diff
@@ -3435,24 +3435,28 @@ packages:
     engines: {node: '>= 10'}
     cpu: [arm64]
     os: [linux]
+    libc: [glibc]
```
ivailop7: I don't think the changes to the pnpm-lock.yaml should be commited

ivailop7:
table tests are failing as well

PR #8132, opened 2026-02-11
[Breaking Change][lexical] Bug Fix: Fix --lexical-indent-base-value CSS variable override 

### LINKED ISSUE #8095: Bug: --lexical-indent-base-value: 20px; not work
## Issue
  --lexical-indent-base-value: 20px; not work, still 40px

Lexical version: 0.39.0

## Steps To Reproduce

1. set :root { --lexical-indent-base-value: 20px; } not work

Image

I think the bug is due to lexical use this code: getComputedStyle(dom).getPropertyValue('--lexical-indent-base-value')

because dom is not attached to the document, so the result is always empty and then will use the default 40px

achaljhawar:
## Breaking change

Setting the `--lexical-indent-base-value` CSS custom property is now read from the editor's root element, so it must be set there or higher (e.g. `:root`) in the CSS. The previous implementation that attempted to read this property from other nodes had race conditions and performance implications.

## Description

fixes --lexical-indent-base-value CSS variable override by reading it from the editor root element instead of detached DOM nodes, and removing element-level declarations that were shadowing :root overrides.

Closes #8095 

### Before

https://github.com/user-attachments/assets/b5a2f879-6086-4dc8-9270-34f85ce4f237

### After

https://github.com/user-attachments/assets/a2699dc3-56ef-4ce8-9f33-5b640c2da55c

### DIFF
--- a/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
+++ b/packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
@@ -6,6 +6,9 @@
  *
  *
  */
+:root {
+  --lexical-indent-base-value: 40px;
+}
 .PlaygroundEditorTheme__paragraph {
   margin: 0;
   position: relative;
@@ -38,9 +41,6 @@
   font-size: 12px;
   margin: 0;
   text-transform: uppercase;
-}
-.PlaygroundEditorTheme__indent {
-  --lexical-indent-base-value: 40px;
 }
 .PlaygroundEditorTheme__textBold {
   font-weight: bold;


packages/lexical-playground/src/themes/PlaygroundEditorTheme.css
```diff
@@ -39,9 +39,6 @@
   margin: 0;
   text-transform: uppercase;
 }
-.PlaygroundEditorTheme__indent {
-  --lexical-indent-base-value: 40px;
-}
```
etrepum, line 44: This should still be set somewhere, otherwise there's no example of its usage.

etrepum (approved):

PR #8138, opened 2026-02-12
[lexical-link] Bug Fix: $toggleLink removes link when selection is collapsed

Senasiko:
## Description

This PR fixs an issue where, when the selection is collapsed, $toggleLink only affects the nearest TextNode
### Before

https://github.com/user-attachments/assets/320a6875-9867-4854-9e1b-03f6ee05db1d

### After

https://github.com/user-attachments/assets/dbc78a77-02d8-40a4-861d-a785df99d52b

### DIFF
--- a/packages/lexical-link/src/LexicalLinkNode.ts
+++ b/packages/lexical-link/src/LexicalLinkNode.ts
@@ -666,8 +666,7 @@ export function $toggleLink(
   }
 
   if (selection.isCollapsed() && url === null) {
-    const node = selection.getNodes()[0];
-    if (node) {
+    for (const node of selection.getNodes()) {
       const parentLink = $findMatchingParent(
         node,
         (parent): parent is LinkNode =>


packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -685,7 +685,14 @@ export function $toggleLink(
           return;
         }
 
-        $splitLinkAtSelection(parentLink, nodes);
+        if (selection.isCollapsed()) {
+          parentLink.getChildren().forEach((child) => {
+            parentLink.insertBefore(child);
+          });
+          parentLink.remove();
+        } else {
```
etrepum: it would probably be better to handle this outside of the nodes.forEach loop

Senasiko: I made this change, please review again

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -665,6 +665,24 @@ export function $toggleLink(
     return;
   }
 
+  if (selection.isCollapsed() && url === null) {
+    const node = selection.getNodes()[0];
+    if (node) {
```
```suggestion
    for (const node of selection.getNodes()) {
```

janpe:
Any idea when this fix might be seen in a release?

etrepum:
It’s already in the nightly release, the monthly release is another week or two out.

PR #8140, opened 2026-02-13
[lexical-markdown] Feature: Support Enter key after ``` to create code block

### LINKED ISSUE #7948: Feature: Support starting a code block by pressing Enter after ``` in markdown plugin.
## Description
Currently, Lexical only recognizes a code block when a space follows “```”, but most Markdown editors allow pressing Enter instead.

achaljhawar:
## Description

This PR adds an Enter key listener so typing ``` and pressing Enter creates a code block. 

Closes #7948 

### Before

https://github.com/user-attachments/assets/b3dec04d-c7c0-4a7e-9a4f-da85f567767a

### After

https://github.com/user-attachments/assets/8e04f3fe-8182-4f55-8bec-987e93d2d43f

### DIFF
--- a/packages/lexical-markdown/src/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/MarkdownShortcuts.ts
@@ -127,9 +127,8 @@ function runMultilineElementTransformers(
     const match = textContent.match(regExpStart);
 
     if (match) {
-      const matchLength = triggerOnEnter
-        ? textContent.length
-        : match[0].endsWith(' ')
+      const matchLength =
+        triggerOnEnter || match[0].endsWith(' ')
           ? anchorOffset
           : anchorOffset - 1;
 
@@ -137,9 +136,8 @@ function runMultilineElementTransformers(
         continue;
       }
 
-      const splitOffset = triggerOnEnter ? textContent.length : anchorOffset;
       const nextSiblings = anchorNode.getNextSiblings();
-      const [leadingNode, remainderNode] = anchorNode.splitText(splitOffset);
+      const [leadingNode, remainderNode] = anchorNode.splitText(anchorOffset);
       const siblings = remainderNode
         ? [remainderNode, ...nextSiblings]
         : nextSiblings;


packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -525,4 +572,63 @@ export function registerMarkdownShortcuts(
...
+
+      return false;
+    },
+    COMMAND_PRIORITY_NORMAL,
+  );
+
+  return () => {
+    removeUpdateListener();
+    removeEnterListener();
+  };
```
etrepum:
```suggestion
  return mergeRegister(
    // but replace these with the definitions, no need to use variables
    removeUpdateListener,
    removeEnterListener,
  );
```

packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -525,4 +572,63 @@ export function registerMarkdownShortcuts(
...
+      ) {
+        if (event !== null) {
+          event.preventDefault();
+        }
+        return true;
+      }
+
+      return false;
+    },
+    COMMAND_PRIORITY_NORMAL,
```
etrepum: Generally these should be registered at the lowest priority that works, I think in this case it would probably be low

```suggestion
    COMMAND_PRIORITY_LOW,
```

packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -525,4 +572,63 @@ export function registerMarkdownShortcuts(
...
+      const anchorOffset = selection.anchor.offset;
+      const anchorNode = selection.anchor.getNode();
+
+      if (!$isTextNode(anchorNode)) {
+        return false;
+      }
+
+      if (!canContainTransformableMarkdown(anchorNode)) {
+        return false;
+      }
```
etrepum:
```suggestion
      if (!$isTextNode(anchorNode) || !canContainTransformableMarkdown(anchorNode)) {
        return false;
      }
```

packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -141,6 +143,51 @@ function runMultilineElementTransformers(
   return false;
 }
 
+function runMultilineElementTransformersOnEnter(
```
etrepum: This looks like almost exactly the same code as `runMultilineElementTransformers`, it would be better to fuse these implementations (maybe with an extra argument to specify the on enter behavior) rather than make the code this much bigger.

packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -120,13 +126,20 @@ function runMultilineElementTransformers(
...
+          : anchorOffset - 1;
+
+      if (match[0].length !== matchLength) {
+        continue;
+      }
+
+      const splitOffset = triggerOnEnter ? textContent.length : anchorOffset;
       const nextSiblings = anchorNode.getNextSiblings();
-      const [leadingNode, remainderNode] = anchorNode.splitText(anchorOffset);
+      const [leadingNode, remainderNode] = anchorNode.splitText(splitOffset);
```
etrepum: I think some of these changes are not necessary since `anchorOffset === textContent.length` should be true when triggerOnEnter is true, since that is checked before calling `runMultilineElementTransformers`

```suggestion
      const matchLength = triggerOnEnter || match[0].endsWith(' ')
          ? anchorOffset
          : anchorOffset - 1;

      if (match[0].length !== matchLength) {
        continue;
      }

      const nextSiblings = anchorNode.getNextSiblings();
      const [leadingNode, remainderNode] = anchorNode.splitText(anchorOffset);
```

Issue #8153, opened 2026-02-19
Bug: Selected text not properly deleted with IME input (e.g. Japanese) on Safari (macOS)


sKawashima:
Lexical version: 0.36.2 (also reproducible on the latest https://playground.lexical.dev )

## Steps To Reproduce

1. Open a Lexical editor on Safari (macOS)
   - Reproducible on the latest https://playground.lexical.dev as well
2. Switch to a Japanese IME (or another IME-based input method such as Chinese/Korean)
3. Type some text and confirm the composition (e.g. type "あいうえお")
4. Select the text using Shift+Arrow keys or Cmd+A
5. Press the Delete (Backspace) key

Link to code example: https://playground.lexical.dev

## The current behavior

The selected text is not properly deleted. Depending on the case, only the last character is removed, or only the first few characters are deleted, instead of the entire selection being cleared.

https://github.com/user-attachments/assets/3d76d9f6-b7ff-4482-aae3-7f0bcc853d4d

This issue does not occur with direct-input (non-IME) languages such as English.

## The expected behavior

The selected text should be fully deleted when pressing Delete, regardless of whether the text was entered via IME composition.

## Impact of fix

- This affects all users who use IME-based input languages (Japanese, Chinese, Korean, etc.) on Safari (macOS).
- Basic text editing operations (select + delete) are broken, significantly degrading the editing experience for these users.
- Safari has a notable market share on macOS, and CJK languages collectively represent a large user base.

PR #8154, opened 2026-02-19
[lexical] Bug Fix: Fix selected text not properly deleted after IME input on Safari (macOS)

sKawashima:
## Description

In Safari on macOS, after confirming IME composition (e.g. Japanese input) with a non-backspace key (Enter, Space, etc.), pressing Backspace to delete a text selection (created via Shift+Arrow or Cmd+A) did not work correctly.

**Root cause:**

Safari fires `compositionend` before `keydown` (unlike other browsers).
`isSafariEndingComposition` is set to `true` on `compositionend` to defer composition finalization to the next `keydown`. However, the flag was only cleared when `keydown` fired with Backspace — leaving it as stale `true` when composition was confirmed with any other key.

When the user later pressed Backspace to delete a selection:
1. `$handleKeyDown` saw `isSafariEndingComposition = true` and called `$onCompositionEndImpl`
2. `$updateTextNodeFromDOMContent` called `selection.setTextNodeRange(anchorNode, anchorOffset, anchorNode, focusOffset)`, forcing anchor and focus onto the same node
3. For multi-node selections (e.g. Cmd+A spanning multiple paragraphs), this corrupted the selection
4. The reconciler updated the DOM selection to match the corrupted Lexical selection
5. The subsequent `beforeinput deleteContentBackward` operated on the wrong selection, producing incorrect deletion

**Fix:** When `isSafariEndingComposition` is `true` but the key is not Backspace, clear the flag without calling `$onCompositionEndImpl`. Composition finalization for non-backspace confirmations is already handled by the `beforeinput` event.

Closes https://github.com/facebook/lexical/issues/8153

### Before

After typing with Japanese IME and confirming with Enter, selecting all text with Cmd+A and pressing Delete does not delete the selected text correctly.

https://github.com/user-attachments/assets/3d76d9f6-b7ff-4482-aae3-7f0bcc853d4d

### After

Deletion works correctly after IME composition confirmed with non-backspace
keys.

https://github.com/user-attachments/assets/c727717b-9ef7-4140-ba96-b7e0fd8d5510

### DIFF
--- a/packages/lexical-playground/__tests__/regression/8153-safari-ime-delete-selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/regression/8153-safari-ime-delete-selection.spec.mjs
@@ -40,9 +40,11 @@ test.describe('Regression #8153', () => {
     page,
     browserName,
     isPlainText,
+    isCollab,
   }) => {
     test.skip(browserName !== 'webkit');
     test.skip(isPlainText);
+    test.skip(isCollab);
 
     await focusEditor(page);
     await page.keyboard.type('Hello');
@@ -72,9 +74,11 @@ test.describe('Regression #8153', () => {
     page,
     browserName,
     isPlainText,
+    isCollab,
   }) => {
     test.skip(browserName !== 'webkit');
     test.skip(isPlainText);
+    test.skip(isCollab);
 
     await focusEditor(page);
     await page.keyboard.type('Hello');


sKawashima:
@etrepum 
Thank you for checking.
I added a WebKit-only regression test.

Note that Playwright has no IME simulation API for WebKit (unlike Chromium's CDP), so the test dispatches a compositionend event programmatically rather than driving real IME input.
It's white-box in nature but reliably exercises the bug's code path.

PR #8155, opened 2026-02-19
[lexical] Bug Fix: REDO_COMMAND not triggered with non-English keyboard layouts (Ctrl+Y / Ctrl+Shift+Z)

### LINKED ISSUE #8147: Bug: REDO_COMMAND not triggered with non-English keyboard layouts (Ctrl+Y / Ctrl+Shift+Z)
Lexical version: 0.40.0

## Steps To Reproduce

0. OS: Windows 10 (22H2), Windows 11 (24H2), Ubuntu 22.04.5 LTS / Browser: Google Chrome 140.0.7339.127, Google Chrome 144.0.7559.133, Firefox Developer Edition 148.0b15
1. Set English and another keyboard layout in the OS (for example, English (US) and Russian (RU)).
2. Enter any text into the lexical input field so that a `historyUndo` action becomes available.
3. Press the keyboard shortcut (Ctrl+Z) to trigger the `historyUndo` native browser event.
4. Press the keyboard shortcut (Ctrl+Y or Ctrl+Shift+Z) to trigger the `historyRedo` native browser event.

## The current behavior

* lexical.js does NOT trigger `REDO_COMMAND` because the native `historyRedo` event is not fired.
* lexical.js does NOT trigger `REDO_COMMAND` because the `event.key` equals to "я" not "z", but it is same physical key button in the keyboard.
* lexical.js detects keyboard shortcuts based on `event.key` instead of `event.code` (https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalUtils.ts#L990).
* The browser menu is not aware of Lexical’s history stack and therefore disables the “Redo” option (I suspect there are browser restrictions).
* My screencast: https://github.com/user-attachments/assets/3e086e9d-95df-4940-84ec-ccb561a5d0bd

## The expected behavior

* lexical.js should process `REDO_COMMAND` when Ctrl+Y or Ctrl+Shift+Z is pressed (with any non-English keyboard  layout).
* lexical.js should determine all shortcuts using `event.code` by default (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#examples).

## Impact of fix

* Affects all users who are not using an English keyboard layout.

kenclaron:
## Description

- `REDO_COMMAND` and other shortcuts are triggered correctly on all keyboard layouts.
- added fallback with `event.code` for single-letter `expectedKey` for non-English keyboard layout.
- added few unit tests for `isExactShortcutMatch()`

Closes #8147

### Before

- screencast: https://github.com/user-attachments/assets/3e086e9d-95df-4940-84ec-ccb561a5d0bd
- no any unit tests for `isExactShortcutMatch()`

### After

- added few unit tests for  `isExactShortcutMatch()` in `/packages/lexical/__tests__/unit/LexicalUtils.test.ts`

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -1103,11 +1103,7 @@ export function isUndo(event: KeyboardEventModifiers): boolean {
 
 export function isRedo(event: KeyboardEventModifiers): boolean {
   if (IS_APPLE) {
-    // Accept both Cmd+Shift+Z and Cmd+Y on macOS (some apps and users expect Cmd+Y).
-    return (
-      isExactShortcutMatch(event, 'z', {metaKey: true, shiftKey: true}) ||
-      isExactShortcutMatch(event, 'y', {metaKey: true})
-    );
+    return isExactShortcutMatch(event, 'z', {metaKey: true, shiftKey: true});
   }
   return (
     isExactShortcutMatch(event, 'y', {ctrlKey: true}) ||


etrepum:
I don't have strong opinions about handling CMD+Y specifically, all of the Mac browsers have it bound to "Show all History" by default. I think the only people who would try that keyboard shortcut are people who are not used to macOS that happen to be using a mac.

I think probably a higher priority should be abstracting all of the built-in keyboard shortcuts so that they are provided by an extension that makes it easier to reconfigure them, right now it's pretty awkward to override any key handling. That way an app can decide if it should behave this way (e.g. for a full page document editor) or if it would be better to follow native conventions.

For the context of this PR specifically my primary concern is that I don't know how it behaves with a keyboard map that maps ascii to ascii, e.g. dvorak or azerty. Possibly it should have a check to make sure that the key is non-ascii before trying the code, or maybe some of these shortcuts should only check the code? I don't have a windows machine to test and playwright doesn't support any keyboard remapping.

levensta:
By the way, Google Docs can workaround the default behavior of Cmd + Y in browsers on MacOS

etrepum:
Yes, google docs can override cmd+y, but if you're using lexical for a form field or a chat input or something like that you might not want it to do that by default.

etrepum:
FWIW at least on macOS the shortcut behavior when using ASCII remapping (e.g. dvorak) is that the shortcuts map to the logical key and not the physical key code. If you press Cmd+KeyB (physical key code) with a dvorak mapping it executes the logical key Cmd+X action (in native apps also, not just chrome). No idea how Windows users expect this to behave, but perhaps the right workaround is to make sure that the key is non-ascii before trying the code fallback.

Screenshot 2026-02-22 at 11 08 19

levensta:
@kenclaron will you continue working on this PR? As a user of the Cyrillic keyboard layout, I am very interested in seeing this fix included in the lexical core

Looks like there's a small fix left to make it work and get compatible

kenclaron:
Thanks for comments!

I added an additional check for non-standard keyboard layouts with ASCII characters (e.g. English/Dvorak) so that it works only on `event.key` as previously (not both `event.key` and `event.code`), by your comments. And added 1 jest-test for that user case.

About "we should incorporate the CMD + Y behavior as a standardized behavior". so I revert it part of code from the PR for now.

PR #8156, opened 2026-02-19
Fix $convertLayoutContainerElement always returning null

### LINKED ISSUE #6813: The $convertLayoutContainerElement function of LayoutContainerNode always returns null
Lexical version:0.19.0

## Steps To Reproduce

1.Create a 2-column example in the editor and export it as HTML
2.Re import the exported HTML content in the editor

Link to code example:

## The current behavior
The window.getComputedStyle used in this function cannot retrieve any value from the un mounted rendered HTML, so the $convertLayoutContainerElement always returns null.
## The expected behavior
successfully returned column instance and rendered column node layout correctly

danielalanbates:
Fixes #6813

## Summary

The `$convertLayoutContainerElement` function in `LayoutContainerNode.ts` always returned `null` when importing HTML because it used `window.getComputedStyle()` to read `grid-template-columns`. Since the DOM nodes being imported are not mounted in the document, `getComputedStyle()` returns empty values for all properties.

## Fix

Changed from `window.getComputedStyle(domNode).getPropertyValue('grid-template-columns')` to reading `domNode.style.gridTemplateColumns` directly. This works because `exportDOM()` already sets `gridTemplateColumns` as an inline style on the exported element, so the value is available on the `style` property without needing computed styles.

## Test plan

- Export a 2-column layout as HTML
- Re-import the exported HTML — layout columns should now be preserved correctly
- Verified that `exportDOM()` sets inline `gridTemplateColumns`, confirming `domNode.style` will have the value

*This PR was created with the assistance of Claude Opus 4.6 by Anthropic. Happy to make any adjustments! Reviewed and submitted by a human.*

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/packages/lexical-playground/src/nodes/LayoutContainerNode.ts
+++ b/packages/lexical-playground/src/nodes/LayoutContainerNode.ts
@@ -31,10 +31,7 @@ export type SerializedLayoutContainerNode = Spread<
 function $convertLayoutContainerElement(
   domNode: HTMLElement,
 ): DOMConversionOutput | null {
-  const styleAttributes = window.getComputedStyle(domNode);
-  const templateColumns = styleAttributes.getPropertyValue(
-    'grid-template-columns',
-  );
+  const templateColumns = domNode.style.gridTemplateColumns;
   if (templateColumns) {
     const node = $createLayoutContainerNode(templateColumns);
     return {node};


danielalanbates:
Thanks for the feedback @etrepum\! I'll look into signing the Meta CLA and adding a test case. I appreciate the guidance on what's needed to get this reviewed.

danielalanbates:
Thank you for your time reviewing this. We're withdrawing this PR. Apologies for any inconvenience.

PR #8158, opened 2026-02-20
[lexical] Bug Fix: Move new paragraph outside inline element in insertParagraph

### LINKED ISSUE #8083: Bug: new paragraph can be inserted inside inline node
Lexical version: v0.39.0

## Steps To Reproduce

1. Copy the title from a google search result and pasting it into the editor.
2. Press Enter at the end of the line

https://github.com/user-attachments/assets/83933df7-a13a-4b1a-861e-2c32091e79c8

Link to code example:

https://playground.lexical.dev/#doc=H4sIAAAAAAAAE51SwW6DMAz9F58jStWKtvmAnXaYxHHawSIeRHMTZEwFqvj3KbCua08Tp-TZeu_5ObkCOa9RSkUlsFeQGDWdVePZCQWw7_8GjhQ9g80NfEY5o87Xc3QEFkKqMBjodORUAANKg4KFVxp8Nfd0bFNrrhu4kHQ-BrDb6cOA80KVzjj0zHePpOSDo7D4_Wg0hM6H-kHGgGKdejtYocg-fD3JCfGNqyg16S_ymkIuoBdOpqptZzcbXtJmji6bNWO0KFgLts1zNBr05b72BMvbqlfYzB_h4Q0mA4ydlnghB3Z7KI7FcZ_vitP-YKCLvVSJ98Y41hL74P6yIc92pyyH6Rs6Z8oscQIAAA

## The current behavior

New paragraph inserted inside LinkNode

```
 root
  └ (47) paragraph 
    └ (48) link "https://lexical.dev/"
      ├ (49) heading 
      | └ (50) text "Lexical"
>     ├ (52) paragraph 
```

## The expected behavior

New paragraph inserted outside LinkNode

```
 root
  ├ (47) paragraph 
  | └ (48) link "https://lexical.dev/"
  |   ├ (49) heading 
  |   | └ (50) text "Lexical"
> └ (52) paragraph 
```

## Impact of fix

I discovered this corner case while reading the bug description in this [PR](https://github.com/facebook/lexical/pull/8078). In cases where a link or any other inline node contains a block element, all new nodes created by pressing Enter will be placed inside that inline element.

This may be normal behavior in the case of a header split or other elements, but the insertion of a new paragraph should be outside the inline element and the parent block

Ideally, I can imagine that normalizing the tree when inserting new nodes could be done in such a way that each inline ElementNode containing block elements is copied for each block and wraps the highest inline ElementNode or TextNode. But maybe it is only necessary to correct the case with the insertion of a new paragraph

achaljhawar:
## Description

Fixed `insertParagraph` to move the new paragraph outside of inline elements when a block is nested inside one.

Closes #8083

### Before

Screenshot 2026-02-21 at 12 33 36 AM

### After

Screenshot 2026-02-23 at 2 42 29 PM

### DIFF
--- a/packages/lexical-link/src/LexicalLinkNode.ts
+++ b/packages/lexical-link/src/LexicalLinkNode.ts
@@ -16,6 +16,8 @@ import type {
   LexicalUpdateJSON,
   NodeKey,
   Point,
+  PointCaret,
+  PointType,
   RangeSelection,
   SerializedElementNode,
 } from 'lexical';
@@ -28,14 +30,17 @@ import {
 } from '@lexical/utils';
 import {
   $applyNodeReplacement,
+  $caretFromPoint,
   $copyNode,
   $getChildCaret,
   $getSelection,
   $isElementNode,
   $isNodeSelection,
   $isRangeSelection,
+  $normalizeCaret,
   $normalizeSelection__EXPERIMENTAL,
   $rewindSiblingCaret,
+  $setPointFromCaret,
   $setSelection,
   createCommand,
   ElementNode,
@@ -290,49 +295,18 @@ export class LinkNode extends ElementNode {
   }
 }
 
-type PointBracket = {
-  leftNode: LexicalNode | null;
-  rightNode: LexicalNode | null;
-};
+type CaretPair = [PointCaret<'next'>, PointCaret<'previous'>];
 
-function $savePointBracket(point: Point): PointBracket | null {
-  if (point.type !== 'element') {
-    return null;
-  }
-  const parent = point.getNode();
-  if (!$isElementNode(parent)) {
-    return null;
-  }
-  const children = parent.getChildren();
-  return {
-    leftNode: point.offset > 0 ? children[point.offset - 1] : null,
-    rightNode: point.offset < children.length ? children[point.offset] : null,
-  };
+function $saveCaretPair(point: PointType): CaretPair {
+  const next = $caretFromPoint(point, 'next');
+  return [next, next.getFlipped()];
 }
 
-function $restorePointFromBracket(
-  point: Point,
-  bracket: PointBracket | null,
-): void {
-  if (bracket === null) {
-    return;
-  }
-  const {rightNode, leftNode} = bracket;
-  if (rightNode !== null && rightNode.isAttached()) {
-    const parent = rightNode.getParent();
-    if (parent) {
-      point.set(parent.getKey(), rightNode.getIndexWithinParent(), 'element');
-      return;
-    }
-  }
-  if (leftNode !== null && leftNode.isAttached()) {
-    const parent = leftNode.getParent();
-    if (parent) {
-      point.set(
-        parent.getKey(),
-        leftNode.getIndexWithinParent() + 1,
-        'element',
-      );
+function $restoreCaretPair(point: PointType, pair: CaretPair): void {
+  for (const caret of pair) {
+    if (caret.origin.isAttached()) {
+      const normalized = $normalizeCaret(caret);
+      $setPointFromCaret(point, normalized);
       return;
     }
   }
@@ -345,11 +319,11 @@ function $restorePointFromBracket(
  */
 export function $linkNodeTransform(link: LinkNode): void {
   const selection = $getSelection();
-  let anchorBracket: PointBracket | null = null;
-  let focusBracket: PointBracket | null = null;
+  let anchorPair: CaretPair | null = null;
+  let focusPair: CaretPair | null = null;
   if ($isRangeSelection(selection)) {
-    anchorBracket = $savePointBracket(selection.anchor);
-    focusBracket = $savePointBracket(selection.focus);
+    anchorPair = $saveCaretPair(selection.anchor);
+    focusPair = $saveCaretPair(selection.focus);
   }
 
   for (const caret of $getChildCaret(link, 'next')) {
@@ -375,8 +349,8 @@ export function $linkNodeTransform(link: LinkNode): void {
   }
 
   if ($isRangeSelection(selection)) {
-    $restorePointFromBracket(selection.anchor, anchorBracket);
-    $restorePointFromBracket(selection.focus, focusBracket);
+    $restoreCaretPair(selection.anchor, anchorPair!);
+    $restoreCaretPair(selection.focus, focusPair!);
     $normalizeSelection__EXPERIMENTAL(selection);
   }
 }


packages/lexical/src/LexicalSelection.ts
```diff
@@ -1473,6 +1473,18 @@ export class RangeSelection implements BaseSelection {
...
+      if ($isElementNode(newBlockParent) && newBlockParent.isInline()) {
+        const topBlock = $findMatchingParent(newBlockParent, (node) => {
+          const parent = node.getParent();
+          return parent !== null && $isRootOrShadowRoot(parent);
+        });
+        if (topBlock) {
+          topBlock.insertAfter(newBlock);
+        }
+      }
+
```
etrepum: While this might work around the problem it seems like the root cause is probably the implementation of HeadingNode.insertNewAfter. Would also be good to test that this works correctly with other types of block nodes (at least ParagraphNode)

etrepum (commented):
Okay so I took a closer look at this problem and the issue is that the document itself has an invalid structure. Inline elements are not allowed to contain blocks in the first place. It doesn't really make sense to add workarounds to improve the behavior when the document's structure is invalid. The only maintainable approach is to prevent these invalid structures from being created in the first place and/or to use transforms to reorganize the document to be valid when the rules are violated.

The two approaches here would be:
1. Fix LinkNode's importDOM to not create invalid documents
2. Add a LinkNode transform that will fix invalid documents (if a block child is encountered it should split the link and the containing block and then an equivalent LinkNode should be wrapped around the children of the previously contained block, if there are any)

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -286,6 +288,66 @@ export class LinkNode extends ElementNode {
...
+ * Extracts block-level children from a LinkNode, wrapping their
+ * contents with a cloned link to preserve the URL. Handles one
+ * block child per pass; transforms re-trigger for the rest.
+ * @param link - The LinkNode to normalize
+ */
+export function $linkNodeTransform(link: LinkNode): void {
+  const children = link.getChildren();
+  let blockChild: ElementNode | null = null;
+  for (const child of children) {
+    if ($isElementNode(child) && !child.isInline()) {
```
etrepum: Could also handle the block decorator case here

etrepum:
also note that there might be some other logic to handle to make sure that the selection remains valid, if either point of the selection is keyed to an ElementNode on the right side of any of those splits the key and/or offset may no longer be valid. The caret code generally won't touch the selection to fix that stuff up, it's lower-level and avoids those kinds of side-effects (e.g. if using RangeSelection.insertNodes it's just going to replace the current selection in many cases, which is likely not what you want in the cases where this transform is applied).

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -286,6 +289,39 @@ export class LinkNode extends ElementNode {
...
+  for (const caret of $getChildCaret(link, 'next')) {
+    const node = caret.origin;
+    if ($isElementNode(node) && !node.isInline()) {
+      const blockChildren = node.getChildren();
+      if (blockChildren.length > 0) {
+        const innerLink = $createLinkNode(link.__url, {
+          rel: link.__rel,
+          target: link.__target,
+          title: link.__title,
+        });
```
```suggestion
        const innerLink = $copyNode(link);
```

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -286,6 +289,39 @@ export class LinkNode extends ElementNode {
...
+      });
+    }
+  }
+  if (link.getChildrenSize() === 0) {
+    const parent = link.getParent();
+    link.remove();
+    if ($isElementNode(parent) && parent.isEmpty()) {
+      parent.remove();
+    }
+  }
```
etrepum: I think this might not be needed, but it can be more concise

```suggestion
  if (link.isEmpty()) {
    const parent = link.getParent();
    link.remove();
    if (parent && parent.isEmpty()) {
      parent.remove();
    }
  }
```

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +887,153 @@ describe('formatUrl', () => {
...
+        const root = $getRoot();
+        const text = $createTextNode('Lexical');
+        const heading = $createHeadingNode('h1');
+        heading.append(text);
+        const link = $createLinkNode('https://lexical.dev');
+        link.append(heading);
+        const paragraph = $createParagraphNode();
+        paragraph.append(link);
+        root.clear().append(paragraph);
+        text.select(3, 3);
```
etrepum: text selections aren't going to be a problem since we are not doing anything that will split a TextNode. The problem I expect to see is when the selection lies on an ElementNode that is split. If that offset is after the split point, the key and the offset both need to change to the right side of the split.

packages/lexical-link/src/index.ts
```diff
@@ -27,6 +27,7 @@ export {
   $createLinkNode,
   $isAutoLinkNode,
   $isLinkNode,
+  $linkNodeTransform,
```
etrepum: I don't think we really need to add this to the public API, the tests can use the LinkExtension

etrepum (commented):
I think there are more edge cases to the selection here. There two base cases that should be tested:

- Selection is in the paragraph that gets split to the right of the LinkNode
- Selection is in the LinkNode to the right of the non-inline node

These cases also branch depending on if there are siblings to the right of the non-inline node (which may even be another non-inline node) and/or to the right of the LinkNode in the cases where there is no sibling (which determines whether or not a new paragraph is created).

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +888,277 @@ describe('formatUrl', () => {
...
+      expect($isElementNode(children[0])).toBe(true);
+      if ($isElementNode(children[0])) {
+        const headingChildren = children[0].getChildren();
+        expect(headingChildren.length).toBe(1);
+        expect($isLinkNode(headingChildren[0])).toBe(true);
+        expect(headingChildren[0].getTextContent()).toBe('Lexical');
+      }
+    });
+
+    editor.dispose();
```
etrepum: You don't really need an explicit dispose in these tests, the extensions aren't doing anything that isn't going to get cleaned up by gc.

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +888,277 @@ describe('formatUrl', () => {
...
+      const children = root.getChildren();
+      expect(children.length).toBe(1);
+      expect(children[0].getType()).toBe('heading');
+      expect($isElementNode(children[0])).toBe(true);
+      if ($isElementNode(children[0])) {
+        const headingChildren = children[0].getChildren();
+        expect(headingChildren.length).toBe(1);
+        expect($isLinkNode(headingChildren[0])).toBe(true);
+        expect(headingChildren[0].getTextContent()).toBe('Lexical');
+      }
```
etrepum: These sorts of tests can be simplified quite a bit using assert. expect doesn't do any type narrowing, but assert does.

```suggestion
      const heading = children[0];
      assert($isHeadingNode(heading), 'First child must be a HeadingNode');
      expect(heading.getChildrenSize()).toBe(1);
      const headingChild = heading.getFirstChild();
      assert($isLinkNode(headingChild), 'Heading child must be a LinkNode');
      expect(headingChild.getTextContent()).toBe('Lexical');
```

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +888,277 @@ describe('formatUrl', () => {
...
+        const link = $createLinkNode('https://lexical.dev');
+        link.append(heading);
+        const textAfter = $createTextNode('after');
+        const paragraph = $createParagraphNode();
+        paragraph.append(textBefore, link, textAfter);
+        root.clear().append(paragraph);
+        const selection = $createRangeSelection();
+        selection.anchor.set(paragraph.getKey(), 2, 'element');
+        selection.focus.set(paragraph.getKey(), 2, 'element');
+        $setSelection(selection);
```
```suggestion
        paragraph.select(2, 2);
```

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -286,6 +290,75 @@ export class LinkNode extends ElementNode {
...
+
+function $getChildAtElementPoint(point: Point): LexicalNode | null {
+  if (point.type !== 'element') {
+    return null;
+  }
+  const node = point.getNode();
+  if ($isElementNode(node)) {
+    return node.getChildAtIndex(point.offset) || null;
+  }
+  return null;
```
```suggestion
  return $caretFromPoint(point).getNodeAtCaret();
```

etrepum (commented):
These tests aren't actually checking the selection at all after the transforms run, so it could be anywhere at all (even be set to null) and the tests would still pass.

I also don't think this is the correct algorithm for fixing up the selection because there may not be a next child when the LinkNode is split at the end. The correct key and offset could be on a new node or any parent node (including the root node) since when the link is split there might not be a right side left at all.

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +891,247 @@ describe('formatUrl', () => {
...
+        heading2.append(heading2Text);
+        const link = $createLinkNode('https://lexical.dev');
+        link.append(heading1, heading2);
+        const paragraph = $createParagraphNode();
+        paragraph.append(link);
+        root.clear().append(paragraph);
+      },
+      {discrete: true},
+    );
+    editor.read(() => {
```
etrepum, line 1142: This test doesn't verify the selection

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +891,247 @@ describe('formatUrl', () => {
...
+        const link = $createLinkNode('https://lexical.dev');
+        link.append(heading);
+        const paragraph = $createParagraphNode();
+        paragraph.append(link);
+        root.clear().append(paragraph);
+        paragraph.select(1, 1);
+      },
+      {discrete: true},
+    );
+    editor.read(() => {
```
etrepum, line 1180: This test doesn't verify the selection

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +891,247 @@ describe('formatUrl', () => {
...
+        'Trailing paragraph child must be a LinkNode',
+      );
+      const selection = $getSelection();
+      assert(
+        $isRangeSelection(selection),
+        'Selection must be a RangeSelection',
+      );
+      expect(selection.anchor.type).toBe('element');
+      expect(selection.anchor.key).toBe(trailingLink.getKey());
+      expect(selection.anchor.offset).toBe(0);
```
etrepum: Most operations will normalize the selection before updating it (via `$normalizeSelection` or `$normalizeCaret`). In this case there would probably be a text selection

packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -884,3 +891,247 @@ describe('formatUrl', () => {
...
+      expect(children.length).toBe(3);
+      expect(children[2].getTextContent()).toBe('after');
+      const selection = $getSelection();
+      assert(
+        $isRangeSelection(selection),
+        'Selection must be a RangeSelection',
+      );
+      expect(selection.anchor.type).toBe('element');
+      expect(selection.anchor.key).toBe(children[2].getKey());
+      expect(selection.anchor.offset).toBe(0);
```
etrepum: This would be a text selection if normalized

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -286,6 +290,97 @@ export class LinkNode extends ElementNode {
...
+      point.set(
+        parent.getKey(),
+        leftNode.getIndexWithinParent() + 1,
+        'element',
+      );
+      return;
+    }
+  }
+}
+
```
etrepum: A simpler data structure for this would be carets pointing in both directions. A similar concept but works better at the edges of an ElementNode since it can be a sibling or child caret. There is also no need to avoid text points with this, the TextPointCaret and its offset will remain valid since we are not removing or splitting any TextNode.

```ts
type CaretPair = [PointCaret<'next'>, PointCaret<'prev'>];

function $saveCaretPair(point: Point): CaretPair {
  const next = $caretFromPoint(point, 'next');
  return [next, next.getFlipped()];
}

function $restoreCaretPair(point: Point, pair: CaretPair): void {
  for (const caret of pair) {
    if (caret.origin.isAttached()) {
      return $setPointFromCaret(point, $normalizeCaret(caret));
    }
  }
}
```

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -286,6 +295,66 @@ export class LinkNode extends ElementNode {
...
+    link.remove();
+    if (parent && parent.isEmpty()) {
+      parent.remove();
+    }
+  }
+
+  if ($isRangeSelection(selection)) {
+    $restoreCaretPair(selection.anchor, anchorPair!);
+    $restoreCaretPair(selection.focus, focusPair!);
+    $normalizeSelection__EXPERIMENTAL(selection);
```
etrepum, line 354: Normalization is already done in $restoreCaretPair

```suggestion
```

achaljhawar: Screenshot 2026-02-25 at 6 27 20 AM

Two tests fail with this change — `$normalizeCaret` (per-caret) doesn't resolve element points to text points like `$normalizeSelection__EXPERIMENTAL` does, so the selection ends up on `root` at offset 1 instead of the correct text node.

Should I keep the `$normalizeSelection__EXPERIMENTAL` call, or is there a different way to handle this?

etrepum: Ok, I can look into that separately after this is merged

etrepum (approved):

PR #8161, opened 2026-02-21
[lexical-markdown] Fix: Prevent markdown shortcut link transformer from being too greedy

### LINKED ISSUE #8129: Bug: markdown link transformer wraps the entire line into a link

etrepum:
## Description

The RegExp used for the link markdown transformer was too greedy and could match a preceding link with unbalanced brackets if two are processed on the same line.

Closes #8129
Closes #8130

## Test plan

New unit tests to confirm that import works as expected and that it now works correctly with shortcuts

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -678,8 +678,9 @@ export const LINK: TextMatchTransformer = {
   importRegExp:
     /(?:\[(.+?)\])(?:\((?:([^()\s]+)(?:\s"((?:[^"]*\\")*[^"]*)"\s*)?)\))/,
   regExp:
-    /(?:\[(.+?)\])(?:\((?:([^()\s]+)(?:\s"((?:[^"]*\\")*[^"]*)"\s*)?)\))$/,
+    /(?:\[([^[\]]*(?:\[[^[\]]*\][^[\]]*)*)\])(?:\((?:([^()\s]+)(?:\s"((?:[^"]*\\")*[^"]*)"\s*)?)\))$/,
   replace: (textNode, match) => {
+    // https://spec.commonmark.org/0.31.2/#inline-link
     const [, linkText, linkUrl, linkTitle] = match;
     const linkNode = $createLinkNode(linkUrl, {title: linkTitle});
     const openBracketAmount = linkText.split('[').length - 1;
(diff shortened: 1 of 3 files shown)


packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -678,8 +678,9 @@ export const LINK: TextMatchTransformer = {
   importRegExp:
     /(?:\[(.+?)\])(?:\((?:([^()\s]+)(?:\s"((?:[^"]*\\")*[^"]*)"\s*)?)\))/,
   regExp:
-    /(?:\[(.+?)\])(?:\((?:([^()\s]+)(?:\s"((?:[^"]*\\")*[^"]*)"\s*)?)\))$/,
+    /(?:\[([^[\]]*(?:\[[^[\]]*\][^[\]]*)*)\])(?:\((?:([^()\s]+)(?:\s"((?:[^"]*\\")*[^"]*)"\s*)?)\))$/,
```
etrepum, line 681: I did let claude update this based on the tests I wrote, I did not thoroughly audit it to make sure that it passes the entire commonmark spec, but it does a better job than what was there before.

PR #8163, opened 2026-02-23
[tests] Chore: Update yaml-language-server in lexical-esm-astro-react integration fixture

PikkaPikkachu:
## Test plan
  - [x] Verified `lodash` is completely absent from the updated `pnpm-lock.yaml`
  - [x] Verified `astro check` runs successfully with the overridden `yaml-language-server@1.20.0`
  - [ ] CI integration tests pass (`pnpm run test-integration`)


etrepum:
It’s not really a vulnerability in any practical sense, this code only runs in CI. Probably a more useful approach would be to update the dependencies to versions where an override isn’t necessary, in case this code ever makes it to the examples folder.

etrepum (approved):

etrepum:
Confirmed that @astrojs/check hasn't been updated yet so in the meantime this override is reasonable. Very low priority to worry about "vulnerabilities" in integration tests though.

PikkaPikkachu:
thanks @etrepum for merging in! Makes sense to not update vulnerabilities in tests :) 

will be putting up some more PRs for more critical vulnerabilities this week!

PR #8164, opened 2026-02-23
[lexical-markdown] Bug Fix: Prevent nesting links creation

levensta:
## Description

Fix for behavior where it is possible to create nested links by typing markdown link text inside LinkNode

https://spec.commonmark.org/0.31.2/#example-518

### Before

https://github.com/user-attachments/assets/e5a7bb0e-963d-432b-8e3c-efbb5675eb9d

### After

https://github.com/user-attachments/assets/5a36f600-bf90-40be-b121-a54309142090

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -35,6 +35,7 @@ import {
 import {
   $createLineBreakNode,
   $createTextNode,
+  $findMatchingParent,
   $getState,
   $setState,
   createState,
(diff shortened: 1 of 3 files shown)


etrepum:
An alternative approach would be that the shortcut still works in these scenarios but it would split the parent link accordingly. Might make sense in the case where AutoLinkNode is involved especially when it's marked as disabled. The transform should probably handle that edge case better (removing disabled AutoLinkNode when they are sufficiently changed to no longer parse as a link).

levensta:
> Might make sense in the case where AutoLinkNode is involved especially when it's marked as disabled. There transform should probably handle that edge case better (removing disabled AutoLinkNode when they are sufficiently changed to no longer parse as a link).


I think the case with AutoLinkNode disabled would be better handled in AutoLinkExtension, because now, with unlinked autolink, you can write any text inside, unlike when the node is linked. Currently, it seems that valid text checking is disabled if the node is unlinked. I think it's right to check the valid text for AutoLinkNode regardless of the linked/unlinked state

https://github.com/user-attachments/assets/e7b7e488-3ed5-4fbc-80e4-64a830364b23

etrepum:
Yes, the transform in the AutoLink extension is what I was referring to in that comment. Not strictly relevant to this transform which is why I thought it made sense to approve and merge as-is

PR #8165, opened 2026-02-23
[lexical-link] Bug Fix: Enable autolink matching when it unlinked

levensta:
## Description

Previously, it was possible to enter invalid characters in the text of an unlinked AutoLinkNode, which made the link invalid. The fix activates parsing of valid links in AutoLinkNode if they are marked as unlinked

### Before

https://github.com/user-attachments/assets/d3100cb7-f297-4d35-80fd-84b45f9536af

### After

https://github.com/user-attachments/assets/16113a8d-c484-4d98-b23b-77bde6c0884d

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
@@ -753,7 +753,7 @@ test.describe.parallel('Auto Links', () => {
     );
   });
 
-  test('Unlinked the autolink should not destruct if add non-spacing text in front or right after it', async ({
+  test('Unlinked autolink is preserved when adding punctuation before or after it', async ({
     page,
     isPlainText,
   }) => {
@@ -840,7 +840,7 @@ test.describe.parallel('Auto Links', () => {
     );
   });
 
-  test('Can destruct unlinked the autolink if add an invalid character inside', async ({
+  test('Adding an invalid character will destruct an unlinked autolink', async ({
     page,
     isPlainText,
   }) => {
@@ -894,7 +894,7 @@ test.describe.parallel('Auto Links', () => {
     );
   });
 
-  test('Can destruct unlinked the autolink if add emoji inside', async ({
+  test('Adding an emoji inside an unlinked autolink will destruct it', async ({
     page,
     isPlainText,
   }) => {


packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
```diff
@@ -753,6 +753,149 @@ test.describe.parallel('Auto Links', () => {
     );
   });
 
+  test('Unlinked the autolink should not destruct if add non-spacing text in front or right after it', async ({
```
levensta: Actually, if you insert non-space characters before or after the link, it also invalidates it if you consider the text as a whole (visually, it is a whole). I don't have a strong opinion on this case, so for now, I've fixed the current behavior with this test

https://github.com/user-attachments/assets/f2ddf49e-9f03-4a87-a0c5-4a5806c72075

The original PR https://github.com/facebook/lexical/pull/6306 was sent with the aim of resetting the link styles, but I don't know if users can have a link and text in a single string without a space between them, and whether this makes sense

etrepum: Other autolink implementations (e.g. google docs, github flavored markdown) will still autolink when the punctuation goes after the link, e.g. `www.google.com!` will autolink `www.google.com`. I don't have a super strong opinion about it, since you can always manually link text, but the GFM style autolink behavior is probably what more people are used to. Either way this PR's scope doesn't need to cover that sort of change.

levensta:
> I'm not sure this fixes the whole problem, e.g. if you have an unlinked AutoLinkNode and add a decorator like date or emoji it will split the AutoLinkNode instead of destroying any of them

If during transformation one of the child nodes is not a TextNode or is not simple, the link will be destroyed. I added a test with emoji

levensta:
The tests are failing for the second time in a specific job `e2e-tests / windows (22.x, chromium, rich-text, legacy-events) / e2e-test`. I don't understand what the problem might be yet, maybe it's worth marking the tests as flaky or rewriting them as unit tests in LexicalAutoLinkExtension

etrepum:
I'm not sure either without looking a lot more closely, which is hard for me since I don't have a windows computer to test with. unit tests would be fine, it's unclear if marking them as flaky would be a sufficient workaround. Really depends on what kind of timing issue it is.

levensta:
I have no idea what's wrong with this test. I couldn't find any information about what exactly changes the legacy events job and what browser version is used, but I tried running the production build on Win 10 Chrome 86 through an online emulator and reproducing the scenario, and I get the expected result

https://github.com/user-attachments/assets/7e554826-d2e3-4001-be15-5b2a7e7a0130

I would suggest skipping this test specifically for legacy events + chromium + windows

levensta:
By the way, the video shows v0.40.dev, although I ran the build and preview on 0.41 🤷‍♂️ 

image

levensta:
Well, it took some time to figure out that the `LEGACY_EVENTS` variable disables the `beforeInput` event, and this has nothing to do with the version of Playwright 😅

The truth turned out to be that even in the current version of lexical, without changes from this PR, there is a problem that this test revealed, and it also occurs on MacOS

https://github.com/user-attachments/assets/cff1d8af-af45-46bd-b46e-1779f3288c1d

https://playground.lexical.dev/?disableBeforeInput=true

levensta:
I also see that in `LEGACY_EVENTS` mode, there is a delay between key type/press events, which may have affected the fact that the tests passed successfully on other operating systems
https://github.com/facebook/lexical/blob/main/packages/lexical-playground/__tests__/utils/index.mjs#L109-L114

I checked locally that the node transform handler is called once in both cases

packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
```diff
@@ -753,6 +753,206 @@ test.describe.parallel('Auto Links', () => {
...
+    page,
+    isPlainText,
+  }) => {
+    test.skip(isPlainText || LEGACY_EVENTS);
```
levensta: As a result, the current changes in this PR do not bring anything new to this scenario. It already fails in prod in legacy_events mode, so I added a condition to skip the test https://github.com/facebook/lexical/pull/8165#issuecomment-3969634838

Tests directly related to the new behavior are written below.

etrepum: I think legacy events are probably vestigial since we don't support browser versions that old anymore. AFAICT Chrome 60+ does not need it, and we don't support anything older than Chrome 86.

packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
```diff
@@ -753,6 +753,206 @@ test.describe.parallel('Auto Links', () => {
...
+        <p dir="auto">
+          <span data-lexical-text="true">http://example.co[m</span>
+        </p>
+      `,
+      undefined,
+      {ignoreClasses: true},
+    );
+  });
+
+  test('Can destruct unlinked the autolink if add emoji inside', async ({
```
etrepum:
```suggestion
  test('Adding an emoji inside an unlinked autolink will destruct it', async ({
```

packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
```diff
@@ -753,6 +753,206 @@ test.describe.parallel('Auto Links', () => {
...
+            <span data-lexical-text="true">http://example.com</span>
+          </span>
+        </p>
+      `,
+      undefined,
+      {ignoreClasses: true},
+    );
+  });
+
+  test('Can destruct unlinked the autolink if add an invalid character inside', async ({
```
etrepum:
```suggestion
  test('Adding an invalid character will destruct an unlinked autolink', async ({
```

packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
```diff
@@ -753,6 +753,206 @@ test.describe.parallel('Auto Links', () => {
     );
   });
 
+  test('Unlinked the autolink should not destruct if add non-spacing text in front or right after it', async ({
```
etrepum:
```suggestion
  test('Unlinked autolink is preserved when adding punctuation before or after it', async ({
```

I am not sure this behavior is strictly desired since this is not how it behaves when it is not unlinked, but I guess it's fine to encode the current behavior in a test until it is intentionally changed.

etrepum (approved):
Just adding some suggestions to the test names to make them a bit easier to understand, I'll go ahead and apply these changes directly

packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
levensta: Sounds like a good reason to remove legacy-events tests jobs. At the same time, I could refactor the tests and remove all stubs for legacy-events mode.

I can work on this and submit a separate PR

packages/lexical-playground/__tests__/e2e/AutoLinks.spec.mjs
levensta: Yep, it's just an additional test that was missing initially

PR #8167, opened 2026-02-25
[security] Fix: Override minimatch@3 to 3.1.3 to resolve CVE-2022-3517

PikkaPikkachu:
Summary                                                                                                                                                  
                                                            
  - Adds pnpm.overrides to pin minimatch@3 to 3.1.3 in the root package.json and three example package.json files that maintain their own lock files       
                                                                                                                                                           
  Why overrides?

  minimatch < 3.1.3 has a HIGH severity ReDoS vulnerability (CVE-2022-3517) via the braceExpand function. The vulnerable version 3.1.2 is pulled in as a transitive dependency by several packages:

  - eslint@8.57.0
  - eslint-plugin-import@2.29.1
  - eslint-plugin-jsx-a11y@6.8.0
  - eslint-plugin-react@7.34.1
  - docusaurus-lunr-search@2.4.2 (via docusaurus-plugin-internaldocs-fb)
  - @eslint/eslintrc@2.1.4, @humanwhocodes/config-array@0.11.14 (transitive from eslint)

  None of these packages have released a version that drops minimatch@^3.1.2. Even the latest releases of all three eslint plugins still depend on it. The override "minimatch@3": "3.1.3" targets only the 3.x resolution, bumping 3.1.2 to the backward-compatible patch 3.1.3. It does not affect minimatch@5.1.6, @9.0.5, or @10.1.1 used by other packages.

  Test plan
  - pnpm run build passes
  - pnpm run test-unit passes (2581/2581 tests)
  - pnpm install --lockfile-only to regenerate lock files (blocked by npm registry outage for @atlaskit packages — unrelated to this change)

### DIFF
--- a/examples/extension-sveltekit-ssr-hydration/package.json
+++ b/examples/extension-sveltekit-ssr-hydration/package.json
@@ -54,6 +54,11 @@
     "vitest": "^3.2.3",
     "vitest-browser-svelte": "^0.1.0"
   },
+  "pnpm": {
+    "overrides": {
+      "minimatch@3": "3.1.3"
+    }
+  },
   "optionalDependencies": {
     "@rollup/rollup-linux-x64-gnu": "4.52.0",
     "@rollup/rollup-darwin-arm64": "4.52.0",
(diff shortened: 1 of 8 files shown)


etrepum:
This should be low priority and is basically busywork, none of this code runs server-side or is shipped as part of lexical. eslint only runs locally and in CI so unexpectedly high CPU usage with untrusted input isn’t a problem of its own (in both cases you have access to run arbitrary code already and you’re not consuming input other than the code itself).

Upgrading to a modern eslint would be a worthwhile task, but it would not be trivial.

PR #8168, opened 2026-02-25
[lexical-link] Feature: Merging two identical links into one node

### LINKED ISSUE #8150: Feature: Merging two identical links into one node
## Description

In lexical, it is possible for two LinkNodes with identical URLs to be adjacent to each other. This situation can be reproduced manually: insert a `TextNode` or `ElementNode` that will split the link into two nodes, and then delete the inserted element.

https://github.com/user-attachments/assets/af851eed-a3c7-4411-9aeb-2ea9c2c7fa4b

## Impact

It would be effective to merge the contents of adjacent links into a single node, as there is no compelling reason to maintain different nodes that lead to the same URL and do not contain any other content between them.

Jynx2004:
## Description

In this approach when a user tries to paste any text inside the link , then the text gets injected inside the text of the link directly instead of creation of new TextNode in between the link thus dividing the text of the same link into 2 halfs . This prevents creation of 2 same links . 

Closes #8150

### Before

Screenshot 2026-02-25 135832

### After

Screenshot 2026-02-25 135749


packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -225,15 +225,62 @@ export function $insertGeneratedNodes(
...
+    const parentElementAfter = endNode.getParent();
+
+    if (parentElementBefore == null || parentElementAfter == null) {
+      return false;
+    }
+
+    return (
+      parentElementAfter.getType() === 'link' &&
+      parentElementAfter.getType() === parentElementBefore.getType()
+    );
```
etrepum, line 260: This isn't an acceptable approach. @lexical/clipboard can't depend on @lexical/link, and you can't use getType to check the type of a node since nodes can be subclasses with a different type.

packages/lexical-clipboard/src/clipboard.ts
```diff
@@ -225,15 +225,62 @@ export function $insertGeneratedNodes(
...
+
+      if (textNodes.length > 0) {
+        const firstText = textNodes[0];
+        selection.insertText(firstText.getTextContent());
+        $updateSelectionOnInsert(selection);
+      }
+    } else {
+      selection.insertNodes(nodes);
+      $updateSelectionOnInsert(selection);
+    }
```
etrepum, line 281: This doesn't look correct since all other nodes will be discarded if one TextNode exists. It also ignores any formatting.

PR #8172, opened 2026-02-25
[Fix] : Linebreak insertion breaks format inheritance

### LINKED ISSUE #8160: Bug: linebreak insertion breaks format inheritance
Lexical version: 0.40.0

## Steps To Reproduce

1. Type bolded "hello" in the editor
2. Move the cursor between "he" and "llo"
3. Press the Shift + Enter key twice
4. Move the cursor between the two linebreaks
5. Start Typing any character

Link to code example: [playground](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE52RwU7DMBBE_2XOVuUAIoo_gDNSjhUHEy-NxdaONpuKKMq_IweqUg5I9GTPrHfmSV5AIWqWVr0S3ALJWcvZ9ZGDUILb_xKB1EeGswZvWY5e4SqDYw4Eh1QchsGoMxcDBkofCoeeyn0eirtZBieSMeYEV61mOc84JnoV8u__f3ArGnP-i-3FIEShTjedJuZLfkmJKVDSrfY7Y_DiD-KH_iroq-_pQlZke6a5oWb7rCvU1YD9qK0_UYCr6rp6bOqqaez9ncGYJ-nK3jP7-SB5SuHnNuzuwe4s1k-gt5gIFQIAAA)

## The current behavior

https://github.com/user-attachments/assets/4d7e1fee-1229-47f9-a53e-f1b6a67da29a

## The expected behavior

If a user inserts linebreaks within a formatted range, those linebreaks should maintain the formatting.

Jynx2004:
## Description
Instead of sending  value 0 as a static value we take the current format from selection.format and pass it to the respective function responsible for format change.
As a result the format gets passed on between multiple linebreaks.

Closes #8160

### Before

Screenshot 2026-02-26 004335

### After

Screenshot 2026-02-26 004320

### DIFF
--- a/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
+++ b/packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
@@ -1394,7 +1394,7 @@ describe('getNodes()', () => {
 
           // Get inserted node
           const nodes = newSelection.getNodes();
-          expect(nodes.length).toBeGreaterThan(0);
+          expect(nodes).toHaveLength(1);
 
           const insertedNode = nodes[0];
           assert($isTextNode(insertedNode), 'Expected TextNode');


Jynx2004:
> I will review this after it has new tests to show that the change has the desired behavior

```
` -$updateSelectionFormatStyle(selection, 0, '');
  +$updateSelectionFormatStyle(selection, selection.format, '');`
```
Here only the passed argument is changed instead of sending static 0 value we have passed the actual format which will not need any new tests . It can be tested on existing tests.

etrepum:
There needs to be a new test that shows that this change does something useful, and that future changes to lexical preserve that behavior. The existing test suite is not sufficient to show that your code works as expected.

Jynx2004:
> There needs to be a new test that shows that this change does something useful, and that future changes to lexical preserve that behavior. The existing test suite is not sufficient to show that your code works as expected.

Will create some tests for testing this issue.

Jynx2004:
> There needs to be a new test that shows that this change does something useful, and that future changes to lexical preserve that behavior. The existing test suite is not sufficient to show that your code works as expected.

I have added the test case to test this feature

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -1367,6 +1368,54 @@ describe('getNodes()', () => {
...
+          const textNode = $createTextNode('hello');
+          textNode.toggleFormat('bold');
+          paragraph.append(textNode);
+          $getRoot().append(paragraph);
+
+          // Place cursor between "he" and "llo"
+          const selection = $createRangeSelection();
+          selection.anchor.set(textNode.getKey(), 2, 'text');
+          selection.focus.set(textNode.getKey(), 2, 'text');
+          $setSelection(selection);
```
etrepum:
```suggestion
          const selection = textNode.select(2, 2);
```

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -1367,6 +1368,54 @@ describe('getNodes()', () => {
...
+          selection.insertLineBreak();
+          selection.insertLineBreak();
+
+          // Get updated selection
+          const newSelection = $getSelection();
+          expect($isRangeSelection(newSelection)).toBe(true);
+
+          if (!$isRangeSelection(newSelection)) {
+            throw new Error('Expected RangeSelection');
+          }
```
etrepum:
```suggestion
          assert($isRangeSelection(newSelection), 'Expected RangeSelection');
```

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -1367,6 +1368,54 @@ describe('getNodes()', () => {
...
+
+          // Type a character between the line breaks
+          newSelection.insertText('X');
+
+          // Get inserted node
+          const nodes = newSelection.getNodes();
+          expect(nodes.length).toBeGreaterThan(0);
```
etrepum:
```suggestion
          expect(nodes).toHaveLength(1);
```

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -1367,6 +1368,54 @@ describe('getNodes()', () => {
...
+
+          const insertedNode = nodes[0];
+          expect($isTextNode(insertedNode)).toBe(true);
+
+          if (!$isTextNode(insertedNode)) {
+            throw new Error('Expected TextNode');
+          }
```
etrepum:
```suggestion
          assert($isTextNode(insertedNode), 'Expected TextNode');
```

PR #8181, opened 2026-02-27
[lexical] Bug Fix: Preserve raw IME input when Chinese composition is cancelled

ywsoul:
## Bug

When a user types pinyin (e.g. `ceshi`) in a Chinese IME on an empty Lexical editor and presses Enter to dismiss the IME without selecting a candidate, the editor clears the text instead of preserving the raw pinyin input.

### Root Cause

On macOS Chrome with the native Chinese IME, pressing Enter to cancel composition triggers:
1. `compositionend(data='')` — the browser clears the DOM text
2. `insertText(data='ceshi')` — the browser restores the raw pinyin as committed text

Previously, Lexical would schedule the text node for removal upon seeing the empty composition result in step 1. By the time step 2 fires, the `insertText` event would either append to stale text (causing duplication like `cesce's`) or insert into a node that was about to be removed (causing text loss).

## Fix

Two changes to `$updateTextNodeFromDOMContent` in `LexicalUtils.ts`:

1. **Clear stale text immediately** — `node.setTextContent('')` ensures the browser's subsequent `insertText` event inserts into a clean node rather than appending to old composition text
2. **Defer removal with content check** — the `setTimeout` callback now checks if the Lexical node has been repopulated by `insertText` before removing it

## Test Plan

- Added E2E test: `Can cancel Chinese pinyin IME on empty editor and preserve raw input via insertText`
- All 16 Composition E2E tests pass
- All 2574 unit tests pass (2 pre-existing failures unrelated to this change)

### DIFF
--- a/packages/lexical/src/LexicalUtils.ts
+++ b/packages/lexical/src/LexicalUtils.ts
@@ -801,11 +801,25 @@ export function $updateTextNodeFromDOMContent(
       if (normalizedTextContent === '') {
         $setCompositionKey(null);
         if (!IS_SAFARI && !IS_IOS && !IS_APPLE_WEBKIT) {
-          // For composition (mainly Android), we have to remove the node on a later update
+          // For composition (mainly Android), we have to remove the node on a later update.
+          // We defer removal to allow the browser to finish processing the
+          // compositionend event. Some browsers (e.g. Chrome on macOS) fire an
+          // additional `insertText` event after compositionend that restores the
+          // user's raw IME input (e.g. pinyin letters). If we remove the node
+          // synchronously, that restored text is lost.
+          //
+          // We also clear the node's text content immediately so that any
+          // subsequent insertText event from the browser inserts into a clean
+          // node rather than appending to stale composition text.
           const editor = getActiveEditor();
+          node.setTextContent('');
           setTimeout(() => {
             editor.update(() => {
               if (node.isAttached()) {
+                const currentText = node.getTextContent();
+                if (currentText !== '') {
+                  return;
+                }
                 node.remove();
               }
             });
--- a/packages/lexical-playground/__tests__/e2e/Composition.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Composition.spec.mjs
@@ -1392,6 +1392,138 @@ test.describe('Composition', () => {
         `,
       );
     });
+
+    test('Can type Chinese pinyin via IME on empty editor and confirm', async ({
+      page,
+      browserName,
+    }) => {
+      // We don't yet support FF.
+      test.skip(browserName !== 'chromium');
+
+      await focusEditor(page);
+      await enableCompositionKeyEvents(page);
+
+      const client = await page.context().newCDPSession(page);
+
+      // Simulate Chinese IME: user types pinyin on an initially empty editor.
+      // This triggers the COMPOSITION_START_CHAR insertion because anchor.type === 'element'.
+      await client.send('Input.imeSetComposition', {
+        selectionStart: 1,
+        selectionEnd: 1,
+        text: 'n',
+      });
+      await client.send('Input.imeSetComposition', {
+        selectionStart: 2,
+        selectionEnd: 2,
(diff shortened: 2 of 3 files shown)


packages/lexical/src/LexicalUtils.ts
```diff
@@ -801,11 +801,25 @@ export function $updateTextNodeFromDOMContent(
...
+          // compositionend event. Some browsers (e.g. Chrome on macOS) fire an
+          // additional `insertText` event after compositionend that restores the
+          // user's raw IME input (e.g. pinyin letters). If we remove the node
+          // synchronously, that restored text is lost.
+          //
+          // We also clear the node's text content immediately so that any
+          // subsequent insertText event from the browser inserts into a clean
+          // node rather than appending to stale composition text.
           const editor = getActiveEditor();
+          node.setTextContent('');
```
etrepum, line 815: The reconciler will remove empty text nodes from the document. Not sure this is working the way you think it is

etrepum:
Can you show a video of a reproduction for this? I'm not able to reproduce on macOS on any browser with the following steps:

* Start with an empty editor and the Pinyin IME https://playground.lexical.dev/?emptyEditor=true
* Type some input (e.g. nihao)
* Press enter to dismiss

etrepum:
There are also no unit test failures, so I'm not sure where that information is coming from in your test plan

ywsoul:
> Can you show a video of a reproduction for this? I'm not able to reproduce on macOS on any browser with the following steps:
> 
> * Start with an empty editor and the Pinyin IME https://playground.lexical.dev/?emptyEditor=true
> * Type some input (e.g. nihao)
> * Press enter to dismiss


Thank you for your reply.

This is the problematic gif I recorded.

emptyInput

This is the gif after my local run repair, and the result is in line with the interaction expectations of the browser.

emptyInput-fix

packages/lexical/src/LexicalUtils.ts
ywsoul: Image

The above gif is the performance after annotating the code.

This line of code actually fixes the duplicate string caused by the input being intercepted.

etrepum: Does node.remove() have the same effect?

etrepum:
I still don't know how to reproduce what you've recorded, the expected behavior happens for me with the current playground:

https://github.com/user-attachments/assets/539b6bc6-a15d-4627-973a-3d332ef2c007

This is using macOS 26.3 and Chrome 145.0.7632.117 with the Pinyin - Simplified IME

ywsoul:
> I still don't know how to reproduce what you've recorded, the expected behavior happens for me with the current playground:
> 
>  chrome-ime.mov 
> This is using macOS 26.3 and Chrome 145.0.7632.117 with the Pinyin - Simplified IME

ok, i get it 

I changed several input methods and finally found that this is an input method compatibility problem. The Xunfei input method I used was developed by an algorithm company in China. This input method is ok in other editors, but it is not effective in lexical. Therefore, this case is up to you to judge whether it needs to be combined.

etrepum:
Maybe it would be better to report this as a bug to Xunfei, rather than add code to lexical to work around a bug in that non-standard IME? Since we don't have any automated CI coverage for mobile devices it can be risky to change this code.

ywsoul:
> Maybe it would be better to report this as a bug to Xunfei, rather than add code to lexical to work around a bug in that non-standard IME? Since we don't have any automated CI coverage for mobile devices it can be risky to change this code.

OK, thank you very much for your patient follow-up. I will turn off the PR later. At that time, I mentioned PR to Lexical because Xunfei was normal in my daily use, but there was a problem in Lexical, and thanks to Lexical, it was open source. Yes, I have found a solution, so I hope to solve this problem that others may also encounter. I understand your concern, so I plan to go to Xunfei to have a look.

Finally, I am a fan of Lexical. I like this editor very much. It realizes all my fantasies about the editor. I am also recommending Lexical to my friends around me. I wish you all the best!

PR #8183, opened 2026-02-27
[lexical-table][lexical-playground] Feature: nested tables resize themselves if hasFitNestedTables: true

randal-atticus:
## Description
#8097 implemented nested table fitting when inserting tables. However, beyond the initial insertion, it didn't handle the nested table itself changing width (1), or the surrounding cell changing width (2).

This PR changes the original implementation to a Node Transform, which handles (1).
To handle (2), the simplest approach is to mark all child tables as dirty when a column resize event occurs, so that the transform fires. I tried to do this with a Transform as well, but had issues getting it to terminate. Open to alternative ideas.

### Before

https://github.com/user-attachments/assets/5074869e-2ca1-471f-9660-13dc3d954919

### After

https://github.com/user-attachments/assets/4a3a591b-67e4-4a1a-b486-60fb35fad1bc

### DIFF
--- a/packages/lexical-table/src/LexicalTablePluginHelpers.ts
+++ b/packages/lexical-table/src/LexicalTablePluginHelpers.ts
@@ -457,9 +457,7 @@ export function registerTablePlugin(
             }
           }
         }
-        const resizeRoots = $calculateResizeRootTables(
-          Array.from(modifiedTables),
-        );
+        const resizeRoots = $calculateResizeRootTables(modifiedTables);
         resizeRoots.forEach((root) => {
           $resizeDOMColWidthsToFit(editor, root);
         });
@@ -758,8 +756,8 @@ function getTotalTableWidth(colWidths: readonly number[]) {
 
 // Returns the subset of tables that are not contained by any of the other tables in
 // the input.
-function $calculateResizeRootTables(tables: TableNode[]) {
-  const inputTables = new Set<LexicalNode>(tables);
+function $calculateResizeRootTables(tables: ReadonlySet<TableNode>) {
+  const inputTables: ReadonlySet<LexicalNode> = tables;
   const roots: TableNode[] = [];
   for (const table of tables) {
     if (


packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -739,8 +739,11 @@ function $getCellWidth(cell: TableCellNode) {
 
 /**
  * Returns horizontal insets of the given cell (padding + border).
+ *
+ * @param cell - The cell to calculate the horizontal insets for.
+ * @returns The horizontal insets of the cell, in pixels.
  */
-function $calculateCellInsets(cell: TableCellNode) {
+function $calculateCellHorizontalInsets(cell: TableCellNode) {
```
randal-atticus: I was thinking of exporting this, and $getCellWidth, so that a similar pattern could be used for things like ImageNodes (basically, any node that can be resized)?

etrepum:
Haven’t looked at the code but node transforms run before the DOM is updated so layout can’t be (correctly) calculated at that time. At best you can read the previously rendered DOM, but there are some big edge cases when you are working with a different version of the document.

packages/lexical-table/src/LexicalTablePluginHelpers.ts
etrepum: Adding exports is fine as long as the functions are suitable for being maintained as public API, or marked as `@internal`

randal-atticus:
Hmm, true. For my awareness, can you elaborate on those edge cases?

If the transform doesn't find the DOM node, at the moment it will just return zero insets, i.e. the resulting nested will be slightly too wide - unfortunate but not fatal.

original table, from `main:`
Screenshot 2026-03-02 at 11 03 12 am
pasted into this branch:
Screenshot 2026-03-02 at 11 03 22 am


If this won't work, some alternatives:
* Replace `$calculateCellHorizontalInsets` with some property  (`tableCellHorizontalInsets: number`) or user-provided callback (`(cell: TableCellNode) => number`). This API is fairly rough/unergonomic so I would probably end up moving this entire thing to another plugin (e.g. `<NestedTableAutoResizerPlugin getTableCellInsets={...} />`
* Keep measuring the DOM but use an update listener, and accept the downside of the nested update.

randal-atticus:
I've gone ahead and added a property to the extension `getCellHorizontalInsets` that allows the user to provide a callback of the insets. It's a callback because different cells might have different styling (e.g. nested tables might have less padding, or users might have configurable table styling via state).

I don't love that you need to provide both properties for this to work, so it might make sense to pull this out to another plugin.

packages/lexical-table/src/LexicalTablePluginHelpers.ts
randal-atticus: Updated to expose `$getCellWidth`. This specific function (`$calculateCellHorizontalInsets`) is now a user-specified callback and no longer needs to be exported.

etrepum:
The edge cases are basically that subsequent versions of the lexical document can have completely arbitrary changes from one to the next. Cells could be moved to different rows or resized, but if you're trying to measure their layout from a transform you're going to either get nothing or whatever the metrics were from the last version. In many cases this could be fine, but in cases where something happens that would affect the metrics of the cell after the new version is rendered you're going to have a stale read that wouldn't be fixed up until some future update that happens to trigger a transform of the same node.

If I were designing something like this I would probably use a mutation and/or update listener to handle it and maybe see if I could do most of it with direct DOM manipulation and not pushing the state back into the nodes (this kind of mechanism would be more responsive, since not all devices would necessarily measure things the same way). If not, then a cascading update to do that state update with history-merge or something like that (only if something actually needs the extra update cycle). Possibly some hybrid of the two approaches where the DOM updates are done eagerly and the render of the cascading update is mostly a no-op.

randal-atticus:
@etrepum Cool idea with direct DOM manipulation. Updated the PR accordingly (triggered by a mutation listener). One effect of this change is that, since the underlying colWidths are not modified, if you shrink then restore a parent cell, the inner table reverts to its original size.

https://github.com/user-attachments/assets/6a98753d-e4e8-4bad-86a8-e46d7739198c

I also made it measure widths against the nearest _root or shadow root_, rather than just the nearest table cell. This makes tables also refit themselves inside things like the collapsible containers or column layouts in the Playground. In effect, that means maybe this feature is more aptly named `hasFitTables` instead?

https://github.com/user-attachments/assets/eeef6296-755d-4925-8b32-a6b077b74456

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -444,6 +441,22 @@ export function registerTablePlugin(
...
+        if (mutation === 'created' || mutation === 'updated') {
+          editor.getEditorState().read(() => {
+            const tableNode = $getNodeByKey<TableNode>(nodeKey);
+            if (!tableNode) {
+              return;
+            }
+            $resizeDOMColWidthsToFit(editor, tableNode);
+          });
+        }
+      }
```
etrepum: Will be more efficient to set up one read instead of N

```suggestion
      editor.getEditorState().read(() => {
        for (const [nodeKey, mutation] of nodeMutations) {
          if (mutation === 'created' || mutation === 'updated') {
            const tableNode = $getNodeByKey<TableNode>(nodeKey);
            if (tableNode) {
              $resizeDOMColWidthsToFit(editor, tableNode);
            }
          }
        }
      });
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -676,76 +688,33 @@ function $insertTableIntoGrid(
...
+ * @returns The horizontal insets of the node, in pixels.
  */
-function $calculateCellInsets(cell: TableCellNode) {
-  const cellDOM = $getEditor().getElementByKey(cell.getKey());
-  if (cellDOM === null) {
-    return 0;
-  }
-  const computedStyle = window.getComputedStyle(cellDOM);
+function $calculateHorizontalInsets(dom: HTMLElement) {
+  const computedStyle = window.getComputedStyle(dom);
```
etrepum: we should be getting the _window from the editor

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -444,6 +441,22 @@ export function registerTablePlugin(
...
+        return;
+      }
+      for (const [nodeKey, mutation] of nodeMutations) {
+        if (mutation === 'created' || mutation === 'updated') {
+          editor.getEditorState().read(() => {
+            const tableNode = $getNodeByKey<TableNode>(nodeKey);
+            if (!tableNode) {
+              return;
+            }
+            $resizeDOMColWidthsToFit(editor, tableNode);
```
etrepum: This should probably be done in two phases since it's already recursive. First phase is to collect the tables and the second phase would be to figure out the table roots that need to be resized (e.g. via `$getCommonAncestor`). This way we're not doing redundant work to recursively resize a subtable we've already resized.

randal-atticus: Done

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -444,6 +442,26 @@ export function registerTablePlugin(
...
+
+      editor.getEditorState().read(() => {
+        const modifiedTables = nodeMutations
+          .entries()
+          .filter(
+            ([, mutation]) => mutation === 'created' || mutation === 'updated',
+          )
+          .map(([nodeKey]) => $getNodeByKey<TableNode>(nodeKey))
+          .filter((table) => table !== null)
+          .toArray();
```
etrepum: These iterator methods are not available in all of the browser versions we support, they are less than a year old (in Safari at least). Probably best to just rewrite this as a for loop that accumulates into a `Set<TableNode>`

etrepum (commented):
Small runtime compat issue that needs to be sorted but otherwise the code looks good and it works well in my testing at https://lexical-playground-2vjr0b74e-fbopensource.vercel.app/?hasNestedTables=true&tableHorizontalScroll=false&hasFitNestedTables=true

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -444,6 +442,29 @@ export function registerTablePlugin(
...
+        for (const [nodeKey, mutation] of nodeMutations) {
+          if (mutation === 'created' || mutation === 'updated') {
+            const tableNode = $getNodeByKey<TableNode>(nodeKey);
+            if (tableNode) {
+              modifiedTables.add(tableNode);
+            }
+          }
+        }
+        const resizeRoots = $calculateResizeRootTables(
+          Array.from(modifiedTables),
```
etrepum: I suggested `Set` for `modifiedTables` because the first thing `$calculateResizeRootTables` does is convert the input `Array` to a `Set` and it only reads from the `Set` so it could take a `ReadonlySet` as input

```suggestion
          modifiedTables,
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -771,47 +752,76 @@ function $calculateCellInsets(cell: TableCellNode) {
...
 
-function $getTotalTableWidth(colWidths: readonly number[]) {
+function getTotalTableWidth(colWidths: readonly number[]) {
   return colWidths.reduce((curWidth, width) => curWidth + width, 0);
 }
 
+// Returns the subset of tables that are not contained by any of the other tables in
+// the input.
+function $calculateResizeRootTables(tables: TableNode[]) {
+  const inputTables = new Set<LexicalNode>(tables);
```
etrepum:
```suggestion
function $calculateResizeRootTables(tables: ReadonlySet<TableNode>) {
  const inputTables: ReadonlySet<LexicalNode> = tables;
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
randal-atticus: Oops, my bad

PR #8186, opened 2026-02-28
[lexical-playground] Fix: add fallback for dimensionless images to prevent collapse

### LINKED ISSUE #7511: Bug: Adding SVG without explicit dimensions fails
Adding an SVG, that doesn't have explicitly set 'height' and 'width' in the svg tag when inserted populates width and height as inherit and does not render the image inside the editor, even though it's present correctly in the editor state.

Some work was done by @kirandash in https://github.com/facebook/lexical/pull/7388 but the fallback for the dimensions do not appear to be working.

https://github.com/user-attachments/assets/e42484f5-1a60-4059-b1f1-3b461a566903

Sa-Te:
## Description

This PR fixes an issue where SVGs inserted without explicit `width` and `height` attributes (such as dimensionless SVGs or base64 data URIs) would visually collapse to `0x0` in the editor.

### The Problem
When a dimensionless SVG or a base64 SVG string is parsed:
1. The browser reports `naturalWidth` and `naturalHeight` as `0`.
2. The `ImageNode` constructor defaults missing dimensions to `'inherit'`.
3. In the React layer (`ImageComponent`), the previous fallback logic relied on a fragile `isSVG` check that failed for base64 data URIs.
4. Because the intrinsic dimensions were `0`, the component rendered with `0px` inline styles, making the node invisible and unselectable.

### The Fix
I pivoted from a global fallback to a **Surgical SVG Fallback** to maintain compatibility with Lexical's existing E2E test suite:
-  Updated the logic to identify SVGs using both file extensions and `data:image/svg+xml` prefixes.
- **Preserved CSS Inheritance:** Standard images (JPG/PNG) continue to return `'inherit'` for dimensions. This ensures that existing E2E tests (which rely on specific HTML snapshots for standard images) remain green and global CSS remains in control of responsive sizing.
- **SVG Math Safety Net:** For SVGs specifically, if no intrinsic dimensions are found, the logic now falls back to the node's `maxWidth`. This ensures a visible bounding box is rendered, allowing the user to see, select, and resize the image.

Fixes #7511

## Test Plan

- **New Test:** Added `Images.spec.ts` -> "Dimensionless SVG renders with a visible bounding box instead of collapsing". This verifies that a base64 encoded SVG without attributes renders with a width/height > 0.
- **Regressions:** Verified that standard image snapshots (e.g., the "Yellow Flower" JPG in `Toolbar.spec.mjs`) still receive `style="width: inherit; height: inherit"` to satisfy existing Playwright assertions.
- **Local Verification:** Ran `npx playwright test packages/lexical-playground/__tests__/e2e/` and confirmed all 600+ tests pass.

### DIFF
--- a/packages/lexical-playground/src/nodes/ImageComponent.tsx
+++ b/packages/lexical-playground/src/nodes/ImageComponent.tsx
@@ -118,7 +118,11 @@ function useSuspenseImage(src: string): ImageStatus {
 }
 
 function isSVG(src: string): boolean {
-  return src.toLowerCase().endsWith('.svg');
+  const lowerCaseSrc = src.toLowerCase();
+  return (
+    lowerCaseSrc.endsWith('.svg') ||
+    lowerCaseSrc.startsWith('data:image/svg+xml')
+  );
 }
 
 function LazyImage({
@@ -140,7 +144,6 @@ function LazyImage({
   width: 'inherit' | number;
   onError: () => void;
 }): JSX.Element {
-  const isSVGImage = isSVG(src);
   const status = useSuspenseImage(src);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/lexical-playground/src/nodes/ImageComponent.tsx
```diff
@@ -155,7 +154,19 @@ function LazyImage({
...
+    if (width !== 'inherit' && height !== 'inherit') {
+      return {
+        height,
+        maxWidth,
+        width,
+      };
+    }
+
+    const isActuallySVG =
+      isSVG(src) || src.toLowerCase().startsWith('data:image/svg+xml');
```
etrepum: Why not change isSVG accordingly?

PR #8187, opened 2026-03-02
[lexical-table] Refactor: use monolithic listener for table SELECTION_CHANGE_COMMAND and deselection handler

randal-atticus:
## Description
As discussed in #8107 the SELECTION_CHANGE_COMMAND and various pointer events are registered for each table in `applyTableHandler`. There is also a lot of recursive updates going on. It would be better to have a monolithic handler that handles table selection events, as it makes it far easier to reason about what is happening when you have nested tables. 

This PR tackles some low hanging fruit:
* SELECTION_CHANGE_COMMAND is hoisted up to `registerTableSelectionObserver` (i.e. only once). It still runs the old logic (`$handleTableSelectionChangeCommand`) on each table in sequence though
* `pointerDownCallback`, which handles clearing table highlight selections, is similarly hoisted up.

Next up is a refactor of `$handleTableSelectionChangeCommand` so that it just runs once and behaves as expected across all the tables.

### Before

N/A, this is a refactor

### After

N/A no behaviour changes, all tests passing

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -86,6 +86,10 @@ async function fillTablePartiallyWithText(page) {
 }
 
 const WRAPPER = IS_TABLE_HORIZONTAL_SCROLL ? [0] : [];
+const nthTableSelector = (nth) =>
+  IS_TABLE_HORIZONTAL_SCROLL
+    ? `div.PlaygroundEditorTheme__tableScrollableWrapper:nth-of-type(${nth}) > table`
+    : `table:nth-of-type(${nth})`;
 
 test.describe.parallel('Tables', () => {
   test(`Can a table be inserted from the toolbar`, async ({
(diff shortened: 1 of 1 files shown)


packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -377,7 +381,61 @@ export function registerTableSelectionObserver(
     {skipInitialization: false},
   );
 
+  const unregisterSelectionChange = editor.registerCommand(
+    SELECTION_CHANGE_COMMAND,
+    () => {
+      for (const [, [tableObserver]] of tableSelections) {
```
etrepum:
```suggestion
      for (const [tableObserver] of tableSelections.values()) {
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -377,7 +381,61 @@ export function registerTableSelectionObserver(
...
+      }
+
+      const rootElement = editor.getRootElement();
+      if (!rootElement) {
+        return;
+      }
+      editor.update(() => {
+        const selection = $getSelection();
+        if ($isTableSelection(selection) && rootElement.contains(target)) {
+          for (const [, [observer]] of tableSelections) {
```
etrepum:
```suggestion
          for (const [observer] of tableSelections.values()) {
```

packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -377,7 +381,61 @@ export function registerTableSelectionObserver(
...
+
+    editorWindow.addEventListener('pointerdown', pointerDownCallback);
+    return () => {
+      editorWindow.removeEventListener('pointerdown', pointerDownCallback);
+    };
+  }
+
+  const unregisterWindowHandlers = applyWindowHandlers();
+
   return () => {
```
etrepum: might make sense to refactor this into a mergeRegister like most of lexical uses

randal-atticus:
I'll add more multi-table coverage. The nested table scenarios are not well tested, because the current behaviour (when using the mouse) is frankly quite broken, with seemingly nondeterministic outcomes. I tried to add some tests for that in #8107 but I feel like it would now be better to include new nested table tests alongside the a more stable selection handler.

etrepum (approved):

PR #8188, opened 2026-03-02
[lexical-playground] Chore: Remove legacy-events mode

### LINKED ISSUE #5673: Bug: Enter incorrectly handled in Safari after inserting new image (legacy events)
Lexical version: 0.13.1

## Steps To Reproduce

1. Open https://playground.lexical.dev/?disableBeforeInput=true in Safari 
2. Clear editor Cmd + A + Delete
3. Click Insert -> Image -> Sample
4. Press Enter

## The current behavior

Nothing happens

https://github.com/facebook/lexical/assets/1125900/e64615b2-6ea6-4fd4-812d-b605e924d890

## The expected behavior

Same as w/o Legacy events. New paragraph should be created

https://github.com/facebook/lexical/assets/1125900/f8482b6e-2465-4457-861a-3be7dee74ae9

Related to #5672

levensta:
## Description

Legacy-events mode tested lexical's functionality by emulating the unavailability of the `beforeInput` event. Currently, lexical supports the minimum browser versions that already support the `beforeInput` event.

- https://github.com/facebook/lexical/pull/8141
- https://developer.mozilla.org/en-US/docs/Web/API/Element/beforeinput_event#browser_compatibility

This PR completely removes GitHub workflows with legacy events mode and any mentions in tests

Closes #5673

### Before

Running tests in Github Actions CI was divided into two modes: modern-events and legacy-events. Some test were skipped in legacy-events mode

### After

All workflows with legacy events have been removed. Test skips for legacy events mode have been removed

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Images.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Images.spec.mjs
@@ -816,4 +816,37 @@ test.describe('Images', () => {
       `,
     );
   });
+
+  test('Dimensionless SVG renders with a visible bounding box instead of collapsing', async ({
+    page,
+    isRichText,
+    isCollab,
+  }) => {
+    test.skip(!isRichText || isCollab);
+    await initialize({page});
+    await focusEditor(page);
+
+    await focusEditor(page);
+
+    // 1. Create a raw SVG without width/height attributes
+    const svgContent = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"><circle cx="50" cy="50" r="40" fill="red"/></svg>`;
+    const base64Svg = `data:image/svg+xml;base64,${Buffer.from(svgContent).toString('base64')}`;
+
+    // 2. Use the repository's native helper instead of hacking the clipboard!
+    await insertUrlImage(page, base64Svg, 'dimensionless-svg');
+
+    // 3. Locator for the image (using the same selector pattern as the other tests)
+    const imageLocator = page.locator(
+      '.editor-image img[alt="dimensionless-svg"]',
+    );
+    await imageLocator.waitFor({state: 'attached', timeout: 5000});
+
+    // 4. Verification: The bounding box should NOT be 0x0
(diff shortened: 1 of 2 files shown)


levensta:
Although the `disableBeforeInput` option is hidden from the UI, it is still available when adding a query parameter in the playground. Should it be kept?

https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/Settings.tsx#L45-L51

etrepum:
> Although the `disableBeforeInput` option is hidden from the UI, it is still available when adding a query parameter in the playground. Should it be kept?
> 
> https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/Settings.tsx#L45-L51

No

etrepum (commented):
I think you should have a close look at all of the conditional statements that have changed, removing just the subexpressions that referenced the event mode is not always the right thing to do depending on whether it evaluates to true or false with modern events and which logical operator was used.

.github/workflows/call-e2e-test.yml
```diff
@@ -19,11 +18,10 @@ jobs:
   e2e-test:
     runs-on: ${{ inputs.os }}
     continue-on-error: ${{ inputs.flaky }}
-    if: (inputs.browser != 'webkit' || inputs.os == 'macos-latest') && (inputs.editor-mode != 'rich-text-with-collab' || inputs.events-mode != 'legacy-events')
+    if: (inputs.browser != 'webkit' || inputs.os == 'macos-latest') && (inputs.editor-mode != 'rich-text-with-collab')
```
etrepum:
```suggestion
    if: (inputs.browser != 'webkit' || inputs.os == 'macos-latest')
```

Incorrect subexpression removal

levensta: `inputs.editor-mode != 'rich-text-with-collab'` OR `inputs.events-mode != 'legacy-events'` are two independent conditions

etrepum: No they’re not. `(_ || true)` is always true.

Issue #123, opened 2021-02-22
Recent regression with textNode.splitText


trueadm:
Since merging https://github.com/facebookexternal/Outline/pull/121, we now encounter issues with focus restoration. Here's a repro using the playground:

- Toggle on character limit
- Type text until you hit limit
- Should fatal

`Unhandled Rejection (IndexSizeError): Failed to execute 'setBaseAndExtent' on 'Selection': The offset 31 is larger than the node's length (30).`

PR #8191, opened 2026-03-04
[lexical-link] Feature: Allow custom separators for AutoLink creation

### LINKED ISSUE #8189: Feature: Add colon to allowed separators for the AutoLinkPlugin
## Description

`AutolinkPlugin` has a hard-coded list of valid separators acceptable at the end of the matched string.
```ts
const PUNCTUATION_OR_SPACE = /[.,;\s]/;
```

With custom matchers which link certain tags into urls (e.g., GH-12345 -> `https://github.com/facebook/lexical/issues/12345 `) it's not possible to write something like this:

```
GH-12345:
* Investigated the issue...
```

This is common style of writing weekly reports - "in issue <tag> I did the following..." - followed by an unordered list.

### Use Cases

GH-12345:
- investigated ...
- resolved ...

JIRA-12345:
- addressed ...

My proposal is to extend this allowlist with a colon (`:`) or allow the list to be modified by the developers via AutoLinkPlugin property.

## Impact

Hundreds of engineers in our company write weekly reports and this is a fairly common syntax to describe what was done in various issues, features, etc. We have several tags defined to reference various external tools.

BryanRPLee:
## Description

**Current behavior:**

`AutoLinkPlugin` / `registerAutoLink` uses a hard-coded internal regex to decide what characters are valid boundaries around an auto-linked match:

```ts
const PUNCTUATION_OR_SPACE = /[.,;\s]/;
```

This means a colon (`:`) is never treated as a valid separator. As a result, patterns like:

```
GH-12345:
- investigated the issue...
```

…do **not** produce an auto-link for `GH-12345`, even with a custom matcher, because the colon immediately following the tag is not recognised as a valid boundary character.

**Changes introduced by this PR:**

- Adds an optional `separators` field to `AutoLinkConfig` (in `LexicalAutoLinkExtension.ts`) that accepts a `RegExp`.
- When provided, this regex replaces the internal `PUNCTUATION_OR_SPACE` pattern used by `isSeparator`, `endsWithSeparator`, `startsWithSeparator`, and all related boundary-validation functions.
- When omitted, behaviour is **fully backwards-compatible** — the existing `/[.,;\s]/` default is used.
- Adds unit tests covering: colon as separator, colon as leading separator, multiple matches, non-separator blocking link creation, default fallback, and link-unwrap on separator removal.

Closes #8189

---

### Before

Typing `GH-123:` with a custom GH-tag matcher produces **no auto-link** — the colon is not a valid separator so the boundary check fails and the text remains plain.

### After
Screenshot 2026-03-04 at 5 30 07 pm

With `separators: /[.,;:\s]/` passed to `registerAutoLink` / `<AutoLinkPlugin>`:

- `GH-123:` -> `[GH-123](:)` — the tag is linked and the colon remains as a plain text node beside it.
- `GH-123 `, `GH-123.`, `GH-123;` continue to work exactly as before.
- `GH-123abc` is **not** linked (non-separator character).
- Removing the colon and replacing it with a non-separator character destroys the link (correct unwrap behaviour).

**Automated tests added:**

```
packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
-  default separators: colon after match prevents link creation
-  custom separators including colon: link created when followed by colon
-  custom separators: link preceded by colon is created
-  custom separators: typical weekly-report pattern "GH-123:\n- item" creates link
-  custom separators: link still created when followed by space (default behaviour preserved)
-  custom separators: link still created when followed by period (default behaviour preserved)
-  custom separators: link NOT created when followed by a non-separator character
-  custom separators: multiple matches in one paragraph all linked
-  omitting separators falls back to default /[.,;\s]/ behaviour
-  link is unwrapped when a non-separator is typed after it
```

### DIFF
--- a/packages/lexical-link/src/LexicalAutoLinkExtension.ts
+++ b/packages/lexical-link/src/LexicalAutoLinkExtension.ts
@@ -83,6 +83,9 @@ function findFirstMatch(
 
 const PUNCTUATION_OR_SPACE = /[.,;\s]/;
 
+const DEFAULT_IS_SEPARATOR: (char: string) => boolean =
+  PUNCTUATION_OR_SPACE.test.bind(PUNCTUATION_OR_SPACE);
+
 function endsWithSeparator(
   textContent: string,
   isSeparator: (char: string) => boolean,
@@ -527,13 +530,13 @@ export interface AutoLinkConfig {
   matchers: LinkMatcher[];
   changeHandlers: ChangeHandler[];
   excludeParents: Array<(parent: ElementNode) => boolean>;
-  isSeparator: (char: string) => boolean;
+  isSeparator?: (char: string) => boolean;
 }
 
 const defaultConfig: AutoLinkConfig = {
   changeHandlers: [],
   excludeParents: [],
-  isSeparator: PUNCTUATION_OR_SPACE.test.bind(PUNCTUATION_OR_SPACE),
+  isSeparator: DEFAULT_IS_SEPARATOR,
   matchers: [],
 };
 
@@ -541,7 +544,8 @@ export function registerAutoLink(
   editor: LexicalEditor,
   config: AutoLinkConfig = defaultConfig,
 ): () => void {
-  const {matchers, changeHandlers, excludeParents, isSeparator} = config;
+  const {matchers, changeHandlers, excludeParents} = config;
+  const isSeparator = config.isSeparator ?? DEFAULT_IS_SEPARATOR;
   const onChange: ChangeHandler = (url, prevUrl) => {
     for (const handler of changeHandlers) {
       handler(url, prevUrl);
--- a/packages/lexical-playground/src/plugins/AutoLinkPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/AutoLinkPlugin/index.tsx
@@ -5,7 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-
 import type {JSX} from 'react';
 
 import {$isCodeNode} from '@lexical/code';
@@ -21,6 +20,8 @@ const URL_REGEX =
 const EMAIL_REGEX =
   /(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))/;
 
+const GH_TAG_REGEX = /GH-\d+/;
+
 const MATCHERS = [
   createLinkMatcherWithRegExp(URL_REGEX, (text) => {
     return text.startsWith('http') ? text : `https://${text}`;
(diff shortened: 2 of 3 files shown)


packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -523,6 +533,7 @@ export function registerAutoLink(
   config: AutoLinkConfig = defaultConfig,
 ): () => void {
   const {matchers, changeHandlers, excludeParents} = config;
+  const separators = config.separators ?? DEFAULT_PUNCTUATION_OR_SPACE;
```
etrepum: If following the general convention of providing the default in the default config then you won't need this logic


```suggestion
const {matchers, changeHandlers, excludeParents, isSeparator} = config;
```

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -510,6 +519,7 @@ export interface AutoLinkConfig {
   matchers: LinkMatcher[];
   changeHandlers: ChangeHandler[];
   excludeParents: Array<(parent: ElementNode) => boolean>;
+  separators?: RegExp;
```
etrepum: The convention for extension config is to always provide a default value whenever reasonable.

The general interface here is based on functions (LinkMatcher is `(text: string) => LinkMatcherResult | null`) and not specifically regex so it would probably make more sense for this to be a `isSeparator: (char: string) => boolean` which could be implemented with a regex or not, e.g. the default could be `PUNCTUATION_OR_SPACE.test.bind(PUNCTUATION_OR_SPACE)`

```suggestion
  isSeparator: (char: string) => boolean;
```

ConditionedHair:
> The build and tests are failing because the react plugin needs to be updated to support the additional config

Hi, I'm currently adding to Playground + passing isSeparator props to AutoLinkPlugin.ts, that was what was failing.

I've also updated the description of this PR to show images of functional tests in Playground

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -510,11 +530,13 @@ export interface AutoLinkConfig {
   matchers: LinkMatcher[];
   changeHandlers: ChangeHandler[];
   excludeParents: Array<(parent: ElementNode) => boolean>;
+  isSeparator?: (char: string) => boolean;
```
etrepum, line 533:
```suggestion
  isSeparator: (char: string) => boolean;
```

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -523,6 +545,7 @@ export function registerAutoLink(
   config: AutoLinkConfig = defaultConfig,
 ): () => void {
   const {matchers, changeHandlers, excludeParents} = config;
+  const isSeparator = config.isSeparator ?? DEFAULT_IS_SEPARATOR;
```
etrepum, line 548: To maintain backwards compatibility something like this should work

```suggestion
  config?: Partial<AutoLinkConfig>,
): () => void {
  const {matchers, changeHandlers, excludeParents, isSeparator} = {...defaultConfig, ...config};
```

etrepum:
It also looks like the changes to the definition of isSeparator in the playground has broken e2e tests that cover autolinking URLs that have an embedded colon (for the port specifier). Might be a good idea to add unit test coverage for those cases so that it's quicker to debug.

Issue #7824, opened 2025-09-17
Bug: backslash injection in numbered list edge case


joh-ku:
Lexical version: v0.35.0

## Description:

I've encountered an issue where the markdown editor mistakenly injects an additional backslash (`\`). This seems to be an edge case that occurs when a numbered list pattern is used within a bullet point list.

### Step 1

Insert the below markdown content into the [lexical playground](https://playground.lexical.dev/):

```
A test:

- 1\. foo
- 5.: bar
```

**Note**: Without the backslash, the first list item, `- 1\. foo` will be interpreted as bullet list and numbered list both at the same time.

### Step 2

Switch between WYSIWYG and markdown mode.

## Expected Behavior:

The escaped dot (`1\.`) should remain unchanged and render correctly without triggering numbered list formatting. The content should display as a simple bullet point list item in the WYSIWYG editor.

## Actual Behavior:

Upon switching back to the normal editor, the markdown content is altered to:

```
- 1\\. minute: that's the first item of a test list
```

The editor adds an extra backslash, resulting in `1\.` in WYSIWYG mode.

## Use Case:

I want to display a bullet point list where each item starts with a number (representing a time-stamp like a minute of an event) followed by a description. The number should not trigger automatic numbered list formatting.

## Additional Context:

This bug was observed using [Sveltia CMS](https://github.com/sveltia/sveltia-cms) and the [downstream bug report can be found here](https://github.com/sveltia/sveltia-cms/issues/512).

## Link to code example

The issue can be reproduced by copying the above (Step 1) sample into the [lexical playground](https://playground.lexical.dev/) and switch between WYSIWYG and markdown mode.

## Impact of fix

Everybody, who wants to mix bullet point and numbered list items, would benefit from this fix.

PR #8192, opened 2026-03-04
[lexical-markdown] Bug Fix: update backslash escape handling to align with CommonMark

kimseongyu:
## Description

This PR aligns with the [CommonMark Spec](https://spec.commonmark.org/0.31.2/#backslash-escapes) regarding backslash escapes.
1. Improved the text parsing regex to handle backslash escapes more accurately.
2. Added the logic that converted backslashes to hard line breaks.
3. Updated incorrect link test cases and refactored related logic.

The following cases can not supported before update of the related functions:
```
`` \[\` ``
<https://example.com?find=\*>
```

Related Issue #7824

Close the related issue after addressed the single line recursive list.

### Before

https://github.com/user-attachments/assets/6ea83beb-eab1-4429-a331-6db25f2e552c

### After

https://github.com/user-attachments/assets/675ea719-64ba-480f-ad53-df421d658fe9

### DIFF
--- a/packages/lexical-markdown/src/MarkdownTransformers.ts
+++ b/packages/lexical-markdown/src/MarkdownTransformers.ts
@@ -693,7 +693,8 @@ export const LINK: TextMatchTransformer = {
     const [, linkText, rawLinkUrl, rawLinkTitle] = match;
 
     const linkUrl = rawLinkUrl != null ? unescapeText(rawLinkUrl) : undefined;
-    const linkTitle = rawLinkTitle != null ? unescapeText(rawLinkTitle) : undefined;
+    const linkTitle =
+      rawLinkTitle != null ? unescapeText(rawLinkTitle) : undefined;
     const linkNode = $createLinkNode(linkUrl, {title: linkTitle});
     const openBracketAmount = linkText.split('[').length - 1;
     const closeBracketAmount = linkText.split(']').length - 1;


packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -685,7 +688,23 @@ export const LINK: TextMatchTransformer = {
...
+    ): string | undefined => {
+      if (value == null) {
+        return undefined;
+      }
+      return value
+        .replace(/\\([!-/:-@[-`{-~])/g, '$1')
+        .replace(/&#(\d+);/g, (_, codePoint) =>
+          String.fromCodePoint(Number(codePoint)),
+        );
+    };
```
etrepum: I see this implemented more than once, should be extracted to an exported function that can also be used by importTextTransformers

packages/lexical-markdown/src/utils.ts
```diff
@@ -462,3 +462,16 @@ export function isEmptyParagraph(node: LexicalNode): boolean {
       MARKDOWN_EMPTY_LINE_REG_EXP.test(firstChild.getTextContent()))
   );
 }
+
+export function unescapeText(
+  value: string | undefined | null,
+): string | undefined {
+  if (value == null) {
+    return undefined;
+  }
```
etrepum:
```suggestion
export function unescapeText(value: string): string {
```

It would be better to do the null checking outside of this function, the call sites can change accordingly

packages/lexical-markdown/src/importTextTransformers.ts
```diff
@@ -135,10 +136,6 @@ export function importTextTransformers(
 
   // Handle escape characters
   const textContent = textNode.getTextContent();
-  const escapedText = textContent
-    .replace(/\\([*_`~\\])/g, '$1')
-    .replace(/&#(\d+);/g, (_, codePoint) => {
-      return String.fromCodePoint(codePoint);
-    });
-  textNode.setTextContent(escapedText);
+  const unescapedText = unescapeText(textContent) ?? '';
```
etrepum:
```suggestion
  const unescapedText = unescapeText(textContent);
```

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -685,7 +690,10 @@ export const LINK: TextMatchTransformer = {
     if ($findMatchingParent(textNode, $isLinkNode)) {
       return;
     }
-    const [, linkText, linkUrl, linkTitle] = match;
+    const [, linkText, rawLinkUrl, rawLinkTitle] = match;
+
+    const linkUrl = unescapeText(rawLinkUrl);
+    const linkTitle = unescapeText(rawLinkTitle ?? '');
```
```suggestion
    const linkUrl = rawLinkUrl != null ? unescapeText(rawLinkUrl) : undefined;
    const linkTitle = rawLinkTitle != null ? unescapeText(rawLinkTitle) : undefined;
```

etrepum:
GitHub actions are currently broken so I will have to check back in later

kimseongyu:
I've updated the code style to comply with Prettier.

PR #8193, opened 2026-03-05
[lexical-table] Bug Fix: Improve nested table selection by using monolithic pointer event handling

randal-atticus:
## Description
Continuation of #8187.

Previously, clicking a nested table would result in each table firing its own pointerDown handler, and also register their own window.onPointerMove handlers, resulting in difficulty clicking on any cells in nested tables (see "Before").

This PR moves to register just a single pointerDown event for all tables (and associated onPointerMove/onPointerUp). Now these handlers only run against the deepest relevant table.

**This is not the final PR in this series**. Some additional handling is needed to correctly exit `isHighlightingCells` mode, as well as handle dragging between inner-table cells and the parent.

Currently no new E2E tests because the previous behaviour is quite hard to test reliably.

### Before

Observe that clicking and dragging on cells in nested tables often results in the parent table being selected as well (because its onPointerMove handler is firing as well)

https://github.com/user-attachments/assets/75a38ae9-14ff-4e90-ab81-7fb264ba643d

### After

https://github.com/user-attachments/assets/c889b47c-c130-471a-876f-107ef8ac2ddc


packages/lexical-table/src/LexicalTablePluginHelpers.ts
```diff
@@ -349,42 +350,8 @@ export function registerTableSelectionObserver(
...
-
-    editorWindow.addEventListener('pointerdown', pointerDownCallback);
-    return () => {
-      editorWindow.removeEventListener('pointerdown', pointerDownCallback);
-    };
-  }
-
   return mergeRegister(
-    registerWindowHandlers(),
+    registerTableWindowHandlers(editor, tableSelections),
```
randal-atticus, line 352: This is leaking the Map of TableObservers out, but felt better than putting a bunch of selection-specific code into LexicalTablePluginHelpers. Perhaps a better way would to pass down a `TableObservers` class that limits the exposure?

etrepum: I think it's fine for internal implementation details to leak across the implementation, it's not part of the API

etrepum:
GitHub actions are currently broken so I will have to check back in later

PR #8195, opened 2026-03-06
[lexical-table] Bug Fix: Prevent single-cell table selection after exiting table selection

randal-atticus:
## Description

Bug introduced in #8193, I think. We were not calling $clearHighlight of a table when clicking on it, meaning it stayed in "highlight mode". This made the table selection mode feel sticky, because any subsequent pointerMove events (which are very eager to fire) would highlight the entire cell. The only way to exit the mode was to very precisely click without moving the mouse.

### Before

https://github.com/user-attachments/assets/5729e097-9777-4e02-9211-fd49c91d4bdb

### After

https://github.com/user-attachments/assets/5314b030-7722-4086-aaf9-52e81c02ca3f

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -7609,13 +7609,13 @@ test.describe.parallel('Tables', () => {
 
     // Click a cell in the same table
     await dragMouse(
-      pageOrFrame,
+      page,
       await selectorBoundingBox(
-        pageOrFrame,
+        page,
         `${nthTableSelector(1)} > tr:first-of-type > th:first-of-type`,
       ),
       await selectorBoundingBox(
-        pageOrFrame,
+        page,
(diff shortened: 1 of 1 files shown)


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -842,9 +837,9 @@ export function $handleTableSelectionChangeCommand(
         return true;
       }
     } else if (
-      focusCell !== tableObserver.anchorCell &&
       tableObserver.anchorCell !== null &&
       tableObserver.anchorCellNodeKey !== null &&
+      focusCell.elem !== tableObserver.anchorCell.elem &&
```
randal-atticus, line 842: also had to change this because even when they were the same cell, they were not referentially equal

etrepum:
Looks like there's some sort of conflict with collab in this PR that needs further investigation

randal-atticus:
Seems to have just been an issue in one test, even though other tests failed. Fixed.

Issue #6575, opened 2024-08-29
Migrate lexical-code from PrismJS to Shiki


ivailop7:
PrismJS appears to be no longer maintained and Shiki is a lighter weight still maintained library for syntax highlighting. A number of people has suggestion the migration and I agree it makes sense, primarily because PrismJS is not ESM and creates issues when used with certain server-side rendering frameworks.

etrepum:
I have looked into this a little bit, the downsides are:

* Shiki is larger since it brings in a wasm build of the oniguruma regex engine (~500kb gzip I think) required to support the TextMate grammars that all of the syntax highlighting is defined with https://shiki.style/guide/bundles
* It's designed more for ahead-of-time highlighting rather than as-you-type highlighting
* It uses inline style rather than semantic classes

Probably not a deal-breaker for everyone but I assume that there are many people that would prefer the dead-end prismjs solution at <100kb

ivailop7:
Not tied to the framework, but it appears everyone I've spoken to recently is switching to this one, short of Monaco and CodeMirror embeds. Size-wise - yes, 5x increase is significant. If there are other reasonable alternatives, I'm all ears.

etrepum:
I didn't find anything better, which was disappointing. Maybe the ideal would be to properly divorce highlighting from the code node model so that either solution (or something bespoke, e.g. a custom treesitter based highlighter) could be used. For the most part it really is just a display concern, shouldn't need such tight coupling.

etrepum:
Looks like https://github.com/wooorm/starry-night is a similar project to shiki but AST and class based

etrepum:
Looks like the size concerns with Shiki are being addressed https://github.com/shikijs/shiki/pull/761

etrepum:
Given that everything in prism is global you should be able to add languages by importing them before you import lexical code. Removing them is a bit trickier but you could configure your bundler to make some of them return an empty module or to eliminate those imports altogether.

vadimkantorov:
Also, I've had to import Prism like so:

```
import Prism from "prismjs"; if (typeof globalThis.Prism === 'undefined') { globalThis.Prism = Prism;}
import {CodeNode} from '@lexical/code';
```

Otherwise, I was getting an exception at runtime of global variable `Prism` not available:
- https://github.com/remix-run/remix/discussions/8182

kyoshino:
[...]

Issue #5424, opened 2023-12-26
Bug: lexical/markdown is hardcoded to transform CODE blocks and not emit highlight nodes


sherrellbc:
I was (am) having issues getting code highlighting to work with Lexical/markdown. It seems to maybe not support it? (Lexical/code _does_ seem to, but the way the markdown code was written seems to be hardcoded to using non-highlight APIs). It also was written strangely such that CODE blocks are hardcoded to always transform even if no transformer is given for them. As such, is it not possible to override the behavior for CODE transforms with the latest version.

1. `$convertFromMarkdownString` and an empty transformer list, CODE blocks are still converted
2. `$convertFromMarkdownString` explicitly uses `$createCodeNode` and not `$createCodeHighlightNode` and as a result code highlighting does not appear to work. Though perhaps I am doing something wrong here.

Regarding (2) I see the `<code>` HTML is created with `data-highlight-language=XXX` but after glancing at the code I'm not convinced highlighting is properly applied? It seems to be a secondary step.

Lexical version: 0.12.2

## Steps To Reproduce

1. Use `$convertFromMarkdownString` with a `transformers` argument as an empty array or with e.g. QUOTE
```
$convertFromMarkdownString('this is text\n> this is a quote\n```c\nthis is code\n```\n> this is a quote', [QUOTE])
```
2. Observe no code highlighting and forced transformations on code blocks
If an empty transformer array is given then no QUOTE transformations occur as expected, but CODE transforms still do.

<img width="397" alt="Screenshot 2023-12-26 at 2 02 03 PM" src="https://github.com/facebook/lexical/assets/8580880/ae95002c-2105-4412-b676-a073f16ca13b">
 

Link to code example: https://codesandbox.io/p/sandbox/awesome-goldberg-vgj4s2

## The current behavior

The function `$convertFromMarkdownString` implicitly transforms CODE blocks with internal logic despite passing no CODE transformer that would otherwise match the blocks.

## The expected behavior

Calling `$convertFromMarkdownString` with an empty array of transformers would effectively be a no-op.

2wheeh:
I found dynamically opting code transformer is marked as TODO here:
https://github.com/facebook/lexical/blob/9d247acc42998b4de3ed079069584ebde701640b/packages/lexical-markdown/src/MarkdownImport.ts#L62-L64

sherrellbc:
[...]

Issue #4550, opened 2023-05-25
Feature: Split markdown shortcuts, import and export


fantactuka:
Current `@lexical/markdown` package supports shortcuts (auto-replace while typing), import and export relying on the same transformers configuration. It also uses RegExp parsing that does not have any external dependencies, which is great for bundle size. On the other hand it has few disadvantages:
- If app only uses shortcuts (no import/export), transformers would still require all import/export related properties
- Transformers API became quite bloated with more expansion requests (#4500)
- RegExp matching is nowhere near to be as good as proper tokenization (#2632, #2214)

Related issues/PRs: #4449, #4500, #4339, #4461, and [more](https://github.com/facebook/lexical/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Amarkdown)

I think we can split these into separate features (or even packages), where shortcuts will keep relying on lightweight RegExp matching and simplified transformers configuration; import/export will use 3rd party dependency for proper tokenization (these features are less common and also can be lazy-loaded for certain scenarios, keeping initial bundle small)

// cc @themagickoala @xinyuan0801 @hanford would appreciate if you guys have any suggestions/ideas (e.g., regarding 3rd party parser) given you've had PPs for Lexical MD

themagickoala:
@fantactuka thanks for thinking of me! https://github.com/themagickoala/lexical-remark is using the remark ecosystem for MD parsing, and it's working a treat. I was able to add video support in a day, and just released collapsible container support in 2 days. These were things that the playground wasn't supporting for import/export, so it was nice to be able to get done!

I can understand the approach you've taken with RegExp, not wanting to add dependencies to the project. Once you start going down that 3rd party route, it definitely bloats things a little. If you do end up being keen on remark, I'm happy to consult as I've become very familiar with it over the past few weeks 😆.

Supporting a separate shortcuts system would be a great idea. I've not started looking at shortcuts in lexical-remark, so I can't give much advice there, but I think RegExp is probably sufficient for matching while typing!

vadimkantorov:
Yeah, plugging a reliable markdown importer/exporter via a widely-used package (like remark?) should be better... Currently many small issues exist with Lexical's support for markdown like https://github.com/facebook/lexical/issues/7234 - likely stemming from the current regexp-based approach and focus on simplicity...

vadimkantorov:
[...]

PR #8196, opened 2026-03-07
[lexical-code][lexical-markdown] Use lightweight code-node import in …

### LINKED ISSUE #5381: Bug: prismjs is included as a dependency of @lexical/markdown

gurkerl83:
## Summary

Decouple `@lexical/markdown` code-block internals from the Prism-heavy `@lexical/code` root export by introducing and consuming a lightweight `@lexical/code/node` subpath.

## Problem

`@lexical/markdown` imported code-node APIs from `@lexical/code` root, which pulls Prism-related modules into markdown-only dependency graphs.

## Changes

- Added new public subpath export: `@lexical/code/node`
  - Exposes only node-focused APIs:
    - `$createCodeNode`
    - `$isCodeNode`
    - `CodeNode`
    - `DEFAULT_CODE_LANGUAGE`
    - `getDefaultCodeLanguage`
    - `SerializedCodeNode` (type)
- Switched markdown internals to import from `@lexical/code/node`:
  - `MarkdownTransformers.ts`
  - `MarkdownShortcuts.ts`
  - `utils.ts`
- Added TS path mappings for `@lexical/code/node`.
- Added regression guardrail test to ensure markdown source does not import `@lexical/code` root.

## Compatibility

No markdown API behavior changes (`TRANSFORMERS`, `CODE`, import/export/shortcuts behavior remain unchanged). 
Apps that need syntax highlighting should continue to set it up explicitly via `@lexical/code` (highlight node/plugin); this PR only removes implicit Prism inclusion from markdown-only paths.

## Validation

- `pnpm exec vitest run packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts packages/lexical-markdown/src/__tests__/unit/MarkdownTransformers.test.ts`
- `pnpm exec vitest run packages/lexical-code/src/__tests__/unit/LexicalCodeNode.test.ts`
- `pnpm exec vitest run scripts/__tests__/unit/build.test.ts scripts/__tests__/unit/markdown-deps.test.ts`

Bundle proof (markdown-only entry via esbuild metafile):
- `packages/lexical-code/src/FacadePrism.ts`: **not present**
- `prismjs/components/prism-*`: **not present**

Fixes #5381
Related to #4550 #5424 #6575

### DIFF
--- a/packages/lexical-code/flow/LexicalCodeNode.js.flow
+++ b/packages/lexical-code/flow/LexicalCodeNode.js.flow
@@ -1,0 +1,17 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+export type {SerializedCodeNode} from './LexicalCode';
+export {
+  $createCodeNode,
+  $isCodeNode,
+  CodeNode,
+  DEFAULT_CODE_LANGUAGE,
+  getDefaultCodeLanguage,
+} from './LexicalCode';
--- a/packages/lexical-code/package.json
+++ b/packages/lexical-code/package.json
@@ -9,7 +9,6 @@
   ],
   "license": "MIT",
   "version": "0.41.0",
-  "main": "LexicalCode.js",
   "types": "index.d.ts",
   "dependencies": {
     "@lexical/utils": "workspace:*",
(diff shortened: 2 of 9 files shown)


etrepum:
I’m not sure that this is the best approach. At some point the prism stuff needs to be moved out. I don’t know if it’s worth making this awkward extra module to support that doesn’t follow other lexical conventions and also doesn’t make any progress into extracting prism highlighting into its own package.

gurkerl83:
Fair point, I can work on this strategy as well 🙂 Another point is that lexical/code is adding also a tone of stuff on lexical/markdown…

etrepum:
The approach here is also not complete for what it is doing, because a lot of tsconfig.json and package.json are maintained by scripts (but due to the nature of json they can't have comments warning about that) so this change would cause problems as soon as those scripts are run.

gurkerl83:
@etrepum Made a second attempt here https://github.com/facebook/lexical/pull/8197

PR #8197, opened 2026-03-07
[lexical-code][lexical-markdown] Extract Prism; make code fences opt-in

### LINKED ISSUE #5381: Bug: prismjs is included as a dependency of @lexical/markdown

gurkerl83:
## Summary
This PR removes the Prism-driven coupling between markdown and code packages by doing two coordinated changes:

1. Extract Prism highlighting ownership from `@lexical/code` into `@lexical/code-prism`.
2. Make markdown code-fence support explicit composition (instead of implicit default behavior in `@lexical/markdown`).

## Context
A previous attempt ([#8196](https://github.com/facebook/lexical/pull/8196)) proposed a lightweight `@lexical/code/node` subpath to avoid importing the Prism-heavy `@lexical/code` root from markdown internals.

Based on maintainer feedback ([comment](https://github.com/facebook/lexical/pull/8196#issuecomment-4016637112)), this PR follows the structural direction instead:
- move Prism out to a dedicated package,
- avoid introducing an interim awkward subpath.

## Problem
`@lexical/markdown` and `@lexical/code` were coupled such that optional code/highlighting paths affected markdown-only use cases. This made modular composition weaker than intended.

### 1) Prism extraction
- Added `@lexical/code-prism` package.
- Moved Prism-specific runtime/highlighting logic there.
- Updated internal consumers (playground/integration fixtures) to import Prism APIs from `@lexical/code-prism`.

### 2) `@lexical/code` slim-down
- `@lexical/code` keeps code-node/core behavior.
- Prism ownership is no longer in `@lexical/code` root.

### 3) Markdown composition change
- `@lexical/markdown` no longer has a direct runtime dependency on `@lexical/code`.
- `TRANSFORMERS` no longer includes code-fence behavior by default.
- Added `createMarkdownCodeBlockTransformer(...)` so code-fence behavior can be composed explicitly by consumers.
- Added multiline transformer `isNode` predicate support so shortcut skip behavior remains correct for composed node types.

### 4) Docs/tests updates
- Updated markdown docs and React plugin docs to show explicit code-fence composition.
- Added/updated unit tests for:
  - default core transformers (no code-fence -> `CodeNode`),
  - explicit composed code-fence behavior.

## Semver-visible behavior change
`@lexical/markdown` default `TRANSFORMERS` is now code-agnostic.  
Consumers that want fenced code block conversion to `CodeNode` must compose it explicitly.

## Migration
```ts
import {
  TRANSFORMERS,
  createMarkdownCodeBlockTransformer,
} from '@lexical/markdown';
import {$createCodeNode, $isCodeNode, CodeNode} from '@lexical/code';

const CODE = createMarkdownCodeBlockTransformer({
  $createCodeNode,
  $isCodeNode,
  dependencies: [CodeNode],
});

const transformers = [...TRANSFORMERS, CODE];
```

## Validation
- `pnpm exec vitest run packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts packages/lexical-markdown/src/__tests__/unit/MarkdownTransformers.test.ts`
- `pnpm exec vitest run packages/lexical-markdown/src/__tests__/unit/LexicalMarkdownCodeComposition.test.ts`
- `pnpm exec vitest run scripts/__tests__/unit/build.test.ts`
- `pnpm run flow`
- `pnpm run tsc`

## Issue links
Fixes #5381  
Related to #4550 #5424 #6575

### DIFF
--- a/packages/lexical-code-prism/README.md
+++ b/packages/lexical-code-prism/README.md
@@ -1,0 +1,5 @@
+# `@lexical/code-prism`
+
+[![See API Documentation](https://lexical.dev/img/see-api-documentation.svg)](https://lexical.dev/docs/api/modules/lexical_code_prism)
+
+This package contains Prism-based syntax highlighting utilities for Lexical code blocks.
--- a/packages/lexical-code/README.md
+++ b/packages/lexical-code/README.md
@@ -2,4 +2,6 @@
 
 [![See API Documentation](https://lexical.dev/img/see-api-documentation.svg)](https://lexical.dev/docs/api/modules/lexical_code)
 
-This package contains the functionality for the code blocks and code highlighting for Lexical.
+This package contains the node and utility APIs for Lexical code blocks.
+
+Prism-based syntax highlighting is provided by [`@lexical/code-prism`](https://www.npmjs.com/package/@lexical/code-prism).
--- a/packages/lexical-markdown/README.md
+++ b/packages/lexical-markdown/README.md
@@ -22,6 +22,27 @@ editor.update(() => {
 });
 ```
 
+### Code Fence Composition
+
+For code-fence support (` ``` ` blocks), compose a code transformer explicitly:
+```js
+import {
+  TRANSFORMERS,
+  createMarkdownCodeBlockTransformer,
+} from '@lexical/markdown';
+import {$createCodeNode, $isCodeNode, CodeNode} from '@lexical/code';
+
+const CODE = createMarkdownCodeBlockTransformer({
+  $createCodeNode,
+  $isCodeNode,
+  dependencies: [CodeNode],
+});
+
+const transformers = [...TRANSFORMERS, CODE];
+```
+
+Use `transformers` for markdown import/export and markdown shortcut registration.
+
 It can also be used for initializing editor's state from markdown string. Here's an example with react `<RichTextPlugin>`
 ```jsx
 <LexicalComposer initialConfig={{
@@ -59,7 +80,6 @@ are processed during import, export or while typing. `@lexical/markdown` package
 ```js
 // Element transformers
 UNORDERED_LIST
-CODE
 HEADING
 ORDERED_LIST
 QUOTE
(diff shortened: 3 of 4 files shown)


etrepum:
This is a better direction but it breaks compatibility without any sort of real deprecation cycle. For a few releases @lexical/code is going to have to work as-is. The problem you want solved is not fixable in a single pass or in a single release, but you should be able to work around it by configuring your bundler to not include the prism stuff in the meantime in situations where you aren’t going to use it.

gurkerl83:
Makes sense, and thanks for the guidance.
I agree this is not something to flip in one pass without a deprecation path. Even if the change is technically small, it changes expected default behavior for existing users.

A major release could be a better place for this kind of split. I also tried bundler-level mitigation in our app, but it was brittle in practice and didn’t reliably keep the Prism/code path out in all cases. 

For now, the compatibility requirement for @lexical/code as-is is clear, so I’ll leave this PR as-is. It may still be the wrong shape (especially around adding a compose layer), but it was intended as one exploration of stronger separation.

etrepum:
Moving prism code to the new package makes sense, that part is directionally correct, but there should be deprecated re-exports from the main code package so that people can upgrade without changing the usage in lockstep.

In the first pass, no behavior should change other than the addition of deprecation warnings and new packages/exports for the upgrade path. The playground and other sibling packages can be upgraded to use the new path to avoid deprecation warnings. No types or exports in @lexical/code should change. No behavior change to markdown should happen at all. No tests should need to change (other than imports, but only to avoid deprecation warnings) and all existing tests should pass.

PR #8200, opened 2026-03-09
[lexical-table] Refactor: Call $handleTableSelectionChangeCommand once instead of per-table

randal-atticus:
## Description
Further refactors the Table's SELECTION_CHANGE_HANDLER ($handleTableSelectionChangeCommand) so that it only needs to run once.

Cheating a bit at the end because there's still some logic that runs over each table, but I think it's preferable to refactor this in pieces.

Key changes:
- nextFocus, shouldCheckSelection are hoisted up into an extension-wide 'TableObservers' class (which now also contains the old map of Table NodeKey -> TableObserver). Think of it as a home for state across all tables.
- nextFocus/shouldCheckSelection therefore only need to be handled once, instead of per-table.
- Split the remaining handler logic into 3 parts.
  - $fixTableSelectionForSelectedTable (this also now only needs to run once)
  - $fixRangeSelectionForTable
  - $syncTableSelectionState 

Note: No existing bugs fixed in this PR but these refactors are a precursor to some important nested table bug fixes.

### Before

### After

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -4712,6 +4712,8 @@ test.describe.parallel('Tables', () => {
       page.setViewportSize({height: 1000, width: 3000});
     }
 
+    const pageOrFrame = getPageOrFrame(page);
+
     await focusEditor(page);
 
     await insertTable(page, 3, 3);
@@ -4784,7 +4786,10 @@ test.describe.parallel('Tables', () => {
       const clipboard = await copyToClipboard(page);
 
       // move caret to the first position of the editor
-      await page.locator('.PlaygroundEditorTheme__paragraph').first().click();
+      await pageOrFrame
+        .locator('.PlaygroundEditorTheme__paragraph')
+        .first()
+        .click();
 
       // move caret to the table cell (2,2)
       await page.keyboard.press('ArrowDown');


packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -939,10 +998,10 @@ export function $handleTableSelectionChangeCommand(
         );
       }
       $setSelection(newSelection);
-      $addHighlightStyleToTable(editor, tableObserver);
```
randal-atticus, line 942: Realised that with the `$syncTableSelectionState` pass happening after all of this, there was no need to explicitly add the styling.

etrepum:
Looks like there's a firefox issue here

packages/lexical-playground/src/plugins/TableActionMenuPlugin/index.tsx
```diff
@@ -159,8 +159,11 @@ function TableActionMenu({
       if ($isTableSelection(selection)) {
         const currentSelectionCounts = computeSelectionCount(selection);
         updateSelectionCounts(computeSelectionCount(selection));
+        const isCollapsedTableSelection = selection.anchor.is(selection.focus);
```
randal-atticus, line 162: If you merge a cell  and then enter a **single-cell table selection**, it would consider the cell mergeable. This just happened to break one of the tests in Firefox because it's easier to set up a single-cell table selection (click, then shift-click a cell - this only works in Firefox).

In the code before, a mergeable cell is not considered un-mergeable so one of the tests was failing in Firefox.

Now:

Image

randal-atticus:
Assuming tests pass on the latest commit, one bug was a silly typo in getAndClearShouldCheckSelectionForTable, another was a selection quirk when unmerging cells, last one was what seemed to be a bug with detecting mergeable/unmergeable table selections in the playground.

etrepum (approved):

PR #8202, opened 2026-03-09
[lexical-extension][lexical-react] Feature: LexicalExtensionEditorComposer, NestedEditorExtension, and fixed SharedHistoryExtension with playground refactor

etrepum:
## Breaking Changes

Unhandled dispatched commands are now only bubbled to parent editors if the parent editor is not currently in an update. The previous behavior was broken as it would trigger a commit and GC in the middle of an in-progress update which can corrupt the state.

## Description

Refactor more of the playground to use extensions, and address some needs it has with nested editors:

`@lexical/react/LexicalExtensionEditorComposer`:

- A new `LexicalExtensionEditorComposer` component to render an editor (nested or otherwise) that was already built with extensions

`@lexical/extension`

- New `NestedEditorExtension` to set the `parentEditor` in the EditorConfig when the nested editor is constructed
- `SharedHistoryExtension` should now work properly (before there was a bootstrapping issue due to the init phase of HistoryExtension)
- `HorizontalRuleExtension` now implements the `INSERT_HORIZONTAL_RULE_COMMAND`
- `LinkExtension` now registers `TOGGLE_LINK_COMMAND` at the lowest editor priority
- `CheckListExtension` now has an output signal for its `disableTakeFocusOnClick` configuration
-  New `SelectionAlwaysOnDisplayExtension` to wrap `selectionAlwaysOnDisplay`

Closes #8203

## Test plan

All existing e2e and unit tests pass (unit tests depending playground plugins were also refactored for extensions)

- [x] Added unit test coverage for SharedHistoryExtension, NestedEditorExtension
- [x] Added unit test coverage for LexicalExtensionEditorComposer

### DIFF
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -66,6 +66,7 @@ import SpeechToTextPlugin from './plugins/SpeechToTextPlugin';
 import TabFocusPlugin from './plugins/TabFocusPlugin';
 import TableCellActionMenuPlugin from './plugins/TableActionMenuPlugin';
 import TableCellResizer from './plugins/TableCellResizer';
+import TableFitNestedTablePlugin from './plugins/TableFitNestedTablePlugin';
 import TableHoverActionsV2Plugin from './plugins/TableHoverActionsV2Plugin';
 import TableOfContentsPlugin from './plugins/TableOfContentsPlugin';
 import TableScrollShadowPlugin from './plugins/TableScrollShadowPlugin';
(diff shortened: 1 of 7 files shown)


packages/lexical-playground/src/Editor.tsx
```diff
@@ -91,8 +82,25 @@ const skipCollaborationInit =
...
+  V,
+  Output extends {[Key in K]: Signal<V>},
+>(extension: OutputExtension<Output>, prop: K, value: V) {
+  const signal = useOptionalExtensionDependency(extension)?.output[prop];
+  useEffect(() => {
+    if (signal) {
+      signal.value = value;
+    }
+  }, [signal, value]);
+}
```
etrepum, line 97: Will probably switch this over to something higher-level in a follow-up PR to propagate settings to output signals in batch but this was straightforward when migrating one plugin to extensions at a time

PR #8203, opened 2026-03-10
[lexical-react][lexical-playground] Fix: correct imports of horizontal rule

dabrpio:
## Description

After deprecation of `HorizontalRuleNode` and `$createHorizontalRuleNode` in `@lexical/react/LexicalHorizontalRuleNode` (#7706) I changed the imports to `@lexical/extension` in my project. It seems it works well in build but today I noticed that on local it does not work. I was getting this error:

obraz

I realized that `HorizontalRulePlugin` from `@lexical/react/LexicalHorizontalRulePlugin` still uses `$createHorizontalRuleNode` deprecated import. This PR updates this and other deprecated imports from  `@lexical/react/LexicalHorizontalRulePlugin`.

Additionally, this PR also updates other imports from this file (`INSERT_HORIZONTAL_RULE_COMMAND`, `$isHorizontalRuleNode`) so that now they are importing from  `@lexical/extension`.

### Suggestion

Apart from deprecated `HorizontalRuleNode` and `$createHorizontalRuleNode`, there are still 3 other things - `INSERT_HORIZONTAL_RULE_COMMAND`, `$isHorizontalRuleNode` and `SerializedHorizontalRuleNode` which are exported from `@lexical/react/LexicalHorizontalRulePlugin`. I suggest to also mark them as deprecated and force everyone to import them from `@lexical/extension`.

### DIFF
--- a/packages/lexical-playground/src/plugins/ComponentPickerPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ComponentPickerPlugin/index.tsx
@@ -9,6 +9,7 @@
 import type {JSX} from 'react';
 
 import {$createCodeNode} from '@lexical/code';
+import {INSERT_HORIZONTAL_RULE_COMMAND} from '@lexical/extension';
 import {
   INSERT_CHECK_LIST_COMMAND,
   INSERT_ORDERED_LIST_COMMAND,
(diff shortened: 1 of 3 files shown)


etrepum:
This is more or less on purpose, the HorizontalRuleNode from extensions is a different non-react implementation from the one in `@lexical/react`. The extension and the react plugin are mutually exclusive. The next step is to migrate the playground to use the extension instead of the plugin.

The command is the same either way, it’s just a re-export. The node class and the plugin vs. extension to support it are the difference.

The WIP PR #8202 addresses most of this already

dabrpio:
Ok, I'm closing this PR since it's now addressed in #8202.

PR #8206, opened 2026-03-10
[lexical-utils][lexical-playground] Feature: Add $insertNodeIntoLeaf and insert deeply DateTimeNode

levensta:
## Description

Added `$insertNodeIntoLeaf` function to insert a node into the outermost leaf node. An example of how to use this function can be seen when inserting DateTimeNode into lexical-playground

### Before

Inserting DateTime into a link split it into two nodes, and DateTime does not become part of the link. Also, inserting DateTime does not take into account the text format. To select DateTime in a link and apply a format to it, you have to select it again

https://github.com/user-attachments/assets/83a0f334-89cb-4d66-aceb-0f968005a54f

### After

DateTime is inserted into the link and becomes part of it. If you enter formatted text, the subsequent insertion of DateTime will preserve the preceding format

https://github.com/user-attachments/assets/382d4e88-8ff5-4fe8-bda9-1db43b200c9b

### DIFF
--- a/packages/lexical-playground/src/plugins/DateTimePlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/DateTimePlugin/index.tsx
@@ -52,11 +52,11 @@ export default function DateTimePlugin(): JSX.Element | null {
           const {dateTime} = payload;
           const dateTimeNode = $createDateTimeNode(dateTime);
 
-          $insertNodeIntoLeaf(dateTimeNode);
           const selection = $getSelection();
           if ($isRangeSelection(selection)) {
             dateTimeNode.setFormat(selection.format);
           }
+          $insertNodeIntoLeaf(dateTimeNode);
           if ($isRootOrShadowRoot(dateTimeNode.getParentOrThrow())) {
             $wrapNodeInElement(dateTimeNode, $createParagraphNode).selectEnd();
           }


packages/lexical-playground/src/plugins/DateTimePlugin/index.tsx
```diff
@@ -47,7 +51,11 @@ export default function DateTimePlugin(): JSX.Element | null {
           const {dateTime} = payload;
           const dateTimeNode = $createDateTimeNode(dateTime);
```
etrepum, line 53: I think it would be simpler and probably more accurate to pick up the selection's format rather than only a bordering TextNode on insert. It's possible to insert a DateTimeNode into an empty ParagraphNode after changing the desired format but before entering any other text.

packages/lexical-playground/src/plugins/DateTimePlugin/index.tsx
```diff
@@ -47,7 +52,11 @@ export default function DateTimePlugin(): JSX.Element | null {
...
 
-          $insertNodes([dateTimeNode]);
+          $insertNodeIntoLeaf(dateTimeNode);
+          const selection = $getSelection();
+          if ($isRangeSelection(selection)) {
+            dateTimeNode.setFormat(selection.format);
+          }
```
etrepum: Makes more sense to do this before changing the selection

```suggestion
          const selection = $getSelection();
          if ($isRangeSelection(selection)) {
            dateTimeNode.setFormat(selection.format);
          }
          $insertNodeIntoLeaf(dateTimeNode);
```

etrepum (approved):

PR #8211, opened 2026-03-11
[lexical-markdown][lexical-playground] Bug Fix: Convert tabs in TabNode at import

lytion:
This is a follow up of #7845. I made a new PR to make it clean.

Reported issue: https://github.com/facebook/lexical/issues/7820
To summarize, if you use indent button and then switch to markdown, indent will be lost.

In this fix, once all transformers have been applied, loop on nodes to convert all '\t' in TabNodes.
I decided to do it here and not in `$importBlocks` as it was gonna create several nodes and break the function flow which only use one TextNode.

Closes https://github.com/facebook/lexical/issues/7820

### DIFF
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -303,13 +303,14 @@ function $importBlocks(
 
 // Look in node for '\t' and create a TabNode for each occurrence.
 function $normalizeMarkdownTextNode(textNode: TextNode): void {
-  const tabOffsets = [];
+  const tabOffsets: Set<number> = new Set();
   const text = textNode.getTextContent();
   let index = text.indexOf('\t');
 
   // Find all tab occurrences
   while (index !== -1) {
-    tabOffsets.push(index, index + 1);
+    tabOffsets.add(index);
+    tabOffsets.add(index + 1);
     index = text.indexOf('\t', index + 1);
   }
 
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -791,6 +791,11 @@ describe('Markdown', () => {
       md: 'foo\t \tbar\n\tbaz',
       skipExport: true,
     },
+    {
+      html: '<p><span style="white-space: pre-wrap;">Hello</span><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;">World</span></p>',
+      md: 'Hello\t\tWorld',
+      skipExport: true,
+    },
   ];
 
   for (const {


packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -84,17 +86,24 @@ export function createMarkdownImport(
...
       ) {
         child.remove();
+        continue;
+      }
+      // Convert all '\t' into TabNode.
+      if ($isParagraphNode(child) && child.getTextContent().includes('\t')) {
+        for (const paragraphChild of child.getChildren()) {
+          $createTabInNode(paragraphChild);
+        }
       }
```
etrepum: This would probably be better written by traversing for all of the TextNodes first, computing paragraphNode.getTextContent() is doing a whole extra traversal of that subtree

```ts
      if ($isElementNode(child)) {
        for (const textNode of child.getAllTextNodes()) {
          $normalizeMarkdownTextNode(textNode);
        }
      }
```

For the simplest correct implementation that would preserve formats and such, the tab replacing code should find the appropriate offsets, use `textNode.splitText()`, and then replace the nodes that are just tabs with `$createTabNode()`.

packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -301,21 +302,22 @@ function $importBlocks(
...
+  // Repeat the operation until no more '\t' is found.
+  while (tabOffset !== -1) {
+    const splitNodes = currNode.splitText(tabOffset, tabOffset + 1);
+    splitNodes.forEach((node) => {
+      if (node.getTextContent() === '\t') {
+        node.replace($createTabNode());
+      }
+    });
+    currNode = splitNodes[splitNodes.length - 1];
+    tabOffset = currNode.getTextContent().indexOf('\t');
```
etrepum: I don't think this logic is correct if the node ends with a tab? It looks like it will loop infinitely. It could also be done a little more clearly if all the tabs are found first and splitText is called at most once

lytion: Could have yes, but it seems there's a trim before because I have no '\t' or space at the end of a string when calling `getTextContent()`.

packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -291,6 +301,27 @@ function $importBlocks(
   }
 }
 
+// Look in node for '\t' and create a TabNode for each occurrence.
+function $normalizeMarkdownTextNode(textNode: TextNode): void {
+  const tabOffsets = [];
```
etrepum:
```suggestion
  const tabOffsets: Set<number> = new Set();
```

packages/lexical-markdown/src/MarkdownImport.ts
```diff
@@ -291,6 +301,27 @@ function $importBlocks(
...
+  const text = textNode.getTextContent();
+  let index = text.indexOf('\t');
+
+  // Find all tab occurrences
+  while (index !== -1) {
+    tabOffsets.push(index, index + 1);
```
etrepum:
```suggestion
    tabOffsets.add(index);
    tabOffsets.add(index + 1);
```

lytion: Indeed, I thought I covered this case in unit test but this test had '\t(space)\t', my bad.
I refactored to use a Set and added a test

PR #8213, opened 2026-03-12
[lexical-list] Fix: create copies ListNode/ListItemNode in split-like operations

levensta:
## Description

`ListNode` and `ListItemNode` internally create new list structures during operations such as **splitting lists, indenting/outdenting**. Previously, these operations relied on factory functions (`$createListNode` / `$createListItemNode`).

This caused issues when users extended ListNode or ListItemNode. Even if an editor contained custom subclasses, split-like operations would always create base ListNode / ListItemNode instances, breaking type consistency.

This change replaces factory-based creation with `$copyNode`, ensuring that new nodes are created from the same class as the source node.

https://github.com/user-attachments/assets/834a897b-60c3-4e48-a778-c8953b24a018

https://codesandbox.io/p/sandbox/lexical-extended-list-rgrcq9

### Before

Split-like operations (list split, indent, outdent) always created base instances ListNode and ListItemNode. Even if the original list used subclasses.

### After

Instances of the same class as the source node are now created using `$copyNode` in the following cases:
- Splitting a list by pressing Enter inside an empty element
- Handling indentation/outdentation
- Using `splice` method
- Using the replace method on a ListItem between other elements

This allows you to save custom `ListNode` and `ListItemNode` subclasses.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownImport.ts
+++ b/packages/lexical-markdown/src/MarkdownImport.ts
@@ -20,12 +20,15 @@ import {$findMatchingParent} from '@lexical/utils';
 import {
   $createLineBreakNode,
   $createParagraphNode,
+  $createTabNode,
   $createTextNode,
   $getRoot,
   $getSelection,
+  $isElementNode,
   $isParagraphNode,
   $isTextNode,
   ElementNode,
+  TextNode,
 } from 'lexical';
 
 import {importTextTransformers} from './importTextTransformers';
@@ -84,17 +87,24 @@ export function createMarkdownImport(
       );
     }
 
-    // By default, removing empty paragraphs as md does not really
-    // allow empty lines and uses them as delimiter.
-    // If you need empty lines set shouldPreserveNewLines = true.
     const children = root.getChildren();
     for (const child of children) {
+      // By default, removing empty paragraphs as md does not really
+      // allow empty lines and uses them as delimiter.
+      // If you need empty lines set shouldPreserveNewLines = true.
       if (
         !shouldPreserveNewLines &&
         isEmptyParagraph(child) &&
         root.getChildrenSize() > 1
       ) {
         child.remove();
+        continue;
+      }
+      // Convert all '\t' into TabNode.
+      if ($isElementNode(child)) {
+        for (const textNode of child.getAllTextNodes()) {
+          $normalizeMarkdownTextNode(textNode);
+        }
       }
     }
 
@@ -289,6 +299,28 @@ function $importBlocks(
       }
     }
   }
+}
+
+// Look in node for '\t' and create a TabNode for each occurrence.
+function $normalizeMarkdownTextNode(textNode: TextNode): void {
+  const tabOffsets: Set<number> = new Set();
+  const text = textNode.getTextContent();
+  let index = text.indexOf('\t');
+
+  // Find all tab occurrences
+  while (index !== -1) {
+    tabOffsets.add(index);
+    tabOffsets.add(index + 1);
+    index = text.indexOf('\t', index + 1);
+  }
+
+  // Split node to isolate each tab then replace '\t' into TabNode
+  const splitNodes = textNode.splitText(...tabOffsets);
+  splitNodes.forEach((node) => {
+    if (node.getTextContent() === '\t') {
+      node.replace($createTabNode());
+    }
+  });
 }
 
 function createTextFormatTransformersIndex(
--- a/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
+++ b/packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
@@ -781,6 +781,21 @@ describe('Markdown', () => {
       md: 'foo\\\nbar',
       mdAfterExport: 'foo\nbar',
     },
+    {
+      html: '<p><span style="white-space: pre-wrap;">foo </span><a href="https://lexical.dev"><span style="white-space: pre-wrap;">bar</span></a><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;">baz</span></p>',
+      md: 'foo [bar](https://lexical.dev)\tbaz',
+      skipExport: true,
+    },
+    {
+      html: '<p><span style="white-space: pre-wrap;">foo</span><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;"> </span><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;">bar</span><br><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;">baz</span></p>',
+      md: 'foo\t \tbar\n\tbaz',
+      skipExport: true,
+    },
+    {
+      html: '<p><span style="white-space: pre-wrap;">Hello</span><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;">\t</span><span style="white-space: pre-wrap;">World</span></p>',
+      md: 'Hello\t\tWorld',
+      skipExport: true,
+    },
   ];
 
   for (const {
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -75,6 +75,7 @@ import SpeechToTextPlugin from './plugins/SpeechToTextPlugin';
 import TabFocusPlugin from './plugins/TabFocusPlugin';
 import TableCellActionMenuPlugin from './plugins/TableActionMenuPlugin';
 import TableCellResizer from './plugins/TableCellResizer';
+import TableFitNestedTablePlugin from './plugins/TableFitNestedTablePlugin';
 import TableHoverActionsV2Plugin from './plugins/TableHoverActionsV2Plugin';
 import TableOfContentsPlugin from './plugins/TableOfContentsPlugin';
 import TableScrollShadowPlugin from './plugins/TableScrollShadowPlugin';
@@ -246,9 +247,9 @@ export default function Editor(): JSX.Element {
               hasCellMerge={tableCellMerge}
               hasCellBackgroundColor={tableCellBackgroundColor}
               hasHorizontalScroll={tableHorizontalScroll && !hasFitNestedTables}
-              hasFitNestedTables={hasFitNestedTables}
               hasNestedTables={hasNestedTables}
             />
+            {hasFitNestedTables ? <TableFitNestedTablePlugin /> : null}
             <TableCellResizer />
             <TableScrollShadowPlugin />
             <ImagesPlugin />
--- a/packages/lexical-playground/src/plugins/TableFitNestedTablePlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/TableFitNestedTablePlugin/index.tsx
@@ -1,0 +1,148 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import type {LexicalNode} from 'lexical';
+
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {$isTableNode, TableNode} from '@lexical/table';
+import {$dfs, $findMatchingParent} from '@lexical/utils';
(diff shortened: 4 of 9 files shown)


etrepum:
Using $copyNode is probably better than accessing the constructor directly

etrepum:
Should also have unit tests that demonstrate that the change allows for the kind of functionality you’re trying to add

levensta:
> Using $copyNode is probably better than accessing the constructor directly

Initially, I tried using `$copyNode`, but it makes a complete copy of all class properties, including node state. At the very least, this caused the saving of markdown list markers to break https://github.com/facebook/lexical/blob/main/packages/lexical-markdown/src/MarkdownTransformers.ts#L223-L225

etrepum:
You'll have to be more specific about "saving of the list markers to break". Not quite sure what that means, it should just re-use the same style of list marker when this copy is made? If the markdown export doesn't check for mismatched list markers then that should be fixed.

There are some potential pitfalls with a full `$copyNode` (but if there are any you'd generally have the same problems when copy & paste is used), but a completely shallow copy has other potential pitfalls (missing state like format and such). Probably what should happen is that this functionality should be put in an instance method similar to `afterCloneFrom` or TextNode's `splitText` to isolate this copy-like functionality and have a place where the instance can be mutated accordingly after construction and if that behavior needs to be overridden in a subclass then super can be called to prevent code duplication.

levensta:
I am referring to a situation where a list is created using markdown, then a new item is created and indented. When indenting, a copy of ListItemNode is created, which also copies the listMarkerState, even though it is already part of another nested list

https://github.com/user-attachments/assets/3e0c3109-763d-4089-ba29-741e7181f984

image

To create a copy using `$copyNode`, you need to reset all properties to their default state, but there is no way to reset the state using a method, except in the following way:
```js
$copyNode(listItem).getWritable().__state = undefined;
````

That's why I thought it would be easier to create a copy using the constructor, since it doesn't require a verbose reset of all properties and states.

**Comment by etrepum** [comment]
That doesn’t seem broken to use the same marker for an indented list? Either way you’re either verbosely copying state from one node to the next or verbosely resetting it.

With either approach, using a method would allow this to be isolated and overridden if necessary. I don’t think adding direct constructor calls all over the place is a good idea, a few strategic places that are easy to find and maintain is ok.

**Comment by levensta** [comment]
> Either way you’re either verbosely copying state from one node to the next or verbosely resetting it.

I agree. If you need to avoid using a constructor, I'd suggest creating a helper function that hides the boilerplate code for resetting properties and state

```js
function $spawnListItemNode(node) {
  const copy = $copyNode(node).setValue(1).setChecked(undefined);
  copy.getWritable().__state = undefined;
  return copy;
}
```

etrepum:
I do not recommend resetting NodeState in this way.

levensta:
> I do not recommend resetting NodeState in this way.

But are there other ways? $setState assumes an explicit prodvide of a known state

I looked at the internal available state API and maybe this approach is acceptable?

```js
const state = $getWritableNodeState(node);
state.knownState.forEach((_, stateConfig) => {
   state.updateFromKnown(stateConfig, null);
});
```

I would also suggest using the `$create` function, but it also requires passing a constructor

etrepum:
You’re misunderstanding me here, using the constructor is ok, but don’t spread it all over the code base. Put it in a *method* so it is easy to find for maintenance and can be overridden.

As for NodeState it shouldn’t be cleared at all, if specific properties should be omitted in this kind of copy-as-template then that’s functionality we’d want to add to the NodeState configuration and support it with some kind of copyNode-like functionality.

packages/lexical/src/nodes/LexicalElementNode.ts
```diff
@@ -343,6 +343,15 @@ export class ElementNode extends LexicalNode {
...
   }
 
+  /**
+   * Creates a new instance of the same class as the current object.
+   * Useful in class hierarchies so that method can return an object of the same specific type.
+   * @returns A new instance of the current class.
+   */
+  create<T extends this>(): T {
+    return new this.constructor() as T;
+  }
```
etrepum: For the use case it would make more sense to be similar to `insertNewAfter` where it assigns properties from the original node as a template. Maybe something like `createSibling`. I don't think we need to have it on ElementNode right now as the use case is specifically ListNode and ListItemNode

For example this is ListItemNode.insertNewAfter

```ts
  insertNewAfter(
    _: RangeSelection,
    restoreSelection = true,
  ): ListItemNode | ParagraphNode {
    const newElement = $createListItemNode()
      .updateFromJSON(this.exportJSON())
      .setChecked(this.getChecked() ? false : undefined);

    this.insertAfter(newElement, restoreSelection);

    return newElement;
  }
```

You can see that it sets up the new element to be similar to the current element (although in this case it will also copy NodeState via the JSON).

Maybe what you really want is to just add something to NodeState such that we can declare that certain state should not be copied when using a node as a template.

levensta: In fact, I even considered moving this method to the LexicalNode base class, since split operations are also used for text nodes:
- https://github.com/facebook/lexical/blob/main/packages/lexical/src/nodes/LexicalTextNode.ts#L1026
- https://github.com/facebook/lexical/blob/main/packages/lexical-code-core/src/CodeNode.ts#L343
- https://github.com/facebook/lexical/blob/main/packages/lexical-mark/src/index.ts#L114-L115

I can imagine a custom user editor having both the base text and its extended class.

levensta: > Maybe what you really want is to just add something to NodeState such that we can declare that certain state should not be copied when using a node as a template

At first glance, it seems like we could add a writable boolean flag to pass to $copyNode, set it in NodeState

```ts
function $copyNode<T extends LexicalNode>(node: T): T {
  const copy = node.constructor.clone(node) as T;
  $setNodeKey(copy, null);
  node.getWritable().__state.mustCopied = true;
  copy.afterCloneFrom(node);
  return copy;
}
```

and then check before copying in the `LexicalNode.afterCloneFrom` method https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalNode.ts#L551

But maybe then it’s worth just creating a `$spawnNode` function that will hide the constructor call within itself?

```ts
function $spawnNode<T extends LexicalNode>(node: T): T {
  return new node.constructor() as T;
}
```

packages/lexical-list/src/LexicalListItemNode.ts
levensta, line 309: I'm a little confused, so let's take it step by step. 

We now have a new method called `resetOnCopyNodeFrom`. By default, it is called after a copy is created within `$copyNode`. This allows us to reset the node's properties to their default values.

In other words, to create a copy of a node without copying its properties, you need to reset those properties in the `resetOnCopyNodeFrom` method which is exactly what we need given the current changes to list processing. But if we reset the values in `resetOnCopyNodeFrom`, it will be a breaking change for `$copyNode`

From all this, I get the impression that it would be better if `$copyNode` did not call `resetOnCopyNodeFrom` by default, but did so only when the option is explicitly passed. This way, `$copyNode` would make a copy with all properties by default, as described in the documentation, and would also have the option to explicitly call reset. But for such an API, we’d need to add `resetOnCopyNodeFrom` for all core nodes. What do you think about this?

etrepum: The goal isn’t to restore all values to defaults, generally when you are copying a node you want another node that’s substantially similar to the original node.  Same formatting, list type, etc. Copying uses the original node as a template. It is not designed to create a node with only defaults.

Certain properties are more unique to a specific node, like whether it is checked/collapsed/etc. or not. Those properties should be reset to defaults when making a copy.

packages/lexical/src/nodes/LexicalElementNode.ts
etrepum: Generally speaking you can't just call `new node.constructor()`, up until the `$config` addition I don't think there was even any documentation encouraging people not to implement constructors with required arguments.

If you look at the TextNode.splitText case it's doing primarily what `$copyNode` does - constructing a new node, then setting all of the properties to be the same (except for `__text` and `__mode`). I think it's generally expected behavior for splitting text to change the class though, particularly in the segmented case. It might make sense to have a lower-level method that splitText calls that allows that behavior to be specified though.

packages/lexical-list/src/__tests__/unit/formatList.test.ts
```diff
@@ -34,9 +35,50 @@ import {
...
+  $createListItemNode,
+  $isListItemNode,
+  ListItemNode,
+} from '../../LexicalListItemNode';
+import {
+  $createListNode,
+  $isListNode,
+  ListNode,
+  ListType,
+} from '../../LexicalListNode';
```
etrepum: I think all of these are public APIs and could just be imported from `@lexical/list`

packages/lexical-list/src/formatList.ts
```diff
@@ -415,10 +416,10 @@ export function $handleIndent(listItemNode: ListItemNode): void {
...
 
     if ($isListNode(parent)) {
-      const newListItem = $createListItemNode()
+      const newListItem = $copyNode(listItemNode)
         .setTextFormat(listItemNode.getTextFormat())
         .setTextStyle(listItemNode.getTextStyle());
-      const newList = $createListNode(parent.getListType())
+      const newList = $copyNode(parent)
         .setTextFormat(parent.getTextFormat())
         .setTextStyle(parent.getTextStyle());
```
etrepum: Setting the format and style should be redundant with `$copyNode`


```suggestion
      const newListItem = $copyNode(listItemNode);
      const newList = $copyNode(parent);
```

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -197,14 +198,21 @@ export class ListNode extends ElementNode {
...
+
     let listItemNodesToInsert = nodesToInsert;
     for (let i = 0; i < nodesToInsert.length; i++) {
       const node = nodesToInsert[i];
       if (!$isListItemNode(node)) {
         if (listItemNodesToInsert === nodesToInsert) {
           listItemNodesToInsert = [...nodesToInsert];
         }
-        listItemNodesToInsert[i] = $createListItemNode().append(
+        listItemNodesToInsert[i] = newListItem.append(
```
etrepum: This change is wrong, newListItem needs to be a function not a single item. It would probably make sense to also add a unit test case that would cover this situation where an example is used (or not) to create at least two ListItemNodes.

```suggestion
        listItemNodesToInsert[i] = $newListItem().append(
```

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -197,14 +198,21 @@ export class ListNode extends ElementNode {
     deleteCount: number,
     nodesToInsert: LexicalNode[],
   ): this {
+    const exampleListItem =
+      nodesToInsert.find($isListItemNode) ??
+      this.getChildren().find($isListItemNode);
+    const newListItem = exampleListItem
+      ? $copyNode(exampleListItem)
+      : $createListItemNode();
```
etrepum:
```suggestion
    const $newListItem = exampleListItem
      ? $copyNode.bind(null, exampleListItem)
      : $createListItemNode;
```

levensta: The `bind` method cannot bind the argument's type to a generic, so I created a callback wrapper

packages/lexical-list/src/LexicalListNode.ts
levensta: I added a new test for `splice`.

I also realized there’s a corner case where a list subclass might be empty, and if `splice` is used with non-`ListItem` nodes, the fallback callback `$createListItemNode` will be triggered. It might be a good idea to find a way to link ListNode to ListItem by adding something like a `createChildElement()` method, similar to how ListItem is linked to ListNode via the `createParentElement` method. But perhaps it would be better to handle this by overriding the `splice` method in a subclass

etrepum: I think that’s something for someone with that use case to worry about and they can propose a method for it (or override splice, or handle it with a transform for their ListNode subclass). Usually best not to make things too abstract until it’s actually needed for something.

etrepum (approved):

etrepum:
Let's update the PR description to be a little more explicit about what this change is and how it affects various list operations before merge, this will help with the release notes

levensta:
> Let's update the PR description to be a little more explicit about what this change is and how it affects various list operations before merge, this will help with the release notes

done

PR #8214, opened 2026-03-12
[lexical-playground] Fix: use inline style for LayoutContainerNode import

### LINKED ISSUE #6813: The $convertLayoutContainerElement function of LayoutContainerNode always returns null

WhyBusyy:
## Summary
- `$convertLayoutContainerElement` used `window.getComputedStyle()` which returns empty values for unmounted DOM nodes during HTML import
- Since `exportDOM()` stores `gridTemplateColumns` as an inline style, read it directly via `domNode.style.gridTemplateColumns`

Fixes #6813

## Test plan
- [ ] Create a 2-column layout in the playground editor
- [ ] Export as HTML (`$generateHtmlFromNodes`)
- [ ] Re-import the exported HTML (`$generateNodesFromDOM`)
- [ ] Verify the layout container node is correctly reconstructed with the original column template

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/lexical-playground/__tests__/unit/LayoutContainerNode.test.ts
+++ b/packages/lexical-playground/__tests__/unit/LayoutContainerNode.test.ts
@@ -1,0 +1,176 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {$generateHtmlFromNodes, $generateNodesFromDOM} from '@lexical/html';
+import {$getRoot, $insertNodes} from 'lexical';
+import {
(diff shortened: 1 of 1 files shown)


etrepum:
I’ll review after the CLA is signed and this PR has tests that demonstrate that the change works as expected

WhyBusyy:
@etrepum Already signed! Thanks for the quick review!

PR #8217, opened 2026-03-12
[lexical] Fix: Consecutive Linebreak insertion resets selection format

### LINKED ISSUE #8160: Bug: linebreak insertion breaks format inheritance

Jynx2004:
## Description

Instead of sending value 0 as a static value we take the current format from selection.format and pass it to the respective function responsible for format change.
As a result the format gets passed on between multiple linebreaks.

Closes #8160 

### Before

Screenshot 2026-02-26 004335

### After

Screenshot 2026-02-26 004320

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -962,7 +962,7 @@ importers:
         version: 2.1.1
       docusaurus-plugin-internaldocs-fb:
         specifier: 1.19.2
-        version: 1.19.2(62eb37513ee458c9f00fe497c6bcc9c4)
+        version: 1.19.2(yx6vk5fx7a57ucxth6gqmura4a)
       docusaurus-plugin-typedoc:
         specifier: ^1.4.2
         version: 1.4.2(typedoc-plugin-markdown@4.9.0(typedoc@0.28.13(typescript@5.9.2)))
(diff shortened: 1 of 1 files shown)


Jynx2004:
This is a new Pull Request which solves the same issue as the first pr that i created but in this i have modified the test script for testing the function.

packages/lexical/src/__tests__/unit/LexicalSelection.test.ts
```diff
@@ -1367,6 +1367,37 @@ describe('getNodes()', () => {
...
+          $getRoot().append(paragraph);
+
+          const selection = textNode.select(2, 2);
+
+          selection.insertLineBreak();
+          selection.insertLineBreak();
+
+          const newSelection = $assertRangeSelection($getSelection());
+
+          newSelection.formatText('bold');
```
etrepum:
```suggestion
```

The purpose of this fix is to eliminate the need to reset the selection's format state, so it doesn't make sense for the test to manually reset the format after inserting the linebreaks

Jynx2004: The test calls the editor API ([selection.insertLineBreak()](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [insertText](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) directly, with no DOM/browser input event. The [collapsedSelectionFormat](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) logic relies on a recent browser event timestamp, so it isn't populated here. As a result the selection.format isn't updated to bold and the inserted character ends up with no format.

etrepum: So that means two things:
1. The testing strategy here isn't sufficient. If you need it to rely on the browser then an e2e test is likely more appropriate. The unit test as-is serves no purpose. It does not do the one thing it's supposed to do, which is to show that this fix works.
2. There's may be a more robust general fix to this problem that would preserve the selection format before entering the collapsedSelectionFormat logic

Jynx2004: Will try to modify the test

PR #8221, opened 2026-03-13
[lexical][lexical-list][lexical-markdown] Feature: resetOnCopyNode configuration to NodeState and LexicalNode.resetOnCopyNodeFrom hook

etrepum:
## Description

It's often useful to use `$copyNode` to create a similar node to the current one, e.g. in implementations of `insertNewAfter`. In these cases there may be state that should get reset when copied, such as the checkmark state of a `ListItemNode`, or markdown parsing annotations.

LexicalNode
* `resetOnCopyFrom` - a method similar to `afterCloneFrom` which can be used to reset a node's state

NodeState `StateValueConfig`
* `resetOnCopy` - boolean configuration

List:
* `ListItemNode` has been changed to use `$copyNode` in its implementation of `insertNewAfter` instead of a round-trip to JSON. The logic to uncheck the ListItemNode has been moved to its `resetOnCopyFrom`

Markdown:
* `listMarkerState` and `codeFenceState` have been changed to use `resetOnCopyNode: true`

As an example, here's how the markdown parsing annotations have been updated to use this new configuration:

```ts
export const listMarkerState = createState('mdListMarker', {
  parse: (v) => (typeof v === 'string' && /^[-*+]$/.test(v) ? v : '-'),
  resetOnCopy: true,
});
```

## Test plan

New unit tests, all existing unit and e2e pass unchanged

### DIFF
--- a/packages/lexical-website/docs/concepts/node-cloning.md
+++ b/packages/lexical-website/docs/concepts/node-cloning.md
@@ -60,6 +60,9 @@ function $duplicateNode(node: MyCustomNode) {
 ### What is `$copyNode`?
 
 `$copyNode` is the public API for creating a copy of a node with a new key. Use this when you need to create a duplicate node.
+By default, all properties and `NodeState` will be copied to the new node, and then `resetOnCopyNodeFrom` will be called to
+allow the node to optionally reset certain properties (and NodeState configured with `resetOnCopyNode: true`) to defaults
+(such as the checked state of a `ListItemNode`).
 
 ```typescript
 // ✅ Correct: Using $copyNode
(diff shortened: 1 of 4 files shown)


zurfyx (approved):

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -300,13 +301,18 @@ export class ListItemNode extends ElementNode {
     }
   }
 
+  resetOnCopyNodeFrom(original: this): void {
+    super.resetOnCopyNodeFrom(original);
+    if (original.getChecked()) {
+      this.setChecked(false);
+    }
+  }
```
levensta, line 309: If the `checked` property is set on a `ListItemNode` and the user attempts to copy the node using `$copyNode`, they will encounter unexpected behavior in which the `checked` property is reset.

I believe the behavior would be more predictable and consistent if resetting properties to their default values were made configurable in `$copyNode`

etrepum: I've added an opt-out argument to $copyNode for the situations where you might want something like a clone but with a new key.

PR #8223, opened 2026-03-14
[lexical] Fix: preserve block equations ($$) in MarkdownShortcutPlugin

### LINKED ISSUE #6936: Bug: Block math equation is falsely converted to inline equation on MarkdownShortcutPlugin

Jynx2004:
## Description

Block math equations `($$...$$)` created via the toolbar were incorrectly converted to inline equations `($...$)` when toggling markdown mode via the MarkdownShortcutPlugin.

Root cause: The EQUATION transformer's export function always returned` $${equation}$` (inline syntax) regardless of whether the node was a block or inline equation. There was also no transformer to import/match `$$...$$ `patterns.

Closes #6936 
## Test plan

Created a new test for testing this functionality .

### Before

 inserting a block equation and clicking the Markdown button converted `$$x^2 + y^2 = z^2$$ → $x^2 + y^2 = z^2$` (inline)

### After

block equations are preserved as `$$x^2 + y^2 = z^2$$ `when toggling markdown mode

### DIFF
--- a/packages/lexical-playground/src/Editor.tsx
+++ b/packages/lexical-playground/src/Editor.tsx
@@ -75,6 +75,7 @@ import SpeechToTextPlugin from './plugins/SpeechToTextPlugin';
 import TabFocusPlugin from './plugins/TabFocusPlugin';
 import TableCellActionMenuPlugin from './plugins/TableActionMenuPlugin';
 import TableCellResizer from './plugins/TableCellResizer';
+import TableFitNestedTablePlugin from './plugins/TableFitNestedTablePlugin';
 import TableHoverActionsV2Plugin from './plugins/TableHoverActionsV2Plugin';
 import TableOfContentsPlugin from './plugins/TableOfContentsPlugin';
 import TableScrollShadowPlugin from './plugins/TableScrollShadowPlugin';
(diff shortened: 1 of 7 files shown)


etrepum:
See also #8194 which is probably a more complete fix but is not yet acceptable because there are no tests.

Jynx2004:
> This only handles export, not import, so it's not complete.

I think having a single import for both inline as well as block case will also work fine  as both have the same trigger `$`

etrepum:
It could possibly be implemented in one transformer or two, but either way this PR is not complete because it doesn’t have test coverage to show that import works. When you write those tests you’ll see that it probably doesn’t work as-is.

Jynx2004:
> It could possibly be implemented in one transformer or two, but either way this PR is not complete because it doesn’t have test coverage to show that import works. When you write those tests you’ll see that it probably doesn’t work as-is.

Made some changes in the unit test to check the import works for both the cases .

packages/lexical-playground/__tests__/unit/MarkdownTransformers.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        const markdown = editor.read(() =>
+          $convertToMarkdownString(transformers),
+        );
+
+        expect(markdown).toBe(`$$${equation}$$`);
+      });
+    },
+    {nodes: [EquationNode]},
+  );
```
etrepum: This test doesn't test anything that is not already tested by your export test and should be removed. What's needed are tests to ensure that the exported markdown can be imported and the result is an EquationNode with the right setting for inline and the correct contents. Of course the import regexes are not used while exporting, it's not meaningful to write tests for such a trivial thing especially when it's very uncommon to change the regexps in these transformers.

```suggestion
```

PR #8226, opened 2026-03-16
[lexical-link]: Refactor: add `afterCloneFrom` method to LinkNode/AutoLinkNode

levensta:
## Description

These changes fix some rare bugs that occur when using subclasses extended from LinkNode and relax the typing constraints of the `canBeEmpty` method

This PR adds the classic `afterCloneFrom` method to `LinkNode` to avoid copying issues, and adds an additional check to the transformation function to prevent unnecessary operations when no mutations occur. 

### Before

### After

### DIFF
--- a/packages/lexical-link/src/LexicalLinkNode.ts
+++ b/packages/lexical-link/src/LexicalLinkNode.ts
@@ -339,7 +339,7 @@ export function $linkNodeTransform(link: LinkNode): void {
         const innerLink = $copyNode(link);
         innerLink.append(...blockChildren);
         node.append(innerLink);
-        transformed ||= true;
+        transformed = true;
       }
       $insertNodeToNearestRootAtCaret(node, $rewindSiblingCaret(caret), {
         $shouldSplit: () => false,
@@ -349,7 +349,7 @@ export function $linkNodeTransform(link: LinkNode): void {
   if (!transformed) {
     return;
   }
-  if (link.isEmpty()) {
+  if (!link.canBeEmpty() && link.isEmpty()) {
     const parent = link.getParent();
     link.remove();
     if (parent && parent.isEmpty()) {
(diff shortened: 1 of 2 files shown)


packages/lexical-link/src/__tests__/unit/LexicalLinkNode.test.ts
```diff
@@ -1266,4 +1266,25 @@ describe('LinkNode transform (Regression #8083)', () => {
       expect(selection.anchor.offset).toBe(7);
     });
   });
+
+  test('an empty link is not deleted if the transformation did not occurred', () => {
```
```suggestion
  test('an empty link is not deleted if the transformation did not occur', () => {
```

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -334,12 +339,16 @@ export function $linkNodeTransform(link: LinkNode): void {
         const innerLink = $copyNode(link);
         innerLink.append(...blockChildren);
         node.append(innerLink);
+        transformed ||= true;
```
etrepum: No reason to check the existing value
```suggestion
        transformed = true;
```

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -334,12 +339,16 @@ export function $linkNodeTransform(link: LinkNode): void {
...
       }
       $insertNodeToNearestRootAtCaret(node, $rewindSiblingCaret(caret), {
         $shouldSplit: () => false,
       });
     }
   }
+  if (!transformed) {
+    return;
+  }
```
etrepum: This should really only affect the selection code below

packages/lexical-link/src/LexicalLinkNode.ts
etrepum: Might be worth checking canBeEmpty in here

PR #8228, opened 2026-03-16
[lexical-playground] Fix: Fixing cursor position after inline equation, fix block equations

### LINKED ISSUE #7606: Bug: Fixing cursor position after equation

Jynx2004:
## Description

Added the display attribute to the parent div/span of the equation node and setting its value according to the `__inline` value taken from the user.

https://github.com/user-attachments/assets/3b520ddc-e0af-4e5e-b966-bb5119f8fd81

Closes #7606 

## Test plan

Not have added any tests as we can see it in our inspect tab in the particular browser.

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/EquationNode.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/EquationNode.spec.mjs
@@ -58,12 +58,13 @@ function equationHtml(inline = true) {
 }
 
 test.describe('EquationNode', () => {
-  test.beforeEach(({isCollab, page}) =>
-    initialize({
+  test.beforeEach(({isCollab, isPlainText, page}) => {
+    test.skip(isPlainText);
+    return initialize({
       isCollab,
       page,
-    }),
-  );
+    });
+  });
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/__tests__/e2e/EquationNode.spec.mjs
Jynx2004, line 1: Why it’s happening:
The selectors that time out are only rendered in rich‑text mode: the toolbar button (Insert specialized editor node) and the equation node (.editor-equation).
Your run is almost certainly in plain‑text E2E mode (E2E_EDITOR_MODE=plain-text), where the toolbar and markdown/equation features are intentionally absent.
Many tests already skip themselves in plain‑text, but EquationNode.spec.mjs didn’t, so it waits forever for elements that will never appear.

Fix applied (safe for rich‑text):
I updated the EquationNode tests to skip when isPlainText is true, matching the rest of the suite.

PR #8230, opened 2026-03-17
[lexical-devtools-core] Bug Fix: Clean up strict mode useLexicalCommandsLog behavior

etrepum:
## Description

The `useLexicalCommandLog` indexes would skip entries when run in dev (React strict mode) because setState callbacks are called twice. The callback had observable side-effects.

### Before

Run `pnpm dev` and observe the command log indices missing numbers in the series (e.g. 1, 3, 5, 7)

### After

Run `pnpm dev` and observe that the command log indices are now dense (e.g. 1, 2, 3, 4)

### DIFF
--- a/packages/lexical-devtools-core/src/useLexicalCommandsLog.ts
+++ b/packages/lexical-devtools-core/src/useLexicalCommandsLog.ts
@@ -9,11 +9,13 @@
 import type {LexicalEditor} from 'lexical';
 
 import {COMMAND_PRIORITY_CRITICAL, LexicalCommand} from 'lexical';
-import {useEffect, useMemo, useState} from 'react';
+import {useEffect, useState} from 'react';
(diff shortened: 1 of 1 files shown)


packages/lexical-devtools-core/src/useLexicalCommandsLog.ts
```diff
@@ -9,41 +9,34 @@
...
-            return newState;
-          });
-
+          index += 1;
+          const entry: LexicalCommandEntry = {
+            index,
+            payload,
+            type: command.type ? command.type : 'UNKNOWN',
+          };
+          setLoggedCommands((state) => [...state.slice(-9), entry]);
```
etrepum, line 39: This is the actual fix, the rest of the PR is just cleanup

PR #8234, opened 2026-03-18
[lexical-table] Bug Fix: handle table selections crossing into/out of nested tables

randal-atticus:
## Description
Further improvements to table selections, and specifically fixing a nested table selection case.

`$fixRangeSelectionForSelectedTable` now only runs once (instead of once per table), in line with previous refactors to avoid running all the logic across every table, thereby making it hard to handle nested tables.

The logic has been updated to correctly handle Range Selections that cross into/out of a nested table. Just as before, the (anchor OR focus) is shifted to the (start OR end) of the (anchor OR focus) table depending on the direction of the selection.

Added Playground tests to assert behaviour for nested and non-nested tables.

### Before
On Playground, observe that range selections crossing a nested table will erroneously expand to the outer table.

https://github.com/user-attachments/assets/b5a28716-ab3a-4d82-b7fb-c7715ee70cf8

### After

As of this PR, range selections crossing a nested table no longer expand to the outer table.

https://github.com/user-attachments/assets/d6c6a9d6-df4b-4d01-9984-afe208ef310b

(Just noticed the "table selection" on the inner table is a bit too sticky, but that's a separate bug)

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Tables.spec.mjs
@@ -7798,12 +7798,17 @@ test.describe.parallel('Tables', () => {
     }
 
     test('Range-select from above nested table into it selects the entire table, but not the outer table', async ({
+      browserName,
       page,
       isPlainText,
       isCollab,
     }) => {
       test.skip(isPlainText);
       test.skip(isCollab);
+      test.fixme(
+        browserName === 'firefox',
+        'Erroneously selects the text after the table as well',
+      );
       await initialize({hasNestedTables: true, page});
 
       await setupTables(page);
@@ -7830,6 +7835,7 @@ test.describe.parallel('Tables', () => {
     });
 
     test('Range-select from below nested table into it selects the entire table, but not the outer table', async ({
+      browserName,
       page,
       isPlainText,
       isCollab,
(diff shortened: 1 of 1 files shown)


randal-atticus:
New tests failing in Firefox, investigating.

Seems like on Firefox, a shift-click selection into or out of a nested table selects more than the user requested, but not for top-level tables.

https://github.com/user-attachments/assets/72143873-b6a4-452b-826b-50d4e8f2a1f8

EDIT: Ended up disabling these tests on Firefox, because this is what was happening before (above video is from live Playground), and I would rather not let perfect get in the way of good.

packages/lexical-table/src/LexicalTableSelectionHelpers.ts
```diff
@@ -921,130 +925,139 @@ export function $handleTableSelectionChangeCommand(
...
+ * Handles cases where range selections cross into, out of, or within tables.
+ */
+function $fixRangeSelectionForSelectedTable(
   selection: RangeSelection,
-  tableNode: TableNode,
-  tableObserver: TableObserver,
+  tableObservers: TableObservers,
 ) {
   const prevSelection = $getPreviousSelection();
-  if ($isRangeSelection(selection)) {
```
randal-atticus, line 944: selection was already a RangeSelection. Unfortunately, this bloated the diff, but reviewing [without whitespace changes](https://github.com/facebook/lexical/pull/8234/changes?w=1) makes this more legible

PR #8235, opened 2026-03-18
[lexical-rich-text] Bug Fix: use writable node in HeadingNode.setTag

karesansui-u:
## Summary

`HeadingNode.setTag()` calls `getWritable()` but writes to `this.__tag` instead of the returned writable clone (`self`). This bypasses Lexical's immutability model — the original (possibly frozen) node is mutated directly.

The fix changes `this.__tag = tag` to `self.__tag = tag`, matching the pattern used elsewhere (e.g. `CodeNode.setLanguage()`).

## Test plan

- Added unit test for `HeadingNode.setTag()` verifying the tag is updated correctly
- All existing tests pass (`pnpm test-unit`: 2112 passed)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3309,24 +3309,28 @@ packages:
     engines: {node: '>= 10'}
     cpu: [arm64]
     os: [linux]
+    libc: [glibc]
 
   '@node-rs/jieba-linux-arm64-musl@1.10.4':
     resolution: {integrity: sha512-Y/tiJ1+HeS5nnmLbZOE+66LbsPOHZ/PUckAYVeLlQfpygLEpLYdlh0aPpS5uiaWMjAXYZYdFkpZHhxDmSLpwpw==}
     engines: {node: '>= 10'}
     cpu: [arm64]
     os: [linux]
+    libc: [musl]
 
   '@node-rs/jieba-linux-x64-gnu@1.10.4':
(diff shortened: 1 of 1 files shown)


karesansui-u:
@etrepum Thank you for the review — you're absolutely right. The original test ran everything in a single `update()`, so `getWritable()` returned the same object and the test would pass regardless of whether `this.__tag` or `self.__tag` was used.

I've updated the test to use two separate `{discrete: true}` updates:

1. First update: create and append an `h1` heading, flush to DOM
2. Second update: retrieve the existing node via `getFirstChildOrThrow()`, call `setTag('h2')`, and verify both the node state and the rendered DOM (`<h1>` → `<h2>`)

With this structure, `getWritable()` returns a cloned node in the second update, so only `self.__tag` (the fix) correctly mutates the writable copy. The buggy `this.__tag` would leave the DOM unchanged, causing the `outerHTML` assertion to fail.

karesansui-u:
@etrepum Sorry about the back and forth — my test didn't properly validate the fix, and I missed the `getLatest()` issue entirely. Thank you for catching both.

I've pushed a new commit that:
- Adds `getLatest()` to `getTag()` to match the pattern in `ListNode.getTag()` and other node getters
- Reverts the test to assert on `heading.getTag()` directly, since `getLatest()` now ensures consistent reads from any node reference

Appreciate the thorough review — learned a lot about Lexical's node versioning model.

pnpm-lock.yaml
etrepum, line 1: This file should not change, remove this from the PR.

karesansui-u: I'm really sorry about that — reverted in the latest push.

PR #8236, opened 2026-03-18
[lexical-link] Feature : Merge adjacent LinkNodes with identical attributes

### LINKED ISSUE #8150: Feature: Merging two identical links into one node

achaljhawar:
## Description

Adjacent `LinkNode`s with identical attributes (url, target, rel, title) are now automatically merged into a single node via the existing `$linkNodeTransform`

Closes #8150

### Before

https://github.com/user-attachments/assets/26c9bea9-e40e-4930-b7d0-c2d2789356fe

### After

https://github.com/user-attachments/assets/53d7a072-994e-4990-9b7c-a93ae6d7bf26

### DIFF
--- a/packages/lexical-link/src/LexicalLinkNode.ts
+++ b/packages/lexical-link/src/LexicalLinkNode.ts
@@ -479,7 +479,7 @@ export class AutoLinkNode extends LinkNode {
     );
   }
 
-  shouldMergeAdjacentLink(): boolean {
+  shouldMergeAdjacentLink(_otherLink: LinkNode): boolean {
     return false;
   }
 


packages/lexical-link/src/__tests__/unit/LinkExtension.test.ts
```diff
@@ -58,4 +69,182 @@ describe('Link', () => {
     );
     editor.dispose();
   });
+
+  describe('merge adjacent LinkNodes', () => {
+    it('merges adjacent LinkNodes with identical attributes', () => {
+      const editor = buildEditorFromExtensions(extension);
+      editor.update(
+        () => {
+          const p = $getRoot().getFirstChild<ElementNode>()!;
```
etrepum: Having an explicit type parameter for this function's return value was a mistake, it's not at all safe (equivalent to using "as ElementNode | null") and shouldn't be used. It doesn't really matter in tests but people will use it as an example and that's not great (and even worse to train LLM models on bad code). Only commenting on the first occurrence of this issue, there are several that should be fixed.

```suggestion
          const p = $getRoot().getFirstChild();
          assert($isParagraphNode(p), 'Expecting a ParagraphNode in the root');
```

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -346,6 +355,17 @@ export function $linkNodeTransform(link: LinkNode): void {
       });
     }
   }
+  if (link.isAttached() && !$isAutoLinkNode(link)) {
+    const nextSibling = link.getNextSibling();
+    if (
+      $isLinkNode(nextSibling) &&
+      !$isAutoLinkNode(nextSibling) &&
+      $haveSameLinkAttributes(link, nextSibling)
```
etrepum: I think we will want to allow this to be overridden because if someone has a subclass of LinkNode in their app they might want to change the conditions for merge (e.g. LinkNode and CustomLinkNode might not ever get merged)

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -346,6 +355,17 @@ export function $linkNodeTransform(link: LinkNode): void {
       });
     }
   }
+  if (link.isAttached() && !$isAutoLinkNode(link)) {
+    const nextSibling = link.getNextSibling();
```
etrepum, line 374: Checking only to the right can miss potential merges under some scenarios, in the case where you have an editor with:

* Link(someUrlA)
* Link(someUrlB)

Then in an subsequent update you change only the attributes of the second link to match the first:

* Link(someUrlA)
* Link(someUrlA)

This will not cause a merge. Only the second link is marked dirty, so the transform is only called on the second link, which does not inspect the link to its left for a potential merge.

The tested scenarios only cover situations where all of the LinkNodes are dirty so the transform is called on each of them.

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -297,6 +297,15 @@ export class LinkNode extends ElementNode {
       this.__url.startsWith('https://') || this.__url.startsWith('http://')
     );
   }
+
+  $shouldMergeAdjacentLink(otherLink: LinkNode): boolean {
```
etrepum: node methods don't generally have a $ prefix

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -297,6 +297,15 @@ export class LinkNode extends ElementNode {
...
+    return (
+      this.getType() === otherLink.getType() &&
+      !$isAutoLinkNode(this) &&
+      !$isAutoLinkNode(otherLink) &&
```
etrepum: Would be simpler to override this method from AutoLinkNode and have it always return false, the type match is sufficient to prevent a LinkNode->AutoLinkNode merge.

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -297,6 +297,15 @@ export class LinkNode extends ElementNode {
...
+      $haveSameLinkAttributes(this, otherLink)
```
etrepum: This can just be inlined, it's not used anywhere else so no need to make it a function

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -346,6 +364,35 @@ export function $linkNodeTransform(link: LinkNode): void {
...
+      $isLinkNode(nextSibling) &&
+      link.$shouldMergeAdjacentLink(nextSibling)
+    ) {
+      link.append(...nextSibling.getChildren());
+      nextSibling.remove();
+      if (!transformed && $isRangeSelection(selection)) {
+        $restoreCaretPair(selection.anchor, anchorPair!);
+        $restoreCaretPair(selection.focus, focusPair!);
+        $normalizeSelection__EXPERIMENTAL(selection);
+      }
```
```suggestion
      transformed = true;
```

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -346,6 +364,35 @@ export function $linkNodeTransform(link: LinkNode): void {
...
+      $isLinkNode(prevSibling) &&
+      prevSibling.$shouldMergeAdjacentLink(link)
+    ) {
+      prevSibling.append(...link.getChildren());
+      link.remove();
+      if ($isRangeSelection(selection)) {
+        $restoreCaretPair(selection.anchor, anchorPair!);
+        $restoreCaretPair(selection.focus, focusPair!);
+        $normalizeSelection__EXPERIMENTAL(selection);
+      }
```
etrepum: Maybe extract this into a function defined earlier, e.g. `$restoreSelection()` so that its implementation doesn't need to be repeated multiple times.

packages/lexical-link/src/LexicalLinkNode.ts
```diff
@@ -451,6 +479,10 @@ export class AutoLinkNode extends LinkNode {
     );
   }
 
+  shouldMergeAdjacentLink(): boolean {
```
```suggestion
  shouldMergeAdjacentLink(_otherLink: LinkNode): boolean {
```

Leaving the signature alone should help with subclassing

etrepum (approved):

PR #8237, opened 2026-03-18
[lexical-playground] Fix :  Clear block formatting if all of its content is selected and handle the case of partial selection.

### LINKED ISSUE #7383: Bug: [lexical-playground] clear-formatting doesn't clear text-left/right/center alignment when no formatting to text

Jynx2004:
## Description

Updated clear-formatting to detect the “no formatting” case by checking selection.extract() instead of comparing anchor/focus. This ensures alignment is cleared even when the selection is collapsed and no text formatting is present.

Closes #7383 

https://github.com/user-attachments/assets/7dad66bb-8cd1-4153-923f-49dd71786312

## Test plan

All the existing unit tests passed . No new tests created.


etrepum (commented):
This PR needs tests to demonstrate that it solves the issue

Jynx2004:
> This PR needs tests to demonstrate that it solves the issue

I have added the unit test to check this functionality

etrepum (commented):
I don't think this is quite a complete solution, it's a bit greedy in that it will always clear formatting from the parent even if you've only selected some of the text. I think the block formatting should only be cleared if the block is selected, or all of its contents are selected.

Jynx2004:
> I don't think this is quite a complete solution, it's a bit greedy in that it will always clear formatting from the parent even if you've only selected some of the text. I think the block formatting should only be cleared if the block is selected, or all of its contents are selected.

I have made changes  where block formatting is applied if all of the block's contents are selected and not for partial selection.

packages/lexical-playground/__tests__/unit/ToolbarPluginClearFormatting.test.ts
```diff
@@ -0,0 +1,60 @@
...
+          paragraphKey = paragraph.getKey();
+          $getRoot().clear().append(paragraph);
+
+          const firstChild = paragraph.getFirstChild();
+          if (!$isTextNode(firstChild)) {
+            throw new Error('Expected first child to be a TextNode.');
+          }
+          const selection = $createRangeSelection();
+          selection.setTextNodeRange(firstChild, 0, firstChild, 3);
+          $setSelection(selection);
```
etrepum:
```suggestion
          firstChild.select(0, 3);
```

packages/lexical-playground/__tests__/unit/ToolbarPluginClearFormatting.test.ts
```diff
@@ -0,0 +1,60 @@
...
+          const paragraph = $createParagraphNode();
+          paragraph.setFormat('center');
+          paragraph.append($createTextNode('Hello'), $createTextNode('World'));
+          paragraphKey = paragraph.getKey();
+          $getRoot().clear().append(paragraph);
+
+          const firstChild = paragraph.getFirstChild();
+          if (!$isTextNode(firstChild)) {
+            throw new Error('Expected first child to be a TextNode.');
+          }
```
etrepum: this should be an `assert`

packages/lexical-playground/__tests__/unit/ToolbarPluginClearFormatting.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        {discrete: true},
+      );
+
+      clearFormatting(editor);
+
+      editor.getEditorState().read(() => {
+        const paragraph = $getNodeByKey(paragraphKey);
+        if (!$isParagraphNode(paragraph)) {
+          throw new Error('Expected a ParagraphNode.');
+        }
```
etrepum: `assert`

packages/lexical-playground/__tests__/unit/ToolbarPluginClearFormatting.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  $setSelection,
+} from 'lexical';
+import {initializeUnitTest} from 'lexical/src/__tests__/utils';
+import {describe, expect, it} from 'vitest';
+
+import {clearFormatting} from '../../src/plugins/ToolbarPlugin/utils';
+
+describe('ToolbarPlugin clearFormatting', () => {
+  initializeUnitTest((testEnv) => {
+    it('keeps block format when selection is partial', () => {
```
etrepum, line 24: I would expect to see more tests than this, e.g. test when there is no text, test with a collapsed selection, test with a complete selection

packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
```diff
@@ -300,6 +304,36 @@ function $findParagraphParent(node: LexicalNode): ElementNode | null {
   return $isElementNode(parent) && $isParagraphNode(parent) ? parent : null;
 }
 
+function $isBlockFullySelected(
```
etrepum, line 309: The implementation of this function doesn't seem quite right, I wouldn't expect to see any text-specific code in here. What I would do is check that the start and end points are at or outside of the beginning and end of the block element. `$normalizeCaret` will be helpful for this also

packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
```diff
@@ -338,6 +372,24 @@ export const clearFormatting = (
...
+      const selectionStart = $createPoint(
+        startPoint.key,
+        startPoint.offset,
+        startPoint.type,
+      );
+      const selectionEnd = $createPoint(
+        endPoint.key,
+        endPoint.offset,
+        endPoint.type,
+      );
```
etrepum: This probably all just be `$caretRangeFromSelection`, no need to create separate points for this. You need to create the `CaretRange` (or inspect the points) *after* the `selection.extract()` operation because it changes the selection when the boundary nodes get split.

packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
```diff
@@ -357,7 +401,14 @@ export const clearFormatting = (
           const nearestBlockElement =
             $getNearestBlockElementAncestorOrThrow(node);
           if (nearestBlockElement.getFormat() !== 0) {
-            nearestBlockElement.setFormat('');
+            // Only clear block format if we have a pre-extract range (i.e. not
+            // a TableSelection) and the block was fully covered by it.
+            if (
+              preExtractCaretRange !== null &&
```
etrepum: The pre-extract range shouldn't be used because it may no longer be valid

packages/lexical-playground/src/plugins/ToolbarPlugin/utils.ts
```diff
@@ -338,11 +362,25 @@ export const clearFormatting = (
...
+        'next',
+      );
+      const focusNext = $getCaretInDirection(
+        postExtractCaretRange.focus,
+        'next',
+      );
+
+      if ($comparePointCaretNext(anchorNext, focusNext) === 0) {
         return;
       }
```
etrepum, line 385: I'm not quite sure this is the correct logic but it looks like you were just trying to check if the selection was collapsed

```suggestion
      if (!$isRangeSelection(postExtractSelection) || postExtractSelection.isCollapsed()) {
        return;
      }
```

etrepum (commented):
I don't think this is working correctly because if I create an empty ParagraphNode, set some formatting (e.g. center alignment) and then clear formatting then it stays center aligned

etrepum:

PR #8240, opened 2026-03-19
[lexical-markdown] Bug Fix: nest heading inside quote when typing `> # SOME HEADER`

### LINKED ISSUE #7407: Bug: Markdown block quotes containing headers dropping quote
I'm not sure if this is by design or if its a bug.  I know in other platforms or tools that support markdown, you can embed a header in a block quote, but with lexical, the header appears to override the quote and cause it to be dropped.  I can format with aspects like bold or italic in the quote, but not a header.  Is this expected behavior?

Lexical version:

## Steps To Reproduce

1. When markdown is enabled, start a block quote by typing ">"
2. Try creating a header by typing "# SOME HEADER"

Link to code example:

This can be done in the playground

## The current behavior
The block quote is dropped and it becomes a header

## The expected behavior
Header is able to be embedded in the quote

achaljhawar:
## Description

Typing `> # SOME HEADER` created a heading but dropped the block quote. The heading transformer was replacing the QuoteNode instead of nesting inside it. Fixed createBlockNode to nest inside non-paragraph parents.

### Before

https://github.com/user-attachments/assets/36122b4d-5a20-4823-976e-d1afaebab28b

### After

https://github.com/user-attachments/assets/0fcd0947-e580-4165-80b7-d26f3a71b247


packages/lexical-markdown/src/__tests__/unit/MarkdownTransformers.test.ts
```diff
@@ -156,3 +160,19 @@ describe('LINK', () => {
     });
   });
 });
+
+describe('BLOCK QUOTE + HEADING', () => {
+  test('typing "> # SOME HEADER" creates a heading inside a quote (issue #7407)', () => {
```
etrepum: this only tests the markdown shortcut, but import and export should also be tested to ensure that it's feature complete

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -241,7 +242,12 @@ const createBlockNode = (
   return (parentNode, children, match, isImport) => {
     const node = createNode(match);
     node.append(...children);
-    parentNode.replace(node);
+    if (!isImport && !$isParagraphNode(parentNode)) {
+      parentNode.clear();
+      parentNode.append(node);
+    } else {
```
etrepum: this seems suspicious, why clear the node and why does this only happen for shortcuts?

etrepum (commented):
The current lexical document model (at least as implemented by the playground) doesn't really support blocks that contain blocks, except in cases where there's a shadow root in between or there's some special code to support it (ListItem -> ListItemNode). There might be other implications of allowing nesting here, it might be best to have this behavior as an option that defaults to false until all of the potential issues (unrelated to markdown export/import specifically) can be evaluated.

etrepum (commented):
This is kind of an awkward way to do it, I would expect these to be distinct transformers that are exported separately and could be used to build an alternate transformers array. Ideally there would be some way to pass options down through to these transformers but that would require a larger refactor of the markdown API to use extensions.

I'm also not sure that it's worth adding this complexity just to support nesting these two types of nodes specifically, certainly not with this specific approach. There is also very likely going to be other problems to fix when using a document that nests these blocks because I believe that there are places in lexical's model that assume blocks aren't nested like this.

achaljhawar:
Hey @etrepum, I looked into the block nesting concern. Core selection/insertion algorithms find HeadingNode as the leaf block before reaching QuoteNode, so they all work correctly. The only issue is the playground toolbar showing "Quote" instead of the heading tag. Let me know if you'd rather I rework this or just close it.

etrepum:
I'll take another look at this next week, my primary concern here is that it's only a partial solution because this structure isn't well supported by the editor and violates some of the assumptions about how elements can typically be nested (block nodes that are not shadow roots can typically only contain inline nodes).

achaljhawar:
Hey @etrepum, any chance you got to look at this? Happy to try a different approach.

potatowagon:
## Review: Nest Heading Inside Quote When Typing `> # SOME HEADER`

### Summary
Refactors `HEADING` and `QUOTE` transformers from static objects into factory functions (`$createHeadingTransformer` and `$createQuoteTransformer`) that accept options. Adds support for typing `> # Header` to create a heading nested inside a blockquote — a common Markdown pattern not previously supported.

### What I Verified
- **Backward compatibility**: The exported `HEADING` and `QUOTE` constants are still present (they call the factories with no options), so existing consumers are unaffected.
- **Factory pattern**: `$createHeadingTransformer({nestInBlockquote: true})` creates a heading transformer aware of quote context. `$createQuoteTransformer({handleNestedHeadings: true})` creates a quote transformer that parses `> # ...` syntax.
- **Regex correctness**: `QUOTE_WITH_HEADING_REGEX = /^>\s(?:(#{1,6})\s)?/` correctly matches optional heading markers after the blockquote prefix.
- **Export roundtrip**: When `handleNestedHeadings` is enabled, a QuoteNode containing a single HeadingNode exports as `> # Text` — proper roundtrip.
- **CI status**: All tests pass across all platforms.

### Concerns
- **API surface change**: This changes two core transformer constants from plain objects to factory-generated objects. While backward-compatible, it shifts the architectural pattern. The default `TRANSFORMERS` array still uses the no-option versions, so the global behavior is unchanged.
- **Interaction complexity**: Users who enable both `nestInBlockquote` on HEADING and `handleNestedHeadings` on QUOTE need to understand they work together. Could benefit from documentation.
- **Potential conflict**: If a user has both the default HEADING transformer and a custom quote transformer with heading support, the heading regex might match before the quote regex gets a chance to parse `> # ...`.

### Verdict
⚠️ **Needs maintainer sign-off on API direction** — the refactor is correct and well-tested, but the factory pattern for core transformers is an architectural decision. The implementation quality is good. If the maintainers are okay with the pattern, it's safe to merge.

potatowagon (commented):
## Review — Nest Heading Inside Quote When Typing `> # SOME HEADER`

### What I verified:

1. **Feature scope**: Enables `> # Heading` markdown shortcut to create a heading nested inside a blockquote. Currently, `> text` creates a quote, and `# text` creates a heading, but the combination doesn't nest them — the heading replaces the quote.

2. **Implementation**: Refactors the static `HEADING` transformer into a factory `$createHeadingTransformer(options)` with an opt-in `nestInBlockquote` flag. When the parent node is a QuoteNode and the flag is set, the heading appends into the quote instead of replacing it. The `QUOTE` transformer is similarly refactored into `$createQuoteTransformer`.

3. **Backward compatibility**: The existing `HEADING` and `QUOTE` exports are preserved as instances of the factory with defaults — no breaking change for existing consumers.

4. **CI status**: Full CI suite green (34 checks pass).

5. **Test coverage**: Tests verify the nesting behavior for both import and shortcut scenarios.

_— via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=8240&ref=potatowagon) on behalf of potatowagon_

etrepum (dismissed):

etrepum:
There are still some issues with this one, will not go into v0.50.0

PR #8247, opened 2026-03-20
[RFC only][lexical-table][lexical-playground][lexical-core][lexical-clipboard] Feature: Table pasting based on range selection

randal-atticus:
## Description

Attempting to implement this functionality, to make it easier to paste nested tables into each other.

https://github.com/user-attachments/assets/b3dd03cd-14c7-44d5-9f5a-2aed1740e0d4

This ended up being harder than I expected because the selection is not actually encoded in the clipboard. Therefore, by the time we get to `LexicalTablePluginHelpers` to insert the table node, all we have is the deserialized TableNode, and no information about _how_ the table was copied.

We don't have a lot of levers to encode the selection into the clipboard yet. I kind of wonder if the originating selection should be serialized into the clipboard payload somehow (although selections refer to nodes by keys, so that's a bit tricky).

I hacked around this by adding a property to the Table (`wholeTable`) but this is pretty bad. I'm looking for suggestions on how this should actually be implemented.

### Before

### After

### DIFF
--- a/packages/lexical-table/src/LexicalTableNode.ts
+++ b/packages/lexical-table/src/LexicalTableNode.ts
@@ -56,6 +56,7 @@ export type SerializedTableNode = Spread<
     rowStriping?: boolean;
     frozenColumnCount?: number;
     frozenRowCount?: number;
+    wholeTable?: boolean;
   },
   SerializedElementNode
 >;
@@ -177,6 +178,7 @@ export class TableNode extends ElementNode {
   __frozenColumnCount: number;
   __frozenRowCount: number;
   __colWidths?: readonly number[];
+  __wholeTable?: boolean; // ephemeral
 
   static getType(): string {
     return 'table';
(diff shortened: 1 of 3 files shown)


packages/lexical-table/src/LexicalTableNode.ts
```diff
@@ -236,15 +240,20 @@ export class TableNode extends ElementNode {
     this.__colWidths = undefined;
   }
 
-  exportJSON(): SerializedTableNode {
+  exportJSON(selection: BaseSelection | null = null): SerializedTableNode {
     return {
       ...super.exportJSON(),
+      // TODO: export appropriate colWidths based on the selection.
```
randal-atticus, line 246: Discovered this bug: because the colWidths are copied from the source table, copying N columns _not_ from the start of the table will use the colWidths from the first N columns.

https://github.com/user-attachments/assets/f62db070-7e6b-4e4a-81d5-4304b3b60677

PR #8248, opened 2026-03-20
[*] Chore: Github CI workflows action updates and run with node 24

### LINKED ISSUE #8244: Bug: Update workflows to use node 24
Node 20 is deprecated, would make sense to make sure it's all running on the latest supported runtime.

>  Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-node@v4, pozetroninc/github-action-get-latest-release@master. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

etrepum:
## Description

Updates the actions we depend on as some of them are using node 20 which is deprecated, also updates our CI matrix to primarily use node 24 but also run unit tests with node 22.

Closes #8244

## Test plan

Workflows should run as before (primarily CI)

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -21,7 +21,7 @@ const config = {
       use: {
         ...devices['Desktop Chrome'],
         launchOptions: {
-          slowMo: 20,
+          slowMo: 50,
         },
       },
     },
@@ -31,7 +31,7 @@ const config = {
       use: {
         ...devices['Desktop Firefox'],
         launchOptions: {
-          slowMo: 20,
+          slowMo: 50,
         },
       },
     },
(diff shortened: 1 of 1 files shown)


etrepum:
It looks like we need to upgrade playwright to use node 24 for some reason

PR #8250, opened 2026-03-20
[lexical] Fix: Correct the display position after creating a code block from selected text.

### LINKED ISSUE #6324: Bug: wrong display position after creating a code block from selected text.
When selecting text and creating a code block, the resulting code block appears in an incorrect position. Instead of being inserted at the location of the selected text, the code block is either displaced or appears in an unintended area of the document. This misalignment disrupts the document's layout and workflow, causing confusion and additional manual adjustments by the user.

## Steps To Reproduce

Please watch this video:

https://github.com/facebook/lexical/assets/64115961/7a51aae3-ae01-4f6e-9ca3-ee8078cf2e75

## The expected behavior
image

Jynx2004:
## Description

Summary

- Capture trailing text after a selection before code formatting mutates the selection.
- Insert the trailing text as a sibling paragraph after the newly created code block.
- Add a unit test that verifies the selected text moves into a code block and the trailing text appears in the next paragraph.

Why

- Ensures code formatting preserves the text after the selection and places it as a sibling paragraph, matching expected editor behavior.
- Locks the new behavior in with a unit test.

Closes #6324 

## Before

https://github.com/user-attachments/assets/5ba0bbfd-ecc4-492b-bb3e-3c2fd23f75e7

## After

https://github.com/user-attachments/assets/392d7fd6-3a70-445c-a989-0cc2342fd853

## Test plan

Unit Test added


etrepum (commented):
I think the only approach worth taking here is to make formatCode behave as expected, rather than have it ignore this known edge case.

Jynx2004:
> I think the only approach worth taking here is to make formatCode behave as expected, rather than have it ignore this known edge case.

I have made the changes in the existing functionality without changing the expected behavior. Also have added the unit test to check that its working as expected.

Jynx2004:
Hey @etrepum let me know if i need to change anything or improvise in this PR .

etrepum (commented):
I think this implementation would be better if it used something like selection.extract to get the nodes that will be placed in the CodeNode and then the remaining nodes in the block can be moved to a new block. This way the nodes can retain any formatting and it will support more than just TextNode.

https://github.com/user-attachments/assets/693cebc0-685d-4532-9b5f-505f41b4a89d

Jynx2004:
> I think this implementation would be better if it used something like selection.extract to get the nodes that will be placed in the CodeNode and then the remaining nodes in the block can be moved to a new block. This way the nodes can retain any formatting and it will support more than just TextNode.
> 
>  Screen.Recording.2026-04-21.at.08.57.05.mov

Have used the selection.extract to get the nodes that will be placed outside and down the code block . This preserves the format of the TextNodes

etrepum:

PR #8278, opened 2026-04-01
[lexical] Fix: Improves the escape key handling when a range selection's anchor and focus points are positioned relative to a table

### LINKED ISSUE #7083: Bug: Table Selection State Persists After Cmd+A and Esc
Lexical version: Latest(Playground)

## Steps To Reproduce

1. Insert a TableNode
2. Select all in the editor using Cmd+A
3. Press the Esc key

https://github.com/user-attachments/assets/c0facf34-3c42-45cd-beba-cebaa56bf7dd

## Link to code example

-   Playground

## The current behavior

-   The selection state of the table remains active.

## The expected behavior

-   The table selection state should be cleared and no color should be applied.

## Impact of fix

-   I don't think it has a major impact.

Jynx2004:
## Description

This fix improves the escape key handling when a range selection's anchor and focus points are positioned relative to a table. Before this change, there were edge cases where table selection highlights would persist or not be properly cleared when the user pressed escape with a range selection that had the table positioned between its endpoints.

### Enhanced the KEY_ESCAPE_COMMAND handler to properly detect and handle two scenarios involving range selections and tables:

**Case 1: Mixed endpoints** — When one endpoint of the range selection is inside the table and the other is outside, the code now correctly clears the table highlight without moving the cursor.

**Case 2: Table between endpoints** — When both endpoints are outside the table but the table falls between them (e.g., selection from a node before the table to a node after it), the code now detects this by comparing node ordering (isBefore checks) and clears the table highlight accordingly.

Closes #7083


etrepum (commented):
I don't think this is really handling the situation in the correct way, it should either be managing the highlight based on focus/blur or using CSS to ensure it's not displayed when not focused per the comment on the issue

etrepum:

PR #8281, opened 2026-04-03
[example] Feature: Add a new agent-example using a local LLM

etrepum:
## Description

Adds a new agent-example that demonstrates using an in-browser LLM to interact with the lexical editor. Two use cases are presented:

- Extract Entities: labels names and places found in the text as decorators
- Generate: generates some gibberish (the model is small) based on the preceding text

Nearly all of this was "authored" by Claude but you can see from the commit history that it took a whole lot of guidance to get here 😆 

[stackblitz link](https://stackblitz.com/github/etrepum/lexical/tree/claude/ai-agent-rich-text-D3N2G/examples/agent-example)
[website preview](https://lexical-git-fork-etrepum-claude-ai-agent-ri-c63730-fbopensource.vercel.app/)

Uses all of the latest features from lexical including Extensions, DecoratorTextNode, NodeState, and $config. Also has some utilities in useExtensionHooks that we should probably ship in `@lexical/react`.

Some novel techniques used here:
* `$collectTextNodeOffsets` used by Extract Entities allows the AI target specific locations in the document to replace based on the string we gave it
* `AICursorNode` (a blinking cursor DecoratorNode) provides an insertion caret for Generate so the LLM has a stable place to insert text even if weird stuff is happening (TextNode merges, markdown shortcuts, whatever). This is temporary and is removed when the generation completes or is aborted. It's also a demonstration of a framework independent decorator that just does its job in `createDOM` and has a no-op `decorate`.
* Generally most editor related code is moved into extensions and out of React which makes the lifecycle simpler and makes it trivial to port to any other frontend ecosystem

https://github.com/user-attachments/assets/066537df-16b7-401e-8696-8fb893ba001e

## Test plan

The example has some unit tests for the more complex functionality it has. Claude wasn't able to successfully one-shot anything so I had to make it write tests.


packages/lexical-website/src/components/HomepageExamples.tsx
```diff
@@ -6,13 +6,18 @@
...
+import BrowserOnly from '@docusaurus/BrowserOnly';
 import ChatEditor from '@examples/website-chat/Editor';
 import NotionEditor from '@examples/website-notion/Editor';
 import RichInputEditor from '@examples/website-rich-input/Editor';
 import React from 'react';
 
 import StackBlitzButton from './StackBlitzButton';
 
+const AgentEditor = React.lazy(() =>
+  import('@examples/agent-example/Editor').then((m) => ({default: m.Editor})),
```
etrepum, line 18: This is mostly to force transformers.js to be loaded browser only and to avoid having to add some SSR code to manage the preloaded editor content. I'm sure this could be worked around differently but it's below the fold so SSR is not very important anyhow.

scripts/update-tsconfig.mjs
```diff
@@ -5,14 +5,13 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-// @ts-check
```
etrepum, line 8: This file was refactored from js to mjs because generally we should be moving everything to esm, but the target change here was to fix a latent issue introduced with the new website where the next update-tsconfig would've removed some manually added paths to support the embedded example editors

zurfyx (approved):

PR #8311, opened 2026-04-08
[lexical-markdown] Bug Fix: Escape ordered-list pattern in bullet list item export

### LINKED ISSUE #7824: Bug: backslash injection in numbered list edge case

Sathvik-Chowdary-Veerapaneni:
## Description

When a bullet or check list item starts with a number-dot-space pattern (e.g. `1. foo`), switching between WYSIWYG and markdown mode causes the exporter to double-escape the backslash: `1\. foo` becomes `1\\. foo` on each round-trip.

The root cause is in `$listExport`: the exporter outputs the list item text as-is without escaping the ordered-list pattern. On re-import, `unescapeText` correctly strips the backslash from `1\.`, but the exporter never re-adds it, causing the inline escape logic to double the existing backslash instead.

Fixed by escaping the dot when a non-number list item's text starts with `\d+\.\s`, producing `1\. foo` in the markdown output.

Closes #7824

### Before

```
- 1\. foo
```
After WYSIWYG round-trip becomes:
```
- 1\\. foo
```

### After

Stable round-trip — `- 1\. foo` stays `- 1\. foo`.

Added a unit test for bullet list items starting with ordered-list patterns.

All existing tests pass (329 markdown tests, 2742 total).


packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -346,6 +346,11 @@ describe('Markdown', () => {
       md: '* Level 1\n    * Level 2\n        * Level 3\n\nHello world',
       skipExport: true,
     },
+    {
+      // Bullet list item starting with number-dot pattern should be escaped (Regression #7824)
```
etrepum:
```suggestion
      // Bullet list item starting with number-dot pattern should be escaped (#7824)
```

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -351,7 +351,11 @@ const $listExport = (
           : listType === 'check'
             ? `${listMarker} [${listItemNode.getChecked() ? 'x' : ' '}] `
             : listMarker + ' ';
-      output.push(indent + prefix + exportChildren(listItemNode));
+      let childrenText = exportChildren(listItemNode);
+      if (listType !== 'number') {
+        childrenText = childrenText.replace(/^(\d+)(\.\s)/, '$1\\$2');
```
etrepum: It should allow for some spaces before the number as well

```suggestion
        childrenText = childrenText.replace(/^(\s{0,3}\d+)(\.\s)/, '$1\\$2');
```

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
Sathvik-Chowdary-Veerapaneni: Fixed.

packages/lexical-markdown/src/MarkdownTransformers.ts
Sathvik-Chowdary-Veerapaneni: Applied both — dropped the label and updated the regex to handle leading spaces.

etrepum (approved):

PR #8312, opened 2026-04-08
[lexical-offset] Bug Fix: Use child index instead of offset value for inline node selection

### LINKED ISSUE #7580: Bug: @lexical/offset makes invalid selection if offset points to inline node

Sathvik-Chowdary-Veerapaneni:
## Description

When `createSelectionFromOffsets` resolves an offset to an inline node (e.g. a DecoratorNode), it used the raw cumulative offset value as the element selection offset. This caused `IndexSizeError: There is no child at offset N` because the parent element might only have 2-3 children while the offset value could be much larger.

Fixed by using the inline node's actual index within its parent (`getIndexWithinParent()`) instead of the raw offset value.

Closes #7580

### Before

Setting an offset that lands on an inline decorator node throws:
```
IndexSizeError: Failed to execute 'setBaseAndExtent' on 'Selection': There is no child at offset 6.
```

### After

Selection is created correctly using the node's child index within its parent.

Added the first unit tests for the `@lexical/offset` package, covering inline node selection at both start and end boundaries.

All existing tests pass (2741 passed, 0 failed).

### DIFF
--- a/packages/lexical-offset/src/__tests__/unit/LexicalOffset.test.ts
+++ b/packages/lexical-offset/src/__tests__/unit/LexicalOffset.test.ts
@@ -98,10 +98,12 @@ describe('LexicalOffset', () => {
         expect(selection.anchor.type).toBe('element');
         const anchorNode = selection.anchor.getNode();
         expect($isElementNode(anchorNode)).toBe(true);
-        // Offset should be a valid child index, not the raw offset value
-        expect(selection.anchor.offset).toBeLessThanOrEqual(
-          anchorNode.getChildrenSize(),
-        );
+        if ($isElementNode(anchorNode)) {
+          // Offset should be a valid child index, not the raw offset value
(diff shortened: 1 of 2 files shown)


etrepum:
Fails ci-check

Sathvik-Chowdary-Veerapaneni:
CI is green now — added the ElementNode type guard.

etrepum (commented):
I think the correct thing to do with this module is to deprecate `OffsetView` and `$createOffsetView`. There are no tests, or documentation for it and it's not used anywhere else in the lexical repo. The best thing to do is throw it out and start over with something that has well known behavior.

PR #8313, opened 2026-04-08
[lexical-list] Bug Fix: Merge nested list into parent <li> during HTML export

### LINKED ISSUE #2951: Feature: Wrap nested lists in previous listitem instead of in a new listitem
**Lexical version:** _0.3.11_

### Current solution

Currently nested lists are added wrapped inside a new listitem:

```html
<ol>
   <li>one</li>
   <li>
      <ol>
         <li>one.one</li>
         <li>one.two</li>
      </ol>
   </li>
   <li>two</li>
</ol>
```

This makes things like prefixes for nested list (1.1, 1.2 [...]) way harder to achieve. We can´t easily make use of incrementing css counters and autoincrements won´t work correctly since there is another list item in the first level list. Actually we need to set `list-style-type:"none";`  on nested lists to get rid of multiple (and incorrect) counters. 

### Alternative solution
An alternative solution would be to wrap the nested list inside the previous listitem. 

```html
<ol>
   <li>one
      <ol>
         <li>one.one</li>
         <li>one.two</li>
      </ol>
   </li>
   <li>two</li>
</ol>
```

This would allow us to achieve things like prefixes way easier since we can just use css counters which won´t get mixed up by a wrong number of listitems. We can also make use of the autoincrement since there is no wrong number of listitems in the previous list level. In general the approach seems more appropriate to me.

### LINKED ISSUE #7207: Bug: Bug Report: Incorrect Numbering in Nested Lists using $generateHtmlFromNodes
When using $generateHtmlFromNodes in Lexical, nested ordered lists are incorrectly numbered. The issue occurs when a nested <ol> is wrapped inside a parent <li>, causing the numbering to break. Specifically, the parent list incorrectly assigns a duplicate number to an item after a nested list.

Lexical version: 0.22.0

## Steps To Reproduce

1. Create an ordered list with a nested list inside one of the items.
2. Convert it to HTML using $generateHtmlFromNodes.
3. Observe that numbering is incorrect, with a duplicate number appearing in the parent list.

Link to code example:

1. Canada  
2. USA  
   a. LA  
   b. TX  
   c. FL  
3. Germany  
4. Italy  

## The current behavior
<ol class="PlaygroundEditorTheme__ol1">
  <li value="1">canada</li>
  <li value="2">usa</li>
  <li value="3"> 
    <ol class="PlaygroundEditorTheme__ol2">
      <li value="1">LA</li>
      <li value="2">TX</li>
      <li value="3">FL</li>
    </ol>
  </li>
  <li value="3">Germany</li> 
  <li value="4">Italy</li>
</ol>

## The expected behavior
<ol class="PlaygroundEditorTheme__ol1">
  <li value="1">canada</li>
  <li value="2">usa
    <ol class="PlaygroundEditorTheme__ol2">
      <li value="1">LA</li>
      <li value="2">TX</li>
      <li value="3">FL</li>
    </ol>
  </li>
  <li value="3">Germany</li>
  <li value="4">Italy</li>
</ol>

Image

Sathvik-Chowdary-Veerapaneni:
## Description

When exporting nested lists with `$generateHtmlFromNodes`, the nested `<ol>`/`<ul>` was wrapped in a separate `<li>` element instead of being placed inside the parent `<li>`. This caused duplicate numbering in ordered lists:

```html
<!-- Before (broken): ghost <li> with duplicate value="3" -->
<li value="2">USA</li>
<li value="3"><ol>...</ol></li>
<li value="3">Germany</li>

<!-- After (fixed): nested list inside parent <li> -->
<li value="2">USA<ol>...</ol></li>
<li value="3">Germany</li>
```

In Lexical's model, a nested list is stored as a separate ListItemNode whose only child is a ListNode. During `exportDOM`, this wrapper now uses an `after` callback to move the nested list into the previous `<li>` sibling and remove the empty wrapper.

Closes #7207
Closes #2951

### Before

Nested ordered lists produce duplicate numbering due to an extra `<li>` wrapping the nested list.

### After

Nested lists are correctly placed inside the parent `<li>`, producing correct numbering.

Added a unit test for nested ordered list HTML export. Updated existing test expectations across 5 files to match the corrected HTML structure.

All existing tests pass (2740 passed, 0 failed).


etrepum:
Fails ci-check

Sathvik-Chowdary-Veerapaneni:
CI is green now — fixed the type check issue.

packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
```diff
@@ -37,6 +42,30 @@ describe('HTML', () => {
       },
       name: 'Empty editor state',
     },
+    {
+      // Regression #7207: nested list should be inside parent <li>, not a separate <li>
```
etrepum: Regression means that the behavior changed in an undesirable way, I don't think it worked like this previously

```suggestion
      // #7207: nested list should be inside parent <li>, not a separate <li>
```

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -182,6 +182,24 @@ export class ListItemNode extends ElementNode {
       element.dir = direction;
     }
 
+    if (isNestedListNode(this)) {
+      return {
+        after(containerElement) {
+          if (containerElement instanceof HTMLElement) {
+            const prevSibling = containerElement.previousElementSibling;
+            if (prevSibling) {
```
etrepum: Might make sense to check the type of this element to make sure it's doing the right thing

etrepum (commented):
This seems to work in the case where it's nested once but if there are multiple levels of nesting without interleaving content it looks like it's still creating the ul -> li -> ul -> li nesting

packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
Sathvik-Chowdary-Veerapaneni: Fixed the comment.

packages/lexical-list/src/LexicalListItemNode.ts
Sathvik-Chowdary-Veerapaneni: Added `instanceof HTMLLIElement` on prevSibling so it only merges into an actual list item — let me know if you had a different check in mind.

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -182,6 +182,24 @@ export class ListItemNode extends ElementNode {
       element.dir = direction;
     }
 
+    if (isNestedListNode(this)) {
+      return {
+        after(containerElement) {
+          if (containerElement instanceof HTMLElement) {
+            const prevSibling = containerElement.previousElementSibling;
+            if (prevSibling instanceof HTMLLIElement) {
```
etrepum: instanceof should never be used for HTML elements because it can fail when multiple documents are used (e.g. iframes). Use the `isHTMLElement` guard and check the nodeName or tagName.

etrepum (approved):

PR #8314, opened 2026-04-08
[Lexical-list] Bug Fix : Exit list on delete for empty non-first list items

### LINKED ISSUE #7577: Bug:  Backspace on Empty ListItem Does Not Convert to Paragraph (Word-like Behavior Broken)
Lexical version: @lexical/react@0.31.2

In Lexical, when pressing Backspace on an empty list item, the expected behavior (like in Microsoft Word or Google Docs) is to exit the list and create a new paragraph block.

**Instead, Lexical**:

Moves the caret to the previous list item.
This breaks the editing UX for any user trying to exit a list intuitively.

```
function BackspaceExitListPlugin() {

    useEffect(() => {
            KEY_BACKSPACE_COMMAND,
            () => {
                debugger;
                if (!$isRangeSelection(selection) || !selection.isCollapsed()) return false;

                const anchor = selection.anchor;
                let node = anchor.getNode();

                while (node && node.getParent() !== null && node.getType() !== 'listitem') {
                    node = node.getParentOrThrow();
                }

                if (node.getType() === 'listitem' && node.getTextContent().trim() === '' && anchor.offset === 0) {
                    editor.update(() => {
                        const text = $createTextNode('');
                        paragraph.append(text);
                        node.replace(paragraph);
                        paragraph.select(0, 0);
                    });

                    return true;
                }

                return false;
            },
            COMMAND_PRIORITY_LOW
        );
    }, [editor]);

    return null;
}

```
here is how I am registering it.
```

                <div className="editor-container">
                    <ToolbarPlugin />
                    <StyledToolbar />
                    <div className="editor-inner">
                        <RichTextPlugin
                            contentEditable={
                                <ContentEditable
                                    className="editor-input"
                                    aria-placeholder={placeholder}
                                    placeholder={
                                        <div className="editor-placeholder">{placeholder}</div>
                                    }
                                />
                            }
                        />
                        <HistoryPlugin />
                        <AutoFocusPlugin />
                        <BackspaceExitListPlugin />
                        <TreeViewPlugin />
                    </div>
                </div>
            </LexicalComposer>
```

Jynx2004:
## Description

Implements the standard word processor behavior (Microsoft Word, Google Docs) where pressing Delete on an empty non-first list item exits the list and creates a new paragraph block.

## Changes
- Added `DELETE_CHARACTER_COMMAND` handler in `registerList()` that:
  - Detects when cursor is at the start of an empty list item
  - Only triggers for non-first list items (preserves existing behavior for first items)
  - Removes the empty list item and creates a paragraph below the list
  - Moves cursor to the newly created paragraph
  - Uses `COMMAND_PRIORITY_LOW` to not interfere with other delete handlers

## Behavior
**Before:** Empty list item would remain in the list
**After:** Pressing Delete on an empty non-first list item exits the list and creates a paragraph

Closes #7577 

## Test plan
- Added comprehensive unit tests covering:
  - Main behavior: exits list and creates paragraph
  - Edge case: first empty list item doesn't exit
  - Edge case: list item with content doesn't exit
  - Edge case: removes list if it becomes empty


packages/lexical-list/src/index.ts
```diff
@@ -180,6 +258,23 @@ export function registerList(
...
+function listNodeHasOverflow(listNode: ListNode): boolean {
+  const stack: Array<LexicalNode> = [listNode];
+  while (stack.length > 0) {
+    const node = stack.pop();
+    if (!node) {
+      continue;
+    }
+    if (node.getType() === 'overflow') {
+      return true;
+    }
```
etrepum: It's *never* acceptable to use `getType()` in this way because it will be false any time there is a subclass of that node. It also does not make sense for lexical-list to depend on lexical-overflow to take the more correct approach of using `$isOverflowNode`. You'll have to find another approach for this.

packages/lexical-list/src/index.ts
```diff
@@ -133,6 +141,76 @@ export function registerList(
       },
       COMMAND_PRIORITY_LOW,
     ),
+    editor.registerCommand(
```
etrepum, line 145: This just seems like a lot of code for what it's doing

packages/lexical-list/src/index.ts
Jynx2004: Have modified the function
Let me know if any changes are needed
```
function listNodeHasOverflow(
  editor: LexicalEditor,
  listNode: ListNode,
): boolean {
  const overflowConfig = editor._nodes.get('overflow');
  if (!overflowConfig) {
    return false;
  }
  const overflowKlass = overflowConfig.klass;
  const stack: Array<LexicalNode> = [listNode];
  while (stack.length > 0) {
    const node = stack.pop()!;
    if (node instanceof overflowKlass) {
      return true;
    }
    if ($isElementNode(node)) {
      let child = (node as ElementNode).getLastChild();
      while (child !== null) {
        stack.push(child);
        child = child.getPreviousSibling();
      }
    }
  }
  return false;
}
```

packages/lexical-list/src/index.ts
Jynx2004: Here we are checking mainly all the conditions as this case only applies for empty lisitem which is not first in the total list items. We need to check these conditions for this particular edge case.

packages/lexical-list/src/index.ts
etrepum: It's still not ok to use these types or introduce this kind of weird cross-package dependency, we don't do that anywhere else.

Jynx2004: Will try to find a better solution for this

packages/lexical-list/src/index.ts
```diff
@@ -180,6 +259,32 @@ export function registerList(
...
+    if ($isElementNode(node)) {
+      let child = (node as ElementNode).getLastChild();
+      while (child !== null) {
+        stack.push(child);
+        child = child.getPreviousSibling();
+      }
+    }
+  }
+  return false;
+}
```
etrepum, line 286: This is still not acceptable for list to depend on overflow implementation details

etrepum:

PR #8315, opened 2026-04-08
[*] Fix: Replace legacy Flow syntax with modern equivalents in .flow …

potatowagon:
…type stubs

The upstream commit 1d349fe7c updated Flow syntax in source files, but the hand-maintained .flow type stubs were not updated to match. This causes Flow errors during www sync.

Replacements:
- `$ReadOnly<>` → `Readonly<>`
- `$ReadOnlyArray<>` → `ReadonlyArray<>`
- `$ReadOnlyMap<>` → `ReadonlyMap<>`
- `$NonMaybeType<>` → `NonNullable<>`
- `mixed` → `unknown`
- `<T: Bound>` → `<T extends Bound>`
- Remove `readonly` keyword from CommunityTeam.tsx interfaces

## Description

Closes #

## Test plan

same changes in [D99817603](https://www.internalfb.com/diff/D99817603) passes flow

### DIFF
--- a/packages/lexical-website/src/components/CommunityTeam.tsx
+++ b/packages/lexical-website/src/components/CommunityTeam.tsx
@@ -12,23 +12,23 @@ import teamData from '@site/src/data/team.json';
 import React from 'react';
 
 interface TeamMember {
-  readonly avatar: string;
-  readonly links: {
-    readonly github: string;
+  avatar: string;
+  links: {
+    github: string;
   };
-  readonly location?: string;
-  readonly name: string;
-  readonly org?: string;
-  readonly orgLink?: string;
-  readonly sponsor?: string;
-  readonly title?: string;
-  readonly username: string;
+  location?: string;
+  name: string;
+  org?: string;
+  orgLink?: string;
+  sponsor?: string;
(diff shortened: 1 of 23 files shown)


etrepum (approved):

etrepum:
Big fan of the new almost-typescript syntax, makes things much easier to maintain! I bet an LLM could be prompted to do it at this point

packages/lexical-website/src/components/CommunityTeam.tsx
```diff
@@ -12,23 +12,23 @@ import teamData from '@site/src/data/team.json';
...
+  links: {
+    github: string;
   };
-  readonly location?: string;
-  readonly name: string;
-  readonly org?: string;
-  readonly orgLink?: string;
-  readonly sponsor?: string;
-  readonly title?: string;
-  readonly username: string;
```
etrepum, line 25: Mostly just curious, but since this isn't flow and the data is only used in a readonly manner, what is the reason that these are changing? It's coming from an imported json file so the rationale for readonly here was to treat it as a frozen constant.

PR #8317, opened 2026-04-09
[lexical-table] Bug Fix: Read background-color from shorthand style a…

tomsarduy:
## Description

  `$convertTableCellNodeElement` only read `style.backgroundColor` when importing table cells from HTML, which misses the two most common ways Excel and some email encode cell background colors:

  - `style="background:#F4B084"` — the CSS `background` shorthand. Browsers do compute this into `style.backgroundColor` on a live DOM, but on detached/fragment DOMs (as used by Lexical's paste engine) some parsers leave `style.backgroundColor` empty.
  - `bgcolor="#F4B084"` — the legacy HTML attribute used by older Excel/Outlook versions.

  The fix reads all three sources in priority order: `style.backgroundColor || style.background ||
  getAttribute('bgcolor')`.

  ### Before

  Pasting a table from Excel or some email clients with background color cells would lose all cell background colors. We use the editor in a ticketing system so there is a lot of copy/paste from Excel

  ### After

  - `style.backgroundColor` (explicit `background-color:`) — preserved ✓
  - `style.background` shorthand — preserved ✓
  - `bgcolor` HTML attribute — preserved ✓
  - Unit tests added covering all four cases (including no background color → `null`)

### DIFF
--- a/packages/lexical-table/src/LexicalTableCellNode.ts
+++ b/packages/lexical-table/src/LexicalTableCellNode.ts
@@ -379,8 +379,12 @@ export function $convertTableCellNodeElement(
   );
 
   tableCellNode.__rowSpan = domNode_.rowSpan;
-  const backgroundColor = domNode_.style.backgroundColor;
-  if (backgroundColor !== '') {
+  const backgroundColor =
+    domNode_.style.backgroundColor ||
+    domNode_.style.background ||
+    domNode_.getAttribute('bgcolor') ||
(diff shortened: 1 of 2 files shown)


hanselabreu:
Nice!

etrepum:
Do you intend to finish this one? It's marked as draft, although other than the conflicts it appears to be fairly complete

etrepum:

PR #8322, opened 2026-04-09
[lexical-playground] Feature: Add PagesExtension

ibastawisi:
## Description
This pull request introduces a comprehensive "Pages" feature to the Lexical Playground, adding support for document pagination, page setup, and print-friendly styling. The main changes include new node types for representing pages and their content, a page setup node for configuring size and margins, and significant CSS updates for print output and UI enhancements.

**Pages feature and node architecture:**

* Added new node types: `PageNode`, `PageContentNode`, and `PageSetupNode`, along with supporting types and constants, to represent document pages, their content, and page setup (size, orientation, margins). These nodes manage pagination logic, page measurement, overflow/underflow handling, and serialization. [[1]](diffhunk://#diff-ec402a463b931b8f634de6de4bdea12d74245ca4e002d6f88eddd50be2e26b08R1-R83) [[2]](diffhunk://#diff-074b22f72790062fce7b26a81cf0ceabfd66671ca1ea0e7ad00e8cde3b23b4a8R1-R310) [[3]](diffhunk://#diff-d434b51c310f5d4a3b630aaae9df42c51047c859f8ac67d009cf203305cce950R1-R224) [[4]](diffhunk://#diff-b105f3dbecad6efd65a1633ef5384b7666b3f5ff70d1e17159c084f7f1942dd5R1-R36) [[5]](diffhunk://#diff-efdea61a4836c0b0a194ef98ceced87df4c7f8e1585477b13d6c811e897d0531R1-R32) [[6]](diffhunk://#diff-0144f34152891bd2faac59efa545bdb1afc302f64d7b4b3e930b3b9af9a8f790R1-R12)
* Integrated the `PagesExtension` into the main playground app, enabling the new pages functionality in the editor. [[1]](diffhunk://#diff-97349b8b721c50fd5c6545f97d0f9f048b0d11a5f533eae37b904fa3879e6fbfR63) [[2]](diffhunk://#diff-97349b8b721c50fd5c6545f97d0f9f048b0d11a5f533eae37b904fa3879e6fbfR178)

**Print and UI enhancements:**

* Added print-specific CSS rules to ensure correct page rendering, hide non-editor elements, and optimize color and layout for printing. [[1]](diffhunk://#diff-82283c8e0d58ea86e246863ebea2a966f073355697ff35c07c0b49ac3fe6615fR98-R143) [[2]](diffhunk://#diff-70c22fab78b3818db22e95c4d01c3f91fe8f8c82a8fa82106da588ae6920ba95R65-R70)
* Updated editor and dropdown styling, including new icons for page setup and improved submenu chevrons, and slightly increased the editor shell's maximum width. [[1]](diffhunk://#diff-82283c8e0d58ea86e246863ebea2a966f073355697ff35c07c0b49ac3fe6615fL55-R55) [[2]](diffhunk://#diff-82283c8e0d58ea86e246863ebea2a966f073355697ff35c07c0b49ac3fe6615fR658-R662) [[3]](diffhunk://#diff-82283c8e0d58ea86e246863ebea2a966f073355697ff35c07c0b49ac3fe6615fR883-R897)

These changes lay the foundation for multi-page document editing and WYSIWYG print output in the Lexical Playground.

image


packages/lexical-playground/src/plugins/PagesExtension/PageContentNode.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+import type {LexicalNode, SerializedElementNode} from 'lexical';
+
+import {ElementNode} from 'lexical';
+
+import {$isPageNode, PageNode} from './PageNode';
+
+export type SerializedPageContentNode = SerializedElementNode;
+
+export class PageContentNode extends ElementNode {
```
etrepum, line 21: Would be good to use the new $config API for this https://lexical.dev/docs/concepts/nodes#creating-custom-nodes-with-config-and-nodestate

packages/lexical-playground/src/plugins/PagesExtension/PageNode.ts
```diff
@@ -0,0 +1,310 @@
...
+  $isPageContentNode,
+  PageContentNode,
+} from './PageContentNode';
+
+export type SerializedPageNode = SerializedElementNode;
+
+const pagesMarkedForMeasurement = new Set<string>();
+const fixedPageHeights = new Map<string, number>();
+
+export class PageNode extends ElementNode {
```
etrepum, line 25: Use new $config API to reduce boilerplate https://lexical.dev/docs/concepts/nodes#creating-custom-nodes-with-config-and-nodestate

packages/lexical-playground/src/nodes/PageNode/PageNode.ts
```diff
@@ -0,0 +1,310 @@
...
+    fixedPageHeights.clear();
+  }
+
+  static markForMeasurement(nodeKey: string): void {
+    pagesMarkedForMeasurement.add(nodeKey);
+  }
+
+  static clearMeasurementFlag(nodeKey: string): void {
+    pagesMarkedForMeasurement.delete(nodeKey);
+  }
```
etrepum: I think we can store this data in the extension scoped to the editor, rather than as module globals?

packages/lexical-playground/src/nodes/PageNode/PageSetupNode.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+export class PageSetupNode extends DecoratorNode<null> {
+  __pageSize: PageSize;
+  __orientation: Orientation;
+  __margins: {
+    top: number;
+    right: number;
+    bottom: number;
+    left: number;
+  };
```
etrepum: Using the new $config API and NodeState would reduce this boilerplate https://lexical.dev/docs/concepts/nodes#creating-custom-nodes-with-config-and-nodestate

packages/lexical-playground/src/plugins/PagesExtension/PageSetupDropdown.tsx
```diff
@@ -0,0 +1,239 @@
...
+      return node.getPageSetup();
+    });
+    setPageSetup(newPageSetup);
+    setOrientationMenuOpen(newPageSetup !== null);
+    setMarginsMenuOpen(newPageSetup !== null);
+  }, [editor]);
+  useEffect(() => {
+    updatePageSetup();
+    editor.registerMutationListener(PageSetupNode, updatePageSetup);
+  }, [editor, updatePageSetup]);
```
etrepum: Probably some of this should live in the extension rather than the component?

packages/lexical-playground/src/plugins/PagesExtension/PagesPlugin.tsx
```diff
@@ -0,0 +1,565 @@
...
+  PageNode,
+  PageSetupNode,
+} from '../../nodes/PageNode';
+import {
+  $createPageContentNode,
+  $isPageContentNode,
+  PageContentNode,
+} from '../../nodes/PageNode/PageContentNode';
+
+export function PagesPlugin(): null {
```
etrepum: Nothing in here is dependent on react, so would be better to just make it part of the extension directly

etrepum (commented):
This is probably one of the most requested features and generally the trickiest to build! I haven't had time to do a close inspection yet, but here are some general notes on the architecture. Basically it could be made a lot less react-dependent (everything but the dropdown UI itself I think) and follow newer lexical conventions more closely to reduce boilerplate required for clone/serialization/etc.

packages/lexical-playground/src/nodes/PageNode/PageSetupNode.tsx
ibastawisi: The sole purpose of page setup node is just to store page setup state.. it’s excluded from copy and renders null.
> You can even add node state to the RootNode to store document-level metadata, which wasn't possible at all before!

I wonder if it’s better to get rid of it and store the page setup state on the root node?

packages/lexical-playground/src/plugins/PagesExtension/PageSetupDropdown.tsx
ibastawisi: These are just to sync the dropdown component state, I guess the names are rather confusing!

packages/lexical-playground/src/plugins/PagesExtension/PagesPlugin.tsx
ibastawisi: You can tell this code is from a pre-extensions era and needs a bit of touch up 😄

packages/lexical-playground/src/nodes/PageNode/PageNode.ts
ibastawisi: Can you share an example of this, I'm just starting to use extensions and there’s a lot to read!

packages/lexical-playground/src/plugins/PagesExtension/PageSetupDropdown.tsx
ibastawisi: Again I'm not sharp on extensions and signals, sharing an example would tremendously help!

packages/lexical-playground/src/nodes/PageNode/PageSetupNode.tsx
etrepum: I think that works a lot better!

etrepum (commented):
I went ahead and committed the sort of refactorings I was thinking of to use the $config and extension APIs. The RootNode state instead of PageSetupNode make a lot of sense so that is done too.

Using `root.splice` instead of `root.clear();root.append(…)` was to fix a selection issue when creating a PageBreakNode.

Since it does somewhat depend on PageBreakNode I went ahead and ported that to an extension and added it as a dependency.

etrepum:
The part that I think could probably be cleaned up is to encapsulate the measurement logic out of the nodes themselves and put that somewhere else (and it could likely be pulled out of the extension output, that was just a stopgap for a quick refactor). It's not really an API that should be called, it's an implementation detail, so having them as methods of the nodes is a bit unwieldy.

ibastawisi:
One thing missing from the measurement logic is splitting text nodes instead of moving last/first paragraph node to next/previous page, probably [pretext](https://github.com/chenglou/pretext) can help with that.
I should look into this tomorrow, thanks for your help 🙏

packages/lexical-playground/src/plugins/PagesExtension/PageSetupDropdown.tsx
```diff
@@ -94,31 +93,22 @@ export function PageSetupDropdownComponent({
...
-    });
+    const newPageSetup = editor.getEditorState().read($getPageSetup);
     setPageSetup(newPageSetup);
     setOrientationMenuOpen(newPageSetup !== null);
     setMarginsMenuOpen(newPageSetup !== null);
   }, [editor]);
   useEffect(() => {
     updatePageSetup();
-    editor.registerMutationListener(PageSetupNode, updatePageSetup);
+    return editor.registerMutationListener(RootNode, updatePageSetup);
```
ibastawisi: maybe we should add a PAGE_SETUP_TAG and return if not found?

packages/lexical-playground/src/plugins/PagesExtension/index.ts
```diff
@@ -0,0 +1,581 @@
...
+
+        const removeCommandListeners = mergeRegister(
+          editor.registerCommand(
+            SELECTION_CHANGE_COMMAND,
+            () => {
+              isTouched = true;
+              const selection = $getSelection();
+              if (!$isRangeSelection(selection)) return false;
+              const anchorNode = selection.anchor.getNode();
+              if ($isRootNode(anchorNode)) {
```
ibastawisi: I think this was to move the PageSetupNode outside the range selection and is no longer needed.

packages/lexical-playground/src/plugins/PagesExtension/PageSetupDropdown.tsx
etrepum: I don't think that would help anything here? It's tapping into the update cycle either way and the bulk of the code only runs when the pageSetupState has changed from one editor state to the next

etrepum: Will look more closely later, afk for a bit

packages/lexical-playground/src/plugins/PagesExtension/index.ts
etrepum: Yeah it does look like some of this can be removed and it might make sense for it to be an update listener instead of `SELECTION_CHANGE_COMMAND` which can fire a few times (and is delegated up from nested editors)

packages/lexical-playground/src/plugins/PagesExtension/PageSetupDropdown.tsx
etrepum: This comment was for a previous version from before it was using a signal so no longer relevant

ibastawisi: Yeah I found out later, the idea was to put an update tag to know when page setup changes to limit running everything again since any mutation is a root mutation.. but I remembered the tag approach wouldn’t work well with history and collaboration..
if the cost of this is low then no need to optimize further.
```ts
 const change = $getStateChange(
    editor.getEditorState().read($getRoot),
    prevEditorState.read($getRoot),
    pageSetupState,
  );
  if (!change) {
    return;
  }
```

etrepum (approved):
I think this looks great, of course the one thing missing here would be adding some e2e tests to show that it works as expected but I think it's also early enough and isolated enough from other playground functionality that we could do that later (e.g. after implementing better TextNode splitting)

ibastawisi:
I'm having a hard time getting text splitting to work as expected, I will try a different approach later today but things look harder than I thought.
If I couldn’t figure it out today, then we'd better open an issue and think about it in another day

packages/lexical-playground/src/plugins/PagesExtension/PagesExtension.ts
```diff
@@ -0,0 +1,755 @@
...
+  pageContentClass: string;
+  pageClass: string;
+}
+
+export const PagesExtension = defineExtension({
+  build: (editor) => {
+    const getPageSetup = () => editor.getEditorState().read($getPageSetup);
+
+    return {
+      Component: PageSetupDropdownComponent,
```
ibastawisi: I think I left this here by mistake

etrepum (approved):
I think we're good to merge this and any improvements to the TextNode splitting or anything else can be addressed in follow-ups!

PR #8326, opened 2026-04-10
[lexical-html][lexical-table] Feature: Inline CSS from <style> tags

tomsarduy:
Address facebook#8318

Apps like Excel and Outlook generate HTML where styles live in <style> class rules (e.g. `.xl65 { background: #FFFF00; color: blue; }`) rather than inline styles. Lexical's import converters only read inline styles, so all class-based formatting was silently lost on paste.

Add `inlineStylesFromStyleSheets()` as a preprocessing step in `$generateNodesFromDOM` that resolves CSS rules from `<style>` tags into inline styles on matching elements before any converters run. This is generic — all existing converters automatically benefit for any CSS property (background-color, color, font-weight, font-size, margins, etc.).

Also extend the table cell import converter's `after` callback to propagate `color` from `<td>` elements to child TextNodes, matching the existing pattern for bold/italic/underline/strikethrough. This is needed because CSS color on a <td> is inherited by child text, but Lexical models text color on TextNode, not on the cell — so the converter must bridge this gap explicitly.

### Pasted table from Excel

| Before | After |
|--------|-------|
| Before | After |

Note: This just parse `<style>` definitions, and I agree with the comment from @etrepum, to support MS Word for example, need a normalise step (something the [outlook client](https://github.com/microsoft/roosterjs) does very well btw) but that's a big task.


packages/lexical-table/src/LexicalTableCellNode.ts
```diff
@@ -431,6 +432,12 @@ export function $convertTableCellNodeElement(
             if (hasUnderlineTextDecoration) {
               child.toggleFormat('underline');
             }
+            if (color) {
+              const existingStyle = child.getStyle();
+              if (!existingStyle.includes('color:')) {
```
tomsarduy, line 437: Propagate color from `<td>` to child TextNodes

packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
```diff
@@ -262,4 +262,66 @@ describe('HTML', () => {
...
+        () => {
+          const root = $getRoot();
+          const dom = parser.parseFromString(
+            `<html><head><style>.highlight { font-weight: bold; }</style></head>` +
+              `<body><p><span class="highlight">Hello</span></p></body></html>`,
+            'text/html',
+          );
+          const nodes = $generateNodesFromDOM(editor, dom);
+          root.append(...nodes);
+          expect(root.getTextContent()).toBe('Hello');
```
etrepum, line 283: This doesn't test that the text is bold?

etrepum (commented):
Overall I think this looks good and is a worthwhile approach for the current importDOM API, but it doesn't appear to test the functionality it's adding. I only see tests that check to make sure it doesn't crash, not that the styles are actually propagated in a way that lexical imports correctly.

etrepum (approved):

PR #8329, opened 2026-04-10
[lexical-react] Bug Fix: Prevent error when editorRef is null

### LINKED ISSUE #8327: Bug: LexicalEditorRefPlugin crashes if input is `null`
Lexical version: 0.42.0 (but the bug is detected in the main branch as well)

## Steps To Reproduce

1. Mount Lexical with EditorRefPlugin attached, but without actually providing a ref
2. Error occurs on [this line](https://github.com/facebook/lexical/blob/2e7d86cd6ad956e39c198fb23f519bbb8a85f3ad/packages/lexical-react/src/LexicalEditorRefPlugin.tsx#L33)

Link to code example:

```
(function ReactComponent(props) {
const ref = null;
// react component with lexical
        <EditorRefPlugin editorRef={ref}/> // where ref is `null`
});
```

## The current behavior
While the typescript docs mention [null as input parameter](https://github.com/facebook/lexical/blob/2e7d86cd6ad956e39c198fb23f519bbb8a85f3ad/packages/lexical-react/src/LexicalEditorRefPlugin.tsx#L26)
the `typeof editorRef === "object"` is missing a `&& editorRef !== null` part.

## The expected behavior
No crash on `editorRef.current = editor;` when editorRef === null

## Impact of fix

Adding the null check fixes everything ;)
My current workaround
` <EditorRefPlugin editorRef={ref || undefined }/>`

Jimmy89:
## Description

editorRef is allowed to be null and must be checked on this type before value allocation. This PR adds that check.

Closes #8327

## Test plan
Add a LexicalEditor with a LexicalEditorRefPlugin where editorRef is value `null`. 

Old situation: error
New situation: line of error skipped

### DIFF
--- a/packages/lexical-react/src/LexicalEditorRefPlugin.tsx
+++ b/packages/lexical-react/src/LexicalEditorRefPlugin.tsx
@@ -30,7 +30,7 @@ export function EditorRefPlugin({
   React.useEffect(() => {
     if (typeof editorRef === 'function') {
       editorRef(editor);
-    } else if (typeof editorRef === 'object') {
+    } else if (typeof editorRef === 'object' && editorRef !== null) {
       editorRef.current = editor;
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps


etrepum (approved):
It looks like editorRef is not allowed to be null according to TypeScript but being a little more defensive here doesn't hurt

PR #8331, opened 2026-04-11
[lexical-ai] Add AI-native editor plugin with provider-agnostic agent loop

automanfromm87:
## Summary

Introduces `@lexical/ai` — a new package that turns any Lexical editor into an AI-native editor. The package is fully **provider-agnostic**: it defines an `AIAdapter` interface that any LLM (Claude, GPT, local models, etc.) can implement.

### Commit Stack

1. **Types, interfaces, and commands** — `AIAdapter`, `AIStreamEvent`, `AgentEvent`, Lexical commands (`AI_EXECUTE_COMMAND`, `AI_CANCEL_COMMAND`, `AI_STATUS_COMMAND`)
2. **Editor tool system** — 10 built-in tools (`get_document`, `replace_text`, `format_text`, `insert_link`, etc.) with text-based node targeting via `$findNodeContaining`
3. **Agent loop, plugin, and React hooks** — ReAct agent loop with sliding window context management (128K token budget), headless `LexicalAINativePlugin`, `useAIAgent` / `useAIAgentState` hooks, and barrel export
4. **Playground integration** — `AnthropicAdapter`, `MockAIAdapter` for testing, chat panel UI, env-var-driven Vite proxy

### Architecture

```
AIAdapter (provider-agnostic interface)
    ↓
AgentLoop (ReAct: stream → tool calls → execute → loop)
    ↓
ToolExecutor (10 built-in editor operations)
    ↓
LexicalEditor (editor.update / editor.read)
```

**Context Management:** Sliding window strategy compacts older conversation turns into structured summaries while preserving system prompt, tool definitions, and recent turns in full. Large tool result payloads (e.g. full document JSON) are automatically stripped from history.

**Extensibility:** Custom tools can be registered via the `customTools` prop with a definition + executor function.

### Usage

```tsx
<LexicalComposer initialConfig={config}>
  <RichTextPlugin ... />
  <LexicalAINativePlugin
    adapter={new AnthropicAdapter({apiKey: '...'})}
    onAgentEvent={(event) => console.log(event)}
  />
</LexicalComposer>
```

## Test plan

- [x] `pnpm run tsc` passes at every commit in the stack
- [x] `pnpm run lint` passes (enforced by pre-commit hooks)
- [ ] `pnpm run dev` — playground starts, AI panel renders
- [ ] MockAIAdapter: verify tool call flow without network
- [ ] AnthropicAdapter: verify end-to-end with `VITE_AI_PROXY_TARGET`


etrepum (commented):
I don't think this is in the stage that it should be a new public package, this is the kind of project that should evolve for a while based on usage and feedback before it becomes a public API. It would make sense to run this kind of experiment self-contained in an example or the playground.

It's also not taking advantage of the latest lexical APIs, it should be exposed as an extension instead of a legacy plugin. An extension could be designed to let other extensions expose custom tools automatically as peer dependency configuration, for example.

automanfromm87:
> I don't think this is in the stage that it should be a new public package, this is the kind of project that should evolve for a while based on usage and feedback before it becomes a public API. It would make sense to run this kind of experiment self-contained in an example or the playground.
> 
> It's also not taking advantage of the latest lexical APIs, it should be exposed as an extension instead of a legacy plugin. An extension could be designed to let other extensions expose custom tools automatically as peer dependency configuration, for example.

thanks for suggesting, let me think about it first based on your proposal,

etrepum:

PR #8344, opened 2026-04-11
[lexical-markdown] Bug Fix: prevent delimiter collision when exporting adjacent text-format nodes

### LINKED ISSUE #7974: Bug: A combination of bold and bold-italic text on the same line is incorrectly converted to Markdown and back.
I use the Lexical Editor, which exports and imports Markdown using built-in tools. The bug can be reproduced in the lexical/playground sandbox; detailed steps are below

Lexical version: 0.38.2

## Steps To Reproduce

1. Open the `lexical/playground` project and the `Editor.tsx` file
2. Add imports `$convertToMarkdownString` and `TRANSFORMERS` from `@lexical/markdown`
3. Add import `OnChangePlugin` from `@lexical/react/LexicalOnChangePlugin`
4. Add `<OnChangePlugin onChange={onChange} />` to editor code and create the next `onChange`  handler:

```
  const onChange = (editorState) => {
    const markdown = editorState.read(() =>
      $convertToMarkdownString(TRANSFORMERS)
    )
    console.log(markdown)
  }
```

5. Run the project, clear editor input and type next subsequence: press `Bold` => type '123' => press `Italic` => type '456' (Result: **123*456*** - a string without spaces, the first three characters are bold, the next three are bold and italicized). Log and copy converted to markdown result - should be `**123*456***` string
6. Go to `App.tsx` and initialize the editor with a value converted from this string:

```
      defineExtension({
        $initialEditorState: () => $convertFromMarkdownString('**123*456***', TRANSFORMERS)
```

7. Check the results

## The current behavior
The editor displays a distorted string `123*456*` (See screenshot):

Image

## The expected behavior
The editor displays the same string we entered in step 5: **123*456***

Yahid-Basha:
### Current behavior

When a paragraph contains adjacent `TextNode`s whose inline format bitmasks share
asterisk-based delimiters — for example, a **bold** node immediately followed by a
**bold+italic** node — `$convertToMarkdownString` naively concatenates their wrapped
strings without checking whether the closing delimiter of one node and the opening
delimiter of the next would merge into an ambiguous asterisk run.

This produces invalid Markdown on export:

| Editor state | Export output |
|---|---|
| `TextNode{bold, "123"}` + `TextNode{bold+italic, "456"}` | `**123*456***` ⚠️ (ambiguous) |

While major CommonMark-compliant parsers handle these sequences correctly through delimiter run leniency, the output is technically non-standard. CommonMark specifies that emphasis delimiters should be explicit and unambiguous. The current export relies on parser leniency rather than producing well-formed output, which creates a portability risk for applications that use stricter or non-standard parsers, and produces output that is harder to read and reason about in raw Markdown.

This bug affects any application that persists Lexical content as Markdown and
re-imports it (CMS pipelines, save/reload flows, copy-paste between editors). It has
been independently reported by multiple downstream projects including
[stacker.news](https://github.com/stackernews/stacker.news/pull/2501) and
[sveltia-cms](https://github.com/sveltia/sveltia-cms/issues/625), both of which are
shipping with this as a known unresolved upstream limitation.

This is the export-side complement of #5758, which fixed the import side in v0.22.0.
The export side was never addressed.

Closes #7974

---

### Changes in this PR

**`MarkdownExport.ts` — two-part fix:**

1. **Boundary space insertion (export loop):** After computing each `TextNode`'s
   formatted string, the export loop now checks whether the previous output ends with
   `*` or `_` and the new string starts with the same character. If so, a single space
   is inserted between them to prevent delimiter merging:

   ```
   bold "123" + bold-italic "456"  →  **123** ***456***  ✅
   ```

2. **Force-close on ambiguous shared tags (`exportTextFormat`):** When two adjacent
   nodes share a format tag (e.g. both are italic) but differ by a single-character
   tag (e.g. one is also bold), the shared tag is normally kept open across the
   boundary to produce tighter output. This PR detects when that kept-open tag would
   create an ambiguous delimiter sequence and force-closes it at the boundary instead,
   emitting a clean close+reopen pair.

   The check is gated on the absence of whitespace between node texts — if there is
   already whitespace at the boundary, no collision is possible and the existing
   behaviour is preserved.

**What is NOT changed:**

- Same-format adjacent nodes still merge into a single span with no extra space
  (`**abcdef**`, not `**abc** **def**`)
- Nodes separated by existing whitespace are unaffected
- The fix is scoped to `*` and `_` delimiter characters only
- The more complex overlapping/crossing format case (e.g. bold → bold+italic → italic
  → plain producing `he**llo*wor**ld*!`) is tracked separately in #4895 and is not
  addressed here

---

## Test plan

10 new unit tests added in `LexicalMarkdown.test.ts` under
`describe('markdown adjacent format boundary export (#7974)')`.

### Before

```
bold "123" + bold-italic "456"  →  **123*456***
bold-italic "123" + bold "456"  →  ***123*456**   (both wrong, formatting lost on re-import)
```

### After

| Scenario | Output |
|---|---|
| Bold → Bold+Italic | `**123** ***456***` |
| Bold+Italic → Bold | `***123*** **456**` |
| Italic → Bold+Italic (shared tag, no collision) | `*123**456***` |
| Bold+Italic → Italic (force-close bold) | `***123*** *456*` |
| Same-format adjacent (no change) | `**abcdef**` |
| Bold → Bold+Italic → Bold (two boundaries) | `**123** ***456*** **789**` |
| Trailing whitespace disables force-close | `*Hello **world***` |
| Roundtrip stability: export → import → re-export | identical output |
| Single-format nodes unaffected (regression check) | `**bold**`, `*italic*`, `***bolditalic***` |

All 10 new tests pass. All 2753 existing tests continue to pass.

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -184,16 +184,30 @@ function exportChildren(
     if ($isLineBreakNode(child)) {
       output.push('\n');
     } else if ($isTextNode(child)) {
-      output.push(
-        exportTextFormat(
-          child,
-          child.getTextContent(),
-          textTransformersIndex,
-          unclosedTags,
-          unclosableTags,
-          shouldPreserveNewLines,
-        ),
+      const textString = exportTextFormat(
+        child,
+        child.getTextContent(),
+        textTransformersIndex,
+        unclosedTags,
+        unclosableTags,
+        shouldPreserveNewLines,
       );
+
+      // When adjacent TextNodes use overlapping asterisk (or underscore) formatting,
+      // their closing and opening delimiters would merge into an ambiguous run.
+      // e.g. bold "123" + bold-italic "456" naively produces **123*456*** instead of
+      // **123** ***456***. Insert a space to prevent delimiter collision.
+      const lastOutput = output.length > 0 ? output[output.length - 1] : '';
+      if (
+        lastOutput.length > 0 &&
+        textString.length > 0 &&
+        ((lastOutput.endsWith('*') && textString.startsWith('*')) ||
+          (lastOutput.endsWith('_') && textString.startsWith('_')))
+      ) {
+        output.push(' ');
+      }
+
+      output.push(textString);
     } else if ($isElementNode(child)) {
       // empty paragraph returns ""
       output.push(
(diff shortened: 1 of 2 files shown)


etrepum (commented):
This just seems incorrect, I don't see any problem with `**123*456*** ` in lexical's parser (as configured in the playground: [link](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE-VTwU7DMAz9l3ecwtQVyrR8xaRxQxxCEtqINJ5SZ1BN_XfkIUG1Y0HiwCn2c_zes5Wc4V1gygc27KHPyEQsp-1CdNkn6MerxHk2IUJXCi-Ue8OXsCfnoZEEiVAYeIwCQIH9O0NjtZJ4PApqyfmbLrRdDG3HUDj5PARK0BuFL_zh8_KxJMvFsNQntdzAM0W30IKlxD7xj-T_evzAJgb7jxfwOw_wScGF7O0l06nE-G1DxEJysiixNNO6UogmtcW0Uu1NfnX0toj58lnnzNMk3AMfzMk76M122-zqXd3UTXWvMFDJVvr20YxtppLcvBvV-u52XWH6AIT4ugEVBAAA)) or the [commonmark dingus](https://spec.commonmark.org/dingus/?text=**123*456***%20%0A) or github markdown.

Example of `**123*456*** ` in github: **123*456*** 

How about instead of trying to fix something that might be working, you scope this down to a failing test and then we can take a closer look.

Yahid-Basha:
You're right, `**123*456***` is valid and I've verified it on CommonMark dingus and GitHub. I found a genuinely broken case while testing — $convertToMarkdownString silently drops the bold closing tag when a bold+italic whitespace node separates a bold node from an italic node, producing `**Hello *World*` instead of `**Hello** *World*`. However my current implementation doesn't fix this case. I'll close this PR and open a focused one targeting the actual failing case once I have a working fix for it.

etrepum:
Sounds great, I am sure there are failing cases, but the manner in which this PR tried to fix them looked like it created different regressions (creating spaces where none exist in the document) instead of making it behave correctly.

Yahid-Basha:
That's a fair point — inserting a space at delimiter boundaries changes the content. I'll come back with a focused fix for that once I have an approach that doesn't introduce content changes.

I appreciate the quick review.

PR #8349, opened 2026-04-12
[lexical] Bug Fix: Resolve Firefox arrow key navigation issue in collapsible nodes (#8348)

### LINKED ISSUE #8348: Bug: Firefox regression in up/down arrow key handling with collapsible
Lexical version: 0.43.0

## Steps To Reproduce

1. Create a collapsible
2. Try and navigate from the title to the content using the down arrow key

In firefox it will skip over the content to the paragraph after the collapsible, all other browsers behave reasonably and go from one block to the next.

This is a firefox quirk, not specifically lexical related, you can reproduce with something as simple as this [jsbin](https://jsbin.com/humigeqayi/1/edit?html,output):

```html
<div contenteditable="true" style="user-select: text;">
  <p>before</p>
  <details open="open">
    <summary>
      <p>a</p>
    </summary>
    <div>
      <p>b</p>
    </div>
  </details>
  <p>after</p>
</div>
```

Ideally we'd have a way to workaround this with just CSS, or maybe we stop using the details element in firefox? Worst case we will have to handle the up/down navigation ourselves and determine if a collapsible section was skipped.

## The current behavior

Arrow down skips over the content and to the next paragraph
Arrow up from the after paragraph skips over the collapsible entirely

## The expected behavior

Arrow keys navigate from block to block as they do in chrome and safari

MohmedIkram:
## Description

This PR fixes a Firefox-specific issue where arrow key navigation skips over collapsible content inside `<details>` elements.

### Current Behavior
- In Firefox, pressing the down arrow from the collapsible title (`<summary>`) skips the content inside the collapsible.
- Similarly, arrow up from the next paragraph skips the entire collapsible block.
- This results in inconsistent navigation compared to Chrome and Safari.

### Changes Made
- Implemented a fix to ensure arrow key navigation correctly moves between blocks inside collapsible nodes in Firefox.
- Added handling to prevent skipping of collapsible content during keyboard navigation.

### Result
- Arrow key navigation now behaves consistently across browsers.
- Users can navigate into and out of collapsible content as expected.

### Before

https://github.com/user-attachments/assets/5d8d26d8-f5f5-4e82-b96f-7557fbc8566d

### After

https://github.com/user-attachments/assets/7ab5f207-9a54-4933-af99-149e3cbc04d2

Closes #8348

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/Selection.spec.mjs
@@ -621,7 +621,10 @@ test.describe.parallel('Selection', () => {
     await moveRight(page, 3);
     await deleteForward(page);
 
-    const collapsibleTag = browserName === 'chromium' ? 'div' : 'details';
+    const collapsibleTag =
+      browserName === 'chromium' || browserName === 'firefox'
+        ? 'div'
+        : 'details';
     await assertHTML(
       page,
       html`
@@ -665,7 +668,10 @@ test.describe.parallel('Selection', () => {
     await moveLeft(page, 'after'.length);
     await deleteBackward(page);
 
-    const collapsibleTag = browserName === 'chromium' ? 'div' : 'details';
+    const collapsibleTag =
+      browserName === 'chromium' || browserName === 'firefox'
+        ? 'div'
+        : 'details';
     await assertHTML(
       page,
       html`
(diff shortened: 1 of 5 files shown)


etrepum (commented):
It's important to make sure that this behavior is tested. I don't see any changes to the relevant e2e tests, which will probably be in a failing state, but there should also be new test code to ensure the up and down arrows work as expected in all browsers with this node.

etrepum (approved):

PR #8360, opened 2026-04-15
[Breaking Change][lexical-code] Bug Fix: move code block escape logic to KEY_ENTER_COMMAND listener

### LINKED ISSUE #8351: Bug: Pasting text with consecutive blank lines into a code block results in text outside of the code block
In the Lexical playground (or another Lexical install that includes `CodeBlockPlugin`), create a new empty code block by typing three backticks. Then copy some plain text from elsewhere, which contains three or more newlines in a row (ie two blank lines), and paste it inside the code block. Text after the newlines will wind up as top-level paragraphs rather than inside the code block.

In [CodeNode.insertNewAfter](https://github.com/facebook/lexical/blob/470d1330aa461a817ad7a619fc1636ad08780f1f/packages/lexical-code-core/src/CodeNode.ts#L295), there is code that checks whether the last two children of the code block are blank lines, and if so, breaks out of the code block and starts a paragraph block. This makes sense if the user is typing (it makes it possible to leave the code block by typing enter a few times at the end), but doesn't make sense while pasting. Adding `!$hasUpdateTag(PASTE_TAG)` to the condition appears to fix it, without changing the intended UI interaction that this bit of code enables.

Lexical version: v0.43.0

anikakpillai:
## Breaking Changes

The "press enter three times to exit the CodeNode" logic has been moved from `CodeNode.insertNewAfter` to a `KEY_ENTER_COMMAND` listener registered by `CodeExtension`.

Using `CodeNode` without `CodeExtension` is now deprecated to allow this functionality to be moved to a command listener. There is a shim in the implementation that will provide backwards compatible behavior and log a warning (in dev) when this scenario is encountered, but you should expect that the node will eventually stop working correctly in a future version if your editor uses `CodeNode` without `CodeExtension`.

## Description

When pasting text containing consecutive blank lines into a code block, 
content after the blank lines incorrectly escapes the code block and 
appears as top-level paragraphs.

The root cause was that the escape-to-paragraph logic in 
`CodeNode.insertNewAfter` ran on all content changes, including paste.

Per maintainer feedback, this fix takes the preferred approach of moving 
the escape logic out of `insertNewAfter` entirely and into a 
`KEY_ENTER_COMMAND` listener registered in `CodeExtension`. This ensures 
the escape behavior only triggers on keyboard input, not paste or other 
programmatic changes.

Closes #8351

### Before
Paste text with consecutive blank lines into a code block — content 
after the blank lines escapes to a top-level paragraph.

### After
- Pasted content stays inside the code block
- Pressing Enter at the end of a code block after two blank lines 
  still correctly exits to a new paragraph
- Four unit tests added covering: escape on Enter, no escape on paste, 
  no escape with single blank line, no escape when cursor is not at end

### DIFF
--- a/packages/lexical-code-core/src/CodeNode.ts
+++ b/packages/lexical-code-core/src/CodeNode.ts
@@ -28,6 +28,7 @@ import {
   $createLineBreakNode,
   $createParagraphNode,
   $createTabNode,
+  $hasUpdateTag,
   $isTabNode,
   $isTextNode,
   addClassNamesToElement,
@@ -298,7 +299,8 @@ export class CodeNode extends ElementNode {
       children[childrenLength - 2].getTextContent() === '\n' &&
       selection.isCollapsed() &&
       selection.anchor.key === this.__key &&
-      selection.anchor.offset === childrenLength
+      selection.anchor.offset === childrenLength &&
+      !$hasUpdateTag('paste')
     ) {
       children[childrenLength - 1].remove();
       children[childrenLength - 2].remove();
--- a/packages/lexical-code-core/src/CodeNode.ts
+++ b/packages/lexical-code-core/src/CodeNode.ts
@@ -28,7 +28,6 @@ import {
   $createLineBreakNode,
   $createParagraphNode,
   $createTabNode,
-  $hasUpdateTag,
   $isTabNode,
   $isTextNode,
   addClassNamesToElement,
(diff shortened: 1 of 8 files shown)


packages/lexical-code-core/src/CodeNode.ts
```diff
@@ -298,7 +299,8 @@ export class CodeNode extends ElementNode {
       children[childrenLength - 2].getTextContent() === '\n' &&
       selection.isCollapsed() &&
       selection.anchor.key === this.__key &&
-      selection.anchor.offset === childrenLength
+      selection.anchor.offset === childrenLength &&
+      !$hasUpdateTag('paste')
```
etrepum:
```suggestion
      !$hasUpdateTag(PASTE_TAG)
```

etrepum (commented):
Generally speaking it's better if PRs add tests to demonstrate that they work as intended

anikakpillai:
Thanks for the review! I've updated the approach per your original suggestion on the issue. The escape logic is now moved out of `insertNewAfter` entirely and into a `KEY_ENTER_COMMAND` listener in `CodeExtension`, so it only fires on keyboard input. Added 4 unit tests as well.

etrepum (commented):
Your latest changes fail integrity checks (looks like type errors) and existing end-to-end tests

etrepum (commented):
This looks good, but I think we will have to think about the compatibility scenarios here since it is a breaking change, if you are using CodeNode in a legacy app that is not using CodeExtension then you will lose the functionality to exit the CodeNode by pressing return three times.

etrepum (approved):

PR #8361, opened 2026-04-16
[lexical][lexical-playground] Fix: Use the fallback matching for number keys in keyboard shortcuts

levensta:
## Description

Added a fallback check for number keys to the `isExactShortcutMatch` function. On macOS, the English keyboard layout produces the following characters when the Option key is pressed in combination with numbers:
- `1` → `¡`
- `2` → `™`
- `£` → `£`

This became apparent after the changes in this PR https://github.com/facebook/lexical/pull/8260, as the keyboard shortcuts for headings in the playground stopped working

### Before

The Option key didn't work with numbers on the English keyboard layout in macOS

### After

Support for the Option + Num hotkeys has been added for macOS. Alternative keyboard layouts, such as Dvorak or other languages, continue working

### DIFF
--- a/packages/lexical-playground/src/plugins/ShortcutsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ShortcutsPlugin/index.tsx
@@ -7,7 +7,6 @@
  */
 
 import {TOGGLE_LINK_COMMAND} from '@lexical/link';
-import {HeadingTagType} from '@lexical/rich-text';
 import {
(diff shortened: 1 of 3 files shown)


etrepum (commented):
it would be cleaner if this was done in one place instead of two, e.g. isFormatHeading could be getFormatHeading which would tell you which heading it was (or undefined)

etrepum (approved):

Issue #4103, opened 2023-03-12
Feature: No well documented way to add Grammarly-like functionality


SikandAlex:
My use case is as follows:

1) Whenever the user updates the text of the editor in any way, a request is made with the new text content of the editor to our custom spellcheck API. This is throttled so that it doesn't actually occur for every character input, only once the user has stopped typing to prevent unnecessary network requests. 

2) The API response returns a list of ranges. Example: `[[0, 5], [5, 9], [9, 15]]`. I also know if the range is "good text" or "bad text". Let's assume the [5, 9] range is a spelling error and the rest are normal text ranges.

3) I want to split the editor content into three elements, two normal `TextNodes` and then a custom node `BadSpellingNode` that extends `TextNode` that when hovered shows a contextual menu. 

I have tried many solutions to accomplish this. I've tried every variation of listening approaches stated in the documentation. I feel this should be way simpler as I was able to easily accomplish this using `slate` but abandoned that library once I found that it wasn't typed properly and had a lot of churn resulting from the `plate` extension project.

Every approach that I utilize either results in an infinite loop: my listener triggers a side effect that updates the text and then triggers the same side effect again. In addition, computing where the caret/selection needs to be on each re-render of the editor (since it is lost and there are now multiple nodes) becomes extremely convoluted.

Am I doing something wrong?

This was previously quite easy in `slate`. Code isn't super clean here but this was my general approach. 

```typescript
const Leaf = ({ attributes, children, leaf }: RenderLeafProps) => {

  if (leaf.bold) {
    children = <strong>{children}</strong>;
  }

  if (leaf.italic) {
    children = <em>{children}</em>;
  }

  if (leaf.underline) {
    children = <u>{children}</u>;
  }

  const ref = useRef(null);

  const [open, setOpen] = useState(false);
  const [anchorEl, setAnchorEl] = useState<HTMLAnchorElement | null>(null);

  const handleMouseEnter = (event: React.MouseEvent<HTMLSpanElement>) => {
    setAnchorEl(event.currentTarget);
    setOpen(true);
  };

  const handleMouseLeave = (event: React.MouseEvent<HTMSpanElement>) => {
    setAnchorEl(null);
    setOpen(false);
  };
const Leaf = ({ attributes, children, leaf }: RenderLeafProps) => {

  if (leaf.bold) {
    children = <strong>{children}</strong>;
  }

  if (leaf.italic) {
    children = <em>{children}</em>;
  }

  if (leaf.underline) {
[...]

PR #8364, opened 2026-04-17
[lexical-text] Chore: Add deprecation notice to $findTextIntersectionFromCharacters

etrepum:
## Description

`$findTextIntersectionFromCharacters` doesn't work correctly and has no tests, usage, or documentation. Like the situation with `@lexical/offset`, it would be nice to have something like this, but it should be implemented correctly with an in-repo use case, tests, and documentation.

## Test plan

N/A


etrepum:
Related https://github.com/facebook/lexical/issues/4103#issuecomment-2361462140

zurfyx (approved):

PR #8372, opened 2026-04-19
[lexical][lexical-code-core][lexical-list][lexical-table][lexical-yjs] Refactor: make runtime style updates CSP-safe

### LINKED ISSUE #8363: Feature: Refactor style management to be CSP-safe
## Description

`.setAttribute('style', …)` and `.style.cssText = …` violate the [style-src-attr](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-attr) CSP directive where you are essentially not allowed to have the browser parse CSS after initial page load.

Most use cases we have are simply convenient ways to set it to specific values or to support user extension, both of which can be done with style objects instead, and with extensions and middleware (#8353) that will get even easier.

## Impact

Will allow lexical to be used on security-conscious pages.

holly-agyei:
## Description
This replaces runtime style string writes in Lexical's DOM update paths with property-based style updates.

- add a shared helper to parse stored style strings and apply or remove individual properties
- update text, code, list, table, yjs cursor, and playground paths that previously used `cssText` or `setAttribute('style', ...)`
- add unit coverage for style creation and update behavior

Closes #8363

### Before
- Runtime DOM updates used `.style.cssText` and `setAttribute('style', ...)` in editor paths.

### After
- `env COREPACK_HOME=/tmp/corepack corepack pnpm run test-unit`
- `env COREPACK_HOME=/tmp/corepack corepack pnpm run tsc`
- `env COREPACK_HOME=/tmp/corepack corepack pnpm --filter @lexical/devtools run compile`
- `env COREPACK_HOME=/tmp/corepack corepack pnpm run build-types`
- `env COREPACK_HOME=/tmp/corepack corepack pnpm run lint`
- `env COREPACK_HOME=/tmp/corepack corepack pnpm run prettier`

### DIFF
--- a/packages/lexical-selection/src/constants.ts
+++ b/packages/lexical-selection/src/constants.ts
@@ -1,8 +1,0 @@
-/**
- * Copyright (c) Meta Platforms, Inc. and affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- *
- */
-export const CSS_TO_STYLES: Map<string, Record<string, string>> = new Map();
--- a/packages/lexical-selection/src/index.ts
+++ b/packages/lexical-selection/src/index.ts
@@ -6,6 +6,8 @@
  *
  */
 
+import {getStyleObjectFromCSS as getStyleObjectFromCSS_} from 'lexical';
+
 import {$trimTextContentFromAnchor} from './lexical-node';
 
 export {
@@ -34,8 +36,9 @@ export {
   createDOMRange,
   createRectsFromDOMRange,
   getCSSFromStyleObject,
-  /** @deprecated moved to the lexical package */ getStyleObjectFromCSS,
 } from './utils';
+/** @deprecated moved to the `lexical` package */
+export const getStyleObjectFromCSS = getStyleObjectFromCSS_;
 /** @deprecated renamed to {@link $trimTextContentFromAnchor} by @lexical/eslint-plugin rules-of-lexical */
 export const trimTextContentFromAnchor = $trimTextContentFromAnchor;
 export {
--- a/packages/lexical-selection/src/index.ts
+++ b/packages/lexical-selection/src/index.ts
@@ -34,7 +34,7 @@ export {
   createDOMRange,
   createRectsFromDOMRange,
   getCSSFromStyleObject,
-  getStyleObjectFromCSS,
+  /** @deprecated moved to the lexical package */ getStyleObjectFromCSS,
 } from './utils';
 /** @deprecated renamed to {@link $trimTextContentFromAnchor} by @lexical/eslint-plugin rules-of-lexical */
 export const trimTextContentFromAnchor = $trimTextContentFromAnchor;
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -20,6 +20,7 @@ import {
   $isTokenOrSegmented,
   BaseSelection,
   ElementNode,
+  getStyleObjectFromCSS,
   LexicalEditor,
   LexicalNode,
   NodeKey,
@@ -28,8 +29,9 @@ import {
   TextNode,
 } from 'lexical';
 import invariant from 'shared/invariant';
+import warnOnlyOnce from 'shared/warnOnlyOnce';
 
-import {getCSSFromStyleObject, getStyleObjectFromCSS} from './utils';
+import {getCSSFromStyleObject} from './utils';
 
 /**
  * Generally used to append text content to HTML and JSON. Grabs the text content and "slices"
@@ -239,9 +241,11 @@ export function $trimTextContentFromAnchor(
 }
 
 /**
- * Kept for backwards compatibility. Styles are parsed on demand now.
+ * @deprecated node styles are parsed on demand and not cached eternally
  */
-export function $addNodeStyle(_node: TextNode): void {}
+export const $addNodeStyle: (_node: TextNode) => void = warnOnlyOnce(
+  '$addNodeStyle is a deprecated no-op and calls should be removed',
+);
 
 /**
  * Applies the provided styles to the given TextNode, ElementNode, or
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -29,10 +29,7 @@ import {
 } from 'lexical';
 import invariant from 'shared/invariant';
 
-import {
-  getCSSFromStyleObject,
-  getStyleObjectFromCSS,
-} from './utils';
+import {getCSSFromStyleObject, getStyleObjectFromCSS} from './utils';
 
 /**
  * Generally used to append text content to HTML and JSON. Grabs the text content and "slices"
--- a/packages/lexical-selection/src/lexical-node.ts
+++ b/packages/lexical-selection/src/lexical-node.ts
@@ -29,11 +29,9 @@ import {
 } from 'lexical';
 import invariant from 'shared/invariant';
 
-import {CSS_TO_STYLES} from './constants';
 import {
   getCSSFromStyleObject,
   getStyleObjectFromCSS,
-  getStyleObjectFromRawCSS,
 } from './utils';
 
 /**
@@ -244,14 +242,9 @@ export function $trimTextContentFromAnchor(
 }
 
 /**
- * Gets the TextNode's style object and adds the styles to the CSS.
- * @param node - The TextNode to add styles to.
+ * Kept for backwards compatibility. Styles are parsed on demand now.
  */
-export function $addNodeStyle(node: TextNode): void {
-  const CSSText = node.getStyle();
-  const styles = getStyleObjectFromRawCSS(CSSText);
-  CSS_TO_STYLES.set(CSSText, styles);
-}
+export function $addNodeStyle(_node: TextNode): void {}
 
 /**
  * Applies the provided styles to the given TextNode, ElementNode, or
@@ -301,7 +294,6 @@ export function $patchStyle(
   } else {
     target.setTextStyle(newCSSText);
   }
-  CSS_TO_STYLES.set(newCSSText, newStyles);
 }
 
 /**
--- a/packages/lexical-selection/src/utils.ts
+++ b/packages/lexical-selection/src/utils.ts
@@ -186,13 +186,6 @@ export function getStyleObjectFromRawCSS(css: string): Record<string, string> {
  * @param css - The CSS property as a string.
  * @returns The value of the given CSS property.
  */
-export {getStyleObjectFromCSS};
-
-/**
- * Gets the CSS styles from the style object.
- * @param styles - The style object containing the styles to get.
- * @returns A string containing the CSS styles and their values.
- */
 export function getCSSFromStyleObject(styles: Record<string, string>): string {
   let css = '';
 
--- a/packages/lexical-selection/src/utils.ts
+++ b/packages/lexical-selection/src/utils.ts
@@ -7,9 +7,12 @@
  */
 import type {ElementNode, LexicalEditor, LexicalNode} from 'lexical';
 
-import {$getEditor, $isRootNode, $isTextNode} from 'lexical';
-
-import {CSS_TO_STYLES} from './constants';
+import {
+  $getEditor,
+  $isRootNode,
+  $isTextNode,
+  getStyleObjectFromCSS,
+} from 'lexical';
 
 function getDOMTextNode(element: Node | null): Text | null {
   let node = element;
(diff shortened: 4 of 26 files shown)


packages/shared/src/setDOMStyle.ts
```diff
@@ -0,0 +1,194 @@
...
+  value: string,
+): void {
+  const priority = IMPORTANT_REG_EXP.test(value) ? 'important' : '';
+  const nextValue =
+    priority === '' ? value : value.replace(IMPORTANT_REG_EXP, '').trim();
+
+  domStyle.setProperty(property, nextValue, priority);
+}
+
+export function getStyleObjectFromCSS(css: string): Record<string, string> {
```
etrepum: This file should be in the lexical package and exported from there, having it in shared means that a separate copy of this function is present in every package that uses it (and it's not available for external re-use).

There's also other style parsing code in lexical that this should possibly be unified with.

etrepum: It may also be worth just taking a dependency on inline-style-parser to implement this feature

holly-agyei: > This file should be in the lexical package and exported from there, having it in shared means that a separate copy of this function is present in every package that uses it (and it's not available for external re-use).
> 
> There's also other style parsing code in lexical that this should possibly be unified with.

Thanks, addressed in the latest update.

I moved the helper into lexical and exported it from there, and I also unified the parsing path so @lexical/selection now reuses the lexical implementation instead of keeping a separate parser. I considered inline-style-parser, but since there was already related style parsing logic in-tree and the needed behavior is small/specific here, I kept it dependency-free.

packages/lexical-selection/src/utils.ts
```diff
@@ -7,7 +7,12 @@
  */
 import type {ElementNode, LexicalEditor, LexicalNode} from 'lexical';
 
-import {$getEditor, $isRootNode, $isTextNode} from 'lexical';
+import {
+  $getEditor,
+  $isRootNode,
+  $isTextNode,
+  getStyleObjectFromCSS as getStyleObjectFromLexicalCSS,
```
etrepum: This should just be an import, you can re-export it with `export {getStyleObjectFromCSS};` later instead of redefining it

holly-agyei: > This should just be an import, you can re-export it with `export {getStyleObjectFromCSS};` later instead of redefining it

Thank you. Updated to do that. getStyleObjectFromCSS is now imported directly from lexical and re-exported, rather than being locally redefined in @lexical/selection.

holly-agyei:
@etrepum All tests are now passing, and I’ve updated the formatting for the lexical-selection imports as discussed. Could you take another look when you have a moment?

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -244,14 +239,9 @@ export function $trimTextContentFromAnchor(
 }
 
 /**
- * Gets the TextNode's style object and adds the styles to the CSS.
- * @param node - The TextNode to add styles to.
+ * Kept for backwards compatibility. Styles are parsed on demand now.
```
etrepum:
```suggestion
 * @deprecated node styles are parsed on demand and not cached eternally
```

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -244,14 +239,9 @@ export function $trimTextContentFromAnchor(
...
  */
-export function $addNodeStyle(node: TextNode): void {
-  const CSSText = node.getStyle();
-  const styles = getStyleObjectFromRawCSS(CSSText);
-  CSS_TO_STYLES.set(CSSText, styles);
-}
+export function $addNodeStyle(_node: TextNode): void {}
```
etrepum: If we are removing this it should probably have a deprecation warning on use since it's not doing anything anymore and the code should be removed.

```suggestion
export const $addNodeStyle: (_node: TextNode) => void = warnOnce('$addNodeStyle is a deprecated no-op and calls should be removed');
```

packages/lexical-selection/src/index.ts
```diff
@@ -34,7 +34,7 @@ export {
   createDOMRange,
   createRectsFromDOMRange,
   getCSSFromStyleObject,
-  getStyleObjectFromCSS,
+  /** @deprecated moved to the lexical package */ getStyleObjectFromCSS,
```
etrepum: These comments on export don't actually work well in editors, they won't show the deprecation message unless it's a separate constant. The best way to do this is something like:

```ts
import {getStyleObjectFromCSS as getStyleObjectFromCSS_} from 'lexical';
/** @deprecated moved to the `lexical` package */
export const getStyleObjectFromCSS = getStyleObjectFromCSS_;
```

packages/lexical-selection/src/lexical-node.ts
```diff
@@ -29,12 +29,7 @@ import {
 } from 'lexical';
 import invariant from 'shared/invariant';
 
-import {CSS_TO_STYLES} from './constants';
-import {
-  getCSSFromStyleObject,
-  getStyleObjectFromCSS,
-  getStyleObjectFromRawCSS,
-} from './utils';
+import {getCSSFromStyleObject, getStyleObjectFromCSS} from './utils';
```
etrepum: getStyleObjectFromCSS should be imported from lexical

packages/lexical-selection/src/utils.ts
```diff
@@ -175,43 +178,15 @@ export function createRectsFromDOMRange(
...
-  }
-
-  if (__DEV__) {
-    // Freeze the value in DEV to prevent accidental mutations
-    Object.freeze(value);
-  }
-
-  return value;
-}
+export {getStyleObjectFromCSS};
```
etrepum: This re-export and documentation should not be here, it should be imported directly from lexical

packages/lexical-selection/src/constants.ts
```diff
@@ -1,8 +0,0 @@
-/**
- * Copyright (c) Meta Platforms, Inc. and affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- *
- */
-export const CSS_TO_STYLES: Map<string, Record<string, string>> = new Map();
```
etrepum, line 8: Removing the style cache has some performance implications (both good and bad). This probably should have some additional consideration.

etrepum (commented):
It's definitely getting close but the change in the caching behavior means that a very thorough audit should be done - possibly adding a different (better) cache.

packages/lexical-selection/src/constants.ts
holly-agyei: I took another pass on the cache behavior here. The parser in lexical now keeps a small bounded cache for repeated style strings, but still returns a fresh object to callers so we don't reintroduce the old eternal selection cache behavior or expose cached objects to mutation.

holly-agyei:
@etrepum  
Thanks for pointing these out to look at.
I have gone through the reviews to fix them. Kindly take a look

etrepum (commented):
Like before I think because this changes the cache fundamentally I think it needs a very close audit. It's also not clear that 100 is a good size for this cache, or even if the cache is needed at all. There isn't any sort of easy to run benchmark that would make it easy to make that kind of decision. It's likely that the caching would be better to do closer to the nodes that the styles are for, if caching is useful, but that would also take careful consideration to avoid causing problems with collab.

packages/lexical-selection/src/constants.ts
holly-agyei: I took a closer pass on the cache question and changed direction. I removed the global parser cache entirely, kept getStyleObjectFromCSS as a pure parse, optimized setDOMStyleFromCSS so it applies parsed properties directly without allocating intermediate style objects, and added a closer-to-the-node fast path in ListItemNode so marker styles are skipped when __textStyle is unchanged.

holly-agyei:
> Like before I think because this changes the cache fundamentally I think it needs a very close audit. It's also not clear that 100 is a good size for this cache, or even if the cache is needed at all. There isn't any sort of easy to run benchmark that would make it easy to make that kind of decision. It's likely that the caching would be better to do closer to the nodes that the styles are for, if caching is useful, but that would also take careful consideration to avoid causing problems with collab.

Thanks, I took a closer pass on the cache behavior with performance as the primary goal and changed direction.

 I removed the global parser cache entirely. After auditing it more closely, I didn’t think I could justify that policy in this PR: the `100` entry size was arbitrary, the hit rate sounds quite unclear(it changes global parser behavior across editors), and because callers still need fresh objects, it would still allocate on cache hits. So it added complexity without a strong enough performance case.

Instead I moved the optimization closer to the actual hot paths:

- `getStyleObjectFromCSS` is pure again and returns a fresh object.
- `setDOMStyleFromCSS` now iterates parsed declarations directly when applying/removing DOM styles, so that path avoids building intermediate style objects.
- `ListItemNode` now skips marker style work entirely when `__textStyle` is unchanged, which felt like a better place to capture repeated work than a global parser cache.

The tradeoff I chose was to avoid a broad cache-policy change and keep the parser semantics simple, while still making the important runtime paths cheaper. That seemed like the safer performance-oriented shape here, especially without a benchmark that would justify a specific cache strategy or size.

packages/lexical/src/index.ts
```diff
@@ -395,3 +395,8 @@ export {
   removeClassNamesFromElement,
 } from './utils/classNames';
 export {mergeRegister} from './utils/mergeRegister';
+export {
+  getStyleObjectFromCSS,
+  setDOMStyleFromCSS,
+  setDOMStyleObject,
```
etrepum: All of these new functions should have API docs, it is especially important to note that these CSS objects are intended to be used with `style.setProperty` so there is an expectation that all of the properties are kebab-case (e.g. `font-size` not `fontSize`) and all units are specified like Solid https://github.com/solidjs/solid/issues/143

packages/lexical/src/utils/setDOMStyle.ts
```diff
@@ -0,0 +1,230 @@
...
+      const value = styleDeclaration.slice(i + 1).trim();
+
+      return property !== '' && value !== '' ? [property, value] : null;
+    }
+  }
+
+  return null;
+}
+
+function forEachStylePropertyInCSS(
```
etrepum: for performance reasons it's probably better to just use for loops rather than abstracting this

packages/lexical/src/utils/setDOMStyle.ts
```diff
@@ -0,0 +1,230 @@
...
+ */
+
+const IMPORTANT_REG_EXP = /\s*!important\s*$/i;
+
+function getStyleDeclarations(css: string): Array<string> {
+  const declarations: Array<string> = [];
+
+  if (!css) {
+    return declarations;
+  }
```
etrepum: We don't really have any use case for tokenizing the declarations separately from the properties so it should really be something like this where we directly parse the css into its final form without intermediate objects that just get garbage collected immediately:

```suggestion
function getStyleObjectFromCSS(css: string): Record<string, string> {
  const styles: Record<string, string> = {};

  if (!css) {
    return styles;
  }
  // parsing code here to find declarations and then parse them
  return styles;
}
```

`getStylePropertyAndValue` could also take the record as an argument and set the value directly, e.g.

```ts
function parseStylePropertyAndValue(
  styles: Record<string, string>,
  styleDeclaration: string,
): void {
  // parsing code omitted
  if (property !== '' && value !== '') {
    styles[property] = value;
  }
}
```

packages/lexical/src/utils/setDOMStyle.ts
```diff
@@ -0,0 +1,230 @@
...
+  if (cssText === prevCSSText) {
+    return;
+  }
+
+  forEachStylePropertyInCSS(prevCSSText, (property) => {
+    domStyle.removeProperty(property);
+  });
+  forEachStylePropertyInCSS(cssText, (property, value) => {
+    setDOMStyleProperty(domStyle, property, value);
+  });
```
etrepum: I think a better algorithm here would be to do something like this to minimize calls to the style object:

```ts
const prevCSS = getStyleObjectFromCSS(prevCSSText);
const nextCSS = getStyleObjectFromCSS(cssText);
for (const property in nextCSS) {
  delete prevCSS[property];
  setDOMStyleProperty(domStyle, property, nextCSS[property]);
}
for (const property in prevCSS) {
  domStyle.removeProperty(property);
}
```

etrepum (commented):
Let's optimize this parser implementation to minimize the impact of not having a cache

I think the place that might be impacted the most by not having a cache is code that uses `$getSelectionStyleValueForProperty` repeatedly for the same selection, but there are other ways to optimize that than a global cache.

packages/lexical/src/utils/setDOMStyle.ts
holly-agyei: Updated in that direction. I removed the callback-based abstraction and tightened the parser to a direct single-pass `for` loop so it parses straight into the final style record.

packages/lexical/src/utils/setDOMStyle.ts
holly-agyei: Thanks, I reworked this to follow that shape more closely. `getStyleObjectFromCSS` now parses directly into its final object in one pass instead of tokenizing declarations first or returning intermediate tuples.

packages/lexical/src/index.ts
holly-agyei: Thanks, updated this. The public exports in `lexical/src/index.ts` now have API docs, and I called out that the returned style objects are intended for `style.setProperty`, so properties are expected to be kebab-case and values need explicit units where applicable.

packages/lexical/src/utils/setDOMStyle.ts
holly-agyei: Updated to use that diffing approach. `setDOMStyleFromCSS` now parses `prevCSSText` and `cssText` into objects, applies `nextCSS` with `setProperty`, deletes matched keys from `prevCSS`, and only removes the remaining stale properties afterward.

packages/lexical/src/index.ts
```diff
@@ -6,6 +6,12 @@
  *
  */
 
+import {
+  getStyleObjectFromCSS as getStyleObjectFromCSS_,
+  setDOMStyleFromCSS as setDOMStyleFromCSS_,
+  setDOMStyleObject as setDOMStyleObject_,
+} from './utils/setDOMStyle';
```
etrepum: This technique only needs to be done when deprecating re-exports, don't do it here. Put the API docs with the implementation of the function and simply export these directly.

packages/lexical/src/index.ts
```diff
@@ -395,3 +401,27 @@ export {
...
+ */
+export const setDOMStyleFromCSS = setDOMStyleFromCSS_;
+/**
+ * Applies a style object to a DOM style declaration using
+ * `CSSStyleDeclaration.setProperty()`.
+ *
+ * Property names are expected to be kebab-case, such as `font-size`, and
+ * values are expected to include explicit units where needed, such as `12px`.
+ */
+export const setDOMStyleObject = setDOMStyleObject_;
```
etrepum: All of this should be colocated with the definitions of these functions

etrepum (approved):

PR #8374, opened 2026-04-20
Fix: resolve invalid offset crash during equation Markdown conversion (#6936)

### LINKED ISSUE #6936: Bug: Block math equation is falsely converted to inline equation on MarkdownShortcutPlugin

Ranjan00001:
Resolves #6936 

## Description
This PR resolves an invariant violation (`$getTextNodeOffset: invalid offset -1`) that occurs when interacting with non-inline/block `EquationNode`s during Markdown conversion, and hardens background syntax highlighting retention logic. 

**Root Cause:**
When a block equation was converted via `$convertToMarkdownString()`, `CodeHighlighterPrism` attempted to automatically re-apply syntax highlighting to the newly created markdown string. The highlighter evaluates surrounding nodes to hold the user's cursor steady. However, when evaluating padding elements (like `LineBreakNode`s naturally present in markdown block equations), the highlighter logic decremented the `textOffset` marker without bounds checking. This caused a `-1` offset to be forwarded into `TextNode.select()`, terminating the TreeView render routine with an invariant violation. 

Also, since the cursor defaults to the `root` index outside the CodeNode upon conversion, the highlighter was improperly capturing this external selection and forcing it inside the CodeNode.

**Changes Made:**
1. **Hardened [CodeHighlighterPrism.ts](cci:7://file:///home/dr/Documents/openS/lexical/packages/lexical-code-prism/src/CodeHighlighterPrism.ts:0:0-0:0) Retainer:** 
   - Scoped [$updateAndRetainSelection](cci:1://file:///home/dr/Documents/openS/lexical/packages/lexical-code-prism/src/CodeHighlighterPrism.ts:210:0-280:1) to ensure the highlighter only attempts to recalculate and shift cursor offsets if the user's active selection is actually localized *inside* the `CodeNode` being evaluated. This prevents nasty "cursor stealing" edge cases during collaborative or API-driven edits.
   - Introduced a `Math.max(0, textOffset)` clamp to cleanly handle offset decrements across non-textual nodes (like markdown line breaks).

2. **Added `EQUATION_BLOCK` Transformer:**
   - Supplemental to fixing the crash, it was observed that block equations lacked proper multi-line markdown support, resulting in them falling back to raw output and fracturing into disconnected paragraphs on import.
   - I implemented `EQUATION_BLOCK` as a [MultilineElementTransformer](cci:2://file:///home/dr/Documents/openS/lexical/packages/lexical-markdown/src/MarkdownTransformers.ts:86:0-153:2) to natively support exporting to block boundary syntax (`$$ ... $$`) and properly digesting multi-line markdown strings back into full `EquationNode`s.

## Test Plan
- Go to the Playground. Insert an Equation and uncheck "Inline".
- Input an expression such as `E=mc^2` and confirm.
- Click "Convert From Markdown" to enter Markdown view. Ensure the editor does not crash and cleanly exports a `$$` bounded block.
- Click "Convert From Markdown" again to re-enter rich text mode. Confirm the multiline block mathematically parses back into a unified `EquationNode` without splitting.


etrepum (commented):
This PR needs to include tests that can confirm that the issue is fixed (the tests should fail without the changes)

Ranjan00001:
@etrepum Have added the tests. Please check and let me know if I'm missing anything.

etrepum (commented):
Looks like the tests are failing, did you run it locally first?

etrepum:
No activity since first review, closing as stale. Re-open or create a new PR if you intend to continue work on this.

PR #8378, opened 2026-04-20
[lexical-link][lexical-react] Feature: Allow custom punctuation for AutoLink boundaries

### LINKED ISSUE #8189: Feature: Add colon to allowed separators for the AutoLinkPlugin

holly-agyei:
Fixes #8189

## Summary
- add an optional punctuation setting to AutoLinkExtension and registerAutoLink
- thread that option through AutoLinkPlugin while preserving the existing default separators
- add coverage for colon-delimited custom matches and ensure URLs with embedded colons still autolink correctly
- sync the published Flow declarations and React plugin docs with the new config

## Test plan
- ./node_modules/.bin/vitest --project unit --run packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts
- ./node_modules/.bin/eslint --no-warn-ignored packages/lexical-link/src/LexicalAutoLinkExtension.ts packages/lexical-link/src/__tests__/unit/LexicalAutoLinkExtension.test.ts packages/lexical-react/src/LexicalAutoLinkPlugin.ts
- ./node_modules/.bin/tsc -p tsconfig.json --noEmit


etrepum (commented):
`pnpm run ci-check` is failing

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -81,18 +81,21 @@ function findFirstMatch(
   return null;
 }
 
-const PUNCTUATION_OR_SPACE = /[.,;\s]/;
+const DEFAULT_PUNCTUATION = '.,;';
 
-function isSeparator(char: string): boolean {
-  return PUNCTUATION_OR_SPACE.test(char);
+function isSeparator(char: string, punctuation: string): boolean {
```
etrepum: for performance reasons punctuation should be a regex as it is here, it can be compiled from a string when the extension is built

holly-agyei: @etrepum  Oh yes, thank you very much, that will also avoid repeated work.
I have switched this back to a compiled separator regex built once in registerAutoLink, and added regression coverage for regex-special punctuation.

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -81,18 +81,36 @@ function findFirstMatch(
   return null;
 }
 
-const PUNCTUATION_OR_SPACE = /[.,;\s]/;
+const DEFAULT_PUNCTUATION = '.,;';
+const DEFAULT_SEPARATOR = /[.,;\s]/;
+const REGEX_SPECIAL_CHARS = /[\\^$.*+?()[\]{}|/-]/g;
```
etrepum: The set of special chars is a bit different inside of a character class, it might be better to just accept a RegExp instead of a string and the user is responsible for making sure it's correct. We can provide an example of the default `/[.,;\s]/` in the api documentation comment. Let's rename punctuation to separatorRegex.

There's no need to support these new options in AutoLinkPlugin, people should be migrating to AutoLinkExtension for new features.

You can take a look at the past attempt at this feature for how that would work https://github.com/facebook/lexical/pull/8191

holly-agyei: @etrepum  Thanks, I checked the PR you linked, and it gave a good signal. 
I've updated this to follow that shape: the option is now separatorRegex on AutoLinkExtension only, with the default documented in the API comment, and I removed the new AutoLinkPlugin/docs surface.

packages/lexical-link/flow/LexicalLink.js.flow
```diff
@@ -122,8 +122,10 @@ declare export function registerClickableLink(
 ): () => void;
 
 export type AutoLinkConfig = {
-  matchers: LinkMatcher[];
   changeHandlers: ChangeHandler[];
+  excludeParents: Array<(parent: ElementNode) => boolean>;
```
etrepum:
```suggestion
  excludeParents: ((parent: ElementNode) => boolean)[];
```

packages/lexical-link/flow/LexicalLink.js.flow
```diff
@@ -122,8 +122,10 @@ declare export function registerClickableLink(
...
+  matchers: LinkMatcher[];
+  separatorRegex?: RegExp;
```
etrepum:
```suggestion
  separatorRegex: RegExp;
```

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -507,22 +525,34 @@ function getTextNodesToMatch(textNode: TextNode): TextNode[] {
 }
 
 export interface AutoLinkConfig {
-  matchers: LinkMatcher[];
   changeHandlers: ChangeHandler[];
   excludeParents: Array<(parent: ElementNode) => boolean>;
```
etrepum:
```suggestion
  excludeParents: ((parent: ElementNode) => boolean)[];
```

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -507,22 +525,34 @@ function getTextNodesToMatch(textNode: TextNode): TextNode[] {
...
+  matchers: LinkMatcher[];
+  /**
+   * The regular expression used to determine whether surrounding
+   * characters count as separators when validating auto-link
+   * boundaries. Defaults to `/[.,;\s]/`.
+   */
+  separatorRegex?: RegExp;
```
etrepum: Configuration of extensions should always use required properties unless undefined is specifically a value that is useful to the extension.

```suggestion
  separatorRegex: RegExp;
```

packages/lexical-link/src/LexicalAutoLinkExtension.ts
```diff
@@ -507,22 +525,34 @@ function getTextNodesToMatch(textNode: TextNode): TextNode[] {
...
 const defaultConfig: AutoLinkConfig = {
   changeHandlers: [],
   excludeParents: [],
   matchers: [],
+  separatorRegex: PUNCTUATION_OR_SPACE,
 };
 
 export function registerAutoLink(
   editor: LexicalEditor,
   config: AutoLinkConfig = defaultConfig,
```
etrepum: This type will have to change accordingly to something like `Partial<AutoLinkConfig>` (which would require some implementation changes below to provide defaults for each value) or `Partial<AutoLinkConfig> & Omit<AutoLinkConfig, 'separatorRegex'>`

```suggestion
  config: Partial<AutoLinkConfig> & Omit<AutoLinkConfig, 'separatorRegex'> = defaultConfig,
```

packages/lexical-react/flow/LexicalAutoLinkPlugin.js.flow
```diff
@@ -8,11 +8,13 @@
  */
 
 import type {LinkAttributes, ChangeHandler, LinkMatcher} from '@lexical/link';
+import type {ElementNode} from 'lexical';
 import {createLinkMatcherWithRegExp} from '@lexical/link';
 export {createLinkMatcherWithRegExp};
 export type {ChangeHandler, LinkMatcher};
 
 declare export function AutoLinkPlugin(props: {
+  excludeParents?: Array<(parent: ElementNode) => boolean>,
```
etrepum:
```suggestion
  excludeParents?: ((parent: ElementNode) => boolean)[],
```

packages/lexical-react/flow/LexicalAutoLinkPlugin.js.flow
```diff
@@ -8,11 +8,13 @@
...
   matchers: Array<LinkMatcher>,
```
etrepum:
```suggestion
  matchers: LinkMatcher[],
```

holly-agyei:
@etrepum I've pushed the updates to address your feedback regarding the Flow array syntax and the AutoLinkConfig properties. Could you please take another look when you have a chance?

packages/lexical-link/flow/LexicalLink.js.flow
etrepum: Still not done.

packages/lexical-link/flow/LexicalLink.js.flow
etrepum: Still not done.

packages/lexical-link/src/LexicalAutoLinkExtension.ts
etrepum: Still not done

packages/lexical-link/src/LexicalAutoLinkExtension.ts
etrepum: Still not done

packages/lexical-link/src/LexicalAutoLinkExtension.ts
etrepum: Still not done

etrepum (commented):
It doesn't look like you addressed any of the last round of suggestions?

etrepum (approved):

PR #8379, opened 2026-04-20
[lexical-playground] Feature: HTML conversion button

levensta:
## Description

Here are some minor UI improvements to the lexical-playground:
- The Markdown mode button now has an active state.
- The settings block can be scrolled, as not all buttons are visible at small viewport, especially with the DevTools panel open at the bottom

I also added a button for converting to and from HTML, similar to Markdown. Unlike the _"export DOM"_ button in the _"debug view"_ panel, this mode clearly demonstrates how serialization to and from HTML works, enabling debugging of `exportDOM`/`importDOM` operations and providing a new way to write tests.

### Before

https://github.com/user-attachments/assets/3e489c71-a251-41c6-ad1c-1808c8231f5d

### After

https://github.com/user-attachments/assets/c2d4aee2-33df-49e9-982c-0fb6ba8f7baa

### DIFF
--- a/packages/lexical-playground/__tests__/e2e/HTML.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/HTML.spec.mjs
@@ -26,9 +26,10 @@ import {
 
 test.describe('HTML', () => {
   test.beforeEach(({isCollab, page}) => initialize({isCollab, page}));
-  test(`Can export HTML using the button`, async ({page}) => {
-    await focusEditor(page);
+  test(`Can export HTML using the button`, async ({page, isPlainText}) => {
+    test.skip(isPlainText);
 
+    await focusEditor(page);
     await page.keyboard.type('Foo');
     await applyHeading(page, 1);
     await page.keyboard.press('Enter');
@@ -135,7 +136,9 @@ test.describe('HTML', () => {
     );
   });
 
-  test(`Can import HTML using the button`, async ({page}) => {
+  test(`Can import HTML using the button`, async ({page, isPlainText}) => {
+    test.skip(isPlainText);
+
     await focusEditor(page);
     await click(page, '.action-button .html');
 
--- a/packages/lexical-playground/__tests__/e2e/HTML.spec.mjs
+++ b/packages/lexical-playground/__tests__/e2e/HTML.spec.mjs
@@ -1,0 +1,221 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {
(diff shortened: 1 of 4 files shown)


etrepum (commented):
All of these look like positive changes, but maybe since it enables adding tests we should write a test or two to confirm that it does the expected thing (convert to html and then convert back from html). We should probably make some sort of change while it's in html view so we can confirm that it's the actual import.

packages/lexical-playground/__tests__/e2e/HTML.spec.mjs
```diff
@@ -0,0 +1,224 @@
...
+    await page.keyboard.press('Enter');
+    await applyCodeBlock(page);
+    await page.keyboard.type('const x = "hello world";');
+    await repeat(3, async () => await page.keyboard.press('Enter'));
+    await page.keyboard.type('Today ');
+    await toggleBold(page);
+    await toggleItalic(page);
+    await insertDateTime(page);
+
+    const expectedViewHtml = `<h1 class="PlaygroundEditorTheme__h1" dir="auto">
```
etrepum, line 44: why not use the html template string here?
```suggestion
    const expectedViewHtml = html`<h1 class="PlaygroundEditorTheme__h1" dir="auto">
```

etrepum (commented):
I think this is a great idea but we probably want to integrate something like prettier in here so that it's more usable. The node-state-style example does something like that for its html view

etrepum (approved):

levensta:
> I think this is a great idea but we probably want to integrate something like prettier in here so that it's more usable. The node-state-style example does something like that for its html view

We could automatically run Prettier when switching to HTML mode, but since text elements with the inline style `white-space: "pre-wrap";` are sensitive to whitespace during re-import, we'll need to add the following settings:
```js
{
  htmlWhitespaceSensitivity: 'css',
  printWidth: Number.MAX_SAFE_INTEGER
}
```

It will look like this. It's not very convenient, but it's better than cramming everything into a single line

image

Another question is which specific Prettier to run. I could try running the Prettier that’s already in the CodeActionPlugin
https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/CodeActionMenuPlugin/components/PrettierButton/index.tsx
Probably it'll have to handle using the React context.

Or run it as a separate instance within the current ActionsPlugin. But then we'll have to set the same Prettier settings in both places so that the result is consistent

etrepum:
We can also normalize the html a bit first to make it less sensitive to whitespace and export fewer theme classes

levensta:
If this is done via DOMExtension, it will mislead users, as it may appear to be a original export of built-in nodes

etrepum:
Well if it was used in a normal way it would affect all exports, unless it was dispatching based on some render context that was only set for the html view

PR #8381, opened 2026-04-22
[lexical-markdown] Bug Fix: Code spans take precedence over inline formatting in shortcuts

potatowagon:
## Description
Per CommonMark spec, code spans (backticks) should take precedence over other inline formatting. Previously, typing `__bold__` would trigger the bold transformer before the closing backtick could complete the code span. Now, text format transformers skip matches that fall inside an unclosed code span.

### Before

https://github.com/user-attachments/assets/7d4b08f6-3257-49e6-a566-9f3062ab57c3

### After

https://github.com/user-attachments/assets/b99e1873-a5de-4718-b1d1-bd150df6a572

### DIFF
--- a/packages/lexical-markdown/src/MarkdownShortcuts.ts
+++ b/packages/lexical-markdown/src/MarkdownShortcuts.ts
@@ -391,22 +391,6 @@ function $isInsideUnclosedCodeSpan(node: TextNode, offset: number): boolean {
     }
   }
 
-  let sibling = node.getPreviousSibling();
-  while (sibling !== null) {
-    if ($isLineBreakNode(sibling)) {
-      break;
-    }
-    if ($isTextNode(sibling)) {
-      const siblingText = sibling.getTextContent();
-      for (let i = 0; i < siblingText.length; i++) {
-        if (siblingText[i] === '`') {
-          backtickCount++;
-        }
-      }
-    }
-    sibling = sibling.getPreviousSibling();
-  }
-
   return backtickCount % 2 !== 0;
 }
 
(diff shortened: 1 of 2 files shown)


etrepum (commented):
would be great to add some tests to this to confirm that it fixes the issue

packages/lexical-markdown/src/MarkdownShortcuts.ts
```diff
@@ -369,6 +377,39 @@ function $runTextFormatTransformers(
...
+    if ($isTextNode(sibling)) {
+      const siblingText = sibling.getTextContent();
+      for (let i = 0; i < siblingText.length; i++) {
+        if (siblingText[i] === '`') {
+          backtickCount++;
+        }
+      }
+    }
+    sibling = sibling.getPreviousSibling();
+  }
```
etrepum: Do we really need to look at adjacent TextNodes?

potatowagon: hello etrepum, yes, the opening tag search at line 261-283 walks previous TextNode siblings to find the open tag, so openNode may be a different node than anchorNode. Consider: a user types `code which creates a code-formatted TextNode, then continues typing __bold__. The backtick is in a previous sibling TextNode. Without checking siblings, $isInsideUnclosedCodeSpan would miss the opening backtick and incorrectly allow the bold transform.

etrepum: yeah but that's a different TextNode and the bold formatting will go away if we treat it as markdown, I think it would work as expected if the shortcut only cared about the current node

etrepum (approved):

packages/lexical-markdown/src/MarkdownShortcuts.ts
potatowagon: thanks for the comment! yea agree it isnt needed, removed

PR #8383, opened 2026-04-22
[lexical-rich-text] Feature: Add opt-in format escape triggers for RichTextExtension

### LINKED ISSUE #4936: Bug: Cannot exit code block without use of toolbar
Lexical version: v0.12.0

## Steps To Reproduce

1. Create code block at end of a line
2. Try to exit with right arrow, clicking to the right of the block, `Enter`, or other key entries/inputs

CleanShot 2023-08-29 at 13 13 42

## The current behavior
Cannot exit a code block without use of the formatting toolbar

## The expected behavior
Most editors allow a right arrow or a click outside (to the leading edge/cursor side) of the code block to 'exit' it and go back to unformatted text. In effect, you're 'stuck' in the code block. 

I've submitted this as a bug as it seems like unexpected behavior, but I could see it as a feature request as well.

vishisht31:
## Description
  - Adds a new `EscapeFormatAtBoundaryExtension` in `@lexical/extension` that clears inline code formatting when the cursor is at the boundary of a code-formatted text node with no adjacent sibling                                                                         
  - Opt-in extension with configurable `triggers` (`'enter'`, `'click'`, `'arrow'`)
  - Exposes `registerEscapeFormatAtBoundary` for non-extension usage                                                                      
  - Removes the hardcoded code-format escape logic from `registerRichText` to preserve backwards compatibility
  - Adds the extension to the playground with all triggers enabled as a reference implementation                                                                                    
                                                     
Closes #4936

### Before

https://github.com/user-attachments/assets/531db7cd-b2f3-4dca-bff5-50dd0487c42d

### After

https://github.com/user-attachments/assets/fced2673-31fe-4a85-86c7-0b8f6f9e6675


etrepum (commented):
I don't think that this makes sense as a default given that it breaks backwards compatibility but it could be an option to RichTextExtension or its own extension

levensta:
I support the idea of ​​adding this option to the extension. Besides, why not do this for any other formats?

A long time ago, I created a workaround like this for my plugin
```js
editor.registerCommand(
  INSERT_PARAGRAPH_COMMAND,
  () => {
      selection.format = 0;
    }
    return false;
  },
  COMMAND_PRIORITY_HIGH
)
```

levensta:
Also I'm not sure that it need to handle clicking or right/left keys to exit the format. Are there any editors with a similar UX?

vishisht31:
Thanks for the feedback! I want to understand the backwards compatibility concern better as this is a bug fix for the case (https://github.com/facebook/lexical/issues/4936) where inline code formatting "sticks" when the cursor moves past a code node boundary.
 cc- @etrepum  @levensta

etrepum:
These aren’t code nodes, they’re just text nodes with code formatting. Not really any different than bold or italic. Trying to make it behave differently is not backwards compatible. Offering an extension or configuration of the RichTextExtension to alter this behavior makes sense, but just changing how it works for everyone with no opt-in doesn’t really make sense.

packages/lexical-extension/src/EscapeFormatAtBoundaryExtension.ts
```diff
@@ -0,0 +1,171 @@
...
+  safeCast,
+} from 'lexical';
+
+import {namedSignals} from './namedSignals';
+import {effect} from './signals';
+
+export interface EscapeFormatAtBoundaryConfig {
+  disabled: boolean;
+  formats: TextFormatType[];
+  triggers: Array<'enter' | 'click' | 'arrow'>;
```
etrepum: there should be a Triggers type so that this isn't repeated several times, also it is typescript (and modern flow) convention to use `T[]` rather than `Array<T>`

Would be good to add some API documentation here too

packages/lexical-extension/src/__tests__/unit/EscapeFormatAtBoundaryExtension.test.ts
```diff
@@ -0,0 +1,442 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {registerEscapeFormatAtBoundary} from '@lexical/extension';
```
etrepum: I don't think we need to expose this at all, there's no reason not to upgrade to extensions. This should be testing `EscapeFormatAtBoundaryExtension` directly.

packages/lexical-extension/src/__tests__/unit/EscapeFormatAtBoundaryExtension.test.ts
```diff
@@ -0,0 +1,442 @@
...
+} from 'lexical';
+import {
+  initializeUnitTest,
+  invariant,
+  KeyboardEventMock,
+} from 'lexical/src/__tests__/utils';
+import {describe, expect, test} from 'vitest';
+
+describe('EscapeFormatAtBoundaryExtension', () => {
+  initializeUnitTest((testEnv) => {
```
etrepum: Since these are extension tests they should be using extension infrastructure, e.g. `using editor = buildEditorFromExtensions(...)`

packages/lexical-extension/src/__tests__/unit/EscapeFormatAtBoundaryExtension.test.ts
```diff
@@ -0,0 +1,442 @@
...
+            textNode.getKey(),
+            textNode.getTextContentSize(),
+            'text',
+          );
+          selection.focus.set(
+            textNode.getKey(),
+            textNode.getTextContentSize(),
+            'text',
+          );
+          $setSelection(selection);
```
etrepum: This is a very long way to write `$getRoot().select()` (or selectEnd, which is the default behavior of select), the selections in the other tests can be created with similar methods (e.g. `textNode.select(3, 3)`)

packages/lexical-extension/src/EscapeFormatAtBoundaryExtension.ts
```diff
@@ -0,0 +1,171 @@
...
+    (direction === 'end' && atEnd) ||
+    (direction === 'start' && atStart) ||
+    (direction === 'both' && (atEnd || atStart))
+  ) {
+    selection.setFormat(0);
+    selection.setStyle('');
+  }
+}
+
+export function registerEscapeFormatAtBoundary(
```
etrepum: No need to export implementation details

packages/lexical-extension/src/EscapeFormatAtBoundaryExtension.ts
```diff
@@ -0,0 +1,171 @@
...
+  editor: LexicalEditor,
+  formats: TextFormatType[],
+  triggers: Array<'enter' | 'click' | 'arrow'>,
```
etrepum: This could be normalized to something like `{ [k in Trigger]?: boolean }`

etrepum (commented):
Looks like it's getting close, the title should be changed to Feature since this is not a bug fix

packages/lexical-extension/src/EscapeFormatAtBoundaryExtension.ts
```diff
@@ -0,0 +1,192 @@
...
+ * - `arrow`: Escape on arrow key navigation (left/right)
+ */
+export type EscapeFormatTrigger = 'enter' | 'click' | 'arrow';
+
+/**
+ * Configuration for {@link EscapeFormatAtBoundaryExtension}.
+ *
+ * @property disabled - When true, the extension is inactive.
+ * @property formats - Text format types to escape at boundaries (e.g. `['code']`).
+ * @property triggers - Which user interactions trigger the escape behavior.
```
etrepum: The default values should be documented

etrepum (commented):
After looking a bit more closely at this I think the configuration should probably be more flexible since you may want to have different triggers for different keys. For example the current `@lexical/rich-text` has a `$resetCapitalization` which is basically this but for `['lowercase', 'uppercase', 'capitalize']` and `['enter', 'space', 'tab']`. We could embed this into the existing RichTextExtension since that functionality is already mostly there.

Possibly a way to configure this would be something like:

```ts
export interface TriggerConfig {
  [K in EscapeFormatTrigger]:? boolean;
}
export interface EscapeFormatTriggerConfig {
  [K in TextFormatType]?: null | TriggerConfig;
}
```

```
escapeFormatTriggers: {
  lowercase: {enter: true, space: true, tab: true},
  uppercase: {enter: true, space: true, tab: true},
  capitalize: {enter: true, space: true, tab: true},
  code: {enter: true},
},
```

This structure would also be suitable for a straightforward mergeConfig implementation where overrides can turn anything on or off.

vishisht31:
Thanks for the feedback! Just want to make sure I'm on the same page before I start working on this:
  1. Restructure the config Instead of having formats and triggers as separate fields, combine them into a single escapeFormatTriggers map where each format gets its own triggers.                                                                                            
  2. Add a mergeConfig So that partial overrides don't blow away the whole config. Like if someone only passes {escapeFormatTriggers:  {code: {arrow: true}}}, it should merge into the defaults instead of replacing everything.                                              
  3. Document the defaults in JSDoc.                        
                                                                                                                                          
  One thing I'd prefer keeping this as a standalone extension rather than putting it inside RichTextExtension. Feels like it makes more sense as something you opt into. But if you'd rather have it there, I'm open to that too.                                               
                                                                                                                                          
  Let me know if I'm understanding this right!   
  CC : @etrepum

etrepum:
It would be fine to put it in RichTextExtension because it overlaps with the $resetCapitalization code, the opt-in would simply be configuring it differently.

If you want to keep it separate for now, that’s fine too, but it will probably end up in RichTextExtension eventually one way or another.

packages/lexical-rich-text/src/index.ts
```diff
@@ -556,19 +557,157 @@ function $isSelectionCollapsedAtFrontOfIndentedBlock(
...
+  if (didEscapeBoundary) {
+    selection.setStyle('');
+  }
+}
+
+function mergeRichTextConfig(
+  config: RichTextConfig,
+  overrides: Partial<RichTextConfig>,
+): RichTextConfig {
+  if (overrides.escapeFormatTriggers === undefined) {
```
etrepum: Generally this should be using `shallowMergeConfig` and then merging nested configuration as needed. It doesn't make a difference now but as soon as other config is added this probably doesn't work

etrepum (approved):

PR #8385, opened 2026-04-23
[lexical-react] Bug Fix: Fall back to root node in NodeContextMenuPlugin when click target has no Lexical node

### LINKED ISSUE #8257: NodeContextMenuPlugin: empty menu when DOM target has no Lexical node; Paste unavailable
## Description

`NodeContextMenuPlugin` from `@lexical/react` always calls `preventDefault()` on `contextmenu` and opens the floating UI, but the list of actions is built only when `$getNearestNodeFromDOMNode(event.target)` returns a node. If that call returns `null` (common when right-clicking the placeholder, padding, or other DOM that does not map to a Lexical node), **no menu items are rendered** while the overlay/popover still appears—effectively an empty menu. Users also lose access to **Paste** / **Paste as plain text** in those cases, even though those actions do not require a resolved node.

## Current behavior

1. Right-click on text or blocks that map to a Lexical node → menu shows Copy, Cut, Paste, etc. ✓  
2. Right-click on “empty” editor chrome (e.g. placeholder element, or targets where `e.target` does not map via `$getNearestNodeFromDOMNode`) → custom menu shell opens but **has zero items**; native context menu is suppressed.

## Expected behavior

Either:

- **A)** Do not call `preventDefault()` / do not open the floating menu when the filtered item list is empty (let the browser show its native context menu), **or**  
- **B)** When no node is resolved, still show a sensible subset (at least **Paste** and **Paste as plain text**), and/or resolve the node using `event.composedPath()` so more right-clicks map to a node.

## Reproduction

1. Use `LexicalComposer` + `RichTextPlugin` + `NodeContextMenuPlugin` with default `NodeContextMenuOption` items (Copy, Cut, Paste, …).  
2. Right-click directly on the **placeholder** (or other non-mapped surface inside the root).  
3. Observe: floating UI opens with no buttons; native menu does not appear.

## Environment

- `@lexical/react`: 0.42.x (behavior observed in `LexicalNodeContextMenuPlugin` implementation)  
- Browser: Chromium / Safari / Firefox (behavior is logical from source, not browser-specific)

## Technical note

In `packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx`, `visibleItems` stays empty when `node` is falsy, but the handler still runs `preventDefault()`, builds `renderableItems` from the empty list, and sets the menu open—so the UI opens with no `menuitem`s.

## Suggested direction

1. Early-return before `preventDefault()` when `visibleItems.length === 0`, **or**  
2. When `node == null`, fall back to clipboard-oriented options (Paste / plain-text paste), **or**  
3. Walk `event.composedPath()` (with fallback to `event.target`) to find the first DOM node that maps to a Lexical node before giving up.

Happy to help test a PR if useful.

mayrang:
## Description

Fixes #8257

`NodeContextMenuPlugin` always calls `preventDefault()` on `contextmenu` events, but when right-clicking on "empty" editor chrome (placeholder, padding, or DOM that doesn't map to a Lexical node), `$getNearestNodeFromDOMNode` returns `null`. This results in an empty custom menu while the native context menu is suppressed.

As suggested by @etrepum in #8257, this falls back to `$getRoot()` when no node is found:

```diff
- const node = $getNearestNodeFromDOMNode(e.target as Element);
+ const node = $getNearestNodeFromDOMNode(e.target as Element) ?? $getRoot();
```

This means:
- Right-clicking on text/blocks works exactly as before (`$getRoot()` fallback is never reached)
- Right-clicking on empty areas now uses the root node for `$showOn` filtering, so items without `$showOn` (e.g. Paste) are shown instead of an empty menu
- No changes to the plugin's public API

## Test Plan

1. Open the Lexical playground
2. Right-click on text inside the editor → context menu shows items as before
3. Right-click on the placeholder or empty padding below content → **Before:** empty menu appears, native menu suppressed. **After:** menu shows applicable items (e.g. Paste)

### DIFF
--- a/packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx
@@ -208,7 +208,8 @@ const NodeContextMenuPlugin = forwardRef<
       let visibleItems: ContextMenuType[] = [];
       if (items) {
         editor.read(() => {
-          const node = $getNearestNodeFromDOMNode(e.target as Element) ?? $getRoot();
+          const node =
+            $getNearestNodeFromDOMNode(e.target as Element) ?? $getRoot();
           if (node) {
             visibleItems = items!.filter((option) =>
               option.$showOn ? option.$showOn(node) : true,
--- a/packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx
+++ b/packages/lexical-react/src/LexicalNodeContextMenuPlugin.tsx
@@ -22,7 +22,7 @@ import {
   useTypeahead,
 } from '@floating-ui/react';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
-import {$getNearestNodeFromDOMNode, LexicalNode} from 'lexical';
+import {$getNearestNodeFromDOMNode, $getRoot, LexicalNode} from 'lexical';
 import {forwardRef, JSX, RefObject, useEffect, useRef, useState} from 'react';
 
 class MenuOption {
@@ -208,7 +208,7 @@ const NodeContextMenuPlugin = forwardRef<
       let visibleItems: ContextMenuType[] = [];
       if (items) {
         editor.read(() => {
-          const node = $getNearestNodeFromDOMNode(e.target as Element);
+          const node = $getNearestNodeFromDOMNode(e.target as Element) ?? $getRoot();
           if (node) {
             visibleItems = items!.filter((option) =>
               option.$showOn ? option.$showOn(node) : true,


etrepum (commented):
Failing pnpm run ci-check

mayrang:
Sorry, missed running Prettier before pushing. Fixed now.

etrepum:
As a first time contributor your checks don't run until a maintainer clicks an approval button, adding empty commits won't help anything.

mayrang:
Got it, thanks for letting me know! I wasn't aware that first-time contributors need manual approval for CI. Sorry for the unnecessary commit.

etrepum (approved):

PR #8386, opened 2026-04-23
[lexical-playground] Feature: Format HTML conversion with prettier

etrepum:
## Description

Format the "Convert to HTML" view with prettier, while being careful to only format it if the content hasn't changed in the meantime.

Follow-up to #8379

## Test plan

New e2e test and existing specs changed to ignore the data-gutter attribute

### DIFF
--- a/packages/lexical-playground/src/plugins/ActionsPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/ActionsPlugin/index.tsx
@@ -49,6 +49,7 @@ import useFlashMessage from '../../hooks/useFlashMessage';
 import useModal from '../../hooks/useModal';
 import Button from '../../ui/Button';
 import {docFromHash, docToHash} from '../../utils/docSerialization';
+import {formatCodeWithPrettier} from '../CodeActionMenuPlugin/components/PrettierButton';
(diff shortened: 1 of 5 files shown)


packages/lexical-playground/src/plugins/ActionsPlugin/index.tsx
```diff
@@ -252,6 +278,15 @@ export default function ActionsPlugin({
...
         setMode('html');
+        if (html) {
+          formatCodeWithPrettier(html, 'html')
+            .then((formattedCode) =>
+              editor.update(() =>
+                handlePrettierResponse(editor, html, formattedCode),
+              ),
+            )
+            .catch((err: Error) => console.error(err));
+        }
```
levensta: It would be best to apply the formatting to the line first, and then insert it into CodeNode. Even with relatively little content, you can see the content jumping around

https://github.com/user-attachments/assets/52916fca-c4b0-4d83-bd2b-dabb7964c291

zurfyx (approved):

packages/lexical-playground/src/plugins/ActionsPlugin/index.tsx
etrepum: I added some useActionState and useOptimistic code to format once but still have the UI make sense

PR #8387, opened 2026-04-24
[lexical-playground] Bug Fix: Prevent floating link editor from overflowing bottom of editor bounds

### LINKED ISSUE #8362: Bug: Floating link editor can overflow editor bounds and be clipped
When you create a link, and when you put the cursor inside a link, a floating link editor appears below the link so you can edit the link destination URL. While this looks like a popup layered over the editor, in the DOM structure it's inside the editor, and if the link is at the bottom of the document, it overflows the editor bounds and gets clipped.

Lexical screenshot playground, showing it both being clipped and also having the wrong Z-index/layering:

Image
 
In Lexical Playground, there is just enough bottom-padding for the popup to be usable anyways (though it does look visually broken). In our (LW's) comment box, we have less bottom-padding, and we added things to the floating link editor that made it taller, so it's more fully broken (the buttons are entirely in the clipped region).

There are two approaches to fixing this. One is to detect when this is happening and flip it from below to above. The other is to place it somewhere else in the DOM, outside the editor. The latter is more complicated but does have precedent; `CommentPlugin_CommentInputBox` works that way, for example.

mayrang:
## Description

Fixes #8362

When a link is near the bottom of the editor, the floating link editor popup overflows the editor scroller's bottom edge and gets clipped. This happens because `setFloatingElemPositionForLinkEditor` checks for top and right boundary collisions but not bottom.

This adds a bottom boundary check that mirrors the existing top boundary check — when the popup would overflow the bottom edge, it flips above the link instead:

```diff
+ if (top + floatingElemRect.height > editorScrollerRect.bottom) {
+   top -= floatingElemRect.height + targetRect.height + verticalGap * 2;
+ }
```

## Test Plan

1. Open the Lexical playground
2. Add text and insert a link near the bottom of the editor
3. Click the link
4. **Before:** Floating link editor overflows below the editor bounds and is clipped
5. **After:** Floating link editor flips above the link and stays within bounds

### DIFF
--- a/packages/lexical-playground/src/utils/setFloatingElemPositionForLinkEditor.ts
+++ b/packages/lexical-playground/src/utils/setFloatingElemPositionForLinkEditor.ts
@@ -38,6 +38,10 @@ export function setFloatingElemPositionForLinkEditor(
     left = editorScrollerRect.right - floatingElemRect.width - horizontalOffset;
   }
 
+  if (top + floatingElemRect.height > editorScrollerRect.bottom) {
+    top -= floatingElemRect.height + targetRect.height + verticalGap * 2;
+  }
+
   top -= anchorElementRect.top;
   left -= anchorElementRect.left;
 


etrepum (commented):
It looks like this does it based on editor bounds but not also screen bounds so it might be worth fixing that too, or maybe just refactoring the link editor to use `@floating-ui/react` if it doesn't already.

mayrang (commented):
Thanks for the review! I think manually patching each edge case isn't sustainable. I'd prefer refactoring to `@floating-ui/react` if that direction works for you — or would you rather I just add a viewport bounds check on top of this for now?

etrepum:
Refactoring for `@floating-ui/react` is the best option, the only reason it isn't done is because nobody has gotten around to it.

mayrang:
I'll open a new PR that refactors the link editor to use `@floating-ui/react` instead. Closing this one.

PR #8390, opened 2026-04-25
[lexical-list] Bug Fix: Toggle checklist items on mobile tap

### LINKED ISSUE #7651: Bug: Unable to check checkboxes on iPhone

jWA86:
## Description

You can't tap to toggle a checklist item on iOS Safari or Android Chrome. The bug is reproducible in the [Lexical playground](https://playground.lexical.dev/) and in Chrome DevTools mobile emulation.

The cause: the `touchstart` listener registered by the CheckList plugin calls `event.preventDefault()` to keep the caret away from the marker, which on mobile also cancels the synthesized `click`, and the click is what `handleClick` is wired to. So the toggle path is never reached. PR #7655 widened the touch hit-test, but the click that the hit-test was meant to handle was already gone by then, so taps still did nothing.

The fix is to also listen for `pointerup` and run the toggle from there when `event.pointerType === 'touch'`. To stay safe on browsers where the click does eventually fire, a 500 ms dedup window swallows the second event so we never toggle twice. Desktop click is left alone.

While I was in there, I widened the handler signatures from `MouseEvent | TouchEvent` to `PointerEvent | MouseEvent | TouchEvent` (they were already accepting pointer events at runtime, this just makes it explicit) and replaced the `(event as PointerEvent).pointerType` cast inside `handleCheckItemEvent` with a `'pointerType' in event` guard.

Closes #7651

### Before

Tap a checklist marker in the playground on iOS Safari / Android Chrome / Chrome DevTools mobile emulation : nothing happens.

The new `packages/lexical-list/src/__tests__/unit/checkList.test.tsx` file fails on `main`:

:red_circle:  touch pointerup over the marker area toggles the item
:red_circle: : pointerup followed by a synthesized click does not double-toggle
:green_circle:  mouse pointerup is ignored (click stays the desktop path)

### After

Same tap toggles the checkbox.

The three new tests pass, and the rest of the `lexical-list` suite still does (86/86). `pnpm tsc` and `pnpm lint` are clean.

### DIFF
--- a/packages/lexical-list/src/__tests__/unit/checkList.test.tsx
+++ b/packages/lexical-list/src/__tests__/unit/checkList.test.tsx
@@ -1,0 +1,171 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {$getRoot} from 'lexical';
+import {initializeUnitTest} from 'lexical/src/__tests__/utils';
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  test,
+  vi,
+} from 'vitest';
+
+import {$createListItemNode, $createListNode} from '../..';
+import {registerCheckList} from '../../checkList';
+
+// Polyfill PointerEvent for jsdom (mirrors LexicalTableMobileSelection.test.tsx).
+interface PointerEventInit extends EventInit {
+  clientX?: number;
+  clientY?: number;
+  pointerType?: string;
+}
+
+(global as unknown as {PointerEvent: unknown}).PointerEvent =
+  class PointerEventPolyfill extends Event {
+    clientX: number;
+    clientY: number;
+    pointerType: string;
+
+    constructor(type: string, options: PointerEventInit = {}) {
+      super(type, options);
+      this.clientX = options.clientX || 0;
+      this.clientY = options.clientY || 0;
+      this.pointerType = options.pointerType || 'mouse';
+    }
+  };
+
+describe('registerCheckList — mobile tap toggle', () => {
+  // jsdom does not implement getComputedStyle with pseudo-elements (see
+  // https://github.com/jsdom/jsdom/issues/1928), but the checklist hit-test
+  // reads the ::before width to size the marker area. Stub a 16px width so
+  // the bounds check resolves to a concrete number; vi.restoreAllMocks
+  // returns the original implementation afterwards so the mock cannot leak
+  // to other test files in the same worker.
+  const originalGetComputedStyle = window.getComputedStyle;
+  beforeAll(() => {
+    vi.spyOn(window, 'getComputedStyle').mockImplementation(
+      (element: Element, pseudoElement?: string | null) => {
+        if (pseudoElement === '::before') {
+          return {width: '16px'} as unknown as CSSStyleDeclaration;
+        }
+        return originalGetComputedStyle.call(window, element, pseudoElement);
+      },
+    );
+  });
+  afterAll(() => {
+    vi.restoreAllMocks();
+  });
+
+  initializeUnitTest((testEnv) => {
+    let cleanup: (() => void) | null = null;
+
+    beforeEach(async () => {
+      cleanup = registerCheckList(testEnv.editor);
+      await testEnv.editor.update(
+        () => {
+          const list = $createListNode('check');
+          list.append($createListItemNode(false));
+          $getRoot().clear().append(list);
+        },
+        {discrete: true},
+      );
+    });
+
+    afterEach(() => {
+      cleanup?.();
+      cleanup = null;
+    });
+
+    function getCheckListItem(): HTMLLIElement {
+      const li = testEnv.container.querySelector('li');
+      if (!li) {
+        throw new Error('No <li> rendered');
+      }
+      return li as HTMLLIElement;
+    }
+
+    function isChecked(li: HTMLLIElement): boolean {
+      return li.getAttribute('aria-checked') === 'true';
+    }
+
+    test('touch pointerup over the marker area toggles the item', async () => {
+      const li = getCheckListItem();
+      expect(isChecked(li)).toBe(false);
+
+      // clientX is well inside the marker hit area for any plausible
+      // ::before width / touch padding combination.
+      li.dispatchEvent(
+        new PointerEvent('pointerup', {
+          bubbles: true,
+          cancelable: true,
+          clientX: 10,
+          clientY: 5,
+          pointerType: 'touch',
+        }),
+      );
+
+      // editor.update() reconciles asynchronously — wait for the DOM rather
+      // than assuming a fixed number of microtasks.
+      await vi.waitFor(() => expect(isChecked(li)).toBe(true));
+    });
+
+    test('mouse pointerup is ignored (click stays the desktop path)', () => {
+      const li = getCheckListItem();
+      expect(isChecked(li)).toBe(false);
+
+      li.dispatchEvent(
+        new PointerEvent('pointerup', {
+          bubbles: true,
+          cancelable: true,
+          clientX: 10,
+          clientY: 5,
+          pointerType: 'mouse',
+        }),
+      );
+
+      expect(isChecked(li)).toBe(false);
+    });
+
+    test('pointerup followed by a synthesized click does not double-toggle', async () => {
+      const li = getCheckListItem();
+      expect(isChecked(li)).toBe(false);
+
+      const baseTimeStamp = 1000;
+      const pointerUp = new PointerEvent('pointerup', {
+        bubbles: true,
+        cancelable: true,
+        clientX: 10,
+        clientY: 5,
+        pointerType: 'touch',
+      });
+      Object.defineProperty(pointerUp, 'timeStamp', {value: baseTimeStamp});
+      li.dispatchEvent(pointerUp);
+      await vi.waitFor(() => expect(isChecked(li)).toBe(true));
+
+      // Browsers where touchstart preventDefault did not suppress click will
+      // also fire it shortly after. The dedup window must absorb it.
+      const click = new MouseEvent('click', {
+        bubbles: true,
+        cancelable: true,
+        clientX: 10,
+        clientY: 5,
+      });
+      Object.defineProperty(click, 'timeStamp', {value: baseTimeStamp + 50});
+      li.dispatchEvent(click);
+
+      // Give any queued reconciliation a chance to run, then assert the
+      // state has not flipped back.
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      expect(isChecked(li)).toBe(true);
+    });
+  });
+});
--- a/packages/lexical-list/src/__tests__/unit/checkList.test.tsx
+++ b/packages/lexical-list/src/__tests__/unit/checkList.test.tsx
@@ -6,8 +6,17 @@
  *
  */
 
+import {buildEditorFromExtensions, defineExtension} from '@lexical/extension';
+import {
+  $createListItemNode,
+  $createListNode,
+  $isListItemNode,
+  $isListNode,
+  CheckListExtension,
+} from '@lexical/list';
+import {RichTextExtension} from '@lexical/rich-text';
 import {$getRoot} from 'lexical';
-import {initializeUnitTest} from 'lexical/src/__tests__/utils';
+import {invariant} from 'lexical/src/__tests__/utils';
 import {
   afterAll,
   afterEach,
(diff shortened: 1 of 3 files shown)


packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+    constructor(type: string, options: PointerEventInit = {}) {
+      super(type, options);
+      this.clientX = options.clientX || 0;
+      this.clientY = options.clientY || 0;
+      this.pointerType = options.pointerType || 'mouse';
+    }
+  };
+
+describe('registerCheckList — mobile tap toggle', () => {
```
etrepum: it would be cleaner to write these tests with newer extension infrastructure, you can look at ListExtension.test.ts for an example that brings in lists.

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  test,
+  vi,
+} from 'vitest';
+
+import {$createListItemNode, $createListNode} from '../..';
+import {registerCheckList} from '../../checkList';
```
etrepum: since these are public it's better to import from the package (but should be using CheckListExtension)

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {$getRoot} from 'lexical';
+import {initializeUnitTest} from 'lexical/src/__tests__/utils';
```
etrepum: don't really need this for new tests

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+    clientY: number;
+    pointerType: string;
+
+    constructor(type: string, options: PointerEventInit = {}) {
+      super(type, options);
+      this.clientX = options.clientX || 0;
+      this.clientY = options.clientY || 0;
+      this.pointerType = options.pointerType || 'mouse';
+    }
+  };
```
etrepum: would be better to polyfill this only if needed, maybe just move it to vitest.setup.mts with most of the other jsdom polyfills

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+      const pointerUp = new PointerEvent('pointerup', {
+        bubbles: true,
+        cancelable: true,
+        clientX: 10,
+        clientY: 5,
+        pointerType: 'touch',
+      });
+      Object.defineProperty(pointerUp, 'timeStamp', {value: baseTimeStamp});
+      li.dispatchEvent(pointerUp);
+      await vi.waitFor(() => expect(isChecked(li)).toBe(true));
```
etrepum:
```suggestion
      await li.dispatchEvent(pointerUp);
      expect(isChecked(li)).toBe(true);
```

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+        cancelable: true,
+        clientX: 10,
+        clientY: 5,
+      });
+      Object.defineProperty(click, 'timeStamp', {value: baseTimeStamp + 50});
+      li.dispatchEvent(click);
+
+      // Give any queued reconciliation a chance to run, then assert the
+      // state has not flipped back.
+      await new Promise((resolve) => setTimeout(resolve, 0));
```
etrepum:
```suggestion
      await li.dispatchEvent(click);
```

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+        new PointerEvent('pointerup', {
+          bubbles: true,
+          cancelable: true,
+          clientX: 10,
+          clientY: 5,
+          pointerType: 'mouse',
+        }),
+      );
+
+      expect(isChecked(li)).toBe(false);
```
etrepum: this probably isn't really testing anything because the update based on that event isn't discrete and wouldn't have reconciled to the DOM yet, there needs to be an await before this

```suggestion
      await li.dispatchEvent(
        new PointerEvent('pointerup', {
          bubbles: true,
          cancelable: true,
          clientX: 10,
          clientY: 5,
          pointerType: 'mouse',
        }),
      );

      expect(isChecked(li)).toBe(false);
```

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+          cancelable: true,
+          clientX: 10,
+          clientY: 5,
+          pointerType: 'touch',
+        }),
+      );
+
+      // editor.update() reconciles asynchronously — wait for the DOM rather
+      // than assuming a fixed number of microtasks.
+      await vi.waitFor(() => expect(isChecked(li)).toBe(true));
```
etrepum: Throwing an `await` on the dispatchEvent line should be sufficient to ensure that any code that happened as a result of that dispatch has run already. Event dispatch is synchronous, lexical is going to do the update of its state synchronously, and then schedule a microtask to reconcile it to the DOM which will complete before the await resumes.

Using an await is better than polling because polling can mask other bugs (or at least incorrect architecture) where updates cause cascading updates after they are reconciled which is not how things should be built.

```suggestion
      await li.dispatchEvent(
        new PointerEvent('pointerup', {
          bubbles: true,
          cancelable: true,
          clientX: 10,
          clientY: 5,
          pointerType: 'touch',
        }),
      );

      expect(isChecked(li)).toBe(true));
```

packages/lexical-list/src/checkList.ts
```diff
@@ -59,10 +59,41 @@ export function registerCheckList(
...
 
-  const configHandleClick = (event: MouseEvent | TouchEvent) => {
+  // Mobile tap fix: the touchstart listener registered below calls
+  // event.preventDefault() to keep the caret away from the marker. On iOS
+  // Safari and Android Chrome that suppression also cancels the synthesized
+  // click, so handleClick never runs and the checkbox cannot be toggled by
+  // tap. We additionally listen for pointerup with pointerType === 'touch'
+  // and run the same toggle logic, deduplicating against any click that
+  // does fire on browsers where preventDefault doesn't suppress it.
+  let lastHandledTimeStamp = 0;
```
etrepum: Would it make more sense to store this on the element itself so you're not prevented from tapping or clicking on several checkboxes really fast?

etrepum (commented):
This is directionally good, I think the tests could be cleaned up and maybe the tests should explicitly test clicking on multiple checkboxes in a short amount of time because I think this global state would make that not work very well.

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Removed in 13e7f91 : migrated to `buildEditorFromExtensions`.

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Switched to `CheckListExtension` from `@lexical/list` 13e7f91

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Migrated to `defineExtension({ dependencies: [CheckListExtension, RichTextExtension], etc... }))` in 13e7f91

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Moved in 13e7f91, gated on the existing `isJsdom` check next to the other jsdom workarounds. Carries `button`, `buttons`, `clientX`, `clientY`, `pointerType` so it covers `LexicalTableMobileSelection.test.tsx` too.

packages/lexical-list/src/checkList.ts
jWA86: Done in 13e7f91 : switched to `WeakMap<EventTarget, number>` keyed by `event.target`. Added a regression test that taps two `<li>` 100ms apart and asserts both toggle (would fail under the previous global timestamp).

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Took a slightly different path in 13e7f91: instead of `await dispatchEvent` + DOM read, the migrated test reads via `editor.read(() => item.getChecked())`. The in-memory toggle is observable synchronously after `dispatchEvent` returns (`before=false sync=true micro=true macro=true`) : DOM reconciliation is on the microtask but state commit isn't, so `editor.read()` doesn't need a flush.

This still has the property you wanted (no polling, deterministic), and it sidesteps DOM reconciliation entirely so it's robust to whatever timing changes happen in the reconciler. Happy to switch to your suggested `await dispatchEvent` + `isChecked(li)` pattern instead if you'd rather keep the test going through the DOM contract.

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Now reads via `editor.read()`, so the mouse-pointerup test actually verifies the negative.

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Covered by the `editor.read()` migration in 13e7f913f3a96e755342ab6690c234733dfac9e0

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
jWA86: Covered by the editor.read() migration in 13e7f913f3a96e755342ab6690c234733dfac9e0

packages/lexical-list/src/checkList.ts
etrepum: I would just write it directly to the element. You can use a symbol if you want to hide it away but lexical already sets several properties on HTMLElements it manages

packages/lexical-list/src/__tests__/unit/checkList.test.tsx
etrepum: editor.read simply forces a reconciliation if one is pending, so any work that would happen in that microtask is executed synchronously before it calls your callback. DOM reconciliation is happening, it’s not sidestepped. It’s equivalent in everything but scheduling to using await, which is fine.

packages/lexical-list/src/checkList.ts
```diff
@@ -59,10 +59,52 @@ export function registerCheckList(
...
+  // event.preventDefault() to keep the caret away from the marker. On iOS
+  // Safari and Android Chrome that suppression also cancels the synthesized
+  // click, so handleClick never runs and the checkbox cannot be toggled by
+  // tap. We additionally listen for pointerup with pointerType === 'touch'
+  // and run the same toggle logic, deduplicating against any click that
+  // does fire on browsers where preventDefault doesn't suppress it.
+  //
+  // Dedup state is per-target: a global window would block tapping a second
+  // checkbox within 500ms of toggling the first.
+  const lastHandledByTarget = new WeakMap<EventTarget, number>();
```
etrepum: I think this would be a bit cleaner and maybe slightly more debuggable if the property was stored directly on the HTMLElement, either with a symbol or a prefixed property. Plenty of other examples of doing this to elements that the editor manages in the codebase, `__lexicalEditor` on root elements, `__lexicalListType` on all checklists, `__lexicalTableSelection` on selected tables, etc.

etrepum (approved):

packages/lexical-list/src/checkList.ts
jWA86: Done in e3ad7fb : switched to `__lexicalCheckListLastHandled` set directly on the event target (with `// @ts-ignore internal field` ).

packages/lexical-list/src/checkList.ts
jWA86: Done e3ad7fbd2491cebb6611f1fc70b30163fc00bd79

etrepum (approved):

PR #8392, opened 2026-04-26
[lexical-playground] Bug Fix: Show floating link editor for single-character links

### LINKED ISSUE #8184: Bug: Selection does not include LinkNode when it contains only one character
Lexical version: 0.41.0

## Steps To Reproduce

1. Create LinkNode with only one character.
2. Insert text before LinkNode and after LinkNode.
3. Move caret, the edit popper does not show.

## The current behavior

https://github.com/user-attachments/assets/a8ed1393-16a3-479d-ab28-6abd895ab1e3

mayrang:
## Description

Fixes #8184

When a link contains only one character, the floating link editor popup never appeared when navigating with arrow keys. The cursor always lands on a node boundary where the browser reports it as outside the link.

Added a `$getAdjacentLinkNode` helper that applies right bias for collapsed selections, as suggested by @etrepum in #8184: when the cursor is at the end of a TextNode and the next sibling is a LinkNode, treat it as being on the link. Applied in `$updateToolbar`, `$updateLinkEditor` (URL), and `$updateLinkEditor` (position).

## Test Plan

1. Type "hello A world", select "A", insert a link
2. Move cursor with arrow keys across the link
3. **Before:** popup never appears for the single-character link
4. **After:** popup appears when cursor enters the link boundary from the left

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
@@ -14,6 +14,7 @@ import {
   $createLinkNode,
   $isAutoLinkNode,
   $isLinkNode,
+  LinkNode,
   TOGGLE_LINK_COMMAND,
 } from '@lexical/link';
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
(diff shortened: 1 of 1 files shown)


etrepum (commented):
Would it clean things up in here if there was a `getSelectedLinkNode` helper function that did all of the checks (selection of link directly, parent is a link, right-biased sibling of text is a link) and returned the best candidate?

mayrang:
That's much cleaner.  refactored into a single $getSelectedLinkNode helper that handles all three cases. Force-pushed. Thanks!

etrepum (approved):

mayrang (commented):
(duplicate comment, please ignore)

Issue #4685, opened 2023-06-21
Bug: can't insert a paragraph above a code block at the beginning of the content


thorn0:
Lexical version: The current version on the Playground. `main`? 0.11.1? 

## Steps To Reproduce

1. open the [playground](https://playground.lexical.dev/)
2. place the cursor at the beginning of the content
3. create a code block with any content there

## The current behavior

There should be a way to add a paragraph above the code block

## The expected behavior

There seems to be no way to do that.

tarngerine:
there technically is, but the cursor position does not accurately update when you're focused on the root at pos 0

https://github.com/facebook/lexical/assets/1629901/0563d078-258f-4e6a-beab-f51dda0568d4

levensta:
Closed by https://github.com/facebook/lexical/pull/8393

@etrepum fyi

PR #8393, opened 2026-04-26
[lexical-code-core][lexical-playground] Feature: Create paragraph around the code node when navigating with the arrow keys

### LINKED ISSUE #7912: Feature: create empty paragraph before/after code block if it's the first/last node

levensta:
### lexical-utils

Handlers `$onEscapeDown` and `$onEscapeUp` have been added to create paragraphs when the cursor is at the end of an element and that element is the last one in the root. Use these handlers for arrow key commands.

### lexical-code-core

The `CodeIndentExtension` now includes an `escapeWithArrows` option (default: false), which improves keyboard usability. When this option is enabled, if CodeNode is the terminal node and the cursor is at the end of the text, pressing the arrow keys will create a new paragraph and move the cursor to it

Closes #7912 #4685 #5968

### Before

### After


etrepum:
This should probably be a configuration option built into a signal for registration because it's not always desired behavior, e.g. when you have the playground in markdown or html mode you do not want to be able to exit that CodeNode at all.

levensta:
> This should probably be a configuration option built into a signal for registration because it's not always desired behavior, e.g. when you have the playground in markdown or html mode you do not want to be able to exit that CodeNode at all.

Currently, the playground [removes](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/ActionsPlugin/index.tsx#L222-L232) any non-CodeNode elements in Markdown/HTML mode, so I didn't add this option initially. But overall, it wouldn't hurt to have this option anyway — the only question is, should this behavior be the default? I would enable it by default because CodeExtension is a new extension and the current feature will not change the behavior of CodeNode unless you enable the extension

levensta:
The current feature is a direct copy of [CollapsibleContainer](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/CollapsibleExtension/index.ts#L50-L107). I also thought it might be a good idea to add similar key handling for table cells, but perhaps we should move the current handlers into generic functions in `lexical/utils`?

etrepum:
Moving shared code to `@lexical/utils` makes sense. I wouldn't necessarily consider the CodeExtension that new anymore, it's been around for half a year, people are may already depend on its current defaults.

etrepum:
Collapsible doesn't have to have any compatibility considerations because it's not a published package, it just lives in the playground

levensta:
I managed to achieve backward compatibility with the current behavior in CodeNode, and to demonstrate this in the current tests, I left the option disabled in the playground.

Additionally, in the current behavior with the option disabled, a bug has been fixed where the cursor is positioned at the end of a CodeNode but cannot move beyond it, even when the block has adjacent elements. e2e test has also been added for this

A commons handlers `$onEscapeDown` and `$onEscapeUp` has been added to lexical-utils and is reused in LayoutContainer in the playground. I think shadow root elements will be the most common use case for these functions

packages/lexical-code-core/src/CodeIndentation.ts
```diff
@@ -541,7 +572,8 @@ export function registerCodeIndentation(
           selection.isCollapsed() &&
           anchor.offset === 0 &&
           anchorNode.getPreviousSibling() === null &&
-          $isCodeNode(anchorNode.getParentOrThrow())
+          $isCodeNode(anchorNode.getParentOrThrow()) &&
+          anchorNode.getParentOrThrow().getPreviousSibling() === null
```
etrepum: anchorNode.getParentOrThrow() should be captured in a variable instead of computing it multiple times

packages/lexical-playground/src/plugins/CollapsibleExtension/index.ts
```diff
@@ -157,7 +158,7 @@ export const CollapsibleExtension = defineExtension({
       // new content even if trailing paragraph is accidentally deleted
       editor.registerCommand(
         KEY_ARROW_DOWN_COMMAND,
-        $onEscapeDown,
+        () => $onEscapeDown(),
```
etrepum: This wrapping seems pointless, unless we are bringing in the versions from `@lexical/utils`?

```suggestion
        $onEscapeDown,
```

etrepum (commented):
It might make sense to enable this in the playground to ensure that it's getting used, with some code to ensure it's not enabled when in markdown or html mode.

levensta:
I enabled the option in the playground and also removed one test that contradicted the behavior with the option enabled. Previously, it was clear that it passed, so the current changes do not break the previous behavior

packages/lexical-utils/src/index.ts
```diff
@@ -1001,3 +1001,87 @@ export function makeStateWrapper<K extends string, V>(
...
+    );
+
+    if (containerNode) {
+      const parent = containerNode.getParent();
+      if (parent !== null && parent.getLastChild() === containerNode) {
+        const contentParagraph = containerNode.getLastDescendant();
+        if (
+          contentParagraph !== null &&
+          selection.anchor.key === contentParagraph.getKey() &&
+          selection.anchor.offset === contentParagraph.getTextContentSize()
```
etrepum: The tests for this must not be very good because this is wrong in most cases, the offset in an element selection is a child index and has nothing to do with the text. If there was any text at the end of the node, the selection would be on the text and not the paragraph too.

packages/lexical-utils/src/index.ts
```diff
@@ -1001,3 +1001,87 @@ export function makeStateWrapper<K extends string, V>(
...
+      $isContainerNode,
+    );
+
+    if (containerNode) {
+      const parent = containerNode.getParent();
+      if (parent !== null && parent.getFirstChild() === containerNode) {
+        const contentParagraph = containerNode.getFirstDescendant();
+        if (
+          contentParagraph !== null &&
+          selection.anchor.key === contentParagraph.getKey()
```
etrepum: If there's any text at all generally the selection won't be on the paragraph

packages/lexical-code-core/src/CodeIndentation.ts
```diff
@@ -487,8 +490,36 @@ function $handleMoveTo(
 export function registerCodeIndentation(
   editor: LexicalEditor,
   tabSize?: number,
+  escapeWithArrows?: boolean,
 ): () => void {
   return mergeRegister(
+    // When node is the last child pressing down/right or up/let arrow will insert paragraph
+    // below it to allow adding more content.
+    // These handlers must be executed before $handleShiftLines
+    ...(escapeWithArrows
```
etrepum, line 514: There are already up/down handlers in here you could just add this code to them in the right place

levensta: I think it's better to keep the group of handlers for the `escapeWithArrows` parameter together and not mix the responsibilities of two different functions into one handler

Otherwise, it would look like this

```js
...(escapeWithArrows
  ? [
      editor.registerCommand(
        KEY_ARROW_RIGHT_COMMAND,
        () => $onEscapeDown($isCodeNode),
        COMMAND_PRIORITY_LOW,
      ),
      editor.registerCommand(
        KEY_ARROW_LEFT_COMMAND,
        () => $onEscapeUp($isCodeNode),
        COMMAND_PRIORITY_LOW,
      ),
    ]
  : []),
editor.registerCommand(
  KEY_ARROW_UP_COMMAND,
  event => {
    if (!event.altKey && escapeWithArrows) {
      $onEscapeUp($isCodeNode);
    }
    // ...
    return $handleShiftLines(KEY_ARROW_UP_COMMAND, event);
  },
  COMMAND_PRIORITY_LOW,
),
editor.registerCommand(
  KEY_ARROW_DOWN_COMMAND,
  event => {
    if (!event.altKey && escapeWithArrows) {
      $onEscapeUp($isCodeNode);
    }
    // ...
    return $handleShiftLines(KEY_ARROW_DOWN_COMMAND, event);
  },
  COMMAND_PRIORITY_LOW,
),
```

levensta:
For the record

The tests in Firefox were failing because the selection remained at the edge of the element, whereas it should have moved to the newly created paragraph after pressing the arrow key

Screenshot 2026-05-19 at 22 57 38
<br><br>
But in real Firefox (not Playwright), as well as in Chrome, the selection consistently moves to the new paragraph

Screenshot 2026-05-19 at 22 58 33

The essence of the fix is to explicitly call `selectEnd()` after passing the new paragraph to `insertAfter` / `insertBefore`

potatowagon:
## Review: Create Paragraph Around Code Node When Navigating with Arrow Keys

### Summary
Adds an `escapeWithArrows` config option to `CodeIndentExtension` that allows users to escape a code block using arrow keys when the cursor is at the beginning/end and the code block has no adjacent sibling elements. This creates a new paragraph above/below the code block for continued editing.

### What I Verified
- **Opt-in behavior**: `escapeWithArrows: false` by default — no change to existing behavior unless explicitly enabled.
- **Logic correctness**: Uses `$onEscapeDown`/`$onEscapeUp` from `@lexical/utils` (established patterns used elsewhere in the codebase). Arrow handlers only trigger when `escapeWithArrows` is enabled AND the code node has no sibling in the escape direction.
- **Fix for existing behavior**: Changed return value from `true` to `false` when the code node has no sibling — previously this would stop the event but not let it propagate to the new escape handlers. Now it correctly falls through.
- **Additional fix**: Added `parent.getPreviousSibling() === null` / `parent.getNextSibling() === null` checks to the UP/DOWN handlers so they only prevent navigation when the code block is truly at the document boundary (not just when cursor is at code boundary).
- **Alt+Arrow excluded**: `event.altKey` check prevents interference with line-moving shortcuts.
- **Test coverage**: 310-line unit test file covering ArrowDown, ArrowRight, ArrowUp, ArrowLeft escape scenarios + verification that escape doesn't trigger when adjacent elements exist.
- **CI status**: All e2e and integration tests pass on all platforms.

### Concern
- This is a UX behavior change that may need project-level consensus — arrow key semantics in code blocks are something users develop muscle memory for. However, since it's opt-in, this isn't blocking.

### Verdict
✅ **Safe to approve** — well-implemented opt-in feature with comprehensive tests and no regression risk for existing users.

packages/lexical-code-core/src/CodeIndentation.ts
```diff
@@ -621,6 +655,15 @@ export interface CodeIndentConfig {
    * this option.
    */
   tabSize: number | undefined;
+  /**
+   * When `true`, When set to true, this enables the ability to exit a code block
```
etrepum:
```suggestion
   * When `true`, this enables the ability to exit a code block
```

etrepum (commented):
Overall this looks good, but it doesn't seem to work with an empty code block

[pr playground link](https://lexical-playground-git-fork-levensta-escape-e87a5e-fbopensource.vercel.app/#doc=H4sIAAAAAAAAE5WOywqDMBBF_-Wug8S-1HxFwWXpIphpG0gzZRwFEf-96Kouu7qcxTncGRSisrTqleBmCLOu271iCkIZ7raDu0GIQp1GznB5SMngwfL2CgcYxBwoK5w10OlDcOg4EAxGkn5zyuX_xvZq11gMku-19SMFuLKq7aFuKntsLo1Bz4N0q3dNfnoKDzn82rDF6VxYLF8VhbEL_gAAAA)

potatowagon (commented):
## Review — Create Paragraph Around Code Node with Arrow Keys

### What I verified:

1. **Feature scope**: Adds arrow-key escape behavior for code blocks — pressing up/left at the start or down/right at the end of a code block creates a new paragraph to allow adding content before/after. Uses the existing `$onEscapeUp`/`$onEscapeDown` utilities from `@lexical/utils`.

2. **Implementation**: The `escapeWithArrows` parameter is added to `registerCodeIndentation` as an opt-in boolean. The handlers register at `COMMAND_PRIORITY_LOW` so they run before `$handleShiftLines`. The fix also correctly changes the return value in `$handleShiftLines` from `true` to `false` when at a boundary (so the command continues propagating to the escape handlers).

3. **Edge case handling**: Alt+arrow is excluded (for line-shift operations). The sibling-null check ensures paragraphs are only created when the code block is at the document boundary.

4. **CI status**: Full CI suite green (42 checks including all e2e tests across all platforms).

5. **Risk**: Opt-in behavior, no change to default code block navigation. Existing tests still pass.

_— via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review&utm_campaign=8393&ref=potatowagon) on behalf of potatowagon_

etrepum (approved):

PR #8395, opened 2026-04-27
[lexical-markdown] Feature: Add $convertSelectionToMarkdownString API

### LINKED ISSUE #6503: Feature: Get markdown content for selection
## Description

I'm working with the markdown module of lexical and I realized there's no way to get the markdown string from only the selection of the text. I mean, If I use `$convertToMarkdownString` method, it only takes the full content of the editor and returns all the text in markdown, but in case of only being interested in the markdown string for the selection, it is not possible.

Would it be possible to change the signature/method or add a new one to allow that functionality? Or is there any possibility to do it currently with some workaround? 

Thnaks!

mayrang:
## Description

Fixes #6503

Adds a new `$convertSelectionToMarkdownString` function that converts only the selected content to a markdown string. Previously, `$convertToMarkdownString` could only convert the entire editor content.

### What changed

- Added `$convertSelectionToMarkdownString(transformers, selection, shouldPreserveNewLines)` to `@lexical/markdown`
- Reuses existing markdown export logic (`exportTopLevelElements`, `exportChildren`, `exportTextFormat`) with selection filtering
- Handles partial text selection by slicing text at anchor/focus offsets
- Returns empty string for null or collapsed selections
- Works inside `editor.read()` — no node creation needed

### Usage

```ts
import {$convertSelectionToMarkdownString, TRANSFORMERS} from '@lexical/markdown';

editor.read(() => {
  const markdown = $convertSelectionToMarkdownString(TRANSFORMERS, $getSelection());
});
```

### Implementation approach

Rather than creating a separate export pipeline, this extends the existing `createMarkdownExport` with an optional `selection` parameter. When provided, nodes are filtered via `isSelected()` and text content is sliced at selection boundaries. This provides a cleaner API for the use case described by @etrepum in #6503, without requiring `editor.update()` or manual node reconstruction.

Open to alternative approaches if a different direction is preferred.

## Test Plan

- Full text selection → returns full markdown
- Partial text selection → returns sliced text
- Bold text selection → preserves formatting
- Null selection → returns empty string
- Collapsed selection → returns empty string
- Backward selection → correct slicing
- Multi-paragraph selection → includes both paragraphs
- List selection → preserves list markdown syntax


mayrang:
Not fully sure this is the right approach — feedback on the direction would be appreciated before going further.

etrepum:
I won't have time to read through the code tonight but if it is following the same approach that HTML and JSON export takes then it's probably the right direction. If it's some other way then maybe not.

mayrang:
Thanks — checked how HTML and JSON export handle this, they both use `$sliceSelectedTextNodeContent` with `'clone'` mode. My initial version did the slicing manually, which worked but missed token/segmented node handling. Switching to `$sliceSelectedTextNodeContent` to stay consistent and cover those cases.

etrepum (commented):
I think there is probably some additional work that should be done to support extractWithChild to support situations like a partial LinkNode selection. I think taking the 'html' path makes the most sense for markdown (not 'clone').

Note that main has changed its prettier config so it might be easiest to copy over the new config, run prettier, commit, and then merge to sync.

mayrang:
For `extractWithChild` — adding the check inside `exportChildren` gets messy with nested structures (e.g. link inside a list item), so I looked into restructuring the export like `$appendNodesToHTML` does. Problem is that changes the return type of `exportChildren` from string to object, which would break `$convertToMarkdownString` and any external transformers using the current callbacks.

Thinking the safest route is building the recursive structure separately for `$convertSelectionToMarkdownString`, keeping the existing export untouched. Some duplication (~50-70 lines) but no regression risk. Does that work, or do you have a different approach in mind?

etrepum:
I think building something separate is probably best, and then we can worry about refactoring or deprecating the current implementation later. Eventually I think the markdown module will be changing a lot for correctness and flexibility reasons, so I don’t think we should care all that much about the existing code other than to keep it backwards compatible.

mayrang:
Added `extractWithChild` support as a separate recursive path (`createSelectionMarkdownExport`), keeping the existing export untouched. Using `$sliceSelectedTextNodeContent` with `'clone'` and `extractWithChild` with `'html'` as suggested.

Went through the transformer implementations looking for edge cases — found one with the list transformer. It iterates children itself and adds `- ` prefixes, so unselected items show up as empty prefixed lines (e.g. `- \n- Item 2` instead of `- Item 2`). The CODE transformer also skips `exportChildren` and calls `getTextContent()` directly, so selection doesn't apply there.

To fix the list issue, the transformer callback would need to support skipping nodes, which means changing the `(node) => string` interface — breaking change for external transformers. Since the markdown module is getting a bigger refactor anyway, maybe this is better addressed then? Let me know what you think.

Added a test showing the current list behavior so you can see the edge case directly.

etrepum:
Yeah, to that point, maybe it would be better to not add this API until we have better markdown infrastructure to support it.

I think in an ideal world we would have separate markdown shortcuts (that operate as you type) and markdown import/export (full document or selection import/export) infrastructure. They are very similar in theory but in practice trying to write code that does the right thing in both scenarios doesn't work very well. It would also be much easier to wire up a fully featured markdown implementation like remark or micromark to the import/export operations.

mayrang:
Makes sense. Would it be useful to keep this as a draft until then, or better to close it entirely? The basic cases (text, formatting, links) work well — just wanted to check if there's value in having it available even with the list limitation.

etrepum:
Up to you whether you keep it draft or close it. I don't think it can work correctly without refactoring other parts as you've noted. Maybe there's a possible refactoring of the existing transformer interface that allows this to work (adding an argument wouldn't necessarily break existing transformers or their types), if you want to look into that it would be a good interim solution for anyone looking for this functionality.

mayrang:
Thanks for the pointer on the transformer interface — added an optional `selection` parameter to `ElementTransformer.export` and updated `$listExport` to skip unselected items when provided. Since the parameter is optional, existing transformers work as-is. Nested lists are handled correctly too.

The list edge case from earlier is now fixed.

packages/lexical-markdown/flow/LexicalMarkdown.js.flow
```diff
@@ -103,6 +104,12 @@ declare export function $convertFromMarkdownString(
   shouldMergeAdjacentLines?: boolean,
 ): void;
 
+declare export function $convertSelectionToMarkdownString(
+  transformers?: Array<Transformer>,
```
etrepum: Current flow syntax and conventions are much closer to Typescript

```suggestion
  transformers?: Transformer[],
```

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -1849,3 +1851,333 @@ describe('markdown Safari compatibility (issue #8012)', () => {
...
+        bold.toggleFormat('bold');
+        const after = $createTextNode(' World');
+        paragraph.append(normal, bold, after);
+        root.append(paragraph);
+        // Select from "Hello " to "Bold"
+        normal.select(0, 6);
+        const selection = $getSelection();
+        if ($isRangeSelection(selection)) {
+          selection.focus.set(bold.getKey(), 4, 'text');
+        }
```
etrepum:
```suggestion
        const selection = normal.select(0, 6);
        selection.focus.set(bold.getKey(), 4, 'text');
```

etrepum: or alternatively you could use the NodeCaret APIs:

```ts
$setSelectionFromCaretRange(
  $getCaretRange(
    $getTextPointCaret(normal, 'next', 0),
    $getTextPointCaret(bold, 'next', 4),
  )
);
```

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -1849,3 +1851,333 @@ describe('markdown Safari compatibility (issue #8012)', () => {
...
+        const root = $getRoot();
+        const p1 = $createParagraphNode();
+        const t1 = $createTextNode('First paragraph');
+        p1.append(t1);
+        const p2 = $createParagraphNode();
+        const t2 = $createTextNode('Second paragraph');
+        p2.append(t2);
+        root.append(p1, p2);
+        // Select from "First" to "Second"
+        t1.select(0, 15);
```
etrepum: see above for simpler ways to create this RangeSelection

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -1849,3 +1851,333 @@ describe('markdown Safari compatibility (issue #8012)', () => {
...
+        const list = $createListNode('bullet');
+        const item1 = $createListItemNode();
+        const t1 = $createTextNode('Item 1');
+        item1.append(t1);
+        const item2 = $createListItemNode();
+        const t2 = $createTextNode('Item 2');
+        item2.append(t2);
+        list.append(item1, item2);
+        root.append(list);
+        t1.select(0, 6);
```
etrepum: see above for simpler ways to create this RangeSelection

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -1849,3 +1851,333 @@ describe('markdown Safari compatibility (issue #8012)', () => {
...
+        const item2 = $createListItemNode();
+        const t2 = $createTextNode('Item 2');
+        item2.append(t2);
+        const item3 = $createListItemNode();
+        const t3 = $createTextNode('Item 3');
+        item3.append(t3);
+        list.append(item1, item2, item3);
+        root.append(list);
+        // Select only "Item 2" and partial "Item 3"
+        t2.select(0, 6);
```
etrepum: see above for simpler ways to create this RangeSelection

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -1849,3 +1851,333 @@ describe('markdown Safari compatibility (issue #8012)', () => {
...
+      () => {
+        const root = $getRoot();
+        const quote = $createQuoteNode();
+        const t1 = $createTextNode('Line 1');
+        const br = $createLineBreakNode();
+        const t2 = $createTextNode('Line 2');
+        quote.append(t1, br, t2);
+        root.append(quote);
+        // Select only "Line 2"
+        t2.select(0, 6);
```
etrepum: fwiw the default offset for both select arguments is the end so `t2.select(0)` also works to select the entire TextNode

packages/lexical-markdown/src/index.ts
```diff
@@ -81,8 +86,30 @@ function $convertToMarkdownString(
   return exportMarkdown(node);
 }
 
+/**
+ * Converts the selected content to a markdown string.
+ */
+function $convertSelectionToMarkdownString(
+  transformers: Array<Transformer> = TRANSFORMERS,
```
etrepum:
```suggestion
  transformers: Transformer[] = TRANSFORMERS,
```

packages/lexical-markdown/src/index.ts
```diff
@@ -81,8 +86,30 @@ function $convertToMarkdownString(
...
+  selection: BaseSelection | null,
+  shouldPreserveNewLines: boolean = false,
+): string {
+  if (!selection) {
+    return '';
+  }
+  if ($isRangeSelection(selection) && selection.isCollapsed()) {
+    return '';
+  }
```
etrepum: These can be fused to save a few bytes


```suggestion
  if (!selection || $isRangeSelection(selection) && selection.isCollapsed()) {
    return '';
  }
```

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -81,6 +88,256 @@ export function createMarkdownExport(
...
 }
 
+/**
+ * Creates a markdown export function that only exports selected content.
+ * Uses a recursive structure similar to $appendNodesToHTML to support
+ * extractWithChild for proper handling of partial selections within
+ * inline elements like links.
+ */
+export function createSelectionMarkdownExport(
+  transformers: Array<Transformer>,
```
etrepum:
```suggestion
  transformers: Transformer[],
```

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -81,6 +88,256 @@ export function createMarkdownExport(
...
+    return output.join('\n');
+  };
+}
+
+function $processNodeForSelection(
+  node: LexicalNode,
+  selection: BaseSelection,
+  elementTransformers: Array<ElementTransformer | MultilineElementTransformer>,
+  textFormatTransformers: Array<TextFormatTransformer>,
+  textMatchTransformers: Array<TextMatchTransformer>,
```
etrepum:
```suggestion
  elementTransformers: (ElementTransformer | MultilineElementTransformer)[],
  textFormatTransformers: TextFormatTransformer[],
  textMatchTransformers: TextMatchTransformer[],
```

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -81,6 +88,256 @@ export function createMarkdownExport(
...
+  // For element transformers (heading, quote, list, code block, etc.)
+  for (const transformer of elementTransformers) {
+    if (!transformer.export) {
+      continue;
+    }
+    const result = transformer.export(
+      node,
+      _node =>
+        $exportChildrenForSelection(
+          _node,
```
etrepum: underscore prefix is for unused variables, shouldn't really be used when just avoiding shadowing

```suggestion
      node_ =>
        $exportChildrenForSelection(
          node_,
```

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -81,6 +88,256 @@ export function createMarkdownExport(
...
+}
+
+function $exportChildrenForSelection(
+  node: ElementNode,
+  selection: BaseSelection,
+  textFormatTransformers: Array<TextFormatTransformer>,
+  textMatchTransformers: Array<TextMatchTransformer>,
+  shouldPreserveNewLines: boolean,
+  unclosedTags?: Array<{format: TextFormatType; tag: string}>,
+  unclosableTags?: Array<{format: TextFormatType; tag: string}>,
```
etrepum:
```suggestion
  textFormatTransformers: TextFormatTransformer[],
  textMatchTransformers: TextMatchTransformer[],
  shouldPreserveNewLines: boolean,
  unclosedTags?: {format: TextFormatType; tag: string}[],
  unclosableTags?: {format: TextFormatType; tag: string}[],
```

etrepum (commented):
Overall this looks great, a welcome addition to the current API. All of the comments are really just syntax nits and simplifications

packages/lexical-markdown/flow/LexicalMarkdown.js.flow
mayrang: Fixed.

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
mayrang: Switched to NodeCaret API.

mayrang: Went with this approach, cleaner to read.

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
mayrang: Updated.

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
mayrang: Updated.

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
mayrang: Updated.

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
mayrang: Done, thanks.

packages/lexical-markdown/src/index.ts
mayrang: Fixed.

packages/lexical-markdown/src/index.ts
mayrang: Fused.

packages/lexical-markdown/src/MarkdownExport.ts
mayrang: Fixed.

packages/lexical-markdown/src/MarkdownExport.ts
mayrang: Fixed.

packages/lexical-markdown/src/MarkdownExport.ts
mayrang: Makes sense, fixed.

packages/lexical-markdown/src/MarkdownExport.ts
mayrang: Fixed.

mayrang:
All addressed, force-pushed. Thanks for the review!

etrepum (approved):

PR #8402, opened 2026-04-27
[lexical-markdown] Bug Fix: Preserve hard line breaks in default markdown import

pro-vi:
## Summary

Preserve CommonMark hard line break markers when default markdown import joins adjacent lines into a `LineBreakNode`.

The preserve-newlines path already keeps hard-break spelling because it preserves source lines directly. In default import, the hard-break marker was stripped before export could distinguish it from a soft line break. In merge-adjacent mode, hard breaks could also be folded into spaces.

This adds markdown provenance to generated `LineBreakNode`s with NodeState, reuses that path for paragraph/list/quote line joins, and exports tagged line breaks with their original marker style.

## Test plan

- `pnpm vitest --project unit packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts packages/lexical-markdown/src/__tests__/unit/MarkdownTransformers.test.ts --run`
- `pnpm run test-unit`
- `pnpm run ci-check`
- `pnpm run build`

### DIFF
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -475,12 +475,7 @@ function $exportChildren(
 }
 
 function $exportLineBreak(node: LineBreakNode): string {
-  const hardLineBreak = $getState(node, hardLineBreakState);
-
-  if (hardLineBreak !== null) {
-    return hardLineBreak + '\n';
-  }
-  return '\n';
+  return $getState(node, hardLineBreakState) + '\n';
 }
 
 function exportTextFormat(
--- a/packages/lexical-markdown/src/MarkdownExport.ts
+++ b/packages/lexical-markdown/src/MarkdownExport.ts
@@ -10,6 +10,7 @@ import type {
   BaseSelection,
   ElementNode,
   LexicalNode,
+  LineBreakNode,
   TextFormatType,
   TextNode,
 } from 'lexical';
@@ -17,6 +18,7 @@ import type {
 import {$sliceSelectedTextNodeContent} from '@lexical/selection';
 import {
   $getRoot,
+  $getState,
   $isDecoratorNode,
   $isElementNode,
   $isLineBreakNode,
@@ -25,6 +27,7 @@ import {
 
 import {
   ElementTransformer,
+  hardLineBreakState,
   MultilineElementTransformer,
   TextFormatTransformer,
   TextMatchTransformer,
@@ -62,7 +65,7 @@ export function createMarkdownExport(
 
     for (let i = 0; i < children.length; i++) {
       const child = children[i];
-      const result = exportTopLevelElements(
+      const result = $exportTopLevelElements(
         child,
         elementTransformers,
         textFormatTransformers,
@@ -285,7 +288,7 @@ function $exportChildrenForSelection(
 
     if ($isLineBreakNode(child)) {
       if (childIncluded) {
-        output.push('\n');
+        output.push($exportLineBreak(child));
         anyChildIncluded = true;
       }
     } else if ($isTextNode(child)) {
@@ -338,7 +341,7 @@ function $exportChildrenForSelection(
   return {markdown: output.join(''), shouldInclude: anyChildIncluded};
 }
 
-function exportTopLevelElements(
+function $exportTopLevelElements(
   node: LexicalNode,
   elementTransformers: Array<ElementTransformer | MultilineElementTransformer>,
   textTransformersIndex: Array<TextFormatTransformer>,
@@ -350,7 +353,7 @@ function exportTopLevelElements(
       continue;
     }
     const result = transformer.export(node, _node =>
-      exportChildren(
+      $exportChildren(
         _node,
         textTransformersIndex,
         textMatchTransformers,
@@ -366,7 +369,7 @@ function exportTopLevelElements(
   }
 
   if ($isElementNode(node)) {
-    return exportChildren(
+    return $exportChildren(
       node,
       textTransformersIndex,
       textMatchTransformers,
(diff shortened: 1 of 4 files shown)


packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -235,6 +237,55 @@ export const codeFenceState = createState('mdCodeFence', {
   resetOnCopyNode: true,
 });
 
+export type MarkdownHardLineBreak = 'backslash' | 'spaces';
```
etrepum: If the goal is to roundtrip exactly maybe the right value for this would be exactly the `'\\'` or run of two or more spaces to preserve exactly the number of spaces?

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -235,6 +237,55 @@ export const codeFenceState = createState('mdCodeFence', {
...
+
+export function $createMarkdownLineBreakNode(
+  previousNode: ElementNode,
+): LineBreakNode {
+  const lineBreakNode = $createLineBreakNode();
+  const lastChild = previousNode.getLastChild();
+
+  if ($isTextNode(lastChild)) {
+    const lastText = lastChild.getTextContent();
+    const hardLineBreak = getMarkdownHardLineBreak(lastText);
```
etrepum: The return value of this could also include the string without the linebreak (e.g. `[string, MarkdownHardLineBreak]`) which would make this simpler and less error-prone

packages/lexical-markdown/src/MarkdownTransformers.ts
pro-vi: Updated to store the literal marker.

packages/lexical-markdown/src/MarkdownTransformers.ts
pro-vi: Replaced with a single parser that returns the stripped line text together with the marker.

Also added `reset-on-copy` behavior parity with tests.

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -235,6 +237,60 @@ export const codeFenceState = createState('mdCodeFence', {
...
+  }
+
+  const spaces = line.match(/ {2,}$/);
+  if (spaces !== null) {
+    const marker = spaces[0];
+    const text = line.slice(0, -marker.length);
+    if (text !== '' && /\S$/.test(text)) {
+      return [text, marker];
+    }
+  }
```
etrepum: The regex could do the parsing here

```suggestion
  const spaces = line.match(/^(.*?\S)( {2,})$/);
  return spaces ? [spaces[1], spaces[2]] : null;
```

packages/lexical-markdown/src/__tests__/unit/LexicalMarkdown.test.ts
```diff
@@ -1698,6 +1717,41 @@ describe('markdown whitespace import (default mode)', () => {
...
+          ]),
+        ),
+    ).toBe(md);
+  }
+
+  function getHardLineBreakMarker(line: string): string {
+    const parsed = parseMarkdownHardLineBreak(line);
+    if (parsed === null) {
+      throw new Error('Expected a hard line break marker');
+    }
```
etrepum: assert is a more concise way to do this

packages/lexical-markdown/src/MarkdownExport.ts
```diff
@@ -214,6 +222,15 @@ function exportChildren(
   return output.join('');
 }
 
+function $exportLineBreak(node: LineBreakNode): string {
+  const hardLineBreak = $getState(node, hardLineBreakState);
+
+  if (hardLineBreak !== null) {
+    return hardLineBreak + '\n';
+  }
+  return '\n';
```
etrepum: If the default was `''` instead of null then it would be even simpler

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -235,6 +237,60 @@ export const codeFenceState = createState('mdCodeFence', {
...
+
+export const hardLineBreakState = createState('mdHardLineBreak', {
+  parse: (val): MarkdownHardLineBreak | null => {
+    if (val === '\\') {
+      return val;
+    }
+    if (typeof val === 'string' && /^ {2,}$/.test(val)) {
+      return val;
+    }
+    return null;
```
etrepum: using `''` instead of null as the default might make usage simpler

packages/lexical-markdown/src/MarkdownTransformers.ts
```diff
@@ -235,6 +237,60 @@ export const codeFenceState = createState('mdCodeFence', {
...
+export type MarkdownHardLineBreak = string;
+
+export const hardLineBreakState = createState('mdHardLineBreak', {
+  parse: (val): MarkdownHardLineBreak | null => {
+    if (val === '\\') {
+      return val;
+    }
+    if (typeof val === 'string' && /^ {2,}$/.test(val)) {
+      return val;
+    }
```
etrepum: Can do both of these tests in the same regex
```suggestion
    if (typeof val === 'string' && /^(\\| {2,})$/.test(val)) {
      return val;
    }
```

etrepum (commented):
There are some conflicts that need to be resolved since #8395 was merged but otherwise I think this is in pretty good shape

pro-vi:
Thanks again for the detailed review. I applied the parser/state cleanup and then checked the import cases more closely and found one follow-up issue: inline transforms can split hard-break syntax into a separate trailing text node after formatted/link content (see new test cases), so the marker could stay as document text instead of moving onto the LineBreakNode.

I added coverage for formatted text, exact space counts, links, blockquotes, and a guard for spaces inside link labels.

etrepum (approved):

PR #8411, opened 2026-04-28
[lexical-utils] Bug Fix: Pass editor context to editorState.read() in markSelection

potatowagon:
## Summary

- PR #8353 renamed `rangeTargetFromPoint` to `$rangeTargetFromPoint`, which now calls `$getEditor()` to use the extensible DOM API. However, the `editorState.read()` call in `markSelection`'s `compute()` function was not updated to pass `{editor}`, so `$getEditor()` returns `null` and throws when the selection's anchor/focus is on an element node (not text).
- This fix passes the `editor` reference to `editorState.read(() => { ... }, {editor})` so that `$getEditor()` can find the active editor inside `$rangeTargetFromPoint`.

## Test plan
pnpm run test-unit
all test pass, including
✓  unit  packages/lexical-utils/src/__tests__/unit/markSelection.test.tsx (4 tests) 193ms

### DIFF
--- a/packages/lexical-utils/src/__tests__/unit/markSelection.test.tsx
+++ b/packages/lexical-utils/src/__tests__/unit/markSelection.test.tsx
@@ -1,0 +1,124 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import {
+  $createParagraphNode,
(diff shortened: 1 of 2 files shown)


etrepum:
A simpler fix would be to make `$rangeTargetFromPoint` take an editor argument since it's already available all the way up the stack

potatowagon:
> A simpler fix would be to make `$rangeTargetFromPoint` take an editor argument since it's already available all the way up the stack

good point, thankyou! simplified it accordingly

zurfyx (approved):

PR #8412, opened 2026-04-28
[lexical][lexical-rich-text][lexical-list] Bug Fix: Import dir attribute in importDOM

### LINKED ISSUE #7765: Feature: `ElementNode` importDOM should import the `dir` attribute
## Description

As off #7727, it's possible to set the `direction` attribute of Element Nodes manually, and this is reflected in `exportDOM` output, however `importDOM` has not been updated to import this attribute.

## Impact

This is mostly a consistency/expectation improvement. `export/importJSON` correctly handle the direction attribute, so it's reasonable to expect that the same would apply for DOM.

mayrang:
## Description

Fixes #7765

The `dir` attribute is exported correctly via `exportDOM` (added in #7727), but `importDOM` conversion functions don't read it back. This means the text direction is lost when importing HTML — for example, RTL content pasted into the editor loses its direction.

### What changed

Added `dir` attribute handling to all ElementNode `importDOM` conversion functions:

- `$convertParagraphElement` (LexicalParagraphNode.ts)
- `$convertHeadingElement` (lexical-rich-text)
- `$convertBlockquoteElement` (lexical-rich-text)
- `$convertListNode` (lexical-list)
- `$convertListItemElement` (lexical-list)

Each reads `element.getAttribute('dir')` and calls `node.setDirection(dir)` when the value is `'ltr'` or `'rtl'`.

## Test Plan

Added tests for HTML import → direction preservation:
- Paragraph with dir="rtl" / dir="ltr" / no dir
- Heading with dir="rtl"
- Blockquote with dir="rtl"
- List with dir="rtl"


etrepum:
tests are failing, expectations probably need to change to match the more correct behavior

mayrang:
Done — fixed the expectations so `dir="ltr"` is preserved on imported `<p>`, `<li>`, and table cell paragraphs.

packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
```diff
@@ -367,4 +368,83 @@ describe('HTML', () => {
...
+        {discrete: true},
+      );
+      let direction: string | null = null;
+      editor.getEditorState().read(() => {
+        const firstChild = $getRoot().getFirstChild();
+        if ($isElementNode(firstChild)) {
+          direction = firstChild.getDirection();
+        }
+      });
+      return direction;
```
etrepum:
```suggestion
      return editor.read(() => {
        const firstChild = $getRoot().getFirstChild();
        return $isElementNode(firstChild) ? firstChild.getDirection() : null;
      });
```

packages/lexical-html/src/__tests__/unit/LexicalHtml.test.ts
```diff
@@ -367,4 +368,83 @@ describe('HTML', () => {
...
+      expect(importAndGetDirection('<h1 dir="rtl">عنوان</h1>')).toBe('rtl');
+    });
+
+    test('blockquote with dir="rtl"', () => {
+      expect(
+        importAndGetDirection('<blockquote dir="rtl">اقتباس</blockquote>'),
+      ).toBe('rtl');
+    });
+
+    test('list with dir="rtl"', () => {
```
etrepum: this looks like it could be an importAndGetDirection test

also these could be simplified a bit further by using a data structure to build the tests https://vitest.dev/api/test.html#test-for

packages/lexical-list/src/LexicalListItemNode.ts
```diff
@@ -646,7 +646,12 @@ function $convertListItemElement(domNode: HTMLElement): DOMConversionOutput {
       : ariaCheckedAttr === 'false'
         ? false
         : undefined;
-  return {node: $createListItemNode(checked)};
+  const node = $createListItemNode(checked);
+  const dir = domNode.getAttribute('dir');
+  if (dir === 'ltr' || dir === 'rtl') {
+    node.setDirection(dir);
+  }
```
etrepum: This seems like a useful helper to export from lexical since it's used across several places, e.g. 

```ts
export function $setDirectionFromDOM<T extends ElementNode>(node: T, domNode: HTMLElement): T {
  const dir = domNode.getAttribute('dir');
  return dir === 'ltr' || dir === 'rtl' ? node.setDirection(dir) : node;
}
```

mayrang:
Thanks for the pointers — all three applied:
- Extracted `$setDirectionFromDOM<T>` in `LexicalUtils` (exported from `lexical`), used in ParagraphNode, ListItemNode, ListNode, and the heading/blockquote converters.
- Switched the test helper to `editor.read(cb)` returning the direction directly.
- Folded the cases (including the list one) into a single `test.for` table.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -2074,6 +2074,14 @@ export function setNodeIndentFromDOM(
   elementNode.setIndent(indent);
 }
 
+export function $setDirectionFromDOM<T extends ElementNode>(
```
etrepum, line 2088: Let's add some API docs for this new export, otherwise I think everything here looks great

mayrang:
Done — added JSDoc for `$setDirectionFromDOM`.

etrepum (approved):

PR #8414, opened 2026-04-28
[lexical][lexical-playground] Bug Fix: Move caret past trailing characters in accepted macOS text replacements

### LINKED ISSUE #8413: Bug: pressing space to accept a macOS text replacement leaves the caret before the space
Lexical version: commit `4a43cb027f55718d6d32cc7fb8e7ec31adf59d2e`

## Steps To Reproduce

1. Type a macOS text replacement and press space.

Link to code example:

https://github.com/user-attachments/assets/16fbe032-dc90-4923-bec2-5a43a4c9e053

## The current behavior

The caret ends up before the space.

## The expected behavior

The caret ends up after the space.

## Impact of fix

This happens for every macOS text replacement. This behavior does not match expectations, leading to users who type fast missing spaces (or, rather, typing before them). This is arguably the fault of how Chrome and Firefox fire the Space `beforeinput` event *before* the replacement event (Safari doesn't, so it's fine).

nmggithub:
## Description

This adds workaround behavior where typing a character to accept a macOS text replacement used to leave the caret before that character instead of after.

(edit: branch name is a bit of a misnomer, as this behavior exhibits in Firefox as well)

(edit 2: this branch now covers the Tab and Enter cases as well, but the title is kept the same)

Closes #8413 

## Test plan

There is a test added.

### Before

https://github.com/user-attachments/assets/0327442b-ba04-47e5-ba95-32a4548f9033

### After

https://github.com/user-attachments/assets/19748694-9c77-43ef-b711-c3618186b33a


etrepum:
`pnpm run ci-check` is failing, looks like a formatting issue that `pnpm run prettier:fix` should sort out

nmggithub:
> `pnpm run ci-check` is failing, looks like a formatting issue that `pnpm run prettier:fix` should sort out

I see no errors on my end for either of those commands.

etrepum:
Your branch is out of sync with main so perhaps you're getting an error that would occur if your branch was updated

nmggithub:
> Your branch is out of sync with main so perhaps you're getting an error that would occur if your branch was updated

Want me to merge in `main` before or after this upcoming set of CI runs?

etrepum:
The CI run is already failing again, it's always a good idea to have your branch in sync with main.

nmggithub:
> The CI run is already failing again, it's always a good idea to have your branch in sync with main.

Got it! My apologies for not doing so earlier.

nmggithub:
On a tangential note, [this bug for macOS text replacement where Backspace accepts the replacement](https://github.com/electron/electron/issues/50432) still exhibits in Chrome, but I have a patch in for it on Electron (at the very least).

nmggithub:
Ok, Prettier should (hopefully) be happy now. Sorry for all the back and forth!

etrepum:
Some of the firefox e2e tests are failing

nmggithub:
> Some of the firefox e2e tests are failing

Yeah I noticed this. It seems to be mostly my own test. Since it is a macOS-specific test, would it be ok if I just put it behind a platform guard?

etrepum:
Ideally any platform guard or detection of misbehavior (better than a platform guard in case the platform fixes the bug) would be in the workaround and the tests would run uniformly across platforms and browsers. This would make it more likely that we don't accidentally cause regressions in environments that didn't have this particular issue to begin with.

nmggithub:
I'll do some testing on my end on Linux to see where the behavior is breaking. If I can get away without a platform guard in the workaround code itself, I'd prefer it.

nmggithub:
Firefox tests work well for me on Linux (Fedora 43). I know that distro isn't officially supported by Playwright, but I got the tests running and working. I'm not sure what's wrong with the CI runners. EDIT: Seems to be a plain-text-mode discrepancy. Currently investigating. EDIT 2: I've figured it out, but have found more edge cases. Working on those now.

nmggithub:
Added fix for plain-text mode. I then realized Enter and Tab were still broken so I added logic for those. I added a test for Enter, but left Tab without one as really only Chrome treats that as an acceptance of a text replacement. Not even Safari does it.

Another note I'll add: the vast majority of this workaround code is because Firefox and Chrome fire events in a different order than Safari. The bugs this is meant to address do not exhibit in Safari at all.

nmggithub:
Closing in favor of a combined effort in #8417.

PR #8417, opened 2026-04-28
[lexical][lexical-playground] Bug Fix: Add workarounds for buggy browser behavior around macOS text replacements

### LINKED ISSUE #8413: Bug: pressing space to accept a macOS text replacement leaves the caret before the space

### LINKED ISSUE #8416: Bug: Backspace accepts macOS text replacement
Lexical version: commit `67cfec5a27108bb865a309db3d0ba5d660ed885c`

## Steps To Reproduce

1. Type a macOS text replacement and press backspace.

Link to code example:

https://github.com/user-attachments/assets/3c38b86b-fead-45c0-8332-a4f1e0882bb9

## The current behavior

Backspace accepts the text replacement, and the caret sometimes ends up somewhere in the middle of the inserted text.

## The expected behavior

Backspace deletes a character.

## Impact of fix

This happens for every macOS text replacement. This only happens on Chrome.

Related: https://github.com/electron/electron/issues/50432

nmggithub:
## Description

This adds workaround behavior where typing a character to accept a macOS text replacement used to leave the caret before that character instead of after. It also adds workaround behavior where pressing Backspace (and performing Select All, apparently) would accept a pending macOS text replacement instead of deleting a character.

Closes #8413
Closes #8416

## Test plan

There are tests added for the caret placement bug. The Backspace / Select All acceptance bug relies on native behavior that can't easily be synthesized, so there are no tests around that.

### Before

https://github.com/user-attachments/assets/0327442b-ba04-47e5-ba95-32a4548f9033

https://github.com/user-attachments/assets/a687a654-cfff-495f-b133-b81299429825

### After

https://github.com/user-attachments/assets/19748694-9c77-43ef-b711-c3618186b33a

https://github.com/user-attachments/assets/1dab46f7-f141-4b0f-bd19-23baee7df3af


nmggithub:
This might not actually be possible due to Chrome's bad heuristics. Still looking into it myself. EDIT: I think I figured it out now. It's a bit wonky but it *does* seem to work.

packages/lexical/src/LexicalEvents.ts
```diff
@@ -607,6 +610,23 @@ function isPossiblyAndroidKeyPress(timeStamp: number): boolean {
...
+  }
+}
+
+function markHandledSelectionCommandInsertText(): void {
+  clearHandledSelectionCommandInsertText();
+  isInsertTextAfterHandledSelectionCommand = true;
+  handledSelectionCommandTimeoutId = setTimeout(() => {
+    isInsertTextAfterHandledSelectionCommand = false;
+    handledSelectionCommandTimeoutId = null;
+  }, 0);
```
etrepum: Looks like this is effectively the same as the above function, should save a few bytes to just call it

```suggestion
  handledSelectionCommandTimeoutId = setTimeout(clearHandledSelectionCommandInsertText, 0);
```

packages/lexical/src/LexicalEvents.ts
```diff
@@ -1296,11 +1339,15 @@ function $handleKeyDown(event: KeyboardEvent): boolean {
...
+        if (dispatchCommand(editor, SELECT_ALL_COMMAND, event)) {
+          markHandledSelectionCommandInsertText();
+        }
       }
     } else if (isSelectAll(event)) {
       event.preventDefault();
-      dispatchCommand(editor, SELECT_ALL_COMMAND, event);
+      if (dispatchCommand(editor, SELECT_ALL_COMMAND, event)) {
+        markHandledSelectionCommandInsertText();
+      }
```
etrepum: As far as I can tell this can be lifted up after isRedo and not repeated

etrepum (commented):
I think we should try and smooth out all of the wonky text replacement behavior we can find in one pass, the behavior in chrome with this one is still a bit wonky. For example pressing return or space should accept the replacement *then* do the behavior of that key. Right now it just accepts the replacement and effectively prevents default (haven't done a thorough audit of that section of the events code to step through what it's doing, just QA). The best justification to do it all at once is to make sure that we add the minimum amount of implicit state machine to make it work rather than add a little bit for each workaround.

Safari seems to have the right behavior before this patch, can also be tested in chrome in any textarea.

nmggithub:
@etrepum yeah, I'm not particularly happy with fix.

This patch of Chromium I submitted for Electron fixes it for Chrome: https://github.com/electron/electron/issues/50432

To note, Chrome is the only browser between it, Firefox, and Safari, that has this bug.

#8414 is a different story. Firefox and Chrome exhibit that bug, but Safari doesn't.


Both bugs closely related, and there's spec questions to be had at how all three browsers implement `InputEvent`s around macOS text replacements, but we have to work around those for both bugs here.

nmggithub:
> I think we should try and smooth out all of the wonky text replacement behavior we can find in one pass, 

I find
1. Backspace accepts text replacement, and
2. caret ends up before the acceptance boundary,

to be two separate bugs (which is why I filed them separately). Bug 1 only exists in Chrome and is arguably because it has the worst implementation of `InputEvent`s here (I could go into detail, but I'll save space here). Bug 2 exists in Chrome and Firefox because they both fire the Space, Return, Etc. event *before* the event that replaces the text. Safari fires the event that replaces the text *first* (hence how I wrote the test for #8414).

Again, we can't really change any of this, only work around it.

etrepum:
I agree that conceptually these are different things because there are different keystrokes involved and have slightly different browser compatibility matrices. However, as far as reviewing goes it would be much easier and more efficient for me if I had one PR to review that worked around all of the known bad behavior with macOS text replacements. This PR for example is shipping without any sort of test infrastructure (playwright doesn't really make it feasible to write real tests for IMEs anyway) and requires manual QA in all of the browsers, so it would be way better if I only had to go through that process for one PR.

nmggithub:
> I agree that conceptually these are different things because there are different keystrokes involved and have slightly different browser compatibility matrices. However, as far as reviewing goes it would be much easier and more efficient for me if I had one PR to review that worked around all of the known bad behavior with macOS text replacements. This PR for example is shipping without any sort of test infrastructure (playwright doesn't really make it feasible to write real tests for IMEs anyway) and requires manual QA in all of the browsers, so it would be way better if I only had to go through that process for one PR.

Understandable, I can combine everything into this PR if you prefer, since we're having the convo here already.

etrepum:
What I would like to see working (with the selection on the rightmost character except in the arrow key cases):
 
- `"omw<space>"` results in `"On My Way! "`
- `"omw<backspace>"` results in `"om"` (dismisses the text replacement and removes w)
- `"omw!"` results in `"On My Way!!"` (most punctuation and symbols should accept and continue according to platform behavior elsewhere)
- `"omw<enter>"` results in `"On My Way!\n"` (newline, in the default lexical case this would be a new paragraph)
- `"omw<arrowkey>"` results in "omw" (dismisses the text replacement) and results in the default navigation for that key
- `"omw<esc>"` results in "omw" (dismisses the text replacement)

Ideally this would just be one PR where I can carefully read through the code and all of the required context and run through all of the scenarios in each browser to verify expected behavior. With multiple PRs I have to repeat all of this.

In the current playground all of these work as expected in Safari, but the behavior is mixed in the other browsers

etrepum:
Rolling it up into this PR works for me

nmggithub:
A long bit of context for these bugs.

### The bugs

The two bugs:
1. Backspace (and Select All, apparently) accepts a pending text replacement, and
2. caret ends up before the acceptance boundary after accepting a pending text replacement,

### The `InputEvent`s

#### Safari (No Bugs)

Safari has, arguably, the most sensible behavior around these replacements. It exhibits neither of these bugs.

Typing `omw<space>` in Safari results in the last two `InputEvent`s being:
1. `{ inputType: "insertReplacementText", dataTransfer: { ... "On my way!" (equivalently) ... })`, and
2. `{ inputType: "insertText", data: " " } `.

This gets two things right:
- it uses `insertReplacementText` for the event that replaces the text (matching [Input Events Level 2](https://www.w3.org/TR/input-events-2/)), and
- it fires the Space event *after* the replacement event (which plays nicely with Lexical's event handling, leaving the caret in the correct spot).

#### Firefox (Bug 2)

Firefox flips the order, but at least gets `insertReplacementText` right: 
1. `{ inputType: "insertText", data: " " } `, and
2. `{ inputType: "insertReplacementText", dataTransfer: { ... "On my way!" (equivalently) ... })`.

#### Chrome (Bug 1, Bug 2)

Chrome doesn't even use `insertReplacementText` for this: 
1. `{ inputType: "insertText", data: " " } `, and
2. `{ inputType: "insertText", data: "On my way!" }`.

Chrome also is the only one that exhibits Bug 1.
From what I can tell, its due to some internal logic error. [I've squashed Bug 1 in Electron by adding a Electron-local Chromium patch,](https://github.com/electron/electron/issues/50432) but it's also possible that logic I patched was/is in plumbing that wouldn't even be hit if Chrome chose to use `insertReplacementText` for this instead (but I can't confirm that yet).

Regardless, Chrome's choice to not use `insertReplacementText` is apparently deliberate, with them only using it for their own first-party replacements. [There is an open issue trying to get them to use `insertReplacementText` for maOS text replacements,](https://issues.chromium.org/issues/40657097) though.

#### Proper Long-Term Solutions

- Getting all browsers to adopt Safari's event ordering would likely finally squash Bug 2 and remove the need for our workaround.
- Getting Chromium to use the `insertReplacementText` plumbing for macOS text replacements *may* squash Bug 1 and remove the need for our workaround, but it might entail a drift in deliberate policy by the Chromium team to not use `insertReplacementText` for third-party replacements (despite how arguably out-of-spec that is).

packages/lexical/src/LexicalEvents.ts
nmggithub: Should be resolved by 2e29782.

packages/lexical/src/LexicalEvents.ts
nmggithub: Should be resolved by 2e29782.

etrepum (commented):
This is much improved! I've found two edge cases that we should consider handling before this is merged:

Chrome - if you complete the text replacement with the same character that the replacement ends with, the selection ends up before the last character you typed.

https://github.com/user-attachments/assets/df788a39-873b-4821-b95d-562120a0f887

Firefox - punctuation doesn't accept the text replacement

https://github.com/user-attachments/assets/4146d3eb-9af7-4e81-903e-50e258fa80c7

nmggithub:
> Firefox - punctuation doesn't accept the text replacement

This appears to be the current behavior before my patches in both `contenteditable` _and_ regular `textarea` and `input` elements in Firefox. I don't know if we want to deliberately stray from that. I would liken this to the fact that Tab only accepts the text replacement in Chrome. It seems to be a consistent browser difference.

I *will* investigate the Chrome bug, though.

nmggithub:
The lingering Chrome bug has been squashed! I left the Firefox bug as-is because it appears to be consistent across all kinds of inputs in that browser.

etrepum (approved):
QA looks great, agreed that it's fine to leave the Firefox behavior as-is. Will merge after taking a closer look at the code when I have the time.

PR #8419, opened 2026-04-29
[lexical-playground] Bug Fix: Position FloatingLinkEditor below the entire link for multi-line and decorator-only links

mayrang:
## Summary

- Fixes #8263. For multi-line links, the floating link toolbar attached to the cursor's range rect (one line tall), so a cursor on line 1 placed the toolbar between line 1 and line 2 — covering line 2 (and the cursor when it sat there).
- The NodeSelection branch already uses the link element's bounding rect; align RangeSelection by looking up the LinkNode and using `getClientRects()[last]` (the last visual line of the link). Falls back to the prior range-rect behavior when the cursor is not inside a link.
- For decorator-only links (e.g. an image wrapped in a link), the link element's line box can be misaligned with the decorator's visual extent (baseline-based positioning makes the `<a>` rect sit on the image's lower half). When the link has a single decorator child, anchor to the decorator's rect directly so the toolbar lands below the image.

### Before
스크린샷 2026-04-30 오전 1 51 52

### After
스크린샷 2026-04-30 오전 1 52 23

## Test plan
- [x] Single-line link: toolbar appears below the link (no regression)
- [x] Multi-line link, cursor on line 1: toolbar appears below the last line, not covering line 2
- [x] Multi-line link, cursor on line 2: toolbar appears below the last line
- [x] Image link (decorator-only link): toolbar appears below the image
- [x] Single-character link (#8392): still works via `$getSelectedLinkNode`'s right-bias path

### DIFF
--- a/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
+++ b/packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
@@ -9,7 +9,14 @@ import type {JSX} from 'react';
 
 import './index.css';
 
-import {autoUpdate, flip, offset, shift, useFloating} from '@floating-ui/react';
+import {
+  autoUpdate,
+  flip,
+  inline,
+  offset,
+  shift,
+  useFloating,
+} from '@floating-ui/react';
 import {
   $createLinkNode,
   $isAutoLinkNode,
@@ -21,6 +28,7 @@ import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {$findMatchingParent, mergeRegister} from '@lexical/utils';
 import {
   $getSelection,
+  $isDecoratorNode,
   $isLineBreakNode,
   $isNodeSelection,
   $isRangeSelection,
@@ -102,6 +110,7 @@ function FloatingLinkEditor({
 
   const {refs, floatingStyles} = useFloating({
     middleware: [
+      inline(),
       offset(10),
       flip({
         boundary: scrollerElem || undefined,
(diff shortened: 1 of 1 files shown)


packages/lexical-playground/src/plugins/FloatingLinkEditorPlugin/index.tsx
```diff
@@ -169,7 +170,38 @@ function FloatingLinkEditor({
...
+              if (decoratorElement) {
+                domRect = decoratorElement.getBoundingClientRect();
+              }
+            }
+            if (!domRect) {
+              const linkElement = editor.getElementByKey(linkNode.getKey());
+              if (linkElement) {
+                const rects = linkElement.getClientRects();
+                domRect =
+                  rects[rects.length - 1] ||
```
etrepum: It looks like floating-ui has a better solution for this sort of situation. https://floating-ui.com/docs/inline

mayrang: Thanks for pointing to `inline`! I tried it and ran into an issue I couldn't resolve cleanly.

After applying `inline`, the toolbar lands at different visual positions across attempts even when the reference is the same.

**Attempt A:**
image
image


**Attempt B:**
image

image


Hello(link: www.google-)
Bye(link: www.naver-)

Both attempts log the same reference:
- text: `bye`
- rect: `{top: 298.27, left: 28, width: 98.33, height: 17.5}`

Same reference, identical rect, different rendered position. The same pattern reproduces with another link in the document. Calling `update()` explicitly after `setReference()` did not change the outcome.

For this PR, I kept the single bounding rect approach (`linkElement.getBoundingClientRect()`) since the issue doesn't reproduce there.

The behavior with `inline` looks like it needs a deeper look — would you prefer to investigate it in this PR, or split it into a separate issue and look into it there? Happy to either way.

etrepum: Slight preference to try and resolve all known positioning issues with the floating link toolbar in this PR, but up to you

mayrang: Dug into this more — turned out it's how `setReference(domElement)` interacts with `inline`, not `inline` itself. Swapped it for `setPositionReference` with a virtual element exposing both rect methods, and the positioning issues went away. Updating the PR.

mayrang: Pushed the change. Switched to `setPositionReference` with a virtual element exposing both `getBoundingClientRect()` and `getClientRects()`, and `inline` is now in the middleware chain. Verified the previously broken cases (viewport-bottom flip, mouse vs keyboard entry, sticky placement) all behave correctly with this approach.

etrepum (approved):
Looks great for the cases I tried!

PR #8422, opened 2026-04-29
[lexical] perf: defer DOM Selection property reads in $updateDOMSelection

rhagigi:
## Summary

Accessing `domSelection.anchorNode`, `.focusNode`, `.anchorOffset`, and `.focusOffset` forces the browser to resolve the selection against the current layout, triggering synchronous style/layout recalculation.

Previously these four properties were read eagerly at the top of `$updateDOMSelection`, even when the function would immediately return via one of two early-exit paths:
1. Collaboration tag / decorator input check
2. `!isRangeSelection(nextSelection)` (the common case on mount)

During initial mount, both `prevSelection` and `nextSelection` are null, so the function hits the `!isRangeSelection` branch and returns — but only after the browser has already paid the layout recalc cost for property reads that are never consumed.

## Performance trace

image

## Changes

- Keep `document.activeElement` at the top (needed for the first early return)
- Inline `domSelection.anchorNode`/`.focusNode` directly in the `isSelectionWithinEditor` call (only reached when `prevSelection !== null`)
- Defer the four property reads (`anchorNode`, `focusNode`, `anchorOffset`, `focusOffset`) to just after the early returns, right before the diff-check block that actually uses them

## Impact

Zero layout-triggering `Selection` reads on mount and on any code path that exits early. This eliminates unnecessary forced style/layout recalculations during Lexical node mounting.

### DIFF
--- a/packages/lexical/src/LexicalSelection.ts
+++ b/packages/lexical/src/LexicalSelection.ts
@@ -3015,10 +3015,6 @@ export function $updateDOMSelection(
   rootElement: HTMLElement,
   nodeCount: number,
 ): void {
-  const anchorDOMNode = domSelection.anchorNode;
-  const focusDOMNode = domSelection.focusNode;
-  const anchorOffset = domSelection.anchorOffset;
-  const focusOffset = domSelection.focusOffset;
   const activeElement = document.activeElement;
 
   // TODO: make this not hard-coded, and add another config option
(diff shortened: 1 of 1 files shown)


etrepum (approved):

levensta:
It will be interesting to look at the performance trace after these changes

PR #8423, opened 2026-04-30
fix(lexical-plain-text): add escape key handler matching rich-text behavior

### LINKED ISSUE #5922: Feature: escape key handler for plain text plugin

matingathani:
Closes #5922

Adds KEY_ESCAPE_COMMAND handler to the plain text plugin, matching the behavior already present in the rich text plugin (line 960-978 of packages/lexical-rich-text/src/index.ts). When Escape is pressed with a range selection, the editor is blurred.

### DIFF
--- a/packages/lexical-plain-text/src/index.ts
+++ b/packages/lexical-plain-text/src/index.ts
@@ -41,6 +41,7 @@ import {
   KEY_BACKSPACE_COMMAND,
   KEY_DELETE_COMMAND,
   KEY_ENTER_COMMAND,
+  KEY_ESCAPE_COMMAND,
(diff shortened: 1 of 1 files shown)


etrepum:
This could use some test coverage, PR titles and descriptions should also match the repository template https://raw.githubusercontent.com/facebook/lexical/refs/heads/main/.github/pull_request_template.md

PR #8424, opened 2026-04-30
fix(lexical-playground): add BLOCK_EQUATION markdown transformer for $$...$$ syntax

### LINKED ISSUE #6936: Bug: Block math equation is falsely converted to inline equation on MarkdownShortcutPlugin

matingathani:
Closes #6936

Adds a new BLOCK_EQUATION TextMatchTransformer that handles `$$...$$` block math equations. The existing EQUATION transformer only handled inline `$...$` equations, causing block equations to be incorrectly converted to inline when using the markdown shortcut.

Changes:
- Added BLOCK_EQUATION transformer (using $([^$]+?)$$ regex) that creates equations with inline=false
- Updated EQUATION export to only match inline equations (inline=true)
- Placed BLOCK_EQUATION before EQUATION in the transformers array so $$ is matched before $

### DIFF
--- a/packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts
+++ b/packages/lexical-playground/src/plugins/MarkdownTransformers/index.ts
@@ -114,7 +114,7 @@ export const EMOJI: TextMatchTransformer = {
 export const EQUATION: TextMatchTransformer = {
   dependencies: [EquationNode],
   export: node => {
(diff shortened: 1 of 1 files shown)


etrepum:
Will review after this has tests and the CLA is accepted.

PR #8426, opened 2026-04-30
fix(lexical-react): prevent false onChange event on initial autofocus with code highlighting

### LINKED ISSUE #4493: Bug: When AutoFocusPlugin is Combined with CodeHighlightPlugin and LexicalOnChangePlugin - an OnChange Event is raised on document load.
When the React AutoFocusPlugin,  (@lexical/react/LexicalAutoFocusPlugin) React OnChangePlugin (@lexical/react/LexicalOnChangePlugin) and the CodeHighlightPlugin (calling {registerCodeHighlighting} from '@lexical/code') are all enabled in the editor - then when the application first loads the document / editor and is 'AutoFocused' - an OnChange event is raised. 

Pretty sure it's being caused by the code highlighter registration - but it only happens when the AutoFocus (or the call to editor.focus()) occurs (i.e. it doesn't happen when the editor is manually focused by clicking inside the editor window).

Lexical version: 0.10.0

## Steps To Reproduce

1. Enable all three plugins - AutoFocusPlugin, OnChangePlugin, and CodeHighlightPlugin
2. Close the document / editor application and then re-open and existing document / editor - the editor will auto-focus, but it will also raise an onChange event - even though the editor state has not changed.

## The current behavior

An onChange event is being raised even though there is no change to editorState

## The expected behavior

There should be no onChange event when the document / editor is loaded (for example when a user is simply opening a document in the application)

matingathani:
Closes #4493

When AutoFocusPlugin, CodeHighlightPlugin, and OnChangePlugin are all used together, an onChange event fires on initial load even though the editor state hasn't changed.

Root cause: LexicalComposer initializes editor state in useMemo (before child useLayoutEffect callbacks). When AutoFocusPlugin later calls editor.focus(), code highlighting transforms can dirty nodes. By this point prevEditorState.isEmpty() returns false because initial content was already loaded.

Fix: Snapshots the editor state at listener registration time. Skips the first update where prevEditorState still matches that snapshot (reference equality). After the auto-focus update creates a new EditorState, subsequent user edits have a different prevEditorState and fire onChange normally.

### DIFF
--- a/packages/lexical-react/src/LexicalOnChangePlugin.ts
+++ b/packages/lexical-react/src/LexicalOnChangePlugin.ts
@@ -10,6 +10,7 @@ import type {EditorState, LexicalEditor} from 'lexical';
 
 import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
 import {HISTORY_MERGE_TAG} from 'lexical';
+import {useRef} from 'react';
 import useLayoutEffect from 'shared/useLayoutEffect';
 
 export function OnChangePlugin({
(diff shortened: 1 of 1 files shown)


etrepum:
It's unclear that this is correct, there aren't any tests or an example to show that this actually solves the issue. PR titles and descriptions should follow the repository template: https://raw.githubusercontent.com/facebook/lexical/refs/heads/main/.github/pull_request_template.md

PR #8427, opened 2026-04-30
[lexical-list] Feature: Add the `createListItemNode` method to the ListNode and use it for children normalization

levensta:
## Description

Follow-up https://github.com/facebook/lexical/pull/8213

Classes that extend ListNode can use the parent `after` method from the conversion function in `importDOM`. To ensure that child normalization works correctly, `createListItemNode` method has been added to the `ListNode` class, which can be overridden in subclasses

### Before

### After


levensta:
Since a copy of ListItemNode is now used for normalization, the default `dir="ltr"` attribute was missing from the test. I think this is a positive change, although the normalized listitem does not affect the direction of the list, so I just updated the test

image

packages/lexical-list/src/utils.ts
```diff
@@ -194,10 +194,14 @@ export function $removeHighestEmptyListParent(
...
  * @param node - The node to be wrapped into a ListItemNode
+ * @param listItemWrapper — The item that the node wraps into.
+ * By default this is ListItemNode, but can be overridden by a subclass
  * @returns The ListItemNode which the passed node is wrapped in.
  */
-export function $wrapInListItem(node: LexicalNode): ListItemNode {
-  const listItemWrapper = $createListItemNode();
+export function $wrapInListItem(
+  node: LexicalNode,
+  listItemWrapper: ListItemNode = $createListItemNode(),
```
etrepum:
```suggestion
  $createWrapper: () => ListItemNode = $createListItemNode,
```

The parameter should be a function to create the ListItemNode, not the ListItemNode itself.

packages/lexical-list/src/LexicalListNode.ts
```diff
@@ -303,12 +303,16 @@ function $normalizeChildren(nodes: Array<LexicalNode>): Array<ListItemNode> {
...
+            normalizedListItems.push(
+              $wrapInListItem(child, () => $copyNode(node)),
+            );
           }
         });
       }
     } else {
-      normalizedListItems.push($wrapInListItem(node));
+      const found = nodes.find($isListItemNode);
+      const $createListItem = found ? () => $copyNode(found) : undefined;
```
etrepum: It does seem strange to just arbitrarily copy some sibling node (that could have some item-specific settings such as style or value), and that we are doing the search every time instead of at most once.

etrepum (commented):
Maybe a better solution to solve the problem at hand is to give ListNode a `createListItem(): ListItemNode { return $createListItem(); }` method by subclasses. I think this is probably the actual behavior you're trying to achieve here (also the splice method could use this)?

packages/lexical-list/src/LexicalListNode.ts
levensta: Yes, the search can indeed be taken out of the condition and a single function for creating a copy can be created based on the found element 👍 The same thing happens in `splice()`

As for specific properties, in this case it doesn't affect the list in any way, although I thought about common ones like `style` and `textFormat` and I think I can add a reset of these properties here and for `splice()`

levensta:
I don't quite understand why the ListItemNode class needs a method that will create an instance of the same class. In this case, it is important to create a copy of the same instance that is already contained in the ListNode

etrepum:
That was a typo on my part, edited to correct ListItem to ListNode

etrepum:
I'd like more context about why it is important to create a copy of some arbitrary other ListItemNode and why the ListNode subclass wouldn't have enough information to create the correct type of ListItemNode wrapper

levensta:
In actually, with the `createListItemNode` method, `ListNode` will be able to create `ListItem` of the desired type, since this method can be overridden. Since the types indicate that a `ListNode` might be null during normalizing list items, I included a fallback that creates a copy

etrepum:
There's already ListItemNode.createParentElementNode() which can be used to create a parent ListNode and be overridden. I haven't audited it to see if it's used everywhere it should be, but that would solve the missing ListNode problem in an extensible way without copy heuristics.

etrepum (approved):

PR #8428, opened 2026-04-30
[lexical] Bug Fix: Respect CSS display style in isBlockDomNode and isInlineDomNode

### LINKED ISSUE #8418: Bug: Pasting from Ghostty terminal inserts unintended linebreaks
When pasting from Ghostty into Lexical you get a code block, because of the `<div style="font-family: monospace; white-space: pre;">` wrapper. However, the contents is full of unintended linebreaks because it treats each `<div style="display: inline">` as having an implicit linebreak.

Ghostty arguably should be using a natively inline element for this, but it should be straightforward to address in the importDOM infrastructure (and whatever replaces it).

Lexical version: v0.44

## Steps To Reproduce

1. Select and copy some text from a Ghostty terminal
2. Paste into the playground

Some example HTML from Ghostty's clipboard code:

```html
<div style="font-family: monospace; white-space: pre;"><div style="display: inline;color: rgb(181, 189, 104);font-weight: bold;">&#10095;</div> cd ~/src/lexical

<div style="display: inline;color: rgb(138, 190, 183);font-weight: bold;">lexical</div> on <div style="display: inline;color: rgb(178, 148, 187);font-weight: bold;">&#62488; main</div> <div style="display: inline;color: rgb(204, 102, 102);font-weight: bold;">[$&#8675;]</div> via <div style="display: inline;color: rgb(181, 189, 104);font-weight: bold;">&#59160; v22.17.0 </div>
<div style="display: inline;color: rgb(181, 189, 104);font-weight: bold;">&#10095;</div> </div>
```

## The current behavior

[playground url](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE81WsU7DMBD9lejEaELSVqTyVyCVrerg2iYxOHbkOFGrqqysjB1g4gP4p34A_AJyKqCUJUoaJVPiu_O9d_Y72xvgTFhtZpZYDngDRmvrvjQRkhmuAM9PBoxbIiTgAMGdNimx1W-qGQcMylkkIMjtWjoDILB8ZQHD_vXdDdaZM1PN-GUi4kSKOLGAoOQmF1oBDrdo8x0lheJLw8nDv4CmJDzKvNosmoI81kJA8GO_PQTrjBtitYE24Fd9gueGdr66vRYo-UrQyncmIbcPGEwpjXtSq_o92RmLz-edlxKh-mcygMWYN-yxrFDUFsQ6fxv8i_3TW-cHyeJMRXYnhFKQAYjh42XnlaNR5_vh9yy6MPKDhhRUkS55u5vljBvd350yjDdePQoLBEwYTivdYFVI-YvjsgnFuDpgHmU7EYIkKi5I7Lz3pCQ5NSKz0CB39e7-w2_rsud2RkrOAIdRFE3C6-l0HI7GCHJdGOrm3Uiyjo0uFDueDYE_mfgBbL8A2sNMgOALAAA)

Image

## The expected behavior

[playground url](https://playground.lexical.dev/#doc=H4sIAAAAAAAAE8VWsU7DMBD9FevEaEJSgoL8FUhlqzq4tkkMjh05TtSqKisrYwc68QH9p34A_AJyiqB0ComiTLbfne_du_NJXoPg0hk7ddQJIGuwxji_skwqboUGMjs7cOGoVEBCDA_G5tQ129xwAQS0RxRgKN1KeQAwOLF0QOCw2yPGkQdWhTcxw8VlJtNMyTRzgKEWtpRGA4k2uDvPcysGDD_4_dHZFMJSZyz0Ib8ak7y0bPDqjipQiaVkja2dyG8vJbVYWEGf_u_QM1NkNPp83aKcSj38y5917E1RaeYq6ry9D__F4eV9cJHzkUWiWlL08bZF9WQyuNhgZLFREoQdU9BVvhD9xr39xAw2x4fdfvwkWtZhjoFLK1jTdaIrpX55fDSpudBHzpNoZ21UVKcVTb31kda0ZFYWDjrEbj4Sf_Lb-Oilm9JacCBRkiRxlIThTXx9i6E0lWX-3p2iq9SaSvPT2xAGcRyEsPkCAULdZrEIAAA)

Image

## Impact of fix

importDOM will work more predictably in cases where `<div>` is used as an inline element

abhishekvishwakarma007:
## Description

Fixes #8418

When pasting from Ghostty terminal into Lexical, `<div style="display: inline">` elements were treated as block elements, inserting unwanted linebreaks. This is because `isBlockDomNode` and `isInlineDomNode` classified elements based solely on tag name, ignoring the CSS `display` property.

The fix checks `element.style.display` before falling back to the tag name. If an explicit display style is set, it takes precedence — a `<div style="display: inline">` is now treated like a `<span>` during import, which matches actual browser rendering behavior.

This also handles `display: inline-block`, `display: inline-flex`, etc., following the existing pattern in `LexicalTextNode.ts`.

## Test plan

**Before:**
- Paste Ghostty HTML with `<div style="display: inline;">` wrappers
- Each inline div creates a new line (unwanted linebreaks)

https://github.com/user-attachments/assets/f4c55101-d582-4b90-b744-8ea7c1155a3f

**After:**
- Same paste produces text on the same line, no extra linebreaks
- Normal `<div>` elements (without inline display) still create new lines as expected

- [x] Added 2 unit tests (generic inline div + Ghostty-specific paste)
- [x] All 2,781 unit tests pass
- [x] `pnpm run ci-check` passes (TypeScript, Flow, Prettier, ESLint)
- [x] Manual testing in playground confirms fix

https://github.com/user-attachments/assets/e27b0b1e-e066-46c5-a63e-9406020b9495


potatowagon:
cmiiw, im gonna assume this fix is for all `<div style="display: inline">` elements which ghostty just happen to have. can u add before and after recordings to test plan to show this works?

can be copying ghostty content and pasting it into the lexical playground. or, go to codepen to render `<div style="display: inline">` html and copy paste that

abhishekvishwakarma007:
Updated the test plan with before/after screen recordings showing the fix in action.

packages/lexical/src/LexicalUtils.ts
```diff
@@ -1894,6 +1894,12 @@ export function isInlineDomNode(
     /^(a|abbr|acronym|b|cite|code|del|em|i|ins|kbd|label|mark|output|q|ruby|s|samp|span|strong|sub|sup|time|u|tt|var|#text)$/,
     'i',
   );
```
etrepum: I know this is existing code but this is inefficient for no reason (it constructs two RegExps every call!), we should just use the regexp syntax and float this constant RegExp out of the function, e.g. `const INLINE_NODE_RE = /^…$/i;`

There's a similar problem with the implementation of `isBlockDomNode`

etrepum (approved):

